Issue #275, opened 2025-09-18
Add BestTransactions for lane ordering


mattsse:
> Let `txs[0..n)` be the block body:

> - There exists an index `k` (0 ≤ k ≤ n) such that  
  > - for all `i < k`: `!isPaymentTx(txs[i])` (non-payment), and  
  > - for all `i ≥ k`: `isPaymentTx(txs[i])` (payment).

> Equivalently: once a payment tx appears, **no** non-payment tx may follow.

after #274 we can identify payment transactions

we still need to order them accordingly,
for this we can start with a `LanedTransactions<I: BestTransantions>` helper type that, buffers payment txs until non-payment gas exhausted.

similar to

https://github.com/paradigmxyz/reth/blob/df9b7a079b460291dfa1048d2e785ab718f2e33f/crates/transaction-pool/src/pool/best.rs#L307-L310

and wraps

https://github.com/tempoxyz/tempo/blob/b1839ccd332ddd8846756008c49d3ce9268e3722/crates/payload/builder/src/lib.rs#L70-L70

## TODO

* introduce helper type that yields/buffers txs accordingly
* builder must notify this type about how much gas is still available
* must adhere to 

> once a payment tx appears, **no** non-payment tx may follow.

PR #299, opened 2025-09-22
feat(transaction-pool): add payment transaction classification with TIP20 prefix check

fgimenez:
Closes #274 

Implements payment transaction classification in the transaction pool to support the PaymentLane specification. Transactions are classified once when entering the pool, with the result stored as a boolean field for instant lookups during block building.

- add `PAYMENT_CLASSIFIER_ID` and `TIP20_PAYMENT_PREFIX` constants to chainspec
- add `is_payment` boolean field to `TempoPooledTransaction` for caching classification
- implement classifier v1: transactions with `to` addresses having the TIP20 prefix (`0x20c000000...`) are marked as payments
- add `is_payment()` method for classification lookup during block building
- tests

PR #306, opened 2025-09-22
feat(payload): add lane ordering for payment transactions with gas tracking

fgimenez:
Closes: #275 
Requires #299 

Implements lane ordering for transaction execution, ensuring all non-payment transactions are included before payment transactions in blocks.

- introduces `LanedTransactions` helper type that wraps `BestTransactions` iterator
- buffers payment transactions while yielding non-payment transactions first up to a configurable gas limit (currently 50% of block gas limit).
- switches to payment lane when non-payment gas is exhausted or block limit approached

### DIFF
--- a/crates/payload/builder/src/lib.rs
+++ b/crates/payload/builder/src/lib.rs
@@ -195,10 +195,6 @@ where
         let block_gas_limit: u64 = builder.evm_mut().block().gas_limit;
         let base_fee = builder.evm_mut().block().basefee;
 
-        // Calculate non-payment gas limit (50% of block gas limit by default)
-        // TODO: Make this configurable via chain spec or attributes
-        let non_payment_gas_limit = block_gas_limit / 2;
-
         let best_txs_inner = best_txs(BestTransactionsAttributes::new(
             base_fee,
             builder


crates/payload/builder/src/laned.rs
```diff
@@ -0,0 +1,211 @@
...
+where
+    I: BestTransactions<Item = Arc<ValidPoolTransaction<TempoPooledTransaction>>>,
+{
+    fn mark_invalid(
+        &mut self,
+        tx: &Arc<ValidPoolTransaction<TempoPooledTransaction>>,
+        error: reth_transaction_pool::error::InvalidPoolTransactionError,
+    ) {
+        self.inner.mark_invalid(tx, error);
+    }
```
klkvr, line 85: ideally we should cache the sender here or invalidate all descendants to make sure we don't emit invalid transactions we've buffered before

crates/payload/builder/src/laned.rs
```diff
@@ -0,0 +1,211 @@
...
+
+            if is_payment {
+                // Buffer payment transactions for later
+                self.payment_buffer.push_back(tx);
+            } else {
+                // Check if this non-payment transaction fits within the gas limit
+                let tx_gas = tx.gas_limit();
+                if tx_gas <= self.non_payment_gas_available {
+                    // We have room for this non-payment transaction
+                    // Note: The actual gas update happens via update_non_payment_gas_used()
```
klkvr: this seems off, I don't think inner iterator yields transactions in ascending gas limit order. i.e if this transaction does not fit, next one might still be fine

crates/payload/builder/src/lib.rs
```diff
@@ -207,6 +215,21 @@ where
...
+                    // Mark this transaction as invalid to skip it
+                    best_txs.mark_invalid(
+                        &pool_tx,
+                        InvalidPoolTransactionError::ExceedsGasLimit(
+                            pool_tx.gas_limit(),
+                            block_gas_limit - cumulative_gas_used,
+                        ),
+                    );
+                    continue;
+                }
```
klkvr: similarly here, I believe such logic might cause a single transaction with high tip and gas limit to block non-payment lane entirely

crates/payload/builder/src/laned.rs
fgimenez: good point, done

crates/payload/builder/src/laned.rs
fgimenez: great catch, we now continue searching non-payment tx that might fit after finding the first that exceeds the gas limit, ptal

crates/payload/builder/src/lib.rs
fgimenez: yep, we now don't force lane switch after the first non-fitting tx and let the itertor to continue, ptal

crates/payload/builder/src/laned.rs
```diff
@@ -0,0 +1,245 @@
...
+                    // We have room for this non-payment transaction
+                    // Reset the skipping flag since we found a usable non-payment tx
+                    self.skipping_non_payment = false;
+                    // Note: The actual gas update happens via update_non_payment_gas_used()
+                    // after successful execution
+                    return Some(tx);
+                } else {
+                    // This non-payment transaction doesn't fit
+                    // Mark that we've started skipping non-payment transactions
+                    self.skipping_non_payment = true;
```
klkvr: this seems odd, I guess in cases when we have transactions getting yielded like

1. non-payment with gas limit 21000
2. non-payment with gas limit `nonPaymentGasLimit`
3. payment tx
4. non-payment with gas limit 21000

this will not include tx 4 while it would make sense to?

fgimenez: indeed, makes total sense. i have extracted the logic for skipping non-payment txs while keeping ttrack of gas used to a separate state machine, and added proper tests to it, including the example you described, ptal

crates/payload/builder/src/laned.rs
```diff
@@ -0,0 +1,803 @@
...
+    /// Processing non-payment transactions normally
+    NonPayment {
+        /// Gas remaining for non-payment transactions
+        gas_available: u64,
+    },
+    /// Skipping large non-payment txs, looking for smaller ones
+    SkippingNonPayment {
+        /// Gas remaining for non-payment transactions
+        gas_available: u64,
+    },
```
klkvr: why do we need those as separate states? looks like they're handled identically and could just be a bool and a u64

crates/payload/builder/src/laned.rs
```diff
@@ -0,0 +1,803 @@
...
+        if self.lane_state.is_draining_buffer() {
+            // Drain buffered payment transactions, skipping those from invalidated senders
+            while let Some(tx) = self.payment_buffer.pop_front() {
+                if !self.invalidated_senders.contains(&tx.sender()) {
+                    return Some(tx);
+                }
+            }
+            // Buffer is empty
+            return None;
+        }
```
klkvr: we should also try draining inner iterator one last time here in case there are new transactions loaded from the pool

crates/payload/builder/src/laned.rs
fgimenez: yep, this was a leftover from a previous approach, fixed

crates/payload/builder/src/laned.rs
fgimenez: good point, done ptal

crates/payload/builder/src/laned.rs
klkvr: sorry I meant that entire enum can be replaced with `payment_lane: bool, non_payment_gas: u64` as it was before

I don't think `skipping` bool is ever read

fgimenez: oh i see, for sure, simpler now ptal

crates/payload/builder/src/laned.rs
```diff
@@ -0,0 +1,812 @@
...
+    I: BestTransactions<Item = Arc<ValidPoolTransaction<TempoPooledTransaction>>>,
+{
+    type Item = Arc<ValidPoolTransaction<TempoPooledTransaction>>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        // If we're draining the buffer (iterator exhausted), check for new transactions first
+        if self.lane_state.is_draining_buffer() {
+            // Try the inner iterator for new transactions that may have been loaded
+            loop {
+                match self.inner.next() {
```
klkvr: we should drain the buffer first here because it mght contain txs with higher priority

crates/payload/builder/src/laned.rs
```diff
@@ -0,0 +1,812 @@
...
+                Some(tx) => tx,
+                None => {
+                    // Iterator exhausted, start draining the buffer
+                    self.lane_state.mark_exhausted();
+                    // Return first buffered transaction (if any)
+                    while let Some(buffered_tx) = self.payment_buffer.pop_front() {
+                        if !self.invalidated_senders.contains(&buffered_tx.sender()) {
+                            return Some(buffered_tx);
+                        }
+                    }
```
klkvr: can we wrap entire thing into the loop so that it's more like
```rust
loop {
    if payments_lane {
        ...drain buffer....
        ...drain pool...
        return None
    }
     
     ...drain pool, fill buffer...
     payments_lane = true
}
```

that way it should be a bit easier to reason about

crates/payload/builder/src/laned.rs
```diff
@@ -0,0 +1,812 @@
...
+    I: BestTransactions<Item = Arc<ValidPoolTransaction<TempoPooledTransaction>>>,
+{
+    /// The inner iterator that provides transactions from the pool
+    inner: I,
+    /// Buffer for payment transactions that arrived while processing non-payment transactions
+    payment_buffer: VecDeque<Arc<ValidPoolTransaction<TempoPooledTransaction>>>,
+    /// Lane state machine managing transitions
+    lane_state: LaneState,
+    /// Total gas used by non-payment transactions
+    non_payment_gas_used: u64,
```
mattsse: I think we can simplify this a bit and perhaps reduce this this to one u64 and a bool instead which determines `non_payment_gas_left` and `non_payment_gas_exhausted`

or perhaps just `non_payment_gas_exhausted` and leave it up to the consumer entirely to track payment gas

the transition can then be triggered two-fold
a) consumer marks `fn payments_only == non_payment_gas_exhausted` 
b) iterator is exhausted and only has payment txs left

crates/payload/builder/src/laned.rs
fgimenez: sounds good, done here https://github.com/tempoxyz/tempo/pull/306/commits/a4b69c5ba1876dc609b4e52bca84b9efcfabe939

crates/payload/builder/src/laned.rs
fgimenez: done as part of https://github.com/tempoxyz/tempo/pull/306/commits/a4b69c5ba1876dc609b4e52bca84b9efcfabe939

crates/payload/builder/src/laned.rs
fgimenez: sure done here https://github.com/tempoxyz/tempo/pull/306/commits/3cce53255edb7f11f39cbf4caed50c7ff63a9283 ptal

crates/payload/builder/src/laned.rs
```diff
@@ -0,0 +1,236 @@
...
+                    // Skip non-payment transactions since we're out of non-payment gas
+                }
+
+                // No more transactions available
+                return None;
+            }
+
+            // Still have non-payment gas - process transactions and fill buffer
+            match self.inner.next() {
+                Some(tx) => {
```
klkvr: nit: can remove one level of indentation here by doing `while let Some(tx) = self.inner.next()` and moving `self.non_payment_gas_exhausted = true;` to the end of the loop

crates/payload/builder/src/laned.rs
```diff
@@ -0,0 +1,236 @@
...
+    #[allow(dead_code)]
+    pub(crate) fn non_payment_gas_used(&self) -> u64 {
+        self.non_payment_gas_used
+    }
+
+    /// Updates the available gas after a transaction has been executed.
+    ///
+    /// This should be called by the builder after successfully executing a non-payment transaction
+    /// to update the gas accounting.
+    pub(crate) fn update_non_payment_gas_used(&mut self, gas_used: u64) {
```
mattsse, line 58: I believe the builder must keep track of non payment gas anyway and can then decide to hit the switch, so what we also need need is `skip_non_payments() == non_payment_exhausted()`

similar to skip_blobs

crates/payload/builder/src/laned.rs
```diff
@@ -0,0 +1,236 @@
...
+                        && !self.invalidated_senders.contains(&tx.sender())
+                    {
+                        return Some(tx);
+                    }
+                    // Skip non-payment transactions since we're out of non-payment gas
+                }
+
+                // No more transactions available
+                return None;
+            }
```
mattsse, line 131: hmm, do we even need this, because next we consume the iter anyway

crates/payload/builder/src/laned.rs
```diff
@@ -0,0 +1,236 @@
...
+                        } else {
+                            // Non-payment transaction doesn't fit, skip it
+                            self.inner.mark_invalid(
+                                &tx,
+                                reth_transaction_pool::error::InvalidPoolTransactionError::ExceedsGasLimit(
+                                    tx_gas,
+                                    self.non_payment_gas_left,
+                                ),
+                            );
+                        }
```
mattsse: this makes sense

crates/payload/builder/src/laned.rs
fgimenez: good point, done ptal

crates/payload/builder/src/laned.rs
fgimenez: i believe we need this inner loop followed by `return None` otherwise how do we know we are done and exit the outer loop, am i missing something?

crates/payload/builder/src/laned.rs
fgimenez: ok, done in https://github.com/tempoxyz/tempo/pull/306/commits/16c07fefce53ee12190a714be7990dfb8b66d027 now the builder keeps track of non payment gas spent and can decide to switch, but also the helper can do that once the inner iterator is exhausted and proceeds to drain the buffer, lmk if that's ok or we should structure this differently

crates/payload/builder/src/laned.rs
mattsse: ah I see, by_ref actually just calls next under the hood, but this is also just `while let Some(tx) = self.inner.next()?

so we either move the payment check out of the loop or have the loop twice, I think the we have it now works

crates/payload/builder/src/laned.rs
```diff
@@ -0,0 +1,659 @@
...
+                                tx_gas,
+                                self.non_payment_gas_left,
+                            ),
+                        );
+                    }
+                }
+            }
+
+            // Iterator exhausted, switch to payment-only mode
+            self.skip_non_payments = true;
```
mattsse, line 161: I see, this toggles

crates/payload/builder/src/lib.rs
```diff
@@ -192,15 +195,21 @@ where
         let block_gas_limit: u64 = builder.evm_mut().block().gas_limit;
         let base_fee = builder.evm_mut().block().basefee;
 
-        let mut best_txs = best_txs(BestTransactionsAttributes::new(
+        // Calculate non-payment gas limit (50% of block gas limit by default)
+        // TODO: Make this configurable via chain spec or attributes
+        let non_payment_gas_limit = block_gas_limit / 2;
```
klkvr: we already initialize this variable on L170, can reuse

crates/payload/builder/src/lib.rs
```diff
@@ -251,7 +259,19 @@ where
...
+                Ok(gas_used) => {
+                    // Update non-payment gas tracking if we're still in non-payment lane
+                    if !best_txs.non_payment_exhausted() && !pool_tx.transaction.is_payment() {
+                        non_payment_gas_used += gas_used;
+                        best_txs.update_non_payment_gas_used(gas_used);
+
+                        // Check if we've exhausted non-payment gas and trigger the switch
+                        if non_payment_gas_used >= non_payment_gas_limit {
+                            best_txs.skip_non_payments();
+                        }
```
klkvr, line 267: this should work for now but I guess eventually we would need some kind of buffer here — e.g if we have <21k gas left it likely doesn't make sense to fill any non-payments

crates/payload/builder/src/lib.rs
fgimenez: good catch, removed

PR #312, opened 2025-09-22
feat(arber): use arbitrary wallet options

Zygimantass:


bin/tempo-arber/src/main.rs
```diff
@@ -3,8 +3,12 @@
 
 use std::time::Duration;
 
-use alloy::{providers::ProviderBuilder, signers::local::MnemonicBuilder};
+use alloy::{
+    network::{EthereumWallet, TxSigner},
+    providers::ProviderBuilder,
+};
 use clap::Parser;
+use foundry_wallets::WalletOpts;
```
gakonst, line 11: realizing we should upstream this into some `alloy-cli` helpers

PR #305, opened 2025-09-22
feat(commonware-node): update forkchoice after propose, verify

SuperFluffy:
Moves updating forkchoice state from the finalizer task to a new forkchoice-updater task.

The finalizer task no longer sends forkchoice-updates on every finalization, but merely tracks the last finalized block digest, making it available through a watch channel (NOTE: breaking convention of only using futures-rs items because tokio watch channels are more ergonomic).

Instead of the finalizer, a forkchoice-updater task is now responsible for updating the canonical chain. It relies on the fact that verifications and proposals are guaranteed to be on notarized parents, and on the finalizer making its latest finalized block available.

1. at the beginning of each proposal, setting `head = proposal.parent` and `finalized = last_finalized_hash`.
2. after a succesful proposal, setting `head = proposal` and `finalized = last_finalized_hash`.
3. at the beginning of each verification, setting `head = proposal.parent` and `finalized = last_finalized_hash`.
4. after a succesful verification, setting `head = proposal` and `finalized = last_finalized_hash`.

We choose to set forkchoice state before and after a proposal to defend against block building taking very long: if we wait until after a block was successfully built to decide whether to set the head to parent or proposal, we can potentially stall a node from progressing for several seconds.

For verification we similarly know that we always want to set the forkchoice state to at least the notarized parent - and, if successful, to the verified proposal derived from it. With forkchoice running in the background there is no need to stall chain evolution on verification running through.

PR #314, opened 2025-09-23
refactor(commonware-node): enforce execution driver invariants

SuperFluffy:
Enforces execution driver initialization at the type level. The execution driver's event loop can now only be entered after its method `ExecutionDriver<Uninit>::into_initialized -> ExecutionDriver<Init>` has been called.

Also restricts a bunch of types and methods to be more private.

### DIFF
--- a/crates/commonware-node/src/consensus/execution_driver/mod.rs
+++ b/crates/commonware-node/src/consensus/execution_driver/mod.rs
@@ -32,7 +32,7 @@ use crate::consensus::execution_driver::executor::ExecutorMailbox;
 
 use super::{View, block::Block};
 
-pub(super) struct Builder<TContext> {
+pub(super) struct ExecutionDriverBuilder<TContext> {
     /// The execution context of the commonwarexyz application (tokio runtime, etc).
     pub(super) context: TContext,
 
@@ -51,14 +51,14 @@ pub(super) struct Builder<TContext> {
     pub(super) execution_node: TempoFullNode,
 }
 
-impl<TContext> Builder<TContext>
+impl<TContext> ExecutionDriverBuilder<TContext>
 where
     TContext: Clock + governor::clock::Clock + Rng + CryptoRng + Spawner + Storage + Metrics,
 {
     /// Builds the uninitialized execution driver.
     pub(super) fn build(self) -> eyre::Result<ExecutionDriver<TContext, Uninit>> {
         let (tx, rx) = mpsc::channel(self.mailbox_size);
-        let my_mailbox = Mailbox::from_sender(tx);
+        let my_mailbox = ExecutionDriverMailbox::from_sender(tx);
 
         let block = self
             .execution_node
@@ -73,7 +73,7 @@ where
 
             fee_recipient: self.fee_recipient,
 
-            from_consensus: rx,
+            mailbox: rx,
             my_mailbox,
             syncer_mailbox: self.syncer_mailbox,
 
@@ -91,8 +91,8 @@ pub(super) struct ExecutionDriver<TContext, TState = Uninit> {
 
     fee_recipient: alloy_primitives::Address,
 
-    from_consensus: mpsc::Receiver<Message>,
-    my_mailbox: Mailbox,
+    mailbox: mpsc::Receiver<Message>,
+    my_mailbox: ExecutionDriverMailbox,
 
     syncer_mailbox: marshal::Mailbox<BlsScheme, Block>,
 
@@ -108,7 +108,7 @@ pub(super) struct ExecutionDriver<TContext, TState = Uninit> {
 }
 
 impl<TContext, TState> ExecutionDriver<TContext, TState> {
-    pub(super) fn mailbox(&self) -> &Mailbox {
+    pub(super) fn mailbox(&self) -> &ExecutionDriverMailbox {
         &self.my_mailbox
     }
 }
@@ -175,7 +175,7 @@ where
         let initialized = ExecutionDriver {
             context: self.context,
             fee_recipient: self.fee_recipient,
-            from_consensus: self.from_consensus,
+            mailbox: self.mailbox,
             my_mailbox: self.my_mailbox,
             syncer_mailbox: self.syncer_mailbox,
             genesis_block: self.genesis_block,
@@ -227,7 +227,7 @@ where
                 // biased and this note here make sense.
                 biased;
 
-                Some(msg) = self.from_consensus.next() => {
+                Some(msg) = self.mailbox.next() => {
                     if let Err(error) =  self.handle_message(msg) {
                         tracing::error_span!("handle message").in_scope(|| tracing::error!(
                             %error,
@@ -694,7 +694,7 @@ async fn verify_block(
     }
 }
 
-impl Automaton for Mailbox {
+impl Automaton for ExecutionDriverMailbox {
     type Context = super::Context;
 
     type Digest = Digest;
@@ -761,7 +761,7 @@ impl Automaton for Mailbox {
     }
 }
 
-impl Relay for Mailbox {
+impl Relay for ExecutionDriverMailbox {
     type Digest = Digest;
 
     async fn broadcast(&mut self, digest: Self::Digest) {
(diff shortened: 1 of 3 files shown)


crates/commonware-node/src/consensus/execution_driver/mod.rs
```diff
@@ -77,17 +78,26 @@ where
...
+#[derive(Clone, Debug)]
+pub(super) struct Uninit(());
+
+/// Carries the runtime initialized state of the execution driver.
+#[derive(Clone, Debug)]
+struct Init {
+    latest_proposed_block: Arc<RwLock<Option<Block>>>,
+    to_finalizer: finalizer::Mailbox,
+    to_forkchoice: forkchoice::Mailbox,
+}
```
mattsse: these type defs and impl blocks are out of order

SuperFluffy: Sorted `Unit` and `Init` after the `ExecutionDriver` impl blocks now.

crates/commonware-node/src/consensus/execution_driver/mod.rs
```diff
@@ -30,30 +30,31 @@ mod canonical_chain;
 
 use super::{View, block::Block};
 
-pub struct Builder<TContext> {
+pub(super) struct Builder<TContext> {
```
0xKitsune: Can we update this to `ExecutionDriverBuilder` or similar?

crates/commonware-node/src/consensus/execution_driver/mod.rs
```diff
@@ -30,30 +30,31 @@ mod canonical_chain;
...
+    pub(super) syncer_mailbox: marshal::Mailbox<BlsScheme, Block>,
 
     /// A handle to the execution node to verify and create new payloads.
-    pub execution_node: TempoFullNode,
+    pub(super) execution_node: TempoFullNode,
 }
 
 impl<TContext> Builder<TContext>
 where
     TContext: Clock + governor::clock::Clock + Rng + CryptoRng + Spawner + Storage + Metrics,
```
0xKitsune, line 56: This is a bit verbose, consider making an `ExecutionCtx` supertrait.

crates/commonware-node/src/consensus/execution_driver/mod.rs
```diff
@@ -96,61 +95,52 @@ pub struct ExecutionDriver<TContext> {
...
+    /// runtime initialized.
+    state: TState,
 }
 
-impl<TContext> ExecutionDriver<TContext>
-where
-    TContext: Clock + governor::clock::Clock + Rng + CryptoRng + Spawner + Storage + Metrics,
-{
+impl<TContext, TState> ExecutionDriver<TContext, TState> {
     pub(super) fn mailbox(&self) -> &Mailbox {
```
0xKitsune: Similar to #305, can we update to make this `ExecutionDriverMailbox` or similar?

crates/commonware-node/src/consensus/execution_driver/mod.rs
```diff
@@ -96,61 +95,52 @@ pub struct ExecutionDriver<TContext> {
...
     ///
     /// This includes:
     ///
     /// 1. reading the last finalized digest from the consensus marshaller.
     /// 2. starting the canonical chain engine and storing its handle.
+    /// 3. starting a backfill task to catch the execution layer up to the
+    ///    consensus finalize block (if necessary).
     #[instrument(skip_all, err)]
-    async fn runtime_init(&mut self) -> eyre::Result<()> {
+    async fn into_initialized(mut self) -> eyre::Result<ExecutionDriver<TContext, Init>> {
```
0xKitsune, line 145: Rather than having an init function here, this seems like a good usecase for a builder that returns an initialized execution driver.

crates/commonware-node/src/consensus/execution_driver/mod.rs
```diff
@@ -429,15 +454,25 @@ where
     }
 }
 
+/// Marker type to signal that the execution driver is not fully initialized.
+#[derive(Clone, Debug)]
+pub(super) struct Uninit(());
```
0xKitsune, line 461: Similar to my prev comment, if we use a builder pattern that always returns an initialized `ExecutionDriver`, can we get rid of this?

crates/commonware-node/src/consensus/execution_driver/mod.rs
```diff
@@ -576,11 +610,11 @@ impl RunPropose {
...
-            let mut lock = latest_proposed_block.write().await;
+            let mut lock = state.latest_proposed_block.write().await;
             *lock = Some(consensus_block);
         }
 
-        if let Err(error) = canonical_chain.update(proposed_block_digest) {
+        if let Err(error) = state.canonical_chain.update(proposed_block_digest) {
             tracing::warn!(
                 %error,
                 "failed updating canonical chain to proposed block",
```
0xKitsune: Consider using `inspect_err` instead. Something like 

```rust
state.canonical_chain
    .update(proposed_block_digest)
    .inspect_err(|error| {
        tracing::warn!(%error, "failed updating canonical chain to proposed block");
    })
    .ok();
```

crates/commonware-node/src/consensus/execution_driver/mod.rs
SuperFluffy: For all intents and purposes this is a builder, just that we use a marker instead of a type. How do you think would this look?

crates/commonware-node/src/consensus/execution_driver/mod.rs
SuperFluffy: I am not sure how - I can't return a fully initialized execution driver. I will only have some of this information available once I start the agent and pass its mailbox to the consensus engine.

crates/commonware-node/src/consensus/execution_driver/mod.rs
SuperFluffy: IMO `if let Err() = ` is easier to read than chaining. `if let Err` signals that processing here is at an end and that the generated event is the desired side effect.

Chaining suggests some more manipulation. In your example, you end up in an `ok()` and I have to remember why (fixing a `#[must_use]` lint in rustc).

crates/commonware-node/src/consensus/execution_driver/mod.rs
SuperFluffy: Not all of these traits are even necessary. This is a remnant of just copying stuff over from commonware's alto. I'd just remove most of these traits tbh. But not here (PR is unrelated).

crates/commonware-node/src/consensus/execution_driver/mod.rs
SuperFluffy: You could of course hide the fact that it's not initailized. But it's all crate-private, so 🤷

crates/commonware-node/src/consensus/execution_driver/mod.rs
SuperFluffy: I am going to do this change, but it IMO does not add any clarity: it's only used in one place (in the `consensus::Engine` builder), and is now qualified as `execution_driver::ExecutionDriverBuilder` because importing it into the head of the file to only use once seems unnecessary.

crates/commonware-node/src/consensus/execution_driver/mod.rs
SuperFluffy: Doing the change - but this type is actually only ever named in one place, so I am not sure if it really adds clarity.

SuperFluffy: [6b5181e](https://github.com/tempoxyz/tempo/pull/314/commits/6b5181e7d310e96821b330c3ff6f4dbedbb289d5)

crates/commonware-node/src/consensus/execution_driver/mod.rs
SuperFluffy: [6b5181e](https://github.com/tempoxyz/tempo/pull/314/commits/6b5181e7d310e96821b330c3ff6f4dbedbb289d5)

crates/commonware-node/src/consensus/execution_driver/mod.rs
0xKitsune: Yeah I dont have a strong opinion here, since nothing is being done with the error other than logging, seems like a fit for `inspect_err`. Will defer if you have a strong preference.

PR #321, opened 2025-09-23
feat(sidecar): add synthetic load generator

Zygimantass:
blocked on #319 

this PR adds a synthetic load generator that:
* picks N wallets
* sets random fee tokens according to zipf distribution (token 1 will be the most common, token 2 will be less common and so on)
* picks a random wallet according to zipf distribution (again, to simulate hotspots)
* sends a transaction according to a poisson distribution (some blocks will have more transactions than others)

### DIFF
--- a/bin/tempo-sidecar/src/synthetic_load/mod.rs
+++ b/bin/tempo-sidecar/src/synthetic_load/mod.rs
@@ -1,12 +1,15 @@
-use tempo_telemetry_util::error_field;
 use alloy::{
     network::EthereumWallet,
-    primitives::{Address, U256, private::rand},
+    primitives::{
+        Address, U256,
+        private::{
+            rand,
+            rand::{RngCore, SeedableRng, rngs::StdRng},
+        },
+    },
     providers::ProviderBuilder,
     signers::local::MnemonicBuilder,
 };
-use alloy::primitives::private::rand::rngs::StdRng;
-use alloy::primitives::private::rand::{RngCore, SeedableRng};
 use eyre::Context;
 use rand_distr::{Distribution, Exp, Zipf};
 use reqwest::Url;
@@ -14,6 +17,7 @@ use tempo_precompiles::{
     TIP_FEE_MANAGER_ADDRESS,
     contracts::{IFeeManager, ITIP20},
 };
+use tempo_telemetry_util::error_field;
 use tracing::{debug, info, warn};
 
 pub struct SyntheticLoadGenerator {
@@ -72,7 +76,8 @@ impl SyntheticLoadGenerator {
         info!("setting fee tokens for load generating wallets");
 
         for address in &addresses {
-            let fee_token_address = zipf_vec_sample(&mut rng, fee_token_zipf, &self.fee_token_addresses)?;
+            let fee_token_address =
+                zipf_vec_sample(&mut rng, fee_token_zipf, &self.fee_token_addresses)?;
             let fee_manager = IFeeManager::new(TIP_FEE_MANAGER_ADDRESS, provider.clone());
             _ = fee_manager
                 .setUserToken(*fee_token_address)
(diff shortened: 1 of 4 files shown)


bin/tempo-sidecar/src/synthetic_load/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+
+        for address in &addresses {
+            let fee_token_address = zipf_vec_sample(fee_token_zipf, &self.fee_token_addresses)?;
+            let fee_manager = IFeeManager::new(TIP_FEE_MANAGER_ADDRESS, provider.clone());
+            _ = fee_manager
+                .setUserToken(*fee_token_address)
+                .from(*address)
+                .send()
+                .await
+                .map_err(|e| {
```
SuperFluffy: You ideally want to avoid explicitly formatting the error:

```rust
.send()
.await
.wrap_err_with(|| format!("failed to set fee token `{address}` for address `{fee_token_address}`"))
```

If you explicitily format the error than you likely use its source chain (there are cases where that's unavoidable, for example if if the `E` in `Result<T, E>` does not implement `std::error::Error`).

bin/tempo-sidecar/src/synthetic_load/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+
+        let exp = Exp::new(self.average_tps as f64)?;
+        let zipf = Zipf::new(self.wallet_count as f64, 1.4)?;
+
+        loop {
+            let sender = zipf_vec_sample(zipf, &addresses)?;
+            let recipient = zipf_vec_sample(zipf, &addresses)?;
+            let token = zipf_vec_sample(fee_token_zipf, &self.fee_token_addresses)?;
+
+            trace!(
```
SuperFluffy: IMO `debug!` here is more appropriate I think. trace is very low level

bin/tempo-sidecar/src/synthetic_load/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+                sender = sender.to_string(),
+                recipient = recipient.to_string(),
```
SuperFluffy: Use `%sender` and `%recipient` instead of `sender = sender.to_string()`. Those use `std::fmt::Display`, just like `ToString::to_string` does.

bin/tempo-sidecar/src/synthetic_load/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+
+            let token = ITIP20::new(*token, provider.clone());
+            if let Err(e) = token
+                .transfer(*recipient, U256::from(10))
+                .from(*sender)
+                .send()
+                .await
+            {
+                error!(
+                    sender = sender.to_string(),
```
SuperFluffy: `%` sigil, not `to_string`

bin/tempo-sidecar/src/synthetic_load/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+                    recipient = recipient.to_string(),
+                    err = e.to_string(),
```
SuperFluffy: It'd be good to not lose the source chain here I think. Depending on whether this is an eyre error or an `E: std::error::Error`, you can do `&error as &dyn std::error::Error` or `AsRef::<dyn std::error::Error>::as_ref(&error)`.

Or you use `tempo_telemetry_util::error_field` :-)

bin/tempo-sidecar/src/synthetic_load/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+            );
+
+            let token = ITIP20::new(*token, provider.clone());
+            if let Err(e) = token
+                .transfer(*recipient, U256::from(10))
+                .from(*sender)
+                .send()
+                .await
+            {
+                error!(
```
SuperFluffy: Probably set this as `warn!`. `error!` is for unrecoverable errors, but you just chugg along after this.

PR #322, opened 2025-09-23
feat: introduce consensus and execution validations for payment lanes

klkvr:
This PR introduces checks relevant for payment lane to consensus and block executor code.

This required to introduce tempo-specific `TempoBlockExecutionCtx`, `TempoNextBlockEnvAttributes` and `TempoBlockAssembler`. All of those wrap inner Ethereum structures and simply extend/override parts of their behavior.

### DIFF
--- a/crates/node/tests/it/eip7702_delegation.rs
+++ b/crates/node/tests/it/eip7702_delegation.rs
@@ -44,9 +44,10 @@ async fn test_auto_7702_delegation() -> eyre::Result<()> {
         .build()?;
     let bob_addr = bob.address();
 
-    let amount = U256::random();
+    let transfer_amount = U256::from(1000);
+    let mint_amount = transfer_amount * U256::from(3);
     token
-        .mint(bob_addr, amount)
+        .mint(bob_addr, mint_amount)
         .send()
         .await?
         .get_receipt()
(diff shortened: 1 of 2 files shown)


crates/evm/src/lib.rs
```diff
@@ -107,30 +117,39 @@ impl ConfigureEvm for TempoEvmConfig {
...
-            parent_beacon_block_root: block.header().parent_beacon_block_root,
-            ommers: &block.body().ommers,
-            withdrawals: block.body().withdrawals.as_ref().map(Cow::Borrowed),
+        self.inner.next_evm_env(parent, &attributes.inner)
+    }
+
+    fn context_for_block<'a>(&self, block: &'a SealedBlock<Block>) -> TempoBlockExecutionCtx<'a> {
+        let non_payment_gas_limit = TempoExtraData::decode(&block.header().extra_data)
+            .map(|data| data.non_payment_gas_limit)
+            .unwrap_or(0);
```
klkvr, line 128: This should be OK because header is always validated before execution. However, we could still benefit from making this faillible

crates/evm/src/assemble.rs
```diff
@@ -0,0 +1,99 @@
...
+            if !tx.is_payment() {
+                non_payment_gas_used = receipt.cumulative_gas_used;
+            } else {
+                break;
+            }
+        }
+
+        // Delegate block building to the inner assembler
+        let mut block = self.inner.assemble_block(BlockAssemblerInput::<
+            EthBlockExecutorFactory<TempoReceiptBuilder, TempoChainSpec, TempoEvmFactory>,
```
klkvr, line 63: This is a bit ugly and can probably be improved on reth side

crates/consensus/src/lib.rs
```diff
@@ -97,6 +119,91 @@ impl FullConsensus<TempoPrimitives> for TempoConsensus {
...
+    pub non_payment_gas_used: u64,
+}
+
+impl TempoExtraData {
+    /// Decodes the extra data from the given bytes.
+    pub fn decode(extra_data: &[u8]) -> Result<Self, ConsensusError> {
+        if extra_data.len() < 21 {
+            return Err(ConsensusError::Other(
+                "Extra data must be 32 bytes".to_string(),
+            ));
```
mattsse: should this error message correspond to <21 check?

crates/consensus/src/lib.rs
```diff
@@ -97,6 +119,91 @@ impl FullConsensus<TempoPrimitives> for TempoConsensus {
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct TempoExtraData {
+    /// Non-payment gas limit.
+    pub non_payment_gas_limit: u64,
+    /// Non-payment gas used.
+    pub non_payment_gas_used: u64,
+}
+
+impl TempoExtraData {
+    /// Decodes the extra data from the given bytes.
```
mattsse, line 167: could we add some docs that describe the format here,

from what I can tell this is `[...<suffix>]`

where suffix is `NPGS<version><nongaslimit><paymentgaslimit>`?

crates/evm/src/assemble.rs
```diff
@@ -0,0 +1,99 @@
...
+        } = input;
+
+        let mut non_payment_gas_used = 0;
+        for (tx, receipt) in transactions.iter().zip(output.receipts.iter()) {
+            if !tx.is_payment() {
+                non_payment_gas_used = receipt.cumulative_gas_used;
+            } else {
+                break;
+            }
+        }
```
mattsse, line 59: this could potentially be optimized with some binary search approach because we only need to find the cutoff index

crates/evm/src/assemble.rs
```diff
@@ -0,0 +1,99 @@
...
+        let prefix = if block.header.extra_data.len()
+            <= MAXIMUM_EXTRA_DATA_SIZE - TEMPO_EXTRA_DATA_SUFFIX_LENGTH
+        {
+            block.header.extra_data
+        } else {
+            block
+                .header
+                .extra_data
+                .slice(..MAXIMUM_EXTRA_DATA_SIZE - TEMPO_EXTRA_DATA_SUFFIX_LENGTH)
+        };
```
mattsse, line 92: I see, this effectively leaves 11bytes for some miner specific extradata?

crates/evm/src/lib.rs
mattsse: yeah makes sense, although we expect this to be wellformed yes

crates/evm/src/assemble.rs
klkvr: yep

crates/consensus/src/lib.rs
klkvr: yep, added a doc

crates/evm/src/assemble.rs
klkvr: yeah but I guess it should be negligible overhead compared to execution, receipt root etc

PR #323, opened 2025-09-23
feat(arber): refactor to initialize with pool data

Zygimantass:


bin/tempo-sidecar/src/cmd/simple_arb.rs
```diff
@@ -0,0 +1,305 @@
...
+    if user_reserve == U256::ZERO && validator_reserve == U256::ZERO {
+        debug!("Pool has zero reserves, skipping rebalance");
+        return Ok(());
+    }
+
+    // Calculate if rebalance is needed - check if user reserves are significantly higher
+    let total_reserve = user_reserve + validator_reserve;
+    if total_reserve == U256::ZERO {
+        debug!("Total reserves are zero, skipping rebalance");
+        return Ok(());
```
0xKitsune: This check seems redundant since we already check that user and validator reserves are > 0.

bin/tempo-sidecar/src/cmd/simple_arb.rs
```diff
@@ -0,0 +1,305 @@
...
+    }
+
+    let user_ratio = user_reserve * U256::from(100) / total_reserve;
+    let threshold = U256::from(60); // If user reserves > 60% of total, rebalance
```
0xKitsune: nit: you can use `uint!(100_U256)`, `uint!(60_U256)` here. This can also be applied everywhere we use `U256::from(val)` where `val` is known ahead of time.

bin/tempo-sidecar/src/cmd/simple_arb.rs
```diff
@@ -0,0 +1,305 @@
...
+    Ok(pools)
+}
+
+#[instrument(skip(signer, pools))]
+async fn rebalance_pool(
+    rpc_url: &str,
+    signer: Arc<EthereumWallet>,
+    token_a: Address,
+    token_b: Address,
+    pools: &DashMap<(Address, Address), Pool>,
```
0xKitsune: I think we can separate this into two functions, one to check if the pool should be rebalanced and another to actually rebalance the pool.

bin/tempo-sidecar/src/cmd/simple_arb.rs
```diff
@@ -0,0 +1,305 @@
...
+    if user_ratio > threshold {
+        info!(
+            %user_ratio,
+            "Rebalancing pool"
+        );
+
+        // Calculate amount to rebalance - take half of the excess
+        let excess = user_reserve.saturating_sub(validator_reserve);
+        let amount_out = excess / U256::from(2);
+
```
0xKitsune: I think we can set `amount_out` to the full user token value here.

bin/tempo-sidecar/src/cmd/simple_arb.rs
```diff
@@ -0,0 +1,305 @@
...
+    rpc_url: &str,
+    signer: Arc<EthereumWallet>,
+    token_a: Address,
+    token_b: Address,
+    pools: &DashMap<(Address, Address), Pool>,
+) -> eyre::Result<()> {
+    let provider = ProviderBuilder::new()
+        .wallet(signer.clone())
+        .connect_http(rpc_url.parse().context("failed to parse RPC URL")?);
+
```
0xKitsune: We can init a provider once and pass it in to all functions rather than initializing this on every function call.

PR #325, opened 2025-09-23
feat: add test for feeToken transaction

klkvr:
Had to slightly update some other parts of code to respect transaction's values

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7718,10 +7718,12 @@ dependencies = [
  "alloy-primitives",
  "alloy-rlp",
  "alloy-rpc-types-engine",
+ "alloy-serde",
  "derive_more",
  "ethereum_ssz",
  "ethereum_ssz_derive",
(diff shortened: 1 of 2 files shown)


crates/precompiles/src/contracts/provider.rs
```diff
@@ -6,41 +6,38 @@ use crate::{
...
-        // Look up user's configured fee token in TIPFeeManager storage
-        let user_token_slot = mapping_slot(user, tip_fee_manager::slots::USER_TOKENS);
-        let mut fee_token = self
-            .storage(TIP_FEE_MANAGER_ADDRESS, user_token_slot.into())?
-            .unwrap_or_default()
-            .into_address();
+    fn get_fee_token_balance(
+        &self,
+        user: Address,
+        tx_fee_token: Option<Address>,
```
mattsse, line 23: this makes sense

PR #326, opened 2025-09-23
feat: tempo rpc header

### LINKED ISSUE #303: Introduce tempo-specific RPC header
```
eth_getBlockBy*: expose parsed nonPaymentGasLimit / nonPaymentGasUsed (decoded from trailer).
```

For this we will need to introduce a custom RPC block header type that will wrap an inner `alloy_rpc_types_eth::Header` and decode the fields from `extraData`

To support RPC conversions we will need to just implement this trait on it:
https://github.com/paradigmxyz/reth/blob/ff51faaeacd519e69ad8bc0c762b961032d8c285/crates/rpc/rpc-convert/src/transaction.rs#L94-L98

klkvr:
Closes #303 

Introduces Tempo-specific RPC block header including `nonPaymentGasUsed` and `nonPaymentGasLimit` fields.

### DIFF
--- a/crates/node/src/rpc/header.rs
+++ b/crates/node/src/rpc/header.rs
@@ -15,8 +15,10 @@ pub struct Header {
     #[serde(flatten)]
     pub inner: alloy_rpc_types_eth::Header,
     /// Non-payment gas limit.
+    #[serde(with = "alloy::serde::quantity")]
     pub non_payment_gas_limit: u64,
     /// Non-payment gas used.
+    #[serde(with = "alloy::serde::quantity")]
     pub non_payment_gas_used: u64,
 }
 


crates/node/src/rpc/header.rs
```diff
@@ -0,0 +1,138 @@
...
+#[serde(rename_all = "camelCase")]
+pub struct Header {
+    /// Inner Ethereum header.
+    #[deref]
+    #[serde(flatten)]
+    pub inner: alloy_rpc_types_eth::Header,
+    /// Non-payment gas limit.
+    pub non_payment_gas_limit: u64,
+    /// Non-payment gas used.
+    pub non_payment_gas_used: u64,
```
mattsse, line 22: these should also be quantity I believe?

crates/node/src/rpc/header.rs
```diff
@@ -0,0 +1,138 @@
...
+
+        let non_payment_gas_limit = extra_data
+            .as_ref()
+            .map(|e| e.non_payment_gas_limit)
+            .unwrap_or_default();
+
+        let non_payment_gas_used = extra_data
+            .as_ref()
+            .map(|e| e.non_payment_gas_used)
+            .unwrap_or_default();
```
mattsse, line 40: these are fine because we can assume that these are correctly encoded into the extradata

crates/node/src/rpc/header.rs
klkvr: oops yep

Issue #315, opened 2025-09-23
Write payment lane integration test


fgimenez:
* Full blocks w/ non-payments transactions, say a “dumb” swap action or something, sustained for N blocks
* Then as the above load continues, we start loading up on payments transactions
* Expectation: Payments transactions can still fit in the next block as there’s capacity, and their fee is also low as it’s not competing w/ DeFi

PR #328, opened 2025-09-24
test: add payment lane integration tests

fgimenez:
Closes: #315

tests added:

- `test_payment_lane_with_mixed_load`: validates payment transactions can be included despite sustained non-payment load, this implements the scenario described in #315
- `test_payment_lane_ordering`: verifies transactions arriving in any order are correctly reordered in blocks
- `test_payment_lane_gas_limits`: confirms payment transactions have separate gas allocation

also adds tracing logs to evm code.

### DIFF
--- a/crates/evm/src/block.rs
+++ b/crates/evm/src/block.rs
@@ -1,5 +1,5 @@
 use crate::{TempoBlockExecutionCtx, evm::TempoEvm};
-use alloy_consensus::Transaction;
+use alloy_consensus::{Transaction, transaction::TxHashRef};
 use alloy_primitives::Bytes;
 use alloy_sol_types::SolCall;
 use reth_evm::{
@@ -18,7 +18,7 @@ use tempo_chainspec::TempoChainSpec;
 use tempo_precompiles::{TIP_FEE_MANAGER_ADDRESS, contracts::IFeeManager::executeBlockCall};
 use tempo_primitives::{TempoReceipt, TempoTxEnvelope};
 use tempo_revm::evm::TempoContext;
-use tracing::{debug, trace};
+use tracing::trace;
 
 /// Builder for [`TempoReceipt`].
 #[derive(Debug, Clone, Copy, Default)]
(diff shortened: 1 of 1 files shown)


crates/evm/src/block.rs
```diff
@@ -129,25 +130,50 @@ where
...
+            target: "tempo::block",
+            is_payment = is_payment,
+            is_system = is_system,
+            gas_limit = gas_limit,
+            non_payment_gas_left = self.non_payment_gas_left,
+            seen_payment = self.seen_payment_tx,
+            seen_system = self.seen_system_tx,
+            tx_to = ?tx.tx().to(),
+            "Executing transaction"
+        );
```
klkvr: those logs would be super noisy during backfill so I'd prefer to either remove them or downgrade to `trace`

crates/node/tests/it/payment_lane.rs
```diff
@@ -0,0 +1,590 @@
...
+
+            // This might indicate the ordering constraint is being violated
+            // or there's another issue
+            panic!("{tx_type} transaction failed - this might indicate improper lane ordering");
+        }
+        println!(
+            "  {} transaction succeeded (gas used: {})",
+            tx_type, receipt.gas_used
+        );
+    }
```
klkvr, line 592: should we assert that all txs have landed in correct order and within a single block?

crates/evm/src/block.rs
fgimenez: good point, removed

crates/node/tests/it/payment_lane.rs
fgimenez: absolutely, the ordering assertion was missing, added, i've also increased the number of txs sent. 

but why on a single block? the block production and tx sending are not in sync, they can land on several blocks

crates/evm/src/block.rs
```diff
@@ -129,25 +130,38 @@ where
...
+                non_payment_gas_left = self.non_payment_gas_left,
+                "Rejecting: non-payment gas limit exceeded"
+            );
             return Err(BlockValidationError::msg(
                 "transaction gas limit exceeds non-payment gas limit",
             )
             .into());
         }
 
+        trace!(target: "tempo::block", "Transaction validation passed, executing");
```
klkvr: this is not super useful without a txhash

crates/evm/src/block.rs
```diff
@@ -158,12 +172,31 @@ where
...
+        let is_system = tx.tx().is_system_tx();
+
+        debug!(
+            target: "tempo::block",
+            is_payment = is_payment,
+            is_system = is_system,
+            gas_used = gas_used,
+            non_payment_gas_left_before = self.non_payment_gas_left,
+            "Committing transaction"
+        );
```
klkvr: this logic would still be noisy

crates/evm/src/block.rs
```diff
@@ -158,12 +172,31 @@ where
...
             self.seen_payment_tx = true;
+            debug!(target: "tempo::block", "Marked payment transaction as seen");
         } else {
-            self.non_payment_gas_left -= gas_used;
+            self.non_payment_gas_left = self.non_payment_gas_left.saturating_sub(gas_used);
+            debug!(
+                target: "tempo::block",
+                non_payment_gas_left_after = self.non_payment_gas_left,
+                "Updated non-payment gas left"
+            );
```
klkvr: same here

PR #330, opened 2025-09-24
feat(commonware-node): add builder timeout when proposing new block

### LINKED ISSUE #329: Pace block building
Right now, blocks are built as fast as possible, meaning that if the execution layer's mempools are empty, the validators will rip through consensus and produce thousands of blocks within no time at all.

To stabilize the block building, we should add a timeout that acts as a barrier before resolving a payload ID.

The following should be called only after a timer is reached: https://github.com/tempoxyz/tempo/blob/b165f70422261f447537b4980ca9f19631d357c9/crates/commonware-node/src/consensus/execution_driver/mod.rs#L452-L462

An appropriate place to configure this value seems here: https://github.com/tempoxyz/tempo/blob/b165f70422261f447537b4980ca9f19631d357c9/crates/commonware-node-config/src/timeouts.rs

To be in line with commonware, we should make use of [`Clock::sleep`](https://docs.rs/commonware-runtime/0.0.62/commonware_runtime/trait.Clock.html#tymethod.sleep) rather than `tokio::sleep`

0xKitsune:
Closes #329 

This PR adds a configurable sleep time between `new_payload()` and `resolve_kind()` calls to the payload builder to allow the block builder to process more transactions.

### DIFF
--- a/crates/commonware-node/src/consensus/execution_driver/mod.rs
+++ b/crates/commonware-node/src/consensus/execution_driver/mod.rs
@@ -54,7 +54,7 @@ pub(super) struct ExecutionDriverBuilder<TContext> {
     /// A handle to the execution node to verify and create new payloads.
     pub(super) execution_node: TempoFullNode,
 
-    /// The amount of time to wait for payload builder before resolving payload.
+    /// The minimum amount of time to wait before resolving a new payload from the builder
     pub(super) new_payload_wait_time: Duration,
 }
 


crates/commonware-node-config/src/timeouts.rs
```diff
@@ -27,6 +27,10 @@ pub struct Config {
 
     /// The number of views until a new leader is elected. Also called a skip timeout.
     pub views_until_leader_skip: u64,
+
+    /// The amount of time to wait for payload builder before resolving payload.
+    #[serde(with = "humantime_serde")]
+    pub payload_builder_timeout: Duration,
```
SuperFluffy: Let's give this a more fitting name. It sounds like we are timing out the payload builder, even though nothing could be farther from the truth.

```suggestion
    pub wait_to_resolve_proposal: Duration
```

crates/commonware-node-config/src/timeouts.rs
```diff
@@ -27,6 +27,10 @@ pub struct Config {
 
     /// The number of views until a new leader is elected. Also called a skip timeout.
     pub views_until_leader_skip: u64,
+
+    /// The amount of time to wait for payload builder before resolving payload.
```
SuperFluffy, line 31: IMO don't talk about the payload builder here because that does not mean anything in the context of consensus:

```suggestion
    /// The minimum amount of time we give the application/state machine/execution layer to propose a block.
```

crates/commonware-node/src/consensus/execution_driver/mod.rs
```diff
@@ -438,6 +443,9 @@ impl Inner<Init> {
             .and_then(|ret| ret.wrap_err("execution layer rejected request"))
             .wrap_err("failed requesting new payload from the execution layer")?;
 
+        // Sleep for the configured payload builder timeout
```
SuperFluffy: Redundant comment. But maybe add a `tracing::debug!` log instead? Might be interesting information in case we get timed out by consensus.

crates/commonware-node-config/src/timeouts.rs
0xKitsune: Updated to `new_payload_wait_time`

Issue #337, opened 2025-09-25
Logs and metrics in payload builder


shekhirin:
Add logs in payload builder: 
1. Trace log for every transaction execution with the time it took to execute it https://github.com/tempoxyz/tempo/blob/0adbfb4bdfbdfc037bf7e304db8460478fff3ced/crates/payload/builder/src/lib.rs#L253
2. Debug log for the time it took to execute all transactions https://github.com/tempoxyz/tempo/blob/0adbfb4bdfbdfc037bf7e304db8460478fff3ced/crates/payload/builder/src/lib.rs#L300-L305
3. Debug log for the time it took to call `builder.finish()` which merges all EVM transitions and calculates the state root https://github.com/tempoxyz/tempo/blob/0adbfb4bdfbdfc037bf7e304db8460478fff3ced/crates/payload/builder/src/lib.rs#L309

For the same three points we should add metrics that will report:
1. Execution time, per transaction
2. Execution time, total
3. `builder.finish()` time

PR #339, opened 2025-09-25
feat(payload): metrics and logs

shekhirin:
Closes https://github.com/tempoxyz/tempo/issues/337

Additionally removes a `tx.clone()`


crates/payload/builder/src/lib.rs
```diff
@@ -198,7 +203,7 @@ where
 
         let chain_spec = self.provider.chain_spec();
 
-        debug!(target: "payload_builder", id=%attributes.payload_id(), parent_header = ?parent_header.hash(), parent_number = parent_header.number, "building new payload");
+        debug!(target: "payload_builder", "building new payload");
```
SuperFluffy: Side-comment on this (not actionable right now because we are not doing OTEL): information like "entered this function" is technically not necessary because the created span itself carries that info.

With tracing-subscribers one can in theory do `with_span_events(FmtSpan::NEW)`, but doing that everywhere leads to hilariously spammy logs.

crates/payload/builder/src/lib.rs
```diff
@@ -256,7 +273,14 @@ where
                 continue;
             }
 
-            let gas_used = match builder.execute_transaction(tx.clone()) {
+            let tx_rlp_length = tx.inner().length();
+            let effective_tip_per_gas = tx.effective_tip_per_gas(base_fee);
+
+            let tx_debug_repr = tracing::enabled!(target: "payload_builder", Level::TRACE)
```
SuperFluffy: That's cool, I never used that.

crates/payload/builder/src/lib.rs
```diff
@@ -276,11 +300,11 @@ where
                 })) => {
                     if error.is_nonce_too_low() {
                         // if the nonce is too low, we can skip this transaction
-                        trace!(target: "payload_builder", %error, ?tx, "skipping nonce too low transaction");
+                        trace!(target: "payload_builder", %error, tx = %tx_debug_repr.unwrap(), "skipping nonce too low transaction");
```
SuperFluffy: Are we safe `unwrap`ping here? Tracing has a contract that code in events must not panic.

Also side-note on `%error` - are we getting the full source chain here? If we don't, you can either do `error = error as &dyn std::error::Error` or use https://github.com/tempoxyz/tempo/blob/893b97f5eb9239ba5957bdbdaef11f5e632a460f/crates/telemetry-util/src/lib.rs#L91

crates/payload/builder/src/lib.rs
```diff
@@ -276,11 +300,11 @@ where
...
                     } else {
                         // if the transaction is invalid, we can skip it and all of its
                         // descendants
-                        trace!(target: "payload_builder", %error, ?tx, "skipping invalid transaction and its descendants");
+                        trace!(target: "payload_builder", %error, tx = %tx_debug_repr.unwrap(), "skipping invalid transaction and its descendants");
```
SuperFluffy: Same comment as above re `%error` and the unwrap.

crates/payload/builder/src/lib.rs
```diff
@@ -293,17 +317,29 @@ where
...
 
-            block_transactions_rlp_length += tx.inner().length();
+            block_transactions_rlp_length += tx_rlp_length;
 
             // update and add to total fees
-            let miner_fee = tx
-                .effective_tip_per_gas(base_fee)
-                .expect("fee is always valid; execution succeeded");
+            let miner_fee =
+                effective_tip_per_gas.expect("fee is always valid; execution succeeded");
```
SuperFluffy, line 344: Q: why is it always valid? EDIT: Ah, because execution succeeded; that does make sense.

crates/payload/builder/src/lib.rs
shekhirin: removed the unwrap and just using an empty string, but this should always be safe, because we're using the same trace level as we're checking above

shekhirin: regarding the error: unsure if we need this here? It's just https://github.com/bluealloy/revm/blob/0e05a3070d2f94b44ce5555604f9ac9ca5d2cb08/crates/context/interface/src/result.rs#L324 underneath

crates/payload/builder/src/lib.rs
shekhirin: it can return `None` only if `max_fee_per_gas < base_fee`, which is not possible if the execution succeeded, because if max fee per gas (the total amount user pays) is less than base fee, it will be out of gas

crates/payload/builder/src/lib.rs
SuperFluffy: > regarding the error: unsure if we need this here? It's just https://github.com/bluealloy/revm/blob/0e05a3070d2f94b44ce5555604f9ac9ca5d2cb08/crates/context/interface/src/result.rs#L324 underneath

Late reply: IMO ideally one should always print the source chain. If somebody decides to wrap this error (for example, giving it some `wrap_err("<context>")`) you otherwise lose the underlying error without noticing until you need it.

PR #346, opened 2025-09-25
feat: `TempoPriorityOrdering`

0xKitsune:
This PR removes the `LanedTransactions` struct with preference for a new custom `TempoPriorityOrdering` policy.

### DIFF
--- a/crates/transaction-pool/src/lib.rs
+++ b/crates/transaction-pool/src/lib.rs
@@ -4,7 +4,6 @@
 #![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
 
 use alloy_consensus::Transaction;
-use alloy_primitives::U256;
 use reth_transaction_pool::{
     Pool, Priority, TransactionOrdering, TransactionValidationTaskExecutor,
     blobstore::DiskFileBlobStore,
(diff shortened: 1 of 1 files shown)


crates/payload/builder/src/lib.rs
```diff
@@ -240,6 +239,20 @@ where
...
+                && pool_tx.gas_limit() + non_payment_gas_used > non_payment_gas_limit
+            {
+                best_txs.mark_invalid(
+                    &pool_tx,
+                    InvalidPoolTransactionError::Other(Box::new(
+                        TempoPoolTransactionError::ExceedsNonPaymentLimit,
+                    )),
+                );
+                continue;
+            }
```
klkvr, line 246: let's move this condition above. we might discard quite a lot of transactions like this and rn every discarded tx will involve a bit of cloning due to `pool_tx.to_consensus()` call above which is somewhat wasteful

crates/transaction-pool/src/lib.rs
```diff
@@ -14,6 +17,91 @@ pub mod validator;
...
+pub struct TempoPriorityOrdering;
+
+/// Tempo priority ordering attributes
+///
+/// The ordering of fields here is important.
+#[derive(Debug, Default, Clone, Ord, PartialOrd, Eq, PartialEq)]
+pub struct TempoPriority {
+    non_payment: bool,
+    effective_tip_per_gas: Option<U256>,
+}
```
klkvr, line 37: should we produce `Priority::None` when effective tip can't be calculated like `CoinbaseTipOrdering` does?

0xKitsune: I updated to return `Priority::None` when effective tip can't be calculated.

crates/payload/builder/src/laned.rs
mattsse, line 1: sorry @fgimenez I haven't thought about this solution which effectively does the same thing but already in the pool itself

it believe all of #328 test cases are still valid

PR #347, opened 2025-09-26
feat(tempo-bench): Simplify `run-max-tps`

0xKitsune:
This PR is a first pass at simplifying the `run-max-tps` subcommand in `tempo-bench`.

These changes simplify the current implementation by removing the old config-driven and worker-based system. The new tool features:
- Simpler CLI arguments instead of external config files
- Pre-generates all txs in parallel
- Rate limits sending txs to the node to ensure a predictable transactions per second 
- Reports TPS for the specified duration

Below is an example of how to use the updated tool:
```bash
 tempo-bench run-max-tps --duration 15 --tps 20000
```


bin/tempo-bench/README.md
```diff
@@ -164,7 +164,7 @@ tempo-bench generate-genesis --accounts 50000 --output genesis.json
 ### 2. Start the Node
 
 ```bash
-    reth node \
+  tempo node \
```
SuperFluffy, line 167: Should we mention `tempo-commonware` here? Maybe not, setup is a bit more involved.

bin/tempo-bench/README.md
```diff
@@ -215,3 +215,43 @@ rm -rf "$HOME/Library/Application Support/reth/"
 # Windows (default: %APPDATA%/reth/)
 rmdir /s "%APPDATA%\reth"
 ```
+
+
+
+
+### Sampling 
+Use the following commands to run the node with sampling.
```
SuperFluffy: This? https://github.com/mstange/samply

bin/tempo-bench/src/opts.rs
```diff
@@ -1,16 +1,15 @@
...
 pub struct TempoBench {
-    // TODO: add node args
     #[command(subcommand)]
     pub cmd: TempoBenchSubcommand,
 }
 
 #[derive(Subcommand, Debug)]
 pub enum TempoBenchSubcommand {
-    RunMaxTPS(TPSArgs),
+    RunMaxTPS(MaxTPSArgs),
```
SuperFluffy: Snakecase? -> `MaxTpsArgs`

crates/evm/src/lib.rs
```diff
@@ -145,29 +145,31 @@ impl ConfigureEvm for TempoEvmConfig {
...
+        };
+
+        Ok(ctx)
     }
 }
 
 impl ConfigureEngineEvm<TempoExecutionData> for TempoEvmConfig {
     fn evm_env_for_payload(&self, payload: &TempoExecutionData) -> EvmEnvFor<Self> {
-        self.evm_env(&payload.0).expect("evm_env should not fail")
+        self.evm_env(&payload.0).expect("TODO: handle error")
```
SuperFluffy: Looks like there is no error handling? IMO leave the `expect` bessage before - it does not contain a lot of extra info ("why is evm_env expected to never fail?!"), but IMO better than leaving this.

crates/evm/src/lib.rs
```diff
@@ -145,29 +145,31 @@ impl ConfigureEvm for TempoEvmConfig {
...
     }
 
     fn context_for_payload<'a>(
         &self,
         payload: &'a TempoExecutionData,
     ) -> ExecutionCtxFor<'a, Self> {
         self.context_for_block(&payload.0)
-            .expect("context_for_block should not fail")
+            .expect("TODO: handle error")
```
SuperFluffy: same for the error msg.

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -1,155 +1,330 @@
...
+    interface ERC20 {
+        function transfer(address to, uint256 amount) external returns (bool);
+    }
+}
 
 /// Run maximum TPS throughput benchmarking
 #[derive(Parser, Debug)]
-pub struct TPSArgs {
-    /// Path to the configuration file
+pub struct MaxTPSArgs {
```
SuperFluffy:
```suggestion
pub struct MaxTPSArgs`
```

Does this actually need to be `pub`? Could we get away with locking all of these fields down?

crates/evm/src/lib.rs
SuperFluffy, line 1: The changes to this file seem orthogonal to the rest of the patch - remove?

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -1,155 +1,330 @@
...
+
+    /// Target URLs for network connections
+    #[arg(long, default_values_t = vec!["http://localhost:8545".to_string()])]
+    pub target_urls: Vec<String>,
+
+    /// Total network connections
+    #[arg(long, default_value = "100")]
+    pub total_connections: u64,
+
+    #[arg(long)]
```
SuperFluffy, line 80: Let's add a doc comment above this, even if its self-explanatory.

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -1,155 +1,330 @@
...
+        let transactions = Arc::new(
+            generate_transactions(
+                total_txs,
+                self.accounts,
+                &self.mnemonic,
+                self.chain_id,
+                self.token_address,
+                &target_urls[0],
+            )
+            .await?,
```
SuperFluffy: error context

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -1,155 +1,330 @@
...
-        config::init(Config::from_file(&self.config)?);
+        // Set file descriptor limit if provided
+        if let Some(fd_limit) = self.fd_limit {
+            increase_nofile_limit(fd_limit)?;
+        }
+
+        let target_urls: Vec<Url> = self
+            .target_urls
+            .iter()
+            .map(|url| url.as_str().parse().expect("Could not parse Url"))
```
SuperFluffy: Let's do error handling here and report which URL failed to be parsed. I am thinking of this:

```rust
.map(|s| s.parse::<Url>().wrap_err_with(|| format!("failed to parse `{s}` as URL")))
.collect::<eyre::Result<Vec<_>>>()
.wrap_err("failed parsing input target URLs")?;
```

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -1,155 +1,330 @@
...
+        // Spawn workers and send transactions
+        send_transactions(
+            transactions,
+            self.workers,
+            self.total_connections,
+            target_urls,
+            self.tps,
+            self.disable_thread_pinning,
+            tx_counter,
+        )?;
```
SuperFluffy: error contet

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -1,155 +1,330 @@
...
+
+        // Wait for all sender threads to finish
+        std::thread::sleep(Duration::from_secs(self.duration));
```
SuperFluffy, line 139: Wonder if we should join the sender threads instead of killing them.

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -1,155 +1,330 @@
...
+        println!("Finished sending transactions");
```
SuperFluffy, line 140: Might be a misnomer if we exit because the sleep finished before the threads.

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -1,155 +1,330 @@
...
 }
 
-impl TPSArgs {
+impl MaxTPSArgs {
     pub async fn run(self) -> eyre::Result<()> {
-        println!("[~] Loading config from {}...", self.config.display());
-        config::init(Config::from_file(&self.config)?);
+        // Set file descriptor limit if provided
+        if let Some(fd_limit) = self.fd_limit {
+            increase_nofile_limit(fd_limit)?;
```
SuperFluffy: context; this might give a very hard to understand error.

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -1,155 +1,330 @@
...
-        tokio::spawn(NETWORK_STATS.start_reporter(
-            Duration::from_secs(config::get().reporters.network_stats_report_interval_secs),
-            cancellation_token.clone(),
-        ));
-
-        tokio::spawn(async move {
-            tokio::time::sleep(Duration::from_secs(config::get().benchmark.run_duration)).await;
-            println!("[*] Stopping reporters...");
-            cancellation_token.cancel();
+            let signature = signer.sign_transaction_sync(&mut tx).unwrap();
```
SuperFluffy: error handling

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -1,155 +1,330 @@
...
+    );
+
+    // Fetch current nonces for all accounts
+    let provider = ProviderBuilder::new().connect_http(rpc_url.clone());
+    println!("Fetching nonces for {} accounts...", signers.len());
+
+    let mut params = Vec::new();
+    for signer in signers {
+        let address = signer.address();
+        let current_nonce = provider.get_transaction_count(address).await?;
```
SuperFluffy: error context

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -1,155 +1,330 @@
...
+    let signers: Vec<PrivateKeySigner> = (0..num_accounts as u32)
+        .into_par_iter()
+        .tqdm()
+        .map(|i| {
+            MnemonicBuilder::<English>::default()
+                .phrase(mnemonic)
+                .index(i)
+                .unwrap()
+                .build()
+                .unwrap()
```
SuperFluffy: let's not panic - we are already in an `eyre::Result` function so we can just propagate this down.

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -1,155 +1,330 @@
...
+        })
+        .collect();
+
+    let txs_per_sender = total_txs / num_accounts;
+    assert!(
```
SuperFluffy:
```suggestion
    eyre::ensure!(
```

bin/tempo-bench/Cargo.toml
```diff
@@ -47,4 +46,5 @@ ratelimit = "0.10"
 rayon = "1.10"
 rlimit = "0.10.2"
 simple-tqdm = { version = "0.2.0", features = ["rayon"] }
-thousands = "0.2.0"
\ No newline at end of file
+thousands = "0.2.0"
+governor = "0.10.1"
```
SuperFluffy, line 49: sort alphabetically?

bin/tempo-bench/README.md
0xKitsune: Yeah I think its probably best to leave this as is for now and we can update in a future PR if needed.

crates/evm/src/lib.rs
0xKitsune: Yeah, we needed to bump reth for `--engine.legacy-state-root` and this broke `evm/src/lib.rs`. However this was patched later and merged to main, Ill update to remove these changes.

bin/tempo-bench/src/cmd/max_tps.rs
0xKitsune: Lets plan to update in a follow on PR.

PR #348, opened 2025-09-26
feat(payload): Check for `interrupt` signal while building

### LINKED ISSUE #254: Add a way to interrupt payload builder
When resolving payload we might want to stop payload builder from draining entire pool and ask it to finalize the block earlier so that `resolve` call does not block.

One way to do this could be to introduce custom `PayloadBuilderAttributes` that would contain an `AtomicBool` or `oneshot::Receiver` that gets notified once we want to resolve the payload.

That way we can just check it here similarly to `cancel` marker and funish block building early
https://github.com/tempoxyz/tempo/blob/05a2d06d7c287c7465fa1be3b35d3a29d0414a21/crates/payload/builder/src/lib.rs#L186-L189

We should introduce new `TempoPayloadBuilderAttributes` here:
https://github.com/tempoxyz/tempo/blob/05a2d06d7c287c7465fa1be3b35d3a29d0414a21/crates/payload/types/src/lib.rs#L59

RomanHodulak:
Closes #254 

This change allows for breaking out of the transactions loop in the payload building.

Since there is no need for notifications, the flag was reimplemented as an atomic bool.

### DIFF
--- a/crates/commonware-node/src/consensus/execution_driver/mod.rs
+++ b/crates/commonware-node/src/consensus/execution_driver/mod.rs
@@ -415,32 +415,34 @@ impl Inner<Init> {
             timestamp = parent.timestamp().saturating_add(1);
         }
 
+        let attrs = TempoPayloadBuilderAttributes::new(EthPayloadBuilderAttributes {
+            // XXX: derives the payload ID from the parent so that
+            // overlong payload builds will eventually succeed on the
+            // next iteration: if all other nodes take equally as long,
+            // the consensus engine will kill the proposal task (see
+            // also `response.cancellation` below). Then eventually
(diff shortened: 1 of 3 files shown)


crates/commonware-node/src/consensus/execution_driver/mod.rs
```diff
@@ -415,32 +415,34 @@ impl Inner<Init> {
             timestamp = parent.timestamp().saturating_add(1);
         }
 
+        let payload = TempoPayloadBuilderAttributes::new(EthPayloadBuilderAttributes {
```
klkvr: nitpick: let's rename this to `attrs`

PR #351, opened 2025-09-26
feat: native fee sponsoring

klkvr:
PoC for native fee sponsoring.

Adds optional `feePayerSignature` field to `TxFeeToken`.


crates/primitives/src/transaction/envelope.rs
```diff
@@ -83,6 +83,22 @@ impl TempoTxEnvelope {
...
+                if let Some(fee_payer_signature) = tx.tx().fee_payer_signature {
+                    fee_payer_signature
+                        .recover_address_from_prehash(&tx.tx().fee_payer_signature_hash(sender))
+                } else {
+                    Ok(sender)
+                }
+            }
+            _ => Ok(sender),
+        }
+    }
```
gakonst, line 100: In an ideal world this would be pre-recovered when the transaction is received so we can do it in parallel vs during serial execution

crates/precompiles/src/contracts/tip_fee_manager/mod.rs
```diff
@@ -193,7 +193,7 @@ impl<'a, S: StorageProvider> TipFeeManager<'a, S> {
     /// Unused gas is later returned via collect_fee_post_tx
     pub fn collect_fee_pre_tx(
```
gakonst, line 196: beautiful

crates/primitives/src/transaction/envelope.rs
klkvr: recovery happens during tx -> TxEnv conversion so this can already be parallelized. also we could cache that in the pool

crates/precompiles/src/contracts/provider.rs
shekhirin, line 95: should we also modify this method to do correct `eth_call` estimations?

klkvr: yep updated the logic here 72de64d

though we're not going to expose any ways to configure `feePayer` on `eth_call`s for now https://ithacaxyz.slack.com/archives/C097VAML3FW/p1759362571412549?thread_ts=1759357210.391789&cid=C097VAML3FW

crates/primitives/src/transaction/fee_token.rs
```diff
@@ -305,12 +390,36 @@ impl SignableTransaction<Signature> for TxFeeToken {
     }
 
     fn encode_for_signing(&self, out: &mut dyn alloy_rlp::BufMut) {
+        let skip_fee_token = self.fee_payer_signature.is_some();
```
fgimenez, line 394: aiui, with this the sender is not aware at all of which fee token will be used right? does not commit to it, it is delegated to the sponsor who commits to pay with fee token to specific user. if so, maybe we could have a comment about this?

crates/revm/src/tx.rs
```diff
@@ -26,6 +27,25 @@ pub struct TempoTxEnv {
...
+    /// - None corresponds to a transaction without a fee payer
+    pub fee_payer: Option<Option<Address>>,
+}
+
+impl TempoTxEnv {
+    /// Resolves fee payer from the signature.
+    pub fn fee_payer(&self) -> Result<Address, InvalidTransaction> {
+        if let Some(fee_payer) = self.fee_payer {
+            // todo: introduce custom error type or Other variant upstream
+            fee_payer.ok_or(InvalidTransaction::OverflowPaymentInTransaction)
```
fgimenez, line 44: before we have custom error type, here we would return `OverflowPaymentInTransaction` when signature recovery fails? would it be better `InvalidTransaction::InvalidSignature`?

klkvr: there's no `InvalidTransaction::InvalidSignature` variant

crates/primitives/src/transaction/fee_token.rs
klkvr: yep makes sense, added doc in 1534af4

crates/revm/src/tx.rs
fgimenez: yeah, closest is `alloy_primitives::SignatureError`, would it make sense to return this instead of `InvalidTransaction` as the error of the result in `fee_payer`?

klkvr: we want this error to get converted into `EVMError::Transaction` variant. imo we need a proper fix for this which'd be implementing https://github.com/tempoxyz/tempo/issues/369

fgimenez: makes sense 👍

crates/revm/src/handler.rs
```diff
@@ -38,24 +38,28 @@ const DEFAULT_7702_DELEGATE_CODE_HASH: B256 =
 /// Tempo EVM [`Handler`] implementation with Tempo specific modifications:
 ///
 /// Fees are paid in fee tokens instead of account balance.
-#[derive(Debug, Clone)]
-pub struct TempoEvmHandler<DB, I> {
+#[derive(Debug)]
+pub struct TempoEvmHandler<DB: reth_evm::Database, I> {
```
mattsse: is enforcing the trait bound required here?

crates/primitives/src/transaction/fee_token.rs
```diff
@@ -121,47 +125,113 @@ impl TxFeeToken {
...
+            addr.encode(out);
+        } else {
+            out.put_u8(EMPTY_STRING_CODE);
+        }
+        encode_signature(&self.fee_payer_signature, out);
+    }
+
+    pub fn fee_payer_signature_hash(&self, sender: Address) -> B256 {
+        let payload_length = self.rlp_encoded_fields_length(|_| sender.length(), false);
+        let mut buf = Vec::new();
```
0xKitsune: nit: are we able to use `Vec::with_capacity` here?

crates/primitives/src/transaction/envelope.rs
0xKitsune, line 100: Is there a reason that we need to pass in the `sender`? Another approach might be to simply return `Result<Option<Address>, SignatureError>` and if `fee_payer?.is_none()`, just use the tx caller. This feels a little cleaner but lmk if I am overlooking something here.

crates/primitives/src/transaction/fee_token.rs
```diff
@@ -305,12 +390,38 @@ impl SignableTransaction<Signature> for TxFeeToken {
     }
 
     fn encode_for_signing(&self, out: &mut dyn alloy_rlp::BufMut) {
+        // We skip encoding the fee token if the signature is present to ensure that user
+        // does not commit to a specific fee token when someone else is paying for the transaction.
```
0xKitsune, line 393: Can we simplify by updating the logic that fetches the fee token to check the token in this order: `fee_payer -> tx fee token -> user fee token -> validator token`. Then I think we would be able to always encode the fee token here.

crates/primitives/src/transaction/fee_token.rs
klkvr: yep 3dc31aa

crates/primitives/src/transaction/envelope.rs
klkvr: we need to know sender to recover the `feePayer` from the signature https://github.com/tempoxyz/tempo/blob/3dc31aa64523486ab3fd11080a986fa16b23bdd0/crates/primitives/src/transaction/fee_token.rs#L179-L195

and it seemed convinient to also return the sender address instead of `None` so that this method always returns the actual tx fee payer. lmk if you think there's a better name for this

0xKitsune: Sounds good, makes sense!

PR #352, opened 2025-09-26
chore: remove `JournalStorageProvider`

klkvr:
We have a way to create `EvmInternals` from context so this can be removed


crates/revm/src/journal_storage_provider.rs
```diff
@@ -1,247 +0,0 @@
...
-    fn get_account_info(&mut self, address: Address) -> Result<AccountInfo, Self::Error> {
-        self.journal.load_account(address)?;
-        let account = self.journal.load_account_code(address)?;
-        Ok(account.info.clone())
-    }
-
-    fn sstore(&mut self, address: Address, key: U256, value: U256) -> Result<(), Self::Error> {
-        self.journal.load_account(address)?;
-        self.journal.touch_account(address);
-        self.journal.sstore(address, key, value)?;
```
mattsse, line 52: hmm, this behaviour is perhaps different on `EvmStorageProvider`?

PR #355, opened 2025-09-29
fix: add serde rename for feeToken field in eth_estimateGas requests

fgimenez:
fixes `eth_estimateGas` for type 0x77 transactions to correctly use the fee token specified in the transaction request, rather than the user's configured default fee token.

when calling `eth_estimateGas` with a type 0x77 transaction that specifies a `feeToken`, the estimation was incorrectly using the user's configured fee token instead. This caused gas estimation to fail with `gas required exceeds allowance (0)` when the user had no balance in their configured token, even if they had sufficient balance in the transaction's specified token.

the `TempoTransactionRequest` struct had `#[serde(rename = "camelCase")]` at the struct level, but this attribute in serde only affects the struct's name when it appears as a field in another struct. Each field needs its own `#[serde(rename)]` attribute, without this, `feeToken` field was ignored during deserialization, `fee_token` became `None` .

also updated `.gitignore`, removed entries from reth-malachite e2e tests and added new ones for the current tests.

### DIFF
--- a/crates/node/src/rpc/request.rs
+++ b/crates/node/src/rpc/request.rs
@@ -18,10 +18,10 @@ use tempo_revm::TempoTxEnv;
 
 /// An Ethereum [`TransactionRequest`] with an optional `fee_token`.
 #[derive(Clone, Debug, Default, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
 pub struct TempoTransactionRequest {
     #[serde(flatten)]
     pub inner: TransactionRequest,
-    #[serde(rename = "feeToken")]
     pub fee_token: Option<Address>,
 }
 


crates/node/src/rpc/request.rs
```diff
@@ -18,10 +18,10 @@ use tempo_revm::TempoTxEnv;
 
 /// An Ethereum [`TransactionRequest`] with an optional `fee_token`.
 #[derive(Clone, Debug, Default, PartialEq, Eq, Hash, Serialize, Deserialize)]
-#[serde(rename = "camelCase")]
```
mattsse, line 21: ah this should have been `rename_all` here

can we double check other container serde attributes?

.gitignore
```diff
@@ -1,11 +1,8 @@
 target
-genesis.json
-# Testnet runtime directories
-testnet/nodes/
-testnet/logs/
 
-# logs
-data
```
mattsse, line 8: any particular reason to changes this here?

crates/node/src/rpc/request.rs
fgimenez: good point, much better. checked and there are no more occurrences of `rename`

.gitignore
fgimenez: yep, these ignores come from `reth-malachite`, no longer relevant, will mention in the PR description

PR #357, opened 2025-09-29
feat(precompiles): `system_transfer_from`

0xKitsune:
This PR adds a `system_transfer_from` method to `TIP20Token`, allowing specific precompiles to transfer tokens without approval checks while keeping normal safety validations. The fee AMM is updated to use this method, removing special cases that skip allowance logic for the fee manager precompile.

### DIFF
--- a/crates/precompiles/src/contracts/tip20.rs
+++ b/crates/precompiles/src/contracts/tip20.rs
@@ -1489,11 +1489,7 @@ mod tests {
 
         assert_eq!(
             storage.events[&token_id_to_address(token_id)].last(),
-            Some(&TIP20Event::Transfer(ITIP20::Transfer {
-                from,
-                to,
-                amount,
-            }).into_log_data())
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/contracts/tip20.rs
```diff
@@ -1428,4 +1442,49 @@ mod tests {
...
+
+        let mut roles = token.get_roles_contract();
+        roles.grant_role_internal(&admin, *ISSUER_ROLE);
+
+        token
+            .mint(&admin, ITIP20::mintCall { to: from, amount })
+            .unwrap();
+
+        let result = token.system_transfer_from(from, to, amount);
+        assert!(result.is_ok());
```
fgimenez, line 1488: maybe we could also check for `Transfer` events emitted? like:
```
assert_eq!(
      storage.events[&token_id_to_address(token_id)].last(),
      Some(&TIP20Event::Transfer(ITIP20::Transfer {
          from,
          to,
          amount,
      }).into_log_data())
  );
```

Issue #356, opened 2025-09-29
Update to latest reth with rpc changes


mattsse:
we introduced https://github.com/paradigmxyz/reth/pull/18685

which changes some of the rpc conversion traits that we also need to update here, and can introduce error types for decoding the fee lane values for the rpc tx receipt

PR #359, opened 2025-09-29
chore: bump reth and fix breaking API changes

fgimenez:
Towards #356

### DIFF
--- a/crates/evm/src/lib.rs
+++ b/crates/evm/src/lib.rs
@@ -122,7 +122,10 @@ impl ConfigureEvm for TempoEvmConfig {
         self.inner.next_evm_env(parent, &attributes.inner)
     }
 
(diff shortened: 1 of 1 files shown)


crates/evm/src/lib.rs
```diff
@@ -122,48 +122,52 @@ impl ConfigureEvm for TempoEvmConfig {
...
             non_payment_gas_limit: attributes.non_payment_gas_limit,
-        }
+        })
     }
 }
 
 impl ConfigureEngineEvm<TempoExecutionData> for TempoEvmConfig {
     fn evm_env_for_payload(&self, payload: &TempoExecutionData) -> EvmEnvFor<Self> {
-        self.evm_env(&payload.0)
+        self.evm_env(&payload.0).expect("evm_env should not fail")
```
mattsse, line 162: @klkvr should we also make these fallible?

klkvr: yep, makes sense

PR #360, opened 2025-09-29
feat: add error types for fee lane decoding in EVM operations

fgimenez:
Closes: #356

### DIFF
--- a/crates/evm/src/error.rs
+++ b/crates/evm/src/error.rs
@@ -7,7 +7,7 @@ use reth_consensus::ConsensusError;
 pub enum TempoEvmError {
     /// Error decoding fee lane data from extra data field.
     #[error("failed to decode fee lane data: {0}")]
-    FeeLandeDecoding(#[from] ConsensusError),
+    FeeLaneDecoding(#[from] ConsensusError),
 
     /// Invalid EVM configuration.
     #[error("invalid EVM configuration: {0}")]


crates/evm/src/error.rs
```diff
@@ -0,0 +1,21 @@
...
+/// Errors that can occur during EVM configuration and execution.
+#[derive(Debug, Clone, PartialEq, Eq, thiserror::Error)]
+pub enum TempoEvmError {
+    /// Error decoding fee lane data from extra data field.
+    #[error("failed to decode fee lane data: {0}")]
+    FeeLaneDecoding(String),
+
+    /// Invalid EVM configuration.
+    #[error("invalid EVM configuration: {0}")]
+    InvalidEvmConfig(String),
```
mattsse: should we just keep box dyn error here @klkvr ?

crates/evm/src/error.rs
```diff
@@ -0,0 +1,21 @@
+//! Error types for Tempo EVM operations.
+
+use reth_consensus::ConsensusError;
+
+/// Errors that can occur during EVM configuration and execution.
+#[derive(Debug, Clone, PartialEq, Eq, thiserror::Error)]
+pub enum TempoEvmError {
+    /// Error decoding fee lane data from extra data field.
+    #[error("failed to decode fee lane data: {0}")]
+    FeeLaneDecoding(String),
```
SuperFluffy: Is `ConsensusError: std::error::Error`?

If yes, let's do:
```suggestion
    FeeLandeDecoding(#[from] ConsensusError),
```

You can then avoid the manual `impl From`

fgimenez: good catch, thx!

crates/evm/src/lib.rs
```diff
@@ -111,32 +113,38 @@ impl ConfigureEvm for TempoEvmConfig {
     }
 
     fn evm_env(&self, header: &Header) -> Result<EvmEnv, Self::Error> {
-        self.inner.evm_env(header)
+        self.inner.evm_env(header).map_err(|_| {
+            TempoEvmError::InvalidEvmConfig("failed to create EVM environment".to_string())
+        })
```
klkvr, line 118: those should be unreachable right? because `inner`'s error is `Infallible`

crates/evm/src/error.rs
klkvr: makes sense, though I think those would never be constructed

crates/evm/src/lib.rs
0xKitsune: It might be nice to keep this, it seems like its technically possible that `evm_env` is updated to return an error without any breaking changes right? In this case we handle this error.

crates/evm/src/error.rs
```diff
@@ -0,0 +1,15 @@
+//! Error types for Tempo EVM operations.
+
+use reth_consensus::ConsensusError;
+
+/// Errors that can occur during EVM configuration and execution.
+#[derive(Debug, Clone, PartialEq, Eq, thiserror::Error)]
+pub enum TempoEvmError {
+    /// Error decoding fee lane data from extra data field.
+    #[error("failed to decode fee lane data: {0}")]
+    FeeLandeDecoding(#[from] ConsensusError),
```
0xKitsune: Looks like this might be a typo.

```suggestion
    FeeLaneDecoding(#[from] ConsensusError),
```

Issue #327, opened 2025-09-24
Add e2e block building test


mattsse:
with payment lane we now have extra rules for how a block is built.

we should add some tests that cover scenarios:

* pool has more payment/nonpayment txs than fit in a block
* pool only has payment/nonpayment
* pool has only a few txs

## TODO
* add e2e test that inject transactions into the pool
* mine a block and ensure the expected outcome/ratio of the block

PR #362, opened 2025-09-29
test: add block building integration tests

fgimenez:
Closes: #327

### DIFF
--- a/crates/node/tests/it/block_building.rs
+++ b/crates/node/tests/it/block_building.rs
@@ -107,26 +107,6 @@ fn extract_user_txs(all_transactions: Vec<TempoTxEnvelope>) -> Vec<TempoTxEnvelo
         .collect()
 }
 
-/// Helper to verify payment transactions come after non-payment transactions
-fn assert_payment_ordering(transactions: &[TempoTxEnvelope]) {
-    let mut payment_found = false;
-    let mut non_payment_after_payment = false;
-
-    for tx in transactions {
-        if tx.is_payment() {
-            payment_found = true;
-        } else if payment_found {
-            non_payment_after_payment = true;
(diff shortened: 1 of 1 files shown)


crates/node/tests/it/block_building.rs
```diff
@@ -0,0 +1,572 @@
...
+            non_payment_after_payment = true;
+            break;
+        }
+    }
+
+    assert!(
+        !non_payment_after_payment,
+        "Non-payment transactions should come before payment transactions"
+    );
+}
```
klkvr: I don't think this will be the case after https://github.com/tempoxyz/tempo/pull/367

crates/node/tests/it/block_building.rs
```diff
@@ -0,0 +1,572 @@
...
+
+/// Helper to setup a test node with optional custom gas limit
+/// Returns the node and TaskManager (which must be kept alive for the node to function)
+async fn setup_test_node(
+    custom_gas_limit: Option<&str>,
+) -> eyre::Result<(
+    reth_e2e_test_utils::NodeHelperType<TempoNode>,
+    reth_ethereum::tasks::TaskManager,
+)> {
```
klkvr, line 117: ideally we find a way to unify the way we setup nodes — right now we do it in 3 different ways iiuc

crates/node/tests/it/block_building.rs
fgimenez: good call, removed

crates/node/tests/it/block_building.rs
fgimenez: sure, created https://github.com/tempoxyz/tempo/issues/382 for tracking

PR #364, opened 2025-09-30
chore: remove malachite

0xKitsune:
This PR removes malachite and related dependencies from the repo.


bin/tempo/src/main.rs
```diff
@@ -1,213 +1,221 @@
...
+    args::FaucetArgs,
+    faucet::{TempoFaucetExt, TempoFaucetExtApiServer},
+};
+use tempo_node::{TempoFullNode, node::TempoNode};
+use tokio::sync::oneshot;
+use tokio_util::either::Either;
+
+// TODO: migrate this to tempo_node eventually.
+#[derive(Debug, Clone, PartialEq, Eq, clap::Args)]
+struct TempoArgs {
```
klkvr, line 46: can we not unify this with `tempo_node::args::TempoArgs` now?

crates/commonware-node/src/consensus/execution_driver/mod.rs
```diff
@@ -443,11 +443,11 @@ impl Inner<Init> {
...
             timestamp,
             suggested_fee_recipient: self.fee_recipient,
-            // XXX(tempo-malachite): for PoS compatibility
+            // XXX(tempo): for PoS compatibility
             prev_randao: B256::ZERO,
-            // XXX(tempo-malachite): empty withdrawals post-shanghai
+            // XXX(tempo): empty withdrawals post-shanghai
             withdrawals: Withdrawals::default(),
-            // TODO: tempo-malachite does this (why?); but maybe we can
+            // TODO: tempo does this (why?); but maybe we can
```
SuperFluffy: That's one confusing TODO now since it reference malachite. :-)

bin/tempo/src/main.rs
SuperFluffy: > can we not unify this with `tempo_node::args::TempoArgs` now?

ACK on unifying this. Should merge `commonware-node` and `node`, but probably in a followup to this (lots of code changes, we won't see what's genuinely new and what's just renamed and moved).

Issue #363, opened 2025-09-30
Block building halts once there's not enough liquidity in the feeAMM pools


Zygimantass:
## What happened?

The transaction load generator drained the FeeAMM LP pools (no arber was running) and the block building process stopped (chain stalled).

## Logs

```
Sep 30 13:28:24 ns5035737 tempo[2369918]: 2025-09-30T13:28:24.395119Z  WARN handle_propose{view=11129968 parent.view=11129692 parent.digest=0x4e6147a36765e2365333c63e865239dd303cf367c909e0b5a6f7996541406d1a}: failed to resolve pending payload err=evm execution error: internal EVM error occurred when executing transaction 0xbdf73f89655462db6bc3d75de25c9a49ff1ee98e6b1e2fe001058f19681c6267: InsufficientLiquidity(InsufficientLiquidity)
Sep 30 13:28:24 ns5035737 tempo[2369918]: 2025-09-30T13:28:24.395146Z  WARN handle_propose{view=11129968 parent.view=11129692 parent.digest=0x4e6147a36765e2365333c63e865239dd303cf367c909e0b5a6f7996541406d1a}: error=[failed creating a proposal, failed getting payload for payload ID `0x4e6147a36765e236`, missing payload]
Sep 30 13:28:36 ns5035737 tempo[2369918]: 2025-09-30T13:28:36.401290Z  INFO update{digest=0x4e6147a36765e2365333c63e865239dd303cf367c909e0b5a6f7996541406d1a}: sending forkchoice-update head_block_hash=0x4e6147a36765e2365333c63e865239dd303cf367c909e0b5a6f7996541406d1a finalized_block_hash=0x4e6147a36765e2365333c63e865239dd303cf367c909e0b5a6f7996541406d1a
Sep 30 13:28:36 ns5035737 tempo[2369918]: 2025-09-30T13:28:36.401355Z  INFO New payload job created id=0x4e6147a36765e236 parent=0x4e6147a36765e2365333c63e865239dd303cf367c909e0b5a6f7996541406d1a
Sep 30 13:28:36 ns5035737 tempo[2369918]: 2025-09-30T13:28:36.401385Z  INFO update{digest=0x4e6147a36765e2365333c63e865239dd303cf367c909e0b5a6f7996541406d1a}: return=PayloadStatus { status: Valid, latest_valid_hash: Some(0x4e6147a36765e2365333c63e865239dd303cf367c909e0b5a6f7996541406d1a) }
[...]

PR #365, opened 2025-09-30
fix: map fee collection errors to transaction validation errors

fgimenez:
Closes: #363

 Fixes block building stall when FeeAMM pools lack liquidity for fee swaps.

Previously, transactions requiring unavailable liquidity caused `InsufficientLiquidity` errors that were wrapped as `InternalBlockExecutionError`, failing the entire block build and stalling the chain.

Now these errors are mapped to `InvalidTransaction` at the handler level (where fee collection happens), properly classifying them as transaction validity errors. The payload builder's existing `InvalidTx` handling then skips these transactions, allowing block building to continue normally.

Adds integration test to verify the fix.

### DIFF
--- a/crates/node/tests/it/liquidity.rs
+++ b/crates/node/tests/it/liquidity.rs
@@ -1,0 +1,177 @@
+use alloy::{primitives::U256, providers::ProviderBuilder, signers::local::MnemonicBuilder};
+
+/// Test block building when FeeAMM pool has insufficient liquidity for payment transactions
+#[tokio::test(flavor = "multi_thread")]
+async fn test_block_building_insufficient_fee_amm_liquidity() -> eyre::Result<()> {
+    reth_tracing::init_test_tracing();
+
+    let source = crate::utils::NodeSource::LocalNode(
+        include_str!("../assets/test-genesis.json").to_string(),
+    );
+    let (http_url, _local_node) = crate::utils::setup_test_node(source).await?;
+
+    let wallet = MnemonicBuilder::from_phrase(crate::utils::TEST_MNEMONIC)
+        .index(0)?
+        .build()?;
+    let sender_address = wallet.address();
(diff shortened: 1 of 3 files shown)


crates/payload/builder/src/lib.rs
```diff
@@ -333,8 +333,23 @@ where
...
-                // this is an error that we should treat as fatal for this attempt
-                Err(err) => return Err(PayloadBuilderError::evm(err)),
+                Err(BlockExecutionError::Validation(err)) => {
+                    // Other validation errors should fail the block build
+                    warn!(%err, "validation error during transaction execution");
+                    return Err(PayloadBuilderError::other(err));
+                }
+                Err(BlockExecutionError::Internal(err)) => {
+                    // Internal errors include EVM execution errors (e.g., InsufficientLiquidity)
+                    // These should skip the transaction rather than fail the entire block
```
klkvr: hmm I'd expect this to end up in `BlockValidationError::InvalidTx` branch above because this is technically a tx validity condition.

fgimenez: imo you're right that conceptually this is a transaction validity condition. however, the error originates from a precompile revert during EVM execution (fee swap happens during transaction execution in the FeeManager precompile), so it flows through as `InternalBlockExecutionError::EVM` rather than `BlockValidationError::InvalidTx`  

do you think we need additional changes for the error to be correctly classified?

klkvr: can we not map the error to `EVMError::Transaction` here? https://github.com/tempoxyz/tempo/blob/fe819a8746b7f5d4eadd0d0de2d9d3613bc367b1/crates/revm/src/handler.rs#L173-L175

fgimenez: yep makes more sense to have the check there, done here https://github.com/tempoxyz/tempo/pull/365/commits/c08527591431b4b6b2e07cdace3c8b83ab53182b ptal

PR #368, opened 2025-10-01
feat(ci): add bench workflow

Zygimantass:
This workflow can be dispatched by anyone to run a benchmark on a 32 core CI runner. The job will the generate a summary with logs where an engineer can see the gas throughput and max tx throughput.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -1,7 +1,6 @@
 name: bench.yml
 on:
   workflow_dispatch:
-  pull_request:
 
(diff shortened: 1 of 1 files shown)


.github/workflows/bench.yml
```diff
@@ -0,0 +1,81 @@
+name: bench.yml
+on:
+  workflow_dispatch:
+  pull_request:
```
0xKitsune: Do we want this to only run on PRs against main?

PR #367, opened 2025-10-01
refactor: `generalGasLimit`

klkvr:
Relaxes constraints on block ordering, allowing payment transactions to appear at any position, and strictly enforcing all transactions to be payments once `generalGasLimit` is exhausted

PR #372, opened 2025-10-01
feat: `TempoHeader`

klkvr:
Based on https://github.com/tempoxyz/tempo/pull/367

Introduces `TempoHeader` with new `generalGasUsed` field which replaces the custom `extraData` encoding

### DIFF
--- a/crates/primitives/src/header.rs
+++ b/crates/primitives/src/header.rs
@@ -23,8 +23,8 @@ use reth_primitives_traits::{InMemorySize, serde_bincode_compat::RlpBincode};
 )]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
-#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
-#[reth_codecs_derive::add_arbitrary_tests(compact)]
+#[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
+#[cfg_attr(test, reth_codecs::add_arbitrary_tests(compact))]
 pub struct TempoHeader {
     /// Non-payment gas limit for the block.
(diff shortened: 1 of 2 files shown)


crates/primitives/src/header.rs
```diff
@@ -0,0 +1,183 @@
...
+
+impl InMemorySize for TempoHeader {
+    fn size(&self) -> usize {
+        self.inner.size()
+    }
+}
+
+impl Sealable for TempoHeader {
+    fn hash_slow(&self) -> B256 {
+        self.inner.hash_slow()
```
mattsse: this isn't accurate, because this skips the extra fields?

crates/primitives/src/header.rs
```diff
@@ -0,0 +1,183 @@
...
+    Hash,
+    Default,
+    derive_more::Deref,
+    derive_more::DerefMut,
+    RlpEncodable,
+    RlpDecodable,
+)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
+pub struct TempoHeader {
```
mattsse, line 28: can we document rlp encoding for this?

crates/primitives/src/header.rs
klkvr: ah yep, good catch

crates/evm/src/assemble.rs
```diff
@@ -49,38 +49,25 @@ impl BlockAssembler<TempoEvmConfig> for TempoBlockAssembler {
             ..
         } = input;
 
+        let parent = SealedHeader::new_unhashed(parent.clone().into_header().inner);
```
gakonst, line 52: kinda meh we need to clone here?

klkvr: yeah, likely negligible given other overhead around block building. I think we can fix this with a small change on reth side

PR #373, opened 2025-10-01
feat(ci): add manual build with custom profiles

Zygimantass:


.github/workflows/build.yml
```diff
@@ -2,6 +2,17 @@ on:
...
+  workflow_dispatch:
+    inputs:
+      profile:
+        default: "maxperf"
+        description: "Build profile"
+        type: choice
+        options:
+          - "release"
+          - "maxperf"
+          - "debug"
```
shekhirin:
```
❯ cargo b --profile debug
error: profile name `debug` is reserved
To configure the default development profile, use the name `dev` as in [profile.dev]
See https://doc.rust-lang.org/cargo/reference/profiles.html for more on configuring profiles.
```

```suggestion
          - "dev"
```

Zygimantass: mb, forgot

PR #375, opened 2025-10-01
chore(ci): don't use cross and build in dev mode by default

Zygimantass:


.github/workflows/build.yml
```diff
@@ -11,7 +11,6 @@ on:
         options:
           - "release"
           - "maxperf"
-          - "dev"
           - "profiling"
```
DaniPopes, line 15: btw this workflow has no name so it shows up as the full path in the github ui

PR #376, opened 2025-10-02
test(e2e): `tempo-commonware` e2e tests

### LINKED ISSUE #361: Consensus E2E Tests
# E2E Tests for Consensus
We need e2e tests that ensure the network maintains liveness/safety with validator failures, and properly rejects invalid proposals without halting.  

### Test Scenarios

1. Single node crash & recovery
   - Stop one node and restart.  
   - Assert it re-syncs to tip and resumes producing/notarizing blocks.  

2. 1/3rd validators crash
   - Stop 1/3 of nodes. 
   - Assert block production continues as normal

3. 2/3rds validators crash
   - Stop 2/3rds of nodes.
   - Assert block production stalls.  
   - On restart, network resumes from the last finalized block.  

4. All validators crash & recover  
   - Stop all nodes, then restart.  
   - Assert the network recovers and resumes from the last finalized block.  

5. Invalid / malformed proposals
   - Inject invalid/malfored proposal
   - Assert nodes reject them without halting the network.  

6. Tx propagation
   - NodeA receives txs 0..`n`
   - NodeB receives txs `n`..`m`
   - NodeC etc.
   - Assert that the block produced contains all txs

0xKitsune:
Closes #361 

This PR introduces e2e tests for `tempo-commonware` node/consensus. Each test spins up 4 peered validators, starts a round-robin transaction generator, and verifies block production/recovery behavior under various failure modes.

#### The following tests are included in this PR:

- `test-partial-network-failure.sh`  
  Single/partial node crash and recovery. This test starts the network and tx generator, verifies blocks, stops one validator, confirms block production continues, restarts the validator, and verifies continued production.

- `test-network-halt-and-recovery.sh`  
  Majority validator halt (2/4 nodes) and recovery. This test starts the network and tx generator, verifies block production, stops two validators to force a chain halt, verifies no block progress, restarts the failed nodes, waits for recovery, and verifies block production resumes from the last finalized state.

- `test-full-network-failure-and-recovery.sh`  
  Full network crash and recovery (4/4 nodes). This test the network and tx generator, verifies progress, stops all validators, confirms the network is unreachable, restarts all, waits for recovery, and verifies block production resumes.
  
  
  
  
 #### Additional scripts to start/stop the node and send txs are also included in this PR:
  
  - `scripts/consensus/start-network.sh`  
  Starts a 4 validator network with Docker.

- `scripts/consensus/stop-network.sh`  
  Stops and cleans up the validator containers and the Docker network.

- `scripts/consensus/tx-generator.sh`  
  TX generator that uses `cast` to send 0 value transfers and confirm tx success. This script rotates nodes to target, ensuring all validator RPCs are used.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,26 +43,6 @@ jobs:
         env:
           RUSTFLAGS: -D warnings
 
-  consensus-e2e:
-    name: consensus e2e tests
-    runs-on: depot-ubuntu-24.04-4
-    timeout-minutes: 45
-    steps:
-      - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@stable
-      - uses: rui314/setup-mold@v1
-      - uses: mozilla-actions/sccache-action@v0.0.9
-      - name: Build tempo image
-        run: docker build -t tempo-commonware:latest .
-      - name: Run consensus network tests
-        run: |
-          cd scripts/consensus
-          ./test-partial-network-failure.sh
-          ./test-network-halt-and-recovery.sh
-          ./test-full-network-failure-and-recovery.sh
-        env:
-          RUST_LOG: debug
-
   test-success:
     name: test success
     runs-on: depot-ubuntu-24.04-4
@@ -70,7 +50,6 @@ jobs:
     needs:
       - test
       - msrv
-      - consensus-e2e
     timeout-minutes: 30
     steps:
       - name: Decide whether the needed jobs succeeded or failed


Dockerfile
```diff
@@ -0,0 +1,41 @@
+FROM rust:1.83-slim-bookworm AS builder
```
SuperFluffy: Let's use 1.88 here (since that's our MSRV).

Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+# Copy workspace files
+COPY Cargo.toml Cargo.lock ./
+COPY bin/ ./bin/
+COPY crates/ ./crates/
+
+# NOTE: Remove xtask from workspace to avoid missing dependency error for now
+RUN sed -i '/xtask/d' Cargo.toml
+
+# Install nightly Rust and build the tempo binary
+RUN rustup toolchain install nightly && rustup default nightly
```
SuperFluffy, line 23: If we use an image with 1.88 preinstalled we can avoid this (and have a bit of an indicator if we bump msrv but forget to bump this containerfile).

Dockerfile
SuperFluffy, line 1: Pet peeve of mine and bikeshed (feel free to ignore): I like to call this `Containerfile` instead. :-)

scripts/consensus/configs/196506320dcb5ea6bf4a5404dae21d6119341e7248e8f7f6dfb14d12301befe2.toml
SuperFluffy, line 1: These are the names provided by the xtask tool, but we should probably name these similarly to how you address them in the tests (i.e. `validator-0` and so on).

I made this for myself so that we can keep the signers, shares, polynomials and hence names stable: https://github.com/tempoxyz/tempo/issues/378

scripts/consensus/start-network.sh
```diff
@@ -0,0 +1,78 @@
...
+
+  echo "  ✓ Started $container_name on port $rpc_port"
+}
+
+# Start all validators
+for i in {0..3}; do
+  start_validator "$i" "${CONFIGS[$i]}"
+done
+
+# Hacky fix: restart validator-0 if it failed
```
SuperFluffy, line 66: This fails because it not the bootstrapper I think (I might be mistaken). Would this be fixed if you started `"1bc6dd47289a73e859502db4233ebcc36976cedf33f4a45635f7cd6a983f9a7b.toml"` first?

scripts/consensus/test-partial-network-failure.sh
```diff
@@ -0,0 +1,81 @@
...
+  start_validator 2
+  echo ""
+
+  # Wait for validator to rejoin
+  echo "Waiting 3 seconds for validator to rejoin..."
+  sleep 3
+  echo ""
+
+  # Check blocks still being produced
+  echo "Checking block production after validator recovery..."
```
SuperFluffy, line 58: I think this test does not yet show if we managed to recover since the network kept producing blocks without validator 2!

But you can kill validator 3 and *then* check block production to ensure recovery:

1. kill validator 2
2. observe block production
3. restart validator 2
4. observe block production
5. kill validator 3
6. observe block production

This way you know that validator 2 successfully joined && synced up to tip && started validating/producing blocks.

0xKitsune: > I think this test does not yet show if we managed to recover since the network kept producing blocks without validator 2

Yes this test is specifically checking that blocks continue to be produced if less than 2f+1 (out of 3f+1) validators crash. 

There is a separate test called `test-network-halt-and-recovery.sh` that asserts that block production stops when >= 2f+1 nodes are down and that block production resumes when validators restart.

SuperFluffy: Right, that makes sense. But then validator 2 restarting is redundant I think. After all, the network kept going, so all is fine. Validator 2 restarting does not add anything - after all, it does not matter if it joined successfully or not given that the other validators kept on chugging along.

But I agree: with `test-network-halt-and-recovery.sh` it's probably not necessary to demonstrate that validators 2 and 3 have "swapped" places.

scripts/consensus/start-network.sh
0xKitsune: It looks like I still run into issues when changing the sequence started.

Dockerfile
0xKitsune: Updated to `rust:1.88-slim-bookworm`

PR #377, opened 2025-10-02
docs: use legacy state root in tempo-bench

shekhirin:
Currently, `--engine.legacy-state-root` performs better than our default state root algorithm (Sparse Trie), so we should use it instead.

### DIFF
--- a/bin/tempo-bench/README.md
+++ b/bin/tempo-bench/README.md
@@ -201,7 +201,7 @@ tempo-bench generate-genesis --accounts 50000 --output genesis.json
 ### 3. Run max TPS benchmark
 
 ```bash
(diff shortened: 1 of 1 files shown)


bin/tempo-bench/README.md
```diff
@@ -219,7 +220,7 @@ rmdir /s "%APPDATA%\reth"
 
 
 
-### Sampling 
+### Sampling
 Use the following commands to run the node with [sampling](https://github.com/mstange/samply):
 ```bash
 	samply record --output tempo.samply -- tempo node \
```
0xKitsune, line 226: Can we also add `--engine.legacy-state-root` to this command?

PR #380, opened 2025-10-02
fix: `FeeManager` fixes

klkvr:
Fixes several bugs in the current fee logic:
1. For `user_token == validator_token` we were actually never paying the validator any fees. Also added test for this.
2. Validators should not be able to change their fee token preference in-between the block they are building.
3. Blacklisted users should not be able to spend their funds on fees. Blaclisted validators should not be able to receive any fees.

### DIFF
--- a/crates/precompiles/src/contracts/tip_fee_manager/mod.rs
+++ b/crates/precompiles/src/contracts/tip_fee_manager/mod.rs
@@ -613,6 +613,11 @@ mod tests {
         let token = token_id_to_address(rand::random::<u64>());
 
         let call = IFeeManager::setValidatorTokenCall { token };
+        let result = fee_manager.set_validator_token(&validator, call.clone());
+        assert_eq!(result, Err(FeeManagerError::cannot_change_within_block()));
+
+        // Now set beneficiary to a random address to avoid `CannotChangeWithinBlock` error
+        fee_manager.beneficiary = Address::random();
         let result = fee_manager.set_validator_token(&validator, call);
         assert!(result.is_ok());
 
@@ -645,9 +650,12 @@ mod tests {
         let mut fee_manager = TipFeeManager::new(TIP_FEE_MANAGER_ADDRESS, validator, &mut storage);
 
         // Set validator token
+        // Set beneficiary to a random address to avoid `CannotChangeWithinBlock` error
+        fee_manager.beneficiary = Address::random();
         fee_manager
             .set_validator_token(&validator, IFeeManager::setValidatorTokenCall { token })
             .unwrap();
+        fee_manager.beneficiary = validator;
 
         // Set user token
         fee_manager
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/contracts/tip_fee_manager/mod.rs
```diff
@@ -136,6 +136,10 @@ impl<'a, S: StorageProvider> TipFeeManager<'a, S> {
             return Err(FeeManagerError::invalid_token());
         }
 
+        if sender == &self.beneficiary {
+            return Err(FeeManagerError::cannot_change_within_block());
+        }
```
0xKitsune, line 141: Noting that at the moment (both in the impl and the spec) it looks like anyone can call this function to set the validator token for a given address.

crates/precompiles/src/contracts/tip_fee_manager/mod.rs
```diff
@@ -291,10 +304,10 @@ impl<'a, S: StorageProvider> TipFeeManager<'a, S> {
         let validator_token = self.get_validator_token();
```
0xKitsune, line 301: I believe this should use `U256::ZERO` instead. The variable naming may have been a bit confusing before. This logic intends to get all pending fees collected during transaction execution and then swap those fees into the validator token.

With this change, instead of only considering fees from the transactions in the current block, we now load the total historically collected fees.

crates/precompiles/src/contracts/tip_fee_manager/mod.rs
klkvr: > for a given address

wdym? this function sets token for the `msg.sender`, right?

crates/precompiles/src/contracts/tip_fee_manager/mod.rs
klkvr: value in `collected_fees_slot()` is cleared in the end of each `executeBlock` call, isn't it? before this PR, the value was also never read

right now fees collected in the same user token as validator token are only recorded in this slot:
https://github.com/tempoxyz/tempo/blob/8dc66fa16b23ab72ca0c92d48e163be2205e6378/crates/precompiles/src/contracts/tip_fee_manager/mod.rs#L266-L268

If I change the `collected_fees` initial value to zero, the `test_set_user_token` test will start to fail because it will only be incremented with fees charged in tokens other than `validator_token`

crates/precompiles/src/contracts/tip_fee_manager/mod.rs
0xKitsune: Ahh yes disregard, got wires crossed.

crates/precompiles/src/contracts/tip_fee_manager/mod.rs
0xKitsune: Yes, my mistake! I had a previous version of the spec  in mind.

PR #390, opened 2025-10-02
feat: introduce TempoInvalidTransaction

### LINKED ISSUE #369: Introduce `TempoInvalidTransaction`
We now have many Tempo-specific checks during transaction validation, e.g
https://github.com/tempoxyz/tempo/blob/d9d5620baa4b338648237bd39e26b4d6bee73116/crates/evm/src/evm.rs#L176-L178
https://github.com/tempoxyz/tempo/blob/d9d5620baa4b338648237bd39e26b4d6bee73116/crates/revm/src/handler.rs#L181-L187

We have an opportunity to clean this up by introducing a custom `TempoInvalidTransaction` error enum, and using it as second generic here 
https://github.com/tempoxyz/tempo/blob/d9d5620baa4b338648237bd39e26b4d6bee73116/crates/evm/src/evm.rs#L136

fgimenez:
Closes #369

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,7 +1,5 @@
 name: Build binaries
 
-# TODO: Vercel requires that the repo is either open sourced or contributors have deployment permissions.
-# This should be uncommented when the repo is public.
 on:
   # push:
   #   branches: [main]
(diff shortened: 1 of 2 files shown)


crates/revm/src/error.rs
```diff
@@ -0,0 +1,103 @@
...
+
+/// Tempo-specific invalid transaction errors.
+///
+/// This enum extends the standard Ethereum [`InvalidTransaction`] with Tempo-specific
+/// validation errors that occur during transaction processing.
+#[derive(Debug, Clone, PartialEq, Eq, Hash, thiserror::Error)]
+pub enum TempoInvalidTransaction {
+    /// Standard Ethereum transaction validation error.
+    #[error(transparent)]
+    Ethereum(#[from] InvalidTransaction),
```
0xKitsune: nit: very minor but consider renaming to `EthInvalidTransaction` for consistency with `TempoInvalidTransaction` naming.

crates/evm/src/evm.rs
```diff
@@ -154,7 +152,7 @@ where
         if tx.is_system_tx {
             let TxKind::Call(to) = tx.inner.kind else {
                 return Err(EVMError::Transaction(
-                    InvalidTransaction::BlobCreateTransaction,
+                    TempoInvalidTransaction::SystemTransactionMustBeCall,
                 ));
```
0xKitsune: nit: Consider adding a from impl to convert from `TempoInvalidTransaction`, Then you can use `into()` to simplify slightly.

crates/revm/src/handler.rs
```diff
@@ -129,9 +129,12 @@ where
...
             is_eip3607_disabled,
             is_nonce_check_disabled,
-        )?;
+        )
+        .map_err(|e| EVMError::Transaction(TempoInvalidTransaction::Ethereum(e)))?;
 
-        let max_balance_spending = tx.max_balance_spending()?;
+        let max_balance_spending = tx
+            .max_balance_spending()
+            .map_err(|e| EVMError::Transaction(TempoInvalidTransaction::Ethereum(e)))?;
```
```suggestion
            .map_err(TempoInvalidTransaction::Ethereum)?;
```

crates/revm/src/error.rs
klkvr: this is revm's type so we can't do that

crates/revm/src/handler.rs
klkvr: ~oh actually this does not work because `EVMError` doesn't implement this conversion bc of a typesystem limitation :/~ it does work if you also do https://github.com/tempoxyz/tempo/pull/390#discussion_r2400264894

crates/revm/src/handler.rs
```diff
@@ -129,9 +129,12 @@ where
             tx.nonce(),
             is_eip3607_disabled,
             is_nonce_check_disabled,
-        )?;
+        )
+        .map_err(|e| EVMError::Transaction(TempoInvalidTransaction::Ethereum(e)))?;
```
```suggestion
        .map_err(TempoInvalidTransaction::Ethereum)?;
```

crates/evm/src/evm.rs
fgimenez: good point, done

crates/revm/src/error.rs
fgimenez: i believe it is about renaming the variant itself, not the revm type, done here https://github.com/tempoxyz/tempo/pull/390/commits/acf9782301a59b5c1ae295bf59c4aa02fc893354

PR #408, opened 2025-10-05
fix(node): forward proposal ancestors if missing

SuperFluffy:
This patch addresses an edge case where notarized (but not yet finalized) blocks are not forwarded to the execution layer, stopping the chain from proposing or validating new blocks.

This fix consists in the execution driver always attempting to backfill the parent block of a given proposal or digest. It does so by starting at the parent and walking its ancestors all the way until reaching the latest block available on the execution layer.

As part of the patch backfilling at node startup and during steady state operation was unified.

This happens under the following scenario:

1. all validators are at a finalized height H and notarized height H+1
2. all validators' execution layers have finalized = H, head = H+1
3. the validators have not witnessed a quorum of finalization receipts
4. all validators are stopped at once
5. upon restart, the validators' consensus layer is caught up to finalized H and notarized H+1
6. the execution layers are at finalized = head = H
7. the EL has not persisted H+1 since it only persists up to the finalized head
8. the consensus layers attempt to transition to round H+2 with parent H+1
9. none of the validators can propose nor verify
10. simplex does not guarantee a direct finalization of H+1, but also no indirect finalization occurs because the chain does not propose new blocks
11. the execution layer will never see block H+1
12. the network is stalled

### DIFF
--- a/crates/commonware-node/src/consensus/execution_driver/executor.rs
+++ b/crates/commonware-node/src/consensus/execution_driver/executor.rs
@@ -1,8 +1,11 @@
-//! Drives the actual execution by sending finalized blocks and forkchoice updates.
+//! Drives the actual execution forwarding blocks and setting forkchoice state.
 //!
 //! This agent forwards finalized blocks from the consensus layer to the
 //! execution layer and tracks the digest of the latest finalized block.
 //! It also advances the canonical chain by sending forkchoice-updates.
+//!
+//! If the agent detects that the execution layer is missing blocks it attempts
+//! to backfill them from the consensus layer.
 
 use std::sync::Arc;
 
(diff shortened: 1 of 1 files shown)


crates/commonware-node/src/consensus/execution_driver/executor.rs
```diff
@@ -22,34 +26,51 @@ pub(super) struct Builder {
...
         } = self;
 
         let (to_me, from_execution_driver) = futures_channel::mpsc::unbounded();
 
         let my_mailbox = ExecutorMailbox { inner: to_me };
 
-        // XXX: this ensures that the initial forkchoice-state with
-        // head = safe = finalized = latest_finalized is the first thing that
-        // the agent sends to the execution layer.
+        // XXX: canonicalizing the latest finalizad digest && starting a backfill
```
0xKitsune:
```suggestion
        // XXX: canonicalizing the latest finalized digest && starting a backfill
```

crates/commonware-node/src/consensus/execution_driver/executor.rs
```diff
@@ -200,62 +317,75 @@ pub(super) struct ExecutorMailbox {
...
+    },
+    Canonicalize {
+        round: Option<Round>,
+        digest: Digest,
+    },
     /// Commands the canonical chain engine to forward a finalized block to
     /// the execution layer.
     ///
     /// The response channel is set if the finalized block is new (i.e. at the
     /// tip of the finalized chain).
```
0xKitsune: nit: this comment looks outdated

PR #411, opened 2025-10-06
fix(commonware-node): don't block executor on backfill

SuperFluffy:
https://github.com/tempoxyz/tempo/pull/408 introduced a backfill mechanism that ensures that notarized blocks eventually get forwarded from the consensus layer to the execution layer. However, it was done in such a way that subscription to the marshaller blocked the event loop of the execution layer, because it was stalling on every single parent block.

This patch pushes the subscription + forward to the latest block onto a queue so that the event loop is no longer stalled.

Note that there is a chance of seeing double-backfills now. This likely will not cause issues but should be replaced by some sort of backfill coalescing.

### DIFF
--- a/crates/commonware-node/src/consensus/execution_driver/executor.rs
+++ b/crates/commonware-node/src/consensus/execution_driver/executor.rs
@@ -167,7 +167,6 @@ impl Executor {
         ),
     )]
     fn backfill(&mut self, cause: Span, round: Option<Round>, digest: Digest) {
-        info!("scheduling backfill");
         is_send(
             backfill_from_consensus(
                 cause,


crates/commonware-node/src/consensus/execution_driver/executor.rs
```diff
@@ -113,20 +117,30 @@ impl Executor {
...
+                    // happen sequentially, so blocking the event loop on await
+                    // is desired.
+                    //
+                    // Backfills will be put on a queue and can happen anytime
+                    // in between.
+                    self.handle_message(msg).await;
+                }
+
+                // Only exists to drain the queue
+                _backfill = self.running_backfills.next() => {}
```
0xKitsune, line 135: Is there a reason we do not just spawn a background task to perpetually process backfilling blocks?

crates/commonware-node/src/consensus/execution_driver/executor.rs
```diff
@@ -140,59 +154,41 @@ impl Executor {
...
+                self.marshal.clone(),
+                round,
+            )
+            .boxed(),
+        );
+        self.running_backfills
+            .push(Box::pin(backfill_from_consensus(
+                Span::current(),
+                digest,
+                self.execution_node.clone(),
```
0xKitsune, line 186: Do we need to clone the entire execution node here?

crates/commonware-node/src/consensus/execution_driver/executor.rs
SuperFluffy: Yes: the background task would be another agent with it's mailbox etc. So more boilerplate. I started with that but went with this solution to get it out quicker.

It can't proactively check for missing blocks - it needs to be told.

crates/commonware-node/src/consensus/execution_driver/executor.rs
SuperFluffy: As for the other comment, I wanted to get this out quickly and didn't want to chase down the specific types I need (PTSD from working with nodes earlier).

Also, the clone here isn't so bad and also not in a perf critical section.

PR #412, opened 2025-10-06
feat(commonware-node): allow running without interrupt

SuperFluffy:
The interrupts introduced in https://github.com/tempoxyz/tempo/pull/348 did not consider the special case of `no interrupt time`.

To make this explicit, this patch allows not setting any interrupt. The case `0ms` will still result in interrupting immediately, but this change makes the behavior a bit more explicit.

NOTE: also changes the name.

### DIFF
--- a/crates/commonware-node/src/consensus/execution_driver/executor.rs
+++ b/crates/commonware-node/src/consensus/execution_driver/executor.rs
@@ -7,7 +7,7 @@
 //! If the agent detects that the execution layer is missing blocks it attempts
 //! to backfill them from the consensus layer.
 
-use std::sync::Arc;
+use std::{pin::Pin, sync::Arc};
 
 use alloy_rpc_types_engine::{ForkchoiceState, PayloadStatus};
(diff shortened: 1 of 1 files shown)


crates/commonware-node/src/consensus/execution_driver/mod.rs
```diff
@@ -468,13 +469,19 @@ impl Inner<Init> {
...
-            timeout_ms = self.new_payload_wait_time.as_millis(),
-            "sleeping for payload builder timeout"
-        );
-        context.sleep(self.new_payload_wait_time).await;
-
-        interrupt_handle.interrupt();
+        if self.new_payload_wait_time.is_zero() {
+            info!(
+                "interrupt time was set to zero; will nlet the payload build \
+                run until completion (this might not finish in time)"
```
0xKitsune, line 475: nit: typo

```suggestion
                "interrupt time was set to zero; will let the payload build
                run until completion (this might not finish in time)"
```

PR #413, opened 2025-10-06
feat: TIP20 linking token

0xKitsune:
This PR introduces linking tokens, as introduced in tempoxyz/specs#30 to enable stablecoin DEX pairs and provide a canonical unit for asset linking. 

Note that this PR introduces `LinkingUSD`, the default linking token and assigns it the address `0x20c0000000000000000000000000000000000000`, effectively deploying the first TIP20 (tokenId 0) as `LinkingUSD`. The TIP20 factory has been updated so new tokens now begin deployment from `tokenId` 1 onward, reserving the initial token slot for the canonical linking token.

### DIFF
--- a/crates/precompiles/src/contracts/linking_usd.rs
+++ b/crates/precompiles/src/contracts/linking_usd.rs
@@ -71,6 +71,14 @@ impl<'a, S: StorageProvider> LinkingUSD<'a, S> {
         self.token.symbol()
     }
 
+    pub fn currency(&mut self) -> String {
+        self.token.currency()
+    }
+
+    pub fn linking_token(&mut self) -> Address {
+        self.token.linking_token()
+    }
+
     pub fn decimals(&mut self) -> u8 {
         self.token.decimals()
     }
--- a/crates/precompiles/src/precompiles/linking_usd.rs
+++ b/crates/precompiles/src/precompiles/linking_usd.rs
@@ -26,6 +26,10 @@ impl<S: StorageProvider> Precompile for LinkingUSD<'_, S> {
             ITIP20::totalSupplyCall::SELECTOR => {
                 metadata::<ITIP20::totalSupplyCall>(self.total_supply())
             }
+            ITIP20::currencyCall::SELECTOR => metadata::<ITIP20::currencyCall>(self.currency()),
+            ITIP20::linkingTokenCall::SELECTOR => {
+                metadata::<ITIP20::linkingTokenCall>(self.linking_token())
+            }
 
             // View functions
             ITIP20::balanceOfCall::SELECTOR => {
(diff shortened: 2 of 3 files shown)


crates/precompiles/src/contracts/linking_usd.rs
```diff
@@ -0,0 +1,152 @@
...
+
+    // Delegate all other methods to the underlying TIP20Token
+    pub fn name(&mut self) -> String {
+        self.token.name()
+    }
+
+    pub fn symbol(&mut self) -> String {
+        self.token.symbol()
+    }
+
```
fgimenez, line 73: nit: metadata fns impls missing 
```suggestion

    pub fn currency(&mut self) -> String {
        self.token.currency()
    }

    pub fn linking_token(&mut self) -> Address {
        self.token.linking_token()
    }

```

crates/precompiles/src/precompiles/linking_usd.rs
```diff
@@ -0,0 +1,77 @@
...
+
+        match selector {
+            // Metadata
+            ITIP20::nameCall::SELECTOR => metadata::<ITIP20::nameCall>(self.name()),
+            ITIP20::symbolCall::SELECTOR => metadata::<ITIP20::symbolCall>(self.symbol()),
+            ITIP20::decimalsCall::SELECTOR => metadata::<ITIP20::decimalsCall>(self.decimals()),
+            ITIP20::totalSupplyCall::SELECTOR => {
+                metadata::<ITIP20::totalSupplyCall>(self.total_supply())
+            }
+
```
fgimenez, line 33: nit missing metadata fns selectors
```suggestion
            ITIP20::currencyCall::SELECTOR => metadata::<ITIP20::currencyCall>(self.currency()),
            ITIP20::linkingTokenCall::SELECTOR => {
                metadata::<ITIP20::linkingTokenCall>(self.linking_token())
            }

```

crates/precompiles/src/contracts/linking_usd.rs
```diff
@@ -0,0 +1,152 @@
...
+    pub fn transfer_with_memo(
+        &mut self,
+        _to: Address,
+        _amount: U256,
+        _memo: [u8; 32],
+    ) -> Result<(), TIP20Error> {
+        Err(TIP20Error::transfers_disabled())
+    }
+
+    pub fn transfer_from_with_memo(
```
fgimenez, line 55: A selector is missing for `transfer_from_with_memo` but it seems there's an inconsistency in the interface definition, here and in the spec it returns a `bool`, but the interface doesn't specify a return type https://github.com/tempoxyz/tempo/blob/c941f54f05d2cf5c3988546386f6f0e57e09014b/crates/contracts/src/precompiles.rs#L51

0xKitsune: Thanks for flagging, updated this to return `bool`.

PR #424, opened 2025-10-07
feat(precompiles): ensure linking token is deployed when creating TIP20 tokens

0xKitsune:
This PR updates the `createToken()` function on the `TIP20Factory` to ensure the linking token specified is currently deployed.

### DIFF
--- a/crates/precompiles/src/precompiles/tip20_factory.rs
+++ b/crates/precompiles/src/precompiles/tip20_factory.rs
@@ -130,12 +130,12 @@ mod tests {
         let result = factory.call(&Bytes::from(calldata), &sender);
         expect_precompile_error(&result, TIP20Error::invalid_currency());
 
-        // Check counter increased again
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/contracts/tip20_factory.rs
```diff
@@ -50,30 +50,24 @@ impl<'a, S: StorageProvider> TIP20Factory<'a, S> {
...
 
-        // increase the token counter
-        self.storage
-            .sstore(
-                TIP20_FACTORY_ADDRESS,
-                slots::TOKEN_ID_COUNTER,
-                token_id + U256::ONE,
-            )
-            .expect("TODO: handle error");
+        if !is_tip20(&call.linkingToken)
```
fgimenez, line 59: nit: maybe a comment here would help to make it clear "token already deployed == its token_id must be <= current counter"

PR #426, opened 2025-10-07
feat(precompiles): implement Order type for stablecoin DEX

### LINKED ISSUE #419: Define `Order` type
Define the stablecoin DEX `Order` type. Each order should contain tick price, quantity and a bool to indicate if the order is a flip order as well as any other necessary metadata.

fgimenez:
Closes #419

Implements Order type for stablecoin DEX with tick price, quantity, flip order support, and all metadata required for price-time priority matching.

### DIFF
--- a/crates/precompiles/src/contracts/stablecoin_dex/order.rs
+++ b/crates/precompiles/src/contracts/stablecoin_dex/order.rs
@@ -1,6 +1,6 @@
 //! Limit order type for the stablecoin DEX.
 //!
-//! This module defines the core `LimitOrder` type used in the stablecoin DEX orderbook.
+//! This module defines the core `Order` type used in the stablecoin DEX orderbook.
 //! Orders support price-time priority matching, partial fills, and flip orders that
 //! automatically place opposite-side orders when filled.
 
@@ -8,12 +8,12 @@ use alloy::primitives::Address;
 
 use super::error::OrderError;
 
-/// Represents a limit order in the stablecoin DEX orderbook.
+/// Represents an order in the stablecoin DEX orderbook.
 ///
 /// # Order Types
-/// - **Bid**: Order to buy the token using its linking token
-/// - **Ask**: Order to sell the token for its linking token
-/// - **Flip**: Order that automatically creates a new order on the opposite side when fully filled
+/// - **Regular orders**: Orders with `flip_tick = None`
+/// - **Flip orders**: Orders with `flip_tick = Some(tick)` that automatically create
+///   a new order on the opposite side when fully filled
 ///
 /// # Order Lifecycle
 /// 1. Order is placed via `place()` or `placeFlip()` and added to pending queue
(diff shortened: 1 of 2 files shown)


crates/precompiles/src/contracts/stablecoin_dex/order.rs
```diff
@@ -0,0 +1,664 @@
...
+/// # Price-Time Priority
+/// Orders are sorted by:
+/// 1. Price (tick)
+/// 2. Block number (when inserted)
+/// 3. Order index within block
+///
+/// Special case: Orders that flipped within a block are added to the
+/// book before other orders that were placed within the block.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Order {
```
0xKitsune, line 37: What do you think about using an enum here for the precompile implementation. Something like:

```rust
pub enum LimitOrder {
    Bid {
        ...
    },

    Ask {
        ...
    },

    Flip {
        ...
    }
}
```

And then we can create corresponding solidity structs for each `Bid`, `Ask`, `Flip` order types that we can keep in EVM storage.

fgimenez: sure, good point, pushed changes for it ptal. we have duplication in the variants, do you think we could use a `LimitOrderCore` type for the common fields?

crates/precompiles/src/contracts/stablecoin_dex/order.rs
```diff
@@ -0,0 +1,1037 @@
...
+        /// Linking token (quote currency) for this pair
+        linking_token: Address,
+        /// Remaining amount of the token to be filled
+        amount: u128,
+        /// Original amount when the order was placed
+        original_amount: u128,
+        /// Price tick: (price - 1) * 1000
+        tick: i16,
+        /// Block number when inserted into orderbook
+        block_number: u64,
```
0xKitsune: We can probably remove `block_number` since the order is already established given there is a `prev_order_id` and `next_order_id`.

crates/precompiles/src/contracts/stablecoin_dex/order.rs
```diff
@@ -0,0 +1,1037 @@
...
+///
+/// # Onchain Storage
+/// Orders are stored onchain in doubly linked lists organized by tick.
+/// Each tick maintains a FIFO queue of orders using `prev_order_id` and `next_order_id` pointers.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum LimitOrder {
+    /// Bid order: buying the token using its linking token
+    Bid {
+        /// Unique identifier for this order
+        order_id: u128,
```
0xKitsune: since `order_id` will be used as the key in storage, do we need to add this to the struct?

crates/precompiles/src/contracts/stablecoin_dex/order.rs
```diff
@@ -0,0 +1,1037 @@
...
+        /// Position within block
+        order_index: u64,
+        /// Previous order ID in the doubly linked list (0 if head)
+        prev_order_id: u128,
+        /// Next order ID in the doubly linked list (0 if tail)
+        next_order_id: u128,
+    },
+
+    /// Flip order: automatically creates a new order on the opposite side when fully filled
+    Flip {
```
0xKitsune: With the recent updates to the orderbook/order structure, we might want to simplify by removing the `Flip` variant and add `flip` related fields to the order struct.

crates/precompiles/src/contracts/stablecoin_dex/order.rs
```diff
@@ -0,0 +1,1037 @@
...
+        tick: i16,
+        /// Block number when inserted into orderbook
+        block_number: u64,
+        /// Position within block
+        order_index: u64,
+        /// Previous order ID in the doubly linked list (0 if head)
+        prev_order_id: u128,
+        /// Next order ID in the doubly linked list (0 if tail)
+        next_order_id: u128,
+    },
```
0xKitsune: Now that we are moving the orderbook/orders onchain, we can probably unify these structs to a single `Order` type again since bids and asks share much of the same fields.

crates/precompiles/src/contracts/stablecoin_dex/order.rs
fgimenez: good call removed

crates/precompiles/src/contracts/stablecoin_dex/order.rs
fgimenez: nope, removed with `block_number`

crates/precompiles/src/contracts/stablecoin_dex/order.rs
fgimenez: makes sense, reverted to the `Order` struct

crates/precompiles/src/contracts/stablecoin_dex/order.rs
fgimenez: yep, done

PR #430, opened 2025-10-07
feat(precompiles): implement two-step linking token update

fgimenez:
ref https://github.com/tempoxyz/specs/pull/30#discussion_r2407687981

Implements two-step linking token update mechanism for TIP20 tokens to enable seamless liquidity transitions.

  - added `nextLinkingToken` storage slot and view function
  - implemented `setNextLinkingToken()` - sets the next linking token with validation
  - implemented `completeLinkingTokenUpdate()` - finalizes the update with loop detection
  - added `NextLinkingTokenSet` and `LinkingTokenUpdate` events
  - initialize `nextLinkingToken` to same value as `linkingToken` on token creation

### DIFF
--- a/crates/contracts/src/precompiles.rs
+++ b/crates/contracts/src/precompiles.rs
@@ -81,8 +81,8 @@ sol! {
         event TransferPolicyUpdate(address indexed updater, uint64 indexed newPolicyId);
         event SupplyCapUpdate(address indexed updater, uint256 indexed newSupplyCap);
         event PauseStateUpdate(address indexed updater, bool isPaused);
-        event NextLinkingTokenSet(address indexed updater, address indexed newLinkingToken);
-        event LinkingTokenUpdate(address indexed updater, address indexed newLinkingToken);
+        event UpdateLinkingToken(address indexed updater, address indexed newLinkingToken);
+        event LinkingTokenUpdateFinalized(address indexed updater, address indexed newLinkingToken);
 
         // Errors
         error InsufficientBalance();
--- a/crates/precompiles/src/contracts/tip20.rs
+++ b/crates/precompiles/src/contracts/tip20.rs
@@ -285,7 +285,7 @@ impl<'a, S: StorageProvider> TIP20Token<'a, S> {
         self.storage
             .emit_event(
                 self.token_address,
-                TIP20Event::NextLinkingTokenSet(ITIP20::NextLinkingTokenSet {
+                TIP20Event::UpdateLinkingToken(ITIP20::UpdateLinkingToken {
                     updater: *msg_sender,
                     newLinkingToken: call.newLinkingToken,
                 })
@@ -328,7 +328,7 @@ impl<'a, S: StorageProvider> TIP20Token<'a, S> {
         self.storage
             .emit_event(
                 self.token_address,
-                TIP20Event::LinkingTokenUpdate(ITIP20::LinkingTokenUpdate {
+                TIP20Event::LinkingTokenUpdateFinalized(ITIP20::LinkingTokenUpdateFinalized {
                     updater: *msg_sender,
                     newLinkingToken: next_linking_token,
                 })
@@ -1732,7 +1732,7 @@ mod tests {
         let events = &storage.events[&token_id_to_address(token_id)];
         assert_eq!(
             events.last().unwrap(),
-            &TIP20Event::NextLinkingTokenSet(ITIP20::NextLinkingTokenSet {
+            &TIP20Event::UpdateLinkingToken(ITIP20::UpdateLinkingToken {
                 updater: admin,
                 newLinkingToken: linking_token_address,
             })
@@ -1906,7 +1906,7 @@ mod tests {
         let events = &storage.events[&token_id_to_address(token_id)];
         assert_eq!(
             events.last().unwrap(),
-            &TIP20Event::LinkingTokenUpdate(ITIP20::LinkingTokenUpdate {
+            &TIP20Event::LinkingTokenUpdateFinalized(ITIP20::LinkingTokenUpdateFinalized {
                 updater: admin,
                 newLinkingToken: linking_token_address,
             })


crates/contracts/src/precompiles.rs
```diff
@@ -56,6 +57,8 @@ sol! {
         function setSupplyCap(uint256 newSupplyCap) external;
         function pause() external;
         function unpause() external;
+        function setNextLinkingToken(address newLinkingToken) external;
+        function completeLinkingTokenUpdate() external;
```
0xKitsune: nit: What do you think about updating these to `updateLinkingToken()` and `finalizeLinkingTokenUpdate()`. We could also update the spec's interface accordingly?

crates/precompiles/src/contracts/tip20.rs
```diff
@@ -244,6 +252,105 @@ impl<'a, S: StorageProvider> TIP20Token<'a, S> {
...
+        self.check_role(msg_sender, DEFAULT_ADMIN_ROLE)?;
+
+        let next_linking_token = self.next_linking_token();
+
+        // Check that this does not create a loop
+        // Loop through linking tokens until we reach the root (LinkingUSD at 0x20C0...)
+        const LINKING_USD_ROOT: Address = Address::new([
+            0x20, 0xC0, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+            0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+        ]);
```
0xKitsune: Lets use the `LINKING_USD_ADDRESS` const here.

crates/precompiles/src/contracts/tip20.rs
```diff
@@ -244,6 +252,105 @@ impl<'a, S: StorageProvider> TIP20Token<'a, S> {
...
+
+        let mut current = next_linking_token;
+        while current != LINKING_USD_ROOT && current != Address::ZERO {
+            // If we encounter this token in the chain, there's a loop
+            if current == self.token_address {
+                return Err(TIP20Error::invalid_linking_token());
+            }
+
+            // Get the linking token of the current token
+            if address_is_token_address(&current) {
```
0xKitsune: We already know that all linking tokens are already tip20 tokens, I think we can remove this check.

crates/precompiles/src/contracts/tip20.rs
```diff
@@ -244,6 +252,105 @@ impl<'a, S: StorageProvider> TIP20Token<'a, S> {
...
+
+        // Check that this does not create a loop
+        // Loop through linking tokens until we reach the root (LinkingUSD at 0x20C0...)
+        const LINKING_USD_ROOT: Address = Address::new([
+            0x20, 0xC0, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+            0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+        ]);
+
+        let mut current = next_linking_token;
+        while current != LINKING_USD_ROOT && current != Address::ZERO {
```
0xKitsune: Can you clarify why we need to check `current != Address:ZERO` here?

crates/precompiles/src/contracts/tip20.rs
```diff
@@ -244,6 +252,105 @@ impl<'a, S: StorageProvider> TIP20Token<'a, S> {
...
+            // Get the linking token of the current token
+            if address_is_token_address(&current) {
+                let token_id = address_to_token_id_unchecked(&current);
+                let mut current_token = TIP20Token::new(token_id, self.storage);
+                current = current_token.linking_token();
+            } else {
+                // If it's not a token address, break (shouldn't happen but safety check)
+                break;
+            }
+        }
```
0xKitsune, line 317: I think we can simplify to the following:

```rust

while current != LINKING_USD_ADDRESS {
    if current == self.token_address{
        return Err(TIP20Error::invalid_linking_token());
    }

    let token_id = address_to_token_id_unchecked(&current);
    current = TIP20Token::new(token_id, self.storage).linking_token();

}
```

crates/precompiles/src/contracts/tip20.rs
```diff
@@ -1532,4 +1653,321 @@ mod tests {
             Some(&TIP20Event::Transfer(ITIP20::Transfer { from, to, amount }).into_log_data())
         );
     }
+
+    #[test]
+    fn test_initialize_sets_next_linking_token() {
```
0xKitsune, line 1645: Can we add a test to assert failure when a new linking token proposed is not a TIP20?

crates/contracts/src/precompiles.rs
fgimenez: makes sense, done

crates/precompiles/src/contracts/tip20.rs
fgimenez: sure, missed it, done

crates/precompiles/src/contracts/tip20.rs
fgimenez: good catch, removed

crates/precompiles/src/contracts/tip20.rs
fgimenez: yeah, this was initially in place while developing the loop detection, i noticed that `LINKING_USD_ROOT` has its linking token set to `Address::ZERO` and wanted to make sure that we would exit before reaching that point. 

it is not needed given that when `current == LINKING_USD_ROOT` we won't enter the while, removed

crates/precompiles/src/contracts/tip20.rs
fgimenez: ok, after the previous simplifications only the intermediate `current_token` variable was remaining, removed

crates/precompiles/src/contracts/tip20.rs
fgimenez: sure done, also added one that asserts failure for non-deployed tokens

crates/contracts/src/precompiles.rs
```diff
@@ -78,6 +81,8 @@ sol! {
         event TransferPolicyUpdate(address indexed updater, uint64 indexed newPolicyId);
         event SupplyCapUpdate(address indexed updater, uint256 indexed newSupplyCap);
         event PauseStateUpdate(address indexed updater, bool isPaused);
+        event NextLinkingTokenSet(address indexed updater, address indexed newLinkingToken);
+        event LinkingTokenUpdate(address indexed updater, address indexed newLinkingToken);
```
0xKitsune: nit: Feel free to ignore, but it might be intuitive to update these to  `UpdateLinkingToken`, `LinkingTokenUpdateFinalized` to correspond with the `updateLinkingToken()` and `finalizeLinkingtokenUpdate()` functions.

fgimenez: good point, done here https://github.com/tempoxyz/tempo/pull/430/commits/1f9f2885f1b86bbc191c4b5ec5715b6563060efc

PR #432, opened 2025-10-08
feat(ci): nightly bench

Zygimantass:


.github/workflows/nightly-bench.yml
```diff
@@ -0,0 +1,66 @@
...
+#          if-no-files-found: "error"
+#          name: ${{ matrix.binary }}
+#          path: target/release/${{ matrix.binary }}
+#      - uses: cloudposse/github-action-matrix-outputs-write@v1
+#        id: out
+#        with:
+#          matrix-step-name: ${{ github.job }}
+#          matrix-key: ${{ matrix.binary }}
+#          outputs: |-
+#            artifact_id: ${{ steps.binary_upload.outputs.artifact-id }}
```
0xKitsune: Can we remove or uncomment if needed?

.github/workflows/nightly-bench.yml
```diff
@@ -0,0 +1,66 @@
...
+        uses: foundry-rs/foundry-toolchain@v1
+      - uses: actions/checkout@v5
+        with:
+          repository: 'tempoxyz/infrastructure'
+          path: "./infrastructure"
+          token: '${{ secrets.INFRASTRUCTURE_REPO_READ_TOKEN }}'
+#      - uses: cloudposse/github-action-matrix-outputs-read@v1
+#        id: out
+#        with:
+#          matrix-step-name: build
```
0xKitsune: Same thing here (as well as anywhere else), can we remove or uncomment?

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -327,6 +361,90 @@ pub fn pin_thread(core_id: CoreId) {
...
+) -> eyre::Result<()> {
+    let provider = ProviderBuilder::new().connect_http(rpc_url.clone());
+
+    let mut last_block_timestamp: Option<u64> = None;
+
+    let mut benchmarked_blocks = Vec::new();
+
+    for number in start_block..=end_block {
+        let block = provider.get_block(number.into()).await?.expect("");
+        let receipts = provider.get_block_receipts(number.into()).await?.expect("");
```
0xKitsune: Can we add specific error messages to `expect()` here?

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -327,6 +361,90 @@ pub fn pin_thread(core_id: CoreId) {
...
+
+pub async fn generate_report(
+    rpc_url: &Url,
+    start_block: BlockNumber,
+    end_block: BlockNumber,
+    args: &MaxTpsArgs,
+) -> eyre::Result<()> {
+    let provider = ProviderBuilder::new().connect_http(rpc_url.clone());
+
+    let mut last_block_timestamp: Option<u64> = None;
```
0xKitsune, line 409: Since we are immediately setting this to `Some(val)` in the next block can we simply set this to `0` and drop the `Option`?

Zygimantass: imo not as easy, since it's used for calculating latency, so you still have to check whether its 0 or not

PR #433, opened 2025-10-08
feat(test): add e2e tests with consensus and execution layers

SuperFluffy:
This patch adds a first set of tests to launch tempo nodes with a consensus and execution layer.

The tests are:

1. 5 nodes with perfect linkage; test success reaching height 5
2. 5 nodes with 75% good links (25% failure); test success is reaching height 5
3. 10 nodes with 98% good links; test success is reaching height 20

The tests were written very close to commonware's examples:

1. alto: https://github.com/commonwarexyz/alto/blob/a831c6eab7ec10451dca025cedb34d25ccb0c2f6/chain/src/lib.rs#L42
2. monorepo: https://github.com/commonwarexyz/monorepo/blob/d049d43c3efe3a6a18e379115e587bb9e03bf895/consensus/src/threshold_simplex/mod.rs#L210

## Downsides of this test setup

1. runs are not completely deterministic. Ideally, we two successive runs with the same input parameters return the same audit result (a hash of all metrics collected by the runtime). However, as we are connecting a deterministic (commonware) runtime to a non-deterministic (tokio/reth) runtime, the completion of futures is not always guaranteed, even if reth always returns the same output given the same inputs. The source of this non-determinism is difficult to pinpoint. To support this feature, commonware recently merged the new `external` feature [1] that allows "pacing" certain futures, making the deterministic runtime await them in real time (and not in simulated time, as it normally would).
2. opening too many reth nodes leads to errors like `failed to open the database: unknown error code: 12 (12)`. This only happens if very many runs are launched righ after another, as is the case in the commonware examples, where runs are repeated in a loop with different seeds. Running the tests in nextest seems to help.
3. Runs take a long time. This is partially because of waiting for real time. Given a proposal time of 2s and a verification time of 1s, one can expect 1 block to take up to 3s in real time. So to reach 5 blocks twice (to verifiy audits) can take of the order of 30s. If one now also includes simulated jitter and communication failures (expected/preconfigured) as well as non-determinism issues (unexpected), this number is expected to go up. Reaching 20 blocks (as is the case for one test) should take about 40s, which indeed it does.

## Note on seemingly orthgonal changes in this PR

This patch adds a non-zero `gasLimit` entry to the `ADAGIO` hardcoded genesis/chainspec. `gasLimit` being absent meant that it took a default value of 0, which made it impossible to run tests. The likely failure point is this check: :https://github.com/tempoxyz/tempo/blob/7166a3d8cfbab6a5174e706a079cfd4f1bee2ae1/crates/consensus/src/lib.rs#L57

[1]: https://github.com/commonwarexyz/monorepo/pull/1848

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2347,7 +2347,7 @@ dependencies = [
 [[package]]
 name = "commonware-broadcast"
 version = "0.0.62"
-source = "git+https://github.com/commonwarexyz/monorepo?rev=b500737b20e75b6e5f486d8da3b5d236e5cb9449#b500737b20e75b6e5f486d8da3b5d236e5cb9449"
+source = "git+https://github.com/commonwarexyz/monorepo?rev=0124b5fb2b3859a702f0ae6d9678ce37ac906db4#0124b5fb2b3859a702f0ae6d9678ce37ac906db4"
 dependencies = [
  "bytes",
  "commonware-codec",
@@ -2365,7 +2365,7 @@ dependencies = [
 [[package]]
 name = "commonware-codec"
 version = "0.0.62"
-source = "git+https://github.com/commonwarexyz/monorepo?rev=b500737b20e75b6e5f486d8da3b5d236e5cb9449#b500737b20e75b6e5f486d8da3b5d236e5cb9449"
+source = "git+https://github.com/commonwarexyz/monorepo?rev=0124b5fb2b3859a702f0ae6d9678ce37ac906db4#0124b5fb2b3859a702f0ae6d9678ce37ac906db4"
 dependencies = [
  "bytes",
  "paste",
@@ -2375,7 +2375,7 @@ dependencies = [
 [[package]]
 name = "commonware-consensus"
 version = "0.0.62"
-source = "git+https://github.com/commonwarexyz/monorepo?rev=b500737b20e75b6e5f486d8da3b5d236e5cb9449#b500737b20e75b6e5f486d8da3b5d236e5cb9449"
+source = "git+https://github.com/commonwarexyz/monorepo?rev=0124b5fb2b3859a702f0ae6d9678ce37ac906db4#0124b5fb2b3859a702f0ae6d9678ce37ac906db4"
 dependencies = [
  "bytes",
  "cfg-if",
@@ -2400,7 +2400,7 @@ dependencies = [
 [[package]]
 name = "commonware-cryptography"
 version = "0.0.62"
-source = "git+https://github.com/commonwarexyz/monorepo?rev=b500737b20e75b6e5f486d8da3b5d236e5cb9449#b500737b20e75b6e5f486d8da3b5d236e5cb9449"
+source = "git+https://github.com/commonwarexyz/monorepo?rev=0124b5fb2b3859a702f0ae6d9678ce37ac906db4#0124b5fb2b3859a702f0ae6d9678ce37ac906db4"
 dependencies = [
  "blake3",
  "blst",
@@ -2425,7 +2425,7 @@ dependencies = [
 [[package]]
 name = "commonware-macros"
 version = "0.0.62"
-source = "git+https://github.com/commonwarexyz/monorepo?rev=b500737b20e75b6e5f486d8da3b5d236e5cb9449#b500737b20e75b6e5f486d8da3b5d236e5cb9449"
+source = "git+https://github.com/commonwarexyz/monorepo?rev=0124b5fb2b3859a702f0ae6d9678ce37ac906db4#0124b5fb2b3859a702f0ae6d9678ce37ac906db4"
 dependencies = [
  "futures",
  "proc-macro2",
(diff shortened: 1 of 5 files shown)


crates/chainspec/src/genesis/adagio.json
```diff
@@ -32,6 +32,7 @@
     "pragueTime": 0
   },
   "baseFeePerGas": "0x2c",
+  "gasLimit": 50000000,
```
SuperFluffy, line 35: Needed because otherwise building on a genesis block fails due to https://github.com/tempoxyz/tempo/blob/7166a3d8cfbab6a5174e706a079cfd4f1bee2ae1/crates/consensus/src/lib.rs#L57

crates/commonware-node/src/consensus/engine.rs
```diff
@@ -45,45 +49,35 @@ const MAX_REPAIR: u64 = 20;
 ///
 // XXX: Mostly a one-to-one copy of alto for now. We also put the context in here
 // because there doesn't really seem to be a point putting it into an extra initializer.
-pub(crate) struct Builder<
```
SuperFluffy, line 48: I removed all values that were hardcoded so that the hardcoded consts don't have to be made public just so the `e2e` crate works.

crates/e2e/src/lib.rs
```diff
@@ -0,0 +1,252 @@
...
+    /// The linkage between individual validators.
+    pub linkage: Link,
+    /// The height that a test has to reach to be considered a success.
+    pub height_to_reach: u64,
+}
+
+/// Runs a test configured by [`Setup`].
+pub fn run(
+    Setup {
+        name,
```
klkvr: can this not be random?

crates/e2e/src/lib.rs
```diff
@@ -0,0 +1,252 @@
...
+    let execution_runtime = ExecutionRuntime::new(name);
+    let mut execution_nodes = Vec::with_capacity(how_many as usize);
+    for i in 0..how_many {
+        let node = execution_runtime
+            .spawn_node_blocking(&format!("node-{i}"))
+            .expect("must be able to spawn nodes on the runtime");
+        execution_nodes.push(node);
+    }
+
+    executor.start(|mut context| async move {
```
klkvr, line 58: is there anything preventing us from spawning reth nodes inside of this future vs adding new runtime?

crates/e2e/src/lib.rs
```diff
@@ -0,0 +1,252 @@
...
+            ops::generate_shares::<_, MinSig>(&mut context, None, how_many, threshold);
+
+        let mut public_keys = HashSet::new();
+        for (signer, share) in signers.into_iter().zip(shares) {
+            let public_key = signer.public_key();
+            public_keys.insert(public_key.clone());
+
+            let uid = format!("validator-{public_key}");
+
+            let engine = tempo_commonware_node::consensus::Builder {
```
klkvr, line 98: is there a nice way to unify that with `run_consensus_stack`? most of `Builder` fields are same as `Config` ones, so maybe it should just take entire config?

crates/e2e/src/lib.rs
```diff
@@ -0,0 +1,252 @@
...
+                let value = parts.next().unwrap();
+
+                if metrics.ends_with("_peers_blocked") {
+                    let value = value.parse::<u64>().unwrap();
+                    assert_eq!(value, 0);
+                }
+
+                // TODO(janis): commonware calls this marshal, we call this sync.
+                // We should rename this to marshal (the actor, that is).
+                if metric.ends_with("_sync_processed_height") {
```
klkvr, line 151: are we parsing some kind of commonware metrics here? could we not assert that height is reached via RPC of EL nodes?

crates/e2e/src/lib.rs
SuperFluffy: Yeah, see the comment before that: I had to do this because node bringup took too long and the deterministic runtime stalls out.

Commonware is working on a fix.

crates/e2e/src/lib.rs
SuperFluffy: We could. I'm trying to stay as close to commonware as possible to benefit from their current examples and future work (so that I can mostly copy and paste).

crates/e2e/src/lib.rs
SuperFluffy: It was random at first - I might revert it to that. I ran into issues with reth complaining that it couldn't open a db and wanted to identify when which datadir was created.

It's very jank. 🤨

crates/e2e/src/lib.rs
SuperFluffy: I would like that! I wanted to go this route and started refactoring, but it started taking this PR into a completely different direction.

In want to clean up the node anyway, so will keep this in mind.

crates/chainspec/src/genesis/adagio.json
0xKitsune: Was this failing due to this being `0` or too high of a number? If the former, lets bump this to a larger value, 50M is pretty low for the throughput we want.

crates/e2e/src/lib.rs
0xKitsune: +1 on this, we explored refactoring to something similar before but other changes were prioritized. Would be nice to revisit at some point.

crates/chainspec/src/genesis/adagio.json
SuperFluffy: It failed because it was `0`. I set it to some arbitrary number just to make this work.

I think a change to a "proper" value should be done in a different PR with an actual argument for the number instead of doing it ad-hoc here? :-)

crates/e2e/src/lib.rs
SuperFluffy: To add to this conversation: with the work in this patch and in https://github.com/tempoxyz/tempo/pull/481 we would have run afoul with the abandoned refactors: the anonymous types `impl Sender` and `impl Receiver` are much easier to work with and fit in better with commonware's APIs (which we are heavily using here).

PR #439, opened 2025-10-09
feat(precompiles): implement place and placeFlip functions for stablecoin DEX

### LINKED ISSUE #418: Implement orderbook `place`, `placeFlip` functions
Implement functions to place orders into the stablecoin DEX.

[Spec for reference](https://github.com/tempoxyz/specs/blob/main/specs/StablecoinExchange.md#functions)

fgimenez:
Closes #418 

only focused on place and placeFlip, left TODOs for:
- balance management
- linking token lookup
- orderbook validation
- order cancellation
- balance queries
- taker execution


crates/precompiles/src/contracts/stablecoin_dex/slots.rs
```diff
@@ -0,0 +1,62 @@
...
+/// Remaining amount field offset
+pub const ORDER_REMAINING_OFFSET: U256 = uint!(5_U256);
+/// Previous order ID field offset
+pub const ORDER_PREV_OFFSET: U256 = uint!(6_U256);
+/// Next order ID field offset
+pub const ORDER_NEXT_OFFSET: U256 = uint!(7_U256);
+/// Is flip order boolean field offset
+pub const ORDER_IS_FLIP_OFFSET: U256 = uint!(8_U256);
+/// Flip tick field offset
+pub const ORDER_FLIP_TICK_OFFSET: U256 = uint!(9_U256);
```
0xKitsune: Lets move these out of `slots` since they are not actually mapped to slots. It might be a footgun where we have two variables that both map to slot 0. Maybe we could create a different mod called `offsets` or take a different approach to handling how structs and storage works.

Whenever we refactor the precompile logic we should make note of this as a good example of things to abstract away.

crates/precompiles/src/contracts/stablecoin_dex/mod.rs
```diff
@@ -41,6 +44,98 @@ impl<'a, S: StorageProvider> StablecoinDex<'a, S> {
...
+        let new_id = current + 1;
+        self.storage
+            .sstore(self.address, slots::PENDING_ORDER_ID, U256::from(new_id))
+            .expect("Storage write failed");
+        new_id
+    }
+
+    /// Store a pending order
+    fn store_pending_order(&mut self, order_id: u128, order: &Order) {
+        let order_slot = mapping_slot(U256::from(order_id).to_be_bytes::<32>(), slots::ORDERS);
```
0xKitsune: I believe this should as a `U256` without converting with `to_be_bytes`.

crates/precompiles/src/contracts/stablecoin_dex/mod.rs
```diff
@@ -41,6 +44,98 @@ impl<'a, S: StorageProvider> StablecoinDex<'a, S> {
...
+        let current = self.get_pending_order_id();
+        let new_id = current + 1;
+        self.storage
+            .sstore(self.address, slots::PENDING_ORDER_ID, U256::from(new_id))
+            .expect("Storage write failed");
+        new_id
+    }
+
+    /// Store a pending order
+    fn store_pending_order(&mut self, order_id: u128, order: &Order) {
```
0xKitsune, line 85: Can we also make sure to update `Order` to reflect the recent changes to the reference implementation?

https://github.com/tempoxyz/docs/pull/41

crates/precompiles/src/contracts/stablecoin_dex/slots.rs
fgimenez: good point, done

crates/precompiles/src/contracts/stablecoin_dex/mod.rs
fgimenez: the incoming `order_id` is `u128`, you mean we don't need to convert to `U256` first and instead use `to_be_bytes` directly? pushed changes for this, i believe we need `to_be_bytes` in any case, as we do for instance for `u64` here https://github.com/tempoxyz/tempo/blob/dbf6106ae9a066b292fcc65a77f91682f92ab927/crates/precompiles/src/contracts/tip403_registry.rs#L335

crates/precompiles/src/contracts/stablecoin_dex/mod.rs
fgimenez: absolutely! done here https://github.com/tempoxyz/tempo/pull/439/commits/58cd2218f43d6c45346634e4c229e5af24cd5ed9 ptal

crates/precompiles/src/contracts/stablecoin_dex/order.rs
```diff
@@ -175,99 +179,93 @@ impl Order {
...
-        self.flip_tick.is_some()
+        self.is_flip
     }
 
-    /// Returns the flip tick if this is a flip order.
-    pub fn flip_tick(&self) -> Option<i16> {
+    /// Returns the flip tick (0 for non-flip orders).
+    pub fn flip_tick(&self) -> i16 {
         self.flip_tick
     }
```
0xKitsune: Is it possible that someone places a flip bid order where bid price is 0.99 and the flip price is 1.00? This would make the flip tick 0 for a valid flip order.

crates/precompiles/src/contracts/stablecoin_dex/order.rs
```diff
@@ -278,73 +276,77 @@ impl Order {
...
-                remaining: self.amount,
+                remaining: self.remaining,
             });
         }
 
         // Create flipped order
+        let new_side = match self.side {
+            Side::Bid => Side::Ask,
+            Side::Ask => Side::Bid,
+        };
```
0xKitsune: nit: consider creating a `side.flip()` function.

crates/precompiles/src/contracts/stablecoin_dex/mod.rs
```diff
@@ -41,6 +45,122 @@ impl<'a, S: StorageProvider> StablecoinDex<'a, S> {
...
+                order_slot + offsets::ORDER_BOOK_KEY_OFFSET,
+                U256::from_be_bytes(order.book_key().0),
+            )
+            .expect("Storage write failed");
+
+        // Store side (Bid = 0, Ask = 1)
+        let side_value = match order.side() {
+            Side::Bid => 0u8,
+            Side::Ask => 1u8,
+        };
```
0xKitsune: nit: Consider creating an `Into<T>` impl and we can use `order.side().into()` here.

crates/precompiles/src/contracts/stablecoin_dex/mod.rs
```diff
@@ -100,19 +221,116 @@ impl<'a, S: StorageProvider> StablecoinDex<'a, S> {
...
+        let linking_token = Address::ZERO;
+
+        // Compute book_key from token pair
+        let book_key = self.compute_book_key(token, linking_token);
+
+        // TODO: Validate pair exists and tick is within bounds
+        // TODO: Balance management - debit from user or transfer from user
+
+        // Create the order
+        let order_id = self.increment_pending_order_id();
```
0xKitsune: nit: if we use the order id after incrementing, we will skip orderId 0.

crates/precompiles/src/contracts/stablecoin_dex/mod.rs
```diff
@@ -123,3 +341,128 @@ impl<'a, S: StorageProvider> StablecoinDex<'a, S> {
...
+        let key = dex.compute_book_key(token, token);
+
+        // Verify it produces a valid hash
+        let mut buf = [0u8; 40];
+        buf[..20].copy_from_slice(token.as_slice());
+        buf[20..].copy_from_slice(token.as_slice());
+        let expected = keccak256(buf);
+
+        assert_eq!(key, expected);
+    }
```
0xKitsune: This test seems like it might be unnecessary

crates/precompiles/src/contracts/stablecoin_dex/mod.rs
```diff
@@ -123,3 +341,128 @@ impl<'a, S: StorageProvider> StablecoinDex<'a, S> {
...
+        let key = dex.compute_book_key(larger, smaller);
+
+        // Compute expected: should sort to smaller, then larger
+        let mut buf = [0u8; 40];
+        buf[..20].copy_from_slice(smaller.as_slice());
+        buf[20..].copy_from_slice(larger.as_slice());
+        let expected = keccak256(buf);
+
+        assert_eq!(key, expected, "Tokens should be sorted before hashing");
+    }
```
0xKitsune: This seems covered in `test_compute_book_key_matches_expected_hash`

crates/precompiles/src/contracts/stablecoin_dex/mod.rs
```diff
@@ -123,3 +341,128 @@ impl<'a, S: StorageProvider> StablecoinDex<'a, S> {
...
+        use crate::contracts::storage::slots::mapping_slot;
+        let order_slot = mapping_slot(order_id.to_be_bytes(), slots::ORDERS);
+        let stored_book_key_u256 = dex.sload(order_slot + offsets::ORDER_BOOK_KEY_OFFSET);
+        let stored_book_key = B256::from(stored_book_key_u256.to_be_bytes::<32>());
+
+        assert_eq!(
+            stored_book_key, expected_book_key,
+            "Stored book_key should match computed book_key"
+        );
+    }
```
0xKitsune: Do we need this? A better test might be store a full order and fetch/assert that it matches.

crates/precompiles/src/precompiles/stablecoin_dex.rs
```diff
@@ -0,0 +1,501 @@
...
+        let stored_side = dex.sload(order_slot + offsets::ORDER_SIDE_OFFSET);
+        assert_eq!(stored_side, U256::from(0u8)); // Bid = 0
+
+        // Check is_flip (false = 0)
+        let stored_is_flip = dex.sload(order_slot + offsets::ORDER_IS_FLIP_OFFSET);
+        assert_eq!(stored_is_flip, U256::ZERO);
+    }
+
+    #[test]
+    fn test_flip_order_storage_persistence() {
```
0xKitsune: Looks like this is essentially the same test as `test_order_storage_persistence`. I think we can remove this one and add tick checks to the test above. Lmk if I am overlooking something though.

crates/precompiles/src/precompiles/stablecoin_dex.rs
```diff
@@ -0,0 +1,501 @@
...
+        dex.call(&Bytes::from(place_call.abi_encode()), &sender)
+            .unwrap();
+
+        // Check pending_order_id incremented to 2
+        let pending_after_second = dex.sload(slots::PENDING_ORDER_ID);
+        assert_eq!(pending_after_second, U256::from(2));
+    }
+
+    #[test]
+    fn test_place_flip_ask_with_valid_flip_tick() {
```
0xKitsune, line 374: We have tests for placing valid/invalid ask flip orders, but only testing invalid bid flip orders. Can we add this case? We can also assert valid/invalid flip orders in a single test since these are largely doing the same thing. No strong preference here though.

crates/precompiles/src/contracts/stablecoin_dex/order.rs
fgimenez: good catch, updated comment

crates/precompiles/src/contracts/stablecoin_dex/order.rs
fgimenez: sure done

crates/precompiles/src/contracts/stablecoin_dex/mod.rs
fgimenez: indeed, nice

crates/precompiles/src/contracts/stablecoin_dex/mod.rs
fgimenez: totally, removed

crates/precompiles/src/contracts/stablecoin_dex/mod.rs
fgimenez: yep, removed

crates/precompiles/src/contracts/stablecoin_dex/mod.rs
fgimenez: we are actually doing that in `test_order_storage_persistence`, removed this one

crates/precompiles/src/precompiles/stablecoin_dex.rs
fgimenez: yep, removed `test_flip_order_storage_persistence` and added flip_tick checking to the other one

crates/precompiles/src/precompiles/stablecoin_dex.rs
fgimenez: good point, added `test_place_flip_bid_with_valid_flip_tick`

crates/precompiles/src/contracts/stablecoin_dex/mod.rs
fgimenez: great catch, fixed

PR #440, opened 2025-10-09
feat: predeploy Arachnid CREATE2 factory

### LINKED ISSUE #434: feat: add support for Foundry's default CREATE2 factory
I think it would make sense if the default CREATE2 factory of Foundry (commonly referred to as Nick’s factory or Arachnid's Deterministic Deployment Proxy) at `0x4e59b44847b379578588920cA78FbF26c0B4956C` was supported as a pre-deploy. This way developers don’t easily run into issues, requiring configuration to address it.

For reference, it also comes pre-installed on OP Stack https://specs.optimism.io/protocol/preinstalls.html

A clear downside is that because the contract is pure Yul so no ABI is available & most block explorers can’t verify the source code.

We are adding primary support for CreateX soon in Foundry (https://github.com/foundry-rs/foundry/issues/2638) but for backwards compatibility this CREATE2 factory will remain in use.

Ref: https://github.com/Arachnid/deterministic-deployment-proxy
Ref: https://github.com/tempoxyz/tempo/blob/main/crates/contracts/src/lib.rs
Ref: https://etherscan.io/address/0x4e59b44847b379578588920ca78fbf26c0b4956c#code
Source: https://github.com/Arachnid/deterministic-deployment-proxy/blob/master/source/deterministic-deployment-proxy.yul

klkvr:
Closes #434

### DIFF
--- a/crates/contracts/src/lib.rs
+++ b/crates/contracts/src/lib.rs
@@ -13,9 +13,14 @@ pub const SAFE_DEPLOYER_ADDRESS: Address = address!("0x914d7Fec6aaC8cd542e72Bca7
 pub const PERMIT2_ADDRESS: Address = address!("0x000000000022d473030f116ddee9f6b43ac78ba3");
 pub const DEFAULT_7702_DELEGATE_ADDRESS: Address =
     address!("0x7702c00000000000000000000000000000000000");
+pub const ARACHNID_CREATE2_FACTORY_ADDRESS: Address =
+    address!("0x4e59b44847b379578588920cA78FbF26c0B4956C");
 
 pub mod contracts {
-    use alloy::sol;
+    use alloy::{
+        primitives::{Bytes, bytes},
+        sol,
+    };
 
     sol!(
         #[allow(missing_docs)]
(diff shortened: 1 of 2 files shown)


crates/contracts/src/lib.rs
```diff
@@ -13,9 +13,14 @@ pub const SAFE_DEPLOYER_ADDRESS: Address = address!("0x914d7Fec6aaC8cd542e72Bca7
 pub const PERMIT2_ADDRESS: Address = address!("0x000000000022d473030f116ddee9f6b43ac78ba3");
 pub const DEFAULT_7702_DELEGATE_ADDRESS: Address =
     address!("0x7702c00000000000000000000000000000000000");
+pub const ARACHNID_CREATE2_FACTORY_ADDRESS: Address =
```
0xKitsune, line 16: Is there a reason Arachnid's deterministic deployment was selected vs pcvaersaccio's?
https://github.com/pcaversaccio/create2deployer

It looks like pcaversaccio's has more active development and wider distribution. Arachnid's latest changes look to be 6 years old.

crates/contracts/src/lib.rs
```diff
@@ -51,6 +56,10 @@ pub mod contracts {
         SafeDeployer,
         "abi/SafeDeployer.json",
     );
+
+    pub const ARACHNID_CREATE2_FACTORY_BYTECODE: Bytes = bytes!(
+        "0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe03601600081602082378035828234f58015156039578182fd5b8082525050506014600cf3"
+    );
```
zerosnacks, line 62: lgtm

ref: https://github.com/foundry-rs/foundry/blob/92d7b3551882272c820df48a38eeb4899517e60d/crates/evm/core/src/constants.rs#L50-L53

ref: https://etherscan.io/address/0x4e59b44847b379578588920ca78fbf26c0b4956c#code

PR #443, opened 2025-10-10
doc(commonware-node): explain backfill from consensus to execution

### LINKED ISSUE #272: Time out consensus block subscriptions when backfilling
The backfill mechanism introduced in https://github.com/tempoxyz/tempo/pull/261 polls the consensus layer for blocks indefinitely. This is bad because no events are emitted to inform the user of the current state, and there is also no guarantee that a block will ever be made available.

Instead, it should be timed out and the backfill requeued.

SuperFluffy:
Explains how backfill works and why it does not stall indefinitely.

Closes #272

### DIFF
--- a/crates/commonware-node/src/consensus/execution_driver/executor.rs
+++ b/crates/commonware-node/src/consensus/execution_driver/executor.rs
@@ -153,8 +153,11 @@ impl Executor {
     /// Backfills a block identified by `digest` by reading it from the
     /// consensus layer.
     ///
-    /// `round` must only be set if `digest` is a notarized block. If it is not,
-    /// this function will stall indefinitely.
+    /// Must only be called if `digest` belongs to a notarized block. If `digest`
+    /// does not identify a notarized block the triggered backfill will never
+    /// complete. `round` is used for blocks which are not yet available in the
+    /// consensus layer to help the marshal agent request it from the consensus
+    /// p2p network.
     ///
     /// Note that this function is not async. Backfills are run in an async
     /// queue and are allowed to finish anytime.


crates/commonware-node/src/consensus/execution_driver/executor.rs
```diff
@@ -167,18 +201,6 @@ impl Executor {
         ),
     )]
     fn backfill(&mut self, cause: Span, round: Option<Round>, digest: Digest) {
-        is_send(
```
SuperFluffy, line 170: This is an oversight from debugging the future returned by `backill_from_consensus`. Was never supposed to be merged (and also doesn't do anything as it is right now, because the future is never polled).

crates/commonware-node/src/consensus/execution_driver/executor.rs
```diff
@@ -150,13 +150,50 @@ impl Executor {
...
 
-    /// Attempts to backfill a block by reading its `digest` from the consensus layer.
+    /// Backfills a block identified by `digest` by reading it from the
+    /// consensus layer.
     ///
-    /// `round` must only be set if `digest` is a notarized block. If it is not
-    /// this function will stall indefinitely.
+    /// Must only be called if `digest` belongs to a notarized block. If `digest`
+    /// does not identify a notarized block the triggered backfill will never
+    /// complete. `round` is used for blocks which are not yet available in the
```
0xKitsune, line 158: Is it possible to safeguard this and return an error in this case?

SuperFluffy: I am afraid we cannot. We rely on the caller of `canonicalize` to supply us with correct information (in this case, correctly forwarding the information supplied by the consensus verify and propose events).

Issue #446, opened 2025-10-10
Bug: Use of system time in `TIP20` precompile


0xKitsune:
Currently, the TIP20 precompile uses system time which could potentially create a chain fork. This should be updated to remove system time from precompiles. 

https://github.com/tempoxyz/tempo/blob/46f6bc28549c8a3e2ff44aae35af3c14e4fe7d6d/crates/precompiles/src/contracts/tip20.rs#L629-L635

PR #447, opened 2025-10-10
chore(clippy): deny systemtime in precompiles

Zygimantass:
to prevent #446 happening in the future, this PR adds a clippy disallowed method for system time calls


crates/precompiles/clippy.toml
```diff
@@ -0,0 +1,4 @@
+disallowed-methods = [
+    { path = "std::time::SystemTime::now", reason = "system time in precompiles can cause determinism, use block time instead" },
+    { path = "std::time::Instant::now", reason = "system time in precompiles can cause determinism, use block time instead" },
```
0xKitsune: nit: should this be "non-determinism" rather than "determinism?

Zygimantass: yes - my bad

PR #454, opened 2025-10-13
feat: complete StablecoinExchange execution logic with balance management and flip orders

fgimenez:
adds missing logic according to spec:
* pending order cancellation in `cancel()` - handle orders in pending queue
* balance management in `place()` and `place_flip()` - escrow tokens with TIP20 integration
* quote token lookup - use `TIP20Token::linking_token()` for pair discovery
* flip order logic in `fill_order()` - auto-create new orders when flip orders fill

### DIFF
--- a/crates/precompiles/src/contracts/stablecoin_dex/mod.rs
+++ b/crates/precompiles/src/contracts/stablecoin_dex/mod.rs
@@ -466,7 +466,7 @@ impl<'a, S: StorageProvider> StablecoinDex<'a, S> {
             .storage
             .sload(self.address, order_slot + offsets::ORDER_MAKER_OFFSET)
             .expect("TODO: handle error");
-        if maker == U256::ZERO {
+        if maker.is_zero() {
             return;
         }
 
@@ -479,7 +479,7 @@ impl<'a, S: StorageProvider> StablecoinDex<'a, S> {
             .storage
             .sload(self.address, order_slot + offsets::ORDER_SIDE_OFFSET)
             .expect("TODO: handle error");
-        let is_bid = side_u256 == U256::ZERO;
+        let is_bid = side_u256.is_zero();
 
         let tick = self
             .storage
@@ -575,7 +575,7 @@ impl<'a, S: StorageProvider> StablecoinDex<'a, S> {
             .storage
             .sload(self.address, order_slot + offsets::ORDER_SIDE_OFFSET)
             .expect("TODO: handle error");
-        let is_bid = side_u256 == U256::ZERO;
+        let is_bid = side_u256.is_zero();
 
         let tick = self
             .storage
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/contracts/stablecoin_dex/mod.rs
```diff
@@ -294,14 +294,31 @@ impl<'a, S: StorageProvider> StablecoinDex<'a, S> {
...
 
         // Compute book_key from token pair
-        let book_key = self.compute_book_key(token, linking_token);
+        let book_key = self.compute_book_key(token, quote_token);
 
-        // TODO: Validate pair exists and tick is within bounds
-        // TODO: Balance management - debit from user or transfer from user
+        // Validate tick is within bounds
+        if !(MIN_TICK..=MAX_TICK).contains(&tick) {
+            panic!("Tick out of bounds");
```
0xKitsune: We should update to return proper errors rather than panicking here.

crates/precompiles/src/contracts/stablecoin_dex/mod.rs
```diff
@@ -347,14 +364,42 @@ impl<'a, S: StorageProvider> StablecoinDex<'a, S> {
...
+        let quote_token =
+            TIP20Token::new(address_to_token_id_unchecked(&token), self.storage).linking_token();
 
         // Compute book_key from token pair
-        let book_key = self.compute_book_key(token, linking_token);
+        let book_key = self.compute_book_key(token, quote_token);
+
+        // Validate tick and flip_tick are within bounds
+        if !(MIN_TICK..=MAX_TICK).contains(&tick) {
+            panic!("Tick out of bounds");
```
0xKitsune: Lets add error handling here and everywhere else that currently uses panic.

crates/precompiles/src/contracts/stablecoin_dex/mod.rs
```diff
@@ -580,6 +621,46 @@ impl<'a, S: StorageProvider> StablecoinDex<'a, S> {
         }
 
         if new_remaining == 0 {
+            // Check if this is a flip order before removing it
+            let is_flip = self
+                .storage
+                .sload(self.address, order_slot + offsets::ORDER_IS_FLIP_OFFSET)
+                .expect("TODO: handle error");
+
+            if is_flip != U256::ZERO {
```
0xKitsune: nit: consider using `!is_flip.is_zero()` or consider coverting into `bool` after loading from storage.

crates/precompiles/src/contracts/stablecoin_dex/mod.rs
fgimenez: sure, added errors and removed panics

crates/precompiles/src/contracts/stablecoin_dex/mod.rs
fgimenez: good point, removed all occurrences

Issue #139, opened 2025-08-29
Remove timestamp > parent.timestamp rule for payloads


mattsse:
The default eth payloadvalidator enforces engineapi rule

>   /// By default, this enforces that the payload attributes timestamp is greater than the
    /// timestamp according to:
    ///   > 7. Client software MUST ensure that payloadAttributes.timestamp is greater than
    ///   > timestamp
    ///   > of a block referenced by forkchoiceState.headBlockHash.
    ///
    /// See also: <https://github.com/ethereum/execution-apis/blob/main/src/engine/common.md#specification-1>

https://github.com/paradigmxyz/reth/blob/5c0c8bb38d369de19160d03ccbb06c18cfd9c3f7/crates/engine/primitives/src/lib.rs#L171-L174

which effectively enshrines a blocktime >= 1s

this can be configured via a `PayloadValidator` implementation, that would also be required if tempo's executionpayload is just the block itself and not like ethereum, the beacon executionpayload type

## TODO
* introduce TempoEngineValidator here

https://github.com/tempoxyz/tempo/blob/d61fa024b26aa3214ac3505ca3263bc74b7eb809/crates/node/src/node.rs#L295-L312
as replacement for:

https://github.com/tempoxyz/tempo/blob/327d0dd16ddad5a11435b22311cb4b70f6f63752/crates/node/src/node.rs#L310-L310

PR #455, opened 2025-10-13
feat: `TempoHeader::timestamp_millis_part`

klkvr:
Adds `timestamp_millis_part` field to `TempoHeader` defining milliseconds portion of the timestamp.

For now it is not propagated to EVM as this would require some more upstream changes alike to https://github.com/alloy-rs/evm/pull/193

### DIFF
--- a/crates/primitives/src/header.rs
+++ b/crates/primitives/src/header.rs
@@ -30,7 +30,10 @@ pub struct TempoHeader {
 impl TempoHeader {
     /// Returns the timestamp in milliseconds.
     pub fn timestamp_millis(&self) -> u64 {
-        self.inner.timestamp() * 1000 + self.timestamp_millis_part
+        self.inner
+            .timestamp()
+            .saturating_mul(1000)
+            .saturating_add(self.timestamp_millis_part)
     }
 }
 


crates/commonware-node/src/consensus/execution_driver/mod.rs
```diff
@@ -423,41 +423,11 @@ impl Inner<Init> {
...
-                prev_randao: B256::ZERO,
-                // XXX(tempo): empty withdrawals post-shanghai
-                withdrawals: Withdrawals::default(),
-                // TODO: tempo-malachite did this (why?); but maybe we can
-                // use the consensus block' digest for this? alternatively somehow
-                // tie this to the threshold simplex view / round / height?;
-                parent_beacon_block_root: Some(B256::ZERO),
-            },
-            context.current().epoch_millis() % 1000,
+            payload_id_from_block_hash(&parent.block_hash()),
```
SuperFluffy, line 431: Let's keep the comment for why we use `payload_if_from_block_hash`?

crates/payload/types/src/attrs.rs
```diff
@@ -33,11 +33,24 @@ pub struct TempoPayloadBuilderAttributes {
...
+        timestamp_millis: u64,
+    ) -> Self {
         Self {
-            inner,
-            interrupt: InterruptHandle(Arc::new(atomic::AtomicBool::new(false))),
-            timestamp_millis_part,
+            inner: EthPayloadBuilderAttributes {
+                id,
+                parent,
+                timestamp: timestamp_millis / 1000,
```
SuperFluffy: I'd move the timestamp construction up one level so that it becomes obvious what flows into the object below:

```rust
let (seconds, millis) = (timestamp_millis / 1000, timestamp_millis % 1000);
```

crates/commonware-node/src/consensus/execution_driver/mod.rs
```diff
@@ -423,41 +423,11 @@ impl Inner<Init> {
...
-        // XXX: ensures the timestamp is strictly monotonically increasing.
-        // This is a requirement to pass eth/reth checks.
-        //
-        // TODO: revisit this to use `SystemTimeExt::epoch_millis` again
-        // once it is/if it becomes possible to use milliseconds in
-        // reth. This is to ensure a consistent view of timestamps in
-        // reth vs the consensus engine.
-        let mut timestamp = context.current().epoch().as_secs();
-        if timestamp <= parent.timestamp() {
-            timestamp = parent.timestamp().saturating_add(1);
```
SuperFluffy, line 435: This is now entirely removed - will that be a problem since now `timestamp = parent.timestamp` in some cases? You mentioned on slack that nothing is yet being done with the millisecond part. So I wonder if this would kill payload builds?

klkvr: It should be fine, we've relaxed the check in https://github.com/tempoxyz/tempo/issues/139

crates/primitives/src/header.rs
```diff
@@ -18,11 +18,22 @@ pub struct TempoHeader {
...
+
     /// Inner Ethereum [`Header`].
     #[cfg_attr(feature = "serde", serde(flatten))]
     pub inner: Header,
 }
 
+impl TempoHeader {
+    /// Returns the timestamp in milliseconds.
+    pub fn timestamp_millis(&self) -> u64 {
+        self.inner.timestamp() * 1000 + self.timestamp_millis_part
```
SuperFluffy: Should this be saturating ops?

And provide unit tests?

PR #465, opened 2025-10-14
feat: add stablecoin DEX executeBlock system transaction

### LINKED ISSUE #421: Implement stablecoin DEX `executeBlock` system transaction
The current Stablecoin DEX spec defines that the `placeOrder` and `placeFlipOrder` functions should enqueue orders into a pending state, which are then committed at the end of the block.

We will need a system transaction that runs at the end of each block committing pending orders to the book, similar to how the feeAMM `executeBlock` function is handled.  We should also have block validation rules ensuring that blocks are only valid if the system transaction has been included.

[Spec for reference](https://github.com/tempoxyz/specs/blob/main/specs/StablecoinExchange.md#order-execution-algorithm--data-structures)

fgimenez:
Closes #421 

- adds executeBlock() system transaction to IStablecoinExchange that commits pending orders at block end
- updates block validation to support two system transactions (fee manager + stablecoin exchange)
- blocks are now only valid if both system transactions are included

### DIFF
--- a/crates/evm/src/block.rs
+++ b/crates/evm/src/block.rs
@@ -114,7 +114,7 @@ where
                 .into_iter()
                 .chain(block)
                 .collect::<Bytes>();
-            if *tx.input() == dex_input && self.seen_fee_manager_system_tx {
+            if *tx.input() == dex_input && self.seen_stablecoin_dex_system_tx {
                 Err(BlockValidationError::msg(
                     "duplicate stablecoin DEX system transaction",
                 ))


crates/payload/builder/src/lib.rs
```diff
@@ -101,6 +104,32 @@ impl<Provider: ChainSpecProvider> TempoPayloadBuilder<Provider> {
...
+                    gas_limit: 0,
+                    to: STABLECOIN_EXCHANGE_ADDRESS.into(),
+                    value: U256::ZERO,
+                    input,
+                },
+                TEMPO_SYSTEM_TX_SIGNATURE,
+            )),
+            TEMPO_SYSTEM_TX_SENDER,
+        )
+    }
```
0xKitsune: Wdyt about updating `build_seal_block_tx()` to `build_seal_block_txs()` and return a `Recovered<TempoTxEnvelope>`. This way if there are additional system transactions needed at the end of the block, we can add them here.

crates/contracts/src/precompiles.rs
```diff
@@ -319,6 +319,9 @@ sol! {
         // Balance management
         function withdraw(address token, uint128 amount) external;
 
+        // System transaction
```
0xKitsune:
```suggestion
        // System transaction called at the end of the block to finalize pending limit orders
```

crates/evm/src/block.rs
```diff
@@ -165,7 +183,12 @@ where
...
 
         if tx.tx().is_system_tx() {
-            self.seen_system_tx = true;
+            // Mark which system transaction we've seen based on the target address
+            if tx.tx().to() == Some(TIP_FEE_MANAGER_ADDRESS) {
+                self.seen_fee_manager_system_tx = true;
+            } else if tx.tx().to() == Some(STABLECOIN_EXCHANGE_ADDRESS) {
+                self.seen_stablecoin_dex_system_tx = true;
+            }
         }
```
0xKitsune:
```suggestion
    let inner_tx = tx.tx();

    if inner_tx.is_system_tx() {
        match inner_tx.to() {
            Some(addr) if addr == TIP_FEE_MANAGER_ADDRESS => {
                self.seen_fee_manager_system_tx = true;
            }
            Some(addr) if addr == STABLECOIN_EXCHANGE_ADDRESS => {
                self.seen_stablecoin_dex_system_tx = true;
            }
            _ => {}
        }
    }
```

crates/evm/src/block.rs
```diff
@@ -82,32 +86,46 @@ where
...
+            if self.seen_stablecoin_dex_system_tx {
+                return Err(BlockValidationError::msg(
+                    "duplicate stablecoin DEX system transaction",
+                ));
+            }
+            return Ok(());
+        }
+
+        Err(BlockValidationError::msg("invalid system transaction"))
     }
```
0xKitsune, line 127:
```suggestion
fn validate_system_tx(&self, tx: &TempoTxEnvelope) -> Result<(), BlockValidationError> {
    let block = self.evm().block().number.to_be_bytes_vec();
    let fee_input = executeBlockCall.abi_encode().into_iter().chain(block.clone()).collect::<Bytes>();
    let dex_input  = IStablecoinExchange::executeBlockCall {}
        .abi_encode().into_iter().chain(block).collect::<Bytes>();

    let to = tx.to();
    let input = tx.input();

    match to {
        Some(TIP_FEE_MANAGER_ADDRESS) if input == &fee_input => {
            if self.seen_fee_manager_system_tx {
                Err(BlockValidationError::msg("duplicate fee manager system transaction"))
            } else {
                Ok(())
            }
        }
        Some(STABLECOIN_EXCHANGE_ADDRESS) if input == &dex_input => {
            if self.seen_stablecoin_dex_system_tx {
                Err(BlockValidationError::msg("duplicate stablecoin DEX system transaction"))
            } else {
                Ok(())
            }
        }
        _ => Err(BlockValidationError::msg("invalid system transaction")),
    }
}


```

crates/payload/builder/src/lib.rs
fgimenez: good point, done here https://github.com/tempoxyz/tempo/pull/465/commits/6d438026f4829b376b8f217a3bd3e00cf1397656 ptal

crates/evm/src/block.rs
```diff
@@ -82,32 +86,47 @@ where
...
+                if self.seen_stablecoin_dex_system_tx {
+                    Err(BlockValidationError::msg(
+                        "duplicate stablecoin DEX system transaction",
+                    ))
+                } else {
+                    Ok(())
+                }
+            }
+            _ => Err(BlockValidationError::msg("invalid system transaction")),
         }
```
0xKitsune, line 126: I think we may be able tosimplify further

```suggestion
        let block = self.evm().block().number.to_be_bytes_vec();
        let to = tx.to().unwrap_or_default();

        if to == TIP_FEE_MANAGER_ADDRESS {
            let fee_input = executeBlockCall
                .abi_encode()
                .into_iter()
                .chain(block)
                .collect::<Bytes>();
            if *tx.input() == fee_input && self.seen_fee_manager_system_tx {
                Err(BlockValidationError::msg(
                    "duplicate fee manager system transaction",
                ))
            } else {
                Ok(())
            }
        } else if to == STABLECOIN_EXCHANGE_ADDRESS {
            let dex_input = IStablecoinExchange::executeBlockCall {}
                .abi_encode()
                .into_iter()
                .chain(block)
                .collect::<Bytes>();

            if *tx.input() == dex_input && self.seen_fee_manager_system_tx {
                Err(BlockValidationError::msg(
                    "duplicate stablecoin DEX system transaction",
                ))
            } else {
                Ok(())
            }
        } else {
            Err(BlockValidationError::msg("invalid system transaction"))
        }

```

fgimenez: nice, good to have `dex_input` and `fee_input` defined only when needed :+1:

PR #466, opened 2025-10-14
test(stablecoin-exchange): add unit tests for bitmap bit_index calculation

fgimenez:


crates/precompiles/src/contracts/stablecoin_exchange/orderbook.rs
```diff
@@ -575,4 +575,221 @@ mod tests {
...
     }
+
+    mod bitmap_tests {
+        use super::*;
+        use crate::contracts::HashMapStorageProvider;
+
+        #[test]
+        fn test_set_and_check_positive_ticks() {
+            let mut storage = HashMapStorageProvider::new(1);
+            let address = address!("0x1111111111111111111111111111111111111111");
```
0xKitsune: nit: we can use `Address::random()`

crates/precompiles/src/contracts/stablecoin_exchange/orderbook.rs
```diff
@@ -575,4 +575,221 @@ mod tests {
...
+            let book_key = B256::ZERO;
+
+            // Set various positive ticks and verify they're marked as initialized
+            for tick in [0, 1, 100, 255, 256, 500, 1000, MAX_TICK] {
+                let mut bitmap = TickBitmap::new(&mut storage, address, book_key);
+                bitmap.set_tick_bit(tick, true);
```
0xKitsune: Currently `set_tick_bit` has a `todo!()`:
```rust
    if !(MIN_TICK..=MAX_TICK).contains(&tick) {
            todo!()
        }

```

Can we update to have proper error handling here and test for these cases?

crates/precompiles/src/contracts/stablecoin_exchange/orderbook.rs
```diff
@@ -575,4 +575,221 @@ mod tests {
...
+                let mut bitmap = TickBitmap::new(&mut storage, address, book_key);
+                assert!(
+                    bitmap.is_tick_initialized(tick, true),
+                    "Tick {tick} should be initialized"
+                );
+            }
+        }
+
+        #[test]
+        fn test_set_clear_and_check() {
```
0xKitsune: It looks like these tests are largely doing the same thing. Wdyt about updating to a single `test_negative_ticks()` test that sets the tick, checks initialized, clears the tick and checks that it is uninitialized. We can also do the same for `test_positive_ticks()`. I think we could even have a single test that covers a range of positive/negative ticks in a single test if we want to simplify further.

crates/precompiles/src/contracts/stablecoin_exchange/orderbook.rs
```diff
@@ -575,4 +575,221 @@ mod tests {
...
+                "Tick should still be initialized for bids"
+            );
+            assert!(
+                bitmap.is_tick_initialized(tick, false),
+                "Tick should now be initialized for asks"
+            );
+        }
+
+        #[test]
+        fn test_ticks_in_same_word() {
```
0xKitsune: I think we already cover this with the existing tests. Is there something this adds that isnt already covered?

crates/precompiles/src/contracts/stablecoin_exchange/orderbook.rs
fgimenez: sure done

crates/precompiles/src/contracts/stablecoin_exchange/orderbook.rs
fgimenez: sure good point, done added `InvalidTick` variant to `OrderError`

crates/precompiles/src/contracts/stablecoin_exchange/orderbook.rs
fgimenez: indeed, refactored into `test_tick_lifecycle`

crates/precompiles/src/contracts/stablecoin_exchange/orderbook.rs
fgimenez: yep, already covered by `test_tick_lifecycle` removed

crates/precompiles/src/contracts/stablecoin_exchange/error.rs
```diff
@@ -105,4 +116,18 @@ mod tests {
...
+        let err = OrderError::InvalidTick {
+            tick: 3000,
+            min: -2000,
+            max: 2000,
+        };
+        let msg = err.to_string();
+        assert!(msg.contains("3000"));
+        assert!(msg.contains("-2000"));
+        assert!(msg.contains("2000"));
+        assert!(msg.contains("out of bounds"));
```
0xKitsune: Is there a reason not to assert the exact error here?

```rust
  let msg = err.to_string();
    assert_eq!(
        msg,
        "tick 3000 is out of bounds (min: -2000, max: 2000)"
    );
```

crates/precompiles/src/contracts/stablecoin_exchange/mod.rs
```diff
@@ -605,7 +605,7 @@ impl<'a, S: StorageProvider> StablecoinExchange<'a, S> {
 
             let mut bitmap =
                 orderbook::TickBitmap::new(self.storage, self.address, B256::from(book_key));
-            bitmap.set_tick_bit(tick, is_bid);
+            bitmap.set_tick_bit(tick, is_bid).unwrap();
```
0xKitsune: Can we add a comment detailing why we can use unwrap here?

crates/precompiles/src/contracts/stablecoin_exchange/orderbook.rs
```diff
@@ -575,4 +591,240 @@ mod tests {
...
+            let result = bitmap.clear_tick_bit(MIN_TICK - 1, true);
+            assert!(result.is_err());
+            assert!(matches!(
+                result.unwrap_err(),
+                OrderError::InvalidTick { .. }
+            ));
+        }
+
+        #[test]
+        fn test_out_of_bounds_tick_is_initialized() {
```
0xKitsune: nit: consider renaming for readability. Same for the above tests.
```suggestion
        fn test_tick_is_initialized_out_of_bounds_() {
```

crates/precompiles/src/contracts/stablecoin_exchange/orderbook.rs
```diff
@@ -575,4 +591,240 @@ mod tests {
...
+            bitmap.set_tick_bit(255, true).unwrap(); // word_index = 0, bit_index = 255
+            bitmap.set_tick_bit(256, true).unwrap(); // word_index = 1, bit_index = 0
+
+            let mut bitmap = TickBitmap::new(&mut storage, address, book_key);
+            assert!(bitmap.is_tick_initialized(255, true).unwrap());
+            assert!(bitmap.is_tick_initialized(256, true).unwrap());
+        }
+
+        #[test]
+        fn test_negative_ticks_different_words() {
```
0xKitsune: Does testing negative tick values across word boundaries provide any additional coverage compared to positive ones? If not, maybe we can update this to `test_ticks_different_words` and test both negative/postive ticks?

crates/precompiles/src/contracts/stablecoin_exchange/error.rs
fgimenez: was trying to not be tight to the concrete error string, but we were checking separately almost every piece :) asserting the exact error now

crates/precompiles/src/contracts/stablecoin_exchange/mod.rs
fgimenez: good point, this was relying on validations in place and place_flip, much better to propagate potential errors properly, removed unwrap and changed `process_pending_order` and `fill_order` to return `Result`

crates/precompiles/src/contracts/stablecoin_exchange/orderbook.rs
fgimenez: sure done for the 3 of them

crates/precompiles/src/contracts/stablecoin_exchange/orderbook.rs
fgimenez: good point done

PR #468, opened 2025-10-14
fix(rpc): update `caller_gas_allowance` to divide by gas price

### LINKED ISSUE #459: High transfer fees on testnet
Sending via the demo app on testnet is not working due to high fees requested. Demo app sends do not specify a fee token, so it's all using `AlphaUSD`.

The amount requested by the network is ~770x the amount you get from topping up at the faucet

**Error Message**

```
Caused by: TransactionExecutionError: The total cost (gas * gas fee + value) of executing this transaction exceeds the balance of the account.

This error could arise when the account does not have enough funds to:
 - pay for the total gas fee,
 - pay for the value to send.
 
The cost of the transaction is calculated as `gas * gas fee + value`, where:
 - `gas` is the amount of gas needed for transaction to execute,
 - `gas fee` is the gas fee,
 - `value` is the amount of ether to send to the recipient.
 
Request Arguments:
  from:  0xbae08717f27078aFAa377a135a970014aEd9779A
  to:    0x20c0000000000000000000000000000000000000
  data:  0xa9059cbb000000000000000000000000beefcafe54750903ac1c8909323af7beb21ea2cb0000000000000000000000000000000000000000000000000000000000b71b00

Details: insufficient funds for gas * price + value: have 1000000000000 want 774056185954260
```

**How to repro**
- go to https://internal-test.tempo.xyz/send (password is in the shared 1password)
- click connect wallet
- click fund wallet
- enter in a send
- see message in console

0xKitsune:
Closes #459 

This PR patches `caller_gas_allowance()` to divide the adjusted balance by the gas price.

### DIFF
--- a/crates/node/src/rpc/mod.rs
+++ b/crates/node/src/rpc/mod.rs
@@ -37,7 +37,6 @@ use reth_rpc_eth_types::{
 use tempo_evm::TempoEvmConfig;
 use tempo_precompiles::contracts::provider::TIPFeeDatabaseExt;
 use tempo_primitives::TempoReceipt;
-use tempo_transaction_pool::validator::USD_DECIMAL_FACTOR;
 use tokio::sync::Mutex;
 
 /// Tempo `Eth` API implementation.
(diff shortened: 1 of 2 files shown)


crates/node/src/rpc/mod.rs
```diff
@@ -219,11 +219,16 @@ impl<N: FullNodeTypes<Types = TempoNode>> Call for TempoEthApi<N> {
 
         // Fee token balance is denominated in USD Decimals and the gas allowance is expected in
         // 10**9 so we must adjust by USD_DECIMAL_FACTOR
-        let adjusted_balance = fee_token_balance
+        let adjusted_balance: U256 = fee_token_balance
             .saturating_mul(USD_DECIMAL_FACTOR)
```
klkvr: why do we need the `USD_DECIMAL_FACTOR`? I don't see us multiplying the balance here https://github.com/tempoxyz/tempo/blob/21ea5550e564e9fc26614129762c7cfa917fa8c7/crates/revm/src/handler.rs#L116-L117

0xKitsune: This is because fee tokens are in USD decimals (6 decimals) and gas has 9 decimals. Maybe we also should be adjusting for `USD_DECIMAL_FACTOR` in `validate_against_state_and_deduct_caller()`? Lmk if I am overlooking something though if there is a reason we shouldnt be adjusting decimals here?

klkvr: This logic should mirror what happens in `validate_against_state_and_deduct_caller` which does not multiply by the decimals right now afaict from looking at the code

Gas doesn't really have a concept of decimals, it's just `required_funds = gas * gas_price`

To me personally this multiplication doesn't make much sense unless we want to introduce more precision to gas prices (i.e right now gas price can not be lower than 1e6 USD)

PR #469, opened 2025-10-15
feat(precompiles): update `LinkingUSD` to allow transfers for auth roles

0xKitsune:
This PR modifies `LinkingUSD` to permit transfers when that both the `to` or `from` accounts meet the required auth conditions.

Ref [tempoxyz/docs/#57](https://github.com/tempoxyz/docs/pull/57)


crates/precompiles/src/contracts/linking_usd.rs
```diff
@@ -20,21 +28,19 @@ impl<'a, S: StorageProvider> LinkingUSD<'a, S> {
...
 
     pub fn transfer(
         &mut self,
         msg_sender: &Address,
         call: ITIP20::transferCall,
     ) -> Result<bool, TIP20Error> {
-        if *msg_sender == STABLECOIN_EXCHANGE_ADDRESS {
+        if self.token.has_role(msg_sender, *TRANSFER_ROLE)
+            || self.token.has_role(&call.to, *RECEIVE_ROLE)
+            || *msg_sender == STABLECOIN_EXCHANGE_ADDRESS
```
fgimenez: very nit, in the ref impl we check first msg.sender, which is just an address comparison, cheaper than token role lookup in terms of gas
```suggestion
        if *msg_sender == STABLECOIN_EXCHANGE_ADDRESS
            || self.token.has_role(msg_sender, *TRANSFER_ROLE)
            || self.token.has_role(&call.to, *RECEIVE_ROLE)
```

crates/precompiles/src/contracts/linking_usd.rs
```diff
@@ -46,7 +52,10 @@ impl<'a, S: StorageProvider> LinkingUSD<'a, S> {
         msg_sender: &Address,
         call: ITIP20::transferFromCall,
     ) -> Result<bool, TIP20Error> {
-        if *msg_sender == STABLECOIN_EXCHANGE_ADDRESS {
+        if self.token.has_role(&call.from, *TRANSFER_ROLE)
```
fgimenez: we check this several times, maybe could be extracted to a helper?

crates/precompiles/src/contracts/linking_usd.rs
```diff
@@ -443,4 +419,286 @@ mod tests {
...
+            spender,
+        });
+
+        assert_eq!(from_balance_after, from_balance_before - amount);
+        assert_eq!(to_balance_after, to_balance_before + amount);
+        assert_eq!(allowance_after, allowance_before - amount);
+    }
+
+    #[test]
+    fn test_transfer_with_memo_with_transfer_role() {
```
fgimenez, line 623: do we need similar tests for `transfer_from_with_memo`?

PR #472, opened 2025-10-15
feat(dev): add localnet just target

Zygimantass:


Justfile
```diff
@@ -19,6 +19,46 @@ build-release binary extra_args="": (build binary "-r " + extra_args)
 build binary extra_args="":
     {{cargo_build_binary}} build {{extra_args}} --bin {{binary}}
 
+[group('localnet')]
+[doc('Generates a genesis file')]
+genesis accounts="1000" output="genesis.json":
+    cargo run --bin tempo-bench --profile maxperf -- generate-genesis --output {{output}} -a {{accounts}}
```
0xKitsune, line 25: nit: Do we need `--profile maxperf` here? I think we should be able to drop this.

Justfile
```diff
@@ -19,6 +19,46 @@ build-release binary extra_args="": (build binary "-r " + extra_args)
...
+    cargo run --bin tempo node \
+                      --chain ./localnet/genesis.json \
+                      --dev \
+                      --dev.block-time 1sec \
+                      --datadir ./localnet/reth \
+                      --http \
+                      --http.addr 0.0.0.0 \
+                      --http.port 8545 \
+                      --http.api all \
+                      --engine.disable-precompile-cache \
```
0xKitsune, line 46: Should we also add `--engine.legacy-state-root`?

Justfile
Zygimantass: yeah, it's faster to build on maxperf and run it than to build on release + run it

Justfile
Zygimantass: yes - good point

PR #470, opened 2025-10-15
chore(precompiles): simplify `StablecoinExchange` logic

0xKitsune:
This PR refactors the stablecoin exchange precompile to simplify the code and make it a bit more readable.

PR #476, opened 2025-10-15
test(precompiles): update `StablecoinExchange` tests and patch implementation

0xKitsune:
Stacks on #470 

This PR updates `StablecoinExchange` and `LinkingUSD` precompile tests and patches various parts of the implementation. Note that tests evaluating precompile dispatched have been stubbed and will be updated in a follow on PR.

### DIFF
--- a/crates/precompiles/src/contracts/stablecoin_exchange/orderbook.rs
+++ b/crates/precompiles/src/contracts/stablecoin_exchange/orderbook.rs
@@ -616,6 +616,31 @@ mod tests {
         assert_eq!(tick_to_price(MAX_TICK), PRICE_SCALE + 2000);
     }
 
+    #[test]
+    fn test_compute_book_key() {
+        let token_a = address!("0x1111111111111111111111111111111111111111");
+        let token_b = address!("0x2222222222222222222222222222222222222222");
+
+        let key_ab = compute_book_key(token_a, token_b);
+        let key_ba = compute_book_key(token_b, token_a);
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/contracts/stablecoin_exchange/orderbook.rs
```diff
@@ -278,15 +273,15 @@ impl Orderbook {
...
+            .to::<u16>() as i16;
 
         let best_ask_tick = storage
             .sload(
                 address,
                 orderbook_slot + offsets::ORDERBOOK_BEST_ASK_TICK_OFFSET,
             )
             .expect("TODO: handle error")
-            .to::<i16>();
+            .to::<u16>() as i16;
```
fgimenez, line 287: can we have a comment about why is this needed?

crates/precompiles/src/contracts/stablecoin_exchange/mod.rs
```diff
@@ -1323,55 +1336,94 @@ impl<'a, S: StorageProvider> StablecoinExchange<'a, S> {
...
 }
 
-#[cfg(test)]
-mod tests {
-    use super::*;
-    use crate::contracts::HashMapStorageProvider;
-    use alloy::primitives::address;
-
-    #[test]
-    fn test_compute_book_key_deterministic() {
```
fgimenez, line 1333: the `compute_book_key` tests are missing in the new location, should we move them?

PR #478, opened 2025-10-15
feat: hook up `StablecoinExchange#createToken` to `call` impl + other small tweaks

jxom:
- Hooks up `StablecoinExchange#createToken` to `call` impl
- Hooks up `StablecoinExchange#pairKey` to `call` impl
- Hooked up `PairCreated` event on precompile
- Added exchange initialization to `genesis.rs`

### DIFF
--- a/bin/tempo-bench/src/cmd/genesis.rs
+++ b/bin/tempo-bench/src/cmd/genesis.rs
@@ -22,7 +22,7 @@ use tempo_contracts::{
 };
 use tempo_evm::evm::{TempoEvm, TempoEvmFactory};
 use tempo_precompiles::{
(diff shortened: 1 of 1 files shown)


bin/tempo-bench/src/cmd/genesis.rs
```diff
@@ -385,6 +389,17 @@ fn initialize_fee_manager(
         .expect("Could not 0x00 validator fee token");
 }
 
+fn initialize_stablecoin_exchange(evm: &mut TempoEvm<CacheDB<EmptyDB>>) -> eyre::Result<()> {
```
0xKitsune, line 392: nit: It looks like we do not need to return `Result` here as there is nothing that can fail within the function.

PR #480, opened 2025-10-15
feat(precompiles): check if pair is already created in`StablecoinExchange::create_pair`

0xKitsune:
This PR introduces checks in `StablecoinExchange::create_pair` to revert if a pair is already created.

### DIFF
--- a/crates/precompiles/src/contracts/stablecoin_exchange/mod.rs
+++ b/crates/precompiles/src/contracts/stablecoin_exchange/mod.rs
@@ -2162,7 +2162,7 @@ mod tests {
     }
 
     #[test]
-    fn test_pair_created_event() {
+    fn test_pair_created() {
         let mut storage = HashMapStorageProvider::new(1);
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/contracts/stablecoin_exchange/mod.rs
```diff
@@ -284,11 +284,16 @@ impl<'a, S: StorageProvider> StablecoinExchange<'a, S> {
             .to::<u128>()
     }
 
-    pub fn create_pair(&mut self, base: &Address) -> B256 {
+    pub fn create_pair(&mut self, base: &Address) -> Result<B256, StablecoinExchangeError> {
         let quote =
             TIP20Token::new(address_to_token_id_unchecked(base), self.storage).linking_token();
 
         let book_key = compute_book_key(*base, quote);
+
```
fgimenez, line 292: would it make sense to add validation here to check that the base token is initialized and has a valid linking token before creating the pair? this would prevent creating pairs with Address::ZERO as the quote token

PR #481, opened 2025-10-16
feat(node): handle epochs

SuperFluffy:
This patch makes the commonware consensus layer fully epoch aware.

It introduces an `orchestrator` actor managing consensus engines on a per-epoch basis. Given a `heights_per_epoch` configuration option, the execution driver (application) actor signals to the orchestrator when the boundary of an era was reached and when a new era was entered.

For a given heights per epoch `H`, all heights `h` are said to belong to an epoch `E` if the satisfy `E*H+1 <= h <= (E+1)*H`. For example, this means that epoch `E=0` contains all heights `h=1` to and including `h=H`, `E=1` contains `h=H+1` to and including `2*H`, and so on.

Because in simplex there is no guarantee to observe a direct finalization of the block at the epoch boundary, spinning up and shutting down a consensus engine for a given epoch is split in two parts. This avoids fractioning the network where some nodes have transitioned to the new epoch while others have been left in the old epoch, and where neither have enough peers to drive consensus:

1. the consensus engine for epoch `E+1` is started when the last height of epoch `E` is finalized.
2. the consensus engine for epoch `E` is stopped when the first height of epoch `E+1` is finalized.

At the epoch boundary (the last height in the epoch), the application will _spin_ on the last block of the epoch: if the application receives a request to propose a new block on top of `parent.digest` and `parent.height == L` (`L` beiing the last height of the epoch), then the application will re-propose `parent` and not ask the execution layer to return a new payload.

The reason for this is that in simplex there is no guarantee of a direct finalization. By re-proposing the last block one can however obtain an indirect finalization via the notarization of a later view, as is the case when re-proposing.

Equivalently, if the application observes a finalization in the new epoch, this means there is now a guarantee that its seed (the last block of the previous epoch) can be indirectly finalized by the other validators.

## Metrics

This patch adds two new gauges:

1. `orchestrator_latest_epoch` that tracks the latest (and thus highest) epoch for which the orchestrator started a consensus engine.
2. `orchestrator_active_epochs` that tracks how many consensus engines are currently active (i.e. how many epochs are backed by an active engine)

## Tests

A new test `transition_three_epochs` was added to `tempo-e2e` to ensure that a validator network indeed enters a new epoch. This test uses the previously defined `orchestrator_latest_epoch` metric as a stop condition.

## Open task / followup

The heights per epoch should not be set in the config but in the chainspec.


crates/e2e/src/tests/epoch.rs
```diff
@@ -0,0 +1,33 @@
...
+
+use crate::{Setup, run};
+
+#[test_traced]
+fn transitions_three_epochs() {
+    let _ = tempo_eyre::install();
+    let linkage = Link {
+        latency: Duration::from_millis(10),
+        jitter: Duration::from_millis(1),
+        success_rate: 1.0,
```
clabby, line 14: Recommend adding a variant of this test with a fallible link (i.e. 90% success.)

crates/commonware-node/src/orchestrator/actor.rs
```diff
@@ -0,0 +1,337 @@
...
+
+        let first = self
+            .epoch_to_engine
+            .first_key_value()
+            .map(|(first, _)| *first)
+            .expect("if there is a last entry, there must be a first entry in the map");
+
+        for epoch in first..last {
+            if let Some(engine) = self.epoch_to_engine.remove(&epoch) {
+                // TODO(janis): this is very not-graceful. Is it ok to just nuke
```
patrick-ogrady: You can just takedown the `engine` ungracefully (once we merge the "task supervision" change in the runtime this guarantees all spawned subtasks will also be aborted) 👍

You may want to delete the consensus data for this epoch (just to keep the ondisk usage in check) after abort.

crates/commonware-node/src/consensus/execution_driver/mod.rs
```diff
@@ -245,8 +255,32 @@ impl Inner<Init> {
...
+            {
+                warn!(%error, "could not signal that the end of an epoch was reached");
+            }
+
+            if is_first_height_of_epoch(finalized.block.height(), epoch, self.heights_per_epoch)
+                && let Err(error) = self.state.orchestrator.epoch_entered(epoch)
+            {
+                warn!(%error, "could not signal that a new epoch was entered");
+            }
+        }
```
clabby: This does work, though if you prefer, you can also use the [`Reporters`](https://docs.rs/commonware-consensus/latest/commonware_consensus/struct.Reporters.html) splitter provided by `commonware-consensus` to have marshal send finalized blocks to both your application as well as the orchestrator.

crates/e2e/src/lib.rs
```diff
@@ -107,13 +108,14 @@ pub fn run(
                 participants: validators.clone(),
                 mailbox_size: 1024,
                 deque_size: 10,
-                leader_timeout: Duration::from_secs(2),
-                notarization_timeout: Duration::from_secs(3),
-                nullify_retry: Duration::from_secs(10),
-                fetch_timeout: Duration::from_secs(2),
-                activity_timeout: 10,
-                skip_timeout: 5,
+                time_to_propose: Duration::from_secs(2),
```
patrick-ogrady, line 111: nit: Just to be crystal clear, the "leader timeout" includes the latency for a leader to enter the same view as other players and to broadcast their proposal to others. You should not aim to take anywhere close to this long just for block building (I don't think you do but felt it was worth clarifying if you were pondering).

crates/commonware-node/src/orchestrator/actor.rs
patrick-ogrady: The only case I could think of where you'd want to handle gracefully is if you are collecting "Activity" and want to ensure you handle all data before deletion.

crates/commonware-node/src/consensus/execution_driver/mod.rs
```diff
@@ -258,12 +292,21 @@ impl Inner<Init> {
...
-        genesis.response.send(genesis_digest).map_err(|_| {
-            eyre!("failed returning genesis block digest: return channel was already closed")
+    async fn handle_genesis(mut self, genesis: Genesis) -> eyre::Result<Digest> {
+        let source = if genesis.epoch == 0 {
+            self.genesis_block.digest()
+        } else {
+            let height = calculate_source_height_of_epoch(genesis.epoch, self.heights_per_epoch);
+            let Some((_, digest)) = self.syncer.get_info(height).await else {
+                // TODO(janis): should we just put the response channel into a map and respond later? Or fail?
+                bail!("no information on the source block at height `{height}` exists yet");
```
clabby: I should improve the example code here, but it is probably best to use marshal::Mailbox::subscribe for this and attempt to respond later rather than bailing.

If the genesis request doesn't come back, the consensus engine will not start.

crates/commonware-node/src/orchestrator/actor.rs
```diff
@@ -0,0 +1,337 @@
...
+            metadata::Config {
+                partition: format!("{}-metadata", self.config.partition_prefix),
+                codec_config: ((), ()),
+            },
+        )
+        .await
+        .expect("failed to initialize orchestrator metadata");
+
+        // Enter the initial epoch
+        let (epoch, seed) = match metadata.get(&METADATA_KEY).cloned() {
```
clabby: `seed` looks unused, can remove; Likely an artifact of the also-unused seed in the `reshare` example.

bg - @brendanchou originally used this value as a random seed for committee selection of the next epoch in https://github.com/commonwarexyz/monorepo/pull/1460, but in the `reshare` example I used the epoch number as the seed for committee selection.

crates/commonware-node/src/orchestrator/actor.rs
```diff
@@ -0,0 +1,337 @@
...
+            .expect("if there is a last entry, there must be a first entry in the map");
+
+        for epoch in first..last {
+            if let Some(engine) = self.epoch_to_engine.remove(&epoch) {
+                // TODO(janis): this is very not-graceful. Is it ok to just nuke
+                // the engine or do we need a cancellation token or some other
+                // mechanism to take down the engine gracefully?
+                info!(epoch, "closing consensus engine for epoch");
+                engine.abort();
+                self.metrics.active_epochs.dec();
```
clabby: Assuming the engine is in a "spinning" state, where online actors are just re-proposing, it should be fine to just nuke the old engine.

Side note - you may observe that on your version of commonware, the engine actually doesn't shut down. There's a WIP PR [here](https://github.com/commonwarexyz/monorepo/pull/1666) + some ongoing runtime updates that will fix this.

crates/commonware-node/src/orchestrator/actor.rs
```diff
@@ -0,0 +1,337 @@
...
+use prometheus_client::metrics::gauge::Gauge;
+use rand::{CryptoRng, Rng};
+use tempo_commonware_node_cryptography::{Digest, PublicKey};
+use tracing::{Level, Span, info, instrument};
+
+use crate::orchestrator::ingress::Activity;
+
+use super::ingress::Message;
+
+const METADATA_KEY: FixedBytes<12> = FixedBytes::new([
```
patrick-ogrady: We often use `U64` for this (and just assign each key a CONST u64).

crates/commonware-node/src/orchestrator/actor.rs
```diff
@@ -0,0 +1,337 @@
...
+                            epoch_boundary_reached.seed,
+                            &mut metadata,
+                            &mut pending_mux,
+                            &mut recovered_mux,
+                            &mut resolver_mux,
+                        )
+                        .await;
+                }
+                Activity::EpochEntered(epoch_entered) => {
+                    let _: Result<_, _> = self.sunset_epoch(cause, epoch_entered.epoch).await;
```
clabby: Not necessary for this PR, but it may be prudent to start a timer after the last height of an epoch is observed but not finalized, to ensure these older engines don't get stuck online.

crates/commonware-node/src/orchestrator/actor.rs
patrick-ogrady: IIUC this `seed` is an artifact of an earlier iteration and can be removed. @clabby will provide detail.

crates/commonware-node/src/consensus/execution_driver/mod.rs
```diff
@@ -386,35 +437,45 @@ impl Inner<Init> {
...
+                "failed getting parent block from syncer; syncer dropped channel before request was fulfilled"
+            ))?;
+
+        debug!(height = parent.height(), "retrieved parent block",);
 
+        // On the last height of the epoch, re-propose the parent instead of
+        // constructing a new one.
+        if is_last_height_of_epoch(parent.height(), round.epoch(), self.heights_per_epoch) {
+            info!("last height of epoch reached; re-proposing parent");
+            return Ok(parent);
```
patrick-ogrady, line 463: In a future revision of the code, we'd like to automatically handle this repropose case for the application.

patrick-ogrady: (The only thing an application could do in this scenario is "mess it up" by not re-proposing, so IMHO we shouldn't expose it).

crates/commonware-node/src/consensus/execution_driver/mod.rs
```diff
@@ -386,35 +437,45 @@ impl Inner<Init> {
...
-                    "failed getting parent block from syncer; syncer dropped channel before request was fulfilled"
-                ))?;
+            .await
+            .map_err(|_| eyre!(
+                "failed getting parent block from syncer; syncer dropped channel before request was fulfilled"
+            ))?;
+
+        debug!(height = parent.height(), "retrieved parent block",);
 
+        // On the last height of the epoch, re-propose the parent instead of
```
patrick-ogrady: I think this would technically be on last height + 1 (we check parent.height() is the last height).

crates/e2e/src/lib.rs
patrick-ogrady: Might be better to call this `time_to_receive_proposal` (better aligned with `time_to_collect_notarizations`)?

crates/commonware-node/src/consensus/execution_driver/mod.rs
```diff
@@ -441,7 +502,7 @@ impl Inner<Init> {
             .and_then(|ret| ret.wrap_err("execution layer rejected request"))
             .wrap_err("failed requesting new payload from the execution layer")?;
 
-        tracing::debug!(
+        debug!(
             timeout_ms = self.new_payload_wait_time.as_millis(),
             "sleeping for payload builder timeout"
         );
```
patrick-ogrady, line 505: May be worth adding that the "sleep + interrupt" call that follows is non-deterministic (as discussed) as soon as there are transactions flowing through the network (just calling out because I've seen you add verbose TODOs/topics to consider elsewhere).

crates/commonware-node/src/consensus/execution_driver/mod.rs
SuperFluffy: I have thought of that, but opted against doing so for 2 reaons:

1. noise and perf: this would mean that the marshaller would send every block over 2 channels, with the orchestrator throwing out almost every block except for the two blocks around the boundary
2. right now, the orchestrator is blissfully unaware of epoch lengths and the app is the single source of truth of epoch boundaries and transitions.

Point 1 could be somewhat alleviated if the marshaller didn't send the full blocks over the channel but just `Arc<Block>` (since the orchestrator need not take ownership of the block).


EDIT: If we relax condition 2 and do make the orchestrator aware of epoch boundaries, then we could make it more resilient: if it receives a finalized block with a height that falls under epoch E but that epoch is not yet running, it will start a consensus engine for E.

Going to think about that.

crates/e2e/src/tests/epoch.rs
SuperFluffy: Good idea, added in  6c66ef0

crates/commonware-node/src/consensus/execution_driver/mod.rs
SuperFluffy: I don't think this is currently possible with `marshal::Mailbox::subscribe` because I'd need the digest/commitment of the block in question (which is also exactly what I am supposed to return to the consensus engine's genesis request).

With that said, I don't think this should be a problem:

1. an epoch transition is triggered by the application receiving the last block at height H of the outgoing epoch from the marshaller (in the form of `Reporter::report`).
2. this means we know the marshaller has the block available and indexed for height H.
3. this means that `get_info(H)` should always succeed.

EDIT: I added this reasoning in e780668

crates/commonware-node/src/epoch/manager/actor.rs
```diff
@@ -0,0 +1,337 @@
...
+    TBlocker: Blocker<PublicKey = PublicKey>,
+    // TODO(janis): are all of these bounds necessary?
+    TContext: Spawner
+        + commonware_runtime::Metrics
+        + Rng
+        + CryptoRng
+        + Clock
+        + governor::clock::Clock
+        + Storage
+        + Network,
```
klkvr, line 51: can we not just use a concrete type here?

crates/commonware-node/src/epoch/manager/actor.rs
```diff
@@ -0,0 +1,337 @@
...
+use super::ingress::Message;
+
+const METADATA_KEY: FixedBytes<12> = FixedBytes::new([
+    b'O', b'R', b'C', b'H', b'E', b'S', b'T', b'R', b'A', b'T', b'O', b'R',
+]);
+
+const REPLAY_BUFFER: NonZeroUsize = NonZeroUsize::new(8 * 1024 * 1024).expect("value is not zero"); // 8MB
+const WRITE_BUFFER: NonZeroUsize = NonZeroUsize::new(1024 * 1024).expect("value is not zero"); // 1MB
+
+pub(crate) struct Actor<TBlocker, TContext> {
```
klkvr, line 31: can we call this `Orchestrator`? also maybe `EpochManager` might be appropriate name here

crates/commonware-node/src/consensus/execution_driver/mod.rs
```diff
@@ -245,8 +255,32 @@ impl Inner<Init> {
...
+                    .epoch_boundary_reached(epoch, finalized.block.digest())
+            {
+                warn!(%error, "could not signal that the end of an epoch was reached");
+            }
+
+            if is_first_height_of_epoch(finalized.block.height(), epoch, self.heights_per_epoch)
+                && let Err(error) = self.state.orchestrator.epoch_entered(epoch)
+            {
+                warn!(%error, "could not signal that a new epoch was entered");
+            }
```
klkvr: will this logic be fine if finalized block validation fails for some reason? e.g node was just started and is still syncing

crates/commonware-node/src/consensus/execution_driver/mod.rs
klkvr: > right now, the orchestrator is blissfully unaware of epoch lengths and the app is the single source of truth of epoch boundaries and transitions

it seems reasonable to make orchestrator aware of epoch lengths?

also to mitigate 1 we could start with only sending number and hash of the finalized block

crates/commonware-node/src/epoch/manager/actor.rs
clabby: It's useful to abstract over the runtime's context to facilitate testing. Commonware has the [`deterministic` runtime](https://docs.rs/commonware-runtime/latest/commonware_runtime/deterministic/index.html) for testing, and a standard [`tokio`-backed runtime](https://docs.rs/commonware-runtime/latest/commonware_runtime/tokio/index.html) for release builds.

klkvr: ah I see, this makes sense. would be nice if there was a way to erase this type by making all bounds dyn-compatible but not sure if that's feasible

SuperFluffy: Main culprit here is `Spawner` requiring `Clone`.

https://docs.rs/commonware-runtime/latest/commonware_runtime/trait.Spawner.html#tymethod.stop

There are some fancy ways allowing you to clone boxed trait objects, but I am indeed not sure if it's worth it: https://github.com/tokio-rs/axum/blob/37a08a17d40690ece6c4c9d9ada9aaf4e5c33296/axum/src/boxed.rs#L81

crates/commonware-node/src/consensus/execution_driver/mod.rs
SuperFluffy: Right now the application considers "syncing" to be a success:
https://github.com/tempoxyz/tempo/blob/bc9c2dae5fb5dd98c4b98f90ab77b434a7eaa464/crates/commonware-node/src/consensus/execution_driver/executor.rs#L348-L353

But even if it isn't, we have a problem: we just drop the return channel -> `<ExecutionDriverMailbox as Reporter>::report` returns `Poll::Ready` immediately -> marshaller sends the next block.

What we'd need to do is to store the response channel somewhere until the execution layer stops syncing (at least for that block) and only then return the ack.

But I am not sure how to do that.

crates/commonware-node/src/epoch/manager/actor.rs
SuperFluffy: I like epoch manager - orchestrator is very non-descript. I would however like to keep the `Actor` name (and also rename the other bits in our system accordingly) to make it clear what the "actor" is and what the "mailbox".

klkvr: imo it's clear enough to have `EpochManager` and `EpochManagerMailbox`

crates/commonware-node/src/consensus/execution_driver/mod.rs
SuperFluffy: I gave this a shot and think I like your suggestions better: the epoch manager is no longer slave to the application but will transition epochs independently of it.


0203c26

crates/commonware-node/src/orchestrator/actor.rs
SuperFluffy: Yupp, noticed this in my tests. I now have this PR poing to your latest HEAD to include https://github.com/commonwarexyz/monorepo/pull/1867

crates/e2e/src/lib.rs
SuperFluffy: Fair point, I will do this in a followup.

crates/commonware-node/src/orchestrator/actor.rs
SuperFluffy: Fair point, I removed all mention of `seed` when talking about the parent/genesis of an epoch.

crates/commonware-node/src/orchestrator/actor.rs
SuperFluffy: Replaced these by `u64` and `into` at the callsite. [6c338f8](https://github.com/tempoxyz/tempo/pull/481/commits/6c338f83f1ae79cf3c0322235b81330ed8116698)

crates/commonware-node/src/orchestrator/actor.rs
SuperFluffy: Removed and all mention of "seed" with it

crates/commonware-node/src/consensus/execution_driver/mod.rs
SuperFluffy: Yupp, that one was confusing. Clarified in 6e7c0f3

crates/commonware-node/src/epoch/manager/actor.rs
SuperFluffy: To be honest, I did the change in another instance and it's confusing me when I look through the code.

At least for now, I would like to maintain my naming convention.

Issue #397, opened 2025-10-03
All reth non-node commands throw an error at the end of running


Zygimantass:
This is a nit, but might be nice to solve - all Reth non-`node` commands (e.g. `tempo-commonware dump-genesis`) throw a consensus task error:

```
consensus task exited with error:
channel closed before consensus-relevant command line args and a handle to the execution node could be received

Caused by:
    channel closed
```

PR #484, opened 2025-10-16
fix: don't fail for non-node commands

klkvr:
Fixes https://github.com/tempoxyz/tempo/issues/397 by immediately exiting consenus thread when any other subcommand is launched

### DIFF
--- a/bin/tempo/src/main.rs
+++ b/bin/tempo/src/main.rs
@@ -22,7 +22,7 @@ use futures::{
     FutureExt as _,
     future::{FusedFuture, pending},
 };
-use reth_ethereum::cli::Cli;
+use reth_ethereum::cli::{Cli, Commands};
 use reth_node_builder::NodeHandle;
 use std::{
     net::{IpAddr, Ipv4Addr, SocketAddr},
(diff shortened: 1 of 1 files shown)


bin/tempo/src/main.rs
```diff
@@ -168,55 +176,54 @@ fn main() -> eyre::Result<()> {
...
-                        let ext = TempoFaucetExt::new(
-                            txpool,
-                            faucet_args.addresses(),
-                            faucet_args.amount(),
-                            faucet_args.provider(),
-                        );
-
-                        ctx.modules.merge_configured(ext.into_rpc())?;
-                    }
+    cli.run_with_components::<TempoNode>(components, async move |builder, args| {
```
klkvr, line 179: only indentation changes here

bin/tempo/src/main.rs
```diff
@@ -83,6 +86,11 @@ fn main() -> eyre::Result<()> {
 
     let shutdown_token_clone = shutdown_token.clone();
     let consensus_handle = thread::spawn(move || {
+        // Exit early if we are not executing `tempo node` command.
+        if !is_node {
+            return Ok(());
+        }
```
mattsse, line 92: makes sense, I think we're okay with always spawning this for convenience

PR #492, opened 2025-10-17
chore: refactor fee calc in deduct_caller with helpers

rakita:
uses `all_mut` so we don't need to fetch values before loading the account.

It uses `calculate_caller_fee` to check the max balance and deducts the balance with the effective fee. It is a nice wrapper used for the mainnet. Additionally, it handles `cfg.is_balance_check_disabled()` correctly.

Tempo specific fee transfer move to the end.

### DIFF
--- a/bin/tempo-bench/src/cmd/genesis.rs
+++ b/bin/tempo-bench/src/cmd/genesis.rs
@@ -290,7 +290,7 @@ fn create_and_mint_token(
                     name: name.into(),
                     symbol: symbol.into(),
                     currency: currency.into(),
-                    linkingToken: LINKING_USD_ADDRESS,
+                    quoteToken: LINKING_USD_ADDRESS,
                     admin,
                 },
             )
--- a/crates/commonware-node/Cargo.toml
+++ b/crates/commonware-node/Cargo.toml
@@ -32,7 +32,7 @@ commonware-cryptography.workspace = true
 commonware-macros.workspace = true
 commonware-p2p.workspace = true
 commonware-resolver.workspace = true
-commonware-runtime.workspace = true
+commonware-runtime = { workspace = true, features = ["external"] }
 commonware-utils.workspace = true
 
 eyre.workspace = true
--- a/crates/commonware-node/src/consensus/execution_driver/mod.rs
+++ b/crates/commonware-node/src/consensus/execution_driver/mod.rs
@@ -650,11 +650,6 @@ async fn verify_block<TContext: Pacer>(
         info!("block's height must be +1 that of the parent but isn't");
         return Ok(false);
     }
-    if block.timestamp() <= parent.timestamp() {
-        info!("block's timestamp must exceed parent's timestamp but doesn't");
-        return Ok(false);
-    }
-
     let block = block.clone().into_inner();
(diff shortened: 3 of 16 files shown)


crates/revm/src/handler.rs
```diff
@@ -438,90 +428,74 @@ where
...
+        // Create storage provider wrapper around journal
+        let internals = EvmInternals::new(journal, &block);
+        let mut storage_provider = EvmStorageProvider::new(internals, cfg.chain_id());
+        let mut fee_manager = TipFeeManager::new(
+            TIP_FEE_MANAGER_ADDRESS,
+            block.beneficiary(),
+            &mut storage_provider,
+        );
+
+        if tx.max_balance_spending().ok() == Some(U256::ZERO) {
```
rakita, line 461: I don't know why we have this check?

klkvr: seems that we want to skip fee collection if there are no fees to charge

0xKitsune: Previously, this check was used to skip collecting fees via the `FeeManager::collect_fee_pre_tx` function since this function does not have checks for if  `amount ==0` and will incur an unnecessary `sstore`

```rust
    pub fn transfer_fee_pre_tx(&mut self, from: &Address, amount: U256) -> Result<(), TIP20Error> {
        let from_balance = self.get_balance(from);
        if amount > from_balance {
            return Err(TIP20Error::insufficient_balance());
        }

        let new_from_balance = from_balance
            .checked_sub(amount)
            .ok_or(TIP20Error::insufficient_balance())?;

        self.set_balance(from, new_from_balance);

        let to_balance = self.get_balance(&TIP_FEE_MANAGER_ADDRESS);
        let new_to_balance = to_balance
            .checked_add(amount)
            .ok_or(TIP20Error::supply_cap_exceeded())?;
        self.set_balance(&TIP_FEE_MANAGER_ADDRESS, new_to_balance);

        Ok(())
    }
```

We could remove this check in the revm handler if we add input validation in `collect_fee_pre_tx` ensuring that `amount > 0`.

rakita: I am more concerned that there is no fee for tx, is this by design?

If I assume that every tx needs to pays fee in terms of the gas spent, we need to deduct max_effective_balance from caller and reimburse it later with unspent gas.

0xKitsune: > I am more concerned that there is no fee for tx

Yeah I think there should always be a fee. This seems to be a residual check from early development when exploring "free" blockspace for payments. IIRC we also explored setting the base fee to 0 before fixing the fee to its current value. 

I think we might be able to remove this check entirely, wdyt @klkvr?

klkvr: we are currently discussing approaches for potentially free transactions from validators https://tempoxyz.slack.com/archives/C094XH2H6NS/p1760670706545979?thread_ts=1760637886.806439&cid=C094XH2H6NS

also I think technically `eth_call`s are mostly free?

rakita: Fine with any, both removing it and staying and can probably can be done in followup PR. In the end this is an optimization.

PR #503, opened 2025-10-20
chore: handler cleanup and revm bump

legion2002:


crates/primitives/src/transaction/account_abstraction.rs
```diff
@@ -130,7 +130,6 @@ pub struct TxAA {
     )]
     pub gas_limit: u64,
 
-    // TODO: What happens if this vec is empty?
     pub calls: Vec<Call>,
```
0xKitsune, line 134: nit: can we add a doc comment for this field?

crates/node/tests/it/account_abstraction.rs
```diff
@@ -196,26 +197,101 @@ async fn verify_tx_not_in_block_via_rpc(
...
+        let wallet = EthereumWallet::from(signer.clone());
+        let provider = ProviderBuilder::new().wallet(wallet).connect_http(http_url);
+
+        (setup, provider, signer, signer_addr)
+    }};
+}
+
+/// Macro to set up P256 test infrastructure with funded account
+/// Creates variables: setup, provider, signing_key, pub_key_x, pub_key_y, signer_addr, funder_signer, funder_addr, chain_id
+macro_rules! setup_test_with_p256_funded_account {
```
0xKitsune: For clarity, why is this a macro vs a function like in other test setups?

crates/node/tests/it/account_abstraction.rs
```diff
@@ -310,22 +386,8 @@ async fn test_aa_basic_transfer_secp256k1() -> eyre::Result<()> {
...
-    let alice_signer = MnemonicBuilder::from_phrase(crate::utils::TEST_MNEMONIC).build()?;
-    let alice_addr = alice_signer.address();
-
-    // Create provider with wallet
-    let wallet = EthereumWallet::from(alice_signer.clone());
-    let provider = ProviderBuilder::new()
-        .wallet(wallet)
-        .connect_http(http_url.clone());
+    let (setup, provider, alice_signer, alice_addr) = setup_test_with_funded_account!();
+    let mut setup = setup;
```
0xKitsune: Looks like this can be simplified to the following: 
```suggestion
    let (mut setup, provider, alice_signer, alice_addr) = setup_test_with_funded_account!();
```

crates/node/tests/it/account_abstraction.rs
```diff
@@ -1447,3 +1432,87 @@ async fn test_aa_fee_payer_tx() -> eyre::Result<()> {
...
+        encoded.len(),
+        encoded[0]
+    );
+
+    // Try to inject transaction - should fail due to empty call batch
+    let result = setup.node.rpc.inject_tx(encoded.clone().into()).await;
+
+    // The transaction should be rejected
+    assert!(
+        result.is_err(),
```
0xKitsune: Can we assert the specific error or that the error contains the expected message?

crates/node/tests/it/account_abstraction.rs
```diff
@@ -1447,3 +1432,87 @@ async fn test_aa_fee_payer_tx() -> eyre::Result<()> {
...
+    if let Err(e) = result {
+        println!("✓ Transaction with empty call batch correctly rejected: {e}");
+
+        // Print the full error chain to see if validation error is buried
+        let mut source = e.source();
+        while let Some(err) = source {
+            println!("  Caused by: {err}");
+            source = err.source();
+        }
+    }
```
0xKitsune: Do we need this if we update to assert the specific error above?

crates/node/tests/it/account_abstraction.rs
legion2002: I was getting large trait bounds with function helpers, so just switching to macros was easier.
But I now read that macros are less manageable, and harder to debug, so updated this to helper functions now.

crates/node/tests/it/account_abstraction.rs
legion2002: Yes, I updated this. The actual empty call error gets wrapped into a tx decode error somewhere, so I was printing the full chain to double check, but removed that now.

crates/node/tests/it/account_abstraction.rs
```diff
@@ -1447,3 +1441,91 @@ async fn test_aa_fee_payer_tx() -> eyre::Result<()> {
...
+        // Empty call batch should fail during decoding/validation
+        let error_msg = e.to_string();
+        assert!(
+            error_msg.contains("decode")
+                || error_msg.contains("empty")
+                || error_msg.contains("call")
+                || error_msg.contains("valid"),
+            "Error should indicate decode/validation failure for empty calls, got: {error_msg}"
+        );
+    }
```
0xKitsune: Since we know this will be an error, lets use `unwrap_err` or similar here.

legion2002: Yup, good point, updated.

PR #510, opened 2025-10-20
chore: add test for a validator joining an already running network 

klkvr:
Adds a test ensuring that if a validator is included into the set but did not participate in the network since genesis they are still able to backfill the chain and start participating in consensus.

I've refactored e2e consensus setup so that validators are created in a separate `setup_validators` fn and can be started lazily

### DIFF
--- a/crates/e2e/src/tests/linkage.rs
+++ b/crates/e2e/src/tests/linkage.rs
@@ -1,14 +1,8 @@
 use std::time::Duration;
 
+use crate::{Setup, run};
 use commonware_macros::test_traced;
 use commonware_p2p::simulated::Link;
-use commonware_runtime::{
-    Clock, Runner as _,
-    deterministic::{self, Runner},
-};
-use reth_ethereum::storage::BlockNumReader;
-
-use crate::{ExecutionRuntime, Setup, run, setup_validators};
 
 #[test_traced]
 fn only_good_links() {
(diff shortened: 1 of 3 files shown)


crates/e2e/src/tests/linkage.rs
```diff
@@ -154,3 +159,48 @@ fn reach_height_20_with_a_few_bad_links() {
 
     std::thread::sleep(Duration::from_secs(1));
 }
+
+#[test_traced]
+fn validator_can_join_later() {
```
SuperFluffy: Let's move this to its own module. This is much more specific than general linkage tests. Maybe backfill or something

crates/e2e/src/tests/linkage.rs
```diff
@@ -154,3 +159,48 @@ fn reach_height_20_with_a_few_bad_links() {
...
+
+        // Start all nodes except the last one
+        let mut last = nodes.pop().unwrap();
+        for node in &mut nodes {
+            node.start().await;
+        }
+
+        // Wait for chain to advance a bit.
+        while nodes[0].node.node.provider.last_block_number().unwrap() < 5 {
+            context.sleep(Duration::from_secs(1)).await;
```
SuperFluffy: I'm always uneasy about explicit sleep in tests.

Can we instead have a callback similar to the success condition in the other tests that allows us to wait until, say, 5 blocks are finalized?

You can do 2 things in that callback: `oneshot.send(())` to hand back control to the test, but also assert that your node has no entries (`validator-{uid}` - put that name into the `ValidatorNode` you define).

output
SuperFluffy, line 1: probably commited by accident

crates/e2e/src/tests/linkage.rs
```diff
@@ -154,3 +159,48 @@ fn reach_height_20_with_a_few_bad_links() {
...
+        }
+
+        assert_eq!(last.node.node.provider.last_block_number().unwrap(), 0);
+
+        // Start the last node.
+        last.start().await;
+
+        // Assert that last node is able to catch up and progress.
+        while last.node.node.provider.last_block_number().unwrap() < 10 {
+            context.sleep(Duration::from_secs(1)).await;
```
SuperFluffy: Similar to above, you can just put the stop condition into the test similar to the other tests - just for your specific later-starter node.

crates/e2e/src/tests/linkage.rs
klkvr: not sure I understand the callback idea, we can't reuse `run` here as it behaves differently

`run` does explicit sleep as well which seems fine to me https://github.com/tempoxyz/tempo/blob/a1f83820c19c5eacb624c5cdaf2a1c0602b38093/crates/e2e/src/lib.rs#L202

PR #511, opened 2025-10-21
chore: add missing txpool arg to `just localnet`

shekhirin:
Without this argument, running `tempo-bench` causes such errors
```console
Failed to send transaction: server returned an error response: error code -32000: gapped-nonce tx from delegated accounts
Failed to send transaction: server returned an error response: error code -32000: in-flight transaction limit reached for delegated accounts
```

Additionally, made it possible to pass a custom profile, features, and args to `just localnet`
1. `maxperf` is very slow to build, `profiling` is good enough for local comparisons
2. `asm-keccak` feature isn't enabled by default


Justfile
```diff
@@ -55,10 +56,23 @@ localnet accounts="1000" reset="true":
...
                       --txpool.basefee-max-size 10000 \
                       --txpool.queued-max-size 10000 \
+                      --txpool.max-new-pending-txs-notifications 10000000 \
+                      --txpool.max-account-slots 500000 \
+                      --txpool.max-pending-txns 10000000000000 \
+                      --txpool.max-new-txns 10000000000000 \
+                      --txpool.disable-transactions-backup \
+                      --txpool.additional-validation-tasks 8 \
+                      --txpool.minimal-protocol-fee 0 \
+                      --txpool.minimum-priority-fee 0 \
```
0xKitsune: Do we need these args?

shekhirin: not really, only `max-accounts-slots` is needed, removed the rest

PR #512, opened 2025-10-21
fix: reject transactions with illiquid fee tokens

### LINKED ISSUE #507: Gas estimation does not fail when fee_token is illiquid on Fee AMM
## Summary
When a transfer is submitted with an illiquid `feeToken`, it's accepted and gets stuck in the mempool. Subsequent writes from that account get stuck behind the stuck transfer. 

Screenshot of discussion in Slack with repro steps and root cause discovery
Image

cc @gorried and @jxom

fgimenez:
Closes #507 

Adds AMM liquidity validation during transaction pool validation to prevent transactions with illiquid fee tokens from getting stuck in the mempool.

### DIFF
--- a/crates/node/tests/it/liquidity.rs
+++ b/crates/node/tests/it/liquidity.rs
@@ -7,7 +7,7 @@ use alloy::{
 use alloy_eips::Encodable2718;
 use alloy_network::TxSignerSync;
 use tempo_contracts::precompiles::IFeeManager::setUserTokenCall;
-use tempo_precompiles::DEFAULT_FEE_TOKEN;
+use tempo_precompiles::{DEFAULT_FEE_TOKEN, LINKING_USD_ADDRESS};
 use tempo_primitives::TxFeeToken;
 
 /// Test block building when FeeAMM pool has insufficient liquidity for payment transactions
@@ -32,10 +32,10 @@ async fn test_block_building_insufficient_fee_amm_liquidity() -> eyre::Result<()
     let payment_token = crate::utils::setup_test_token(provider.clone(), sender_address).await?;
     let payment_token_addr = *payment_token.address();
 
-    // Get validator token address (USDC from genesis)
+    // Get validator token address (LINKING_USD - validators always use this)
     use tempo_contracts::precompiles::ITIPFeeAMM;
     use tempo_precompiles::TIP_FEE_MANAGER_ADDRESS;
-    let validator_token_addr = DEFAULT_FEE_TOKEN;
+    let validator_token_addr = LINKING_USD_ADDRESS;
 
     let fee_amm = ITIPFeeAMM::new(TIP_FEE_MANAGER_ADDRESS, provider.clone());
     let validator_token =
@@ -145,19 +145,18 @@ async fn test_block_building_insufficient_fee_amm_liquidity() -> eyre::Result<()
         .await?;
 
     // Now try to send payment transactions that require fee swaps
-    // With insufficient liquidity, these should be excluded from blocks
+    // With insufficient liquidity, these should be rejected by the transaction pool validator
     let num_payment_txs = 5;
     println!("Sending {num_payment_txs} payment transactions that require fee swaps...");
 
-    let mut transactions_included = 0;
-    let mut transactions_timed_out = 0;
+    let mut transactions_rejected = 0;
 
     for i in 0..num_payment_txs {
         let transfer = payment_token.transfer(sender_address, U256::from((i + 1) as u64));
         match transfer.send().await {
             Ok(pending_tx) => {
                 let tx_num = i + 1;
-                println!("Transaction {tx_num} sent, waiting for receipt...");
+                println!("Transaction {tx_num} unexpectedly sent! Waiting for receipt...");
                 match tokio::time::timeout(
                     std::time::Duration::from_secs(10),
                     pending_tx.get_receipt(),
@@ -166,39 +165,39 @@ async fn test_block_building_insufficient_fee_amm_liquidity() -> eyre::Result<()
                 {
                     Ok(Ok(receipt)) => {
                         let status = receipt.status();
-                        println!("Transaction {tx_num} included with status: {status:?}");
-                        transactions_included += 1;
+                        println!(
+                            "Transaction {tx_num} unexpectedly included with status: {status:?}"
+                        );
                     }
                     Ok(Err(e)) => {
                         println!("Transaction {tx_num} receipt error: {e}");
                     }
                     Err(_) => {
-                        println!("Transaction {tx_num} timed out waiting for receipt");
-                        transactions_timed_out += 1;
-                        break; // Stop trying if we timeout
+                        println!("Transaction {tx_num} timed out");
                     }
                 }
             }
             Err(e) => {
                 let tx_num = i + 1;
-                println!("Transaction {tx_num} failed to send: {e}");
+                let error_msg = e.to_string();
+                println!("Transaction {tx_num} rejected: {e}");
+                // Verify it's rejected due to insufficient liquidity
+                if error_msg.contains("Insufficient liquidity in FeeAMM pool for fee token swap") {
+                    transactions_rejected += 1;
+                }
             }
         }
     }
 
-    println!("Transactions included: {transactions_included}, timed out: {transactions_timed_out}");
+    println!("Transactions rejected due to insufficient liquidity: {transactions_rejected}");
 
-    // Verify that transactions requiring unavailable liquidity were NOT included
+    // Verify that all transactions requiring unavailable liquidity were rejected
     assert_eq!(
-        transactions_included, 0,
-        "Transactions requiring unavailable liquidity should be excluded from blocks"
-    );
-    assert!(
-        transactions_timed_out > 0,
-        "At least one transaction should have timed out (indicating it was excluded)"
+        transactions_rejected, num_payment_txs,
+        "All transactions requiring unavailable liquidity should be rejected by the pool validator"
     );
 
-    println!("Test completed: block building continued without stalling");
+    println!("Test completed: transactions with insufficient AMM liquidity correctly rejected");
 
     Ok(())
 }
--- a/crates/precompiles/src/tip_fee_manager/amm.rs
+++ b/crates/precompiles/src/tip_fee_manager/amm.rs
@@ -17,6 +17,12 @@ pub const SCALE: U256 = uint!(10000_U256);
 pub const SQRT_SCALE: U256 = uint!(100000_U256);
 pub const MIN_LIQUIDITY: U256 = uint!(1000_U256);
 
+/// Computes the output amount for a given input amount using the constant product formula
+/// with a 0.3% fee (M=9970, SCALE=10000).
+pub fn compute_amount_out(amount_in: U256) -> Option<U256> {
+    amount_in.checked_mul(M).map(|p| p / SCALE)
+}
+
 /// Pool structure matching the Solidity implementation
 #[derive(Debug, Clone, Default, Storable)]
 pub struct Pool {
@@ -720,6 +726,30 @@ mod tests {
     };
     use alloy::primitives::{Address, uint};
     use tempo_contracts::precompiles::TIP20Error;
+
+    #[test]
+    fn test_compute_amount_out() {
+        // Test basic calculation: 1000 * 9970 / 10000 = 997
+        let amount_in = U256::from(1000);
+        let amount_out = compute_amount_out(amount_in);
+        assert_eq!(amount_out, Some(U256::from(997)));
+
+        // Test with larger amount
+        let amount_in = U256::from(1_000_000);
+        let amount_out = compute_amount_out(amount_in);
+        assert_eq!(amount_out, Some(U256::from(997_000)));
+
+        // Test with zero
+        let amount_in = U256::ZERO;
+        let amount_out = compute_amount_out(amount_in);
+        assert_eq!(amount_out, Some(U256::ZERO));
+
+        // Test with max value (should not overflow since we divide by SCALE)
+        let amount_in = U256::MAX;
+        let amount_out = compute_amount_out(amount_in);
+        // This should overflow during multiplication
+        assert_eq!(amount_out, None);
+    }
 
     #[test]
     fn test_mint_identical_addresses() {
--- a/crates/revm/src/common.rs
+++ b/crates/revm/src/common.rs
@@ -10,7 +10,7 @@ use tempo_contracts::precompiles::IFeeManager;
 use tempo_precompiles::{
     DEFAULT_FEE_TOKEN, LINKING_USD_ADDRESS, TIP_FEE_MANAGER_ADDRESS,
     storage::slots::mapping_slot,
-    tip_fee_manager,
+    tip_fee_manager::{self, amm::Pool},
     tip20::{self, is_tip20},
 };
 use tempo_primitives::TempoTxEnvelope;
@@ -180,6 +180,31 @@ pub trait TempoStateAccess<T> {
         // Load fee token account to ensure that we can load storage for it.
         self.basic(token)?;
         self.sload(token, balance_slot)
+    }
+
+    /// Returns the FeeAMM pool reserves for the given token pair.
+    fn get_fee_amm_pool(
+        &mut self,
+        user_token: Address,
+        validator_token: Address,
+    ) -> Result<Pool, Self::Error> {
+        let pool_id = tip_fee_manager::amm::PoolKey::new(user_token, validator_token).get_id();
+        let pool_slot = mapping_slot(pool_id, tip_fee_manager::slots::POOLS);
+
+        // Load FeeManager account to ensure we can load storage for it
+        self.basic(TIP_FEE_MANAGER_ADDRESS)?;
+        let pool_data = self.sload(TIP_FEE_MANAGER_ADDRESS, pool_slot)?;
+
+        // Unpack reserves: lower 128 bits = user token, upper 128 bits = validator token
+        const MASK_128: U256 = uint!(0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF_U256);
+        let reserve_user_token: u128 = (pool_data & MASK_128).to();
+        let shifted: U256 = pool_data >> 128;
+        let reserve_validator_token: u128 = shifted.to();
+
+        Ok(Pool {
+            reserve_user_token,
+            reserve_validator_token,
+        })
     }
 }
 
--- a/crates/transaction-pool/src/validator.rs
+++ b/crates/transaction-pool/src/validator.rs
@@ -1,5 +1,6 @@
 use crate::transaction::{TempoPoolTransactionError, TempoPooledTransaction};
 use alloy_consensus::Transaction;
+use alloy_primitives::{Address, U256};
 use reth_chainspec::{ChainSpecProvider, EthereumHardforks};
 use reth_primitives_traits::{
     Block, GotExpected, SealedBlock, transaction::error::InvalidTransactionError,
@@ -9,6 +10,7 @@ use reth_transaction_pool::{
     EthTransactionValidator, PoolTransaction, TransactionOrigin, TransactionValidationOutcome,
     TransactionValidator, error::InvalidPoolTransactionError,
 };
+use tempo_precompiles::{LINKING_USD_ADDRESS, tip_fee_manager::amm::compute_amount_out};
 use tempo_revm::TempoStateAccess;
 
 /// Validator for Tempo transactions.
@@ -101,6 +103,12 @@ where
             }
         }
 
+        // For testnet, we assume all validators use LinkingUSD as their fee token.
+        // We hardcode this assumption here to avoid expensive header lookups on every validation.
+        // TODO: For mainnet, this will need to be updated to use actual validator addresses.
+        const VALIDATOR_FEE_TOKEN: Address = LINKING_USD_ADDRESS;
+
+        // Check user balance first
         let balance = match state_provider.get_token_balance(fee_token, fee_payer) {
             Ok(balance) => balance,
             Err(err) => {
@@ -122,6 +130,52 @@ where
                 )
                 .into(),
             );
+        }
+
+        // Validate AMM liquidity for fee token swap.
+        // Skip liquidity check only for LINKING_USD as it's the primary validator token
+        // and doesn't require a swap. All other tokens (including DEFAULT_FEE_TOKEN)
+        // need sufficient AMM liquidity to swap to LINKING_USD.
+        // Note: There is no fallback mechanism - if a token doesn't have liquidity
+        // against LINKING_USD, the transaction will fail.
+        // TODO: For mainnet, this will need to be updated to check liquidity
+        // against actual validator tokens or implement a more sophisticated validation strategy.
+        if fee_token != VALIDATOR_FEE_TOKEN {
+            let fee_amount = transaction.fee_token_cost();
+
+            // Check liquidity: get pool reserves
+            let pool = match state_provider.get_fee_amm_pool(fee_token, VALIDATOR_FEE_TOKEN) {
+                Ok(pool) => pool,
+                Err(err) => {
+                    return TransactionValidationOutcome::Error(*transaction.hash(), Box::new(err));
+                }
+            };
+
+            // Calculate output: amount_out = amount_in * M / SCALE (M=9970, SCALE=10000)
+            let amount_out = match compute_amount_out(fee_amount) {
+                Some(amount) => amount,
+                None => {
+                    return TransactionValidationOutcome::Error(
+                        *transaction.hash(),
+                        Box::new(std::io::Error::new(
+                            std::io::ErrorKind::InvalidData,
+                            "overflow",
+                        )),
+                    );
+                }
+            };
+
+            // Check if pool has sufficient liquidity for the swap
+            let has_liquidity = amount_out <= U256::from(pool.reserve_validator_token);
+
+            if !has_liquidity {
+                return TransactionValidationOutcome::Invalid(
+                    transaction,
+                    InvalidPoolTransactionError::other(
+                        TempoPoolTransactionError::InsufficientLiquidity,
+                    ),
+                );
+            }
         }
 
         self.inner.validate_one(origin, transaction)
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11757,6 +11757,7 @@ dependencies = [
  "reth-provider",
  "reth-storage-api",
  "reth-transaction-pool",
+ "tempo-precompiles",
  "tempo-primitives",
  "tempo-revm",
  "thiserror 2.0.17",
--- a/crates/transaction-pool/Cargo.toml
+++ b/crates/transaction-pool/Cargo.toml
@@ -14,6 +14,7 @@ workspace = true
 [dependencies]
 tempo-primitives = { workspace = true, features = ["serde", "reth-codec"] }
 tempo-revm = { workspace = true, features = ["reth"] }
+tempo-precompiles.workspace = true
 
 reth-transaction-pool.workspace = true
 reth-chainspec.workspace = true
--- a/crates/transaction-pool/src/transaction.rs
+++ b/crates/transaction-pool/src/transaction.rs
@@ -73,6 +73,9 @@ pub enum TempoPoolTransactionError {
 
     #[error("No fee token preference configured")]
     MissingFeeToken,
+
+    #[error("Insufficient liquidity in FeeAMM pool for fee token swap")]
+    InsufficientLiquidity,
 }
 
 impl PoolTransactionError for TempoPoolTransactionError {
@@ -81,6 +84,7 @@ impl PoolTransactionError for TempoPoolTransactionError {
             Self::ExceedsNonPaymentLimit => false,
             Self::InvalidFeeToken(_) => false,
             Self::MissingFeeToken => false,
+            Self::InsufficientLiquidity => false,
         }
     }
 


crates/node/src/rpc/mod.rs
```diff
@@ -213,8 +312,10 @@ impl<N: FullNodeTypes<Types = TempoNode>> Call for TempoEthApi<N> {
         evm_env: &EvmEnvFor<Self::Evm>,
         tx_env: &TxEnvFor<Self::Evm>,
     ) -> Result<u64, Self::Error> {
-        let fee_token_balance =
-            self.caller_fee_token_allowance(&mut db, tx_env, evm_env.block_env.beneficiary)?;
+        let validator = evm_env.block_env.beneficiary;
+        let fee_token_balance = self.caller_fee_token_allowance(&mut db, tx_env, validator)?;
+
+        self.check_fee_token_liquidity(&mut db, tx_env, validator)?;
```
klkvr: not sure if it makes sense for this validation to happen in `estimateGas`. We can just do it when adding transaction in `TempoTransactionValidator`

also `caller_gas_allowance` will only be called when gas_price > 0 so I'm not sure if it's an appropriate place for the check

fgimenez: yep, the PR scope is just for estimate gas (or send tx without gas limit, which calls estimate_gas too), maybe when we decide how to move forward we can put all the validation in a single place ofc

0xKitsune: > We can just do it when adding transaction in `TempoTransactionValidator`

Yeah I think tx validation may be the best place for this validation rather than in estimate gas. Thinking through the UX, it seems beneficial for a user to still be able to simulate calls/estimate gas even if there is no liquidity in the feeAMM pool but ultimately tx validation should reject the transaction.

This also highlights a clunky UX for fee tokens. If there is no liquidity in the pool and the user has their `feeToken` set in the fee manager contract, they either must wait for liquidity to be provided into the `feeAMM` pool or use a `FeeTokenTx`. We should think through if there is a better way to approach this when feeAMM liquidity is exhausted. 

For now wdyt about moving this to tx validation?

fgimenez: makes total sense, done here https://github.com/tempoxyz/tempo/pull/512/commits/394595cdfa2c299ed414b4160ebfac248505be22 ptal

crates/precompiles/src/contracts/provider.rs
```diff
@@ -49,6 +49,30 @@ pub trait TIPFeeStateProviderExt: StateProvider {
...
+    ///
+    /// Returns true if the AMM pool exists and has liquidity available.
+    fn has_fee_token_liquidity(
+        &self,
+        user_token: Address,
+        validator_token: Address,
+    ) -> ProviderResult<bool> {
+        if user_token == validator_token {
+            return Ok(true);
+        }
```
0xKitsune: nit: technically when this returns true this could be interpreted as there being a pool for the pair, however this is not the case when `user_token == validator_token`. Consider returning an error here and only calling this function if `user_token != validator_token`.

crates/precompiles/src/contracts/provider.rs
```diff
@@ -49,6 +49,30 @@ pub trait TIPFeeStateProviderExt: StateProvider {
...
+
+        let pool_id = PoolKey::new(user_token, validator_token).get_id();
+
+        let pool_slot = mapping_slot(pool_id, tip_fee_manager::amm::slots::POOLS);
+        let pool_data: U256 = self
+            .storage(TIP_FEE_MANAGER_ADDRESS, pool_slot.into())?
+            .unwrap_or_default();
+
+        let reserve_validator_token = (pool_data >> U256::from(128)).saturating_to::<u128>();
+
```
0xKitsune: Maybe we should implement a helper function to get reserves here.

crates/transaction-pool/src/validator.rs
```diff
@@ -19,12 +24,81 @@ pub struct TempoTransactionValidator<Client> {
...
     }
 
+    /// Resolves the user's fee token based on transaction and state
+    fn get_user_fee_token(
+        &self,
+        state_provider: &impl StateProvider,
+        fee_payer: alloy_primitives::Address,
+        validator: alloy_primitives::Address,
+        tx_fee_token: Option<alloy_primitives::Address>,
+    ) -> Result<alloy_primitives::Address, Box<dyn std::error::Error + Send + Sync>> {
```
0xKitsune: Should we return a `TransactionValidationOutcome::Error` here instead of `Box<dyn Error>`?

crates/transaction-pool/src/validator.rs
```diff
@@ -19,12 +24,81 @@ pub struct TempoTransactionValidator<Client> {
...
+            };
+
+            if validator_fee_token.is_zero() {
+                Ok(DEFAULT_FEE_TOKEN)
+            } else {
+                Ok(validator_fee_token)
+            }
+        } else {
+            Ok(user_fee_token)
+        }
```
0xKitsune: We have this logic in a few places throughout the codebase. Can we write a helper function in the `fee_manager` module or somewhere similar and use this directly rather than reimplementing this logic?

crates/transaction-pool/src/validator.rs
```diff
@@ -19,12 +24,81 @@ pub struct TempoTransactionValidator<Client> {
...
+    }
+
+    /// Gets the validator's fee token
+    fn get_validator_fee_token(
```
0xKitsune: Same thing here, can we have a single helper function to get the validator fee token? Lmk if I am overlooking something that makes this challenging/not possible.

crates/transaction-pool/src/validator.rs
```diff
@@ -45,6 +119,79 @@ where
...
+                    transaction,
+                    InvalidTransactionError::InsufficientFunds(
+                        GotExpected {
+                            got: alloy_primitives::U256::ZERO,
+                            expected: alloy_primitives::U256::from(1),
+                        }
+                        .into(),
+                    )
+                    .into(),
+                );
```
0xKitsune, line 177: Can we move this logic into a separate `validate_fee_token_liquidity()` function?

crates/precompiles/src/contracts/provider.rs
fgimenez: good point done

crates/precompiles/src/contracts/provider.rs
fgimenez: ok added `unpack_pool_reserves`

crates/transaction-pool/src/validator.rs
fgimenez: this is no longer needed since we removed the wrapper methods entirely as part of addressing https://github.com/tempoxyz/tempo/pull/512#discussion_r2453189732 the error handling happens inline in validate_one where we construct `TransactionValidationOutcome::Error` directly from the `ProviderError` returned by `get_fee_token` ptal

crates/transaction-pool/src/validator.rs
fgimenez: good point, done

crates/transaction-pool/src/validator.rs
fgimenez: good catch, i've removed both wrapper methods entirely, now we call `state_provider.get_fee_token()` directly in `validate_one` for user fee token and validator fee token. the `get_fee_token` helper in `TIPFeeStateProviderExt` already handles both cases, so the wrapper methods were unnecessary indirection

crates/transaction-pool/src/validator.rs
fgimenez: for sure, done

crates/precompiles/src/provider.rs
```diff
@@ -1,16 +1,73 @@
...
 };
 
+/// Unpacks pool reserves from packed storage.
+///
+/// Pool reserves are stored as a single U256 with:
+/// - Lower 128 bits: reserve_user_token
+/// - Upper 128 bits: reserve_validator_token
+fn unpack_pool_reserves(pool_data: U256) -> (u128, u128) {
+    let reserve_user_token = (pool_data & U256::from(u128::MAX)).saturating_to::<u128>();
+    let reserve_validator_token = (pool_data >> U256::from(128)).saturating_to::<u128>();
```
0xKitsune: nit: can we use consts here instead of `U256::from`?

crates/transaction-pool/src/validator.rs
```diff
@@ -45,6 +88,65 @@ where
...
+            }
+        };
+
+        // Get validator's fee token
+        let validator_fee_token = match state_provider.get_fee_token(validator, validator, None) {
+            Ok(token) => token,
+            Err(err) => {
+                return TransactionValidationOutcome::Error(*transaction.hash(), Box::new(err));
+            }
+        };
```
klkvr: wouldn't this return validator's user token? (if they have one configured)

crates/precompiles/src/provider.rs
```diff
@@ -99,6 +184,97 @@ pub trait TIPFeeDatabaseExt: Database {
 
         Ok(balance)
     }
+
+    /// Check if there's sufficient AMM liquidity to swap fee tokens.
+    ///
+    /// Returns true if the AMM pool exists and has liquidity available.
+    /// This is a basic check to ensure the pool is initialized and can handle swaps.
+    fn has_fee_token_liquidity(
```
klkvr: we really need to find a way to unify those helpers

crates/precompiles/src/provider.rs
```diff
@@ -1,16 +1,73 @@
...
+}
+
 /// Trait to provide [`StateProvider`] access to TIPFeeManager storage to fetch fee token data and balances
 #[cfg(feature = "reth")]
 pub trait TIPFeeStateProviderExt: StateProvider {
+    /// Resolves the fee token for a transaction.
+    ///
+    /// Falls back to validator's token if user has no token configured,
+    /// and to DEFAULT_FEE_TOKEN if validator has no token.
+    fn get_fee_token(
```
klkvr: can we unify the logic below with that? https://github.com/tempoxyz/tempo/blob/fbcf998286eb4a42b6dde34bdce3a89a8e2305a1/crates/precompiles/src/provider.rs#L82-L99

crates/transaction-pool/src/validator.rs
```diff
@@ -45,6 +88,65 @@ where
             }
         };
 
+        // Get the validator (beneficiary) from the latest block header.
+        // Note: This is a heuristic - we don't know which validator will actually include
+        // this transaction in a future block, so we use the current block's validator as an
+        // approximation. This allows us to reject obviously invalid transactions (e.g., those
+        // with illiquid fee tokens) before they enter the mempool.
+        let block_num = match self.inner.client().best_block_number() {
```
klkvr: this might produce false positives if there are validators that accept user's fee token or can swap it into theirs (besides the one at the current block)

this will also produce false negatives when there's not enough liquidity in the corresponding pool (right now we only check if there's 1 wei)

this will probably fix issues with the current setup but I'd prefer if we at least agree on a potential long-term solution for this

crates/transaction-pool/src/validator.rs
fgimenez: good catch, fixed

crates/precompiles/src/provider.rs
fgimenez: sure, done

crates/transaction-pool/src/validator.rs
fgimenez: agree, this might also not work well when the subsequent validator has a different fee token than the previous one. would make more sense to come up with a long-term solution for sure.

crates/node/src/rpc/mod.rs
klkvr, line 1: I believe changes in this file need to be reverted

crates/payload/builder/src/lib.rs
klkvr, line 1: here too

crates/node/src/rpc/mod.rs
fgimenez: ok done

crates/payload/builder/src/lib.rs
fgimenez: i believe we need these, without them i get:
```
error[E0599]: the method `best_transactions_with_attributes` exists for struct `Pool<TransactionValidationTaskExecutor<TempoTransactionValidator<Provider>>, ..., ...>`, but its trait bounds were not satisfied
   --> crates/payload/builder/src/lib.rs:243:36
    |
243 |             |attributes| self.pool.best_transactions_with_attributes(attributes),
    |                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ method cannot be called due to unsatisfied trait bounds
    |
   ::: /home/home-link/.cargo/git/checkouts/reth-e231042ee7db3fb7/1cd5b50/crates/transaction-pool/src/lib.rs:339:1
    |
339 | pub struct Pool<V, T: TransactionOrdering, S> {
    | --------------------------------------------- doesn't satisfy `_: TransactionPool`
    |
   ::: /home/home-link/workspace/paradigm/tempo/crates/transaction-pool/src/validator.rs:59:1
    |
 59 | pub struct TempoTransactionValidator<Client> {
    | -------------------------------------------- doesn't satisfy `_: TransactionValidator`
    |
   ::: /home/home-link/.cargo/git/checkouts/reth-e231042ee7db3fb7/1cd5b50/crates/transaction-pool/src/validate/task.rs:99:1
    |
 99 | pub struct TransactionValidationTaskExecutor<V> {
    | ----------------------------------------------- doesn't satisfy `_: TransactionValidator`
    |
    = note: the following trait bounds were not satisfied:
            `tempo_transaction_pool::validator::TempoTransactionValidator<Provider>: reth_transaction_pool::TransactionValidator`
            which is required by `reth_transaction_pool::Pool<reth_transaction_pool::TransactionValidationTaskExecutor<tempo_transaction_pool::validator::TempoTransactionValidator<Provider>>, reth_transaction_pool::CoinbaseTipOrdering<tempo_transaction_pool::transaction::TempoPooledTransaction>, reth_transaction_pool::blobstore::DiskFileBlobStore>: reth_transaction_pool::TransactionPool`
            `Provider: reth_storage_api::HeaderProvider`
            which is required by `reth_transaction_pool::Pool<reth_transaction_pool::TransactionValidationTaskExecutor<tempo_transaction_pool::validator::TempoTransactionValidator<Provider>>, reth_transaction_pool::CoinbaseTipOrdering<tempo_transaction_pool::transaction::TempoPooledTransaction>, reth_transaction_pool::blobstore::DiskFileBlobStore>: reth_transaction_pool::TransactionPool`
            `reth_transaction_pool::TransactionValidationTaskExecutor<tempo_transaction_pool::validator::TempoTransactionValidator<Provider>>: reth_transaction_pool::TransactionValidator`
            which is required by `reth_transaction_pool::Pool<reth_transaction_pool::TransactionValidationTaskExecutor<tempo_transaction_pool::validator::TempoTransactionValidator<Provider>>, reth_transaction_pool::CoinbaseTipOrdering<tempo_transaction_pool::transaction::TempoPooledTransaction>, reth_transaction_pool::blobstore::DiskFileBlobStore>: reth_transaction_pool::TransactionPool`
    = note: the full name for the type has been written to '/home/home-link/workspace/paradigm/tempo/target/debug/deps/tempo_payload_builder-d756fc6653a2c815.long-type-671122387977680231.txt'
    = note: consider using `--verbose` to print the full type name to the console

error[E0599]: the method `best_transactions_with_attributes` exists for struct `Pool<TransactionValidationTaskExecutor<TempoTransactionValidator<Provider>>, ..., ...>`, but its trait bounds were not satisfied
   --> crates/payload/builder/src/lib.rs:243:36
    |
243 |             |attributes| self.pool.best_transactions_with_attributes(attributes),
    |                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ method cannot be called due to unsatisfied trait bounds
    |
   ::: /home/home-link/.cargo/git/checkouts/reth-e231042ee7db3fb7/1cd5b50/crates/transaction-pool/src/lib.rs:339:1
    |
339 | pub struct Pool<V, T: TransactionOrdering, S> {
    | --------------------------------------------- doesn't satisfy `_: TransactionPool`
    |
   ::: /home/home-link/workspace/paradigm/tempo/crates/transaction-pool/src/validator.rs:59:1
    |
 59 | pub struct TempoTransactionValidator<Client> {
    | -------------------------------------------- doesn't satisfy `_: TransactionValidator`
    |
   ::: /home/home-link/.cargo/git/checkouts/reth-e231042ee7db3fb7/1cd5b50/crates/transaction-pool/src/validate/task.rs:99:1
    |
 99 | pub struct TransactionValidationTaskExecutor<V> {
    | ----------------------------------------------- doesn't satisfy `_: TransactionValidator`
    |
    = note: the following trait bounds were not satisfied:
            `tempo_transaction_pool::validator::TempoTransactionValidator<Provider>: reth_transaction_pool::TransactionValidator`
            which is required by `reth_transaction_pool::Pool<reth_transaction_pool::TransactionValidationTaskExecutor<tempo_transaction_pool::validator::TempoTransactionValidator<Provider>>, reth_transaction_pool::CoinbaseTipOrdering<tempo_transaction_pool::transaction::TempoPooledTransaction>, reth_transaction_pool::blobstore::DiskFileBlobStore>: reth_transaction_pool::TransactionPool`
            `Provider: reth_storage_api::HeaderProvider`
            which is required by `reth_transaction_pool::Pool<reth_transaction_pool::TransactionValidationTaskExecutor<tempo_transaction_pool::validator::TempoTransactionValidator<Provider>>, reth_transaction_pool::CoinbaseTipOrdering<tempo_transaction_pool::transaction::TempoPooledTransaction>, reth_transaction_pool::blobstore::DiskFileBlobStore>: reth_transaction_pool::TransactionPool`
            `reth_transaction_pool::TransactionValidationTaskExecutor<tempo_transaction_pool::validator::TempoTransactionValidator<Provider>>: reth_transaction_pool::TransactionValidator`
            which is required by `reth_transaction_pool::Pool<reth_transaction_pool::TransactionValidationTaskExecutor<tempo_transaction_pool::validator::TempoTransactionValidator<Provider>>, reth_transaction_pool::CoinbaseTipOrdering<tempo_transaction_pool::transaction::TempoPooledTransaction>, reth_transaction_pool::blobstore::DiskFileBlobStore>: reth_transaction_pool::TransactionPool`
    = note: the full name for the type has been written to '/home/home-link/workspace/paradigm/tempo/target/debug/deps/tempo_payload_builder-1bf08758e5376009.long-type-16532105956320651562.txt'
    = note: consider using `--verbose` to print the full type name to the console
```

crates/precompiles/src/provider.rs
0xKitsune: Also flagging that we will need to align this with [687](https://github.com/tempoxyz/tempo/pull/687) once merged.

crates/precompiles/src/provider.rs
```diff
@@ -48,13 +97,84 @@ pub trait TIPFeeStateProviderExt: StateProvider {
...
+        user_token: Address,
+        validator_token: Address,
+    ) -> ProviderResult<bool> {
+        // Return error if tokens are identical - no swap needed, so checking for a pool
+        // doesn't make sense. Caller should ensure tokens differ before calling this.
+        if user_token == validator_token {
+            return Err(ProviderError::other(std::io::Error::new(
+                std::io::ErrorKind::InvalidInput,
+                "has_fee_token_liquidity called with identical tokens",
+            )));
```
0xKitsune: Can we update this to `ProviderError::other(TempoPrecompileError::FeeManagerError(FeeManagerError::insufficient_liquidity()))`

crates/precompiles/src/provider.rs
```diff
@@ -1,44 +1,93 @@
...
+/// Unpacks pool reserves from packed storage.
+///
+/// Pool reserves are stored as a single U256 with:
+/// - Lower 128 bits: reserve_user_token
+/// - Upper 128 bits: reserve_validator_token
+fn unpack_pool_reserves(pool_data: U256) -> (u128, u128) {
+    let reserve_user_token = (pool_data & U256::from(u128::MAX)).saturating_to::<u128>();
+    let reserve_validator_token = (pool_data >> U256::from(128)).saturating_to::<u128>();
+    (reserve_user_token, reserve_validator_token)
+}
```
0xKitsune: Is this function available from `tip_fee_manager`? If not it might make sesen to move this there and import as a helper?

crates/precompiles/src/provider.rs
fgimenez: sure, introduced `LOWER_128_MASK` and `RESERVE_SHIFT`

crates/precompiles/src/provider.rs
fgimenez: sure done

crates/precompiles/src/provider.rs
fgimenez: good call, done

crates/precompiles/src/tip_fee_manager/amm.rs
```diff
@@ -16,6 +16,23 @@ pub const SCALE: U256 = uint!(10000_U256);
...
+/// Unpacks pool reserves from packed storage.
+///
+/// Pool reserves are stored as a single U256 with:
+/// - Lower 128 bits: reserve_user_token
+/// - Upper 128 bits: reserve_validator_token
+pub fn unpack_pool_reserves(pool_data: U256) -> (u128, u128) {
+    let reserve_user_token = (pool_data & LOWER_128_MASK).saturating_to::<u128>();
+    let reserve_validator_token = (pool_data >> RESERVE_SHIFT).saturating_to::<u128>();
+    (reserve_user_token, reserve_validator_token)
+}
```
klkvr: would be nice to unify this with https://github.com/tempoxyz/tempo/blob/8c12335a4a68307d400c334be50f9fb5ee8cd81b/crates/precompiles/src/tip_fee_manager/amm.rs#L113-L124

crates/transaction-pool/src/validator.rs
```diff
@@ -63,49 +101,60 @@ where
...
-            };
-
-            if account.is_none_or(|acc| !acc.has_bytecode()) {
-                return TransactionValidationOutcome::Invalid(
-                    transaction,
-                    InvalidPoolTransactionError::other(TempoPoolTransactionError::InvalidFeeToken(
-                        fee_token,
-                    )),
-                );
+        let validator = header.beneficiary();
```
klkvr: given that the current iteration of PR assumes LinkingUSD as validator fee token I don't think we need to actually know the validator

fetching header on every tx validation is also quite expensive

crates/precompiles/src/provider.rs
```diff
@@ -1,51 +1,89 @@
...
-                    && tip20::is_tip20(*to)
-                {
-                    *to
-                } else {
-                    DEFAULT_FEE_TOKEN
-                }
-            } else {
-                fee_token
-            }
-        };
```
klkvr: was this removed intentionally?

crates/precompiles/src/provider.rs
```diff
@@ -55,13 +93,83 @@ pub trait TIPFeeStateProviderExt: StateProvider {
...
+
+        let pool_id = PoolKey::new(user_token, validator_token).get_id();
+        let pool_slot = mapping_slot(pool_id, tip_fee_manager::amm::slots::POOLS);
+        let pool_data = self
+            .storage(TIP_FEE_MANAGER_ADDRESS, pool_slot.into())?
+            .unwrap_or_default();
+
+        let (_reserve_user_token, reserve_validator_token) = unpack_pool_reserves(pool_data);
+
+        Ok(reserve_validator_token > 0)
```
klkvr: should we compare reserves with actual transaction gas usage?

crates/precompiles/src/provider.rs
```diff
@@ -55,13 +93,83 @@ pub trait TIPFeeStateProviderExt: StateProvider {
...
+        user_token: Address,
+        validator_token: Address,
+    ) -> ProviderResult<bool> {
+        // Return error if tokens are identical - no swap needed, so checking for a pool
+        // doesn't make sense. Caller should ensure tokens differ before calling this.
+        if user_token == validator_token {
+            return Err(ProviderError::other(TempoPrecompileError::FeeManagerError(
+                FeeManagerError::insufficient_liquidity(),
+            )));
+        }
```
klkvr: I'd rather just return true here

crates/transaction-pool/src/validator.rs
```diff
@@ -1,21 +1,59 @@
...
+    TransactionValidator,
 };
+use std::io::{Error as IoError, ErrorKind};
+use tempo_precompiles::{DEFAULT_FEE_TOKEN, provider::TIPFeeStateProviderExt};
 use tempo_primitives::TempoTxEnvelope;
 
+/// Validates that there's sufficient AMM liquidity to swap fee tokens.
+///
+/// Returns None if validation passes, or a ValidationOutcome if it fails.
+fn validate_fee_token_liquidity(
```
klkvr: imo it would be more readable if we'd just directly invoke `state_provider.has_fee_token_liquidity` in the `TempoTransactionValidator::validate_one`

crates/precompiles/src/provider.rs
```diff
@@ -55,13 +93,83 @@ pub trait TIPFeeStateProviderExt: StateProvider {
...
+            self.get_validator_fee_token(validator)
+        } else {
+            Ok(user_fee_token)
+        }
+    }
+
+    /// Gets the validator's configured fee token.
+    ///
+    /// Falls back to DEFAULT_FEE_TOKEN if validator has no token configured.
+    fn get_validator_fee_token(&mut self, validator: Address) -> Result<Address, Self::Error> {
```
klkvr: I don't think those helpers are used right now, so would prefer to not add them just to avoid duplicating this logic even more

crates/precompiles/src/provider.rs
```diff
@@ -109,6 +217,97 @@ pub trait TIPFeeDatabaseExt: Database {
 
         Ok(balance)
     }
+
+    /// Check if there's sufficient AMM liquidity to swap fee tokens.
+    ///
+    /// Returns true if the AMM pool exists and has liquidity available.
+    /// This is a basic check to ensure the pool is initialized and can handle swaps.
+    fn has_fee_token_liquidity(
```
klkvr: same here

crates/precompiles/src/tip_fee_manager/amm.rs
fgimenez: good call done

crates/transaction-pool/src/validator.rs
fgimenez: indeed, left a TODO comment and using now `Address::ZERO` for validators, which will return `DEFAULT_FEE_TOKEN` as the fee token

crates/precompiles/src/provider.rs
fgimenez: nope, was part of a rebase, restored

crates/precompiles/src/provider.rs
fgimenez: totally, we must check if there's enough liquidity, not just any liquidity, pushed changes for this

crates/precompiles/src/provider.rs
fgimenez: good catch fixed

crates/transaction-pool/src/validator.rs
fgimenez: ok, inlined

crates/precompiles/src/provider.rs
fgimenez: ok removed

crates/precompiles/src/provider.rs
fgimenez: done

crates/transaction-pool/src/validator.rs
```diff
@@ -63,48 +60,59 @@ where
...
-        } else if !transaction.inner().is_aa()
-            && fee_payer == transaction.sender()
-            && transaction.inner().kind().to() == Some(&TIP_FEE_MANAGER_ADDRESS)
-            && let Ok(call) = setUserTokenCall::abi_decode(transaction.inner().input())
-        {
-            Some(call.token)
-        } else {
-            None
-        };
+        // For now, we assume all validators use DEFAULT_FEE_TOKEN (LinkingUSD).
```
klkvr: hmm I don't think `DEFAULT_FEE_TOKEN` is LinkingUSD?

https://github.com/tempoxyz/tempo/blob/ffd44f9099931954e836f66e3dde3919f1e43238/crates/contracts/src/precompiles/mod.rs#L23-L24

fgimenez: sure, pushed changes for using the right value here https://github.com/tempoxyz/tempo/pull/512/commits/630b9c7b006138b8c024fe64c9b9f4e495c6c5d9 ptal

crates/transaction-pool/src/validator.rs
```diff
@@ -63,48 +60,66 @@ where
...
+        };
+
+        // Validate AMM liquidity for fee token swap.
+        // Skip liquidity check for system stablecoins (LINKING_USD and DEFAULT_FEE_TOKEN),
+        // as validators will accept both. For other tokens, check if there's sufficient
+        // AMM liquidity to swap to either system stablecoin (validators may use either).
+        // TODO: For mainnet, this will need to be updated to check liquidity
+        // against actual validator tokens or implement a more sophisticated validation strategy.
+        let is_system_stablecoin =
+            user_fee_token == VALIDATOR_FEE_TOKEN || user_fee_token == DEFAULT_FEE_TOKEN;
```
klkvr: not sure this makes sense, why woul we skip the check when user is paying fees in AlphaUSD? which is `DEFAULT_FEE_TOKEN` iiuc

crates/transaction-pool/src/validator.rs
```diff
@@ -63,48 +60,66 @@ where
...
-            let account = match state_provider.basic_account(&fee_token) {
-                Ok(code) => code,
-                Err(err) => {
-                    return TransactionValidationOutcome::Error(*transaction.hash(), Box::new(err));
-                }
-            };
+            // Check liquidity to LINKING_USD (primary validator token)
+            let has_linking_usd_liquidity = state_provider
+                .has_fee_token_liquidity(user_fee_token, VALIDATOR_FEE_TOKEN, fee_amount)
+                .unwrap_or(false);
```
klkvr: instead of `.unwrap_or(false)` we should properly handle errror like e.g here 
https://github.com/tempoxyz/tempo/blob/fc93d6d12d125cf1a176adc92a03db4c20613c65/crates/transaction-pool/src/validator.rs#L105-L114

crates/transaction-pool/src/validator.rs
```diff
@@ -63,48 +60,66 @@ where
...
 
-            if account.is_none_or(|acc| !acc.has_bytecode()) {
+            // Check liquidity to DEFAULT_FEE_TOKEN (fallback validator token)
+            let has_default_token_liquidity = state_provider
+                .has_fee_token_liquidity(user_fee_token, DEFAULT_FEE_TOKEN, fee_amount)
+                .unwrap_or(false);
```
klkvr: hmm I don't think this fallback makes sense right now? we don't have any liquidity fallbacks implemented yet, so if there's no liquidity against linking USD transaction will just fail

crates/transaction-pool/src/validator.rs
fgimenez: indeed, added this to make it easier to adapt tests, updated to only skip linking usd

crates/transaction-pool/src/validator.rs
fgimenez: sure done

crates/transaction-pool/src/validator.rs
fgimenez: yep done

crates/node/tests/assets/test-genesis.json
klkvr, line 1: what's the tldr on why we had to change this? we aren't changing any precompile logic so seems weird that we need to change the genesis state

crates/precompiles/src/provider.rs
```diff
@@ -55,6 +112,72 @@ pub trait TIPFeeStateProviderExt: StateProvider {
...
+        // Get pending swaps to calculate effective reserve
+        let pending_slot = mapping_slot(pool_id, tip_fee_manager::amm::slots::PENDING_FEE_SWAP_IN);
+        let pending_fee_swap_in = self
+            .storage(TIP_FEE_MANAGER_ADDRESS, pending_slot.into())?
+            .unwrap_or_default();
+
+        let pending_out = pending_fee_swap_in
+            .checked_mul(M)
+            .and_then(|product| product.checked_div(SCALE))
+            .ok_or(ProviderError::other(TempoPrecompileError::under_overflow()))?;
```
klkvr: I don't think we ened to account for this here as pending swaps are only non-zero mid-block

crates/transaction-pool/src/validator.rs
```diff
@@ -63,59 +60,154 @@ where
...
+
+        // Validate AMM liquidity for fee token swap.
+        // Skip liquidity check only for LINKING_USD as it's the primary validator token
+        // and doesn't require a swap. All other tokens (including DEFAULT_FEE_TOKEN)
+        // need sufficient AMM liquidity to swap to LINKING_USD.
+        // Note: There is no fallback mechanism - if a token doesn't have liquidity
+        // against LINKING_USD, the transaction will fail.
+        // TODO: For mainnet, this will need to be updated to check liquidity
+        // against actual validator tokens or implement a more sophisticated validation strategy.
+        if user_fee_token != VALIDATOR_FEE_TOKEN {
```
klkvr: the logic here lgtm

crates/node/tests/assets/test-genesis.json
fgimenez: yep, test genesis file was being regenerated from the genesis generator tool with the changes in https://github.com/tempoxyz/tempo/pull/781 to include linking usd pools (now required with the checks in this PR). I've extracted these changes to a separate PR https://github.com/tempoxyz/tempo/pull/791 and rebased here

crates/precompiles/src/provider.rs
fgimenez: great catch removed

crates/node/tests/it/tip_fee_manager.rs
```diff
@@ -60,10 +60,25 @@ async fn test_set_user_token() -> eyre::Result<()> {
...
+        .mintWithValidatorToken(
+            *user_token.address(),
+            *validator_token.address(),
+            U256::from(1e8),
+            user_address,
+        )
+        .send()
+        .await?
+        .get_receipt()
+        .await?;
```
klkvr: I don't think we need that extra mint here

let's revert functional changes to the test

crates/transaction-pool/src/validator.rs
```diff
@@ -101,6 +106,138 @@ where
...
+            };
+
+            let pending_fee_swap_in = match state_provider
+                .storage(TIP_FEE_MANAGER_ADDRESS, pending_slot.into())
+            {
+                Ok(data) => data.unwrap_or_default(),
+                Err(err) => {
+                    return TransactionValidationOutcome::Error(*transaction.hash(), Box::new(err));
+                }
+            };
```
klkvr: it doesn't make much sense to track pending fee swaps here because they would always be zero

crates/transaction-pool/src/validator.rs
```diff
@@ -242,6 +407,85 @@ mod tests {
         TempoPooledTransaction::new(tx.try_into_recovered().unwrap())
     }
 
+    #[tokio::test]
+    async fn test_insufficient_balance() {
```
klkvr: I've refactored this test away into https://github.com/tempoxyz/tempo/blob/262b5318a71dd645307064464223bf414ddb0f3b/crates/node/tests/it/pool.rs#L74

want to avoid doing those weird state manipulations here given that we already have proper node setup available in node/tests/it

crates/transaction-pool/src/validator.rs
```diff
@@ -202,6 +339,28 @@ where
...
+    buf[32..].copy_from_slice(&mapping_slot.to_be_bytes::<32>());
+    U256::from_be_bytes(keccak256(buf).0)
+}
+
+/// Helper function to compute pool ID from token pair.
+/// Computes: keccak256(abi.encode(user_token, validator_token))
+#[inline]
+fn compute_pool_id(user_token: Address, validator_token: Address) -> B256 {
+    keccak256((user_token, validator_token).abi_encode())
+}
```
klkvr: can we import this from `tempo-precompiles`?

crates/transaction-pool/src/validator.rs
```diff
@@ -68,6 +72,7 @@ where
             }
         };
 
+        // Get fee token using main's approach
```
klkvr:
```suggestion
        // Get fee token
```

crates/transaction-pool/src/validator.rs
```diff
@@ -101,6 +106,138 @@ where
...
+        // TODO: For mainnet, this will need to be updated to check liquidity
+        // against actual validator tokens or implement a more sophisticated validation strategy.
+        if fee_token != VALIDATOR_FEE_TOKEN {
+            // Calculate the fee amount using the same logic as execution
+            // Use max_fee_per_gas as an upper bound since we don't know the actual basefee yet
+            // This is conservative - we check for more liquidity than might be needed,
+            // but ensures the transaction won't fail during execution due to insufficient liquidity
+            let gas_limit = transaction.gas_limit();
+            let max_fee_per_gas = transaction.max_fee_per_gas();
+            let fee_amount = calc_gas_balance_spending(gas_limit, max_fee_per_gas);
```
klkvr:
```suggestion
            let fee_amount = tx.fee_token_cost();
```

crates/transaction-pool/src/validator.rs
```diff
@@ -101,6 +106,138 @@ where
...
+            let pool_id = compute_pool_id(fee_token, VALIDATOR_FEE_TOKEN);
+            // Storage layout after #[contract] macro changes:
+            // slot 0: validator_tokens
+            // slot 1: user_tokens
+            // slot 2: collected_fees
+            // slot 3: tokens_with_fees
+            // slot 4: token_in_fees_array
+            // slot 5: pools
+            // slot 6: pending_fee_swap_in
+            const POOLS_SLOT: U256 = uint!(5_U256);
```
klkvr: this should be imported as `tempo-precompiles` constant

crates/node/tests/it/liquidity.rs
```diff
@@ -166,39 +165,41 @@ async fn test_block_building_insufficient_fee_amm_liquidity() -> eyre::Result<()
...
                 }
             }
             Err(e) => {
                 let tx_num = i + 1;
-                println!("Transaction {tx_num} failed to send: {e}");
+                let error_msg = e.to_string();
+                println!("Transaction {tx_num} rejected: {e}");
+                // Verify it's rejected due to insufficient liquidity (which manifests as insufficient funds)
+                if error_msg.contains("insufficient funds")
+                    || error_msg.contains("insufficient liquidity")
```
klkvr: hmm given that this test is supposed to assert txs failing due to lack of liquidity specifically this condition seems strange? should be just `insufficient liquidity`

crates/transaction-pool/src/validator.rs
fgimenez: indeed, done

crates/transaction-pool/src/validator.rs
fgimenez: cool removed

crates/transaction-pool/src/validator.rs
fgimenez: sure done

crates/transaction-pool/src/validator.rs
fgimenez: done

crates/node/tests/it/liquidity.rs
fgimenez: yeah, the problem is that the transaction pool validator is currently using reth's InvalidTransactionError, and there is no pool liquidity variant there. i've updated the test to just check the `InvalidTransactionError::InsuficientFunds` and dropped a note, lmk if we should change this

crates/node/tests/it/tip_fee_manager.rs
fgimenez: yep done

crates/node/tests/it/liquidity.rs
```diff
@@ -166,39 +165,41 @@ async fn test_block_building_insufficient_fee_amm_liquidity() -> eyre::Result<()
...
                 }
             }
             Err(e) => {
                 let tx_num = i + 1;
-                println!("Transaction {tx_num} failed to send: {e}");
+                let error_msg = e.to_string();
+                println!("Transaction {tx_num} rejected: {e}");
+                // Verify it's rejected due to insufficient liquidity
+                // Note: In the pool validator, this manifests as InsufficientFunds since the pool
+                // validator uses reth's error types which don't have a specific liquidity variant
```
klkvr: we have this enum which we can use here https://github.com/tempoxyz/tempo/blob/8fb10e8aa649552a627fdbdc42bdeddf0d50221e/crates/transaction-pool/src/transaction.rs#L66-L76

https://github.com/tempoxyz/tempo/blob/8fb10e8aa649552a627fdbdc42bdeddf0d50221e/crates/transaction-pool/src/validator.rs#L91-L96

crates/transaction-pool/src/validator.rs
```diff
@@ -101,6 +107,76 @@ where
...
+                }
+            };
+
+            // Unpack reserves: lower 128 bits = user token, upper 128 bits = validator token
+            // The pool_id encoding always puts user_token first, validator_token second,
+            // so the reserves are always in this fixed order regardless of address comparison.
+            const LOWER_128_MASK: U256 = uint!(0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF_U256);
+            const RESERVE_SHIFT: usize = 128;
+            let _reserve_user_token = (pool_data & LOWER_128_MASK).saturating_to::<u128>();
+            let reserve_validator_token = (pool_data >> RESERVE_SHIFT).saturating_to::<u128>();
```
klkvr:
```suggestion
            // Unpack reserves: lower 128 bits = user token, upper 128 bits = validator token
            // The pool_id encoding always puts user_token first, validator_token second,
            // so the reserves are always in this fixed order regardless of address comparison.
            let reserve_validator_token = (pool_data >> 128).to::<u128>();
```

crates/transaction-pool/src/validator.rs
```diff
@@ -101,6 +107,76 @@ where
...
+
+            // Calculate output: amount_out = amount_in * M / SCALE (M=9970, SCALE=10000)
+            const M: U256 = uint!(9970_U256);
+            const SCALE: U256 = uint!(10000_U256);
```
klkvr: those consts are available here https://github.com/tempoxyz/tempo/blob/8fb10e8aa649552a627fdbdc42bdeddf0d50221e/crates/precompiles/src/tip_fee_manager/amm.rs#L14-L16

maybe we should add a helper `fn compute_amout_out(amount_in: U256)` to tempo-precompiles

crates/node/tests/it/liquidity.rs
fgimenez: ok introduced `TempoPoolTransactionError::InsufficientLiquidity`

crates/transaction-pool/src/validator.rs
```diff
@@ -101,6 +107,73 @@ where
...
+            let pool_id = PoolKey::new(fee_token, VALIDATOR_FEE_TOKEN).get_id();
+            let pool_slot = mapping_slot(pool_id, POOLS);
+
+            let pool_data = match state_provider.storage(TIP_FEE_MANAGER_ADDRESS, pool_slot.into())
+            {
+                Ok(data) => data.unwrap_or_default(),
+                Err(err) => {
+                    return TransactionValidationOutcome::Error(*transaction.hash(), Box::new(err));
+                }
+            };
```
0xKitsune: It might be nice to `provider.get_fee_amm_pool(token_a, token_b)` function here.

crates/transaction-pool/src/validator.rs
```diff
@@ -101,6 +107,73 @@ where
...
+
+            // Unpack reserves: lower 128 bits = user token, upper 128 bits = validator token
+            // The pool_id encoding always puts user_token first, validator_token second,
+            // so the reserves are always in this fixed order regardless of address comparison.
+            let reserve_validator_token = (pool_data >> 128).to::<u128>();
```
0xKitsune: It might be nice to `provider.get_fee_amm_pool(token_a, token_b)` function here that returns a `Pool` type rather than having to manually unpack reserves here.

crates/transaction-pool/src/validator.rs
0xKitsune: Yeah lets use the existing constants, otherwise if these ever change this could be accidentially overlooked. 

Agreed that a helper here would be nice.

fgimenez: sure, helper added

crates/transaction-pool/src/validator.rs
fgimenez: good call done

crates/precompiles/src/tip_fee_manager/amm.rs
```diff
@@ -17,6 +17,12 @@ pub const SCALE: U256 = uint!(10000_U256);
 pub const SQRT_SCALE: U256 = uint!(100000_U256);
 pub const MIN_LIQUIDITY: U256 = uint!(1000_U256);
 
+/// Computes the output amount for a given input amount using the constant product formula
+/// with a 0.3% fee (M=9970, SCALE=10000).
+pub fn compute_amount_out(amount_in: U256) -> Option<U256> {
+    amount_in.checked_mul(M).and_then(|p| p.checked_div(SCALE))
```
klkvr:
```suggestion
    amount_in.checked_mul(M).map(|p| p / SCALE)
```

crates/transaction-pool/src/validator.rs
fgimenez: yep done as part of https://github.com/tempoxyz/tempo/pull/512#discussion_r2528008215

crates/precompiles/src/tip_fee_manager/amm.rs
klkvr: we can also use that in FeeAMM code now

crates/revm/src/common.rs
```diff
@@ -181,6 +181,31 @@ pub trait TempoStateAccess<T> {
...
+
+    /// Returns the FeeAMM pool reserves for the given token pair.
+    ///
+    /// Returns (reserve_user_token, reserve_validator_token) where reserves are packed
+    /// in a single U256 storage slot (lower 128 bits = user token, upper 128 bits = validator token).
+    fn get_fee_amm_pool(
+        &mut self,
+        user_token: Address,
+        validator_token: Address,
+    ) -> Result<(u128, u128), Self::Error> {
```
klkvr: probably better if this would return a `Pool` 
https://github.com/tempoxyz/tempo/blob/fee3ef9fa3a38066fda400ccadda628d8190f80b/crates/precompiles/src/tip_fee_manager/amm.rs#L20-L26

crates/transaction-pool/src/validator.rs
```diff
@@ -101,6 +103,12 @@ where
             }
         }
 
+        // For testnet, we assume all validators use LinkingUSD as their fee token.
+        // We hardcode this assumption here to avoid expensive header lookups on every validation.
+        // TODO: For mainnet, this will need to be updated to use actual validator addresses.
+        const VALIDATOR_FEE_TOKEN: Address = LINKING_USD_ADDRESS;
```
klkvr, line 109: I guess we need to either hold this off or set this to `DEFAULT_FEE_TOKEN`

crates/revm/src/common.rs
fgimenez: sure done

PR #518, opened 2025-10-21
feat: `TempoTransactionRequest::calls`

klkvr:
Adds a way to specify `calls` field to `TempoTransactionRequest` and trigger AA codepath

### DIFF
--- a/crates/node/src/rpc/request.rs
+++ b/crates/node/src/rpc/request.rs
@@ -33,6 +33,7 @@ pub struct TempoTransactionRequest {
     pub fee_token: Option<Address>,
 
     /// Optional calls array, for AA transactions.
+    #[serde(default)]
     pub calls: Vec<Call>,
 }
 


crates/node/src/rpc/request.rs
```diff
@@ -15,16 +15,25 @@ use reth_rpc_convert::{
...
 pub struct TempoTransactionRequest {
+    /// Inner [`TransactionRequest`]
     #[serde(flatten)]
     pub inner: TransactionRequest,
+
+    /// Optional fee token preference
     pub fee_token: Option<Address>,
+
+    /// Optional calls array, for AA transactions.
+    pub calls: Vec<Call>,
```
mattsse, line 37: I think this field is now mandatory?

should we make this `Option`?

I prefer Option over `default` because then you can differentiate between missing and empty

klkvr: ah yep good catch

I've thought about Vec vs Option and it seemed like there are no cases when we would want to handle missing and empty differently but it allows to simplify code in some places, so imo it should be fine to keep it as is for now

PR #520, opened 2025-10-21
fix: Intrinsic gas calculations for AA txs

### LINKED ISSUE #491: TxAA initial gas calculation
Few additional gas calculations were not taken into account:
* Gas for call's len. In initial gas calculation we should take additional `call.len() * AA_CALL_GAS`
  * I am unsure what value to use, but it needs to account for cold load of account of 2900gas + some static gas for bytes increase (Not sure how much, it shouldn't be a big number)
* Additional gas for call inputs is not measured. Would imagine this is calculated similar to ordinary tx input, token per ones and zeros multiplied by constant.  ([Ref](https://github.com/bluealloy/revm/blob/8f185ecab59191f9b8349a8b7449a14adf67589d/crates/interpreter/src/gas/calc.rs#L402))
* if Value is != 0 and TxKind::Call, additional gas for the account change should be added. It is 9000gas rn for CALL opcode.
* if it is TxKind::Create check limit of input ([Ref](https://github.com/bluealloy/revm/blob/8f185ecab59191f9b8349a8b7449a14adf67589d/crates/interpreter/src/gas/calc.rs#L425C28-L425C41)) and charge additional gas for analysis. 
* If it is TxKind::Create charge additional gas for contract creation: [Ref](https://github.com/bluealloy/revm/blob/8f185ecab59191f9b8349a8b7449a14adf67589d/crates/interpreter/src/gas/calc.rs#L414)

Calculate additional floor_gas by adding all inputs len and few bytes per code.len() so we could get a new floor. Gas floor is a ddos prevention of abusing inputs as DA without using CPU (aka gas for execution).  

All of this gas accounting should be done and deducted before call loop starts. `execute_single` should have zero initial gas, as initial gas was already deducted. [This](https://github.com/tempoxyz/tempo/blob/33cfd19d39d9030fb18347ef225986c2244c9a43/crates/revm/src/handler.rs#L183-L192) needs to be removed as it should already be deducted.

We should recheck these gas numbers above, but at least it should be in ballpark.

legion2002:
Resolves #491

### DIFF
--- a/crates/revm/src/handler.rs
+++ b/crates/revm/src/handler.rs
@@ -616,7 +616,7 @@ fn calculate_aa_batch_intrinsic_gas<'a>(
         if call.to.is_create() {
             // CREATE costs 32000 additional gas
             gas.initial_gas += CREATE; // 32000 gas
-            
+
             // EIP-3860: Initcode analysis gas using revm helper
             gas.initial_gas += initcode_cost(call.input.len());
         }
@@ -642,10 +642,8 @@ fn calculate_aa_batch_intrinsic_gas<'a>(
         gas.initial_gas += storages * ACCESS_LIST_STORAGE_KEY; // 1900 per storage
     }
 
-    // 6. Floor gas (EIP-7623, Prague+) using revm helper
-    if spec.is_enabled_in(RevmSpecId::PRAGUE) {
-        gas.floor_gas = calc_tx_floor_cost(total_tokens); // tokens * 10 + 21000
-    }
+    // 6. Floor gas  using revm helper
+    gas.floor_gas = calc_tx_floor_cost(total_tokens); // tokens * 10 + 21000
 
     // TODO: Add 7702 gas costs here, once we implement it.
 
@@ -707,10 +705,7 @@ where
     }
 
     // Validate floor gas (Prague+)
-    if spec.is_enabled_in(RevmSpecId::PRAGUE)
-        && !evm.ctx.cfg.is_eip7623_disabled()
-        && gas_limit < batch_gas.floor_gas
-    {
+    if !evm.ctx.cfg.is_eip7623_disabled() && gas_limit < batch_gas.floor_gas {
         return Err(TempoInvalidTransaction::InsufficientGasForIntrinsicCost {
             gas_limit,
             intrinsic_gas: batch_gas.floor_gas,


crates/revm/src/handler.rs
```diff
@@ -591,82 +543,189 @@ where
...
+                spec.is_enabled_in(RevmSpecId::ISTANBUL),
+            );
+            P256_VERIFY_GAS + tokens * STANDARD_TOKEN_COST
+        }
+    };
+    gas.initial_gas += signature_gas;
+
+    // 3. Per-call overhead: cold account access
+    // TODO: Maybe we should optimize this for the user, and only charge cold access cost
+    // if the `to` address has not appeared in the call batch before.
```
rakita, line 604: Intristric gas should cover the worst case and be simple/fast to run (it needs to be run in txpool for ddos prevention). Maybe we can optimise it when the calls start executing to add back the cold gas calc if the account is already warm

crates/revm/src/handler.rs
```diff
@@ -591,82 +543,189 @@ where
...
+    gas.initial_gas += COLD_ACCOUNT_ACCESS_COST * calls.len() as u64;
```
rakita, line 605: This looks fine, data is already accounted for so only value and type is needed.

In EIP-2930 https://eips.ethereum.org/EIPS/eip-2930#parameters access list item for he ccount is 2400gas
While gas for cold account load is 2600gas (Not sure why it is more :D).

Just to say accounting here is not precise, and very hard to decide what is correct, the most important part is that we are in ballpark.

crates/revm/src/handler.rs
```diff
@@ -591,82 +543,189 @@ where
...
+            // EIP-3860: Initcode analysis gas (Shanghai+) using revm helper
+            if spec.is_enabled_in(RevmSpecId::SHANGHAI) {
+                gas.initial_gas += initcode_cost(call.input.len());
+            }
+        }
+
+        // 4c. Value transfer cost using revm constant
+        if !call.value.is_zero() && !call.to.is_create() {
+            gas.initial_gas += CALLVALUE; // 9000 gas
+        }
```
```suggestion
            // CREATE costs 32000 additional gas
            gas.initial_gas += CREATE; // 32000 gas
            
            // EIP-3860: Initcode analysis gas using revm helper
            gas.initial_gas += initcode_cost(call.input.len());
        }

        // 4c. Value transfer cost using revm constant
        if !call.value.is_zero() && call.to.is_call() {
            gas.initial_gas += CALLVALUE; // 9000 gas
        }
```

nit, we dont need spec check here

crates/revm/src/handler.rs
```diff
@@ -591,82 +543,189 @@ where
...
         }
         .into());
     }
 
-    Ok(InitialAndFloorGas::new(
-        total_intrinsic_gas,
-        standard_gas.floor_gas,
-    ))
+    // Validate floor gas (Prague+)
+    if spec.is_enabled_in(RevmSpecId::PRAGUE)
```
rakita: nit, we dont need spec check as it will be always enabled

crates/revm/src/handler.rs
```diff
@@ -591,82 +543,189 @@ where
...
+/// - Base 21k stipend (once for the transaction)
+/// - Signature verification gas (P256: 5k, WebAuthn: 5k + webauthn_data)
+/// - Per-call account access cost (COLD_ACCOUNT_ACCESS_COST * calls.len())
+/// - Per-call input data gas (calldata tokens * 4 gas)
+/// - Per-call CREATE costs (if applicable):
+///   - Additional 32k base (CREATE constant)
+///   - Initcode analysis gas (2 per 32-byte chunk, Shanghai+)
+/// - Per-call value transfer cost (9k if value > 0 and TxKind::Call)
+/// - Access list costs (shared across batch)
+/// - Floor gas calculation (EIP-7623, Prague+)
```
rakita, line 575: We should probably put this in docs, it is very readable and nice to follow comment

legion2002: Yup, will do

Issue #508, opened 2025-10-20
Refactor precompiles


0xKitsune:
## Overview
Tempo’s precompiles enable core protocol features (e.g. stablecoin gas payments, fee AMM, account abstraction) and are injected at `TempoEvm` initialization. 

The current implementation is functional but exposes too many low-level EVM storage details and has inconsistent error handling. This makes precompiles harder to reason about, extend, and test. The precompiles should be refactored  to abstract away EVM storage footguns and unify error propagation so that precompile implementations are safer, easier to test and easier to build on. Also note that this refactor should ensure minimal dependencies for Foundry compatibility.

### EVM Storage Abstractions
- Provide a clean abstraction for specifying storage variables for a given contract.
- Create macros to auto implement functions to read/write to specific storage variables. 
- Implement traits for generic types generated from `sol!` bindings to define methods for reading/writing to EVM storage.
- Implement helper functions for mappings, arrays and structs to emulate solidity builtins (eg. `insert`, `delete`, `push`).

### Error Propagation
- All precompile functions exposed via the contract's public interface should return `Result<T, PrecompileError>`.  Ensure that errors from storage ops, internal precompile functions, or other calls issues propagate back to the caller.
- Replace all `unwrap` and `excepts` with explicit error handling.

PR #521, opened 2025-10-22
chore(precompiles): restructure precompiles

0xKitsune:
Ref #508 

This PR restructures the precompile implementation to improve clarity and maintainability. Each precompile now lives in its own directory, containing a `bindings.rs` file for contract bindings, a `dispatch.rs` file that handles routing to the correct precompile functions when the public interface is invoked, and additional modules for the core precompile logic. This PR also begins updating error handling, but this PR focuses on structural reorganization to keep the diff smaller between refactoring PRs.

### DIFF
--- a/crates/precompiles/src/stablecoin_exchange/mod.rs
+++ b/crates/precompiles/src/stablecoin_exchange/mod.rs
@@ -562,7 +562,8 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
             current_order_id += 1;
         }
 
-        self.set_active_order_id(pending_order_id);
+        self.set_active_order_id(pending_order_id)
+            .expect("TODO: handle error");
 
         Ok(())
     }
@@ -645,10 +646,10 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
         order.update_remaining(new_remaining, self.storage, self.address);
 
         if order.is_bid() {
-            self.increment_balance(order.maker(), orderbook.base, fill_amount);
+            self.increment_balance(order.maker(), orderbook.base, fill_amount)?;
         } else {
             let quote_amount = (fill_amount * price as u128) / orderbook::PRICE_SCALE as u128;
-            self.increment_balance(order.maker(), orderbook.quote, quote_amount);
+            self.increment_balance(order.maker(), orderbook.quote, quote_amount)?;
         }
 
         let amount_out = if order.is_bid() {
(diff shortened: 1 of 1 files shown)


bin/tempo-bench/src/cmd/genesis.rs
```diff
@@ -363,7 +367,7 @@ fn initialize_fee_manager(
     let block = evm.block.clone();
 
     let evm_internals = EvmInternals::new(evm.journal_mut(), &block);
-    let mut provider = EvmStorageProvider::new(evm_internals, 1);
+    let mut provider = EvmPrecompileStorageProvider::new(evm_internals, 1);
```
mattsse, line 367: @klkvr wondering how we could move access to the chainid into evminternals, we could maybe just copy it?
adding `Box<dyn Cfg>` isnt feasible due to the specid AT.

but adding chain_id would be nice to have imo

crates/precompiles/src/nonce/mod.rs
```diff
@@ -1,11 +1,16 @@
-use crate::contracts::{StorageProvider, types::INonce};
+pub mod bindings;
```
mattsse: should we simply re-export everything from bindings? then we don't need to use `use <precompile>::bindings::{}` 

if we group everything under `<precompile>::` this is sufficient imo

bin/tempo-bench/src/cmd/genesis.rs
klkvr: yep we can just put `chain_id` there, both copying and referencing it should be possible

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -51,62 +53,56 @@ impl<'a, S: StorageProvider> StablecoinExchange<'a, S> {
...
+    fn get_active_order_id(&mut self) -> Result<u128, PrecompileError> {
+        Ok(self
+            .storage
+            .sload(self.address, slots::ACTIVE_ORDER_ID)?
+            .to::<u128>())
     }
 
     /// Set active order ID
-    fn set_active_order_id(&mut self, order_id: u128) {
+    fn set_active_order_id(&mut self, order_id: u128) -> Result<(), PrecompileError> {
```
klkvr, line 87: I think eventually we might want our own error type for this, revm's `PrecompileError` is somewhat limited and only allows to express halts

crates/precompiles/src/tip20_factory/dispatch.rs
```diff
@@ -0,0 +1,34 @@
...
+            ITIP20Factory::createTokenCall::SELECTOR => {
+                mutate::<ITIP20Factory::createTokenCall, _>(calldata, msg_sender, |s, call| {
+                    self.create_token(s, call)
+                })
+            }
+            _ => Err(PrecompileError::Other(
+                "Unknown function selector".to_string(),
+            )),
+        }
+    }
```
onbjerg, line 33: you can actually simplify this quite a bit, as `sol!` generates an enum that has `abi_decode` on it which will automatically handle the selector & parameter decoding we do manually now, which surely will have knockon effects on `mutate`/`view` etc (goes for the other dispatch modules as well).

0xKitsune: Sounds good! I think we can do this in a follow-on when cleaning up the rest of the precompile implementation.

PR #533, opened 2025-10-22
feat: add multi-hop trading support to stablecoin exchange

### LINKED ISSUE #522: Implement multi-hop trades
ref https://tempoxyz.slack.com/archives/C09KCGR4LQ4/p1761069305765029?thread_ts=1761069171.674129&cid=C09KCGR4LQ4

* Have the swap() and quote() functions support trades where the two tokens are not directly paired against each other.
  * The DEX should find the shortest path between the tokens (through quote tokens), by finding their lowest common ancestor on the liquidity tree.
  * These intermediate trades should not involve transfers to the user or updates to their stored balance on the DEX; the intermediate balances should be entirely transitory.
* No need to make any changes to the place() or cancel() functions.

fgimenez:
Closes #522

Adds multi-hop trades using lowest-common-ancestor-based path-finding through the liquidity tree. Users can swap between any two tokens in a single transaction (e.g., USDC -> LinkingUSD -> EURC), with intermediate balances remaining transitory, never persisted to wallet or exchange balances.

### DIFF
--- a/crates/precompiles/src/stablecoin_exchange/mod.rs
+++ b/crates/precompiles/src/stablecoin_exchange/mod.rs
@@ -1234,17 +1234,11 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
             return Err(StablecoinExchangeError::identical_tokens());
         }
 
-        // Check if direct pair exists (token_in -> token_out)
-        let token_in_quote = TIP20Token::from_address(token_in, self.storage).quote_token();
+        // Check if direct or reverse pair exists
+        let in_quote = TIP20Token::from_address(token_in, self.storage).quote_token();
+        let out_quote = TIP20Token::from_address(token_out, self.storage).quote_token();
 
-        if token_in_quote == token_out {
-            return self.validate_and_build_route(&[token_in, token_out]);
-        }
-
-        // Check if reverse pair exists (token_out -> token_in)
-        let token_out_quote = TIP20Token::from_address(token_out, self.storage).quote_token();
-
-        if token_out_quote == token_in {
+        if in_quote == token_out || out_quote == token_in {
             return self.validate_and_build_route(&[token_in, token_out]);
         }
 


crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -299,19 +344,39 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
         amount_out: u128,
         max_amount_in: u128,
     ) -> Result<u128, StablecoinExchangeError> {
-        let book_key = compute_book_key(token_in, token_out);
-        let orderbook = Orderbook::from_storage(book_key, self.storage, self.address);
+        let path = self.find_trade_path(token_in, token_out)?;
```
0xKitsune: I think we probably want to return the book keys along the path from the `get_route` function. IMO this function should also handle if the pair doesnt exist and return an error.

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -1210,6 +1275,102 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
...
+            path.push(current);
+
+            // Check if we've reached LinkingUSD
+            if current == LINKING_USD_ADDRESS {
+                break;
+            }
+
+            // Get the quote token for the current token
+            let token_id = address_to_token_id_unchecked(&current);
+            let quote_token = TIP20Token::new(token_id, self.storage).quote_token();
```
0xKitsune: This pattern exists frequently throughout the codebase. Maybe it makes sense to add a `TIP20Token::from_address()` function (or similar)?

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -1210,6 +1275,102 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
...
+    /// Find the trade path between two tokens
+    /// Returns a vector of addresses representing the trading route
+    fn find_trade_path(
+        &mut self,
+        token_in: Address,
+        token_out: Address,
+    ) -> Result<Vec<Address>, StablecoinExchangeError> {
+        // Cannot trade same token
+        if token_in == token_out {
+            return Err(StablecoinExchangeError::pair_does_not_exist());
```
0xKitsune: Can we return an error indicating that the tokens are identical rather than the pair doesnt exist?

crates/precompiles/src/stablecoin_exchange/mod.rs
fgimenez: makes sense, now we precompute book keys and validate pairs in `find_trade_pair`

crates/precompiles/src/stablecoin_exchange/mod.rs
fgimenez: sure, added `TIP20Token::from_address` and refactored

crates/precompiles/src/stablecoin_exchange/mod.rs
fgimenez: yep, much better, done

crates/precompiles/src/tip20/mod.rs
```diff
@@ -756,6 +755,12 @@ impl<'a, S: PrecompileStorageProvider> TIP20Token<'a, S> {
         }
     }
 
+    /// Create a TIP20Token from an address
+    pub fn from_address(address: Address, storage: &'a mut S) -> Self {
+        let token_id = address_to_token_id_unchecked(&address);
```
0xKitsune, line 760: Should we have input validation ensuring that the address is a valid tip20 token address?

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -272,19 +274,29 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
...
+        for (book_key, base_for_quote) in route {
+            // Fill orders for this hop - no min check on intermediate hops
+            current_amount =
+                self.fill_orders_exact_in(book_key, base_for_quote, current_amount, 0)?;
         }
 
-        let base_for_quote = token_in == orderbook.base;
-        let amount_out =
-            self.fill_orders_exact_in(book_key, base_for_quote, amount_in, min_amount_out)?;
+        let amount_out = current_amount;
```
0xKitsune: nit: Are we binding to a new variable for readability? If so consider updating `current_amount` to `amount` and comparing `amount < min_amount_out` or similar.

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -1210,6 +1229,128 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
...
+            if current == LINKING_USD_ADDRESS {
+                break;
+            }
+
+            // Get the quote token for the current token
+            let quote_token = TIP20Token::from_address(current, self.storage).quote_token();
+
+            // Move to the parent (quote token)
+            current = quote_token;
+        }
```
0xKitsune: nit: consider simplifying to something like the following:

```rust
    let mut path = vec![];

    while token != LINKING_USD_ADDRESS {
        path.push(token);
        token = TIP20Token::from_address(token, self.storage).quote_token();
    }
```

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -1210,6 +1229,128 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
...
+        if token_in_quote == token_out {
+            return self.validate_and_build_route(&[token_in, token_out]);
+        }
+
+        // Check if reverse pair exists (token_out -> token_in)
+        let token_out_quote = TIP20Token::from_address(token_out, self.storage).quote_token();
+
+        if token_out_quote == token_in {
+            return self.validate_and_build_route(&[token_in, token_out]);
+        }
```
0xKitsune: Consider simplifying to the following or similar.

```rust
let in_quote = TIP20Token::from_address(token_in, self.storage).quote_token();
let out_quote = TIP20Token::from_address(token_out, self.storage).quote_token();

if in_quote == token_out || out_quote == token_in {
    return self.validate_and_build_route(&[token_in, token_out]);
}
```

crates/precompiles/src/tip20/mod.rs
fgimenez: not sure, we are replacing a pattern without validation here, just refactoring existing code. i believe we should keep the current behavior and don't vaidate, wdyt?

crates/precompiles/src/stablecoin_exchange/mod.rs
fgimenez: good catch, no need for this binding, removed

crates/precompiles/src/stablecoin_exchange/mod.rs
fgimenez: makes sense, done with a small change to include LinkingUSD in the path

crates/precompiles/src/stablecoin_exchange/mod.rs
fgimenez: much better, done

PR #534, opened 2025-10-22
fix(rpc): eth_estimateGas issue

### LINKED ISSUE #516: Native AA: `eth_estimateGas` returning insufficient gas values for non-secp256k1 keys
## Summary

When calling `eth_estimateGas` for AA transactions with a webauthn key, the RPC method does not support passing a key type parameter. This causes the gas estimation to return lower values than what is actually required to execute the transaction as it is estimating the gas with a "secp256k1 key", leading to transactions that fail due to insufficient gas.

## Steps to Reproduce

The following RPC call demonstrates the issue:

```bash
cast rpc eth_estimateGas \
  '{"from":"0x0cc33662d67a318304231ee06e8948500cf52590","nonce":"0x2","type":"0x76","chainId":"0xa5bb","calls":[{"to":"0x20c0000000000000000000000000000000000001","value":"0x","data":"0xa9059cbb0000000000000000000000000cc33662d67a318304231ee06e8948500cf5259000000000000000000000000000000000000000000000000000000000000f4240"}]}' \
  --rpc-url https://eng:zealous-mayer@rpc-andante.tempoxyz.dev/

> "0xd221"
```

This estimate is insufficient for the actual transaction execution.

legion2002:
Resolves #516

### DIFF
--- a/crates/node/tests/it/account_abstraction.rs
+++ b/crates/node/tests/it/account_abstraction.rs
@@ -1589,7 +1589,7 @@ async fn test_aa_estimate_gas_with_key_types() -> eyre::Result<()> {
     tx_request_webauthn
         .as_object_mut()
         .unwrap()
-        .insert("keyType".to_string(), serde_json::json!("webauthn"));
+        .insert("keyType".to_string(), serde_json::json!("webAuthn"));
     tx_request_webauthn
         .as_object_mut()
         .unwrap()


crates/node/tests/it/account_abstraction.rs
```diff
@@ -1522,3 +1523,102 @@ async fn test_aa_empty_call_batch_should_fail() -> eyre::Result<()> {
...
+        "nonce": "0x0",
+        "nonceKey": "0x0",
+        "type": "0x76",
+        "chainId": format!("0x{:x}", chain_id),
+        "calls": [{
+            "to": recipient.to_string(),
+            "value": "0x0",
+            "input": {
+                "input": "0x"
+            }
```
klkvr: oh that's a bug in serde impl

legion2002: Yeah, think so. test doesn't work in any other format

crates/node/src/rpc/request.rs
```diff
@@ -202,22 +227,117 @@ impl TryIntoTxEnv<TempoTxEnv, TempoBlockEnv> for TempoTransactionRequest {
...
+            // key_data contains the total size of webauthn_data (excluding 128 bytes for public keys)
+            // Default: 200 bytes if no key_data provided
+
+            // Base clientDataJSON template (50 bytes): {"type":"webauthn.get","challenge":"","origin":""}
+            // Authenticator data (37 bytes): 32 rpIdHash + 1 flags + 4 signCount
+            // Minimum total: 87 bytes
+            const BASE_CLIENT_JSON: &str = r#"{"type":"webauthn.get","challenge":"","origin":""}"#;
+            const AUTH_DATA_SIZE: usize = 37;
+            const MIN_WEBAUTHN_SIZE: usize = AUTH_DATA_SIZE + BASE_CLIENT_JSON.len(); // 87 bytes
+            const DEFAULT_WEBAUTHN_SIZE: usize = 200; // Default when no key_data provided
```
jxom:
```suggestion
            const DEFAULT_WEBAUTHN_SIZE: usize = 800; // Default when no key_data provided
```

crates/node/src/rpc/request.rs
```diff
@@ -21,6 +21,17 @@ use tempo_primitives::{
...
 
+/// Key type for AA transaction signature verification.
+/// Used in gas estimation to calculate accurate signature verification costs.
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[serde(rename_all = "lowercase")]
+// TODO: Check if this can be unified
+pub enum KeyType {
+    Secp256k1,
+    P256,
+    WebAuthn,
```
jxom: I believe the serialized version of this would be `"webauthn"`, which would be inconsistent with `"webAuthn"` used in other places. is it meant to be `camelCase` instead of `lowercase`?

klkvr: yep fixed in 3d053b2

PR #538, opened 2025-10-23
feat(consensus): on chain DKG

SuperFluffy:
Adds on-chain distributed key generation.

This patch adds two actors to the consensus engine:

1. the epoch manager, that is entering and exiting epochs by starting/stopping consensus engines for a specific epoch.
2. the dkg manager, that is running DKG ceremonies.

Building on top of https://github.com/tempoxyz/tempo/pull/481, the epoch manager has been significantly simplified. Because in order to transition epochs one needs to know the new set of participants (validator identities) as well as the public polynomial and the private key share for the new epoch, the epoch manager is now directly slaved to the dkg manager.

# New config options

Adds a new config option `epoch_length`, the number of heights in an epoch.

# New metrics and changed metrics

+ `epoch_manager_latest_epoch` that tracks the latest (and thus highest) epoch for which the orchestrator started a consensus engine.
+ `epoch_manager_active_epochs` that tracks how many consensus engines are currently active (i.e. how many epochs are backed by an active engine)
+ `epoch_manager_latest_participants` that tracks how many participants (i.e. validators) make up the latest epoch
+ `dkg_manager_ceremony_failures` that tracks how many DKG ceremonies failed
+ `dkg_manager_ceremony_successes` that tracks how many DKG ceremonies succeeded
+ `dkg_manager_ceremony_dealers` that tracks how many dealers participate in the active ceremony
+ `dkg_manager_ceremony_players` that tracks how many players participate in the active ceremony

All metrics relating to the commonware marshal actor have been changed from `*_syncer_*` to `*_marshal_*`.

# Ceremony details

The DKG manager is running DKG ceremonies in the following way:

1. First, the ceremony is started by creating a new commitment and set of private key shares.
2. In the first half of the epoch, the DKG manager sends the key shares to all players for acknowledgment. The DKG manager collects these acknowledgments.
3. At the epoch midpoint, the DKG manager constructs the outcome of its dealings (its commitment, the acknowledgments it received, and the key shares for which it did not receive acks and which it now reveals).
4. During the second half of the epoch, every time the node is asked to propose, it attempts to fetch the dealing outcomes from its DKG manager to write it to the `extraData` field of its proposal header.
5. Concurrently, in the second half of the epoch the DKG manager reads these dealing outcomes from the `extraData` field of each block header.
6. If the DKG manager finds its own dealing outcome in a block, it removes it so that the node does not write it to a block again (note that there is a chance that it finds the same dealing outcomes twice because of finalization delays).
7. On the pre-to-last block of an epoch, the DKG manager generates the overall outcome of the ceremony so it can be written to the last block of the epoch. It also already starts the ceremony for the next epoch here and requests the epoch manager to start a new epoch.
8. If a node is the proposer on the last block of the epoch, it writes this ceremony outcome to the `extraData` header of its proposal. If the ceremony outcome is not yet ready, it refuses to propose.
9. If a node is the validator on the last block of the epoch, it verifies that the last block indeed contains the same ceremony outcome.

### DIFF
--- a/crates/commonware-node/src/consensus/execution_driver/mod.rs
+++ b/crates/commonware-node/src/consensus/execution_driver/mod.rs
@@ -344,11 +344,7 @@ impl Inner<Init> {
         )?;
 
         if proposal.digest() != parent_digest {
-            if epoch::is_last_height_of_epoch(
-                proposal.height(),
-                round.epoch(),
-                self.epoch_length,
-            ) {
+            if epoch::is_last_height_of_epoch(proposal.height(), round.epoch(), self.epoch_length) {
                 let outcome = self
                     .state
                     .dkg_manager
@@ -607,8 +603,7 @@ impl Inner<Init> {
         // immediately, and happen very rarely. It's better to optimize for the
         // general case.
         if payload == parent_digest {
-            if epoch::is_last_height_of_epoch(block.height(), round.epoch(), self.epoch_length)
-            {
+            if epoch::is_last_height_of_epoch(block.height(), round.epoch(), self.epoch_length) {
                 return Ok((block, true));
             } else {
                 return Ok((block, false));
--- a/crates/commonware-node/src/dkg/manager/actor.rs
+++ b/crates/commonware-node/src/dkg/manager/actor.rs
@@ -327,8 +327,7 @@ where
         TSender: Sender<PublicKey = PublicKey>,
     {
         // Special case height == 0
-        let Some(block_epoch) = epoch::of_height(block.height(), self.config.epoch_length)
-        else {
+        let Some(block_epoch) = epoch::of_height(block.height(), self.config.epoch_length) else {
             return ceremony;
         };
 
@@ -400,10 +399,7 @@ where
         // stored on chain.
         //
         // This starts a new ceremony.
-        if epoch::is_last_height(
-            block.height().saturating_add(1),
-            self.config.epoch_length,
-        ) {
+        if epoch::is_last_height(block.height().saturating_add(1), self.config.epoch_length) {
             info!("on pre-to-last height of epoch; finalizing ceremony");
 
             let next_epoch = ceremony.epoch().saturating_add(1);
--- a/crates/e2e/src/tests/dkg.rs
+++ b/crates/e2e/src/tests/dkg.rs
@@ -21,7 +21,7 @@ fn transitions_with_perfect_links() {
         how_many: 3,
         seed: 0,
         linkage,
-        epoch_length: epoch_length,
+        epoch_length,
     };
 
     let mut epoch_reached = false;


crates/e2e/src/tests/epoch.rs
SuperFluffy, line 1: Removed because testing epoch transitions no longer works with on chain dkg (observing this takes too long and we can't artificially derease the epoch lenght).

crates/commonware-node/src/consensus/block.rs
```diff
@@ -24,6 +29,33 @@ use crate::consensus::Digest;
 pub(crate) struct Block(SealedBlock<tempo_primitives::Block>);
 
 impl Block {
+    pub(crate) fn insert_ceremony_deal_outcome(self, deal: DealingOutcome) -> Self {
+        let sealed = self.into_inner();
+        let (mut header, body) = sealed.split_header_body();
+        // XXX: extra step via vec because commonware writes bytes::Bytes, but
+        // alloy expects alloy_primitives::Bytes
+        header.inner.extra_data = deal.encode().freeze().to_vec().into();
+        Self(SealedBlock::seal_parts(header, body))
```
klkvr, line 38: hmm this should be fine for now but might eventually become problematic because from reth pov the built block is different

crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -0,0 +1,401 @@
...
+            .take()
+            .expect("there must be a ceremony active at all times");
+
+        if let Some(block_epoch) = epoch::of_height(block.height(), self.config.heights_per_epoch)
+            && block_epoch == this_ceremony.epoch()
+        {
+            if epoch::is_first_height(block.height(), self.config.heights_per_epoch) {
+                // Notify the epoch manager that the first height of the new epoch
+                // was entered and the previous epoch can be exited.
+                self.config
```
klkvr: are epoch transitions now fully handled in DKG manager?

crates/commonware-node/src/dkg/ceremony/mod.rs
```diff
@@ -0,0 +1,809 @@
...
+                continue;
+            }
+
+            let share = dealer_me
+                .shares
+                .get(player)
+                .cloned()
+                .expect("invariant violated: all players must have an entry in the shares map");
+
+            if let Some(player) = &mut self.player_me {
```
klkvr: should we not only enter this when the loop `player` (the loop item) is us?

crates/commonware-node/src/consensus/execution_driver/mod.rs
```diff
@@ -339,16 +343,56 @@ impl Inner<Init> {
...
+                    .wrap_err(
+                        "failed getting public dkg ceremony outcome; refusing to generate proposal",
+                    )?;
+                proposal = proposal.insert_public_ceremony_outcome(outcome);
+            } else {
+                match self
+                    .state
+                    .dkg_manager
+                    .get_intermediate_dealing(request.round.epoch())
+                    .await
```
klkvr, line 366: would it be nicer if were just asking `dkg_manager` for an `Option<Bytes>`? that way we could encapsulate more epoch/height based decisions in it?

crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -0,0 +1,401 @@
...
+            let cause = message.cause;
+            match message.command {
+                super::Command::GetIntermediateDealing(get_ceremony_deal) => {
+                    let _: Result<_, _> = self
+                        .handle_get_intermediate_dealing(
+                            cause,
+                            get_ceremony_deal,
+                            ceremony
+                                .as_mut()
+                                .expect("there must be a ceremony active at all times"),
```
klkvr: any reason why `ceremony` has to be an `Option`? also seems like it could be easier if it eas just a field on `self`?

crates/commonware-node/src/dkg/ceremony/mod.rs
```diff
@@ -0,0 +1,809 @@
...
+            arbiter,
+            ceremony_metadata,
+            receiver,
+            sender,
+        })
+    }
+
+    /// Requests players to sign acks.
+    #[instrument(skip_all, fields(epoch = self.config.epoch), err)]
+    pub(super) async fn request_acks(&mut self) -> eyre::Result<()> {
```
klkvr: would `distribute_shares` be a better name for this?

crates/commonware-node/src/consensus/block.rs
SuperFluffy: I was wondering about this - what is the downside? And doesn't any consensus layer (no matter if tempo or actual eth beacon) change the block if it updates the header? I was thinking that a post-build modification of the block header is a normal operation, but probably it isn't?

Related to this - I think reth does not really cache the block either. In `handle_propose`, right after the execution driver receives the block from reth and forwards it to the consensus engine, it sends an FCU back to reth, updating `head_hash = proposal.hash`. But the response is always `SYNCING`. This leads me to believe there is no connection between payload building and caching blocks?

crates/commonware-node/src/dkg/manager/actor.rs
SuperFluffy: Yes. The epoch manager is still responsible for spinning a consensus engine up and down, but indeed `dkg-manager ---> (transition epoch) --> epoch-manager`.

This is because the dkg manager generates the items that are necessary to do the transition.

crates/commonware-node/src/dkg/ceremony/mod.rs
SuperFluffy: I am not sure I fully understand the question, so please correct me if the following doesn't answer it:

`request_acks` (will change to `distribute_shares`) is only executed if we are a dealer (see the condition `let Some(dealer_me) = self.dealer_me else {return}`). But in every round we can be a dealer, or player, or both. And so if we are both dealer and player, we don't need to send the dealings to ourselves over p2p.

crates/commonware-node/src/consensus/execution_driver/mod.rs
SuperFluffy: So the execution driver just places opaque objects in it? Yeah, that would definitely be possible. I was thinking of going that route, but I felt like passing the types around is a bit easier to understand.

Would you like to see this in this PR or in a followup? Even if we change it to opaque bytes, we'd still need to change the read- and write-paths in lockstep.

There are 2 open questions:

1. how do we version this so that this can evolve in time (as you mention, encapsulating more information in there)?
2. should this be RLP encoded? Does that matter?

crates/commonware-node/src/dkg/manager/actor.rs
SuperFluffy: Yes, unfortunately.  We have `Ceremony::finalize(self)` since we also have `Arbiter::finalize(self)`, and so we need to take the ceremony "out of the box" and put it back in. I could of course abstract these things away by having a `structure Ceremony { inner: Option<CeremonyImpl>, }` and enforce this invariant (== that a ceremony must always be set), but I can't really get around it.

EDIT: Another issue is that the cermony gets sub-p2p-channels and so is parameterized over `TSender` and `TReceiver`. But these are only passed into the dkg actor on `dkg::manager::Actor::run` and are hence not known on `Actor::init`. I could also work around this by doing something like this:

```rust
impl Actor<TContext> {
    fn run(self,
        (sender, receiver): (
            impl Sender<PublicKey = PublicKey>,
            impl Receiver<PublicKey = PublicKey>,
        ),
    ) {
        self.into_running(sender, receiver).await.run();
    }
}

struct Running<TContext, TReceiver, TSender> {
    // all actor fields
    ceremony: Option<Ceremony<TContext, TReceiver, TSender>>,
}
```
I guess this is a tiny bit cleaner?

crates/commonware-node/src/dkg/ceremony/mod.rs
SuperFluffy: Yes, it would! 🙏   I will change it.

SuperFluffy: [e513915](https://github.com/tempoxyz/tempo/pull/538/commits/e5139157168d3de4dcc844dad2da568e85698e31)

crates/commonware-node/src/dkg/ceremony/mod.rs
SuperFluffy: EDIT: I think I realize where my confusion came from - you are right, we should only do this when we are the player. I found this during testing and pushing it while you were reviewing: [72acc3d](https://github.com/tempoxyz/tempo/pull/538/commits/72acc3d8c9e8d618007a599bc0494986be48f80f)

This now asserts that player key is us.

crates/commonware-node/src/dkg/manager/actor.rs
klkvr: for `Arbiter::finalize` I'd say we can just make it store `Option<Arbiter>` internally?

for sender/receiver generics it's indeed kind of tricky because adding those to `Actor` would leak all the way up to the `Engine` which'd be super ugly

imo proper solution here is to erase `impl Sender`/`impl Receiver` types so that we can only store something like `pub type BoxedReceiver = Box<dyn Receiver<PublicKey = ed25519::PublicKey, Error = UnifiedP2PError>>`. Though I think this is out of scope for this PR and ideally should be powered by some helpers on Commonware side

just getting rid of `Option` would still be nice imo

crates/commonware-node/src/consensus/execution_driver/mod.rs
klkvr: > Would you like to see this in this PR or in a followup? Even if we change it to opaque bytes, we'd still need to change the read- and write-paths in lockstep.

it's not a hard blocker obv so up to you

> how do we version this so that this can evolve in time (as you mention, encapsulating more information in there)?

I'd say that any format/version changes would be done based on timestamp given that's how execution layer changes would work. i.e if timetamp < X, expect version 1, if timestamp >= X, expect version 2

> should this be RLP encoded? Does that matter?

no strict requirements here, but I'd say that rlp encoding is preferred because rlp traits are available and implemented by most parts of codebase, so it'd just be easier to reason about

crates/commonware-node/src/consensus/block.rs
klkvr: for now there's no specific downside but reth [does have](https://github.com/kumaryash90/reth/blob/8d952abae290f91b70947a296dbbbe743c0eb05d/crates/node/builder/src/launch/engine.rs#L280-L285) a mechanism for caching blocks that were built by the same node which we might want to use eventually

right now it's not active because we're mostly using Ethereum components where block building is handled by third parties, so it does not rely on this optimization

so probably for now it's fine but ideally `extraData` is pased to builder as part of `TempoPayloadAttributes`, this would make EL and CL pov similar and would also allow us to keep a way for block builders to include custom data suffix/prefix

crates/commonware-node/src/dkg/manager/actor.rs
SuperFluffy: > for `Arbiter::finalize` I'd say we can just make it store `Option<Arbiter>` internally?

I found a nicer way to doing this: `Actor::handle_finalize(ceremony) -> ceremony` :-)

[aa6a628](https://github.com/tempoxyz/tempo/pull/538/commits/aa6a6287c8073351fc208550ab60f04c385ce06a)

crates/commonware-node/src/consensus/execution_driver/mod.rs
SuperFluffy: OK, I would like to leave this as a follow-up to not extend this PR further (it's already so massive because it's build on top of the epocher one).

crates/commonware-node/src/consensus/block.rs
SuperFluffy: Only (extremely minor) downside here is that we wouldn't give the consensus layer ("dkg actor") more time to collect acks because we'd collect at the beginning of a proposal, rather than at the end.

Created a followup here: https://github.com/tempoxyz/tempo/issues/560

crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -0,0 +1,473 @@
...
+                        }
+                        .into(),
+                    )
+                    .await;
+            }
+        }
+
+        match epoch::relative_position(block.height(), self.config.heights_per_epoch) {
+            epoch::RelativePosition::FirstHalf => {
+                let _ = ceremony.distribute_shares().await;
```
klkvr, line 385: is my understanding correct that this logic will be entered even when backfilling? should we try to avoid doing any p2p here if this is happening for an already passed epoch?

crates/commonware-node/src/consensus/execution_driver/mod.rs
SuperFluffy: Created followup issues #561 and #562

crates/commonware-node/src/dkg/manager/actor.rs
SuperFluffy: Yes, you will enter this logic when backfiling and attempt to distribute shares etc. However, avoiding p2p is harder than it sounds, because you cannot really know what the true current epoch is before getting all finalization certificates of the intermediate epoch boundary heights (so that you can verify certs given the public polynomial generated at the end of each epoch).

Similarly, you still want to participate in the ceremony if you went down and restarted during the an epoch.

I did want to do this myself, but would now prefer to leave it as a followup because it's harder than it sounds.

The current behavior should be as follows:

1. other nodes: will ignore you (because you are sending on the wrong subchannel).
2. yourself: will be a verifier but not a signer in the epoch.

crates/commonware-node/src/dkg/ceremony/mod.rs
```diff
@@ -0,0 +1,849 @@
...
+                    dkg_dealer.ack(ack.player.clone()).wrap_err_with(|| {
+                        format!(
+                            "failed updating dealer information with ack for \
+                             player `{player}` recovered from disk",
+                            player = ack.player,
+                        )
+                    })?;
+                }
+                dealer_me = Some(Dealer {
+                    inner: dkg_dealer,
```
klkvr: can we remove the `Dealer::inner` field? looks like it's only used to record acks which are already recoreded in `Dealer::acks` and are never read from `inner`

SuperFluffy: Interesting, we are not using `Dealer::finalize`. I thought we did. So this indeed becomes pretty useless.

Let me have a look if I am not missing any of the (very limited) validation that this dealer does.

crates/commonware-node/src/dkg/ceremony/mod.rs
```diff
@@ -0,0 +1,849 @@
...
+                    State {
+                        num_players: config.players.len() as u64,
+                        ..State::default()
+                    },
+                )
+                .await
+                .expect("must always be able to initialize the ceremony state to disk");
+
+            let (dkg_dealer, commitment, shares) =
+                dkg::Dealer::new(context, Some(share), config.players.clone());
```
klkvr: during backfill won't this be problematic because of randomness in new polynomial generation?

SuperFluffy: No, because the contribution of this node should either not play a role (everybody will ignore us and so our contribution to the DKG commitment doesn't matter), or we have already started the ceremony and recover our own dealings from disk.

The way this will look is that we generate generate these commitments and shares and try to distribute them, but effectively we will be yelling into the void. Toward the end of the ceremony, your arbiter will fail your ceremony (or technically: collect the dealings of the other dealers and generate a valid polynomial from that).

klkvr: I see, so `arbiter` only sees data from the block outcomes

SuperFluffy: Yeah, I renamed the function from `process_block` to `process_dealings_in_block` to make this clearer:

1. for each ack in the "intermediate outcome" (the outcome of a dealer's dealings), we check if the player that has signed the ack was in our list of players, and then we verify the ack's signature: if either of these fails, our arbiter disqualifies the dealer
2. then, we track the dealer, it's commitment, acks and reveals in the arbiter.

So for our backfilling node this would mean that we never put our dealings into the arbiter.

crates/commonware-node/src/dkg/ceremony/mod.rs
SuperFluffy: Dropped it: [00f137d](https://github.com/tempoxyz/tempo/pull/538/commits/00f137dc86d6ffb92fce212b3815caaa55874100)

crates/commonware-node/src/dkg/ceremony/mod.rs
```diff
@@ -0,0 +1,848 @@
...
+        }
+
+        Ok(())
+    }
+
+    /// Constructs and stores the intermediate ceremony outcome.
+    ///
+    /// If the node is not a dealer, then this is a no-op.
+    #[instrument(skip_all, fields(epoch = self.epoch()), err)]
+    pub(super) async fn construct_intermediate_outcome(&mut self) -> eyre::Result<()> {
```
klkvr, line 648: is there anything preventing us from constructing this once we receive a request? not super important but might simplfy logic a bit

SuperFluffy: I guess one advantage is that this does not eat into the proposal window. The operations are fairly cheap though (serialize a bunch of fields by writing them to a buffer, sign), so this should probably not matter.

I leave this for a followup though. https://github.com/tempoxyz/tempo/issues/565

Issue #536, opened 2025-10-22
Add dex_getOrders endpoint


mattsse:
For simplicity we can add this into just `TempoRpcExt`

# TODO
* define trait and function dex_getOrdersByAddress
* define rpc request types, we can put them here for now (new module) https://github.com/tempoxyz/tempo/blob/7d798a633ccf92aee06481b3cddc5252f15e4dc4/crates/node/src/rpc/mod.rs#L1-L1 
* we likely need some helpers to fetch orders from storage https://github.com/tempoxyz/tempo/blob/7d798a633ccf92aee06481b3cddc5252f15e4dc4/crates/precompiles/src/contracts/stablecoin_exchange/order.rs#L171-L171
* 

see also https://tempo-docs-git-jxom-rpc-tempoxyz.vercel.app/rpc https://tempo-docs-git-jxom-rpc-tempoxyz.vercel.app/rpc/dex_getOrders

PR #540, opened 2025-10-23
feat: Add an empty `tempo-order-book` crate

RomanHodulak:
Part of #536 

Dedicated for JSON-RPC API of orders


mattsse (commented):
unsure we need a new crate for this just now, I think we can define all of this in the existing node crate for now under rpc?

wdyt @klkvr

PR #542, opened 2025-10-23
feat: more missing linkingusd functions

howydev:
Adds other calls in TIP20 outside of `quote` related functions to LinkingUSD

This PR also removes `quoteToken()` from LinkingUSD as part of above

### DIFF
--- a/crates/precompiles/src/linking_usd/dispatch.rs
+++ b/crates/precompiles/src/linking_usd/dispatch.rs
@@ -28,6 +28,7 @@ impl<S: PrecompileStorageProvider> Precompile for LinkingUSD<'_, S> {
                 metadata::<ITIP20::totalSupplyCall>(self.total_supply())
             }
             ITIP20::currencyCall::SELECTOR => metadata::<ITIP20::currencyCall>(self.currency()),
+            ITIP20::quoteTokenCall::SELECTOR => metadata::<ITIP20::quoteTokenCall>(Address::ZERO),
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/linking_usd/dispatch.rs
mattsse, line 24: now that we have a bunch of variants

@DaniPopes do you think it's worth rearranging them based on which ones are likely to be used most often?

crates/precompiles/src/linking_usd/dispatch.rs
```diff
@@ -28,10 +28,13 @@ impl<S: PrecompileStorageProvider> Precompile for LinkingUSD<'_, S> {
                 metadata::<ITIP20::totalSupplyCall>(self.total_supply())
             }
             ITIP20::currencyCall::SELECTOR => metadata::<ITIP20::currencyCall>(self.currency()),
-            ITIP20::quoteTokenCall::SELECTOR => {
-                metadata::<ITIP20::quoteTokenCall>(self.quote_token())
-            }
             ITIP20::pausedCall::SELECTOR => metadata::<ITIP20::pausedCall>(self.paused()),
+            ITIP20::supplyCapCall::SELECTOR => {
```
0xKitsune, line 33: I think we still want to expose the `quoteToken` function, but simply return `Address::ZERO` since the `LinkingUSD` contract inherits the TIP20 contract.

PR #546, opened 2025-10-23
feat: Add `dex_getOrders` endpoint and schema

RomanHodulak:
Part of #536 

Adds a trait, parameters and response type.

### DIFF
--- a/crates/node/src/rpc/dex/mod.rs
+++ b/crates/node/src/rpc/dex/mod.rs
@@ -1,0 +1,13 @@
+pub use types::{
+    FilterRange, Order, OrdersFilters, OrdersParams, OrdersResponse, OrdersSort, OrdersSortOrder,
+};
+
+use jsonrpsee::{core::RpcResult, proc_macros::rpc};
+
+pub mod types;
+
+#[rpc(server, namespace = "dex")]
+pub trait TempoDexApi {
+    #[method(name = "getOrders")]
+    async fn orders(&self, params: Vec<OrdersParams>) -> RpcResult<OrdersResponse>;
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11708,6 +11708,7 @@ dependencies = [
  "clap",
  "eyre",
  "futures",
+ "jsonrpsee",
  "p256",
  "rand 0.9.2",
  "reth-chainspec",
--- a/crates/node/Cargo.toml
+++ b/crates/node/Cargo.toml
@@ -51,6 +51,7 @@ clap.workspace = true
 eyre.workspace = true
 tokio.workspace = true
 serde.workspace = true
+jsonrpsee.workspace = true
 
 [dev-dependencies]
 tempo-precompiles.workspace = true
@@ -62,10 +63,10 @@ serde_json.workspace = true
 tempo-contracts.workspace = true
 tokio = { workspace = true, features = ["rt", "rt-multi-thread", "macros"] }
 alloy = { workspace = true, features = [
-  "full",
-  "providers",
-  "signers",
-  "signer-mnemonic-all-languages",
+    "full",
+    "providers",
+    "signers",
+    "signer-mnemonic-all-languages",
 ] }
 rand = "0.9.2"
 futures = "0.3"
@@ -81,8 +82,8 @@ vergen-git2.workspace = true
 [features]
 default = []
 asm-keccak = [
-  "alloy/asm-keccak",
-  "alloy-primitives/asm-keccak",
-  "reth-node-core/asm-keccak",
-  "reth-node-ethereum/asm-keccak",
+    "alloy/asm-keccak",
+    "alloy-primitives/asm-keccak",
+    "reth-node-core/asm-keccak",
+    "reth-node-ethereum/asm-keccak",
 ]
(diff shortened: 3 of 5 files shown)


crates/node/src/rpc/orders.rs
```diff
@@ -0,0 +1,98 @@
+use alloy::primitives::{Address, B256};
+use jsonrpsee::{core::RpcResult, proc_macros::rpc};
+
+#[rpc(server, namespace = "dex")]
+pub trait TempoOrdersExtApi {
```
0xKitsune: nit: consider updating this to `TempoDexExtApi`since this will likely be expanded to fetch other dex related data in addition to orders.

crates/node/src/rpc/orders.rs
```diff
@@ -0,0 +1,98 @@
...
+    /// Filter flip orders
+    is_flip: Option<bool>,
+    /// Last filled timestamp (seconds) in range
+    last_filled_at: Option<FilterTimestampRange>,
+    /// Filter by maker address
+    maker: Option<B256>,
+    /// Filter by quote token
+    quote_token: Option<Address>,
+    /// Remaining amount in range
+    remaining: Option<FilterNumberRange>,
```
0xKitsune: Order amounts are `u128`. Is there a reason we are using `B256` for `FilterNumberRange`?

crates/node/src/rpc/orders.rs
```diff
@@ -0,0 +1,98 @@
...
+    /// Tick in range (from -2000 to 2000)
+    tick: Option<FilterNumberRange>,
```
0xKitsune: Similar to the above, ticks are `int16`.

crates/node/src/rpc/orders.rs
```diff
@@ -0,0 +1,98 @@
...
+    /// Created timestamp (seconds) in range
+    created_at: Option<FilterTimestampRange>,
+    /// Filter by order side (true=buy, false=sell)
+    is_bid: Option<bool>,
+    /// Filter flip orders
+    is_flip: Option<bool>,
+    /// Last filled timestamp (seconds) in range
+    last_filled_at: Option<FilterTimestampRange>,
+    /// Filter by maker address
+    maker: Option<B256>,
```
0xKitsune: Should this be `Address`?

crates/node/src/rpc/dex.rs
```diff
@@ -0,0 +1,98 @@
...
+}
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct OrdersResponse {
+    next_cursor: Option<String>,
+    orders: Vec<Order>,
+}
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct Order {
```
0xKitsune: Are we able to use the `Order` struct from the precompile crate directly?

RomanHodulak: The RPC represents it differently and has features that wouldn't make sense for the precompile type (for example, why would the precompile representation be serializable to JSON?).

crates/node/src/rpc/dex.rs
```diff
@@ -0,0 +1,92 @@
...
+}
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub enum OrdersSortOrder {
+    Asc,
+    Desc,
+}
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct OrdersFilters {
```
jxom: need camelCase serde for all these defs

crates/node/src/rpc/dex.rs
mattsse, line 1: all of these fields can be pub

crates/node/src/rpc/dex.rs
mattsse: we could potentially unify those if we add serde support for this order type

crates/node/src/rpc/dex.rs
```diff
@@ -0,0 +1,100 @@
...
+
+#[rpc(server, namespace = "dex")]
+pub trait TempoDexExtApi {
+    #[method(name = "getOrders")]
+    async fn orders(
+        &self,
+        cursor: Option<String>,
+        filters: Option<OrdersFilters>,
+        limit: Option<usize>,
+        sort: Option<OrdersSort>,
```
mattsse: I think this is a single request object and need a struct for this

```
 params: [{
    cursor: response.nextCursor,
    limit: 100
  }]
```

crates/node/src/rpc/dex/mod.rs
```diff
@@ -0,0 +1,13 @@
+pub use types::{
+    FilterRange, Order, OrdersFilters, OrdersParams, OrdersResponse, OrdersSort, OrdersSortOrder,
+};
+
+use jsonrpsee::{core::RpcResult, proc_macros::rpc};
+
+pub mod types;
+
+#[rpc(server, namespace = "dex")]
+pub trait TempoDexExtApi {
```
mattsse:
```suggestion
pub trait TempoDexApi {
```

Ext would indicate that have already have a namesapce for this

PR #548, opened 2025-10-24
feat: add validator config precompile

howydev:
Adding a validator config contract as a precompile

Notes:
1. We have no events. This should not be necessary since we're doing `eth_call`s for data
2. The interface is different from the proposed one in slack. We can only catch duplicates for `address`, but duplicated ed25519 keys would not be possible. This should be handled on the admin/adding keys side
3. Gas prices are currently derived from lib.rs with `VIEW_FUNC_GAS = 100` and `MUTATE_FUNC_GAS = 1000`. These are significantly mispriced, but given that the testnets are a friendly environment this is not urgent

cc @legion2002 let's do a pricing sync sometime 1-2 weeks from now to get accurate numbers for these ops

### DIFF
--- a/crates/contracts/src/precompiles/validator_config.rs
+++ b/crates/contracts/src/precompiles/validator_config.rs
@@ -18,10 +18,10 @@ sol! {
             uint64 index;
             address validatorAddress;
             /// Address where other validators can connect to this validator.
-            /// Format: <hostname|ip>:<port>
+            /// Format: `<hostname|ip>:<port>`
             string inboundAddress;
             /// IP address for firewall whitelisting by other validators.
-            /// Format: <ip>:<port> - must be an IP address, not a hostname.
+            /// Format: `<ip>:<port>` - must be an IP address, not a hostname.
             string outboundAddress;
         }
 
@@ -32,15 +32,15 @@ sol! {
         /// Add a new validator (owner only)
         /// @param newValidatorAddress The address of the new validator
         /// @param publicKey The validator's communication public publicKey
-        /// @param inboundAddress The validator's inbound address <hostname|ip>:<port> for incoming connections
-        /// @param outboundAddress The validator's outbound IP address <ip>:<port> for firewall whitelisting (IP only, no hostnames)
+        /// @param inboundAddress The validator's inbound address `<hostname|ip>:<port>` for incoming connections
+        /// @param outboundAddress The validator's outbound IP address `<ip>:<port>` for firewall whitelisting (IP only, no hostnames)
         function addValidator(address newValidatorAddress, bytes32 publicKey, bool active, string calldata inboundAddress, string calldata outboundAddress) external;
 
         /// Update validator information (only validator)
         /// @param newValidatorAddress The new address for this validator
         /// @param publicKey The validator's new communication public publicKey
-        /// @param inboundAddress The validator's inbound address <hostname|ip>:<port> for incoming connections
-        /// @param outboundAddress The validator's outbound IP address <ip>:<port> for firewall whitelisting (IP only, no hostnames)
+        /// @param inboundAddress The validator's inbound address `<hostname|ip>:<port>` for incoming connections
+        /// @param outboundAddress The validator's outbound IP address `<ip>:<port>` for firewall whitelisting (IP only, no hostnames)
         function updateValidator(address newValidatorAddress, bytes32 publicKey, string calldata inboundAddress, string calldata outboundAddress) external;
 
         /// Change validator active status (owner only)


crates/precompiles/src/validator_config/dispatch.rs
```diff
@@ -0,0 +1,183 @@
...
+            }
+            IValidatorConfig::changeValidatorStatusCall::SELECTOR => {
+                mutate_void::<IValidatorConfig::changeValidatorStatusCall, ValidatorConfigError>(
+                    calldata,
+                    msg_sender,
+                    |s, call| self.change_validator_status(s, call),
+                )
+            }
+            IValidatorConfig::changeOwnerCall::SELECTOR => {
+                mutate_void::<IValidatorConfig::changeOwnerCall, ValidatorConfigError>(
```
0xKitsune: Should we use the equivalent of `Ownable2Step`?

crates/contracts/src/precompiles/validator_config.rs
```diff
@@ -0,0 +1,81 @@
...
+    /// and the owner can manage validator status.
+    #[derive(Debug, PartialEq, Eq)]
+    #[sol(rpc)]
+    interface IValidatorConfig {
+        /// Validator information
+        struct Validator {
+            bytes32 key;
+            bool active;
+            uint64 index;
+            address validatorAddress;
```
klkvr, line 19: this would mean that we'd have to associate each validator with an addresses (this is not currently the case)

is this something we've decided to do? I might be missing some context but I thought for now the plan is to just store validators in the contract

cc @SuperFluffy @0xKitsune

crates/precompiles/src/validator_config/mod.rs
```diff
@@ -0,0 +1,838 @@
...
+        let len = bytes[31] as usize / 2;
+        if len > 31 {
+            return Err(ValidatorConfigError::unauthorized());
+        }
+        Ok(String::from_utf8_lossy(&bytes[..len]).to_string())
+    }
+
+    fn write_string(&mut self, slot: U256, value: String) -> Result<(), ValidatorConfigError> {
+        let bytes = value.as_bytes();
+        if bytes.len() > 31 {
```
klkvr: will `ipAddressOrDns` always fit into a single slot? this is partially why I think it might've been easier to do as just a Solidity contract for now, even writing strings in precompiles is pretty tricky rn

bin/tempo-bench/src/cmd/genesis.rs
```diff
@@ -22,16 +23,56 @@ use tempo_contracts::{
...
 };
 
+/// Initial validator configuration
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct InitialValidator {
+    /// Validator address
+    pub address: Address,
+    /// Communication key (32 bytes)
+    #[serde(with = "serde_bytes_32")]
+    pub key: [u8; 32],
```
klkvr: we can just use `alloy_primitives::B256` here

crates/contracts/src/precompiles/validator_config.rs
```diff
@@ -0,0 +1,81 @@
...
+        /// Update validator information (only validator)
+        /// @param newValidatorAddress The new address for this validator
+        /// @param key The validator's new communication public key
+        /// @param ipAddressOrDns The validator's new IP address
+        function updateValidator(address newValidatorAddress, bytes32 key, string calldata ipAddressOrDns) external;
+
+        /// Change validator active status (owner only)
+        /// @param validator The validator address
+        /// @param active Whether the validator should be active
+        function changeValidatorStatus(address validator, bool active) external;
```
klkvr, line 49: why would we want to store inactive validators?

xtask/src/genesis.rs
```diff
@@ -22,16 +23,56 @@ use tempo_contracts::{
...
+/// Initial validator configuration
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct InitialValidator {
+    /// Validator address
+    pub address: Address,
+    /// Communication key (32 bytes)
+    #[serde(with = "serde_bytes_32")]
+    pub key: [u8; 32],
+    /// IP address or DNS name
+    pub ip_address_or_dns: String,
```
0xKitsune: Consider using an enum here. Something like the following or similar:

```rust

#[derive(Debug, Clone, Serialize, Deserialize)]
pub enum IpOrDns {
    Ip(std::net::IpAddr),
    Dns(Url),
}

#[derive(Debug, Clone, Serialize, Deserialize)]
pub struct InitialValidator {
    // --snip--

    pub endpoint: IpOrDns

    // --snip--
}
```

bin/tempo-bench/src/cmd/genesis.rs
```diff
@@ -75,6 +120,18 @@ impl GenesisArgs {
...
+        // Load initial validators if config file provided
+        let initial_validators = if let Some(config_path) = &self.validators_config {
+            println!("Loading validators from {:?}", config_path);
+            let config_content = fs::read_to_string(config_path)?;
+            let validators: Vec<InitialValidator> = serde_json::from_str(&config_content)?;
+            println!("Loaded {} initial validators", validators.len());
+            validators
+        } else {
+            Vec::new()
+        };
```
0xKitsune: nit: since `initialize_validator_config` is the only place we pass this in, consider moving this logic into the function instead.

crates/contracts/src/precompiles/validator_config.rs
howydev: The address is what they'd use to rotate their keys, its easier to do it this way instead of implementing ed25519 validation

crates/contracts/src/precompiles/validator_config.rs
0xKitsune: Yeah unless there is a reason to keep inactive validators in the set, I think we we should remove these instead.

crates/contracts/src/precompiles/validator_config.rs
```diff
@@ -0,0 +1,81 @@
...
+    pub const fn validator_not_found() -> Self {
+        Self::ValidatorNotFound(IValidatorConfig::ValidatorNotFound {})
+    }
+}
+
+impl From<ValidatorConfigError> for PrecompileError {
+    fn from(val: ValidatorConfigError) -> Self {
+        Self::Other(format!("{:?}", val.selector()))
+    }
+}
```
0xKitsune: IIRC we do not use this anymore and this can be removed since precompile dispatch functions now return `Result<_, TempoPrecompileError>`. Not positive though so feel free to lmk if this we need this somewhere.

crates/precompiles/src/validator_config/mod.rs
```diff
@@ -0,0 +1,838 @@
...
+    pub fn initialize(&mut self, owner: Address) {
+        trace!(address=%self.precompile_address, %owner, "Initializing validator config precompile");
+
+        // must ensure the account is not empty, by setting some code
+        self.storage
+            .set_code(
+                self.precompile_address,
+                Bytecode::new_legacy(Bytes::from_static(&[0xef])),
+            )
+            .expect("TODO: handle error");
```
0xKitsune: Can we propagate all of these errors? You can update this to return `TempoPrecompileError`.

crates/precompiles/src/validator_config/mod.rs
```diff
@@ -0,0 +1,838 @@
...
+        call: IValidatorConfig::changeOwnerCall,
+    ) -> Result<(), ValidatorConfigError> {
+        self.check_owner(sender)?;
+        self.storage
+            .sstore(
+                self.precompile_address,
+                slots::OWNER,
+                call.newOwner.into_word().into(),
+            )
+            .map_err(|_| ValidatorConfigError::unauthorized())?;
```
0xKitsune: This should be propagated directly to `TempoPrecompileError` rather than mapped to `ValidatorConfigError::unauthorized`.

validators.example.json
```diff
@@ -0,0 +1,20 @@
...
+    "ip_address_or_dns": "192.168.1.100",
+    "active": true
+  },
+  {
+    "address": "0x3333333333333333333333333333333333333333",
+    "key": "0xcccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc",
+    "ip_address_or_dns": "validator3.example.com:26656",
+    "active": false
+  }
+]
```
0xKitsune: Looks like this file may have been accidentially commited?

howydev: hmm i meant for this to be like an `.env.example` equivalent of the validators format. Do we do stuff like that in this codebase?

crates/contracts/src/precompiles/validator_config.rs
howydev: let's keep this in, as per slack convo!

validators.example.json
SuperFluffy: I think this is generally fine, as long as it is clear how this can be used (which from a quick glance is not the case here, but I might be blind.)

I personally prefer to not only have an example, but also have it backed by actual code / a test that parses it  - and even better, calls the contract with these arguments.

crates/contracts/src/precompiles/validator_config.rs
howydev: Ah i think those changes were just merged into main. Will rebase and update the pattern

xtask/src/genesis.rs
SuperFluffy, line 1: This has migrated to the xtasks binary at `xtask/`

xtask/src/genesis.rs
SuperFluffy: An IP address is probably not sufficient - we need a socket address, which is `<ip>:<port>`

xtask/src/genesis.rs
```diff
@@ -423,6 +446,47 @@ fn initialize_stablecoin_exchange(evm: &mut TempoEvm<CacheDB<EmptyDB>>) -> eyre:
...
+            .add_validator(
+                &owner,
+                IValidatorConfig::addValidatorCall {
+                    newValidatorAddress: validator.address,
+                    key: FixedBytes::from(validator.key.0),
+                    active: validator.active,
+                    ipAddressOrDns: validator.ip_address_or_dns.clone(),
+                },
+            )
+            .expect("Failed to add validator");
```
SuperFluffy: expect messages should not contain information that something failed, but why you expected something to not fail.

Either way, you are already in a function that returns an `eyre::Result`, so just:

```suggest
.wrap_err("contract call to add to add validator failed")?;
```

xtask/src/genesis.rs
SuperFluffy: Just a plain URL is also not sufficient, because it need not have a port.

You could parse the URL, then grab only its host and port, and finally ensure that `parsed_url.to_string == format!("{host}:{port}")`

crates/precompiles/src/validator_config/dispatch.rs
howydev: we wouldnt have an admin in the future, this is just a stopgap so i dont think its necessary

xtask/src/genesis.rs
howydev: TIL - i'm just following the format for the other errors in the file, e.g. `.expect("Could not 0x00 validator fee token")`

Changed as requested

xtask/src/genesis.rs
SuperFluffy: Added this in [6271772](https://github.com/tempoxyz/tempo/pull/548/commits/6271772f43aea1728f6fefdb5497434ac574b88f)

crates/contracts/src/precompiles/validator_config.rs
SuperFluffy: Consider making a follow up or even just a tracking issue for this, @howydev (we can still decide to keep it the way it is, but move the final decision elsehwere)

crates/contracts/src/precompiles/validator_config.rs
```diff
@@ -0,0 +1,87 @@
...
+    ///
+    /// This precompile manages the set of validators that participate in consensus.
+    /// Validators can update their own information, rotate their identity to a new address,
+    /// and the owner can manage validator status.
+    #[derive(Debug, PartialEq, Eq)]
+    #[sol(rpc)]
+    interface IValidatorConfig {
+        /// Validator information
+        struct Validator {
+            bytes32 key;
```
SuperFluffy: Is this the public key? Can we:

```suggestion
            bytes32 publicKey;
```

xtask/Cargo.toml
```diff
@@ -41,3 +41,4 @@ alloy = { workspace = true, features = [
 ] }
 rayon = "1.10"
 simple-tqdm = { version = "0.2.0", features = ["rayon"] }
+url = "2.5.7"
```
SuperFluffy: remove

crates/precompiles/src/validator_config/dispatch.rs
```diff
@@ -0,0 +1,184 @@
...
+
+impl<'a, S: PrecompileStorageProvider> Precompile for ValidatorConfig<'a, S> {
+    fn call(&mut self, calldata: &[u8], msg_sender: &Address) -> PrecompileResult {
+        let selector: [u8; 4] = calldata
+            .get(..4)
+            .ok_or_else(|| {
+                PrecompileError::Other("Invalid input: missing function selector".to_string())
+            })?
+            .try_into()
+            .unwrap();
```
SuperFluffy: handle this error before merge

crates/precompiles/src/validator_config/mod.rs
```diff
@@ -0,0 +1,1264 @@
...
+
+            let (active, index) = slots::unpack_active_index(active_and_idx);
+
+            let inbound_address =
+                self.read_string(slot + slots::VALIDATOR_INBOUND_ADDRESS_OFFSET)?;
+
+            let outbound_address =
+                self.read_string(slot + slots::VALIDATOR_OUTBOUND_ADDRESS_OFFSET)?;
+
+            validators.push(IValidatorConfig::Validator {
```
SuperFluffy, line 181: Should this be prefiltered or not? Right now it's on the user to filter `active`.

crates/precompiles/src/validator_config/dispatch.rs
howydev: Hmm, the other precompiles have the same error

Not sure what to error here - if this was a smart contract and this happened, it would throw with a generic "EVM error: revert"

crates/precompiles/src/validator_config/mod.rs
howydev: I think that's fine, no way to find inactive validators to set as active otherwise (we don't have events to track when ecdsa key rotation happens)

crates/precompiles/src/validator_config/dispatch.rs
0xKitsune: You cam just use `map_err(PrecompileError::Other("Invalid function selector length"))?` or similar for now. This will be updated in [550](https://github.com/tempoxyz/tempo/pull/550).

Ref: 
https://github.com/tempoxyz/tempo/blob/feat/contract-macro/crates/precompiles-macros/src/dispatcher.rs#L41-L43

xtask/src/genesis.rs
```diff
@@ -426,6 +452,50 @@ fn initialize_stablecoin_exchange(evm: &mut TempoEvm<CacheDB<EmptyDB>>) -> eyre:
...
+    };
+
+    // Add initial validators
+    for validator in initial_validators.iter().tqdm() {
+        validator_config
+            .add_validator(
+                &owner,
+                IValidatorConfig::addValidatorCall {
+                    newValidatorAddress: validator.address,
+                    publicKey: FixedBytes::from(validator.key.0),
```
0xKitsune: nit: you can use `validator.key` directly here.
```suggestion
                    publicKey: validator.key,
```

crates/precompiles/src/validator_config/mod.rs
```diff
@@ -0,0 +1,1273 @@
...
+                slot + slots::VALIDATOR_ACTIVE_INDEX_OFFSET,
+            )?;
+
+            let (active, index) = slots::unpack_active_index(active_and_idx);
+
+            let inbound_address =
+                self.read_string(slot + slots::VALIDATOR_INBOUND_ADDRESS_OFFSET)?;
+
+            let outbound_address =
+                self.read_string(slot + slots::VALIDATOR_OUTBOUND_ADDRESS_OFFSET)?;
```
0xKitsune, line 179: You could consider updating this to be similar to how we handle the [Orderbook struct](https://github.com/tempoxyz/tempo/blob/main/crates/precompiles/src/stablecoin_exchange/orderbook.rs#L54-L91) in the Stable DEX precompile. However I would also be fine with not blocking on this since all of this will come for free with the `contract` macro from [550](https://github.com/tempoxyz/tempo/pull/550).

crates/precompiles/src/validator_config/mod.rs
```diff
@@ -0,0 +1,1273 @@
...
+                self.precompile_address,
+                old_slot + slots::VALIDATOR_ACTIVE_INDEX_OFFSET,
+                U256::ZERO,
+            )?;
+
+            // Clear old validator's inboundAddress
+            self.delete_string(old_slot + slots::VALIDATOR_INBOUND_ADDRESS_OFFSET)?;
+
+            // Clear old validator's outboundAddress
+            self.delete_string(old_slot + slots::VALIDATOR_OUTBOUND_ADDRESS_OFFSET)?;
```
0xKitsune, line 320: Similar to the comment above, the precompile refactor should make this much nicer but it might be nice to abstract this away with `Validator` struct methods in the meantime.

crates/precompiles/src/validator_config/mod.rs
```diff
@@ -0,0 +1,1273 @@
...
+    use crate::storage::slots::mapping_slot;
+    use alloy::primitives::{Address, U256, uint};
+
+    // Simple values
+    pub const OWNER: U256 = uint!(0_U256);
+    pub const VALIDATOR_COUNT: U256 = uint!(1_U256);
+
+    // Mappings
+    /// Maps index -> validator address (for iteration)
+    pub const VALIDATORS_ARRAY: U256 = uint!(2_U256);
```
0xKitsune, line 21: nit: If this is a mapping, can we update to `VALIDATORS_MAPPING` or update the comment?

howydev: good call

howydev: Whoops, a little bit of morning brain - i think might be an array, we track the latest slot with `VALIDATORS` and we only append AFAICT

The solidity compiler stores arrays values with a similar-ish scheme as a solidity map, something like `keccak(slot, idx)` so it does look like a map

crates/precompiles/src/validator_config/mod.rs
howydev: Interesting pattern, thanks for sharing!

crates/precompiles/src/validator_config/mod.rs
howydev: That would make it more readable yeah

Would prefer to leave this as tech debt for now, I'll pick the simplification in a follow up PR once things slow down a little

howydev: https://github.com/tempoxyz/tempo/issues/622 and assigned myself

PR #551, opened 2025-10-24
feat(node): Launch node with `TempoDex` RPC module

RomanHodulak:
Part of #536 

Adds a placeholder implementation of the generated jsonrpsee API trait.

Registers this implementation during node launch.

### DIFF
--- a/crates/node/src/rpc/dex/mod.rs
+++ b/crates/node/src/rpc/dex/mod.rs
@@ -3,6 +3,8 @@ pub use types::{
 };
 
 use jsonrpsee::{core::RpcResult, proc_macros::rpc};
+use reth_node_core::rpc::result::internal_rpc_err;
+use reth_rpc_eth_api::RpcNodeCore;
 
 pub mod types;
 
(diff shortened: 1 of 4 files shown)


crates/node/src/rpc/dex/mod.rs
```diff
@@ -11,3 +11,13 @@ pub trait TempoDexApi {
...
     async fn orders(&self, params: Vec<OrdersParams>) -> RpcResult<OrdersResponse>;
 }
+
+#[derive(Debug, Clone, Default)]
+pub struct TempoDex;
+
+#[async_trait::async_trait]
+impl TempoDexApiServer for TempoDex {
+    async fn orders(&self, _params: Vec<OrdersParams>) -> RpcResult<OrdersResponse> {
+        todo!()
```
mattsse: for now let's return:

   Err(internal_rpc_err("unimplemented"))

see 

https://github.com/paradigmxyz/reth/blob/4a24cb3b499b997f93de9b5a01bfd0ecfe8339c1/crates/rpc/rpc/src/otterscan.rs#L293-L293

then this endpoint doesnt crash when called

crates/node/src/rpc/dex/mod.rs
```diff
@@ -11,3 +11,13 @@ pub trait TempoDexApi {
     #[method(name = "getOrders")]
     async fn orders(&self, params: Vec<OrdersParams>) -> RpcResult<OrdersResponse>;
 }
+
+#[derive(Debug, Clone, Default)]
+pub struct TempoDex;
```
mattsse: let's add a oneliner doc

like

```
Rpc Handlers for the `dex_` API.
```

PR #553, opened 2025-10-24
feat(node): Add `dex_getOrderbooks` JSON-RPC endpoint and schema

RomanHodulak:
Part of #536 

Adds the boilerplate for https://tempo-docs-git-jxom-rpc-tempoxyz.vercel.app/rpc/dex_getOrderbooks


crates/node/src/rpc/dex/books.rs
mattsse, line 1: I think we need pub fields for all of these

PR #555, opened 2025-10-24
feat: Add authorization list support for AA Tx Type

legion2002:
AA Tx types need a different authorization list because of multiple signature types

### DIFF
--- a/crates/node/src/rpc/request.rs
+++ b/crates/node/src/rpc/request.rs
@@ -17,7 +17,7 @@ use serde::{Deserialize, Serialize};
 use tempo_evm::TempoBlockEnv;
 use tempo_primitives::{
     AASignature, SignatureType, TempoTxEnvelope, TxAA, TxFeeToken,
-    transaction::{Call, TempoTypedTransaction},
+    transaction::{AASignedAuthorization, Call, TempoTypedTransaction},
 };
 use tempo_revm::{AATxEnv, TempoTxEnv};
 
@@ -43,6 +43,10 @@ pub struct TempoTransactionRequest {
     /// Optional key-specific data for gas estimation (e.g., webauthn authenticator data).
     /// Required when key_type is WebAuthn to calculate calldata gas costs.
     pub key_data: Option<Bytes>,
+
+    /// Optional AA authorization list for AA transactions (supports multiple signature types)
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub aa_authorization_list: Vec<AASignedAuthorization>,
 }
 
 impl TempoTransactionRequest {
@@ -145,6 +149,7 @@ impl TempoTransactionRequest {
             fee_token: self.fee_token,
             access_list: self.inner.access_list.unwrap_or_default(),
             calls,
+            aa_authorization_list: self.aa_authorization_list,
             ..Default::default()
         })
     }
@@ -173,6 +178,7 @@ impl TryIntoSimTx<TempoTxEnvelope> for TempoTransactionRequest {
                 calls,
                 key_type,
                 key_data,
+                aa_authorization_list,
             } = self;
             let envelope =
                 match TryIntoSimTx::<EthereumTxEnvelope<TxEip4844>>::try_into_sim_tx(inner.clone())
@@ -185,6 +191,7 @@ impl TryIntoSimTx<TempoTxEnvelope> for TempoTransactionRequest {
                             calls,
                             key_type,
                             key_data,
+                            aa_authorization_list,
                         }));
                     }
                 };
@@ -198,6 +205,7 @@ impl TryIntoSimTx<TempoTxEnvelope> for TempoTransactionRequest {
                         calls,
                         key_type,
                         key_data,
+                        aa_authorization_list,
                     })
                 })?)
         }
@@ -218,13 +226,14 @@ impl TryIntoTxEnv<TempoTxEnv, TempoBlockEnv> for TempoTransactionRequest {
             calls,
             key_type,
             key_data,
+            aa_authorization_list,
         } = self;
         Ok(TempoTxEnv {
             inner: inner.try_into_tx_env(cfg_env, &block_env.inner)?,
             fee_token,
             is_system_tx: false,
             fee_payer: None,
-            aa_tx_env: (!calls.is_empty()).then(|| {
+            aa_tx_env: (!calls.is_empty() || !aa_authorization_list.is_empty()).then(|| {
                 // Create mock signature for gas estimation
                 // If key_type is not provided, default to secp256k1
                 let mock_signature = key_type
@@ -235,6 +244,7 @@ impl TryIntoTxEnv<TempoTxEnv, TempoBlockEnv> for TempoTransactionRequest {
                 Box::new(AATxEnv {
                     aa_calls: calls,
                     signature: mock_signature,
+                    aa_authorization_list,
                     ..Default::default()
                 })
             }),
@@ -350,9 +360,7 @@ impl From<TransactionRequest> for TempoTransactionRequest {
         Self {
             inner: value,
             fee_token: None,
-            calls: vec![],
-            key_type: None,
-            key_data: None,
+            ..Default::default()
         }
     }
 }
@@ -421,9 +429,7 @@ impl<T: TransactionTrait + FeeToken> From<Signed<T>> for TempoTransactionRequest
         Self {
             fee_token: value.tx().fee_token(),
             inner: TransactionRequest::from_transaction(value),
-            calls: vec![],
-            key_type: None,
-            key_data: None,
+            ..Default::default()
         }
     }
 }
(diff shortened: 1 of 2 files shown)


crates/primitives/src/transaction/aa_authorization.rs
```diff
@@ -0,0 +1,239 @@
...
+/// The structure and methods mirror `SignedAuthorization` exactly to maintain
+/// compatibility with the EIP-7702 spec.
+#[derive(Clone, Debug, Eq, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct AASignedAuthorization {
+    /// Inner authorization (reuses alloy's Authorization)
+    #[cfg_attr(feature = "serde", serde(flatten))]
+    inner: Authorization,
+    /// The AA signature (Secp256k1, P256, or WebAuthn)
+    signature: AASignature,
```
mattsse, line 28: ah I see, we need this because this has a diff signature than 7702

imo not worth making 7702 sig generic, so this additional type is fine

crates/primitives/src/transaction/aa_authorization.rs
```diff
@@ -0,0 +1,239 @@
...
+#[cfg(feature = "reth-codec")]
+impl reth_codecs::Compact for AASignedAuthorization {
+    fn to_compact<B>(&self, buf: &mut B) -> usize
+    where
+        B: alloy_rlp::BufMut + AsMut<[u8]>,
+    {
+        // Encode using RLP
+        let mut rlp_buf = Vec::new();
+        self.encode(&mut rlp_buf);
+        buf.put_slice(&rlp_buf);
```
mattsse: this is likely not optimal but I think this is fine, not worth adding compact derives here I guess @klkvr @joshieDo ?

crates/primitives/src/transaction/account_abstraction.rs
```diff
@@ -442,6 +461,8 @@ impl TxAA {
             return Err(alloy_rlp::Error::InputTooShort);
         };
 
+        let authorization_list = Decodable::decode(buf)?;
```
mattsse: I think makes sense at this position

crates/primitives/src/transaction/aa_authorization.rs
```diff
@@ -0,0 +1,239 @@
...
+    }
+}
+
+impl Hash for AASignedAuthorization {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.inner.hash(state);
+        // Hash the signature bytes
+        self.signature.to_bytes().hash(state);
+    }
+}
```
klkvr: can we just derive this?

crates/primitives/src/transaction/aa_authorization.rs
klkvr: I think rlp should fine here, though can we pass the `buf` directly into `encode`?

crates/primitives/src/transaction/aa_authorization.rs
```diff
@@ -0,0 +1,239 @@
...
+        self.encode(&mut rlp_buf);
+        buf.put_slice(&rlp_buf);
+        rlp_buf.len()
+    }
+
+    fn from_compact(buf: &[u8], len: usize) -> (Self, &[u8]) {
+        let rlp_data = &buf[..len];
+        let auth = Self::decode(&mut &rlp_data[..]).expect("valid RLP encoding");
+        (auth, &buf[len..])
+    }
```
klkvr, line 180: same here, `decode(&mut buf)` should work

crates/primitives/src/transaction/aa_authorization.rs
```diff
@@ -0,0 +1,239 @@
...
+        let signature = AASignature::default(); // Use secp256k1 test signature
+        let signed = AASignedAuthorization::new_unchecked(auth, signature);
+
+        let mut buf = Vec::new();
+        signed.encode(&mut buf);
+
+        let decoded = AASignedAuthorization::decode(&mut buf.as_slice()).unwrap();
+        assert_eq!(buf.len(), signed.length());
+        assert_eq!(decoded, signed);
+    }
```
klkvr, line 206: let's just add `#[cfg_attr(test, reth_codecs::add_arbitrary_tests(compact, rlp))]`

crates/revm/src/handler.rs
```diff
@@ -352,6 +370,105 @@ where
...
+            .tx()
+            .aa_tx_env
+            .as_ref()
+            .map(|aa_env| !aa_env.authorization_list.is_empty())
+            .unwrap_or(false);
+
+        // If it's an AA transaction with authorization list, we need to apply it manually
+        // since the default implementation only checks for TransactionType::Eip7702
+        if has_aa_auth_list {
+            // TODO: @rakita could we have a helper function for this logic in revm?
```
klkvr: @rakita worth evaluating which logic from here can be encapsulated as helpers in revm

crates/primitives/src/transaction/account_abstraction.rs
```diff
@@ -169,6 +168,9 @@ pub struct TxAA {
     /// Transaction can only be included in a block after this timestamp
     #[cfg_attr(feature = "serde", serde(with = "alloy_serde::quantity::opt"))]
     pub valid_after: Option<u64>,
+
+    /// Authorization list (EIP-7702 style with AA signatures)
+    pub authorization_list: Vec<AASignedAuthorization>,
```
klkvr: I guess this would break deserialization for some of the tooling, not sure if matters much

legion2002: maybe we should rename this to authorization_list_aa, just to make sure it's handled correctly.

jxom: I don't think it matters too much (on TS side).

crates/primitives/src/transaction/account_abstraction.rs
```diff
@@ -199,6 +201,17 @@ impl TxAA {
             return Err("only protocol nonce (nonce_key = 0) is currently supported");
         }
 
+        // Authorization list validation: Cannot have Create in any call when authorization_list is non-empty
+        // This follows EIP-7702 semantics - when using delegation
+        // TODO: Why is this restricted in 7702?
```
rakita: Can't remember why, but it feels like an artificial restriction. Some explanation can be found here: https://eips.ethereum.org/EIPS/eip-7702#no-blobs-no-contract-creation

An empty list is just a sanity check to ensure that the new transaction type is not used without purpose.

crates/revm/src/handler.rs
rakita: Probably all of it, will refactor this later and make a PR in tempo

crates/primitives/src/transaction/aa_signature.rs
```diff
@@ -75,6 +75,13 @@ impl Encodable for AASignature {
     }
 }
 
+impl Decodable for AASignature {
+    fn decode(buf: &mut &[u8]) -> alloy_rlp::Result<Self> {
+        let bytes: Bytes = Decodable::decode(buf)?;
+        Self::from_bytes(&bytes).map_err(alloy_rlp::Error::Custom)
```
mattsse, line 81: looking at

I believe this should be okay `AASignature::from_bytes` I believe this is okay

crates/primitives/src/transaction/account_abstraction.rs
mattsse: @klkvr I believe this matters for things like eth_call server side, because unless we specialize serde for TempoTxrequest this would clash with the 7702 field?

crates/primitives/src/transaction/account_abstraction.rs
mattsse: imo we could perhaps allow this, but also doesnt really matter

crates/primitives/src/transaction/account_abstraction.rs
klkvr: yeah, we'd have to specialize serde or require those to be passed in a separate field

legion2002: yeah, I'm happy to just rename this. This fixes the issue right?

crates/primitives/src/transaction/account_abstraction.rs
legion2002: Okay doesn't seem like there is a strong reason to restrict it, but will just leave as is for now.

crates/primitives/src/transaction/account_abstraction.rs
legion2002: updated.

crates/node/src/rpc/request.rs
```diff
@@ -43,6 +43,10 @@ pub struct TempoTransactionRequest {
     /// Optional key-specific data for gas estimation (e.g., webauthn authenticator data).
     /// Required when key_type is WebAuthn to calculate calldata gas costs.
     pub key_data: Option<Bytes>,
+
+    /// Optional AA authorization list for AA transactions (supports multiple signature types)
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub aa_authorization_list: Vec<AASignedAuthorization>,
```
legion2002, line 49: @klkvr Just checking, we do need this extra field in the TempoTxRequest right? So that eth_estimateGas works correctly with the AA authorization lists?

Issue #543, opened 2025-10-23
Add tempo hardfork support


mattsse:
if we want to persist testnets while still being able to ship changes, we need hardfork transitions

we'll need this for mainnet anyway I believe.

for this we need

## TODO
* define TempoHardforks 
* integrate in evm

see also:

https://github.com/alloy-rs/hardforks/blob/6dd26f64e8d21228aac9ee6830074e026bcfe4d1/crates/op-hardforks/src/lib.rs#L24-L31

and 

https://github.com/berachain/bera-reth/blob/4b8f9af4ee8f250143266e104cde04b504e5941c/src/hardforks/mod.rs

PR #556, opened 2025-10-24
feat: add tempo hardfork support

fgimenez:
Closes: #543 

- define `TempoHardfork` enum using alloy-hardforks `hardfork!`
- add `Adagio` as baseline hardfork (name comes from `genesis/adagio.json` chainspec)
- add `TempoHardforks` trait for querying hardfork activations
- integrate hardfork configuration into `TempoChainSpec`

### DIFF
--- a/crates/chainspec/src/spec.rs
+++ b/crates/chainspec/src/spec.rs
@@ -3,9 +3,8 @@ use alloy_eips::eip7840::BlobParams;
 use alloy_genesis::Genesis;
 use alloy_primitives::{Address, B256, U256};
 use reth_chainspec::{
-    BaseFeeParams, Chain, ChainHardforks, ChainSpec, DepositContract, EthChainSpec,
-    EthereumHardfork, EthereumHardforks, ForkCondition, ForkFilter, ForkId, Hardfork, Hardforks,
-    Head,
+    BaseFeeParams, Chain, ChainSpec, DepositContract, EthChainSpec, EthereumHardfork,
+    EthereumHardforks, ForkCondition, ForkFilter, ForkId, Hardfork, Hardforks, Head,
 };
 use reth_cli::chainspec::{ChainSpecParser, parse_genesis};
 use reth_ethereum::evm::primitives::eth::spec::EthExecutorSpec;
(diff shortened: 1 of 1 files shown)


crates/chainspec/src/hardfork.rs
```diff
@@ -0,0 +1,118 @@
...
+//! ## Current State
+//!
+//! The `Adagio` variant is a placeholder representing the pre-hardfork baseline.
+
+use alloy_hardforks::hardfork;
+use reth_chainspec::ForkCondition;
+
+hardfork!(
+    /// Tempo-specific hardforks for network upgrades.
+    TempoHardfork {
```
0xKitsune, line 25: Should we add `Serialize`/`Deserialize` similar to [OpHardforks](https://github.com/alloy-rs/hardforks/blob/1e727a2539af5893e504239773392defa22773c7/crates/op-hardforks/src/lib.rs#L29-L30)?

crates/chainspec/src/spec.rs
```diff
@@ -78,6 +80,8 @@ pub static DEV: LazyLock<Arc<TempoChainSpec>> = LazyLock::new(|| {
 pub struct TempoChainSpec {
     /// [`ChainSpec`].
     pub inner: ChainSpec<TempoHeader>,
+    /// Tempo-specific hardfork configuration
+    pub tempo_hardforks: TempoChainHardforks,
```
klkvr: doing this via an extra field would mean that p2p forkhashes would not be affected by Tempo hardforks

not sure if this matters much but I guess ideally we want to merge the forks into inner `ChainHardforks`? @mattsse wdyt?

crates/chainspec/src/hardfork.rs
fgimenez: good point, done

crates/chainspec/src/spec.rs
mattsse: yeah exactly

@fgimenez we'd need to parse this from chainspec anyway, so we'd need to do what we do for optimism

https://github.com/paradigmxyz/reth/blob/ffeaa4772d01313062b0b127eb4dcc980550aec0/crates/optimism/chainspec/src/lib.rs#L412-L422

and insert the type erased hardfork types

crates/chainspec/src/spec.rs
```diff
@@ -179,8 +184,16 @@ impl EthExecutorSpec for TempoChainSpec {
     }
 }
 
+impl TempoHardforks for TempoChainSpec {
+    fn tempo_fork_activation(&self, fork: TempoHardfork) -> ForkCondition {
+        self.tempo_hardforks.tempo_fork_activation(fork)
```
mattsse: this then can simply forward to inner, like

https://github.com/paradigmxyz/reth/blob/ffeaa4772d01313062b0b127eb4dcc980550aec0/crates/optimism/chainspec/src/lib.rs#L334-L338

crates/chainspec/src/hardfork.rs
```diff
@@ -0,0 +1,133 @@
...
+    /// Tempo-specific hardforks for network upgrades.
+    #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+    TempoHardfork {
+        /// Placeholder representing the baseline (pre-hardfork) state.
+        Adagio,
+    }
+);
+
+/// Trait for querying Tempo-specific hardfork activations.
+pub trait TempoHardforks {
```
mattsse: we can model this after OpHardforks: EthereumHardforks, then we can easier add this bound

crates/chainspec/src/spec.rs
fgimenez: makes sense done in dba8781a4d75b11121ba96d5152ed7636b9f762c, ptal

crates/chainspec/src/spec.rs
fgimenez: indeed, already done as part of dba8781a4d75b11121ba96d5152ed7636b9f762c

crates/chainspec/src/hardfork.rs
fgimenez: good point, done

crates/chainspec/src/hardfork.rs
```diff
@@ -0,0 +1,134 @@
...
+    /// Convenience method to check if Adagio hardfork is active at a given timestamp
+    fn is_adagio_active_at_timestamp(&self, timestamp: u64) -> bool {
+        self.tempo_fork_activation(TempoHardfork::Adagio)
+            .active_at_timestamp(timestamp)
+    }
+}
+
+/// Configuration for Tempo-specific hardfork activations
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TempoChainHardforks {
```
klkvr: do we still need this type?

crates/chainspec/src/spec.rs
```diff
@@ -83,8 +105,48 @@ pub struct TempoChainSpec {
...
+        let tempo_hardfork_opts = [(
+            TempoHardfork::Adagio.boxed(),
+            tempo_genesis_info.adagio_time,
+        )];
+
+        let mut tempo_hardforks = tempo_hardfork_opts
+            .into_iter()
+            .filter_map(|(hardfork, opt)| {
+                opt.map(|time| (hardfork, ForkCondition::Timestamp(time)))
+            })
```
mattsse: @klkvr I feel like all of this should be easier

any ideas how to simplify?

```
let mut base_spec = ChainSpec::from_genesis(genesis);
```

already does the ordering so I think we only need to something like `ChainHardforks::insert_reorder` ?

does that make sense @fgimenez ?

since we only have timestamps, this should be fairly simple to reorder the list

crates/chainspec/src/spec.rs
```diff
@@ -83,8 +105,48 @@ pub struct TempoChainSpec {
...
+            .collect::<Vec<_>>();
+
+        all_hardforks.append(&mut tempo_hardforks);
```
klkvr: I believe we want this to be ordered by `ForkCondition`

Which makes we wonder if we should improve `ChainHardforks` API so that we could just use `ChainHardforks::insert` here which right now does not ensure the correct order

crates/chainspec/src/spec.rs
klkvr: just noticed your comment, yep I think we need to add better API for this to reth

fgimenez: > since we only have timestamps, this should be fairly simple to reorder the list

indeed, pushed changes to insert tempo hardforks in the right positions for now, ptal

crates/chainspec/src/hardfork.rs
fgimenez: good catch, removed

crates/chainspec/src/spec.rs
fgimenez: yep, already addressed

crates/chainspec/src/spec.rs
fgimenez: bumped reth to use new `ChainHardforks` API, much nicer now @mattsse @klkvr ptal

crates/chainspec/src/spec.rs
```diff
@@ -83,8 +104,24 @@ pub struct TempoChainSpec {
...
+        // Create base chainspec from genesis (already has ordered Ethereum hardforks)
+        let mut base_spec = ChainSpec::from_genesis(genesis);
+
+        // Collect Tempo hardforks to insert
+        let tempo_forks: Vec<_> = [tempo_genesis_info
+            .adagio_time
+            .map(|time| (TempoHardfork::Adagio, ForkCondition::Timestamp(time)))]
+        .into_iter()
+        .flatten()
+        .collect();
```
mattsse, line 119: this makes sense

this currently doesn't take affect because the genesis generator doesnt know about this, but we can do this separately once this becomes relevant, but worth tracking in an issue

fgimenez: makes sense, created here https://github.com/tempoxyz/tempo/issues/583

PR #558, opened 2025-10-26
feat(precompiles): propagate errors through call context

0xKitsune:
This PR introduces an initial implementation of precompile error handling. It replaces `unwrap`s, `expect`s and panics with proper error propagation, ensuring that errors from nested calls are surfaced through the call context.

Previously, precompile calls were dispatched via the [`view`, `mutate`, `mutate_void`](https://github.com/tempoxyz/tempo/blob/main/crates/precompiles/src/lib.rs#L201-L231), etc. functions, which each returned a single `E: SolInterface` type corresponding to one contract interface (e.g. `IStablecoinExchangeError`).

While this allowed errors from individual precompiles to propagate, nested call errors were not propagated.

To address this, a new `TempoPrecompileError` type was introduced to wrap all precompile specific errors. The `view`, `mutate`, and related functions now return this unified error type, allowing full propagation across nested calls.

```rust
/// Top-level error type for all Tempo precompile operations
#[derive(Debug, Clone, PartialEq, Eq, thiserror::Error)]
pub enum TempoPrecompileError {
    /// Error from stablecoin exchange
    #[error("Stablecoin exchange error: {0:?}")]
    StablecoinExchange(StablecoinExchangeError),

    /// Error from TIP20 token
    #[error("TIP20 token error: {0:?}")]
    TIP20(TIP20Error),

    /// Error from roles auth
    #[error("Roles auth error: {0:?}")]
    RolesAuthError(RolesAuthError),

    /// Error from 403 registry
    #[error("Roles auth error: {0:?}")]
    TIP403RegistryError(TIP403RegistryError),

    /// Error from TIP  fee manager
    #[error("TIP fee manager error: {0:?}")]
    FeeManagerError(FeeManagerError),

    /// Error from TIP fee AMM
    #[error("TIP fee AMM error: {0:?}")]
    TIPFeeAMMError(TIPFeeAMMError),

    /// Error from TIP account registrar
    #[error("TIP account registrar error: {0:?}")]
    TIPAccountRegistrarError(TIPAccountRegistrarError),

    /// Error from native AA nonce manager
    #[error("Native AA nonce error: {0:?}")]
    NonceError(NonceError),

    #[error("Fatal precompile error: {0:?}")]
    Fatal(String),
}
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2256,17 +2256,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b94f61472cee1439c0b966b47e3aca9ae07e45d070759512cd390ea2bebc6675"
 
 [[package]]
-name = "clocksource"
-version = "0.8.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "129026dd5a8a9592d96916258f3a5379589e513ea5e86aeb0bd2530286e44e9e"
-dependencies = [
- "libc",
- "time",
- "winapi",
-]
-
-[[package]]
 name = "coins-bip32"
 version = "0.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2954,27 +2943,6 @@ dependencies = [
  "generic-array",
  "rand_core 0.6.4",
  "typenum",
-]
-
-[[package]]
-name = "csv"
-version = "1.3.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "acdc4883a9c96732e4733212c01447ebd805833b7275a73ca3ee080fd77afdaf"
-dependencies = [
- "csv-core",
- "itoa",
- "ryu",
- "serde",
-]
-
-[[package]]
-name = "csv-core"
-version = "0.1.12"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7d02f3b0da4c6504f86e9cd789d8dbafab48c2321be74e9987593de5a894d93d"
-dependencies = [
- "memchr",
 ]
 
 [[package]]
@@ -7210,17 +7178,6 @@ dependencies = [
  "unicode-segmentation",
  "unicode-truncate",
  "unicode-width 0.2.0",
-]
-
-[[package]]
-name = "ratelimit"
-version = "0.10.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "36ea961700fd7260e7fa3701c8287d901b2172c51f9c1421fa0f21d7f7e184b7"
-dependencies = [
- "clocksource",
- "parking_lot",
- "thiserror 1.0.69",
 ]
 
 [[package]]
@@ -11478,7 +11435,6 @@ dependencies = [
  "clap",
  "core_affinity",
  "crossbeam-utils",
- "csv",
  "dashmap 6.1.0",
  "eyre",
  "fastrand",
@@ -11491,20 +11447,15 @@ dependencies = [
  "hyper-util",
  "mimalloc",
  "parking_lot",
- "ratelimit",
  "rayon",
- "reth",
- "reth-evm",
  "rlimit",
  "serde",
  "serde_json",
  "simple-tqdm",
  "tempo-chainspec",
  "tempo-contracts",
- "tempo-evm",
  "tempo-precompiles",
  "thiserror 2.0.17",
- "thousands",
  "tokio",
  "tokio-util",
  "toml",
@@ -11918,6 +11869,7 @@ dependencies = [
 name = "tempo-xtask"
 version = "0.1.0"
 dependencies = [
+ "alloy",
  "alloy-primitives",
  "camino",
  "clap",
@@ -11926,8 +11878,19 @@ dependencies = [
  "eyre",
  "indexmap 2.11.4",
  "rand 0.8.5",
+ "rayon",
+ "reth",
+ "reth-evm",
+ "serde",
+ "serde_json",
+ "simple-tqdm",
+ "tempo-chainspec",
  "tempo-commonware-node-config",
  "tempo-commonware-node-cryptography",
+ "tempo-contracts",
+ "tempo-evm",
+ "tempo-precompiles",
+ "tokio",
  "toml",
 ]
 
@@ -12061,12 +12024,6 @@ dependencies = [
  "quote",
  "syn 2.0.106",
 ]
-
-[[package]]
-name = "thousands"
-version = "0.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3bf63baf9f5039dadc247375c29eb13706706cfde997d0330d05aa63a77d8820"
 
 [[package]]
 name = "thread_local"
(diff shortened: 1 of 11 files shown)


crates/contracts/src/precompiles/tip20.rs
```diff
@@ -116,6 +116,19 @@ sol! {
...
+    pub const fn unauthorized() -> Self {
+        Self::Unauthorized(IRolesAuth::Unauthorized {})
+    }
+}
+
+impl From<RolesAuthError> for PrecompileError {
+    fn from(val: RolesAuthError) -> Self {
+        Self::Other(format!("{:?}", val.selector()))
+    }
+}
```
klkvr: why do we need this impl? I'd expect that `RolesAuthError` should map to a revert vs `::Other` variant

crates/contracts/src/precompiles/tip_account_registrar.rs
```diff
@@ -14,3 +15,26 @@ sol! {
...
+    pub const fn nonce_not_zero() -> Self {
+        Self::NonceNotZero(ITipAccountRegistrar::NonceNotZero {})
+    }
+}
+
+impl From<TIPAccountRegistrarError> for PrecompileError {
+    fn from(val: TIPAccountRegistrarError) -> Self {
+        Self::Other(format!("{:?}", val.selector()))
+    }
+}
```
klkvr: same here

crates/precompiles/src/linking_usd/mod.rs
```diff
@@ -26,148 +28,173 @@ impl<'a, S: PrecompileStorageProvider> LinkingUSD<'a, S> {
...
     fn is_transfer_from_authorized(
         &mut self,
         sender: &Address,
         from: &Address,
         recipient: &Address,
-    ) -> bool {
-        *sender == STABLECOIN_EXCHANGE_ADDRESS
-            || self.token.has_role(from, *TRANSFER_ROLE)
-            || self.token.has_role(recipient, *RECEIVE_ROLE)
+    ) -> Result<bool, TempoPrecompileError> {
```
klkvr, line 53: might be worth introducing a `Result` alias in tempo-precompiles to simplify this a bit, like
```
type Result<T, E = TempoPrecompileError> = Result<T, E>;
```

crates/revm/src/handler.rs
```diff
@@ -427,19 +428,21 @@ where
...
-                                fee: Box::new(gas_balance_spending),
-                                balance: Box::new(account_balance),
-                            },
-                        )
-                    }
+                            balance: Box::new(account_balance),
+                        },
+                    ),
+
                     _ => EVMError::Custom(format!("{e:?}")),
```
klkvr: I'm a bit worried about this logic — right now if we encounter e.g an error due to `fee_manager` being blacklisted we'd end up in this `EVMError::Custom` branch which would fail entire block execution

should we maybe just ensure that if `collect_fee_pre_tx` reverts, the tx is invalid?

crates/precompiles/src/lib.rs
```diff
@@ -249,16 +245,16 @@ pub fn mutate<T: SolCall, E: SolInterface>(
         )),
         Err(e) => Ok(PrecompileOutput::new_reverted(
             MUTATE_FUNC_GAS,
-            E::abi_encode(&e).into(),
+            e.abi_encode(),
         )),
```
klkvr: we most likely want to propagate some of the errors here as `PrecompileError` as https://github.com/tempoxyz/tempo/pull/559 suggests

crates/precompiles/src/error.rs
```diff
@@ -0,0 +1,106 @@
...
+    fn from(err: TIPAccountRegistrarError) -> Self {
+        Self::TIPAccountRegistrarError(err)
+    }
+}
+
+impl From<NonceError> for TempoPrecompileError {
+    fn from(err: NonceError) -> Self {
+        Self::NonceError(err)
+    }
+}
```
klkvr, line 97: those can be just `#[from]` attribute, like 
```rust
/// Error from TIP20 token
#[error("TIP20 token error: {0:?}")]
TIP20(#[from] TIP20Error),
```

crates/precompiles/src/error.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    /// Error from native AA nonce manager
+    #[error("Native AA nonce error: {0:?}")]
+    NonceError(NonceError),
+
+    #[error("Fatal precompile error: {0:?}")]
+    Fatal(String),
+}
+
+impl From<StablecoinExchangeError> for TempoPrecompileError {
```
fgimenez, line 51: we should add conversion here (either impl or `#[from]` as suggested) for `RolesAuthError` too

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -273,7 +273,7 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
         token_out: Address,
         amount_in: u128,
         min_amount_out: u128,
-    ) -> Result<u128, StablecoinExchangeError> {
+    ) -> Result<u128, TempoPrecompileError> {
```
fgimenez, line 276: there are still `expect` calls in this function and other places in this file, are we planning to handle them in a follow-up?

crates/precompiles/src/tip_fee_manager/amm.rs
```diff
@@ -255,20 +250,20 @@ impl<'a, S: PrecompileStorageProvider> TIPFeeAMM<'a, S> {
         amount_user_token: U256,
         amount_validator_token: U256,
         to: Address,
-    ) -> Result<U256, TIPFeeAMMError> {
+    ) -> Result<U256, TempoPrecompileError> {
```
fgimenez, line 253: same here, this function still has `expect` calls

crates/precompiles/src/error.rs
```diff
@@ -0,0 +1,106 @@
...
+    /// Error from TIP20 token
+    #[error("TIP20 token error: {0:?}")]
+    TIP20(TIP20Error),
+
+    /// Error from roles auth
+    #[error("Roles auth error: {0:?}")]
+    RolesAuthError(RolesAuthError),
+
+    /// Error from 403 registry
+    #[error("Roles auth error: {0:?}")]
```
fgimenez:
```suggestion
    #[error("TIP403 registry error: {0:?}")]
```

crates/contracts/src/precompiles/tip20.rs
0xKitsune: Yes this is unneeded, I have updated to remove.

crates/precompiles/src/error.rs
0xKitsune: It looks like when trying to use `#[from]`, I am running into unsatisfied trait bounds and I am unable to implement them since this is a foreign type. Lmk if there is a way you have in mind to resolve or if its best to stick with the current approach.

mattsse: this might be because TIP20Error isnt actually `std::error::Error`?

0xKitsune: Yes I believe this is the case. We are not able to implement this directly since `TIP20` error is a type alias for a foreign type. Lmk if I am overlooking something here though.

klkvr: ah right, didn't realize we don't implement `Error` for this

we could also use `derive_more::From` for this but manual impls are fine too

PR #559, opened 2025-10-27
chore(precompiles): dispatch fatal errors

0xrusowsky:
i was reviewing the PR and thought we should cascade fatal errors directly, as they are not abi-encodable (business logic related)

### DIFF
--- a/crates/precompiles/src/error.rs
+++ b/crates/precompiles/src/error.rs
@@ -87,26 +87,6 @@ impl From<NonceError> for TempoPrecompileError {
     }
 }
 
-impl TempoPrecompileError {
-    pub fn into_precompile_err(self, gas: u64) -> PrecompileResult {
-        let bytes = match self {
-            Self::StablecoinExchange(err) => err.abi_encode().into(),
-            Self::TIP20(err) => err.abi_encode().into(),
-            Self::RolesAuthError(err) => err.abi_encode().into(),
-            Self::TIP403RegistryError(err) => err.abi_encode().into(),
-            Self::TIPAccountRegistrarError(err) => err.abi_encode().into(),
-            Self::FeeManagerError(err) => err.abi_encode().into(),
-            Self::TIPFeeAMMError(err) => err.abi_encode().into(),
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/error.rs
```diff
@@ -97,9 +97,9 @@ impl TempoPrecompileError {
             Self::FeeManagerError(err) => err.abi_encode().into(),
             Self::TIPFeeAMMError(err) => err.abi_encode().into(),
             Self::NonceError(err) => err.abi_encode().into(),
-            Self::Fatal(_) => {
-                // TODO: decide what to return here
-                Bytes::new()
+            Self::Fatal(err) => {
+                // should be returned as `PrecompileError::Fatal` instead
+                panic!("Fatal errors should not be ABI-encoded: {err}");
```
mattsse: is this even reachable from within the evm?

feel like we shouldnt panic here

klkvr: I'd rather have `From<TempoPrecompileError> for PrecompileResult` impl, or we could change this fn to `revert_bytes(&self) -> Option<Bytes>` but panic is not ideal here yeah

crates/precompiles/src/error.rs
```diff
@@ -87,8 +88,8 @@ impl From<NonceError> for TempoPrecompileError {
 }
 
 impl TempoPrecompileError {
-    pub fn abi_encode(&self) -> Bytes {
-        match self {
+    pub fn into_precompile_err(self, gas: u64) -> PrecompileResult {
```
0xKitsune: nit: Is there a reason we are using `into_precompile_err` for converting to `PrecompileResult` rather than `impl From<TempoPrecompileError> for PrecompileResult`?

0xrusowsky: afaik `From<T>` must stick to the `fn from()` sig, which cannot have the gas param.

funnily, i pushed an update to try to make things cleaner while u were reviewing. could u have another look @0xKitsune ?

PR #569, opened 2025-10-28
chore: 2d pool poc

legion2002:


crates/transaction-pool/src/twod/combined.rs
```diff
@@ -0,0 +1,161 @@
...
+/// Combined transaction pool supporting both vanilla and 2D nonce transactions
+///
+/// This pool manages two separate transaction pools:
+/// - Vanilla pool: Standard Ethereum transactions (nonce key 0)
+/// - 2D pool: Parallel transactions with multiple nonce keys (keys 1-N)
+///
+/// The best_transactions() method merges both pools, returning transactions
+/// in descending order of their effective tip per gas, regardless of which
+/// pool they come from.
+pub struct CombinedPool<Client, S = DiskFileBlobStore> {
```
klkvr, line 30: hmm this doesn't implement the pool trait which makes it somewhat harder to reason about

crates/transaction-pool/src/twod/combined.rs
```diff
@@ -0,0 +1,161 @@
...
+        Self {
+            vanilla_pool,
+            twod_pool,
+        }
+    }
+
+    /// Add a transaction to the appropriate pool
+    pub fn add_transaction(
+        &mut self,
+        tx: ValidPoolTransaction<TempoPooledTransaction>,
```
klkvr, line 65: right now `ValidPoolTransaction` can only be produced by inner vanilla pool which'd require doing some nonce validations, how should we avoid this checks for the 2d transactions?

crates/transaction-pool/src/twod/pool.rs
```diff
@@ -0,0 +1,209 @@
...
+    twod::types::{SenderKey, U192},
+};
+
+/// 2D Nonce Transaction Pool - Minimal POC Implementation
+pub struct TwoDimensionalPool {
+    /// Pending transactions ready for execution
+    pending: HashMap<SenderKey, BTreeMap<u64, Arc<ValidPoolTransaction<TempoPooledTransaction>>>>,
+
+    /// Queued transactions with nonce gaps
+    queued: HashMap<SenderKey, BTreeMap<u64, Arc<ValidPoolTransaction<TempoPooledTransaction>>>>,
```
klkvr, line 26: I see, basefee is fixed and we don't have blobs so we only need two subpools which simplifies things a bit

crates/transaction-pool/src/twod/pool.rs
```diff
@@ -0,0 +1,209 @@
...
+
+            if queued_txs.is_empty() {
+                self.queued.remove(&sender_key);
+            }
+        }
+
+        for (seq, tx) in promotions {
+            self.add_to_pending(sender_key, seq, tx.clone());
+
+            let tip = tx.effective_tip_per_gas(0).unwrap_or(0);
```
klkvr, line 174: this should use the actual basefee, otherwise it will always return the `maxFeePerGas`

crates/transaction-pool/src/twod/pool.rs
```diff
@@ -0,0 +1,209 @@
...
+/// 2D Nonce Transaction Pool - Minimal POC Implementation
+pub struct TwoDimensionalPool {
+    /// Pending transactions ready for execution
+    pending: HashMap<SenderKey, BTreeMap<u64, Arc<ValidPoolTransaction<TempoPooledTransaction>>>>,
+
+    /// Queued transactions with nonce gaps
+    queued: HashMap<SenderKey, BTreeMap<u64, Arc<ValidPoolTransaction<TempoPooledTransaction>>>>,
+
+    /// Ordering for best transaction selection (sorted by tip, then by TransactionId)
+    ordering: BTreeMap<(u128, TransactionId), Arc<ValidPoolTransaction<TempoPooledTransaction>>>,
```
klkvr, line 29: anything preventing us from storing `PendingTransaction`s in `pending` set? seems like it would allow to reuse the `CoinbaseTipOrdering`?

klkvr (commented):
I think this does makes sense directionally

essentially we introduce an entirely separate datastructure responsible for storing the transactions with 2d nonces. we've explored something similar for relay and it did make sense

imo as long as we're able to re-use the validation codepaths re-implementing a bunch of nonce tracking is fine

crates/transaction-pool/src/twod/tempo_pool.rs
```diff
@@ -0,0 +1,84 @@
...
+
+/// Simple merge iterator that picks highest fee transaction from either pool
+///
+/// At each step, this iterator:
+/// 1. Peeks at the next transaction from both pools
+/// 2. Compares their effective tips per gas
+/// 3. Returns the transaction with the higher tip
+/// 4. Continues until both pools are exhausted
+pub struct MergeByTip<I1, I2>
+where
```
mattsse, line 17: this makes me think, we could potentially upstream this impl, but perhaps not worth it

crates/transaction-pool/src/twod/types.rs
```diff
@@ -0,0 +1,27 @@
...
+
+use alloy_primitives::Address;
+
+/// 192-bit nonce key (24 bytes)
+pub type U192 = [u8; 24];
+
+/// Unique identifier for a sender + nonce_key combination
+#[derive(Debug, Hash, Eq, PartialEq, Clone, Copy)]
+pub struct SenderKey {
+    pub sender: Address,
```
mattsse, line 11: doing lookups and ordering by address can become fairly expensive, we can def expose the address -> u64 map in the reth pool if it isnt exposed yet

PR #570, opened 2025-10-28
feat(precompiles): TIP20 rewards

### LINKED ISSUE #544: TIP20 Rewards
Update the TIP20 precompile to add a mechanism for pro rata token distributions, allowing anyone to distribute tokens proportionally to all opted-in holders of the token.

Ref: tempoxyz/docs#58

0xKitsune:
Closes #544 

This PR introduces TIP-20 streaming rewards, as described in [tempoxyz/docs#58](https://github.com/tempoxyz/docs/pull/58).

This feature adds an opt-in, streaming reward mechanism to TIP-20 tokens. Funders can start reward streams using `startReward(amount, seconds)`, which linearly distribute `amount` over `seconds` to all opted-in token holders. Rewards are streamed continuously and accrue proportionally to each user’s delegated balance. Streams can be canceled by the funder at any time before completion.

The reward system is fully integrated into normal token operations. Users opt in by calling `setRewardRecipient(recipient)`, which assigns their balance to a reward recipient. When opted in, a user’s balance contributes to the total opted-in supply, which determines how rewards are split. Accrual happens automatically during any balance changing operations (mints, transfers, burns, or changing reward recipients)  so users do not need to need to manually claim rewards.

At a high level overview of the flow:
- `setRewardRecipient` opts the user in or out and updates delegated balances accordingly.  
- `startReward` creates a stream that emits rewards over time, tracked via `totalRewardPerSecond`.  
- `accrue()` updates the global `rewardPerTokenStored` based on elapsed time and total active rewards.  
- `updateRewards(recipient)` settles accrued rewards for that recipient and transfers them from the contract balance. 

Additionally this PR introduces the `TIP20RewardsRegistry` precompile which keeps track of all tip20 token streams ending at a given timestamp. This precompile exposes a `finalize_streams()` function which is called via a system tx at the end of every block.

### DIFF
--- a/crates/precompiles/src/tip20/rewards.rs
+++ b/crates/precompiles/src/tip20/rewards.rs
@@ -1088,7 +1088,10 @@ mod tests {
         assert!(reward_per_token_after >= reward_per_token_before);
         assert_eq!(user_reward_per_token_paid_after, reward_per_token_after);
         assert_eq!(token.get_opted_in_supply()?, mint_amount + reward_amount);
-        assert_eq!(token.get_delegated_balance(&alice)?, mint_amount + reward_amount);
+        assert_eq!(
+            token.get_delegated_balance(&alice)?,
+            mint_amount + reward_amount
+        );
 
         Ok(())
     }
@@ -1218,7 +1221,10 @@ mod tests {
 
         token.update_rewards(&alice)?;
         assert_eq!(token.get_opted_in_supply()?, mint_amount + reward_amount);
-        assert_eq!(token.get_delegated_balance(&alice)?, mint_amount + reward_amount);
+        assert_eq!(
+            token.get_delegated_balance(&alice)?,
+            mint_amount + reward_amount
+        );
         assert_eq!(
             token.get_user_reward_per_token_paid(&alice)?,
             reward_per_token_stored


crates/precompiles/src/tip20/mod.rs
```diff
@@ -658,6 +695,103 @@ impl<'a, S: PrecompileStorageProvider> TIP20Token<'a, S> {
...
+            // TODO: update this to overflow error
+            .ok_or(TIP20Error::supply_cap_exceeded())?;
+        self.set_balance(&contract_address, new_to_balance)?;
+
+        self.accrue()?;
+
+        if call.seconds == 0 {
+            let opted_in_supply = self.get_opted_in_supply()?;
+            if opted_in_supply.is_zero() {
+                return Ok(0);
```
fgimenez: aiui this is different from both the spec and the reference implementation, here we make the transaction succeed but would be returning without emitting `RewardScheduled`. the reference impl reverts here, which does not match the spec either, do we need to align the 3 of them first?

crates/precompiles/src/tip20/mod.rs
```diff
@@ -1028,998 +1169,2137 @@ impl<'a, S: PrecompileStorageProvider> TIP20Token<'a, S> {
...
+            }
+
+            let new_contract_balance = contract_balance - accrued;
+            self.set_balance(&token_address, new_contract_balance)?;
+
+            let recipient_balance = self.get_balance(recipient)?;
+            let new_recipient_balance = recipient_balance + accrued;
+            self.set_balance(recipient, new_recipient_balance)?;
+
+            self.set_user_reward_per_token_paid(recipient, reward_per_token_stored)?;
```
fgimenez: we should emit a `Transfer` event here see https://github.com/tempoxyz/docs/pull/58/files#diff-b6ac723971818592aadd6f17b711e879145451b63effbc4afa36e8f1cf5a83d7R259 and https://github.com/tempoxyz/docs/pull/58/files#diff-6f8f190fa8407d9a69e0a8d928f89c60b677bf618846ac33c3979313a86f73abR542

crates/precompiles/src/tip20/mod.rs
```diff
@@ -1028,998 +1169,2137 @@ impl<'a, S: PrecompileStorageProvider> TIP20Token<'a, S> {
...
+            (delegated * (reward_per_token_stored - user_reward_per_token_paid)) / ACC_PRECISION;
 
-        // Build digest
-        let nonce_slot = mapping_slot(owner, super::slots::NONCES);
-        let nonce = token
+        if accrued > U256::ZERO {
+            let token_address = self.token_address;
+            let contract_balance = self.get_balance(&token_address)?;
+
+            if accrued > contract_balance {
```
fgimenez: in the reference implementation we set `accrued` to `contract_balance` instead of returning an error https://github.com/tempoxyz/docs/pull/58/files#diff-6f8f190fa8407d9a69e0a8d928f89c60b677bf618846ac33c3979313a86f73abR542 i believe returning an error could be problematic, we are not calling `self.set_user_reward_per_token_paid` and next time the tx would revert too (token might be bricked), wdyt? clamping sounds better here

0xKitsune: This has now been updated to 

```
if accrued > contract_balance {
     accrued = contract_balance;
}
```

crates/precompiles/src/tip20/mod.rs
0xKitsune: This has been updated to emit a `Transfer` event

crates/precompiles/src/tip20/mod.rs
0xKitsune: This has been updated to match the reference implementation.

crates/precompiles/src/tip20/rewards.rs
```diff
@@ -0,0 +1,1341 @@
...
+
+        if call.amount == U256::ZERO {
+            return Err(TIP20Error::invalid_amount().into());
+        }
+
+        let from_balance = self.get_balance(msg_sender)?;
+        let new_from_balance = from_balance
+            .checked_sub(call.amount)
+            .ok_or(TempoPrecompileError::under_overflow())?;
+        self.set_balance(msg_sender, new_from_balance)?;
```
howydev: we should probably check if the transfer amount is part of `opted_in_supply` or not, and decrement `opted_in_supply` if it is. I would also move `accrue` to above this logic

crates/precompiles/src/tip20/rewards.rs
```diff
@@ -0,0 +1,1341 @@
...
+            if accrued > contract_balance {
+                accrued = contract_balance;
+            }
+
+            let new_contract_balance = contract_balance - accrued;
+            self.set_balance(&token_address, new_contract_balance)?;
+
+            let recipient_balance = self.get_balance(recipient)?;
+            let new_recipient_balance = recipient_balance + accrued;
+            self.set_balance(recipient, new_recipient_balance)?;
```
howydev: Hmm, just noticed that we don't update the delegated balance here

I think this means that users that refresh their delegated balance would be able to compound rewards, or essentially a penalty for AFK. I generally don't like systems that punish inactivity, but I don't see an easy way to solve this rn

howydev: Actually, is this a bug? Since actual balance > delegated balance, when we do a `transfer` with any amount > delegated balance, i think that would cause an negative underflow

crates/precompiles/src/tip20/rewards.rs
```diff
@@ -0,0 +1,1341 @@
...
+
+        Ok(refund)
+    }
+
+    /// Finalizes expired reward streams by updating the total reward rate.
+    ///
+    /// This function is called to clean up streams that have reached their end time,
+    /// reducing the total reward per second rate by the amount of the expired streams.
+    pub fn finalize_streams(&mut self, end_time: u128) -> Result<(), TempoPrecompileError> {
+        let rate_decrease = self.get_scheduled_rate_decrease_at(end_time);
```
fgimenez, line 488: do we need here an authorization check? https://github.com/tempoxyz/docs/pull/58/files#diff-b6ac723971818592aadd6f17b711e879145451b63effbc4afa36e8f1cf5a83d7R329 also timestamp like https://github.com/tempoxyz/docs/pull/58/files#diff-b6ac723971818592aadd6f17b711e879145451b63effbc4afa36e8f1cf5a83d7R334
```suggestion
        if *caller != TIP20_REWARDS_REGISTRY_ADDRESS {
          return Err(TIP20Error::unauthorized().into());
        }

        let current_time = self.storage.timestamp().to::<u128>();
        if end_time != current_time {
            return Err(TIP20Error::invalid_timestamp().into());
        }

        let rate_decrease = self.get_scheduled_rate_decrease_at(end_time);
```

crates/precompiles/src/tip20/rewards.rs
```diff
@@ -0,0 +1,1341 @@
...
+            .checked_add(call.amount)
+            .ok_or(TempoPrecompileError::under_overflow())?;
+        self.set_balance(&contract_address, new_to_balance)?;
+
+        self.accrue()?;
+
+        if call.seconds == 0 {
+            let opted_in_supply = self.get_opted_in_supply()?;
+            if opted_in_supply.is_zero() {
+                return Err(TIP20Error::no_reward_supplied().into());
```
fgimenez: not sure if returning an error here is ok, according to the spec aiui these should be accepted but permanently locked (not distrubuted) https://github.com/tempoxyz/docs/pull/58/files#diff-b6ac723971818592aadd6f17b711e879145451b63effbc4afa36e8f1cf5a83d7R30 is this correct?

crates/precompiles/src/tip20/dispatch.rs
```diff
@@ -123,6 +123,21 @@ impl<'a, S: PrecompileStorageProvider> Precompile for TIP20Token<'a, S> {
...
+            ITIP20::setRewardRecipientCall::SELECTOR => {
+                mutate_void::<ITIP20::setRewardRecipientCall>(calldata, msg_sender, |s, call| {
+                    self.set_reward_recipient(s, call)
+                })
+            }
+            ITIP20::cancelRewardCall::SELECTOR => {
+                mutate::<ITIP20::cancelRewardCall>(calldata, msg_sender, |s, call| {
+                    self.cancel_reward(s, call)
+                })
+            }
```
fgimenez, line 140: we are missing view functions for `totalRewardPerSecond` and `getStream`
```suggestion
            }
            ITIP20::totalRewardPerSecondCall::SELECTOR => {
                metadata::<ITIP20::totalRewardPerSecondCall>(|| self.total_reward_per_second())
            }
            ITIP20::getStreamCall::SELECTOR => {
                view::<ITIP20::getStreamCall>(calldata, |call| self.get_stream(call))
            }
```

crates/precompiles/src/tip20/rewards.rs
```diff
@@ -0,0 +1,1341 @@
...
+                    id: 0,
+                    amount: call.amount,
+                    durationSeconds: 0,
+                })
+                .into_log_data(),
+            )?;
+
+            Ok(0)
+        } else {
+            let rate = (call.amount * ACC_PRECISION) / U256::from(call.seconds);
```
fgimenez: would be good to add checked arithmetic here

crates/contracts/src/precompiles/tip20.rs
```diff
@@ -79,6 +106,9 @@ sol! {
         event PauseStateUpdate(address indexed updater, bool isPaused);
         event UpdateQuoteToken(address indexed updater, address indexed newQuoteToken);
         event QuoteTokenUpdateFinalized(address indexed updater, address indexed newQuoteToken);
+        event RewardScheduled(address indexed funder, uint64 indexed id, uint256 amount, uint64 durationSeconds);
```
fgimenez: nit, `durationSeconds` should be uint32 according to https://github.com/tempoxyz/docs/pull/58/files#diff-b6ac723971818592aadd6f17b711e879145451b63effbc4afa36e8f1cf5a83d7R67
```suggestion
        event RewardScheduled(address indexed funder, uint64 indexed id, uint256 amount, uint32 durationSeconds);
```

crates/precompiles/src/tip20/rewards.rs
0xKitsune: This has been updated to match the latest spec.

crates/precompiles/src/tip20/dispatch.rs
```diff
@@ -123,6 +123,27 @@ impl<'a, S: PrecompileStorageProvider> Precompile for TIP20Token<'a, S> {
...
+            ITIP20::totalRewardPerSecondCall::SELECTOR => {
+                metadata::<ITIP20::totalRewardPerSecondCall>(|| self.total_reward_per_second())
+            }
+            ITIP20::getStreamCall::SELECTOR => {
```
howydev: There might be a few more missing ones - `optedInSupply`, `lastUpdateTime`, `streams`, `scheduledRateDecrease`, `rewardRecipientOf`, `userRewardPerTokenPaid`, `delegatedBalance`, and would also like the new `accruedBalanceOf` to be public (will likely need that in DEX rewards distribution work)

crates/precompiles/src/tip20/rewards.rs
```diff
@@ -0,0 +1,1353 @@
...
+        }
+
+        let delegated = self.get_delegated_balance(recipient)?;
+        let reward_per_token_stored = self.get_reward_per_token_stored()?;
+        let user_reward_per_token_paid = self.get_user_reward_per_token_paid(recipient)?;
+
+        let mut accrued =
+            (delegated * (reward_per_token_stored - user_reward_per_token_paid)) / ACC_PRECISION;
+
+        self.set_user_reward_per_token_paid(recipient, reward_per_token_stored)?;
```
howydev, line 244: Really small nit: if we compare `reward_per_token_stored` to `user_reward_per_token_paid` before doing the store, we might avoid an sstore in the case when the value is latest. I'm not able to see immediately how often this case would happen though

crates/precompiles/src/tip20/rewards.rs
0xKitsune: Updated to use checked math everywhere.

crates/precompiles/src/tip20/rewards.rs
0xKitsune: This has been updated to reflect the latest changes in the spec.

crates/precompiles/src/tip20/rewards.rs
0xKitsune: This function is not actually publicly exposed and is only callable by other precompiles, we can add a check here regardless to ensure other precompiles cant call this.

crates/precompiles/src/tip20/rewards.rs
0xKitsune: With the latest changes the implementation now matches the spec 
https://github.com/tempoxyz/docs/blob/dr/reward-distribution/specs/src/TIP20.sol#L595-L597

@danrobinson can you clarify if this is the expected behavior and if the spec/docs should be updated?

crates/precompiles/src/tip20/dispatch.rs
0xKitsune: At the moment this PR is blocking the precompile refactor. Wdyt about adding this in a fast follow PR to not block?

crates/precompiles/src/tip20/rewards.rs
howydev: Whoops, i did not see the spec

samczsun's review comment was that we should just revert, so it wouldn't be locked. That made sense to me and seemed like a small lift to not brick tokens, so I updated the TIP20 ref impl that way

I'm leaning towards the approach in ref impl, but I might not have full context, will defer to Dan on this

crates/precompiles/src/tip20/dispatch.rs
howydev: Works for me! I don't think these would be used for testnet 5

crates/precompiles/src/tip20/rewards.rs
danrobinson: Yeah I guess I agree that if seconds is 0 and opted in supply is 0 then revert makes sense. But if seconds > 0 then don't.

PR #571, opened 2025-10-28
feat(xtask): add devnet generator task

Zygimantass:
Adds a devnet generator task for generating configs necessary for a Kubernetes-hosted cluster.

The difference from the old Commonware generator is that it:
- generates JSON output instead of separate TOML files
- also generates `discv5` identities for Reth

### DIFF
--- a/bin/tempo-bench/src/cmd/max_tps.rs
+++ b/bin/tempo-bench/src/cmd/max_tps.rs
@@ -248,7 +248,7 @@ fn send_transactions(
                         Ok(Ok(_)) => {
                             tx_counter.fetch_add(1, Ordering::Relaxed);
                         }
-                        Ok(Err(e)) => eprintln!("Failed to send transaction: {}", e),
+                        Ok(Err(e)) => eprintln!("Failed to send transaction: {e}"),
                         Err(_) => eprintln!("Tx send timed out"),
                     }
                 }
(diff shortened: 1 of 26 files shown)


xtask/src/devnet.rs
```diff
@@ -0,0 +1,166 @@
...
+use commonware_cryptography::{PrivateKeyExt as _, Signer as _};
+use eyre::{Context, ensure};
+use indexmap::IndexMap;
+use itertools::multizip;
+use rand::SeedableRng;
+use reth_network_peers::pk2id;
+use secp256k1::SECP256K1;
+use serde::Serialize;
+use tempo_commonware_node_config::Config;
+use tempo_commonware_node_cryptography::PrivateKey;
```
SuperFluffy: This crate no longer exists. Uuse `commonware_cryptography::ed25519::PrivateKey` instead (maybe double check the import path).

xtask/src/devnet.rs
```diff
@@ -0,0 +1,166 @@
...
+            })?
+            .next()
+            .is_none();
+        ensure!(
+            target_is_empty,
+            "target directory `{output}` is not empty; delete all its contents or rerun command with --force",
+        );
+    }
+
+    let mut rng = rand::rngs::StdRng::seed_from_u64(rand::random::<u64>());
```
SuperFluffy: IMO just use the thread rng directly instead of this extra step, see my commnet below

xtask/src/devnet.rs
```diff
@@ -0,0 +1,166 @@
...
+    let mut signers = (0..nodes.len())
+        .map(|_| PrivateKey::from_rng(&mut rng))
```
SuperFluffy:
```suggestion
        .map(|_| PrivateKey::from_rng(&mut rand::thread_rng()))
```

PR #573, opened 2025-10-28
feat(precompiles): add logic to get stablecoin exchange orderbook keys

0xKitsune:
ref #536

This PR introduces a new storage variable called `bookKeys`, to the stablecoin exchange. This new array keeps track of all order book keys associated with created trading pairs. A helper function has also been added to the interface to retrieve the full list of book keys.

### DIFF
--- a/crates/precompiles/src/stablecoin_exchange/mod.rs
+++ b/crates/precompiles/src/stablecoin_exchange/mod.rs
@@ -367,6 +367,8 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
     }
 
     /// Add book key to book keys array
+    /// This function adds the specified book key to the `book_keys` array in storage
+    /// and increments the length of the array
     fn push_to_book_keys(&mut self, book_key: B256) -> Result<(), TempoPrecompileError> {
         let length = self.storage.sload(self.address, slots::BOOK_KEYS_LENGTH)?;
         self.storage.sstore(


crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -366,6 +366,34 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
...
+        let mut i = U256::ZERO;
+        while i < length {
+            let book_key = self
+                .storage
+                .sload(self.address, slots::BOOK_KEYS_BASE + i)?;
+            book_keys.push(B256::from(book_key));
+            i += U256::ONE;
+        }
+        Ok(book_keys)
+    }
```
klkvr, line 397: should we start exploring proper approach for pricing methods like this? once we have a lot of pairs transactions invoking this might be pretty resource-consuming

mattsse, line 397: for rpc we might not necessarily need this, but this also doesnt hurt, at the very least useful as a reference

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -366,6 +366,34 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
         Orderbook::from_storage(pair_key, self.storage, self.address)
     }
 
+    /// Add book key to book keys array
+    fn push_to_book_keys(&mut self, book_key: B256) -> Result<(), TempoPrecompileError> {
```
mattsse, line 372: can document that this increases the book counter and adds a new mapping entry

PR #578, opened 2025-10-29
feat(rpc): impl `dex_` namespace

Rjected:
This implements both `dex_getOrders` and `dex_getOrderbooks`. These are paginated and have filtering. They do not support sorting.

Currently `dex_getOrders` only returns one type of order, either bids or asks, depending on the filter.

### DIFF
--- a/crates/node/src/rpc/dex/types.rs
+++ b/crates/node/src/rpc/dex/types.rs
@@ -66,9 +66,34 @@ pub struct OrdersFilters {
     /// Filter by quote token
     pub quote_token: Option<Address>,
     /// Remaining amount in range
-    pub remaining: Option<FilterRange<u128>>,
+    pub remaining: Option<RemainingFilterRange>,
     /// Tick in range (from -2000 to 2000)
     pub tick: Option<FilterRange<Tick>>,
+}
+
+/// FilterRange type for u128, so that we can serialize the u128s as QUANTITY.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct RemainingFilterRange {
+    #[serde(with = "alloy_serde::quantity::opt")]
+    pub min: Option<u128>,
+    #[serde(with = "alloy_serde::quantity::opt")]
+    pub max: Option<u128>,
+}
+
+impl RemainingFilterRange {
+    /// Checks if a value is within this range (inclusive)
+    pub fn in_range(&self, value: u128) -> bool {
+        if self.min.as_ref().is_some_and(|min| &value < min) {
+            return false;
+        }
+
+        if self.max.as_ref().is_some_and(|max| &value > max) {
+            return false;
+        }
+
+        true
+    }
 }
 
 #[derive(Clone, Debug, Default, PartialEq, Eq, Hash, Serialize, Deserialize)]


crates/node/src/rpc/dex/types.rs
jxom, line 1: We will need to serialize `remaining` filter range as quantity btw, I think.

crates/node/src/rpc/dex/mod.rs
```diff
@@ -1,26 +1,223 @@
...
+fn get_book_keys_for_iteration(
+    exchange: &mut StablecoinExchange<'_, EvmPrecompileStorageProvider<'_>>,
+    base_token: Option<Address>,
+    quote_token: Option<Address>,
+) -> Result<Vec<B256>, DexApiError> {
+    match (base_token, quote_token) {
+        (Some(base), Some(quote)) => Ok(vec![compute_book_key(base, quote)]),
+        _ => exchange.get_book_keys().map_err(DexApiError::Precompile),
+    }
+}
```
mattsse: can we move this end of file

crates/node/src/rpc/dex/mod.rs
```diff
@@ -35,23 +232,494 @@ impl<EthApi> TempoDex<EthApi> {
...
+
 #[async_trait::async_trait]
-impl<EthApi: RpcNodeCore> TempoDexApiServer for TempoDex<EthApi> {
-    async fn orders(&self, _params: PaginationParams<OrdersFilters>) -> RpcResult<OrdersResponse> {
-        Err(internal_rpc_err("unimplemented"))
+impl<
+    EthApi: RpcNodeCore<Evm = TempoEvmConfig, Primitives: NodePrimitives<BlockHeader = TempoHeader>>,
+> TempoDexApiServer for TempoDex<EthApi>
+{
```
mattsse, line 68: this should be the first thing in this file

crates/node/src/rpc/dex/mod.rs
```diff
@@ -35,23 +232,494 @@ impl<EthApi> TempoDex<EthApi> {
...
+        .is_some_and(|tick_range| !tick_range.in_range(order.tick))
+    {
+        return false;
+    }
+
+    true
+}
+
+/// Parses a cursor string into a u128 for orders
+fn parse_order_cursor(cursor: &str) -> Result<u128, DexApiError> {
```
mattsse: can we mention that this parses quantity

crates/node/src/rpc/dex/mod.rs
```diff
@@ -1,26 +1,223 @@
...
+/// Gets book keys to iterate over. If both base and quote are specified, returns only that book.
+/// Otherwise returns all book keys (filtering happens later during iteration).
+fn get_book_keys_for_iteration(
+    exchange: &mut StablecoinExchange<'_, EvmPrecompileStorageProvider<'_>>,
+    base_token: Option<Address>,
+    quote_token: Option<Address>,
+) -> Result<Vec<B256>, DexApiError> {
+    match (base_token, quote_token) {
+        (Some(base), Some(quote)) => Ok(vec![compute_book_key(base, quote)]),
+        _ => exchange.get_book_keys().map_err(DexApiError::Precompile),
```
mattsse: for a followup, I believe these keys can be partially cached, so that the only thing the request needs to check is the length of the order book

because afaik, the keys dont change?

we could also monitor the BOOK_KEYS_LENGTH slot and then hotload the books in the background, but loading the for now should be okay

crates/node/src/rpc/dex/mod.rs
```diff
@@ -35,23 +232,494 @@ impl<EthApi> TempoDex<EthApi> {
...
     pub fn provider(&self) -> &EthApi::Provider {
         self.eth_api.provider()
     }
+
+    /// Creates an `EvmPrecompileStorageProvider` at the given block.
+    /// This handles the boilerplate of creating the EVM context and state provider.
+    fn with_storage_at_block<F, R>(&self, at: BlockId, f: F) -> Result<R, DexApiError>
+    where
+        F: FnOnce(&mut EvmPrecompileStorageProvider<'_>) -> Result<R, DexApiError>,
+    {
```
mattsse, line 210: this fn does a lot of IO and must be performed on a blocking task, currently this is executed in a regular task

crates/node/src/rpc/dex/mod.rs
```diff
@@ -35,23 +232,494 @@ impl<EthApi> TempoDex<EthApi> {
...
+        self.with_exchange_at_block(BlockNumberOrTag::Latest.into(), |exchange| {
+            let book_key = compute_book_key(base, quote);
+            exchange.books(book_key).map_err(DexApiError::Precompile)
+        })
+    }
+}
+
+/// Checks if an orderbook matches the given base and/or quote token filters
+fn orderbook_matches_tokens(
+    book: &PrecompileOrderbook,
```
mattsse: can we make this a fn of `PrecompileOrderbook`?

crates/node/src/rpc/dex/mod.rs
```diff
@@ -35,23 +232,494 @@ impl<EthApi> TempoDex<EthApi> {
...
+                // Use the last item's cursor as the next cursor
+                orderbooks
+                    .get(limit)
+                    .map(|book| format!("0x{}", compute_book_key(book.base, book.quote)))
+            } else {
+                None
+            };
+
+            // Return only up to limit items
+            let items = orderbooks.into_iter().take(limit).collect();
```
mattsse: I think this section is the same as

```
            let mut iter = orderbooks.into_iter();
            let items = iter.by_ref().take(limit).collect();
            let next_cursor = iter.next().map( |book| format!("0x{}", compute_book_key(book.base, book.quote)));
```

crates/node/src/rpc/dex/mod.rs
```diff
@@ -35,23 +232,494 @@ impl<EthApi> TempoDex<EthApi> {
...
+            // Take limit + 1 to check if there's a next page
+            for key in keys.into_iter().skip(start_idx).take(limit + 1) {
+                let book = exchange.books(key).map_err(DexApiError::Precompile)?;
+
+                // Apply filters if present
+                if let Some(ref filter) = params.filters
+                    && !orderbook_matches_filter(&book, filter)
+                {
+                    continue;
+                }
```
mattsse, line 293: not exactly sure this is correct because this skips books that don't match but the limit is applied to _all_ books

I assume the limit means `limit of matching books`?

if so then we can remove the `.take(limit + 1)` here because we already have 

```
                if orderbooks.len() > limit {
                    break;
                }
```

crates/node/src/rpc/dex/mod.rs
```diff
@@ -35,23 +232,494 @@ impl<EthApi> TempoDex<EthApi> {
...
+                // Determine which books to iterate based on filter
+                let base_token = params.filters.as_ref().and_then(|f| f.base_token);
+                let quote_token = params.filters.as_ref().and_then(|f| f.quote_token);
+                let book_keys =
+                    get_book_keys_for_iteration(&mut exchange, base_token, quote_token)?;
+
+                let is_bid = params
+                    .filters
+                    .as_ref()
+                    .is_none_or(|f| f.is_bid.unwrap_or(false));
```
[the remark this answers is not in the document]
Rjected: not in scope, this PR will do only one side as determined by the filter or the default (asks)

crates/node/src/rpc/dex/mod.rs
Rjected: right, removed the take

crates/node/src/rpc/dex/mod.rs
Rjected: yep

PR #581, opened 2025-10-29
feat: validator shared blockspace

klkvr:
Implements support for https://github.com/tempoxyz/docs/blob/main/specs/reserved-gas.md

Changes in the PR:
- Subblocks service is introduced. The way it works is the following:
    - Collect subblock transactions
    - Listen to all consensus events and record next block's parent (current tip) and proposer
    - Once next block parent is available, start building subblock by executing all transactions on top of latest block state. Once subblock is built, sign and send it to next proposer
    - Proposer is expected to collect and verify all received blocks, and expose them to payload building task
- Block executor is changed to track `BlockSection` and infer section boundaries from transactions that are executed. It is also responsible for tracking all of the submitted subblocks and verifying the subblocks system transaction.

Proper subblocks verification requires knowledge of the current validator set which is only available to consensus layer. For that reason, `TempoBlockExecutor` expects to be provided with optional set of validators:
https://github.com/tempoxyz/tempo/blob/31d6b942889d87c2a1079bfa6cd5ea07171bdff4/crates/evm/src/block.rs#L96
When it is not provided (primarily when processing finalized blocks), this check will be skipped. This means we should now be careful with sending FCU with untrusted `head` block (which I believe we never do right now anyway)

This also includes 2 drive-by fixes:
- Blocks that were just built are now sent as `newPayload` which is a backport of https://github.com/tempoxyz/tempo/commit/9cf5be04ee03c6da0c8f880f695eac3dfa802711
- Block executor is changed to only account for non-payment transactions in `generalGasLimit` which was a divergence from the spec that is now fixed

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,7 @@ members = [
   "bin/tempo",
   "bin/tempo-bench",
   "bin/tempo-sidecar",
+  "crates/alloy",
   "crates/chainspec",
   "crates/commonware-node",
   "crates/commonware-node-config",
@@ -90,6 +91,8 @@ lto = "fat"
 codegen-units = 1
 
 [workspace.dependencies]
+tempo-alloy = { path = "crates/alloy" }
+
 tempo-node = { path = "crates/node" }
 
 tempo-chainspec = { path = "crates/chainspec" }
@@ -108,72 +111,73 @@ tempo-contracts = { path = "crates/contracts" }
 tempo-telemetry-util = { path = "crates/telemetry-util" }
 tempo-transaction-pool = { path = "crates/transaction-pool" }
 
-reth = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-basic-payload-builder = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-chainspec = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-cli = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-cli-commands = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-cli-util = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-codecs = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-codecs-derive = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-consensus = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-consensus-common = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-db = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-db-api = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-engine-primitives = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-errors = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-ethereum-engine-primitives = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-evm = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-evm-ethereum = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-ethereum-forks = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-ethereum-primitives = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1", default-features = false }
-reth-execution-types = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-metrics = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-node-builder = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-node-core = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-node-ethereum = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-network-peers = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-payload-builder = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-payload-primitives = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-primitives-traits = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1", default-features = false }
-reth-provider = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-rpc-builder = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-rpc-convert = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-rpc-eth-api = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-storage-api = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-tasks = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-tracing = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-trie-db = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-node-api = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-rpc = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-rpc-api = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-rpc-eth-types = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-rpc-server-types = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-cli-runner = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-node-metrics = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-network = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-transaction-pool = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-engine-local = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-ethereum-consensus = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-ethereum = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-ethereum-cli = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
-reth-e2e-test-utils = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1" }
+reth = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-basic-payload-builder = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-chainspec = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-cli = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-cli-commands = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-cli-util = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-codecs = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-codecs-derive = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-consensus = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-consensus-common = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-db = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-db-api = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-engine-primitives = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-errors = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-ethereum-engine-primitives = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-evm = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-evm-ethereum = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-ethereum-forks = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-ethereum-primitives = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb", default-features = false }
+reth-execution-types = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-metrics = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-node-builder = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-node-core = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-node-ethereum = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-network-peers = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-payload-builder = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-payload-primitives = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-primitives-traits = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb", default-features = false }
+reth-provider = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-rpc-builder = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-rpc-convert = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-rpc-eth-api = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-storage-api = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-tasks = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-tracing = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-trie-db = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-node-api = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-rpc = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-rpc-api = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-rpc-eth-types = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-rpc-server-types = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-cli-runner = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-node-metrics = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-network = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-transaction-pool = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-engine-local = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-ethereum-consensus = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-ethereum = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-ethereum-cli = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
+reth-e2e-test-utils = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb" }
 
 alloy-network = { version = "1.0.41" }
 alloy-rpc-types-eth = { version = "1.0.41" }
-reth-revm = { git = "https://github.com/paradigmxyz/reth", rev = "3827e5cb1", features = [
+reth-revm = { git = "https://github.com/paradigmxyz/reth", rev = "bb694fb", features = [
   "std",
   "optional-checks",
 ] }
-revm = "30.2.0"
+revm = "31.0"
 
 alloy = { version = "1.0.41", default-features = false }
-alloy-evm = "0.22.3"
+alloy-evm = "0.23.0"
 alloy-consensus = "1.0.41"
 alloy-eips = "1.0.41"
 alloy-genesis = "1.0.41"
-alloy-hardforks = "0.4.0"
+alloy-hardforks = "0.4.3"
 alloy-primitives = "1.4.1"
+alloy-provider = { version = "1.0.41", default-features = false }
 alloy-sol-types = "1.4.1"
 alloy-rlp = "0.3.12"
 alloy-rpc-types-engine = "1.0.41"
--- a/crates/revm/src/handler.rs
+++ b/crates/revm/src/handler.rs
@@ -26,7 +26,7 @@ use revm::{
         instructions::utility::IntoAddress,
         interpreter::EthInterpreter,
     },
-    primitives::{KECCAK_EMPTY, eip7702, hardfork::SpecId as RevmSpecId},
+    primitives::{eip7702, hardfork::SpecId as RevmSpecId},
     state::Bytecode,
 };
 use tempo_contracts::{DEFAULT_7702_DELEGATE_ADDRESS, precompiles::FeeManagerError};
@@ -418,7 +418,7 @@ where
                 };
 
                 // 4. Add `authority` to `accessed_addresses` (warm the account)
-                let mut authority_acc = journal.load_account_code(authority)?;
+                let mut authority_acc = journal.load_account_with_code_mut(authority)?;
 
                 // 5. Verify the code of `authority` is either empty or already delegated.
                 if let Some(bytecode) = &authority_acc.info.code {
@@ -440,23 +440,11 @@ where
                     refunded_accounts += 1;
                 }
 
-                // 8. Set the code of `authority` to be `0xef0100 || address`
+                // 8. Set the code of `authority` to be `0xef0100 || address`. This is a delegation designation.
                 //  * As a special case, if `address` is `0x0000000000000000000000000000000000000000` do not write the designation.
                 //    Clear the accounts code and reset the account's code hash to the empty hash `0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470`.
-                let address = authorization.address;
-                let (bytecode, hash) = if address.is_zero() {
-                    (Bytecode::default(), KECCAK_EMPTY)
-                } else {
-                    let bytecode = Bytecode::new_eip7702(address);
-                    let hash = bytecode.hash_slow();
-                    (bytecode, hash)
-                };
-                authority_acc.info.code_hash = hash;
-                authority_acc.info.code = Some(bytecode);
-
                 // 9. Increase the nonce of `authority` by one.
-                authority_acc.info.nonce = authority_acc.info.nonce.saturating_add(1);
-                authority_acc.mark_touch();
+                authority_acc.delegate(*authorization.address());
             }
 
             let refunded_gas =
@@ -479,13 +467,12 @@ where
         let account_balance = get_token_balance(journal, self.fee_token, self.fee_payer)?;
 
         // Load caller's account
-        let caller_account = journal.load_account_code(tx.caller())?.data;
+        let mut caller_account = journal.load_account_with_code_mut(tx.caller())?.data;
 
-        let account_info = &mut caller_account.info;
-        if account_info.has_no_code_and_nonce() {
-            account_info.set_code_and_hash(
-                Bytecode::new_eip7702(DEFAULT_7702_DELEGATE_ADDRESS),
+        if caller_account.info.has_no_code_and_nonce() {
+            caller_account.set_code(
                 DEFAULT_7702_DELEGATE_CODE_HASH,
+                Bytecode::new_eip7702(DEFAULT_7702_DELEGATE_ADDRESS),
             );
         }
 
@@ -497,7 +484,7 @@ where
 
         // Validate account nonce and code (EIP-3607) using upstream helper
         pre_execution::validate_account_nonce_and_code(
-            &mut caller_account.info,
+            &caller_account.info,
             tx.nonce(),
             cfg.is_eip3607_disabled(),
             // skip nonce check if 2D nonce is used
@@ -505,7 +492,7 @@ where
         )?;
 
         // modify account nonce and touch the account.
-        caller_account.mark_touch();
+        caller_account.touch();
 
         if !nonce_key.is_zero() {
             let internals = EvmInternals::new(journal, block);
@@ -559,7 +546,7 @@ where
             //
             // Always bump nonce for AA transactions.
             if tx.aa_tx_env.is_some() || tx.kind().is_call() {
-                caller_account.info.nonce = caller_account.info.nonce.saturating_add(1);
+                caller_account.bump_nonce();
             }
         }
 
@@ -1103,7 +1090,7 @@ mod tests {
 
         // Set up initial balance
         let balance_slot = mapping_slot(account, tip20::slots::BALANCES);
-        journal.warm_account(token)?;
+        journal.load_account(token)?;
         journal
             .sstore(token, balance_slot, expected_balance)
             .unwrap();
@@ -1123,7 +1110,7 @@ mod tests {
         let initial_balance = U256::random();
 
         let sender_slot = mapping_slot(sender, tip20::slots::BALANCES);
-        journal.warm_account(token)?;
+        journal.load_account(token)?;
         journal.sstore(token, sender_slot, initial_balance).unwrap();
         let sender_balance = get_token_balance(&mut journal, token, sender).unwrap();
         assert_eq!(sender_balance, initial_balance);
@@ -1155,7 +1142,7 @@ mod tests {
 
         // Set validator token
         let validator_slot = mapping_slot(validator, tip_fee_manager::slots::VALIDATOR_TOKENS);
-        ctx.journaled_state.warm_account(TIP_FEE_MANAGER_ADDRESS)?;
+        ctx.journaled_state.load_account(TIP_FEE_MANAGER_ADDRESS)?;
         ctx.journaled_state
             .sstore(
                 TIP_FEE_MANAGER_ADDRESS,
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -45,7 +45,7 @@ jobs:
   
   crate-checks:
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 60
     if: github.event_name == 'push'
     steps:
       - uses: actions/checkout@v5
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,7 +27,7 @@ jobs:
       - uses: mozilla-actions/sccache-action@v0.0.9
       - uses: taiki-e/install-action@nextest
       - name: Run tests
-        run: cargo nextest run --no-fail-fast
+        run: cargo nextest run --no-fail-fast -j num-cpus
 
   msrv:
     name: MSRV
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,15 +3,6 @@
 version = 4
 
 [[package]]
-name = "addr2line"
-version = "0.25.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1b5d307320b3181d6d7954e663bd7c774a838b8220fe0593c86d9fb09f498b4b"
-dependencies = [
- "gimli",
-]
-
-[[package]]
 name = "adler2"
 version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -59,7 +50,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5a15f179cd60c4584b8a8c596927aadc462e27f2ca70c04e0071964a73ba7a75"
 dependencies = [
  "cfg-if",
- "getrandom 0.3.3",
+ "getrandom 0.3.4",
  "once_cell",
  "version_check",
  "zerocopy",
@@ -67,11 +58,20 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.3"
+version = "1.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8e60d3430d3a69478ad0993f19238d2df97c507009a52b3c10addcd7f6bcb916"
+checksum = "ddd31a130427c27518df266943a5308ed92d4b226cc639f5a8f1002816174301"
 dependencies = [
  "memchr",
+]
+
+[[package]]
+name = "aligned-vec"
+version = "0.6.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dc890384c8602f339876ded803c97ad529f3842aba97f6392b3dba0dd171769b"
+dependencies = [
+ "equator",
 ]
 
 [[package]]
@@ -97,9 +97,9 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "alloy"
-version = "1.0.41"
+version = "1.0.42"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ae62e633fa48b4190af5e841eb05179841bb8b713945103291e2c0867037c0d1"
+checksum = "0e84ab07ca0c3210734019e4d0e5392952ed574196ab0f904ab9c7ba0ae15595"
 dependencies = [
  "alloy-consensus",
  "alloy-contract",
@@ -123,9 +123,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-chains"
-version = "0.2.14"
+version = "0.2.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bf01dd83a1ca5e4807d0ca0223c9615e211ce5db0a9fd1443c2778cacf89b546"
+checksum = "6068f356948cd84b5ad9ac30c50478e433847f14a50714d2b68f15d052724049"
 dependencies = [
  "alloy-primitives",
  "alloy-rlp",
@@ -138,9 +138,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-consensus"
-version = "1.0.41"
+version = "1.0.42"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b9b151e38e42f1586a01369ec52a6934702731d07e8509a7307331b09f6c46dc"
+checksum = "3abecb92ba478a285fbf5689100dbafe4003ded4a09bf4b5ef62cca87cd4f79e"
 dependencies = [
  "alloy-eips",
  "alloy-primitives",
@@ -165,9 +165,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-consensus-any"
-version = "1.0.41"
+version = "1.0.42"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6e2d5e8668ef6215efdb7dcca6f22277b4e483a5650e05f5de22b2350971f4b8"
+checksum = "2e864d4f11d1fb8d3ac2fd8f3a15f1ee46d55ec6d116b342ed1b2cb737f25894"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
@@ -180,9 +180,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-contract"
-version = "1.0.41"
+version = "1.0.42"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "630288cf4f3a34a8c6bc75c03dce1dbd47833138f65f37d53a1661eafc96b83f"
+checksum = "c98d21aeef3e0783046c207abd3eb6cb41f6e77e0c0fc8077ebecd6df4f9d171"
 dependencies = [
  "alloy-consensus",
  "alloy-dyn-abi",
@@ -203,9 +203,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-core"
-version = "1.4.0"
+version = "1.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "575053cea24ea8cb7e775e39d5c53c33b19cfd0ca1cf6c0fd653f3d8c682095f"
+checksum = "5ca96214615ec8cf3fa2a54b32f486eb49100ca7fe7eb0b8c1137cd316e7250a"
 dependencies = [
  "alloy-dyn-abi",
  "alloy-json-abi",
@@ -277,9 +277,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-eips"
-version = "1.0.41"
+version = "1.0.42"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e5434834adaf64fa20a6fb90877bc1d33214c41b055cc49f82189c98614368cc"
+checksum = "07d9a64522a0db6ebcc4ff9c904e329e77dd737c2c25d30f1bdc32ca6c6ce334"
 dependencies = [
  "alloy-eip2124",
  "alloy-eip2930",
@@ -302,9 +302,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-evm"
-version = "0.22.5"
+version = "0.23.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "28bd79e109f2b3ff81ed1a93ed3d07cf175ca627fd4fad176df721041cc40dcc"
+checksum = "88d4291974e3564db30f1d2bcb3ba4a53dbc927e9a6fce2edaf389a712204fbd"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
@@ -324,9 +324,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-genesis"
-version = "1.0.41"
+version = "1.0.42"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "919a8471cfbed7bcd8cf1197a57dda583ce0e10c6385f6ff4e8b41304b223392"
+checksum = "675b163946b343ed2ddde4416114ad61fabc8b2a50d08423f38aa0ac2319e800"
 dependencies = [
  "alloy-eips",
  "alloy-primitives",
@@ -338,9 +338,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-hardforks"
-version = "0.4.0"
+version = "0.4.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4b16ee6b2c7d39da592d30a5f9607a83f50ee5ec2a2c301746cc81e91891f4ca"
+checksum = "51e7f93a60ef3d867c93d43442ef3f2d8a1095450131c3d4e16bbbbf2166b9bd"
 dependencies = [
  "alloy-chains",
  "alloy-eip2124",
@@ -364,9 +364,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-json-rpc"
-version = "1.0.41"
+version = "1.0.42"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d7c69f6c9c68a1287c9d5ff903d0010726934de0dac10989be37b75a29190d55"
+checksum = "f87b774478fcc616993e97659697f3e3c7988fdad598e46ee0ed11209cd0d8ee"
 dependencies = [
  "alloy-primitives",
  "alloy-sol-types",
@@ -379,9 +379,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-network"
-version = "1.0.41"
+version = "1.0.42"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8eaf2ae05219e73e0979cb2cf55612aafbab191d130f203079805eaf881cca58"
+checksum = "d5d6ed73d440bae8f27771b7cd507fa8f10f19ddf0b8f67e7622a52e0dbf798e"
 dependencies = [
  "alloy-consensus",
  "alloy-consensus-any",
@@ -405,9 +405,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-network-primitives"
-version = "1.0.41"
+version = "1.0.42"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e58f4f345cef483eab7374f2b6056973c7419ffe8ad35e994b7a7f5d8e0c7ba4"
+checksum = "219dccd2cf753a43bd9b0fbb7771a16927ffdb56e43e3a15755bef1a74d614aa"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
@@ -429,9 +429,9 @@ dependencies = [
  "const-hex",
  "derive_more",
  "foldhash 0.2.0",
- "getrandom 0.3.3",
+ "getrandom 0.3.4",
  "hashbrown 0.16.0",
- "indexmap 2.11.4",
+ "indexmap 2.12.0",
  "itoa",
  "k256",
  "keccak-asm",
@@ -448,9 +448,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-provider"
-version = "1.0.41"
+version = "1.0.42"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "de2597751539b1cc8fe4204e5325f9a9ed83fcacfb212018dfcfa7877e76de21"
+checksum = "f0ef8cbc2b68e2512acf04b2d296c05c98a661bc460462add6414528f4ff3d9b"
 dependencies = [
  "alloy-chains",
  "alloy-consensus",
@@ -494,9 +494,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-pubsub"
-version = "1.0.41"
+version = "1.0.42"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "06e45a68423e732900a0c824b8e22237db461b79d2e472dd68b7547c16104427"
+checksum = "be028fb1c6c173f5765d0baa3580a11d69826ea89fe00ee5c9d7eddb2c3509cd"
 dependencies = [
  "alloy-json-rpc",
  "alloy-primitives",
@@ -533,14 +533,14 @@ checksum = "64b728d511962dda67c1bc7ea7c03736ec275ed2cf4c35d9585298ac9ccf3b73"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.106",
+ "syn 2.0.108",
 ]
 
 [[package]]
 name = "alloy-rpc-client"
-version = "1.0.41"
+version = "1.0.42"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "edf8eb8be597cfa8c312934d2566ec4516f066d69164f9212d7a148979fdcfd8"
+checksum = "2a0f67d1e655ed93efca217213340d21cce982333cc44a1d918af9150952ef66"
 dependencies = [
  "alloy-json-rpc",
  "alloy-primitives",
@@ -564,9 +564,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-rpc-types"
-version = "1.0.41"
+version = "1.0.42"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "339af7336571dd39ae3a15bde08ae6a647e62f75350bd415832640268af92c06"
+checksum = "fe106e50522980bc9e7cc9016f445531edf1a53e0fdba904c833b98c6fdff3f0"
 dependencies = [
  "alloy-primitives",
  "alloy-rpc-types-anvil",
@@ -581,9 +581,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-rpc-types-admin"
-version = "1.0.41"
+version = "1.0.42"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "19b33cdc0483d236cdfff763dae799ccef9646e94fb549a74f7adac6a7f7bb86"
+checksum = "e8b67bf1ed8cac6fde7dd017ca0a1c33be846e613a265956089f983af1354f13"
 dependencies = [
  "alloy-genesis",
  "alloy-primitives",
@@ -593,9 +593,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-rpc-types-anvil"
-version = "1.0.41"
+version = "1.0.42"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "83d98fb386a462e143f5efa64350860af39950c49e7c0cbdba419c16793116ef"
+checksum = "c1cf94d581b3aa13ebacb90ea52e0179985b7c20d8a522319e7d40768d56667a"
 dependencies = [
  "alloy-primitives",
  "alloy-rpc-types-eth",
@@ -605,9 +605,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-rpc-types-any"
-version = "1.0.41"
+version = "1.0.42"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fbde0801a32d21c5f111f037bee7e22874836fba7add34ed4a6919932dd7cf23"
+checksum = "425e14ee32eb8b7edd6a2247fe0ed640785e6eba75af27db27f1e6220c15ef0d"
 dependencies = [
  "alloy-consensus-any",
  "alloy-rpc-types-eth",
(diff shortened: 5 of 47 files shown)


crates/payload/builder/src/lib.rs
```diff
@@ -379,6 +386,45 @@ where
...
+                to_execute.push(subblock.clone());
+            }
+
+            drop(subblocks);
+
+            for subblock in to_execute {
+                for tx in subblock.transactions_recovered() {
+                    builder
+                        .execute_transaction(tx.cloned())
+                        .map_err(PayloadBuilderError::evm)?;
```
mattsse: this wouldn't skip txs with an incorrect nonce, right?

crates/payload/builder/src/lib.rs
```diff
@@ -379,6 +386,45 @@ where
...
         }
 
+        let mut selected_subblocks = HashMap::new();
+
+        // Subblocks will most likely arrive after we've started building the payload, 
+        // so we need to wait for a bit to ensure we collect as many as possible.
+        loop {
+            let Ok(subblocks) = attributes.subblocks().lock() else {
+                break;
+            };
```
mattsse: I see, we can't take the subblocks out here, in case we're doing another iteration

crates/payload/builder/src/lib.rs
```diff
@@ -379,6 +386,45 @@ where
...
+            for subblock in subblocks.iter() {
+                if selected_subblocks.contains_key(&subblock.validator()) {
+                    continue;
+                }
+                to_execute.push(subblock.clone());
+            }
+
+            drop(subblocks);
+
+            for subblock in to_execute {
```
mattsse: I think we want some traces for this part here

crates/payload/builder/src/lib.rs
```diff
@@ -379,6 +386,45 @@ where
             cumulative_gas_used += gas_used;
         }
 
+        let mut selected_subblocks = HashMap::new();
+
+        // Subblocks will most likely arrive after we've started building the payload, 
+        // so we need to wait for a bit to ensure we collect as many as possible.
+        loop {
+            let Ok(subblocks) = attributes.subblocks().lock() else {
```
mattsse: at this point we're basically racing against the interrupt signal and sleeping 100ms, which means we might miss out on subblocks if this happes


```
0: sleep 100ms
10: new subblock arrives
90: interrupt trigger
100: new attempt
110: subblock finishes + interrupted
```


for this use case it might be better to use something like a cross beam select over an interrupt channel and incoming subblocks channel

the builder can then store already received subblocks in its own Arc Mutex store for the next attempt

crates/payload/builder/src/lib.rs
klkvr: yeah it wouldn't but we want to include those

Cargo.toml
```diff
@@ -204,6 +204,7 @@ camino = "1.1.12"
 clap = { version = "4.5.45", features = ["derive"] }
 const-hex = { version = "1.15.0" }
 derive_more = { version = "2.0.0" }
+ed25519-consensus = "2"
```
SuperFluffy: You are probably constructing pub keys from bytes? I ran into the same issue and decided to not import ed25519-consensus directly and instead rely on the fact that commonware's `Read` and `Write` impls use the underlying 32 bytes.

That felt like the better choice in that moment instead of relying on dependencies of dependencies. But it'd be better if they exposed contructors.

crates/commonware-node/src/config.rs
```diff
@@ -21,6 +21,7 @@ pub(crate) const RESOLVER_CHANNEL_IDENT: commonware_p2p::Channel = 2;
 pub(crate) const BROADCASTER_CHANNEL_IDENT: commonware_p2p::Channel = 3;
 pub(crate) const BACKFILL_BY_DIGEST_CHANNEL_IDENT: commonware_p2p::Channel = 4;
 pub(crate) const DKG_CHANNEL_IDENT: commonware_p2p::Channel = 5;
+pub(crate) const SUBBLOCKS_CHANNEL_IDENT: commonware_p2p::Channel = 6;
```
SuperFluffy: Headsup - this probably needs to become `7` because the boundary-cert channel uses 6.

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,403 @@
...
+impl<Ctx: Spawner> SubBlocksService<Ctx> {
+    pub async fn run(
+        mut self,
+        (mut network_tx, mut network_rx): (
+            impl Sender<PublicKey = PublicKey>,
+            impl Receiver<PublicKey = PublicKey>,
+        ),
+    ) {
+        loop {
+            tokio::select! {
```
SuperFluffy: That's a source of non-determinism - can we make use of commonware's `commonware_macros::select`?

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,403 @@
...
+                // First, handle messages from consensus engine and handle.
+                Some(action) = self.actions_rx.recv() => {
+                    self.on_new_message(action);
+                }
+                // Handle messages from the network.
+                Ok((sender, message)) = network_rx.recv() => {
+                    let Ok(subblock) = SignedSubBlock::decode(&mut &*message) else {
```
SuperFluffy: Can you stuff all this logic into a handler an instrument it? Reason is that all of your events are outside of a root span. That works with tracing-subscriber's fmt, but won't work with OTEL, where every event needs a span.

crates/commonware-node/src/epoch/manager/actor.rs
```diff
@@ -35,12 +38,19 @@ use super::ingress::Message;
...
+    pub scheme: Scheme<MinSig>,
+}
+
 pub(crate) struct Actor<TBlocker, TContext> {
     active_epochs: BTreeMap<Epoch, Handle<()>>,
     config: super::Config<TBlocker>,
     context: ContextCell<TContext>,
     mailbox: mpsc::UnboundedReceiver<Message>,
     metrics: Metrics,
+    context_tx: watch::Sender<Option<EpochContext>>,
```
SuperFluffy: To be consistent with the rest of the crate (and fit into the commonware model), can we make this take a `T: Reporter<Activity = EpochContext>` instead?

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,403 @@
...
+
+    /// Sends a new transaction to the subblocks service.
+    AddTransaction(Box<TempoTxEnvelope>),
+
+    /// Reports a new consensus event.
+    Consensus(Box<Activity<Scheme<MinSig>, Digest>>),
+}
+
+/// Task managing collected subblocks.
+pub struct SubBlocksService<Ctx> {
```
SuperFluffy: Follow the naming convention of the rest of the crate?

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,403 @@
...
+                    let _ = network_tx.send(Recipients::One(next_proposer.clone()), alloy_rlp::encode(&signed_subblock).into(), true).await;
+                }
+            }
+        }
+    }
+}
+
+/// Handle to the spawned subblocks service.
+#[derive(Clone)]
+pub struct SubBlocksHandle {
```
SuperFluffy:
```suggestion
pub struct Mailbox {
```

SuperFluffy: On it being `pub`:

I understand the only purpose of making this `pub` is so that it can be used in `e2e` as an escape hatch?

If that's the case, I think I'd rather 1) lock this down/make it `pub(crate)`, and b) move some of the e2e tests into here instead. Because we'd have circular dependencies we can create a `e2e-utils` crate or something.

My reasoning is that e2e tests should be as close to running a binary as possible, and the tests defined using externally visible effects/APIs (like metrics; I'll give the reth handle a pass because we could do those checks using an RPC provider).

crates/commonware-node/src/epoch/manager/actor.rs
SuperFluffy: Actually come to think of it, it'd be best if this was exposed through the epoch manager's `Mailbox` so that there is a single ingress point. Fundamentally this is the same as requesting the latest epoch state from the actor.

So you'd do `epoch::manager::Mailbox::get_latest_epoch_context`, which would return the epoch context.

klkvr: I've replaced this with `SchemeProvider` usage

crates/commonware-node/src/subblocks.rs
klkvr: used `tokio::select! { biased; }` instead which I think should be deterministic?

crates/commonware-node/src/subblocks.rs
klkvr: I think it should be fine to keep as is for now, eventually we will have RPC APIs for subblock transactions submission  https://tempoxyz.slack.com/archives/C09KCGR4LQ4/p1761921622448169 which would remove the need for using the raw handle in e2e tests

Though we'd still need some kind of handle on Reth RPC side

crates/commonware-node/src/subblocks.rs
klkvr: 59ca68f

crates/e2e/src/execution_runtime.rs
```diff
@@ -213,6 +215,10 @@ pub async fn launch_execution_node<P: AsRef<Path>>(
         .with_datadir_args(DatadirArgs {
             datadir: datadir.as_ref().to_path_buf().into(),
             ..DatadirArgs::default()
+        })
+        .with_payload_builder(PayloadBuilderArgs {
+            interval: Duration::from_millis(100),
+            ..Default::default()
```
klkvr, line 221: default interval is `1s` which causes us to pretty much never do multiple intervals for payload building

@mattsse I'm wondering if it's worth exposing a way to override default CLI values for reth args?

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,534 @@
...
+                Round::new(round.epoch(), round.view() + 1),
+                scheme.seed(round, &certificate),
+            )
+        };
+
+        let Some(scheme) = self.scheme_provider.scheme(next_block_round.epoch()) else {
+            return;
+        };
+
+        let (next_proposer, _) = select_leader::<Scheme<PublicKey, MinSig>, _>(
```
klkvr: @SuperFluffy would appreciate a closer look here

the idea for that logic is to always know the next block proposer as quickly as possible

crates/commonware-node-config/src/timeouts.rs
```diff
@@ -31,6 +31,14 @@ pub struct Config {
     /// The amount of time to wait for payload builder before resolving payload.
     #[serde(with = "humantime_serde")]
     pub new_payload_wait_time: Duration,
+
+    /// Timeout for subblock building.
+    #[serde(with = "humantime_serde", default = "default_time_to_build_subblock")]
```
mattsse, line 36: should this type just use serde(default) ?

crates/commonware-node/src/consensus/engine.rs
```diff
@@ -281,6 +299,9 @@ where
     marshal: crate::alias::marshal::Actor<TContext>,
 
     epoch_manager: epoch::manager::Actor<TBlocker, TContext>,
+
+    subblocks: subblocks::Actor<TContext>,
+    subblocks_handle: subblocks::Mailbox,
```
mattsse: can we document what these types are, a bit confusing that we have both, looks like subblocks_handle is an actual handle.
side note: naming this mailbox kinda always confuses me because this is like a CW thing when this type here is not

EDIT: I see that actor is just a service

ngl, I'll likely never getting used to this

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -529,20 +555,48 @@ impl Inner<Init> {
...
+        // Spawn a task to update the subblocks set for the payload builder.
+        //
+        // It is likely that the subblocks will be arriving as we are building the block,
+        // thus we rely on multiple iterations of payload building to happen, and build
+        // blocks with as many subblocks as possible.
+        {
+            let interrupt_handle = interrupt_handle.clone();
+            context.clone().spawn(move |context| async move {
+                loop {
+                    context.sleep(Duration::from_millis(100)).await;
```
mattsse: this should be wired to the config arg instead of hard codec value?

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,534 @@
...
+use tempo_node::{TempoFullNode, consensus::TEMPO_SHARED_GAS_DIVISOR, evm::evm::TempoEvm};
+use tempo_primitives::{
+    RecoveredSubBlock, SignedSubBlock, SubBlock, SubBlockVersion, TempoTxEnvelope,
+};
+use tokio::sync::{mpsc, oneshot};
+use tracing::{debug, instrument, warn};
+
+/// Actions processed by the subblocks service.
+#[derive(Debug)]
+enum Message {
```
mattsse: this type def is out of order, should be moved below main type

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,534 @@
...
+    node: TempoFullNode,
+    fee_recipient: Address,
+    time_to_build_subblock: Duration,
+    epoch_length: u64,
+
+    next_proposer: Option<PublicKey>,
+    next_parent_hash: Option<BlockHash>,
+    next_block_epoch: Option<Epoch>,
+    subblocks: HashMap<B256, RecoveredSubBlock>,
+    subblock_transactions: Arc<Mutex<HashMap<TxHash, Arc<TempoTxEnvelope>>>>,
```
mattsse: can we add some docs here?

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,534 @@
+use std::{
+    sync::{Arc, Mutex},
```
mattsse: lets use parking lot mutex, this avoids lock poising error handling

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,534 @@
...
+    /// of the service or if no subblocks have been collected yet.
+    GetSubBlocks {
+        /// Parent block to return subblocks for.
+        parent: BlockHash,
+        /// Response channel.
+        response: oneshot::Sender<Vec<RecoveredSubBlock>>,
+    },
+
+    /// Sends a new transaction to the subblocks service.
+    AddTransaction(Box<TempoTxEnvelope>),
```
mattsse: should this instead only receive valid recovered transactions?

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,534 @@
...
+        let node = self.node.clone();
+        let Some(parent_hash) = self.next_parent_hash else {
+            return;
+        };
+        let num_validators = scheme.participants().len();
+        let signer = self.signer.clone();
+        let fee_recipient = self.fee_recipient;
+        let timeout = self.time_to_build_subblock;
+        let handle = self.context.clone().shared(true).spawn(move |_| {
+            build_subblock(
```
mattsse: how often is this expected to be invoked?

we could definitely reuse 

https://github.com/tempoxyz/tempo/blob/623ebd3347992bd3f2e3e462162591f136b4e037/crates/commonware-node/src/subblocks.rs#L420-L422

we could even keep this updated in the background by fetching from the cache or from the new tips

crates/revm/src/handler.rs
```diff
@@ -713,6 +715,10 @@ where
             validate_time_window(aa_env.valid_after, aa_env.valid_before, block_timestamp)?;
         }
 
+        if tx.subblock_transaction && tx.max_fee_per_gas() > 0 {
+            return Err(TempoInvalidTransaction::SubblockTransactionMustHaveZeroFee.into());
```
mattsse: can we move this up, we always need to check this and this is pretty cheap

crates/revm/src/tx.rs
```diff
@@ -59,14 +59,16 @@ pub struct TempoTxEnv {
 
     /// AA-specific transaction environment (boxed to keep TempoTxEnv lean for non-AA tx)
     pub aa_tx_env: Option<Box<AATxEnv>>,
+
+    /// Whether the transaction is a subblock transaction.
+    pub subblock_transaction: bool,
```
mattsse: only AAtxs can be subblock transactions, right?

then this should be moved into aa_tx_env I think

crates/primitives/src/subblock.rs
```diff
@@ -0,0 +1,259 @@
...
+        if buf.len() + header.payload_length != remaining {
+            return Err(alloy_rlp::Error::UnexpectedLength);
+        }
+
+        Ok(this)
+    }
+}
+
+/// A subblock with recovered senders.
+#[derive(Debug, Clone, RlpEncodable, RlpDecodable, derive_more::Deref, derive_more::DerefMut)]
```
mattsse: do we need rlp here?

imo we should avoid encoding the recovered signers

crates/primitives/src/subblock.rs
```diff
@@ -0,0 +1,259 @@
...
+    /// Recipient of the fees for the subblock.
+    pub fee_recipient: Address,
+    /// Transactions included in the subblock.
+    pub transactions: Vec<TempoTxEnvelope>,
+}
+
+impl SubBlock {
+    /// Returns the hash for the signature.
+    pub fn signature_hash(&self) -> B256 {
+        let mut buf = Vec::new();
```
mattsse: we can pre-alloc a few bytes here

crates/primitives/src/subblock.rs
```diff
@@ -0,0 +1,259 @@
...
+impl Encodable for SubBlock {
+    fn encode(&self, out: &mut dyn BufMut) {
+        self.rlp_header().encode(out);
+        self.rlp_encode_fields(out);
+    }
+}
+
+/// A subblock with a signature.
+#[derive(Debug, Clone, derive_more::Deref, derive_more::DerefMut)]
+pub struct SignedSubBlock {
```
mattsse, line 131: lets add arbitrary support and add a few proptest roundtrip tests?

crates/primitives/src/header.rs
```diff
@@ -15,9 +15,16 @@ use reth_primitives_traits::InMemorySize;
...
+    )]
     pub general_gas_limit: u64,
 
+    /// Shared gas limit allocated for the subblocks section of the block.
+    #[cfg_attr(feature = "serde", serde(with = "alloy_serde::quantity"))]
+    pub shared_gas_limit: u64,
+
     /// Sub-second (milliseconds) portion of the timestamp.
     #[cfg_attr(feature = "serde", serde(with = "alloy_serde::quantity"))]
     pub timestamp_millis_part: u64,
```
mattsse, line 30: unrelated, should we move them below the flattened header, so that these appear latest in json

I guess not because this would break compact derive perhaps @joshieDo ?

crates/payload/types/src/attrs.rs
```diff
@@ -5,8 +5,9 @@ use reth_ethereum_engine_primitives::EthPayloadBuilderAttributes;
 use reth_node_api::PayloadBuilderAttributes;
 use std::{
     convert::Infallible,
-    sync::{Arc, atomic, atomic::Ordering},
+    sync::{Arc, Mutex, atomic, atomic::Ordering},
```
mattsse: same here, parkinglot

crates/payload/builder/src/lib.rs
```diff
@@ -478,3 +532,24 @@ where
...
+        .transactions
+        .iter()
+        .rev()
+        .find_map(|tx| Vec::<SubBlockMetadata>::decode(&mut tx.input().as_ref()).ok())
+    else {
+        return false;
+    };
+
+    subblocks.len() > best_metadata.len()
+}
```
mattsse, line 619: I think we can avoid this and instead track SubblockInfo in our TempoBuiltPayload?

crates/payload/builder/src/lib.rs
```diff
@@ -85,7 +85,11 @@ impl<Provider: ChainSpecProvider> TempoPayloadBuilder<Provider> {
     /// Returns a vector of system transactions that must be executed at the end of each block:
     /// 1. Fee manager executeBlock - processes collected fees
     /// 2. Stablecoin exchange executeBlock - commits pending orders
```
mattsse, line 104: need to update docs here, because we now add additional txs

crates/e2e/src/execution_runtime.rs
mattsse: tricky, the way to override it would be via the cli itself

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,534 @@
...
+    pub fn get_subblocks(&self, parent: BlockHash) -> oneshot::Receiver<Vec<RecoveredSubBlock>> {
+        let (tx, rx) = oneshot::channel();
+        let _ = self.tx.send(Message::GetSubBlocks {
+            parent,
+            response: tx,
+        });
+        rx
+    }
+
+    pub fn add_transaction(&self, tx: TempoTxEnvelope) {
```
mattsse: rn we dont have a way to submit transactions, right? still tbd?

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,534 @@
...
+
+    /// Sends a new transaction to the subblocks service.
+    AddTransaction(Box<TempoTxEnvelope>),
+
+    /// Reports a new consensus event.
+    Consensus(Box<Activity<Scheme<PublicKey, MinSig>, Digest>>),
+}
+
+/// Task managing collected subblocks.
+pub struct Actor<Ctx> {
```
mattsse: can we document here what this actually does?

not immediately obvious that this type actually builds them based on consensus events

looks like this collects subblocks from the network and is responsible for building a local subblock if any

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,534 @@
...
+                transactions.insert(*transaction.tx_hash(), Arc::new(*transaction));
+            }
+            Message::Consensus(activity) => {
+                self.on_consensus_event(*activity);
+            }
+        }
+    }
+
+    #[instrument(skip_all, fields(event.epoch = event.epoch(), event.view = event.view()))]
+    fn on_consensus_event(&mut self, event: Activity<Scheme<PublicKey, MinSig>, Digest>) {
```
mattsse, line 204: when is this triggered, and how often per block?

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,534 @@
...
+                node,
+                parent_hash,
+                num_validators,
+                signer,
+                fee_recipient,
+                timeout,
+            )
+        });
+
+        self.subblock_builder_handle = Some(handle);
```
mattsse: I believe this is fine because we only ever build one?

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,534 @@
...
+    Consensus(Box<Activity<Scheme<PublicKey, MinSig>, Digest>>),
+}
+
+/// Task managing collected subblocks.
+pub struct Actor<Ctx> {
+    actions_rx: mpsc::UnboundedReceiver<Message>,
+    scheme_provider: SchemeProvider,
+    subblock_builder_handle: Option<Handle<RecoveredSubBlock>>,
+    validated_subblocks_tx: mpsc::UnboundedSender<RecoveredSubBlock>,
+    validated_subblocks_rx: mpsc::UnboundedReceiver<RecoveredSubBlock>,
```
mattsse: hmm, if I understand this right, then these are just used to communicate back to this service here?

then this could also just be another message variant? and this service could hold on to its handle?

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,534 @@
...
+        .build();
+    let block = node
+        .provider
+        .sealed_block_with_senders(hash.into(), Default::default())?
+        .ok_or(ProviderError::BestBlockNotFound)?;
+
+    Ok(node.evm_config.evm_for_block(db, block.sealed_block())?)
+}
+
+async fn build_subblock(
```
mattsse, line 496: can we document why we're doing this?

this is basically pre-filtering, right?

then it might also make sense to flag any transactions that are invalid,
might be relevent depending on how we decide to inject them in here

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,534 @@
...
+    };
+
+    RecoveredSubBlock::new_unchecked(
+        signed_subblock,
+        senders,
+        B256::from_slice(&signer.public_key()),
+    )
+}
+
+async fn validate_subblock(
```
mattsse, line 570: same docs here, basically what does it mean to be a valid subblock,

e.g. when building we skip invalid txs (for our own subblock) but if this happens with a subblock received from another validator the entire subblock is invalid

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,534 @@
...
+}
+
+/// Handle to the spawned subblocks service.
+#[derive(Clone)]
+pub struct Mailbox {
+    tx: mpsc::UnboundedSender<Message>,
+}
+
+impl Mailbox {
+    pub fn get_subblocks(&self, parent: BlockHash) -> oneshot::Receiver<Vec<RecoveredSubBlock>> {
```
mattsse: would it make sense to introduce an enum  here that represents whether this is a local subblock or remote

for local we might be able to adjust txs on the fly, but perhaps this isnt really useful

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,534 @@
...
+
+            let txs = transactions.lock().unwrap().clone();
+            for (tx_hash, tx) in txs {
+                if tx.gas_limit() > gas_left {
+                    continue;
+                }
+                let Ok(sender) = tx.try_recover() else {
+                    continue;
+                };
+                if evm.transact_commit(tx.with_signer_ref(sender)).is_err() {
```
mattsse: I think we need finer error handling here because AA txs might be delayed via the valid_after setting, so we should put these back or have some form of maintenance for them

crates/commonware-node-config/src/timeouts.rs
klkvr: yep this would work too, wasn't sure if we're OK with that

crates/commonware-node/src/consensus/application/actor.rs
klkvr: I think ideally we could derive it somehow from `new_payload_wait_time` and payload builder job interval

configuring this specific sleep seems like an overkill given that this just clones some stuff to mutex once in a while

maybe we should just give the entire subblocks handle to the payload builder after all

crates/commonware-node/src/subblocks.rs
klkvr: `Notarization`/`Nullification` should only happen once per view/round. i.e once per block or a missed slot

crates/commonware-node/src/subblocks.rs
klkvr: yep we're only buillding one at a time on top of the latest tip

crates/commonware-node/src/subblocks.rs
klkvr: yep good point

crates/commonware-node/src/subblocks.rs
klkvr: this should only really fetch the current tip block which we already cache, so I think it should be fine?

also changed this to just use header instead of entire block

crates/primitives/src/subblock.rs
klkvr: yeah no need for rlp here, removed

crates/primitives/src/subblock.rs
klkvr: yep added

crates/commonware-node/src/consensus/application/actor.rs
klkvr: I've removed that logic now and changed payload builder to talk with subblocks service in a somewhat hacky but easier way than this loop:
https://github.com/tempoxyz/tempo/blob/70675da151d60031d8039edf1e863a5d7633c803/crates/payload/types/src/attrs.rs#L49
https://github.com/tempoxyz/tempo/blob/70675da151d60031d8039edf1e863a5d7633c803/crates/commonware-node/src/consensus/application/actor.rs#L558-L562

Mutex was problematic in cases when same validator is getting two separate proposal requests for the same epoch, and we end up re-using the same payload id

https://github.com/tempoxyz/tempo/blob/70675da151d60031d8039edf1e863a5d7633c803/crates/commonware-node/src/consensus/application/actor.rs#L547-L554

crates/commonware-node/src/subblocks.rs
klkvr: I think we can check this trivially by looking at `subblock.validator`

crates/payload/builder/src/lib.rs
klkvr: yep this makes sense, though PR is already pretty big so would prefer to leave this for a follow-up https://github.com/tempoxyz/tempo/issues/635

crates/commonware-node/src/subblocks.rs
klkvr: this is actually a bit hacky right now because we don't track included transactions, so we'd always attempt to include them again and filter out for the next subblock

we might want a proper pool for this eventually

this is also somewhat relevant to the general effort of smarter payload builder

crates/payload/builder/src/lib.rs
mattsse: makes sense

crates/commonware-node/src/subblocks.rs
klkvr: ah this is a good catch actually

I was assuming that it's fine to validate subblock transactions once on top of the parent because all of them are free (balance can't get invalidated) and use a separate nonce lane (nonce can't get invalidated)

I guess `valid_before`/`valid_after` are a special cases here that we need to handle properly as well

klkvr: handled this edge case here e98763f

crates/commonware-node-config/src/timeouts.rs
SuperFluffy: Just to be clear - if we used `serde(default)`, this would mean we don't leave any time to build a block? For `new_payload_wait_time` this lead to confusion, where it was not clear if `Duration:ZERO` meant uninterrupted block building or immediate interrupt (it's the latter).

Personally happy to use `Duration::from_millis`

crates/commonware-node/src/config.rs
```diff
@@ -42,5 +43,7 @@ pub(crate) const RECOVERED_LIMIT: Quota =
     Quota::per_second(NonZeroU32::new(128).expect("value is not zero"));
 pub(crate) const RESOLVER_LIMIT: Quota =
     Quota::per_second(NonZeroU32::new(128).expect("value is not zero"));
+pub(crate) const SUBBLOCKS_LIMIT: Quota =
```
SuperFluffy, line 46: I don't have a good intuition what a good quota here would be.

If I recall correctly, this quota should be per-peer. This means a peer can send up to 128 values. Is that reasonable? Should it be lower?

Should double check if it is indeed per-peer or if I am mistaken.

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -51,8 +52,9 @@ use super::{
 use crate::{
     consensus::{Digest, block::Block},
     dkg::PublicOutcome,
-    epoch,
-    marshal_utils::UpdateExt as _,
+    epoch::{self, SchemeProvider},
+    marshal_utils::UpdateExt,
```
SuperFluffy: Very nit (my personal thing about not cluttering the namespace):

```suggestion
    marshal_utils::UpdateExt as _,
```

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -371,7 +378,26 @@ impl Inner<Init> {
...
+                .clone(),
+            &proposal,
+            parent_digest.0,
+            &self.scheme_provider,
+        )
+        .await
+        .wrap_err("failed verifying block against execution layer")?;
+
+        if !is_good {
+            eyre::bail!("proposal failed validation");
```
SuperFluffy: Let's make it extra clear that this is weird:

```suggestion
        .wrap_err("failed verifying just-proposed block against execution layer")?;

        if !is_good {
            eyre::bail!("validation reported that that just-proposed block is invalid");
```

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -451,7 +477,7 @@ impl Inner<Init> {
         }
     }
 
-    async fn propose<TContext: Pacer>(
+    async fn propose<TContext: Pacer + Spawner>(
```
SuperFluffy: New bound necessary? Are you spawning something in the propose body now that you didn't before?

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -509,14 +535,19 @@ impl Inner<Init> {
...
+            },
         );
 
         let interrupt_handle = attrs.interrupt_handle().clone();
 
         let payload_id = self
             .execution_node
             .payload_builder_handle
-            .send_new_payload(attrs)
+            .send_new_payload(attrs.clone())
```
SuperFluffy: Cloning necessary?

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -736,27 +768,37 @@ async fn verify_block<TContext: Pacer>(
     epoch_length: u64,
     engine: ConsensusEngineHandle<TempoPayloadTypes>,
     block: &Block,
-    parent: &Block,
+    parent_hash: B256,
```
SuperFluffy: Let's call this `parent_digest: Digest` here - hashes and digests map 1-to-1 so you are going to get the same output, but we are dealing with commonware consensus and I am trying to be consistent when using digest vs hash (the other instrumentation bits use digest).

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -371,7 +378,26 @@ impl Inner<Init> {
 
         {
             let mut lock = self.state.latest_proposed_block.write().await;
```
SuperFluffy, line 345: Comment on this line and the `response.send(proposal_digest)` above, which is outside the diff:

If `verify_block` below fails for whatever reason, should we really propose the block and store it as `self.state.latest_proposed_block`?.

Or should we consider `propose + verify` to be one step?

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -371,7 +378,26 @@ impl Inner<Init> {
...
-            *lock = Some(proposal);
+            *lock = Some(proposal.clone());
+        }
+
+        let is_good = verify_block(
```
SuperFluffy, line 350: Add comment that the main purpose of this is to provide the payload to the EL?

Can this ever fail or return `false`?

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -509,14 +535,19 @@ impl Inner<Init> {
             parent.block_hash(),
             self.fee_recipient,
             context.current().epoch_millis(),
+            move || {
```
SuperFluffy, line 548: The more fields we add to this constructor, the more I wish we'd just make all of its fields `pub`.

crates/consensus/src/lib.rs
```diff
@@ -136,8 +137,12 @@ impl HeaderValidator<TempoHeader> for TempoConsensus {
     fn validate_header(&self, header: &SealedHeader<TempoHeader>) -> Result<(), ConsensusError> {
         self.eth_beacon_validate_header(header)?;
 
+        let shared_gas_limit = header.gas_limit() / TEMPO_SHARED_GAS_DIVISOR;
+
         // Validate the non-payment gas limit
-        if header.general_gas_limit != header.gas_limit() / TEMPO_GENERAL_GAS_DIVISOR {
+        if header.general_gas_limit
+            != (header.gas_limit() - shared_gas_limit) / TEMPO_GENERAL_GAS_DIVISOR
```
SuperFluffy, line 144: Should this be saturating or checked arithmetic?

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,597 @@
...
+            time_to_build_subblock,
+            epoch_length,
+            consensus_tip: None,
+            subblocks: Default::default(),
+            subblock_transactions: Default::default(),
+        }
+    }
+
+    /// Returns a handle to the subblocks service.
+    pub fn mailbox(&self) -> Mailbox {
```
SuperFluffy, line 131: This was a design error on my part for the application actor. By doing this, there is no trigger to shut down (or even error) if the channel is closed, because there we will always ourselves hold on to a sender.

 I have to fix this in the application actor, but let's do it properly here and either have `Actor::new -> (Self, Mailbox)` or - my personal preference - provide a `subscriber::init -> (Actor, Mailbox)`

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,597 @@
...
+    consensus_tip: Option<(Round, BlockHash, bls12381_threshold::Signature<MinSig>)>,
+
+    /// Collected subblocks keyed by validator public key.
+    subblocks: HashMap<B256, RecoveredSubBlock>,
+    /// Subblock candidate transactions.
+    subblock_transactions: Arc<Mutex<HashMap<TxHash, Arc<Recovered<TempoTxEnvelope>>>>>,
+}
+
+impl<Ctx: Spawner> Actor<Ctx> {
+    pub(crate) fn new(
```
SuperFluffy, line 101: Please create a `Config` struct and pass that in instead for consistency with the rest of the crate.

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,597 @@
...
+    /// Current consensus tip. Includes highest observed round, digest and certificate.
+    consensus_tip: Option<(Round, BlockHash, bls12381_threshold::Signature<MinSig>)>,
+
+    /// Collected subblocks keyed by validator public key.
+    subblocks: HashMap<B256, RecoveredSubBlock>,
+    /// Subblock candidate transactions.
+    subblock_transactions: Arc<Mutex<HashMap<TxHash, Arc<Recovered<TempoTxEnvelope>>>>>,
+}
+
+impl<Ctx: Spawner> Actor<Ctx> {
```
SuperFluffy: Nit, but also for consistency: please call this generic `TContext`.

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,597 @@
...
+use reth_revm::database::StateProviderDatabase;
+use std::{
+    sync::{Arc, mpsc::RecvError},
+    time::{Duration, Instant},
+};
+use tempo_node::{TempoFullNode, consensus::TEMPO_SHARED_GAS_DIVISOR, evm::evm::TempoEvm};
+use tempo_primitives::{
+    RecoveredSubBlock, SignedSubBlock, SubBlock, SubBlockVersion, TempoTxEnvelope,
+};
+use tokio::sync::mpsc;
```
SuperFluffy: Let's use `futures::mpsc` everywhere for consistency.

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,597 @@
...
+
+    pub async fn run(
+        mut self,
+        (mut network_tx, mut network_rx): (
+            impl Sender<PublicKey = PublicKey>,
+            impl Receiver<PublicKey = PublicKey>,
+        ),
+    ) {
+        loop {
+            tokio::select! {
```
SuperFluffy: As noted above, let's use either `futures` or actually `commonware_macros::select!` for consistency.

`tokio::select! { biased ... }` should do the trick, but IMO it's better to not mix and match (I started with tokio but got rid of it as much as possible).

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,597 @@
...
+                evm.block().gas_limit / TEMPO_SHARED_GAS_DIVISOR / num_validators as u64;
+
+            let txs = transactions.lock().clone();
+            for (tx_hash, tx) in txs {
+                if tx.gas_limit() > gas_left {
+                    continue;
+                }
+                if evm.transact_commit(&*tx).is_err() {
+                    // Remove invalid transactions from the set.
+                    transactions.lock().remove(&tx_hash);
```
mattsse: this makes sense here imo

alternative would be to include this info in the response, but we can simply filter them out like this

although we lose info here, could b relevant for things like monitoring, but we can figure this out later

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,597 @@
...
+                    && *round < n.proposal.round
+                {
+                    if *tip != n.proposal.payload.0 {
+                        // Clear collected subblocks if we have a new tip.
+                        self.subblocks.clear();
+                    }
+                    *round = n.proposal.round;
+                    *tip = n.proposal.payload.0;
+                    *cert = n.certificate;
+                } else if self.consensus_tip.is_none() {
```
SuperFluffy: Nit: technically not necessary: you are already in the `else` branch of the `if let Some ... = consensus_tip` condition.

crates/payload/builder/src/lib.rs
```diff
@@ -403,16 +442,27 @@ where
...
+        // `valid_before` field.
+        subblocks.retain(|subblock| {
+            !subblock.transactions.iter().any(|tx| {
+                tx.as_aa().is_some_and(|tx| {
+                    tx.tx()
+                        .valid_before
+                        .is_some_and(|valid| valid < attributes.timestamp())
+                })
+            })
+        });
```
mattsse, line 498: this makes sense here, we should check this early

we could perhaps move this into attributes because this knows the timestamps, but can also keep it here for vis

crates/payload/builder/src/lib.rs
```diff
@@ -422,8 +472,24 @@ where
             });
         }
 
+        for subblock in &subblocks {
+            for tx in subblock.transactions_recovered() {
+                builder
+                    .execute_transaction(tx.cloned())
+                    .map_err(PayloadBuilderError::evm)?;
```
mattsse: we need to exit the entire builder fn, because we might have already committed a partial subblock?

this is potentially problematic, because I dont think we're excluding this subblock in a followup attempt?

so if a subblock fails, this would prevent us from building a block indefinitely ?

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,597 @@
...
+                    .lock()
+                    .insert(*transaction.tx_hash(), Arc::new(*transaction));
+            }
+            Message::Consensus(activity) => self.on_consensus_event(*activity),
+            Message::ValidatedSubblock(subblock) => self.on_validated_subblock(subblock),
+        }
+    }
+
+    /// Tracking of the current sconsensus state by listening to notarizations and nullifications.
+    #[instrument(skip_all, fields(event.epoch = event.epoch(), event.view = event.view()))]
```
SuperFluffy, line 203: You could add `err(level = Level::DEBUG)` and instead of returning early on the various conditions, you can do `eyre::ensure!(condition, "<msg>")` to report why something was skipped.

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,597 @@
...
+
+        let Some((round, tip, certificate)) = &self.consensus_tip else {
+            return;
+        };
+
+        let Ok(Some(header)) = self.node.provider.header(*tip) else {
+            return;
+        };
+
+        let next_epoch = if epoch::is_last_height(header.number(), self.epoch_length) {
```
SuperFluffy: This is the epoch of the next block? Should we call it e.g. `epoch_of_next_block`? Was confused by this, because `next_epoch = this_epoch + 1`.

Does the line below express the same thing?

```rust
let epoch_of_next_block = epoch::of_height(
    header.number().saturating_add(1),
    self.epoch_length,
)
.expect("non-zero heights are guaranteed to have an epoch");
```

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,597 @@
...
+            epoch::of_height(header.number(), self.epoch_length).unwrap() + 1
+        } else {
+            round.epoch()
+        };
+
+        // Can't proceed without knowing a validator set for the current epoch.
+        let Some(scheme) = self.scheme_provider.scheme(next_epoch) else {
+            return;
```
SuperFluffy, line 251: In reference to my comment on `instrument(err)`:

This could just be `return Err(eyre!("scheme provider returned no validator set for epoch of next block `{next_epoch}`; can't continue without it"));` and you get documentation && early return at the same time.

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,597 @@
...
+        };
+
+        let next_round = if round.epoch() == next_epoch {
+            Round::new(round.epoch(), round.view() + 1)
```
SuperFluffy, line 255: Please saturating/checked arithmetic.

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,597 @@
...
+        } else {
+            Round::new(next_epoch, 1)
```
SuperFluffy: Commonware confirmed it always starts at 1?

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,597 @@
...
+        };
+
+        let seed = if next_round.view() == 1 {
+            // First view does not have a seed.
```
SuperFluffy, line 261: Even new epochs don't have a seed?

EDIT: checked - that seems indeed the case. Let's link the relevant line?
https://github.com/commonwarexyz/monorepo/blob/4c6db774ace5da5ec3cabd59069ecbac1dc4a13c/consensus/src/simplex/actors/voter/actor.rs#L1581

I wonder if commonware might change this in the future.

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,597 @@
...
+            Activity::Nullification(n) => {
+                // On a newer nullification, update round and certificate.
+                if let Some((round, _, cert)) = &mut self.consensus_tip
+                    && *round < n.round
+                {
+                    *round = n.round;
+                    *cert = n.certificate;
+                }
+            }
+            _ => return,
```
SuperFluffy, line 213: Do we need to also consider Finalizations? They probably happen less frequently and are usually lagging nullications and notarizations, but commonware are also considering them:

https://github.com/commonwarexyz/monorepo/blob/4c6db774ace5da5ec3cabd59069ecbac1dc4a13c/consensus/src/simplex/actors/voter/actor.rs#L1231

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,597 @@
...
+        };
+
+        let seed = if next_round.view() == 1 {
+            // First view does not have a seed.
+            None
+        } else {
+            scheme.seed(*round, certificate)
+        };
+
+        let (next_proposer, _) = select_leader::<Scheme<PublicKey, MinSig>, _>(
```
SuperFluffy, line 267: This seems fragile - the moment this logic is changed upstream I worry that this line fails.

I don't have a better idea. Let's emit a debug log for `next_proposer`?

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,597 @@
...
+    fn on_consensus_event(&mut self, event: Activity<Scheme<PublicKey, MinSig>, Digest>) {
+        match event {
+            Activity::Notarization(n) => {
+                // If we have a newer notarization, record new tip and certificate.
+                if let Some((round, tip, cert)) = &mut self.consensus_tip
+                    && *round < n.proposal.round
+                {
+                    if *tip != n.proposal.payload.0 {
+                        // Clear collected subblocks if we have a new tip.
+                        self.subblocks.clear();
```
SuperFluffy: A metric or event informing us that subblocks were cleared without being sent to a proposer would be good (i.e. block building took too long).

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,597 @@
...
+        });
+
+        self.subblock_builder_handle = Some(BuildSubblockTask {
+            handle,
+            parent_hash,
+            proposer: next_proposer,
+        });
+    }
+
+    #[instrument(skip_all, fields(sender = %sender))]
```
SuperFluffy: Add `err(level = Level::WARN)` (or whichever you feel appropriate), and emit errors instead of early returns. Then we have some indication if somebody is sending us garbage.

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,597 @@
...
+}
+
+/// Validates a subblock and reports it to the subblocks service.
+///
+/// Validation checks include:
+/// 1. Signature verification
+/// 2. Ensuring that sender is a validator for the block's epoch
+/// 3. Ensuring that all transactions have corresponding nonce key set.
+/// 4. Ensuring that all transactions are valid.
+async fn validate_subblock(
```
SuperFluffy, line 570: Instrument this and also add `err(level = Level::WARN)` (unless `Level::Error` is more appropraite) - then you get all of these errors in the correct context for free!

Maybe add a hash for the subblock and the sender to identify the task?

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,597 @@
...
+        .header(hash)?
+        .ok_or(ProviderError::BestBlockNotFound)?;
+
+    Ok(node.evm_config.evm_for_block(db, &header)?)
+}
+
+/// Builds a subblock from candidate transactions we've collected so far.
+///
+/// This will include as many valid transactions as possible within the given timeout.
+async fn build_subblock(
```
SuperFluffy, line 496: Instrument this future to guarantee that the events inside it have a span.

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,597 @@
...
+            for (tx_hash, tx) in txs {
+                if tx.gas_limit() > gas_left {
+                    continue;
+                }
+                if evm.transact_commit(&*tx).is_err() {
+                    // Remove invalid transactions from the set.
+                    transactions.lock().remove(&tx_hash);
+                    continue;
+                }
+                gas_left -= tx.gas_limit();
```
SuperFluffy, line 524: saturating sub?

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,597 @@
...
+    timeout: Duration,
+) -> RecoveredSubBlock {
+    let start = Instant::now();
+
+    let (transactions, senders) = match evm_at_block(&node, parent_hash) {
+        Ok(mut evm) => {
+            let mut selected_transactions = Vec::new();
+            let mut senders = Vec::new();
+            let mut gas_left =
+                evm.block().gas_limit / TEMPO_SHARED_GAS_DIVISOR / num_validators as u64;
```
SuperFluffy, line 512: checked division?

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,597 @@
...
+    /// Length of an epoch in blocks.
+    epoch_length: u64,
+
+    /// Current consensus tip. Includes highest observed round, digest and certificate.
+    consensus_tip: Option<(Round, BlockHash, bls12381_threshold::Signature<MinSig>)>,
+
+    /// Collected subblocks keyed by validator public key.
+    subblocks: HashMap<B256, RecoveredSubBlock>,
+    /// Subblock candidate transactions.
+    subblock_transactions: Arc<Mutex<HashMap<TxHash, Arc<Recovered<TempoTxEnvelope>>>>>,
```
SuperFluffy: Let's remove this source of non-determinism and make this an `IndexMap`? Then when iterating over this at least we will iterate from oldest to newest. I guess BTreeMap works as well, but is probably too expensive.

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,597 @@
...
+    /// Timeout for building a subblock.
+    time_to_build_subblock: Duration,
+    /// Length of an epoch in blocks.
+    epoch_length: u64,
+
+    /// Current consensus tip. Includes highest observed round, digest and certificate.
+    consensus_tip: Option<(Round, BlockHash, bls12381_threshold::Signature<MinSig>)>,
+
+    /// Collected subblocks keyed by validator public key.
+    subblocks: HashMap<B256, RecoveredSubBlock>,
```
SuperFluffy: Same as below - can we make this an indexmap or btreemap to remove a source of nondeterminism?

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,597 @@
...
+///
+/// This actor is responsible for tracking consensus events and determining
+/// current tip of the chain and next block's proposer.
+///
+/// Once next block proposer is known, we immediately start building a new subblock.
+/// Once it's built, we broadcast it to the next proposer directly.
+///
+/// Upon receiving a subblock from the network, we ensure that we are
+/// the proposer and verify the block on top of latest state.
+pub struct Actor<Ctx> {
```
SuperFluffy: For consistency, can you call this `TContext`?

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,597 @@
...
+        scheme: Arc<Scheme<PublicKey, MinSig>>,
+    ) {
+        let transactions = self.subblock_transactions.clone();
+        let node = self.node.clone();
+        let num_validators = scheme.participants().len();
+        let signer = self.signer.clone();
+        let fee_recipient = self.fee_recipient;
+        let timeout = self.time_to_build_subblock;
+        let handle = self.context.clone().shared(true).spawn(move |_| {
+            build_subblock(
```
SuperFluffy: Please attach a span to this spawned future by passing a `Span::current` and using `Instrument::instrument`

crates/commonware-node-config/src/timeouts.rs
klkvr: > Just to be clear - if we used serde(default), this would mean we don't leave any time to build a block?

I think Matt meant adding `serde(default)` attr to entire `Config` type that would mean that all fields will be filled from `Config::default()` if missing, and we'll be able to avoid `default_time_to_build_subblock` hack

crates/commonware-node/src/subblocks.rs
klkvr: > Actor::new -> (Self, Mailbox)

this was the interface before, was changed to address https://github.com/tempoxyz/tempo/pull/581#discussion_r2482834067

crates/commonware-node/src/subblocks.rs
klkvr: it is necessary because there's `&& *round < n.proposal.round` part of the condition

crates/commonware-node/src/subblocks.rs
klkvr: are we expecting view `u64::MAX` to be reachable? commonware does not use checked arithmetic for that

crates/commonware-node/src/subblocks.rs
klkvr: https://github.com/commonwarexyz/monorepo/blob/f6fc8ec7eac95bc6508a64caaca0a2a042826816/consensus/src/simplex/actors/voter/actor.rs#L1579-L1582

crates/commonware-node/src/subblocks.rs
SuperFluffy: Overlooked, good point.

crates/commonware-node/src/subblocks.rs
SuperFluffy: If we did I'd consider that a bug. But doing `round.view().checked_add(1).unwrap())` or w/e guards against that always.

crates/commonware-node/src/subblocks.rs
klkvr: we check this above

https://github.com/tempoxyz/tempo/blob/b98eadf1eef79682a08d6006e68fa29bf5ac977c/crates/commonware-node/src/subblocks.rs#L495-L497

crates/commonware-node/src/subblocks.rs
SuperFluffy: Ah, you are indeed holding and using the channel to yourself. That makes sense then and I didn't notice that.

crates/commonware-node/src/consensus/application/actor.rs
klkvr: if `verify_block` fails it means something went very wrong and we have a bug in payload builder

in happy case it should never happen so imo it's fine to optimize for it and not spend time on extra verification when proposing

crates/consensus/src/lib.rs
klkvr: it should be fine, `shared_gas_limit < header.gas_limit()` always holds

crates/commonware-node/src/subblocks.rs
klkvr: would you be open to opting out of this generics naming conversion? it's not consistent with neither reth nor commonware

SuperFluffy: In general yes if I get enough pushback. :-) But if we do that, we should do it in one go.

crates/commonware-node/src/subblocks.rs
klkvr: I think my preference would be to keep this as is, this can be considered unreachable, and the current handling is also consistent with commonware

crates/commonware-node/src/subblocks.rs
klkvr: this can't be reached if we have 0 validators, we only build a subblock when we have a next proposer to send it to

crates/commonware-node/src/subblocks.rs
klkvr: opted for explicit `debug!` traces in 3bd7ab84

feels strange to make fn return a `Result` just for the traces

crates/commonware-node/src/subblocks.rs
klkvr: once this will get changed upstream we should see subblocks test failing

ideally API for this should be something like `scheme.leader(round, certificate)` imo

crates/commonware-node/src/subblocks.rs
klkvr: this feels more like a domain of payload builder (i.e log if we didn't have enough time to collect subblocks)

for now we will always fetch them when building a payload so it's not very useful, but eventually might become relevant if we have a more adavanced builder

crates/commonware-node/src/config.rs
klkvr: >  This means a peer can send up to 128 values

128 values over which time period?

crates/payload/builder/src/lib.rs
klkvr: > we need to exit the entire builder fn, because we might have already committed a partial subblock?

yep

> so if a subblock fails, this would prevent us from building a block indefinitely ?

yep

this should never happen because we're properly validating transactions and invalid txs should never get this far, but this is indeed quite bad if reached and quite easy to protect from so I've added logic for this here e7c4b81a

crates/commonware-node/src/subblocks.rs
SuperFluffy: OK, if you promise :-)

crates/commonware-node/src/config.rs
SuperFluffy: This is `Quota::per_second`, so 128 per second.

crates/commonware-node/src/subblocks.rs
SuperFluffy: It's the cleanest way to get error events emitted in the same span they were encountered.

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -745,27 +778,37 @@ async fn verify_block<TContext: Pacer>(
...
         );
         return Ok(false);
     }
-    if block.height() != parent.height().saturating_add(1) {
-        info!("block's height must be +1 that of the parent but isn't");
-        return Ok(false);
-    }
+    let scheme = scheme_provider
+        .scheme(epoch)
+        .ok_or_eyre("epoch's scheme not found")?;
```
SuperFluffy: The event will be very hard to understand when reading logs (I don't like the name `scheme` because nobody will know what this means in this context):
```suggestion
    .ok_or_eyre("cannot determine participants in the current epoch)?;
```

crates/commonware-node/src/consensus/engine.rs
```diff
@@ -341,10 +364,14 @@ where
...
                 dkg_channel,
                 boundary_certificates_channel,
+                subblocks_channel,
             )
         })
     }
 
+    /// Start the `simplex` consensus engine.
+    ///
+    /// This will also rebuild the state of the engine from provided `Journal`.
```
SuperFluffy: This doc comment might be confusing, because the actual simplex engine is started much deeper. What this is is technically the "tempo consensus engine"

Also, no `Journal` argument is found anywhere.

crates/commonware-node/src/consensus/engine.rs
```diff
@@ -407,11 +442,17 @@ where
...
         ])
         .await
         .map(|_| ())
         // TODO: look into adding error context so that we know which
         // component failed.
         .wrap_err("one of the consensus engine's actors failed")
     }
+
+    /// Returns a handle to the subblocks service.
+    pub fn subblocks_handle(&self) -> subblocks::Mailbox {
```
SuperFluffy: :-)
```suggestion
    pub fn subblocks_mailbox(&self) -> subblocks::Mailbox {
```

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,614 @@
...
+    let mut evm = evm_at_block(&node, subblock.parent_hash)?;
+
+    let epoch = epoch::of_height(evm.block().number.to::<u64>() + 1, epoch_length)
+        .ok_or_eyre("failed to compute epoch from block number")?;
+
+    let scheme = scheme_provider
+        .scheme(epoch)
+        .ok_or_eyre("scheme not found")?;
+
+    if !scheme.participants().iter().any(|p| p == &sender) {
```
SuperFluffy: Could you `eyre::ensure!` here.

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,614 @@
...
+    if subblock.transactions.iter().any(|tx| {
+        tx.subblock_proposer()
+            .is_none_or(|proposer| !proposer.matches(&sender))
+    }) {
+        return Err(eyre::eyre!(
+            "all transactions must specify the subblock validator"
+        ));
+    }
+
+    let subblock = subblock.try_into_recovered(B256::from_slice(&sender))?;
```
SuperFluffy, line 598: context?

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,614 @@
...
+#[instrument(skip_all, err(level = Level::WARN), fields(sender = %sender))]
+async fn validate_subblock(
+    sender: PublicKey,
+    node: TempoFullNode,
+    subblock: SignedSubBlock,
+    actions_tx: mpsc::UnboundedSender<Message>,
+    scheme_provider: SchemeProvider,
+    epoch_length: u64,
+) -> eyre::Result<()> {
+    let Ok(signature) = Signature::decode(&mut subblock.signature.as_ref()) else {
```
SuperFluffy: You don't want to leak the error? I mean `signature::decode(...).wrap_err(...)`?

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,614 @@
...
+            ));
+        }
+
+        // Spawn task to validate the subblock.
+        let node = self.node.clone();
+        let validated_subblocks_tx = self.actions_tx.clone();
+        let scheme_provider = self.scheme_provider.clone();
+        let epoch_length = self.epoch_length;
+        let span = Span::current();
+        self.context.clone().shared(true).spawn(move |_| {
```
SuperFluffy, line 345: Could give that spawned task a name, i.e. `self.context.with_label("validate subblock")`

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,614 @@
...
+        });
+
+        self.subblock_builder_handle = Some(BuildSubblockTask {
+            handle,
+            parent_hash,
+            proposer: next_proposer,
+        });
+    }
+
+    #[instrument(skip_all, err(level = Level::WARN), fields(sender = %sender))]
```
SuperFluffy: length of the message is often interesting, especially if empty `fields(msg_bytes = message.len())`

crates/commonware-node/src/subblocks.rs
SuperFluffy: Let's emit this as an event still? If this fails, then this will be the first thing we will look at.

crates/commonware-node/src/subblocks.rs
SuperFluffy: Bump on this one :-)

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,614 @@
...
+                    let Some(action) = action else { break; };
+                    self.on_new_message(action);
+                },
+                // Handle messages from the network.
+                message = network_rx.recv() => {
+                    let Ok((sender, message)) = message else { continue; };
+                    let _ = self.on_network_message(sender, message);
+                },
+                // Handle built subblocks.
+                subblock = if let Some(task) = self.subblock_builder_handle.as_mut() {
```
SuperFluffy, line 157: Yeah, tokio select would make this nicer.

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,614 @@
...
+    pub fn get_subblocks(&self, parent: BlockHash) -> Result<Vec<RecoveredSubBlock>, RecvError> {
+        let (tx, rx) = std::sync::mpsc::sync_channel(1);
+        let _ = self.tx.unbounded_send(Message::GetSubBlocks {
+            parent,
+            response: tx,
+        });
+        rx.recv()
+    }
+
+    pub fn add_transaction(&self, tx: Recovered<TempoTxEnvelope>) {
```
SuperFluffy, line 460: This is only called from tests?

Can'ts send it over the p2p channel?

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,614 @@
...
+        let Ok(subblock) = SignedSubBlock::decode(&mut &*message) else {
+            return Err(eyre::eyre!("failed to decode subblock"));
+        };
+
+        let Some(tip) = self.tip() else {
+            return Err(eyre::eyre!("missing tip of the chain"));
+        };
+
+        // Skip subblocks that are not built on top of the tip.
+        if subblock.parent_hash != tip {
```
SuperFluffy: `eyre::ensure!` could make this nicer.

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,614 @@
...
+            )
+            .instrument(span)
+        });
+
+        Ok(())
+    }
+
+    #[instrument(skip_all, fields(subblock.validator = %subblock.validator(), subblock.parent_hash = %subblock.parent_hash))]
+    fn on_validated_subblock(&mut self, subblock: RecoveredSubBlock) {
+        // SKip subblock if we are already past its parent
```
SuperFluffy:
```suggestion
        // Skip subblock if we are already past its parent
```

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,614 @@
...
+            }
+        };
+
+        if Some(subblock.parent_hash) != self.tip() {
+            return;
+        }
+
+        debug!(
+            ?subblock,
+            ?next_proposer,
```
SuperFluffy: Implements display
```suggestion
            %next_proposer,
```

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,614 @@
...
+    async fn on_built_subblock(
+        &mut self,
+        subblock: Result<RecoveredSubBlock, commonware_runtime::Error>,
+        next_proposer: PublicKey,
+        network_tx: &mut impl Sender<PublicKey = PublicKey>,
+    ) {
+        let subblock = match subblock {
+            Ok(subblock) => subblock,
+            Err(err) => {
+                warn!(%err, "failed to build subblock");
```
SuperFluffy: Easier to grep events
```suggestion
                warn!(%error, "failed to build subblock");
```

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,614 @@
...
+                    let _ = self.on_network_message(sender, message);
+                },
+                // Handle built subblocks.
+                subblock = if let Some(task) = self.subblock_builder_handle.as_mut() {
+                    (&mut task.handle).fuse()
+                } else {
+                    futures::future::Fuse::terminated()
+                } => {
+                    let task = self.subblock_builder_handle.take().unwrap();
+                    self.on_built_subblock(subblock, task.proposer, &mut network_tx).await;
```
SuperFluffy, line 163: The `.await` is surprising/unfortunate because all other branches don't need it.

They should provide a `.try_send` that's not async.

crates/commonware-node/src/subblocks.rs
```diff
@@ -0,0 +1,614 @@
...
+    pub async fn run(
+        mut self,
+        (mut network_tx, mut network_rx): (
+            impl Sender<PublicKey = PublicKey>,
+            impl Receiver<PublicKey = PublicKey>,
+        ),
+    ) {
+        loop {
+            commonware_macros::select! {
+                // Handle messages from consensus engine and service handle.
```
SuperFluffy, line 146: IMO the comments on all the branches are not necessary because they are mostly self-explanatory.

Issue #565, opened 2025-10-27
Consider generating intermediate DKG ceremony dealings on request, not eagerly


SuperFluffy:
As a followup to https://github.com/tempoxyz/tempo/pull/538, we can consider generating the intermediate DKG outcome upon request from the execution driver, and not eagerly as is done right now.

Issue #562, opened 2025-10-27
Use RLP to write dkg ceremony extra data


SuperFluffy:
As a followup to https://github.com/tempoxyz/tempo/pull/538, we want to use RLP encoding to write ceremony data to the extraData header (instead of the commonware coding/encoding mechanism). This way, third parties will have an easier time parsing this data.

PR #585, opened 2025-10-29
feat!(consensus): add a validator dynamically

### LINKED ISSUE #310: Investigate merging tempo and reth datadirs
Tempo right now can store consensus and execution data in two different places, depending on how either config/command line arguments are configured. A node can be started in such a way that both end up under the same parent directory, but maybe it makes sense to make this a stronger requirement.

`xtask generate-config` shows how this is done manually by setting `--datadir` to write to the same place as consensus `Config.storage_directory`:

https://github.com/tempoxyz/tempo/blob/9bfb25275c02d1bab3a381e92fd349acbbdc627e/xtask/src/main.rs#L190-L195

### LINKED ISSUE #563: Store epoch length, initial BLS key, initial validator set public keys in chainspec
Right now, epoch lengths, the initial BLS key, and the initial validator set are passed to the node via config: https://github.com/tempoxyz/tempo/tree/6a303fbbee8df3208012e8a917deca79d04a1ff0/crates/commonware-node-config

This should be moved onchain, by writing it to the chainspec.

### LINKED ISSUE #883: consensus: write full peer set for an epoch to chain
### Describe the feature

### Additional context

### Sidequests:

SuperFluffy:
This patch allows dynamically adding and removing validators from Tempo by using the `ValidatorConfig` smart contract.

There is a high level overview of the protocol provided in `docs/network-reconfiguration.md` provided with this patch.

### Configuration

1. the public polynomial was moved from config to the `genesis.extra_data` field (to make reading it consistent with how the public polynomials are written during DKG ceremonies)
2. the initial set of peers was moved from config to `genesis.extra_data`
3. the socket addresses / host names of the initial set of peers are moved to the `ValidatorConfig` smart contract
4. the epoch length is moved to `genesis.config.extra_fields`

### Startup and execution

1. the consensus layer is now configured via command line arguments, moving it closer to reth / the execution layer.
2. the crate `tempo-commonware-node-config` was replaced by `tempo-consesus-args`.l

### DKG ceremonies

1. where before `dealers == players` in each ceremony, the players in a new ceremony are now established by doing a `getValidators` call of the `ValidatorConfig` smart contract (technically, the validators returned by the contract call will be the players in the next ceremony, giving them time to sync for one entire epoch; see the spec provided with this patch).
2. the p2p peers are similarly established through `getValidators`, where we track the peers of the last 3 epochs (current, parent, grandparent). The peers in epoch N is the union of the peers of `N`, `N+1` and `N+2`.

### Xtasks

1. update xtask genesis generator to inject validators, public polynomial into the smart contract and thus genesis
2. rework the `generate-genesis` task by breaking out genesis-generation into its own set of arguments
3. rework the `generate-devnet` task in terms of this this new genesis-generation

### e2e testing

1. rework e2e tests to get access to configs, execute smart contract calls from within the execution/tokio runtime
2. provide an e2e test for adding a validator, an e2e test for a validator having lost its share and regaining it in a subsequent round

## Followup issues

https://github.com/tempoxyz/tempo/issues/562
https://github.com/tempoxyz/tempo/issues/565
https://github.com/tempoxyz/tempo/issues/792
https://github.com/tempoxyz/tempo/issues/884
https://github.com/tempoxyz/tempo/issues/924
https://github.com/tempoxyz/tempo/issues/928
https://github.com/tempoxyz/tempo/issues/930
https://github.com/tempoxyz/tempo/issues/933

Closes #310 
Closes #563 
Closes #883


crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -481,3 +609,263 @@ struct Metrics {
...
+    };
+
+    // FIXME(janis): pushing this finalized block into the execution node and
+    // starting a new ceremony happens concurrently. We need to ensure that
+    // we wait until the block is actually part of the execution layer to
+    // access this state.
+    let db = State::builder()
+        .with_database(StateProviderDatabase::new(
+            node.provider
+                .state_by_block_id(block_or_genesis.to_block_id())
```
SuperFluffy: This is inherently racy - we operate on finalized blocks here and cannot be 100% sure that the block was actually already included in reth. Unsure how to enforce proper order of operations.

crates/primitives/src/header.rs
```diff
@@ -22,6 +22,14 @@ pub struct TempoHeader {
     #[cfg_attr(feature = "serde", serde(with = "alloy_serde::quantity"))]
     pub timestamp_millis_part: u64,
 
+    /// The public BLS polynomial used for sucinct consensus certificates.
+    // FIXME(janis): explicit hex string or should the deserializer handle this?
+    pub public_polynomial: Vec<u8>,
+
+    /// The length of measures in heights.
+    #[cfg_attr(feature = "serde", serde(with = "alloy_serde::quantity"))]
+    pub epoch_length: u64,
```
klkvr: what's the motivation for adding those here? looks like we only read them from genesis header and never validate/set those values properly for other blocks

SuperFluffy: No motivation other than sifting through `ChainSpec<TempoHeader>` and noticing that this header is the only type I can actually change.

Is there a better escape hatch I should use?

I also only really need these during genesis. The public polynomial is otherwise read from the extra data field on the epoch boundary (the read is not actually done right now, but should be), and the epoch length should never change.

klkvr: could we not just add a field to `TempoChainSpec`?

SuperFluffy: Ah, that works? I thought it always needs to be constructed from genesis, like we do right now:

```
pub static ADAGIO: LazyLock<Arc<TempoChainSpec>> = LazyLock::new(|| {
    let genesis: Genesis = serde_json::from_str(include_str!("./genesis/adagio.json"))
        .expect("`./genesis/adagio.json` must be present and deserializable");
    TempoChainSpec::from_genesis(genesis).into()
});
```

Is there a mechanism that ensures that all nodes parse the same genesis? I don't see anything here at first glance that generates a hash over the entire tempo chain spec?

crates/commonware-node/src/dkg/manager/actor.rs
SuperFluffy: No longer racy: when processing the boundary height N, we read the validator config at N-1.

crates/primitives/src/header.rs
SuperFluffy: `epoch_length` is now part of the `genesis.config.extra_fields`.

The initial DKG values are now part of `genesis.extra_data`

Cargo.toml
```diff
@@ -242,3 +244,15 @@ test-case = "3"
 # build deps
 vergen = "9"
 vergen-git2 = "1"
+
+[patch.crates-io]
+commonware-broadcast = { git = "https://github.com/commonwarexyz/monorepo", rev = "554fd5adca87404903326b20a95d8893c038e2f8" }
```
SuperFluffy: ~~Not necessary. Will change this to upstream~~

This is right after https://github.com/commonwarexyz/monorepo/pull/2123 was merged.

crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -408,89 +459,494 @@ where
...
+            EvmPrecompileStorageProvider::new_max_gas(internals, node.chain_spec().chain_id());
+
+        let mut validator_config =
+            ValidatorConfig::new(VALIDATOR_CONFIG_ADDRESS, &mut precompile_storage);
+        validator_config
+            .get_validators(IValidatorConfig::getValidatorsCall {})
+            .wrap_err("failed to query contract for validator config")?
+    };
+
+    let addresses = decode_contract_validators(contract_validators)
```
joshieDo: i think we should be able to use the `interface IValidatorConfig` as any other contract using the execution_node provider?  similar to https://github.com/alloy-rs/examples/blob/b2f76090ad98f59f29746a5ca6953f651d843a63/examples/contracts/examples/interact_with_abi.rs#L19-L26

joshieDo: https://github.com/tempoxyz/tempo/blob/69d7d3e73e7437f30ca6a708b298c0e295a2fe19/crates/contracts/src/precompiles/validator_config.rs#L28-L30

klkvr: this would mean we'd have to construct an alloy provider and go through RPC code for this which doesn't seem ideal

but also might be fine given that it'd be pretty much the same thing

SuperFluffy: Need some guidance on this - should I or should I not go through alloy?

I wrote this looking at the various integration tests and threads on slack.

joshieDo: >this would mean we'd have to construct an alloy provider and go through RPC code for this which doesn't seem ideal

yeah, true. this is already fetching directly from storage and going through RPC would add additional unnecessary overhead.

feels a bit weird having all that evm initialization here though, i'd like a specific provider trait that would do this - but outside the scope of this PR

crates/chainspec/src/spec.rs
```diff
@@ -19,12 +19,16 @@ use tempo_primitives::TempoHeader;
...
-struct TempoGenesisInfo {
+pub struct TempoGenesisInfo {
     /// Timestamp of Adagio hardfork activation
     #[serde(skip_serializing_if = "Option::is_none")]
-    adagio_time: Option<u64>,
+    pub adagio_time: Option<u64>,
+
+    /// The length of a Tempo epoch as measured in blocks.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub epoch_length: Option<u64>,
```
joshieDo: shouldn't this be a required field?

SuperFluffy: Yeah, it should from the point of view of the consensus layer. But it screws with so many of our tests and I figured that the execution layer does not really need to care about this.

crates/chainspec/src/spec.rs
```diff
@@ -305,7 +353,7 @@ mod tests {
 
     #[test]
     fn test_parse_tempo_hardforks_from_genesis_extra_fields() {
-        // Create a genesis with Tempo hardfork timestamps as extra fields in config
```
SuperFluffy, line 316: undo

crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -98,22 +139,74 @@ where
...
             config,
             context,
             mailbox,
             ceremony_metadata: Arc::new(Mutex::new(ceremony_metadata)),
             epoch_metadata,
+            p2p_metadata,
+            epoch_state,
             metrics,
-        }
+            p2p_states: P2pStates::new(),
```
joshieDo: should be using the `p2p_states` above

joshieDo: still not super familiar with this bit, so the fact that tests are passing means we don't have test coverage here ig?

SuperFluffy: a) good catch, and b) yes. This is the recovery case, for which we want to use https://github.com/tempoxyz/tempo/pull/783.

Killing a single node became possible just a few days ago.

SuperFluffy: Fixed in afadaf9

crates/commonware-node/src/consensus/application/actor.rs
joshieDo, line 668: this is confusing, could we do proper error handling/report here with a error enum variant (in this PR or another)

SuperFluffy: The reason we have done it this way is to distinguish outright errors (can't communicate with subparts of the system) from a valid result (the block being invalid). So the difference is between being able to render a decision (block is valid/invalid), and not being able to render a decision (don't have information on the block, internal system error).

But happy to sync on how to change this in a future PR (doesn't fit this PR).

joshieDo, line 425: iiuc, right now if the `PublicOutcome` is invalid (`!= our_outcome`, for whatever reason) we just wait for the next proposer to propose a valid one once this is nullified? 
(just for my understanding)

SuperFluffy: Yes, indeed. This is going round until we have an honest proposer proposing our result.

SuperFluffy: After your comment I reconsidered how the DKG actor should deal with finalized blocks. It is now also blocking on finalization, similar to the application: [646951f](https://github.com/tempoxyz/tempo/pull/585/commits/646951f54406a562afa9abbc9be3b9c4ff78ac04)

Cargo.toml
```diff
@@ -13,7 +13,7 @@ members = [
   "crates/alloy",
   "crates/chainspec",
   "crates/commonware-node",
-  "crates/commonware-node-config",
```
SuperFluffy, line 16: This is weird - why did node-config fall out of members list?

crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -131,72 +223,57 @@ where
...
-                .set(epoch_state.participants.len() as i64);
-            self.metrics
-                .ceremony_players
-                .set(epoch_state.participants.len() as i64);
-            ceremony::Ceremony::init(
-                &mut self.context,
-                &mut ceremony_mux,
-                self.ceremony_metadata.clone(),
-                config,
+        let mut ceremony = Some(self.start_new_ceremony(&mut ceremony_mux).await);
```
SuperFluffy: @klkvr Flagging this - I made the `ceremony: Option<Ceremony>` again, because everything else turned out to be a cumbersome workaround and this expressed it best.

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -656,8 +657,10 @@ impl Inner<Init> {
             };
             if our_outcome != block_outcome {
                 warn!(
+                    our.epoch = our_outcome.epoch,
                     our.participants = ?our_outcome.participants,
                     our.public = ?our_outcome.public,
+                    block.epoch = block_outcome.epoch,
                     block.participants = ?block_outcome.participants,
                     block.public = ?block_outcome.public,
                     "our public dkg ceremont outcome does not match what's stored in the block; failing block",
```
joshieDo:
```suggestion
                    "our public dkg ceremony outcome does not match what's stored in the block; failing block",
```

crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -408,89 +461,490 @@ where
...
+                .state_by_block_id(at_height.into())
+                .wrap_err_with(|| {
+                    format!("failed to get state from node provider for height `{at_height}`")
+                })?,
+        ))
+        .build();
+
+    // XXX: Ensure that evm and internals go out of scope before the await point
+    // below.
+    let contract_validators = {
```
joshieDo: as a new player, we are initially creating a peerset from the genesis block (or snapshotted block )- however if all validators have rotated their keys then i guess we'll never progress  ?

joshieDo: i wonder (in a follow-up), if we could add an external rpc provider to help bootstrap new players and/or a cli/config option

docs/network-reconfiguration.md
```diff
@@ -0,0 +1,150 @@
...
+epoch is then defined by:
+
+1. the public threshold key polynomial;
+2. the set of private threshold key shares (one per validator);
+3. the list `[ed25519-public-key]` of public ed25519 keys uniquely identifying
+   each validator, socket-address)`.
+
+At genesis, the network is launched with a known public threshold key
+polynomial, a set of private threshold key shares (one per validator), a list
+of `(ed25519-public-key, socket-address)` to identify and coontact each
```
shekhirin:
```suggestion
of `(ed25519-public-key, socket-address)` to identify and contact each
```

crates/commonware-node/src/consensus/application/actor.rs
joshieDo: to me they just seem like different error variants (yeah not this PR) that would need to be handled, but would be way more readable imo

crates/commonware-node/src/dkg/manager/actor.rs
SuperFluffy: > as a new player, we are initially creating a peerset from the genesis block (or snapshotted block )- however if all validators have rotated their keys then i guess we'll never progress ?

Commonware subsumes this issue under "synchrony assumptions", and this is also the reason why we register the union of the last 3 epochs for the current epoch.

Now naturally this is not enough: we also need to ensure that our peer sets do not change too much within an epoch so that enough validators stay online for a node to catch up. Commonware suggested that up to 5% of validators should be able to rotate their keys.

SuperFluffy: > i wonder (in a follow-up), if we could add an external rpc provider to help bootstrap new players

@Zygimantass what do you think?

crates/commonware-node/src/consensus/application/actor.rs
SuperFluffy: Yeah, I see that. It's a bit of a pain in a dynamic-error world.

crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -408,89 +461,490 @@ where
...
+    use reth_provider::{BlockReader as _, StateProviderFactory as _};
+    use tempo_precompiles::{
+        VALIDATOR_CONFIG_ADDRESS,
+        storage::evm::EvmPrecompileStorageProvider,
+        validator_config::{IValidatorConfig, ValidatorConfig},
+    };
+
+    let block = node
+        .provider
+        .block_by_number(at_height)
```
joshieDo: is there any guarantee that EL has processed this block?  lets say a scenario where we had some network downtime, and the moment we get back online a peer sends us a finalization - wouldnt both the application and dkg see this in their mailbox at the "same time"?

crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -408,89 +461,490 @@ where
...
+    use reth_ethereum::evm::revm::{
+        State, context::ContextTr as _, database::StateProviderDatabase,
+    };
+    use reth_node_builder::{Block as _, ConfigureEvm as _};
+    use reth_provider::{BlockReader as _, StateProviderFactory as _};
+    use tempo_precompiles::{
+        VALIDATOR_CONFIG_ADDRESS,
+        storage::evm::EvmPrecompileStorageProvider,
+        validator_config::{IValidatorConfig, ValidatorConfig},
+    };
```
joshieDo: nit: move top level

crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -408,89 +461,490 @@ where
...
+
+    Ok(P2pState { peers })
+}
+
+struct Addresses {
+    inbound: SocketAddr,
+    _outbound: SocketAddr,
+}
+
+use tempo_precompiles::validator_config::IValidatorConfig;
```
joshieDo: top level

crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -1,55 +1,82 @@
...
+    config: super::Config<TPeerManager>,
     context: ContextCell<TContext>,
     mailbox: mpsc::UnboundedReceiver<super::Message>,
 
     ceremony_metadata: Arc<Mutex<Metadata<ContextCell<TContext>, U64, CeremonyState>>>,
     epoch_metadata: Metadata<ContextCell<TContext>, U64, EpochState>,
+    p2p_metadata: Metadata<ContextCell<TContext>, U64, P2pState>,
+
+    epoch_state: EpochState,
+    p2p_states: P2pStates,
```
klkvr: what is the difference between `p2p_states`-`p2p_metadata`, `epoch_state`-`epoch_metadata`? are the `state` ones just a caching layer?

crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -252,145 +332,118 @@ where
...
-        if ceremony.epoch().saturating_sub(1) == block_epoch
-            && utils::is_last_block_in_epoch(self.config.epoch_length, block.height())
-                .is_some_and(|e| e == block_epoch)
-        {
-            debug!(
-                "reached last height of outgoing epoch; reporting that a \
-                new epoch can be entered"
-            );
+            maybe_ceremony.replace(self.start_new_ceremony(ceremony_mux).await);
+            let merged_peer_set = self.p2p_states.construct_merged_peerset();
```
klkvr: should `start_new_ceremony` handle peerset update as well?

crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -408,89 +461,490 @@ where
...
+
+    fn highest(&self) -> Option<&P2pState> {
+        self.inner.last_key_value().map(|(_, p2p)| p2p)
+    }
+
+    fn highest_epoch(&self) -> Option<Epoch> {
+        self.inner.last_key_value().map(|(epoch, _)| *epoch)
+    }
+
+    /// Merges the peers in the current, parent, and grandparent p2p states.
```
patrick-ogrady: > The peers in epoch N is the union of the peers of N, N-1 and N-2 (this is to ensure that a validator that fell behind 2 epochs can still catch up).

You should instead go `N, N+1, N+2` (the validators that fell behind are handled by `tracked_peer_sets`): https://github.com/commonwarexyz/monorepo/blob/27cee9d0dee2c1b0c27528ac7228572079ed0e92/p2p/src/authenticated/lookup/config.rs#L84-L92).

Here is a good example (N == Dealers, N+1 == Players, N+2 == Syncing Players): https://github.com/commonwarexyz/monorepo/blob/27cee9d0dee2c1b0c27528ac7228572079ed0e92/examples/reshare/src/dkg/actor.rs#L224-L238

crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -408,89 +461,490 @@ where
...
+
+    fn add(&mut self, epoch: Epoch, new: P2pState) -> Option<(Epoch, P2pState)> {
+        use std::collections::btree_map::Entry;
+        match self.inner.entry(epoch) {
+            Entry::Vacant(vacant_entry) => {
+                vacant_entry.insert(new);
+            }
+            Entry::Occupied(mut slot) => {
+                if slot.get() != &new {
+                    warn!(
```
patrick-ogrady: You may want to assert this doesn't happen.

crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -408,89 +461,490 @@ where
...
+        // XXX(!!!): This is critical: start_new_epoch *MUST* be called on the
+        // first block, and the validator config *MUST* be read on the
+        // *BOUNDARY* of the previous epoch. This ensures that the block is
+        // firmly committed on the execution layer.
+        let boundary = self.epoch_state.epoch.checked_sub(1).map_or(0, |previous| {
+            utils::last_block_in_epoch(self.config.epoch_length, previous)
+        });
+
+        let new_p2p_state =
+            match read_validator_config_from_contract(self.config.execution_node.clone(), boundary)
```
patrick-ogrady: I think you may want to loop on this until you get the data (if execution is behind or something -> not sure if possible here)?

crates/commonware-node/src/dkg/manager/actor.rs
patrick-ogrady: As noted elsewhere, you should be using `tracked_peer_sets` for this case (and doing a merged peer set on a go forward basis).

If you want to backfill with old peer sets (before the epoch you are in on restart), that will work just make sure to do it in order.

patrick-ogrady: As discussed, we think there is a straightforward way to help new peers sync with a finalization cert on some boundary and either a proof of the contract state at that point (who will be the current peers) or by putting the info in the extra data field.

crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -408,89 +461,490 @@ where
...
+            "resolved inbound and outbound addresses to socket addresses; \
+            always taking the last one in case there are more than one"
+        );
+
+        if parsed_vals
+            .insert(
+                peer.clone(),
+                Addresses {
+                    inbound,
+                    _outbound: outbound,
```
patrick-ogrady: Hope to allow registering inbound/outbound unique IPs for each peer soon.

crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -252,145 +332,118 @@ where
...
+    ///
+    /// The processing of all other blocks depends on which part of the epoch
+    /// they fall in:
+    ///
+    /// + first half: if we are a dealer, distribute the generated DKG shares
+    ///   to the players and collect their acks. If we are a player, receive
+    ///   DKG shares and respond with an ack.
+    /// + exact middle of an epoch: if we are a dealer, generate the dealing
+    ///   (the intermediate outcome) of the ceremony.
+    /// + second half of an epoch: if we are a dealer, send it to the application
```
patrick-ogrady, line 436: May at some point be useful to reject duplicate submissions but that would require tracking which dealings are in the pending ancestry

crates/commonware-node/src/dkg/manager/actor.rs
patrick-ogrady: I "think" it is more correct to setup the peer set for the next epoch between the ending and start of the new (otherwise you'll be playing with one hand tied behind your back when trying to finalize the first block).

patrick-ogrady: I suspect it is also better to start ceremony in this "in-between" time as well...

crates/commonware-node/src/dkg/ceremony/mod.rs
```diff
@@ -561,29 +572,29 @@ where
...
+                    block_outcome.dealer(),
+                    block_outcome.commitment(),
                 )
         }) {
-            self.arbiter.disqualify(block_outcome.dealer.clone());
+            self.arbiter.disqualify(block_outcome.dealer().clone());
             bail!("invalid ack signatures; disqualifying dealer");
         }
 
         // Check dealer commitment
```
patrick-ogrady, line 593: You may want to avoid ack'ing shares from a dealer that had their share revealed in a previous round (while it may provide randomness to the new polynomial it is publicly revealed and may not be "best practice"). We can discuss this at length in the future but may want to create an issue to track.

crates/dkg-onchain-artifacts/src/lib.rs
```diff
@@ -0,0 +1,408 @@
...
+        _cfg: &Self::Cfg,
+    ) -> Result<Self, commonware_codec::Error> {
+        let n_players = UInt::read(buf)?.into();
+        Ok(Self {
+            n_players,
+            dealer: PublicKey::read(buf)?,
+            dealer_signature: Signature::read(buf)?,
+            epoch: UInt::read(buf)?.into(),
+            commitment: Public::<MinSig>::read_cfg(buf, &(quorum(n_players as u32) as usize))?,
+            acks: Vec::read_cfg(buf, &(RangeCfg::from(0..=usize::MAX), ()))?,
```
patrick-ogrady: You should use `Cfg` to bound the length of this Vec (and the one below). Otherwise, a malicious dealer could post an outcome and ask you to preallocate a Vec of `usize::MAX` (what you are specifying in that range config).

patrick-ogrady: `n_players` could probably be used to limit this?

crates/dkg-onchain-artifacts/src/lib.rs
```diff
@@ -0,0 +1,408 @@
...
+        n_players: u64,
+        dealer_signer: &PrivateKey,
+        namespace: &[u8],
+        epoch: Epoch,
+        commitment: Public<MinSig>,
+        acks: Vec<Ack>,
+        reveals: Vec<group::Share>,
+    ) -> Self {
+        // Sign the resharing outcome
+        let payload = Self::signature_payload_from_parts(epoch, &commitment, &acks, &reveals);
```
patrick-ogrady: You need to include all fields in the signature or else it can be trivially tampered with (if you use `n_players` to parse fields below this would allow an observer to make what looks like a valid `IntermediateOutcome` look invalid).

crates/dkg-onchain-artifacts/src/lib.rs
patrick-ogrady: You'd still need to cap it at a reasonable value tho

crates/e2e/src/lib.rs
```diff
@@ -7,79 +7,177 @@
...
+        let public_key = self.public_key.clone();
+        let uid = self.uid.clone();
+        let engine = self
+            .engine_config
+            .try_init()
+            .await
+            .expect("must be able to start the engine");
+        let pending = oracle
+            .control(public_key.clone())
+            .register(0)
```
patrick-ogrady, line 73: nit: recommend referring to `const` for these channels

docs/network-reconfiguration.md
```diff
@@ -0,0 +1,150 @@
...
+
+The state of a network (and thus the outcome of a DKG ceremony) for a given
+epoch is then defined by:
+
+1. the public threshold key polynomial;
+2. the set of private threshold key shares (one per validator);
+3. the list `[ed25519-public-key]` of public ed25519 keys uniquely identifying
+   each validator, socket-address)`.
+
+At genesis, the network is launched with a known public threshold key
```
patrick-ogrady: May be worth noting that initial polynomial's constant term will stay constant across reshares (and can be used to verify certificates from any epoch).

docs/network-reconfiguration.md
```diff
@@ -0,0 +1,150 @@
...
+separate purpose-build chain to generate the initial public polynomial and set
+of initial key shares without revealing them.
+
+[1]: https://docs.rs/commonware-consensus/0.0.63/commonware_consensus/simplex/index.html#protocol-description
+[2]: https://docs.rs/commonware-consensus/0.0.63/commonware_consensus/simplex/index.html#signing_schemebls12381_threshold
+[3]: https://docs.rs/commonware-p2p/0.0.63/commonware_p2p/authenticated/lookup/index.html
+
+## Genesis data
+
+At genesis, `epoch_length` is read from `genesis.config.extra_fields`. The
```
patrick-ogrady: May want to note this shouldn't be changed during the lifetime of the chain

docs/network-reconfiguration.md
```diff
@@ -0,0 +1,150 @@
...
+[6]: ../crates/contracts/src/precompiles/validator_config.rs
+
+## Protocol of a DKG ceremony
+
+The process of initializing, running, and completing a DKG ceremony is closely
+coupled to finalized blocks. The steps that Tempo is taking to run such a DKG
+protocol are this explained in terms of finalized blocks below. One strong
+assumption: each finalized block will be processed in strict sequential order.
+
+For a detailed description of the role of players, dealers
```
patrick-ogrady: nit: dangling

docs/network-reconfiguration.md
```diff
@@ -0,0 +1,150 @@
...
+contract on-chain. On success, they will become the validators of the next epoch.
+
+For a lower level overview of DKG ceremony and the role of players and dealers,
+see [7]
+
+For each finalized block, the following actions will be taken depending on its
+height H.
+
+1. on `H == C*E` (first height of the epoch):
+   a. read the ceremony players using the `getValidators` call. Note:
```
patrick-ogrady: as noted elsewhere, recommend doing this before the epoch starts (the last block of the last)

crates/chainspec/src/spec.rs
patrick-ogrady: Agreed 👍

patrick-ogrady (commented):
Did an initial scan!

Primary topics to discuss:
* Peer set contents
* Height to update peer set
* Limiting the size of vector parsing from untrusted sources

crates/commonware-node/src/dkg/manager/actor.rs
SuperFluffy: Not at the level of this function, but in `fn start_new_ceremony`. Specifically, we call `start_new_ceremony` on block 1 of epoch E. `start_new_ceremony` then calls this function with block N of epoch E-1 (the boundary block/last block of the previous epoch).

So this relies on us receiving new finalized blocks from the marshal actor in strictly sequential order, with `block 1 / Epoch E` coming in only after `block N / Epoch E-1` was processed.


But in the light of this backfill issue, I should probably loop on this function.

crates/commonware-node/src/dkg/manager/actor.rs
SuperFluffy: I have reworked this logic in [ba7e31b](https://github.com/tempoxyz/tempo/pull/585/commits/ba7e31b7ccf46ccbbaba93ccea034d05fdbde53e). Now tracking forward instead of backward.

crates/commonware-node/src/dkg/manager/actor.rs
SuperFluffy: Reworked in [ba7e31b](https://github.com/tempoxyz/tempo/pull/585/commits/ba7e31b7ccf46ccbbaba93ccea034d05fdbde53e)

crates/commonware-node/src/dkg/manager/actor.rs
SuperFluffy: Updated so that we are looping on this: [52d7521](https://github.com/tempoxyz/tempo/pull/585/commits/52d7521d5891d46de4a00b848e01979b8c839158)

We know that we are processing the finalized height at the same time as the application/automaton actor. This means its "safe" to loop here since it will eventually process it.

Note that I believe that the version before was fine: on the first height of E we were reading the boundary of E-1. Given that finalized blocks are processed in order, it was guaranteed that the application ack'd the finalization.

However, I think this solution is cleaner and needs less explanation.

crates/commonware-node/src/dkg/manager/actor.rs
SuperFluffy: It does now with [54a1b39](https://github.com/tempoxyz/tempo/pull/585/commits/54a1b3983682c530f33fe1dec63f87611b6cddf5), but more importantly it was expanded on in [ba7e31b](https://github.com/tempoxyz/tempo/pull/585/commits/ba7e31b7ccf46ccbbaba93ccea034d05fdbde53e)

crates/commonware-node/src/dkg/manager/actor.rs
SuperFluffy: Yes indeed. `epoch_state` is in-memory. `epoch_metadata` is on-disk. Same with `p2p-states` - although that was changed to `registered_peers_metadata` and now slightly orthogonal to the new `all_participants`.

crates/e2e/src/lib.rs
SuperFluffy: I chose not to do that because I wanted to keep these crate-private (in commonware-node) - and the tags themselves are transparent/unkonwn to the application itself.

crates/dkg-onchain-artifacts/src/lib.rs
SuperFluffy: Good point and my oversight. I was planning to rework this to use RLP encoding.

crates/dkg-onchain-artifacts/src/lib.rs
SuperFluffy: Overlooked this when doing a quick fix. Good catch! Signing over `n_players` as well now - and I limited `n_players: u16`:  f56a621

crates/commonware-node/src/dkg/manager/actor.rs
SuperFluffy: > As noted elsewhere, you should be using tracked_peer_sets for this case (and doing a merged peer set on a go forward basis).
> If you want to backfill with old peer sets (before the epoch you are in on restart), that will work just make sure to do it in order.
Reworked this in [ba7e31b](https://github.com/tempoxyz/tempo/pull/585/commits/ba7e31b7ccf46ccbbaba93ccea034d05fdbde53e). Now the actually registered peer set is cached/written to disk for epochs E-2, E-1. Epoch E is always determined from reading the on-chain data.

That would be extremely valuable. I will add this in a follow-up. https://github.com/tempoxyz/tempo/issues/883

crates/commonware-node/src/dkg/manager/actor.rs
SuperFluffy: > May at some point be useful to reject duplicate submissions but that would require tracking which dealings are in the pending ancestry

Yes - would provide this in a followup. Right now this value is unset once we have seen it in a finalized block.

crates/commonware-node/src/dkg/manager/actor.rs
SuperFluffy: I used to have this, but shuffled it around before your review. I have moved this logic back into [54a1b39](https://github.com/tempoxyz/tempo/pull/585/commits/54a1b3983682c530f33fe1dec63f87611b6cddf5)

Now the ceremony (and hence registered the peers) is done on the boundary of the last epoch.

crates/commonware-node/src/dkg/ceremony/mod.rs
SuperFluffy: Yeah, very good point: https://github.com/tempoxyz/tempo/issues/884

crates/dkg-onchain-artifacts/src/lib.rs
SuperFluffy: Addressed in bfeeae9.

I also think codec could be changed to not pre-allocate the entire buffer, but instead read in chunks?

crates/commonware-node/src/consensus/engine.rs
```diff
@@ -124,6 +115,14 @@ where
             },
         );
 
+        let Some(epoch_length) = self.execution_node.chain_spec().info.epoch_length else {
+            return Err(eyre::eyre!(
+                "epochLength not set in genesis / chainspec; cannot continue without it"
+            ));
```
clabby: This should likely be an `expect`.

crates/commonware-node/src/dkg/ceremony/mod.rs
```diff
@@ -667,7 +678,7 @@ where
         );
 
         let dealing_outcome = Some(IntermediateOutcome::new(
-            self.config.players.len() as u64,
+            self.config.players.len().clamp(0, u16::MAX as usize) as u16,
```
clabby:
```suggestion
            u16::try_from(self.players.len()).expect("too many players")
```

clabby: Unrealistic that > `u16::MAX` players could be registered, but should be an invalid state (and prevented within the contract.)

crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -99,21 +143,128 @@ where
...
+        let players = read_validator_config_from_contract(0, &config.execution_node, players_epoch)
+            .await
+            .wrap_err_with(|| {
+                format!(
+                    "validator config could not be read for epoch \
+                    `{players_epoch}` (latest known epoch `{}`)",
+                    epoch_state.epoch,
+                )
+            })?;
+        all_participants.push(players);
```
clabby: 1. It looks like we're using the epoch as the block height for `read_validator_config_from_contract` - is this intentional? it seems like this will return the wrong result, just fetching the state of the contract at block heights `epoch-2` & `epoch-1` rather than the first block height of each epoch. (Seeing that internally to `read_validator_config_from_contract`, the third argument is passed directly to `block_by_number`.)

2. Does the contract handle failures in the ceremony? It looks like we're directly loading the state from the contract, but that state isn't persisted by the protocol. What happens if e.g. the ceremony fails 3 times in a row? Wouldn't we read the wrong set of dealers/players?

SuperFluffy: Re 1: addressed in [f23d55b](https://github.com/tempoxyz/tempo/pull/585/commits/f23d55b4e8d8eafc0c12fd01394b66fc375f7d48). Always reading in terms of `boundary(E-1)` now.

Re 2: Good catch: the outcome of the public polynomial is persisted in the boundary height of the ending epoch. I.e. `boundary(E-1)` contains the outcome of `ceremony(E-1)` upon entering epoch `E`. This means the following:

if the ceremony fails 3 times, the participants of the ceremony E-3 will be used in epoch E. This means that the participants can have fallen out of the set.

This would be gracefully addressed by tracking the actual participants/peers in https://github.com/tempoxyz/tempo/issues/883

crates/commonware-node/src/consensus/engine.rs
SuperFluffy: I don't see a reason to panic the program if I can just exit the constructor at this point - I can handle the condition (by stopping construction), and hence I should.

docs/network-reconfiguration.md
SuperFluffy: 7acac3e9

crates/commonware-node/src/consensus/application/actor.rs
SuperFluffy: 7acac3e9

crates/commonware-node/src/consensus/engine.rs
SuperFluffy: Resolved this by making `epochLength` a required field on the chainspec.

crates/chainspec/src/spec.rs
SuperFluffy: Made this a required field. Wasn't as bad as I thought: a1f0c039

crates/commonware-node/src/dkg/manager/actor.rs
SuperFluffy: We are looping now in [52d7521](https://github.com/tempoxyz/tempo/pull/585/commits/52d7521d5891d46de4a00b848e01979b8c839158)

crates/commonware-node/src/dkg/manager/actor.rs
SuperFluffy: Reorganized everything in 54f81ccc

crates/commonware-node/src/dkg/manager/actor.rs
SuperFluffy: 54f81ccc

crates/commonware-node/src/dkg/ceremony/mod.rs
SuperFluffy: [21fcb8a](https://github.com/tempoxyz/tempo/pull/585/commits/21fcb8ace3078c48dcf6e9a7a6fc05d98c04d0f1)

> Unrealistic that > `u16::MAX` players could be registered, but should be an invalid state (and prevented within the contract.)

https://github.com/tempoxyz/tempo/issues/928

docs/network-reconfiguration.md
SuperFluffy: 3892a73a, good thing to emphasize.

docs/network-reconfiguration.md
SuperFluffy: 3892a73a

docs/network-reconfiguration.md
SuperFluffy: 3892a73a

docs/network-reconfiguration.md
SuperFluffy: 3892a73a

xtask/src/devnet.rs
SuperFluffy, line 1: I still need to address this object here.

SuperFluffy: This is now addressed by commits [f79fea9](https://github.com/tempoxyz/tempo/pull/585/commits/f79fea9364df8ffb132014a0e243f94753d094fd) and [47c54b8](https://github.com/tempoxyz/tempo/pull/585/commits/47c54b82e96d9612bc2cd701888fed6b26aeedce)

crates/commonware-node/src/args.rs
```diff
@@ -0,0 +1,100 @@
+//! Command line arguments for configuring the consensus layer of a tempo node.
+use std::{net::SocketAddr, path::PathBuf};
+
+/// Arguments for configuring the consensus layer of a tempo node.
+/// Command line arguments for configuring the consensus layer of a tempo node.
+#[derive(Debug, Clone, PartialEq, Eq, clap::Args)]
+pub struct ConsensusArgs {
+    /// The file containing the ed25519 signing key for p2p communication.
+    #[arg(long = "consensus-signing-key")]
+    pub signing_key: PathBuf,
```
klkvr: should we follow reth cli arg naming conversions here? i.e `--consensus.signing-key` instead?

crates/commonware-node/src/lib.rs
```diff
@@ -63,33 +87,53 @@ pub async fn run_consensus_stack(
     );
     let subblocks = network.register(SUBBLOCKS_CHANNEL_IDENT, SUBBLOCKS_LIMIT, message_backlog);
 
+    let fee_recipient = config.fee_recipient.unwrap_or_else(|| {
+        // FIXME: outside of root span; put in root span.
+        info!("argument fee-recipient not set; using coinbase set in genesis");
+        execution_node.chain_spec().genesis().coinbase
+    });
```
klkvr: we want a hard error here, right now if validator forgets to set recipient their fees will be sent to genesis coinbase (which is likely a dummy address)

crates/commonware-node/src/args.rs
klkvr: I also think this breaks `tempo --follow` API because this is no longer optional https://github.com/tempoxyz/tempo/blob/8ab3810bfecbcbbe6454c7443d47624d4d09782b/bin/tempo/src/main.rs#L55-L56

bin/tempo/src/main.rs
```diff
@@ -96,55 +93,44 @@ fn main() -> eyre::Result<()> {
...
+                .with_worker_threads(args.consensus.worker_threads)
+                .with_storage_directory(
+                    &node
+                        .config
+                        .datadir
+                        .datadir
+                        .unwrap_or_default()
+                        .as_ref()
+                        .join("consensus"),
+                )
```
klkvr: this should properly infer reth datadir resolution logic, which is basically `config.datadir.resolve_datadir(chain_spec.chain())`

https://github.com/paradigmxyz/reth/blob/6fa11ff2f88c1135622660fe25dbfa8dbccb45a4/crates/cli/commands/src/download.rs#L134

right now if `--datadir` is not set explicitly (which is often the case) it will write to something weird

also I think we want an `info!` log with consensus datadir, like
https://github.com/paradigmxyz/reth/blob/6fa11ff2f88c1135622660fe25dbfa8dbccb45a4/crates/cli/commands/src/node.rs#L198

bin/tempo/src/main.rs
```diff
@@ -96,55 +93,44 @@ fn main() -> eyre::Result<()> {
...
-                        )
-                    }
-                    None => Either::Right(pending()),
-                }.fuse();
-                let consensus_stack = run_consensus_stack(&ctx, &consensus_config, node);
+                let mut metrics_server = tempo_commonware_node::metrics::install(
+                    ctx.with_label("metrics"),
+                    args.consensus.metrics_address,
+                )
+                .fuse();
```
klkvr, line 141: I guess it's not worth the effort trying to merge this with reth's prometheus server

crates/commonware-node/src/dkg/ceremony/mod.rs
```diff
@@ -212,38 +214,47 @@ where
...
             }
-        } else if let Some(share) = config.share.clone() {
-            // XXX: It is critical to write the initial ceremony state with the
-            // correct number of players to disk. The upsert operations in the
-            // rest of the ceremony write the default value for the number of
-            // players, with is 0.
+        } else {
+            info!("starting a fresh ceremony");
+
+            if let Some(share) = config.share.clone() {
```
klkvr, line 220: nitpick: I find naming this `config` a bit confusing because I usually associate config with global values passed to the binary. here it's even tricker because there's a `share` arg in our global config and this line always takes me a second to make sense of :D

could we maybe name this `opts`/`args` so that it's easier to reason about?

crates/commonware-node/src/dkg/manager/ingress.rs
```diff
@@ -95,21 +90,22 @@ pub(super) struct GetIntermediateDealing {
...
+    pub(super) response: oneshot::Sender<PublicOutcome>,
 }
 
 impl Reporter for Mailbox {
     type Activity = Update<Block>;
 
     async fn report(&mut self, update: Self::Activity) {
+        let Self::Activity::Block(block) = update else {
+            return;
+        };
```
klkvr:
```suggestion
        let Self::Activity::Block(block) = update else {
            // Skip tip updates as DKG manager is only interested in full blocks
            return;
        };
```

crates/commonware-node/src/dkg/manager/mod.rs
```diff
@@ -1,33 +1,50 @@
...
+pub(crate) async fn init<TContext, TPeerManager>(
+    context: TContext,
+    config: Config<TPeerManager>,
+) -> eyre::Result<(Actor<TContext, TPeerManager>, Mailbox)>
 where
     TContext: Clock + CryptoRngCore + Metrics + Spawner + Storage,
+    TPeerManager: commonware_p2p::Manager<
+            PublicKey = PublicKey,
+            Peers = OrderedAssociated<PublicKey, SocketAddr>,
+        >,
```
klkvr, line 36: would be cool to make this dyn-compatible

crates/commonware-node/src/lib.rs
```diff
@@ -135,48 +182,17 @@ async fn instantiate_network(
...
-        })
-        .collect::<FuturesOrdered<_>>();
-    let resolved = resolve_all
-        .try_collect::<Vec<(_, _)>>()
-        .await
-        .wrap_err("failed resolving at least one peer")?;
-    Ok(resolved.into())
+fn read_from_file<T: DecodeExt<X>, X: IsUnit, P: AsRef<Path>>(path: P) -> eyre::Result<T> {
+    let raw_bytes = std::fs::read(path).wrap_err("failed reading file")?;
+    let decoded = const_hex::decode(&raw_bytes).wrap_err("failed decoding file contents as hex")?;
```
klkvr: `const_hex` dep should not be needed, it's available as `alloy_primitives::hex`

crates/consensus-args/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "tempo-consensus-args"
```
klkvr: why do we need an entire crate for a single struct? `tempo-commonware-node` seems like a right place for it?

crates/e2e/src/execution_runtime.rs
```diff
@@ -285,9 +411,158 @@ pub async fn launch_execution_node<P: AsRef<Path>>(
...
+    fn from(value: AddValidator) -> Self {
+        Self::AddValidator(value.into())
+    }
+}
+
+impl From<SpawnNode> for Message {
+    fn from(value: SpawnNode) -> Self {
+        Self::SpawnNode(value.into())
+    }
+}
```
klkvr, line 504: I don't really like those `From` impls because they make it harder to understand where the messages are actually sent when you go to references of a message variant

crates/e2e/src/execution_runtime.rs
```diff
@@ -285,9 +411,158 @@ pub async fn launch_execution_node<P: AsRef<Path>>(
...
+                        admin(),
+                        IValidatorConfig::addValidatorCall {
+                            newValidatorAddress: validator(i as u32),
+                            publicKey: peer.encode().freeze().as_ref().try_into().unwrap(),
+                            active: true,
+                            inboundAddress: addr.to_string(),
+                            outboundAddress: addr.to_string(),
+                        },
+                    )
+                    .unwrap();
```
klkvr, line 597: could we somehow reuse `generate-genesis` logic here?

crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -1,55 +1,83 @@
...
+    TPeerManager: commonware_p2p::Manager,
 {
-    config: super::Config,
+    config: super::Config<TPeerManager>,
     context: ContextCell<TContext>,
     mailbox: mpsc::UnboundedReceiver<super::Message>,
 
     ceremony_metadata: Arc<Mutex<Metadata<ContextCell<TContext>, U64, CeremonyState>>>,
     epoch_metadata: Metadata<ContextCell<TContext>, U64, EpochState>,
+    registered_peers_metadata: Metadata<ContextCell<TContext>, U64, PeersRegistered>,
```
klkvr: let's add docs for all fields here, especially the metadata ones, so that it's obvious what we store on disk and for which epochs

crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -99,21 +144,138 @@ where
...
+        for epoch in epoch_state.epoch.saturating_sub(2)..epoch_state.epoch {
+            if let Some::<PeersRegistered>(peers) =
+                registered_peers_metadata.get(&epoch.into()).cloned()
+            {
+                info!(epoch, ?peers, "peers recovered; registering");
+                config.peer_manager.update(epoch, peers.into_inner()).await;
+            } else {
+                info!(epoch, "no peers found on disk");
+            }
+        }
```
klkvr: we have 2 dimensions of peers here so maybe it is worth introducing an additional term here like a `peer_set`/`epoch_peer_set`

for me it would be easier to understand if `all_participants` were `current_epoch_peer_set` and `registered_peers_metadata` was `peer_sets`/`epoch_to_peer_set`

crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -250,147 +398,107 @@ where
...
-                    .epoch_manager
-                    .report(
-                        epoch::Exit {
-                            epoch: previous_epoch,
-                        }
-                        .into(),
-                    )
-                    .await;
-            }
+            self.report_epoch_entered().await;
```
klkvr: maybe `utils::is_first_block_in_epoch` for consistency?

crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -408,89 +516,295 @@ where
...
+        TReceiver: Receiver<PublicKey = PublicKey>,
+        TSender: Sender<PublicKey = PublicKey>,
+    {
+        let syncers = read_validator_config_with_retry(
+            &self.context,
+            &self.config.execution_node,
+            self.epoch_state.epoch,
+            self.config.epoch_length,
+        )
+        .await;
```
klkvr: could we call those `next_players`?

crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -408,89 +516,295 @@ where
...
+            .syncers
+            .set(self.all_participants.syncers().len() as i64);
+        self.metrics
+            .how_often_dealer
+            .inc_by(ceremony.is_dealer() as u64);
+        self.metrics
+            .how_often_player
+            .inc_by(ceremony.is_player() as u64);
+
+        let peers_to_register = self.all_participants.construct_peers_to_register();
```
klkvr: this is semantically just `self.all_participants.peers()`? i.e we just return addresses of all peers in all sets

klkvr (changes requested):
some notes and naming nitpicks

didn't look very closely at xtask code yet as I didn't write most of it, would appreciate @0xKitsune or @Zygimantass review there

I'm wondering if a path to hardfork-gating this could be
1. Remove the assumption of inital DKG outcome to be stored in genesis by adding it to e.g chainspec. Same for validator config initial state
2. Before hardfork boundary, assume that `next_players == current_dealers`
3. After hardfork, start loading next players from the contract

crates/commonware-node/src/dkg/manager/actor.rs
SuperFluffy: It'd be a bit confusing because here we'd call them `next_players`, but the next line right below would call be `players: some_other_players`. I like `syncers` because it becomes clear what they role is - maybe `syncing_players`?

bin/tempo/src/main.rs
SuperFluffy: I don't think it's possible - reth is using `metrics`, but commonware is using `prometheus-client` and further has a bunch of abstractions on top of it.

I brought it up with them, but it's not a high priority.

bin/tempo/src/main.rs
SuperFluffy: TIL, thank you for catching this. Done both in 10f121be

crates/commonware-node/src/dkg/manager/actor.rs
klkvr: yeah `syncing_players`/`pending_players` makes sense to me

crates/commonware-node/src/dkg/manager/actor.rs
SuperFluffy: Added doc comments on every field in  eac35792

While doing that, I believe I found two potential bugs - see the TODOs in  eac35792. I will commit fixes after going through all review comments.

crates/commonware-node/src/dkg/manager/actor.rs
SuperFluffy: The more I think about this, the more this logic is problematic. I wanted to deal with this in a followup, https://github.com/tempoxyz/tempo/issues/933, but will rework this logic now.

crates/commonware-node/src/dkg/manager/actor.rs
SuperFluffy: Commonware does not provide that - it's a bit overkill to have this for just one call, but I added this utility to `crates::epoch` in 4b0adb12

crates/commonware-node/src/dkg/manager/actor.rs
SuperFluffy: Done in 9051fc10

crates/commonware-node/src/dkg/manager/actor.rs
SuperFluffy: Yes indeed. Are you asking to rename it? `peers` sounds so innocent, which is why I went with `construct_peers`. Maybe I should go back to `construct_merged_peerset`?

SuperFluffy: Yes indeed. Are you asking to rename it? `peers` sounds so innocent, which is why I went with `construct_peers`. Maybe I should go back to `construct_merged_peerset`?

crates/commonware-node/src/dkg/manager/ingress.rs
SuperFluffy: One-upped this by emitting `trace!` events: ad541fac

crates/commonware-node/src/dkg/manager/mod.rs
SuperFluffy: Let's ask commonware :)

crates/commonware-node/src/dkg/manager/actor.rs
klkvr: should we rename `syncers` -> `syncing_players` everywhere else too?

crates/commonware-node/src/lib.rs
SuperFluffy: Cool, good to know! ed91a7c4

crates/commonware-node/src/dkg/manager/actor.rs
klkvr: maybe just `merged_peers`? `construct` implies that there's a more complex logic logic going on while it's just a `flat_map` essentially

crates/commonware-node/src/lib.rs
SuperFluffy: Feels odd using re-exports from an unrelated crate (not doing anything related to alloy here), but did the change here: 9c3d54ad

crates/consensus-args/Cargo.toml
SuperFluffy: Yeah, that indeed seems unnecessary. Just took the `-config` crate and migrated it over. 3142e891

Broken compile: 32110c69

crates/commonware-node/src/args.rs
SuperFluffy: Yeah, good point. I added `required_unless_present_any` to both `--consensus.signing-key` and `--consensus.fee-recipient`: d91279dc

crates/commonware-node/src/dkg/ceremony/mod.rs
SuperFluffy: I have similar qualms about `opts` and `args`. 🙈 They scream `cli args` or `cli options` at me (and well, most fields are not even optional and we are literally "configuring" the ceremony here). Because we make heavy use of `configuration` (also through commonware) here, I think I'd prefer keeping this.

I was using `builder` before, but y'all didn't like that one either - although I would actually like to go back to that. Doing a full builder pattern is overkill, but at the same time it's technically what it is. This seems nice:

```rust
ceremony::Builder {
    share,
    polynomial,
    epoch,
}.init(context)
.await?;
```

crates/commonware-node/src/dkg/manager/actor.rs
SuperFluffy: We should! 74d8c59a

crates/commonware-node/src/dkg/manager/actor.rs
SuperFluffy: More logic is indeed happening: we are also looking up the hostnames, so `construct` seems too innocent. `lookup_and_merge_peers`?

crates/e2e/src/execution_runtime.rs
SuperFluffy: Yeah, that'd be nice. I would like to leave breaking that out of xtask as a follow if possible.

crates/commonware-node/src/dkg/ceremony/mod.rs
klkvr: maybe `params`? :D

I don't like builder because it is quite verbose, requires an extra struct per type and introduces indirection when you're trying to find where fields are actually set

crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -408,89 +549,295 @@ where
...
+        TReceiver: Receiver<PublicKey = PublicKey>,
+        TSender: Sender<PublicKey = PublicKey>,
+    {
+        let syncing_players = read_validator_config_with_retry(
+            &self.context,
+            &self.config.execution_node,
+            self.epoch_state.epoch,
+            self.config.epoch_length,
+        )
+        .await;
```
klkvr: does this effectively require all validators to run an archive node?

thinking about a situation like this:
1. you run a full node and trigger backfill based on `::Tip` update from marshal
2. by the time the node is synced, it prunes state for older blocks (below tip-10064)
3. DKG actor is not yet finished, attempts to fetch validator config from a pruned block and ends up stuck

SuperFluffy: For completeness: I do think that that is currently the case.

Commonware has two inflight PRs that will help our nodes catch up:

https://github.com/commonwarexyz/monorepo/issues/1706
https://github.com/commonwarexyz/monorepo/issues/2032

crates/commonware-node/src/dkg/manager/actor.rs
SuperFluffy: I have moved this around into a new type `ValidatorState`, calling this method `resolve_addresses_and_merge_peers` [2852340](https://github.com/tempoxyz/tempo/pull/585/commits/2852340bceb6bce84ab4e483f559b2fbff14a63c)

crates/commonware-node/src/dkg/manager/actor.rs
SuperFluffy: I have reworked all of this logic because it was confusing me as well:

We no longer have `epoch_state`, `all_participants`. Instead, we only have:

1. `epoch_metadata` at key `CURRENT_EPOCH_KEY`, which tracks the DKG outcome that will be used in epoch `i` and set of validators that will be registered on eopch `i` (with full information); this value is set on the last height of epoch `i-1`
2. `epoch_metadata` at key `PREVIOUS_EPOCH_KEY`, which tracks the previous DKG outcome for the brief time-period in between starting the epoch `i` on the last height of `i-1`, and fully entering `i` on the first height of `i`.
3. `validators_metadata`, which tracks all validators, but is only used to register validators for epochs `i-1` and `i-2` on restart

[2852340](https://github.com/tempoxyz/tempo/pull/585/commits/2852340bceb6bce84ab4e483f559b2fbff14a63c)

crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -406,87 +521,479 @@ where
...
+                .expect("must always be able to persist data");
+        }
+    }
+
+    #[instrument(skip_all)]
+    async fn update_current_epoch_state(&mut self) {
+        let old_epoch_state = self
+            .epoch_metadata
+            .remove(&CURRENT_EPOCH_KEY.into())
+            .expect("there must always exist an epoch state");
```
klkvr, line 786: so this basically just changes the `CURRENT_EPOCH_KEY` to contain value for newly entered epoch

should we also register the new peers right here?

crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -406,87 +521,479 @@ where
...
+            "registered p2p peers by merging dealers, players, syncing players",
+        );
+    }
+
+    /// Reports that a new epoch can be entered.
+    ///
+    /// This should trigger the epoch manager to start a new consensus engine
+    /// backing the epoch stored by the DKG manager.
+    #[instrument(skip_all, fields(epoch = self.current_epoch_state().epoch()))]
+    async fn report_current_epoch_state(&mut self) {
```
klkvr: `enter_epoch_and_register_peers`?

crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -406,87 +521,479 @@ where
...
+        );
+        self.register_validators(epoch_state.epoch(), epoch_state.validator_state)
+            .await;
+    }
+
+    /// Reports that the previous epoch should be entered.
+    ///
+    /// This method is called on startup to ensure that a consensus engine for
+    /// the previous epoch i-1 is started in case the node went down before the
+    /// new epoch i was firmly locked in.
```
klkvr, line 714: hmm what does this mean? it seems that it's only possible for `CURRENT_EPOCH_KEY` to be updated if last height of previous epoch was finalized which means we don't need the engine for previous epoch anymore?

SuperFluffy: The flow is like this, boundary `B` is the last height of epoch `i-1`:

+ on `B-1`, we finalize the DKG ceremony and store in on disk at `DKG_OUTCOME_KEY`.
+ on `B`, we read the new syncing validators from the smart contract, take the data in `DKG_OUTCOME_KEY`, and with both write a new state to `CURRENT_EPOCH_KEY`. The previous value we store in `PREVIOUS_EPOCH_KEY`.
+ also on `B`, we report this new value and have the epoch manager spin up a new engine.
+ only on `B+1` (first height of epoch `i`) do we report that epoch `i-1` can be closed.

So `PREVIOUS_EPOCH_KEY` ensures that if we go down right after `B` but before `B+1`, that we still start the engine.

This is an edge case to ensure that all validators get to see the finalization of `B` (not the one that went down, but its peers).

crates/commonware-node/src/dkg/manager/actor.rs
SuperFluffy: Yeah, makes sense: cfa70e9b

crates/commonware-node/src/dkg/manager/actor.rs
SuperFluffy: `enter_current_epoch_state_and_register_peers`: cfa70e9b

crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -406,87 +521,479 @@ where
...
+
+        self.epoch_metadata
+            .sync()
+            .await
+            .expect("must always be able to persists epoch state");
+
+        self.validators_metadata
+            .put_sync(epoch.into(), new_validator_state)
+            .await
+            .expect("must always be able to persist validator metadata");
```
joshieDo: what happens on restart, if the node shutsdown inbetween these two?

SuperFluffy: If the node shuts down in between both syncs there are two safeguards that cover this (exhaustively?):

1. First, and most importantly, the validator state is now moved into the epoch state for the current epoch. For the current epoch, we always register the peer set from the validators stored in this atomically updated cell. The validators written to validator-specific metadate cell are just a nice to have for older peers to catch up, but there is no effect on the core functionality of the actor.
2. Second, even if the above weren't in place: if the node shuts down without handling to completion the finalized block that triggered these syncs, then the finalization will not be acknowledged. And if the block is not acknowledged, then on restart the marshal actor will resend the same finalized block, and these functions will be entered again.

SuperFluffy: Because of your comment I took a closer look at this logic and I found some issues:

1. I had a regression in the code where I removed a check for blocks matching the current epoch. I returned that in 671367a to ensure that last finalized block does not get replayed and the already-written `CURRENT_EPOCH_KEY` not updated again.
2. But in addition I noticed that if we go down before we get to write the validator set, that we will forever lose it. I have moved that logic around to make sure that the validator set always gets written: e4abb91

PR #589, opened 2025-10-30
chore: bump commonware

klkvr:
bumps commonware to latest main commit to make changes from https://github.com/commonwarexyz/monorepo/pull/2053 available

### DIFF
--- a/crates/commonware-node/src/alias.rs
+++ b/crates/commonware-node/src/alias.rs
@@ -1,18 +1,13 @@
 //! A collection of aliases for frequently used (primarily commonware) types.
 
-use commonware_consensus::simplex::signing_scheme::bls12381_threshold::Scheme;
-use commonware_cryptography::{bls12381::primitives::variant::MinSig, ed25519::PublicKey};
-
 pub(crate) mod marshal {
-    use super::*;
-    use commonware_consensus::marshal;
+    use commonware_consensus::{marshal, simplex::signing_scheme::bls12381_threshold::Scheme};
+    use commonware_cryptography::{bls12381::primitives::variant::MinSig, ed25519::PublicKey};
 
     use crate::consensus::block::Block;
 
     pub(crate) type Actor<TContext> =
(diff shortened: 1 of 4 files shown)


crates/commonware-node/src/alias.rs
```diff
@@ -1,13 +1,18 @@
...
 
     pub(crate) type Actor<TContext> =
-        marshal::Actor<TContext, Block, crate::epoch::SchemeProvider, Scheme<MinSig>>;
+        marshal::Actor<TContext, Block, crate::epoch::SchemeProvider, ThresholdScheme>;
 
-    pub(crate) type Mailbox = marshal::Mailbox<Scheme<MinSig>, Block>;
+    pub(crate) type Mailbox = marshal::Mailbox<ThresholdScheme, Block>;
 }
+
+pub(crate) type ThresholdScheme = Scheme<PublicKey, MinSig>;
```
SuperFluffy: I think I don't like this. I had a lot of these aliases in an earlier version and decided that I am more confused by them than anything.

Right now we are using `MinSig` in a few other places (like `Public<MinSig>` for the public polynomial). If we keep `Scheme<PublicKey, MinSig>`, then it's evident that these are related.

If instead you use `ThresholdScheme`, that distinction is hidden.

Can I suggest this as an alternative?

```suggestion
pub(crate) type Scheme<T> = Scheme<PublicKey, T>;
```

crates/commonware-node/src/epoch/manager/actor.rs
```diff
@@ -234,22 +233,21 @@ where
...
         );
 
+        self.metrics
+            .latest_participants
+            .set(participants.len() as i64);
+
         // Register the new signing scheme with the scheme provider.
         let scheme = if let Some(share) = share {
-            Scheme::new(participants.as_ref(), &public, share)
+            Scheme::new(participants, &public, share)
```
SuperFluffy, line 241: Related to my comment above re `ThresholdScheme` - here you are using `Scheme::new`, but this ends up becoming a `ThresholdScheme`. If you do the alias, then IMO this should become `ThresholdScheme::new`

PR #592, opened 2025-10-30
feat: support 2d nonces in EVM

klkvr:
Changes in the PR:
- `nonceKey` is changed to `U256`
- AA transactions with non-zero `nonceKey` are now rejected at the txpool level instead of returning an error during decoding.
- Logic for bumping 2D nonce is added to the EVM.
- AA transactions nonce is now always bumped before execution to fix a bug when reverting CREATE transaction did not bump a nonce. A test is added for this edge case

### DIFF
--- a/crates/primitives/src/transaction/account_abstraction.rs
+++ b/crates/primitives/src/transaction/account_abstraction.rs
@@ -1002,7 +1002,6 @@ mod tests {
             calls: vec![dummy_call],
             ..Default::default()
         };
-        assert!(tx2.validate().is_err());
         assert_eq!(tx2.nonce(), 0);
     }
 


crates/transaction-pool/src/validator.rs
```diff
@@ -31,13 +32,25 @@ where
...
+        // Reject system transactions, those are never allowed in the pool.
         if transaction.inner().is_system_tx() {
             return TransactionValidationOutcome::Error(
                 *transaction.hash(),
                 InvalidTransactionError::TxTypeNotSupported.into(),
             );
         }
 
+        // Reject AA transactions with non-zero nonce keys, we don't have
+        // a nice way to track them in the pool yet.
```
mattsse, line 44: mental note here, this should be removed once we have #569 ?

crates/revm/src/handler.rs
```diff
@@ -371,19 +372,78 @@ where
...
+                .increment_nonce(&tx.caller(), nonce_key)
+                .map_err(|err| match err {
+                    TempoPrecompileError::Fatal(err) => EVMError::Custom(err),
+                    err => TempoInvalidTransaction::NonceManagerError(err.to_string()).into(),
+                })?;
+        } else {
+            // Bump the nonce for calls. Nonce for CREATE will be bumped in `make_create_frame`.
+            //
+            // Always bump nonce for AA transactions.
+            if tx.aa_tx_env.is_some() || tx.kind().is_call() {
```
legion2002, line 561: Why do we always bump nonce for AA tx type here?
More accurate would be to bump if all calls in the batch, are of type call

crates/revm/src/handler.rs
```diff
@@ -371,19 +372,78 @@ where
...
+                            },
+                        )
+                        .into());
+                    }
+                    _ => {}
+                }
+            }
+
+            // Always increment nonce for AA transactions with non-zero nonce keys.
+            nonce_manager
```
legion2002, line 551: should we also disallow any CREATE calls in the batch here?

crates/revm/src/handler.rs
klkvr: we can't rely on CREATE item to bump the nonce because right now if that CREATE reverts we don't retain the nonce bump in the journal

this is the case on main right now and I don't see a nice way to fix this

crates/revm/src/handler.rs
klkvr: why would we want to disallow them specifically for non-zero keys?

crates/revm/src/handler.rs
legion2002: but doesn't this mean that for successful CREATE, the nonce is bumped twice? 
So if an AA tx has just a single CREATE, then nonce is bumped by 2.

But ig it doesn't that matter, it is the same behavior as an EVM CREATE.

klkvr: > So if an AA tx has just a single CREATE, then nonce is bumped by 2.

yep

> But ig it doesn't that matter, it is the same behavior as an EVM CREATE.

it is not, if you send a eip1559 tx that does a CREATE, nonce will only be numped once

legion2002: yeah by EVM create I meant, if you have a delegated account, and that account uses the CREATE opcode.

klkvr: I guess we could special-case this — i.e if AA tx's first frame is CREATE and it is using protocol nonce we will rely on CREATE frame to bump it, and then if it fails we will still forcefully bump it

though this is a bit inconsistent, because for non-zero keys we'd still bump nonce 2 times — once for the key and once for the protocol nonce

crates/revm/src/handler.rs
legion2002: Oh yeah nvm

crates/revm/src/handler.rs
legion2002: Yeah I think it's fine.
Don't think any major applications assume the sequentiality of protocol nonce, if someone wants complete control over the nonce, they would just use CREATE2

PR #593, opened 2025-10-30
fix(consensus): ensure boundary certs are gossiped for non-committee members

SuperFluffy:
This patch follows https://github.com/commonwarexyz/monorepo/pull/2022 by inverting the relationship between receiving votes and sending out finalization certificates for boundary heights (the final heights of an epoch):

1. the epoch manager now listens for votes in newer epochs
2. if it finds out about it being on an older epoch this way, it requests the finalization certificate for the boundary height of its old epoch from the voter
3. the voter sends this certificate to the running simplex engine backing the old epoch of the node
4. this allows the node to move forward to the end of the epoch and transition to the new epoch.

This process is repeated until the node catches up.

### DIFF
--- a/crates/commonware-node/src/epoch/manager/actor.rs
+++ b/crates/commonware-node/src/epoch/manager/actor.rs
@@ -3,9 +3,9 @@
 //! This actor is responsible for:
 //!
 //! 1. entering and exiting epochs given messages it receives from the DKG
-//! manager.
+//!    manager.
 //! 2. catching the node up by listening to votes for unknown epoch and
-//! requesting finalizations for the currently known boundary height.
+//!    requesting finalizations for the currently known boundary height.
 //!
 //! # Entering and exiting epochs
 //!
--- a/crates/commonware-node/src/consensus/engine.rs
+++ b/crates/commonware-node/src/consensus/engine.rs
@@ -296,6 +296,10 @@ where
         + Spawner
         + Storage,
 {
+    #[expect(
+        clippy::too_many_arguments,
+        reason = "following commonware's style of writing"
+    )]
     pub fn start(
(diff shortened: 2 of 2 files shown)


crates/commonware-node/src/epoch/manager/actor.rs
```diff
@@ -319,55 +399,101 @@ where
...
+            self.config
+                .marshal
+                .get_finalization(boundary_height)
+                .await
+                .is_none(),
+            "finalization certificate for epoch `{our_epoch}` at boundary \
+            height `{boundary_height}` is already known; no action necessary",
         );
 
         // Forward the finalization to the sender. This operation is best-effort.
```
klkvr: should probably update that comment

crates/commonware-node/src/epoch/manager/actor.rs
```diff
@@ -319,55 +399,101 @@ where
...
     /// commonware p2p muxer will send the message to the backup channel, tagged
-    /// with unknown epoch.
-    #[instrument(skip_all, fields(msg.epoch = epoch, msg.from = %from), err(level = Level::WARN))]
+    /// with the unknown epoch.
+    #[instrument(skip_all, fields(msg.epoch = their_epoch, msg.from = %from), err(level = Level::INFO))]
     async fn handle_msg_for_unregistered_epoch(
         &mut self,
-        recovered_global_sender: &mut GlobalSender<impl Sender<PublicKey = PublicKey>>,
-        epoch: Epoch,
+        // recovered_global_sender: &mut GlobalSender<impl Sender<PublicKey = PublicKey>>,
```
klkvr:
```suggestion
```

PR #598, opened 2025-10-30
!feat: add a keychain natively for each account

legion2002:
This allows users to temporarily issue access keys to DApps, making the UX much better.
This also means that we can remove the IthacaAccount bytecode completely from the chain, hence remove the DAA mechanism.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -9,6 +9,9 @@ on:
   pull_request:
   workflow_dispatch:
   workflow_call:
+  schedule:
+    - cron: "5 9 * * *"
+    - cron: "30 20 * **"
 
 env:
   REGISTRY: ghcr.io/tempoxyz
@@ -48,6 +51,7 @@ jobs:
         with:
           images: ${{ env.REGISTRY }}/${{ matrix.binary.name }}
           tags: |
+            type=schedule,pattern=nightly
             type=semver,pattern={{version}}
             type=semver,pattern={{major}}.{{minor}}
             type=semver,pattern={{major}}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11390,6 +11390,7 @@ dependencies = [
  "const-hex",
  "humantime-serde",
  "indexmap 2.12.1",
+ "reth-consensus-common",
  "serde",
  "thiserror 2.0.17",
  "toml",
--- a/crates/commonware-node-config/Cargo.toml
+++ b/crates/commonware-node-config/Cargo.toml
@@ -17,8 +17,9 @@ commonware-codec.workspace = true
 commonware-cryptography.workspace = true
 commonware-utils.workspace = true
 const-hex = { workspace = true, features = ["serde"] }
+humantime-serde.workspace = true
 indexmap = { workspace = true, features = ["serde"] }
+reth-consensus-common.workspace = true
 serde = { workspace = true, features = ["derive"] }
 thiserror.workspace = true
 toml.workspace = true
-humantime-serde.workspace = true
--- a/crates/commonware-node-config/src/p2p.rs
+++ b/crates/commonware-node-config/src/p2p.rs
@@ -1,5 +1,12 @@
 //! Config options specific to the p2p layer.
 
+use reth_consensus_common::validation::MAX_RLP_BLOCK_SIZE;
+
+/// Set the maximum permitted message size to Reth's maximum RLP-encoded block size.
+///
+/// It is enforced on the builder side in `crates/payload/builder/src/lib.rs`.
+const DEFAULT_MAX_MESSAGE_SIZE_BYTES: usize = MAX_RLP_BLOCK_SIZE;
+
 #[derive(Clone, Debug, serde::Deserialize, serde::Serialize)]
 pub struct Config {
     /// The maximum permitted message size in bytes for messages sent over the
@@ -10,7 +17,19 @@ pub struct Config {
 impl Default for Config {
     fn default() -> Self {
         Self {
-            max_message_size_bytes: 1024 * 1024,
+            max_message_size_bytes: DEFAULT_MAX_MESSAGE_SIZE_BYTES,
         }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// Dummy test to catch any upstream Reth changes of max RLP block size.
+    #[test]
+    fn test_default_config() {
+        let config = Config::default();
+        assert_eq!(config.max_message_size_bytes, 8_388_608);
     }
 }
--- a/crates/commonware-node/src/subblocks.rs
+++ b/crates/commonware-node/src/subblocks.rs
@@ -45,6 +45,12 @@ use tempo_primitives::{
 use tokio::sync::broadcast;
 use tracing::{Instrument, Level, Span, debug, instrument, warn};
 
+/// Maximum number of stored subblock transactions. Used to prevent DOS attacks.
+///
+/// NOTE: included txs are organically cleared when building the next subblock
+/// because they become invalid once their nonce is used.
+const MAX_SUBBLOCK_TXS: usize = 100_000;
+
 pub(crate) struct Config<TContext> {
     pub(crate) context: TContext,
     pub(crate) signer: PrivateKey,
@@ -202,9 +208,11 @@ impl<TContext: Spawner + Metrics + Pacer> Actor<TContext> {
         {
             return;
         }
-        self.subblock_transactions
-            .lock()
-            .insert(*transaction.tx_hash(), Arc::new(transaction));
+        let mut txs = self.subblock_transactions.lock();
+        if txs.len() >= MAX_SUBBLOCK_TXS {
+            return;
+        }
+        txs.insert(*transaction.tx_hash(), Arc::new(transaction));
     }
 
     /// Tracking of the current sconsensus state by listening to notarizations and nullifications.


crates/revm/src/handler.rs
```diff
@@ -398,9 +402,187 @@ where
...
 
-        // Create storage provider wrapper around journal
+        // If the transaction includes a KeyAuthorization, validate and authorize the key
+        if let Some(aa_tx_env) = tx.aa_tx_env.as_ref() {
+            if let Some(key_auth) = &aa_tx_env.key_authorization {
+                // Validate that the KeyAuthorization is signed by the root account
+                let root_account = &tx.caller;
+
+                // Compute the message hash for the KeyAuthorization
+                // Message format: keccak256(key_type || key_id || expiry || limits)
```
jxom: It's a bit odd that this isn't in same format as the RLP version

crates/primitives/src/transaction/aa_signature.rs
```diff
@@ -42,6 +43,25 @@ pub struct WebAuthnSignature {
...
+/// The handler validates that user_address has authorized the access key in the KeyChain precompile.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
+#[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
+pub struct KeychainSignature {
+    /// Root account address that this transaction is being executed for
+    pub user_address: Address,
+    /// The actual signature from the access key (can be Secp256k1, P256, or WebAuthn, but NOT another Keychain)
+    pub signature: Box<AASignature>,
```
jxom: we can't avoid this in any way? nesting may get a bit wild (3 `signature`s)

```
{
  ...,
  signature: {
    userAddress: '0x...',
    signature: {
      type: 'secp256k1,
      signature: {
        ...
      }
    }
  }
}
```

jxom: why don't we add `userAddress` to `AASignature` itself?

crates/revm/src/handler.rs
legion2002: Yeah I can make this RLP, good point.

crates/primitives/src/transaction/aa_signature.rs
legion2002: Yeah nesting was causing issues in roundtrip tests.
I have added a new PrimitiveSignature type now for this, which does not have the keychain.

crates/contracts/src/precompiles/account_keychain.rs
```diff
@@ -0,0 +1,148 @@
...
+        uint256 amount;
+    }
+
+    /// Key information structure
+    #[derive(Debug, PartialEq, Eq)]
+    struct KeyInfo {
+        SignatureType signatureType;
+        address keyId;
+        uint64 expiry;
+    }
```
jxom: Should this be inside the `interface`? (I use rely on this for codegen)

legion2002: Oh yes, will move.

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -0,0 +1,318 @@
...
+    pub signature_type: u8, // 0: secp256k1, 1: P256, 2: WebAuthn
+    pub expiry: u64,        // Block timestamp when key expires
+    pub is_active: bool,    // Whether key is active
+}
+
+/// Account Keychain contract for managing authorized keys
+#[contract]
+pub struct AccountKeychain {
+    // keys[account][keyId] -> AuthorizedKey
+    keys: Mapping<Address, Mapping<Address, AuthorizedKey>>,
```
klkvr, line 55: what's the motivation for using `address` as keyId?

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -0,0 +1,318 @@
...
+            return Err(AccountKeychainError::key_already_exists().into());
+        }
+
+        // Convert SignatureType enum to u8 for storage
+        let signature_type = match call.signatureType {
+            SignatureType::Secp256k1 => 0,
+            SignatureType::P256 => 1,
+            SignatureType::WebAuthn => 2,
+            _ => return Err(AccountKeychainError::invalid_signature_type().into()),
+        };
```
mattsse, line 148: I was considering making this a tryfrom u8 impl, but I think being explicit here might actually be better

crates/precompiles/src/tip20/mod.rs
```diff
@@ -515,6 +516,30 @@ impl<'a, S: PrecompileStorageProvider> TIP20Token<'a, S> {
...
+                let old_allowance = self.get_allowance(msg_sender, call.spender)?;
+
+                // Calculate the increase in approval (only deduct if increasing)
+                // If old approval is 100 and new approval is 120, deduct 20 from spending limit
+                // If old approval is 100 and new approval is 80, deduct 0 (decreasing approval is free)
+                let approval_increase = if call.amount > old_allowance {
+                    call.amount - old_allowance
+                } else {
+                    U256::ZERO
+                };
```
mattsse: this is effecitvely just `.saturating_sub`, but we can also keep this as is

crates/primitives/src/transaction/account_abstraction.rs
```diff
@@ -106,7 +109,264 @@ impl Decodable for Call {
...
+        key_id: Address,
+        expiry: u64,
+        limits: &[TokenLimit],
+    ) -> B256 {
+        let mut auth_message = Vec::new();
+        let key_type_byte: u8 = match key_type {
+            SignatureType::Secp256k1 => 0,
+            SignatureType::P256 => 1,
+            SignatureType::WebAuthn => 2,
+        };
```
mattsse: we can make this a `From<SignatureType> for u8`

crates/primitives/src/transaction/account_abstraction.rs
```diff
@@ -106,7 +109,264 @@ impl Decodable for Call {
...
+
+        // Convert limits slice to Vec for encoding
+        let limits_vec = limits.to_vec();
```
mattsse: unsure if we need this alloc,

I think we can directly use:

```
impl<T: Encodable> Encodable for Vec<T> {
    #[inline]
    fn length(&self) -> usize {
        list_length(self)
    }

    #[inline]
    fn encode(&self, out: &mut dyn BufMut) {
        encode_list(self, out)
    }
}
```

here

crates/primitives/src/transaction/account_abstraction.rs
```diff
@@ -106,7 +109,264 @@ impl Decodable for Call {
...
+
+impl Encodable for KeyAuthorization {
+    fn encode(&self, out: &mut dyn BufMut) {
+        self.rlp_header().encode(out);
+        // Encode key_type as u8
+        let sig_type_byte: u8 = match self.key_type {
+            SignatureType::Secp256k1 => 0,
+            SignatureType::P256 => 1,
+            SignatureType::WebAuthn => 2,
+        };
```
mattsse: then we can reuse this as .into()

crates/precompiles/src/account_keychain/dispatch.rs
```diff
@@ -0,0 +1,69 @@
...
+                })
+            }
+
+            IAccountKeychain::getTransactionKeyCall::SELECTOR => {
+                view::<IAccountKeychain::getTransactionKeyCall>(calldata, |call| {
+                    self.get_transaction_key(call, msg_sender)
+                })
+            }
+
+            _ => Err(PrecompileError::Other("Unknown function selector".into())),
```
klkvr: we handle those a bit differently now https://github.com/tempoxyz/tempo/blob/cf1d487690524bd391b10ffda4a52446f0337da1/crates/precompiles/src/stablecoin_exchange/dispatch.rs#L183

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -0,0 +1,318 @@
...
+/// Account Keychain contract for managing authorized keys
+#[contract]
+pub struct AccountKeychain {
+    // keys[account][keyId] -> AuthorizedKey
+    keys: Mapping<Address, Mapping<Address, AuthorizedKey>>,
+    // spendingLimits[(account, keyId)][token] -> amount
+    // Using a hash of account and keyId as the key to avoid triple nesting
+    spending_limits: Mapping<B256, Mapping<Address, U256>>,
+    // transactionKey[account] -> keyId (Address::ZERO for main key)
+    transaction_key: Mapping<Address, Address>,
```
klkvr: why does this have to be a mapping? isn't it always set once for the current `tx.origin`?

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -0,0 +1,318 @@
...
+
+    /// Initializes the account keychain contract.
+    pub fn initialize(&mut self) -> Result<()> {
+        self.storage.set_code(
+            ACCOUNT_KEYCHAIN_ADDRESS,
+            Bytecode::new_legacy(Bytes::from_static(&[0xef])),
+        )?;
+
+        Ok(())
+    }
```
klkvr, line 117: I believe this is first precompile that needs to be deployed based at a hardfork vs at genesis so it's likely we need to add some logic to block executor to handle that

and ideally we will also need a test that makes sure that transition happens correctly

crates/precompiles/src/tip20/mod.rs
```diff
@@ -515,6 +516,26 @@ impl<'a, S: PrecompileStorageProvider> TIP20Token<'a, S> {
 
     // Standard token functions
     pub fn approve(&mut self, msg_sender: Address, call: ITIP20::approveCall) -> Result<bool> {
+        // Only check access keys after Moderato hardfork
+        if self.storage.spec().is_moderato() {
```
klkvr: reminder to replace all of those with `is_allegretto`

crates/precompiles/src/tip20/mod.rs
```diff
@@ -770,6 +800,44 @@ impl<'a, S: PrecompileStorageProvider> TIP20Token<'a, S> {
...
+    /// sufficient spending limit for the approval increase.
+    fn check_spending_limit(
+        &mut self,
+        account: &Address,
+        transaction_key: Address,
+        amount: U256,
+    ) -> Result<()> {
+        // Create AccountKeychain instance to read transaction key
+        let mut keychain = AccountKeychain::new(self.storage);
+
```
klkvr: can we encapsulate all of that in `AccountKeychain`? maybe `AccountKeychain::authorize_transfer`/`authorize_approve`?

crates/primitives/src/transaction/aa_signature.rs
```diff
@@ -64,25 +113,196 @@ impl Default for AASignature {
...
+        bytes.to_compact(buf)
+    }
+
+    fn from_compact(buf: &[u8], len: usize) -> (Self, &[u8]) {
+        // Delegate to Bytes::from_compact which handles variable-length decoding
+        let (bytes, rest) = Bytes::from_compact(buf, len);
+        let signature = Self::from_bytes(&bytes).unwrap_or_else(|_| {
+            // For proptest, return a default value based on the type identifier
+            // Try to preserve the signature type even if decoding fails
+            if bytes.len() == SECP256K1_SIGNATURE_LENGTH {
```
klkvr: hmm not sure I understand what's the idea here, what's wrong with proptest values?

crates/primitives/src/transaction/aa_signature.rs
```diff
@@ -56,6 +99,12 @@ pub enum AASignature {
 
     /// WebAuthn signature with variable-length authenticator data
     WebAuthn(WebAuthnSignature),
```
klkvr: can we merge those into `Primitive(PrimitiveSignature)` variant?

crates/primitives/src/transaction/account_abstraction.rs
```diff
@@ -106,7 +119,253 @@ impl Decodable for Call {
...
+
+/// Key authorization for provisioning access keys
+///
+/// Used in TxAA to add a new key to the AccountKeychain precompile.
+/// The transaction must be signed by the root key to authorize adding this access key.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
+#[cfg_attr(test, reth_codecs::add_arbitrary_tests(rlp))]
+pub struct KeyAuthorization {
```
klkvr: I'd prefer this to live in a separate file

crates/primitives/src/transaction/account_abstraction.rs
```diff
@@ -106,7 +119,253 @@ impl Decodable for Call {
...
+#[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
+#[cfg_attr(feature = "reth-codec", derive(reth_codecs::Compact))]
+#[cfg_attr(test, reth_codecs::add_arbitrary_tests(compact, rlp))]
+pub struct TokenLimit {
+    /// TIP20 token address
+    pub token: Address,
+
+    /// Maximum spending amount for this token (enforced over the key's lifetime)
+    pub limit: U256,
+}
```
klkvr: can we just derive `RlpEncodable`/`RlpDecodable` here and for `KeyAuthorization`?

crates/primitives/src/transaction/account_abstraction.rs
```diff
@@ -106,7 +119,253 @@ impl Decodable for Call {
...
+#[cfg(any(test, feature = "arbitrary"))]
+impl<'a> arbitrary::Arbitrary<'a> for KeyAuthorization {
+    fn arbitrary(u: &mut arbitrary::Unstructured<'a>) -> arbitrary::Result<Self> {
+        // Generate key_type independently - this is the type of KEY being authorized,
+        // not the type of the authorization signature
+        let key_type = u.arbitrary()?;
+
+        // Generate a simple Secp256k1 signature for the authorization (from root key)
+        // Using Secp256k1 is most common for root keys
+        let signature = AASignature::Secp256k1(Signature::test_signature());
```
klkvr: why not generate an actual arbitrary sig here? this means we never test encoding of this with non-secp256k1 keys

crates/primitives/src/transaction/account_abstraction.rs
```diff
@@ -106,7 +119,253 @@ impl Decodable for Call {
...
+}
+
+/// Key authorization for provisioning access keys
+///
+/// Used in TxAA to add a new key to the AccountKeychain precompile.
+/// The transaction must be signed by the root key to authorize adding this access key.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
+#[cfg_attr(test, reth_codecs::add_arbitrary_tests(rlp))]
```
klkvr:
```suggestion
#[cfg_attr(test, reth_codecs::add_arbitrary_tests(compact, rlp))]
```

crates/primitives/src/transaction/account_abstraction.rs
```diff
@@ -229,6 +494,17 @@ impl TxAA {
...
         mem::size_of::<Option<u64>>() + // valid_after
+        // key_authorization (optional)
+        if let Some(key_auth) = &self.key_authorization {
+            mem::size_of::<u64>() + // expiry
+            mem::size_of::<Address>() + // key_id
+            key_auth.signature.to_bytes().len() + // signature
+            key_auth.limits.iter().map(|_limit| {
+                mem::size_of::<Address>() + mem::size_of::<U256>()
+            }).sum::<usize>()
+        } else {
```
klkvr: let's impl `InMemorySize` for `KeyAuthorization`

crates/revm/src/tx.rs
```diff
@@ -39,6 +39,17 @@ pub struct AATxEnv {
 
     /// Whether the transaction is a subblock transaction.
     pub subblock_transaction: bool,
+
+    /// Optional key authorization for provisioning access keys
+    pub key_authorization: Option<tempo_primitives::transaction::KeyAuthorization>,
+
+    /// Transaction signature hash (for signature verification)
+    pub tx_hash: B256,
```
klkvr: let's call this `signature_hash` instead of `tx_hash` which usually refers to the actual tx hashed with signature

crates/revm/src/handler.rs
```diff
@@ -51,21 +57,41 @@ const P256_VERIFY_GAS: u64 = 5_000;
...
-/// Calculates the gas cost for verifying an AA signature.
+/// Calculates the gas cost for verifying a primitive signature.
 ///
 /// Returns the additional gas required beyond the base transaction cost:
 /// - Secp256k1: 0 (already included in base 21k)
 /// - P256: 5000 gas
 /// - WebAuthn: 5000 gas + calldata cost for webauthn_data
 #[inline]
-fn aa_signature_verification_gas(signature: &AASignature) -> u64 {
+fn primitive_signature_verification_gas(signature: &PrimitiveSignature, is_istanbul: bool) -> u64 {
```
klkvr: we don't need `is_istanbul` here, on Tempo it's always true

crates/revm/src/handler.rs
```diff
@@ -587,10 +613,191 @@ where
...
+                    .recover_signer(&aa_tx_env.tx_hash)
+                    .map_err(|_| {
+                        EVMError::Transaction(
+                            TempoInvalidTransaction::AccessKeyAuthorizationFailed {
+                                reason:
+                                    "Failed to recover access key address from Keychain signature"
+                                        .to_string(),
+                            },
+                        )
+                    })?;
```
klkvr: it might be worth exploring doing this during `Tx -> TxEnv` conversion, like we do with fee payers: 
https://github.com/tempoxyz/tempo/blob/caa6cfa6fe91c2fa446ed2e7b84d8bb2e8ce5fe2/crates/revm/src/tx.rs#L297-L300
https://github.com/tempoxyz/tempo/blob/caa6cfa6fe91c2fa446ed2e7b84d8bb2e8ce5fe2/crates/revm/src/tx.rs#L68-L73

the benefit here is that because we are doing the conversion to `TxEnv` in parallel, this allows to parallelize recovery and not block execution on potentially heavy operation

crates/revm/src/handler.rs
```diff
@@ -587,10 +613,191 @@ where
...
+
+                // Only allow if authorizing the same key that's being used (same-tx auth+use)
+                if access_key_addr != key_auth.key_id {
+                    return Err(EVMError::Transaction(
+                            TempoInvalidTransaction::AccessKeyAuthorizationFailed {
+                                reason: "Access keys cannot authorize other keys. Only the root key can authorize new keys.".to_string(),
+                            },
+                        ));
+                }
```
klkvr, line 638: ah so we are using addresses for the keyIds because we can only recover an `Address` from the sig?

legion2002: Yes, we needed a natural ID for the signature. We already have that in the form of the recovered address, so we just use that.

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -42,6 +47,24 @@ impl<'a, S: PrecompileStorageProvider> AccountKeychain<'a, S> {
         Self::_new(ACCOUNT_KEYCHAIN_ADDRESS, storage)
     }
 
+    /// Load transaction key from transient storage
+    fn tload_transaction_key(&mut self) -> Result<Address> {
+        let value = self
+            .storage
+            .tload(ACCOUNT_KEYCHAIN_ADDRESS, TRANSACTION_KEY_SLOT)?;
```
0xrusowsky: fyi, you can use `self.address()` rather than `ACCOUNT_KEYCHAIN_ADDRESS`, but both work fine

crates/revm/src/handler.rs
```diff
@@ -587,10 +613,191 @@ where
...
+        if let Some(aa_tx_env) = tx.aa_tx_env.as_ref()
+            && let Some(key_auth) = &aa_tx_env.key_authorization
+        {
+            // Check if this TX is using a Keychain signature (access key)
+            // Access keys cannot authorize new keys UNLESS it's the same key being authorized (same-tx auth+use)
+            if let tempo_primitives::AASignature::Keychain(keychain_sig) = &aa_tx_env.signature {
+                // Recover the access key address from the inner signature
+                let access_key_addr = keychain_sig
+                    .signature
+                    .recover_signer(&aa_tx_env.tx_hash)
```
rakita: Is tx_hash with or without signature? Maybe worth renaming it to tx_sig_hash

crates/revm/src/handler.rs
```diff
@@ -587,10 +613,191 @@ where
...
+                    .map_err(|_| {
+                        EVMError::Transaction(
+                            TempoInvalidTransaction::AccessKeyAuthorizationFailed {
+                                reason:
+                                    "Failed to recover access key address from Keychain signature"
+                                        .to_string(),
+                            },
+                        )
```
rakita, line 628: From slack `But since these are revokable keys, we cannot guarantee that the sig will be valid at the time of inclusion.` If we can't guarantee recovery, we shouldn't invalidate the tx?

Still wrapping my head around it, not sure if this is checked in txpool or not

crates/revm/src/handler.rs
```diff
@@ -587,10 +613,191 @@ where
...
+
+            // Recover the signer of the KeyAuthorization
+            let auth_signer = key_auth
+                .signature
+                .recover_signer(&auth_message_hash)
+                .map_err(|_| {
+                    EVMError::Transaction(TempoInvalidTransaction::AccessKeyAuthorizationFailed {
+                        reason: "Failed to recover signer from KeyAuthorization signature"
+                            .to_string(),
+                    })
```
rakita: Should we continue here, and not make a tx invalid?

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -0,0 +1,470 @@
...
+        };
+
+        // Create and store the new key
+        let new_key = AuthorizedKey {
+            signature_type,
+            expiry: call.expiry,
+            is_active: true,
+        };
+
+        self.sstore_keys(msg_sender, call.keyId, new_key)?;
```
rakita, line 157: Couldn't quickly find this fn, but would it store to multiple storage slots?

crates/revm/src/handler.rs
```diff
@@ -587,10 +613,191 @@ where
...
+
+            // Call precompile to authorize the key (same phase as nonce increment)
+            keychain
+                .authorize_key(*root_account, authorize_call)
+                .map_err(|err| match err {
+                    TempoPrecompileError::Fatal(err) => EVMError::Custom(err),
+                    err => TempoInvalidTransaction::AccessKeyAuthorizationFailed {
+                        reason: err.to_string(),
+                    }
+                    .into(),
```
rakita, line 704: This shouldn't fail, right?

crates/precompiles/src/account_keychain/mod.rs
legion2002: Yes I think this was a mapping from an older design, updated now.

crates/revm/src/handler.rs
legion2002: Yes this is now renamed to signature_hash

crates/precompiles/src/account_keychain/mod.rs
legion2002: No, I think it just does one SSTORE, but on a nested mapping.  This is generated as part of the new macros

crates/revm/src/handler.rs
legion2002: Yeah we started doing this check in the tx pool too now, so ig we should be okay with invalidating it here?
I think I am not very familiar with this nuance of when to invalidate tx in these cases or not.
Let me know if you have any strong opinions here, and I can update the impl accordingly CC: @klkvr

crates/revm/src/handler.rs
legion2002: After caching access key recovery during recover_signer, I believe just doing this should work?
https://github.com/tempoxyz/tempo/pull/598/commits/5f8a40037a30b753535be44585ff62692effb36a

crates/primitives/src/transaction/aa_signature.rs
legion2002: Yeah this is a relic of some old weird encoding rountrip issues I was facing. This has been removed now.

crates/primitives/src/transaction/account_abstraction.rs
legion2002: Yup this was also some old issue, also has been updated.

crates/chainspec/src/genesis/andantino.json
```diff
@@ -9085,6 +9085,11 @@
         "0xfffca9576adb611994d270130f886f4652f02d0bdd2bdf4f4f3053770ce08b26": "0x000000000000000000000000000000000000000000000000ffffffffffffffff"
       }
     },
+    "0xAAAAAAAA00000000000000000000000000000000": {
+      "nonce": "0x0",
+      "balance": "0x0",
+      "code": "0xef"
+    },
```
mattsse: if we add more precompiles for a hardfork we need to do this at runtime, because this will alter the genesis

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -0,0 +1,600 @@
...
+/// - bytes 1-8: expiry (u64, little-endian)
+/// - byte 9: is_active (bool)
+#[derive(Debug, Clone, Default, PartialEq, Eq, Storable)]
+pub struct AuthorizedKey {
+    pub signature_type: u8, // 0: secp256k1, 1: P256, 2: WebAuthn
+    pub expiry: u64,        // Block timestamp when key expires
+    pub is_active: bool,    // Whether key is active
+}
+
+// TODO: Can we get some precompile macro utilities for this, which are compatible with read-only contexts?
```
mattsse, line 40: is this still relevant?

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -0,0 +1,600 @@
...
+///
+/// Storage layout (packed into single slot, right-aligned):
+/// - byte 0: signature_type (u8)
+/// - bytes 1-8: expiry (u64, little-endian)
+/// - byte 9: is_active (bool)
+#[derive(Debug, Clone, Default, PartialEq, Eq, Storable)]
+pub struct AuthorizedKey {
+    pub signature_type: u8, // 0: secp256k1, 1: P256, 2: WebAuthn
+    pub expiry: u64,        // Block timestamp when key expires
+    pub is_active: bool,    // Whether key is active
```
mattsse: would like to convert these to proper /// docs at the line above

crates/precompiles/src/account_keychain/mod.rs
mattsse: yeah we'd need to force set the code at activation, e.g. if code is empty

we can't really make the account not empty because we can't send it some balance, unsure if there's another way.

crates/precompiles/src/lib.rs
```diff
@@ -95,6 +101,9 @@ pub fn extend_tempo_precompiles(precompiles: &mut PrecompilesMap, cfg: &CfgEnv<T
             Some(NoncePrecompile::create(chain_id, spec))
         } else if *address == VALIDATOR_CONFIG_ADDRESS {
             Some(ValidatorConfigPrecompile::create(chain_id, spec))
+        } else if *address == ACCOUNT_KEYCHAIN_ADDRESS && spec.is_moderato() {
+            // AccountKeychain is only available after Moderato hardfork
+            Some(AccountKeychainPrecompile::create(chain_id, spec))
```
mattsse: we need to be careful here if alegretto is active we'd need to potentially initialize the account if we dont do it somewhere else.

we had a previous bug where a set_code call triggered a checkedpoint that then caused issues if a tx reverts.

unsure what the best solution to this is

crates/primitives/src/transaction/aa_signature.rs
```diff
@@ -42,11 +45,127 @@ pub struct WebAuthnSignature {
     pub webauthn_data: Bytes,
 }
 
+/// Primitive signature types that can be used standalone or within a Keychain signature.
+/// This enum contains only the base signature types: Secp256k1, P256, and WebAuthn.
+/// It does NOT support Keychain signatures to prevent recursion.
+///
+/// Note: This enum uses custom RLP encoding via `to_bytes()` and does NOT derive Compact.
```
mattsse, line 54: I think this is appropriate

crates/transaction-pool/src/validator.rs
```diff
@@ -40,6 +87,116 @@ where
...
         }
 
+        // For AA transactions with Keychain signatures, validate authorization
+        enum ValidationResult {
+            ValidateKeychain(Address, Address),
+            Skip,
+            Reject(String),
+        }
+
+        match (|| -> ValidationResult {
```
mattsse: is this outer match here needed?
this looks a bit strange

crates/primitives/src/transaction/account_abstraction.rs
```diff
@@ -766,6 +826,181 @@ mod serde_input {
...
+        }
+
+        fn from_compact(buf: &[u8], len: usize) -> (Self, &[u8]) {
+            // HACK: for OldTxAA 5th byte is highest non-zero chainid byte. For NewTxAA its
+            // either 1 (when keyAuthorization is Some) or 0 (when keyAuthorization is None)
+            //
+            // We infer the encoding version by checking this byte. The assumption here is that this
+            // encoding will only be used for chains which chain_id's highest non-zero byte is >1.
+            //
+            // This is very hacky and should be removed ASAP.
```
mattsse, line 943: can only drop this once we retire andantino or if manually checked that this isnt an issue

legion2002: Yup, we'll do it before mainnet

crates/precompiles/src/account_keychain/mod.rs
legion2002: Yes, @0xrusowsky can you PTAL, and suggest if there's a better way to read precompile state in the validator file.

crates/revm/src/handler.rs
rakita: General rule is, if it is passing in txpool, it should be valid tx and should not fail on block building

crates/revm/src/handler.rs
legion2002: I think it should fail here, because we do the same in the tx pool validator now. And a valid key authorization will never expire, because root keys cannot expire.

crates/transaction-pool/src/validator.rs
```diff
@@ -26,6 +38,133 @@ where
...
+            // The KeyAuthorization MUST be signed by the root account
+            if let Some(key_auth) = aa_signed.tx().key_authorization.as_ref() {
+                // Compute the message hash for the KeyAuthorization
+                let auth_message_hash =
+                    tempo_primitives::transaction::KeyAuthorization::authorization_message_hash(
+                        key_auth.key_type.clone(),
+                        key_auth.key_id,
+                        key_auth.expiry,
+                        &key_auth.limits,
+                    );
```
rakita: Would be nice to have a function `key_auth.sig_hash()`

crates/revm/src/handler.rs
legion2002: If the control flow reaches here, then yes authorize_key should never revert, provided the signature type is correct.
Because we have already checked that the tx is signed by the root key.

crates/transaction-pool/src/validator.rs
```diff
@@ -26,6 +38,133 @@ where
...
+
+                // Recover the signer of the KeyAuthorization
+                match key_auth.signature.recover_signer(&auth_message_hash) {
```
rakita: I am a fan of the let else branching. 
```rust
let Ok(auth_signer) = key_auth.signature.recover_signer(&auth_message_hash) {
    return ValidationResult::Reject(format!(
                            "Invalid KeyAuthorization signature: failed to recover signer: {e}"
    ));
}

if auth_signer != signer {}

return ValidationResult::Skip
```

crates/transaction-pool/src/validator.rs
legion2002: Oh yes, good point. Updated both comments

crates/evm/src/block.rs
```diff
@@ -447,7 +451,33 @@ where
...
+                .load_account_with_code_mut(ACCOUNT_KEYCHAIN_ADDRESS)
+                .map_err(BlockExecutionError::other)?;
+
+            // Only initialize if the account has no code
+            if keychain_account.data.info.is_empty_code_hash() {
+                let code_hash = Bytecode::new_legacy(Bytes::from_static(&[0xef])).hash_slow();
+                keychain_account
+                    .data
+                    .set_code(code_hash, Bytecode::new_legacy(Bytes::from_static(&[0xef])));
+            }
```
mattsse: @klkvr I believe this should work

this isnt dependent on any of the transactions, but has the sideffect that this always loads the account, but I believe here this inst actually problematic

crates/precompiles/src/lib.rs
```diff
@@ -95,6 +101,9 @@ pub fn extend_tempo_precompiles(precompiles: &mut PrecompilesMap, cfg: &CfgEnv<T
             Some(NoncePrecompile::create(chain_id, spec))
         } else if *address == VALIDATOR_CONFIG_ADDRESS {
             Some(ValidatorConfigPrecompile::create(chain_id, spec))
+        } else if *address == ACCOUNT_KEYCHAIN_ADDRESS && spec.is_allegretto() {
+            // AccountKeychain is only available after Allegretto hardfork
+            Some(AccountKeychainPrecompile::create(chain_id, spec))
```
mattsse, line 106: this means at this point the contract should always exists

crates/primitives/src/transaction/aa_signature.rs
```diff
@@ -58,31 +65,200 @@ pub enum AASignature {
...
+        // Decode user_address and signature, initialize cached_key_id as empty
+        let (user_address, rest) = Address::from_compact(buf, len);
+        let remaining_len = len - (buf.len() - rest.len());
+        let (signature, rest) = PrimitiveSignature::from_compact(rest, remaining_len);
+
+        (
+            Self {
+                user_address,
+                signature,
+                cached_key_id: OnceLock::new(),
```
mattsse: this makes sense

crates/transaction-pool/src/validator.rs
```diff
@@ -26,6 +38,133 @@ where
...
+            return ValidationResult::Skip;
+        }
+
+        let sender = transaction.sender();
+        let sig_hash = aa_signed.signature_hash();
+
+        // Recover and validate the access key ID
+        let access_key_addr = match aa_signed.signature().key_id(&sig_hash).ok().flatten() {
+            Some(addr) => addr,
+            None => return ValidationResult::Skip,
```
rakita: Does key_id recover the signature here? If this is true, should we return error on invalid signature?

legion2002: good catch, fixed.

crates/evm/src/block.rs
```diff
@@ -449,7 +453,33 @@ where
...
```
klkvr: I don't think using `journal` is correct here because then if we don't end up executing any transactions this will never be commited iiuc 

instead we probably want to do something similar to how OP handles create2 deployer at Canyon https://github.com/alloy-rs/evm/blob/ff6db1912088f421c8bab7ee6ae2b9f097d746a0/crates/op-evm/src/block/canyon.rs#L19

crates/primitives/src/transaction/aa_signature.rs
```diff
@@ -58,31 +65,200 @@ pub enum AASignature {
...
+///
+/// Format: 0x03 || user_address (20 bytes) || inner_signature
+///
+/// The user_address is the root account this transaction is being executed for.
+/// The inner signature proves an authorized access key signed the transaction.
+/// The handler validates that user_address has authorized the access key in the KeyChain precompile.
+#[derive(Clone, Debug)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
+pub struct KeychainSignature {
```
klkvr: this needs roundtrip tests

crates/primitives/src/transaction/aa_signature.rs
```diff
@@ -42,12 +45,16 @@ pub struct WebAuthnSignature {
...
+/// It does NOT support Keychain signatures to prevent recursion.
+///
+/// Note: This enum uses custom RLP encoding via `to_bytes()` and does NOT derive Compact.
+/// The Compact encoding is handled at the parent struct level (e.g., KeyAuthorization).
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "serde", serde(tag = "type", rename_all = "camelCase"))]
-#[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
-pub enum AASignature {
+pub enum PrimitiveSignature {
```
klkvr, line 64: this as well

crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -0,0 +1,223 @@
...
+///
+/// Defines a per-token spending limit for an access key provisioned via key_authorization.
+/// This limit is enforced by the AccountKeychain precompile when the key is used.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
+#[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
+#[cfg_attr(feature = "reth-codec", derive(reth_codecs::Compact))]
+#[cfg_attr(test, reth_codecs::add_arbitrary_tests(compact, rlp))]
+#[derive(alloy_rlp::RlpEncodable, alloy_rlp::RlpDecodable)]
```
klkvr: let's move this to the `derive` attr list above

crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -0,0 +1,223 @@
...
+        self.expiry.encode(out);
+        self.limits.encode(out);
+        self.key_id.encode(out);
+        self.signature.encode(out);
+    }
+
+    fn length(&self) -> usize {
+        self.rlp_header().length_with_payload()
+    }
+}
```
klkvr: why can't we derive this?

crates/revm/src/handler.rs
klkvr: > General rule is, if it is passing in txpool, it should be valid tx and should not fail on block building

this can't always be the case as you don't know whether one of the transactions you've executed will invalidate yours

crates/revm/src/handler.rs
```diff
@@ -587,10 +607,199 @@ where
...
+                    })?;
+
+                // Only allow if authorizing the same key that's being used (same-tx auth+use)
+                if access_key_addr != key_auth.key_id {
+                    return Err(EVMError::Transaction(
+                            TempoInvalidTransaction::AccessKeyAuthorizationFailed {
+                                reason: "Access keys cannot authorize other keys. Only the root key can authorize new keys.".to_string(),
+                            },
+                        ));
+                }
```
klkvr, line 638: what's the motivation here? why can't session key A use authorization for session key B given that key B was signed by the root EOA anyway

crates/revm/src/handler.rs
```diff
@@ -587,10 +607,199 @@ where
...
+                })?
+                .ok_or_else(|| {
+                    EVMError::Transaction(TempoInvalidTransaction::AccessKeyAuthorizationFailed {
+                        reason: "Expected Keychain signature but got different type".to_string(),
+                    })
+                })?;
+
+            // Check if this transaction includes a KeyAuthorization for the same key
+            // If so, skip validation here - the key will be authorized during execution
+            let is_authorizing_this_key = aa_tx_env
```
klkvr, line 746: hmm didn't we just insert the key that's being authorized into the contract? https://github.com/tempoxyz/tempo/blob/b5e53adc08642cb34a267840706743074bd77379/crates/revm/src/handler.rs#L712-L721

not sure I understand why we're skipping the validation here

crates/revm/src/handler.rs
legion2002: This is because, this is "free work" we are doing before tx validation. So we want to limit the extent of this work, just like in precompiles.
The only usecase for allowing protocol level key authorization _before_ sig validation is for the same tx auth + use case, for all other cases you can just directly call the precompile to authorize the key in advance.

crates/revm/src/handler.rs
legion2002: We only allow key authorizations in-protocol, if the access key being authorized, is actually being used in the same tx.
This means when we authorize the key above, we have already verified it to be correct.

So we don't need to validate it again. We just do this for efficiency.

crates/primitives/src/transaction/key_authorization.rs
legion2002: It was because of the sig type field being an enum. 
But now I added a manual implementation for the enum, and automatically derived this.

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -0,0 +1,593 @@
...
+}
+
+/// Account Keychain contract for managing authorized keys
+#[contract]
+pub struct AccountKeychain {
+    // keys[account][keyId] -> AuthorizedKey
+    keys: Mapping<Address, Mapping<Address, AuthorizedKey>>,
+    // spendingLimits[(account, keyId)][token] -> amount
+    // Using a hash of account and keyId as the key to avoid triple nesting
+    spending_limits: Mapping<B256, Mapping<Address, U256>>,
```
rakita, line 58: Not sure how `contract` macro works, but should this be `(account, KeyId, token)` map to U256?

crates/revm/src/handler.rs
rakita: Maybe I am mistaken but it should be on mainnet (maybe it is not on tempo), EIP-7702 even introduced new rule to only have one tx per sender bcs of this. PR: https://github.com/paradigmxyz/reth/pull/15312

crates/revm/src/handler.rs
rakita: Makes sense if we are invalidating this in txpool

crates/revm/src/handler.rs
rakita: Would be good to leave a comment here

crates/revm/src/handler.rs
klkvr: yes but I'm not sure if this actually helps

e.g if you have a tx from Bob that has a delegated account, and then Alice submits a tx that bumps Bob's nonce by doing a CREATE from their delegated account, nothing prevents you from yielding Bob's tx after Alice's (unless you explicitly track Bob's nonce in the tx iterator which we don't do currently)

Issue #549, opened 2025-10-24
Feature Request: Indexed RPC endpoints


jxom:
On top of #536, we will also need the following endpoints as listed on the [proposed specs here](https://github.com/tempoxyz/docs/pull/71):

- [ ] - `amm_getLiquidityPools`  
- [ ] - `dex_getOrderbooks`  
- [ ] - `eth_getTransactions`  
- [ ] - `policy_getAddresses`  
- [ ] - `token_getTokens`  
- [ ] - `token_getTokensByAddress`  
- [ ] - `token_getRoleHistory`

PR #599, opened 2025-10-30
feat(node): Add generic response type for pagination JSON-RPC API endpoints

RomanHodulak:
Part of #549

Unifies the response type for pagination under `PaginationResponse`.

This response is commonly used in the custom JSON-RPC API extensions so it would be helpful to have it defined once with a generic item type.

### DIFF
--- a/crates/node/src/rpc/dex/types.rs
+++ b/crates/node/src/rpc/dex/types.rs
@@ -1,8 +1,37 @@
 use alloy::primitives::{Address, B256};
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Serialize, Serializer, ser::SerializeStruct};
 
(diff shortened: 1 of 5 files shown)


crates/node/src/rpc/dex/types.rs
```diff
@@ -1,9 +1,38 @@
...
 pub type OrdersParams = PaginationParams<OrdersFilters>;
 pub type Tick = i16;
 
+#[derive(Clone, Debug, Default, PartialEq, Eq, Hash)]
+pub struct PaginationResponse<Item> {
+    /// Cursor for next page, null if no more results
+    pub next_cursor: Option<String>,
+    /// Array of items matching the input query
+    pub items: Vec<Item>,
+}
```
mattsse, line 13: I think we dont really need this abstraction and can instead just duplicate container types for the responses

PR #601, opened 2025-10-30
fix(consensus):use grandparent as finalized hash, only do backfill on init

SuperFluffy:
This patch addresses the backfill issues one of our users is experiencing when starting a node that is far behind:

Right now, the executor actor is starting a new backfill process on every FCU it is sending to the execution layer and for which the EL is responding with `SYNCING`. Say the node is on height 0 but network is on height 1_000, then FCU triggers a backfill (one on 1_001, one on 1_002, and so on). All of these backfills are racing each other.

Further, each backfill is walking the ancestry from the newest missing height to the oldest. This is a problem because reth needs cannot have gaps when backfilling.

Finally, the executor is only setting `FCU.finalized_hash` after it forwarded a finalized block from the marshal actor to the execution layer. Doing this means that we never trigger reth p2p backups, which are faster than feeding finalized blocks bit by bit.

This patch addresses this issue in several ways:

1. A backfill is only started done on startup. Originally, it was done to address the issue that on shutdown/restart, the execution layer (reth) is not always persisting all finalized blocks sent to it. This logic remains.
2. `FCU.finalized_hash` is now set on every verification and propose request: when receiving either, we are guaranteed that the parent digest in the request is notarized. But this implies that the grandparent is finalized, and so we are allowed to the finalized hash to the grandparent.
3. When forwarding a finalized block from the marshal actor to the execution layer, also send an FCU if the finalized block hash and height are ahead of the last FCU. Similarly, if the finalized hash would end up ahead of head, also update head. This relies on the fact that the executor agent is doing all action strictly sequentially.

### DIFF
--- a/crates/commonware-node/src/consensus/execution_driver/mod.rs
+++ b/crates/commonware-node/src/consensus/execution_driver/mod.rs
@@ -29,7 +29,7 @@ use eyre::{OptionExt, WrapErr as _, bail, ensure, eyre};
 use futures::{
     SinkExt as _, StreamExt as _, TryFutureExt as _,
     channel::{mpsc, oneshot},
-    future::{Either, always_ready, try_join},
+    future::{Either, always_ready, ready, try_join},
 };
 use rand::{CryptoRng, Rng};
 use reth::payload::EthBuiltPayload;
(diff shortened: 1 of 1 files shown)


bin/tempo-bench/src/cmd/genesis.rs
```diff
@@ -25,11 +25,10 @@ use tempo_precompiles::{
     LINKING_USD_ADDRESS, TIP_FEE_MANAGER_ADDRESS,
     contracts::{
         EvmStorageProvider, IFeeManager, ITIP20, ITIP20Factory, TIP20Factory, TIP20Token,
-        linking_usd::LinkingUSD, stablecoin_exchange::StablecoinExchange,
+        TIP403Registry, linking_usd::LinkingUSD, stablecoin_exchange::StablecoinExchange,
```
SuperFluffy, line 28: I have no idea why that's here.

EDIT: Ah, fmt

PR #604, opened 2025-10-30
feat(precompiles): enforce dust limit check on DEX orders

fgimenez:
- adds `MIN_ORDER_AMOUNT` constant (10 USD = 10_000_000 with 6 decimals)
- adds `BelowMinimumOrderSize` error to prevent orders below the dust limit
- enforces minimum order size in `place()` and `place_flip()` functions
- updates existing test to use amounts above the new dust limit

### DIFF
--- a/crates/precompiles/src/stablecoin_exchange/mod.rs
+++ b/crates/precompiles/src/stablecoin_exchange/mod.rs
@@ -29,8 +29,15 @@ use crate::{
 use alloy::primitives::{Address, B256, Bytes, IntoLogData, U256};
 use revm::state::Bytecode;
 
-/// Minimum order size (dust limit): $10 with 6 decimals = 10_000_000
-pub const MIN_ORDER_AMOUNT: u128 = 10_000_000;
+/// Minimum order size in token units (10 units of the token)
+pub const MIN_ORDER_UNITS: u128 = 10;
+
+/// Calculate minimum order amount based on token decimals
+/// Returns MIN_ORDER_UNITS scaled by the token's decimals
+/// For example: 6 decimals -> 10_000_000, 18 decimals -> 10_000_000_000_000_000_000
+pub fn calculate_min_order_amount(decimals: u8) -> u128 {
+    MIN_ORDER_UNITS * 10u128.pow(decimals as u32)
+}
 
 /// Calculate quote amount from base amount and tick price using checked arithmetic
 ///
@@ -452,8 +459,10 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
         is_bid: bool,
         tick: i16,
     ) -> Result<u128, TempoPrecompileError> {
-        // Lookup quote token from TIP20 token
-        let quote_token = TIP20Token::from_address(token, self.storage).quote_token()?;
+        // Lookup quote token and decimals from TIP20 token
+        let mut token_contract = TIP20Token::from_address(token, self.storage);
+        let quote_token = token_contract.quote_token()?;
+        let decimals = token_contract.decimals()?;
 
         // Compute book_key from token pair
         let book_key = compute_book_key(token, quote_token);
@@ -467,8 +476,9 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
             return Err(StablecoinExchangeError::tick_out_of_bounds(tick).into());
         }
 
-        // Validate order amount meets dust limit
-        if amount < MIN_ORDER_AMOUNT {
+        // Validate order amount meets minimum requirement
+        let min_order_amount = calculate_min_order_amount(decimals);
+        if amount < min_order_amount {
             return Err(StablecoinExchangeError::below_minimum_order_size(amount).into());
         }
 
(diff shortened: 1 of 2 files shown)


crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -536,6 +546,13 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
             return Err(StablecoinExchangeError::invalid_flip_tick().into());
         }
 
+        // Validate order amount meets dust limit
+        if amount < DUST_LIMIT {
+            return Err(
+                StablecoinExchangeError::below_minimum_order_size(amount, DUST_LIMIT).into(),
```
0xKitsune: If the minimum order size is a constant, do we need to include data in the error response?

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -29,6 +29,9 @@ use crate::{
 use alloy::primitives::{Address, B256, Bytes, IntoLogData, U256};
 use revm::state::Bytecode;
 
+/// Minimum order size (dust limit): $100 with 6 decimals = 100_000_000
+pub const DUST_LIMIT: u128 = 100_000_000;
```
0xKitsune: nit: can we rename to `MIN_ORDER_AMOUNT` to be a bit more explicit? Also feel free to ignore if preferred.

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -2016,7 +2033,7 @@ mod tests {
             &admin,
             &alice,
             exchange.address,
-            1_000_000u128,
+            100_000_000u128,
```
0xKitsune: Consider using `MIN_ORDER_AMOUNT` here. If we ever want to update this value in the future, tests will automatically work without needing to update each individual test.

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -459,6 +462,13 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
             return Err(StablecoinExchangeError::tick_out_of_bounds(tick).into());
         }
 
+        // Validate order amount meets dust limit
+        if amount < DUST_LIMIT {
+            return Err(
+                StablecoinExchangeError::below_minimum_order_size(amount, DUST_LIMIT).into(),
+            );
+        }
```
0xKitsune: Can we add a test that tries to place an order with `amount < min_amount` and asserts this error?

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -536,6 +546,13 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
             return Err(StablecoinExchangeError::invalid_flip_tick().into());
         }
 
+        // Validate order amount meets dust limit
+        if amount < DUST_LIMIT {
```
0xKitsune: Same here, can we add a test that tries to place an order with `amount < min_amount` and asserts this error?

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -29,6 +29,9 @@ use crate::{
 use alloy::primitives::{Address, B256, Bytes, IntoLogData, U256};
 use revm::state::Bytecode;
 
+/// Minimum order size (dust limit): $100 with 6 decimals = 100_000_000
```
0xKitsune: Just flagging that this is only $100 when the underlying asset is USD. In the future if we support other underlying assets (eg. EUR, etc.) the minimum value will depend on the decimals.

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -464,6 +462,13 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
             return Err(StablecoinExchangeError::tick_out_of_bounds(tick).into());
         }
 
+        // Validate order amount meets dust limit
+        if amount < DUST_LIMIT {
```
mattsse: I assume we should use the token's decimals here @0xKitsune ?

crates/precompiles/src/stablecoin_exchange/mod.rs
fgimenez: good point, removed the constant

crates/precompiles/src/stablecoin_exchange/mod.rs
fgimenez: yep done

crates/precompiles/src/stablecoin_exchange/mod.rs
fgimenez: yep, done, also using `MIN_ORDER_AMOUNT * 5` and `MIN_ORDER_AMOUNT * 10`

crates/precompiles/src/stablecoin_exchange/mod.rs
fgimenez: yep added `test_place_order_below_minimum_amount` and `test_place_flip_order_below_minimum_amount`

crates/precompiles/src/stablecoin_exchange/mod.rs
fgimenez: currently the constant is hardcoded to a USD value, including decimals. instead we could define the constant as token units, and dynamically calculate the actual minimum value depending on each token decimals, wdyt @0xKitsune @mattsse should we implement that now?

0xKitsune: Yeah IMO lets implement this now, otherwise it might be a footgun if we forget to change in the future.

fgimenez: @0xKitsune ok done here https://github.com/tempoxyz/tempo/pull/604/commits/f94913c41ad6f2a65522ff51b18a9a35c44e3764 ptal

0xKitsune: Just took a look, lgtm

PR #606, opened 2025-10-30
feat(precompiles)!: add tip20 error data

tmm:
Adds error data to TIP-20 `InsufficientBalance`. Goal is to make error messages better downstream for devs, including in `tempo.ts` (ideally passing data we already have so no need to fetch anything new). Started with `InsufficientBalance` to keep this initial PR small. If cool to move forward, will create a larger PR for rest of the TIP-20 errors.


crates/node/tests/it/tip20.rs
```diff
@@ -69,9 +69,13 @@ async fn test_tip20_transfer() -> eyre::Result<()> {
     }
 
     // Attempt to transfer more than the balance
-    for (account, _, _) in &account_data {
-        let balance = token.balanceOf(*account).call().await?;
```
tmm, line 73: `balance` wasn't really used before since `account` wasn't used in `token.transfer`

crates/contracts/src/precompiles/tip20.rs
```diff
@@ -141,8 +141,11 @@ impl RolesAuthError {
 
 impl TIP20Error {
     /// Creates an error for insufficient token balance.
-    pub const fn insufficient_balance() -> Self {
-        Self::InsufficientBalance(ITIP20::InsufficientBalance {})
+    pub const fn insufficient_balance(required: Uint<256, 4>, available: Uint<256, 4>) -> Self {
```
mattsse: I believe this should just be
```suggestion
    pub const fn insufficient_balance(required: U256, available: U256) -> Self {
```

crates/node/tests/it/tip20.rs
```diff
@@ -69,9 +69,13 @@ async fn test_tip20_transfer() -> eyre::Result<()> {
...
-        let Err(result) = token
+    for (_, wallet, balance) in &account_data {
+        let account_provider = ProviderBuilder::new()
+            .wallet(wallet.clone())
+            .connect_http(http_url.clone());
```
mattsse, line 75: I see, we want a new token instance with the given wallet

crates/precompiles/src/tip20/mod.rs
```diff
@@ -867,12 +867,12 @@ impl<'a, S: PrecompileStorageProvider> TIP20Token<'a, S> {
     ) -> Result<(), TempoPrecompileError> {
         let from_balance = self.get_balance(&TIP_FEE_MANAGER_ADDRESS)?;
         if refund > from_balance {
-            return Err(TIP20Error::insufficient_balance().into());
+            return Err(TIP20Error::insufficient_balance(refund, from_balance).into());
```
jxom: Is this meant to be the other way round?

```
error InsufficientBalance(uint256 available, uint256 required);
```

```suggestion
            return Err(TIP20Error::insufficient_balance(from_balance, refund).into());
```

tmm: `TIP20Error::insufficient_balance` has the args in the opposite order. Will make it the same as the Solidity signature.

tmm: https://github.com/tempoxyz/tempo/pull/606/commits/16c72969dc1b382189294d8c33171d305938b17d

crates/contracts/src/precompiles/tip20.rs
```diff
@@ -113,7 +113,7 @@ sol! {
         event RewardRecipientSet(address indexed holder, address indexed recipient);
 
         // Errors
-        error InsufficientBalance();
+        error InsufficientBalance(uint256 available, uint256 required);
```
joshieDo: would it make sense to specify which tip20 address?

PR #608, opened 2025-10-30
chore: spawn dex calls blocking

### LINKED ISSUE #603: spawn dex rpc calls blocking
with #578 we have dex endpoints that can perform a bit of IO and especially the iterator based endpoints should be performed on a blocking task

## TODO:
* spawn dex_ calls blocking

mattsse:
closes #603

moves a few things around, and spawns the actual calls blocking

I think we could get rid of some error mapping by directly using TempoEthApi here

### DIFF
--- a/crates/node/src/rpc/dex/error.rs
+++ b/crates/node/src/rpc/dex/error.rs
@@ -1,27 +1,12 @@
-use std::convert::Infallible;
-
 use alloy_eips::BlockId;
 use alloy_primitives::B256;
-use alloy_rpc_types_eth::BlockError;
 use jsonrpsee::types::ErrorObject;
-use reth_evm::revm::context::result::{EVMError, HaltReason};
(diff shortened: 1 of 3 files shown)


crates/node/src/rpc/dex/error.rs
```diff
@@ -61,34 +42,31 @@ pub enum DexApiError {
...
+impl DexApiError {
+    /// Returns the rpc error for this error
+    const fn error_code(&self) -> i32 {
+        match self {
+            Self::InvalidOrderbookCursor(_)
+            | Self::InvalidOrderCursor(_)
+            | Self::ParseOrderCursor(_) => jsonrpsee::types::error::INVALID_PARAMS_CODE,
+            _ => jsonrpsee::types::error::INTERNAL_ERROR_CODE,
+        }
+    }
```
Rjected, line 54: makes sense

PR #612, opened 2025-10-31
feat(node): Add new endpoints and schema for JSON-RPC API

RomanHodulak:
Part of #549

Adds a trait and a dummy implementation for the endpoints with an accompanying schema.

Need your help with the proper types for the numeric fields as the spec doesn't have them.

https://tempo-docs-git-jxom-rpc-tempoxyz.vercel.app/rpc/

### DIFF
--- a/crates/node/src/rpc/policy/addresses.rs
+++ b/crates/node/src/rpc/policy/addresses.rs
@@ -1,4 +1,4 @@
-use crate::rpc::{pagination::PaginationParams, token::role_history::RoleChange};
+use crate::rpc::pagination::PaginationParams;
 use alloy_primitives::Address;
 use serde::{Deserialize, Serialize};
 
@@ -8,7 +8,7 @@ pub struct AddressesResponse {
     /// Cursor for next page, null if no more results
     pub next_cursor: Option<String>,
     /// Array of items matching the input query
-    pub addresses: Vec<RoleChange>,
+    pub addresses: Vec<PolicyAddress>,
 }
 
 #[derive(Clone, Debug, Default, PartialEq, Eq, Hash, Serialize, Deserialize)]


crates/node/src/rpc/token/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+    ) -> RpcResult<PaginationResponse<RoleChange>>;
+
+    #[method(name = "getTokens")]
+    async fn tokens(
+        &self,
+        params: PaginationParams<TokensFilters>,
+    ) -> RpcResult<PaginationResponse<Token>>;
+
+    #[method(name = "getTokensByAddress")]
+    async fn tokens_by_address(
```
mattsse, line 42: for all the function defs in the various namespace traits, can we copy paste the docs from the spec?

like we have for dex:

https://github.com/tempoxyz/tempo/blob/e605c39add3cc30b854adb7a754f2f434980d323/crates/node/src/rpc/dex/api.rs#L9-L11

crates/node/src/rpc/token/tokens.rs
```diff
@@ -0,0 +1,64 @@
...
+    /// Token decimals (from TIP-4217 registry based on currency)
+    pub decimals: u32,
+    /// Token name
+    pub name: String,
+    /// Whether token transfers are paused
+    pub paused: bool,
+    /// Quote token address for trading pairs
+    pub quote_token: Address,
+    /// Maximum token supply
+    pub supply_cap: B256,
```
mattsse: I believe this should be U256

crates/node/src/rpc/token/tokens.rs
```diff
@@ -0,0 +1,64 @@
...
+    pub total_supply: Option<FilterRange<U256>>,
+}
+
+#[derive(Clone, Debug, Default, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Token {
+    /// Token contract address (deterministic vanity address based on tokenId)
+    pub address: Address,
+    /// Timestamp when token was created
+    pub created_at: u64,
```
mattsse, line 44: unsure if we want to stick to hex quantity for numbers,

imo we should, like we did for all the dex response fields:

https://github.com/tempoxyz/tempo/blob/e605c39add3cc30b854adb7a754f2f434980d323/crates/node/src/rpc/dex/types.rs#L132-L153

but defer to @jxom here

crates/node/src/rpc/pagination.rs
```diff
@@ -1,9 +1,38 @@
...
+}
+
+/// A trait whose implementation determines a field name for [`PaginationResponse`] implementation
+/// of [`Serialize`]r.
+pub trait FieldName {
+    /// Returns the camel case plural name for this field.
+    ///
+    /// For example a struct called `RoleChange` would likely return `"roleChanges"`.
+    fn field_plural_camel_case() -> &'static str;
+}
```
mattsse: unsure if we really need this, imo we can just copy paste a new `<Method>Response` container type where needed

this is now defined in the dex/types but also used in the other modules

crates/node/src/rpc/token/tokens.rs
```diff
@@ -0,0 +1,64 @@
...
+    /// Quote token address for trading pairs
+    pub quote_token: Address,
+    /// Maximum token supply
+    pub supply_cap: U256,
+    /// Token symbol
+    pub symbol: String,
+    /// Unique token ID from factory
+    pub token_id: B256,
+    /// Current total supply
+    pub total_supply: B256,
```
fgimenez: should be `U256`?

crates/node/src/rpc/amm/pools.rs
```diff
@@ -0,0 +1,48 @@
...
+    pub validator_token: Option<Address>,
+}
+
+#[derive(Clone, Debug, Default, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Pool {
+    /// Effective reserve of validator token after pending swaps
+    pub effective_reserve_validator_token: U256,
+    /// Pool ID (keccak256 of userToken and validatorToken)
+    pub pool_id: U256,
```
fgimenez: given that the comment says it comes from `keccak256`, this should be `B256`?

crates/node/src/rpc/token/tokens.rs
RomanHodulak: Do you mean that every number should be `u128` with `#[serde(with = "alloy_serde::quantity")]` annotation?

crates/node/src/rpc/token/tokens_by_address.rs
```diff
@@ -0,0 +1,34 @@
...
+    token::tokens::{Token, TokensFilters},
+};
+use alloy_primitives::{Address, B256, U256};
+use jsonrpsee::core::Serialize;
+use serde::Deserialize;
+
+#[derive(Clone, Debug, Default, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct TokensByAddressParams {
+    /// Filter by account address that received/lost role.
```
fgimenez:
```suggestion
    /// Account address to query tokens for.
```

crates/node/src/rpc/pagination.rs
```diff
@@ -0,0 +1,152 @@
...
+#[derive(Clone, Debug, Default, PartialEq, Eq, Hash)]
+pub struct PaginationResponse<Item> {
+    /// Cursor for next page, null if no more results
+    pub next_cursor: Option<String>,
+    /// Array of items matching the input query
+    pub items: Vec<Item>,
+}
+
+/// A trait whose implementation determines a field name for [`PaginationResponse`] implementation
+/// of [`Serialize`]r.
```
fgimenez:
```suggestion
/// of [`Serialize`].
```

crates/node/src/rpc/token/tokens_by_address.rs
```diff
@@ -0,0 +1,34 @@
...
+
+#[derive(Clone, Debug, Default, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct AccountToken {
+    /// Account's balance in this token
+    pub balance: U256,
+    /// Roles the account has for this token
+    pub roles: Vec<B256>,
+    /// Token details
+    pub token: Token,
```
fgimenez, line 35: we are including here the full `Token` (10 fields), if a user has multiple tokens a query might return here a lot of data that might be not needed, maybe better just adding a token address reference? question probably for @jxom

crates/node/src/rpc/pagination.rs
RomanHodulak: Moved to a better module

crates/node/src/rpc/pagination.rs
```diff
@@ -0,0 +1,152 @@
...
+pub trait FieldName {
+    /// Returns the camel case plural name for this field.
+    ///
+    /// For example a struct called `RoleChange` would likely return `"roleChanges"`.
+    fn field_plural_camel_case() -> &'static str;
+}
```
mattsse: I still think this abstractions isnt very useful and I'd rather have dedicated response types for all the endpoints, imo doesnt hurt to copy paste response container types

crates/node/src/rpc/token/tokens_by_address.rs
jxom: yeah, that works.

crates/node/src/rpc/token/tokens.rs
jxom: don't mind

crates/node/src/rpc/token/tokens_by_address.rs
RomanHodulak: Feel a bit iffy about this change because it deviates from the spec which might be a surprise to @gorried whom might have already expected to get exactly what's in the vercel

Issue #611, opened 2025-10-31
feat(precompiles): type-safe storage + improved devex


0xrusowsky:
## Overview

provide a safe, maintainable, and ergonomic abstraction layer for building precompile contracts in Tempo, eliminating low-level storage slot manipulation and reducing boilerplate code.

## Motivation

address the current pain points:
- building precompiles requires manual low-level storage slot manipulation
- high cognitive overhead when implementing new contracts or reviewing/auditing them
- verbose boilerplate patterns (dispatcher, slot constants, storage access methods), which are prone to human-errors

### Safe Storage Primitives

- [x] type-safe storage access that prevents common bugs at compile-time
	- [X] simple evm words
	- [X] mappings
    - [x] custom structs
- [ ] first-class devex support for single-slot values and mappings (including nested mappings)
	- [X] via helper methods
	- [ ] with index-like API (to be done later on)

### 2. Reduce Boilerplate

- [x] define contract storage layout with a struct
   - [X] automatic slot assignation
   - [X] support for custom storage layout
	   - [X] decimal, hex and string (auto-computes the hash)
	   - [X] `slot` (just for the field) + `base_slot` (shifts all subsequent fields)
- [x] interface-driven development: compiler enforces implementation of all interface methods
   - [x] still able to split contract impl in several files (1 impl = 1 file = 1 interface)
   - [x] single storage layout despite requiring storage vars for each interface
   - [x] all storage vars are accessible to all methods, regardless of interface

auto-generation of:
- [x] slot constants
- [x] storage access methods
- [x] getter functions that directly map to storage fields
- [x] call dispatching logic
- [x] helpers to easily emit events and throw errors 
  
## Success criteria

- [ ] all precompiles use the new API
- [ ] existing tests pass with new implementations (may need some minor adjustments)
- [X] full compatibility with existing `PrecompileStorageProvider` and `StorageOps` traits
- [ ] zero or minimal runtime overhead compared to hand-written code

## Future enhancements
- [ ] enhance the macro to automatically digest the interface (kinda like the `sol!` macro) without relying on an hardcoded interface registry
- [ ] enhance storage abstractions and storage layout macro to support:
   - [ ] fixed-size arrays
   - [ ] dynamic arrays (vecs)

PR #615, opened 2025-10-31
[wip] feat(precompiles): storage primitives + contract devex

0xrusowsky:
towards #611


crates/precompiles/src/tip20_migrated/mod.rs
```diff
@@ -0,0 +1,1347 @@
...
+    #[map = "allowance"]
+    allowances: Mapping<Address, Mapping<Address, U256>>, // slot 11
+    nonces: Mapping<Address, U256>, // slot 12
+    paused: bool,       // slot 13
+    supply_cap: U256,   // slot 14
+    salts: Mapping<B256, bool>, // slot 15
+
+    // TIP20 Rewards
+    last_update_time: u64, // slot 16
+    opted_in_supply: U256, // slot 17
```
0xKitsune, line 93: It would be nice (maybe in V2) to extend the `contract` macro to automatically add getter functions to precompile dispatch if struct fields are marked as `pub`. Looks like these are already getting generated, but once we remove the `sol!` macro to define the interface, it would be nice to have this feature to easily know which variables on the precompile are public.

crates/precompiles/src/storage/types/mapping.rs
```diff
@@ -0,0 +1,469 @@
...
+///     allowances: Mapping<Address, Mapping<Address, U256>, [11, 0, 0, 0]>,
+/// }
+/// ```
+/// Note: For nested mappings, only the outermost mapping needs a slot specified.
+#[derive(Debug, Clone, Copy)]
+pub struct Mapping<K, V, const SLOT: [u64; 4]> {
+    _phantom: PhantomData<(K, V)>,
+}
+
+impl<K, V, const SLOT: [u64; 4]> Mapping<K, V, SLOT> {
```
0xKitsune: nit: consider using `U256` instead of `[u64; 4]` as this is slightly more readable.

crates/precompiles/src/storage/types/traits.rs
```diff
@@ -0,0 +1,618 @@
...
+    }
+}
+
+impl Storable for bool {
+    const SLOT_COUNT: usize = 1;
+
+    #[inline]
+    fn load<S: StorageOps>(storage: &mut S, base_slot: U256) -> Result<Self> {
+        let value = storage.sload(base_slot)?;
+        Ok(value != U256::ZERO)
```
0xKitsune, line 159: nit: You can use `value.to::<Self>` here.

crates/precompiles/src/storage/types/traits.rs
```diff
@@ -0,0 +1,618 @@
...
+
+    fn load<S: StorageOps>(storage: &mut S, base_slot: U256) -> Result<Self> {
+        let value = storage.sload(base_slot)?;
+        let bytes = value.to_be_bytes::<32>();
+        let len = bytes[31] as usize / 2; // Length stored as len * 2
+
+        if len > 31 {
+            return Err(TempoPrecompileError::Fatal(
+                "String too long for short string encoding".into(),
+            ));
```
0xKitsune: Would this return `Fatal` in Solidity? Ideally we should return the same errors that a user would encounter on the EVM.

crates/precompiles/src/storage/types/traits.rs
```diff
@@ -0,0 +1,618 @@
...
+        }
+
+        let utf8_bytes = &bytes[..len];
+        Self::from_utf8(utf8_bytes.to_vec()).map_err(|e| {
+            TempoPrecompileError::Fatal(format!("Invalid UTF-8 in stored string: {e}"))
```
0xKitsune: Same thing here, is this the error that would be encountered on the vanilla EVM? If not, we should document when users can hit this error on Tempo and what error message to expect.

crates/precompiles/src/storage/mod.rs
```diff
@@ -22,7 +26,35 @@ pub trait PrecompileStorageProvider {
...
+    fn sstore(&mut self, slot: U256, value: U256) -> Result<(), TempoPrecompileError> {
+        let address = self.address();
+        self.storage().sstore(address, slot, value)
+    }
+
+    fn sload(&mut self, slot: U256) -> Result<U256, TempoPrecompileError> {
+        let address = self.address();
+        self.storage().sload(address, slot)
+    }
+}
```
0xKitsune, line 62: Nice

crates/precompiles/src/tip20_migrated/mod.rs
```diff
@@ -0,0 +1,1347 @@
...
+
+    // Admin functions
+    fn change_transfer_policy_id(
+        &mut self,
+        msg_sender: Address,
+        new_policy_id: u64,
+    ) -> crate::error::Result<()> {
+        self.check_role(msg_sender, DEFAULT_ADMIN_ROLE)?;
+        self._set_transfer_policy_id(new_policy_id)?;
+        self._emit_transfer_policy_update(msg_sender, new_policy_id)
```
0xKitsune: Is there a reason that many of these functions are prefixed with `_`? If not can we remove this?

crates/precompiles/src/tip20_migrated/mod.rs
```diff
@@ -0,0 +1,1347 @@
...
+    }
+
+    fn set_supply_cap(
+        &mut self,
+        msg_sender: Address,
+        new_supply_cap: U256,
+    ) -> crate::error::Result<()> {
+        self.check_role(msg_sender, DEFAULT_ADMIN_ROLE)?;
```
0xKitsune, line 134: We could consider using a `#[modifier = modifier_name]` (not a necessity for V1) for modifiers rather than replicating this logic for all functions. This would reduce repition in many of the functions that check role, if the contract is paused, etc. This also might be overkill, lmk what you think.

crates/precompiles/src/tip20_migrated/mod.rs
```diff
@@ -0,0 +1,1347 @@
...
+        if new_supply_cap < self.total_supply()? {
+            return Err(TIP20Error::supply_cap_exceeded().into());
+        }
+
+        self._set_supply_cap(new_supply_cap)?;
+        self._emit_supply_cap_update(msg_sender, new_supply_cap)
+    }
+
+    fn pause(&mut self, msg_sender: Address) -> crate::error::Result<()> {
```
0xKitsune: Can we add docs for all functions that are currently missing doc comments?

crates/precompiles/src/tip20_migrated/mod.rs
```diff
@@ -0,0 +1,1347 @@
...
+    fn unpause(&mut self, msg_sender: Address) -> crate::error::Result<()> {
+        self.check_role(msg_sender, *PAUSE_ROLE)?;
+        self._set_paused(false)?;
+        self._emit_pause_state_update(msg_sender, false)
+    }
+    fn update_quote_token(
+        &mut self,
+        msg_sender: Address,
+        new_quote_token: Address,
+    ) -> crate::error::Result<()> {
```
0xKitsune: nit: can we use `Result<()>` directly rather than `crate::error::Result`? It looks like we are already using `Result<()>` directly in a few other functions.

crates/precompiles/src/tip20_migrated/mod.rs
```diff
@@ -0,0 +1,1347 @@
...
+
+        // Update the quote token
+        self._set_quote_token(next_quote_token)?;
+        self._emit_quote_token_update_finalized(msg_sender, next_quote_token)
+    }
+
+    // Token operations
+    /// Mints new tokens to specified address
+    fn mint(&mut self, msg_sender: Address, to: Address, amount: U256) -> crate::error::Result<()> {
+        self._mint(msg_sender, to, amount)
```
0xKitsune: Are we able to remove this pattern and just have a single canonical `mint` impl instead of delegating through `_mint`. `mint_with_memo` can then just call `mint` directly rather than both calling the helper.

crates/precompiles/src/tip20_migrated/mod.rs
```diff
@@ -0,0 +1,1347 @@
...
+        amount: U256,
+        memo: B256,
+    ) -> crate::error::Result<()> {
+        self._mint(msg_sender, to, amount)?;
+        self._emit_transfer_with_memo(msg_sender, to, amount, memo)
+    }
+
+    /// Burns tokens from sender's balance and reduces total supply
+    fn burn(&mut self, msg_sender: Address, amount: U256) -> crate::error::Result<()> {
+        self._burn(msg_sender, amount)
```
0xKitsune: Same thing here, see the note about `mint`.

crates/precompiles/src/tip20_migrated/mod.rs
```diff
@@ -0,0 +1,1347 @@
...
+            // Only allow burning from addresses that are blocked from transferring
+            return Err(TIP20Error::policy_forbids().into());
+        }
+
+        self._transfer(from, Address::ZERO, amount)?;
+
+        let total_supply = self.total_supply()?;
+        let new_supply = total_supply
+            .checked_sub(amount)
+            .ok_or(TIP20Error::insufficient_balance())?;
```
0xKitsune: This should be `.ok_or(TempoPrecompileError::under_ovewrflow())`

crates/precompiles/src/tip20_migrated/mod.rs
```diff
@@ -0,0 +1,1347 @@
...
+        }
+
+        // Set default values
+        self._set_supply_cap(U256::MAX)?;
+        self._set_transfer_policy_id(1)?;
+
+        // Initialize roles system and grant admin role
+        self.roles_initialize()?;
+        self.roles_grant_default_admin(admin)
+    }
```
0xKitsune, line 444: All precompiles have an `initialize` function that acts similar to a constructor. Can we add this function to a trait definition to ensure that it is not possible to implement a precompile without the `initalize` function? We should also consider renaming this to `constructor` or similar. Constructor might not be the best name since this function is typically setting account code.

crates/precompiles/src/tip20_migrated/mod.rs
```diff
@@ -0,0 +1,1347 @@
...
+        let new_from_balance = from_balance
+            .checked_sub(amount)
+            .ok_or(TIP20Error::insufficient_balance())?;
+
+        self._set_balances(from, new_from_balance)?;
+
+        let to_balance = self._get_balances(TIP_FEE_MANAGER_ADDRESS)?;
+        let new_to_balance = to_balance
+            .checked_add(amount)
+            .ok_or(TIP20Error::supply_cap_exceeded())?;
```
0xKitsune: This should be updated to `TempoPrecompileError::under_overflow()`. The current error is left over from when we did not have the `TempoPrecompileError`(and was the wrong error in general).

crates/precompiles/src/tip20_migrated/mod.rs
```diff
@@ -0,0 +1,1347 @@
...
+        actual_used: U256,
+    ) -> Result<()> {
+        let from_balance = self._get_balances(TIP_FEE_MANAGER_ADDRESS)?;
+        if refund > from_balance {
+            return Err(TIP20Error::insufficient_balance().into());
+        }
+
+        let new_from_balance = from_balance
+            .checked_sub(refund)
+            .ok_or(TIP20Error::insufficient_balance())?;
```
0xKitsune: Same here can we update to `TempoPrecompileError::under_overflow()`

crates/precompiles/src/tip20_migrated/mod.rs
```diff
@@ -0,0 +1,1347 @@
...
+
+        self._set_balances(TIP_FEE_MANAGER_ADDRESS, new_from_balance)?;
+
+        let to_balance = self._get_balances(to)?;
+        let new_to_balance = to_balance
+            .checked_add(refund)
+            .ok_or(TIP20Error::supply_cap_exceeded())?;
```
0xKitsune: Update to `TempoPrecompileError::under_overflow()`

crates/precompiles/src/tip20_migrated/rewards.rs
```diff
@@ -0,0 +1,907 @@
...
+    ITIP20, ITIP20Rewards, TIP20Error, TIP20Event, TIP20RewardsError,
+};
+use tempo_precompiles_macros::Storable;
+
+use super::{TIP20Token, TIP20Token_ITIP20Rewards};
+
+pub const ACC_PRECISION: U256 = uint!(1000000000000000000_U256);
+
+/// Reward stream data structure occupying 5 consecutive storage slots.
+/// Fields are stored in declaration order, one field per slot.
```
0xKitsune: This does not match Solidity's storage semantics for structs. Solidity will pack values into a single storage slot for structs when possible. If we do not follow the same convention, this might break existing tooling.

crates/precompiles/src/tip20_migrated/rewards.rs
```diff
@@ -0,0 +1,907 @@
...
+
+        // Emit reward recipient set event
+        self._emit_reward_recipient_set(msg_sender, recipient)
+    }
+
+    /// Cancels an active reward stream and refunds remaining tokens.
+    ///
+    /// This function allows the funder of a reward stream to cancel it early,
+    /// stopping future reward distribution and refunding unused tokens.
+    fn cancel_reward(&mut self, msg_sender: Address, stream_id: u64) -> crate::error::Result<U256> {
```
0xKitsune: Can we use `Result` directly rather than `crate::error::Result`.

crates/precompiles/src/tip20_migrated/rewards.rs
```diff
@@ -0,0 +1,907 @@
...
+
+            self.storage.emit_event(
+                self.address,
+                TIP20Event::Transfer(ITIP20::Transfer {
+                    from: token_address,
+                    to: recipient,
+                    amount: accrued,
+                })
+                .into_log_data(),
+            )?;
```
0xKitsune: Can we update event emission to use the helpers generated by the macro?

crates/precompiles/src/tip20_migrated/rewards.rs
```diff
@@ -0,0 +1,907 @@
...
+    /// This function manages the opted-in supply adjustments when tokens are transferred
+    /// between addresses with different reward recipient settings. It returns the net
+    /// change to the opted-in supply.
+    pub fn handle_rewards_on_transfer(
+        &mut self,
+        from: Address,
+        to: Address,
+        amount: U256,
+    ) -> Result<()> {
+        let mut opted_in_delta = alloy::primitives::I256::ZERO;
```
0xKitsune: Lets update to use `I256` directly rather than `alloy::primitives::I256`.

crates/precompiles/src/lib.rs
```diff
@@ -210,12 +212,15 @@ impl LinkingUSDPrecompile {
 }
 
 #[inline]
-fn metadata<T: SolCall>(f: impl FnOnce() -> Result<T::Return>) -> PrecompileResult {
+pub fn metadata<T: SolCall>(f: impl FnOnce() -> Result<T::Return>) -> PrecompileResult {
```
0xKitsune, line 222: We should update `metadata`, `view` `mutate` etc or remove these functions in favor of an approach that supports specifying the gas cost for each precompile call.

Ideally we are able to do this with macro over each contract function. For example:

```rust


pub struct Contract{}

impl Contract {

    #[gas_cost = TRANSFER_GAS_COST]
    pub fn transfer(){
        // --snip--
    }
}

File: crates/precompiles/src/storage/types/mapping.rs (hunk 2 above)
Comment (0xrusowsky) [reply crates/precompiles/src/storage/types/mapping.rs]: unfortunately i was unable to do it... apparently it is a rustc limitation

will look more into it just in case it was a skill issue from my side

File: crates/precompiles/src/storage/types/traits.rs (hunk 4 above)
Comment (0xrusowsky) [reply crates/precompiles/src/storage/types/traits.rs]: afaik we diverge from solidity here, as in solidity +31byte strings are stored in another set of consecutive slots at `keccak(base_slot)`

i kept ur current impl, but we could aim for a full match with solidity if wanted 

note that since our impl is a subset of the solidity one, this non-breaking (just a feature regression where ur evm frontend only supports 31byte strings at most)

File: crates/precompiles/src/tip20_migrated/mod.rs (hunk 7 above)
Comment (0xrusowsky) [reply crates/precompiles/src/tip20_migrated/mod.rs]: mainly due to name colision, for example the low-level `self.set_supply_cap(new_supply_cap: U256)` would collude with the trait fn `fn set_supply_cap(msg_sender: Address, new_supply_cap: U256)`

i will do the change, but at the expense of having to rename variables of the storage layout to avoid the clash.

for instance, we can avoid the problem with:
```rs
    #[map = "supply_cap"]
    max_supply_cap: U256, // slot 14
```
so that the low-level fn becomes: `self.set_max_supply_cap(..)`

crates/precompiles/src/tip20_migrated/mod.rs
0xrusowsky: yes, this is a cool idea to explore.
i kinda had the same thought when migrating the fns

crates/precompiles/src/tip20_migrated/mod.rs
0xrusowsky: yes, my bad.

this is case thanks to the trait impl, when i click tab, i get the fully typed fn signature and forgot to clean-up

crates/precompiles/src/tip20_migrated/mod.rs
0xrusowsky: excellent call

crates/precompiles/src/tip20_migrated/rewards.rs
0xrusowsky: as discussed internally, this should be fixable by changing the storage primitive impl `fn store/load/delete()`

crates/precompiles/src/tip20_migrated/rewards.rs
0xrusowsky: must have forgottten this one, good catch

crates/precompiles/src/tip20_migrated/mod.rs
0xrusowsky: pushed the commit to address this, but lmk if u like it better (or if u prefer the old way)

crates/precompiles/src/tip20_migrated/mod.rs
```diff
@@ -0,0 +1,1304 @@
...
+pub fn token_id_to_address(token_id: u64) -> Address {
+    let mut address_bytes = [0u8; 20];
+    address_bytes[..12].copy_from_slice(&TIP20_TOKEN_PREFIX);
+    address_bytes[12..20].copy_from_slice(&token_id.to_be_bytes());
+    Address::from(address_bytes)
+}
+
+pub fn address_to_token_id_unchecked(address: Address) -> u64 {
+    u64::from_be_bytes(address.as_slice()[12..20].try_into().unwrap())
+}
```
0xKitsune, line 61: Can we also add a function that is `address_to_token_id` which checks if the address is a valid `TIP20`? There are instances throughout the codebase that we use the unchecked version where we also should be checking if the address is TIP20.

crates/precompiles/src/tip20_migrated/mod.rs
```diff
@@ -0,0 +1,1304 @@
...
+        let factory_token_id_counter = TIP20Factory::new(self.storage)
+            .token_id_counter()?
+            .to::<u64>();
+
+        // Ensure the quote token has been deployed (token_id < counter)
+        if new_token_id >= factory_token_id_counter {
+            return Err(TIP20Error::invalid_quote_token().into());
+        }
+
+        self.set_next_quote_token(new_quote_token)?;
```
0xKitsune: Lets update `set_`, `get_` functions to be prefixed with `sstore_`, `sload_`. This is more explicit that variables are being written/loaded from storage.

crates/precompiles/src/tip20_migrated/mod.rs
```diff
@@ -0,0 +1,1304 @@
...
+
+    // Token operations
+    /// Mints new tokens to specified address
+    fn mint(&mut self, msg_sender: Address, to: Address, amount: U256) -> Result<()> {
+        self.check_role(msg_sender, *ISSUER_ROLE)?;
+        let total_supply = self.total_supply()?;
+
+        let new_supply = total_supply
+            .checked_add(amount)
+            .ok_or(TempoPrecompileError::under_overflow())?;
```
0xKitsune: Given that we use this pattern everywhere throughout the precompiles, lets add a small extension trait to make this cleaner. Something like the following

```rust
trait CheckedMath {
    fn checked_add(self, rhs: Self) -> Result<Self, TempoPrecompileError>;

    // --snip-- other operations

}

impl CheckedMath for U256 {
    fn checked_add(self, rhs: Self) -> Result<Self, TempoPrecompileError> {
        self.checked_add(rhs).ok_or(TempoPrecompileError::under_overflow())
    }
}
```

crates/precompiles/src/tip20_migrated/mod.rs
0xKitsune: Id also like to find a way to have the constructor run automatically without having to manually call this function. Similar to how in solidity it simply runs the constructor whenever a contract is deployed. Maybe as a part of the `new()` function, we can check if the code is set already and if not, call the constructor. Also open to other ideas, I havent thought deeply about this yet.

crates/precompiles/src/tip20_migrated/roles.rs
```diff
@@ -0,0 +1,213 @@
...
+        self.emit_role_membership_updated(role, msg_sender, msg_sender, false)
+    }
+
+    fn set_role_admin(&mut self, msg_sender: Address, role: B256, admin_role: B256) -> Result<()> {
+        let current_admin_role = self.get_role_admin_internal(role)?;
+        self.check_role_internal(msg_sender, current_admin_role)?;
+
+        self.set_role_admin_internal(role, admin_role)?;
+        self.emit_role_admin_updated(role, admin_role, msg_sender)
+    }
```
0xKitsune, line 43: Can we add doc comments for all these functions?

crates/precompiles/src/storage/types/traits.rs
```diff
@@ -0,0 +1,618 @@
...
+    ///
+    /// The default implementation sets each slot to zero individually.
+    /// Types may override this for optimized bulk deletion.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the storage write fails.
+    fn delete<S: StorageOps>(storage: &mut S, base_slot: U256) -> Result<()> {
+        for offset in 0..Self::SLOT_COUNT {
+            storage.sstore(base_slot + U256::from(offset), U256::ZERO)?;
```
0xKitsune, line 90: Are we able to update `SLOT_COUNT` to be `U256` rather than `usize`? Seems like this would save us a `U256::from` conversion.

crates/precompiles/src/storage/types/traits.rs
```diff
@@ -0,0 +1,618 @@
...
+    }
+
+    #[inline]
+    fn store<S: StorageOps>(&self, storage: &mut S, base_slot: U256) -> Result<()> {
+        let value = if *self { U256::ONE } else { U256::ZERO };
+        storage.sstore(base_slot, value)
+    }
+}
+
+impl Storable for u64 {
```
0xKitsune: We should have this implemented for all uints, not just `u64` and `u128`. Lets make a macro for this and implement for all types.

crates/precompiles/src/storage/types/traits.rs
```diff
@@ -0,0 +1,618 @@
...
+        Ok(value.to::<Self>())
+    }
+
+    #[inline]
+    fn store<S: StorageOps>(&self, storage: &mut S, base_slot: U256) -> Result<()> {
+        storage.sstore(base_slot, U256::from(*self))
+    }
+}
+
+impl Storable for i16 {
```
0xKitsune: Same as above, we should have this implemented for all ints. Lets make a macro for this and implement for all.

crates/precompiles/src/storage/types/traits.rs
0xKitsune: We should have parity with default EVM storage semantics for all types. We should ensure that these storage abstractions are written in a way where you never have to think about a diff or what could/might break. Lets plan to update this, but make this lower prio if non breaking. Ideally we have this in place before public testnet but if we need to de-scope for more mission critical tasks, this should be OK to patch in a follow on PR. Lets make sure we have an issue to track in the latter case.

crates/precompiles/src/storage/types/traits.rs
```diff
@@ -0,0 +1,618 @@
...
+        self.as_slice()
+    }
+}
+
+impl StorageKey for FixedBytes<4> {
+    #[inline]
+    fn as_storage_bytes(&self) -> &[u8] {
+        self.as_slice()
+    }
+}
```
0xKitsune: Lets implement this for all concrete solidity types (uints, ints, etc.), not just ones we are currently using. We can make a macro for cases where the impl is repeated across types frequently.

crates/precompiles/src/storage/types/traits.rs
```diff
@@ -0,0 +1,618 @@
...
+        }
+
+        BUFFER.with(|buf| {
+            let mut buffer = buf.borrow_mut();
+            *buffer = self.to_be_bytes();
+            // SAFETY: The buffer lives in TLS and we're returning a reference
+            // that cannot outlive this function call. The caller must use it
+            // immediately before any other code can access the TLS buffer.
+            unsafe { std::slice::from_raw_parts(buffer.as_ptr(), 16) }
+        })
```
0xKitsune: @DaniPopes Do we handle similar cases in a better way in `solar`, `alloy` etc?

crates/precompiles/src/storage/types/traits.rs
```diff
@@ -0,0 +1,618 @@
...
+    fn test_u256_round_trip() {
+        let mut storage = HashMapStorageProvider::new(1);
+        let addr = Address::random();
+        let mut contract = TestContract {
+            address: addr,
+            storage: &mut storage,
+        };
+
+        let value = U256::from(12345u64);
+        let slot = U256::from(0);
```
0xKitsune: Lets use `U256::random()` instead.

crates/precompiles/src/storage/types/traits.rs
```diff
@@ -0,0 +1,618 @@
...
+    #[test]
+    fn test_address_round_trip() {
+        let mut storage = HashMapStorageProvider::new(1);
+        let contract_addr = Address::random();
+        let mut contract = TestContract {
+            address: contract_addr,
+            storage: &mut storage,
+        };
+
+        let addr = address!("0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266");
```
0xKitsune: Is there a reason we aren't using `Address::random()`?

crates/precompiles/src/storage/types/traits.rs
```diff
@@ -0,0 +1,618 @@
...
+    #[test]
+    fn test_b256_round_trip() {
+        let mut storage = HashMapStorageProvider::new(1);
+        let addr = Address::random();
+        let mut contract = TestContract {
+            address: addr,
+            storage: &mut storage,
+        };
+
+        let value = B256::from([0x42u8; 32]);
```
0xKitsune: Same here, can we use `B256::random()`? To avoid adding repetitive comments, Ill note here that we should update all tests to use random values where possible/practical to ensure that storage operations and encoding/decoding do not break. We should also prop test round trip conversions.

crates/precompiles/src/storage/types/traits.rs
```diff
@@ -0,0 +1,618 @@
...
+
+        // Positive value
+        let pos = i16::MAX;
+        pos.store(&mut contract, slot).unwrap();
+        assert_eq!(i16::load(&mut contract, slot).unwrap(), pos);
+
+        // Negative value (two's complement)
+        let neg = i16::MIN;
+        neg.store(&mut contract, slot).unwrap();
+        assert_eq!(i16::load(&mut contract, slot).unwrap(), neg);
```
0xKitsune: Lets have separate tests for negative and positive values here. Lets prop test as well.

crates/precompiles/src/lib.rs
```diff
@@ -1,6 +1,7 @@
 //! Tempo precompile implementations.
 #![cfg_attr(not(test), warn(unused_crate_dependencies))]
 #![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
+#![feature(adt_const_params)]
```
0xKitsune: Lets add a lint to deny unchecked arithmetic.

crates/precompiles/src/lib.rs
```diff
@@ -60,9 +62,9 @@ pub const STABLECOIN_EXCHANGE_ADDRESS: Address =
 pub const NONCE_PRECOMPILE_ADDRESS: Address =
     address!("0x4E4F4E4345000000000000000000000000000000");
 
-const METADATA_GAS: u64 = 50;
-const VIEW_FUNC_GAS: u64 = 100;
-const MUTATE_FUNC_GAS: u64 = 1000;
+pub const METADATA_GAS: u64 = 50;
+pub const VIEW_FUNC_GAS: u64 = 100;
+pub const MUTATE_FUNC_GAS: u64 = 1000;
```
0xKitsune, line 62: Commenting here since the snippet below is not currently part of the diff.


This pattern is followed for all precompiles, lets add a macro for this to reduce repetitive code/boilerplate.

```rust

pub struct TIP20FactoryPrecompile;
impl TIP20FactoryPrecompile {
    pub fn create(chain_id: u64) -> DynPrecompile {
        tempo_precompile!("TIP20Factory", |input| TIP20Factory::new(
            &mut EvmPrecompileStorageProvider::new(input.internals, chain_id)
        ))
    }
}

pub struct TIP20Precompile;
impl TIP20Precompile {
    pub fn create(address: Address, chain_id: u64) -> DynPrecompile {
        let token_id = address_to_token_id_unchecked(address);
        tempo_precompile!("TIP20Token", |input| TIP20Token::new(
            token_id,
            &mut EvmPrecompileStorageProvider::new(input.internals, chain_id),
        ))
    }
}
```

crates/precompiles-macros/src/dispatcher.rs
```diff
@@ -0,0 +1,158 @@
...
+                        ::revm::precompile::PrecompileError::Other(
+                            "Invalid selector format".to_string()
+                        )
+                    })?;
+
+                match selector {
+                    #(#match_arms)*
+                    _ => Err(::revm::precompile::PrecompileError::Other(
+                        "Unknown function selector".to_string()
+                    )),
```
0xKitsune, line 64: Lets update these to match the exact errors that Solidity would return here.

crates/precompiles/src/storage/types/traits.rs
DaniPopes: We don't. I don't particularly like this either, I think you can simply return the u8 array as an `impl AsRef<[u8]>` since we only use this as a compile time helper

0xrusowsky: nice, will do. thanks dani!

0xrusowsky: addressed in [`cb35c49` (#615)](https://github.com/tempoxyz/tempo/pull/615/commits/cb35c4932d4392ed1fcac0146dca58cb567eac24)

crates/precompiles/src/tip20_migrated/mod.rs
0xrusowsky: do u foresee any other math utils?

[`8fdd404` (#615)](https://github.com/tempoxyz/tempo/pull/615/commits/8fdd4047fb7fd480d58574689559bc8812c67242)

0xrusowsky: also, maybe i should rename `fn mul_div` to `fn mul_div_checked`?

i think if u don't foresee any other math utils besides those checked math-related, i'll rename to `CheckedMath` as suggested

crates/precompiles/src/tip20_migrated/mod.rs
0xrusowsky: agreed, but what i had in mind was slightly different:

we keep the impl split into several blocks/traits, so that we leverage traits to define the contract's interface.

imo this design is better because it allows us to easily reason about the contract's public interface at a single glance (traits could still leave under `contracts/`)

when we do the impl the blocks, we know that only the trait fns are dispatchable on-chain, the other public fns are only callable within rust code

0xrusowsky: agreed, but what i had in mind was slightly different:

we keep the impl split into several blocks/traits, so that we leverage traits to define the contract's interface.

imo this design is better because it allows us to easily reason about the contract's public interface at a single glance

when we do the impl the blocks, we know that only the trait fns are dispatchable on-chain, the other public fns are only callable within rust code

crates/precompiles/src/storage/types/traits.rs
0xrusowsky: implemented here: [`3e960d5` (#615)](https://github.com/tempoxyz/tempo/pull/615/commits/3e960d5ed830ed3ab2ceffb64cc291a45868a3fe)

crates/precompiles/src/lib.rs
0xrusowsky: granted the `contract` macro's dispatcher impl the ability to use custom gas here: [`a47af3a` (#615)](https://github.com/tempoxyz/tempo/pull/615/commits/a47af3a6ea79bb6a730affc481fab4d185c967bd)

pending ingestion of the `gas` attribute once we drop the `sol!` dependency

crates/precompiles/src/tip20_migrated/mod.rs
0xrusowsky: addressed here: [`6e618ec` (#615)](https://github.com/tempoxyz/tempo/pull/615/commits/6e618eca9039794c85ae26dd5905f012291f0023)

crates/precompiles/src/storage/types/traits.rs
0xrusowsky: addressed in [`1054b81` (#615)](https://github.com/tempoxyz/tempo/pull/615/commits/1054b812c6738bc92a1135d29610329a637391ce)

crates/precompiles/src/storage/types/traits.rs
0xrusowsky: addressed in [`1054b81` (#615)](https://github.com/tempoxyz/tempo/pull/615/commits/1054b812c6738bc92a1135d29610329a637391ce)

crates/precompiles/src/storage/types/traits.rs
0xrusowsky: addressed in [`1054b81` (#615)](https://github.com/tempoxyz/tempo/pull/615/commits/1054b812c6738bc92a1135d29610329a637391ce)

crates/precompiles/src/storage/types/traits.rs
0xrusowsky: addressed in [`1054b81` (#615)](https://github.com/tempoxyz/tempo/pull/615/commits/1054b812c6738bc92a1135d29610329a637391ce)

crates/precompiles/src/storage/types/traits.rs
0xrusowsky: addressed in [`1054b81` (#615)](https://github.com/tempoxyz/tempo/pull/615/commits/1054b812c6738bc92a1135d29610329a637391ce)

crates/precompiles/src/storage/types/traits.rs
0xrusowsky: addressed in [`1054b81` (#615)](https://github.com/tempoxyz/tempo/pull/615/commits/1054b812c6738bc92a1135d29610329a637391ce)

0xrusowsky: switched to proptest in [`c2c06f9` (#615)](https://github.com/tempoxyz/tempo/pull/615/commits/c2c06f9a0e9932a8f0a68f907795f0f29b664e3a)

crates/precompiles/src/tip20_migrated/rewards.rs
0xrusowsky: addressed with [`3e960d5` (#615)](https://github.com/tempoxyz/tempo/pull/615/commits/3e960d5ed830ed3ab2ceffb64cc291a45868a3fe) and [`75df847` (#615)](https://github.com/tempoxyz/tempo/pull/615/commits/75df847fe36e329231ee7afd2591827e1875aa66)

crates/precompiles/src/tip20_migrated/mod.rs
0xrusowsky: addressed in [`f8bdf4c` (#615)](https://github.com/tempoxyz/tempo/pull/615/commits/f8bdf4c29b54b6d7385b058fbddc142c6cb83ed2)

crates/precompiles/src/tip20_migrated/mod.rs
0xrusowsky: addressed in [`f8bdf4c` (#615)](https://github.com/tempoxyz/tempo/pull/615/commits/f8bdf4c29b54b6d7385b058fbddc142c6cb83ed2)

crates/precompiles/src/tip20_migrated/mod.rs
0xrusowsky: addressed in [`f8bdf4c` (#615)](https://github.com/tempoxyz/tempo/pull/615/commits/f8bdf4c29b54b6d7385b058fbddc142c6cb83ed2)

crates/precompiles/src/tip20_migrated/mod.rs
0xrusowsky: addressed in [`f8bdf4c` (#615)](https://github.com/tempoxyz/tempo/pull/615/commits/f8bdf4c29b54b6d7385b058fbddc142c6cb83ed2)

crates/precompiles/src/lib.rs
0xrusowsky: addressed in [`f8bdf4c` (#615)](https://github.com/tempoxyz/tempo/pull/615/commits/f8bdf4c29b54b6d7385b058fbddc142c6cb83ed2)

PR #616, opened 2025-10-31
chore: bump commonware to v0.0.63

SuperFluffy:
Bumps commonware to the latest v0.0.63 release.

Notable changes affecting our codebase: the marshal reporter activity was changed from `Block` to `Update<Block>`. This patch quietly drops `Update::Tip`, but we will make use of it in a follow-up feature.

This patch adds the `UpdateExt` trait to provide utilities `Update::as_block` and `Update::into_block`. These should be upstreamed.

The `Coordinator` type is removed, since commonware is now using its p2p oracle in its place (technically, any type implementing `T: commonware_p2p::Manager`, which the various p2p oracles do).

The simulated p2p oracle had its method `Oracle::register` removed in favor of `Oracle::control` and `Control::register`.
The peer set used in the test is now also explicitly registered on the oracle using `commonware_p2p::Manager::update`.

### DIFF
--- a/crates/e2e/src/lib.rs
+++ b/crates/e2e/src/lib.rs
@@ -14,7 +14,10 @@ use commonware_cryptography::{
     bls12381::{dkg::ops, primitives::variant::MinSig},
     ed25519::{PrivateKey, PublicKey},
 };
-use commonware_p2p::simulated::{self, Link, Network, Oracle};
+use commonware_p2p::{
+    Manager,
+    simulated::{self, Link, Network, Oracle},
+};
 
 use commonware_runtime::{
     Clock, Metrics as _, Runner as _,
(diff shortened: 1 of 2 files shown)


crates/e2e/src/lib.rs
```diff
@@ -91,6 +95,7 @@ pub async fn setup_validators(
     }
     validators.sort();
     signers.sort_by_key(|s| s.public_key());
+    oracle.update(0, validators.clone().into()).await;
```
patrick-ogrady, line 98: Now that the `resolver` operates directly over peer sets (automatically populated by `p2p`'s knowledge of said peer sets), you must now inform p2p of the participants AND configure links.

patrick-ogrady: (tests that don't rely on peer sets will still pass without this configuration although in your repo I suspect always specifying this will make the most sense)

SuperFluffy: Thank you very much. Appreciate the fix and the insight.

PR #618, opened 2025-10-31
fix(consensus): always send finalized tip to execution layer

SuperFluffy:
Always sends the latest finalized tip that the consensus engine knows about to the execution layer. This is to ensure that the execution layer can trigger an e2e backfill and does not rely on the consensus engine trickle feeding blocks to it.

Makes use a the new feature introduced in https://github.com/commonwarexyz/monorepo/pull/2065.

# Follow-up

With this patch we would still trickle feed individual blocks from the consensus engine to the execution layer. A followup will skip this as to not unnecessarily calculate state roots.

### DIFF
--- a/crates/commonware-node/src/consensus/application/executor.rs
+++ b/crates/commonware-node/src/consensus/application/executor.rs
@@ -113,7 +113,7 @@ impl LastCanonicalized {
         this.update_head(height, hash)
     }
 
-    /// UPdates the head height and head block hash to `height` and `hash`.
+    /// Updates the head height and head block hash to `height` and `hash`.
     ///
     /// `head` must be ahead of the latest canonicalized head height. If it is
     /// not, then this is a no-op.


crates/commonware-node/src/consensus/application/executor.rs
```diff
@@ -61,25 +59,74 @@ impl Builder {
...
+        let mut this = self;
+        if height > this.finalized_height {
+            this.finalized_height = height;
+            this.forkchoice.safe_block_hash = hash;
+            this.forkchoice.finalized_block_hash = hash;
+        }
+        this.update_head(height, hash)
+    }
+
+    /// UPdates the head height and head block hash to `height` and `hash`.
```
joshieDo:
```suggestion
    /// Updates the head height and head block hash to `height` and `hash`.
```

PR #619, opened 2025-10-31
feat(consensus): track execution layer sync target and drop consensus blocks below it

SuperFluffy:
This patch tracks the sync target of the execution layer:

If the node sends a Forkchoice Update to update `forkchoice_state.finalized_hash`, and if the execution layer responds with `SYNCING`, the application's executor actor will track the height corresponding to this block hash as the `SYNC TARGET`.

It will subsequently drop all finalized blocks it receives from the marshal actor that are below `SYNC TARGET`.

The application will still send `FCU`s to update `forkchoice_state.head_hash` and the finalized hash.

The application will also send `NEW PAYLOAD` requests to the execution layer for finalized blocks that satisfy `BLOCK.HEIGHT > SYNC TARGET`.

The application will still send `NEW PAYLOAD` requests on blocks the consensus layer requests it to verify.

### DIFF
--- a/crates/commonware-node/src/consensus/application/executor.rs
+++ b/crates/commonware-node/src/consensus/application/executor.rs
@@ -10,7 +10,7 @@
 use std::{sync::Arc, time::Duration};
 
 use alloy_primitives::B256;
-use alloy_rpc_types_engine::{ForkchoiceState, PayloadStatus};
+use alloy_rpc_types_engine::ForkchoiceState;
 use alloy_rpc_types_eth::SyncStatus;
 use commonware_consensus::{
     Block as _,
@@ -19,7 +19,7 @@ use commonware_consensus::{
 
 use commonware_macros::select;
 use commonware_runtime::{ContextCell, FutureExt, Handle, Metrics, Pacer, Spawner, spawn_cell};
-use eyre::{WrapErr as _, bail, ensure, eyre};
+use eyre::{Report, WrapErr as _, bail, ensure, eyre};
 use futures::{
     StreamExt as _,
     channel::{mpsc, oneshot},
@@ -27,7 +27,7 @@ use futures::{
 use reth_provider::BlockNumReader as _;
 use reth_rpc_eth_api::helpers::EthApiSpec as _;
 use tempo_node::{TempoExecutionData, TempoFullNode};
-use tracing::{Level, Span, debug, info, instrument, warn};
+use tracing::{Level, Span, debug, error, info, instrument, warn};
 
 use crate::consensus::{Digest, block::Block};
 
@@ -329,7 +329,7 @@ where
         );
 
         if fcu_response.is_invalid() {
-            return Err(eyre::Report::msg(fcu_response.payload_status)
+            return Err(Report::msg(fcu_response.payload_status)
                 .wrap_err("execution layer responded with error for forkchoice-update"));
         }
 
@@ -524,12 +524,6 @@ where
 enum HeadOrFinalized {
     Head,
     Finalized,
-}
-
-impl HeadOrFinalized {
-    fn is_finalized(&self) -> bool {
-        matches!(self, Self::Finalized)
-    }
 }
 
 impl std::fmt::Display for HeadOrFinalized {


crates/commonware-node/src/consensus/application/executor.rs
```diff
@@ -411,24 +435,50 @@ where
...
+        'new_payload: {
+            match self.sync_target {
+                Some(sync_target) if sync_target.saturating_sub(block.height()) < 65 => {
+                    debug!(
+                        sync_target,
+                        "execution layer has a sync target set, but the \
+                        consensus block is at most 64 heights below it or \
+                        ahead; forwarding it to the execution lasyer",
+                    );
+                }
```
shekhirin: I don't think this will ever get triggered, because you set the `sync_target` only once.

Imagine you're starting a genesis sync:
- Curent Reth block = 0, last consensus finalized block = 10_000, `sync_target` = 10_000
- While Reth is syncing, `sync_target` isn't getting updated because all FCU requests return `SYNCING`
- When Reth finishes syncing the range `0..=10_000`, last consensus finalized block will be in the future by far more than 64 blocks
- New FCU (forkchoice update) for block, say, 11_000 will trigger another sync on Reth side, **but we will not update the `sync_target` Reth never replied `VALID`**

crates/commonware-node/src/consensus/application/executor.rs
```diff
@@ -150,7 +151,14 @@ pub(super) struct Executor<TContext> {
...
     my_mailbox: ExecutorMailbox,
 
+    /// The last forkchoice state executor sent to the execution layer,
+    /// including the heights corresponding to the head and the finalized block
+    /// hashes.
     last_canonicalized: LastCanonicalized,
+
+    /// The execution layer's sync target. This is set on the first
+    /// forkchoice-state that returned SYNCING.
+    sync_target: Option<u64>,
```
klkvr: this is more like `finalize_sync_target`, right?

crates/commonware-node/src/consensus/application/executor.rs
```diff
@@ -411,24 +435,50 @@ where
...
-                query payload status of finalized block",
-            )?;
+        'new_payload: {
+            match self.sync_target {
+                Some(sync_target) if sync_target.saturating_sub(block.height()) < 65 => {
+                    debug!(
+                        sync_target,
+                        "execution layer has a sync target set, but the \
+                        consensus block is at most 64 heights below it or \
+                        ahead; forwarding it to the execution lasyer",
```
klkvr: why do we not drop all blocks that are below the sync target? seems fine to expect reth to handle all of them?

crates/commonware-node/src/consensus/application/executor.rs
SuperFluffy: > New FCU (forkchoice update) for block, say, 11_000 will trigger another sync on Reth side, but we will not update the sync_target Reth never replied VALID

Ah, excellent point! That's a hole in my logic. Fixing it now.

crates/commonware-node/src/consensus/application/executor.rs
SuperFluffy: Yeah, but since there is no other sync target I thought it's fine. Should I fix the comment?

crates/commonware-node/src/consensus/application/executor.rs
```diff
@@ -411,24 +416,58 @@ where
...
+        'new_payload: {
+            match self.execution_node.eth_api().sync_status() {
+                Err(error) => error!(
+                    error = %Report::new(error),
+                    "could not query sync status of execution layer; going to \
+                    send the finalized block as a NEW PAYLOAD request; but \
+                    this really should not happen",
+                ),
+                Ok(SyncStatus::Info(info))
+                    if info.current_block >= block.height().saturating_sub(64) =>
```
SuperFluffy, line 428: So this technically means that there is likely an overlap.

If EL is at 0 and syncs to 1000. If CL sends blocks 0 to 1000 this will likely trigger.

shekhirin, line 428: I'd move this to a constant, maybe we'll change it because Reth's block buffer is 256 right now https://github.com/paradigmxyz/reth/blob/e0489aa1e3965fd4db223241246623920b4ab5ea/crates/engine/primitives/src/config.rs#L17

crates/commonware-node/src/consensus/application/executor.rs
SuperFluffy: Removed this explicit tracking in favor of asking the node for its sync status.

crates/commonware-node/src/consensus/application/executor.rs
SuperFluffy: No longer explicitly tracking the sync target.

Changed the logic to look at `sync_status.current` and forward the consensus layer block if we're either not syncing or if `sync_status.current` is within 64 blocks of `block.height` (will change to 32 after sync with @shekhirin )

crates/commonware-node/src/consensus/application/executor.rs
SuperFluffy: Sync with @shekhirin : going to leave this / not special case around this. Will get a few redundant `NEW PAYLOAD`, but it doesn't seem worth the complexity working around that.

klkvr, line 428: not sure I understand the expected flow here

if EL is currently in backfill sync I believe `currentBlock` will just be stuck at the number that was the tip before backfill started

so it seems unlikely that `info.current_block` will ever be behind the chain updates we send here?

iiuc what we want to fetch here is current sync target? it doesn't seem like we expose it on `sync_status` though

cc @shekhirin @joshieDo please correct me if I'm wrong

shekhirin: It can be behind when the node has just started and consensus is backfilling all blocks from genesis to finalised. We want to send a single FCU with finalised block, and discard all blocks before it without sending a newPayload for each of them.

This is main motivation IIUC cc @SuperFluffy

SuperFluffy: So first the point of this change: replace our trickle-feed CL -> EL mechanism by EL stream sync.

@shekhirin investigated and it seems Lighthouse sends every payload (finalized block) to the EL. But this doesn't seem feasible because our block times are so low, we'd probably fill reth's buffer.

So we need a mechanism to drop those blocks that are covered by stream sync.

Question: what do we use as the check? It seems @shekhirin and @joshieDo assumed that `currentHeight` gets updated during stream sync? Or does it stay fixed and only gets bumped after?

Alternatively, we can use the initial CL-provided finalized tip height that caused the sync, and drop all blocks below that.

Is there another heuristic we can target?

PR #627, opened 2025-10-31
feat: pass extra data header field via payload builder attributes

### LINKED ISSUE #560: Pass extra data header field via payload builder attributes
As a followup to https://github.com/tempoxyz/tempo/pull/538, we want to pass in the DKG data from the consensus layer / DKG actor to execution layer as part of the payload builder attributes. As of https://github.com/tempoxyz/tempo/pull/538, the extraData header is modified on the (sealed) block returned by reth, which means we are changing the block's hash. This prevents us from using future caching-infrastructure for self-built block in the execution layer.

fgimenez:
Closes #560 

- add `extra_data` field to `TempoPayloadBuilderAttributes` as required parameter
- `extra_data` flows through `TempoNextBlockEnvAttributes` and `TempoBlockExecutionCtx`
- application actor queries DKG before building and passes data in attributes
- remove post-build `insert_*` methods

### DIFF
--- a/crates/payload/builder/src/lib.rs
+++ b/crates/payload/builder/src/lib.rs
@@ -34,7 +34,7 @@ use reth_transaction_pool::{
     BestTransactions, BestTransactionsAttributes, TransactionPool, ValidPoolTransaction,
     error::InvalidPoolTransactionError,
 };
-use std::{borrow::Cow, sync::Arc, time::Instant};
+use std::{sync::Arc, time::Instant};
 use tempo_chainspec::TempoChainSpec;
 use tempo_consensus::TEMPO_GENERAL_GAS_DIVISOR;
 use tempo_evm::{TempoEvmConfig, TempoNextBlockEnvAttributes};
@@ -246,18 +246,8 @@ where
 
         let general_gas_limit = parent_header.gas_limit() / TEMPO_GENERAL_GAS_DIVISOR;
 
-        // If extra data is provided in attributes, configure the EVM with it
-        let evm_config = if !attributes.extra_data().is_empty() {
-            Cow::Owned(
-                self.evm_config
-                    .clone()
-                    .with_extra_data(attributes.extra_data().clone()),
-            )
-        } else {
-            Cow::Borrowed(&self.evm_config)
-        };
-
-        let mut builder = evm_config
+        let mut builder = self
+            .evm_config
             .builder_for_next_block(
                 &mut db,
                 &parent_header,
@@ -272,6 +262,7 @@ where
                     },
                     general_gas_limit,
                     timestamp_millis_part: attributes.timestamp_millis_part(),
+                    extra_data: attributes.extra_data().clone(),
                 },
             )
             .map_err(PayloadBuilderError::other)?;
--- a/crates/evm/src/assemble.rs
+++ b/crates/evm/src/assemble.rs
@@ -36,6 +36,7 @@ impl BlockAssembler<TempoEvmConfig> for TempoBlockAssembler {
                 TempoBlockExecutionCtx {
                     inner,
                     general_gas_limit,
+                    extra_data,
                 },
             parent,
             transactions,
@@ -50,8 +51,12 @@ impl BlockAssembler<TempoEvmConfig> for TempoBlockAssembler {
 
         let timestamp_millis_part = evm_env.block_env.timestamp_millis_part;
 
+        // Set extra_data on the inner assembler before building
+        let mut assembler = self.inner.clone();
+        assembler.extra_data = extra_data;
+
         // Delegate block building to the inner assembler
-        let block = self.inner.assemble_block(BlockAssemblerInput::<
+        let block = assembler.assemble_block(BlockAssemblerInput::<
             EthBlockExecutorFactory<TempoReceiptBuilder, TempoChainSpec, TempoEvmFactory>,
         >::new(
             evm_env,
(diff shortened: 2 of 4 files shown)


crates/payload/types/src/attrs.rs
```diff
@@ -24,11 +24,13 @@ impl InterruptHandle {
 /// Container type for all components required to build a payload.
 ///
 /// The `TempoPayloadBuilderAttributes` has an additional feature of interrupting payload.
+/// It also carries optional DKG data to be included in the block's extra_data field.
 #[derive(Debug, Clone)]
 pub struct TempoPayloadBuilderAttributes {
     inner: EthPayloadBuilderAttributes,
     interrupt: InterruptHandle,
     timestamp_millis_part: u64,
+    optional_extra_data: Option<Vec<u8>>,
```
SuperFluffy:
```suggestion
    extra_data: Vec<u8>
```

Or is there a benefit wrapping it in an option?

Also, `Bytes`? Unless it doesn't work with some encoding?

crates/payload/types/src/attrs.rs
```diff
@@ -52,9 +54,28 @@ impl TempoPayloadBuilderAttributes {
             },
             interrupt: InterruptHandle::default(),
             timestamp_millis_part: millis,
+            optional_extra_data: None,
         }
     }
 
+    /// Creates new `TempoPayloadBuilderAttributes` with optional DKG extra data.
+    pub fn with_optional_extra_data(
```
SuperFluffy: Should we just remove these constructors and make the fields public so we can set them at the call site?

Adding more constructors to set an extra field seems unsustainable.

crates/payload/builder/src/lib.rs
```diff
@@ -246,8 +246,18 @@ where
 
         let general_gas_limit = parent_header.gas_limit() / TEMPO_GENERAL_GAS_DIVISOR;
 
-        let mut builder = self
-            .evm_config
+        // If extra data is provided in attributes, configure the EVM with it
+        let evm_config = if let Some(extra_data) = attributes.optional_extra_data() {
```
SuperFluffy: Re my option comment elsewhere - here, we could check if `.is_empty()`

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -497,7 +462,50 @@ impl Inner<Init> {
...
+        let optional_extra_data = if epoch::is_last_height_of_epoch(
+            parent.height() + 1,
+            round.epoch(),
+            self.epoch_length,
+        ) {
+            // At epoch boundary: include public ceremony outcome
+            let outcome = self
+                .state
+                .dkg_manager
+                .get_public_ceremony_outcome(round.epoch())
```
SuperFluffy, line 476: Probably for a followup: we should just ask the dkg actor for something like `get_bytes_to_include(height: u64)` and not have the application worry about the specifics.

I'm not completely sure if that's the right thing though because when verifying the application needs to check at least the extra data contents of the boundary block.

crates/payload/types/src/attrs.rs
mattsse: yeah this should just bet Bytes

mattsse: let's drop the `optional_` and instead document when this can be missing

crates/payload/types/src/attrs.rs
mattsse: I would rather require the arg in the existing new fn

crates/payload/builder/src/lib.rs
```diff
@@ -246,8 +246,18 @@ where
...
+            Cow::Owned(
+                self.evm_config
+                    .clone()
+                    .with_extra_data(Bytes::from(extra_data.clone())),
```
mattsse: I think instead of doing this, we could move this to `TempoNextBlockEnvAttributes`

@klkvr rn, we don't have a way to set the extradata per blockbuilder instance, so I think moving this to the input makes sense here?

crates/payload/types/src/attrs.rs
SuperFluffy: So many unnamed arguments. :-/

crates/payload/builder/src/lib.rs
klkvr: yeah, we should make it part of `TempoBlockExecutionCtx` and use it here https://github.com/tempoxyz/tempo/blob/9f8e1d21003559655e13f673fbb1d39911ead5f4/crates/evm/src/assemble.rs#L36-L39

crates/payload/types/src/attrs.rs
fgimenez: makes sense, changed to `Byte`

crates/payload/types/src/attrs.rs
fgimenez: ok added as required to `new` and remove the new constructor

crates/payload/builder/src/lib.rs
fgimenez: done as part of the refactoring to `Bytes`, still using `is_empty()` in the new type

crates/payload/types/src/attrs.rs
fgimenez: ok, added comment about when `extra_data` can be empty

crates/payload/builder/src/lib.rs
fgimenez: makes sense done here https://github.com/tempoxyz/tempo/pull/627/commits/cd1919bbcd754f2beeca561eebac61e5fa83416a ptal

PR #630, opened 2025-11-01
fix: remove decimal calculation from min order amount

0xKitsune:
Currently we are adjusting for decimals when calculating the minimum order quantity on the `StablecoinExchange`. This PR updates to remove decimals from the calculation since the `StablecoinExchange` assumes `USD_DECIMALS`.

### DIFF
--- a/crates/precompiles/src/stablecoin_exchange/mod.rs
+++ b/crates/precompiles/src/stablecoin_exchange/mod.rs
@@ -29,8 +29,8 @@ use crate::{
 use alloy::primitives::{Address, B256, Bytes, IntoLogData, U256};
 use revm::state::Bytecode;
 
-/// Minimum order size of $100 USD
-pub const MIN_ORDER_AMOUNT: u128 = 100_000_000;
+/// Minimum order size of $10 USD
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -29,15 +29,8 @@ use crate::{
...
-pub const MIN_ORDER_UNITS: u128 = 10;
-
-/// Calculate minimum order amount based on token decimals
-/// Returns MIN_ORDER_UNITS scaled by the token's decimals
-/// For example: 6 decimals -> 10_000_000, 18 decimals -> 10_000_000_000_000_000_000
-pub fn calculate_min_order_amount(decimals: u8) -> u128 {
-    MIN_ORDER_UNITS * 10u128.pow(decimals as u32)
-}
+/// Minimum order size of $100 USD
+pub const MIN_ORDER_AMOUNT: u128 = 100_000_000;
```
danrobinson: Sorry, we just changed this to $10—can you make it `10_000_000`?

PR #632, opened 2025-11-01
feat(cli): add follow arg

mattsse:
pending reth bump with https://github.com/paradigmxyz/reth/pull/19436

### DIFF
--- a/bin/tempo/src/main.rs
+++ b/bin/tempo/src/main.rs
@@ -23,7 +23,7 @@ use futures::{
     future::{FusedFuture, pending},
 };
 use reth_ethereum::cli::{Cli, Commands};
-use reth_node_builder::NodeHandle;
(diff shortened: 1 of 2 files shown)


bin/tempo/src/main.rs
```diff
@@ -47,6 +47,9 @@ struct TempoArgs {
     /// Start the node without consensus
     #[arg(long)]
     pub no_consensus: bool,
+    /// Follow a remote node
+    #[arg(long, requires = "no_consensus")]
+    pub follow: bool,
```
0xKitsune, line 52: Is there a case where a node should run with `--no-consensus` but not `--follow`? Should `--no-consensus` just be updated to follow consensus if `--debug` is not passed?

Zygimantass: yep, or maybe `--follow` to replace `--no-consensus`. I prefer that, but don't have a super strong opinion

PR #636, opened 2025-11-02
feat(consensus): switch from discovery p2p to lookup p2p

SuperFluffy:
This patch changes the consensus p2p network from discovery to lookup. Where before, one or more known bootstrapper nodes were provided, now all nodes have to be known at startup in order to communicate with each other.

This change is a requirement for dynamic validator sets using on-chain lookup and is done in this patch to de-risk https://github.com/tempoxyz/tempo/pull/585.

As part of this patch the following changes were done to the consensus config file:

1. the `bootstrappers` entry was removed.
2. `listen_port` was removed
3. `listen_addr` was added
4. `dialable_addr` was added

Our commonware x reth e2e tests use a simulated network, and so we are not testing this part of the flow. 

However, I spun up a local network of nodes using this and the nodes were able to connect and produce blocks.


crates/commonware-node/src/lib.rs
```diff
@@ -40,9 +37,12 @@ pub async fn run_consensus_stack(
         .await
         .wrap_err("failed to start network")?;
 
-    oracle
-        .update(0, config.peers.keys().cloned().collect())
-        .await;
+    let all_resolved_peers = resolve_all_peers(&config.peers)
```
SuperFluffy, line 40: We don't check if our pubkey is among the peers. It's not necessary for lookup p2p.

PR #641, opened 2025-11-03
chore(consensus): set static peer set dynamically

SuperFluffy:
This patch demonstrates that a new set of peers can be registered from within one of tempo actors and demonstrates how we will be able to test dynamic reconfiguration.

Instead of registering the peer set before the consensus engine starts, the peers and their resolved socket addresses are now registered after the engine starts.

At the behavior level nothing has changed: we still use a static peer set.

The main achievement of this patch is changing the `crates/e2e` tests to use the `commonware_p2p::authenticated::lookup` network, the same as is used in the main binary. A few tests needed to be commented out (those concerned with link failures), but the primary tests are still persent.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -236,13 +236,13 @@ vergen = "9"
 vergen-git2 = "1"
 
 [patch.crates-io]
-commonware-broadcast = { version = "0.0.63", git = "https://github.com/superfluffy/monorepo", rev = "76a6e0b1a05a66e109c245480c6991a8eebbe4eb" }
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -234,3 +234,15 @@ test-case = "3"
 # build deps
 vergen = "9"
 vergen-git2 = "1"
+
+[patch.crates-io]
+commonware-broadcast = { version = "0.0.63", git = "https://github.com/superfluffy/monorepo", rev = "5e455d83de519432a304e655615c28d19c7ed0d7" }
```
SuperFluffy, line 239: Move this to the tempo org?

PR #644, opened 2025-11-03
feat(node): Implement `policy` endpoints using log filtering

RomanHodulak:
Part of #549

Adds a function that can traverse the logs from emitted events.

Implements `policy_getAddresses` using log filtering to extract data and construct the response.

https://tempo-docs-git-jxom-rpc-tempoxyz.vercel.app/rpc/policy_getAddresses


crates/node/src/rpc/policy/logs.rs
```diff
@@ -0,0 +1,110 @@
...
+        + SpawnBlocking
+        + RpcNodeCoreExt<Provider: BlockIdReader, Pool: TransactionPool>
+        + EthApiTypes
+        + 'static,
+{
+    pub async fn addresses_using_logs(
+        &self,
+        params: AddressesParams,
+    ) -> RpcResult<AddressesResponse> {
+        let mut whitelist = HashSet::new();
```
fgimenez, line 37: nit: we are always scanning all whitelist events, no matter of the policy type, maybe we could have a match upfront? like:
```rust
 let policy_data = query_policy_data(params.policy_id).await?;

  match policy_data.policy_type {
      PolicyType::WHITELIST => {
          // Only scan WhitelistUpdated events
          let filter = Filter::new()
              .select(0u64..)
              .event_signature(WhitelistUpdated::SIGNATURE_HASH)
              .topic1(U256::from(params.policy_id));
          // ... build whitelist
      }
      PolicyType::BLACKLIST => {
          // Only scan BlacklistUpdated events  
          let filter = Filter::new()
              .select(0u64..)
              .event_signature(BlacklistUpdated::SIGNATURE_HASH)
              .topic1(U256::from(params.policy_id));
          // ... build blacklist
      }
  }
``` 
wdyt?

crates/node/src/rpc/policy/logs.rs
```diff
@@ -0,0 +1,110 @@
...
+
+        let filter = Filter::new()
+            .select(0u64..)
```
fgimenez, line 40: this filters from genesis? if so maybe we should try to add some type of caching

PR #645, opened 2025-11-03
fix: more precise block size control during block building

klkvr:
Based on https://github.com/tempoxyz/tempo/pull/581

Changes block building to correctly account for subblocks and system transactions sizes when ensuring that total block size does not exceed the EIP-7934 limit

### DIFF
--- a/crates/payload/builder/src/lib.rs
+++ b/crates/payload/builder/src/lib.rs
@@ -110,7 +110,7 @@ impl<Provider: ChainSpecProvider> TempoPayloadBuilder<Provider> {
     fn build_seal_block_txs(
         &self,
         block_env: &BlockEnv,
-        subblocks: Vec<RecoveredSubBlock>,
+        subblocks: &[RecoveredSubBlock],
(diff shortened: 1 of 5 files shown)


crates/payload/builder/src/lib.rs
```diff
@@ -301,10 +301,25 @@ where
...
+        let block_gas_limit: u64 = parent_header.gas_limit();
+        let shared_gas_limit = block_gas_limit / TEMPO_SHARED_GAS_DIVISOR;
+        let non_shared_gas_limit = block_gas_limit - shared_gas_limit;
         let general_gas_limit =
             (parent_header.gas_limit() - shared_gas_limit) / TEMPO_GENERAL_GAS_DIVISOR;
 
+        let mut cumulative_gas_used = 0;
+        let mut non_payment_gas_used = 0;
+        // initial block size usage - size of withdrawals plus 1Kb of overhead for the block header
+        let mut block_size_used = attributes.withdrawals().length() + 1024;
```
mattsse, line 322: this is sufficient, even with the additional header fields

PR #647, opened 2025-11-03
chore(tests): connect execution nodes on e2e tests via reth p2p

joshieDo:
To add any pipeline sync test we need to make sure that the e2e setup nodes are connected on reth p2p


crates/e2e/src/execution_runtime.rs
```diff
@@ -175,6 +178,31 @@ pub struct ExecutionNode {
...
+        self.node
+            .network
+            .peers_handle()
+            .add_peer(other_record.id, other_record.tcp_addr());
+
+        other
+            .node
+            .network
+            .peers_handle()
+            .add_peer(self_record.id, self_record.tcp_addr());
```
mattsse: this works, but both peers would race to connect to each other

so we can make use of adding one and await the connection event from both peers?

mattsse: can also use add_trusted_peer?

crates/e2e/src/execution_runtime.rs
```diff
@@ -175,6 +185,29 @@ pub struct ExecutionNode {
...
+    /// Connect peers bidirectionally.
+    pub async fn connect_peer(&self, other: &Self) {
+        let self_record = self.node.network.local_node_record();
+        let other_record = other.node.network.local_node_record();
+
+        self.node
+            .network
+            .add_trusted_peer(other_record.id, other_record.tcp_addr());
+
+        match self.node.network.event_listener().next().await {
```
mattsse: the listener should be obtained before adding peers, so that this catches all events generated by the peer actions

crates/e2e/src/execution_runtime.rs
```diff
@@ -203,13 +242,20 @@ pub async fn launch_execution_node<P: AsRef<Path>>(
...
 ) -> eyre::Result<ExecutionNode> {
     let node_config = NodeConfig::new(chainspec())
-        .with_unused_ports()
+        .with_network(NetworkArgs {
+            discovery: DiscoveryArgs {
+                disable_discovery: true,
+                ..DiscoveryArgs::default()
+            },
+            ..NetworkArgs::default()
+        })
```
mattsse: can we change this to `.apply(|c| {
   c.network.disable_discovery = true;
   c
})`

crates/e2e/src/lib.rs
```diff
@@ -102,16 +102,27 @@ pub async fn setup_validators(
...
 
+    for i in 0..how_many {
         let node = execution_runtime
             .spawn_node_blocking(&format!("node-{i}"))
             .expect("must be able to spawn nodes on the runtime");
 
+        // ensure EL p2p connectivity for backfill syncs
+        for existing_node in &execution_nodes {
+            existing_node.connect_peer(&node).await;
+        }
```
mattsse, line 123: this only established a 1 to many connection, should we peer all nodes with each other instead?

crates/e2e/src/execution_runtime.rs
```diff
@@ -175,6 +185,35 @@ pub struct ExecutionNode {
...
+            .add_trusted_peer(other_record.id, other_record.tcp_addr());
+
+        match events.next().await {
+            Some(NetworkEvent::Peer(PeerEvent::PeerAdded(_))) => (),
+            ev => panic!("Expected a peer added event, got: {ev:?}"),
+        }
+
+        match events.next().await {
+            Some(NetworkEvent::ActivePeerSession { .. }) => (),
+            ev => panic!("Expected an active peer session event, got: {ev:?}"),
```
mattsse, line 206: these should be emitted in this exact order so the panic is fine here but we could also skip them

like

https://github.com/paradigmxyz/reth/blob/338c1d4393d4b30cdf492bcfaf396c45691f3541/crates/net/network/src/test_utils/testnet.rs#L759-L767

PR #650, opened 2025-11-03
feat(precompile): Precompile gas checks for sstore/sload/log operations

rakita:
Gas cost for precompile operations. Loading of account, sload, sstore and set code

### DIFF
--- a/crates/precompiles/src/tip_account_registrar/mod.rs
+++ b/crates/precompiles/src/tip_account_registrar/mod.rs
@@ -50,7 +50,7 @@ impl<'a, S: PrecompileStorageProvider> TipAccountRegistrar<'a, S> {
         }
 
         // EIP-7702 gas cost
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/tip4217_registry/dispatch.rs
```diff
@@ -6,6 +6,10 @@ use crate::tip4217_registry::{ITIP4217Registry, TIP4217Registry};
 
 impl Precompile for TIP4217Registry {
     fn call(&mut self, calldata: &[u8], _msg_sender: Address) -> PrecompileResult {
+        // self.storage
+        //     .deduct_gas(input_cost(calldata.len()))
+        //     .map_err(|_| PrecompileError::OutOfGas)?;
```
klkvr, line 11: this should either be removed or uncommented

rakita: It will get removed in https://github.com/tempoxyz/tempo/pull/689

PR #652, opened 2025-11-04
chore(tests): ensure pipeline sync was triggered

joshieDo:
On top of https://github.com/tempoxyz/tempo/pull/647

Checks `metrics` to ensure pipeline was executed.

Test is kinda slow right now - how can we speed up the consensus chain? 100ms block time or smth @SuperFluffy ?

### DIFF
--- a/crates/e2e/src/tests/backfill.rs
+++ b/crates/e2e/src/tests/backfill.rs
@@ -4,18 +4,18 @@ use commonware_macros::test_traced;
 use commonware_p2p::simulated::Link;
 use commonware_runtime::{
     Clock, Runner as _,
-    deterministic::{self, Runner},
+    deterministic::{self, Context, Runner},
 };
 use reth_ethereum::storage::BlockNumReader;
-use reth_node_metrics::recorder::{PrometheusRecorder, install_prometheus_recorder};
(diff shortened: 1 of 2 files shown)


crates/e2e/src/tests/backfill.rs
```diff
@@ -7,8 +7,9 @@ use commonware_runtime::{
     deterministic::{self, Runner},
 };
 use reth_ethereum::storage::BlockNumReader;
+use reth_node_metrics::recorder::install_prometheus_recorder;
 
-use crate::{ExecutionRuntime, Setup, setup_validators};
+use crate::{ExecutionRuntime, Setup, get_pipeline_runs, setup_validators};
 
 #[test_traced]
 fn validator_can_join_later() {
```
SuperFluffy: Can we do a second test for this? Keep the smaller "trickle backfill" as it is right now, but then also add the pipeline sync one that specifically targets the metric?

crates/node/Cargo.toml
```diff
@@ -58,11 +58,13 @@ serde.workspace = true
 jsonrpsee.workspace = true
 
 [dev-dependencies]
+tempo-e2e.workspace = true
```
SuperFluffy, line 61: Should this create really depend on e2e?

PR #653, opened 2025-11-04
fix(rewards-registry): return early from finalize_stream on invalid state

fgimenez:
ref https://tempoxyz.slack.com/archives/C09KCGR4LQ4/p1762157514260509?thread_ts=1761930432.103249&cid=C09KCGR4LQ4

### DIFF
--- a/crates/precompiles/src/tip20/rewards.rs
+++ b/crates/precompiles/src/tip20/rewards.rs
@@ -394,13 +394,20 @@ impl<'a, S: PrecompileStorageProvider> TIP20Token<'a, S> {
             return Ok(());
         }
 
+        let total_rps = self.get_total_reward_per_second()?;
+
+        // Return early if state is invalid (rate_decrease > total_rps)
+        // This can happen if streams were cancelled or state was corrupted
+        if rate_decrease > total_rps {
+            return Ok(());
+        }
+
         self.accrue(U256::from(end_time))?;
 
-        let total_rps = self
-            .get_total_reward_per_second()?
+        let new_total_rps = total_rps
             .checked_sub(rate_decrease)
             .ok_or(TempoPrecompileError::under_overflow())?;
-        self.set_total_reward_per_second(total_rps)?;
+        self.set_total_reward_per_second(new_total_rps)?;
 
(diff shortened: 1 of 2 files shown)


crates/precompiles/src/tip20_rewards_registry/mod.rs
```diff
@@ -194,18 +198,52 @@ impl<'a, S: PrecompileStorageProvider> TIP20RewardsRegistry<'a, S> {
...
+                // Update the stream index mapping
+                let stream_key = keccak256((token, next_timestamp).abi_encode());
+                let index_slot = mapping_slot(stream_key, slots::STREAM_INDEX);
+                self.storage
+                    .sstore(self.address, index_slot, U256::from(i))?;
+            }
+
+            // Update the array length
+            self.storage
+                .sstore(self.address, array_slot, U256::from(failed_tokens.len()))?;
```
0xKitsune: Im not sure if this will work since we are updating the `last_updated_timestamp` to `current_timestamp` but rebuilding the array for `next_timestamp` which will be in the past the next time this gets processed. Rather than match here (and try catch in the reference impl), it looks like we could update to [return early here](https://github.com/tempoxyz/docs/blob/main/specs/src/TIP20.sol#L708-L709) and make this call infallible instead?

crates/precompiles/src/tip20/rewards.rs
```diff
@@ -394,13 +394,20 @@ impl<'a, S: PrecompileStorageProvider> TIP20Token<'a, S> {
             return Ok(());
         }
 
+        let total_rps = self.get_total_reward_per_second()?;
+
+        // Return early if state is invalid (rate_decrease > total_rps)
+        // This can happen if streams were cancelled or state was corrupted
+        if rate_decrease > total_rps {
+            return Ok(());
+        }
```
0xKitsune, line 403: The only check in the reference contract is `require(rateDecrease > 0, "No streams to finalize");`. Should this be removed or what is this doing?

fgimenez: yep not needed, we already have the required logic and we should match this behaviour in sol :+1:

PR #654, opened 2025-11-04
feat(block): execute rewards registry system tx at start of block

fgimenez:
ref https://tempoxyz.slack.com/archives/C09KCGR4LQ4/p1762157514260509?thread_ts=1761930432.103249&cid=C09KCGR4LQ4

Changes block execution order to place rewards registry system transaction at the start of the block, allowing regular transactions to execute after it.

the execution order is now:
  1. rewards registry `finalizeStreams` (start of block)
  2. regular transactions
  3. fee manager `executeBlock` (end of block)
  4. stablecoin DEX `executeBlock` (end of block)

the changes:
- add `build_start_block_txs()` to create rewards registry system tx
- execute start-of-block system txs after `apply_pre_execution_changes()` but before regular transactions
- update validation logic to enforce ordering: rewards registry must come first, regular txs in middle, fee manager/dex at end

### DIFF
--- a/crates/evm/src/block.rs
+++ b/crates/evm/src/block.rs
@@ -355,13 +355,18 @@ where
             self.validate_system_tx(tx)
         } else if let Some(tx_proposer) = tx.subblock_proposer() {
             match self.section {
+                BlockSection::StartOfBlock {
+                    seen_tip20_rewards_registry,
+                } if !seen_tip20_rewards_registry => Err(BlockValidationError::msg(
+                    "TIP20 rewards registry system transaction was not seen",
+                )),
+                BlockSection::GasIncentive | BlockSection::System { .. } => {
+                    Err(BlockValidationError::msg("subblock section already passed"))
+                }
                 BlockSection::StartOfBlock { .. } | BlockSection::NonShared => {
                     Ok(BlockSection::SubBlock {
                         proposer: tx_proposer,
                     })
-                }
-                BlockSection::GasIncentive | BlockSection::System { .. } => {
-                    Err(BlockValidationError::msg("subblock section already passed"))
                 }
                 BlockSection::SubBlock { proposer } => {
                     if proposer == tx_proposer
@@ -379,6 +384,11 @@ where
             }
         } else {
             match self.section {
+                BlockSection::StartOfBlock {
+                    seen_tip20_rewards_registry,
+                } if !seen_tip20_rewards_registry => Err(BlockValidationError::msg(
+                    "TIP20 rewards registry system transaction was not seen",
+                )),
                 BlockSection::StartOfBlock { .. } | BlockSection::NonShared => {
                     if gas_used > self.non_shared_gas_left
                         || (!tx.is_payment() && gas_used > self.non_payment_gas_left)
--- a/crates/payload/builder/src/lib.rs
+++ b/crates/payload/builder/src/lib.rs
@@ -412,13 +412,11 @@ where
         ));
 
         // Execute start-of-block system transactions (rewards registry finalize)
-        let start_block_txs = self.build_start_block_txs(builder.evm().block());
-        for tx in &start_block_txs {
+        for tx in self.build_start_block_txs(builder.evm().block()) {
             block_size_used += tx.inner().length();
-        }
-        for system_tx in start_block_txs {
+
             builder
-                .execute_transaction(system_tx)
+                .execute_transaction(tx)
                 .map_err(PayloadBuilderError::evm)?;
         }
 
--- a/crates/node/tests/it/backfill.rs
+++ b/crates/node/tests/it/backfill.rs
@@ -6,16 +6,13 @@ use alloy::{
 };
 use alloy_eips::{BlockNumberOrTag, Encodable2718};
 use alloy_network::TxSignerSync;
-use alloy_primitives::{Address, Sealable};
+use alloy_primitives::Address;
 use alloy_rpc_types_engine::ForkchoiceState;
-use futures::StreamExt;
 use reth_e2e_test_utils::wallet::Wallet;
 use reth_node_api::EngineApiMessageVersion;
 use reth_node_metrics::recorder::install_prometheus_recorder;
-use reth_primitives_traits::{AlloyBlockHeader as _, Block as _, transaction::TxHashRef};
-use reth_provider::BlockReaderIdExt;
+use reth_primitives_traits::{AlloyBlockHeader as _, transaction::TxHashRef};
 use tempo_chainspec::spec::TEMPO_BASE_FEE;
-use tempo_primitives::{TempoTxEnvelope, transaction::envelope::TEMPO_SYSTEM_TX_SIGNATURE};
 
 /// Test that verifies backfill sync works correctly.
 ///
@@ -94,39 +91,14 @@ async fn test_backfill_sync() -> eyre::Result<()> {
         let payload = node1.advance_block().await?;
 
         // Verify the transaction was included
-        let block_hash = payload.block().hash();
         let block_number = payload.block().number();
 
-        // Get head block from notifications stream and verify the tx is present
-        let head = node1.canonical_stream.next().await.unwrap();
-
-        // Find the first non-system transaction
-        let user_tx = head
-            .tip()
-            .body()
-            .transactions()
-            .find(|tx| {
-                !matches!(tx, TempoTxEnvelope::Legacy(signed) if signed.signature() == &TEMPO_SYSTEM_TX_SIGNATURE)
-            })
-            .expect("Should have at least one user transaction");
-
-        assert_eq!(user_tx.tx_hash().as_slice(), tx_hash.as_slice());
-
-        loop {
-            // wait for the block to commit
-            tokio::time::sleep(std::time::Duration::from_millis(20)).await;
-            let latest_block = node1
-                .inner
-                .provider
-                .block_by_number_or_tag(BlockNumberOrTag::Latest)?;
-            if let Some(latest_block) = latest_block
-                && latest_block.header().number() == block_number
-            {
-                // make sure the block hash we submitted via FCU engine api is the new latest block
-                assert_eq!(latest_block.header().hash_slow(), block_hash);
-                break;
-            }
-        }
+        let block = provider1
+            .get_block(block_number.into())
+            .full()
+            .await?
+            .unwrap();
+        assert!(block.into_transactions_vec()[1].inner.tx_hash() == &tx_hash);
 
         if block_number % 10 == 0 {
             println!("Advanced to block {block_number}");


crates/consensus/src/lib.rs
```diff
@@ -201,27 +201,39 @@ impl Consensus<Block> for TempoConsensus {
...
+                return Err(ConsensusError::Other(
+                    "Block must start with system txs".to_string(),
+                ));
+            }
+            if tx.to().unwrap_or_default() != expected_to {
+                return Err(ConsensusError::Other(
+                    "Invalid start-of-block system tx order".to_string(),
+                ));
+            }
+        }
```
0xKitsune: nit: we could also update to just get the first tx rather than `let start_txs = &transactions[..TEMPO_SYSTEM_TX_START_OF_BLOCK.len()];`. However it might make sesen to leave this here if we intend to have additional system txs at the top of the block.

crates/consensus/src/lib.rs
```diff
@@ -201,27 +201,39 @@ impl Consensus<Block> for TempoConsensus {
...
+                return Err(ConsensusError::Other(
+                    "Block must end with system txs".to_string(),
+                ));
+            }
             if tx.to().unwrap_or_default() != expected_to {
-                return Err(ConsensusError::Other("Invalid system tx order".to_string()));
+                return Err(ConsensusError::Other(
+                    "Invalid end-of-block system tx order".to_string(),
+                ));
             }
```
0xKitsune: Looks like these are doing the same thing with the same types. Can we make a helper function called `validate_system_txs` or similar for this?

crates/evm/src/block.rs
```diff
@@ -87,51 +87,64 @@ where
...
                 Err(BlockValidationError::msg(
-                    "duplicate stablecoin TIP20 rewards registry system transaction",
+                    "duplicate stablecoin DEX system transaction",
+                ))
+            } else if !self.seen_fee_manager_system_tx {
+                Err(BlockValidationError::msg(
+                    "stablecoin DEX system transaction must come after fee manager system transaction",
                 ))
             } else {
                 Ok(())
```
0xKitsune: nit: I think we can probably clean this up a bit. This will get a bit verbose if we continue to add system txs. Something like the following to start could work. We can also do this in a separate PR.


```rust
enum SysTxKind {
    RewardsRegistry,
    FeeManager,
    StablecoinDex,
}

impl SysTxKind {
    fn encode(&self, block: &[u8]) -> Bytes {
        match self {
            Self::RewardsRegistry => ITIP20RewardsRegistry::finalizeStreamsCall{}.abi_encode(),
            Self::FeeManager => IFeeManager::executeBlockCall.abi_encode(),
            Self::StablecoinDex => IStablecoinExchange::executeBlockCall{}.abi_encode(),
        }.into_iter().chain(block.iter().copied()).collect()
    }
}
```

crates/consensus/src/lib.rs
```diff
@@ -201,27 +201,39 @@ impl Consensus<Block> for TempoConsensus {
...
+        if transactions.len() < TEMPO_SYSTEM_TX_COUNT {
             return Err(ConsensusError::Other(
                 "Block must contain Tempo system txs".to_string(),
             ));
         }
 
-        // Valdiate that the sequence of system txs is correct
-        for (tx, expected_to) in system_txs.into_iter().zip(TEMPO_SYSTEM_TX_ADDRESSES) {
+        // Validate start-of-block system transactions (first transaction: rewards registry)
+        let start_txs = &transactions[..TEMPO_SYSTEM_TX_START_OF_BLOCK.len()];
```
0xKitsune: Can you remind me why we need to process this at the top of the block rather than with the rest of the system txs?

crates/node/tests/it/utils.rs
```diff
@@ -335,3 +341,57 @@ fn default_attributes_generator(timestamp: u64) -> TempoPayloadBuilderAttributes
...
+                .provider
+                .block_by_number_or_tag(BlockNumberOrTag::Latest)?
+                && latest_block.header().number() == block_number
+            {
+                // make sure the block hash we submitted via FCU engine api is the new latest
+                // block using an RPC call
+                assert_eq!(latest_block.header().hash_slow(), block_hash);
+                break;
+            }
+        }
```
0xKitsune: Why do we need this logic? Are tests broken otherwise?

crates/node/tests/it/utils.rs
```diff
@@ -335,3 +341,57 @@ fn default_attributes_generator(timestamp: u64) -> TempoPayloadBuilderAttributes
...
+        // get head block from notifications stream and verify the tx has been pushed to the
+        // pool is actually present in the canonical block
+        let head = self.canonical_stream.next().await.unwrap();
+
+        // Find the first non-system transaction
+        let user_tx = head
+            .tip()
+            .body()
+            .transactions()
+            .find(|tx| !matches!(tx, TempoTxEnvelope::Legacy(signed) if signed.signature() == &TEMPO_SYSTEM_TX_SIGNATURE))
```
0xKitsune: IIUC this will skip the first system tx but not the final system txs?

crates/consensus/src/lib.rs
fgimenez: simplified for now, we can add back the loop if needed

crates/consensus/src/lib.rs
fgimenez: good point, done

crates/evm/src/block.rs
fgimenez: yep good call, added

crates/consensus/src/lib.rs
fgimenez: these changes basically follow @samczsun suggestion in https://tempoxyz.slack.com/archives/C09KCGR4LQ4/p1762157514260509?thread_ts=1761930432.103249&cid=C09KCGR4LQ4 `also the reward registry finalize call should happen top of block, so there should be regular transactions after it (unlike the fee manager and dex system txs)`

crates/node/tests/it/utils.rs
fgimenez: yep tests were timing out/failing without it, see https://github.com/tempoxyz/tempo/actions/runs/19065472787/job/54454923087 and  https://github.com/tempoxyz/tempo/actions/runs/19067656696/job/54462152520

crates/node/tests/it/utils.rs
fgimenez: it will stop at the first user tx which is hat we want, skipping the first system tx, no need to skip the final ones, or am i missing something?

crates/node/tests/it/utils.rs
```diff
@@ -335,3 +341,57 @@ fn default_attributes_generator(timestamp: u64) -> TempoPayloadBuilderAttributes
...
+    async fn assert_new_block_tempo(
+        &mut self,
+        tip_tx_hash: B256,
+        block_hash: B256,
+        block_number: u64,
+    ) -> eyre::Result<()>;
+}
+
+impl TempoNodeTestExt for NodeHelperType<TempoNode> {
+    async fn assert_new_block_tempo(
```
klkvr: we only use `assert_new_block` once so I'd rather just change logic here to fetch the block directly https://github.com/tempoxyz/tempo/blob/5b8682c934a756155f75c15b54f694fd707d7d86/crates/node/tests/it/backfill.rs#L72-L80

crates/payload/builder/src/lib.rs
```diff
@@ -351,21 +380,28 @@ where
...
+
         let mut block_transactions_rlp_length = 0;
         let is_osaka = chain_spec.is_osaka_active_at_timestamp(attributes.timestamp());
 
         let execution_start = Instant::now();
         let mut payment_transactions = 0;
         while let Some(pool_tx) = best_txs.next() {
             // ensure we still have capacity for this transaction
-            if cumulative_gas_used + pool_tx.gas_limit() > non_shared_gas_limit {
+            if cumulative_gas_used + pool_tx.gas_limit() > block_gas_limit {
```
klkvr: seems like rebase went wrong here, this is not equivalent

crates/payload/builder/src/lib.rs
```diff
@@ -374,7 +410,7 @@ where
             // If the tx is not a payment and will exceed the general gas limit
             // mark the tx as invalid and continue
             if !pool_tx.transaction.is_payment()
-                && non_payment_gas_used + pool_tx.gas_limit() > general_gas_limit
+                && cumulative_gas_used + pool_tx.gas_limit() > general_gas_limit
```
klkvr: same here

crates/evm/src/block.rs
```diff
@@ -133,88 +136,102 @@ where
...
+                    seen_subblocks_signatures,
+                ),
+                _ => (false, false, false),
+            };
+
+        if to == TIP20_REWARDS_REGISTRY_ADDRESS {
+            return Err(BlockValidationError::msg(
+                "rewards registry system transaction must be at start of block",
+            ));
+        } else if to == TIP_FEE_MANAGER_ADDRESS {
```
klkvr:
```suggestion
        if to == TIP_FEE_MANAGER_ADDRESS {
```

should be fine to let this be handled in the fallback branch bellow

crates/evm/src/block.rs
```diff
@@ -370,6 +389,16 @@ where
...
+                    // Transition from start-of-block to non-shared section for regular transactions
+                    if gas_used > self.non_shared_gas_left
+                        || (!tx.is_payment() && gas_used > self.non_payment_gas_left)
+                    {
+                        Ok(BlockSection::GasIncentive)
+                    } else {
+                        Ok(BlockSection::NonShared)
+                    }
+                }
                 BlockSection::NonShared => {
```
klkvr:
```suggestion
                BlockSection::StartOfBlock { seen_tip20_rewards_registry } && !seen_tip20_rewards_registry => {
                    Err(BlockValidationError::msg(
                        "cannot include non-system transaction first in the block",
                    )
                }
                BlockSection::StartOfBlock { .. } | BlockSection::NonShared => {
```

we need make sure you can't send non-system/non-subblock tx before tip20 registry was sent

crates/evm/src/block.rs
```diff
@@ -348,6 +364,9 @@ where
             self.validate_system_tx(tx)
         } else if let Some(tx_proposer) = tx.subblock_proposer() {
             match self.section {
+                BlockSection::StartOfBlock { .. } => Err(BlockValidationError::msg(
+                    "cannot include subblock transactions before rewards registry",
+                )),
                 BlockSection::GasIncentive | BlockSection::System { .. } => {
```
klkvr:
```suggestion
                BlockSection::StartOfBlock { seen_tip20_rewards_registry } && !seen_tip20_rewards_registry => Err(BlockValidationError::msg(
                    "cannot include subblock transactions before rewards registry",
                )),
BlockSection::GasIncentive | BlockSection::System { .. } => {
```

below we also need to handle `StartOfBlock { .. }` in the same branch as `NonShared`

crates/evm/src/lib.rs
```diff
@@ -14,6 +14,10 @@ pub use error::TempoEvmError;
 pub mod evm;
 use std::{borrow::Cow, sync::Arc};
 
+// tempo_consensus is used in context.rs under #[cfg(feature = "rpc")]
```
klkvr: should it be an optional dependency?

crates/node/tests/it/utils.rs
fgimenez: ok inlined here

crates/payload/builder/src/lib.rs
fgimenez: good catch, recovered

crates/payload/builder/src/lib.rs
fgimenez: done

crates/evm/src/lib.rs
fgimenez: nope, `tempo_consensus` is used unconditionally in lib.rs, removed

crates/evm/src/block.rs
```diff
@@ -348,6 +360,14 @@ where
             self.validate_system_tx(tx)
         } else if let Some(tx_proposer) = tx.subblock_proposer() {
             match self.section {
+                BlockSection::StartOfBlock {
+                    seen_tip20_rewards_registry,
+                } if !seen_tip20_rewards_registry => Err(BlockValidationError::msg(
+                    "rewards registry system transaction must be first in the block",
+                )),
+                BlockSection::StartOfBlock { .. } => Err(BlockValidationError::msg(
```
klkvr: this needs to be in the same branch as `NonShared` 

block is allowed to only contain subblocks, so we don't want to return an error here

crates/payload/builder/src/lib.rs
```diff
@@ -402,6 +411,13 @@ where
                 .map(|gasprice| gasprice as u64),
         ));
 
+        // Execute start-of-block system transactions (rewards registry finalize)
+        for system_tx in self.build_start_block_txs(builder.evm().block()) {
+            builder
+                .execute_transaction(system_tx)
+                .map_err(PayloadBuilderError::evm)?;
+        }
```
klkvr, line 421: we also need to account for those transactions in `block_size_used`, like in https://github.com/tempoxyz/tempo/pull/645

crates/evm/src/block.rs
fgimenez: makes sense, done here and for regular transactions

crates/payload/builder/src/lib.rs
fgimenez: ok done ptal

Issue #651, opened 2025-11-04
extract rpc related types from network to alloy-tempo


mattsse:
we have all rpc related types and alloy trait impls here:

https://github.com/tempoxyz/tempo/blob/main/crates/node/src/network.rs

this depends on primitive types, which itself requires some reth impls (already feature gated)

but this we should move to `crates/alloy-tempo`

## TODO

* extract rpc/alloy specific trait impls to a dedicated crate
* this can get additional features for provider api ext etc.

PR #656, opened 2025-11-04
chore: pull out `TempoNetwork` into own crate

onbjerg:
Pulls out `TempoNetwork` and `TempoTransactionRequest` into a new crate.

As a follow up, we should make most EVM-related things optional for consumers that are not the Tempo node. Other follow ups would be things like provider extension traits.

Closes https://github.com/tempoxyz/tempo/issues/651

### DIFF
--- a/crates/alloy/src/lib.rs
+++ b/crates/alloy/src/lib.rs
@@ -1,3 +1,8 @@
+//! Tempo types for Alloy.
+
+#![cfg_attr(not(test), warn(unused_crate_dependencies))]
+#![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
+
 mod network;
 pub use network::*;
 


crates/alloy/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+rust-version.workspace = true
+publish.workspace = true
+
+[lints]
+workspace = true
+
+[dependencies]
+tempo-primitives.workspace = true
+tempo-evm.workspace = true # todo: make this optional
+tempo-revm.workspace = true # todo: make this optional
```
mattsse, line 16: makes sense

more features are fine

crates/alloy/src/lib.rs
```diff
@@ -0,0 +1,4 @@
+mod network;
```
mattsse, line 6: can we add the same header that we have in other crates: one line doc +  cfgs

https://github.com/tempoxyz/tempo/blob/81799da741aa107794842530a9f7ccfba336b5eb/crates/evm/src/lib.rs#L1-L4

onbjerg: oh yeah ofc

PR #658, opened 2025-11-04
deps: make reth and tempo deps optional in `tempo-alloy`

onbjerg:
Follow up on #656 

Feature gates some dependencies that most consumers outside of the Tempo node won't use


crates/alloy/src/rpc/request.rs
```diff
@@ -152,6 +158,7 @@ impl TempoTransactionRequest {
     }
 }
 
+#[cfg(feature = "tempo-compat")]
```
mattsse: can we group this into a mod compat module?

crates/alloy/Cargo.toml
```diff
@@ -35,3 +34,12 @@ default = []
 asm-keccak = [
     "alloy-primitives/asm-keccak",
 ]
+reth-compat = [
+    "dep:reth-evm",
+    "dep:reth-rpc-convert"
+]
+tempo-compat = [
```
mattsse: yeah I think we only need one?

PR #662, opened 2025-11-04
deps: no default features for tempo crates

onbjerg:
Even though a bunch of deps in all tempo crates were marked as optional, they were also part of the `default` feature, which means all of them were being added anyway for downstream consumers. As an example, depending on `tempo-alloy` at the moment compiles `reth-cli-commands` and everything downstream (most of Reth).

Adding `default-features = false` to the workspace for the tempo crates fixes this and depending on `tempo-alloy` now only compiles two reth crates related to primitives, which is not that bad.


bin/tempo/Cargo.toml
```diff
@@ -14,13 +14,13 @@ workspace = true
...
-tempo-consensus.workspace = true
-tempo-evm.workspace = true
-tempo-faucet.workspace = true
+tempo-node = { workspace = true, features = ["default"] }
+tempo-commonware-node = { workspace = true }
+tempo-commonware-node-config = { workspace = true }
+tempo-chainspec = { workspace = true, features = ["default"] }
+tempo-consensus = { workspace = true }
+tempo-evm = { workspace = true }
+tempo-faucet = { workspace = true  }
```
klkvr:
```suggestion
tempo-node = { workspace = true, features = ["default"] }
tempo-commonware-node.workspace = true
tempo-commonware-node-config.workspace = true
tempo-chainspec = { workspace = true, features = ["default"] }
tempo-consensus.workspace = true
tempo-evm.workspace = true
tempo-faucet.workspace = true
```

PR #663, opened 2025-11-04
feat: scale gasprice by 1e12

### LINKED ISSUE #629: Support sub-cent transactions
Right now minimum basefee value is 1 wei and minimum transaction gas usage is 21.000 which means that it's not possible to send a transaction that will pay less than $0.021

We can address this by scaling the basefee by some factor meaning that 1 wei basefee would actually translate to less than 1 wei balance spending per gas.

This likely requires proper specification that would mention specific factor used, how this should affect existing codepaths relying on basefee (e.g `BASEFEE` opcode), and how rounding should be handled

klkvr:
Closes #629 

changes all relevant code to calculate transaction cost as `(gas * gas_price).div_ceil(1e12)`

### DIFF
--- a/crates/precompiles/src/tip_fee_manager/mod.rs
+++ b/crates/precompiles/src/tip_fee_manager/mod.rs
@@ -237,9 +237,6 @@ impl<'a, S: PrecompileStorageProvider> TipFeeManager<'a, S> {
         if !refund_amount.is_zero() {
             let mut tip20_token = TIP20Token::from_address(fee_token, self.storage);
             tip20_token.transfer_fee_post_tx(fee_payer, refund_amount, actual_spending)?;
-            println!(
-                "transfer_fee_post_tx {fee_token:?} {fee_payer:?} {refund_amount:?} {actual_spending:?}"
-            );
         }
 
         // Execute fee swap and track collected fees


crates/primitives/src/transaction/mod.rs
```diff
@@ -16,3 +16,15 @@ pub use account_abstraction::{
 pub use alloy_eips::eip7702::Authorization;
 pub use envelope::{TempoTxEnvelope, TempoTxType, TempoTypedTransaction};
 pub use fee_token::{FEE_TOKEN_TX_TYPE_ID, TxFeeToken};
+
+use alloy_primitives::U256;
+
+/// Factor by which we scale the gas price for gas spending calculations.
+pub const TEMPO_GAS_PRICE_SCALING_FACTOR: u128 = 1_000_000_000_000;
```
0xKitsune: nit: Should we make this a `U256`? It looks like we are converting to `U256` in most places we are using this cost.

crates/chainspec/src/genesis/adagio.json
```diff
@@ -32,7 +32,7 @@
     "pragueTime": 0,
     "adagioTime": 0
   },
-  "baseFeePerGas": "0x2c",
+  "baseFeePerGas": "0x28048c5ec000",
```
0xKitsune, line 35: We should adjust this to ensure transfers are < ~$0.0001.

@danrobinson to confirm this value.

crates/chainspec/src/spec.rs
```diff
@@ -13,7 +13,7 @@ use std::sync::{Arc, LazyLock};
 use tempo_contracts::DEFAULT_7702_DELEGATE_ADDRESS;
 use tempo_primitives::TempoHeader;
 
-pub const TEMPO_BASE_FEE: u64 = 44;
+pub const TEMPO_BASE_FEE: u64 = 44_000_000_000_000;
```
mattsse, line 16: can we add the unit here and document and also add the hex value in docs (helps me find stuff)

this reduces mental load when reviewing this

crates/chainspec/src/genesis/adagio.json
danrobinson: <$0.001 (tenth of a cent, not hundredth)! And should be slightly less

klkvr: should we go with 0.04 wei per gas? would mean that min tx cost is `21000 * 0.04 = $0.00084`

danrobinson: What's the gas cost of a transaction that includes a TIP-20 token transfer, though?

klkvr: looks like around 23k gas

https://andante.cloud.blockscout.com/tx/0x65cb275cafb2e91241bd9645d4cb4adb29331d1d1bc17f44c0158e9a07bcfb77

klkvr: likely to change soon once we start charging for storage reads properly, with policy/paused checks might be closer to 50k

PR #665, opened 2025-11-04
feat(cli): add follow flag and default follow URL

Zygimantass:
adds a follow flag and automatically sets `--debug.rpc-consensus-url` if the follow flag is set


bin/tempo/src/main.rs
```diff
@@ -42,11 +42,19 @@ use tokio_util::either::Either;
 // TODO: migrate this to tempo_node eventually.
 #[derive(Debug, Clone, PartialEq, Eq, clap::Args)]
 struct TempoArgs {
-    /// Start the node without consensus
+    /// Start the node without consensus and follow an RPC node
     #[arg(long)]
-    pub no_consensus: bool,
+    pub follow: bool,
```
mattsse, line 47: this is better

PR #669, opened 2025-11-04
feat!(precompiles): storage primitives

0xrusowsky:
## Motivation

ref: https://linear.app/tempoxyz/issue/CHAIN-27/precompile-refactor

## Solution

this PR implements storage primitives to safely interact with the EVM storage model.

### Storage primitives

the base of the storage primitives is the `trait Storable`, which is an API for loading/storing/deleting types with solidity-like encoding.

https://github.com/tempoxyz/tempo/blob/7de6140b85b715698d73a6f2dfd11200498612cb/crates/precompiles/src/storage/types/mod.rs#L56

the trait has been implemented for all rust/alloy primitives via macros:

- https://github.com/tempoxyz/tempo/blob/rus/storage-primitives/crates/precompiles-macros/src/storable_primitives.rs

it has also been manually implemented for `String/Bytes` and `Vec<T: Storable>`:

- [Bytes and String impl](https://github.com/tempoxyz/tempo/blob/rus/storage-primitives/crates/precompiles/src/storage/types/bytes_like.rs)

- [dynamic arrays impl](https://github.com/tempoxyz/tempo/blob/rus/storage-primitives/crates/precompiles/src/storage/types/vec.rs)

- [relevant slot-packing utilities](https://github.com/tempoxyz/tempo/blob/rus/storage-primitives/crates/precompiles/src/storage/packing.rs)

### Storage macros

on top of the storage primitives, there is a an abstraction layer, which leverages macros, that allows for easy interaction with the EVM storage model and ensures no slot collisions at compile-time.

the abstraction layer is built on top of the `trait SlotId` + 2 wrapper types: `Slot` and `Mapping`, which are capable of performing `trait StorageOperations` (sstore and sload).

https://github.com/tempoxyz/tempo/blob/7de6140b85b715698d73a6f2dfd11200498612cb/crates/precompiles/src/storage/types/slot.rs#L51

https://github.com/tempoxyz/tempo/blob/7de6140b85b715698d73a6f2dfd11200498612cb/crates/precompiles/src/storage/types/mapping.rs#L30

#### [Storable macro](https://github.com/tempoxyz/tempo/blob/rus/storage-primitives/crates/precompiles-macros/src/storable.rs)

derive macro that impls `Storable` to user-defined structs (if all their fields are `Storable`) and assigns slots to each of them (generates a mod with slot info for each field).
Slot assignment has solidity-like packing into account.

#### [Contract macro](https://github.com/tempoxyz/tempo/blob/rus/storage-primitives/crates/precompiles-macros/src/lib.rs#L30-L62)

transforms a struct that represents a storage layout into a contract with helper methods to easily read/write/clear its fields. Its packing and encoding schemes aim to be an exact representation of the storage model used by Solidity. Like `Storable` it also generates a mod with slot info for each field.

```rs
#[derive(Debug, Storable)]
struct Foo {
    field1: U256,
    field2: Address,
    field3: u64,
}

#[contract]
pub struct Layout {
    pub field_a: U256,                   // Auto: slot 0
    #[slot(5)]
    pub field_b: U256,                   // Explicit: slot 5
    pub field_c: Foo                     // Auto: slot 1-2 
    pub field_c: Mapping<Address, U256>, // Auto: slot 3
    #[slot(bar)]
    pub field_d: U256,                   // Explicit: slot keccak("bar")
}
```

## Testing

- i've implemented some manual unit tests to verify the logic, and used Claude to expand. Lmk if u feel like some area needs further coverage.
- slot testing with proptest is tricky as slots are consts generated at compile-time, so we can only fuzz the storage values, but not the layout
- to increase the confidence of the impl i have tried to mimic what `solar` does with solc's and the ui-tests. So i created an [integration folder that runs solc](https://github.com/tempoxyz/tempo/blob/rus/storage-primitives/crates/precompiles/tests/storage_tests/solidity/mod.rs#L1) with the `storage-layout` flag to ensure that we match its slot assignation schema


crates/contracts/src/precompiles/tip20.rs
0xrusowsky, line 1: ignore, as this is leftover from the `contract` macro refactor

crates/node/tests/it/tip20.rs
0xrusowsky, line 1: ignore, as this is leftover from the contract macro refactor

crates/precompiles/src/tip20/dispatch.rs
0xrusowsky, line 1: ignore, as this is leftover from the contract macro refactor

crates/precompiles/src/tip20/rewards.rs
0xrusowsky, line 1: ignore, as this is leftover from the contract macro refactor

crates/precompiles/src/storage/types/mapping.rs
```diff
@@ -0,0 +1,595 @@
...
+
+    // Test SlotId implementations
+    struct TestSlot0;
+    impl SlotId for TestSlot0 {
+        const SLOT: U256 = U256::ZERO;
+    }
+
+    struct TestSlot1;
+    impl SlotId for TestSlot1 {
+        const SLOT: U256 = U256::from_limbs([1, 0, 0, 0]);
```
0xKitsune: nit: I beleive you should be able to use `uint!(1_U256)` here.

crates/precompiles/src/storage/types/mapping.rs
```diff
@@ -0,0 +1,595 @@
...
+    }
+
+    struct TestSlot2;
+    impl SlotId for TestSlot2 {
+        const SLOT: U256 = U256::from_limbs([2, 0, 0, 0]);
```
0xKitsune: nit: same thing here, you should be able to use `uint!(2_U256)` here.

crates/precompiles/src/storage/types/slot.rs
```diff
@@ -0,0 +1,399 @@
...
+        let loaded = MaxSlot::read(&mut contract).unwrap();
+        assert_eq!(loaded, value);
+    }
+
+    #[test]
+    fn test_slot_read_write_u256() {
+        let mut storage = HashMapStorageProvider::new(1);
+        let mut contract = setup_test_contract(&mut storage);
+        type TestSlot = Slot<U256, TestSlot1>;
+        let test_value = U256::from(12345u64);
```
0xKitsune: nit: can we use `U256::random` here?

crates/precompiles/src/storage/types/primitives.rs
```diff
@@ -0,0 +1,1061 @@
...
+    const BYTE_COUNT: usize = 1;
+}
+
+impl Storable<1> for bool {
+    const SLOT_COUNT: usize = 1;
+
+    #[inline]
+    fn load<S: StorageOps>(storage: &mut S, base_slot: U256) -> Result<Self> {
+        let value = storage.sload(base_slot)?;
+        Ok(value != U256::ZERO)
```
0xKitsune: nit: consider using `!value.is_zero()`.

crates/precompiles/src/storage/types/primitives.rs
```diff
@@ -0,0 +1,1061 @@
...
+    }
+
+    #[inline]
+    fn to_evm_words(&self) -> Result<[U256; 1]> {
+        Ok([if *self { U256::ONE } else { U256::ZERO }])
+    }
+
+    #[inline]
+    fn from_evm_words(words: [U256; 1]) -> Result<Self> {
+        Ok(words[0] != U256::ZERO)
```
0xKitsune: nit: consider using `words[0].is_zero()` here.

crates/precompiles/src/storage/packing.rs
```diff
@@ -0,0 +1,326 @@
...
+    bytes: usize,
+) -> Result<T> {
+    // Calculate how many bits to shift right to align the value
+    let shift_bits = (32 - offset - bytes) * 8;
+
+    // Create mask for the value's bit width
+    let mask = if bytes == 32 {
+        U256::MAX
+    } else {
+        (U256::from(1) << (bytes * 8)) - U256::from(1)
```
0xKitsune: nit: use `U256::ONE`

crates/precompiles-macros/tests/layout.rs
```diff
@@ -0,0 +1,739 @@
...
+//    ```
+//
+// 3. Mismatched slot_count:
+//    ```rust
+//    #[contract]
+//    pub struct Layout {
+//        #[slot_count(5)]       // Wrong! TestBlock::SLOT_COUNT = 3
+//        pub block: TestBlock,  // ERROR: compile-time validation fails
+//    }
+//    ```
```
klkvr: how do we prevent this? and if we do, why do we need the explicit `slot_count` attribute at all?

0xrusowsky: it will throw a compile-time error, the LSP/rustc will tell u that it should be 3.

i know its sub-optimal, but i haven't been able to figure out how to do it without this workaround yet

0xrusowsky: with that being said, since i was trying to move fast, i didn't revisit it.
will give it another go, i think it should be addressable and was probably a skill issue

0xrusowsky: not needed anymore

crates/precompiles-macros/src/layout.rs
```diff
@@ -0,0 +1,725 @@
...
+    match &allocated.kind {
+        FieldKind::Direct => {
+            // Manual slots and dynamic types are never packed (always at offset 0)
+            if matches!(allocated.assigned_slot, SlotAssignment::Manual(_))
+                || is_dynamic_type(field_ty)
+            {
+                quote! {
+                    impl<'a, S: crate::storage::PrecompileStorageProvider> #struct_name<'a, S> {
+                        #[inline]
+                        fn #getter_name(&mut self) -> crate::error::Result<#field_ty> {
```
mattsse, line 210: these are all private, are these intended to remain private?

crates/precompiles-macros/src/lib.rs
mattsse, line 1: not a blocker but maintaining macros is a pita if you haven't written them, so what would be helpful here is a dev explainer which section generates which code

crates/precompiles/src/storage/mod.rs
```diff
@@ -31,7 +39,35 @@ pub trait PrecompileStorageProvider {
...
 }
+
+/// Trait providing access to a contract's address and storage provider.
+///
+/// Abstracts the common pattern of contracts needing both an address and a mutable reference
+/// to a storage provider. It is automatically implemented by the `#[contract]` macro.
+pub trait ContractStorage {
+    type Storage: PrecompileStorageProvider;
+    fn address(&self) -> Address;
+    fn storage(&mut self) -> &mut Self::Storage;
```
mattsse, line 64: these need some docs

crates/precompiles/src/storage/types/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+/// Implementations must ensure that:
+/// - Round-trip conversions preserve data: `load(store(x)) == Ok(x)`
+/// - `N` accurately reflects the number of slots used
+/// - `store` and `load` access exactly `N` consecutive slots
+/// - `to_evm_words` and `from_evm_words` produce/consume exactly `N` words
+pub trait Storable<const N: usize>: Sized + StorableType {
+    /// The number of consecutive storage slots this type occupies.
+    ///
+    /// Must be equal to `N`, and is provided as a convenient type-level access constant.
+    const SLOT_COUNT: usize;
```
mattsse, line 60: if this must be set to N then we can do

`const SLOT_COUNT: usize = N;` here?

crates/precompiles/src/storage/types/mapping.rs
```diff
@@ -0,0 +1,725 @@
...
+/// Mappings use Solidity's storage layout:
+/// - Base slot: `Id::SLOT`
+/// - Actual slot for key `k`: `keccak256(k || base_slot)`
+///
+/// # Compile-Time Guarantees
+///
+/// - Different mappings have distinct types even with same K,V
+/// - Base slot encoded in type system via `Id::SLOT`
+#[derive(Debug, Clone, Copy)]
+pub struct Mapping<K, V, Id: SlotId> {
```
mattsse: should we just name Id `Base`?

crates/precompiles/src/storage/types/primitives.rs
```diff
@@ -0,0 +1,811 @@
...
+    fn load<S: StorageOps>(storage: &mut S, base_slot: U256) -> Result<Self> {
+        let value = storage.sload(base_slot)?;
+        Ok(value != U256::ZERO)
+    }
+
+    #[inline]
+    fn store<S: StorageOps>(&self, storage: &mut S, base_slot: U256) -> Result<()> {
+        let value = if *self { U256::ONE } else { U256::ZERO };
+        storage.sstore(base_slot, value)
+    }
```
mattsse, line 41: this means we can also invert this, like instead of doing

```
            b.store(&mut contract, base_slot)?;
            let loaded = bool::load(&mut contract, base_slot)?;
```

we can do

```
let val = false;
contract.write(&val, base_slot);
let loaded: bool = contract.load(base_slot);
```

I find this a bit more intuitive than perform ops on the value.

this can be done with an extension trait I think that autoimpls

crates/precompiles-macros/src/layout.rs
0xrusowsky: yes, they are meant to only be internal fns for the contract implementation, and we should have proper public fns

crates/precompiles-macros/src/lib.rs
0xrusowsky: sounds good, will do!

crates/precompiles-macros/src/layout.rs
```diff
@@ -0,0 +1,725 @@
...
+/// Generate the constructor method
+pub(crate) fn gen_constructor(name: &Ident) -> proc_macro2::TokenStream {
+    quote! {
+        impl<'a, S: crate::storage::PrecompileStorageProvider> #name<'a, S> {
+            #[inline(always)]
+            fn _new(address: ::alloy::primitives::Address, storage: &'a mut S) -> Self {
+                Self {
+                    address,
+                    storage,
+                    msg_sender: ::alloy::primitives::Address::ZERO,
```
klkvr: is this ever used?

crates/precompiles/src/storage/types/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+/// - Slot 2: `base_slot + 2`
+///
+/// # Safety
+///
+/// Implementations must ensure that:
+/// - Round-trip conversions preserve data: `load(store(x)) == Ok(x)`
+/// - `N` accurately reflects the number of slots used
+/// - `store` and `load` access exactly `N` consecutive slots
+/// - `to_evm_words` and `from_evm_words` produce/consume exactly `N` words
+pub trait Storable<const N: usize>: Sized + StorableType {
```
klkvr, line 56: why do we need the `N` generic here? assuming this trait can only be implemented once per type it seems fine to remove the generic and infer it from `SLOT_COUNT`?

crates/precompiles-macros/src/layout.rs
```diff
@@ -0,0 +1,725 @@
...
+            let other_slot_id = format_ident!("{}", other_field.slot_id_name());
+            let other_field_name = &other_field.info.name;
+
+            // Generate slot count expressions
+            let current_count_expr =
+                generate_slot_count_expr(&current_field.kind, &current_field.info.ty);
+            let other_count_expr =
+                generate_slot_count_expr(&other_field.kind, &other_field.info.ty);
+
+            // Generate a debug assertion that checks for overlap
```
klkvr, line 689: hmm seems like those checks are never invoked?

crates/precompiles/src/storage/types/slot.rs
```diff
@@ -0,0 +1,399 @@
...
+///
+/// The sealed pattern prevents external crates from implementing this trait,
+/// ensuring type safety and compile-time slot guarantees.
+mod sealed {
+    use alloy::primitives::U256;
+
+    pub trait SlotId {
+        const SLOT: U256;
+    }
+}
```
klkvr, line 22: what's the motivation here? cant't we just use constants from `mod slots` instead of a new type and trait impl per field?

crates/precompiles-macros/src/layout.rs
```diff
@@ -0,0 +1,725 @@
...
+                            };
+                            (slot_expr, quote! { 0 })
+                        } else {
+                            // Previous field is a primitive
+                            if allocated.kind.is_mapping()
+                                || is_dynamic_type(&allocated.info.ty)
+                                || is_array_type(&allocated.info.ty)
+                                || is_custom_struct(&allocated.info.ty)
+                            {
+                                // Current is non-primitive: must start on slot boundary
```
klkvr: this makes me wonder if it could make sense to unify the way layout is handled between primitive and non-primitive types, e.g each type could define something like

```rust
const LAYOUT: Layout;

enum Layout {
    /// Fits within a single slot, defines length in bytes
    Bytes(usize),
    /// Requires one or more slots, defines number of slots
    Slots(usize),
}
```

One benefit of this would be that we wouldn't have to special-case primitives like this https://github.com/tempoxyz/tempo/blob/e0b283c3f974d5995243165cb416fc6dddcadc3b/crates/precompiles-macros/src/utils.rs#L250-L261 which became pretty annoying with reth storage macros

crates/precompiles-macros/src/layout.rs
```diff
@@ -0,0 +1,725 @@
...
+                        let slot_expr = quote! {
+                            ::alloy::primitives::U256::from_limbs([#(#limbs),*])
+                        };
+                        (slot_expr, quote! { 0 })
+                    } else {
+                        // Previous was also auto - compute based on its slot/offset/bytes
+                        if prev.kind.is_mapping() || is_dynamic_type(&prev.info.ty) {
+                            // Previous field occupies exactly 1 full slot
+                            let slot_expr = quote! {
+                                #prev_slot.saturating_add(::alloy::primitives::U256::from_limbs([1, 0, 0, 0]))
```
klkvr: this can be just `U256::ONE`

crates/precompiles-macros/src/storable.rs
```diff
@@ -0,0 +1,580 @@
...
+            #(#field_byte_sizes)*
+            #(#field_layouts)*
+
+            pub const SLOT_COUNT: usize = #last_slot_const + 1;
+        }
+    }
+}
+
+/// Generate the `fn load()` implementation with unpacking logic.
+fn gen_load_impl(fields: &[(&Ident, &Type)], packing: &Ident) -> TokenStream {
```
klkvr: I believe this currently always loads all fields of a struct from storage?

I'm wondering if there's a way here to somehow support Solidity behavior when doing `myStruct.myField` or `myStructs[key].myField` would only do a single storage read

crates/precompiles-macros/src/storable.rs
```diff
@@ -0,0 +1,580 @@
...
+    let load_fields = fields.iter().enumerate().map(|(idx, (name, ty))| {
+        let slot_const = Ident::new(&format!("FIELD_{idx}_SLOT"), proc_macro2::Span::call_site());
+        let offset_const = Ident::new(&format!("FIELD_{idx}_OFFSET"), proc_macro2::Span::call_site());
+        let bytes_const = Ident::new(&format!("FIELD_{idx}_BYTES"), proc_macro2::Span::call_site());
+
+        // Struct, dynamic type, and array fields always use `load()` directly (never packed)
+        if is_array_type(ty) || is_dynamic_type(ty) || is_custom_struct(ty) {
```
klkvr: it would be really nice if we could have a helpre type that would encapsulate that context, so that we could do e.g `parsed_type.can_be_packed()`

it seems like we do a lot of such `is_*` checks a lot of which are somewhat redundant

crates/precompiles-macros/src/storable_primitives.rs
```diff
@@ -0,0 +1,961 @@
...
+    match strategy {
+        StorableConversionStrategy::Unsigned => {
+            quote! {
+                impl Storable<1> for #type_path {
+                    const SLOT_COUNT: usize = 1;
+
+                    #[inline]
+                    fn load<S: StorageOps>(storage: &mut S, base_slot: U256) -> Result<Self> {
+                        let value = storage.sload(base_slot)?;
+                        Ok(value.to::<Self>())
```
klkvr, line 83: we must handle overflow properly here

crates/precompiles-macros/src/storable_primitives.rs
```diff
@@ -0,0 +1,961 @@
...
+        StorableConversionStrategy::SignedRust(unsigned_type) => {
+            quote! {
+                impl Storable<1> for #type_path {
+                    const SLOT_COUNT: usize = 1;
+
+                    #[inline]
+                    fn load<S: StorageOps>(storage: &mut S, base_slot: U256) -> Result<Self> {
+                        let value = storage.sload(base_slot)?;
+                        // Read as unsigned then cast to signed (preserves bit pattern)
+                        Ok(value.to::<#unsigned_type>() as Self)
```
klkvr, line 139: same here

crates/precompiles-macros/src/layout.rs
0xrusowsky: good catch, `msg_sender` was used for the auto-dispatcher impl, but i forgot to delete it when it was descoped

crates/precompiles/src/storage/types/slot.rs
0xrusowsky: yeah that makes, i think i overcomplicated myself...

for context, initially i didn't generate `mod slots` and i was embedding the slot number into the actual `Slot/Mapping` struct using the u256 limbs as consts.
then i generated the slots for backwards compatibility.
lastly Daniel told me if it was possible to use U256 directly, but i was unable, and i ended up finding this workaround... which in retrospective we could skip now that we have the slots lol

crates/precompiles-macros/src/layout.rs
0xrusowsky: oh, that's interesting! i will think about it, looks promising

crates/precompiles-macros/src/storable.rs
0xrusowsky: good call, we should definitely do it.
@0xKitsune pointed out the same issue for `Vec<T>`

crates/precompiles-macros/src/layout.rs
```diff
@@ -0,0 +1,725 @@
...
+        base_slot: U256,
+        /// Whether this field can participate in slot packing (primitive types only)
+        is_primitive: bool,
+    },
+}
+
+/// Allocate slots to fields (explicit + auto-assignment)
+pub(crate) fn allocate_slots(fields: &[FieldInfo]) -> syn::Result<Vec<AllocatedField<'_>>> {
+    let mut allocated_fields = Vec::new();
+    let mut last_auto_slot = U256::ZERO;
```
klkvr, line 50: this is more like `next_auto_slot` I think

crates/precompiles-macros/src/storable.rs
```diff
@@ -0,0 +1,580 @@
...
+    let combined = quote! {
+        #expanded
+        #array_impls
+    };
+
+    Ok(combined)
+}
+
+/// Generate a compile-time module that calculates the packing layout.
+fn gen_packing_module(fields: &[(&Ident, &Type)], mod_ident: &Ident) -> TokenStream {
```
klkvr, line 345: could we not unify this with `gen_packing_constants_for_slots_module`?

crates/precompiles-macros/src/lib.rs
```diff
@@ -0,0 +1,409 @@
...
+/// - Rust integers: u8-u128, i8-i128
+/// - Alloy integers: U8-U256, I8-I256
+/// - Address
+/// - FixedBytes<20>, FixedBytes<32>
+///
+/// Each array gets:
+/// - `StorableType` impl with `BYTE_COUNT = SLOT_COUNT * 32`
+/// - `Storable<N>` impl where N is computed from element packing
+#[proc_macro]
+pub fn storable_arrays(_input: TokenStream) -> TokenStream {
```
klkvr, line 317: do we need to store fixed-sized arrays anywhere? my impression was that we only use dynamic but mabe I'm wrong

it would be nicer if there was a way for us to have blanket impls for arrays

crates/precompiles-macros/src/utils.rs
```diff
@@ -0,0 +1,382 @@
...
+///
+/// Arrays, like structs and dynamic types, force slot boundaries:
+/// - Start at offset 0 of a new slot
+/// - Force the next field to start at a new slot
+/// - Cannot be packed with other fields
+///
+/// This ensures arrays maintain contiguous storage layout.
+pub(crate) fn is_array_type(ty: &Type) -> bool {
+    matches!(ty, Type::Array(_))
+}
```
mattsse, line 331: this can also just be `Type::is_array`?

crates/precompiles-macros/src/utils.rs
```diff
@@ -0,0 +1,382 @@
...
+
+/// Checks if a type is a dynamic type that forces slot boundaries.
+///
+/// Dynamic types (like `String`, `Bytes`, and `Vec`) always:
+/// - Start at offset 0 of a new slot
+/// - Force the next field to start at a new slot
+/// - Cannot be packed with other fields
+///
+/// This matches Solidity's storage layout rules for dynamic types.
+pub(crate) fn is_dynamic_type(ty: &Type) -> bool {
```
mattsse, line 306: same for this,

not a blocker for this pr tho

crates/precompiles-macros/src/lib.rs
0xrusowsky: my reasoning was that the compiler would simply drop the unused types, but i will look into that

PR #672, opened 2025-11-04
feat(genesis): mint linkingUSD to all recepients

### LINKED ISSUE #621: Give users access to LinkingUSD via faucet
### Problem Statement
Users cannot access LinkingUSD from the faucet, preventing them from testing DEX swap functionality.

### Context
Unable to test buy orders and swaps because only AlphaUSD is available via faucet. LinkingUSD is required as the quote token for trading pairs.

### What Needs to be Fixed
Add LinkingUSD to faucet backend

shekhirin:
Closes #621

All recipients are now granted a transfer role for linkingUSD and minted `u64::MAX` of it.

### DIFF
--- a/xtask/src/genesis.rs
+++ b/xtask/src/genesis.rs
@@ -161,7 +161,7 @@ impl GenesisArgs {
         )?;
 
         println!("Initializing LinkingUSD");
-        initialize_linking_usd(admin, &mut evm)?;
+        initialize_linking_usd(admin, &addresses, &mut evm)?;
 
         println!("Initializing TIP20RewardsRegistry");
         initialize_tip20_rewards_registry(&mut evm)?;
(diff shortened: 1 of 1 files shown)


xtask/src/genesis.rs
```diff
@@ -407,9 +408,25 @@ fn initialize_linking_usd(
...
+        roles.grant_role_internal(*recipient, *TRANSFER_ROLE)?;
+    }
+
+    for recipient in recipients.iter().tqdm() {
+        linking_usd
+            .mint(
+                admin,
+                ITIP20::mintCall {
+                    to: *recipient,
+                    amount: U256::from(u64::MAX),
```
mattsse, line 425: we might want to change alloc and could move to fn arg, but seems fine

PR #675, opened 2025-11-04
fix(consensus): allow arbitrarily deep reorgs

SuperFluffy:
This relaxes the condition for sending a forkchoice update to update the head hash:

1. Now: set `ForkchoiceState.head_hash` to any `block.hash` as long as `block.height > last_finalized`.
2. Before: set `ForkchoiceState.head_hash` to `block.hash` if and only if `block.height >  latest_sent_head_height`

So this means the finalized hed acts as the only floor, and above it arbitary reorgs can happen.

This is done because simplex does not provide any guarantees for notarized or nullified rounds - only finalized rounds are firmly set.

### DIFF
--- a/crates/chainspec/src/bootnodes.rs
+++ b/crates/chainspec/src/bootnodes.rs
@@ -1,0 +1,12 @@
+use reth_network_peers::{NodeRecord, parse_nodes};
+
+pub(crate) static ANDANTINO_BOOTNODES: [&str; 4] = [
+    "enode://386269ddd50fd95143ad8f33f60ff67c7bd17da1adb46aed19eba5c6ae5326316643595e3aa99784f41b7d861fedde78594b1f764e56a4659917b916c1c3e321@148.113.193.123:30004",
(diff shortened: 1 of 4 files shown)


crates/commonware-node/src/consensus/application/executor.rs
```diff
@@ -115,11 +115,15 @@ impl LastCanonicalized {
...
+    ///
+    /// If `height <= self.finalized_height`, then this method will return
+    /// `self` unchanged.
     fn update_head(self, height: u64, hash: B256) -> Self {
         let mut this = self;
-        if height >= this.head_height {
+        if height > this.finalized_height {
             this.head_height = height;
             this.forkchoice.head_block_hash = hash;
         }
```
joshieDo, line 129: maybe logging with `warn!` on `else {}` ?

SuperFluffy: I didn't get an email for this comment and didn't see it before merge. I will add this in a followup.

PR #680, opened 2025-11-04
chore: Upgrade to `Application` interface

clabby:
Upgrades the `tempo-commonware-node` to use the new `Application` interface and `Marshaled` application wrapper offered by `commonware-consensus`. This interface removes the need to manually handle epoching.

Additionally, this PR replaces `tempo-commonware-node`'s own epoch utilities with those offered by `commonware-consensus`.


crates/commonware-node/src/consensus/application/core.rs
```diff
@@ -0,0 +1,473 @@
...
+        if let Err(error) = self
+            .executor_mailbox
+            .canonicalize_head(built_block.height(), built_block.commitment())
+        {
+            warn!(
+                %error,
+                proposal_digest = %built_block.commitment(),
+                "failed making the proposal the head of the canonical chain",
+            );
+        }
```
clabby, line 266: Note - before using the `Application` interface, you guys notified consensus prior to importing the built payload into the EL and canonicalizing it. With `Marshaled` handling this for you now, you no longer have this capability.

I'm interested to see if this causes a substantial performance degradation; If it does, we can definitely add the capability to inform consensus early upstream.

(Side note - since you're extending reth, maybe consider adding a new engine API method that persists built blocks within `engine_forchoiceUpdated` rather than having to chain another `engine_newPayload` afterwards to import.)

crates/commonware-node/src/epoch/manager/actor.rs
```diff
@@ -140,6 +145,10 @@ where
         }
     }
 
+    pub(crate) fn set_application(&mut self, app: Marshaled<TContext>) {
+        self.application = Some(app);
+    }
```
clabby, line 150: Don't love this, but quick & dirty patch to remove that typestate lazy-initialization y'all used to have on the application actor.

crates/commonware-node/src/consensus/application/core.rs
patrick-ogrady: Another approach here would be to extend `update()` to return when a block was broadcast by consensus to call this canonicalize head here.

This would ensure we only do so if consensus cared (an improvement) and wouldn't block on notifying consensus (like we do in this PR).

PR #681, opened 2025-11-04
chore!(precompiles): prevent setting supply cap to beyond u128 max

Rjected:
PR corresponding to spec in https://github.com/tempoxyz/docs/pull/121

### DIFF
--- a/crates/precompiles/src/tip20/mod.rs
+++ b/crates/precompiles/src/tip20/mod.rs
@@ -20,7 +20,7 @@ use crate::{
 };
 use alloy::{
     hex,
-    primitives::{Address, B256, Bytes, IntoLogData, U256, keccak256},
+    primitives::{Address, B256, Bytes, IntoLogData, U256, keccak256, uint},
 };
 use revm::{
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/tip20/mod.rs
```diff
@@ -181,7 +184,7 @@ impl<'a, S: PrecompileStorageProvider> TIP20Token<'a, S> {
         call: ITIP20::setSupplyCapCall,
     ) -> Result<(), TempoPrecompileError> {
         self.check_role(msg_sender, DEFAULT_ADMIN_ROLE)?;
-        if call.newSupplyCap < self.total_supply()? {
+        if call.newSupplyCap < self.total_supply()? || call.newSupplyCap > U128_MAX {
             return Err(TIP20Error::supply_cap_exceeded().into());
```
0xKitsune, line 188: Lgtm, just flagging that `call.newSupplyCap < self.total_supply()?` doesnt exceed the supply cap so we might want to update to a different error both in the reference implementation and the precompile.

Rjected: yep makes sense

PR #682, opened 2025-11-05
chore!(precompiles): change to InvalidSupplyCap in u128 max check

Rjected:
changes this to be consistent with [tempoxyz/docs@`98bf60a` (#121)](https://github.com/tempoxyz/docs/pull/121/commits/98bf60aa02b7c07a512f03dbb02f40a494029487) by introducing `InvalidSupplyCap`

### DIFF
--- a/crates/precompiles/src/linking_usd/mod.rs
+++ b/crates/precompiles/src/linking_usd/mod.rs
@@ -193,6 +193,7 @@ impl<'a, S: PrecompileStorageProvider> LinkingUSD<'a, S> {
 #[cfg(test)]
 mod tests {
 
+    use alloy_primitives::uint;
     use tempo_contracts::precompiles::RolesAuthError;
 
     use super::*;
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/tip20/mod.rs
```diff
@@ -184,9 +184,14 @@ impl<'a, S: PrecompileStorageProvider> TIP20Token<'a, S> {
...
         self.check_role(msg_sender, DEFAULT_ADMIN_ROLE)?;
-        if call.newSupplyCap < self.total_supply()? || call.newSupplyCap > U128_MAX {
+        if call.newSupplyCap < self.total_supply()? {
             return Err(TIP20Error::supply_cap_exceeded().into());
         }
+
+        if call.newSupplyCap > U128_MAX {
+            return Err(TIP20Error::invalid_supply_cap().into());
+        }
+
```
0xKitsune, line 194: Looks like these errors should be flipped:


```rust

 if call.newSupplyCap < self.total_supply()? {
            return Err(TIP20Error::invalid_supply_cap().into());
        }

        if call.newSupplyCap > U128_MAX {
        }
```

Rjected: right, flipped

PR #684, opened 2025-11-05
[e2e/tests] Integrate `SocketManager`

patrick-ogrady:
Blocking: https://github.com/commonwarexyz/monorepo/pull/2123
Related: https://github.com/tempoxyz/tempo/pull/677

### DIFF
--- a/crates/e2e/src/lib.rs
+++ b/crates/e2e/src/lib.rs
@@ -83,7 +83,7 @@ pub async fn setup_validators(
 ) -> (Vec<ValidatorNode>, simulated::Oracle<PublicKey>) {
     let threshold = quorum(how_many);
 
-    let (network, oracle) = Network::new(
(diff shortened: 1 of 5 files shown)


crates/e2e/src/lib.rs
```diff
@@ -119,6 +112,14 @@ pub async fn setup_validators(
         .map(|validator| (validator, SocketAddr::from(([127, 0, 0, 1], 0)).to_string()))
         .collect::<Vec<_>>()
         .into();
+    link_validators(&mut oracle, unresolved_peers.keys().as_ref(), linkage, None).await;
```
patrick-ogrady, line 115: All possible peers can now be linked before the start of the test (and then activated dynamically) 👍 

If you prefer, you can still link in a wrapper but it seems clearer to lay out the entire network topology before the test starts.

Issue #683, opened 2025-11-05
Transaction Fee Token Order Preference


0xKitsune:
Currently when checking a user’s fee token preference, the protocol resolves in this order:

- Fee token specified by tx type
- Fee payer’s specified fee token in FeeManager
- Validator’s specified fee token in FeeManager

This should be updated to match the latest spec. The new precedence resolution should be:

1. Fee token specified by tx type  
2. `fee_payer`’s fee token preference stored in FeeManager  
   - If the fee payer is also the `msg.sender` **and** the transaction is calling FeeManager to set a new preference, the newly set preference should be used immediately instead of the previously stored one  
3. If `tx.to()` is a TIP-20 token, use that token as the fee token  
4. Otherwise, fall back to the validator’s fee token preference

ref: tempoxyz/docs#110

PR #687, opened 2025-11-05
feat!(revm): add expanded fee token preference

Rjected:
ref https://github.com/tempoxyz/tempo/issues/683

This implements the fee token preference in `get_fee_token` according to the new rules

### DIFF
--- a/crates/revm/src/handler.rs
+++ b/crates/revm/src/handler.rs
@@ -957,7 +957,8 @@ where
     // If the fee payer is also the msg.sender and the transaction is calling FeeManager to set a
     // new preference, the newly set preference should be used immediately instead of the
     // previously stored one
-    if ctx.tx().fee_payer()? == ctx.tx().caller()
+    if ctx.tx().aa_tx_env.is_none()
+        && ctx.tx().fee_payer()? == ctx.tx().caller()
         && ctx.tx().kind().to() == Some(&TIP_FEE_MANAGER_ADDRESS)
         && let Ok(call) = IFeeManager::setUserTokenCall::abi_decode(ctx.tx().input())
     {
@@ -978,7 +979,8 @@ where
     }
 
     // If tx.to() is a TIP-20 token, use that token as the fee token
-    if let Some(&to_addr) = ctx.tx().kind().to()
+    if ctx.tx().aa_tx_env.is_none()
+        && let Some(&to_addr) = ctx.tx().kind().to()
         && is_tip20(to_addr)
     {
         return Ok(to_addr);
--- a/crates/node/tests/it/eth_call.rs
+++ b/crates/node/tests/it/eth_call.rs
@@ -346,7 +346,7 @@ async fn test_eth_estimate_gas_different_fee_tokens() -> eyre::Result<()> {
 
     // Estimate gas when user fee token differs from validator fee token
     let gas = provider.estimate_gas(tx.clone()).await?;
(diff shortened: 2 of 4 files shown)


crates/revm/src/handler.rs
```diff
@@ -903,32 +907,57 @@ pub fn get_fee_token<DB>(
...
+        // Try to decode setUserToken call
+        if let Ok(call) = IFeeManager::setUserTokenCall::abi_decode(ctx.tx().input()) {
+            call.token
+        } else {
+            stored_user_token
+        }
     } else {
-        Ok(user_fee_token)
+        stored_user_token
+    };
```
0xKitsune: nit: consider separating the `ctx.tx().kind().to() == Some(&TIP_FEE_MANAGER_ADDRESS) && ctx.tx().caller() == ctx.tx().fee_payer()` check for readability.



```rust
// If the fee payer is also the msg.sender and the tx is calling the FeeManager
let is_self_update =
    ctx.tx().kind().to() == Some(&TIP_FEE_MANAGER_ADDRESS)
        && ctx.tx().caller() == ctx.tx().fee_payer()?;

let effective_user_token = if is_self_update {
    IFeeManager::setUserTokenCall::abi_decode(ctx.tx().input())
        .map(|c| c.token)
        .unwrap_or(stored_user_token)
} else {
    stored_user_token
};
```

crates/revm/src/handler.rs
```diff
@@ -903,32 +907,57 @@ pub fn get_fee_token<DB>(
...
-        Ok(validator_fee_token)
+    // If the fee payer is also the msg.sender and the transaction is calling FeeManager to set a
+    // new preference, the newly set preference should be used immediately instead of the
+    // previously stored one
+    let effective_user_token = if ctx.tx().kind().to() == Some(&TIP_FEE_MANAGER_ADDRESS)
+        && ctx.tx().caller() == ctx.tx().fee_payer()?
+    {
+        // Try to decode setUserToken call
+        if let Ok(call) = IFeeManager::setUserTokenCall::abi_decode(ctx.tx().input()) {
+            call.token
```
klkvr: should we ensure that token is a tip20 here? not sure what's the behavior when feeToken is set to the value which doesn't make sense but `setUserToken` reverts in that case  https://github.com/tempoxyz/tempo/blob/c7c8a9deba737f619ddf6a1bb2624a511bc16395/crates/precompiles/src/tip_fee_manager/mod.rs#L167-L169

crates/node/tests/it/liquidity.rs
```diff
@@ -110,13 +121,27 @@ async fn test_block_building_insufficient_fee_amm_liquidity() -> eyre::Result<()
     // Now set the user's fee token to our custom payment token (not USDC)
     // This ensures subsequent transactions will require a swap through the drained FeeAMM
     println!("Setting user's fee token preference...");
-    use tempo_contracts::precompiles::IFeeManager;
-    let fee_manager = IFeeManager::new(TIP_FEE_MANAGER_ADDRESS, provider.clone());
-    fee_manager
-        .setUserToken(payment_token_addr)
-        .send()
+    let mut tx = TxFeeToken {
```
0xKitsune, line 124: For clarity, why did we need to update to use `TxFeeToken` rather than setting the user token? If the goal is to test fee token preference ordering, maybe we should have a separate test for this? Feel free to clarify though.

Issue #686, opened 2025-11-05
Update TIP-20 currency checks & remove TIP-4217


0xKitsune:
The following should be implemented to match recent changes to the spec:
ref: tempoxyz/docs#108

- All TIP-20 tokens default to 6 decimals
- Remove the TIP-4217 precompile and related documentation
- TIP-20 currency symbol may now be an arbitrary string (no registry / no reserved namespace requirement)
- All TIP-20 tokens must define a quote token
  - If the TIP-20 token is `USD` ->  the quote token must also be `USD`
- Non-USD tokens:
  - cannot be selected as a fee token (by validator or user)
  - cannot be used on the FeeAMM
- StableExchange pairs cannot be created if either token or its quote token are non-USD

PR #691, opened 2025-11-05
feat(precompiles): enforce TIP20 currency checks

0xKitsune:
Ref #686 

Stacks on #689 

This PR enforces that non-USD tokens:
- Cannot be selected as a fee token by a user or validator
- Cannot be used on the FeeAMM
- StableExchange pairs cannot be created if either token or its quote token are non-USD

### DIFF
--- a/crates/precompiles/src/tip20/mod.rs
+++ b/crates/precompiles/src/tip20/mod.rs
@@ -103,7 +103,7 @@ pub fn validate_usd_currency<S: PrecompileStorageProvider>(
 ) -> Result<(), TempoPrecompileError> {
     let mut tip20_token = TIP20Token::from_address(token, storage);
     let currency = tip20_token.currency()?;
-    if keccak256(currency.as_bytes()) != keccak256(USD_CURRENCY.as_bytes()) {
+    if currency != USD_CURRENCY {
         return Err(TIP20Error::invalid_currency().into());
     }
     Ok(())


crates/precompiles/src/tip20/mod.rs
```diff
@@ -91,6 +96,19 @@ pub static UNPAUSE_ROLE: LazyLock<B256> = LazyLock::new(|| keccak256(b"UNPAUSE_R
...
 pub static BURN_BLOCKED_ROLE: LazyLock<B256> = LazyLock::new(|| keccak256(b"BURN_BLOCKED_ROLE"));
 
+/// Validates that a token has USD currency
+pub fn validate_usd_currency<S: PrecompileStorageProvider>(
+    token: Address,
+    storage: &mut S,
+) -> Result<(), TempoPrecompileError> {
+    let mut tip20_token = TIP20Token::from_address(token, storage);
+    let currency = tip20_token.currency()?;
+    if keccak256(currency.as_bytes()) != keccak256(USD_CURRENCY.as_bytes()) {
```
shekhirin: why do we compare keccaks here, and not just raw strings?

crates/precompiles/src/tip20/mod.rs
```diff
@@ -663,6 +689,16 @@ impl<'a, S: PrecompileStorageProvider> TIP20Token<'a, S> {
         self.write_string(slots::NAME, name.to_string())?;
         self.write_string(slots::SYMBOL, symbol.to_string())?;
         self.write_string(slots::CURRENCY, currency.to_string())?;
+
+        // If the currency is USD, the quote token must also be USD
```
shekhirin, line 693: nit: can be moved to a helper method, like `self.ensure_quote_token_matches()`, and reused in `update_quote_token`

crates/precompiles/src/tip20/mod.rs
shekhirin: ah, it's coming from the Solidity implementation. Unsure if we need to replicate this.

fgimenez: good point, we are already comparing strings in other places

PR #702, opened 2025-11-05
test(precompiles): add dispatch tests for stablecoin exchange

fgimenez:
fills in test stubs in the stablecoin exchange's dispatch module

### DIFF
--- a/crates/precompiles/src/stablecoin_exchange/dispatch.rs
+++ b/crates/precompiles/src/stablecoin_exchange/dispatch.rs
@@ -270,7 +270,8 @@ mod tests {
 
         // Should dispatch to place function (may fail due to business logic, but dispatch works)
         let result = exchange.call(&Bytes::from(calldata), sender);
-        assert!(result.is_ok() || result.is_err()); // Just verify it dispatches
+        // Ok indicates successful dispatch (either success or TempoPrecompileError)
+        assert!(result.is_ok());
     }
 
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/stablecoin_exchange/dispatch.rs
```diff
@@ -152,64 +152,348 @@ impl<'a, S: PrecompileStorageProvider> Precompile for StablecoinExchange<'a, S>
...
+            amount: 100u128,
+            isBid: true,
+            tick: 0,
+            flipTick: 10,
+        };
+        let calldata = call.abi_encode();
+
+        // Should dispatch to place_flip function
+        let result = exchange.call(&Bytes::from(calldata), sender);
+        assert!(result.is_ok() || result.is_err()); // Just verify it dispatches
```
0xKitsune: IIUC this will always be true even when we hit errors related to "Unknown selector", etc. We should assert that the result is not a `PrecompileError`, or that the result is either ok or a `TempoPrecompileError`.

fgimenez: absolutely, this was a leftover sorry, updated all the occurrences to assert `result.is_ok()` which should account for success and `TempoPrecompileError` ( == `Ok(PrecompileOutput { reverted: true, ...})`) ptal

PR #705, opened 2025-11-05
fix(consensus): block only on finalized blocks

SuperFluffy:
This patch ensures that the `Reporter for application::Mailbox` implementation only blocks on finalized blocks, not on updates to the tip of the finalized chain.

PR https://github.com/commonwarexyz/monorepo/pull/2065 changed the commonware marshal actor to not only report a sequential stream of finalized blocks to the application, but to now also send the the highest known finalized tip of the the consensus chain.

Where before it expected only blocks to be explicitly acknowledged by the application (by awaiting `application.report(block).await`), this is explicitly not the case for the tip updates, and the marshal actor expects to not block in these calls:

https://github.com/commonwarexyz/monorepo/blob/f6fc8ec7eac95bc6508a64caaca0a2a042826816/consensus/src/marshal/actor.rs#L298
https://github.com/commonwarexyz/monorepo/blob/f6fc8ec7eac95bc6508a64caaca0a2a042826816/consensus/src/marshal/actor.rs#L785

The first instance, before the marshaller enters its event loop, reports the finalized tip to the application. The application (specifically, the application's executor actor) however queries the marshal actor for its latest finalized tip:
https://github.com/tempoxyz/tempo/blob/91ae859f8d937195ffcfcc670d31fa78496d710c/crates/commonware-node/src/consensus/application/executor.rs#L198

This creates a deadlock: in order for the application to serve the marshal actor, it must enter its event loop. Likewise, in order for the marshal actor to serve the application, it must in turn enter its event loop.

But both are blocked right before their respective loops start.

### DIFF
--- a/crates/commonware-node/src/consensus/engine.rs
+++ b/crates/commonware-node/src/consensus/engine.rs
@@ -415,13 +415,13 @@ where
         let resolver =
             marshal::resolver::p2p::init(&self.context, self.resolver_config, marshal_channel);
 
+        let application = self.application.start(self.dkg_manager_mailbox.clone());
+
         let marshal = self.marshal.start(
             Reporters::from((self.application_mailbox, self.dkg_manager_mailbox.clone())),
             self.broadcast_mailbox,
             resolver,
         );
-
-        let application = self.application.start(self.dkg_manager_mailbox.clone());
 
         let epoch_manager = self.epoch_manager.start(
             pending_channel,


crates/commonware-node/src/consensus/application/ingress.rs
```diff
@@ -181,18 +197,35 @@ impl Reporter for Mailbox {
...
 
         // XXX: This is used as an acknowledgement that the application
         // finalized the block:
         // Response on this channel -> future returns -> marshaller gets an ack
-        //
-        // TODO(janis): report if this channel gets dropped?
-        let _ = rx.await;
+        if let Some(rx) = rx {
+            // TODO(janis): commonware does not distinguish between the channel
+            // getting dropped or receiving a response. But this is something
```
patrick-ogrady, line 226: related: https://github.com/commonwarexyz/monorepo/issues/2135

crates/commonware-node/src/consensus/engine.rs
```diff
@@ -409,17 +412,17 @@ where
...
             marshal::resolver::p2p::init(&self.context, self.resolver_config, marshal_channel);
 
         let marshal = self.marshal.start(
-            Reporters::from((self.application_mailbox, self.dkg_manager_mailbox)),
+            Reporters::from((self.application_mailbox, self.dkg_manager_mailbox.clone())),
             self.broadcast_mailbox,
             resolver,
         );
 
+        let application = self.application.start(self.dkg_manager_mailbox.clone());
```
patrick-ogrady: I think you should keep the prior order 👍

Issue #708, opened 2025-11-05
bug: `tip20_factory` `TokenCreated` event does not log `quoteToken` and does not expose `isTIP20` per Solidity reference implementation


zerosnacks:
Reference:

https://github.com/tempoxyz/docs/blob/ae57961228dd8481efa6da16c903bf0ebd864d47/specs/src/TIP20Factory.sol#L10-L18

misses the `quoteToken` field in `TokenCreated`

Implementation:

https://github.com/tempoxyz/tempo/blob/91ae859f8d937195ffcfcc670d31fa78496d710c/crates/contracts/src/precompiles/tip20_factory.rs#L8

And we are also missing the `isTIP20` view function that should be available:

https://github.com/tempoxyz/docs/blob/ae57961228dd8481efa6da16c903bf0ebd864d47/specs/src/TIP20Factory.sol#L51-L54

PR #711, opened 2025-11-05
fix!(`ITIP20Factory` precompile): add `quoteToken` to `TokenCreated` event per specification, add `isTIP20` view method per specification

zerosnacks:
Fixes: https://github.com/tempoxyz/tempo/issues/708

Added smoke test for `isTIP20` 

Ref: https://github.com/tempoxyz/docs/blob/ae57961228dd8481efa6da16c903bf0ebd864d47/specs/src/TIP20Factory.sol#L10-L18

Ref: https://github.com/tempoxyz/docs/blob/ae57961228dd8481efa6da16c903bf0ebd864d47/specs/src/TIP20Factory.sol#L51-L54


crates/precompiles/src/tip20_factory/mod.rs
```diff
@@ -106,13 +107,17 @@ impl<'a, S: PrecompileStorageProvider> TIP20Factory<'a, S> {
             Ok(counter)
         }
     }
+
+    pub fn is_tip20(&mut self, token: Address) -> Result<bool> {
+        Ok(is_tip20(token))
+    }
```
0xKitsune: We will also need to add this to dispatch [here](https://github.com/tempoxyz/tempo/blob/main/crates/precompiles/src/tip20_factory/dispatch.rs#L24). We can probably also add `is_tip20` directly in dispatch, rather than creating a wrapper here. 

Then we can move the test below to `dispatch.mod` tests.

zerosnacks: Ah yes, inlined `is_tip20`, moved test https://github.com/tempoxyz/tempo/pull/711/commits/0ffa56a77aaa77d2db59979123674d5b269d539b

zerosnacks: Ended up moving the test to `tip20/mod.rs` as thats where `is_tip20` is defined https://github.com/tempoxyz/tempo/pull/711/commits/0d2be5de86f35d75dd59601849dd65fb67bb424b; feel free to move  where you see fit:

PR #718, opened 2025-11-05
feat(revm): preserve logs even on halt/revert

rakita:
Preserve logs that would be removed if halt/revert happens on tx

### DIFF
--- a/crates/revm/src/handler.rs
+++ b/crates/revm/src/handler.rs
@@ -374,18 +374,20 @@ where
         }
     }
 
+    /// Take logs from the Journal if outcome is Halt Or Revert.
     #[inline]
-    fn last_frame_result(
+    fn execution_result(
         &mut self,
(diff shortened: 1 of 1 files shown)


crates/revm/src/handler.rs
```diff
@@ -374,6 +383,20 @@ where
         }
     }
 
+    #[inline]
+    fn last_frame_result(
```
klkvr: we want this to happen later because we include the fee transfer logs here in `reimburse_caller`
https://github.com/tempoxyz/tempo/blob/716c25b4a33e1315a8585c5f4abc884842297f55/crates/revm/src/handler.rs#L658

rakita: Yeah, have moved it to `Handler::execution_result` it is the step where output gets created. One of last steps, or maybe even the last one.

PR #728, opened 2025-11-05
chore!(consensus): Migrate to `commonware-consensus` epoch helpers

clabby:
## Overview

Migrates the epoch helpers in `tempo-commonware-node` to use those in the `commonware-consensus` crate. Tempo currently handles the semantics of epochs differently than upstream, which may make future integrations more difficult if a long-lived network is launched (i.e. integrating with the new `Application` interface, as in #680.)

### DIFF
--- a/crates/chainspec/src/genesis/adagio.json
+++ b/crates/chainspec/src/genesis/adagio.json
@@ -32,7 +32,7 @@
     "pragueTime": 0,
     "adagioTime": 0
   },
-  "baseFeePerGas": "0x28048c5ec000",
+  "baseFeePerGas": "0x2540be400",
   "gasLimit": 50000000,
   "alloc": {
     "0x7702c00000000000000000000000000000000000": {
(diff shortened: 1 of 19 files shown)


crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -253,7 +254,11 @@ impl Inner<Init> {
         let source = if genesis.epoch == 0 {
             self.genesis_block.digest()
         } else {
-            let height = epoch::parent_height(genesis.epoch, self.epoch_length);
+            // The last block of the *previous* genesis provides the "genesis"
+            // of the *current* epoch. Only epoch 0 is special cased above.
+            let height =
```
patrick-ogrady, line 259: `marshal` uses our `utils` to determine the signing scheme for certificate validation: https://github.com/commonwarexyz/monorepo/blob/40de291edb20b1cc60d8b4afef73da0ebe667a3d/consensus/src/marshal/actor.rs#L613-L616

In `threshold`, misalignment shouldn't matter because we'll verify using the constant key but this is technically "wrong" (and is only working because of that constant verification component).

patrick-ogrady: Disregard. Read this wrong ("epoch 0 is special cased" put me in a bender).

PR #733, opened 2025-11-05
fix: add missing fn

howydev:


crates/precompiles/src/tip20/dispatch.rs
```diff
@@ -171,6 +171,12 @@ impl<'a, S: PrecompileStorageProvider> Precompile for TIP20Token<'a, S> {
                 })
             }
 
+            ITIP20::optedInSupplyCall::SELECTOR => {
+                view::<ITIP20::optedInSupplyCall>(calldata, |_call| {
+                    self.get_opted_in_supply()
```
joshieDo: needs `Ok(value.to::<u128>())` inside `get_opted_in_supply`, but there seems to be other places that use it as u256

are we sure its its being used as u128

PR #747, opened 2025-11-06
chore!: rename receive_role

### LINKED ISSUE #724: bug: `RECEIVE_ROLE` for `LinkingUSD` is specified as `RECEIVE_WITH_MEMO_ROLE` in the Solidity reference implementation
The `RECEIVE_ROLE` in the specification is `RECEIVE_WITH_MEMO_ROLE` for `LinkingUSD`

https://github.com/tempoxyz/docs/blob/d73975cc8d72a274f69661d7183395b4061d2b1a/specs/src/LinkingUSD.sol#L12

https://github.com/tempoxyz/tempo/blob/41fdaf11154daa03cafd906d0ce630cdf8072686/crates/precompiles/src/linking_usd/mod.rs#L14

Not clear what is preferred @danrobinson

mattsse:
closes #724

pending decision

### DIFF
--- a/crates/precompiles/src/linking_usd/mod.rs
+++ b/crates/precompiles/src/linking_usd/mod.rs
@@ -34,7 +34,24 @@ impl<'a, S: PrecompileStorageProvider> LinkingUSD<'a, S> {
             .initialize(NAME, SYMBOL, CURRENCY, Address::ZERO, admin)
     }
 
-    fn is_transfer_authorized(&mut self, sender: Address, recipient: Address) -> Result<bool> {
+    fn is_transfer_authorized(&mut self, sender: Address) -> Result<bool> {
+        let authorized =
+            sender == STABLECOIN_EXCHANGE_ADDRESS || self.token.has_role(sender, *TRANSFER_ROLE)?;
+
+        Ok(authorized)
+    }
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/linking_usd/mod.rs
```diff
@@ -36,7 +37,7 @@ impl<'a, S: PrecompileStorageProvider> LinkingUSD<'a, S> {
     fn is_transfer_authorized(&mut self, sender: Address, recipient: Address) -> Result<bool> {
         let authorized = sender == STABLECOIN_EXCHANGE_ADDRESS
             || self.token.has_role(sender, *TRANSFER_ROLE)?
-            || self.token.has_role(recipient, *RECEIVE_ROLE)?;
+            || self.token.has_role(recipient, *RECEIVE_WITH_MEMO_ROLE)?;
```
0xKitsune, line 57: Looks like this has drifted from the spec, this should only be checked during `transfer_from_with_memo` and `transfer_with_memo`. Otherwise, transfers should only check the transfer role.

https://github.com/tempoxyz/docs/blob/main/specs/src/LinkingUSD.sol#L18-L37

PR #751, opened 2025-11-06
fix: update receive with memo checks

0xKitsune:


crates/precompiles/src/linking_usd/mod.rs
```diff
@@ -34,15 +34,32 @@ impl<'a, S: PrecompileStorageProvider> LinkingUSD<'a, S> {
...
+            sender == STABLECOIN_EXCHANGE_ADDRESS || self.token.has_role(sender, *TRANSFER_ROLE)?;
+
+        Ok(authorized)
+    }
+
+    fn is_transfer_from_authorized(&mut self, sender: Address, from: Address) -> Result<bool> {
+        let authorized =
+            sender == STABLECOIN_EXCHANGE_ADDRESS || self.token.has_role(from, *TRANSFER_ROLE)?;
+        Ok(authorized)
+    }
```
mattsse, line 48: this is the same fn?

PR #754, opened 2025-11-06
feat: include fee token into RPC transaction receipts

klkvr:
Adds `TempoTransactionReceipt` RPC type that extends basic `TransactionReceipt` with `feeToken` field.

For now we assume that fee token is the address that emitted the last log

### DIFF
--- a/crates/node/tests/it/tip20_gas_fees.rs
+++ b/crates/node/tests/it/tip20_gas_fees.rs
@@ -9,6 +9,7 @@ use alloy_network::TransactionBuilder;
 use alloy_primitives::Bytes;
 use alloy_rpc_types_eth::TransactionRequest;
 use std::env;
+use tempo_alloy::rpc::TempoTransactionReceipt;
 use tempo_contracts::precompiles::{IFeeManager, ITIP20};
 use tempo_precompiles::TIP_FEE_MANAGER_ADDRESS;
 use tempo_primitives::transaction::calc_gas_balance_spending;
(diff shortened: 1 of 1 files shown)


crates/node/src/rpc/mod.rs
```diff
@@ -272,6 +277,59 @@ impl<N: FullNodeTypes<Types = TempoNode>> EthTransactions for TempoEthApi<N> {
...
+                    fee_token: None,
+                };
+                if receipt.effective_gas_price == 0 || receipt.gas_used == 0 {
+                    return receipt;
+                }
+
+                // Set fee token to the address that emitted the last log.
+                //
+                // Assumption is that every non-free transaction will end with a
+                // fee token transfer to TIPFeeManager.
```
shekhirin, line 325: can we add a test that will start failing if this assumption breaks?

shekhirin: not blocking if we want to get this PR in fast, but let's create an issue to not forget

mattsse: yeah can do

klkvr: we kind of already aserted this here https://github.com/tempoxyz/tempo/blob/ffd44f9099931954e836f66e3dde3919f1e43238/crates/node/tests/it/tip20_gas_fees.rs#L53-L58

updated the tests to also deserialize `TempoTransactionReceipt` and compare the `feeToken`

PR #757, opened 2025-11-06
feat: expose `timestampMillis` on RPC header

klkvr:
Exposes `timestampMillis` field on RPC header response


crates/alloy/src/rpc/compat.rs
```diff
@@ -198,3 +200,12 @@ impl SignableTxRequest<TempoTxEnvelope> for TempoTransactionRequest {
         SignableTxRequest::<TempoTxEnvelope>::try_build_and_sign(self.inner, signer).await
     }
 }
+
+impl FromConsensusHeader<TempoHeader> for TempoHeaderResponse {
+    fn from_consensus_header(header: SealedHeader<TempoHeader>, block_size: usize) -> Self {
+        Self {
+            timestamp_millis: header.timestamp_millis(),
```
mattsse, line 207: ah this makes sense

Issue #744, opened 2025-11-06
Add checks for value == 0


rakita:
Value should always be zero as nobody has native token.

This check should propagate to AATx

PR #759, opened 2025-11-06
feat!: error on none zero tx.value

rakita:
closing https://github.com/tempoxyz/tempo/issues/744

### DIFF
--- a/crates/revm/src/handler.rs
+++ b/crates/revm/src/handler.rs
@@ -1051,51 +1051,6 @@ where
     Ok(balance)
 }
 
-/// Transfers `amount` from the sender's to the receivers balance inside the token contract.
-///
-/// Caution: assumes the `token` address is already loaded
-pub fn transfer_token<JOURNAL>(
-    journal: &mut JOURNAL,
-    token: Address,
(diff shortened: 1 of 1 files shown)


crates/revm/src/handler.rs
```diff
@@ -1599,4 +1608,31 @@ mod tests {
             "Should calculate floor gas for Prague matching revm"
         );
     }
+
+    #[test]
+    fn test_zero_value_transfer() -> eyre::Result<()> {
```
mattsse, line 1549: can we add a note here that this will start failing once we no longer enforce this? ideally with this pr's url

crates/revm/src/handler.rs
```diff
@@ -823,7 +828,12 @@ fn calculate_aa_batch_intrinsic_gas<'a>(
             gas.initial_gas += initcode_cost(call.input.len());
         }
 
+        if !call.value.is_zero() {
+            return Err(TempoInvalidTransaction::ValueTransferNotAllowedInAATx);
+        }
```
mattsse, line 838: I think we should also make a note here

PR #761, opened 2025-11-06
fix!: correct mapping slots

onbjerg:
The `mapping_slot` function incorrectly encoded the key as little endian, while the EVM is big endian. This made the mapping slot incorrect vs Solidity.

Additionally, `pad_to_32` was applying right padding, when it should be applying left padding.

The slot fixture is from Solidity obtained by @howydev

### DIFF
--- a/crates/precompiles/src/storage/slots.rs
+++ b/crates/precompiles/src/storage/slots.rs
@@ -1,6 +1,6 @@
 use alloy::primitives::{U256, keccak256};
 
-fn right_pad_to_32(data: &[u8]) -> [u8; 32] {
+fn left_pad_to_32(data: &[u8]) -> [u8; 32] {
     let mut buf = [0u8; 32];
     buf[32 - data.len()..].copy_from_slice(data);
     buf
@@ -10,7 +10,7 @@ fn right_pad_to_32(data: &[u8]) -> [u8; 32] {
 #[inline]
 pub fn mapping_slot<T: AsRef<[u8]>>(key: T, mapping_slot: U256) -> U256 {
     let mut buf = [0u8; 64];
-    buf[..32].copy_from_slice(&right_pad_to_32(key.as_ref()));
+    buf[..32].copy_from_slice(&left_pad_to_32(key.as_ref()));
     buf[32..].copy_from_slice(&mapping_slot.to_be_bytes::<32>());
     U256::from_be_bytes(keccak256(buf).0)
 }
@@ -24,7 +24,7 @@ pub fn double_mapping_slot<T: AsRef<[u8]>, U: AsRef<[u8]>>(
 ) -> U256 {
     let intermediate_slot = mapping_slot(key1, base_slot);
     let mut buf = [0u8; 64];
-    buf[..32].copy_from_slice(&right_pad_to_32(key2.as_ref()));
+    buf[..32].copy_from_slice(&left_pad_to_32(key2.as_ref()));
     buf[32..].copy_from_slice(&intermediate_slot.to_be_bytes::<32>());
     U256::from_be_bytes(keccak256(buf).0)
 }


crates/precompiles/src/storage/slots.rs
```diff
@@ -1,14 +1,8 @@
 use alloy::primitives::{U256, keccak256};
 
-pub const fn pad_to_32(x: &[u8]) -> [u8; 32] {
+pub fn pad_to_32(data: &[u8]) -> [u8; 32] {
```
DaniPopes: please rename to right_pad_32

crates/precompiles/src/storage/slots.rs
```diff
@@ -39,8 +33,25 @@ pub fn double_mapping_slot<T: AsRef<[u8]>, U: AsRef<[u8]>>(
...
+
+        let slot = double_mapping_slot(account, role.to_be_bytes::<32>(), roles_slot);
+
+        assert_eq!(
+            slot,
+            U256::from_be_bytes(
+                b256!("863fcb99fb78098ba65a07c5dcf7196ac60e997b3f3793ab79b7db9297b2b1c6").0
+            )
+        );
+    }
```
zerosnacks, line 53: For reference:

```
cast index uint256 0 $(cast index address 0x7fa9385be102ac3eac297483dd6233d62b3e1496 0)
```

results in 

```
0x863fcb99fb78098ba65a07c5dcf7196ac60e997b3f3793ab79b7db9297b2b1c6
```

which matches the above

crates/precompiles/src/storage/slots.rs
```diff
@@ -1,23 +1,17 @@
...
+fn right_pad_to_32(data: &[u8]) -> [u8; 32] {
     let mut buf = [0u8; 32];
-    let mut i = 0;
-    // Note: This is not idiomatic but it's the cleanest
-    // way to make this function const as far as I can tell.
-    while i < x.len() && i < 32 {
-        buf[i] = x[i];
-        i += 1;
-    }
+    buf[32 - data.len()..].copy_from_slice(data);
```
fgimenez, line 5: is this right or left pad? zeros to the left?

onbjerg: sorry yes, left, typo

crates/precompiles/src/storage/slots.rs
```diff
@@ -1,23 +1,17 @@
 use alloy::primitives::{U256, keccak256};
 
-pub const fn pad_to_32(x: &[u8]) -> [u8; 32] {
+fn right_pad_to_32(data: &[u8]) -> [u8; 32] {
```
onbjerg:
```suggestion
fn left_pad_to_32(data: &[u8]) -> [u8; 32] {
```

crates/precompiles/src/storage/slots.rs
```diff
@@ -1,23 +1,17 @@
...
     buf
 }
 
 /// Compute storage slot for a mapping
 #[inline]
 pub fn mapping_slot<T: AsRef<[u8]>>(key: T, mapping_slot: U256) -> U256 {
     let mut buf = [0u8; 64];
-    buf[..32].copy_from_slice(&pad_to_32(key.as_ref()));
-    buf[32..].copy_from_slice(&mapping_slot.to_le_bytes::<32>());
+    buf[..32].copy_from_slice(&right_pad_to_32(key.as_ref()));
```
onbjerg:
```suggestion
    buf[..32].copy_from_slice(&left_pad_to_32(key.as_ref()));
```

crates/precompiles/src/storage/slots.rs
```diff
@@ -30,7 +24,7 @@ pub fn double_mapping_slot<T: AsRef<[u8]>, U: AsRef<[u8]>>(
 ) -> U256 {
     let intermediate_slot = mapping_slot(key1, base_slot);
     let mut buf = [0u8; 64];
-    buf[..32].copy_from_slice(&pad_to_32(key2.as_ref()));
+    buf[..32].copy_from_slice(&right_pad_to_32(key2.as_ref()));
```
onbjerg:
```suggestion
    buf[..32].copy_from_slice(&left_pad_to_32(key2.as_ref()));
```

PR #768, opened 2025-11-06
fix(consensus): relax starting conditions on height mismatch between execution, consensus layers

SuperFluffy:
The starting conditions were too strict. It's probably fine if our execution layer is ahead of the consensus layer. We just need to set the forkchoice state to whatever is higher. Consensus will then forward one block or more blocks, and execution will process or drop it.

This change is not breaking.

Followup: https://github.com/tempoxyz/tempo/issues/769

### DIFF
--- a/crates/commonware-node/src/consensus/application/executor.rs
+++ b/crates/commonware-node/src/consensus/application/executor.rs
@@ -7,7 +7,7 @@
 //! If the agent detects that the execution layer is missing blocks it attempts
 //! to backfill them from the consensus layer.
 
-use std::{sync::Arc, time::Duration};
+use std::{cmp::Ordering, sync::Arc, time::Duration};
 
 use alloy_primitives::B256;
 use alloy_rpc_types_engine::ForkchoiceState;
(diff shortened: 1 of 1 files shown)


crates/commonware-node/src/consensus/application/executor.rs
```diff
@@ -208,61 +209,80 @@ where
                 (0, self.genesis_block.digest())
             });
 
+        let execution_height = self.execution_node.provider.last_block_number().wrap_err(
+            "failed getting last block number from execution layer; cannot \
+                continue without it",
+        )?;
+
+        let block = Block::from_execution_block(SealedBlock::seal_slow(
```
joshieDo, line 217: can we move this down to where its needed

crates/commonware-node/src/consensus/application/executor.rs
```diff
@@ -208,61 +209,80 @@ where
...
+                    (execution_height, block.digest(), false)
+                }
+            };
         self.canonicalize(
             Span::current(),
             HeadOrFinalized::Finalized,
-            finalized_consensus_height,
-            finalized_consensus_digest,
+            finalized_height,
+            finalized_digest,
```
joshieDo, line 272: there's a weird edge case that shouldnt matter, but may be important to highlight.

if we are restarting in the middle of a pipeline sync, `last_block_number` will not be returning the sync target (maybe here we want to actually check `sync_status` as well) . 

And so, we may be sending an FCU with a lower finalized block hash than what our node is syncing towards to. Shouldn't be an issue since it's part of the canonical chain.

crates/commonware-node/src/consensus/application/executor.rs
SuperFluffy: Well, we are immediately reporting on the digest in the info event right after.

Is there a call that would give me latest height && corresponding hash in one go?

crates/commonware-node/src/consensus/application/executor.rs
SuperFluffy: OK, but sync status only returns heights IIRC. What would I take as the hashes?

PR #770, opened 2025-11-06
test(bench): Include DEX orders and swaps in `max_tps`

### LINKED ISSUE #614: Extend transaction spammer to support stableDEX transactions
### Context
Currently we only send TIP20 transfers in our txn spammer, for more realistic load we should add stableDEX support for:
-   Multi-node network benchmarking
-   Realistic orderbook load testing
-   Devnet/testnet automated performance testing

**Tools:**
- `tempo-bench` (`bin/tempo-bench/src/cmd/max_tps.rs`) - high-throughput bursts (25k TPS)
- `tempo-sidecar` (`bin/tempo-sidecar/src/synthetic_load/mod.rs`) - continuous realistic load (50 TPS)

### Requirements

**Transaction types:**
1. ✅ TIP20 transfers (existing)
2. StableDEX limit orders: `place(token, amount, isBid, tick)`
3. StableDEX swaps: `swapExactAmountIn(tokenIn, tokenOut, amountIn, minOut)`

**Configuration:**
- Weighted distribution via CLI: `--tip20-weight=0.7 --place-order-weight=0.2 --swap-weight=0.1`
- Configurable parameters: amounts, tick ranges, token pairs

**Setup:**
- Fund accounts with multiple tokens
- Approve tokens for stableDEX contract
- Seed initial liquidity for swaps

### Priorities

Priority should be to get out an MVP with StableDEX support (setup + transactions), potentially with weights. Later on, we can focus on configurable params for swaps.

### References
- Inspiration: [Flashbots Contender](https://github.com/flashbots/contender)
- StableDEX precompile: `crates/precompiles/src/stablecoin_exchange/mod.rs`
- Example usage: `crates/node/tests/it/stablecoin_exchange.rs`

RomanHodulak:
Closes #614

Adds transactions to setup the DEX.

Includes DEX `place` and `spawn_exact_amount_in` transactions along with TIP20 transfers in the `generate_transactions` function randomly distributed with equal weights.

### DIFF
--- a/bin/tempo-bench/src/cmd/max_tps.rs
+++ b/bin/tempo-bench/src/cmd/max_tps.rs
@@ -6,17 +6,9 @@ use tempo_alloy::TempoNetwork;
 use alloy::{
     consensus::BlockHeader,
     eips::{BlockNumberOrTag::Latest, Decodable2718},
-    network::{
-        Ethereum, EthereumWallet, Network, ReceiptResponse, TransactionBuilder, TxSignerSync,
-    },
+    network::{Ethereum, Network, ReceiptResponse, TransactionBuilder, TxSignerSync},
     primitives::{Address, BlockNumber, ChainId, Signature, TxKind, U256},
-    providers::{
-        PendingTransactionBuilder, Provider, ProviderBuilder, RootProvider,
-        fillers::{
-            BlobGasFiller, ChainIdFiller, FillProvider, GasFiller, JoinFill, NonceFiller,
-            WalletFiller,
-        },
-    },
+    providers::{PendingTransactionBuilder, Provider, ProviderBuilder},
     sol_types::{SolCall, SolEvent},
     transports::http::reqwest::Url,
 };
--- a/bin/tempo-bench/src/cmd/max_tps/dex.rs
+++ b/bin/tempo-bench/src/cmd/max_tps/dex.rs
@@ -1,20 +1,10 @@
 use super::*;
+use alloy::providers::DynProvider;
 use std::pin::Pin;
 use tempo_contracts::precompiles::IStablecoinExchange;
 use tempo_precompiles::stablecoin_exchange::{MAX_TICK, MIN_TICK, price_to_tick};
 
 const GAS_LIMIT: u64 = 500_000;
-
-type DexProvider = FillProvider<
-    JoinFill<
-        JoinFill<
-            alloy::providers::Identity,
-            JoinFill<GasFiller, JoinFill<BlobGasFiller, JoinFill<NonceFiller, ChainIdFiller>>>,
-        >,
-        WalletFiller<EthereumWallet>,
-    >,
-    RootProvider,
->;
 
 /// This method performs a one-time setup for sending a lot of transactions:
 /// * Adds a quote token and a couple of user tokens paired with the quote token.
@@ -27,7 +17,7 @@ pub(super) async fn setup(
     signers: Vec<PrivateKeySigner>,
     max_concurrent_requests: usize,
 ) -> eyre::Result<(
-    IStablecoinExchangeInstance<DexProvider>,
+    IStablecoinExchangeInstance<DynProvider>,
     Address,
     Address,
     Address,
@@ -135,7 +125,7 @@ pub(super) async fn setup(
 
     join_all(futures, &tx_count, max_concurrent_requests).await?;
 
-    let exchange = IStablecoinExchange::new(STABLECOIN_EXCHANGE_ADDRESS, provider.clone());
+    let exchange = IStablecoinExchange::new(STABLECOIN_EXCHANGE_ADDRESS, provider.clone().erased());
 
     Ok((
         exchange,


bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -306,35 +316,324 @@ async fn generate_transactions(
...
+            JoinFill<
+                alloy::providers::Identity,
+                JoinFill<GasFiller, JoinFill<BlobGasFiller, JoinFill<NonceFiller, ChainIdFiller>>>,
+            >,
+            WalletFiller<EthereumWallet>,
+        >,
+        RootProvider,
+    >;
+
+    pub(super) async fn setup(
```
0xKitsune: Can we add some docs describing what this function is doing?

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -306,35 +316,324 @@ async fn generate_transactions(
...
+            5 => tip20::transfer(signer.clone(), nonce, chain_id, base2),
+            v => unreachable!("Number {v} is outside the random range"),
         })
         .collect::<eyre::Result<Vec<_>>>()?;
 
     println!("Generated {} transactions", transactions.len());
     Ok(transactions)
 }
 
+mod dex {
```
0xKitsune: Do we need to add a module here or can we just have a function like `setup_stablecoin_exchange()`? I think simpler is better for this purpose.

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -306,35 +316,324 @@ async fn generate_transactions(
...
+                    .send()
+                    .await?,
+                quote
+                    .approve(STABLECOIN_EXCHANGE_ADDRESS, U256::MAX)
+                    .gas_price(TEMPO_BASE_FEE as u128)
+                    .gas(300_000)
+                    .send()
+                    .await?,
+            ]);
+        }
```
0xKitsune: Can we simplify to something like the following? I dont think we need to do this twice but let me know if I am overlooking something.


```rust

let mut receipts = Vec::new();

for signer in signers {
    let provider = ProviderBuilder::new()
        .wallet(signer)
        .connect_http(url.clone());

    let tokens = [
        ITIP20::new(*base1.address(), provider.clone()),
        ITIP20::new(*base2.address(), provider.clone()),
        ITIP20::new(*quote.address(), provider.clone()),
    ];

    for token in tokens {
        receipts.push(
            token
                .mint(provider.address(), mint_amount).send().await?,
        );
        receipts.push(
            token
                .approve(STABLECOIN_EXCHANGE_ADDRESS, U256::MAX)
                .send()
                .await?,
        );
    }
}

```

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -306,35 +316,324 @@ async fn generate_transactions(
...
+                base1
+                    .mint(signer.address(), mint_amount)
+                    .gas_price(TEMPO_BASE_FEE as u128)
+                    .gas(300_000)
+                    .send()
+                    .await?,
+                base2
+                    .mint(signer.address(), mint_amount)
+                    .gas_price(TEMPO_BASE_FEE as u128)
+                    .gas(300_000)
```
0xKitsune: The fill provider should take care of gas price and gas. I dont think we need this.

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -306,35 +316,324 @@ async fn generate_transactions(
...
+    ) -> eyre::Result<(
+        IStablecoinExchangeInstance<DexProvider>,
+        Address,
+        Address,
+        Address,
+    )> {
+        println!("Sending DEX setup transactions...");
+
+        let tx_count = ProgressBar::new(6 + 11 * signers.len() as u64);
+        tx_count.tick();
```
0xKitsune: IMO I dont think we need a progress bar, in this case, simple print statements seem a bit cleaner here to help the user know what stage the setup is at.

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -306,35 +316,324 @@ async fn generate_transactions(
...
+        let mint_amount = U256::from(1000000000000000u128);
+        let first_order_amount = 1000000000000u128;
+
+        await_receipts(
+            &mut vec![
+                exchange.createPair(*base1.address()).send().await?,
+                exchange.createPair(*base2.address()).send().await?,
+            ],
+            &tx_count,
+        )
```
0xKitsune: We dont have to wait for this receipt here, we can continue with the other txs concurrently I believe.

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -306,35 +316,324 @@ async fn generate_transactions(
...
+
+            receipts.extend([
+                exchange
+                    .place(*base1.address(), first_order_amount, true, 0)
+                    .send()
+                    .await?,
+                exchange
+                    .place(*base2.address(), first_order_amount, true, 0)
+                    .send()
+                    .await?,
```
0xKitsune: IIUC we want to place large flip orders here via `placeFlip()` at price levels `0.999` and `1.001` for initial liquidity. You can use `priceToTick()` function to derive the tick needed for these price levels.

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -306,35 +316,324 @@ async fn generate_transactions(
...
+        >,
+        P: Provider<N>,
+    {
+        let min_order_amount = MIN_ORDER_AMOUNT;
+
+        // Place an order at exactly the dust limit (should succeed)
+        let mut tx = exchange
+            .place(token_address, min_order_amount, true, 0)
+            .into_transaction_request()
+            .with_gas_limit(1_000_000)
```
0xKitsune: Can we use a smaller gas limit? This might influence how many txs we are able to include in a block.

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -306,35 +316,324 @@ async fn generate_transactions(
...
+        P: Provider<N>,
+    {
+        let min_amount_out = 0;
+        let min_order_amount = MIN_ORDER_AMOUNT;
+
+        // Place an order at exactly the dust limit (should succeed)
+        let mut tx = exchange
+            .swapExactAmountIn(token_in, token_out, min_order_amount, min_amount_out)
+            .into_transaction_request()
+            .with_gas_limit(1_000_000)
```
0xKitsune: Same thing here can we adjust this gas limit? I do not believe we will need this large of a limit, lmk if I am overlooking something here though.

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -306,35 +316,324 @@ async fn generate_transactions(
...
+            .grantRole(*ISSUER_ROLE, caller)
+            .send()
+            .await?
+            .get_receipt()
+            .await?;
+        tx_count.inc(1);
+
+        Ok(token)
+    }
+}
```
0xKitsune: It looks like this is the same function used in the e2e test utils. Wdyt about importing this from there instead? @klkvr any reason not to?

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -475,3 +774,16 @@ fn monitor_tps(tx_counter: Arc<AtomicU64>) -> thread::JoinHandle<()> {
...
+    tx_count: &ProgressBar,
+) -> eyre::Result<()> {
+    for tx in pending_txs.drain(..) {
+        let receipt = tx.get_receipt().await?;
+        tx_count.inc(1);
+        assert!(receipt.status());
+    }
+
+    Ok(())
+}
```
0xKitsune: Same thing here, I think we could import this from the e2e test utils.

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -306,35 +316,324 @@ async fn generate_transactions(
...
+            .with_chain_id(chain_id)
+            .with_nonce(nonce)
+            .build_unsigned()?;
+
+        let signature = signer
+            .sign_transaction_sync(&mut tx)
+            .map_err(|e| eyre::eyre!("Failed to sign transaction: {e}"))?;
+        let mut payload = Vec::new();
+        tx.into_signed(signature).eip2718_encode(&mut payload);
+        Ok(payload)
```
0xKitsune: nit: this logic is repeated for each of the various functions creating txs. I think we could encapsulate this to be cleaner so we arent repeating this logic in many places.

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -306,35 +316,324 @@ async fn generate_transactions(
...
+            3 => dex::swap_in(&exchange, signer.clone(), nonce, chain_id, base2, quote),
+            4 => tip20::transfer(signer.clone(), nonce, chain_id, base1),
+            5 => tip20::transfer(signer.clone(), nonce, chain_id, base2),
+            v => unreachable!("Number {v} is outside the random range"),
         })
         .collect::<eyre::Result<Vec<_>>>()?;
 
     println!("Generated {} transactions", transactions.len());
     Ok(transactions)
 }
```
0xKitsune: nit: This works, you could also consider simplifying to something like to the following

```rust
use rand::{seq::SliceRandom, thread_rng};

let tx_builders: &[&dyn Fn(&SignerType, u64) -> eyre::Result<Vec<u8>>] = &[
    &|s, n| dex::place(&exchange, s.clone(), n, chain_id, base1),
    &|s, n| dex::place(&exchange, s.clone(), n, chain_id, base2),
    &|s, n| dex::swap_in(&exchange, s.clone(), n, chain_id, base1, quote),
    &|s, n| dex::swap_in(&exchange, s.clone(), n, chain_id, base2, quote),
    &|s, n| tip20::transfer(s.clone(), n, chain_id, base1),
    &|s, n| tip20::transfer(s.clone(), n, chain_id, base2),
];

let transactions: Vec<Vec<u8>> = params
    .into_par_iter()
    .tqdm()
    .map_init(thread_rng, |rng, (signer, nonce)| {
        tx_builders.choose(rng).unwrap()(signer, nonce)
    })
    .collect::<eyre::Result<_>>()?;
```

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -306,35 +316,324 @@ async fn generate_transactions(
...
+        token_out: Address,
+    ) -> eyre::Result<Vec<u8>>
+    where
+        N: Network<
+            UnsignedTx: SignableTransaction<alloy::signers::Signature> + RlpEcdsaEncodableTx,
+        >,
+        P: Provider<N>,
+    {
+        let min_amount_out = 0;
+        let min_order_amount = MIN_ORDER_AMOUNT;
```
0xKitsune: nit: we can pass these into the function directly.

bin/tempo-bench/src/cmd/max_tps.rs
RomanHodulak: Can't import from test-only code in another crate. The `setup_test_token` would need to be extracted outside of the test-only code, like under a feature flag.

bin/tempo-bench/src/cmd/max_tps.rs
RomanHodulak: Same response here + this one keeps track of the transaction progress.

bin/tempo-bench/src/cmd/max_tps.rs
RomanHodulak: This progress bar is already here via `rayon` and if you run the benchmark you'll see a couple of progress bars other than this one so it would seem weird to not have this here like it's stuck.

bin/tempo-bench/src/cmd/max_tps.rs
RomanHodulak: Using modules improves the organization and readability of this file by reducing the amount of concerns. They should be in their own files though, so I changed that.

bin/tempo-bench/src/cmd/max_tps.rs
RomanHodulak: At least for the first one I think it's more readable to have a name on that value instead of just `0`.

bin/tempo-bench/src/cmd/max_tps.rs
RomanHodulak: Nice idea, I might do something like this when I add the weighted random next.

bin/tempo-bench/src/cmd/max_tps/dex.rs
```diff
@@ -0,0 +1,270 @@
...
+type DexProvider = FillProvider<
+    JoinFill<
+        JoinFill<
+            alloy::providers::Identity,
+            JoinFill<GasFiller, JoinFill<BlobGasFiller, JoinFill<NonceFiller, ChainIdFiller>>>,
+        >,
+        WalletFiller<EthereumWallet>,
+    >,
+    RootProvider,
+>;
```
shekhirin: can probably use https://docs.rs/alloy/latest/alloy/providers/trait.Provider.html#method.erased

bin/tempo-bench/src/cmd/max_tps/dex.rs
```diff
@@ -0,0 +1,270 @@
...
+
+/// This method performs a one-time setup for sending a lot of transactions:
+/// * Adds a quote token and a couple of user tokens paired with the quote token.
+/// * Mints some large amount for all `signers` and approves unlimited spending for stablecoin
+///   exchange contract.
+/// * Seeds initial liquidity by placing flip orders
+pub(super) async fn setup(
```
shekhirin, line 14: can we make use of `buffer_unordered(max_concurrent_requests)` here that will be called on a chain of streams that do mint and approve calls? And then you consume the receipts from that stream, and update the progress bar

PR #775, opened 2025-11-06
chore: remove `backfill` and `run_pre_event_loop_init`

joshieDo:
Reth only requires a FCU or newPayload to move forward, including backfilling from its own p2p network. 

When the node starts, eventually CL will receive a `Finalization` and/or `Notarize` message that will send the FCU/nP to reth triggering whatever it needs to move forward.

Unless there is a non-obvious dependency on why we cant just remove this.

### DIFF
--- a/crates/commonware-node/src/consensus/application/executor.rs
+++ b/crates/commonware-node/src/consensus/application/executor.rs
@@ -11,7 +11,7 @@ use std::{sync::Arc, time::Duration};
 
 use alloy_primitives::B256;
 use alloy_rpc_types_engine::ForkchoiceState;
-use commonware_consensus::{Block as _};
+use commonware_consensus::Block as _;
 
 use commonware_macros::select;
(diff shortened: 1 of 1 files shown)


crates/commonware-node/src/consensus/application/executor.rs
```diff
@@ -192,80 +172,6 @@ where
...
-                "backfilling from consensus layer to execution layer \
-                failed; cannot recover from that",
-            )?;
-        } else {
-            bail!(
-                "execution layer is ahead of consensus layer; cannot deal \
-                with that; \
-                execution_height: `{latest_execution_block_number}`, \
-                consensus_height: `{finalized_consensus_height}`"
-            );
```
joshieDo, line 264: this is confusing to me. the latest is not really finalized (can be head), and there should be no issue on the head being further than finalized

PR #776, opened 2025-11-06
fix: update fee token checks to match latest spec

klkvr:
applies updates from https://github.com/tempoxyz/docs/pull/131/files

This adds some helpers to `TempoTxEnv` which assume that AA calls vec is not empty. To ensure this never panics I've added corresponding check to `TempoTransactionRequest -> TempoTxEnv` conversion

### DIFF
--- a/crates/revm/src/tx.rs
+++ b/crates/revm/src/tx.rs
@@ -83,12 +83,13 @@ impl TempoTxEnv {
     }
 
     /// Returns the first top-level call in the transaction.
-    pub fn first_call(&self) -> (&TxKind, &[u8]) {
+    pub fn first_call(&self) -> Option<(&TxKind, &[u8])> {
         if let Some(aa) = self.aa_tx_env.as_ref() {
-            let call = aa.aa_calls.first().unwrap();
-            (&call.to, &call.input)
+            aa.aa_calls
+                .first()
+                .map(|call| (&call.to, call.input.as_ref()))
         } else {
-            (&self.inner.kind, &self.inner.data)
+            Some((&self.inner.kind, &self.inner.data))
         }
     }
 
--- a/crates/revm/src/handler.rs
+++ b/crates/revm/src/handler.rs
@@ -1001,7 +1001,7 @@ where
     }
 
     // If tx.to() is a TIP-20 token, use that token as the fee token
-    if let Some(to) = ctx.tx.first_call().0.to().copied()
+    if let Some(to) = ctx.tx.first_call().and_then(|(kind, _)| kind.to().copied())
         && ctx.tx().calls().all(|(kind, _)| kind.to() == Some(&to))
         && is_valid_fee_token(ctx, to)?
     {


crates/revm/src/tx.rs
```diff
@@ -81,6 +81,26 @@ impl TempoTxEnv {
             .as_ref()
             .is_some_and(|aa| aa.subblock_transaction)
     }
+
+    /// Returns the first top-level call in the transaction.
+    pub fn first_call(&self) -> (&TxKind, &[u8]) {
+        if let Some(aa) = self.aa_tx_env.as_ref() {
+            let call = aa.aa_calls.first().unwrap();
```
mattsse: is this unwrap here safe?

crates/revm/src/tx.rs
```diff
@@ -81,6 +81,26 @@ impl TempoTxEnv {
...
+    }
+
+    /// Invokes the given closure for each top-level call in the transaction and
+    /// returns true if all calls returned true.
+    pub fn for_all_calls(&self, mut f: impl FnMut(&TxKind, &[u8]) -> bool) -> bool {
+        if let Some(aa) = self.aa_tx_env.as_ref() {
+            aa.aa_calls.iter().all(|call| f(&call.to, &call.input))
+        } else {
+            f(&self.inner.kind(), self.inner.input())
+        }
```
mattsse: should we make this an impl iterator instead? then we dont need to change this when we have more use cases for this

crates/revm/src/tx.rs
klkvr: yeah, we validate that `calls` are non-empty when decoding AA transaction, and when building TxEnv

maybe we should also enforce this during `AATxEnv` construction somehow, though it'd be a more invasive change

0xKitsune: Would it be safer to use `unwrap_or_default()` here instead so that we don't accidentically panic if there are changes to input validation?

crates/alloy/src/rpc/compat.rs
```diff
@@ -91,21 +91,27 @@ impl TryIntoTxEnv<TempoTxEnv, TempoBlockEnv> for TempoTransactionRequest {
...
                 // If key_type is not provided, default to secp256k1
                 let mock_signature = key_type
                     .as_ref()
                     .map(|kt| create_mock_aa_signature(kt, key_data.as_ref()))
                     .unwrap_or_else(|| create_mock_aa_signature(&SignatureType::Secp256k1, None));
 
-                Box::new(AATxEnv {
+                if calls.is_empty() {
+                    return Err(EthApiError::InvalidParams("empty calls list".to_string()));
+                }
```
0xKitsune, line 104: Should this be invalid? Im not positive here but thought that AA txs will still be valid even if calls is empty, I could be misremembering here though so feel free to lmk if this is not the case.

klkvr: this is already enforced during decoding of `TxAA` 

https://github.com/tempoxyz/tempo/blob/cd985a5196bde7cb124fc83a3b24366c6622c85e/crates/primitives/src/transaction/account_abstraction.rs#L185-L190

couldn't find this in the spec but I do think this is intentional. should still be possible to change though

0xKitsune: If this is per the spec then let's keep this for now and make a note to align on the expected behavior/update the spec/impl accordingly if needed.

crates/revm/src/tx.rs
klkvr: changed this to return an `Option` 14356138, it's not too bad and we can maybe re-introduce `.unwrap()` when handling of `None` here becomes more annoying

PR #777, opened 2025-11-06
fix(download/follow): change defaults to actual testnet values

Zygimantass:


bin/tempo/src/main.rs
```diff
@@ -46,7 +46,7 @@ struct TempoArgs {
     #[arg(
         long,
         value_name = "URL",
-        default_missing_value = "http://rpc-andante.tempoxyz.dev",
+        default_missing_value = "wss://eng:zealous-mayer@rpc.testnet.tempo.xyz",
```
fgimenez, line 49: :see_no_evil:

Zygimantass: yep, intentional, auth will be gone by the time it's public, so nbd

PR #780, opened 2025-11-07
feat(subblocks): forward transactions from RPC to subblocks service

klkvr:
Overrides `send_transaction` implementation so that subblock transactions are forwarded to subblocks service and all others are handled by inner `EthApi::send_transaction`

@SuperFluffy I wasn't able to figure out how to make `tokio::sync::broadcast` work with commonware/futures select! macro, so changed it to `tokio::select! { biased; }` which should be equivalent

### DIFF
--- a/crates/commonware-node/src/subblocks.rs
+++ b/crates/commonware-node/src/subblocks.rs
@@ -23,7 +23,7 @@ use commonware_cryptography::{
     ed25519::{PrivateKey, PublicKey, Signature},
 };
 use commonware_p2p::{Receiver, Recipients, Sender};
-use commonware_runtime::{FutureExt, Handle, Metrics, Pacer, Spawner};
+use commonware_runtime::{Handle, Metrics, Pacer, Spawner};
 use eyre::{Context, OptionExt};
 use futures::{FutureExt as _, StreamExt, channel::mpsc};
 use indexmap::IndexMap;


Cargo.toml
```diff
@@ -113,60 +113,60 @@ tempo-contracts = { path = "crates/contracts", default-features = false }
...
-reth-cli-runner = { git = "https://github.com/paradigmxyz/reth", rev = "99fe175" }
-reth-node-metrics = { git = "https://github.com/paradigmxyz/reth", rev = "99fe175" }
-reth-network = { git = "https://github.com/paradigmxyz/reth", rev = "99fe175" }
-reth-transaction-pool = { git = "https://github.com/paradigmxyz/reth", rev = "99fe175" }
-reth-engine-local = { git = "https://github.com/paradigmxyz/reth", rev = "99fe175" }
-reth-ethereum-consensus = { git = "https://github.com/paradigmxyz/reth", rev = "99fe175" }
-reth-ethereum = { git = "https://github.com/paradigmxyz/reth", rev = "99fe175" }
-reth-ethereum-cli = { git = "https://github.com/paradigmxyz/reth", rev = "99fe175" }
-reth-e2e-test-utils = { git = "https://github.com/paradigmxyz/reth", rev = "99fe175" }
+reth = { git = "https://github.com/paradigmxyz/reth", rev = "4d9d712" }
```
SuperFluffy, line 116: Obligatory question 🙂: necessary bumping reth to make this work?

crates/commonware-node/src/subblocks.rs
```diff
@@ -141,15 +145,19 @@ impl<TContext: Spawner + Metrics> Actor<TContext> {
...
+                biased;
+
                 // Handle messages from consensus engine and service handle.
-                action = self.actions_rx.next() => {
-                    let Some(action) = action else { break; };
+                Some(action) = self.actions_rx.next() => {
                     self.on_new_message(action);
                 },
+                // Handle new subblock transactions.
+                Ok(transaction) = self.subblock_transactions_rx.recv() => {
```
SuperFluffy, line 156: Can you throw a `Pacer::pace` call on here, similar to how we do it for other interactions with reth? I don't know if that will actually do anything, but let it serve as a reminder that we are "outside" of reth.

crates/commonware-node/src/consensus/engine.rs
```diff
@@ -450,9 +450,4 @@ where
         // component failed.
         .wrap_err("one of the consensus engine's actors failed")
     }
-
-    /// Returns a handle to the subblocks service.
-    pub fn subblocks_mailbox(&self) -> subblocks::Mailbox {
```
SuperFluffy, line 455: I think with this you can lock down the entire subblocks module once more, make it crate-private?

crates/e2e/src/tests/subblocks.rs
```diff
@@ -106,8 +107,12 @@ fn submit_subblock_tx(node: &ValidatorNode) -> TxHash {
 
     let tx = TempoTxEnvelope::AA(tx.into_signed(signature.into()));
     let tx_hash = *tx.tx_hash();
-    node.subblocks
-        .add_transaction(tx.try_into_recovered().unwrap());
+    node.node
+        .node
+        .eth_api()
+        .send_raw_transaction(tx.encoded_2718().into())
```
SuperFluffy, line 113: Cool, so it figures out itself that it will send this as a subblock?

Cargo.toml
klkvr: yeah this depends on https://github.com/paradigmxyz/reth/pull/19564

crates/e2e/src/tests/subblocks.rs
klkvr: yep

crates/commonware-node/src/subblocks.rs
```diff
@@ -127,29 +131,33 @@ impl<TContext: Spawner + Metrics> Actor<TContext> {
...
+                biased;
+
                 // Handle messages from consensus engine and service handle.
-                action = self.actions_rx.next() => {
-                    let Some(action) = action else { break; };
+                Some(action) = self.actions_rx.next() => {
                     self.on_new_message(action);
                 },
+                // Handle new subblock transactions.
+                Ok(transaction) = self.subblock_transactions_rx.recv().pace(&self.context, Duration::from_millis(20)) => {
```
mattsse: I believe this will now block until we receive a new subblock preventing any further action handling in the deterministic rt

because in determ runtime pace is actually a block_in_place effectively if I understood it correctly

SuperFluffy: https://github.com/commonwarexyz/monorepo/blob/512369fe57fa470471577f18d6468c52fa3f7b18/runtime/src/deterministic.rs#L1213-L1227

Yeah, that's right. The idea is to ensure that we get exactly the same order of operations, ideally within the duration argument.

But good catch. My request is definitely wrong. We cannot rely on getting any blocks, and it shows in the tests not completing.

Now I wonder if there is any way to make this work at all or if we simply need to forget the idea of creating reproducible tests. :(

@klkvr Sorry for the suggestion.

crates/commonware-node/src/subblocks.rs
SuperFluffy: This was wrong

crates/commonware-node/src/subblocks.rs
```diff
@@ -96,7 +99,7 @@ pub struct Actor<TContext> {
     subblock_transactions: Arc<Mutex<IndexMap<TxHash, Arc<Recovered<TempoTxEnvelope>>>>>,
```
mattsse, line 99: this is now effectively a pool that only evicts if the transactions are invalid, or included?

I think this should work and drain pretty quickly
but maybe it isn't the best experience for the user if this tx silently drops, for example due to lack of fees

imo for now this is okay, but perhaps we want to do smth here

could we add an issue for this so that we dont forget?

klkvr: right now all transactions are free so main reason why transactions would fail is nonce issues I believe

tracking here https://github.com/tempoxyz/tempo/issues/784

PR #782, opened 2025-11-07
chore: canonicalize after backfill & docs to CL backfill edge case

joshieDo:
CL backfill is required for the following low probability edge case.
```rust
    /// Recovers from a consensus-execution layer height mismatch at startup and backfills missing blocks from CL to EL.
    ///
    /// Normally, Reth's EL P2P can handle backfilling missing blocks from other nodes just fine.
    /// However, this function addresses a specific edge case that occurs when:
    /// 1. All nodes in the network crash simultaneously (e.g., hitting the same bug at the same block)
    /// 2. The execution layer (EL) failed to persist the last N finalized blocks before shutdown
    ///    (e.g., due to --persistence-threshold, Reth may keep recent blocks in memory only)
    /// 3. On restart, CL believes it sent block H to EL, but EL only has H-N
    /// 4. Because all nodes experienced the same issue, no peer has these blocks persisted in their EL either
    /// 5. Sending the finalized tip hash from CL to EL does not help - no node has the missing
    ///    blocks persisted in EL
    ///
    /// Without this recovery mechanism, the network would deadlock: without the last finalized
    /// block, no proposer can propose the next block. CL expects EL to have blocks it doesn't,
    /// and EL's normal P2P backfill won't help because no peer has them in their execution
    /// layer either.
```


crates/commonware-node/src/consensus/application/executor.rs
```diff
@@ -281,8 +294,18 @@ where
...
                 )?;
+
+            self.canonicalize(
+                Span::current(),
+                HeadOrFinalized::Finalized,
+                finalized_height,
+                finalized_digest,
+            )
+            .await
+            .wrap_err("failed setting initial canonical state; can't go on like this")?;
```
joshieDo, line 306: i think we only need it here right? only if we needed to backfill, and only when the EL has the blocks buffered?

SuperFluffy: Cases:

1. CL > EL - need to send FCU with head = finalized = CL.finalized
2. CL = EL - no need to FCU
3. CL < EL - no need to FCU

(I am assuming that in cases 2. and 3. EL will have the appropriate forkchoice state).

Yupp, I think that's right.

crates/commonware-node/src/consensus/application/executor.rs
```diff
@@ -193,8 +197,25 @@ where
...
+    /// However, this function addresses a specific edge case that occurs when:
+    /// 1. All nodes in the network crash simultaneously (e.g., hitting the same bug at the same block)
+    /// 2. The execution layer (EL) failed to persist the last N finalized blocks before shutdown
+    ///    (e.g., due to --persistence-threshold, Reth may keep recent blocks in memory only)
+    /// 3. On restart, CL believes it sent block H to EL, but EL only has H-N
+    /// 4. Because all nodes experienced the same issue, no peer has these blocks persisted in their EL either
+    /// 5. Sending the finalized tip hash from CL to EL does not help - no node has the missing
+    ///    blocks persisted in EL
+    ///
+    /// Without this recovery mechanism, the network would deadlock: without the last finalized
```
mattsse: can we also clarify what the actual recovery mechanism is here, didnt fully get this from the docs

crates/commonware-node/src/consensus/application/executor.rs
SuperFluffy: As an aside: this FCU is indeed only necessary in the EL < CL case on startup - but here it *must* sent so that the subsequent NEW PAYLOAD requests go through.

Once the actor enters its event loop, CL immediately sends the tip of the finalized chain, and the actor forwards it to the EL.

crates/commonware-node/src/consensus/application/executor.rs
SuperFluffy: Good feedback, thank you @mattsse.

Not always seeing what's obvious and what's not anymore with this code.

crates/commonware-node/src/consensus/application/executor.rs
SuperFluffy: As a sidenote - I'd rather have this particular change in a different PR, because it is a business logic change vs just adding a couple docs.

Issue #769, opened 2025-11-06
e2e test to ensure node starts after pipeline sync if consensus layer falls behind


SuperFluffy:
We ran into a crash-loop when the execution layer was ahead of the consensus layer. On startup, the consensus layer would refuse to start, claiming it could not handle this scenario.

PR https://github.com/tempoxyz/tempo/pull/768 addresses the issue.

A test covering this scenario would be good:

1. start a network of 4 nodes
2. let them progress a while
3. shut down 1 node
4. let the network continue to run until the dead node is sufficiently far behind to a trigger a pipeline sync
5. restart the node, observe the pipeline sync
6. immediately shut down the node once pipeline sync completes
7. here, we should be able to observe that consensus is behind execution by tracking the last finalized height according to the node's marshal actor
8. restart the node
9. observe that it runs normally and catches up to the network.

0xAysh:
Hello , @SuperFluffy. I'd like to work in this, just need a green light.

0xAysh:
Just went ahead made a pr(#3712) with the test described @SuperFluffy

PR #783, opened 2025-11-07
test(e2e): implement validator restart and recovery test scenarios

fgimenez:
Adds e2e tests for validator restart/recovery scenarios, adapted from https://github.com/commonwarexyz/monorepo/pull/2140. These tests verify that validators can be killed and successfully rejoin the network, catching up to the current state.

As part of this patch node setup was reworked: `setup_validators` now returns a `Vec<PreparedNode>`. And each prepared validator exposes `PreparedNode::start -> RunningNode`. Similarly, `RunningNode::stop` returns a `PreparedNode`, which makes restarting it easy.

## Issues/future work

Right now only the consensus layer gets stopped/is restarted. We actually need to also restart the execution layer to make this test complete. Follow-up issue is here: https://github.com/tempoxyz/tempo/issues/909

There are more test scenarios that could be tested. One such test is described in https://github.com/tempoxyz/tempo/issues/769

### DIFF
--- a/crates/e2e/src/lib.rs
+++ b/crates/e2e/src/lib.rs
@@ -43,7 +43,7 @@ pub const CONSENSUS_NODE_PREFIX: &str = "consensus";
 pub const EXECUTION_NODE_PREFIX: &str = "execution";
 
 /// A Tempo node with lazily started consensus engine.
-pub struct Node {
+pub struct PreparedNode {
     pub uid: String,
 
     /// Execution-layer node. Spawned in the background but won't progress unless consensus engine is started.
@@ -57,44 +57,7 @@ pub struct Node {
     pub oracle: simulated::Oracle<PublicKey>,
 }
 
-/// A Tempo node with lazily started consensus engine.
-pub struct RunningNode {
-    pub uid: String,
-
-    pub consensus_config: consensus::Builder<Control<PublicKey>, Context, SocketManager<PublicKey>>,
-    pub consensus_handle: Handle<eyre::Result<()>>,
-
-    /// Execution-layer node. Spawned in the background but won't progress unless consensus engine is started.
-    pub execution_node: ExecutionNode,
-
-    /// Public key of the validator.
-    pub public_key: PublicKey,
-
-    pub oracle: simulated::Oracle<PublicKey>,
-}
-
-impl RunningNode {
-    pub fn stop(self) -> Node {
-        let Self {
-            uid,
-            execution_node,
-            public_key,
-            oracle,
-            consensus_config,
-            consensus_handle,
-        } = self;
-        consensus_handle.abort();
-        Node {
-            uid,
-            execution_node,
-            public_key,
-            consensus_config,
-            oracle,
-        }
-    }
-}
-
-impl Node {
+impl PreparedNode {
     pub async fn start(self) -> RunningNode {
         let Self {
             uid,
@@ -173,6 +136,43 @@ impl Node {
     }
 }
 
+/// A Tempo node with lazily started consensus engine.
+pub struct RunningNode {
+    pub uid: String,
+
+    pub consensus_config: consensus::Builder<Control<PublicKey>, Context, SocketManager<PublicKey>>,
+    pub consensus_handle: Handle<eyre::Result<()>>,
+
+    /// Execution-layer node. Spawned in the background but won't progress unless consensus engine is started.
+    pub execution_node: ExecutionNode,
+
+    /// Public key of the validator.
+    pub public_key: PublicKey,
+
+    pub oracle: simulated::Oracle<PublicKey>,
+}
+
+impl RunningNode {
+    pub fn stop(self) -> PreparedNode {
+        let Self {
+            uid,
+            execution_node,
+            public_key,
+            oracle,
+            consensus_config,
+            consensus_handle,
+        } = self;
+        consensus_handle.abort();
+        PreparedNode {
+            uid,
+            execution_node,
+            public_key,
+            consensus_config,
+            oracle,
+        }
+    }
+}
+
 /// The test setup run by [`run`].
 #[derive(Clone)]
 pub struct Setup {
@@ -199,7 +199,7 @@ pub async fn setup_validators(
         connect_execution_layer_nodes,
         ..
     }: Setup,
-) -> (Vec<Node>, simulated::Oracle<PublicKey>) {
+) -> (Vec<PreparedNode>, simulated::Oracle<PublicKey>) {
     let (network, oracle) = Network::new(
         context.with_label("network"),
         simulated::Config {
@@ -296,7 +296,7 @@ pub async fn setup_validators(
             time_to_build_subblock: Duration::from_millis(100),
         };
 
-        nodes.push(Node {
+        nodes.push(PreparedNode {
             execution_node,
             public_key: private_key.public_key(),
             consensus_config: engine_config,


crates/e2e/src/tests/restart.rs
```diff
@@ -0,0 +1,420 @@
...
+        );
+        drop(killed_node);
+
+        // Wait for remaining validators to reach restart height
+        info!(
+            "Waiting for {} remaining validators to reach restart height {}",
+            setup.how_many - 1,
+            setup.restart_height
+        );
+        wait_for_height(&context, setup.how_many - 1, setup.restart_height).await;
```
SuperFluffy: For the case `how_many = 3`, this should not work: quorum for 3 validators is 3. If one of them drops out, then no progress should have been made in the meantime, and `wait_for_height` should never resolve.

crates/e2e/src/tests/restart.rs
```diff
@@ -0,0 +1,420 @@
...
+
+    ValidatorNode {
+        node,
+        public_key: signer.public_key(),
+        start_engine: Some(Box::pin(async move {
+            // Use new channel IDs for the restarted validator to avoid conflicts
+            // Add offset of 100 to all channel IDs
+            let pending = oracle_clone
+                .control(signer.public_key())
+                .register(100)
```
SuperFluffy: This should break the original validator node: the channels registered here must exactly match the original channels of the other validators.

If there are conflicts registering here, this likely means that the original validator has not been shutdown.

crates/e2e/src/tests/restart.rs
```diff
@@ -0,0 +1,420 @@
...
+        let idx = context.gen_range(0..nodes.len());
+        let killed_node = nodes.remove(idx);
+        let killed_public_key = killed_node.public_key.clone();
+
+        // Kill the validator by dropping it
+        info!(
+            "Killing validator at index {} with public key {:?}",
+            idx, killed_public_key
+        );
+        drop(killed_node);
```
SuperFluffy: Dropping the node is not sufficient - its handle must be actively aborted (like a tokio task).

fixed - aborting now

crates/e2e/src/lib.rs
SuperFluffy, line 1: reworked this crate by backporting some changes introduced in #585

crates/e2e/src/lib.rs
```diff
@@ -7,78 +7,199 @@
...
+            consensus_config,
+            oracle,
+        }
+    }
 }
 
-impl ValidatorNode {
-    pub fn start(&mut self) -> impl Future<Output = ()> + use<> {
-        self.start_engine.take().expect("engine already started")
+impl Node {
```
klkvr: let's move this impl block to be after `Node` definition

crates/e2e/src/tests/backfill.rs
SuperFluffy, line 1: test conditions are unchanged, just catching up to the e2e crate refactor

crates/e2e/src/tests/dkg.rs
SuperFluffy, line 1: test conditions are unchanged, just catching up to the e2e crate refactor

crates/e2e/src/tests/linkage.rs
SuperFluffy, line 1: test conditions are unchanged, just catching up to the e2e crate refactor

crates/e2e/src/tests/mod.rs
SuperFluffy, line 1: test conditions are unchanged, just catching up to the e2e crate refactor

crates/e2e/src/tests/subblocks.rs
SuperFluffy, line 1: test conditions are unchanged, just catching up to the e2e crate refactor

crates/e2e/src/lib.rs
```diff
@@ -7,78 +7,199 @@
...
+    pub fn stop(self) -> PreparedNode {
+        let Self {
+            uid,
+            execution_node,
+            public_key,
+            oracle,
+            consensus_config,
+            consensus_handle,
+        } = self;
+        consensus_handle.abort();
```
SuperFluffy, line 165: This here is not complete: we only stop the consensus layer and keep the execution layer running.

We actually need to stop the execution node as well and restart it (keeping in mind the non-persistence of finalized blocks, etc - not sure if this artificial setup would test that tho).

crates/e2e/src/tests/restart.rs
SuperFluffy: fixed - the consensus engine is now properly aborted and the network makes no progress in the meantime.

crates/e2e/src/tests/restart.rs
SuperFluffy: fixed - the node now registers the same subchannels.

crates/e2e/src/tests/backfill.rs
```diff
@@ -6,60 +6,86 @@ use commonware_runtime::{
...
     // Wait for chain to advance before starting the last node
-    while nodes[0].node.node.provider.last_block_number().unwrap() < blocks_before_join {
+    while running[0]
+        .execution_node
+        .node
+        .provider
+        .last_block_number()
+        .unwrap()
+        < blocks_before_join
+    {
```
klkvr, line 54: some helper getters would be useful here

crates/e2e/src/tests/backfill.rs
```diff
@@ -6,60 +6,86 @@ use commonware_runtime::{
...
 
     // Start all nodes except the last one
-    let mut last = nodes.pop().unwrap();
-    for node in &mut nodes {
-        node.start().await;
-    }
+    let last = nodes.pop().unwrap();
+    let mut running = join_all(nodes.into_iter().map(|node| node.start())).await;
+
+    link_validators(&mut oracle, &running, linkage.clone(), None).await;
```
klkvr, line 44: can we keep the `link_validators` call inside the `node.start()`?

SuperFluffy: No, because conceptually that's not how the linking is intended to work (and if you stare at logs, it's actually complaining that all the links are being duplicates).

Commonware mentioned that the intended use is to establish the architecture of the network ahead of time by establishing links, and only then start the nodes.

Having said that, some tests don't do a good job at following through with that.

crates/e2e/src/tests/backfill.rs
SuperFluffy: Yeah, but I'd leave that as a followup (I didn't touch the test itself, just what followed from restructuring the crate).

crates/e2e/src/lib.rs
SuperFluffy: Didn't see the comment - already done (and the node types renamed).

crates/e2e/src/tests/backfill.rs
klkvr: > Commonware mentioned that the intended use is to establish the architecture of the network ahead of time by establishing links, and only then start the nodes.

hmm then it doesn't make sense for `link_validators` to take `&[RunningNode]`? right now none of the tests actually link validators before starting the nodes

klkvr: I'd be OK with moving this call to `setup_validators` if that's possible, but relying on tests to invoke that as an extra mandatory setup step seems wrong/non-ergonomic

SuperFluffy: > hmm then it doesn't make sense for link_validators to take &[RunningNode]? right now none of the tests actually link validators before starting the nodes

Yes, that's right and yes on the running node. I wanted to just get rid of that extra step extracting all the pubkeys to make it work with the first version of that function so I don't have to write yet another `iter.map.collect`

> I'd be OK with moving this call to setup_validators if that's possible, but relying on tests to invoke that as an extra mandatory setup step seems wrong/non-ergonomic

Ambivalent on putting it in `setup_validator`, because there is a real test case of not linking certain validators (right now we just do all-to-all connections).

I guess I could make it take `Some(Link)` though and leave an escape hatch. Wdyt?

SuperFluffy: Just to add: none of this was intended to be a full refactor, but mainly a way to make certain things easier/possible in the context of dynamic dkg. And it helped here.

Let's crate followup issues to clear this up?

klkvr: > I guess I could make it take Some(Link) though and leave an escape hatch. Wdyt?

yeah this sgtm, tracking here https://github.com/tempoxyz/tempo/issues/920

PR #787, opened 2025-11-08
feat!(precompiles):  migrate TIP20

0xrusowsky:


crates/precompiles/src/tip20/rewards.rs
```diff
@@ -149,7 +136,10 @@ impl<'a, S: PrecompileStorageProvider> TIP20Token<'a, S> {
             return Ok(());
         }
 
-        self.set_last_update_time(accrue_to_timestamp)?;
+        // NOTE(rusowsky): first limb = u64, so it should be fine.
+        // however, it would be easier to always work with U256, since
+        // there is no possible slot packing in this slot (surrounded by U256)
+        self.set_last_update_time(accrue_to_timestamp.as_limbs()[0])?;
```
0xKitsune: Lets use `accrue_to_timestamp.to::<u64>()` here or `try_into` if this is fallible.

crates/precompiles/src/tip20/rewards.rs
```diff
@@ -333,7 +327,7 @@ impl<'a, S: PrecompileStorageProvider> TIP20Token<'a, S> {
             .ok_or(TempoPrecompileError::under_overflow())?;
         self.set_scheduled_rate_decrease_at(end_time, new_rate)?;
 
-        stream.delete(self.storage, self.token_address)?;
+        self.clear_streams(stream_id)?;
```
0xKitsune, line 330: nit: Just a note for the future, similar to how `set_streams` was renamed to `sstore_streams` previously, it might be nice to rename `clear_streams` to `delete_streams` to match Solidity semantics. Very minor and we don't need to do this now.

crates/precompiles/src/tip20/rewards.rs
```diff
@@ -470,82 +473,64 @@ impl<'a, S: PrecompileStorageProvider> TIP20Token<'a, S> {
...
+        self.sstore_global_reward_per_token(value)
     }
 
     /// Gets the timestamp of the last reward update from storage.
     fn get_last_update_time(&mut self) -> Result<u64> {
-        Ok(self
-            .storage
-            .sload(self.token_address, slots::LAST_UPDATE_TIME)?
-            .to::<u64>())
+        self.sload_last_update_time()
```
0xKitsune, line 498: Now that we have better abstractions for storage, I think we should remove these smaller getter functions and use `sload_var` directly. This applies to all getter/setter functions that just call `sload`/`sstore`.

crates/precompiles/src/tip20/rewards.rs
```diff
@@ -563,13 +548,21 @@ impl<'a, S: PrecompileStorageProvider> TIP20Token<'a, S> {
...
             }
         } else if !to_delegate.is_zero() {
             let opted_in_supply = U256::from(self.get_opted_in_supply()?)
                 .checked_add(amount)
                 .ok_or(TempoPrecompileError::under_overflow())?;
-            self.set_opted_in_supply(opted_in_supply)?;
+            self.set_opted_in_supply(
+                opted_in_supply
+                    .try_into()
+                    .map_err(|_| TempoPrecompileError::under_overflow())?,
```
0xKitsune, line 564: nit: Note for the future, we use this pattern frequently now where we `try_into().map_err()`. We could consider adding a trait abstractions for this to make this cleaner, similar to what I mentioned previously for `.checked_add(amount).ok_or(TempoPrecompileError::under_overflow())?;`.

Something like:

```rust 
pub trait TryToU128 {
    fn try_to_u128(self) -> Result<u128, TempoPrecompileError>;
}

impl TryToU128 for T {
    fn try_to_u128(self) -> Result<u128, TempoPrecompileError> {
        u128::try_from(self).map_err(|_| TempoPrecompileError::under_overflow())
    }
}

```

crates/precompiles/src/tip20/rewards.rs
0xrusowsky: sounds good. 

for convenience tough i'd keep the getters that are used by the dispatcher

0xKitsune: Sounds good to me, yeah lets focus on just migrating the storage abstractions for now, we can clean up in a separate PR.

crates/precompiles/src/tip20/mod.rs
```diff
@@ -61,34 +59,39 @@ pub fn address_to_token_id_unchecked(address: Address) -> u64 {
...
+    domain_separator: B256,
+    quote_token: Address,
+    next_quote_token: Address,
+    transfer_policy_id: u64,
+
+    // TIP20 Token
+    total_supply: U256,
+    balances: Mapping<Address, U256>,
+    allowances: Mapping<Address, Mapping<Address, U256>>,
+    nonces: Mapping<Address, U256>,
```
0xKitsune, line 81: nonces and salts are no longer in the reference contract. We can update this in a subsequent PR.

crates/precompiles/src/tip20/mod.rs
```diff
@@ -336,27 +320,24 @@ impl<'a, S: PrecompileStorageProvider> TIP20Token<'a, S> {
         self.check_role(msg_sender, DEFAULT_ADMIN_ROLE)?;
 
         let next_quote_token = self.next_quote_token()?;
+        println!("LOADED NEXT TOKEN: {}", next_quote_token);
```
0xKitsune: Lets remove this `println`

PR #789, opened 2025-11-08
fix: block withdrawals that would drain liquidity for current transaction

klkvr:
Applies latest spec changes from https://github.com/tempoxyz/docs/pull/140

- `has_liquidity` is renamed to `reserve_liquidity`. it now returns error when there's not enough liquidity, and also increments `pending_fee_swap_in` with the max swap amount
- `fee_swap` is renamed to `reserve_liquidity` and now its only task is to decremenet `pending_fee_swap_in` with `refund_amount`
- In `calculate_burn_amounts`, `amount_user_token` is no longer validated against pending swaps as this check is redundant, see https://github.com/tempoxyz/docs/blob/c4d0f5fc2fd09a391867a6c973e28922b3111f5a/specs/src/FeeAMM.sol#L278

Not sure if it's actually breaking but I think this change might change gas usage in some edge cases, so marking as such

### DIFF
--- a/crates/precompiles/src/tip_fee_manager/amm.rs
+++ b/crates/precompiles/src/tip_fee_manager/amm.rs
@@ -127,7 +127,7 @@ impl<'a, S: PrecompileStorageProvider> TipFeeManager<'a, S> {
             .ok_or(TempoPrecompileError::under_overflow())
     }
 
-    /// Execute a swap from one fee token to another
+    /// Releases `refund_amount` of liquidity that was locked by `reserve_liquidity`
     pub fn release_liquidity(
         &mut self,
         user_token: Address,
--- a/crates/precompiles/src/tip_fee_manager/mod.rs
+++ b/crates/precompiles/src/tip_fee_manager/mod.rs
@@ -127,7 +127,8 @@ impl<'a, S: PrecompileStorageProvider> TipFeeManager<'a, S> {
 
     /// Collects fees from user before transaction execution.
     ///
-    /// Determines fee token, verifies pool liquidity for swaps if needed, and transfers max fee amount.
+    /// Determines fee token, verifies pool liquidity for swaps if needed, reserves liquidity
+    /// for the max fee amount and transfers it to the fee manager.
     /// Unused gas is later returned via collect_fee_post_tx
     pub fn collect_fee_pre_tx(
         &mut self,


crates/precompiles/src/tip_fee_manager/amm.rs
```diff
@@ -123,21 +123,33 @@ impl<'a, S: PrecompileStorageProvider> TIPFeeAMM<'a, S> {
         })
     }
 
-    /// Checks if pool has enough liquidity for a fee swap
-    pub fn has_liquidity(
+    /// Ensures that pool has enough liquidity for a fee swap
+    pub fn reserve_liquidity(
```
mattsse: can we also document that this sets the pending fee swap in to pool for the user+val token

crates/precompiles/src/tip_fee_manager/amm.rs
```diff
@@ -123,21 +123,33 @@ impl<'a, S: PrecompileStorageProvider> TIPFeeAMM<'a, S> {
...
         validator_token: Address,
-        amount_in: U256,
-    ) -> Result<bool> {
+        max_amount: U256,
+    ) -> Result<()> {
         let pool_id = PoolKey::new(user_token, validator_token).get_id();
-        let amount_out = amount_in
+        let amount_out = max_amount
             .checked_mul(M)
             .and_then(|product| product.checked_div(SCALE))
```
mattsse: checked_div should always return some here because rhs is non zero?

crates/precompiles/src/tip_fee_manager/amm.rs
```diff
@@ -123,21 +123,33 @@ impl<'a, S: PrecompileStorageProvider> TIPFeeAMM<'a, S> {
...
             .ok_or(TempoPrecompileError::under_overflow())?;
```
mattsse, line 84: so this can only overflow

crates/precompiles/src/tip_fee_manager/amm.rs
```diff
@@ -580,12 +592,11 @@ impl<'a, S: PrecompileStorageProvider> TipFeeManager<'a, S> {
...
 
         // Check that withdrawal does not violate pending swaps
-        let available_user_token = self.get_effective_user_reserve(pool_id)?;
-        let available_validator_token = self.get_effective_validator_reserve(pool_id)?;
-
-        if amount_user_token > available_user_token {
-            return Err(TIPFeeAMMError::insufficient_reserves().into());
+        if self.storage.spec() < TempoHardfork::Moderato {
+            // Keep this for gas usage to be backwards compatible
+            self.get_effective_user_reserve(pool_id)?;
```
klkvr: this seems to be the only change that needs to be hardfork-gated

klkvr: reverted it as it's not directly relevant for the PR and allows to make it non-breaking

crates/precompiles/src/tip_fee_manager/amm.rs
```diff
@@ -112,28 +128,23 @@ impl<'a, S: PrecompileStorageProvider> TipFeeManager<'a, S> {
     }
 
     /// Execute a swap from one fee token to another
-    pub fn fee_swap(
+    pub fn release_liquidity(
```
0xKitsune, line 131: Looks like the comments for this function might be outdated, can we update?

crates/precompiles/src/tip_fee_manager/mod.rs
```diff
@@ -141,10 +141,8 @@ impl<'a, S: PrecompileStorageProvider> TipFeeManager<'a, S> {
         let validator_token = self.get_validator_token(beneficiary)?;
 
         // Verify pool liquidity if user token differs from validator token
-        if user_token != validator_token
-            && !self.has_liquidity(user_token, validator_token, max_amount)?
-        {
-            return Err(FeeManagerError::insufficient_liquidity().into());
+        if user_token != validator_token {
```
0xKitsune, line 145: Can we update the comment above to mention that this is not only checking for pool liquidity now?

PR #791, opened 2025-11-08
chore: regenerate test genesis, fix tests, and enforce genesis consistency in CI

fgimenez:
Updates the test genesis file to match the latest generator output and ensures it stays in sync going forward.

  - regenerated test-genesis.json using the genesis generator tool with 100 accounts
  - fixed test assertions that were incompatible with the updated genesis
  - removed duplicate base-fee-test.json (consolidated to single test genesis)
  - added CI check that validates test-genesis.json matches generator output before running tests

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -58,7 +58,7 @@ jobs:
       - uses: rui314/setup-mold@v1
       - uses: mozilla-actions/sccache-action@v0.0.9
       - name: Generate genesis
-        run: cargo run --bin tempo-xtask -- generate-genesis --accounts 100 --output generated-test-genesis.json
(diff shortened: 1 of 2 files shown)


crates/node/tests/assets/test-genesis.json
```diff
@@ -28,7 +28,7 @@
   "gasLimit": "0xf761ef61000",
   "difficulty": "0x0",
   "mixHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
-  "coinbase": "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266",
+  "coinbase": "0x0000000000000000000000000000000000000000",
```
klkvr, line 31: should we revert this instead of changing the test?

klkvr: looks like it's default behavior of `generate-genesis` though, so maybe it's fine

PR #796, opened 2025-11-08
feat: TIP20 rewards registry + Vec<T> helper traits

0xrusowsky:
This PR:
- migrates TIP20 rewards registry
- introduces `trait VecMappingExt`, `trait VecSlotExt` which provide qol methods to interact with individual elements of `Vec<T>` without having to always access its totality
- sets up the base for enums (still wip)

### DIFF
--- a/bin/tempo-sidecar/src/cmd/simple_arb.rs
+++ b/bin/tempo-sidecar/src/cmd/simple_arb.rs
@@ -9,10 +9,8 @@ use eyre::Context;
 use itertools::Itertools;
 use std::{collections::HashSet, time::Duration};
 use tempo_precompiles::{
-    TIP_FEE_MANAGER_ADDRESS,
-    TIP20_FACTORY_ADDRESS, // tip_fee_manager::ITIPFeeAMM,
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/tip20_rewards_registry/mod.rs
```diff
@@ -54,135 +46,50 @@ impl<'a, S: PrecompileStorageProvider> TIP20RewardsRegistry<'a, S> {
...
 
-        let array_slot = mapping_slot(end_time.to_be_bytes(), slots::STREAMS_ENDING_AT);
-        let index = self.storage.sload(self.address, array_slot)?;
-        self.set_stream_index(stream_key, index)?;
-
-        self.push_stream_ending_at_timestamp(token, end_time)?;
-
-        Ok(())
+        self.sstore_stream_index(stream_key, U256::from(length))?;
+        StreamEndingAt::push(self, end_time, token)
```
0xKitsune, line 55: This should work for now but lets plant to update to match the same syntax as other storage variables in a future PR.

PR #803, opened 2025-11-09
feat!(precompiles): migrate TIP403 registry

0xrusowsky:
since the `contract` macro was getting exponentially larger and more complex due to having to distinguish between user-defined structs (not packable in the layout) and enums (packable, as they are equivalent to u8s), i ended up not supporting them in the `contract` macro.

Instead, we define them as u8s, and leverage the sol! macro's `TryInto<u8>` impl, as well as their `as u8` capabilities.


crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -1,45 +1,36 @@
...
-    pub(super) const POLICY_DATA: U256 = uint!(1_U256);
-    pub(super) const POLICY_SET: U256 = uint!(2_U256);
-}
-
-#[derive(Debug)]
-pub struct TIP403Registry<'a, S: PrecompileStorageProvider> {
-    storage: &'a mut S,
+#[contract]
+pub struct TIP403Registry {
+    policy_id_counter: u64,
```
0xKitsune, line 16: Note that if a user inpsects this slot direclty rather than using getter functions, it will return `0` on default.

PR #806, opened 2025-11-10
feat!(precompile): gas refund

rakita:
Adds refund to the sstore function in precompiles. This would make a repeated sstore operations chepear to do and it alignes gas account to sstore on mainnet.

Pending on revm PR: https://github.com/bluealloy/revm/pull/3152

### DIFF
--- a/crates/precompiles/src/lib.rs
+++ b/crates/precompiles/src/lib.rs
@@ -285,7 +285,7 @@ fn fill_precompile_output(
     output.gas_used = storage.gas_used();
 
     // add refund only if it is not reverted
-    if !output.reverted && storage.spec().is_moderato() {
+    if !output.reverted && storage.spec().is_allegretto() {
         output.gas_refunded = storage.gas_refunded();
     }
     output


crates/precompiles/src/lib.rs
```diff
@@ -255,6 +255,21 @@ fn mutate_void<T: SolCall>(
...
+    mut output: PrecompileOutput,
+    storage: &mut impl PrecompileStorageProvider,
+) -> PrecompileOutput {
+    output.gas_used = storage.gas_used();
+
+    // add refund only if it is not reverted
+    // currently waiting for revm commit https://github.com/bluealloy/revm/pull/3152
+    // if !output.reverted {
+    //     result.gas_refunded = storage.gas_refunded();
+    // }
```
0xKitsune: It looks like this is merged, can we update this?

rakita: Need to wait for revm bump PR, but have merged with maid and added fork check.

crates/chainspec/src/hardfork.rs
```diff
@@ -33,6 +33,13 @@ hardfork!(
     }
 );
 
+impl TempoHardfork {
+    #[inline]
+    pub fn is_enabled_in_spec(&self, spec: Self) -> bool {
```
klkvr: this is a pretty confusing naming, I'd prefer to just have `.is_moderato`

crates/precompiles/src/lib.rs
```diff
@@ -259,6 +259,20 @@ fn mutate_void<T: SolCall>(
...
+#[inline]
+fn fill_precompile_output(
+    mut output: PrecompileOutput,
+    storage: &mut impl PrecompileStorageProvider,
+) -> PrecompileOutput {
+    output.gas_used = storage.gas_used();
+
+    // add refund only if it is not reverted
+    if !storage.spec().is_enabled_in_spec(TempoHardfork::Moderato) {
+        // output.gas_refunded = storage.gas_refunded();
```
rakita:
```suggestion
    if storage.spec().is_enabled_in_spec(TempoHardfork::Moderato) {
        output.gas_refunded = storage.gas_refunded();
```

crates/chainspec/src/hardfork.rs
rakita: `is_moderato_enabled` maybe? `is_moderato` does not convey if previous fork is enabled

rakita: I see there is already `is_moderato` fn, so will just use it

crates/precompiles/src/lib.rs
```diff
@@ -259,6 +259,21 @@ fn mutate_void<T: SolCall>(
...
+/// Fills the precompile output with the used and refunded gas.
+#[inline]
+fn fill_precompile_output(
+    mut output: PrecompileOutput,
+    storage: &mut impl PrecompileStorageProvider,
+) -> PrecompileOutput {
+    output.gas_used = storage.gas_used();
+
+    // add refund only if it is not reverted
+    if !output.reverted && storage.spec().is_moderato() {
```
rakita:
```suggestion
    if !output.reverted && storage.spec().is_allegretto() {
```

crates/precompiles/src/lib.rs
```diff
@@ -277,6 +277,20 @@ fn mutate_void<T: SolCall>(
...
 
+#[inline]
+fn fill_precompile_output(
+    mut output: PrecompileOutput,
+    storage: &mut impl PrecompileStorageProvider,
+) -> PrecompileOutput {
+    output.gas_used = storage.gas_used();
+
+    // add refund only if it is not reverted
+    if !output.reverted && storage.spec().is_moderato() {
```
0xKitsune: Can we update this to `allegretto`?

PR #824, opened 2025-11-10
refactor: unify codepaths for fee token operations

klkvr:
Right now fee token preferences logic is duplicated in mulitple places:
- EVM https://github.com/tempoxyz/tempo/blob/a7b95b622116ca51a640bd0c20c11e09b3813c01/crates/revm/src/handler.rs#L981
- Pool https://github.com/tempoxyz/tempo/blob/a7b95b622116ca51a640bd0c20c11e09b3813c01/crates/transaction-pool/src/validator.rs#L76-L86
- extension traits consumed by RPC https://github.com/tempoxyz/tempo/blob/a7b95b622116ca51a640bd0c20c11e09b3813c01/crates/precompiles/src/provider.rs#L75-L102

This PR unifies this logic under a single `tempo_revm::TempoStateAccess` traits that provides blanket implementations for `Journal`, `Database` and `StateProvider` implementations. Allowing to unify logic from extension traits with actual EVM codepaths

The new trait is using dummy generic to avoid conflicting blanket impls

### DIFF
--- a/crates/transaction-pool/src/validator.rs
+++ b/crates/transaction-pool/src/validator.rs
@@ -9,11 +9,7 @@ use reth_transaction_pool::{
     EthTransactionValidator, PoolTransaction, TransactionOrigin, TransactionValidationOutcome,
     TransactionValidator, error::InvalidPoolTransactionError,
 };
-use tempo_precompiles::{
-    TIP_FEE_MANAGER_ADDRESS,
-    tip_fee_manager::IFeeManager::setUserTokenCall,
-};
-use tempo_primitives::TempoTxEnvelope;
+use tempo_precompiles::DEFAULT_FEE_TOKEN;
 use tempo_revm::TempoStateAccess;
 
 /// Validator for Tempo transactions.
--- a/crates/revm/src/handler.rs
+++ b/crates/revm/src/handler.rs
@@ -1059,7 +1059,7 @@ mod tests {
     #[test]
     fn test_get_fee_token() -> eyre::Result<()> {
         let journal = create_test_journal();
-        let mut ctx = Context::mainnet()
+        let mut ctx: TempoContext<_> = Context::mainnet()
             .with_db(CacheDB::new(EmptyDB::default()))
             .with_block(TempoBlockEnv::default())
             .with_cfg(Default::default())


crates/revm/src/common.rs
mattsse, line 1: this makes a lot of sense

crates/transaction-pool/src/validator.rs
```diff
@@ -291,29 +266,6 @@ mod tests {
...
-            U256::from_be_bytes(fee_token.into_word().0),
-        )];
-        let fee_manager_acct = ExtendedAccount::new(0, U256::ZERO).extend_storage(storage);
-        provider.add_account(TIP_FEE_MANAGER_ADDRESS, fee_manager_acct);
-
-        let outcome = validator
-            .validate_transaction(TransactionOrigin::External, transaction)
-            .await;
-
-        assert!(outcome.is_valid());
```
klkvr, line 316: this is no longer passing because the fee token is not properly initialized and triggeres `InvalidFeeToken` error

the positive path is covered via other tests sending transactions

crates/node/src/rpc/mod.rs
```diff
@@ -249,14 +249,14 @@ impl<N: FullNodeTypes<Types = TempoNode>> Call for TempoEthApi<N> {
         evm_env: &EvmEnvFor<Self::Evm>,
         tx_env: &TxEnvFor<Self::Evm>,
     ) -> Result<u64, Self::Error> {
+        let fee_payer = tx_env
+            .fee_payer()
+            .map_err(EVMError::<ProviderError, _>::from)?;
```
mattsse, line 254: is this the same as `map_err(ProviderError::from)?` ?

crates/revm/src/common.rs
```diff
@@ -0,0 +1,224 @@
...
+            Either::Right(core::iter::once((self.inner.kind, &self.inner.data)))
+        }
+    }
+
+    fn caller(&self) -> Address {
+        self.inner.caller
+    }
+}
+
+impl TempoTx for Recovered<TempoTxEnvelope> {
```
mattsse, line 60: can be
```suggestion
impl<T: TempoTx> TempoTx for Recovered<T> {
```
?

crates/node/src/rpc/mod.rs
klkvr: not exactly because `fee_payer` returns `TempoInvalidTransaction`

crates/revm/src/common.rs
klkvr: we can't implement `TempoTx` for `TempoTxEnvelope` because of the `caller` fn

mattsse: ah I see

PR #833, opened 2025-11-10
feat: preparation for next hardfork

klkvr:
Introduces new `Andantino` hardfork variant (are we ok with naming collision?)

Propagates hardfork generic to EVM

### DIFF
--- a/crates/revm/src/evm.rs
+++ b/crates/revm/src/evm.rs
@@ -224,11 +224,10 @@ mod tests {
             nonce: 0,
             ..Default::default()
         };
-        let res = tempo_evm.transact_raw(tx_env.into())?;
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/lib.rs
```diff
@@ -113,100 +119,102 @@ macro_rules! tempo_precompile {
 
 pub struct TipFeeManagerPrecompile;
 impl TipFeeManagerPrecompile {
-    pub fn create(chain_id: u64) -> DynPrecompile {
+    pub fn create(chain_id: u64, spec: TempoHardfork) -> DynPrecompile {
```
mattsse, line 122: ah cool, with this we get direct access to the current active hardfork

PR #834, opened 2025-11-10
fix: validate `shared_gas_limit`

klkvr:
Ensures `shared_gas_limit` is correctly set

### DIFF
--- a/crates/consensus/src/lib.rs
+++ b/crates/consensus/src/lib.rs
@@ -137,11 +137,15 @@ impl HeaderValidator<TempoHeader> for TempoConsensus {
     fn validate_header(&self, header: &SealedHeader<TempoHeader>) -> Result<(), ConsensusError> {
         self.eth_beacon_validate_header(header)?;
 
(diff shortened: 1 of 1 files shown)


crates/consensus/src/lib.rs
```diff
@@ -137,11 +137,15 @@ impl HeaderValidator<TempoHeader> for TempoConsensus {
...
+        if header.shared_gas_limit != header.gas_limit() / TEMPO_SHARED_GAS_DIVISOR {
+            return Err(ConsensusError::Other(
+                "Shared gas limit does not match header gas limit".to_string(),
+            ));
+        }
 
         // Validate the non-payment gas limit
         if header.general_gas_limit
-            != (header.gas_limit() - shared_gas_limit) / TEMPO_GENERAL_GAS_DIVISOR
+            != (header.gas_limit() - header.shared_gas_limit) / TEMPO_GENERAL_GAS_DIVISOR
```
mattsse, line 148: this can't overflow because we ensured above that this is < gaslimit, so we dont need saturating/checked here

PR #836, opened 2025-11-10
fix: validate system transactions fields

klkvr:
Enforces that system transactions have zero values for `gas_limit`, `gas_price`, `value` and `nonce` fields, and correct `chain_id` set.

This shouldn't really be important and can be set to any values by block builder but enforcing it doesn't hurt

### DIFF
--- a/crates/consensus/src/lib.rs
+++ b/crates/consensus/src/lib.rs
@@ -207,12 +207,7 @@ impl Consensus<Block> for TempoConsensus {
         let transactions = &block.body().transactions;
 
         if let Some(tx) = transactions.iter().find(|&tx| {
-            tx.is_system_tx()
-                && (tx.max_fee_per_gas() != 0
-                    || tx.gas_limit() != 0
-                    || !tx.value().is_zero()
-                    || tx.chain_id() != Some(self.inner.chain_spec().chain().id())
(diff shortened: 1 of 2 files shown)


crates/consensus/src/lib.rs
```diff
@@ -206,6 +206,20 @@ impl Consensus<Block> for TempoConsensus {
     fn validate_block_pre_execution(&self, block: &SealedBlock<Block>) -> Result<(), Self::Error> {
         let transactions = &block.body().transactions;
 
+        if let Some(tx) = transactions.iter().find(|&tx| {
+            tx.is_system_tx()
```
mattsse: this should be just one u256 cmp for non system txs so relative inexpensive

crates/consensus/src/lib.rs
```diff
@@ -206,6 +206,20 @@ impl Consensus<Block> for TempoConsensus {
...
+                && (tx.max_fee_per_gas() != 0
+                    || tx.gas_limit() != 0
+                    || !tx.value().is_zero()
+                    || tx.chain_id() != Some(self.inner.chain_spec().chain().id())
+                    || tx.nonce() != 0)
```
mattsse: these make sense, but should we make this a dedicated function instead?

PR #844, opened 2025-11-11
fix!(stablecoin-exchange): validate base token is TIP20 in createPair

fgimenez:

```
https://github.com/tempoxyz/docs/blob/a4a5b5bf9c0b296b9d7b54b12ec2688bf1930c9b/specs/src/StablecoinExchange.sol#L155-L163
https://github.com/tempoxyz/tempo/blob/13a3f1f6826b6490caa7099ea06450a4d9997541/crates/precompiles/src/stablecoin_exchange/mod.rs#L337-L342
this should validate that the base token is a tip20
```

breaking, changes error message and gas consumption in receipt

### DIFF
--- a/crates/precompiles/src/stablecoin_exchange/mod.rs
+++ b/crates/precompiles/src/stablecoin_exchange/mod.rs
@@ -20,7 +20,7 @@ use crate::{
     linking_usd::LinkingUSD,
     stablecoin_exchange::orderbook::compute_book_key,
     storage::{PrecompileStorageProvider, Slot, Storable, VecSlotExt},
-    tip20::{ITIP20, TIP20Error, TIP20Token, is_tip20, validate_usd_currency},
+    tip20::{ITIP20, TIP20Token, is_tip20, validate_usd_currency},
 };
 use alloy::primitives::{Address, B256, Bytes, IntoLogData, U256};
 use revm::state::Bytecode;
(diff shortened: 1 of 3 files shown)


crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -335,6 +336,11 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
     }
 
     pub fn create_pair(&mut self, base: Address) -> Result<B256> {
+        // Validate that base is a TIP20 token (only after Moderato hardfork)
+        if self.storage.spec() >= TempoHardfork::Moderato && !is_tip20(base) {
+            return Err(TIP20Error::invalid_quote_token().into());
```
0xKitsune: Should this error reflect that the base token is invalid rather than the quote token? Can we also make sure the spec reference is using custom errors?

fgimenez: makes sense, done here https://github.com/tempoxyz/tempo/pull/844/commits/52ff231ef33c8067e4658c555bc9ec81d10e9354 also proposed changes to the docs PR to introduce the custom error too https://github.com/tempoxyz/docs/pull/180

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -335,6 +336,11 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
     }
 
     pub fn create_pair(&mut self, base: Address) -> Result<B256> {
+        // Validate that base is a TIP20 token (only after Moderato hardfork)
+        if self.storage.spec() >= TempoHardfork::Moderato && !is_tip20(base) {
+            return Err(TIP20Error::invalid_base_token().into());
```
0xKitsune: I think this should be a `StablecoinExchangeError` since `invalid_base_token()` only relates to the Stablecoin Exchange.

Opened a PR for this here:
https://github.com/tempoxyz/tempo/pull/885

PR #845, opened 2025-11-11
fix!(stablecoin-exchange): add missing book existence check to place_flip

fgimenez:

```
https://github.com/tempoxyz/tempo/blob/13a3f1f6826b6490caa7099ea06450a4d9997541/crates/precompiles/src/stablecoin_exchange/mod.rs#L468-L470
missing check that book exists, also why is place_flip implemented separately from place?
```
addressing here the missing check.

breaking, without the fix a place tx with unexisting book would actually succeed

### DIFF
--- a/crates/precompiles/src/stablecoin_exchange/mod.rs
+++ b/crates/precompiles/src/stablecoin_exchange/mod.rs
@@ -392,12 +392,9 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
         // Compute book_key from token pair
         let book_key = compute_book_key(token, quote_token);
 
-        // Check book existence (only after Moderato hardfork)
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -467,6 +468,11 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
 
         // Compute book_key from token pair
         let book_key = compute_book_key(token, quote_token);
+        let book = self.sload_books(book_key)?;
```
0xKitsune: Will this cause gas to differ? Can we sload the book inside the conditional checking if spec >= hardfork?

fgimenez: good catch, done ptal

PR #851, opened 2025-11-11
feat(arbbot): add simple metrics

Zygimantass:


bin/tempo-sidecar/src/cmd/simple_arb.rs
```diff
@@ -29,6 +34,10 @@ pub struct SimpleArbArgs {
     /// Interval between checking pools for rebalancing. This should be set to the block time.
     #[arg(long, default_value_t = 2)]
     poll_interval: u64,
+
+    /// Number of failed transactions executed by the arb bot
```
mattsse: should be 
```suggestion
    /// Prometheus port for metrics
```

or similar

PR #855, opened 2025-11-11
fix!(precompiles): add reward handling to TIP20 fee transfer functions

fgimenez:

```
https://github.com/tempoxyz/docs/blob/344d08b043ff2da3720302bceb550ca9f47179db/specs/src/TIP20.sol#L388-L430
https://github.com/tempoxyz/tempo/blob/13a3f1f6826b6490caa7099ea06450a4d9997541/crates/precompiles/src/tip20/mod.rs#L792-L866
the reward handling logic disappeared from the precompiles
```

breaking, state changes in balances, `opted_in_supply`, global reward state and user rewards

### DIFF
--- a/crates/precompiles/src/tip20/mod.rs
+++ b/crates/precompiles/src/tip20/mod.rs
@@ -23,6 +23,7 @@ use alloy::{
 };
 use revm::state::Bytecode;
 use std::sync::LazyLock;
+use tempo_chainspec::hardfork::TempoHardfork;
 use tempo_precompiles_macros::contract;
 use tracing::trace;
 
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/tip20/mod.rs
```diff
@@ -835,6 +858,25 @@ impl<'a, S: PrecompileStorageProvider> TIP20Token<'a, S> {
             return Ok(());
         }
 
+        // Handle rewards (only after Moderato hardfork)
+        if self.storage.spec() >= TempoHardfork::Moderato {
```
0xKitsune, line 862: Very minor, but for the future we could consider adding a helper function for evaluating hardforks. For example `if self.spec().is_moderato()`.

PR #863, opened 2025-11-11
fix!: correct off-by-one in TIP20Factory quote token validation, gate behind Moderato

fgimenez:

```
no changes needed in tempo, we already have https://github.com/tempoxyz/tempo/blob/main/crates/precompiles/src/tip20_factory/mod.rs#L62
this conditional is off-by-one, token_id is the id of the to be deployed token, so it should be >=
```

### DIFF
--- a/crates/precompiles/src/tip20_factory/mod.rs
+++ b/crates/precompiles/src/tip20_factory/mod.rs
@@ -57,14 +57,23 @@ impl<'a, S: PrecompileStorageProvider> TIP20Factory<'a, S> {
 
         trace!(%sender, %token_id, ?call, "Create token");
 
-        // Ensure that the quote token is a valid TIP20 that is currently deployed (only after Moderato hardfork).
-        // Note that the token Id increments on each deployment which ensures that the quote
-        // token id must always be < the current token_id (strictly less than, not less than or equal).
-        if self.storage.spec() >= TempoHardfork::Moderato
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/tip20_factory/mod.rs
```diff
@@ -56,10 +57,13 @@ impl<'a, S: PrecompileStorageProvider> TIP20Factory<'a, S> {
...
-        // Ensure that the quote token is a valid TIP20 that is currently deployed.
+        // Ensure that the quote token is a valid TIP20 that is currently deployed (only after Moderato hardfork).
         // Note that the token Id increments on each deployment which ensures that the quote
-        // token id must always be <= the current token_id
-        if !is_tip20(call.quoteToken) || address_to_token_id_unchecked(call.quoteToken) > token_id {
+        // token id must always be < the current token_id (strictly less than, not less than or equal).
+        if self.storage.spec() >= TempoHardfork::Moderato
+            && (!is_tip20(call.quoteToken)
+                || address_to_token_id_unchecked(call.quoteToken) >= token_id)
+        {
```
0xKitsune: Shouldnt we still be checking for the previous condition if `spec() < TempoHardfork::Moderato`?

fgimenez: sure, fixed ptal

Issue #857, opened 2025-11-11
chore(precompiles): drop `SLOT_COUNT` and `BYTE_COUNT` in favor of `LAYOUT`


0xrusowsky:
## Motivation

introduce a new `enum Layout` to replace `BYTE_COUNT` and `SLOT_COUNT` in all `Storable` and `StorableType` types

```rs
pub enum Layout {
    /// Single slot primitives. Can be packed with other fields if N < 32.
    Bytes(usize),

    /// Occupies N full slots (each 32 bytes). Cannot be packed.
    ///
    /// Used for:
    /// - Multi-slot types (structs, arrays)
    /// - Dynamic types (String, Bytes, Vec) that store their base slot
    Slots(usize),
}
```

Issue #856, opened 2025-11-11
feat(precompiles): simplify storage primitives and macros


0xrusowsky:
## Motivation

now that the storage primitives were introduced and merged, we need to do a follow-up effort to refactor a little bit and improve maintainability.

currently, the contract macro has intricate code to conditionally handle packing manually which makes it hard to audit and maintain.

additionally, in order to quickly add support for new storage access patterns we had to add a set of methods with repeated/similar logic `read_at_offset()`, `read_at_offset_packed()`, `write_at_offset()`, `write_at_offset_packed()` used for array/vec indexing and single field access on structs

## Solution

by implementing all the sub-issues, we should address the presented pain points

PR #864, opened 2025-11-11
chore(precompiles): use const LAYOUT

0xrusowsky:
towards https://github.com/tempoxyz/tempo/issues/856

closes https://github.com/tempoxyz/tempo/issues/857

### DIFF
--- a/crates/precompiles-macros/src/layout.rs
+++ b/crates/precompiles-macros/src/layout.rs
@@ -363,7 +363,7 @@ pub(crate) fn gen_getters_and_setters(
                 impl<'a, S: crate::storage::PrecompileStorageProvider> #struct_name<'a, S> {
                     #[inline]
                     fn #getter_name(&mut self) -> crate::error::Result<#ty> {
-                        <#ty as crate::storage::Storable<{ <#ty as crate::storage::StorableType>::LAYOUT.slots() }>>::load(
+                        <#ty as crate::storage::Storable<{ <#ty as crate::storage::StorableType>::SLOTS }>>::load(
                             self,
                             <#slot_id as crate::storage::SlotId>::SLOT,
                         )
@@ -371,7 +371,7 @@ pub(crate) fn gen_getters_and_setters(
 
                     #[inline]
                     fn #cleaner_name(&mut self) -> crate::error::Result<()> {
-                        <#ty as crate::storage::Storable<{ <#ty as crate::storage::StorableType>::LAYOUT.slots() }>>::delete(
+                        <#ty as crate::storage::Storable<{ <#ty as crate::storage::StorableType>::SLOTS }>>::delete(
                             self,
                             <#slot_id as crate::storage::SlotId>::SLOT,
                         )
@@ -489,7 +489,7 @@ fn gen_packing_constants_for_slots_module(
         let byte_count_expr = match &allocated.kind {
             FieldKind::Mapping { .. } | FieldKind::NestedMapping { .. } => quote! { 32 },
             _ if is_dynamic_type(field_ty) => quote! { 32 },
-            _ => quote! { <#field_ty as crate::storage::StorableType>::LAYOUT.bytes() },
+            _ => quote! { <#field_ty as crate::storage::StorableType>::BYTES },
         };
 
         constants.extend(quote! {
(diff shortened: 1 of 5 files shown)


crates/precompiles/src/storage/types/mod.rs
```diff
@@ -119,7 +175,12 @@ pub trait Storable<const N: usize>: Sized + StorableType {
     /// When multiple small fields are packed into a single slot, they are
     /// extracted from the appropriate word using bit shifts and masks.
     /// The derive macro handles this automatically.
-    fn from_evm_words(words: [U256; N]) -> Result<Self>;
+    fn from_evm_words(words: [U256; SLOTS]) -> Result<Self>;
+
+    /// Test helper to ensure `LAYOUT` and `SLOTS` are in sync.
+    fn validate_layout() {
+        debug_assert_eq!(<Self as StorableType>::LAYOUT.slots(), SLOTS)
+    }
```
0xrusowsky, line 192: i tried adding the `#[cfg(test)]` flag, but then the fn wouldn't be available in `tests/storage/tests/packing.rs`

i think fine, as the optimizer should get rid of it in the release build

crates/precompiles/src/storage/types/mod.rs
```diff
@@ -11,18 +11,79 @@ pub mod vec;
...
+}
+
+impl Layout {
+    /// Returns true if this field can be packed with adjacent fields.
+    ///
+    /// Only `Bytes` variants with size < 32 can be packed.
+    /// Full-slot types (`Bytes(32)` and all `Slots` variants) cannot be packed.
+    pub const fn is_packable(&self) -> bool {
+        match self {
+            Self::Bytes(n) => *n < 32,
```
klkvr, line 42: should this be an invariant? why would you use `Bytes` with `n > 32`?

crates/precompiles-macros/src/layout.rs
```diff
@@ -363,15 +363,15 @@ pub(crate) fn gen_getters_and_setters(
                 impl<'a, S: crate::storage::PrecompileStorageProvider> #struct_name<'a, S> {
                     #[inline]
                     fn #getter_name(&mut self) -> crate::error::Result<#ty> {
-                        <#ty as crate::storage::Storable<{ <#ty>::SLOT_COUNT }>>::load(
+                        <#ty as crate::storage::Storable<{ <#ty as crate::storage::StorableType>::LAYOUT.slots() }>>::load(
```
klkvr: should we add a helper getter on `StorableType`?

crates/precompiles/src/storage/types/mod.rs
```diff
@@ -11,18 +11,79 @@ pub mod vec;
...
+        match self {
+            Self::Bytes(n) => *n,
+            Self::Slots(n) => {
+                // Compute n * 32 using repeated addition for const compatibility
+                let (mut i, mut result) = (0, 0);
+                while i < *n {
+                    result += 32;
+                    i += 1;
+                }
+                result
```
mattsse, line 69: ah I see const mul isnt stable yet

crates/precompiles/src/storage/types/mod.rs
0xrusowsky: it skips `U256` which i labeled as `Bytes(32)` rather than `Slot(1)`.
maybe i was trying to be too purist, as in `Bytes` == primitives, but since it's a const fn i thought it didn't matter

crates/precompiles/src/storage/types/mod.rs
0xrusowsky: yes, the compiler wasn't happy about it 😞

crates/precompiles/src/storage/types/mod.rs
klkvr: ah I see, `U256` is an edge case

imo it's fine to return `is_packable` true for it? we will just never be able to actually pack it

but I guess also doesn't matter that much

crates/precompiles-macros/src/layout.rs
0xrusowsky: addressed in [`01c75bb` (#864)](https://github.com/tempoxyz/tempo/pull/864/commits/01c75bb93efbdada753af355607c25a346c291dc)

crates/precompiles/src/storage/types/mod.rs
0xrusowsky: it was just to help the compiler optimize haha i guess it's kinda the same in the end

Issue #858, opened 2025-11-11
chore(precompiles): update `trait Storable` to use `LayoutCtx`


0xrusowsky:
introduce `enum LayoutCtx` and modify `trait Storable` to achieve a unified API that allows us to totally integrate the storage interaction logic inside the trait methods
```rs
pub enum LayoutCtx {
  /// Load/store entire value at base_slot (no SLOAD for writes)
  Full,

  /// Load/store packed primitive at byte offset (requires SLOAD for writes)
  Packed(usize)
}
```
note:
- all `Storable` types can be loaded with `LayoutCtx::Full`.
- `LayoutCtx::Packed` is only for primitives
- `LayoutCtx::Index` is only for arrays/vecs

```rs
trait Storable<const N: usize>: Sized + StorableType {
  fn load<S: StorageOps>(storage: &mut S, base_slot: U256, ctx: LayoutCtx);
  fn store<S: StorageOps>(&self, storage: &mut S, base_slot: U256, ctx: LayoutCtx);
  fn delete<S: StorageOps>(storage: &mut S, base_slot: U256, ctx: LayoutCtx);
```

PR #872, opened 2025-11-12
chore(precompiles): update `trait Storable` to use `LayoutCtx`

0xrusowsky:
towards https://github.com/tempoxyz/tempo/issues/856

follows https://github.com/tempoxyz/tempo/issues/857

closes https://github.com/tempoxyz/tempo/issues/858

### DIFF
--- a/crates/precompiles/tests/storage_tests/solidity/primitives.rs
+++ b/crates/precompiles/tests/storage_tests/solidity/primitives.rs
@@ -54,9 +54,17 @@ fn test_arrays_layout() {
         field_a: U256,
         large_array: [U256; 5],
         field_b: U256,
+        nested_array: [[u8; 4]; 8],
+        another_nested_array: [[u16; 2]; 6],
     }
 
-    let rust_layout = layout_fields!(field_a, large_array, field_b);
+    let rust_layout = layout_fields!(
+        field_a,
+        large_array,
+        field_b,
+        nested_array,
+        another_nested_array
+    );
 
     // Compare against expected layout from Solidity
     let solc_layout = load_solc_layout(&testdata("arrays.sol"));
--- a/crates/precompiles/tests/storage_tests/solidity/testdata/arrays.layout.json
+++ b/crates/precompiles/tests/storage_tests/solidity/testdata/arrays.layout.json
@@ -26,19 +26,69 @@
             "offset": 0,
             "slot": "6",
             "type": "t_uint256"
+          },
+          {
+            "astId": 16,
+            "contract": "tests/storage_tests/solidity/testdata/arrays.sol:Arrays",
+            "label": "nestedArray",
+            "offset": 0,
+            "slot": "7",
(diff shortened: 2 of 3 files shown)


crates/precompiles-macros/src/storable_primitives.rs
```diff
@@ -69,21 +69,42 @@ fn gen_storage_key_impl(type_path: &TokenStream, strategy: &StorageKeyStrategy)
...
+                        match ctx {
+                            LayoutCtx::Full => {
+                                storage.sstore(base_slot, U256::from(*self))?;
+                                Ok(())
+                            }
+                            LayoutCtx::Packed(offset) => {
+                                let current = storage.sload(base_slot)?;
+                                let value = U256::from(*self);
+                                let updated = crate::storage::packing::insert_packed_value(current, &value, offset, #byte_count)?;
+                                storage.sstore(base_slot, updated)?;
```
klkvr, line 104: wondering if `storage.store_packed` could be useful

crates/precompiles-macros/src/storable_primitives.rs
```diff
@@ -542,18 +625,7 @@ fn gen_unpacked_array_store() -> TokenStream {
...
-    }
-}
-
-/// Generate delete implementation for unpacked arrays
-fn gen_unpacked_array_delete(array_size: &usize) -> TokenStream {
-    quote! {
-        // For unpacked single-slot elements, just zero out each slot
-        for i in 0..#array_size {
-            storage.sstore(base_slot + U256::from(i), U256::ZERO)?;
+            elem.store(storage, elem_slot, LayoutCtx::Full)?;
```
klkvr, line 628: how does this work with arrays of arrays? or do we not support that?

crates/precompiles-macros/src/storable_primitives.rs
0xrusowsky: i think this is fine as it is, because:
- it is macro code (users never see it)
- the complexity is abstracted away and isolated under the `packing` module

crates/precompiles-macros/src/storable_primitives.rs
0xrusowsky: they are supported, yes.
the macro treats inner arrays as if they're 32-byte types, which forces
unpacked layout for the outer array (i.e. `[[u8; 4]; 8]` becomes 8 slots despite it could fit in 1 single slot)

0xrusowsky: tbh idk if solidity is smarter about it (i think it does the same), and i've just realized that we don't have solidity compatibility tests for nested arrays.

i'll add tests and figure out if we are solidity-compatible or if we need to adjust the behavior

0xrusowsky: yes, we are solidity compatible, see: [`7cdbbef` (#872)](https://github.com/tempoxyz/tempo/pull/872/commits/7cdbbefa85405e24db132a12ae8d015c157940c2)

Issue #875, opened 2025-11-12
chore(deps): bump revm to 33.0.0


zerosnacks:
### Describe the feature

Reth has been updated: https://github.com/paradigmxyz/reth/blob/3f1a7b37e91c0e6509a3145d9e4ac69302cf2d12/Cargo.toml#L469

`tempo-foundry` requires it to merge from upstream: https://github.com/foundry-rs/foundry/pull/12538

It may require applying this specific patch: https://github.com/bluealloy/revm/pull/3164

Revm 33.0.0 may have some breaking changes

PR #878, opened 2025-11-12
chore(deps): bump to revm 33, latest Reth + fix zepter lint

zerosnacks:
Closes: https://github.com/tempoxyz/tempo/issues/875

Bumps Reth to latest commit: https://github.com/paradigmxyz/reth/commit/95b8a8535b9bf0bad3222aaa6e24aa1056c6bda9, please change if you have a policy around this!

Bumps to Revm: 33.0.0
Bumps to Alloy-Evm: 0.24.1

Note: I've ran `cargo update` to fix the lockfile

I've used `.into()` to convert `String` into `Cow<'_, str>` assuming this will be optimized at compile time

No logic changes but please carefully review

Some internals likely have to be updated for the emitting of the events, this has not yet been included as I don't know the details. Currently marked as `draft` because of this. Feel free to take over the PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7398,7 +7398,7 @@ checksum = "6b3789b30bd25ba102de4beabd95d21ac45b69b1be7d14522bab988c526d6799"
 [[package]]
 name = "reth-basic-payload-builder"
 version = "1.9.2"
-source = "git+https://github.com/paradigmxyz/reth?rev=95b8a85#95b8a8535b9bf0bad3222aaa6e24aa1056c6bda9"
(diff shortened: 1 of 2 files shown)


crates/precompiles/src/nonce/dispatch.rs
```diff
@@ -15,7 +15,7 @@ impl<S: PrecompileStorageProvider> Precompile for NonceManager<'_, S> {
         let selector: [u8; 4] = calldata
             .get(..4)
             .ok_or_else(|| {
-                PrecompileError::Other("Invalid input: missing function selector".to_string())
+                PrecompileError::Other("Invalid input: missing function selector".into())
```
rakita, line 18: `into()` is good, it will use the `&'static str` of Cow and it would not need to allocate the string. It is more performant that way.

PR #890, opened 2025-11-13
fix!(stablecoin-exchange): correct unit conversions and avoid rounding in swap order fills

fgimenez:
from https://tempoxyz.slack.com/archives/C09KCGR4LQ4/p1763009606739239?thread_ts=1762895220.576359&cid=C09KCGR4LQ4

```
if remainingOut is 1:
if price < PRICE_SCALE - baseNeeded and fillAmount is 1 and (fillAmount * price) / PRICE_SCALE is 0
if price > PRICE_SCALE - baseNeeded and fillAmount is 0 and (fillAmount * price) / PRICE_SCALE is also 0
only way it works is if price == PRICE_SCALE, it seems
yikes
Doing the conversion twice is bad, we are basically doing:
filledAmount = ((remainingOut * PRICE_SCALE) / price) * price / PRICE_SCALE;
if currentOrder.remaining > base we should be able to just zero remainingOut without the div, should fix the problem then 
PR with fix in solidity: https://github.com/tempoxyz/docs/pull/221
```

### DIFF
--- a/crates/precompiles/src/stablecoin_exchange/mod.rs
+++ b/crates/precompiles/src/stablecoin_exchange/mod.rs
@@ -24,7 +24,6 @@ use crate::{
 };
 use alloy::primitives::{Address, B256, Bytes, IntoLogData, U256};
 use revm::state::Bytecode;
-use tempo_chainspec::hardfork::TempoHardfork;
 use tempo_precompiles_macros::contract;
 
 /// Minimum order size of $10 USD
@@ -259,7 +258,7 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
         let mut amount = amount_in;
         for (book_key, base_for_quote) in route {
             // Fill orders for this hop - no min check on intermediate hops
-            amount = if self.storage.spec() >= TempoHardfork::Moderato {
+            amount = if self.storage.spec().is_moderato() {
                 self.fill_orders_exact_in_post_moderato(book_key, base_for_quote, amount)?
             } else {
                 self.fill_orders_exact_in_pre_moderato(book_key, base_for_quote, amount, 0)?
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -781,16 +781,40 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
...
-            let amount_in = if bid {
-                fill_amount
+
+            let (fill_amount, amount_in) = if is_moderato {
+                // Post-Moderato: correct unit conversion
+                if bid {
+                    let base_needed = amount_out
+                        .checked_mul(orderbook::PRICE_SCALE as u128)
+                        .and_then(|v| v.checked_div(price as u128))
+                        .expect("Base needed calculation overflow");
```
mattsse: hmm, not clear to me why this is safe, because isnt amount_out unchecked input?

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -843,23 +892,74 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
         let mut order = self.sload_orders(level.head)?;
 
         let mut total_amount_out: u128 = 0;
+        let is_moderato = self.storage.spec() >= TempoHardfork::Moderato;
+
         while amount_in > 0 {
-            if amount_in < order.remaining() {
-                let amount_out = self.partial_fill_order(&mut order, &mut level, amount_in)?;
+            let price = tick_to_price(order.tick());
```
mattsse, line 939: isn't this constant for the entire duration of the loop?

fgimenez: this is indeed recalculated on every iteration, and it should only be required to change when we move to a new order. however this is how the code worked before the changes in this PR to fix the conversions, i can address this optimization separately in a followup, wdyt?

crates/precompiles/src/stablecoin_exchange/mod.rs
fgimenez: absolutely, fixed

crates/precompiles/src/stablecoin_exchange/mod.rs
0xKitsune: This has the potential to change if the order liquidity is exhausted and the next best order is at a new tick.

```rust
if let Some((new_level, new_order)) = next_order_info {
    level = new_level;
    order = new_order;

// --snip--
```

PR #891, opened 2025-11-13
feat(precompiles): simplify macros

0xrusowsky:
## Motivation

towards #856 

builds on top of #864 and #872 to simplify the macros leveraging the new `const: LAYOUT` and the enhanced `trait Storable` with `LayoutCtx` support. These 2 changes allows us to outsource most of the packing logic to the types that `impl Storage` and the compiler (thanks to compile-time const evaluation)

## Solution

refactors the `contract` and `Storable` macros to uses a common `packing` module that handles the whole slot allocation and constant generation by leveraging an IR `struct LayoutField` that captures all important info.

additionally it enhances the `contract`'s built-in constructor method to perform slot collision checks (optimized away in released builds).


crates/precompiles-macros/src/packing.rs
```diff
@@ -0,0 +1,550 @@
...
+        } else {
+            // Auto-assignment with packing support
+            let is_primitive = kind.is_direct();
+
+            // Non-primitives always start a new slot
+            let base_slot = if index == 0 || !is_primitive {
+                let slot = last_auto_slot;
+                last_auto_slot += U256::ONE;
+                slot
+            }
```
klkvr: can we avoid doing this entirely? i.e just store "group" of fields with base slots

e.g fields 0, 1, 2 go into a group with `base_slot` 0, and determine the concrete slots at compile-time based on layouts of each other

crates/precompiles-macros/src/packing.rs
```diff
@@ -0,0 +1,550 @@
...
+            SlotAssignment::Manual(manual_slot) => {
+                let slot_expr = match slot_type {
+                    SlotType::Usize => {
+                        let value = manual_slot.to::<usize>();
+                        quote! { usize = #value }
+                    }
+                    SlotType::U256 => {
+                        let hex_value = format!("{manual_slot}_U256");
+                        let slot_lit = syn::LitInt::new(&hex_value, proc_macro2::Span::call_site());
+                        quote! { ::alloy::primitives::U256 = ::alloy::primitives::uint!(#slot_lit) }
```
klkvr: what's the motivation for usize/U256 differences here? can we unify?

crates/precompiles-macros/src/packing.rs
```diff
@@ -0,0 +1,550 @@
...
+            SlotType::Usize => quote! { #prev_slot_expr + 1 },
+            SlotType::U256 => {
+                quote! { #prev_slot_expr.checked_add(::alloy::primitives::U256::ONE).expect("slot overflow") }
+            }
+        };
+        return (slot_expr, quote! { 0 });
+    }
+
+    // If current field is a mapping, it must start on a new slot
+    if is_curr_mapping {
```
klkvr, line 325: ah so mappings are not actual types so we can't just use `prev_ty::LAYOUT` here? that's unfortunate, curious if we could unify this eventually

crates/precompiles-macros/src/packing.rs
```diff
@@ -0,0 +1,550 @@
...
+    }
+
+    // Standard packing logic for non-mapping fields
+    let can_pack_expr = quote! {
+        const PREV_LAYOUT: crate::storage::Layout = <#prev_ty as crate::storage::StorableType>::LAYOUT;
+        const CURR_LAYOUT: crate::storage::Layout = <#curr_ty as crate::storage::StorableType>::LAYOUT;
+
+        // Compute packing decision at compile-time
+        const PREV_END: usize = #prev_offset_expr + PREV_LAYOUT.bytes();
+        const CAN_PACK: bool = PREV_LAYOUT.is_packable() && CURR_LAYOUT.is_packable() && PREV_END + CURR_LAYOUT.bytes() <= 32;
```
klkvr: what's the motivation for generating some extra consts here vs a big expr? just curious

crates/precompiles-macros/src/packing.rs
0xrusowsky: yes, we can unify, i will have to change the struct helper fns, but probably better to unify

crates/precompiles-macros/src/packing.rs
0xrusowsky: yes they are, the problem is that mapping right now requires the `SlotId` . But is hidden from the user cause structs drop all fields and use accessors (see how in the precompiles we simply say `Mapping<K, V>` and we don't get LSP errors).

however, this means that i cannot evaluate the mapping types in the macro cause i get compiler errrors because of the to missing `SlotId`.

i left a todo somewhere in the macros where i explain this, and mention that once we drop `SlotId` we can simplify this further

crates/precompiles-macros/src/packing.rs
0xrusowsky: code readibility, but let's minimize consts

crates/precompiles-macros/src/packing.rs
0xrusowsky: the "primitive" wording is stale, as this already handles multi-slot structs and dynamic types in the other arm.

however, u are right that this shouldn't be needed at this point, as we already do some manual codegen to ensure mappings eval to 32 bytes, so i should be able to make it fully generic.

good catch!

crates/precompiles-macros/src/packing.rs
0xrusowsky: for vis, this is the cmnt i was referring to: https://github.com/tempoxyz/tempo/blob/4e4ed4955389895c45f8e2c53ffbfad0fbd76be3/crates/precompiles-macros/src/packing.rs#L339-L347

will add it everywhere that can be simplified though

crates/precompiles-macros/src/packing.rs
0xrusowsky: addressed in [`e790381` (#891)](https://github.com/tempoxyz/tempo/pull/891/commits/e790381e0ec88e5538bd887e9edcf97dbbc2f28e)

crates/precompiles-macros/src/packing.rs
0xrusowsky: addressed here: [`a1a9974` (#891)](https://github.com/tempoxyz/tempo/pull/891/commits/a1a99741414dee87eb47a6704360ac754692a69c)

crates/precompiles-macros/src/packing.rs
0xrusowsky: addressed in [`a1a9974` (#891)](https://github.com/tempoxyz/tempo/pull/891/commits/a1a99741414dee87eb47a6704360ac754692a69c)

crates/precompiles-macros/src/packing.rs
0xrusowsky: update: this is totally dropped in #921

crates/precompiles-macros/src/packing.rs
```diff
@@ -0,0 +1,475 @@
...
+            let base_slot = if index == 0 {
+                let slot = last_auto_slot;
+                last_auto_slot += U256::ONE;
+                slot
+            } else {
+                let prev: &LayoutField<'_> = &result[index - 1];
+
+                // If previous also was auto-assigned, reuse base slot (becomes packing candidate)
+                if let SlotAssignment::Auto { base_slot } = &prev.assigned_slot {
+                    *base_slot
```
klkvr: does this mean that field that had `base_slot` specified (and thus has `SlotAssignment::Manual`) will never be packed with another one?

crates/precompiles-macros/src/packing.rs
```diff
@@ -0,0 +1,475 @@
...
+                }
+                // Otherwise, start new slot
+                else {
+                    let slot = last_auto_slot;
+                    last_auto_slot += U256::ONE;
```
klkvr: `last_auto_slot` is kind of confusing here because it seems like it's pretty easy for it to become out of sync?

i.e if I have 3 auto fields in a row, the last field will be at slot 2, and `last_auto_slot` would still be 1

crates/precompiles-macros/src/packing.rs
```diff
@@ -0,0 +1,475 @@
...
+
+    // All non-mapping fields use the same accessor pattern
+    Ok(FieldKind::Slot(ty))
+}
+
+/// Generate byte count expression for a field type.
+///
+/// Returns a const expression that evaluates to the byte size of the type.
+/// For mapping types, returns `32` directly to avoid type resolution issues.
+pub(crate) fn gen_byte_count_expr(ty: &Type, is_mapping: bool) -> TokenStream {
```
klkvr, line 285: should this be a helper on `LayoutField`?

crates/precompiles-macros/src/packing.rs
0xrusowsky: yes, this was my assumption. should we try packing it?

klkvr: seems reasonable? we do try packing into slot 0 right? not that important but worth documenting imo

crates/precompiles-macros/src/packing.rs
0xrusowsky: mmm ok, what do u propose? renaming it? or using a different approach which is easier to reason about?

crates/precompiles-macros/src/packing.rs
0xrusowsky: actually i got rid of it, in #921

crates/precompiles-macros/src/packing.rs
0xrusowsky: i thought if we individually moved one slot it would had a good reason.

if we wanted to move all slots we would use `base_slot` they would pack, but it's true that the very first one wouldn't and maybe that's not desirable

crates/precompiles-macros/src/packing.rs
klkvr: do we actually need to increment it? iiuc if we have "groups" of fields with the same base slot we could just keep it same for different fields?

it's also be more like `current_base_slot` Ig

0xrusowsky: addressed in [`ca727d8` (#891)](https://github.com/tempoxyz/tempo/pull/891/commits/ca727d80f19b167b943ebd6a33844762150a10fd)

crates/precompiles-macros/src/packing.rs
0xrusowsky: addressed in [`ca727d8` (#891)](https://github.com/tempoxyz/tempo/pull/891/commits/ca727d80f19b167b943ebd6a33844762150a10fd)

basically `base_slot` becomes an auto assigned slot, and we simply bump its reference slot before hand

crates/precompiles/tests/storage_tests/solidity/precompiles.rs
```diff
@@ -224,3 +224,243 @@ fn test_tip20_layout() {
...
+/// This test is marked with #[ignore] so it doesn't run in CI.
+/// To run it manually:
+/// ```bash
+/// cargo test export_all_storage_constants -- --ignored --nocapture
+/// ```
+///
+/// The output will be written to `current_branch_constants.json` in the workspace root.
+#[test]
+#[ignore]
+fn export_all_storage_constants() {
```
0xKitsune: Is this test actually asserting anything? Seems useful to have a test run in CI that ensures the storage layout does not change.

crates/precompiles/tests/storage_tests/layouts.rs
```diff
@@ -347,3 +347,50 @@ proptest! {
         prop_assert_eq!(s.storage.sload(s.address, U256::from(201))?, val_e); // field_e
     }
 }
+
+#[test]
+#[should_panic(expected = "Storage slot collision")]
+fn test_collision_same_slot() {
+    // Two fields with identical slot assignments should panic in debug builds
```
0xKitsune, line 354: It seems like its a footgun if its possible to assign two variables the same storage slots like this. Does this only panic in debug builds or other build profiles as well?

scripts/storage/compare_constants.sh
```diff
@@ -0,0 +1,174 @@
...
+#   ./scripts/compare_constants.sh <comparison-branch>
+#
+# Example:
+#   ./scripts/compare_constants.sh feat/simplify-storage
+#
+# This script will:
+# 1. Export constants from the current branch
+# 2. Export constants from the comparison branch
+# 3. Save a diff file showing differences
+# 4. Display the diff to stdout
```
0xKitsune: Instead of comparing constants across branches, its probably better to keep a snapshot of the latest contract layouts and add a test that verifies the current layout matches that snapshot. When a layout change is intentional, you just update the snapshot in the same PR. If it isnt, the test will fail and catch it.

Is there a reason that this is a script instead?

crates/precompiles/tests/storage_tests/layouts.rs
0xrusowsky: only in debug build, so that they don't hurt performance in release mode.
do you think this is an issue?

note that only manually assigned slots can cause collisions (vs themselves, or vs a multi-slot auto-assigned type (i.e. struct or fixed array)

crates/precompiles/tests/storage_tests/solidity/precompiles.rs
0xrusowsky: i was manually comparing the storage layouts, and this seemed a convenient way to do it, but having it on CI is prob better

scripts/storage/compare_constants.sh
0xrusowsky: that sounds better yes.

the motivation behind this, is that i was investigating why the CI broke. Despite we already had checks in place to ensure that the storage layout matched the solidity one, i wanted to double-check vs the main branch.

it turns out the diff is not in the storage layout but in the access pattern (optimization of SLOAD/SSTORE calls). The initial version that i shipped was too naive on whether a field should use slot packing or not, and thus consumes more gas than needed.

@klkvr suggested we keep the exact same pattern for the time being, and eventually hardfork to make it more efficient

with all this in mind, do you think it is worth keeping the layout snapshot test idea? or is the check vs solidity enough?

0xrusowsky: addressed here, feel free to merge if it looks good to u
https://github.com/tempoxyz/tempo/pull/943

crates/precompiles/tests/storage_tests/solidity/precompiles.rs
0xrusowsky: addressed here, feel free to merge if it looks good to u
https://github.com/tempoxyz/tempo/pull/943

PR #892, opened 2025-11-13
fix: `FeeAMM` fixes

### LINKED ISSUE #901: Bug: FeeAMM should validate that an address is a valid TIP20
Currently, the FeeAMM does not validate that an address is a valid TIP20. This should be fixed and gated via the `Moderato` hardfork.

[spec reference](https://github.com/tempoxyz/docs/blob/a8fb7a9fbd76a3718006ffcff2fcda942abe8bfe/specs/src/FeeAMM.sol#L20-L24) · [impl link](https://github.com/tempoxyz/tempo/blob/3af8d19bee800b5f18b8980d1f331da8bea13c1c/crates/precompiles/src/tip20/mod.rs#L100-L110)

### LINKED ISSUE #902: Bug: FeeAMM incorrect calculation
Currently, the FeeAMM contains an incorrect average calculation. This should be `(amount_user_token + amount_validator_token) / 2`, not `(amount_user_token * amount_validator_token) / 2`.  

  [Ref](https://github.com/tempoxyz/tempo/blob/3af8d19bee800b5f18b8980d1f331da8bea13c1c/crates/precompiles/src/tip_fee_manager/amm.rs#L251-L254)

### LINKED ISSUE #903: Bug: FeeManager `token_in_fees_array` is never cleared
Currently the FeeManager does not clear `token_in_fees_array` properly. This should be fixed and gated behind the `Moderato` hardfork.

[Ref](https://github.com/tempoxyz/tempo/blob/3af8d19bee800b5f18b8980d1f331da8bea13c1c/crates/precompiles/src/tip_fee_manager/mod.rs#L262-L272)

klkvr:
Closes #901 
Closes #902
Closes #903

Applies fixes to FeeManager and FeeAMM post-Moderato

- `validate_usd_currency` should check that token is a valid TIP20
- `drain_tokens_with_fees` should unset `token_in_fees_array`. that one is super bad because we're currently not collecting any fees on testnet because of this — added e2e test for it
- initially liquidity computation is not correct (should be `add` not `mul`)

### DIFF
--- a/crates/precompiles/src/tip_fee_manager/amm.rs
+++ b/crates/precompiles/src/tip_fee_manager/amm.rs
@@ -8,7 +8,6 @@ use alloy::{
     primitives::{Address, B256, IntoLogData, U256, keccak256, uint},
     sol_types::SolValue,
 };
-use tempo_chainspec::hardfork::TempoHardfork;
 use tempo_precompiles_macros::Storable;
 
 /// Constants from the Solidity reference implementation


crates/precompiles/src/tip20/mod.rs
```diff
@@ -101,6 +102,10 @@ pub fn validate_usd_currency<S: PrecompileStorageProvider>(
     token: Address,
     storage: &mut S,
 ) -> Result<()> {
+    if storage.spec() >= TempoHardfork::Moderato && !is_tip20(token) {
```
0xKitsune: nit: can we use `if storage.spec().is_moderato()`? This should be included in the latest changes to main.

crates/precompiles/src/tip_fee_manager/amm.rs
```diff
@@ -248,10 +249,17 @@ impl<'a, S: PrecompileStorageProvider> TipFeeManager<'a, S> {
 
         let liquidity = if total_supply.is_zero() {
             // Use checked math for multiplication and division
-            let mean = amount_user_token
-                .checked_mul(amount_validator_token)
-                .and_then(|product| product.checked_div(uint!(2_U256)))
-                .ok_or(TIPFeeAMMError::invalid_amount())?;
+            let mean = if self.storage.spec() >= TempoHardfork::Moderato {
```
0xKitsune: nit: same thing here can we use `spec.is_moderato()`?

crates/precompiles/src/tip_fee_manager/mod.rs
```diff
@@ -266,6 +266,9 @@ impl<'a, S: PrecompileStorageProvider> TipFeeManager<'a, S> {
         let mut tokens = Vec::new();
         while let Some(token) = TokensWithFees::pop(self)? {
             tokens.push(token);
+            if self.storage.spec() >= TempoHardfork::Moderato {
```
0xKitsune: nit: same here, can we use `storage.spec().is_moderato()`

PR #893, opened 2025-11-13
fix!(precompiles): add explicit invalidToken() error 

legion2002:
Currently, the TIP20::from_address function, does not check if the address actually has the TIP20 prefix.
This leads to random errors like PrecompileOOG, when it is used in places like the StablecoinDex.  
Eg. https://github.com/tempoxyz/docs/pull/182#discussion_r2515563659

We add an InvalidToken() error specifically for this case in this PR.

TBD: @0xKitsune is there a reason why we use unchecked address to tokenId conversion here - https://github.com/tempoxyz/tempo/blob/main/crates/precompiles/src/tip20/mod.rs#L640

If not, then we could move this invalid token check inside the TIP20 fns


crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -1138,6 +1138,14 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
             return Err(StablecoinExchangeError::identical_tokens().into());
         }
 
+        // Validate that both tokens are TIP20 tokens
+        if !is_tip20(token_in) {
+            return Err(StablecoinExchangeError::invalid_token().into());
+        }
+        if !is_tip20(token_out) {
+            return Err(StablecoinExchangeError::invalid_token().into());
+        }
```
0xKitsune, line 1424: This can be simplified to the following.

```rust
  if !is_tip20(token_in) || !is_tip20(token_out) {
            return Err(StablecoinExchangeError::invalid_token().into());
  }
```

We also need to gate this behind the `Moderato` hardfork since this called in `swap_exact_amount*` functions.

PR #908, opened 2025-11-13
feat: use cargo docs doc_cfg

mattsse:


scripts/build-cargo-docs.sh
```diff
@@ -0,0 +1,14 @@
...
+# Script to build cargo docs with the same flags as used in CI
+
+# Navigate to the root directory
+cd .. || exit 1
+
+echo "Building cargo docs..."
+
+# Build the documentation
+export RUSTDOCFLAGS="--cfg docsrs --show-type-layout --generate-link-to-definition --enable-index-page -Zunstable-options"
+cargo docs
```
onbjerg, line 12: is this not `cargo doc`?

mattsse: no, check new alias

.cargo/config.toml
```diff
@@ -1,3 +1,4 @@
 [alias]
 x = "run --package tempo-xtask --"
 xtask = "run --package tempo-xtask --"
+docs = "doc --workspace --all-features --no-deps"
```
mattsse, line 4: shorthand for

PR #916, opened 2025-11-13
fix: reject transactions without a fee token preference

klkvr:
rejects transactions if no transaction-level or user-level fee token preference is found

### DIFF
--- a/crates/transaction-pool/src/transaction.rs
+++ b/crates/transaction-pool/src/transaction.rs
@@ -71,7 +71,7 @@ pub enum TempoPoolTransactionError {
     #[error("Invalid fee token: {0}")]
     InvalidFeeToken(Address),
 
-    #[error("no fee token preference configured")]
+    #[error("No fee token preference configured")]
     MissingFeeToken,
 }
 
--- a/crates/transaction-pool/src/validator.rs
+++ b/crates/transaction-pool/src/validator.rs
@@ -71,6 +71,9 @@ where
         let fee_token = match state_provider.user_or_tx_fee_token(transaction.inner(), fee_payer) {
             Ok(Some(fee_token)) => fee_token,
             Ok(None) => {
+                // If no fee token preference was specified at transaction or user level, this means that we will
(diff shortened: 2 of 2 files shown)


crates/transaction-pool/src/transaction.rs
```diff
@@ -70,13 +70,17 @@ pub enum TempoPoolTransactionError {
 
     #[error("Invalid fee token: {0}")]
     InvalidFeeToken(Address),
+
+    #[error("no fee token preference configured")]
```
mattsse:
```suggestion
    #[error("No fee token preference configured")]
```

crates/transaction-pool/src/validator.rs
```diff
@@ -69,35 +68,36 @@ where
...
         };
 
-        let maybe_tx_fee_token =
-            match state_provider.user_or_tx_fee_token(transaction.inner(), fee_payer) {
-                Ok(maybe_tx_fee_token) => maybe_tx_fee_token,
-                Err(err) => {
-                    return TransactionValidationOutcome::Error(*transaction.hash(), Box::new(err));
-                }
-            };
+        let fee_token = match state_provider.user_or_tx_fee_token(transaction.inner(), fee_payer) {
```
mattsse, line 71: let's add a oneliner doc here why we're doing this check and the one in L84

crates/transaction-pool/src/validator.rs
```diff
@@ -69,35 +68,36 @@ where
...
+            Ok(Some(fee_token)) => fee_token,
+            Ok(None) => {
+                return TransactionValidationOutcome::Invalid(
+                    transaction,
+                    InvalidPoolTransactionError::other(TempoPoolTransactionError::MissingFeeToken),
+                );
+            }
```
mattsse, line 81: there is one issue with this branch, because this can race the actual `setUserToken(address)` tx if that one isn't mined yet

worth documenting here as well, imo we can ignore this case because as user you want to wait for the initial setUserToken(address) tx to confirm during onboarding

PR #921, opened 2025-11-14
chore(precompiles): drop `SlotId` in favor of runtime slot const

### LINKED ISSUE #859: chore(precompiles): drop `trait SlotId`
deprecate the `SlotId` sealed trait in favor of direct usage of macro-generated constants for each slot.

since slots will only be used by macro-generated fns (not by devs), the simplification is desirable

0xrusowsky:
## Motivation

towards #856 

closes #859 

deprecate the `SlotId` sealed trait in favor of direct usage of macro-generated constants for each slot.

since slots will only be used by macro-generated fns (not by devs), the simplification is desirable

## Solution

this simplification also allows us to evaluate `Mapping<K, V>` in the macro's compile-time constant resolution, which unlocks further simplifications (on top of #891) when allocationg slots. This is because mappings become another generic type which is allocated 32 bytes, and thus cannot be packed.

### DIFF
--- a/crates/precompiles/src/stablecoin_exchange/mod.rs
+++ b/crates/precompiles/src/stablecoin_exchange/mod.rs
@@ -18,7 +18,7 @@ use crate::{
         MAX_PRICE_POST_MODERATO, MAX_PRICE_PRE_MODERATO, MIN_PRICE_POST_MODERATO,
         MIN_PRICE_PRE_MODERATO, compute_book_key,
     },
-    storage::{PrecompileStorageProvider, Slot, VecSlotExt},
+    storage::{Mapping, PrecompileStorageProvider, Slot, VecSlotExt},
     tip20::{ITIP20, TIP20Token, is_tip20, validate_usd_currency},
 };
 use alloy::primitives::{Address, B256, Bytes, IntoLogData, U256};


crates/precompiles-macros/src/packing.rs
```diff
@@ -308,32 +276,12 @@ pub(crate) fn gen_slot_packing_logic(
...
-
-    // If current field is a mapping, it must start on a new slot
-    // TODO(rusowsky): Necessary to avoid type resolution issues. Remove once `SlotId` is dropped
-    if is_curr_mapping {
-        let slot_expr = quote! {{
-            #prev_slot_expr.checked_add(#prev_layout_slots).expect("slot overflow")
-        }};
-        return (slot_expr, quote! { 0 });
-    }
-
```
klkvr, line 331: nice

crates/precompiles-macros/src/packing.rs
```diff
@@ -88,28 +77,16 @@ pub(crate) struct LayoutField<'a> {
...
-        self.1
-    }
-    fn manual_slot(&self) -> Option<U256> {
-        None
-    }
-    fn base_slot_attr(&self) -> Option<U256> {
-        None
+impl<'a> AllocInfoExt for (&'a Ident, &'a Type) {
+    fn alloc_info(&self) -> (&Ident, &Type, Option<U256>, Option<U256>) {
+        (self.0, self.1, None, None)
```
klkvr: could we not just use `FieldInfo` here as well? https://github.com/tempoxyz/tempo/blob/d818dacf71bbe5d390142ad3e2ee2815f085b81e/crates/precompiles-macros/src/storable.rs#L49-L56

crates/precompiles-macros/src/packing.rs
```diff
@@ -249,9 +216,24 @@ pub(crate) fn gen_constants_from_ir(fields: &[LayoutField<'_>]) -> TokenStream {
...
-            pub const #slot_const: #slot_expr;
+            pub const #slot_const: ::alloy::primitives::U256 = #slot_expr;
             pub const #offset_const: usize = #offset_expr;
         });
+
+        // For the `Storable` macro, also generate the location constant
+        if gen_location {
+            constants.extend(quote! {
+                pub const #loc_const: crate::storage::packing::FieldLocation =
+                    crate::storage::packing::FieldLocation::new(#slot_const.as_limbs()[0] as usize, #offset_const, #bytes_expr);
```
klkvr, line 197: does this mean that we can't have slots > `usize::MAX`? not sure if it's reasonable given that there are e.g namespaced storage layouts that use slots much higher

crates/precompiles/src/storage/types/mapping.rs
```diff
@@ -579,37 +564,29 @@ impl<K1, K2, V, Base: SlotId, DummyId: SlotId> Mapping<K1, Mapping<K2, V, DummyI
...
+        Self::new(U256::ZERO)
     }
 }
 
 // Mappings occupy a full 32-byte slot in the layout (used as a base for hashing),
 // even though they don't store data in that slot directly.
 //
 // **NOTE:** Necessary to allow it to participate in struct layout calculations.
-impl<K, V, Base: SlotId> StorableType for Mapping<K, V, Base> {
+impl<K, V> StorableType for Mapping<K, V> {
```
klkvr, line 586: cool, I guess this also unlocks `Storable` impl for this eventually?

crates/precompiles-macros/src/packing.rs
0xrusowsky: this is the slot offset of a struct field vs the struct's base slot, so it's always safe to assume that a struct has less than 2**64 fields

(structs don't support custom slot alloc among its fields)

crates/precompiles/src/storage/types/mapping.rs
0xrusowsky: yes, although with the new API work that we'll do, i think the plan is for mappings to simply become a smart wrapper around slot

still need to explore this more

crates/precompiles-macros/src/packing.rs
klkvr: ah I see, this is not entered for `#[storage]` expander. let's add a note that `#slot_const` here is index of slot in the struct and is limited by the number of fields

crates/precompiles-macros/src/packing.rs
0xrusowsky: addressed in [1a2e449](https://github.com/tempoxyz/tempo/pull/921/commits/1a2e449c5fd675cd170bf869695faf8fa98d2129)

PR #923, opened 2025-11-14
feat(alloy): re-export `tempo-contracts`

onbjerg:
Re-exports `tempo-contracts` in `tempo-alloy`, since API consumers will likely want to interact with the enshrined contracts without having to also depend on `tempo-contracts`.

Note that this puts some requirements on `tempo-contracts` down the line, i.e. a relative stability guarantee, and good docs.

### DIFF
--- a/crates/alloy/Cargo.toml
+++ b/crates/alloy/Cargo.toml
@@ -11,7 +11,7 @@ publish.workspace = true
 workspace = true
 
 [dependencies]
-tempo-contracts = { workspace = true }
+tempo-contracts.workspace = true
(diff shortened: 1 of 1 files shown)


crates/alloy/Cargo.toml
```diff
@@ -11,6 +11,7 @@ publish.workspace = true
 workspace = true
 
 [dependencies]
+tempo-contracts = { workspace = true }
```
klkvr:
```suggestion
tempo-contracts.workspace = true
```

Issue #909, opened 2025-11-13
e2e tests: for restarts, do execution layer as well


SuperFluffy:
### Describe the feature

https://github.com/tempoxyz/tempo/pull/783 introduces restart e2e tests.

However, they are only taking down the consensus layer while keeping the node itself running.

We need to also take down and restart the execution layer.

One of the first tests this should implement is consensus layer to execution layer backfill. This simulates the case where the node was taken down but where the execution layer did not flush its state to disk.

PR #936, opened 2025-11-14
chore(e2e): use `TestingNode` instead

joshieDo:
* replace `RunningNode` and `PreparedNode` with an unified `TestingNode`
* adds shutdown to `execution_node` (closes https://github.com/tempoxyz/tempo/issues/909)


crates/e2e/src/execution_runtime.rs
```diff
@@ -12,18 +18,83 @@ use reth_ethereum::{
...
+    pub fn generate_many(count: u32, connect_peers: bool) -> Vec<Self> {
+        // Reserve ports by binding to them
+        let ports: Vec<u16> = (0..count)
+            .map(|_| {
+                let listener = TcpListener::bind("127.0.0.1:0").unwrap();
+                let port = listener
+                    .local_addr()
+                    .expect("failed to get local addr")
+                    .port();
+                drop(listener);
```
mattsse, line 113: this might work, but alos might not

I guess since we're running this in nextest this is probably fine, but worth documenting that this behaviour in case we see flaky tests

crates/e2e/src/execution_runtime.rs
```diff
@@ -142,17 +238,31 @@ impl Default for ExecutionRuntime {
...
 /// Can be used to spawn nodes.
+#[derive(Clone)]
 pub struct ExecutionRuntimeHandle {
     to_runtime: tokio::sync::mpsc::UnboundedSender<Message>,
+    nodes_dir: PathBuf,
 }
 
 impl ExecutionRuntimeHandle {
+    /// Returns the base directory where execution node data is stored.
+    pub fn nodes_dir(&self) -> &Path {
```
mattsse, line 388: let's name this datadir?

crates/e2e/src/lib.rs
```diff
@@ -242,40 +109,50 @@ pub async fn setup_validators(
...
-
     let mut private_keys = private_keys.into_iter();
-
-    // First, process the signers
-    for (private_key, share) in private_keys
+    let execution_configs =
+        ExecutionNodeConfig::generate_many(how_many_signers, connect_execution_layer_nodes);
+
+    // Spawn a dummy execution node to use as a placeholder in consensus configs
+    // This will be replaced when TestingNode::start_execution() is called
```
mattsse: this is a bit strange, why is this necessary exactly?
what's the impact of not doing this?

crates/e2e/src/execution_runtime.rs
```diff
@@ -12,18 +18,83 @@ use reth_ethereum::{
...
+    pub secret_key: B256,
+    /// List of trusted peer enode URLs to connect to.
+    pub trusted_peers: Vec<String>,
+    /// Port for the network service.
+    pub port: u16,
+}
+
+impl ExecutionNodeConfig {
+    /// Generate multiple execution node configs with optional trusted peer connections.
+    pub fn generate_many(count: u32, connect_peers: bool) -> Vec<Self> {
```
mattsse: this will be a pita to extend when we need more settings so would recommend doing a builder style pattern here instead for the args

crates/e2e/src/lib.rs
```diff
@@ -318,12 +196,12 @@ pub fn run(setup: Setup, mut stop_condition: impl FnMut(&str, &str) -> bool) ->
...
         // Setup and run all validators.
-        let (nodes, mut oracle) =
+        let (mut nodes, mut oracle) =
             setup_validators(context.clone(), &execution_runtime, setup).await;
 
-        let running = join_all(nodes.into_iter().map(|node| node.start())).await;
+        join_all(nodes.iter_mut().map(|node| node.start())).await;
 
-        link_validators(&mut oracle, &running, linkage, None).await;
+        link_validators(&mut oracle, &nodes, linkage, None).await;
```
mattsse: we have this repeated a bunch

wonder if it's worthing dooing a helper type for the response of setup_validators?

joshieDo: yeah definitely, didn't want to include it here tho

crates/e2e/src/lib.rs
joshieDo: no need. added `Builder::with_execution_node`

crates/e2e/src/execution_runtime.rs
joshieDo: added, for whats worth, havent had any flakiness yet

crates/e2e/src/execution_runtime.rs
joshieDo: makes sense

.github/workflows/test.yml
```diff
@@ -50,7 +50,7 @@ jobs:
       - uses: mozilla-actions/sccache-action@v0.0.9
       - uses: taiki-e/install-action@nextest
       - name: Run tests
-        run: cargo nextest run --no-fail-fast -j num-cpus
+        run: cargo nextest run --no-fail-fast -j num-cpus --retries 2
```
joshieDo, line 55: we are trying to reserve ports before we even spawn the nodes, so there's a chance/race condition inbetween the time we reserve a port and spawn the node that we might lose the port.

SuperFluffy: I have added a ton of tests in https://github.com/tempoxyz/tempo/pull/1249 because I need to test consensus hardfork behavior at genesis.

I think reserving ports will just keep on getting worse/more flaky with time?

joshieDo: changed it so we only try and reserve for tests that require EL p2p interconnected https://github.com/tempoxyz/tempo/pull/936/commits/93a4b2892fa16dd9fd5c2466c72e591b8fe0ca48

crates/e2e/src/execution_runtime.rs
```diff
@@ -392,33 +633,41 @@ pub async fn launch_execution_node<P: AsRef<Path>>(
         })
         .apply(|mut c| {
             c.network.discovery.disable_discovery = true;
+            c.network.trusted_peers = config
+                .trusted_peers
+                .into_iter()
+                .map(|s| TrustedPeer::from_str(&s).expect("invalid trusted peer enode"))
```
SuperFluffy: nit (official suggestino is to never use `FromStr::from_str` directly, but `str::parse` instead):
```suggestion
                .map(|s| s.parse::<TrustedPeer>().expect("invalid trusted peer enode"))
```

crates/e2e/src/execution_runtime.rs
```diff
@@ -259,13 +411,81 @@ impl ExecutionRuntime {
...
+                response: tx,
+            })
+            .wrap_err("the execution runtime went away")?;
+        rx.await.wrap_err(
+            "the execution runtime dropped the response channel before sending an execution node",
+        )
+    }
+
+    /// Requests a new execution node and blocks until its returned.
+    pub fn spawn_node_blocking(
```
SuperFluffy: I removed this in the meantime - probably made it here because of merge conflicts.

IMO we should remove these `*_blocking` methods. They are pretty useless/never used.

crates/e2e/src/lib.rs
```diff
@@ -400,53 +250,17 @@ pub async fn setup_validators(
...
-
-        nodes.push(PreparedNode {
-            execution_node,
-            public_key,
-            consensus_config,
-            oracle,
-            uid,
-        });
-    }
+    link_validators(&mut oracle, &nodes, linkage, None).await;
```
SuperFluffy, line 263: Sanity check (big PR): this is only run once during `setup_validators`, correct? I was originally doing this in all sorts of places and only recently moved it into the setup (as that's the way commonware intended).

crates/e2e/src/testing_node.rs
```diff
@@ -0,0 +1,474 @@
...
+    providers::{BlockchainProvider, StaticFileProvider},
+};
+use reth_node_builder::NodeTypesWithDBAdapter;
+use std::{path::PathBuf, sync::Arc};
+use tempo_commonware_node::consensus;
+use tempo_node::node::TempoNode;
+use tracing::debug;
+
+/// A testing node that can start and stop both consensus and execution layers.
+pub struct TestingNode {
```
SuperFluffy, line 22: Maybe just `Node`? :-) Little less verbose and IMO clear that this is just for testing.

```suggestion
pub struct Node {
```

crates/e2e/src/tests/backfill.rs
```diff
@@ -17,51 +17,29 @@ async fn run_validator_late_join_test(
...
-    assert_eq!(
-        last.execution_node
-            .node
-            .provider
-            .last_block_number()
-            .unwrap(),
-        0
-    );
+
+    let metrics_recorder = install_prometheus_recorder();
```
SuperFluffy: Install closer to the setup phase?

.github/workflows/test.yml
SuperFluffy: Do we need the `--retries` still? Probably yes because there is still a potential for flakiness?

crates/commonware-node/src/consensus/engine.rs
```diff
@@ -105,25 +105,32 @@ where
...
+        self.execution_node = Some(execution_node);
+        self
+    }
+
     pub async fn try_init(self) -> eyre::Result<Engine<TBlocker, TContext, TPeerManager>> {
-        let epoch_length = self
+        let execution_node = self
             .execution_node
+            .clone()
+            .ok_or_else(|| eyre::eyre!("execution_node must be set using with_execution_node()"))?;
```
SuperFluffy: From `eyre::OptionExt` (I think, one of their extension traits):
```suggestion
            .ok_or_eyre("execution_node must be set using with_execution_node()")
```

crates/commonware-node/src/consensus/engine.rs
```diff
@@ -66,7 +66,7 @@ pub struct Builder<TBlocker, TContext, TPeerManager> {
 
     pub fee_recipient: alloy_primitives::Address,
 
-    pub execution_node: TempoFullNode,
+    pub execution_node: Option<TempoFullNode>,
```
SuperFluffy, line 69: If it helps, we can make this an argument to `fn try_init` instead. Then we don't need to do these `Option` shennanigans.

crates/e2e/src/testing_node.rs
```diff
@@ -0,0 +1,474 @@
...
+
+    /// Stop only the consensus engine.
+    ///
+    /// # Panics
+    /// Panics if consensus is not running.
+    async fn stop_consensus(&mut self) {
+        let handle = self
+            .consensus_handle
+            .take()
+            .unwrap_or_else(|| panic!("consensus is not running for {}, cannot stop", self.uid));
```
SuperFluffy, line 268: bit less verbose (and don't care about perf/lazy exec):
```suggestion
            .expect(format!("consensus is not running for {}, cannot stop", self.uid))
```

crates/e2e/src/tests/restart.rs
```diff
@@ -58,28 +56,29 @@ fn run_restart_test(
         wait_for_height(&context, node_setup.how_many_signers, shutdown_height).await;
 
         // Randomly select a validator to kill
-        let idx = context.gen_range(0..running.len());
-        let to_restart = running.remove(idx).stop();
+        let idx = context.gen_range(0..nodes.len());
```
SuperFluffy, line 59: This is very non-random. We can as well just kill node `0` (nothing to do with this PR obv, just pointing it out).

crates/e2e/src/tests/restart.rs
```diff
@@ -58,28 +56,29 @@ fn run_restart_test(
...
+        nodes[idx].stop().await;
```
SuperFluffy, line 60: OK nice - with this we now truly shut down the execution layer.

crates/e2e/src/tests/restart.rs
```diff
@@ -58,28 +56,29 @@ fn run_restart_test(
...
         debug!(
             height = restart_height,
             "waiting for remaining validators to reach target height before restarting validator",
         );
-        wait_for_height(&context, running.len() as u32, restart_height).await;
+        let num_running = nodes.iter().filter(|n| n.is_running()).count() as u32;
+        wait_for_height(&context, num_running, restart_height).await;
 
-        running.push(to_restart.start().await);
+        nodes[idx].start().await;
```
SuperFluffy, line 71: Now that we started it, can we somehow ensure that we are not just syncing from genesis?

On the consensus side, we can check that the `marshal_processed_block` metric is not less than what we had when we stopped. Probably the same for the execution layer?

crates/e2e/src/testing_node.rs
```diff
@@ -0,0 +1,474 @@
...
+        // Verify RPC is accessible
+        let provider = ProviderBuilder::new().connect_http(rpc_url.parse().unwrap());
+        let block_number = provider.get_block_number().await;
+        assert!(block_number.is_ok(), "RPC should be accessible after start");
+
+        rpc_url
+    }
+
+    #[tokio::test]
+    async fn test_restart() {
```
SuperFluffy, line 417: We are already in a `tests` module; no need to remind the reader that this is a test (matches style of the other tests too):
```suggestion
    async fn node_stops_and_restarts() {
```

crates/e2e/src/testing_node.rs
```diff
@@ -0,0 +1,474 @@
...
+    use std::time::Duration;
+    use tokio::sync::{oneshot, oneshot::Sender};
+
+    enum Message {
+        Stop(Sender<()>),
+        Start(Sender<std::net::SocketAddr>),
+    }
+
+    /// Start node and verify RPC is accessible
+    async fn start_and_verify(tx_msg: &tokio::sync::mpsc::UnboundedSender<Message>) -> String {
```
SuperFluffy, line 402:
```suggestion
    async fn assert_node_starts_and_is_reachable(tx_msg: &tokio::sync::mpsc::UnboundedSender<Message>) -> String {
```

And can we give `tx_msg` a more descriptive name?

crates/e2e/src/testing_node.rs
```diff
@@ -0,0 +1,474 @@
...
+
+    #[tokio::test]
+    async fn test_restart() {
+        // Ensures that the node can be stopped completely and brought up inside a test.
+        let _ = tempo_eyre::install();
+
+        let runner = Runner::from(Config::default().with_seed(0));
+        let (tx_msg, mut rx_msg) = tokio::sync::mpsc::unbounded_channel::<Message>();
+
+        std::thread::spawn(move || {
```
SuperFluffy, line 424: This ends up a bit roundabout (thread -> deterministic runtime -> thread with tokio). Can we flatten this and remove the first thread?

crates/e2e/src/lib.rs
```diff
@@ -17,26 +17,26 @@ use commonware_cryptography::{
...
 use futures::future::join_all;
 use itertools::Itertools as _;
 use reth_node_metrics::recorder::PrometheusRecorder;
 use tempo_chainspec::TempoChainSpec;
-use tempo_commonware_node::consensus;
-use tracing::debug;
 
 pub mod execution_runtime;
+pub use execution_runtime::ExecutionNodeConfig;
+pub mod testing_node;
```
SuperFluffy, line 34: Nit (IMO clear we are not running any sort of full node).
```suggestion
pub mod node;
```

crates/e2e/src/testing_node.rs
```diff
@@ -0,0 +1,474 @@
...
+        rpc_url
+    }
+
+    #[tokio::test]
+    async fn test_restart() {
+        // Ensures that the node can be stopped completely and brought up inside a test.
+        let _ = tempo_eyre::install();
+
+        let runner = Runner::from(Config::default().with_seed(0));
+        let (tx_msg, mut rx_msg) = tokio::sync::mpsc::unbounded_channel::<Message>();
```
SuperFluffy, line 422: Can you give these more descriptive names? `transaction_message` is very non-descript.

crates/e2e/src/testing_node.rs
```diff
@@ -0,0 +1,474 @@
...
+                    connect_execution_layer_nodes: false,
+                    allegretto_in_seconds: None,
+                };
+
+                let (mut nodes, _execution_runtime) =
+                    setup_validators(context.clone(), setup).await;
+
+                let mut node = nodes.pop().unwrap();
+
+                loop {
```
SuperFluffy, line 445: For context with my other comments - I think just flattening this operations into the test would make it a bit easier to understand. I think the channels are not necessary.

crates/e2e/src/tests/restart.rs
```diff
@@ -173,51 +172,50 @@ async fn ensure_no_progress(context: &Context, tries: u32) {
...
-    for seed in 0..3 {
-        let setup = Setup::new()
-            .how_many_signers(3) // quorum for 3 validators is 3.
-            .seed(seed)
-            .epoch_length(100);
+    let setup = Setup::new()
+        .how_many_signers(3) // quorum for 3 validators is 3.
+        .seed(0)
+        .epoch_length(100)
+        .connect_execution_layer_nodes(true);
```
joshieDo, line 179: * https://github.com/tempoxyz/tempo/pull/1173 introduced a bug/edge-case where if the reth EL is not p2p connected AND on restart no finalization comes, there won't be a way for the restarted node to recover the blocks that it kept in memory

crates/e2e/src/lib.rs
joshieDo: yeah

crates/e2e/src/tests/restart.rs
SuperFluffy: Let's create an issue to track this.

Right now the assumption is that we trigger backfill syncs on FCU if the block delta > 64 (or >= 64?). We should just perform a backfill for cases <64 (or =< 64).

crates/e2e/src/tests/restart.rs
joshieDo: i'd be really surprised since we keep the tempdir on the ExecutionRuntime - still rn the node will record the last block on disk before shutting down, and compare against it when it starts up again.

crates/e2e/src/testing_node.rs
joshieDo: not allowed, clippy changes it

crates/e2e/src/tests/restart.rs
SuperFluffy: NOTE: we shouldn't have connected the nodes here just to get the test to pass.

This is hiding a real bug. Mentioning this line here: https://github.com/tempoxyz/tempo/issues/1309

PR #938, opened 2025-11-14
feat: payload builder defaults

klkvr:
Default payload builder interval is `1s` which effectively means that currently we are never attempting to build a better payload

This changes default to `300ms`

This is particularly useful for subblocks as lower interval means we are able to collect more of them.

### DIFF
--- a/bin/tempo/src/defaults.rs
+++ b/bin/tempo/src/defaults.rs
@@ -20,7 +20,7 @@ fn init_download_urls() {
 
 fn init_payload_builder_defaults() {
     DefaultPayloadBuilderValues::default()
-        .with_interval(Duration::from_millis(300))
+        .with_interval(Duration::from_millis(100))
         .try_init()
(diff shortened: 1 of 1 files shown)


bin/tempo/src/defaults.rs
```diff
@@ -17,3 +17,15 @@ pub(crate) fn init_download_urls() {
         .try_init()
         .expect("failed to initialize download URLs");
 }
+
+fn init_payload_builder_defaults() {
+    DefaultPayloadBuilderValues::default()
+        .with_interval(Duration::from_millis(300))
```
mattsse: could even go lower here?
if we target 400ms then perhaps do 75 or 100?

this is only respected after the first candidate is built so we can keep this pretty tight

klkvr: set to 100ms

not sure what's our current timeout for interrupting payload cc @Zygimantass, I believe 500ms?

PR #941, opened 2025-11-17
feat!(precompiles): return ABI-encoded UnknownFunctionSelector error (Moderato+)

fgimenez:
Improves error messages for unknown function selectors by returning ABI-encoded error data with the specific selector that was called, gated by the Moderato hardfork.

### DIFF
--- a/crates/contracts/src/precompiles/common_errors.rs
+++ b/crates/contracts/src/precompiles/common_errors.rs
@@ -1,22 +1,7 @@
-pub use ICommonPrecompileErrors::ICommonPrecompileErrorsErrors as CommonPrecompileError;
-
 use alloy::sol;
 
 sol! {
-    /// Common errors shared across all precompiles
+    /// Error returned when a function selector is not recognized
     #[derive(Debug, PartialEq, Eq)]
-    #[sol(rpc)]
-    interface ICommonPrecompileErrors {
-        /// Error returned when a function selector is not recognized
-        error UnknownFunctionSelector(bytes4 selector);
-    }
-}
-
-impl CommonPrecompileError {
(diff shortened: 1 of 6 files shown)


crates/precompiles/src/tip_account_registrar/dispatch.rs
```diff
@@ -23,33 +21,29 @@ impl<'a, S: PrecompileStorageProvider> Precompile for TipAccountRegistrar<'a, S>
...
-                    msg_sender,
-                    |_, call| {
-                        if self.storage.spec().is_moderato() {
-                            self.delegate_to_default_v2(call)
-                        } else {
-                            Err(TempoPrecompileError::UnknownFunctionSelector)
-                        }
-                    },
-                )
+                if self.storage.spec().is_moderato() {
```
mattsse, line 36: I see, this is effectively the same just with the if lifted out of the inner call

crates/contracts/src/precompiles/common_errors.rs
```diff
@@ -0,0 +1,22 @@
...
+
+sol! {
+    /// Common errors shared across all precompiles
+    #[derive(Debug, PartialEq, Eq)]
+    #[sol(rpc)]
+    interface ICommonPrecompileErrors {
+        /// Error returned when a function selector is not recognized
+        error UnknownFunctionSelector(bytes4 selector);
+    }
+}
```
klkvr: I think this could be just `sol! { error UnknownFunctionSelector(bytes4 selector) }`

fgimenez: sure done

PR #942, opened 2025-11-17
fix!(stablecoin-exchange): round up quote amounts (Moderato+)

### LINKED ISSUE #929: Bug: `calculate_quote_amount()` does not round up
### Describe the bug

[calculate_quote_amount() ](https://github.com/tempoxyz/tempo/blob/b3d9b23f0e80a6f6db6e18e3d33be55b339d36cb/crates/precompiles/src/stablecoin_exchange/mod.rs#L32-L38)does not round up, favoring the user instead of the exchange.

fgimenez:
Closes #929

Fixes a bug where `calculate_quote_amount()` used floor division (rounding down), which favored users over the exchange. Post-Moderato hardfork, the function now uses ceiling division (rounds up).

### DIFF
--- a/crates/precompiles/src/stablecoin_exchange/mod.rs
+++ b/crates/precompiles/src/stablecoin_exchange/mod.rs
@@ -33,9 +33,7 @@ pub const MIN_ORDER_AMOUNT: u128 = 10_000_000;
 /// Pre-Moderato behavior
 fn calculate_quote_amount_floor(amount: u128, tick: i16) -> Option<u128> {
     let price = tick_to_price(tick) as u128;
-    let scale = PRICE_SCALE as u128;
-    let product = amount.checked_mul(price)?;
-    product.checked_div(scale)
+    amount.checked_mul(price)?.checked_div(PRICE_SCALE as u128)
 }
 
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -29,12 +29,23 @@ use tempo_precompiles_macros::contract;
...
+
+/// Calculate quote amount using ceiling division (rounds up)
+/// Post-Moderato behavior
+fn calculate_quote_amount_ceil(amount: u128, tick: i16) -> Option<u128> {
+    let price = tick_to_price(tick) as u128;
+    let scale = PRICE_SCALE as u128;
+    let product = amount.checked_mul(price)?;
+    // Ceiling division: (a + b - 1) / b
+    product.checked_add(scale - 1)?.checked_div(scale)
 }
```
0xKitsune, line 44: Are we able to just use `amount.checked_mul(price)?.div_ceil(PRICE_SCALE as u128` here instead?

fgimenez: sure done

PR #943, opened 2025-11-17
chore(precompiles): harden storage layout tests

### LINKED ISSUE #949: test(precompiles): harden storage layout tests
## Motivation

harden precompile storage tests by ensuring we don't change slots by mistake

### Additional context

rather than commiting a snapshot file, we should have a dedicacted CI workflow that computes it from `main` on the fly. Ref: https://github.com/tempoxyz/tempo/pull/943#discussion_r2534955168

0xrusowsky:
## Motivation

harden precompile storage tests by ensuring we don't change slots by mistake.

## Solution
adds an extra validation layer to compare against snapshot of the precompile's storage layout.

additionally, updates all the solidity comparison tests to pull from the generated snapshot so that everything is in sync.


crates/precompiles/tests/testdata/solidity/arrays.layout.json
klkvr, line 1: in reth for that kind of tests we don't store snapshots in the git but just make a more complex ci job that checks out main and prepares testcases for the PR branch https://github.com/paradigmxyz/reth/blob/main/.github/workflows/compact.yml

imo ideally we should follow this pattern here

PR #948, opened 2025-11-17
fix: add delegateToDefaultV2 to prevent signature forgery

### LINKED ISSUE #932: Bug: `delegate_to_default()` signatures can be forged
### Describe the bug

In [delegate_to_default()](https://github.com/tempoxyz/tempo/blob/b3d9b23f0e80a6f6db6e18e3d33be55b339d36cb/crates/precompiles/src/tip_account_registrar/mod.rs#L40), given that the hash is unconstrained, signatures for any arbitrary public key can be forged as explained [here](https://crypto.stackexchange.com/a/44517). To mitigate this, the pre-image of the hash must be verified.

fgimenez:
Closes #932

Add `delegateToDefaultV2` to fix ECDSA signature forgery vulnerability.

The original `delegateToDefault` accepted an unconstrained hash parameter, allowing attackers to forge signatures for arbitrary addresses. This PR adds `delegateToDefaultV2` which:
  - Accepts a nonce instead of hash
  - Computes hash internally using domain separation
  - Includes nonce, delegate address, chain ID, and contract address in hash

The vulnerable v1 function is preserved for backward compatibility.


crates/precompiles/src/tip_account_registrar/mod.rs
```diff
@@ -80,6 +81,83 @@ impl<'a, S: PrecompileStorageProvider> TipAccountRegistrar<'a, S> {
...
+    ) -> Result<Address> {
+        let ITipAccountRegistrar::delegateToDefaultV2Call { nonce, signature } = call;
+
+        // Compute the hash internally with domain separation
+        let hash = compute_delegation_hash(
+            nonce,
+            DEFAULT_7702_DELEGATE_ADDRESS,
+            self.storage.chain_id(),
+            TIP_ACCOUNT_REGISTRAR,
+        );
```
klkvr: hmm I'm not sure that's how we want precompile to work

my understanding is that it should be possible to take any valid signature of a public key and use it to delegate the account vs it signing a specific payload

I guess we could make it take arbtirary `bytes` and then recover signature against `keccak256(bytes)`?

crates/precompiles/src/tip_account_registrar/dispatch.rs
```diff
@@ -26,6 +26,13 @@ impl<'a, S: PrecompileStorageProvider> Precompile for TipAccountRegistrar<'a, S>
                     |_, call| self.delegate_to_default(call),
                 )
             }
+            ITipAccountRegistrar::delegateToDefaultV2Call::SELECTOR => {
```
klkvr: should we deprecate the previous function at Moderato/Alegretto?

crates/precompiles/src/tip_account_registrar/mod.rs
fgimenez: good point, pushed changes for this ptal

crates/precompiles/src/tip_account_registrar/dispatch.rs
fgimenez: ok, pushed changes to make it return `Err(TempoPrecompileError::UnknownFunctionSelector)` post moderato, ptal

crates/contracts/src/precompiles/tip_account_registrar.rs
```diff
@@ -5,8 +5,12 @@ sol! {
     #[derive(Debug, PartialEq, Eq)]
     #[sol(rpc, abi)]
     interface ITipAccountRegistrar {
+        /// Pre-Moderato: accepts arbitrary hash (vulnerable to signature forgery)
         function delegateToDefault(bytes32 hash, bytes calldata signature) external returns (address authority);
 
+        /// Post-Moderato: accepts arbitrary message bytes, computes keccak256(bytes) internally
+        function delegateToDefaultV2(bytes calldata message, bytes calldata signature) external returns (address authority);
```
0xKitsune: nit: since we now return `UnknownSelector` if `delegateToDefault(bytes32 hash, bytes calldata signature)` is called, we could updated  `delegateToDefaultV2` to be `delegateToDefault(bytes calldata message, bytes calldata signature) external returns (address authority)`.

crates/precompiles/src/tip_account_registrar/mod.rs
```diff
@@ -153,13 +208,38 @@ mod tests {
...
+        // Encode the call
+        let call = ITipAccountRegistrar::delegateToDefaultCall {
+            hash,
+            signature: signature.as_bytes().into(),
+        };
+        let calldata = call.abi_encode();
+
+        // Should fail with UnknownFunctionSelector after Moderato
+        let result = registrar.call(&calldata, signer.address());
+        assert!(result.is_err());
```
0xKitsune: Can we assert the specific error here?

crates/precompiles/src/tip_account_registrar/mod.rs
```diff
@@ -245,4 +325,78 @@ mod tests {
...
+        match result {
+            Ok(recovered_addr) => {
+                // Should recover a different address than the actual signer
+                assert_ne!(recovered_addr, expected_address);
+            }
+            Err(_) => {
+                // Also acceptable - signature verification can fail
+            }
+        }
+    }
```
0xKitsune: This test should be deterministic right? Can we update to assert the epxected condition?

crates/contracts/src/precompiles/tip_account_registrar.rs
fgimenez: good point, added the overloaded function to the interface, both named `delegateToDefault`, and updated logic to reject v2 pre-moderato and v1 post-moderato, ptal

crates/precompiles/src/tip_account_registrar/mod.rs
fgimenez: sure, done

crates/precompiles/src/tip_account_registrar/mod.rs
fgimenez: good point done

crates/precompiles/src/tip_account_registrar/mod.rs
```diff
@@ -35,13 +35,65 @@ impl<'a, S: PrecompileStorageProvider> TipAccountRegistrar<'a, S> {
...
+    pub fn delegate_to_default_v2(
         &mut self,
-        call: ITipAccountRegistrar::delegateToDefaultCall,
+        call: ITipAccountRegistrar::delegateToDefault_1Call,
     ) -> Result<Address> {
-        let ITipAccountRegistrar::delegateToDefaultCall { hash, signature } = call;
+        let ITipAccountRegistrar::delegateToDefault_1Call { message, signature } = call;
+
+        // Compute the hash internally from the provided message
+        let hash = alloy::primitives::keccak256(&message);
```
klkvr, line 96: can we just reuse the fn above here?

fgimenez: sure, done

Issue #860, opened 2025-11-11
feat(precompiles): support individual struct field access


0xrusowsky:
- [x] #950 
- [ ] enhance the `Storable` macro to auto-generate individual field accessor methods

PR #951, opened 2025-11-17
chore: simplify types and storage helper fns

### LINKED ISSUE #950: chore(precompiles): simplify types and storage helper fns
## Motivation

in order to unlock https://github.com/tempoxyz/tempo/issues/860 we first need to further enhance/simplify our storage primitives.

we need to:
- [ ] refactor `Mapping<K, V>` to become an access wrapper around `Slot<T: Storable>`
- [ ] drop all the helper fns (`read_at_offset()`, `read_at_offset_packed()`, `write_at_offset()`, `write_at_offset_packed()`) and fully embrace `Slot` and `SlotPacked`

### Additional context

_No response_

0xrusowsky:
towards #860 

closes #950 

## Motivation

now that we have simplified the storage macros and dropped `SlotId`, we can start improving the devex.

this PR checks addresses all the subitems defined in #950
- [X] refactor `Mapping<K, V>` to become an access wrapper around `Slot<T: Storable>`
- [X] ~introduce a new `SlotPacked<T: Storable>` for packed fields~
   - since the `contract` macro relies on constant evaluation to determine packability, i had to modify `Slot` to have both, a `slot: U256` and a `ctx: LayoutCtx`.
   - to reduce the memory size of `Slot` i turned `enum LayoutCtx` into `struct LayoutCtx(u8)`
- [X] drop all the helper fns (`read_at_offset()`, `read_at_offset_packed()`, `write_at_offset()`, `write_at_offset_packed()`) and fully embrace `Slot` and `SlotPacked`

### DIFF
--- a/crates/precompiles/src/storage/types/mod.rs
+++ b/crates/precompiles/src/storage/types/mod.rs
@@ -69,8 +69,17 @@ impl Layout {
 
 /// Describes the context in which a `Storable` value is being loaded or stored.
 ///
-/// This determines whether the value occupies an entire storage slot or is packed
+/// Determines whether the value occupies an entire storage slot or is packed
 /// with other values at a specific byte offset within a slot.
+///
+/// **NOTE:** This type is not an enum to minimize its memory size, but its
+/// implementation is equivalent to:
+/// ```rs
+/// enum LayoutCtx {
+///    Full,
+///    Packed(usize)
+/// }
+/// ```
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 #[repr(transparent)]
 pub struct LayoutCtx(usize);
@@ -80,8 +89,7 @@ impl LayoutCtx {
     ///
     /// For writes, this directly overwrites the entire slot without needing SLOAD.
     /// All `Storable` types support this context.
-    #[allow(non_upper_case_globals)]
-    pub const Full: Self = Self(usize::MAX);
+    pub const FULL: Self = Self(usize::MAX);
 
     /// Load/store a packed primitive at the given byte offset within a slot.
     ///
@@ -90,8 +98,7 @@ impl LayoutCtx {
     /// packed fields in the same slot.
     ///
     /// Only primitive types with `Layout::Bytes(n)` where `n < 32` support this context.
-    #[allow(non_snake_case)]
-    pub const fn Packed(offset: usize) -> Self {
+    pub const fn packed(offset: usize) -> Self {
         debug_assert!(offset < 32);
         Self(offset)
     }
@@ -164,8 +171,8 @@ pub trait Storable<const SLOTS: usize>: Sized + StorableType {
     ///
     /// # Context
     ///
-    /// - `LayoutCtx::Full`: Load the entire value from `base_slot` (and subsequent slots if multi-slot)
-    /// - `LayoutCtx::Packed(offset)`: Load a packed primitive from byte `offset` within `base_slot`
+    /// - `LayoutCtx::FULL`: Load the entire value from `base_slot` (and subsequent slots if multi-slot)
+    /// - `LayoutCtx::packed(offset)`: Load a packed primitive from byte `offset` within `base_slot`
     ///
     /// # Errors
     ///
@@ -181,8 +188,8 @@ pub trait Storable<const SLOTS: usize>: Sized + StorableType {
     ///
     /// # Context
     ///
-    /// - `LayoutCtx::Full`: Write the entire value to `base_slot` (overwrites full slot)
-    /// - `LayoutCtx::Packed(offset)`: Write a packed primitive at byte `offset` (read-modify-write)
+    /// - `LayoutCtx::FULL`: Write the entire value to `base_slot` (overwrites full slot)
+    /// - `LayoutCtx::packed(offset)`: Write a packed primitive at byte `offset` (read-modify-write)
     ///
     /// # Errors
     ///
@@ -197,8 +204,8 @@ pub trait Storable<const SLOTS: usize>: Sized + StorableType {
     ///
     /// # Context
     ///
-    /// - `LayoutCtx::Full`: Clear entire slot(s) by writing zero
-    /// - `LayoutCtx::Packed(offset)`: Clear only the bytes at the offset (read-modify-write)
+    /// - `LayoutCtx::FULL`: Clear entire slot(s) by writing zero
+    /// - `LayoutCtx::packed(offset)`: Clear only the bytes at the offset (read-modify-write)
     ///
     /// The default implementation handles both contexts appropriately.
     ///
--- a/crates/precompiles/src/storage/types/slot.rs
+++ b/crates/precompiles/src/storage/types/slot.rs
@@ -46,7 +46,7 @@ impl<T> Slot<T> {
     pub const fn new(slot: U256) -> Self {
         Self {
             slot,
-            ctx: crate::storage::types::LayoutCtx::Full,
+            ctx: crate::storage::types::LayoutCtx::FULL,
             _ty: PhantomData,
         }
     }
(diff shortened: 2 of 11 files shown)


crates/precompiles/src/storage/types/mod.rs
```diff
@@ -86,7 +90,35 @@ pub enum LayoutCtx {
     /// packed fields in the same slot.
     ///
     /// Only primitive types with `Layout::Bytes(n)` where `n < 32` support this context.
-    Packed(usize),
+    #[allow(non_snake_case)]
+    pub const fn Packed(offset: usize) -> Self {
+        debug_assert!(offset < 32);
+        Self(offset)
+    }
```
0xrusowsky, line 104: i did this workaround to preserve the enum-like API (but having the mem size gain of simply using a `usize`)

personally i like it, but i acknowledge it may be weird. wdyt?

crates/precompiles-macros/src/layout.rs
```diff
@@ -160,23 +153,17 @@ pub(crate) fn gen_getters_and_setters(
                 impl<'a, S: crate::storage::PrecompileStorageProvider> #struct_name<'a, S> {
                     #[inline]
                     fn #getter_name(&mut self, key1: #key1_ty, key2: #key2_ty) -> crate::error::Result<#value_ty> {
-                        crate::storage::Mapping::<#key1_ty, crate::storage::Mapping<#key2_ty, #value_ty>>::new(slots::#slot_const).read_nested(
-                            self, key1, key2,
-                        )
+                        crate::storage::NestedMapping::<#key1_ty, #key2_ty, #value_ty>::new(slots::#slot_const).at(key1).at(key2).read(self)
```
klkvr, line 156: do we have a path for removing `NestedMapping` entirely and just operating on `Mapping<U256, Mapping<...>>`?

crates/precompiles/src/tip20_rewards_registry/mod.rs
```diff
@@ -49,36 +49,36 @@ impl<'a, S: PrecompileStorageProvider> TIP20RewardsRegistry<'a, S> {
     /// Add a token to the registry for a given stream end time
     pub fn add_stream(&mut self, token: Address, end_time: u128) -> Result<()> {
         let stream_key = keccak256((token, end_time).abi_encode());
-        let stream_ending_at = StreamEndingAt::new(slots::STREAMS_ENDING_AT);
-        let length = stream_ending_at.len(self, end_time)?;
+        let stream_ending_at = StreamEndingAt::new(slots::STREAMS_ENDING_AT).at(end_time);
```
klkvr, line 52: should we have a helper getter for mappings? so that instead of `StreamEndingAt::new(slots::STREAMS_ENDING_AT)` we can do `self.stream_endpoint_at()`?

crates/precompiles/src/stablecoin_exchange/order.rs
```diff
@@ -173,12 +170,9 @@ impl Order {
...
         let order_base_slot = mapping_slot(order_id.to_be_bytes(), super::slots::ORDERS);
-        OrderId::write_at_offset_packed(
-            storage,
-            order_base_slot,
-            __packing_order::NEXT_LOC,
-            new_next,
-        )?;
+        OrderId::new_at_offset(order_base_slot, __packing_order::NEXT_LOC.offset_slots)
+            .packed(__packing_order::NEXT_LOC.offset_bytes)
+            .write(storage, new_next)?;
```
klkvr: should we have `Slot::new_at_loc`? also I guess eventually this will be just a getter on `Order`?

crates/precompiles-macros/src/layout.rs
0xrusowsky: the reason why i introduced `NestedMapping` was so that it becomes accessible for users when they do `self.<nested_mapping_field>`.

this will be `NestedMapping` which only has `fn slot()` and `fn at() -> Mapping`

so it naturally guides users towards `Slot` (`NestedMapping -> Mapping -> Slot`). Once they have a slot, they can read/write/delete it with the a unified API.

crates/precompiles/src/tip20_rewards_registry/mod.rs
0xrusowsky: as mentioned above, the idea is for users to do `let stream_ending_at: Slot<u64> = self.stream_ending.at(end_time)?;`

all this helpers will go away in the next PR once i touch the `contract` macro

crates/precompiles-macros/src/layout.rs
0xrusowsky: note that this doesn't work with `Mapping<K1, Mapping<K2, V>>` cause we can't have `fn at()` implemented twice (one would return `Slot` and the other `Mapping`)

crates/precompiles/src/stablecoin_exchange/order.rs
0xrusowsky: yes, that's the plan, but i agree that `Slot::new_at_loc` is cleaner that the builder pattern that i implemented

0xrusowsky: addressed in [95edd48](https://github.com/tempoxyz/tempo/pull/951/commits/95edd487cdf3c63f196fd451fc9fa54e636fb21b)

crates/precompiles/src/storage/types/slot.rs
```diff
@@ -27,18 +27,62 @@ use crate::{
...
+    }
+
+    /// Creates a new `Slot` with the given base slot number with the given offset.
+    ///
+    /// This is a convenience method for accessing struct fields.
+    /// Creates a full-slot accessor. For packed fields, use `new_at_loc` instead.
+    #[inline]
+    pub const fn new_at_offset(base_slot: U256, offset_slots: usize) -> Self {
+        Self {
+            slot: base_slot.saturating_add(U256::from_limbs([offset_slots as u64, 0, 0, 0])),
```
mattsse, line 61: isn't this just `U256::from(offset_slots)` ?

klkvr: I think won't work in `const`

crates/precompiles/src/stablecoin_exchange/order.rs
```diff
@@ -158,12 +158,8 @@ impl Order {
...
     ) -> Result<(), TempoPrecompileError> {
         let order_base_slot = mapping_slot(order_id.to_be_bytes(), super::slots::ORDERS);
-        OrderAmount::write_at_offset_packed(
-            storage,
-            order_base_slot,
-            __packing_order::REMAINING_LOC,
-            new_remaining,
-        )?;
+        OrderAmount::new_at_loc(order_base_slot, __packing_order::REMAINING_LOC)
+            .write(storage, new_remaining)?;
```
0xKitsune, line 162: Are there still plans to support `Order::write_order_id()` in the future? Or is `OrderAmount::new_at_loc` the proposed solution for the forseeable future?

crates/precompiles/src/storage/types/mod.rs
0xKitsune: I dont have a very strong opinion but when reading through the code this might convey that this is an enum when it is actually not which could be a potential source of confusion. Id adjust to remove any possible ambiguity.

crates/precompiles/src/stablecoin_exchange/order.rs
0xrusowsky: same as here: https://github.com/tempoxyz/tempo/pull/951#discussion_r2538242355

the idea is for users to not have to deal with slots/offsets/packing at all, this should be invisible to consumers of the storage primitives

crates/precompiles/src/storage/types/mod.rs
0xrusowsky: my intent was to clearly showcase the duality of `Full` or `Packed`, which is better reflected with an enum.

but to your point, i guess it doesn't hurt to use regular casing and let people check the doc comments if unclear.

addressed in [`ab39f3a` (#951)](https://github.com/tempoxyz/tempo/pull/951/commits/ab39f3af357c965498447ca47a020d2fcd002a56)

PR #953, opened 2025-11-18
fix(precompiles): disable scheduled TIP20 rewards  

### LINKED ISSUE #935: Bug: Malicious user could potentially cause system tx to exceed 30M gas and halt block production
### Describe the bug

A user can mint a large number of new TIP20 tokens, create tiny reward streams for each, and schedule all of them to end at the same timestamp. When all streams end at once, the system transaction for that block triggers `finalize_streams()`, which must iterate over every pending stream. System transactions run with a fixed 30M gas limit via `system_call_with_caller`, so the traversal can exceed this limit and cause the system transaction to fail. That results in an invalid block and effectively halts the chain.

0xKitsune:
Closes #935 

This PR disables TIP20 scheduled reward and removes the `TIP20RewardRegistry` system transaction. By eliminating scheduled reward streams, we remove the DOS vector described in #935.

With scheduled rewards disabled, the registry system transaction is no longer needed and is removed post `Moderato` hardfork. TIP20 rewards where `duration=0` continue to function as normal. We will reintroduce scheduled rewards in a future hardfork with a design that avoids usage of system txs.

### DIFF
--- a/crates/evm/src/evm.rs
+++ b/crates/evm/src/evm.rs
@@ -231,3 +231,42 @@ where
         )
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use reth_revm::context::BlockEnv;
+    use revm::{context::TxEnv, database::EmptyDB};
+
+    use super::*;
+
+    #[test]
+    fn can_execute_system_tx() {
+        let mut evm = TempoEvm::new(
+            EmptyDB::default(),
+            EvmEnv {
+                block_env: TempoBlockEnv {
+                    inner: BlockEnv {
+                        basefee: 1,
+                        ..Default::default()
+                    },
+                    ..Default::default()
+                },
+                ..Default::default()
+            },
+        );
+        let result = evm
+            .transact(TempoTxEnv {
+                inner: TxEnv {
+                    caller: Address::ZERO,
+                    gas_price: 0,
+                    gas_limit: 21000,
+                    ..Default::default()
+                },
+                is_system_tx: true,
+                ..Default::default()
+            })
+            .unwrap();
+
+        assert!(result.result.is_success());
+    }
+}


crates/consensus/src/lib.rs
```diff
@@ -219,14 +219,23 @@ impl Consensus<Block> for TempoConsensus {
             )));
         }
 
-        // Check for optional rewards registry system transaction at the start
-        if let Some(first_tx) = transactions.first()
-            && first_tx.is_system_tx()
-            && first_tx.to().unwrap_or_default() != TIP20_REWARDS_REGISTRY_ADDRESS
+        // If the moderator hardfork is not active, valdiate that the TIP20 Rewards Registry system
```
fgimenez:
```suggestion
        // If the moderator hardfork is not active, validate that the TIP20 Rewards Registry system
```

crates/evm/src/block.rs
```diff
@@ -350,16 +358,25 @@ where
         tx: &TempoTxEnvelope,
         gas_used: u64,
     ) -> Result<BlockSection, BlockValidationError> {
+        let block = self.evm().block();
+        let block_timestamp = block.timestamp.to::<u64>();
+        let post_moderato = self
```
fgimenez, line 363: very nit, this moderato check is done per tx, we could do it once in the calling site

0xKitsune: Sounds good, lets plan to update this in a follow on PR.

crates/consensus/src/lib.rs
```diff
@@ -219,14 +219,23 @@ impl Consensus<Block> for TempoConsensus {
...
-            && first_tx.is_system_tx()
-            && first_tx.to().unwrap_or_default() != TIP20_REWARDS_REGISTRY_ADDRESS
+        // If the moderator hardfork is not active, validate that the TIP20 Rewards Registry system
+        // tx is the first transaction in the block. If the block timestamp is post moderato, skip
+        // this step
+        if !self
+            .inner
+            .chain_spec()
+            .is_moderato_active_at_timestamp(block.timestamp())
         {
```
mattsse, line 229: I see, this just wraps the activation check

PR #959, opened 2025-11-18
test(bench): Include DEX orders and swaps in `max_tps`

RomanHodulak:
Ref #614

Includes these steps in `tempo-bench`
* Setup DEX to be able to send DEX specific transactions.
* Generate DEX specific transactions in the spammer. 
* Add weights and other parameters to control the benchmark

### DIFF
--- a/bin/tempo-bench/src/cmd/max_tps/dex.rs
+++ b/bin/tempo-bench/src/cmd/max_tps/dex.rs
@@ -49,12 +49,12 @@ pub(super) async fn setup(
     let user_tokens = [*base1.address(), *base2.address()];
     let base = [base1.clone(), base2.clone()];
 
-    let quote = ITIP20Instance::new(token_id_to_address(0), provider.clone());
+    let quote_address = token_id_to_address(0);
 
     let mint_amount = U256::from(1000000000000000u128);
     let first_order_amount = 1000000000000u128;
 
-    let tokens = [&base1, &base2, &quote];
+    let tokens = [&base1, &base2];
     let mut futures = Vec::new();
     let nonce = provider.get_transaction_count(wallet.address()).await?;
 
@@ -102,7 +102,7 @@ pub(super) async fn setup(
 
     for signer in signers {
         let nonce = provider.get_transaction_count(signer.address()).await?;
-        let quote = ITIP20::new(*quote.address(), provider.clone());
+        let quote = ITIP20::new(quote_address, provider.clone());
         let tokens = base.clone().into_iter().chain(std::iter::once(quote));
         let tokens_count = base.len() as u64 + 1;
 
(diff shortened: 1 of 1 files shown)


Justfile
```diff
@@ -21,13 +21,13 @@ build binary extra_args="":
 
 [group('localnet')]
 [doc('Generates a genesis file')]
-genesis accounts="1000" output="genesis.json" profile="maxperf":
+genesis accounts="1000000" output="genesis.json" profile="maxperf":
```
shekhirin: I think it's not necessary for a default, will take way too long

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -119,6 +119,14 @@ pub struct MaxTpsArgs {
     #[arg(long, default_value = "8")]
```
shekhirin: this can likely be bumped to 1024? Hard limit in Linux is 4096, so we should be able to open at least 1024.

RomanHodulak: Without any tinkering with my machine, I'm getting 429 too many connections error. This does not seem like a reasonable default.

To be defensive and use a value that will most likely work, even if it's not the most optimal one nor sub-optiomal, i just bumped it to 128.

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -99,15 +113,50 @@ pub struct MaxTpsArgs {
     /// Benchmark mode for metadata (e.g., "max_tps", "stress_test")
     #[arg(long)]
     benchmark_mode: Option<String>,
+
+    /// A limit of the maximum amount of concurrent requests, prevents issues with too many
+    /// connections open at once.
+    #[arg(long, default_value = "128")]
+    max_concurrent_requests: usize,
```
shekhirin: can we use `total_connections` instead of a new argument? Or is it not the same?

bin/tempo-bench/src/cmd/max_tps/dex.rs
```diff
@@ -0,0 +1,359 @@
...
+    let base2 = setup_test_token(provider.clone(), caller, &tx_count).await?;
+    let user_tokens = [*base1.address(), *base2.address()];
+    let base = [base1.clone(), base2.clone()];
+
+    let quote = ITIP20Instance::new(token_id_to_address(0), provider.clone());
+
+    let mint_amount = U256::from(1000000000000000u128);
+    let first_order_amount = 1000000000000u128;
+
+    let tokens = [&base1, &base2, &quote];
```
shekhirin: Minting the quote token (linkingUSD) will not work on a real network, we should either make it optional, or remove altogether. Unsure if you even need it for swaps?

PR #962, opened 2025-11-18
poc(precompiles): thread local example

0xrusowsky:
## Motivation

POC to explore how a threadlocal impl to avoid passing storage around would look like

## Note

this is a prototype built on top of the current architecture, but obviously we would have to modify the existing traits/fns/APIs rather than having 2 impls that cohexist


crates/precompiles/src/thread_local_example/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+        {
+            // uses rewards' address from `rewards._address_guard`
+            let rewards = ThreadLocalRewards::new(REWARDS_ADDRESS)?;
+            rewards.distribute(amount)?;
+        }
+
+        // once rewards is dropped, its guard is also dropped and we switch back to `token._address_guard`
+
+        Ok(())
+    }
```
0xrusowsky, line 118: how do we feel about the scope of cross-contract calls? is this API safe enough, or is it a footgun?

note that if we did:
```rs
self.transfer(from, to, amount)?;
let rewards = ThreadLocalRewards::new(REWARDS_ADDRESS)?;
rewards.distribute(amount)?;
self.do_something()?; // this would target rewards' storage, as that's still the 
                      // top-level address of the stack and hasn't been dropped

Comment (0xrusowsky) [reply crates/precompiles/src/thread_local_example/mod.rs:118]: alternatives that we could explore:

1. change contract's `fn new` for a closure-like API `fn call<F>(address, f: F)` which is clearly scoped:
```rs
ThreadLocalRewards::call(REWARDS_ADDRESS, |rewards| {
   rewards.distribute(amount)
})?;
```

2. store address in `Slot` struct (this increases the memory usage by at least 20bytes per slot

3. (hardest to impl) rather than creating the `AddressGuard` on `fn new()` rely on the contract macro to inject it on every public method, so that we prevent any scoping issues. I feel like this one has the same devex as 2. without the increased memory size (but feels more hacky) and may be more difficult to implement

0xrusowsky: i think my preferred option may be 1. because of the clear intent

0xrusowsky: example of how 1. would look like: https://github.com/tempoxyz/tempo/pull/966/files

with closures we can differentiate between calls and staticcalls, providing compile-time guarantees, which is quite cool.
(note that calling `call_mut` from an immutable fn that takes `&self` is still possible, but i think we could further restrict that)

0xrusowsky: UPDATE: [`d1cf227` (#966)](https://github.com/tempoxyz/tempo/pull/966/commits/d1cf227def774853a2011c2bfe5db6e2338142f9) introduces `trait CallCtx` and `struct CallBuilder` to enforce x-contract call mutability consistency

note that if we didn't want to go with the closure approach, we should also be able to use this pattern/type system to enforce it in other alternatives

PR #965, opened 2025-11-18
refactor: pull pagination types into `tempo-alloy`

onbjerg:
Pulls pagination related types into `tempo-alloy` (`PaginatedParams<T>`, `FilterRange`, sorting) into `tempo-alloy` so we can reuse these types in the provider extensions.

There is no functional change.

This also includes an empty file for the provider extensions since I am basing my work off of this locally (and it avoids some annoying merge conflicts).

### DIFF
--- a/crates/commonware-node/src/consensus/application/actor.rs
+++ b/crates/commonware-node/src/consensus/application/actor.rs
@@ -718,6 +718,9 @@ impl Inner<Uninit> {
         }
         .build(context.with_label("executor"));
 
+        let executor_mailbox = executor.mailbox().clone();
+        let executor_handle = executor.start();
+
         let initialized = Inner {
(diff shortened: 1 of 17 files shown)


crates/node/src/rpc/dex/orders.rs
```diff
@@ -1,28 +1,10 @@
...
-    pub order: OrdersSortOrder,
-}
-
-#[derive(Clone, Debug, Default, PartialEq, Eq, Hash, Serialize, Deserialize)]
-#[serde(rename_all = "camelCase")]
-pub enum OrdersSortOrder {
-    Asc,
-    #[default]
-    Desc,
-}
```
onbjerg, line 24: I pulled these types into the pagination module as well as they are used in there, and gave them more general names not related to dex orders.

PR #968, opened 2025-11-19
feat: add basic joinset type

mattsse:
same motivation as #967

this adds a very basic joinset type that groups the tasks together and aborts them if the `Engine::start` future itself is dropped (e.g. manually dropped for restart).

because if we only abort:

https://github.com/tempoxyz/tempo/blob/afa9423403e50fc1ec52364e153fb99edb41095c/crates/commonware-node/src/consensus/engine.rs#L359-L359

then this doesn't propagate to all the tasks that `start` itself spawned.

ideally we can get this primitive into cw eventually

### DIFF
--- a/crates/commonware-node/src/consensus/engine.rs
+++ b/crates/commonware-node/src/consensus/engine.rs
@@ -25,7 +25,6 @@ use commonware_runtime::{
 };
 use commonware_utils::set::Ordered;
 use eyre::WrapErr as _;
-use futures::future::try_join_all;
 use rand::{CryptoRng, Rng};
 use tempo_node::TempoFullNode;
 
@@ -438,18 +437,20 @@ where
 
         let dkg_manager = self.dkg_manager.start(dkg_channel);
 
-        try_join_all(vec![
+        let handles = vec![
             application,
             broadcast,
             epoch_manager,
             marshal,
             dkg_manager,
             subblocks,
-        ])
-        .await
-        .map(|_| ())
-        // TODO: look into adding error context so that we know which
-        // component failed.
-        .wrap_err("one of the consensus engine's actors failed")
+        ];
+
+        let mut set = crate::util::JoinSet::from_vec(handles);
+        while let Some(res) = set.join_next().await {
+            res.wrap_err("one of the consensus engine's actors failed")?;
+        }
+
+        Ok(())
     }
 }
--- a/crates/commonware-node/src/util.rs
+++ b/crates/commonware-node/src/util.rs
@@ -1,0 +1,219 @@
+//! Commonware utils.
+
+use commonware_runtime::Handle;
+use futures::FutureExt;
+use std::task::{Context, Poll};
+
+/// A collection of tasks spawned on the commonware runtime.
+///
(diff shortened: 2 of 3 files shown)


joshieDo (approved):

crates/commonware-node/src/util.rs
```diff
@@ -0,0 +1,219 @@
...
+/// All of the tasks must have the same return type `T`.
+///
+/// When the `JoinSet` is dropped, all tasks in the `JoinSet` are immediately aborted.
+///
+/// This is primarily intended to group small sets of [`Handle`] to ensure all get aborted on drop.
+pub struct JoinSet<T>
+where
+    T: Send + 'static,
+{
+    tasks: Vec<Handle<T>>,
```
SuperFluffy, line 21: Could we just toss this into a `FuturesUnordered` instead? What's the advantage of using this over that primitive?

crates/commonware-node/src/util.rs
```diff
@@ -0,0 +1,219 @@
...
+
+            // No more results
+            let third = set.join_next().await;
+            assert!(third.is_none());
+        });
+    }
+
+    #[test]
+    fn tasks_are_aborted_on_drop() {
+        struct SendOnDrop(tokio::sync::mpsc::UnboundedSender<()>);
```
SuperFluffy: futures unboundedsender for consistency?

crates/commonware-node/src/util.rs
mattsse: that should also work, iirc this also gives us internal access

mattsse: doesn't really matter here because this fn will only be polled once

SuperFluffy: true

crates/commonware-node/src/consensus/engine.rs
```diff
@@ -438,18 +437,20 @@ where
...
             subblocks,
-        ])
-        .await
-        .map(|_| ())
-        // TODO: look into adding error context so that we know which
-        // component failed.
-        .wrap_err("one of the consensus engine's actors failed")
+        ];
+
+        let mut set = crate::util::JoinSet::from_vec(handles);
```
SuperFluffy, line 449: Wonder if we should turn this into a primitive `JoinMap` to return which of the tasks failed.

crates/commonware-node/src/util.rs
```diff
@@ -0,0 +1,220 @@
...
+        Self::new()
+    }
+}
+
+impl<T> Drop for JoinSet<T>
+where
+    T: Send + 'static,
+{
+    fn drop(&mut self) {
+        self.tasks.drain(..).for_each(|handle| handle.abort());
```
SuperFluffy, line 107: Mentioned elsewhere: something is odd if we actually need to do that. It's clear that one of the tasks going down should by itself not kill the other tasks.

But what should happen (and apprently doesn't):

1. task exits
2. `consensus::Engine::start` returns
3. "root context" completes
4. all other tasks get aborted

https://github.com/commonwarexyz/monorepo/blob/889339894ca53f3bdc694ecd94a555ab86230d29/runtime/src/lib.rs#L158-L170

PR #976, opened 2025-11-20
feat(chainspec): add Allegretto hardfork

fgimenez:
Same as we have for `Moderato` also added a missing test for `tempo_hardfork_at`

### DIFF
--- a/crates/chainspec/src/hardfork.rs
+++ b/crates/chainspec/src/hardfork.rs
@@ -154,7 +154,7 @@ mod tests {
         assert!(!TempoHardfork::Moderato.is_allegretto());
 
         assert!(TempoHardfork::Allegretto.is_allegretto());
-       
+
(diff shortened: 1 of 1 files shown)


crates/chainspec/src/hardfork.rs
```diff
@@ -129,4 +147,12 @@ mod tests {
 
         assert!(TempoHardfork::Moderato.is_moderato());
     }
+
+    #[test]
+    fn test_is_allegretto() {
+        assert!(!TempoHardfork::Adagio.is_allegretto());
+        assert!(!TempoHardfork::Moderato.is_allegretto());
+
+        assert!(TempoHardfork::Allegretto.is_allegretto());
```
rakita, line 156: It would be good to have one more case.


```suggestion
        assert!(TempoHardfork::Allegretto.is_allegretto());
       
        assert!(TempoHardfork::Allegretto.is_moderato());
```

PR #980, opened 2025-11-20
fix(sidecar): add graceful shutdowns

Zygimantass:
adds response to CTRL+C and SIGTERM which will make restarts in Kubernetes faster


bin/tempo-sidecar/src/cmd/monitor.rs
```diff
@@ -57,13 +58,28 @@ impl MonitorArgs {
...
+            .context("failed to install SIGINT handler")?;
+
+        tokio::select! {
+            _ = sigterm.recv() => tracing::info!("Received SIGTERM, shutting down gracefully"),
+            _ = sigint.recv() => tracing::info!("Received SIGINT, shutting down gracefully"),
+        }
+
+        // Abort tasks
+        monitor_handle.abort();
+        server_handle.abort();
```
mattsse, line 80: do we need the monitor task to actually end gracefully?

because this can return before the task was fully terminated, but the tokio rt also drops them when the program ends

PR #982, opened 2025-11-20
Add transaction latency monitor to tempo-sidecar

Zygimantass:
## Summary
- add a transaction latency monitoring command to tempo-sidecar that measures time from pending observation to on-chain inclusion
- expose Prometheus histogram and pending gauge metrics with a chain label, alongside a metrics endpoint
- wire the new monitor into the CLI subcommands

## Testing
- cargo fmt
- cargo check -p tempo-sidecar *(fails: network error while fetching git dependency)*

------
[Codex Task](https://chatgpt.com/codex/tasks/task_b_691f4b5a7608832a832f993f208139dd)


bin/tempo-sidecar/src/cmd/tx_latency.rs
```diff
@@ -0,0 +1,199 @@
...
+        let tracked_hashes: Vec<B256> = self.pending.keys().cloned().collect();
+
+        for hash in tracked_hashes {
+            match provider.get_transaction_receipt(hash).await {
+                Ok(Some(_)) => {
+                    if let Some(seen_at) = self.pending.remove(&hash) {
+                        let latency = seen_at.elapsed();
+                        histogram!("tempo_tx_landing_latency_seconds")
+                            .record(latency.as_secs_f64());
+                    }
```
klkvr: I think instead of measuring it via an `Instant` and thus having accuracy depend on the configured `poll_interval` we could measure it by fetching block and looking at `timestamp_millis`

bin/tempo-sidecar/src/cmd/tx_latency.rs
```diff
@@ -0,0 +1,235 @@
...
+    async fn observe_pending<P>(&mut self, provider: &P) -> Result<()>
+    where
+        P: Provider,
+    {
+        gauge!("tempo_tx_latency_pending_observed").set(self.pending.len() as f64);
+
+        let tracked_hashes: Vec<B256> = self.pending.keys().cloned().collect();
+
+        for hash in tracked_hashes {
+            match provider.get_transaction_receipt(hash).await {
```
mattsse: once we get more TPS then this will be a bottleneck eventually

e.g. if time it takes to fetch X receipts takes longer than the actual block time

bin/tempo-sidecar/src/cmd/tx_latency.rs
```diff
@@ -0,0 +1,235 @@
...
+                                .await
+                                .context("failed to resubscribe to pending transactions")?;
+                            stream = subscription.into_stream();
+                            continue;
+                        }
+                    }
+                }
+                _ = receipt_interval.tick() => {
+                    self.observe_pending(&provider).await?;
+                }
```
mattsse: since we're only interested in the time to inclusion effectively, we can make this more efficient by observing new blocks and check the included hashes, then we only need 1 request per block

there's `.subscribe_full_blocks()`

PR #984, opened 2025-11-20
feat: expose `feePayer` on RPC receipt

klkvr:


crates/node/src/rpc/mod.rs
```diff
@@ -334,26 +334,32 @@ impl ReceiptConverter<TempoPrimitives> for TempoReceiptConverter {
         &self,
         receipts: Vec<ConvertReceiptInput<'_, TempoPrimitives>>,
     ) -> Result<Vec<Self::RpcReceipt>, Self::Error> {
-        let receipts = self.inner.convert_receipts(receipts)?;
-        Ok(receipts
+        let txs = receipts.iter().map(|r| r.tx).collect::<Vec<_>>();
+        self.inner
```
mattsse, line 338: I see, this collect is pretty cheap because a ref of tx

PR #985, opened 2025-11-21
feat!(stablecoin-exchange): create new dex pair on first order

### LINKED ISSUE #970: Create new dex Pair on first order
### Describe the feature

Instead of requiring that pairs are created explicitly via `createPair`, create a pair between a token / its quote token if one does not exist when an order is placed

This will make the stablecoin exchange behave in the same way as the FeeAMM

Other considerations
- will need to decide what to return if someone requests a quote for a pair that does not exist yet, is it still PairDoesNotExist or do we return InsufficientLiquidity

### Additional context

_No response_

0xKitsune:
Closes #970 

This PR updates the `StablecoinExchange` precompile to remove the `createPair()` function from the public interface and automatically create a new pair when placing an order if the orderbook is not already created. This PR gates this logic behind the `Allegretto` hardfork.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11240,7 +11240,9 @@ dependencies = [
  "alloy-serde",
  "alloy-signer",
  "alloy-signer-local",
+ "alloy-transport",
  "derive_more",
+ "eyre",
  "reth-evm",
  "reth-primitives-traits",
  "reth-rpc-convert",
@@ -11251,6 +11253,7 @@ dependencies = [
  "tempo-primitives",
  "tempo-revm",
  "test-case",
+ "tokio",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -189,6 +189,7 @@ alloy-serde = "1.1.2"
 alloy-signer = "1.1.2"
 alloy-signer-local = "1.1.2"
 alloy-sol-types = "1.4.1"
+alloy-transport = "1.1.2"
 
 commonware-broadcast = "0.0.63"
 commonware-codec = "0.0.63"
--- a/bin/tempo-bench/src/cmd/max_tps.rs
+++ b/bin/tempo-bench/src/cmd/max_tps.rs
@@ -3,7 +3,7 @@ mod dex;
 use itertools::Itertools;
 use reth_tracing::{
     RethTracer, Tracer,
-    tracing::{debug, error, info, warn},
+    tracing::{debug, error, info},
 };
 use tempo_alloy::TempoNetwork;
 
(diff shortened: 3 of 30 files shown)


crates/precompiles/src/stablecoin_exchange/dispatch.rs
```diff
@@ -88,7 +90,13 @@ impl<'a, S: PrecompileStorageProvider> Precompile for StablecoinExchange<'a, S>
 
             IStablecoinExchange::createPairCall::SELECTOR => {
                 mutate::<IStablecoinExchange::createPairCall>(calldata, msg_sender, |_s, call| {
-                    self.create_pair(call.base)
+                    if self.storage.spec().is_allegretto() {
```
fgimenez: we could avoid decoding calldata post-allegretto (just to reject it) by using the `unknown_selector` helper, we would also get accurate gas accounting, instead of
```rust
    IStablecoinExchange::createPairCall::SELECTOR => {
      mutate::<IStablecoinExchange::createPairCall>(calldata, msg_sender, |_s, call| {
          if self.storage.spec().is_allegretto() {
              Err(TempoPrecompileError::UnknownFunctionSelector(
                  IStablecoinExchange::createPairCall::SELECTOR,
              ))
          } else {
              self.create_pair(call.base)
          }
      })
    }
```
we could have
```rust
    IStablecoinExchange::createPairCall::SELECTOR => {
      if self.storage.spec().is_allegretto() {
          unknown_selector(selector, self.storage.gas_used(), self.storage.spec())
      } else {
          mutate::<IStablecoinExchange::createPairCall>(calldata, msg_sender, |_s, call| {
              self.create_pair(call.base)
          })
      }
    }
```
sorry, can't suggest including unchanged lines, wdyt?

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -491,7 +491,12 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
 
         let book = self.sload_books(book_key)?;
         if book.base.is_zero() {
-            return Err(StablecoinExchangeError::pair_does_not_exist().into());
```
fgimenez, line 494: this logic is duplicated in `place` and `place_flip`, could we refactor?

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -4666,4 +4677,105 @@ mod tests {
...
+        base.transfer(
+            user,
+            ITIP20::transferCall {
+                to: exchange.address,
+                amount: U256::from(MIN_ORDER_AMOUNT),
+            },
+        )
+        .expect("Base token transfer failed");
+
+        // Place a order which should also create the pair
```
fgimenez:
```suggestion
        // Place an order which should also create the pair
```

crates/precompiles/src/stablecoin_exchange/dispatch.rs
0xKitsune: I updated to allow `createPair` calls post allegretto to make this change non breaking. If the pair does not exist when placing the first order, the book will still be initialized.

PR #986, opened 2025-11-21
feat(precompiles): clean storage api

### LINKED ISSUE #860: feat(precompiles): support individual struct field access

0xrusowsky:
## Motivation

closes #860 and achieves the final goal of #856 (pending precompile migration to the new API).

Improve the devex by introducing an API that feels like native Rust and, at the same time, closer to the usual Solidity patterns.

so if we had these types:
 ```rs
 #[derive(Storable)]
 pub struct MyStruct {
    foo: Address,
    bar: u64,
    bazz: U256,
 }

 #[contract]
 pub struct MyContract {
    my_uint: U256,                       // handler: Slot<U256>
    my_struct: MyStruct,                 // handler: MyStructHandler
    my_mapping: Mapping<Address, bool>,  // handler: Mapping<Address, bool>
 }
```

we would use them like:
```rs
 impl MyContract {
    pub fn do_something(&mut self, addr: Address) -> Result<()> {
       // handler for primitive types is `Slot<T>`
       let my_uint: Slot<U256> = self.my_uint;
       let old_value = my_uint.read()?;
       my_uint.write(old_value + U256::ONE)?;

       // custom structs have their own handler
       let my_struct: MyStructHandler = self.my_struct;
       let full_struct: MyStruct = my_struct.read()?; // loads the full struct
       let single_field: Slot<Address> = my_struct.foo;
       single_field.write(addr)?; // only writes to `MyStruct.foo`

       // mappings are their own handler, indexable using `at()`
       // nested mappings just return another Mapping
       let my_mapping: Slot<bool> = self.my_mapping.at(addr);
       my_mapping.write(true)?;
    }
 }
```

## Solution

 - Extend StorableType so that, besides the const LAYOUT, it also defines a type Handler, which is what the contract struct will hold.
 - Turn Slot into a unified proxy type for storage access, which implements StorageOps. From now on, Slot deals with all storage operations
 (packed and unpacked). Other handlers perform their storage access via Slot (it's the simplest possible handler).
 - Handlers hold a pointer (Rc<Address>) to the contract address. This is a deliberate memory tradeoff (8 bytes per handler) that enables direct field access without getter methods, avoiding the need to pass addresses around.
 - To avoid passing storage around, we define a thread-local storage instance (more details below).

### Thread-local Storage

To avoid passing storage references throughout the precompile call stack, we use thread-local storage via the scoped-tls crate in thread_local.rs:

**`StorageContext`**: A zero-sized type providing scoped access to a `PrecompileStorageProvider`. It exposes a single entry point `enter()` that takes a closure, establishing an explicit critical section:
```rs
 let mut storage = HashMapStorageProvider::new(1);
 StorageContext::enter(&mut storage, || {
     // All storage operations happen within this closure
     layout.block.write(data)?;
     layout.profile.read()?;
 });
 // Storage context automatically cleaned up when closure exits
```

key characteristics:
 - closure-scoped: Storage context lifetime bounded by the closure, making the critical section explicit
 - split mutability: mirrors `PrecompileStorageProvider` methods with split mutability semantics:
   - read operations (sload, tload, chain_id, etc.) take `&self`
   - write operations (sstore, tstore, emit_event, etc.) take `&mut self`
 - ee-entrancy protection: Uses `RefCell::borrow_mut()` internally, panicking on nested `StorageContext::enter()` calls
 - private access: internal `with_storage()` method is private, ensuring all storage access goes through public API

safety invariants:
the impl uses `unsafe` to transmute the storage pointer lifetime to 'static for the duration of the closure. This is sound if these invariants hold:

 | Invariant                                   | Enforcement                               | Risk             |
 |---------------------------------------------|-------------------------------------------|------------------|
 | single `enter()` per thread                   | `RefCell::borrow_mut()` panic               | pointer aliasing |
 | closure completes before storage dropped    | closure lifetime bounded by `enter()` scope | use-after-free   |
 | storage access only via `StorageContext`      | `with_storage()` is private                 | re-entrancy      |
 | parallel threads cannot share storage addrs | caller responsibility (not enforced)      | data races       |

the `scoped-tls` crate provides compile-time guarantees that the storage pointer is only accessible within the closure scope.

### Type (trait) system

the storage abstraction is built around these core traits:

| Trait            | Purpose                                                                                      |
 |------------------|----------------------------------------------------------------------------------------------|
 | `StorableType`   | compile-time layout info (`LAYOUT`, `SLOTS`, `BYTES`, `IS_PACKABLE`) + `Handler` type + factory method |
 | `Handler<T>`     | abstraction for `read()`, `write()`, `delete()` operations on storable types                 |
 | `Storable`       | low-level storage ops: `load()`, `store()`, `delete()` with `LayoutCtx` support              |
 | `Packable`       | infallible single-word conversion via `to_word()`/`from_word()` for packable primitives      |
 | `StorageKey`     | types that can be used as mapping keys (provides byte representation for hashing)            |

some clarifying notes:
  - `Storable` has a blanket impl for all `Packable` types, so primitives get storage operations automatically
  - `StorageKey` is now implemented for structs via the derive macro, enabling multi-word keys

## Handler APIs
```rs
 // Base handler for all storage operations. Other handlers delegate to `Slot`.
 struct Slot<T> {
     pub fn read(&self) -> Result<T>
     pub fn write(&mut self, value: T) -> Result<()>
     pub fn delete(&mut self) -> Result<()>
 }
```
```rs
 // Returns a handler for the given key.
 // For nested mappings (Mapping<K1, Mapping<K2, V>>), returns another Mapping.
 struct Mapping<K, V> {
     pub fn at(&self, key: K) -> <V as StorableType>::Handler
 }
```
```rs
 // Generated by `#[derive(Storable)]`, its fields are handlers.
 struct MyStructHandler {
     pub foo: Slot<Address>,
     pub bar: Mapping<u64, U256>,
 }
 impl Handler<MyStruct> for MyStructHandler {
     pub fn read(&self) -> Result<MyStruct>
     pub fn write(&mut self, value: MyStruct) -> Result<()>
     pub fn delete(&mut self) -> Result<()>
 }
```
```rs
 // Fixed-size array. Element access is checked (returns an option).
 struct ArrayHandler<T, N> {
     pub fn read(&self) -> Result<[T; N]>
     pub fn write(&mut self, value: [T; N]) -> Result<()>
     pub fn delete(&mut self) -> Result<()>
     pub fn at(&self, index: usize) -> Option<<T as StorableType>::Handler>
     pub const fn len(&self) -> usize
 }
```
```rs
 // Dynamic array with stack-like ops. Supports multi-slot element types.
 struct VecHandler<T> {
     pub fn read(&self) -> Result<Vec<T>>
     pub fn write(&mut self, value: Vec<T>) -> Result<()>
     pub fn delete(&mut self) -> Result<()>
     pub fn at(&self, index: usize) -> <T as StorableType>::Handler
     pub fn len(&self) -> Result<usize>
     pub fn pop(&self) -> Result<Option<T>>
 }
```


crates/precompiles/src/storage/thread_local.rs
```diff
@@ -0,0 +1,213 @@
...
+use tempo_chainspec::hardfork::TempoHardfork;
+
+use crate::{
+    error::{Result, TempoPrecompileError},
+    storage::PrecompileStorageProvider,
+};
+
+// Thread-local storage for accessing `PrecompileStorageProvider`
+thread_local! {
+    static STORAGE: Cell<Option<*mut dyn PrecompileStorageProvider>> = const { Cell::new(None) };
```
DaniPopes: why are we not using the [`scoped_tls`](https://docs.rs/scoped-tls/latest/scoped_tls/) crate? e.g. https://github.com/paradigmxyz/solar/blob/8ed9a10f1efe85cc06115a5889414faa63bc49a1/crates/interface/src/globals.rs

crates/precompiles/src/storage/types/vec.rs
```diff
@@ -42,6 +42,8 @@ impl<T> Encodable<1> for Vec<T>
 where
     T: Storable + MaybePackable,
 {
+    const VALIDATE_LAYOUT: () = assert!(Self::SLOTS == 1, "SLOTS must equal WORDS");
```
DaniPopes: this constant needs to be mentioned somewhere, anywhere, for it to run

crates/precompiles/src/storage/types/mod.rs
```diff
@@ -134,133 +143,159 @@ pub trait StorableType {
...
+
+    /// The handler type that provides storage access for this type.
+    ///
+    /// For primitives, this is `Slot<Self>`.
+    /// For mappings, this is `Self` (mappings are their own handlers).
+    /// For user-defined structs, this is a generated handler type (e.g., `MyStructHandler`).
+    type Handler;
+
+    /// Creates a handler for this type at the given storage location.
+    fn handle(slot: U256, ctx: LayoutCtx, address: Rc<Address>) -> Self::Handler;
```
DaniPopes: why is this Rc?

0xrusowsky: we wanted to reduce memsize and thought it was preferable to us `Rc` to only use 8bytes, but we could (maybe should) use `Address` and have it be 20 bytes

we may have midcurved. wdyt its best? small rutime overhead of rc and lower memsize or the other way around?

0xrusowsky: obviously we haven't run any benches yet

crates/precompiles/src/storage/thread_local.rs
```diff
@@ -0,0 +1,213 @@
...
+        let storage_ptr = STORAGE
+            .with(|s| s.get())
+            .ok_or(TempoPrecompileError::Fatal(
+                "No storage context. 'StorageGuard' must be initialized".to_string(),
+            ))?;
+
+        // SAFETY:
+        // - Caller must ensure NO recursive calls.
+        // - Type system ensures the storage pointer is valid.
+        let storage = unsafe { &mut *storage_ptr };
```
DaniPopes: i would recommend a separate function for better type safety since all you're doing is `&'a mut T -> &'b mut T`

crates/precompiles/src/storage/thread_local.rs
DaniPopes: i guess it wouldn't work with a guard but with a closure

crates/precompiles/src/storage/types/mod.rs
DaniPopes: it really doesn't matter, just use address

crates/precompiles/src/storage/types/vec.rs
0xrusowsky: not relevant anymore after we removed `MaybePackable` and `Encodable<N>` infavor of `Packable`

crates/precompiles/src/storage/types/mod.rs
0xrusowsky: addressed in [`1579306` (#986)](https://github.com/tempoxyz/tempo/pull/986/commits/157930680edd441245759751f922e84b06b38483)

crates/precompiles/src/storage/thread_local.rs
0xrusowsky: addressed in [`7e301b0` (#986)](https://github.com/tempoxyz/tempo/pull/986/commits/7e301b0e0caddb089eae4e37e2a1828fac19bdb6)

crates/precompiles/src/storage/thread_local.rs
0xrusowsky: addressed in [`7e301b0` (#986)](https://github.com/tempoxyz/tempo/pull/986/commits/7e301b0e0caddb089eae4e37e2a1828fac19bdb6)

crates/precompiles/src/storage/thread_local.rs
```diff
@@ -0,0 +1,216 @@
...
+use crate::{
+    error::{Result, TempoPrecompileError},
+    storage::PrecompileStorageProvider,
+};
+
+scoped_thread_local!(static STORAGE: RefCell<*mut dyn PrecompileStorageProvider>);
+
+/// Converts a raw pointer to a mutable reference.
+#[inline]
+unsafe fn ptr_as_mut<'a>(
```
DaniPopes: this doesn't do much, just do it inline

crates/precompiles/src/storage/thread_local.rs
```diff
@@ -0,0 +1,216 @@
...
+
+    pub fn get_account_info(&self, address: Address) -> Result<&'_ AccountInfo> {
+        // SAFETY: The returned reference is valid for the duration of the
+        // `StorageContext::enter` closure. Since `StorageContext` can only be used
+        // while inside enter(), the reference remains valid.
+        Self::with_storage(|s| {
+            let info = s.get_account_info(address)?;
+            // SAFETY: Extend the lifetime to match `&'_ self`.
+            // Underlying storage data outlives the accessor.
+            let info: &'_ AccountInfo = unsafe { &*(info as *const AccountInfo) };
```
DaniPopes: this should be a util `&'a T -> &'b T` for more type safety

crates/precompiles/src/storage/thread_local.rs
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    #[cfg(any(test, feature = "test-utils"))]
+    pub fn get_events(&self, address: Address) -> &Vec<LogData> {
+        // SAFETY: The returned reference is valid for the duration of the
+        // `StorageContext::enter` closure. Since `StorageContext` can only be used
+        // while inside enter(), the reference remains valid.
+        Self::with_storage(|s| {
+            let events = s.get_events(address);
+            let events: &'_ Vec<LogData> = unsafe { &*(events as *const Vec<LogData>) };
```
DaniPopes: same here

crates/precompiles/src/storage/thread_local.rs
```diff
@@ -0,0 +1,216 @@
...
+        .unwrap()
+    }
+
+    #[cfg(any(test, feature = "test-utils"))]
+    pub fn set_nonce(&mut self, address: Address, nonce: u64) {
+        Self::with_storage(|s| {
+            s.set_nonce(address, nonce);
+            Ok(())
+        })
+        .unwrap()
```
DaniPopes: you could have `with_storage() -> T` and `try_with_storage() -> Result` to remove these `Ok(())+unwrap`s

DaniPopes (commented):
nits

crates/precompiles/src/storage/thread_local.rs
0xrusowsky: addressed in: [`1f4e9f6` (#986)](https://github.com/tempoxyz/tempo/pull/986/commits/1f4e9f6fa9ff7288b7415a085ea882b1d1c9f3f3)

crates/precompiles/src/storage/thread_local.rs
0xrusowsky: addressed in: [`1f4e9f6` (#986)](https://github.com/tempoxyz/tempo/pull/986/commits/1f4e9f6fa9ff7288b7415a085ea882b1d1c9f3f3)

crates/precompiles/src/storage/thread_local.rs
0xrusowsky: addressed in: [`1f4e9f6` (#986)](https://github.com/tempoxyz/tempo/pull/986/commits/1f4e9f6fa9ff7288b7415a085ea882b1d1c9f3f3)

crates/precompiles/src/storage/thread_local.rs
0xrusowsky: addressed in: [`1f4e9f6` (#986)](https://github.com/tempoxyz/tempo/pull/986/commits/1f4e9f6fa9ff7288b7415a085ea882b1d1c9f3f3)

crates/precompiles-macros/src/storable.rs
```diff
@@ -124,21 +138,18 @@ pub(crate) fn derive_impl(input: DeriveInput) -> syn::Result<TokenStream> {
...
-                use crate::storage::Storable;
-
-                #to_evm_words_impl
-            }
-
-            fn from_evm_words(words: [::alloy::primitives::U256; { #mod_ident::SLOT_COUNT }]) -> crate::error::Result<Self> {
-                use crate::storage::Storable;
-
-                #from_evm_words_impl
+            // delete uses default implementation from trait
```
klkvr: is this right? wouldn't this miss proper cleanup of e.g `Vec` field?

crates/precompiles-macros/src/storable.rs
```diff
@@ -124,21 +138,18 @@ pub(crate) fn derive_impl(input: DeriveInput) -> syn::Result<TokenStream> {
...
+        // `StorageKey` implementation: byte encoding for mapping keys
+        impl #impl_generics crate::storage::StorageKey for #strukt #ty_generics #where_clause {
+            fn as_storage_bytes(&self) -> impl AsRef<[u8]> {
+                use crate::storage::StorageKey;
+                let mut bytes = Vec::with_capacity(#mod_ident::SLOT_COUNT * 32);
+                #(
+                    bytes.extend_from_slice(self.#direct_names.as_storage_bytes().as_ref());
+                )*
+                bytes
             }
```
klkvr, line 165: imo we can omit this impl unless it's used anywhere

crates/precompiles-macros/src/storable_primitives.rs
```diff
@@ -66,274 +71,103 @@ fn gen_storage_key_impl(type_path: &TokenStream, strategy: &StorageKeyStrategy)
...
-                        Ok([U256::from(*self)])
+                    fn to_word(&self) -> U256 {
+                        U256::from(*self)
                     }
 
                     #[inline]
-                    fn from_evm_words(words: [U256; 1]) -> Result<Self> {
-                        Ok(words[0].to::<Self>())
+                    fn from_word(word: U256) -> Self {
+                        word.to::<Self>()
```
klkvr: this might panic which I don't think is desired

crates/precompiles-macros/src/storable_primitives.rs
```diff
@@ -66,274 +71,103 @@ fn gen_storage_key_impl(type_path: &TokenStream, strategy: &StorageKeyStrategy)
...
                     #[inline]
-                    fn to_evm_words(&self) -> Result<[U256; 1]> {
-                        Ok([U256::from(*self as #unsigned_type)])
-                    }
-
-                    #[inline]
-                    fn from_evm_words(words: [U256; 1]) -> Result<Self> {
-                        Ok(words[0].to::<#unsigned_type>() as Self)
+                    fn from_word(word: U256) -> Self {
+                        word.to::<#unsigned_type>() as Self
```
klkvr: does this properly handle all of the two's complement stuff?

crates/precompiles/src/storage/types/slot.rs
```diff
@@ -24,42 +26,58 @@ use crate::{
...
+    ctx: LayoutCtx,
+    address: Address,
     _ty: PhantomData<T>,
 }
 
 impl<T> Default for Slot<T> {
     fn default() -> Self {
-        Self::new(U256::ZERO)
+        Self::new(U256::ZERO, Address::ZERO)
     }
```
klkvr: this seems dangerous, what do we need it for?

crates/precompiles/src/storage/thread_local.rs
```diff
@@ -0,0 +1,205 @@
...
+    pub fn get_account_info(&self, address: Address) -> Result<&'_ AccountInfo> {
+        // SAFETY: The returned reference is valid for the duration of the
+        // `StorageContext::enter` closure. Since `StorageContext` can only be used
+        // while inside enter(), the reference remains valid.
+        Self::try_with_storage(|s| {
+            let info = s.get_account_info(address)?;
+            // SAFETY: Underlying storage data outlives the accessor.
+            Ok(unsafe { extend_lifetime(info) })
+        })
+    }
```
klkvr: consumers of this are quite limited so we can maybe avoid this by adding more precise getters like `has_nonce`/`has_code` or just clonning `AccountInfo` which is very cheap

another solution is `fn inspect_account_info(&self, f: impl FnOnce(&AccountInfo))` I guess

crates/precompiles/src/storage/thread_local.rs
```diff
@@ -0,0 +1,205 @@
...
+        if !STORAGE.is_set() {
+            return Err(TempoPrecompileError::Fatal(
+                "No storage context. 'StorageContext::enter' must be called first".to_string(),
+            ));
+        }
+        STORAGE.with(|cell| {
+            // SAFETY: `scoped_tls` ensures the pointer is only accessible within the closure scope.
+            // Holding the guard prevents re-entrant borrows.
+            let mut guard = cell.borrow_mut();
+            f(unsafe { &mut **guard })
```
klkvr: now that we're anyway ending up doing `RefCell::borrow_mut` I'm wondering if we could just pass an `Rc<RefCell<*mut dyn PrecompileStorageProvider>>` around

crates/precompiles/src/storage/thread_local.rs
```diff
@@ -0,0 +1,205 @@
...
+use scoped_tls::scoped_thread_local;
+use std::cell::RefCell;
+use tempo_chainspec::hardfork::TempoHardfork;
+
+use crate::{
+    error::{Result, TempoPrecompileError},
+    storage::PrecompileStorageProvider,
+};
+
+scoped_thread_local!(static STORAGE: RefCell<*mut dyn PrecompileStorageProvider>);
```
klkvr: would it be sound to change this to `RefCell<&'static mut dyn PrecompileStorageProvider>` given that we're anyway casting to `'static`?

crates/precompiles/src/storage/mod.rs
```diff
@@ -1,100 +1,118 @@
...
+        unimplemented!("set_beneficiary only available for test storage providers")
+    }
+    #[cfg(any(test, feature = "test-utils"))]
+    fn set_spec(&mut self, _spec: TempoHardfork) {
+        unimplemented!("set_spec only available for test storage providers")
+    }
+    #[cfg(any(test, feature = "test-utils"))]
+    fn clear_transient(&mut self) {
+        unimplemented!("clear_transient only available for test storage providers")
+    }
```
klkvr: makes me wonder if instead of this we should've just added a `StorageContext::cast_to_hashmap` or smth

klkvr (commented):
nice, this mostly looks good to me

ideally we need to run the re-execute tests with this ASAP just to confirm that all of the new logic is correct

crates/precompiles-macros/src/storable.rs
0xrusowsky: initially added it cause the tip fee manager uses `TokenPair` as a mapping key.

as discussed in dms, it's better to impl it manually when (rarely) needed

crates/precompiles/src/storage/thread_local.rs
0xrusowsky: agreed, let's use the precise getters

crates/precompiles/src/storage/thread_local.rs
0xrusowsky: personally i prefer using the tls approach as it's way more ergonomic, and we all agree that it is sound in this case

crates/precompiles/src/storage/types/slot.rs
0xrusowsky: good call, it's not needed

crates/precompiles/src/storage/thread_local.rs
0xrusowsky: i realized that the precise getters could have an impact in gas usage, so i used teh closure approach instead

crates/precompiles/src/storage/mod.rs
0xrusowsky: addressed in [`f30e85c` (#986)](https://github.com/tempoxyz/tempo/pull/986/commits/f30e85cc0e23376bffc30adf392f3674c704a022)

crates/precompiles/src/storage/thread_local.rs
0xrusowsky: addressed in addressed in [`f30e85c` (#986)](https://github.com/tempoxyz/tempo/pull/986/commits/f30e85cc0e23376bffc30adf392f3674c704a022)

crates/precompiles/src/storage/thread_local.rs
0xrusowsky: addressed in [`f30e85c` (#986)](https://github.com/tempoxyz/tempo/pull/986/commits/f30e85cc0e23376bffc30adf392f3674c704a022)

crates/precompiles-macros/src/storable_primitives.rs
0xrusowsky: yes, i think so

added tests in [`859f60f` (#986)](https://github.com/tempoxyz/tempo/pull/986/commits/859f60fb9a500cf5ff990cadf9643614eade7183) (note that since we already test roundrip conversions we don't need to check both methods)

crates/precompiles-macros/src/storable_primitives.rs
0xrusowsky: made fallible in [`80fcb12` (#986)](https://github.com/tempoxyz/tempo/pull/986/commits/80fcb129490238c525d4935b4e8bcda29a6d3f95) (note that i committed it together with the `Storable` macro fix for `fn delete()` by mistake)

crates/precompiles-macros/src/storable.rs
0xrusowsky: addressed in [`80fcb12` (#986)](https://github.com/tempoxyz/tempo/pull/986/commits/80fcb129490238c525d4935b4e8bcda29a6d3f95)

crates/precompiles/src/storage/types/slot.rs
0xrusowsky: dropped in [`69331a3` (#986)](https://github.com/tempoxyz/tempo/pull/986/commits/69331a3473fc07c7f73f77c822045d88cb6e5500)

PR #999, opened 2025-11-21
feat(consensus): include validator addresses in PublicOutcome for peer bootstrapping

### LINKED ISSUE #883: consensus: write full peer set for an epoch to chain

fgimenez:
Closes #883 

- Replace `participants: Ordered<PublicKey>` with `validator_state: ValidatorState` in `PublicOutcome`
- Move `DecodedValidator` and `ValidatorState` from commonware-node to dkg-onchain-artifacts crate
- Add `resolve_addresses_and_merge_peers()` method to extract peer addresses from chain data
- Add e2e test verifying validator addresses can be decoded from `PublicOutcome` in block extra_data


crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -402,10 +401,21 @@ where
...
+        let validator_state = self
+            .validators_metadata
+            .get(&dkg_outcome.epoch.into())
+            .cloned()
+            .unwrap_or_else(|| {
+                warn!(
+                    epoch = dkg_outcome.epoch,
+                    "no validator state for epoch, using current"
+                );
+                self.current_epoch_state().validator_state.clone()
```
joshieDo: hmm, is this really equivalent  ?

fgimenez: absolutely great catch, those were from the wrong state. now during ceremony finalization, we read validators from the latest block, compute the next epoch's `validator_state` (applying the rotation), and store it in `validators_metadata[epoch+1]`, so `handle_get_outcome` always finds the correct state.

i've also updated the existing `validator_is_added` e2e test with assertions around this, ptal

PR #1001, opened 2025-11-21
fix: validate blocktime is not in the future

klkvr:
Closes https://github.com/tempoxyz/tempo/issues/1010

Changes in the PR:
- Bump reth
- Add check to `TempoConsensus` to make sure that no blocks are in the future further than 10s from our timestamp
- Add `TempoPayloadAttributes` to allow propagating millisecond timestamp part in devmode


crates/consensus/src/lib.rs
```diff
@@ -51,6 +54,18 @@ impl HeaderValidator<TempoHeader> for TempoConsensus {
     fn validate_header(&self, header: &SealedHeader<TempoHeader>) -> Result<(), ConsensusError> {
         self.inner.validate_header(header)?;
 
+        let present_timestamp = std::time::SystemTime::now()
+            .duration_since(std::time::SystemTime::UNIX_EPOCH)
+            .unwrap()
```
SuperFluffy: This is not necessarily obvious to everyone. Let's document this:
```suggestion
            .expect("system time should never be before UNIX EPOCH")
```

crates/consensus/src/lib.rs
```diff
@@ -30,6 +30,9 @@ const END_OF_BLOCK_SYSTEM_TX_ADDRESSES: [Address; END_OF_BLOCK_SYSTEM_TX_COUNT]
     Address::ZERO,
 ];
 
+/// How far in the future the block timestamp can be.
+pub const ALLOWED_FUTURE_BLOCK_TIME_SECONDS: u64 = 10;
```
SuperFluffy:
```suggestion
pub const ALLOWED_FUTURE_BLOCK_TIME: Duration = Duration::from_secs(10);
```

crates/consensus/src/lib.rs
```diff
@@ -51,6 +54,18 @@ impl HeaderValidator<TempoHeader> for TempoConsensus {
     fn validate_header(&self, header: &SealedHeader<TempoHeader>) -> Result<(), ConsensusError> {
         self.inner.validate_header(header)?;
 
+        let present_timestamp = std::time::SystemTime::now()
+            .duration_since(std::time::SystemTime::UNIX_EPOCH)
+            .expect("system time should never be before UNIX EPOCH")
+            .as_secs();
```
SuperFluffy, line 60: Can get rid of this if you use the const fn for `Duration::from_secs`

mattsse, line 60: this only allows up to 0.9... s local clock drift, which I think should be okay with NTP.

but we could even go higher a bit

crates/consensus/src/lib.rs
```diff
@@ -30,6 +30,9 @@ const END_OF_BLOCK_SYSTEM_TX_ADDRESSES: [Address; END_OF_BLOCK_SYSTEM_TX_COUNT]
     Address::ZERO,
 ];
 
+/// How far in the future the block timestamp can be.
+pub const ALLOWED_FUTURE_BLOCK_TIME_SECONDS: u64 = 1;
```
mattsse: is this based on anything?

PR #1006, opened 2025-11-21
feat: add new tx pool for 2d nonces 

legion2002:
We use a hybrid approach, which aims to not interfere at all with the battle hardened reth tx pool. So txs that use normal nonces just go directly to the reth pool.

If the transaction type is AA and it has a nonce_key != 0, this means tha the tx uses a 2d nonce, and so it is routed to the simpler `Pool2d`.

The TempoPool is just a routing wrapper struct on top of the 2 pools, it implements the `TransactionPool` trait from reth. 

changes in the precompile contract are tracked and maintained externally, mapping back to addresses

### DIFF
--- a/crates/node/tests/it/account_abstraction.rs
+++ b/crates/node/tests/it/account_abstraction.rs
@@ -1085,7 +1085,7 @@ async fn test_aa_2d_nonce_pool_comprehensive() -> eyre::Result<()> {
     for tx_hash in &sent {
         // Assert that transactions were removed from the pool and included in the block
         assert!(block1_txs.iter().any(|tx| tx.tx_hash() == tx_hash));
-        assert!(!setup.node.inner.pool.contains(&tx_hash));
+        assert!(!setup.node.inner.pool.contains(tx_hash));
     }
     println!("  ✓ All 3 transactions from different pools included in block");
 
@@ -1195,7 +1195,7 @@ async fn test_aa_2d_nonce_pool_comprehensive() -> eyre::Result<()> {
 
     for tx_hash in &sent {
         // Assert that transactions were removed from the pool
-        assert!(!setup.node.inner.pool.contains(&tx_hash));
+        assert!(!setup.node.inner.pool.contains(tx_hash));
     }
 
     // ===========================================================================


crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -0,0 +1,822 @@
...
+/// Minimal 2D nonce pool for user nonces only
+mod pool_2d;
+use pool_2d::Pool2D;
+
+/// Tempo transaction pool that routes based on nonce_key
+pub struct TempoTransactionPool<Client, S>
+where
+    Client: StateProviderFactory,
+    S: BlobStore,
+{
```
klkvr: I don't think we need this to be generic over `BlobStore`, can just hardcode `DiskFileBlobStore` generic for the inner pool

crates/transaction-pool/src/tempo_pool/pool_2d.rs
```diff
@@ -0,0 +1,880 @@
...
+                    InvalidTransactionError::TxTypeNotSupported,
+                )),
+            })?;
+
+        // Storage returns Option<U256>, unwrap to U256 then convert to u64
+        Ok(nonce_value.unwrap_or_default().to::<u64>())
+    }
+
+    /// Add a transaction to the pool
+    pub(super) async fn add_transaction(
```
klkvr: does this not perform any validations besides the nonce check? we want all transactions to go through `TempoTransactionValidator`

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -0,0 +1,822 @@
...
+        transactions: Vec<Self::Transaction>,
+    ) -> Vec<PoolResult<AddedTransactionOutcome>> {
+        let mut results = Vec::with_capacity(transactions.len());
+
+        for tx in transactions {
+            results.push(self.add_transaction(origin, tx).await);
+        }
+
+        results
+    }
```
klkvr, line 281: we likely want to divide txs into 2 separate sets here and invoke `add_transactions` for 2 pools in parallel

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -0,0 +1,822 @@
...
+
+    fn pending_transactions_listener_for(
+        &self,
+        kind: reth_transaction_pool::TransactionListenerKind,
+    ) -> tokio::sync::mpsc::Receiver<B256> {
+        // Delegate to protocol pool for now
+        // TODO: Merge events from both pools
+        self.protocol_pool.pending_transactions_listener_for(kind)
+    }
```
klkvr, line 288: ah this is a bit tricky but I guess we can spawn an extra task that merges events from 2 receivers here

another option is to add a helper to reth pool to return both sender and receiver and just re-use the sender

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -0,0 +1,822 @@
...
+        // TODO: Check both pools
+        self.protocol_pool.get_pooled_transaction_element(tx_hash)
+    }
+
+    fn best_transactions(
+        &self,
+    ) -> Box<dyn BestTransactions<Item = Arc<ValidPoolTransaction<Self::Transaction>>>> {
+        let protocol_iter = Box::new(self.protocol_pool.best_transactions())
+            as Box<dyn Iterator<Item = Arc<ValidPoolTransaction<TempoPooledTransaction>>> + Send>;
+        let user_iter = Box::new(self.user_nonce_pool.best_transactions())
```
klkvr: do we need the extra `Box::new` here? I think inner pools already return boxed `BestTransactions`

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -0,0 +1,822 @@
...
+    ) -> Vec<Arc<ValidPoolTransaction<Self::Transaction>>> {
+        let mut txs = self.protocol_pool.get_queued_transactions_by_sender(sender);
+        txs.extend(self.user_nonce_pool.get_queued_for_address(&sender));
+        txs
+    }
+
+    fn get_highest_transaction_by_sender(
+        &self,
+        sender: Address,
+    ) -> Option<Arc<ValidPoolTransaction<Self::Transaction>>> {
```
klkvr, line 570: hmm this kind of doesn't fit here because with 2d nonces there's no really a concept of "highest" nonce

maybe we should just assume highest protocol nonce here and return the result from regular pool

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -0,0 +1,822 @@
...
+        if let Some(tx) = self
+            .protocol_pool
+            .get_transaction_by_sender_and_nonce(sender, nonce)
+        {
+            return Some(tx);
+        }
+
+        // Check 2D pool for any nonce_key
+        self.user_nonce_pool.get_by_sender_and_nonce(&sender, nonce)
+    }
```
klkvr, line 594: 2d nonces also kind of don't fit here and a bit unclear what's the expected response for 2d nonces, this one is actually exposed on RPC side

crates/transaction-pool/src/tempo_pool.rs
mattsse: yeah imo we should either relax this on the reth trait or add a way to trigger emits on the vanilla pool

mattsse: https://github.com/paradigmxyz/reth/pull/19918

crates/transaction-pool/src/tempo_pool/pool_2d.rs
```diff
@@ -0,0 +1,880 @@
...
+            removed.push(tx);
+        }
+        removed
+    }
+}
+
+/// Minimal 2D nonce pool
+pub(super) struct Pool2D<S: StateProviderFactory> {
+    /// State provider for nonce queries
+    state: S,
```
mattsse: we shouldn't use state access here and instead do all of this in the validator itself, so that this type never has to do any IO directly

crates/transaction-pool/src/tempo_pool/pool_2d.rs
```diff
@@ -0,0 +1,880 @@
...
+            state,
+            by_sender: Arc::new(RwLock::new(HashMap::new())),
+            by_hash: Arc::new(RwLock::new(HashMap::new())),
+            address_to_keys: Arc::new(RwLock::new(HashMap::new())),
+            tx_origin: Arc::new(RwLock::new(HashMap::new())),
+        }
+    }
+
+    /// Get the current nonce from state for (address, nonce_key)
+    fn get_state_nonce(&self, address: Address, nonce_key: U256) -> Result<u64, PoolError> {
```
mattsse: this we need to move to tempotxvalidator so that for 2d txs we fetch this and provide this as part of the tempopooledtx

crates/transaction-pool/src/tempo_pool.rs
legion2002: Yup, we'll just return from the regular pool

crates/transaction-pool/src/tempo_pool.rs
legion2002: We'll just return protocol nonce for this now

crates/payload/builder/src/lib.rs
```diff
@@ -57,17 +57,14 @@ use tempo_primitives::{
...
+use tempo_transaction_pool::transaction::{TempoPoolTransactionError, TempoPooledTransaction};
 use tracing::{Level, debug, error, info, instrument, trace, warn};
 
 use crate::metrics::TempoPayloadBuilderMetrics;
 
 #[derive(Debug, Clone)]
-pub struct TempoPayloadBuilder<Provider> {
-    pool: TempoTransactionPool<Provider>,
+pub struct TempoPayloadBuilder<Provider, Pool> {
+    pool: Pool,
```
klkvr: this can be kept as is I think?

crates/transaction-pool/src/validator.rs
```diff
@@ -1,29 +1,62 @@
...
+    fn get_2d_nonce(
+        &self,
+        state_provider: &impl StateProvider,
+        address: alloy_primitives::Address,
+        nonce_key: U256,
+    ) -> ProviderResult<(U256, u64)> {
+        // Compute storage slot for 2D nonce
+        // Based on: mapping(address => mapping(uint256 => uint64)) at slot 0
+        let outer_slot = mapping_slot(address.as_slice(), U256::ZERO);
+        let slot = mapping_slot(nonce_key.as_le_bytes(), outer_slot);
```
klkvr: I think we have `double_mapping_slot` for this

crates/transaction-pool/src/validator.rs
```diff
@@ -50,17 +83,6 @@ where
...
-        // a nice way to track them in the pool yet.
-        if let Some(tx) = transaction.inner().as_aa()
-            && !tx.tx().nonce_key.is_zero()
-        {
-            return TransactionValidationOutcome::Error(
-                *transaction.hash(),
-                InvalidTransactionError::TxTypeNotSupported.into(),
-            );
-        }
-
```
klkvr, line 165: I think we need to also reject txs with subblock-specific nonce keys here

crates/transaction-pool/src/aa_2d_pool.rs
```diff
@@ -0,0 +1,2371 @@
...
+    ///
+    /// NOTE: The range is _inclusive_: if the transaction that belongs to `id` it field be the
+    /// first value.
+    fn descendant_txs_mut<'a, 'b: 'a>(
+        &'a mut self,
+        id: &'b AA2dTransactionId,
+    ) -> impl Iterator<Item = (&'a AA2dTransactionId, &'a mut AA2dInternalTransaction)> + 'a {
+        self.by_id
+            .range_mut(id..)
+            .take_while(|(other, _)| id.sender == other.sender)
```
klkvr: wasn't immeditealy obvious that `id.sender` here is not just sender but also a nonce id

maybe we can name this to `id.sender_and_nonce` or smth? or `seq_id`

crates/transaction-pool/src/aa_2d_pool.rs
```diff
@@ -0,0 +1,2371 @@
...
+                discarded: self.discard(),
+            }));
+        }
+
+        Ok(AddedTransaction::Parked {
+            transaction,
+            replaced: replaced.map(|tx| tx.inner.transaction),
+            subpool: SubPool::Queued,
+            queued_reason: Some(QueuedReason::NonceGap),
+        })
```
klkvr, line 238: hmm looks like the loop above might promote some transactions but we will never return them here

probably requires a race with maintenance task but still seems possible

crates/transaction-pool/src/aa_2d_pool.rs
```diff
@@ -0,0 +1,2371 @@
...
+
+    fn remove_transaction_by_id(
+        &mut self,
+        id: &AA2dTransactionId,
+    ) -> Option<Arc<ValidPoolTransaction<TempoPooledTransaction>>> {
+        let tx = self.by_id.remove(id)?;
+        self.independent_transactions.remove(id);
+        self.by_hash.remove(tx.inner.transaction.hash());
+        Some(tx.inner.transaction)
+    }
```
klkvr, line 406: should this also unset `is_pending` for descendant transactions? do we even do this in reth?

crates/transaction-pool/src/aa_2d_pool.rs
```diff
@@ -0,0 +1,2371 @@
...
+
+        // scan all the transactions with the same nonce key starting with the on chain nonce
+        for (existing_id, existing_tx) in self.descendant_txs_mut(&on_chain_id) {
+            let mut is_promoted = false;
+            if existing_id.nonce == next_nonce {
+                // if this was previously not pending we need to promote the transaction
+                is_promoted = !std::mem::replace(&mut existing_tx.is_pending, true);
+                next_nonce = existing_id.nonce.saturating_add(1);
+            } else {
+                existing_tx.is_pending = false;
```
klkvr: should we just break here? I guess if we've found a gap it means that everything further is already non-pending, and our new tx should not be either

crates/transaction-pool/src/aa_2d_pool.rs
```diff
@@ -0,0 +1,2371 @@
...
+                if existing_id.nonce == next_nonce {
+                    // Promote if transaction was previously queued (not pending)
+                    if !std::mem::replace(&mut existing_tx.is_pending, true) {
+                        promoted.push(existing_tx.inner.transaction.clone());
+                    }
+
+                    if existing_id.nonce == on_chain_nonce {
+                        // if this is the on chain nonce we can mark it as the next independent transaction
+                        self.independent_transactions
+                            .insert(*existing_id, existing_tx.inner.clone());
```
klkvr: I'm wondering if those should be keyed by `AASenderId` just so we definitely can't have multiple txs per sender in there

crates/transaction-pool/src/aa_2d_pool.rs
```diff
@@ -0,0 +1,2371 @@
...
+                    }
+
+                    next_nonce = next_nonce.saturating_add(1);
+                } else {
+                    // Gap detected - remaining transactions should not be pending
+                    existing_tx.is_pending = false;
```
klkvr, line 586: I think we can break here too?

crates/transaction-pool/src/maintain.rs
```diff
@@ -0,0 +1,54 @@
...
+            .collect::<Vec<_>>();
+        if changed_slots.is_empty() {
+            continue;
+        }
+
+        // we can now map the slots back to addresses and nonce keys and then update the pool
+        let updates = nonce_keys.update_tracked(changed_slots);
+        trace!(target: "txpool::2d",  ?updates, "update tracked 2d nonce changes");
+
+        pool.on_aa_2d_nonce_changes(updates);
```
klkvr, line 52: nice

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -0,0 +1,716 @@
...
+    ) -> PoolResult<TransactionEvents> {
+        let tx = self
+            .protocol_pool
+            .validator()
+            .validate_transaction(origin, transaction)
+            .await;
+        let res = self.add_validated_transaction(origin, tx)?;
+        self.transaction_event_listener(res.hash)
+            .ok_or_else(|| PoolError::new(res.hash, PoolErrorKind::DiscardedOnInsert))
+    }
```
klkvr, line 241: this should have default impl in reth I guess

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -0,0 +1,716 @@
...
+
+    fn get_pending_transactions_with_predicate(
+        &self,
+        predicate: impl FnMut(&ValidPoolTransaction<Self::Transaction>) -> bool,
+    ) -> Vec<Arc<ValidPoolTransaction<Self::Transaction>>> {
+        // TODO: support 2d pool
+        self.protocol_pool
+            .get_pending_transactions_with_predicate(predicate)
+    }
```
klkvr, line 541: unused in reth so I guess can skip for now?

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -0,0 +1,716 @@
...
+
+        txs
+    }
+
+    fn get_queued_transactions_by_sender(
+        &self,
+        sender: Address,
+    ) -> Vec<Arc<ValidPoolTransaction<Self::Transaction>>> {
+        self.protocol_pool.get_queued_transactions_by_sender(sender)
+    }
```
klkvr, line 565: same here?

crates/transaction-pool/src/aa_2d_pool.rs
mattsse: > multiple txs per sender in there

isn't this the entire point of the 2d nonces?

crates/transaction-pool/src/aa_2d_pool.rs
mattsse: yeah we can because the remaining once will be updated either if mined or the missing tx is filled

crates/transaction-pool/src/aa_2d_pool.rs
mattsse: yep that does make sense

crates/transaction-pool/src/aa_2d_pool.rs
mattsse: good point, we do it in the pending pool only

crates/transaction-pool/src/aa_2d_pool.rs
mattsse: this doesn't include any promoted, because if a tx was parked, this means it is not pending, so it also couldnt promote anything

klkvr: yeah but we might promote things just because `on_chain_nonce` got higher

i.e if on-chain nonce was 0 and pool had a queued tx with nonce 1

then some tx with nonce 0 was mined and at the same time tx with nonce 3 was submitted

then we will end up promoting nonce 1 to pending but parking nonce 3

crates/transaction-pool/src/validator.rs
```diff
@@ -1,32 +1,67 @@
...
+    /// Returns `(slot, nonce)` for the given address and nonce key.
+    fn get_2d_nonce(
+        &self,
+        state_provider: &impl StateProvider,
+        address: alloy_primitives::Address,
+        nonce_key: U256,
+    ) -> ProviderResult<(U256, u64)> {
+        // Compute storage slot for 2D nonce
+        // Based on: mapping(address => mapping(uint256 => uint64)) at slot 0
+        let slot = double_mapping_slot(address.as_slice(), nonce_key.as_le_bytes(), U256::ZERO);
```
klkvr: shouldn't it be 
```suggestion
        let slot = double_mapping_slot(address.as_slice(), nonce_key.to_be_bytes::<32>(), U256::ZERO);
```
?

do we have test coverage for this?

crates/transaction-pool/src/best.rs
```diff
@@ -0,0 +1,139 @@
...
+            // Only left has an item - take it
+            (Some(_), None) => {
+                let (item, priority) = self.next_left.take()?;
+                Some((item, priority))
+            }
+            // Only right has an item - take it
+            (None, Some(_)) => {
+                let (item, priority) = self.next_right.take()?;
+                Some((item, priority))
+            }
```
klkvr, line 89:
```suggestion
            // Only one item available - take it
            (Some(item), None) | (None, Some(item)) => {
                let (item, priority) = self.next_left.take()?;
                Some((item, priority))
            }
```

crates/transaction-pool/src/best.rs
```diff
@@ -0,0 +1,139 @@
...
+
+impl<L, R, T> BestTransactions for MergeBestTransactions<L, R, T>
+where
+    L: BestPriorityTransactions<T, Item: Send> + Send,
+    R: BestPriorityTransactions<T, Item = L::Item> + Send,
+    T: TransactionOrdering,
+{
+    fn mark_invalid(&mut self, transaction: &Self::Item, kind: &InvalidPoolTransactionError) {
+        self.left.mark_invalid(transaction, kind);
+        self.right.mark_invalid(transaction, kind);
```
klkvr, line 126: I'm wondering if this could be problematic in cases when we receive invalid tx from 2d pool which is getting marked invalid in the protocol iter as well, thus skipping all protocol pool transactions that might be OK for the sender wrt protocol nonce

crates/transaction-pool/src/validator.rs
mattsse: tbh idk

crates/transaction-pool/src/best.rs
mattsse: yeah probably

I guess we need a specialized impl after all

crates/transaction-pool/src/best.rs
```diff
@@ -0,0 +1,138 @@
...
+        }
+    }
+}
+
+impl<L, R, T> MergeBestTransactions<L, R, T>
+where
+    L: BestPriorityTransactions<T>,
+    R: BestPriorityTransactions<T, Item = L::Item>,
+    T: TransactionOrdering,
+{
```
legion2002, line 65:
```suggestion

```
can be merged.

crates/transaction-pool/src/transaction.rs
```diff
@@ -157,6 +196,17 @@ impl PoolTransaction for TempoPooledTransaction {
     fn encoded_length(&self) -> usize {
         self.inner.encoded_length
     }
+
+    fn requires_nonce_check(&self) -> bool {
```
legion2002, line 200: nice, this solves the validation duplication problem

crates/transaction-pool/src/validator.rs
```diff
@@ -223,7 +251,76 @@ where
...
+                propagate,
+                authorities,
+            } => {
+                // Additional 2D nonce validations
+                // Check for 2D nonce validation (nonce_key > 0)
+                if let Some(nonce_key) = transaction.transaction().nonce_key()
+                    && !nonce_key.is_zero()
+                {
+                    // ensure the nonce key isn't prefixed with the sub-block prefix
+                    if has_sub_block_nonce_key_prefix(&nonce_key) {
```
legion2002, line 269: How are subblock txs meant to be sent to a validator? 
Shouldn't the validator allow the nonces for their own address here?

mattsse: they are gossiped over CL p2p I believe

crates/transaction-pool/src/transaction.rs
```diff
@@ -78,6 +112,10 @@ pub enum TempoPoolTransactionError {
         "Native transfers are not supported, if you were trying to transfer a stablecoin, please call TIP20::Transfer"
     )]
     NonZeroValue,
+
+    /// Thrown if a AA transaction with a nonce key prefixed with the sub-block prefix marker added to the pool
+    #[error("AA transaction with subblock nonce key prefix aren't supported in the pool")]
```
Zygimantass, line 117: https://docs.tempo.xyz/errors/tx/SubblockNonceKey

klkvr: this is unreachable for RPC transactions I believe, only from p2p

crates/transaction-pool/src/validator.rs
klkvr: they are supposed to be sent directly to validator

PR #1018, opened 2025-11-22
feat(subblocks): propagate subblock feeRecipient to EVM

klkvr:
ref https://github.com/tempoxyz/docs/pull/303

As preparation to non-free subblock transactions changes `TempoBlockExecutor` to contain mapping from validator public key to the configured fee recipient for each subblock.

This allows EVM to eventually use this context once we have necessary `FeeManager` changes in-place

### DIFF
--- a/crates/revm/src/evm.rs
+++ b/crates/revm/src/evm.rs
@@ -33,7 +33,7 @@ pub struct TempoEvm<DB: Database, I> {
     >,
     /// Preserved logs from the last transaction
     pub logs: Vec<Log>,
-    /// Subblock fee recipient, if we executing a subblock transaction.
+    /// Subblock fee recipient, if executing a subblock transaction.
     pub subblock_fee_recipient: Option<Address>,
 }
 


crates/revm/src/evm.rs
```diff
@@ -33,6 +33,8 @@ pub struct TempoEvm<DB: Database, I> {
     >,
     /// Preserved logs from the last transaction
     pub logs: Vec<Log>,
+    /// Subblock fee recipient, if we executing a subblock transaction.
```
SuperFluffy:
```suggestion
    /// Subblock fee recipient, if executing a subblock transaction.
```

crates/payload/builder/src/lib.rs
```diff
@@ -403,6 +375,46 @@ where
...
+                    shared_gas_limit,
+                    timestamp_millis_part: attributes.timestamp_millis_part(),
+                    extra_data: attributes.extra_data().clone(),
+                    subblock_fee_recipients,
+                },
+            )
+            .map_err(PayloadBuilderError::other)?;
+
+        builder.apply_pre_execution_changes().map_err(|err| {
+            warn!(%err, "failed to apply pre-execution changes");
```
SuperFluffy, line 412: Note that this is not going to give you the full source chain, but only the display impl for `err` (i.e. the topmost error).

Other than using workarounds like our eyre error hook (if you have an `eyre::Report`), the unfortunately verbose way of doing this is:

```suggestion
            warn!(error = &err as &dyn std::error::Error, "failed to apply pre-execution changes");
```

Alternatively you can use https://github.com/tempoxyz/tempo/blob/1392d6cad0f24f9a72c7c36d19aeced0908e6102/crates/telemetry-util/src/lib.rs#L91

```suggestion
            warn!(error = error_field(&err), "failed to apply pre-execution changes");
```

---

Unrelated point: I think we should coalesce on a common  field name for errors (sometimes we do `err`, sometimes `e`, sometimes `error`). I am partial to `error = ` :-)

crates/evm/src/lib.rs
```diff
@@ -167,6 +172,24 @@ impl ConfigureEvm for TempoEvmConfig {
...
         block: &'a SealedBlock<Block>,
     ) -> Result<TempoBlockExecutionCtx<'a>, Self::Error> {
+        // Decode validator -> fee_recipient mapping from the subblock metadata system transaction.
+        let subblock_fee_recipients = block
+            .body()
+            .transactions
+            .iter()
+            .rev()
+            .filter(|tx| (*tx).to() == Some(Address::ZERO))
+            .find_map(|tx| Vec::<SubBlockMetadata>::decode(&mut tx.input().as_ref()).ok())
```
mattsse, line 182: this allows trailing bytes in the input, which should be okay, but we could check buf empty if this is relevant

PR #1021, opened 2025-11-23
fix(consensus): retain old epoch information until full transition

### LINKED ISSUE #1020: Bug: node cannot recover if it goes down before processing last height of an epoch
### Describe the bug

As of current main, if a node goes down right after finalizing its local DKG ceremony and persisting its updated epoch information (public polynomial, private key share), it will not be able to recover because it lost all information about the not-yet-ended epoch:
https://github.com/tempoxyz/tempo/blob/3a585a46eba17d4378b3160f1b839d81c228f88f/crates/commonware-node/src/dkg/manager/actor.rs#L443

If enough nodes go down in this way and at the same time, then the network suffers a catastrophic halt because it cannot recover.

This is already fixed in #971.

### Steps to reproduce

Kill a node right after its finalized the pre-to-last height of an epoch, observe that it does not start a consensus engine backing the outgoing epoch.

### Logs

```text

```

### Platform(s)

_No response_

### Container Type

Not running in a container

3a585a46eba17d4378b3160f1b839d81c228f88f

_No response_

### Code of Conduct

SuperFluffy:
Retains the epoch state of the outgoing epoch until the node is fully transitioned to the new (by observing a finalized block in the new epoch).

Uses the retained outgoing epoch state to start a consensus engine backing the old epoch on restart.

While working on this patch, it was discovered that the node would also not start into the incoming epoch: because the boundary block was not yet finalized and present, the new incoming epoch did not have its "genesis" and refused to start.

This patch also provides a test to cover this scenario: by looping over the consensus metrics every 100ms, it attempts to catch a node just as it processed the boundary-1-th block and restarts it.

Closes #1020

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 version = "0.5.2"
 edition = "2024"
 rust-version = "1.88"
-license = ""
+license = "MIT OR Apache-2.0"
 publish = false
 
 [workspace]


crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -132,27 +133,63 @@ where
...
+                    }
+                    .into(),
+                )
+                .await;
+        }
+
+        if let Some(epoch_state) = self.epoch_metadata.get(&PREVIOUS_EPOCH_KEY) {
+            self.config
+                .epoch_manager
+                .report(
```
klkvr, line 185: let's document here that this will only be entered if we did not observe a finalization of first block of a new epoch yet

crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -376,16 +413,20 @@ where
         // Recall, for an epoch length E the first heights are 0E, 1E, 2E, ...
         if block.height().is_multiple_of(self.config.epoch_length) {
             // Special case epoch == 0
-            if let Some(previous_epoch) = ceremony.epoch().checked_sub(1) {
+            if let Some(old_epoch_state) = self.epoch_metadata.remove(&PREVIOUS_EPOCH_KEY) {
```
klkvr:
```suggestion
            if let Some(old_epoch_state) = self.epoch_metadata.remove(&PREVIOUS_EPOCH_KEY) {
```

crates/commonware-node/src/dkg/manager/actor.rs
```diff
@@ -433,19 +474,27 @@ where
...
+            self.epoch_metadata.put(PREVIOUS_EPOCH_KEY, old_epoch_state);
+            let new_epoch_state = EpochState {
                 epoch: next_epoch,
                 participants: ceremony_outcome.participants,
                 public,
                 share,
             };
             self.epoch_metadata
-                .put_sync(EPOCH_KEY.into(), epoch_state.clone())
+                .put(CURRENT_EPOCH_KEY, new_epoch_state.clone());
```
klkvr, line 491: it's a bit confusing that we flip `CURRENT_EPOCH_KEY` on the pre-to-last block given that the new epoch can't even be started yet

I guess this has similar motivation as ceremony finalization logic but imo worth exploring if we can make this more intuitive

SuperFluffy: ~I agree. I went through a couple of naming schemes but none of them made me happy (neither does this one). Let's bikeshed this.~

Actually, it's all fixed in the dynamic validator rework. There, we are only updating the current epoch state on the last block.

crates/commonware-node/src/dkg/manager/actor.rs
SuperFluffy: b1cdb6fd

crates/commonware-node/src/dkg/manager/actor.rs
SuperFluffy: [cb117e9](https://github.com/tempoxyz/tempo/pull/1021/commits/cb117e91e2598b61199cfbbb166a12b10ae8bfe5)

PR #1022, opened 2025-11-24
feat(revm): Update fee token selection logic

### LINKED ISSUE #992: Changes to user token preference rules
### Describe the feature

- Add rule (at same preference level as the rules when the transaction calls a TIP-20 contract) that if a transaction calls swapExactAmountOut() or swapExactAmountIn() on the Stablecoin Exchange, then the fee token is the input token specified (the token that that function will pull from the user).
- Remove fallback to the validator’s preferred token (currently undocumented but implemented).
- Add fallback to pathUSD as the final step in the cascade.

### Additional context

_No response_

### LINKED ISSUE #993: Changes to validator token preference rules
### Describe the feature

If the validator has not specified a fee token preference, fall back to pathUSD as the validator token preference.

### Additional context

_No response_

0xKitsune:
Closes #992
Closes  #993
Closes CHAIN-191

Stacks on #1015 

This PR updates `user_or_tx_fee_token()` to handle if `tx.to` is the `StablecoinExchange`. Additionally this PR removes the condition where the fee token falls back to the validator fee token and sets the `DEFAULT_FEE_TOKEN` (ie. first TIP20 deployed) as a fallback when no fee token is found.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7241,7 +7241,7 @@ checksum = "1e061d1b48cb8d38042de4ae0a7a6401009d6143dc80d2e2d6f31f0bdd6470c7"
 [[package]]
 name = "reth-basic-payload-builder"
 version = "1.9.3"
-source = "git+https://github.com/paradigmxyz/reth?rev=587b1f4#587b1f45db2aa90af1c8059d6927bea2c1b623ac"
+source = "git+https://github.com/paradigmxyz/reth?rev=8528769#85287698963e8640dd63995b49fff59096f21adc"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
@@ -7265,7 +7265,7 @@ dependencies = [
 [[package]]
 name = "reth-chain-state"
 version = "1.9.3"
-source = "git+https://github.com/paradigmxyz/reth?rev=587b1f4#587b1f45db2aa90af1c8059d6927bea2c1b623ac"
+source = "git+https://github.com/paradigmxyz/reth?rev=8528769#85287698963e8640dd63995b49fff59096f21adc"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
@@ -7296,7 +7296,7 @@ dependencies = [
 [[package]]
 name = "reth-chainspec"
 version = "1.9.3"
-source = "git+https://github.com/paradigmxyz/reth?rev=587b1f4#587b1f45db2aa90af1c8059d6927bea2c1b623ac"
+source = "git+https://github.com/paradigmxyz/reth?rev=8528769#85287698963e8640dd63995b49fff59096f21adc"
 dependencies = [
  "alloy-chains",
  "alloy-consensus",
@@ -7316,7 +7316,7 @@ dependencies = [
 [[package]]
 name = "reth-cli"
 version = "1.9.3"
-source = "git+https://github.com/paradigmxyz/reth?rev=587b1f4#587b1f45db2aa90af1c8059d6927bea2c1b623ac"
+source = "git+https://github.com/paradigmxyz/reth?rev=8528769#85287698963e8640dd63995b49fff59096f21adc"
 dependencies = [
  "alloy-genesis",
  "clap",
@@ -7330,7 +7330,7 @@ dependencies = [
 [[package]]
 name = "reth-cli-commands"
 version = "1.9.3"
-source = "git+https://github.com/paradigmxyz/reth?rev=587b1f4#587b1f45db2aa90af1c8059d6927bea2c1b623ac"
+source = "git+https://github.com/paradigmxyz/reth?rev=8528769#85287698963e8640dd63995b49fff59096f21adc"
 dependencies = [
  "alloy-chains",
  "alloy-consensus",
@@ -7411,7 +7411,7 @@ dependencies = [
 [[package]]
 name = "reth-cli-runner"
 version = "1.9.3"
-source = "git+https://github.com/paradigmxyz/reth?rev=587b1f4#587b1f45db2aa90af1c8059d6927bea2c1b623ac"
+source = "git+https://github.com/paradigmxyz/reth?rev=8528769#85287698963e8640dd63995b49fff59096f21adc"
 dependencies = [
  "reth-tasks",
  "tokio",
@@ -7421,7 +7421,7 @@ dependencies = [
 [[package]]
 name = "reth-cli-util"
 version = "1.9.3"
-source = "git+https://github.com/paradigmxyz/reth?rev=587b1f4#587b1f45db2aa90af1c8059d6927bea2c1b623ac"
+source = "git+https://github.com/paradigmxyz/reth?rev=8528769#85287698963e8640dd63995b49fff59096f21adc"
 dependencies = [
  "alloy-eips",
  "alloy-primitives",
@@ -7439,7 +7439,7 @@ dependencies = [
 [[package]]
 name = "reth-codecs"
 version = "1.9.3"
-source = "git+https://github.com/paradigmxyz/reth?rev=587b1f4#587b1f45db2aa90af1c8059d6927bea2c1b623ac"
+source = "git+https://github.com/paradigmxyz/reth?rev=8528769#85287698963e8640dd63995b49fff59096f21adc"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
@@ -7459,7 +7459,7 @@ dependencies = [
 [[package]]
 name = "reth-codecs-derive"
 version = "1.9.3"
-source = "git+https://github.com/paradigmxyz/reth?rev=587b1f4#587b1f45db2aa90af1c8059d6927bea2c1b623ac"
+source = "git+https://github.com/paradigmxyz/reth?rev=8528769#85287698963e8640dd63995b49fff59096f21adc"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -7469,7 +7469,7 @@ dependencies = [
 [[package]]
 name = "reth-config"
 version = "1.9.3"
-source = "git+https://github.com/paradigmxyz/reth?rev=587b1f4#587b1f45db2aa90af1c8059d6927bea2c1b623ac"
+source = "git+https://github.com/paradigmxyz/reth?rev=8528769#85287698963e8640dd63995b49fff59096f21adc"
 dependencies = [
  "eyre",
  "humantime-serde",
@@ -7485,7 +7485,7 @@ dependencies = [
 [[package]]
 name = "reth-consensus"
 version = "1.9.3"
-source = "git+https://github.com/paradigmxyz/reth?rev=587b1f4#587b1f45db2aa90af1c8059d6927bea2c1b623ac"
+source = "git+https://github.com/paradigmxyz/reth?rev=8528769#85287698963e8640dd63995b49fff59096f21adc"
 dependencies = [
  "alloy-consensus",
  "alloy-primitives",
@@ -7498,7 +7498,7 @@ dependencies = [
 [[package]]
 name = "reth-consensus-common"
 version = "1.9.3"
-source = "git+https://github.com/paradigmxyz/reth?rev=587b1f4#587b1f45db2aa90af1c8059d6927bea2c1b623ac"
+source = "git+https://github.com/paradigmxyz/reth?rev=8528769#85287698963e8640dd63995b49fff59096f21adc"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
@@ -7510,7 +7510,7 @@ dependencies = [
 [[package]]
 name = "reth-consensus-debug-client"
 version = "1.9.3"
-source = "git+https://github.com/paradigmxyz/reth?rev=587b1f4#587b1f45db2aa90af1c8059d6927bea2c1b623ac"
+source = "git+https://github.com/paradigmxyz/reth?rev=8528769#85287698963e8640dd63995b49fff59096f21adc"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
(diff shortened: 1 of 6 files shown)


crates/revm/src/common.rs
```diff
@@ -94,22 +94,36 @@ pub trait TempoStateAccess<T> {
...
+            // use the input token as the fee token (the token that will be pulled from the user)
+            if kind.to() == Some(&STABLECOIN_EXCHANGE_ADDRESS) {
+                if let Ok(call) = IStablecoinExchange::swapExactAmountInCall::abi_decode(input) {
+                    return Ok(call.tokenIn);
+                } else if let Ok(call) =
+                    IStablecoinExchange::swapExactAmountOutCall::abi_decode(input)
+                {
+                    return Ok(call.tokenIn);
+                }
+            }
```
klkvr: as per https://github.com/tempoxyz/tempo/issues/992

> Add rule (at same preference level as the rules when the transaction calls a TIP-20 contract) that if a transaction calls swapExactAmountOut() or swapExactAmountIn() on the Stablecoin Exchange, then the fee token is the input token specified (the token that that function will pull from the user).

this sounds like this logic should be added below? right now it's added at the same level as preference inferred from `setUserToken` call, this means that stablecoing exchange preference would a. override the user token preference, b. not be respected for AA transactions

crates/revm/src/common.rs
klkvr: let's rename this to `get_fee_token` now that there're no other preference sources?

crates/revm/src/common.rs
```diff
@@ -120,41 +134,15 @@ pub trait TempoStateAccess<T> {
...
-        // Otherwise fall back to the validator fee token preference
-        let validator_slot = mapping_slot(validator, tip_fee_manager::slots::VALIDATOR_TOKENS);
-        let validator_fee_token = self
-            .sload(TIP_FEE_MANAGER_ADDRESS, validator_slot)?
-            .into_address();
-
-        if !validator_fee_token.is_zero() {
-            return Ok(validator_fee_token);
+            return Ok(to);
         }
```
klkvr, line 158: I don't think we can just remove this, it'd need to be hardfork-gated, which is likely why re-execution tests are failing

crates/node/tests/assets/test-genesis.json
```diff
@@ -20,6 +20,8 @@
     "terminalTotalDifficulty": 0,
     "terminalTotalDifficultyPassed": true,
     "adagioTime": 0,
+    "allegrettoTime": 0,
+    "moderatoTime": 0,
```
klkvr: is this necessary? I don't mind enabling moderato here but would prefer enabling allegretto later

in https://github.com/tempoxyz/tempo/pull/598 I've added some helpers to node setup builder to enable Allegretto for specific tests

crates/transaction-pool/src/validator.rs
```diff
@@ -68,21 +70,15 @@ where
...
-        let fee_token = match state_provider.user_or_tx_fee_token(transaction.inner(), fee_payer) {
-            Ok(Some(fee_token)) => fee_token,
-            Ok(None) => {
-                // If no fee token preference was specified at transaction or user level, this means that we will
-                // fallback to validator token. On Andantino testnet, all validators use LINKING_USD as their fee token,
-                // which will cause the transaction to be rejected because it can't be used to pay fees.
-                return TransactionValidationOutcome::Invalid(
-                    transaction,
-                    InvalidPoolTransactionError::other(TempoPoolTransactionError::MissingFeeToken),
-                );
```
klkvr, line 182: I guess if we're fine with this error not being returned before Allegretto, let's remove the variant from `TempoPoolTransactionError`?

crates/revm/src/common.rs
```diff
@@ -120,44 +126,52 @@ pub trait TempoStateAccess<T> {
...
-    ) -> Result<Address, Self::Error> {
-        // First check transaction or user preference
-        if let Some(fee_token) = self.user_or_tx_fee_token(tx, fee_payer)? {
-            return Ok(fee_token);
+        // If calling swapExactAmountOut() or swapExactAmountIn() on the Stablecoin Exchange,
+        // use the input token as the fee token (the token that will be pulled from the user)
+        if spec.is_allegretto()
+            && !tx.is_aa()
+            && let Some((kind, input)) = tx.calls().next()
+            && kind.to() == Some(&STABLECOIN_EXCHANGE_ADDRESS)
```
klkvr: is this correct that this does not apply to AA transactions? I'd imagine that we might want this to still work for AA tx with exactly one call to stablecoin DEX

crates/precompiles/src/tip_fee_manager/mod.rs
```diff
@@ -63,8 +63,8 @@ impl<'a, S: PrecompileStorageProvider> TipFeeManager<'a, S> {
     pub fn get_validator_token(&mut self, beneficiary: Address) -> Result<Address> {
         let token = self.sload_validator_tokens(beneficiary)?;
 
-        if token.is_zero() {
-            Ok(DEFAULT_FEE_TOKEN)
+        if self.storage.spec().is_allegretto() && token.is_zero() {
+            Ok(DEFAULT_FEE_TOKEN_PRE_ALLEGRETTO)
```
klkvr: this will return `Address::ZERO` pre-allegretto

crates/precompiles/src/tip_fee_manager/mod.rs
```diff
@@ -300,8 +300,8 @@ impl<'a, S: PrecompileStorageProvider> TipFeeManager<'a, S> {
     pub fn validator_tokens(&mut self, call: IFeeManager::validatorTokensCall) -> Result<Address> {
         let token = self.sload_validator_tokens(call.validator)?;
 
-        if token.is_zero() {
-            Ok(DEFAULT_FEE_TOKEN)
+        if self.storage.spec().is_allegretto() && token.is_zero() {
+            Ok(DEFAULT_FEE_TOKEN_PRE_ALLEGRETTO)
```
klkvr: same here

crates/node/tests/assets/test-genesis.json
fgimenez: nope, removed

crates/transaction-pool/src/validator.rs
fgimenez: good point, removed

crates/revm/src/common.rs
fgimenez: ok, pushed changes to allow fee selection for single-call AA txs, ptal

crates/precompiles/src/tip_fee_manager/mod.rs
fgimenez: good point, just checking now `if token.is_zero()`

crates/precompiles/src/tip_fee_manager/mod.rs
fgimenez: fixed

crates/precompiles/src/tip_fee_manager/mod.rs
```diff
@@ -64,7 +64,7 @@ impl<'a, S: PrecompileStorageProvider> TipFeeManager<'a, S> {
         let token = self.sload_validator_tokens(beneficiary)?;
 
         if token.is_zero() {
-            Ok(DEFAULT_FEE_TOKEN)
+            Ok(DEFAULT_FEE_TOKEN_PRE_ALLEGRETTO)
```
klkvr: should this not be `pathUSD` post-Allegretto?

crates/precompiles/src/tip_fee_manager/mod.rs
```diff
@@ -301,7 +301,7 @@ impl<'a, S: PrecompileStorageProvider> TipFeeManager<'a, S> {
         let token = self.sload_validator_tokens(call.validator)?;
 
         if token.is_zero() {
-            Ok(DEFAULT_FEE_TOKEN)
+            Ok(DEFAULT_FEE_TOKEN_PRE_ALLEGRETTO)
```
klkvr: same here and let's deduplicate those getters

crates/revm/src/common.rs
```diff
@@ -120,44 +126,56 @@ pub trait TempoStateAccess<T> {
...
+                    && self.is_valid_fee_token(call.tokenIn)?
+                {
+                    return Ok(call.tokenIn);
+                } else if let Ok(call) =
+                    IStablecoinExchange::swapExactAmountOutCall::abi_decode(input)
+                    && self.is_valid_fee_token(call.tokenIn)?
+                {
+                    return Ok(call.tokenIn);
+                }
+            }
```
klkvr, line 159: is there a spec for this? given that this behavior was just changed based on my review I guess we need to update if there is

xtask/src/genesis.rs
```diff
@@ -98,13 +98,13 @@ pub(crate) struct GenesisArgs {
...
+    /// Moderato hardfork activation timestamp (defaults to 0 = active at genesis)
+    #[arg(long, default_value_t = 0)]
+    pub moderato_time: u64,
 
-    /// Allegretto hardfork activation timestamp
-    #[arg(long)]
-    pub allegretto_time: Option<u64>,
+    /// Allegretto hardfork activation timestamp (defaults to 0 = active at genesis)
+    #[arg(long, default_value_t = 0)]
+    pub allegretto_time: u64,
```
klkvr: I think we can only default to 0 for moderato for now?

xtask/src/genesis.rs
```diff
@@ -309,17 +313,6 @@ impl GenesisArgs {
...
-            chain_config
-                .extra_fields
-                .insert("moderatoTime".to_string(), serde_json::json!(moderato_time));
-        }
-        if let Some(allegretto_time) = self.allegretto_time {
-            chain_config.extra_fields.insert(
-                "allegrettoTime".to_string(),
-                serde_json::json!(allegretto_time),
-            );
-        }
```
klkvr, line 322: this should be kept?

crates/transaction-pool/src/validator.rs
```diff
@@ -170,21 +173,15 @@ where
...
-            }
-            Err(err) => {
-                return TransactionValidationOutcome::Error(*transaction.hash(), Box::new(err));
-            }
-        };
+        let spec = TempoHardfork::Allegretto;
```
klkvr: ideally this should be fetched based on chainspec from inner validator

crates/contracts/src/precompiles/mod.rs
```diff
@@ -27,7 +27,12 @@ pub use validator_config::*;
 
 pub const TIP_FEE_MANAGER_ADDRESS: Address = address!("0xfeec000000000000000000000000000000000000");
 pub const PATH_USD_ADDRESS: Address = address!("0x20C0000000000000000000000000000000000000");
-pub const DEFAULT_FEE_TOKEN: Address = address!("0x20C0000000000000000000000000000000000001");
+/// Default fee token pre allegretto hardfork
+pub const DEFAULT_FEE_TOKEN_PRE_ALLEGRETTO: Address =
+    address!("0x20C0000000000000000000000000000000000001");
+/// Default fee token post allegretto hardfork
+pub const DEFAULT_FEE_TOKEN_POST_ALLEGRETTO: Address =
+    address!("0x20C0000000000000000000000000000000000000");
```
0xKitsune: nit: maybe we can update to `DEFAULT_FEE_TOKEN_POST_ALLEGRETTO: Address = PATH_USD_ADDRESS` to retain "default token" semantics throughout the codebase but also ensure there is no way the default fee token post allegretto can be anything other than pathUSD.

crates/precompiles/src/tip_fee_manager/mod.rs
0xKitsune: Yes this should be pathUSD. We should update this to `DEFAULT_FEE_TOKEN_POST_ALLEGRETTO` or `PATH_USD_ADDRESS` as these are the same thing.

crates/precompiles/src/tip_fee_manager/mod.rs
0xKitsune: Also does this need to be hardforked? Likely not an issue at the moment but if this was ever called onchain we would need to hardfork this.

crates/precompiles/src/tip_fee_manager/mod.rs
fgimenez: yep fixed

crates/revm/src/common.rs
0xKitsune: +1, I think the specs need to be aligned here. [#1050](https://github.com/tempoxyz/tempo/issues/1050) for reference. If this differs from FeeManager spec alignment and applies more to general "Fee Token Mechanics" we should add a sub issue to the tracking issue.

crates/precompiles/src/tip_fee_manager/mod.rs
fgimenez: done

crates/transaction-pool/src/validator.rs
0xKitsune: Agreed, can we update to this @fgimenez?

xtask/src/genesis.rs
0xKitsune: IIRC you can specify `time=0` for all hardforks active at genesis? Not sure if this has an affect though, I am under the impression that the latest hardfork will take precedence in any case.

fgimenez: yep fixed

xtask/src/genesis.rs
fgimenez: yep adapted for non-option moderato time

crates/precompiles/src/tip_fee_manager/mod.rs
klkvr: pretty sure this was triggered yeah

crates/transaction-pool/src/validator.rs
fgimenez: sure done

crates/contracts/src/precompiles/mod.rs
fgimenez: good point done

PR #1023, opened 2025-11-24
feat(precompiles): Add `feeRecipient` to `TIP20`

### LINKED ISSUE #994: Changes to TIP-20
### Describe the feature

Add the ability for the admin of the token to specify a feeRecipient address using a setFeeRecipient function.

### Additional context

_No response_

0xKitsune:
Closes #994

Stacks on #1015

This PR updates `TIP20` contracts to specify a `feeRecipient` and exposes relevant functions to update this value via admin controls.

### DIFF
--- a/crates/node/tests/it/tip_fee_amm.rs
+++ b/crates/node/tests/it/tip_fee_amm.rs
@@ -89,27 +89,31 @@ async fn test_mint_liquidity() -> eyre::Result<()> {
     let total_supply = fee_amm.totalSupply(pool_id).call().await?;
     let lp_balance = fee_amm.liquidityBalances(pool_id, caller).call().await?;
 
-    // With mintWithValidatorToken, liquidity = amount - MIN_LIQUIDITY (for first mint)
-    let expected_liquidity = amount - MIN_LIQUIDITY;
+    // With mintWithValidatorToken, liquidity = (amount / 2) - MIN_LIQUIDITY (for first mint)
+    // Only validator tokens are transferred, creating a one-sided pool
+    let half_amount = amount / U256::from(2);
+    let expected_liquidity = half_amount - MIN_LIQUIDITY;
     assert_eq!(lp_balance, expected_liquidity);
-    let expected_total_supply = amount;
+    let expected_total_supply = half_amount;
     assert_eq!(total_supply, expected_total_supply);
 
+    // Only validator reserve is updated (user reserve stays 0)
     let pool = fee_amm.pools(pool_id).call().await?;
-    assert_eq!(pool.reserveUserToken, amount.to::<u128>());
+    assert_eq!(pool.reserveUserToken, 0);
     assert_eq!(pool.reserveValidatorToken, amount.to::<u128>());
 
+    // User token balance unchanged (not transferred)
     let final_token0_balance = token_0.balanceOf(caller).call().await?;
-    assert_eq!(final_token0_balance, user_token0_balance - amount);
+    assert_eq!(final_token0_balance, user_token0_balance);
+    // Validator token balance decreased
     let final_token1_balance = token_1.balanceOf(caller).call().await?;
     assert_eq!(final_token1_balance, user_token1_balance - amount);
 
+    // User token not transferred to fee manager
     let final_fee_manager_token0_balance =
         token_0.balanceOf(TIP_FEE_MANAGER_ADDRESS).call().await?;
-    assert_eq!(
-        final_fee_manager_token0_balance,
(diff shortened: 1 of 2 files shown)


crates/precompiles/src/tip20/dispatch.rs
```diff
@@ -113,6 +113,16 @@ impl<'a, S: PrecompileStorageProvider> Precompile for TIP20Token<'a, S> {
...
             }
+
+            ITIP20::feeRecipientCall::SELECTOR => {
+                view::<ITIP20::feeRecipientCall>(calldata, |_call| self.sload_fee_recipient())
+            }
+            ITIP20::setFeeRecipientCall::SELECTOR => {
+                mutate_void::<ITIP20::setFeeRecipientCall>(calldata, msg_sender, |s, call| {
+                    self.set_fee_recipient(s, call.newRecipient)
+                })
+            }
```
0xKitsune, line 138: Technically this should be gated behind the hardfork. It is very unlikely that this function was called pre allegretto but it might make sense to gate this anyway.

crates/precompiles/src/tip20/mod.rs
```diff
@@ -346,9 +349,18 @@ impl<'a, S: PrecompileStorageProvider> TIP20Token<'a, S> {
         )
     }
 
+    /// Sets a new fee recipient
+    pub fn set_fee_recipient(&mut self, msg_sender: Address, new_recipient: Address) -> Result<()> {
+        self.check_role(msg_sender, DEFAULT_ADMIN_ROLE)?;
+        self.sstore_fee_recipient(new_recipient)?;
+
+        Ok(())
+    }
```
0xKitsune, line 367: Lets emit an event here

crates/precompiles/src/tip20/mod.rs
```diff
@@ -346,9 +349,18 @@ impl<'a, S: PrecompileStorageProvider> TIP20Token<'a, S> {
...
+
     // Token operations
     /// Mints new tokens to specified address
     pub fn mint(&mut self, msg_sender: Address, call: ITIP20::mintCall) -> Result<()> {
+        self.check_role(msg_sender, *ISSUER_ROLE)?;
```
0xKitsune: Can we remove this? This is already checked in the `_mint()` function

xtask/src/genesis.rs
```diff
@@ -367,13 +367,12 @@ fn create_and_mint_token(
...
-                    currency: currency.into(),
-                    quoteToken: PATH_USD_ADDRESS,
-                    admin,
-                },
+                name.into(),
+                symbol.into(),
+                currency.into(),
+                PATH_USD_ADDRESS,
+                admin,
+                Address::ZERO,
```
0xKitsune, line 375: Lets confirm if we should set this to anything or if we should leave this as `address(0)`. We can handle this in a subsequent PR.

crates/precompiles/src/tip20/dispatch.rs
fgimenez: sure done

crates/precompiles/src/tip20/mod.rs
fgimenez: ok introduced FeeRecipientUpdated we need to update spec/ref impl with this

crates/precompiles/src/tip20/mod.rs
fgimenez: yep removed

xtask/src/genesis.rs
fgimenez: ok sounds good

crates/contracts/src/precompiles/tip20_factory.rs
```diff
@@ -5,14 +5,30 @@ sol! {
   #[derive(Debug, PartialEq, Eq)]
     #[sol(rpc, abi)]
     interface ITIP20Factory {
+        /// TokenCreated event pre Allegretto hardfork (without feeRecipient)
         event TokenCreated(address indexed token, uint256 indexed tokenId, string name, string symbol, string currency, address quoteToken, address admin);
 
+        /// TokenCreatedWithFeeRecipient event post Allegretto hardfork (with feeRecipient)
+        event TokenCreatedWithFeeRecipient(address indexed token, uint256 indexed tokenId, string name, string symbol, string currency, address quoteToken, address admin, address feeRecipient);
```
klkvr: can we not name this `TokenCreated` as well?

fgimenez: indeed, done

PR #1024, opened 2025-11-24
chore(deny): apply `MIT OR Apache-2.0` license to crates

zerosnacks:
This was raised by a review from `cargo deny` indicating the crates were missing their license

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 version = "0.5.2"
 edition = "2024"
 rust-version = "1.88"
-license = ""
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -2,7 +2,7 @@
 version = "0.5.2"
 edition = "2024"
 rust-version = "1.88"
-license = ""
```
mattsse, line 5: have we decided on a license @jenpaff ?

jenpaff: Yes Apache mit

Issue #983, opened 2025-11-20
Unify DKG storage


jenpaff:
### Describe the feature

Ensure that state changes are atomic.

PR #1029, opened 2025-11-24
feat: add `MetadataDatabase`

joshieDo:
towards https://github.com/tempoxyz/tempo/issues/983 | stacked: `self` <- https://github.com/tempoxyz/tempo/pull/1041

Right now, the DKG actor has two different storages which can be out of sync/inconsistent if a restart happens as `handle_finalized` is executing. The [dynamic validator set PR](https://github.com/tempoxyz/tempo/pull/585) adds 3 or 4 more. 

https://github.com/tempoxyz/tempo/blob/155bcec3e57f820e200b35d84b9f2a37e9ca5883/crates/commonware-node/src/dkg/manager/actor.rs#L38-L39 

Keeping track of all the potential out-of-sync/consistency scenarios is a needless headache, so this PR unifies them under a small DB wrapper API that provides DB-like transaction capabilities.

In follow-up PRs:
* integrate `MetadataDatabase` and pass the `Tx` to the `handle_finalized`, and commit it after (alongside a `acknowledgment.acknowledge(); `
* migrate the current storage of the current active validators on spawn

https://github.com/tempoxyz/tempo/blob/155bcec3e57f820e200b35d84b9f2a37e9ca5883/crates/commonware-node/src/dkg/manager/actor.rs#L236-L240

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -201,7 +201,6 @@ auto_impl = "1"
 axum = "0.8.4"
 base64 = "0.22"
 blake3 = "1.8"
-rapidhash = "1.4"
 bytes = "1.8"
 clap = { version = "4.5.45", features = ["derive"] }
 const-hex = { version = "1.15.0" }


crates/commonware-node/src/db/mod.rs
```diff
@@ -0,0 +1,397 @@
...
+        Ok(Tx::new(self.inner.clone(), Some(self.tx_active.clone())))
+    }
+
+    /// Begin a new read-only transaction.
+    ///
+    /// Multiple read-only transactions can be active concurrently.
+    /// Read-only transactions cannot call `insert()` or `remove()`.
+    pub fn read_only(&self) -> Tx<TContext> {
+        Tx::new(self.inner.clone(), None)
+    }
```
joshieDo: i dont necessarily think we need this, id rather remove it and only add if we ever need it. A read-only tx done right would snapshot the underlying store (not happening rn)

Thoughts?

crates/commonware-node/src/db/mod.rs
```diff
@@ -0,0 +1,394 @@
...
+
+use alloy_primitives::keccak256;
+use bytes::Bytes;
+use commonware_codec::{EncodeSize, Read, Write as CodecWrite};
+use commonware_runtime::{Clock, Metrics, Storage};
+use commonware_storage::metadata::Metadata;
+use commonware_utils::sequence::FixedBytes;
+use parking_lot::RwLock;
+
+type B256 = FixedBytes<32>;
```
mattsse: this alias is redundant?

crates/commonware-node/src/db/mod.rs
```diff
@@ -0,0 +1,394 @@
...
+    }
+
+    /// Commit all buffered writes to the store and sync.
+    ///
+    /// Holds the write lock across the async sync operation to ensure atomicity - all writes
+    /// are applied and flushed to disk without other transactions interleaving.
+    #[allow(clippy::await_holding_lock)]
+    pub async fn commit(mut self) -> Result<(), eyre::Error> {
+        // Acquire write lock and apply all buffered writes and deletions
+        let mut store = self.store.write();
```
mattsse: this is holding a lock across an await point, this is big no no hence the clippy lint

crates/commonware-node/src/db/mod.rs
joshieDo: redudant in what way? unfortunately this is not the same B256 that alloy uses. And it can not be used, since it needs to impl a CW specific trait

mattsse: ah sigh.....

crates/commonware-node/src/db/mod.rs
mattsse: yeah this does indeed some kinda pointless

crates/commonware-node/src/db/mod.rs
joshieDo: changed for `async_lock`

crates/commonware-node/src/db/epoch.rs
```diff
@@ -0,0 +1,68 @@
...
+use eyre::Result;
+
+fn current_epoch_key(regime: HardforkRegime) -> &'static str {
+    match regime {
+        HardforkRegime::PreAllegretto => "pre_allegretto_epoch_current",
+        HardforkRegime::PostAllegretto => "post_allegretto_epoch_current",
+    }
+}
+
+fn previous_epoch_key(regime: HardforkRegime) -> &'static str {
```
mattsse: if these are defined in the same crate, can we not make this native fns?

crates/commonware-node/src/db/validators.rs
```diff
@@ -0,0 +1,45 @@
...
+    fn get_validators(
+        &mut self,
+        epoch: u64,
+    ) -> impl Future<Output = Result<Option<ValidatorState>>> + Send;
+
+    /// Set validators state for a specific epoch.
+    fn set_validators(&mut self, epoch: u64, state: ValidatorState) -> Result<()>;
+
+    /// Remove validators state for a specific epoch.
+    fn remove_validators(&mut self, epoch: u64);
```
mattsse: should these return the removed value?

I guess not useful

crates/commonware-node/src/db/ceremony.rs
```diff
@@ -0,0 +1,45 @@
...
+use crate::dkg::ceremony;
+use commonware_runtime::{Clock, Metrics, Storage};
+use eyre::Result;
+
+fn ceremony_key(epoch: u64) -> String {
+    format!("ceremony_{epoch}")
+}
+
+/// Trait for ceremony-related database operations.
+pub trait CeremonyStore<TContext>
```
SuperFluffy: Let's keep this `pub(crate)`. Can always make it public once we want to use it in other crates.

crates/commonware-node/src/dkg/ceremony/persisted.rs
```diff
@@ -14,7 +14,7 @@ use super::IntermediateOutcome;
 
 /// Information on a ceremony that is persisted to disk.
 #[derive(Clone, Default, Debug, PartialEq, Eq)]
-pub(in crate::dkg) struct State {
+pub struct State {
```
SuperFluffy: Design-wise it's IMO problematic, that this is now public (technically, crate-public), where before it was confined to the dkg module.

I would like to keep it that way.

crates/commonware-node/src/db/mod.rs
```diff
@@ -0,0 +1,372 @@
...
+        Ok(store.get(&key_hash).cloned())
+    }
+
+    /// Get a value from the store, deserializing it.
+    ///
+    /// This checks pending writes first, then falls back to the store.
+    pub async fn get<K, V>(&mut self, key: K) -> Result<Option<V>, eyre::Error>
+    where
+        K: AsRef<[u8]>,
+        V: Read<Cfg = ()>,
```
SuperFluffy: This requirement is probably fine but restricts the usage of `Metadata` to only types that don't need any extra configuration.

That's probably fine though.

crates/commonware-node/src/db/mod.rs
```diff
@@ -0,0 +1,372 @@
...
+        }
+
+        let store = self.store.read().await;
+        Ok(store.get(&key_hash).cloned())
+    }
+
+    /// Get a value from the store, deserializing it.
+    ///
+    /// This checks pending writes first, then falls back to the store.
+    pub async fn get<K, V>(&mut self, key: K) -> Result<Option<V>, eyre::Error>
```
SuperFluffy: Making this both `async` and fallible is problematic, since a lot of the code assumes neither (especially when recording span fields).

crates/commonware-node/src/db/mod.rs
```diff
@@ -0,0 +1,372 @@
...
+    where
+        K: AsRef<[u8]>,
+        V: Read<Cfg = ()>,
+    {
+        let Some(value_bytes) = self.get_raw(key).await? else {
+            return Ok(None);
+        };
+
+        Ok(Some(deserialize_from_bytes::<V>(&value_bytes)?))
```
SuperFluffy: This means that every read now incurs a deserialization step? IMO that too is a regression compared to what we had before.

crates/commonware-node/src/db/mod.rs
```diff
@@ -0,0 +1,372 @@
...
+                None => {
+                    store.remove(&key);
+                }
+            }
+        }
+
+        store
+            .sync()
+            .await
+            .map_err(|e| eyre::eyre!("sync failed: {}", e))?;
```
SuperFluffy: `wrap_err`?

crates/commonware-node/src/db/mod.rs
```diff
@@ -0,0 +1,372 @@
...
+}
+
+/// Deserialize a value from Bytes.
+fn deserialize_from_bytes<T>(bytes: &Bytes) -> Result<T, eyre::Error>
+where
+    T: Read<Cfg = ()>,
+{
+    let slice: &[u8] = bytes.as_ref();
+    let mut cursor: &[u8] = slice;
+    T::read_cfg(&mut cursor, &()).map_err(|e| eyre::eyre!("deserialization failed: {}", e))
```
SuperFluffy: `wrap_err`?

crates/commonware-node/src/db/mod.rs
SuperFluffy: Why do we use `B256` in the first place? Can't we just use `FixedBytes<32>` instead? IMO the name causes confusion since it has nothing(?) to do with the alloy `B256`.

crates/commonware-node/src/db/mod.rs
```diff
@@ -0,0 +1,429 @@
...
+use commonware_storage::metadata::Metadata;
+use commonware_utils::sequence::FixedBytes;
+
+type B256 = FixedBytes<32>;
+
+/// Key used to store the node version that last wrote to the database.
+const NODE_VERSION_KEY: &str = "_node_version";
+
+/// Inner state of the metadata database.
+pub struct MetadataDatabaseInner<TContext>
```
SuperFluffy: The name suggestes that this should be private and only the higher level `MetadataBase` exposed

crates/commonware-node/src/db/mod.rs
```diff
@@ -0,0 +1,429 @@
...
+    /// Cache to avoid repeated deserialization.
+    pub cache: HashMap<B256, Box<dyn CachedValue>>,
+}
+
+/// A database wrapper around `Metadata<B256, Bytes>` that provides transactional operations.
+///
+/// The underlying [`Metadata`] storage always keeps all data in memory and persists
+/// it to disk during commit/sync operations.
+#[derive(Clone)]
+pub struct MetadataDatabase<TContext>(pub Arc<RwLock<MetadataDatabaseInner<TContext>>>)
```
SuperFluffy: Hide the innards of `MetadataBase`?

crates/commonware-node/src/db/mod.rs
```diff
@@ -0,0 +1,429 @@
...
+mod validators;
+pub use validators::ValidatorsStore;
+
+use eyre::WrapErr;
+use std::{any::Any, collections::HashMap, sync::Arc};
+
+use alloy_primitives::keccak256;
+use async_lock::RwLock;
+use bytes::Bytes;
+use commonware_codec::{EncodeSize, Read, Write as CodecWrite};
```
SuperFluffy:
```suggest
use commonware_codec::{EncodeSize, Read, Write};
```
(because I think there are no other instances of `Write` being used?);

crates/commonware-node/src/db/ceremony.rs
```diff
@@ -0,0 +1,62 @@
...
```
klkvr: don't think we need the context generic on the traits? also are we planning to support alternative storage impls? traits introduce a bit of boilerplate which we could avoid if we've just implement methods directly on `Tx`

crates/commonware-node/src/db/mod.rs
```diff
@@ -0,0 +1,429 @@
...
+        let key_hash = key_to_b256(key.as_ref());
+
+        // Check pending writes first
+        if let Some(value_opt) = self.writes.get(&key_hash) {
+            return match value_opt {
+                Some(cached) => match cached.as_any().downcast_ref::<V>() {
+                    Some(value) => Ok(Some(value.clone())),
+                    None => Err(eyre::eyre!(
+                        "type mismatch: pending write cannot be downcast to requested type"
+                    )),
```
klkvr, line 173: should this panic? doesn't look like we actually handle this error anywhere

klkvr (approved):
this looks good to me

Cargo.toml
```diff
@@ -194,6 +194,7 @@ commonware-runtime = "0.0.63"
 commonware-storage = "0.0.63"
 commonware-utils = "0.0.63"
 arbitrary = { version = "1.3", features = ["derive"] }
+async-lock = "3"
```
SuperFluffy: nit: I prefer to write the exact version of dependencies I used. It provides a lower bound because Cargo.lock is not always reliable

crates/commonware-node/src/db/ceremony.rs
SuperFluffy: Ack, I would like to have these as newtype wrappers around the inuner DB (mentioned before).

crates/commonware-node/src/db/ceremony.rs
```diff
@@ -0,0 +1,62 @@
...
+
+    /// Update ceremony state for a specific epoch using a closure.
+    ///
+    /// This reads the current state (or creates a default if none exists),
+    fn update_ceremony<F>(&mut self, epoch: u64, f: F) -> impl Future<Output = Result<()>> + Send
+    where
+        F: FnOnce(&mut ceremony::State) + Send;
+}
+
+impl<TContext> CeremonyStore<TContext> for Tx<TContext>
```
SuperFluffy: Related to the trait solution - this seems very heavy / lots of boilerplate for single-implementor types.

crates/commonware-node/src/db/mod.rs
```diff
@@ -0,0 +1,571 @@
...
+            store,
+            cache: HashMap::new(),
+        })))
+    }
+
+    /// Get a value from cache or storage.
+    ///
+    /// If the value exists in storage but not in cache, it will be deserialized
+    /// and added to the cache for future access.
+    pub async fn get<K, V>(&self, key: K) -> Result<Option<V>, eyre::Error>
```
SuperFluffy: Had crazy idea to replace `K, V` by single `T` from which you derive the key and to which you downcast:

I am thinking of doing something like this instead:

```rust
pub async fn get<T>(&self) -> Result<Option<T>, eyre::Error>
where
    T: ToKey
```

But I think it's too hard doing that with indexed states like `CeremonyState<const EPOCH: u64>` because you can't do construct a type like `CeremonyState<n>` where `n` is `epoch_state.epoch`.

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -105,31 +85,27 @@ where
...
 
-        let ceremony_metadata = Metadata::init(
-            context.with_label("ceremony_metadata"),
-            commonware_storage::metadata::Config {
-                partition: format!("{}_ceremony", config.partition_prefix),
-                codec_config: (),
+        // Initialize the unified metadata database
+        let metadata: Metadata<ContextCell<TContext>, FixedBytes<32>, Bytes> = Metadata::init(
+            context.with_label("dkg_db"),
+            MetadataConfig {
```
SuperFluffy: Instead of single-use aliases
```suggestion
            metadata::Config {
```

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -105,31 +85,27 @@ where
...
         let context = ContextCell::new(context);
 
-        let ceremony_metadata = Metadata::init(
-            context.with_label("ceremony_metadata"),
-            commonware_storage::metadata::Config {
-                partition: format!("{}_ceremony", config.partition_prefix),
-                codec_config: (),
+        // Initialize the unified metadata database
+        let metadata: Metadata<ContextCell<TContext>, FixedBytes<32>, Bytes> = Metadata::init(
+            context.with_label("dkg_db"),
```
SuperFluffy: We already name the context `dkg_manager`, so this would become `dkg_manager_dkg_db`.
```suggestion
            context.with_label("database"),
```

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -105,31 +85,27 @@ where
...
+            MetadataConfig {
+                partition: format!("{}_dkg_db", config.partition_prefix),
```
SuperFluffy: Similar reasoning - the prefix is already `*_dkg_manager`:

```suggestion
                partition: format!("{}_database", config.partition_prefix),
```

crates/commonware-node/src/dkg/ceremony/mod.rs
```diff
@@ -155,12 +155,7 @@ where
 
         debug!("attempting to read ceremony state from disk");
         // TODO(janis): move this "recovery" logic to a function.
-        // Clone in order to not hold onto the lock too long.
-        let recovered = ceremony_metadata
-            .lock()
-            .await
-            .get(&config.epoch.into())
-            .cloned();
+        let recovered = tx.get_ceremony(config.epoch).await?;
```
SuperFluffy: context

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -105,31 +85,27 @@ where
...
+                codec_config: RangeCfg::from(0..=usize::MAX),
```
SuperFluffy, line 93: This one is unfortunate. Where does it come from? Because we are writing vectors of bytes?

crates/commonware-node/src/db/mod.rs
```diff
@@ -0,0 +1,571 @@
...
+        K: AsRef<[u8]>,
+    {
+        let key_hash = key_to_b256(key.as_ref());
+        self.writes.insert(key_hash, None);
+    }
+
+    /// Check if a key exists in the store.
+    ///
+    /// This checks pending writes first, then falls back to the database.
+    pub async fn exists<K>(&self, key: K) -> bool
```
SuperFluffy: Similar to hashmap and btreemap:
```suggestion
    pub async fn contains_key<K>(&self, key: K) -> bool
```

crates/commonware-node/src/db/mod.rs
```diff
@@ -0,0 +1,571 @@
...
+    /// This should be called when the database schema is initialized or migrated
+    /// to track which node version last modified the database.
+    pub fn set_node_version(&mut self, version: String) {
+        self.insert(NODE_VERSION_KEY, version.into_bytes())
+    }
+
+    // ── Ceremony Store ──────────────────────────────────────────────────────
+
+    /// Get ceremony state for a specific epoch.
+    pub async fn get_ceremony(&self, epoch: u64) -> Result<Option<ceremony::State>, eyre::Error> {
```
SuperFluffy: Do these need to be on the low level tx? Can we make this a new-type wrapper that is specific to the DKG actor? I.e. `crate::dkg::manager::Tx(crate::db::Tx).`

I like that that the DKG actor has its own database where it need not worry about anybody else messing with it.

crates/commonware-node/src/db/mod.rs
```diff
@@ -0,0 +1,571 @@
...
+///
+/// # Example
+/// ```ignore
+/// let mut tx = db.read_write();
+/// tx.insert("epoch", epoch_state);
+/// tx.insert("ceremony_0", ceremony_state);
+/// tx.remove("old_data");
+/// tx.commit().await?;  // All changes serialized and applied atomically
+/// ```
+pub struct Tx<TContext>
```
SuperFluffy:
```suggestion
pub struct ReadWriteTransaction<TContext>
```

crates/commonware-node/src/dkg/manager/migrate.rs
```diff
@@ -0,0 +1,269 @@
...
+    // Check if already migrated
+    if tx.get_node_version().await?.is_some() {
+        return Ok(());
+    }
+
+    info!("migrating old metadata stores to unified database");
+
+    let current_epoch = get_current_epoch_for_migration(context, partition_prefix).await;
+
+    migrate_ceremony_metadata(context, partition_prefix, tx, current_epoch).await?;
```
SuperFluffy: Add context which migrations failed (applies to all of these lines)

crates/commonware-node/src/dkg/manager/migrate.rs
```diff
@@ -0,0 +1,269 @@
...
+///
+/// This function migrates the following old metadata stores:
+/// 1. `{prefix}_ceremony` - ceremony state per epoch
+/// 2. `{prefix}_current_epoch` - pre-allegretto epoch state (current/previous)
+/// 3. `{prefix}_post_allegretto_current_epoch` - post-allegretto epoch state (current/previous)
+/// 4. `{prefix}_next_dkg_outcome` - DKG outcome
+/// 5. `{prefix}_validators` - validator state per epoch
+///
+/// The caller is responsible for setting the version and committing the transaction.
+pub(super) async fn maybe_migrate_to_db<TContext>(
```
SuperFluffy, line 57:
```suggestion
#[instrument(skip_all, err)]
pub(super) async fn maybe_migrate_to_db<TContext>(
```

crates/commonware-node/src/dkg/manager/validators.rs
```diff
@@ -130,10 +130,10 @@ async fn decode_from_contract(
 /// 2. the player, that will become dealers in the next ceremony
 /// 3. the syncing players, that will become players in the next ceremony
 #[derive(Clone, Debug)]
-pub(super) struct ValidatorState {
-    dealers: OrderedAssociated<PublicKey, DecodedValidator>,
-    players: OrderedAssociated<PublicKey, DecodedValidator>,
-    syncing_players: OrderedAssociated<PublicKey, DecodedValidator>,
+pub struct ValidatorState {
```
SuperFluffy: Please don't make all of this pub.

crates/commonware-node/src/dkg/manager/validators.rs
```diff
@@ -289,28 +289,28 @@ impl Read for ValidatorState {
 /// `<host>` is either an IPv4 or IPV6 address, or a fully qualified domain name.
 /// `<ip>` is an IPv4 or IPv6 address.
 #[derive(Clone, Debug, PartialEq, Eq)]
-pub(super) struct DecodedValidator {
+pub struct DecodedValidator {
```
SuperFluffy: Same as before -  I would like to have these stay module-private.

If you take the newtype wrapper route for the read-write tx and put that into the dkg module, then all of these can stay private.

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -105,31 +85,27 @@ where
...
-        .await
-        .expect("must be able to initialize metadata on disk to function");
+        // Run migration from old metadata stores if needed
+        {
+            let mut tx = db.read_write();
+            super::migrate::maybe_migrate_to_db(&context, &config.partition_prefix, &mut tx)
+                .await?;
+            tx.set_node_version(env!("CARGO_PKG_VERSION").to_string());
+            tx.commit().await?;
+        }
```
SuperFluffy, line 111: context everywhere?

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -105,31 +85,27 @@ where
...
-            },
-        )
-        .await
-        .expect("must be able to initialize metadata on disk to function");
+        // Run migration from old metadata stores if needed
+        {
+            let mut tx = db.read_write();
+            super::migrate::maybe_migrate_to_db(&context, &config.partition_prefix, &mut tx)
+                .await?;
+            tx.set_node_version(env!("CARGO_PKG_VERSION").to_string());
```
SuperFluffy, line 107: Probably a good thing to have this looking forward.

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -399,50 +367,68 @@ where
...
+                *block,
+                maybe_ceremony,
+                ceremony_mux,
+                &mut tx,
+            )
+            .await;
         }
+
+        tx.commit()
+            .await
```
SuperFluffy, line 416: I like that this is explicit now.

crates/commonware-node/src/db/mod.rs
joshieDo: alloy also uses `FixedBytes<_>`, so in my head it's the same thing. I'd rather read `HashMap<B256...` than `HashMap<FixedBytes<32...`

crates/commonware-node/src/db/mod.rs
```diff
@@ -0,0 +1,571 @@
...
+        if self.writes.is_empty() {
+            return Ok(());
+        }
+
+        let mut db = self.db.0.write().await;
+
+        for (key, value_opt) in self.writes {
+            match value_opt {
+                Some(cached) => {
+                    let bytes = cached.serialize()?;
```
SuperFluffy: context

crates/commonware-node/src/db/mod.rs
```diff
@@ -0,0 +1,571 @@
...
+    /// This serializes all pending writes, persists them to storage,
+    /// and updates the shared cache with the new values.
+    pub async fn commit(self) -> Result<(), eyre::Error> {
+        if self.writes.is_empty() {
+            return Ok(());
+        }
+
+        let mut db = self.db.0.write().await;
+
+        for (key, value_opt) in self.writes {
```
SuperFluffy, line 271: A summary of the changes performed here would be extremely cool. Maybe at debug level?

joshieDo: rn i added just number of inserts and deletes - could probably add the keys, but unsure if worth it? what did you have in mind?

SuperFluffy: Was thinking of literally keys and values being written. Might be overly verbose though for DEBUG.

crates/commonware-node/src/db/mod.rs
SuperFluffy: I understand, but you have to acknowledge that that's confusing since `B256` is a well-known alloy alias. If you do a `rg B256` and this lights up, you will at first be confused why this alias has nothing to do with what you are expecting.

crates/commonware-node/src/db/mod.rs
```diff
@@ -0,0 +1,425 @@
...
+    /// Returns None if no version has been written yet (new database).
+    pub async fn get_node_version(&self) -> Result<Option<String>, eyre::Error> {
+        let key_hash = key_to_b256(NODE_VERSION_KEY.as_bytes());
+
+        // Check pending writes first
+        if let Some(value_opt) = self.writes.get(&key_hash) {
+            return match value_opt {
+                Some(cached) => Ok(Some(
+                    String::from_utf8(cached.serialize()?.to_vec())
+                        .map_err(|e| eyre::eyre!("invalid utf8: {}", e))?,
```
SuperFluffy:
```suggestion
                        .wrap_err("node version was invalid utf8")?
```

crates/commonware-node/src/db/mod.rs
```diff
@@ -0,0 +1,425 @@
...
+            .get_raw(NODE_VERSION_KEY)
+            .await
+            .and_then(|b| String::from_utf8(b.to_vec()).ok()))
+    }
+
+    /// Set the current node version in the database.
+    ///
+    /// This should be called when the database schema is initialized or migrated
+    /// to track which node version last modified the database.
+    pub fn set_node_version(&mut self, version: String) {
```
SuperFluffy, line 252: We can do this in a follow - but should this be done in some kind of builder?

crates/commonware-node/src/db/mod.rs
```diff
@@ -0,0 +1,425 @@
...
+
+    /// In-memory write buffer that accumulates all operations before commit.
+    /// - Key: `B256`
+    /// - Value: `Some(Bytes)` for insert/update, `None` for delete
+    ///
+    /// This map provides both write buffering and serves as a read cache,
+    /// ensuring that reads within a transaction see uncommitted writes.
+    ///
+    /// On commit, these are serialized to storage and merged into the shared cache.
+    writes: HashMap<B256, Option<Box<dyn CachedValue>>>,
```
SuperFluffy: Do we really need 256 bit keys? I think u64 (`U64` commonware type) is plenty. We can then also use much simpler non-cryptographic hash functions like FxHash.

crates/commonware-node/src/db/mod.rs
```diff
@@ -0,0 +1,425 @@
...
+        db.store.sync().await.wrap_err("sync failed")?;
+
+        Ok(())
+    }
+}
+
+/// Convert key to B256.
+///
+/// Zero-pads if key is <= 32 bytes, hashes with [`keccak256`] if > 32 bytes.
+fn key_to_b256(key: &[u8]) -> B256 {
```
SuperFluffy: Mentioned elsewhere: keccak + 256 bits seems complete overkill for our needs.

crates/commonware-node/src/dkg/manager/mod.rs
```diff
@@ -11,14 +11,16 @@ use futures::channel::mpsc;
 use rand_core::CryptoRngCore;
 use tempo_node::TempoFullNode;
 
-mod actor;
+pub mod actor;
 mod ingress;
-mod validators;
+mod migrate;
+pub(super) mod tx;
```
SuperFluffy: Let's also rename this to match the main type of the module:

```suggestion
pub(super) mod read_write_transaction;
```

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -399,50 +370,68 @@ where
...
+            .expect("must be able to commit finalize tx");
         acknowledgment.acknowledge();
```
SuperFluffy, line 418: We have an opportunity to bake replay protection into the DB: we should tag each commit written to the database with the block number that was processed.

If the node goes down immediately after `.commit` but before `acknowledgment.acknowledge` was processed by the marshal actor, then the last block will be replayed on startup. But we can then just check if `block.height > committed.height` and jump straight to acknowledge otherwise.

SuperFluffy (approved):
I would like to see a few changes:

1. let's `s/FixedBytes<u32>/U64` and also get rid of the cryptographic keccak hasher (both 256 bit keys and keccak seem complete overkill for on-data storage that we control)
2. I think all keys should go through a hasher (basically, treat the DB as a hashmap) without an escape hatch - at least on standard methods like `.get`
3. bake in rewrite protection by tagging commits with heights

Further, if this should end up in main, can you please first do a test where you 1. start with node built from HEAD@main, and then 2. stop it and restart it with a node based off of this branch?

This is to check that the handover indeed works.

crates/commonware-node/src/db/mod.rs
joshieDo: changed to u64 + foldhash

crates/commonware-node/src/db/mod.rs
joshieDo: changed to u64 + foldhash

Cargo.toml
```diff
@@ -195,18 +195,20 @@ commonware-runtime = "0.0.63"
 commonware-storage = "0.0.63"
 commonware-utils = "0.0.63"
 arbitrary = { version = "1.3", features = ["derive"] }
+async-lock = "3.4.1"
 async-trait = "0.1"
 auto_impl = "1"
 axum = "0.8.4"
 base64 = "0.22"
+blake3 = "1.8"
+rapidhash = "1.4"
```
SuperFluffy: forgot to rm this line

PR #1030, opened 2025-11-24
fix: clear prev order + test

### LINKED ISSUE #1007: Bug: `fill_order` does not clear `Order.prev`
Severity: Medium
Status: Open

**Notes:** ✅

0xrusowsky:
## Motivation

closes #1007 

## Solution

clear previous order when an order is filled

### DIFF
--- a/crates/precompiles/src/stablecoin_exchange/mod.rs
+++ b/crates/precompiles/src/stablecoin_exchange/mod.rs
@@ -4127,8 +4127,7 @@ mod tests {
     }
 
     #[test]
-    // https://github.com/tempoxyz/tempo/issues/1007
-    fn test_audit_issue_1007() -> eyre::Result<()> {
+    fn test_clear_order_post_allegretto() -> eyre::Result<()> {
         const AMOUNT: u128 = 1_000_000_000;
 
         // Test that fill_order properly clears the prev pointer when advancing to the next order


crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -1524,78 +1525,82 @@ mod tests {
...
 
-        // Initialize base token  and mint amount
+    fn setup_test_tokens<S: PrecompileStorageProvider>(
+        storage: &mut S,
+        admin: Address,
+        user: Address,
+        exchange_address: Address,
+        amount: u128,
+    ) -> (Address, Address) {
+        // Initialize quote token (LinkingUSD)
```
fgimenez: this might need to be aligned with https://github.com/tempoxyz/tempo/pull/1015, maybe gating behind allegretto, wdyt @0xKitsune

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -802,6 +802,7 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
         } else {
             // If there are subsequent orders at tick, advance to next order
             level.head = order.next();
+            Order::update_prev_order(self, order.next(), 0)?;
```
klkvr: should this be gated behind Allegretto?

crates/precompiles/src/stablecoin_exchange/mod.rs
0xKitsune: Yes this should be aligned with 1015 which updates linkingUSD to pathUSD. I don't think we need to gate anything here since this is for tests.

crates/precompiles/src/stablecoin_exchange/mod.rs
0xKitsune: Yes this needs to be gated behind Allegretto.

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -4118,4 +4125,58 @@ mod tests {
 
         Ok(())
     }
+
+    #[test]
+    // https://github.com/tempoxyz/tempo/issues/1007
+    fn test_audit_issue_1007() -> eyre::Result<()> {
```
0xKitsune: nit: can we rename to `test_clear_order_post_allegretto` or similar? We can still link the above issue.

Issue #1012, opened 2025-11-21
Bug: New validators cannot sync due to 1MB p2p message limit conflicting with 8MB Reth blocks


0xKitsune:
### Describe the bug

Validator nodes cannot join or sync the chain because the p2p layer enforces a ~1MB message size, while Reth block payloads can be up to 8MB. This prevents block propagation to new nodes, blocking chain synchronization.

PR #1037, opened 2025-11-24
feat(p2p): increase default message limit to 8MB

shekhirin:
Closes https://github.com/tempoxyz/tempo/issues/1012

A block with 20k transactions that are a mix of DEX and TIP-20 transfers takes 3-4MB. Using Reth's ~8MB limit feels safe.


crates/commonware-node-config/src/p2p.rs
```diff
@@ -1,5 +1,12 @@
 //! Config options specific to the p2p layer.
 
+use reth_consensus_common::validation::MAX_RLP_BLOCK_SIZE;
+
+/// Set the maximum permitted message size to Reth's maximum RLP-encoded block size.
+///
+/// It is enforced on the builder side in `crates/payload/builder/src/lib.rs`.
+const DEFAULT_MAX_MESSAGE_SIZE_BYTES: usize = MAX_RLP_BLOCK_SIZE;
```
SuperFluffy, line 8: This const seems a bit useless? :-)

shekhirin: just to be able to add a small margin in the future if we need, also easier to see it right at the top of the file

Issue #933, opened 2025-11-14
feat: ensure DKG failures don't make validators fall out of peer set


SuperFluffy:
### Describe the feature

Context: https://github.com/tempoxyz/tempo/pull/585#discussion_r2519985511

@clabby 
> It looks like we're using the epoch as the block height for read_validator_config_from_contract - is this intentional? it seems like this will return the wrong result, just fetching the state of the contract at block heights epoch-2 & epoch-1 rather than the first block height of each epoch. (Seeing that internally to read_validator_config_from_contract, the third argument is passed directly to block_by_number.)

> Does the contract handle failures in the ceremony? It looks like we're directly loading the state from the contract, but that state isn't persisted by the protocol. What happens if e.g. the ceremony fails 3 times in a row? Wouldn't we read the wrong set of dealers/players?

@SuperFluffy 
> Re 2: Good catch: the outcome of the public polynomial is persisted in the boundary height of the ending epoch. I.e. boundary(E-1) contains the outcome of ceremony(E-1) upon entering epoch E. This means the following:
> if the ceremony fails 3 times, the participants of the ceremony E-3 will be used in epoch E. This means that the participants can have fallen out of the set.
> This would be gracefully addressed by tracking the actual participants/peers in https://github.com/tempoxyz/tempo/issues/883

Issue #930, opened 2025-11-14
consensus: reject duplicate submissions of dkg dealings


SuperFluffy:
May at some point be useful to reject duplicate submissions but that would require tracking which dealings are in the pending ancestry

_Originally posted by @patrick-ogrady in https://github.com/tempoxyz/tempo/pull/585#discussion_r2516318186_

Issue #928, opened 2025-11-14
consensus: enforce upper bound on number of possible validators


SuperFluffy:
### Describe the feature

Context: https://github.com/tempoxyz/tempo/pull/585#discussion_r2519877922

> Unrealistic that > `u16::MAX` players could be registered, but should be an invalid state (and prevented within the contract.)

Issue #924, opened 2025-11-14
consensus: finalize DKG ceremony speculatively


SuperFluffy:
### Describe the feature

Finalization of DKG ceremonies should happen speculatively and only on request from the application. Further, this construction should not end the ceremony: in case validators do not agree on the same outcome, this leaves the option to process more finalized blocks in the meantime.

Right now, the DKG ceremony is finalized on the pre-to-last finalized block of an epoch to be included in the last block of that epoch. This is problematic: when constructing a proposal on top of some parent, simplex only guarantees that the parent is notarized, not finalized. 

Related: #565, https://github.com/commonwarexyz/monorepo/pull/2214

PR #1039, opened 2025-11-24
feat(consensus): dynamic validator sets

### LINKED ISSUE #310: Investigate merging tempo and reth datadirs

### LINKED ISSUE #563: Store epoch length, initial BLS key, initial validator set public keys in chainspec

### LINKED ISSUE #883: consensus: write full peer set for an epoch to chain

SuperFluffy:
This patch allows dynamically adding and removing validators from Tempo by using the `ValidatorConfig` smart contract. The consensus layer is hardfork enabled, meaning that it only transitions to the new behavior after `allegrettoTime`.

There is a high level overview of the protocol provided in `docs/network-reconfiguration.md` provided with this patch.

### Configuration

1. consensus config was removed in favor of command line arguments, see `tempo node --help` and `crates/commonware-node/src/args.rs`
2 the public polynomial was moved to chainspec `"publicPolynomial"` entry
3. the initial set of peers was moved to chainspec `"validators"` entry
4. epoch length was moved to chainspec `"epochLength"` entry

### Adding and removing validators

1. where before `dealers == players` in each ceremony (and where both were defined through chainspec), the players in a new ceremony are now established by doing a `getValidators` call of the `ValidatorConfig` smart contract (technically, the validators returned by the contract call will be the players in the ceremony of the epoch after, giving them time to sync for one entire epoch; see the spec provided with this patch).
2. the p2p peers are similarly established through `getValidators`, where we track the peers of the last 3 epochs (current, parent, grandparent). The peers in epoch N is the union of the peers of `N`, `N+1` and `N+2`.
3. validators are removed by using the `changeValidatorStatus` smart contract call setting `active = false`. Similar to how adding validators work, this will remove the validator from first the `syncing players` list, and then from the `players` of the ceremony in the epoch after the next.

### DKG ceremonies and hardfork behavior

1. at startup, DKG ceremonies are run with the validator set specified in the chainspec. This matches the behavior before this patch (except that peers were provided through a config file)
2. when a new DKG ceremony is started (on the last height of the epoch) and if `boundary_block.timestamp >= allegretto_time`, the DKG manager reads the entries of the `ValidatorConfig` smart contract
4. if the entries match exactly the validators in the chainspec (i.e. `chainspec.validators == contract.validators` with both being a list `[(pubkey:socketaddr)]`), then a DKG ceremony according to post-hardfork logic is started.
5. from then on, validators can be added and removed by using the `ValidatorConfig` smart contract. 
6. in all other cases (timestamp is before allegretto time, the contract validators don't match), the pre-hardfork logic stays in place until the next epoch. This gives the operator time to fix the onchain contract state and provide the correct validator data.

Note 1: this requires an operator (admin account) to issue explicit `addValidator` calls to add all validators to the contract state.
Note 2: at the moment, a node always starts out in the pre-hardfork logic even if allegretto time is 0. This will be fixed in a followup.

### Xtasks

1. update xtask genesis generator to generate and inject validators, public polynomial into genesis/chainspec
2. rework the `generate-genesis` task by breaking out genesis-generation into its own set of arguments
3. rework the `generate-devnet` task in terms of this this new genesis-generation
4. remove `generate-config`

### e2e testing

1. rework e2e tests to get access to configs, execute smart contract calls from within the execution/tokio runtime
2. provide an e2e test for adding and removing a validator, a validator having lost its share and regaining it in a subsequent round, transition from pre- to post-allegretto logic

## Followup issues

https://github.com/tempoxyz/tempo/issues/562
https://github.com/tempoxyz/tempo/issues/565
https://github.com/tempoxyz/tempo/issues/792
https://github.com/tempoxyz/tempo/issues/884
https://github.com/tempoxyz/tempo/issues/924
https://github.com/tempoxyz/tempo/issues/928
https://github.com/tempoxyz/tempo/issues/930
https://github.com/tempoxyz/tempo/issues/933
https://github.com/tempoxyz/tempo/issues/1087

Closes #310 
Closes #563 
Closes #883

### DIFF
--- a/crates/commonware-node/src/dkg/manager/actor/pre_allegretto.rs
+++ b/crates/commonware-node/src/dkg/manager/actor/pre_allegretto.rs
@@ -14,7 +14,7 @@ use commonware_utils::{
     sequence::U64,
     set::{Ordered, OrderedAssociated},
 };
-use eyre::{WrapErr as _, ensure};
+use eyre::WrapErr as _;
 use rand_core::CryptoRngCore;
 use tempo_chainspec::hardfork::TempoHardforks;
 use tempo_dkg_onchain_artifacts::PublicOutcome;
@@ -32,8 +32,6 @@ use crate::{
 
 const CURRENT_EPOCH_KEY: U64 = U64::new(0);
 const PREVIOUS_EPOCH_KEY: U64 = U64::new(1);
-
-const CEREMONY_SUCCESS_KEY: U64 = U64::new(0);
 
 impl<TContext, TPeerManager> super::Actor<TContext, TPeerManager>
 where
@@ -283,11 +281,6 @@ where
 
         let ceremony_outcome = match ceremony.finalize() {
             Ok(outcome) => {
-                self.pre_allegretto_metadatas
-                    .ceremony_success_metadata
-                    .put_sync(CEREMONY_SUCCESS_KEY, true)
-                    .await
-                    .expect("must always be able to persist state");
                 self.metrics.ceremony_successes.inc();
                 info!(
                     "ceremony was successful; using the new participants, polynomial and secret key"
@@ -295,11 +288,6 @@ where
                 outcome
             }
             Err(outcome) => {
-                self.pre_allegretto_metadatas
-                    .ceremony_success_metadata
-                    .put_sync(CEREMONY_SUCCESS_KEY, false)
-                    .await
-                    .expect("must always be able to persist state");
                 self.metrics.ceremony_failures.inc();
                 warn!(
                     "ceremony was a failure; using the old participants, polynomial and secret key"
@@ -417,16 +405,6 @@ where
         TReceiver: Receiver<PublicKey = PublicKey>,
         TSender: Sender<PublicKey = PublicKey>,
     {
-        ensure!(
-            self.pre_allegretto_metadatas
-                .ceremony_success_metadata
-                .get(&CEREMONY_SUCCESS_KEY)
-                .copied()
-                .unwrap_or_default(),
-            "last DKG ceremony was not successful; waiting another epoch before \
-            transitioning"
-        );
-
         let epoch_state = self
             .pre_allegretto_metadatas
             .epoch_metadata
@@ -453,11 +431,6 @@ where
     /// Persisted information on the current epoch for DKG ceremonies that were
     /// started after the allegretto hardfork.
     epoch_metadata: Metadata<TContext, U64, EpochState>,
-
-    /// Persisted information on whether the last ceremony was successful or not.
-    /// The network transitions to dynamic validator sets only if the last
-    /// ceremony was successful.
-    ceremony_success_metadata: Metadata<TContext, U64, bool>,
 }
 
 impl<TContext> Metadatas<TContext>
@@ -480,22 +453,7 @@ where
         .await
         .expect("must be able to initialize metadata on disk to function");
 
-        let ceremony_success_metadata = Metadata::init(
-            context.with_label("post_allegretto_ceremony_success_metadata"),
-            commonware_storage::metadata::Config {
-                // XXX: the prefix of this partition must stay fixed to be
-                // backward compatible with the pre-allegretto hardfork.
-                partition: format!("{partition_prefix}_ceremony_success"),
-                codec_config: (),
-            },
-        )
-        .await
-        .expect("must be able to initialize metadata on disk to function");
-
-        Self {
-            epoch_metadata,
-            ceremony_success_metadata,
-        }
+        Self { epoch_metadata }
     }
 
     pub(super) fn dkg_outcome(&self) -> Option<PublicOutcome> {


crates/commonware-node/src/dkg/manager/actor/post_allegretto.rs
```diff
@@ -0,0 +1,599 @@
...
+        ceremony::{self, Ceremony},
+        manager::{
+            actor::{DkgOutcome, pre_allegretto},
+            validators::ValidatorState,
+        },
+    },
+    epoch::{self, is_first_block_in_epoch},
+};
+
+const CURRENT_EPOCH_KEY: U64 = U64::new(0);
```
fgimenez, line 34: these constants are also defined in pre_allegreto.rs, could be moved one level up

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -0,0 +1,763 @@
...
+    /// The actor configuration passed in when constructing the actor.
+    config: super::Config<TPeerManager>,
+
+    /// The runtime context passed in when constructing the actor.
+    context: ContextCell<TContext>,
+
+    /// The channel over which the actor will receive messages.
+    mailbox: mpsc::UnboundedReceiver<super::Message>,
+
+    /// Persisted information on the currently running ceremmony and its
```
fgimenez:
```suggestion
    /// Persisted information on the currently running ceremony and its
```

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -0,0 +1,763 @@
...
+
+    /// Persisted information on the current epoch for DKG ceremonies that were
+    /// started after the allegretto hardfork.
+    post_allegretto_metadatas: post_allegretto::Metadatas<ContextCell<TContext>>,
+
+    /// Persisted information on the current epoch for DKG ceremonies that were
+    /// started before the allegretto hardfork.
+    pre_allegretto_metadatas: pre_allegretto::Metadatas<ContextCell<TContext>>,
+
+    /// Information on the peers registered on the p2p peer mamnager for a given
```
fgimenez:
```suggestion
    /// Information on the peers registered on the p2p peer manager for a given
```

crates/commonware-node/src/dkg/manager/actor/post_allegretto.rs
```diff
@@ -0,0 +1,599 @@
...
+
+        // Remove it?
+        let dkg_outcome = self
+            .post_allegretto_metadatas
+            .dkg_outcome_metadata
+            .get(&DKG_OUTCOME_KEY.into())
+            .cloned()
+            .expect(
+                "when updating the current epoch state, there must be a DKG \
+                outcome of some ceremomny",
```
fgimenez:
```suggestion
                outcome of some ceremony",
```

crates/commonware-node/src/args.rs
```diff
@@ -0,0 +1,118 @@
+//! Command line arguments for configuring the consensus layer of a tempo node.
+use std::{net::SocketAddr, path::PathBuf};
+
+/// Arguments for configuring the consensus layer of a tempo node.
```
fgimenez:
```suggestion
```

crates/commonware-node/src/dkg/manager/actor/post_allegretto.rs
```diff
@@ -0,0 +1,599 @@
...
+                    .expect("metadata must always be writable");
+            }
+        }
+
+        let mut ceremony = maybe_ceremony.take().expect(
+            "past this point a ceremony must always be defined; the only \
+                time a ceremony is not permitted to exist is exactly on the \
+                boundary; did the code after ensure that the ceremony is \
+                returned to its Option?",
+        );
```
fgimenez, line 146: can this expect panic? in other words, does the code after ensure that the ceremony is returned to its Option?

crates/commonware-node/src/dkg/manager/validators.rs
```diff
@@ -0,0 +1,456 @@
...
+    /// [`commonware_p2p::authenticated::lookup`]). If the inbound value is a
+    /// socket address, then the conversion is immediate. If is a domain name,
+    /// the domain name is resolved. If DNS resolution returns more than 1 value,
+    /// the last one is taken.
+    #[instrument(skip_all, fields(public_key = %self.public_key, inbound = %self.inbound), err)]
+    fn inbound_to_socket_addr(&self) -> eyre::Result<SocketAddr> {
+        let all_addrs = self
+            .inbound
+            .to_string()
+            .to_socket_addrs()
```
fgimenez: this is in the hot path when registering peers right? when using hostnames for peers this will involve dns lookups, could this be a problem?

crates/commonware-node/src/dkg/manager/actor/post_allegretto.rs
SuperFluffy: They share the same names and values but I want them to be confined (module-private) and not leak out.

crates/commonware-node/src/dkg/manager/validators.rs
SuperFluffy: It's in the hotpath, but finalization cannot conclude until the hostnames are resolved.

You are right though: this should be performed in a blocking task/extra thread.

crates/commonware-node/src/dkg/manager/actor/post_allegretto.rs
SuperFluffy: Yeah, see below.

It's very manual - I attempted to work around this with a guard, but the UX of that was awful.

crates/commonware-node/src/dkg/manager/actor/post_allegretto.rs
fgimenez: sure, but they could be defined once in `crates/commonware-node/src/dkg/manager/actor/mod.rs` and still could be module private and be accessible pre and post allegretto?

crates/commonware-node/src/dkg/manager/actor/post_allegretto.rs
fgimenez: great thx, could we update the expect message to make this clear?

crates/commonware-node/src/dkg/manager/actor/post_allegretto.rs
```diff
@@ -0,0 +1,600 @@
...
+        //
+        // This relies on the fact that the actor updates its tracked epoch
+        // state on the last block of the epoch.
+        if block_epoch != self.current_epoch_state().epoch() {
+            info!(
+                block_epoch,
+                actor_epoch = self.current_epoch_state().epoch(),
+                "block was for an epoch other than what the actor is currently tracking; ignoring",
+            );
+        }
```
joshieDo, line 106: should we return early here ?

SuperFluffy: Yes, good catch! I need to add a restart test for conditions post-allegretto && shutdown right after boundary-block finalization.

fad16919

crates/commonware-node/src/dkg/manager/actor/post_allegretto.rs
```diff
@@ -0,0 +1,600 @@
...
+                    participants: ceremony_outcome.participants,
+                    public,
+                    share,
+                },
+            )
+            .await
+            .expect("must always be able to persist the DKG outcome");
+
+        // Prune older ceremony.
+        if let Some(epoch) = current_epoch.checked_sub(1) {
```
joshieDo, line 217: we used to prune -2, any special reason on why we prune the previous one now?

crates/commonware-node/src/dkg/manager/actor/post_allegretto.rs
```diff
@@ -0,0 +1,600 @@
...
+            epoch::RelativePosition::FirstHalf => {
+                let _ = ceremony.distribute_shares().await;
+                let _ = ceremony.process_messages().await;
+            }
+            epoch::RelativePosition::Middle => {
+                let _ = ceremony.process_messages().await;
+                let _ = ceremony.construct_intermediate_outcome().await;
+            }
+            epoch::RelativePosition::SecondHalf => {
+                let _ = ceremony.process_dealings_in_block(&block).await;
```
joshieDo: why don't we stop here if there's an error?

can we `inspect_err`  and log if we're not planning to exit here

crates/commonware-node/src/dkg/manager/actor/post_allegretto.rs
SuperFluffy: Yeah, there is no point in keeping it around: we are at the end of of epoch `E`, having just finalized the ceremony that will give us the state for epoch `E+1`. There is nothing to do with the ceremony for `E-1`. I think technically we can even remove the ceremony for `E`.

The reason I shifted it here is to ensure that the ceremony for pre-allegretto is truly gone.

crates/commonware-node/src/dkg/manager/actor/post_allegretto.rs
SuperFluffy: We do provide an error event. The event is emitted in the scope of `process_dealings_in_block` where it belongs, see` #[instrument(err)]`

That's an unfortunate wart of how tracing works in Rust. :-( After the `process_dealings_in_block` ends, I have no way of adding an event to its scope.

crates/commonware-node/src/dkg/manager/validators.rs
```diff
@@ -0,0 +1,456 @@
...
+        // IMPORTANT: Starting with the syncing players to ensure that the
+        // latest address for a validator with a given pubkey is used.
+        // OrderedAssociated takes the first instance of a key it sees and
+        // drops the later instances.
+        self.syncing_players()
+            .iter_pairs()
+            .chain(self.players().iter_pairs())
+            .chain(self.dealers().iter_pairs())
+            .filter_map(|(pubkey, validator)| {
+                let addr = validator.inbound_to_socket_addr().ok()?;
```
joshieDo: any DNS issue will result in us ignoring peers with domain names. unsure if this is truly desireable

SuperFluffy: Yes, that is a problem. I don't really know how to address this. Worse, the DNS problems can be localized to a single node.

Should we just spin on errors until all validators can be resolved?

SuperFluffy: To followup: note that the peers are only not registered on the peer manager (so we will not talk to them). We still track them and we still consider them potential players in a ceremony.

crates/commonware-node/src/args.rs
```diff
@@ -0,0 +1,118 @@
...
+    #[arg(long = "consensus.signing-share")]
+    pub signing_share: Option<PathBuf>,
+
+    /// The TOML formatted file containing the bls12-381 threshold public key
+    /// polynomial and the list of peers.
+    #[arg(
+        long = "consensus.peers-and-public-polynomial",
+        required_unless_present_any = ["follow", "dev"],
+    )]
+    pub peers_and_public_polynomial: Option<PathBuf>,
```
klkvr: could this not be part of chainspec? also seems that we can keep that in genesis? just make the rule something like `last block of last non-Allegretto epoch must contain the dkg outcome`

that way for chains that were started before Allegretto we will just read it at transition

crates/commonware-node/src/dkg/ceremony/persisted.rs
```diff
@@ -15,7 +15,7 @@ use super::IntermediateOutcome;
 /// Information on a ceremony that is persisted to disk.
 #[derive(Clone, Default, Debug, PartialEq, Eq)]
 pub(in crate::dkg) struct State {
-    pub(super) num_players: u64,
+    pub(super) num_players: u16,
```
joshieDo, line 18: is this change backwards compatible?

joshieDo: ok yes it is, its varint encoding

crates/commonware-node/src/dkg/manager/validators.rs
```diff
@@ -0,0 +1,452 @@
...
+            inboundAddress,
+            outboundAddress,
+            ..
+        }: IValidatorConfig::Validator,
+    ) -> eyre::Result<Self> {
+        ensure!(
+            active,
+            "field `active` is set to false; this method should only be called \
+            for active validators"
+        );
```
joshieDo: unsafe, decoding shouldnt do any checks imo

SuperFluffy: 8962ca13, good point. Removed also because it's redundant (filtering is done one level up).

crates/commonware-node/src/dkg/ceremony/persisted.rs
SuperFluffy: yes :-)

crates/commonware-node/src/args.rs
SuperFluffy: All in chainspec as per [e307882](https://github.com/tempoxyz/tempo/pull/1039/commits/e307882c4ac131a54555654194a216702332535b)

I made it optional because writing some default values is pretty annoying.

crates/commonware-node/src/dkg/manager/validators.rs
SuperFluffy: This resolved itself: we no longer have hostnames in the inbound addresses as per [c9f055e](https://github.com/tempoxyz/tempo/pull/1039/commits/c9f055eca950f58186d0803cc1efd70a6f93911a) and  f9608776

crates/commonware-node/src/dkg/manager/validators.rs
SuperFluffy: Inbound addresses must be socket addresses as per [c9f055e](https://github.com/tempoxyz/tempo/pull/1039/commits/c9f055eca950f58186d0803cc1efd70a6f93911a) and  f9608776

crates/commonware-node/src/dkg/manager/actor/post_allegretto.rs
```diff
@@ -0,0 +1,609 @@
...
+                static validators = {static_validators:?}; \
+                on chain validators = {on_chain_validators:?}",
+            );
+        }
+
+        // NOTE: it would be good to compare the static validators addresses to
+        // what was read from the contract. But this would mean that some nodes
+        // transition to the dynamic validator regime while others don't,
+        // depending on how they are launched. Only the ed25519 keys are
+        // guaranteed to be the same across nodes.
```
joshieDo: is this whats happening above?

SuperFluffy: ty, added that check yesterdat and forgot the comment below it.  33bcf9c7

crates/dkg-onchain-artifacts/src/lib.rs
```diff
@@ -0,0 +1,562 @@
...
+
+    fn read_cfg(
+        buf: &mut impl bytes::Buf,
+        _cfg: &Self::Cfg,
+    ) -> Result<Self, commonware_codec::Error> {
+        let n_players: u16 = UInt::read(buf)?.into();
+        let dealer = PublicKey::read(buf)?;
+        let dealer_signature = Signature::read(buf)?;
+        let epoch = UInt::read(buf)?.into();
+        let commitment = Public::<MinSig>::read_cfg(buf, &(quorum(n_players as u32) as usize))?;
```
joshieDo, line 386: If this is where IntermediateOutcome::decode ends up (i think?), and if a malicious validator builds this IntermediateOutcome with n_players == 0 , this will panic and disconnect every validator from the network

joshieDo: https://github.com/commonwarexyz/monorepo/blob/1ffaa1e6fb7c683b8b0ccc22adc8fe341499c24d/utils/src/lib.rs#L115-L118

joshieDo: need to validate n_players before we try the rest

crates/dkg-onchain-artifacts/src/lib.rs
```diff
@@ -0,0 +1,562 @@
...
+        self.public.write(buf);
+    }
+}
+
+impl Read for PublicOutcome {
+    type Cfg = ();
+
+    fn read_cfg(buf: &mut impl Buf, _cfg: &Self::Cfg) -> Result<Self, commonware_codec::Error> {
+        let epoch = UInt::read(buf)?.into();
+        let participants = Ordered::read_cfg(buf, &(RangeCfg::from(0..=usize::MAX), ()))?;
```
joshieDo: this needs to be capped, or else a malicious validator might craft an outcome that will try to allocate >130GB of memory

https://github.com/commonwarexyz/monorepo/blob/1ffaa1e6fb7c683b8b0ccc22adc8fe341499c24d/codec/src/types/vec.rs#L47-L48

joshieDo: we should review any other usage of `RangeCfg::from(0..=usize::MAX)` that would be passed by other validators

crates/dkg-onchain-artifacts/src/lib.rs
```diff
@@ -0,0 +1,562 @@
...
+        let public =
+            Public::<MinSig>::read_cfg(buf, &(quorum(participants.len() as u32) as usize))?;
```
joshieDo, line 124: same issue as IntermediateOutcome, if participants is 0, this will panic every validator

crates/dkg-onchain-artifacts/src/lib.rs
SuperFluffy: Capped this value at `u16::MAX`, still big, but manageable: ef1120b5

crates/dkg-onchain-artifacts/src/lib.rs
SuperFluffy: Enforcing that `n_players != 0` now: ef1120b5

crates/dkg-onchain-artifacts/src/lib.rs
SuperFluffy: Good catch: this is enforcing that `n_players != 0`: ef1120b5

crates/dkg-onchain-artifacts/src/lib.rs
SuperFluffy: Finally fixed in [6610d25](https://github.com/tempoxyz/tempo/pull/1039/commits/6610d259d2908bcb689746f69f4358f9b1f865e3) - we now enforce that `RangeCfg::from(1..u16::MAX)` for the number of participants.

crates/commonware-node/src/dkg/manager/validators.rs
```diff
@@ -0,0 +1,415 @@
...
+/// The inbound and outbound addresses stored herein are guaranteed to be of the
+/// form `<host>:<port>` for inbound, and `<ip>:<port>` for outbound. Here,
+/// `<host>` is either an IPv4 or IPV6 address, or a fully qualified domain name.
+/// `<ip>` is an IPv4 or IPv6 address.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub(super) struct DecodedValidator {
+    pub(super) public_key: PublicKey,
+    pub(super) inbound: SocketAddr,
+    pub(super) outbound: SocketAddr,
+    pub(super) index: u64,
```
joshieDo, line 286: this index is not being used anywhere right? can we just remove it? Initially i was confused if this was somehow [wrongly] used in the ceremony.

SuperFluffy: It's not used, but I am displaying the validators in logs (see the `Display` impl) to show what exactly was read

joshieDo: okay probably cant remove it otherwise the new binary will have a bad time, can we just document this

joshieDo:
```suggestion
    /// Contract's internal array position; unused in node logic.
    pub(super) index: u64,
```

SuperFluffy: I exhaustively commented all the fields. :-)

PR #1040, opened 2025-11-24
fix(mempool): reject invalid AA txs to prevent mempool DOS

### LINKED ISSUE #1009: Bug: Tx Pool can be DOSed via AA txs
### Describe the bug

The tx pool can be permanently DOSed with AA txs. AA txs with far future `valid_after` timestamps permanently occupy mempool, and will never be executed or evicted

0xrusowsky:
## Motivation

closes #1009 

## Solution

as proposed in the issue discussion, we reject txs which have a `valid_after` beyond a `const AA_VALID_AFTER_MAX_SECS`.

## Questions

- is 1h a good value?

### DIFF
--- a/crates/precompiles/src/nonce/mod.rs
+++ b/crates/precompiles/src/nonce/mod.rs
@@ -67,7 +67,6 @@ impl<'a, S: PrecompileStorageProvider> NonceManager<'a, S> {
     /// Internal: Increment nonce for a specific account and nonce key
     pub fn increment_nonce(&mut self, account: Address, nonce_key: U256) -> Result<u64> {
         if nonce_key == 0 {
-            // TODO: Should this be a different error?
             return Err(NonceError::invalid_nonce_key().into());
         }
 
@@ -83,6 +82,18 @@ impl<'a, S: PrecompileStorageProvider> NonceManager<'a, S> {
             .ok_or_else(NonceError::nonce_overflow)?;
 
         self.sstore_nonces(account, nonce_key, new_nonce)?;
+
+        if self.storage.spec().is_allegretto() {
+            self.storage.emit_event(
+                self.address,
+                NonceEvent::NonceIncremented(INonce::NonceIncremented {
+                    account,
+                    nonceKey: nonce_key,
+                    newNonce: new_nonce,
+                })
+                .into_log_data(),
+            )?;
+        }
 
         Ok(new_nonce)
     }
@@ -289,36 +300,70 @@ mod tests {
 
     #[test]
     fn test_active_key_count_event_not_emitted_pre_moderato() {
-        // Test with Adagio (pre-Moderato) - event should NOT be emitted
         let mut storage = HashMapStorageProvider::new(1).with_spec(TempoHardfork::Adagio);
         let account = address!("0x1111111111111111111111111111111111111111");
         let nonce_key = U256::from(5);
 
-        // First increment should NOT emit ActiveKeyCountChanged event pre-Moderato
         {
             let mut nonce_mgr = NonceManager::new(&mut storage);
             nonce_mgr.increment_nonce(account, nonce_key).unwrap();
         }
 
-        // Check that NO events were emitted
         let events = storage.events.get(&NONCE_PRECOMPILE_ADDRESS);
         assert!(
             events.is_none() || events.unwrap().is_empty(),
             "No events should be emitted pre-Moderato"
         );
 
-        // Increment on different key should also NOT emit event
         let nonce_key2 = U256::from(10);
         {
             let mut nonce_mgr = NonceManager::new(&mut storage);
             nonce_mgr.increment_nonce(account, nonce_key2).unwrap();
         }
 
-        // Still no events
         let events = storage.events.get(&NONCE_PRECOMPILE_ADDRESS);
         assert!(
             events.is_none() || events.unwrap().is_empty(),
             "No events should be emitted pre-Moderato"
         );
+    }
+
+    #[test]
+    fn test_increment_nonce_post_allegretto() {
+        let mut storage = HashMapStorageProvider::new(1).with_spec(TempoHardfork::Allegretto);
+        let account = address!("0x1111111111111111111111111111111111111111");
+        let nonce_key = U256::from(5);
+
+        {
+            let mut nonce_mgr = NonceManager::new(&mut storage);
+            nonce_mgr.increment_nonce(account, nonce_key).unwrap();
+        }
+
+        let events = storage
+            .events
+            .get(&NONCE_PRECOMPILE_ADDRESS)
+            .expect("Could not get events");
+        assert_eq!(events.len(), 2,);
+
+        let expected_nonce_event = NonceEvent::NonceIncremented(INonce::NonceIncremented {
+            account,
+            nonceKey: nonce_key,
+            newNonce: 1,
+        });
+        assert_eq!(events[1], expected_nonce_event.into_log_data());
+
+        {
+            let mut nonce_mgr = NonceManager::new(&mut storage);
+            nonce_mgr.increment_nonce(account, nonce_key).unwrap();
+        }
+        let events = storage.events.get(&NONCE_PRECOMPILE_ADDRESS).unwrap();
+        assert_eq!(events.len(), 3);
+
+        let expected_nonce_event2 = NonceEvent::NonceIncremented(INonce::NonceIncremented {
+            account,
+            nonceKey: nonce_key,
+            newNonce: 2,
+        });
+        assert_eq!(events[2], expected_nonce_event2.into_log_data());
     }
 }


crates/transaction-pool/src/validator.rs
```diff
@@ -11,6 +11,10 @@ use reth_transaction_pool::{
 };
 use tempo_revm::TempoStateAccess;
 
+// Reject AA transactions where `valid_after` is too far in the future to prevent mempool DOS.
+// ref: <https://github.com/tempoxyz/tempo/issues/1009>
+const AA_VALID_AFTER_MAX_SECS: u64 = 3600;
```
mattsse: this is 6 hrs?

crates/transaction-pool/src/transaction.rs
```diff
@@ -81,6 +84,7 @@ impl PoolTransactionError for TempoPoolTransactionError {
             Self::ExceedsNonPaymentLimit => false,
             Self::InvalidFeeToken(_) => false,
             Self::MissingFeeToken => false,
+            Self::InvalidValidAfter { .. } => true,
```
mattsse: this should also be false here imo

because this is still a good tx technically, just not yet

crates/transaction-pool/src/validator.rs
0xrusowsky: as discussed in dms, will turn it into a configurable param

crates/transaction-pool/src/validator.rs
```diff
@@ -11,6 +13,14 @@ use reth_transaction_pool::{
 };
 use tempo_revm::TempoStateAccess;
 
+/// Maximum allowed `valid_after` offset for AA transactions (in seconds).
+static AA_VALID_AFTER_MAX_SECS: LazyLock<u64> = LazyLock::new(|| {
+    std::env::var("AA_VALID_AFTER_MAX_SECS")
```
mattsse: let's not introduce env vars for settings

this should be a field in the validator instead and can be wired to the cli via TempoArgs but we can also do that later

0xrusowsky: addressed in [`e5da853` (#1040)](https://github.com/tempoxyz/tempo/pull/1040/commits/e5da853c4512f8c4547f5d77f729fa77142cd623)

bin/tempo/src/main.rs
```diff
@@ -36,10 +36,28 @@ use tempo_faucet::{
...
+};
 use tokio::sync::oneshot;
 use tokio_util::either::Either;
 
+/// Transaction pool validator CLI arguments.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, clap::Args)]
+struct ValidatorArgs {
+    /// Maximum allowed `valid_after` offset for AA txs.
+    #[arg(long, default_value_t = DEFAULT_AA_VALID_AFTER_MAX_SECS)]
+    pub aa_valid_after_max_secs: u64,
```
mattsse: let's prefix this with "txpool." see also TxpoolArgs

bin/tempo/src/main.rs
```diff
@@ -36,10 +36,28 @@ use tempo_faucet::{
...
-use tempo_node::{TempoFullNode, node::TempoNode};
+use tempo_node::{
+    DEFAULT_AA_VALID_AFTER_MAX_SECS, TempoFullNode, ValidatorConfig, node::TempoNode,
+};
 use tokio::sync::oneshot;
 use tokio_util::either::Either;
 
+/// Transaction pool validator CLI arguments.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, clap::Args)]
+struct ValidatorArgs {
```
mattsse: we can name this TempTxPoolArgs

this should be moved below the main type and made a pub

crates/transaction-pool/src/validator.rs
```diff
@@ -61,6 +69,28 @@ where
             );
         }
 
+        // Reject AA transactions where `valid_after` is too far in the future to prevent mempool DOS.
+        if let Some(tx) = transaction.inner().as_aa()
+            && let Some(valid_after) = tx.tx().valid_after
+        {
+            let current_time = std::time::SystemTime::now()
```
klkvr: we could also use `self.inner.fork_tracker().tip_timestamp()` to get the latest block timestamp, not sure what's better here

crates/transaction-pool/src/validator.rs
```diff
@@ -61,6 +69,28 @@ where
...
+                .duration_since(std::time::UNIX_EPOCH)
+                .expect("system time before UNIX EPOCH")
+                .as_secs();
+            let max_allowed = current_time.saturating_add(self.aa_valid_after_max_secs);
+            if valid_after > max_allowed {
```
klkvr, line 215: I think we also need to check `valid_before` here

0xrusowsky: yup, this is addressed in #1055

bin/tempo/src/main.rs
0xrusowsky: addressed in [`9bb86d6` (#1040)](https://github.com/tempoxyz/tempo/pull/1040/commits/9bb86d6ffe147dd0512acee1a469d9df8d0ca057)

bin/tempo/src/main.rs
0xrusowsky: addressed in [`9bb86d6` (#1040)](https://github.com/tempoxyz/tempo/pull/1040/commits/9bb86d6ffe147dd0512acee1a469d9df8d0ca057)

crates/transaction-pool/src/validator.rs
0xrusowsky: @mattsse thoughts?

crates/node/src/node.rs
```diff
@@ -41,19 +41,42 @@ use tempo_payload_types::TempoPayloadAttributes;
...
 
+/// Default maximum allowed `valid_after` offset for AA txs (1 hour).
+pub const DEFAULT_AA_VALID_AFTER_MAX_SECS: u64 = 3600;
+
+/// Validator configuration.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct ValidatorConfig {
+    /// Maximum allowed `valid_after` offset for AA txs.
+    pub aa_valid_after_max_secs: u64,
+}
```
0xKitsune: Do we need a separate config for this? Or can this simply be a field on the `TempoPoolBuilder`?

crates/transaction-pool/src/validator.rs
0xKitsune: This is probably fine in most cases IMO we should use `tip_timestamp()` as we can not assume `SystemTime::now()` is monotonic. It can jump backward/forward due to NTP, VM clock drift, etc. and hit a panic due to the expect here.

crates/node/src/node.rs
0xrusowsky: will integrate on `TempoPoolBuilder`, as that will align better with the changes that mattsse requested for the CLI inputs

0xrusowsky: address with [`fe5d007` (#1040)](https://github.com/tempoxyz/tempo/pull/1040/commits/fe5d00761ce5cd5b2be6f360c66e840634461ead)

crates/transaction-pool/src/validator.rs
0xrusowsky: addressed with [`edc0252` (#1040)](https://github.com/tempoxyz/tempo/pull/1040/commits/edc0252e1ff66df411682ff29026abceff57e097)

crates/transaction-pool/src/maintain.rs
```diff
@@ -0,0 +1,81 @@
...
+use reth_transaction_pool::{PoolTransaction, TransactionPool};
+use std::collections::BTreeMap;
+use tracing::debug;
+
+/// Spawns a background task that evicts expired AA transactions.
+///
+/// - Listens for new blocks to get chain timestamp
+/// - Listens for new transactions to track `valid_before` timestamps
+/// - Evicts transactions when `valid_before <= tip_timestamp`
+pub async fn evict_expired_aa_txs<P, C>(pool: P, client: C)
```
klkvr, line 23: seems that when we spawn this we're racing with backup task which re-inserts all txs from disk into the pool

this means that if backup txs will be re-inserted before this is spawned, we will miss some of them

I guess what we could do here is start by loading all txs from the pool and inserting them into the `expiry_map`

0xrusowsky: i see, makes sense. will fix it

0xrusowsky: addressed in [`f4575e6` (#1040)](https://github.com/tempoxyz/tempo/pull/1040/commits/f4575e6b6196c6a56d76043e54184704632c0bb7)

bin/tempo/src/main.rs
```diff
@@ -188,7 +208,7 @@ fn main() -> eyre::Result<()> {
             node,
             node_exit_future,
         } = builder
-            .node(TempoNode::new())
+            .node(TempoNode::new(args.tx_pool_args.into()))
```
mattsse: this conversion will lead do some additional refactor work in case we need more settings not limited to the pool, imo it'd be easier to just propagate the settings/args here

crates/node/src/node.rs
```diff
@@ -41,19 +41,27 @@ use tempo_payload_types::TempoPayloadAttributes;
...
-    pub const fn new() -> Self {
-        Self
+    pub fn new(tx_pool_config: TempoPoolBuilder) -> Self {
+        Self { tx_pool_config }
     }
 
     /// Returns a [`ComponentsBuilder`] configured for a regular Tempo node.
-    pub fn components<Node>() -> ComponentsBuilder<
+    pub fn components<Node>(
+        tx_pool_config: TempoPoolBuilder,
```
mattsse: this is misnamed

crates/transaction-pool/src/tasks.rs
```diff
@@ -0,0 +1,92 @@
...
+                    .range(..=tip_timestamp)
+                    .map(|(ts, _)| *ts)
+                    .collect();
+
+                let mut to_remove = Vec::new();
+                for ts in expired {
+                    if let Some(hashes) = expiry_map.remove(&ts) {
+                        to_remove.extend(hashes);
+                    }
+                }
```
mattsse: I believe you can simplify this with a

```
while let Some(entry) = expiry_map.first_entry() {

}
```

crates/transaction-pool/src/validator.rs
```diff
@@ -466,4 +538,58 @@ mod tests {
...
+        let tx_no_valid_after = create_aa_transaction(None);
+        let validator = setup_validator(&tx_no_valid_after);
+        let outcome = validator
+            .validate_transaction(TransactionOrigin::External, tx_no_valid_after)
+            .await;
+
+        if let TransactionValidationOutcome::Invalid(_, err) = outcome {
+            let error_msg = format!("{err}");
+            assert!(
+                !error_msg.contains("valid_after timestamp")
```
fgimenez: is this string right? `valid_after` should be included but `timestamp`? if this string is wrong, then the assertion is always true `assert!(!false || anything) == assert!(true || anything) == ok`

crates/transaction-pool/src/validator.rs
```diff
@@ -466,4 +538,58 @@ mod tests {
...
+        let tx_within_limit = create_aa_transaction(Some(current_time + 1800));
+        let validator = setup_validator(&tx_within_limit);
+        let outcome = validator
+            .validate_transaction(TransactionOrigin::External, tx_within_limit)
+            .await;
+
+        if let TransactionValidationOutcome::Invalid(_, err) = outcome {
+            let error_msg = format!("{err}");
+            assert!(
+                !error_msg.contains("valid_after timestamp")
```
fgimenez: same here

crates/transaction-pool/src/validator.rs
```diff
@@ -466,4 +538,58 @@ mod tests {
             panic!("Expected Invalid outcome with InsufficientFunds error");
         }
     }
+
+    #[tokio::test]
+    async fn test_aa_valid_after_check() {
+        let current_time = std::time::SystemTime::now()
```
fgimenez, line 577: is it ok to use `SystemTime::now()` here? the validator will use `self.inner.fork_tracker().tip_timestamp()` which returns the mock provider's tip timestamp, this might be very far away (much smaller) than system time

0xrusowsky: addressed in [`da3cc46` (#1040)](https://github.com/tempoxyz/tempo/pull/1040/commits/da3cc46fd6f25e7bcddd871967a9836d9ee57b0e)

crates/transaction-pool/src/validator.rs
0xrusowsky: addressed in [`da3cc46` (#1040)](https://github.com/tempoxyz/tempo/pull/1040/commits/da3cc46fd6f25e7bcddd871967a9836d9ee57b0e)

crates/transaction-pool/src/validator.rs
0xrusowsky: good catch, without this error tests were failing!

addressed in [`da3cc46` (#1040)](https://github.com/tempoxyz/tempo/pull/1040/commits/da3cc46fd6f25e7bcddd871967a9836d9ee57b0e)

bin/tempo/src/main.rs
0xrusowsky: addressed in [`ca4bc40` (#1040)](https://github.com/tempoxyz/tempo/pull/1040/commits/ca4bc404db5ddaae209a6bb156ef0902dc021a50)

crates/node/src/node.rs
0xrusowsky: addressed in [`ca4bc40` (#1040)](https://github.com/tempoxyz/tempo/pull/1040/commits/ca4bc404db5ddaae209a6bb156ef0902dc021a50)

crates/transaction-pool/src/tasks.rs
0xrusowsky: addressed in [`285bc3c` (#1040)](https://github.com/tempoxyz/tempo/pull/1040/commits/285bc3ca8b735c3e6df10e75b66c056f100168ed)

crates/node/src/node.rs
```diff
@@ -45,19 +45,47 @@ use tempo_transaction_pool::{
...
+/// Default maximum allowed `valid_after` offset for AA txs (1 hour).
+pub const DEFAULT_AA_VALID_AFTER_MAX_SECS: u64 = 3600;
+
+/// Tempo node CLI arguments.
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq, clap::Args)]
+#[command(next_help_heading = "TxPool")]
+pub struct TempoNodeArgs {
+    /// Maximum allowed `valid_after` offset for AA txs.
+    #[arg(long = "txpool.aa-valid-after-max-secs", default_value_t = DEFAULT_AA_VALID_AFTER_MAX_SECS)]
+    pub aa_valid_after_max_secs: u64,
```
mattsse, line 57: this is better!

crates/transaction-pool/src/maintain.rs
mattsse, line 1: nitpick, can we keep this as maintain.rs

crates/transaction-pool/src/maintain.rs
```diff
@@ -0,0 +1,133 @@
...
+    // Track valid_before timestamp -> Vec<TxHash>
+    let mut expiry_map: BTreeMap<u64, Vec<TxHash>> = BTreeMap::new();
+
+    // Populate expiry map to prevent race condition at start-up
+    pool.all_transactions()
+        .all()
+        .for_each(|tx| track_expiry(&mut expiry_map, tx.inner().as_aa()));
+
+    // Subscribe to new transactions and blocks
+    let mut new_txs = pool.new_transactions_listener();
```
mattsse, line 45: there's technically a smol race condition here, so this could be moved above the ` pool.all_transactions()` call

crates/transaction-pool/src/maintain.rs
```diff
@@ -0,0 +1,133 @@
...
+            && let Some(valid_before) = aa_tx.tx().valid_before
+        {
+            let hash = *aa_tx.hash();
+            map.entry(valid_before).or_default().push(hash);
+        }
+    };
+
+    // Track valid_before timestamp -> Vec<TxHash>
+    let mut expiry_map: BTreeMap<u64, Vec<TxHash>> = BTreeMap::new();
+
```
mattsse, line 40: I think we could sleep here for a sec or two just to give the backup task some time?

crates/transaction-pool/src/maintain.rs
```diff
@@ -0,0 +1,133 @@
...
+        if let Some(aa_tx) = maybe_aa_tx
+            && let Some(valid_before) = aa_tx.tx().valid_before
+        {
+            let hash = *aa_tx.hash();
+            map.entry(valid_before).or_default().push(hash);
+        }
+    };
+
+    // Track valid_before timestamp -> Vec<TxHash>
+    let mut expiry_map: BTreeMap<u64, Vec<TxHash>> = BTreeMap::new();
```
mattsse, line 39: I think this should be okay

crates/transaction-pool/src/maintain.rs
0xrusowsky: addressed in [`f3dd6c2` (#1040)](https://github.com/tempoxyz/tempo/pull/1040/commits/f3dd6c260134073ae78700c36eb7c2a8db010b49)

crates/transaction-pool/src/maintain.rs
0xrusowsky: addressed in [`f3dd6c2` (#1040)](https://github.com/tempoxyz/tempo/pull/1040/commits/f3dd6c260134073ae78700c36eb7c2a8db010b49)

PR #1042, opened 2025-11-24
feat(dkg): add metrics for shares, acks, and dealings

### LINKED ISSUE #792: Add metrics for DKG ceremony
Our DKG ceremony has too few metrics.

We need more metrics, for example:

1. number of shares distributed (if a dealer)
2. number of acks received (if a dealer)
3. number of acks sent (if a player)
5. number of dealings read from the blockchain

fgimenez:
Closes #792 

Added these gauges to track DKG ceremony activity (all reset to 0 on every new epoch/ceremony because total counts would not be very interesting):

- `consensus_engine_dkg_manager_ceremony_shares_distributed` - Number of shares distributed (when acting
as dealer)
- `consensus_engine_dkg_manager_ceremony_shares_received` - Number of shares received (when acting
as player)
- `consensus_engine_dkg_manager_ceremony_acks_received` - Number of acknowledgments received (when
acting as dealer)
- `consensus_engine_dkg_manager_ceremony_acks_sent` - Number of acknowledgments sent (when acting as
player)
  - `consensus_engine_dkg_manager_ceremony_dealings_read` - Number of dealings read from the blockchain
  - `consensus_engine_dkg_manager_ceremony_dealings_empty` - Blocks that contained no dealings, were empty (for long epoch, this is normal)
  - `consensus_engine_dkg_manager_ceremony_bad_dealings` - DKG dealings that could not be parsed or with invalid sigantures.

Also moved some ceremony specific metrics into the new `ceremony::Metrics` type and made `Block::try_read_ceremony_deal_outcome` a fallible operation so that it can be tracked better.


crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -147,6 +147,11 @@ where
         let post_allegretto_ceremonies = Counter::default();
         let failed_allegretto_transitions = Counter::default();
 
+        let shares_distributed = Counter::default();
```
SuperFluffy: I am wondering if it makes sense to make all these `Gauge` and reset them to 0 at the beginning of each ceremony.

The reason is that these metrics are most interesting on a per-epoch basis.

Once we cross epoch boundaries it is almost guaranteed that for example `shares_distributed != acks_received`, and so one needs to know the starting point at the beginning of an epoch to calculate how many shares actually got an ack.

crates/commonware-node/src/dkg/ceremony/mod.rs
```diff
@@ -360,6 +372,7 @@ where
                     })
                     .await
                     .expect("must be able to persists acks");
+                self.metrics.shares_distributed.inc();
```
SuperFluffy, line 379: If we do this here, then we must `acks_received + 1` and` acks_sent + 1`, otherwise it becomes weird (after all, if we "send" a share to ourselves, we also "receive" it and "return" it.)

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -669,6 +698,21 @@ struct Metrics {
     post_allegretto_ceremonies: Counter,
     failed_allegretto_transitions: Counter,
     syncing_players: Gauge,
+    shares_distributed: Counter,
```
SuperFluffy: What if you do `ceremony_metrics: CeremonyMetrics`, then you don't need to do the manual clone below.

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -669,6 +698,21 @@ struct Metrics {
...
+    acks_received: Counter,
+    acks_sent: Counter,
+    dealings_read: Counter,
+}
+
+impl Metrics {
+    fn ceremony_metrics(&self) -> CeremonyMetrics {
```
SuperFluffy: Basically a manual clone - what if you do `self.ceremony_metrics.clone()` instead (see comment above on putting `CeremonyMetrics` in instead of folding the fields into the `Metrics` superstructure.

crates/commonware-node/src/dkg/ceremony/mod.rs
```diff
@@ -67,6 +68,14 @@ pub(super) struct Config {
     pub(super) players: Ordered<PublicKey>,
 }
 
+#[derive(Clone)]
+pub(super) struct CeremonyMetrics {
```
SuperFluffy: My personal preference is to call this `Metrics` and refer to it as `ceremony::Metrics` in `manager/actor/mod.rs`

crates/e2e/src/tests/dkg.rs
```diff
@@ -122,7 +126,39 @@ fn assert_static_transitions(how_many: u32, epoch_length: u64, transitions: u64)
...
+        }
+        if metric.ends_with("_dkg_manager_dealings_read_total") {
+            let value = value.parse::<u64>().unwrap();
+            dealings_read_seen |= value > 0;
+        }
+
+        // In single-validator scenarios, acks_received and acks_sent will always be 0
+        // because there are no other validators to exchange messages with.
+        // These metrics are only incremented when receiving/sending messages over p2p.
+        let acks_condition = if how_many == 1 {
```
SuperFluffy: I mentioned it at the incremental point of `shares_distributed` inside the metrics handler: I think you should also increment these metrics to be symmetric - then you don't need to handle this special case.

crates/e2e/src/tests/dkg.rs
```diff
@@ -122,7 +126,39 @@ fn assert_static_transitions(how_many: u32, epoch_length: u64, transitions: u64)
             let value = value.parse::<u64>().unwrap();
             dkg_successful |= value >= transitions;
         }
-        epoch_reached && dkg_successful
+
+        // Verify new DKG ceremony metrics are exposed and incremented
+        if metric.ends_with("_dkg_manager_shares_distributed_total") {
+            let value = value.parse::<u64>().unwrap();
+            shares_distributed_seen |= value > 0;
```
SuperFluffy: For both this and the other metrics - should we count `shares_distributed_seen` and require that `shares_distributed_seen = how_many * how_many`? That's essentially a test that all nodes have indeed seen the shares (since we have a perfect-connectivity test this makes sense?).

crates/commonware-node/src/dkg/ceremony/mod.rs
```diff
@@ -67,6 +68,14 @@ pub(super) struct Config {
...
+    pub(super) shares_distributed: Counter,
+    pub(super) acks_received: Counter,
+    pub(super) acks_sent: Counter,
+    pub(super) dealings_read: Counter,
```
SuperFluffy: I think I would like 2 more counters here:

1. dealings that were empty (i.e. `extra_data.is_empty()`)
2. dealings that we failed to read (i.e. `!extra_data.is_empty()` but something went wrong when reading)

crates/commonware-node/src/dkg/manager/actor/mod.rs
fgimenez: yep, better `Gauge` updated and added reset calls ptal

crates/commonware-node/src/dkg/ceremony/mod.rs
fgimenez: good catch, done, this simplifies the test changes

crates/commonware-node/src/dkg/manager/actor/mod.rs
fgimenez: this is better, done

crates/commonware-node/src/dkg/manager/actor/mod.rs
fgimenez: yep already done

crates/commonware-node/src/dkg/ceremony/mod.rs
fgimenez: done, also updated the existing `Metrics` from `commonware_runtime` to `commonware_runtime::Metrics` to disambiguate

crates/e2e/src/tests/dkg.rs
fgimenez: yep already fixed

crates/e2e/src/tests/dkg.rs
fgimenez: with `shares_distributed_seen = how_many * how_many` all the test except the single node case are failing. pushed changes for `shares_distributed_seen = how_many` and this works fine ptal

crates/commonware-node/src/dkg/ceremony/mod.rs
fgimenez: makes sense done

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -147,6 +146,13 @@ where
         let post_allegretto_ceremonies = Counter::default();
         let failed_allegretto_transitions = Counter::default();
 
+        let shares_distributed = Gauge::default();
```
SuperFluffy: Last thing: move this into a constructor `ceremony::Metrics::new` so that the dkg manager does not know about this.

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -669,6 +716,19 @@ struct Metrics {
     post_allegretto_ceremonies: Counter,
     failed_allegretto_transitions: Counter,
     syncing_players: Gauge,
+    ceremony_metrics: ceremony::Metrics,
+}
+
+impl Metrics {
+    /// Resets ceremony-specific metrics to zero at the start of a new ceremony.
+    fn reset_ceremony_metrics(&self) {
```
SuperFluffy: ~This should a method of `ceremony::Metrics`, i.e. `ceremony::Metrics::reset`, and moved into `Ceremony::init` so that the DKG actor does not have to worry about that.~

Let's just re-create the metrics on every `Ceremony::init`

crates/commonware-node/src/dkg/manager/actor/post_allegretto.rs
```diff
@@ -115,6 +115,7 @@ where
         if utils::is_last_block_in_epoch(self.config.epoch_length, block.height()).is_some() {
             self.update_and_register_current_epoch_state().await;
 
+            self.metrics.reset_ceremony_metrics();
```
SuperFluffy: Move into `Ceremony::init`

crates/commonware-node/src/dkg/ceremony/mod.rs
```diff
@@ -561,8 +583,17 @@ where
         block: &Block,
         hardfork_regime: HardforkRegime,
     ) -> eyre::Result<()> {
+        // Check if extra_data is empty before trying to read
+        if block.header().extra_data().is_empty() {
+            info!("block contained no dealing (extra_data empty)");
```
SuperFluffy: This is already emitted when doing `try_read_cermony_deal_outcome`, so it's odd to also have this here as well. (also see comment below).

crates/commonware-node/src/dkg/ceremony/mod.rs
```diff
@@ -561,8 +583,17 @@ where
...
+            self.metrics.dealings_empty.inc();
+            return Ok(());
+        }
+
         let Some(block_outcome) = block.try_read_ceremony_deal_outcome() else {
-            info!("block contained no usable intermediate deal outcome");
+            // extra_data was not empty but decode failed
```
SuperFluffy: If we do this, then we need to change the signature of `try_read_cermony_deal_outcome` to be `eyre::Result<...>`.  Because otherwise this becomes weird quickly.

In fact I think we should do that. Let's change the signature of `try_read_ceremony_deal_outcome` to be `eyre::Result<...>` and remove its check for `extra_data().is_empty()` - it seems better to keep this in the actor as you have done here.

crates/commonware-node/src/dkg/ceremony/mod.rs
```diff
@@ -561,8 +583,17 @@ where
...
+            info!("block contained dealing data but failed to decode");
```
SuperFluffy: This should be `warn!`

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -207,6 +213,36 @@ where
             "how many transitions from pre- to post-allegretto failed",
             failed_allegretto_transitions.clone(),
         );
+        context.register(
```
SuperFluffy: Hm, I think this is technically the wrong context to register this on - if we register this on the ceremony context, then we'd get `dkg_manager_ceremony_<metric-name>` which is IMO better. Downside is that you'd need to register these metrics every time you init the ceremony, but I think that's fine.

In fact, let's move the metric construction to `Ceremony::init` entirely. If we keep resetting them to 0 every time a ceremony starts, we can as well just re-create the metrics.


We can do this in a followup, too: just prefix all of these metrics with `ceremony_*`, i.e. `ceremony_acks_received"`, `"ceremony_acks_sent"`

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -207,6 +213,36 @@ where
...
+            "the number of dealings read from the blockchain in the current ceremony",
+            dealings_read.clone(),
+        );
+        context.register(
+            "dealings_empty",
+            "the number of blocks with empty extra_data (no dealing) in the current ceremony",
+            dealings_empty.clone(),
+        );
+        context.register(
+            "dealings_failed",
```
SuperFluffy: This could be easily confused with "I failed to create my own dealings".

```suggestion
            "dealings_bad_decode",
```

(or with my suggestion re putting this into the `ceremony` namespace:)

```suggestion
            "ceremony_dealings_bad_decode",
```

crates/commonware-node/src/dkg/manager/actor/mod.rs
fgimenez: sure done

crates/commonware-node/src/dkg/manager/actor/mod.rs
fgimenez: good point done

crates/commonware-node/src/dkg/manager/actor/post_allegretto.rs
fgimenez: yep already done

crates/commonware-node/src/dkg/ceremony/mod.rs
fgimenez: yep removed from here

crates/commonware-node/src/dkg/ceremony/mod.rs
fgimenez: makes sense, changed to return `Result`

crates/commonware-node/src/dkg/manager/actor/mod.rs
fgimenez: sure makes total sense, done

crates/commonware-node/src/dkg/manager/actor/mod.rs
fgimenez: gone for `ceremony_` at the end

crates/commonware-node/src/dkg/ceremony/mod.rs
```diff
@@ -117,6 +181,9 @@ where
         ceremony_metadata: Arc<Mutex<Metadata<TContext, U64, State>>>,
         config: Config,
     ) -> eyre::Result<Self> {
+        // Create fresh metrics for this ceremony
+        let metrics = Metrics::new(context);
```
SuperFluffy: I think this is fundamentally the right thing to do, but I don't think this plays well with how prometheus thinks about metrics.

If I look where this ends up:

https://github.com/commonwarexyz/monorepo/blob/f6ac06a8895020492f8c4b55b333dda033bf02ce/runtime/src/tokio/runtime.rs#L556

And eventually here:

https://github.com/prometheus/client_rust/blob/82ab3c45ee04b5e1c40a4898855a25d885cbfb15/src/registry.rs#L172

So this looks like we just keep pushing these endlessly into the registry?

So over time, this just keeps growing and growing and never gets pruned.

PR #1055, opened 2025-11-25
fix(mempool): prevent short-lived aa txs from invalidating blocks

### LINKED ISSUE #1013: Bug: Subblock can be invalidated using short lived `valid_before` AA transactions
### Describe the bug

AA transactions with `valid_before` timestamps that pass initial validation but fail during block building cause the entire subblock to be discarded.

0xrusowsky:
## Motivation

closes #1013 

## Solution

similar approach as what was proposed for #1009.

invalidate AA txs which have a `valid_before` that is smaller than `const AA_VALID_BEFORE_MIN_SECS`. Require a minimum buffer of 3s. Otherwise they should simply pass `None` and aim for direct inclusion.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,25 @@
+docs/ @Zygimantass
+bin/tempo @0xKitsune @klkvr @Zygimantass @SuperFluffy 
+bin/tempo-bench @shekhirin
+bin/tempo-sidecar @Zygimantass
+contrib/ @Zygimantass
+scripts/ @0xKitsune
+xtask/ @0xKitsune @Zygimantass @SuperFluffy 
+crates/alloy/ @klkvr @onbjerg @mattsse
+crates/chainspec/ @0xKitsune @klkvr @mattsse
+crates/commonware-node/ @joshieDo @SuperFluffy
+crates/commonware-node-config/ @joshieDo @SuperFluffy
+crates/consensus/ @joshieDo @SuperFluffy
+crates/contracts/ @0xKitsune @fgimenez @klkvr @mattsse
+crates/e2e/ @joshieDo @SuperFluffy
+crates/evm/ @0xKitsune @klkvr @mattsse
+crates/eyre/ @SuperFluffy
+crates/faucet/ @klkvr @mattsse
+crates/node/ @0xKitsune @klkvr @mattsse
+crates/payload/ @0xKitsune @klkvr @mattsse
+crates/precompiles/ @0xKitsune @fgimenez @klkvr @mattsse
+crates/precompiles-macros/ @0xrusowsky @mattsse
+crates/primitives/ @klkvr @mattsse
+crates/revm/ @klkvr @mattsse @rakita @0xKitsune
+crates/telemetry-util/ @SuperFluffy
+crates/transaction-pool/ @0xKitsune @klkvr @mattsse


crates/transaction-pool/src/validator.rs
```diff
@@ -61,6 +65,29 @@ where
             );
         }
 
+        // Reject AA transactions where `valid_before` is too close to current time (or already expired)
+        // to prevent subblock invalidation. ref: <https://github.com/tempoxyz/tempo/issues/1013>
```
mattsse: I think we should exclude this reference here

crates/transaction-pool/src/transaction.rs
```diff
@@ -81,6 +86,7 @@ impl PoolTransactionError for TempoPoolTransactionError {
             Self::ExceedsNonPaymentLimit => false,
             Self::InvalidFeeToken(_) => false,
             Self::MissingFeeToken => false,
+            Self::InvalidValidBefore { .. } => true,
```
mattsse: I think we should be a bit more graceful here, because we would end up penalizing peers if it broadcasted a tx that by the time it reached the other peer is now invalidbefore.

so imo we can treat this as non bad transaction

crates/transaction-pool/src/validator.rs
0xrusowsky: i've been adding gh issue refs to all my audit-related PRs
should i remove all of them?

mattsse: hmm, I see, let's keep then

0xrusowsky: np, i can delete them. will leave the comments explaining why the checks are in place though

crates/transaction-pool/src/validator.rs
```diff
@@ -14,6 +14,9 @@ use tempo_chainspec::{TempoChainSpec, hardfork::TempoHardforks};
 use tempo_precompiles::{ACCOUNT_KEYCHAIN_ADDRESS, AuthorizedKey, compute_keys_slot};
 use tempo_revm::TempoStateAccess;
 
+// Reject AA txs where `valid_before` is too close to current time (or already expired) to prevent block invalidation.
+const AA_VALID_BEFORE_MIN_SECS: u64 = 3;
```
0xKitsune, line 29: Are there use cases where users would want to submit txs that either complete or are invalid within a shorter time horizon? Since we have sub second blocks I am under the impression that this might be the case. 

Is there a downside to only checking if the valid before timestamp is elapsed, and then running some task to clean up the tx pool for AA txs that expire after tx pool inclusion but before block inclusion?

crates/transaction-pool/src/validator.rs
```diff
@@ -163,6 +166,28 @@ where
...
         }
 
+        // Reject AA txs where `valid_before` is too close to current time (or already expired) to prevent block invalidation.
+        if let Some(tx) = transaction.inner().as_aa()
+            && let Some(valid_before) = tx.tx().valid_before
+        {
+            let current_time = std::time::SystemTime::now()
+                .duration_since(std::time::UNIX_EPOCH)
+                .expect("system time before UNIX EPOCH")
+                .as_secs();
```
0xKitsune: Same thing here as https://github.com/tempoxyz/tempo/pull/1040#discussion_r2560959220

Can we use `tip_timestamp` instead?

crates/transaction-pool/src/validator.rs
legion2002: I think 3 seconds is reasonable for most real world use cases.
Bots would def like a lower threshold for MEV purposes, but idt that's the intended use case of the AA Tx type right now, and if we see requests for this, then we can lower this margin without a hardfork anyways.

crates/transaction-pool/src/validator.rs
mattsse: I think local clock is okay, local head can also drift technically

0xrusowsky: used `tip_timestamp` to compute both bounds (lower and upper) from the same value

crates/transaction-pool/src/validator.rs
```diff
@@ -206,22 +209,39 @@ where
...
-                        TempoPoolTransactionError::InvalidValidAfter {
-                            valid_after,
-                            max_allowed,
-                        },
-                    ),
-                );
+        if let Some(tx) = transaction.inner().as_aa() {
+            // Reject AA txs where `valid_before` is too close to current time (or already expired) to prevent mempool DOS attacks.
+            if let Some(valid_before) = tx.tx().valid_before {
+                let current_time = self.inner.fork_tracker().tip_timestamp();
```
mattsse: this works but this value will be outdated in case the node is lagging behind, but then we'll catch those in the maintenance task anyway

crates/transaction-pool/src/validator.rs
```diff
@@ -206,22 +209,39 @@ where
...
+                            },
+                        ),
+                    );
+                }
+            }
+
+            // Reject AA txs where `valid_after` is too far in the future to prevent mempool DOS attacks.
+            if let Some(valid_after) = tx.tx().valid_after {
+                let current_time = self.inner.fork_tracker().tip_timestamp();
+                let max_allowed = current_time.saturating_add(self.aa_valid_after_max_secs);
```
mattsse: for valid after this is slightly more problematic, because if this is lagging then this might get rejected

crates/transaction-pool/src/validator.rs
```diff
@@ -206,22 +209,39 @@ where
...
-                return TransactionValidationOutcome::Invalid(
-                    transaction,
-                    InvalidPoolTransactionError::other(
-                        TempoPoolTransactionError::InvalidValidAfter {
-                            valid_after,
-                            max_allowed,
-                        },
-                    ),
-                );
+        if let Some(tx) = transaction.inner().as_aa() {
```
mattsse: could we extract this to `fn ensure_valid_conditionals(&self, tx: &AATx) -> Result<(), TempoPoolTransactionError>`

0xrusowsky: addressed in [`c5227f9` (#1055)](https://github.com/tempoxyz/tempo/pull/1055/commits/c5227f984200e3dd657a94a237301c64b779a641)

crates/transaction-pool/src/validator.rs
0xrusowsky: addressed in [`c5227f9` (#1055)](https://github.com/tempoxyz/tempo/pull/1055/commits/c5227f984200e3dd657a94a237301c64b779a641)

PR #1056, opened 2025-11-25
fix(precompiles): prevent reverts in `FeeManager` sys-tx

### LINKED ISSUE #1014: Bug: FeeManager systemTx can fail via AA calls from malicious proposer
### Describe the bug

A malicious proposer can create an AA transaction whose inner calls intentionally corrupt the protocol’s fee accounting invariants by abusing the fact that AA inner calls have no gas reservation and therefore may freely mutate fee related state before the protocol’s post-transaction fee settlement in `collect_fees_post_tx`. 

Inside the AA transaction, the validator uses a malicious fee token implementation together with `burn_blocked` mechanics to burn only a portion of the gas fees that were just collected for that transaction. This puts the system in an inconsistent state where the FeeManager’s recorded `collected_fees` value is higher than the actual spendable balance remaining after the partial burn. 

When the system transaction at the end of the block attempts to transfer the full expected fee amount to the validator, the FeeManager does not actually hold that amount anymore due to the earlier partial burn, causing the system transaction to revert.

0xrusowsky:
## Motivation

closes #1014 

## Solution

despite i believe the issue is fixed by forbidding `burn_blocked()` calls on `FeeManager`, i also added an upper bound to the `FeeManager` collected fee transfer (in case that there could be other ways to burn fees).

### DIFF
--- a/crates/precompiles/src/tip20/mod.rs
+++ b/crates/precompiles/src/tip20/mod.rs
@@ -454,7 +454,7 @@ impl<'a, S: PrecompileStorageProvider> TIP20Token<'a, S> {
     ) -> Result<()> {
         self.check_role(msg_sender, *BURN_BLOCKED_ROLE)?;
 
-        // Prevent burning from `FeeManager` and `StablecoinExchanfe` to protect accounting invariants
+        // Prevent burning from `FeeManager` and `StablecoinExchange` to protect accounting invariants
         if self.storage.spec().is_allegretto()
             && matches!(
                 call.from,


crates/precompiles/src/tip20/mod.rs
```diff
@@ -453,6 +453,11 @@ impl<'a, S: PrecompileStorageProvider> TIP20Token<'a, S> {
     ) -> Result<()> {
         self.check_role(msg_sender, *BURN_BLOCKED_ROLE)?;
 
+        // Prevent burning from `FeeManager` to protect fee accounting invariants
+        if call.from == TIP_FEE_MANAGER_ADDRESS {
+            return Err(TIP20Error::protected_address().into());
+        }
```
klkvr: this seems quite restrictive and might need a spec update/sync with @danrobinson 

ideally we just explicitly track that kind of situations and not allow burning any liquidity that is reserved for fee payments

0xrusowsky: i think we can still address the audit concern by only adding the upper-bound in the `FeeManager` transfer. Should we do that? or wait for dan's feedback @klkvr?

danrobinson: Good catch here. I think any burning from the FeeManager will probably cause issues with the accounting, even if it's not going to lead to a revert this particular block, so I think it's correct to forbid burning from it. I think we should also probably forbid burning tokens from the FeeManager, the token contract, and the StablecoinExchange?

0xrusowsky: do i need to open a PR for specs too?

crates/precompiles/src/tip_fee_manager/mod.rs
```diff
@@ -232,19 +232,26 @@ impl<'a, S: PrecompileStorageProvider> TipFeeManager<'a, S> {
...
-                    )
-                    .map_err(|_| {
-                        IFeeManager::IFeeManagerErrors::InsufficientFeeTokenBalance(
-                            IFeeManager::InsufficientFeeTokenBalance {},
+                // Bound fee transfer to contract balance
+                let balance = token.balance_of(ITIP20::balanceOfCall {
+                    account: self.address,
+                })?;
+
+                if !balance.is_zero() {
```
0xKitsune, line 240: Makes sense. Just for clarity to restate the problem this PR addresses:

- A malicious user deploys their own TIP20 and sets it as their preferred fee token.
- They submit an AA bundle that:
  - Adds the FeeManager to the blacklist.
  - Calls `burn_blocked` to burn the funds reserved for the transaction.
- When `execute_block` runs, it can't refund the reserved funds because they've been burned, so the block can't complete.

Given that, I believe proposed fix seems correct. ~~However why does the fee manager need to be restricted from calling `burn_blocked`?. Im under the impression that this precompile will never call this function?~~

crates/precompiles/src/tip20/mod.rs
```diff
@@ -454,6 +454,14 @@ impl<'a, S: PrecompileStorageProvider> TIP20Token<'a, S> {
     ) -> Result<()> {
         self.check_role(msg_sender, *BURN_BLOCKED_ROLE)?;
 
+        // Prevent burning from `FeeManager` and `StablecoinExchanfe` to protect accounting invariants
+        if matches!(
+            call.from,
+            TIP_FEE_MANAGER_ADDRESS | STABLECOIN_EXCHANGE_ADDRESS
+        ) {
+            return Err(TIP20Error::protected_address().into());
+        }
```
0xKitsune: While unlikely to cause any issues, we should be gating this logic behind a hardfork.

0xrusowsky: addressed in [`92369ab` (#1056)](https://github.com/tempoxyz/tempo/pull/1056/commits/92369ababe09d50abdf46467ed4e6a7049be409b)

crates/precompiles/src/tip20/mod.rs
```diff
@@ -454,6 +454,16 @@ impl<'a, S: PrecompileStorageProvider> TIP20Token<'a, S> {
     ) -> Result<()> {
         self.check_role(msg_sender, *BURN_BLOCKED_ROLE)?;
 
+        // Prevent burning from `FeeManager` and `StablecoinExchanfe` to protect accounting invariants
```
0xKitsune:
```suggestion
        // Prevent burning from `FeeManager` and `StablecoinExchange` to protect accounting invariants
```

PR #1057, opened 2025-11-25
feat: update default `TempoHardfork` to latest hardfork

0xKitsune:
This PR updates the default `TempoHardfork` to `Allegretto` and patches the test suite accordingly.

### DIFF
--- a/crates/node/tests/it/liquidity.rs
+++ b/crates/node/tests/it/liquidity.rs
@@ -28,104 +28,32 @@ async fn test_block_building_insufficient_fee_amm_liquidity() -> eyre::Result<()
         .wallet(wallet.clone())
         .connect_http(http_url);
 
-    // Setup payment token
-    let payment_token = crate::utils::setup_test_token(provider.clone(), sender_address).await?;
-    let payment_token_addr = *payment_token.address();
-
-    // Get validator token address (USDC from genesis)
-    use tempo_contracts::precompiles::ITIPFeeAMM;
     use tempo_precompiles::TIP_FEE_MANAGER_ADDRESS;
-    let validator_token_addr = DEFAULT_FEE_TOKEN_POST_ALLEGRETTO;
 
-    let fee_amm = ITIPFeeAMM::new(TIP_FEE_MANAGER_ADDRESS, provider.clone());
-    let validator_token =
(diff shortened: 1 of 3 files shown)


crates/chainspec/src/hardfork.rs
```diff
@@ -26,11 +26,11 @@ hardfork!(
     #[derive(Default)]
     TempoHardfork {
         /// Placeholder representing the baseline (pre-hardfork) state.
-        #[default]
         Adagio,
         /// Testnet hardfork for Andantino. To be removed before mainnet launch.
         Moderato,
         /// Allegretto hardfork.
+        #[default]
         Allegretto,
```
klkvr, line 34: general rule I've sticked to is to keep set default to highest hardfork activated on the "primary" chain (i.e testnet right now, mainnet later) which is `Moderato` rn

I think at this point e.g tempo-foundry relies on this default impl, and it also seems reasonable for our tests to run on `Moderato` to make sure that we test the same codepaths that are actually run on testnet

this also encourages expicitly specifying `Allegretto` when writing tests for Allegretto behavior vs relying on default

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -1817,7 +1817,7 @@ mod tests {
 
     #[test]
     fn test_place_order_below_minimum_amount() -> eyre::Result<()> {
-        let mut storage = HashMapStorageProvider::new(1);
+        let mut storage = HashMapStorageProvider::new_with_spec(1, TempoHardfork::Adagio);
```
klkvr, line 1820: not sure if it makes sense to test things at `Adagio`, entire pre-Adagio history is covered by re-execute CI check and setting this in tests means we will not know about breakages to the logic in newer hardforks

crates/transaction-pool/src/validator.rs
```diff
@@ -70,7 +70,8 @@ where
             }
         };
 
-        let spec = TempoHardfork::Allegretto;
+        // Get the latest hardfork
+        let spec = TempoHardfork::default();
```
klkvr, line 74: we have access to the chainspec here as`self.inner.chain_spec()` so we can do this properly

PR #1059, opened 2025-11-25
feat: `FeeManager` changes

klkvr:
applies `FeeManager` changes specified in https://github.com/tempoxyz/docs/pull/303

main difference is that instead of `tokens_with_fees` we now track `pools_with_fee`, and post-Allegrettto we calculate validator's earnings in `collectFeePostTx`

`collected_fees` was directly changed to be a mapping. this should not be a breaking change because the slot is always zeroed out at the end of the block


crates/precompiles/src/tip_fee_manager/mod.rs
```diff
@@ -10,30 +10,58 @@ pub use tempo_contracts::precompiles::{
...
+        bytes
+    }
+}
+
 #[contract]
 pub struct TipFeeManager {
     validator_tokens: Mapping<Address, Address>,
     user_tokens: Mapping<Address, Address>,
-    collected_fees: U256,
+    collected_fees: Mapping<Address, U256>,
```
joshieDo, line 54: i have no idea, is this populated already? is it a backwards compatible change?

klkvr: yeah backwards compatible because `collected_fees` slot is always zero at the start and at the end of block

re-execute CI covers that

crates/precompiles/src/tip_fee_manager/amm.rs
```diff
@@ -17,6 +17,14 @@ pub const SCALE: U256 = uint!(10000_U256);
 pub const SQRT_SCALE: U256 = uint!(100000_U256);
 pub const MIN_LIQUIDITY: U256 = uint!(1000_U256);
 
+/// Compute amount out for a fee swap
+pub fn compute_amount_out(amount_in: U256) -> Result<U256> {
```
mattsse, line 22: probably something we should inline?

crates/precompiles/src/tip_fee_manager/mod.rs
```diff
@@ -215,29 +264,41 @@ impl<'a, S: PrecompileStorageProvider> TipFeeManager<'a, S> {
...
+        } else {
+            self.drain_pools_with_fees()?
+                .into_iter()
+                .map(|pair| {
+                    (
+                        token_id_to_address(pair.user_token),
+                        token_id_to_address(pair.validator_token),
+                    )
+                })
+                .collect()
```
mattsse, line 283: if we only call drain_pools_with_fees once,
should we directly return the addresses and avoid the additional collect?

crates/precompiles/src/tip_fee_manager/mod.rs
```diff
@@ -215,29 +264,41 @@ impl<'a, S: PrecompileStorageProvider> TipFeeManager<'a, S> {
...
+        };
+        for (user_token, validator_token) in pools {
+            total_amount_out += self.execute_pending_fee_swaps(user_token, validator_token)?;
```
mattsse, line 286: this is safe to do unchecked because this cant overflow afaict

crates/precompiles/tests/storage_tests/solidity/utils.rs
```diff
@@ -342,3 +358,27 @@ pub(super) fn compare_struct_members(
         Err(errors)
     }
 }
+
+/// Panics with a detailed error message when a storage layout mismatch is detected.
+///
+/// Includes instructions for updating the Solidity test file when the spec changes.
+pub(super) fn panic_layout_mismatch(context: &str, errors: Vec<String>, sol_path: &Path) -> ! {
```
mattsse, line 365: cool

PR #1060, opened 2025-11-25
feat(tempo-bench): faucet and fee token

### LINKED ISSUE #1038: When running tempo-bench with `--faucet` argument, mint tokens to cover fees
### Describe the feature

When running `tempo-bench` on an existing network, we don't necessarily have the access to accounts form genesis allocation, meaning that we need to have some tokens for cover the fees for transactions that we'll be spamming.

When `--faucet` flag is passed to `tempo-bench`, it should call `tempo_fundAddress` for each of the accounts derived from the mnemonic.

### Additional context

_No response_

shekhirin:
Closes #1038

1. `--faucet` funds the specified number of accounts
2. `--fee-token` (default being `DEFAULT_FEE_TOKEN_PRE_ALLEGRETTO`) specifies the fee token that will be set as default fee token on `FeeManager` contract. It should be one of the tokens that you get from the faucet.
3. `--mnemonic` allows to specify a custom mnemonic to derive the accounts from. Defaults to `--mnemonic random` which is useful when running `just localnet` that uses the default Foundry mnemonic in faucet, as this can cause nonce issues.
4. A separate quote token is created and used for other tokens as a quote token instead of pathUSD. This will not be needed after Allegretto where pathUSD becomes transferrable.

image


bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -122,20 +122,56 @@ pub struct MaxTpsArgs {
...
+    /// Calls tempo_fundAddress for each account.
+    #[arg(long)]
+    faucet: bool,
+}
+
+#[derive(Debug, Clone)]
+enum MnemonicArg {
+    Mnemonic(String),
+    Random,
+}
```
mattsse: this type is out of order

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -149,32 +185,62 @@ impl MaxTpsArgs {
...
+        let mnemonic = self.mnemonic.resolve();
+        let signer_providers = (self.from_mnemonic_index
+            ..(self.from_mnemonic_index + self.accounts as u32))
+            .into_par_iter()
+            .progress()
+            .map(|i| {
+                let signer = MnemonicBuilder::<English>::default()
+                    .phrase(&mnemonic)
+                    .index(i)?
+                    .build()?;
```
mattsse: there should be a shortcut for this now

this should just be

```
MnemonicBuilder::from_phrase_nth
```

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -455,6 +506,43 @@ async fn generate_transactions(
     Ok(transactions)
 }
 
+async fn fund_accounts(
```
mattsse, line 519: can we document this

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -635,30 +722,40 @@ async fn join_all<
...
 }
 
+async fn assert_receipts<R: ReceiptResponse, F: Future<Output = eyre::Result<R>>>(
+    receipts: impl IntoIterator<Item = F>,
+    max_concurrent_requests: usize,
+) -> eyre::Result<()> {
+    stream::iter(receipts.into_iter())
+        .buffer_unordered(max_concurrent_requests)
+        .try_for_each(async |receipt| {
+            eyre::ensure!(receipt.status(), "Transaction failed");
```
mattsse: this doesnt tell you which tx failed

bin/tempo-bench/Cargo.toml
```diff
@@ -53,3 +53,4 @@ rayon = "1.10"
 rlimit = "0.10.2"
 governor = "0.10.1"
 rand = "0.9.2"
+rand_085 = { package = "rand", version = "0.8.5" }
```
mattsse: can we use rand08 like we do in reth, unsure if this would pull in another rand variant

crates/alloy/src/rpc/request.rs
```diff
@@ -301,3 +311,17 @@ impl From<TempoTypedTransaction> for TempoTransactionRequest {
         }
     }
 }
+
+/// Extension trait for [`CallBuilder`]
+pub trait TempoTransactionCallBuilderExt {
```
mattsse: this makes sense

onbjerg: nit: perhaps just `TempoCallBuilderExt`?

PR #1064, opened 2025-11-25
fix(primitives): use minicbor to validate CBOR-encoded data

### LINKED ISSUE #1008: Bug: Potential for RPC Node DOS by providing bloated webauthn signatures
### Describe the bug

The current node implementation has a potential for RPC Node DOS by providing bloated webauthn signatures which invoke O(n^2) CBOR parsing.

### LINKED ISSUE #1107: WebAuthn: Support extensions
### Describe the feature

WebAuthn extensions are currently forbidden for simplicity and security. This tracks potential future support if needed.

### Context
- Porto doesn't use WebAuthn extensions (confirmed by @jxom, @omar)
- Forbidden to minimize attack surface and keep validation simple
- Can enable later if use cases emerge (credential protection, HMAC secrets, large blob storage, etc.)

### Implementation
PR #1064 has complete implementation using `minicbor` for proper CBOR parsing. Can be revived if needed.

### Decision Criteria

Re-evaluate if:
- Porto adds extension usage
- Users request extension features
- Standards make extensions valuable

**References:** #1064, #1008

### Additional context

_No response_

0xrusowsky:
## Motivation

closes #1008 

## Solution

uses the `minicbor` crate to validate the CBOR data.
i based my impl on:
- https://www.w3.org/TR/webauthn-2/#sctn-authenticator-data
- https://www.w3.org/TR/webauthn-2/#attested-credential-data

but i was completely unfamiliar with this before, so please review carefully

note: to use `minicbor` i had to add `BlueOak-1.0.0` to the list of allowed licences in `deny.toml`


crates/primitives/src/transaction/aa_signature.rs
```diff
@@ -327,70 +328,52 @@ fn verify_webauthn_data_internal(
...
-    let flags = authenticator_data[32];
-    if (flags & 0x01) == 0 {
-        return Err("User Presence (UP) flag not set in authenticatorData");
+    // Determine authenticatorData length
+    let auth_data_len = if ed_flag == 0 {
+        // If ED flag is not set, exactly 37 bytes (no extensions)
+        MIN_AUTH_DATA_LEN
+    } else {
+        // Otherwise, we need to parse CBOR data
+        parse_auth_data_length_with_ed(webauthn_data)?
```
legion2002, line 715: Note for later, we should research in which cases the ED flag is set, and if it is not common maybe just disallow it.
This would also allow us to reduce our max webauthn signature size to be a much tighter value.

imo starting off conservative with the types of webauthn signatures we allow should be fine, we can increase progressively if users ask for it.

crates/primitives/src/transaction/aa_signature.rs
```diff
@@ -415,6 +398,29 @@ fn verify_webauthn_data_internal(
...
 }
 
+/// Parse authenticatorData length when ED flag is set.
+/// ref: <https://www.w3.org/TR/webauthn-2/#sctn-authenticator-data>
+///
+/// When the ED (Extension Data) flag is set, authenticatorData contains:
+/// - 37 bytes base (rpIdHash + flags + signCount)
+/// - CBOR-encoded extensions (variable length)
+fn parse_auth_data_length_with_ed(webauthn_data: &[u8]) -> Result<usize, &'static str> {
+    const MIN_AUTH_DATA_LEN: usize = 37;
```
legion2002: MIN_AUTH_DATA_LEN can be defined just once at the top.

legion2002 (approved):
small nit, rest lgtm

crates/primitives/src/transaction/aa_signature.rs
0xrusowsky: totally agreed, already started a discussion with @jxom and @o-az

crates/primitives/src/transaction/aa_signature.rs
0xrusowsky: addressed in [`380f413` (#1064)](https://github.com/tempoxyz/tempo/pull/1064/commits/380f41323bcfdb04e1dbb3ff20c1b506ea865c93)

fgimenez (approved):
lgtm

PR #1065, opened 2025-11-25
fix(test): more integration test fixes after bumping default hardfork

fgimenez:
fixes remaining integration tests

### DIFF
--- a/crates/node/tests/it/tip_fee_amm.rs
+++ b/crates/node/tests/it/tip_fee_amm.rs
@@ -409,14 +409,12 @@ async fn test_first_liquidity_provider() -> eyre::Result<()> {
     let validator_token = setup_test_token(provider.clone(), alice).await?;
     let fee_amm = ITIPFeeAMM::new(TIP_FEE_MANAGER_ADDRESS, provider.clone());
 
-    // Define amounts (100000 * 1e18)
-    let amount0 = uint!(100000_000000000000000000_U256);
-    let amount1 = uint!(100000_000000000000000000_U256);
+    // Define amount (100000 * 1e18)
+    let amount = uint!(100000_000000000000000000_U256);
 
-    // Mint tokens to alice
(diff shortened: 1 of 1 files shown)


crates/node/tests/it/tip_fee_manager.rs
```diff
@@ -144,8 +145,8 @@ async fn test_set_validator_token() -> eyre::Result<()> {
         .validatorTokens(validator_address)
         .call()
         .await?;
-    // Initial token should be predeployed token
-    assert_eq!(initial_token, token_id_to_address(1));
+    // Initial token should be default fee token (PathUSD in Allegretto)
```
0xKitsune: nit: can we update this comment to mention that we set this in the genesis file for clarity? (Rather than this being a protocol mechanic)

crates/node/tests/it/tip_fee_amm.rs
```diff
@@ -426,12 +428,11 @@ async fn test_first_liquidity_provider() -> eyre::Result<()> {
     assert_eq!(pool.reserveUserToken, 0);
     assert_eq!(pool.reserveValidatorToken, 0);
 
-    // Add liquidity which creates the pool
+    // Add liquidity which creates the pool (using mintWithValidatorToken)
     let mint_receipt = fee_amm
-        .mint(
+        .mintWithValidatorToken(
```
0xKitsune: nit: now that we are using `mintWithValidatorToken`, we dont need `amount0` anymore. We can also rename `amount1` to amount. This is minor and not blocking though.

crates/node/tests/it/tip_fee_manager.rs
fgimenez: yep done

crates/node/tests/it/tip_fee_amm.rs
fgimenez: done too

PR #1073, opened 2025-11-25
feat(node): add custom rpc revert errors

Zygimantass:
Returns errors like these:

```Error: server returned an error response: error code 3: execution reverted: TIP20 token error: TransfersDisabled(TransfersDisabled), data: "0x544950323020746f6b656e206572726f723a205472616e736665727344697361626c6564285472616e736665727344697361626c656429```

Instead of just execution reverted. Ideally in a follow-up we figure out how to get docs links into these errors, but I put it out of scope for this PR. The only foot gun I can think of here is forgetting to add the new errors to the ErrorRegistry.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -6,7 +6,6 @@ on:
       - main
     tags:
       - "v*.*.*"
-  pull_request:
   workflow_dispatch:
   workflow_call:
   schedule:
@@ -27,7 +26,7 @@ jobs:
     runs-on: depot-ubuntu-latest-16
     strategy:
       matrix:
-        binary: 
+        binary:
           - name: tempo
             features: "asm-keccak,jemalloc,otlp"
           - name: tempo-bench
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11266,6 +11266,8 @@ dependencies = [
  "alloy-provider",
  "alloy-rpc-types-eth",
  "alloy-serde",
+ "alloy-signer",
+ "alloy-signer-local",
  "derive_more",
  "reth-evm",
  "reth-primitives-traits",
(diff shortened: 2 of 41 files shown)

### CODE REVIEW

File: crates/precompiles/src/error.rs [hunk 1]
```diff
@@ -81,6 +88,61 @@ impl TempoPrecompileError {
...
+        let converter = Arc::clone(&converter);
+        registry.insert(
+            selector.into(),
+            Box::new(move |data: &[u8]| T::abi_decode(data).ok().map(|error| converter(error))),
+        );
+    }
+}
+
+pub type TempoPrecompileErrorRegistry =
+    HashMap<Selector, Box<dyn Fn(&[u8]) -> Option<TempoPrecompileError> + Send + Sync>>;
```
klkvr: I see, this optimizes `selector -> error` lookup

crates/node/src/rpc/error.rs
```diff
@@ -0,0 +1,81 @@
...
+        Self::EthApiError(RpcInvalidTransactionError::halt(halt, gas_limit).into())
+    }
+}
+
+impl FromRevert for TempoEthApiError {
+    fn from_revert(revert: Bytes) -> Self {
+        match tempo_precompiles::error::decode_error(&revert.0) {
+            Some(error) => Self::EthApiError(EthApiError::from_revert(Bytes::copy_from_slice(
+                error.to_string().as_bytes(),
+            ))),
```
klkvr: I think this would mess with the actual revert bytes? i.e this would change both `message` and `data` fields but we want to keep the `data`

Zygimantass: oh, maybe we should just do PrecompileError instead of RevertError here?

Zygimantass: ah, that doesnt seem to exist, nvm

Zygimantass: but you're right in that it messes with the data

PR #1075, opened 2025-11-26
fix(precompiles): ensure `Orderbook`'s best tick is always up-to-date

### LINKED ISSUE #1062: Bug: The `best_tick` in `orderbook` may not be updated in time
### Describe the bug

In crates/precompiles/src/stablecoin_exchange/mod.rs, orders with the same tick are arranged in a linked list, corresponding to one level. In the `process_pending_order function`, when a level is newly created, its tick is checked and the best_tick may be updated.

```
let prev_tail = level.tail;
if prev_tail == 0 {
    level.head = order_id;
    level.tail = order_id;

    Orderbook::set_tick_bit(self, order.book_key(), order.tick(), order.is_bid())
        .expect("Tick is valid");

    if order.is_bid() {
        if order.tick() > orderbook.best_bid_tick {
            Orderbook::update_best_bid_tick(self, order.book_key(), order.tick())?; // ===> update best tick
        }
    } else if order.tick() < orderbook.best_ask_tick {
        Orderbook::update_best_ask_tick(self, order.book_key(), order.tick())?; // ===> update best tick
    }
}
```

In subsequent transactions, the contract matches orders from the best_tick queue.However, in both the `fill_order` and `cancel_active_order` functions, the orders at a certain tick may be exhausted, but neither place checks whether the best_tick should be updated at this time:

```
let next_tick_info = if order.next() == 0 {
    Orderbook::delete_tick_level(self, book_key, order.is_bid(), order.tick())?;

    Orderbook::clear_tick_bit(self, order.book_key(), order.tick(), order.is_bid())
        .expect("Tick is valid");

    let (tick, has_liquidity) =
        Orderbook::next_initialized_tick(self, book_key, order.is_bid(), order.tick());
    // ===> just delete, clear and search next
}
......
// If this was the last order at this tick, clear the bitmap bit
if level.head == 0 {
    Orderbook::clear_tick_bit(self, order.book_key(), order.tick(), order.is_bid()) // ===> just clear bit_map
        .expect("Tick is valid");
}
```

Then, `get_best_price_level` may return an incorrect level. Since the level.head is 0 at this time, the 0th order is used first in subsequent transactions, which may lead to:
1. Transactions deviate from the expected processing logic, starting from order 0 instead of the best order;
2. The tick 0 queue may be mistakenly deleted and the corresponding bitmap cleared, so the orders in this queue will never be selected for trading and can only be canceled;
3. best_tick may be occupied by an incorrect value for a long time until a better queue is created next time.For reference, in the Solidity implementation, best_tick is updated in `fillOrder`, but not in the cancel function:

0xrusowsky:
## Motivation

closes #1062 

## Solution

ensure that `best_tick` updated when:
- filling orders that consume all its liquidity
- cancelling an active order that represented all the liquidity at the best tick

### DIFF
--- a/crates/precompiles/src/stablecoin_exchange/mod.rs
+++ b/crates/precompiles/src/stablecoin_exchange/mod.rs
@@ -4219,7 +4219,7 @@ mod tests {
 
     #[test]
     fn test_best_tick_updates_on_fill() -> eyre::Result<()> {
-        let mut storage = HashMapStorageProvider::new(1);
+        let mut storage = HashMapStorageProvider::new(1).with_spec(TempoHardfork::Allegretto);
         let mut exchange = StablecoinExchange::new(&mut storage);
         exchange.initialize()?;
 
@@ -4303,7 +4303,7 @@ mod tests {
 
     #[test]
     fn test_best_tick_updates_on_cancel() -> eyre::Result<()> {
-        let mut storage = HashMapStorageProvider::new(1);
+        let mut storage = HashMapStorageProvider::new(1).with_spec(TempoHardfork::Allegretto);
         let mut exchange = StablecoinExchange::new(&mut storage);
         exchange.initialize()?;
 


crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -791,6 +791,15 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
...
                 Orderbook::next_initialized_tick(self, book_key, order.is_bid(), order.tick());
 
+            // Update best_tick when tick is exhausted
+            if order.is_bid() {
+                let new_best = if has_liquidity { tick } else { i16::MIN };
+                Orderbook::update_best_bid_tick(self, book_key, new_best)?;
+            } else {
+                let new_best = if has_liquidity { tick } else { i16::MAX };
+                Orderbook::update_best_ask_tick(self, book_key, new_best)?;
+            }
```
0xKitsune: We need to gate this behind a hardfork.

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -1212,6 +1221,31 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
...
+                );
+
+                if order.is_bid() {
+                    let new_best = if has_liquidity { next_tick } else { i16::MIN };
+                    Orderbook::update_best_bid_tick(self, order.book_key(), new_best)?;
+                } else {
+                    let new_best = if has_liquidity { next_tick } else { i16::MAX };
+                    Orderbook::update_best_ask_tick(self, order.book_key(), new_best)?;
+                }
+            }
```
0xKitsune, line 1252: Same here this needs to be gated behind a hardfork.

PR #1080, opened 2025-11-26
refactor(tempo-bench): simplifications, tracing logs, idiomatic Alloy

shekhirin:
1. Removes manual nonce, gas limit, and gas price management
2. Adds tracing logs
3. Simplifies weighted transaction type choosing
4. Makes the Alloy usage more idiomatic by having contract instances initialized with the providers, and sending the transactions through them, instead of manually signing first and then broadcasting

### DIFF
--- a/bin/tempo-bench/src/cmd/max_tps/dex.rs
+++ b/bin/tempo-bench/src/cmd/max_tps/dex.rs
@@ -51,6 +51,7 @@ pub(super) async fn setup(
             .iter()
             .map(|(signer, _)| signer.address())
             .flat_map(|signer| {
+                #[expect(clippy::redundant_iter_cloned)] // False positive
                 user_tokens.iter().cloned().map(move |token| {
                     Box::pin(async move {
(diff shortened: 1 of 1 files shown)


bin/tempo-bench/src/cmd/max_tps/dex.rs
```diff
@@ -1,156 +1,105 @@
...
+    futures.extend(
+        signer_providers
+            .iter()
+            .map(|(signer, _)| signer.address())
+            .flat_map(|signer| {
+                user_tokens.iter().cloned().map(move |token| {
+                    Box::pin(async move {
+                        let tx = token.mint(signer, mint_amount);
+                        tx.send().await
+                    }) as BoxFuture<'static, _>
```
mattsse, line 59: actually unsure if these are still needed after the first type hint

shekhirin: they are, yeah ☹️

PR #1081, opened 2025-11-26
!feat: consistent handling of spending limits and expiry

legion2002:
* If limits is set to None, then we consider the key to have unlimited spending limit.
* If limits is set to empty vec, then we consider they key to have NO spending limits.
* If expiry is set to None, then we consider the key to have NO expiry.
* If expiry on any key is 0, then it means that the key is not active. We remove the explicity is_active flag.
* Setting `expiry < current_timestamp` is not allowed.

This PR also adds replay protection, by disallowing keys to be reauthorized once they have been revoked.

### DIFF
--- a/crates/precompiles/src/tip_fee_manager/amm.rs
+++ b/crates/precompiles/src/tip_fee_manager/amm.rs
@@ -17,6 +17,15 @@ pub const SCALE: U256 = uint!(10000_U256);
 pub const SQRT_SCALE: U256 = uint!(100000_U256);
 pub const MIN_LIQUIDITY: U256 = uint!(1000_U256);
 
+/// Compute amount out for a fee swap
+#[inline]
+pub fn compute_amount_out(amount_in: U256) -> Result<U256> {
+    amount_in
+        .checked_mul(M)
+        .map(|product| product / SCALE)
+        .ok_or(TempoPrecompileError::under_overflow())
+}
+
 /// Pool structure matching the Solidity implementation
 #[derive(Debug, Clone, Default, Storable)]
 pub struct Pool {
@@ -33,7 +42,7 @@ impl From<Pool> for ITIPFeeAMM::Pool {
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Storable)]
 pub struct PoolKey {
     pub user_token: Address,
     pub validator_token: Address,
@@ -78,10 +87,7 @@ impl<'a, S: PrecompileStorageProvider> TipFeeManager<'a, S> {
         max_amount: U256,
     ) -> Result<()> {
         let pool_id = PoolKey::new(user_token, validator_token).get_id();
-        let amount_out = max_amount
-            .checked_mul(M)
-            .map(|product| product / SCALE)
-            .ok_or(TempoPrecompileError::under_overflow())?;
+        let amount_out = compute_amount_out(max_amount)?;
         let available_validator_token = self.get_effective_validator_reserve(pool_id)?;
 
         if amount_out > available_validator_token {
@@ -107,10 +113,7 @@ impl<'a, S: PrecompileStorageProvider> TipFeeManager<'a, S> {
     fn get_effective_validator_reserve(&mut self, pool_id: B256) -> Result<U256> {
         let pool = self.sload_pools(pool_id)?;
         let pending_fee_swap_in = self.get_pending_fee_swap_in(pool_id)?;
-        let pending_out = U256::from(pending_fee_swap_in)
-            .checked_mul(M)
-            .and_then(|product| product.checked_div(SCALE))
-            .ok_or(TempoPrecompileError::under_overflow())?;
+        let pending_out = compute_amount_out(U256::from(pending_fee_swap_in))?;
 
         U256::from(pool.reserve_validator_token)
             .checked_sub(pending_out)
@@ -623,10 +626,7 @@ impl<'a, S: PrecompileStorageProvider> TipFeeManager<'a, S> {
         let mut pool = self.sload_pools(pool_id)?;
 
         let amount_in = U256::from(self.get_pending_fee_swap_in(pool_id)?);
-        let pending_out = amount_in
-            .checked_mul(M)
-            .and_then(|product| product.checked_div(SCALE))
-            .ok_or(TempoPrecompileError::under_overflow())?;
+        let pending_out = compute_amount_out(amount_in)?;
 
         // Use checked math for these operations
         let new_user_reserve = U256::from(pool.reserve_user_token)
--- a/crates/precompiles/src/tip_fee_manager/mod.rs
+++ b/crates/precompiles/src/tip_fee_manager/mod.rs
@@ -10,30 +10,58 @@ pub use tempo_contracts::precompiles::{
 use crate::{
     DEFAULT_FEE_TOKEN_POST_ALLEGRETTO, DEFAULT_FEE_TOKEN_PRE_ALLEGRETTO, PATH_USD_ADDRESS,
     error::{Result, TempoPrecompileError},
-    storage::{PrecompileStorageProvider, Slot, Storable, VecSlotExt},
-    tip_fee_manager::amm::Pool,
-    tip20::{ITIP20, TIP20Token, is_tip20, validate_usd_currency},
+    storage::{PrecompileStorageProvider, Slot, Storable, StorageKey, VecSlotExt},
+    tip_fee_manager::amm::{Pool, compute_amount_out},
+    tip20::{
+        ITIP20, TIP20Token, address_to_token_id_unchecked, is_tip20, token_id_to_address,
+        validate_usd_currency,
+    },
 };
 
 // Re-export PoolKey for backward compatibility with tests
 use alloy::primitives::{Address, Bytes, IntoLogData, U256, uint};
 use revm::state::Bytecode;
-use tempo_precompiles_macros::contract;
+use tempo_precompiles_macros::{Storable, contract};
 
 /// Helper type to easily interact with the `tokens_with_fees` array
 type TokensWithFees = Slot<Vec<Address>, TokensWithFeesSlot>;
+
+/// Helper type to easily interact with the `pools_with_fees` array
+type PoolsWithFees = Slot<Vec<TokenPair>, PoolsWithFeesSlot>;
+
+/// Helper type to easily interact with the `validators_with_fees` array
+type ValidatorsWithFees = Slot<Vec<Address>, ValidatorsWithFeesSlot>;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Storable)]
+pub struct TokenPair {
+    pub user_token: u64,
+    pub validator_token: u64,
+}
+
+impl StorageKey for TokenPair {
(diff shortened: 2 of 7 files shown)


crates/transaction-pool/src/validator.rs
```diff
@@ -155,12 +155,17 @@ where
...
+        if authorized_key.is_revoked {
+            return Ok(Err("access key has been revoked"));
         }
 
+        // Check if key exists (key exists if expiry > 0)
+        if authorized_key.expiry == 0 {
+            return Ok(Err("access key does not exist"));
+        }
+
+        // Expiry checks are skipped here, they are done in the EVM handler where block timestamp is easily available.
```
mattsse, line 175: I see, then this is also something we should potentially monitor with @0xrusowsky txpool task?

crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -101,12 +127,32 @@ impl KeyAuthorization {
         chain_id.encode(&mut auth_message);
         key_type_byte.encode(&mut auth_message);
         key_id.encode(&mut auth_message);
-        expiry.encode(&mut auth_message);
-        encode_list(limits, &mut auth_message);
+
+        // Encode expiry: None as 0x80, Some(v) as the value
+        match expiry {
+            None => auth_message.push(0x80), // RLP empty string
```
mattsse: can we use EMPTY_STRING_CODE for this instead?

crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -63,32 +74,47 @@ impl KeyAuthorization {
...
+        // Option<u64>: None = 1 byte (0x80), Some(v) = v.length()
+        let expiry_length = match expiry {
+            None => 1, // RLP empty string 0x80
+            Some(v) => v.length(),
+        };
+        // Option<&[TokenLimit]>: None = 1 byte (0x80), Some(v) = list_length(v)
+        let limits_length = match limits {
+            None => 1, // RLP empty string 0x80
+            Some(v) => list_length(v),
+        };
```
mattsse: this is basically .map_or or map().unwrap_or(1)

but I don't mind the match here that much

crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -27,29 +27,40 @@ pub struct TokenLimit {
...
+/// - Non-optional fields come first, followed by optional (trailing) fields
+/// - `expiry`: `None` (omitted or 0x80) = key never expires, `Some(timestamp)` = expires at timestamp
+/// - `limits`: `None` (omitted or 0x80) = unlimited spending, `Some([])` = no spending, `Some([...])` = specific limits
 #[derive(Clone, Debug, PartialEq, Eq, Hash, alloy_rlp::RlpEncodable, alloy_rlp::RlpDecodable)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
-#[cfg_attr(test, reth_codecs::add_arbitrary_tests(compact, rlp))]
 #[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
+#[rlp(trailing)]
+// TODO: Macro RLP tests don't work here because of the trailing flag.
```
mattsse: I don't think trailing works here because this will allow missing empty string makers

so for this we need to impl manually I believe @klkvr

crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -27,29 +27,40 @@ pub struct TokenLimit {
...
     pub chain_id: u64,
 
     /// Type of key being authorized (Secp256k1, P256, or WebAuthn)
     pub key_type: SignatureType,
 
-    /// Unix timestamp when key expires (0 = never expires)
-    pub expiry: u64,
-
-    /// TIP20 spending limits for this key
-    pub limits: Vec<TokenLimit>,
```
mattsse, line 46: move these will alter rlp, which would be breaking but if they become optional we can still do this but we need to keep them here

crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -27,29 +27,40 @@ pub struct TokenLimit {
 ///
 /// Used in TxAA to add a new key to the AccountKeychain precompile.
 /// The transaction must be signed by the root key to authorize adding this access key.
+///
+/// RLP encoding: `[key_type, key_id, signature, expiry?, limits?]`
+/// - Non-optional fields come first, followed by optional (trailing) fields
```
mattsse, line 33: optional in rlp is hard -.- hence the empty string marker is a thing

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -224,11 +249,26 @@ impl<'a, S: PrecompileStorageProvider> AccountKeychain<'a, S> {
...
+            return Err(AccountKeychainError::key_already_revoked().into());
+        }
 
-        if !key.is_active {
-            return Err(AccountKeychainError::key_inactive().into());
+        if key.expiry == 0 {
+            return Err(AccountKeychainError::key_not_found().into());
+        }
+
+        let current_timestamp = self.storage.timestamp().to::<u64>();
```
mattsse:
```suggestion
        let current_timestamp = self.storage.timestamp().saturating_to::<u64>();
```

crates/node/tests/it/account_abstraction.rs
```diff
@@ -3952,6 +4002,624 @@ async fn test_transaction_key_authorization_and_spending_limits() -> eyre::Resul
...
+    envelope.encode_2718(&mut encoded);
+
+    setup.node.rpc.inject_tx(encoded.into()).await?;
+    setup.node.advance_block().await?;
+    nonce += 1;
+
+    // Check the receipt to understand the result
+    let receipt: Option<serde_json::Value> = provider
+        .raw_request("eth_getTransactionReceipt".into(), [tx_hash])
+        .await?;
```
mattsse: this works but we can also create a provider like this

https://github.com/tempoxyz/tempo/blob/5d7430af2ca23cfd72ca2ac8096c49dba8276fe3/bin/tempo-bench/src/cmd/max_tps.rs#L537-L538

then we have type bindings for this

and dont need to deal with json

crates/node/tests/it/account_abstraction.rs
```diff
@@ -3952,6 +4002,624 @@ async fn test_transaction_key_authorization_and_spending_limits() -> eyre::Resul
...
+    let mut encoded = Vec::new();
+    envelope.encode_2718(&mut encoded);
+    let tx_hash = *envelope.tx_hash();
+
+    setup.node.rpc.inject_tx(encoded.into()).await?;
+    setup.node.advance_block().await?;
+
+    let receipt: Option<serde_json::Value> = provider
+        .raw_request("eth_getTransactionReceipt".into(), [tx_hash])
+        .await?;
```
mattsse: same

crates/primitives/src/transaction/key_authorization.rs
legion2002: having breaking changes wouldn't matter, because this feature wasn't in the last hardfork.
But if we're going to do manual RLP anyways, might as well keep them here.

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -256,12 +296,14 @@ impl<'a, S: PrecompileStorageProvider> AccountKeychain<'a, S> {
     pub fn get_key(&mut self, call: getKeyCall) -> Result<KeyInfo> {
         let key = self.sload_keys(call.account, call.keyId)?;
 
-        // If the key is not active, return default (non-existent key)
-        if !key.is_active {
+        // Key exists if expiry > 0 (but might be revoked)
+        if key.expiry == 0 && !key.is_revoked {
```
klkvr: is this supposed to be `if key.expiry == 0 || key.is_revoked`?

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -224,11 +249,26 @@ impl<'a, S: PrecompileStorageProvider> AccountKeychain<'a, S> {
...
-        if !key.is_active {
-            return Err(AccountKeychainError::key_inactive().into());
+        if key.expiry == 0 {
+            return Err(AccountKeychainError::key_not_found().into());
+        }
+
+        let current_timestamp = self.storage.timestamp().to::<u64>();
+        if current_timestamp >= key.expiry {
+            return Err(AccountKeychainError::key_expired().into());
+        }
```
klkvr, line 258: this is identical to `validate_keychain_authorization` and `verify_and_update_spending` logic I believe, can we unify this? like `fn enforce_valid_key(key_id) -> Key`

crates/primitives/src/transaction/key_authorization.rs
klkvr: I'm wondering if we should just extract all of that into a separate unsigned auth struct and rely on rlp derives entirely

crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -27,29 +27,40 @@ pub struct TokenLimit {
...
+/// - `expiry`: `None` (omitted or 0x80) = key never expires, `Some(timestamp)` = expires at timestamp
+/// - `limits`: `None` (omitted or 0x80) = unlimited spending, `Some([])` = no spending, `Some([...])` = specific limits
 #[derive(Clone, Debug, PartialEq, Eq, Hash, alloy_rlp::RlpEncodable, alloy_rlp::RlpDecodable)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
-#[cfg_attr(test, reth_codecs::add_arbitrary_tests(compact, rlp))]
```
klkvr, line 33: hmm the tests seem to be working fine for me?

crates/revm/src/handler.rs
```diff
@@ -692,21 +692,45 @@ where
                 SignatureType::WebAuthn => PrecompileSignatureType::WebAuthn,
             };
 
-            // Convert limits to the format expected by the precompile
+            // Handle expiry: None means never expires (store as u64::MAX)
+            let expiry = key_auth.expiry.unwrap_or(u64::MAX);
+
+            // Validate expiry is not in the past
+            let current_timestamp = block.timestamp().to::<u64>();
```
klkvr: this should be `saturating_to`

crates/revm/src/handler.rs
```diff
@@ -692,21 +692,45 @@ where
...
+            if expiry <= current_timestamp {
+                return Err(EVMError::Transaction(
+                    TempoInvalidTransaction::AccessKeyAuthorizationFailed {
+                        reason: format!(
+                            "Key expiry {expiry} is in the past (current timestamp: {current_timestamp})"
+                        ),
+                    },
+                ));
+            }
```
klkvr, line 708: is this check necessary? it's just like `valid_before` but kinda indirect

klkvr: I guess every keychain tx for expiring key has this sort of timer anyway so not much we can do here

legion2002: Yeah this check is for authorizing keys, and it wouldn't make sense to authorize keys in the past.

PR #1084, opened 2025-11-26
feat(stablecoin-exchange): enforce tick spacing

0xKitsune:
Ref tempoxyz/docs#316

This PR introduces `TICK_SPACING` and enforces that new orders placed after the allegretto hardfork adhere to the specified tick spacing. Note that fill logic still iterates through each tick in the tick bitmap, and this should be updated in the future for more a efficient search over existing liquidity.


crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -454,6 +457,11 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
             return Err(StablecoinExchangeError::tick_out_of_bounds(tick).into());
         }
 
+        // Post allegretto, enforce that the tick adheres to tick spacing
+        if self.storage.spec().is_allegretto() && !(tick as u64).is_multiple_of(TICK_SPACING) {
```
fgimenez: this does not work fine for negative ticks, we could just check modulo without the u64 cast
```suggestion
        if self.storage.spec().is_allegretto() && tick % TICK_SPACING as i16 != 0 {
```

0xKitsune: I have updated `TICK_SPACING` to an `i16` and adjusted the check to be `tick % TICK_SPACING != 0`

Issue #1026, opened 2025-11-24
Tracking: Enable paid sub-blocks


jenpaff:
### Describe the feature

In Andantino we shipped sub-blocks infrastructure allowing validators to include transactions in every block (not just when they're the proposer). Currently, only free transactions (maxFeePerGas=0) are supported in sub-blocks. The limitation exists because the Fee Manager cannot handle fee distribution across multiple validators within a single block.

Tasks involved: 
- [x] merge spec https://github.com/tempoxyz/docs/pull/303 
- [x] enable sub-blocks with fees https://github.com/tempoxyz/tempo/pull/1018
- [x] incorporate TipFeeManager changes https://github.com/tempoxyz/tempo/pull/1059
- [x] incorporate in sub-blocks to enable paid sub blocks https://github.com/tempoxyz/tempo/pull/1086
- [ ] Extend tempo-bench to support sub-blocks

PR #1086, opened 2025-11-26
feat: paid subblocks

klkvr:
Based on https://github.com/tempoxyz/tempo/pull/1059
ref https://github.com/tempoxyz/tempo/issues/1026

Implements logic for paid subblocks, changes in the PR:
- `TempoHaltReason` added which wraps basic Ethereum `HaltReason` and adds custom variant of `SubblockTxFeePayment` representing a halt due to subblock transaction failing to pay fees.
- `FeePaymentError` error type is introduced that encapsulates all failures that might happen during `collectFeePreTx` call
- Instead of a separate `subblock_fee_recipient` field, `block_env.beneficiary` is now temporarily set to subblock's configured `fee_recipient`
- Subblock transactions are no longer exempt from basefee checks and are required to pay fees under regular rules post-Allegretto
- If subblock transaction encounters a failure in `collectFeePreTx` call, this failure is catched in `catch_error` hook and treated as a halt with `SubblockTxFeePayment` reason

### DIFF
--- a/crates/revm/src/handler.rs
+++ b/crates/revm/src/handler.rs
@@ -845,7 +845,7 @@ where
             // Map fee collection errors to transaction validation errors since they
             // indicate the transaction cannot be included (e.g., insufficient liquidity
             // in FeeAMM pool for fee swaps)
-            return Err(match err {
+            Err(match err {
                 TempoPrecompileError::TIPFeeAMMError(TIPFeeAMMError::InsufficientLiquidity(_)) => {
                     FeePaymentError::InsufficientAmmLiquidity {
                         fee: gas_balance_spending,
@@ -866,10 +866,12 @@ where
                 TempoPrecompileError::Fatal(e) => EVMError::Custom(e),
 
                 _ => EVMError::Transaction(FeePaymentError::Other(err.to_string()).into()),
-            });
-        }
+            })
+        } else {
+            journal.checkpoint_commit();
 
-        Ok(())
+            Ok(())
+        }
     }
 
     fn reimburse_caller(


crates/revm/src/handler.rs
```diff
@@ -968,6 +979,36 @@ where
...
+            //
+            // `collectFeePreTx` call will happen after the nonce bump so this will only commit the nonce increment.
+            evm.ctx.journaled_state.commit_tx();
+
+            evm.ctx().local_mut().clear();
+            evm.ctx().journal_mut().discard_tx();
+            evm.frame_stack().clear();
+
+            Ok(ExecutionResult::Halt {
+                reason: TempoHaltReason::SubblockTxFeePayment,
```
joshieDo, line 1049: why erase the real err?

crates/commonware-node/src/subblocks.rs
```diff
@@ -529,11 +529,13 @@ async fn build_subblock(
                 if tx.gas_limit() > gas_left {
                     continue;
                 }
-                if evm.transact_commit(&*tx).is_err() {
+                if let Err(err) = evm.transact_commit(&*tx) {
+                    warn!(%err, tx_hash = %tx_hash, "invalid subblock candidate transaction");
```
mattsse, line 533: I remember running into a case where some of the tx traits didnt satisfy debug, should we add this to alloy-evm?
I believe we can easily do this

crates/evm/src/evm.rs
```diff
@@ -25,7 +25,7 @@ impl EvmFactory for TempoEvmFactory {
     type Tx = TempoTxEnv;
     type Error<DBError: std::error::Error + Send + Sync + 'static> =
         EVMError<DBError, TempoInvalidTransaction>;
-    type HaltReason = HaltReason;
+    type HaltReason = TempoHaltReason;
```
mattsse, line 28: pretty cool that we thought of this earlier already

crates/revm/src/error.rs
```diff
@@ -169,6 +148,67 @@ impl<DBError> From<TempoInvalidTransaction> for EVMError<DBError, TempoInvalidTr
...
+/// Tempo-specific halt reason.
+///
+/// Used to extend basic [`HaltReason`] with an edge case of a subblock transaction fee payment error.
+#[derive(Debug, Clone, PartialEq, Eq, Hash, derive_more::From)]
+pub enum TempoHaltReason {
+    /// Basic Ethereum halt reason.
+    #[from]
+    Ethereum(HaltReason),
+    /// Subblock transaction failed to pay fees.
+    SubblockTxFeePayment,
```
mattsse, line 205: we need this because if a subblock tx fails to pay, we need to halt and consume all gas right?

this similar to out of gas i assume?

crates/revm/src/handler.rs
mattsse, line 1: @rakita could you ptal at these as well?

crates/revm/src/handler.rs
```diff
@@ -783,52 +776,59 @@ where
                 .map_err(|e| EVMError::Custom(e.to_string()))?;
         }
 
-        let mut fee_manager = TipFeeManager::new(&mut storage_provider);
+        let balance_spending = tx.effective_balance_spending(
+            block.basefee as u128,
+            block.blob_gasprice().unwrap_or_default(),
+        )?;
```
rakita: Previously it was
```
        // calculate the new balance after the fee is collected.
        let new_balance = calculate_caller_fee(account_balance, tx, block, cfg)?;
        // doing max to avoid underflow as new_balance can be more than
        // account balance if `cfg.is_balance_check_disabled()` is true.
        let gas_balance_spending = core::cmp::max(account_balance, new_balance) - new_balance;
```
This logic takes into account check for max spending (limit*max_gas_price), and takes into account `is_balance_check_disabled` (I think this is used in foundry)

crates/revm/src/handler.rs
```diff
@@ -968,6 +979,36 @@ where
...
+            && evm.cfg.spec.is_allegretto()
+            && let Some(TempoInvalidTransaction::CollectFeePreTx(_)) = error.as_invalid_tx_err()
+        {
+            // Commit the transaction.
+            //
+            // `collectFeePreTx` call will happen after the nonce bump so this will only commit the nonce increment.
+            evm.ctx.journaled_state.commit_tx();
+
+            evm.ctx().local_mut().clear();
+            evm.ctx().journal_mut().discard_tx();
```
```suggestion
```

Should be removed commit_tx() above does the job.

crates/revm/src/handler.rs
klkvr: I see, good catch

I was trying to simplify the set of errors we need to catch but I guess it would be eaiser to keep this as is, ptal at e11dde4

crates/revm/src/error.rs
klkvr: yep, essentially we want a way to not invalidate transaction but also don't execute anything, though we return 0 gas usage for now

crates/revm/src/handler.rs
rakita: It is not that obvious and could do some cleanup, but that `calculate_caller_fee` is used on mainnet so it give us confidence

crates/revm/src/handler.rs
```diff
@@ -783,52 +785,54 @@ where
...
+                        balance: available,
+                    }
+                    .into(),
+                ),
+
+                TempoPrecompileError::Fatal(e) => EVMError::Custom(e),
+
+                _ => EVMError::Transaction(FeePaymentError::Other(err.to_string()).into()),
+            });
+        }
```
rakita: We should do `journal.checkpoint_commit()` if not err, it internally would decrement the nonce that is used by Inspectors.

rakita: Additionally doing the commit here, would not revert the changes if the call reverts. Not sure if it is expected or not

PR #1099, opened 2025-11-27
fix(primitives): reject AT and ED flags in webauthn validation

### LINKED ISSUE #1008: Bug: Potential for RPC Node DOS by providing bloated webauthn signatures

0xrusowsky:
## Motivation

closes #1008 

## Solution

when performing webauthn validation, reject signatures with the following flags:
- AT: it must NOT be set for assertion signatures (`webauthn.get`)
- ED: we purposely want to deny extensions, as we don't have any usecase for them yet

note: by rejecting extensions, we do not need to worry about CBOR-encoded data

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -69,30 +69,29 @@ jobs:
         env:
           RUSTDOCFLAGS: --cfg docsrs -D warnings --show-type-layout --generate-link-to-definition --enable-index-page -Zunstable-options
       - name: Setup Pages
-        if: github.ref_name == 'main' && github.event_name == 'push'
+        if: false && github.ref_name == 'main' && github.event_name == 'push'
         uses: actions/configure-pages@v5
       - name: Upload artifact
-        if: github.ref_name == 'main' && github.event_name == 'push'
+        if: false && github.ref_name == 'main' && github.event_name == 'push'
         uses: actions/upload-pages-artifact@v4
         with:
           path: ./target/doc
 
-  # Disabled until public
-  # deploy-docs:
-  #   if: github.ref_name == 'main' && github.event_name == 'push'
-  #   needs: [docs]
-  #   runs-on: depot-ubuntu-latest
-  #   timeout-minutes: 30
-  #   permissions:
-  #     pages: write
-  #     id-token: write
-  #   environment:
-  #     name: github-pages
-  #     url: ${{ steps.deployment.outputs.page_url }}
-  #   steps:
-  #     - name: Deploy to GitHub Pages
-  #       id: deployment
-  #       uses: actions/deploy-pages@v4
+  deploy-docs:
+    if: false && github.ref_name == 'main' && github.event_name == 'push'
+    needs: [docs]
+    runs-on: depot-ubuntu-latest
+    timeout-minutes: 30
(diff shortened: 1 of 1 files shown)


crates/primitives/src/transaction/aa_signature.rs
```diff
@@ -684,71 +688,51 @@ fn verify_webauthn_data_internal(
...
 
+    let authenticator_data = &webauthn_data[..auth_data_len];
+    let client_data_json = &webauthn_data[auth_data_len..];
+
     // Validate clientDataJSON
     let json_str =
         core::str::from_utf8(client_data_json).map_err(|_| "clientDataJSON is not valid UTF-8")?;
 
+    // Basic JSON structure validation
+    if !json_str.starts_with('{') || !json_str.ends_with('}') {
```
legion2002, line 728: Why is this validation necessary here? I don't think we do it in the onchain webauthn verifiers.
If we do want to check for clientDataJSON structure, this doesn't seem enough.

crates/primitives/src/transaction/aa_signature.rs
```diff
@@ -684,71 +688,51 @@ fn verify_webauthn_data_internal(
...
+        // If ED flag is not set, exactly 37 bytes (no extensions)
+        MIN_AUTH_DATA_LEN
+    } else {
+        // ED flag must NOT be set, as Tempo AA doesn't support extensions
+        // NOTE: If we ever want to support extensions, we will have to parse CBOR data
+        return Err("ED flag must not be set, as Tempo doesn't support extensions");
+    };
+
+    // Ensure that we have clientDataJSON after authenticatorData
+    if auth_data_len >= webauthn_data.len() {
```
legion2002: This can be removed here, and just added to the check above.

crates/primitives/src/transaction/aa_signature.rs
```diff
@@ -684,71 +688,51 @@ fn verify_webauthn_data_internal(
     webauthn_data: &[u8],
     tx_hash: &B256,
 ) -> Result<B256, &'static str> {
-    // Minimum authenticatorData is 37 bytes (32 rpIdHash + 1 flags + 4 signCount)
-    if webauthn_data.len() < 37 {
+    if webauthn_data.len() < MIN_AUTH_DATA_LEN {
```
legion2002:
```suggestion
    if webauthn_data.len() <= MIN_AUTH_DATA_LEN + 32 {
```

This allows us to remove the clientDataJson check below. This is also a tighter check, because all webauthn data must contain the 37 byte auth data, and the 32 byte tx hash.

crates/primitives/src/transaction/aa_signature.rs
0xrusowsky: that's the check u had before the patch, i only moved it around.

what do u propose?

legion2002: oh okay, I thought this was some audit finding.
Yeah we don't have this check onchain, but I think fine to keep here, it is just an extra layer of validation.

PR #1106, opened 2025-11-27
feat(precompiles): add missing FeeManager interface functions and pending fees validation

fgimenez:
ref https://github.com/tempoxyz/tempo/issues/1050

- Add `CannotChangeWithPendingFees` error and validation in `setValidatorToken` (post-Allegretto)
- Expose `M()`, `N()`, `SCALE()` constants as view functions per spec
- Add `collectFeePreTx/collectFeePostTx` to interface with dispatch handlers (protocol-only)

### DIFF
--- a/crates/node/src/rpc/dex/mod.rs
+++ b/crates/node/src/rpc/dex/mod.rs
@@ -9,6 +9,7 @@ use reth_provider::{BlockReaderIdExt, StateProviderFactory};
 use reth_rpc_eth_api::{RpcNodeCore, helpers::SpawnBlocking};
 use reth_rpc_eth_types::{EthApiError, error::FromEthApiError};
 use tempo_alloy::rpc::pagination::PaginationParams;
+use tempo_chainspec::hardfork::TempoHardfork;
 use tempo_evm::TempoEvmConfig;
 use tempo_precompiles::{
     stablecoin_exchange::{
@@ -503,8 +504,13 @@ impl<'a, 'b> BookIterator<'a, 'b> {
     /// Get the next initialized tick after the given tick
     /// Returns None if there are no more ticks
     pub fn get_next_tick(&mut self, tick: i16) -> Option<i16> {
-        let (next_tick, more_ticks) =
-            PrecompileOrderbook::next_initialized_tick(self, self.book_key, self.bids, tick);
+        let (next_tick, more_ticks) = PrecompileOrderbook::next_initialized_tick(
+            self,
+            self.book_key,
+            self.bids,
+            tick,
+            TempoHardfork::default(),
+        );
 
         if more_ticks { Some(next_tick) } else { None }
     }
--- a/crates/precompiles/src/stablecoin_exchange/mod.rs
+++ b/crates/precompiles/src/stablecoin_exchange/mod.rs
@@ -835,8 +835,13 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
             Orderbook::clear_tick_bit(self, order.book_key(), order.tick(), order.is_bid())
                 .expect("Tick is valid");
 
-            let (tick, has_liquidity) =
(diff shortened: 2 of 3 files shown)


crates/contracts/src/precompiles/tip_fee_manager.rs
```diff
@@ -34,6 +34,20 @@ sol! {
         // Fee functions
         function getFeeTokenBalance(address sender, address validator) external view returns (address, uint256);
         function executeBlock() external;
+        function collectFeePreTx(
+            address user,
+            address txToAddress,
+            uint256 maxAmount,
+            address feeRecipient
+        ) external returns (address userToken);
```
klkvr: let's remove `txToAddress` and `feeRecipient` from signature

`txToAddress` is unused and `feeRecipient` can be fetched as current `beneficiary`

crates/contracts/src/precompiles/tip_fee_manager.rs
```diff
@@ -34,6 +34,20 @@ sol! {
...
+        function collectFeePostTx(
+            address user,
+            uint256 maxAmount,
+            uint256 actualUsed,
+            address userToken,
+            address validatorToken,
+            address feeRecipient
+        ) external;
```
klkvr: same here, `feeRecipient` and `validatorToken` can be inferred from environment

crates/contracts/src/precompiles/tip_fee_manager.rs
fgimenez: done, will prepare changes in docs

crates/precompiles/src/tip_fee_manager/dispatch.rs
```diff
@@ -93,6 +101,46 @@ impl<'a, S: PrecompileStorageProvider> Precompile for TipFeeManager<'a, S> {
                     self.execute_block(s, self.storage.beneficiary())
                 })
             }
+            // Protocol-only functions - these are called directly by the protocol,
+            // not through external transactions. Return error if called externally.
+            IFeeManager::collectFeePreTxCall::SELECTOR => {
```
0xKitsune: Is the rationale that we should add these to the interface/dispatch to expose these for fuzz testing? If so, maybe we feature flag these or similar, I dont think exposing these via dispatch is necessary. If we are only adding this to have 1:1 parity with the reference implementation, I think we can add a dev comment on these functions within the referece contract to mention that these are not actually exposed via the public interface.

Let me know if I am overlooking something here.

crates/precompiles/src/tip_fee_manager/dispatch.rs
```diff
@@ -93,6 +101,46 @@ impl<'a, S: PrecompileStorageProvider> Precompile for TipFeeManager<'a, S> {
...
+                            crate::tip_fee_manager::FeeManagerError::only_system_contract().into(),
+                        );
+                    }
+                    // Determine fee token from user preferences or default
+                    let user_token = self.sload_user_tokens(call.user)?;
+                    let fee_token = if user_token.is_zero() {
+                        self.default_fee_token()
+                    } else {
+                        user_token
+                    };
```
0xKitsune: This only covers the case if the fee token is set, but does not account for the other fee token selection rules.

crates/precompiles/src/tip_fee_manager/dispatch.rs
fgimenez: good point, removed and added a note, aolso done in docs https://github.com/tempoxyz/docs/pull/347/commits/a7a7747805867ea41dfaa3003b4999a618453311

crates/precompiles/src/tip_fee_manager/dispatch.rs
fgimenez: yep, but all this was already removed

PR #1113, opened 2025-11-28
feat(tempo-bench): receipts sampling with timeout, `--clear-txpool`

shekhirin:
Two main chages:
1. If any of the transactions get stuck and do not land on-chain, our current receipts sampling will get stuck as well. This PR adds a 5 second timeout when waiting for receipts.
2. If any of the previous benchmark runs transactions got stuck, you can run with `--clear-txpool` and `tempo-bench` will clear the transaction pool for you before sending any transactions. This should be the preferred way of running `tempo-bench` on networks intended for benchmarking only, where clearing the transaction pool is acceptable.

Additionally:
1. Added logs with the number of succeeded, failed, timeout transactions. Same for receipts sampling.
2. `monitor_tps` is now a future that's getting cancelled with a cancellation token

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -69,30 +69,29 @@ jobs:
         env:
           RUSTDOCFLAGS: --cfg docsrs -D warnings --show-type-layout --generate-link-to-definition --enable-index-page -Zunstable-options
       - name: Setup Pages
-        if: github.ref_name == 'main' && github.event_name == 'push'
+        if: false && github.ref_name == 'main' && github.event_name == 'push'
         uses: actions/configure-pages@v5
       - name: Upload artifact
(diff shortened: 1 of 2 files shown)


bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -222,51 +247,60 @@ impl MaxTpsArgs {
...
+        stream::iter(0..sample_size)
+            .map(|_| {
+                let idx = random_range(0..pending_txs.len());
+                pending_txs.remove(idx)
+            })
+            .map(|pending_tx| {
+                let hash = *pending_tx.tx_hash();
+                pending_tx
+                    .with_timeout(Some(Duration::from_secs(5)))
+                    .get_receipt()
```
mattsse, line 276: I see

PR #1114, opened 2025-11-28
fix(mempool): ensure fee payer is not blacklisted

### LINKED ISSUE #1110: bug: mempool DOS from blacklisted user
### Describe the bug

the mempool validator only checks if the user has sufficient balance, but does not check if he is allowed to transfer tokens. Because of this, malicious blacklisted users can spam the mempool with transactions, which will fail at the collect_fee_pre_tx function in the handler -> free spamming

0xrusowsky:
## Motivation

closes #1110 

## Solution

update the tx pool validator to ensure that the fee payer is not blacklisted for the fee token

since the logic is more complex than simply reading a slot, i moved the TIP403 registry internal logic into a fn that takes a closure to perform the sload, so that both are in sync.

PS: I have realized that we read precompile-storage data in several places outside precompiles, so i'll address that in the precompile-refactor PR

### DIFF
--- a/crates/alloy/src/network.rs
+++ b/crates/alloy/src/network.rs
@@ -11,11 +11,16 @@ use alloy_primitives::{Address, Bytes, ChainId, TxKind, U256};
 use alloy_provider::fillers::{
     ChainIdFiller, GasFiller, JoinFill, NonceFiller, RecommendedFillers,
 };
-use alloy_rpc_types_eth::AccessList;
+use alloy_rpc_types_eth::{AccessList, Block, Transaction};
 use alloy_signer_local::PrivateKeySigner;
 use tempo_primitives::{
     TempoHeader, TempoReceipt, TempoTxEnvelope, TempoTxType, transaction::TempoTypedTransaction,
 };
+
+/// Set of recommended fillers.
+///
+/// `N` is a nonce filler.
+pub(crate) type TempoFillers<N> = JoinFill<GasFiller, JoinFill<N, ChainIdFiller>>;
 
 /// The Tempo specific configuration of [`Network`] schema and consensus primitives.
 #[derive(Default, Debug, Clone, Copy)]
@@ -29,13 +34,10 @@ impl Network for TempoNetwork {
     type ReceiptEnvelope = TempoReceipt;
     type Header = TempoHeader;
     type TransactionRequest = TempoTransactionRequest;
-    type TransactionResponse = alloy_rpc_types_eth::Transaction<TempoTxEnvelope>;
+    type TransactionResponse = Transaction<TempoTxEnvelope>;
     type ReceiptResponse = TempoTransactionReceipt;
     type HeaderResponse = TempoHeaderResponse;
-    type BlockResponse = alloy_rpc_types_eth::Block<
-        alloy_rpc_types_eth::Transaction<TempoTxEnvelope>,
-        Self::HeaderResponse,
-    >;
+    type BlockResponse = Block<Transaction<TempoTxEnvelope>, Self::HeaderResponse>;
 }
 
 impl TransactionBuilder<TempoNetwork> for TempoTransactionRequest {
@@ -331,7 +333,7 @@ impl TempoTransactionRequest {
 }
 
 impl RecommendedFillers for TempoNetwork {
-    type RecommendedFillers = JoinFill<GasFiller, JoinFill<NonceFiller, ChainIdFiller>>;
+    type RecommendedFillers = TempoFillers<NonceFiller>;
 
     fn recommended_fillers() -> Self::RecommendedFillers {
         Default::default()
--- a/crates/alloy/src/provider/ext.rs
+++ b/crates/alloy/src/provider/ext.rs
@@ -1,1 +1,54 @@
+use alloy_provider::{
+    Identity, ProviderBuilder,
+    fillers::{JoinFill, RecommendedFillers},
+};
 
+use crate::{TempoFillers, TempoNetwork, fillers::Random2DNonceFiller};
+
+/// Extension trait for [`ProviderBuilder`] with Tempo-specific functionality.
+pub trait TempoProviderBuilderExt {
+    /// Returns a provider builder with the recommended Tempo fillers and the random 2D nonce filler.
+    ///
+    /// See [`Random2DNonceFiller`] for more information on random 2D nonces.
+    fn with_random_2d_nonces(
+        self,
+    ) -> ProviderBuilder<
+        Identity,
+        JoinFill<Identity, TempoFillers<Random2DNonceFiller>>,
+        TempoNetwork,
+    >;
+}
+
+impl TempoProviderBuilderExt
+    for ProviderBuilder<
+        Identity,
+        JoinFill<Identity, <TempoNetwork as RecommendedFillers>::RecommendedFillers>,
+        TempoNetwork,
+    >
+{
+    fn with_random_2d_nonces(
+        self,
+    ) -> ProviderBuilder<
+        Identity,
+        JoinFill<Identity, TempoFillers<Random2DNonceFiller>>,
+        TempoNetwork,
+    > {
+        ProviderBuilder::default().filler(TempoFillers::default())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use alloy_provider::{Identity, ProviderBuilder, fillers::JoinFill};
+
+    use crate::{
+        TempoFillers, TempoNetwork, fillers::Random2DNonceFiller,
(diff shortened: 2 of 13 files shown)


crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -307,28 +309,58 @@ impl<'a, S: PrecompileStorageProvider> TIP403Registry<'a, S> {
...
+///
+/// # Returns
+/// - `Ok(true)` if the user is authorized
+/// - `Ok(false)` if not authorized or if policy data cannot be decoded
+/// - `Err(E)` if the storage loading operation fails
+pub fn is_authorized_with<E>(
+    policy_id: u64,
+    user: Address,
+    mut load_storage: impl FnMut(U256) -> core::result::Result<U256, E>,
+) -> core::result::Result<bool, E> {
```
mattsse: this should not need the full path?

crates/transaction-pool/src/transaction.rs
```diff
@@ -109,6 +109,12 @@ pub enum TempoPoolTransactionError {
     )]
     InvalidFeeToken(Address),
 
+    #[error("Fee payer {fee_payer} is blacklisted for fee token: {fee_token}")]
+    BlackListedFeePayer {
```
mattsse: this needs docs, why this happened what this means

crates/revm/src/common.rs
```diff
@@ -200,6 +201,36 @@ pub trait TempoStateAccess<T> {
...
+            tip20::slots::TRANSFER_POLICY_ID_BYTES,
+        ) else {
+            // Should be infallible, but if unable to extract packed value, assume blacklisted.
+            return Ok(false);
+        };
+        self.is_action_authorized(transfer_policy_id, fee_payer)
+    }
+
+    /// Checks if an action is authorized by the TIP403 registry
+    fn is_action_authorized(&mut self, policy_id: u64, user: Address) -> Result<bool, Self::Error> {
```
mattsse: what does action mean here

crates/revm/src/common.rs
```diff
@@ -200,6 +201,36 @@ pub trait TempoStateAccess<T> {
         Ok(self.sload(fee_token, tip20::slots::CURRENCY)? == USD_CURRENCY_SLOT_VALUE)
     }
 
+    /// Checks if the fee payer is not blacklisted for the given token.
+    fn is_valid_fee_payer(
```
mattsse: to be very nitpicky here, we have blacklist vs valid, is this the same or is this different?

crates/revm/src/common.rs
```diff
@@ -200,6 +201,36 @@ pub trait TempoStateAccess<T> {
...
+        // Ensure the fee payer is not blacklisted
+        let Ok(transfer_policy_id) = storage::packing::extract_packed_value(
+            self.sload(fee_token, tip20::slots::TRANSFER_POLICY_ID)?,
+            tip20::slots::TRANSFER_POLICY_ID_OFFSET,
+            tip20::slots::TRANSFER_POLICY_ID_BYTES,
+        ) else {
+            // Should be infallible, but if unable to extract packed value, assume blacklisted.
+            return Ok(false);
+        };
+        self.is_action_authorized(transfer_policy_id, fee_payer)
```
mattsse: this now checks if the transfer policy is authorized, so we're mixing

* valid
* blacklisted
* authorized

and treating this as one and the same, can we unify naming for all of this, because this is a bit confusing, atleast to me

because what this check basically does is "check if the fee payer is authorized to transfer"?

crates/precompiles/src/tip403_registry/mod.rs
mattsse, line 1: is this change here required for the mempool validation logic?

crates/revm/src/common.rs
0xrusowsky: should i rename to call?
basically whatever the policy checks (in this case a transfer, but could be any policy)

crates/revm/src/common.rs
0xrusowsky: yes, i only made the policy check its own fn as i thought we may want to reuse it later.

but maybe i was overabstracting

crates/precompiles/src/tip403_registry/mod.rs
0xrusowsky: i did it to ensure that both impls are in sync.

as i mentioned in the PR description, the way precompile validations are used outside precompile crates is kinda concerning to me, as we went back to manually reading storage slots (which is error prone and dups code)

i already thought of a solution (after the storage API refactor PR), so maybe it is fine to have dupped logic for the time being, as in the other fns that check precompile state

crates/precompiles/src/tip403_registry/mod.rs
0xrusowsky: all the precompile fns use `crate::error::Result` so otherwise there is a type conflict.

if i leave this as `Result` i need to change all the others

crates/revm/src/common.rs
0xrusowsky: dropped the fn completely as it's not needed for anything else for the time being.

addressed in [`082d7f3` (#1114)](https://github.com/tempoxyz/tempo/pull/1114/commits/082d7f3c6b663e218dfb247d01e07f6f13b7708b)

crates/revm/src/common.rs
0xrusowsky: renamed to `can_fee_payer_transfer` and made docs more explicit [`082d7f3` (#1114)](https://github.com/tempoxyz/tempo/pull/1114/commits/082d7f3c6b663e218dfb247d01e07f6f13b7708b)

crates/revm/src/common.rs
0xrusowsky: dropped the fn completely as it's not needed for anything else for the time being.

addressed in [`082d7f3` (#1114)](https://github.com/tempoxyz/tempo/pull/1114/commits/082d7f3c6b663e218dfb247d01e07f6f13b7708b)

crates/transaction-pool/src/validator.rs
```diff
@@ -742,4 +762,129 @@ mod tests {
             panic!("Expected invalid outcome with InvalidValidAfter error");
         }
     }
+
+    #[tokio::test]
+    async fn test_blacklisted_fee_payer_rejected() {
```
klkvr, line 767: imo we should remove all of those tests and rewrite them as e2e tests that deploy actual tokens

right now most of the test is just configuring the storage manually which is somewhat redundant

can be done in a follow-up though

crates/revm/src/common.rs
```diff
@@ -200,6 +201,71 @@ pub trait TempoStateAccess<T> {
         Ok(self.sload(fee_token, tip20::slots::CURRENCY)? == USD_CURRENCY_SLOT_VALUE)
     }
 
+    /// Checks if the fee payer can transfer a given token (is not blacklisted).
+    fn can_fee_payer_transfer(
```
klkvr, line 205: yeah we should definitely find a way to unify this

crates/transaction-pool/src/validator.rs
0xKitsune: +1, sgtm to do this in a follow up.

crates/revm/src/common.rs
```diff
@@ -200,6 +201,71 @@ pub trait TempoStateAccess<T> {
...
+        }
+
+        // Ensure the fee payer is not blacklisted
+        let Ok(transfer_policy_id) = storage::packing::extract_packed_value::<1, u64>(
+            self.sload(fee_token, tip20::slots::TRANSFER_POLICY_ID)?,
+            tip20::slots::TRANSFER_POLICY_ID_OFFSET,
+            <u64 as StorableType>::BYTES,
+        ) else {
+            // Should be infallible, but if unable to extract packed value, assume blacklisted.
+            return Ok(false);
```
0xKitsune, line 223: nit: can we add logging here to warn if this is ever hit? We can also do this in a follow on PR.

crates/revm/src/common.rs
```diff
@@ -200,6 +201,71 @@ pub trait TempoStateAccess<T> {
...
+                    transfer_policy_id.to_be_bytes(),
+                    tip403_registry::slots::POLICY_DATA,
+                ),
+            )?;
+            let Ok(data) = tip403_registry::PolicyData::from_evm_words([policy_data_word]) else {
+                return Ok(false);
+            };
+            let Ok(policy_type) = data.policy_type.try_into() else {
+                return Ok(false);
+            };
```
0xKitsune, line 252: nit: same thing here we should add proper logging in the case that we hit the else case here. This can be done in a future PR.

PR #1115, opened 2025-11-28
feat: fee AMM liqudiity tracker

klkvr:
Introduces and integrates `AmmLiquidityCache` that is a small tracker caching following vlaues:
- Latest fee tokens seen being used by validators, currently in last 100 blocks
- Validator fee token preferences (to optimize populating the above)
- AMM pool -> liquidity mappings

During transaction validation, we check if its fee token matches any of the recently used by validators or if amm pool with any of validator tokens has enough liquidity with user token.

This should allow to reliably reject transactions which are very unlikely to get included without many false positives.

Note that this primarily protects us from transacitons lacking fee AMM liquditiy entering the pool but doesn't enforce that transactions will get evicted once liquidity they depend on is exhausted

### DIFF
--- a/bin/tempo-bench/src/cmd/max_tps.rs
+++ b/bin/tempo-bench/src/cmd/max_tps.rs
@@ -3,7 +3,7 @@ mod dex;
 use itertools::Itertools;
 use reth_tracing::{
     RethTracer, Tracer,
-    tracing::{debug, error, info, warn},
+    tracing::{debug, error, info},
 };
 use tempo_alloy::TempoNetwork;
 
@@ -13,11 +13,11 @@ use alloy::{
     primitives::{Address, B256, BlockNumber, U256},
     providers::{
         DynProvider, PendingTransactionBuilder, PendingTransactionError, Provider, ProviderBuilder,
-        WatchTxError,
+        WatchTxError, fillers::NonceFiller,
     },
     rpc::client::NoParams,
     signers::local::{
-        MnemonicBuilder, PrivateKeySigner,
+        PrivateKeySigner,
         coins_bip39::{English, Mnemonic, MnemonicError},
     },
     transports::http::reqwest::Url,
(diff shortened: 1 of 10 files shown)


crates/transaction-pool/src/amm.rs
```diff
@@ -0,0 +1,99 @@
...
+    storage::mapping_slot,
+    tip_fee_manager::{
+        amm::{Pool, PoolKey, compute_amount_out},
+        slots,
+    },
+    tip20::{address_to_token_id_unchecked, token_id_to_address},
+};
+
+#[derive(Debug, Default)]
+struct AmmLiquidityCacheInner {
```
mattsse: can we put this below the pub outer type

crates/transaction-pool/src/transaction.rs
```diff
@@ -133,6 +133,9 @@ pub enum TempoPoolTransactionError {
     /// Thrown if a AA transaction with a nonce key prefixed with the sub-block prefix marker added to the pool
     #[error("AA transaction with subblock nonce key prefix aren't supported in the pool")]
     SubblockNonceKey,
+
+    #[error("Insufficient liquidity for fee token: {0}")]
```
mattsse: can we elaborate in docs, why this exists, when this is thrown and when this would be resolved, basically "how to get liquidity"

crates/transaction-pool/src/amm.rs
```diff
@@ -0,0 +1,99 @@
...
+            }
+        }
+
+        // If no cache misses were hit, just return false
+        if missing_in_cache.is_empty() {
+            return Ok(false);
+        }
+
+        // Otherwise, load pools that weren't found in cache and check if they have enough liquidity
+        let mut inner = self.inner.write();
```
mattsse: this can deadlock, right? because the read lock is still active

crates/transaction-pool/src/amm.rs
```diff
@@ -0,0 +1,99 @@
...
+            {
+                return Ok(true);
+            }
+
+            let pool_key =
+                PoolKey::new(token_id_to_address(user_id), token_id_to_address(token)).get_id();
+            let slot = mapping_slot(pool_key, slots::POOLS);
+            let pool = state_provider
+                .storage(TIP_FEE_MANAGER_ADDRESS, slot.into())?
+                .unwrap_or_default();
```
mattsse, line 104: we should avoid doing this while holding the lock

crates/transaction-pool/src/amm.rs
```diff
@@ -0,0 +1,237 @@
...
+    /// Reverse index for mapping AMM slot to a pool.
+    slot_to_pool: HashMap<U256, (u64, u64)>,
+
+    /// Latest observed validator tokens.
+    last_seen_tokens: VecDeque<Address>,
+
+    /// Unique tokens that have been seen in the last_seen_tokens.
+    ///
+    /// Ordered by the number of times they've been seen.
+    unique_tokens: Vec<Address>,
```
mattsse, line 230: I see, this is effectively bound by LAST_SEEN_TOKENS_WINDOW so iterating over those items should be quite cheap

crates/transaction-pool/src/amm.rs
```diff
@@ -0,0 +1,237 @@
...
+
+        let preference = if let Some(cached) = cached_preference {
+            cached
+        } else {
+            // If no cached preference, load from state
+            state
+                .state_by_block_hash(header.hash())?
+                .storage(
+                    TIP_FEE_MANAGER_ADDRESS,
+                    mapping_slot(beneficiary, slots::VALIDATOR_TOKENS).into(),
```
mattsse, line 174: these changes should be present in the executionoutcome as well I believe and could combine this with the `Notification::Chain` I think which has the headers

but this also works and having this separate is easier to reason about

PR #1124, opened 2025-11-28
fix!: stablecoin exchange overflow error

legion2002:
Fix arithmetic overflow in StablecoinExchange when querying empty orderbooks

  ### Problem

  The quote_exact_in and quote_exact_out functions in the StablecoinExchange precompile were panicking with attempt
   to add with overflow when querying swap quotes on orderbooks with no liquidity on the ask side.

 ### Root Cause

  When an orderbook has no asks, best_ask_tick is initialized to i16::MAX (32767). The code only checked for
  i16::MIN (no bids) before proceeding:
```
  if current_tick == i16::MIN {
      return Err(StablecoinExchangeError::insufficient_liquidity().into());
  }
```

  When querying the ask side with best_ask_tick = i16::MAX, the code would proceed to next_initialized_ask_tick()
  which attempts tick + 1, causing an integer overflow.

  Fix

  1. In mod.rs: Added check for i16::MAX alongside i16::MIN in both quote_exact_in and quote_exact_out:
  ```
  if current_tick == i16::MIN || current_tick == i16::MAX {
      return Err(StablecoinExchangeError::insufficient_liquidity().into());
  }
```
  3. In orderbook.rs: Added defensive guards in next_initialized_ask_tick and next_initialized_bid_tick to prevent
  overflow/underflow if called with edge values.

 #### Note

  The Solidity spec implementation already handles this correctly by checking type(int16).max for asks and
  type(int16).min for bids separately in each code path. This fix aligns the Rust precompile behavior with the
  Solidity spec.

### DIFF
--- a/bin/tempo-bench/src/cmd/max_tps.rs
+++ b/bin/tempo-bench/src/cmd/max_tps.rs
@@ -3,7 +3,7 @@ mod dex;
 use itertools::Itertools;
 use reth_tracing::{
     RethTracer, Tracer,
-    tracing::{error, info, warn},
+    tracing::{debug, error, info, warn},
 };
 use tempo_alloy::TempoNetwork;
 
@@ -35,6 +35,7 @@ use rayon::iter::{IntoParallelIterator, ParallelIterator};
 use rlimit::Resource;
 use serde::Serialize;
 use std::{
+    collections::VecDeque,
     fs::File,
     io::BufWriter,
     num::NonZeroU32,
@@ -163,7 +164,9 @@ impl MaxTpsArgs {
 
         let target_url = self.target_urls[0].clone();
         if self.target_urls.len() > 1 {
-            warn!("Multiple target URLs provided, but only the first one will be used")
+            warn!(
+                "Multiple target URLs provided, but only the first one will be used for sending transactions"
+            )
         }
         let provider = ProviderBuilder::new_with_network::<TempoNetwork>()
             .connect_http(target_url.clone())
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: crates/precompiles/src/stablecoin_exchange/mod.rs [hunk 1]
```diff
@@ -1556,7 +1557,8 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
             orderbook.best_ask_tick
         };
 
-        if current_tick == i16::MIN {
+        // Check for no liquidity: i16::MIN means no bids, i16::MAX means no asks
+        if current_tick == i16::MIN || current_tick == i16::MAX {
```
fgimenez: nit maybe we could use a logic similar to `get_best_price_level` that checks each bid-ask case separately:
```
  let current_tick = if is_bid {
      if orderbook.best_bid_tick == i16::MIN {
          return Err(...);
      }
      orderbook.best_bid_tick
  } else {
      if orderbook.best_ask_tick == i16::MAX {
          return Err(...);
      }
      orderbook.best_ask_tick
  };
```
instead of the combined check (`i16::MIN` is the sentinel for bids and `i16::MAX` is the sentinel for asks) wdyt

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -1361,7 +1361,8 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
         } else {
             orderbook.best_ask_tick
         };
-        if current_tick == i16::MIN {
+        // Check for no liquidity: i16::MIN means no bids, i16::MAX means no asks
+        if current_tick == i16::MIN || current_tick == i16::MAX {
```
0xKitsune: Technically this would require a hardfork if the function was ever called onchain. In practice it should be fine as long as the snapshot tests pass, but calling it out for visibility. Since the repo isn’t public yet and partners are not likely to invoke this onchain, the risk is low. If we want to be extra safe, we can just gate it behind Allegretto.

Wdyt @fgimenez @klkvr

crates/precompiles/src/stablecoin_exchange/orderbook.rs
```diff
@@ -409,6 +409,10 @@ impl Orderbook {
         book_key: B256,
         tick: i16,
     ) -> (i16, bool) {
+        // Guard against overflow when tick is at or above MAX_TICK
+        if tick >= MAX_TICK {
+            return (MAX_TICK, false);
+        }
```
0xKitsune, line 418: Same thing here, this should be fine as it is unlikely this path has been hit in current persistent testnet, but this would require a hardfork to not break syncing.

crates/precompiles/src/stablecoin_exchange/mod.rs
0xrusowsky: i already opened a PR to address this as it was reported by zellic https://github.com/tempoxyz/tempo/pull/1129

PR #1126, opened 2025-11-28
feat: re-broadcast built subblocks multiple times

klkvr:
It appears that right now on testnet there's a race condition when some of the validators might send a subblock to proposer earlier than proposer processed the current tip. This causes us to hit this error:
https://github.com/tempoxyz/tempo/blob/52044ec67081f730c452ca0d47e4ff11d24cda66/crates/commonware-node/src/subblocks.rs#L365-L371

This PR changes logic so that subblocks are attempted to be broadcasted again (by default every 50ms) to make sure that proposer accumulates all of the subblocks once they catch up to the tip

Upon receiving a subblock that matches our chain view, we send and acknowledgment — 32 bytes of subblock hash. Upon receiving an acknowledgment, subblock sender stops broadcasting it to us to avoid extra networking overhead

### DIFF
--- a/bin/tempo-bench/src/cmd/max_tps.rs
+++ b/bin/tempo-bench/src/cmd/max_tps.rs
@@ -3,7 +3,7 @@ mod dex;
 use itertools::Itertools;
 use reth_tracing::{
     RethTracer, Tracer,
-    tracing::{error, info, warn},
+    tracing::{debug, error, info, warn},
 };
 use tempo_alloy::TempoNetwork;
 
@@ -35,6 +35,7 @@ use rayon::iter::{IntoParallelIterator, ParallelIterator};
 use rlimit::Resource;
 use serde::Serialize;
 use std::{
+    collections::VecDeque,
     fs::File,
     io::BufWriter,
     num::NonZeroU32,
@@ -163,7 +164,9 @@ impl MaxTpsArgs {
 
         let target_url = self.target_urls[0].clone();
         if self.target_urls.len() > 1 {
-            warn!("Multiple target URLs provided, but only the first one will be used")
+            warn!(
+                "Multiple target URLs provided, but only the first one will be used for sending transactions"
+            )
         }
         let provider = ProviderBuilder::new_with_network::<TempoNetwork>()
             .connect_http(target_url.clone())
@@ -191,13 +194,16 @@ impl MaxTpsArgs {
             .collect::<eyre::Result<Vec<_>>>()?;
 
         if self.clear_txpool {
-            let transactions: u64 = provider
-                .raw_request("admin_clearTxpool".into(), NoParams::default())
-                .await
-                .context(
-                    "Failed to clear transaction pool. Is `admin_clearTxpool` RPC method available?",
-                )?;
-            info!(transactions, "Cleared transaction pool");
+            for target_url in &self.target_urls {
+                let transactions: u64 = ProviderBuilder::new()
+                    .connect_http(target_url.clone())
+                    .raw_request("admin_clearTxpool".into(), NoParams::default())
+                    .await
+                    .context(
+                        "Failed to clear transaction pool for {target_url}. Is `admin_clearTxpool` RPC method available?",
+                    )?;
+                info!(%target_url, transactions, "Cleared transaction pool");
+            }
         }
 
         // Fund accounts from faucet if requested
@@ -248,7 +254,6 @@ impl MaxTpsArgs {
         .context("Failed to generate transactions")?;
 
         // Send transactions
-        let start_block_number = provider.get_block_number().await?;
         let mut pending_txs = send_transactions(
             transactions,
             self.max_concurrent_requests,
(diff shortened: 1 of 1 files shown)


crates/commonware-node/src/subblocks.rs
```diff
@@ -167,13 +179,21 @@ impl<TContext: Spawner + Metrics + Pacer> Actor<TContext> {
...
-                    let task = self.subblock_builder_handle.take().unwrap();
-                    self.on_built_subblock(subblock, task.proposer, &mut network_tx).await;
+                    let task = self.our_subblock.take_task().unwrap();
+                    self.on_built_subblock(subblock, task.proposer).await;
+                }
+                // Handle subblocks broadcast.
+                _ = if let Some(broadcast_interval) = broadcast_interval {
+                    broadcast_interval.fuse()
+                } else {
+                    futures::future::Fuse::terminated()
```
SuperFluffy, line 194: Had to do a double take of this pattern.

This is fundamentally another way of doing this, right?

```rust
_ = async { maybe_fut.as_mut().unwrap().await }, if maybe_fut.is_some() => { ... }
```

It's a bit confusing seeing this at first (because fusing the `&mut Fut` is kinda pointless I was unsure what this is doing).

crates/commonware-node/src/subblocks.rs
```diff
@@ -151,6 +157,12 @@ impl<TContext: Spawner + Metrics + Pacer> Actor<TContext> {
         ),
     ) {
         loop {
+            let (subblock_task, broadcast_interval) = match &mut self.our_subblock {
```
SuperFluffy, line 160: Could this be made nicer by having `PendingSubblock` be a future that returns an enum?

crates/commonware-node/src/subblocks.rs
```diff
@@ -353,7 +353,28 @@ impl<TContext: Spawner + Metrics + Pacer> Actor<TContext> {
...
     #[instrument(skip_all, err(level = Level::WARN), fields(sender = %sender, msg_bytes = message.len()))]
-    fn on_network_message(&mut self, sender: PublicKey, message: bytes::Bytes) -> eyre::Result<()> {
+    async fn on_network_message(
+        &mut self,
+        sender: PublicKey,
+        message: bytes::Bytes,
+        network_tx: &mut impl Sender<PublicKey = PublicKey>,
+    ) -> eyre::Result<()> {
+        // Process acknowledgements
+        if message.len() == 32 {
```
SuperFluffy: Create an enum for the different kinds of messages we can get over the network? I guess the ship has sailed for tagging them like we do in the `ceremony::payload::Message` and `ceremony::payload::Payload` because that would break the wire format, but it would still be nicer to read them here.

crates/commonware-node/src/args.rs
```diff
@@ -114,4 +114,12 @@ pub struct Args {
...
     )]
     pub allow_unregistered_handshakes: bool,
+
+    /// The interval at which to broadcast subblocks to the next proposer.
+    /// Each built subblock is immediately broadcasted to the next proposer (if it's known).
+    /// We broadcast subblock every `subblock-broadcast-interval` to ensure the next
+    /// proposer is aware of the subblock even if they were slightly behind the chain
+    /// once we sent it in the first time.
+    #[arg(long = "consensus.subblock-broadcast-interval", default_value = "50ms")]
+    pub subblock_broadcast_interval: jiff::SignedDuration,
```
SuperFluffy, line 124: Let's align this with the other args (`wait_for_proposal`, `wait_to_rebrodcast_nullify`).
```suggestion
    pub wait_to_rebroadcast_subblock: jiff::SignedDuration,
```

(and the CLI name of course)

crates/commonware-node/src/subblocks.rs
klkvr: I don't think this will work because the `unwrap` will be invoked anyway

> Additionally, each branch may include an optional if precondition. If the precondition returns false, then the branch is disabled. The provided <async expression> is still evaluated but the resulting future is never polled. This capability is useful when using select! within a loop.

https://docs.rs/tokio/latest/tokio/macro.select.html

another option here is to have `future::pending().await` in the else branch but both are kinda similar code-wise

crates/commonware-node/src/subblocks.rs
klkvr: it could but this match seemed easier to understand than forcing reader to go look at another `Future` impl and an extra type it returns

crates/commonware-node/src/args.rs
klkvr: this is kinda different because here it's an actual `interval` vs a `timeout`. naming is partially inspired by reth's payload builder args which we also use

https://github.com/paradigmxyz/reth/blob/78f8dddfc2a09a41ea1f4627cef55617a6440049/crates/node/core/src/args/payload_builder.rs#L94-L105

crates/commonware-node/src/subblocks.rs
klkvr: added in 86f2768

crates/commonware-node/src/args.rs
SuperFluffy: Yeah, but this is in the consensus context. You are also prefixing it with `consensus.`

PR #1130, opened 2025-11-28
fix(precompiles): account for `is_bid` in `quote_exact_in` calc

### LINKED ISSUE #1127: bug(precompiles): stablecoin exchange doesn't account for `is_bid` when calculating returns
### Describe the feature

In `crates/precompiles/src/stablecoin_exchange/mod.rs`, the `quote_exact_in` function does not take the is_bid into account at all when calculating returns. This may lead to a significant deviation between the `min_amount_out` passed by the user in subsequent transactions and the actual value, thereby causing transaction failures.
By contrast, `quote_exact_out` considers the `is_bid` of the order.
```rs
fn quote_exact_out(&mut self, book_key: B256, amount_out: u128, is_bid: bool) -> Result<u128> {
    let (fill_amount, amount_in_tick) = if is_bid { // ===> with is_bid
        ......
    } else {
        ......
    };
}
......
fn quote_exact_in(&mut self, book_key: B256, amount_in: u128, is_bid: bool) -> Result<u128> {
    let fill_amount = if remaining_in > level.total_liquidity { // ===> no is_bid
        level.total_liquidity
    } else {
        remaining_in
    };
    ......
}
```

### Additional context

_No response_

0xrusowsky:
## Motivation

closes #1127 

## Solution

properly account for `is_bid` also in `quote_extact_in`

### DIFF
--- a/crates/precompiles/src/tip403_registry/mod.rs
+++ b/crates/precompiles/src/tip403_registry/mod.rs
@@ -306,6 +306,7 @@ impl<'a, S: PrecompileStorageProvider> TIP403Registry<'a, S> {
         self.sstore_policy_set(policy_id, account, value)
     }
 
+    // NOTE: must be synced with `fn can_fee_payer_transfer` @crates/revm/src/common.rs
     fn is_authorized_internal(&mut self, policy_id: u64, user: Address) -> Result<bool> {
         // Special case for always-allow and always-reject policies
         if policy_id < 2 {
@@ -326,7 +327,6 @@ impl<'a, S: PrecompileStorageProvider> TIP403Registry<'a, S> {
             ITIP403Registry::PolicyType::BLACKLIST => !is_in_set,
             ITIP403Registry::PolicyType::__Invalid => false,
         };
-
         Ok(auth)
     }
 }
--- a/crates/revm/src/common.rs
+++ b/crates/revm/src/common.rs
@@ -9,13 +9,14 @@ use revm::{
 use tempo_chainspec::hardfork::TempoHardfork;
 use tempo_contracts::precompiles::{
     DEFAULT_FEE_TOKEN_POST_ALLEGRETTO, DEFAULT_FEE_TOKEN_PRE_ALLEGRETTO, IFeeManager,
-    IStablecoinExchange, PATH_USD_ADDRESS, STABLECOIN_EXCHANGE_ADDRESS,
+    IStablecoinExchange, ITIP403Registry, PATH_USD_ADDRESS, STABLECOIN_EXCHANGE_ADDRESS,
 };
 use tempo_precompiles::{
-    TIP_FEE_MANAGER_ADDRESS,
-    storage::slots::mapping_slot,
+    TIP_FEE_MANAGER_ADDRESS, TIP403_REGISTRY_ADDRESS,
+    storage::{self, Storable, StorableType, double_mapping_slot, slots::mapping_slot},
     tip_fee_manager,
     tip20::{self, is_tip20},
+    tip403_registry,
 };
 use tempo_primitives::TempoTxEnvelope;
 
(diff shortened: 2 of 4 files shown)


crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -1577,18 +1577,38 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
...
+                    level.total_liquidity
+                } else {
+                    base_to_get
+                };
+                let quote_consumed = fill_amount
+                    .checked_mul(price as u128)
+                    .and_then(|v| v.checked_div(orderbook::PRICE_SCALE as u128))
+                    .ok_or(TempoPrecompileError::under_overflow())?;
+                (fill_amount, fill_amount, quote_consumed)
             };
```
0xKitsune: I believe this needs to be gated behind a hardfork.

0xrusowsky: i forgot to push, will do once back at my laptop

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -1611,39 +1611,98 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
...
+                let (fill_amount, amount_out_tick, amount_consumed) = if is_bid {
+                    // For bids: remaining_in is base, amount_out is quote
+                    let fill_amount = if remaining_in > level.total_liquidity {
+                        level.total_liquidity
+                    } else {
+                        remaining_in
+                    };
+                    let quote_out = fill_amount
+                        .checked_mul(price as u128)
+                        .and_then(|v| v.checked_div(orderbook::PRICE_SCALE as u128))
```
mattsse: we dont need checked div here because rhs is > 0

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -1611,39 +1611,98 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
...
+                        .and_then(|v| v.checked_div(price as u128))
+                        .ok_or(TempoPrecompileError::under_overflow())?;
+                    let fill_amount = if base_to_get > level.total_liquidity {
+                        level.total_liquidity
+                    } else {
+                        base_to_get
+                    };
+                    let quote_consumed = fill_amount
+                        .checked_mul(price as u128)
+                        .and_then(|v| v.checked_div(orderbook::PRICE_SCALE as u128))
```
mattsse: same

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -1611,39 +1611,98 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
...
-                .expect("Amount out calculation overflow");
+            if self.storage.spec().is_allegretto() {
+                // Post-allegretto: logic accounts for `is_bid`
+                let (fill_amount, amount_out_tick, amount_consumed) = if is_bid {
+                    // For bids: remaining_in is base, amount_out is quote
+                    let fill_amount = if remaining_in > level.total_liquidity {
+                        level.total_liquidity
+                    } else {
+                        remaining_in
+                    };
```
0xKitsune: nit: can be simplified to `let fill = remaining_in.min(level.total_liquidity);`

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -1611,39 +1611,98 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
...
+                    // For asks: remaining_in is quote, amount_out is base
+                    let base_to_get = remaining_in
+                        .checked_mul(orderbook::PRICE_SCALE as u128)
+                        .and_then(|v| v.checked_div(price as u128))
+                        .ok_or(TempoPrecompileError::under_overflow())?;
+                    let fill_amount = if base_to_get > level.total_liquidity {
+                        level.total_liquidity
+                    } else {
+                        base_to_get
+                    };
```
0xKitsune: nit: can be simplified to `base_to_get.min(level.total_liquidity)`.

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -1611,39 +1611,98 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
...
 
-                if !initialized && remaining_in > 0 {
-                    return Err(StablecoinExchangeError::insufficient_liquidity().into());
+                    if !initialized && remaining_in > 0 {
+                        return Err(StablecoinExchangeError::insufficient_liquidity().into());
+                    }
+                    current_tick = next_tick;
+                } else {
+                    break;
                 }
```
0xKitsune: It looks like this is the same logic pre allegretto. Can we refactor to consolidate this?

Issue #1133, opened 2025-11-28
Bug: malicious proposer can include dealings with unknown dealer key, causing DKG ceremony to panic


SuperFluffy:
### Describe the bug

A malicious dealer can insert intermediate dealings into the header `extra_data` field with a dealer that does not participate in the DKG ceremony.

If the dealing is constructed such that the dealing would lead to a disqualification, the ceremony tracks the dealer in its arbiter state:

https://github.com/tempoxyz/tempo/blob/f6cb7bf062ada3c69f09594ac2d5566f3f195c69/crates/commonware-node/src/dkg/ceremony/mod.rs#L605

On finalization, the arbiter assumes the invariant that only participating dealers are members of its disqualified list, leading to a panic:
https://github.com/commonwarexyz/monorepo/blob/f6ac06a8895020492f8c4b55b333dda033bf02ce/cryptography/src/bls12381/dkg/arbiter.rs#L212

This is tracked upstream by commonware: https://github.com/commonwarexyz/monorepo/issues/1996

Solution: the ceremony must filter ignore public keys that are not dealers. Furthermore, blocks that contain intermediate dealing outcomes where the dealer is not the proposer should be blocked.

### Steps to reproduce

Create a proposer with a malicious payload. Observe the panic.

### Logs

```text

```

### Container Type

Not running in a container

### What version/commit are you on?

This problem was reported for https://github.com/tempoxyz/tempo/commit/4004c3f0dfad2fec3fa17dcd5c9ca8d214006e42, but exists as of https://github.com/tempoxyz/tempo/commit/f6cb7bf062ada3c69f09594ac2d5566f3f195c69

### Code of Conduct

- [x] I agree to follow the Code of Conduct

PR #1134, opened 2025-11-28
fix(consensus): reject proposals, DKG dealings from wrong dealers

SuperFluffy:
Closes https://github.com/tempoxyz/tempo/issues/1133

This patch addresses #1133 in two ways:

1. proposals that contain intermediate DKG dealings that do not belong to the proposer are rejected. This blocks malicious dealers from getting any blocks in that contain unaccounted for information.
2. intermediate DKG dealings are not tracked/included in the DKG arbiter if they are for a dealer that is not a member of the ceremony dealer committee. This ensures the that DKG arbiter invariant is not violated.

This is also tracked upstream by https://github.com/commonwarexyz/monorepo/issues/1996

### DIFF
--- a/crates/alloy/src/network.rs
+++ b/crates/alloy/src/network.rs
@@ -11,11 +11,16 @@ use alloy_primitives::{Address, Bytes, ChainId, TxKind, U256};
 use alloy_provider::fillers::{
     ChainIdFiller, GasFiller, JoinFill, NonceFiller, RecommendedFillers,
 };
-use alloy_rpc_types_eth::AccessList;
+use alloy_rpc_types_eth::{AccessList, Block, Transaction};
 use alloy_signer_local::PrivateKeySigner;
 use tempo_primitives::{
     TempoHeader, TempoReceipt, TempoTxEnvelope, TempoTxType, transaction::TempoTypedTransaction,
 };
+
+/// Set of recommended fillers.
+///
+/// `N` is a nonce filler.
+pub(crate) type TempoFillers<N> = JoinFill<GasFiller, JoinFill<N, ChainIdFiller>>;
 
 /// The Tempo specific configuration of [`Network`] schema and consensus primitives.
 #[derive(Default, Debug, Clone, Copy)]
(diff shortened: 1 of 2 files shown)


crates/commonware-node/src/dkg/ceremony/mod.rs
```diff
@@ -574,6 +574,12 @@ where
             self.epoch(),
         );
 
+        ensure!(
```
patrick-ogrady, line 612: nit: You shouldn't need to re-verify anything about the outcome here (AFAICT `process_dealings_in_block` only called post-finalization, which means you will only get here if at least f+1 honest have already verified) but no harm in being cautious.

SuperFluffy: I am going to leave this here until this is addressed upstream: I should not be able to violate the arbiter invariants. For the time being, I will let this serve as a reminder.

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -848,6 +832,68 @@ async fn verify_block<TContext: Pacer>(
         }
     }
 }
+
+#[instrument(skip_all, err(Display))]
+async fn verify_header_extra_data(
```
klkvr, line 837: this is not necessary right? i.e if we just let any extradata and skip invalid dealings in DKG actor we will also be fine?

it does kind of make sense to de the validation here though because we don't want validators being able to fill every built block with 10Kb of useless data

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -229,9 +235,40 @@ where
...
+                // is in. This is a request when verifying proposals. It relies on the
+                // fact that a new epoch (and hence a different hardfork regime) will
+                // only be entered once the finalized height of the current epoch was seen.
+                //
+                // Furthermore, extra data headers are only checked for intermediate
+                // dealings up but excluding the last height of an epoch.
+                //
+                // In other words: no dealing will ever have to be verified if it is
+                // for another epoch than the currently latest one.
+                super::Command::VerifyDealing(verify_dealing) => {
```
klkvr, line 249: should this ensure that there's one dealing per validator per epoch?

crates/commonware-node/src/consensus/application/actor.rs
SuperFluffy: So right now I am actually pretty lax about this:

1. the last block *must* contain the DKG outcome (this one is strict - no DKG outcome, no notarization)
2. but the other blocks *can* contain a dealing. If we cannot parse the dealing, then we just skip over and ignore the extra data - you can write any junk you want into it. If and only if we can read the dealing we perform extra checks on it (mainly that the signature pans out and that the proposer included its own proposal).

Do you think I should make this more strict so that blocks may only contain the DKG outcome and the dealings?

crates/commonware-node/src/dkg/manager/actor/mod.rs
SuperFluffy: You mean for the case where a proposer has already included their dealing into the ceremony (and where we have already read it from a finalize block)? That's a good idea! I guess extra data is free, right?

SuperFluffy: This requires more changes than I thought; I will do that in a followup: :https://github.com/tempoxyz/tempo/issues/1159

crates/commonware-node/src/consensus/application/actor.rs
SuperFluffy: https://github.com/tempoxyz/tempo/issues/1161

PR #1137, opened 2025-11-28
test(precompiles): assert unique error selectors

shekhirin:
`Unauthorized` selector exists in four interfaces. This can result in such error message when trying to mint tokens on a TIP-20 that you're unauthorized to do so:
```console
2025-11-28T18:49:22.507718Z  INFO Minting tokens mint_amount=340282366920938463463374607431768211455
Error: Failed to mint tokens

Caused by:
    server returned an error response: error code 3: execution reverted: Validator config error: Unauthorized(Unauthorized), data: "0x82b42900"

Location:
    bin/tempo-bench/src/cmd/max_tps.rs:761:27
```

This PR doesn't solve the issue, but rather just adds an assertion and a test that fail with the current selectors.


crates/precompiles/src/error.rs
```diff
@@ -106,6 +107,7 @@ pub fn add_errors_to_registry<T: SolInterface>(
                     })
             }),
         );
+        assert!(old_selector.is_none(), "Duplicate selector {selector:x?}");
```
mattsse: this assert isn't great, because we would encounter this at runtime I belive:

https://github.com/tempoxyz/tempo/blob/cf857ae280a0b3347e98c4098319bc61784b6c0b/crates/precompiles/src/error.rs#L145-L146

can we make this a debug_assert and ~~also add a test that obtains the error registry~~

mattsse (changes requested):

mattsse (approved):
I guess the assert is covered via the test

0xKitsune (approved):
Lgtm

Just noting that we should expect the need to change the test suite and align spec/reference contracts/docs if the error signature is changing.

PR #1138, opened 2025-11-28
feat(alloy): random 2d nonce filler

shekhirin:
https://github.com/tempoxyz/tempo/blob/39ee8535c4b0cbeb7f7dca487b84f319b0a8f83b/crates/alloy/src/fillers/nonce.rs#L10-L12

### DIFF
--- a/crates/node/src/node.rs
+++ b/crates/node/src/node.rs
@@ -41,7 +41,7 @@ use tempo_payload_builder::TempoPayloadBuilder;
 use tempo_payload_types::TempoPayloadAttributes;
 use tempo_primitives::{TempoHeader, TempoPrimitives, TempoTxEnvelope, TempoTxType};
 use tempo_transaction_pool::{
-    AA2dNonceKeys, AA2dPool, AA2dPoolConfig, TempoTransactionPool,
+    AA2dNonceKeys, AA2dPool, AA2dPoolConfig, TempoTransactionPool, amm::AmmLiquidityCache,
     validator::TempoTransactionValidator,
 };
 
@@ -442,9 +442,15 @@ where
         };
         let nonce_keys = AA2dNonceKeys::default();
         let aa_2d_pool = AA2dPool::new(aa_2d_config, nonce_keys.clone());
+        let amm_liquidity_cache = AmmLiquidityCache::new(ctx.provider())?;
 
         let validator = validator.map(|v| {
-            TempoTransactionValidator::new(v, nonce_keys.clone(), self.aa_valid_after_max_secs)
+            TempoTransactionValidator::new(
+                v,
+                nonce_keys.clone(),
+                self.aa_valid_after_max_secs,
+                amm_liquidity_cache.clone(),
+            )
         });
         let protocol_pool = TxPoolBuilder::new(ctx)
             .with_validator(validator)
@@ -468,6 +474,12 @@ where
                 transaction_pool.clone(),
                 ctx.provider().canonical_state_stream(),
             ),
+        );
+
+        // Spawn AMM liquidity cache maintenance task
+        ctx.task_executor().spawn_critical(
+            "txpool maintenance - amm liquidity cache",
+            tempo_transaction_pool::maintain::maintain_amm_cache(transaction_pool.clone()),
         );
 
         info!(target: "reth::cli", "Transaction pool initialized");
--- a/crates/node/tests/it/liquidity.rs
+++ b/crates/node/tests/it/liquidity.rs
@@ -155,43 +155,31 @@ async fn test_block_building_insufficient_fee_amm_liquidity() -> eyre::Result<()
     println!("Sending {num_payment_txs} payment transactions that require fee swaps...");
 
     let mut transactions_included = 0;
-    let mut transactions_timed_out = 0;
+    let mut transactions_rejected = 0;
+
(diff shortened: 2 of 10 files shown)


crates/alloy/src/fillers/nonce.rs
```diff
@@ -0,0 +1,93 @@
...
+    fn status(&self, tx: &N::TransactionRequest) -> FillerControlFlow {
+        if tx.nonce().is_some() || tx.nonce_key.is_some() {
+            return FillerControlFlow::Finished;
+        }
+        FillerControlFlow::Ready
+    }
+
+    fn fill_sync(&self, tx: &mut SendableTx<N>) {
+        if let Some(builder) = tx.as_mut_builder() {
+            builder.set_nonce_key(U256::random());
```
klkvr: we need to make sure that this doesn't use subblock nonce prefix by accident

https://github.com/tempoxyz/tempo/blob/efe2861a5bee8e09c19839f8ff4f4e5c16191059/crates/primitives/src/subblock.rs#L9-L16
https://github.com/tempoxyz/tempo/blob/efe2861a5bee8e09c19839f8ff4f4e5c16191059/crates/transaction-pool/src/validator.rs#L335-L343

crates/alloy/src/fillers/nonce.rs
```diff
@@ -0,0 +1,93 @@
...
+#[derive(Clone, Copy, Debug, Default)]
+pub struct Random2DNonceFiller;
+
+impl<N: Network<TransactionRequest = TempoTransactionRequest>> TxFiller<N> for Random2DNonceFiller {
+    type Fillable = ();
+
+    fn status(&self, tx: &N::TransactionRequest) -> FillerControlFlow {
+        if tx.nonce().is_some() || tx.nonce_key.is_some() {
+            return FillerControlFlow::Finished;
+        }
```
klkvr, line 23: given that `ProviderBuilder::new_with_network` configures default `NonceFiller` by default I guess this won't work properly?

I guess we could add an ext trait that would enable `ProviderBuilder::new_with_network::<TempoNetwork>().with_random_nonces()` API that'd internally just replace the default nonce filler with `Random2DNonceFiller`

crates/alloy/src/fillers/nonce.rs
shekhirin: TIL, fixed

crates/alloy/src/fillers/nonce.rs
shekhirin: Hmm, I don't think we can replace existing fillers? Will create a separate issue to resolve this DX problem

crates/alloy/src/fillers/nonce.rs
```diff
@@ -0,0 +1,101 @@
...
+    fillers::{FillerControlFlow, TxFiller},
+};
+use alloy_transport::TransportResult;
+use tempo_primitives::subblock::has_sub_block_nonce_key_prefix;
+
+/// A [`TxFiller`] that populates the [`TxAA`](`tempo_primitives::TxAA`) transaction with a random `nonce_key`, and `nonce` set to `0`.
+///
+/// This filler can be used to avoid nonce gaps by having a random 2D nonce key that doesn't conflict with any other transactions.
+#[derive(Clone, Copy, Debug, Default)]
+pub struct Random2DNonceFiller;
```
mattsse, line 15: I see so this should ideally always produce an executable tx because new key and nonce 0

crates/alloy/src/fillers/nonce.rs
mattsse: we can however change the 

```
impl RecommendedFillers for Ethereum {
    type RecommendedFillers =
        JoinFill<GasFiller, JoinFill<BlobGasFiller, JoinFill<NonceFiller, ChainIdFiller>>>;

    fn recommended_fillers() -> Self::RecommendedFillers {
        Default::default()
    }
}
```

impl for temponetwork

crates/alloy/src/fillers/nonce.rs
```diff
@@ -0,0 +1,101 @@
...
+
+impl<N: Network<TransactionRequest = TempoTransactionRequest>> TxFiller<N> for Random2DNonceFiller {
```
mattsse, line 17: this assumes that only aa txs will be used, should we clarify this or can we derive this info from a different field?

crates/alloy/src/fillers/nonce.rs
shekhirin: I thought about it, but think it can be an unexpected behaviour to have this 2D nonce filler a default. Maybe no filler at all, and an extension trait to either have normal 1D filler, or a 2D?

crates/alloy/src/fillers/nonce.rs
shekhirin: this comment is enough I think? https://github.com/tempoxyz/tempo/blob/1ff0528b9b3cc2aa8bf3d52a201d4178496b576b/crates/alloy/src/fillers/nonce.rs#L11

crates/alloy/src/fillers/nonce.rs
klkvr: I guess we could wrap an default nonce filler and make the new one delegate to it by default

and then we could maybe have a flag on our custom filler that we can flip through an ext trait to enable random nonces

PR #1139, opened 2025-11-28
feat(tempo-bench): send AA transactions and use 2D nonces

shekhirin:
Use AA transactions with random 2D nonces in `tempo-bench`. DEX setup transactions are still done with 1D nonces.

This is needed for two reasons:
1. Test 2D nonces
2. Not worry about nonce gaps caused by transactions stuck in the transaction pool

### DIFF
--- a/bin/tempo-bench/src/cmd/max_tps/dex.rs
+++ b/bin/tempo-bench/src/cmd/max_tps/dex.rs
@@ -58,7 +58,6 @@ pub(super) async fn setup(
                 let exchange = exchange.clone();
                 Box::pin(async move {
                     let tx = exchange.createPair(token);
-
                     tx.send().await
                 }) as BoxFuture<'static, _>
             })


bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -847,24 +851,27 @@ async fn assert_receipts<R: ReceiptResponse, F: Future<Output = eyre::Result<R>>
...
 }
 
+async fn assert_receipt<R: ReceiptResponse>(receipt: R) -> eyre::Result<()> {
+    eyre::ensure!(
+        receipt.status(),
+        "Transaction {} failed",
+        receipt.transaction_hash()
+    );
+    Ok(())
+}
```
mattsse, line 865: gonna upstream this to alloy as fn ensure_success

PR #1140, opened 2025-11-28
feat: add `burnFrom` to TIP-20

danrobinson:
Updates implementation to match https://github.com/tempoxyz/docs/pull/349


crates/precompiles/src/tip20/mod.rs
```diff
@@ -518,6 +527,43 @@ impl<'a, S: PrecompileStorageProvider> TIP20Token<'a, S> {
...
+
+        // Prevent burning from `FeeManager` and `StablecoinExchange` to protect accounting invariants
+        if self.storage.spec().is_allegretto()
+            && matches!(
+                call.from,
+                TIP_FEE_MANAGER_ADDRESS | STABLECOIN_EXCHANGE_ADDRESS
+            )
+        {
+            return Err(TIP20Error::protected_address().into());
+        }
```
0xKitsune, line 540: Since this function did not exist pre `allegretto`, we do not need to gate this logic behind the hardfork here. We do need to gate the `burn_from` call during dispatch though.

crates/precompiles/src/path_usd/dispatch.rs
```diff
@@ -79,6 +79,9 @@ impl<S: PrecompileStorageProvider> Precompile for PathUSD<'_, S> {
             ITIP20::BURN_BLOCKED_ROLECall::SELECTOR => {
                 view::<ITIP20::BURN_BLOCKED_ROLECall>(calldata, |_| Ok(Self::burn_blocked_role()))
             }
+            ITIP20::BURN_FROM_ROLECall::SELECTOR => {
+                view::<ITIP20::BURN_FROM_ROLECall>(calldata, |_| Ok(Self::burn_from_role()))
+            }
```
0xKitsune: We do not need to add this here since all post `allegretto` calls will use the TIP20 implementation.

```rust
   if self.token.storage().spec().is_allegretto()
            && selector != ITIP20::nameCall::SELECTOR
            && selector != ITIP20::symbolCall::SELECTOR
        {
            return self.token.call(calldata, msg_sender);
        }
```

crates/precompiles/src/path_usd/dispatch.rs
```diff
@@ -119,6 +122,11 @@ impl<S: PrecompileStorageProvider> Precompile for PathUSD<'_, S> {
                     self.token.burn_blocked(sender, call)
                 })
             }
+            ITIP20::burnFromCall::SELECTOR => {
+                mutate_void::<ITIP20::burnFromCall>(calldata, msg_sender, |sender, call| {
+                    self.token.burn_from(sender, call)
+                })
+            }
```
0xKitsune: Same thing here, we can remove this since the PathUSD precompile is not used post `allegretto`.

crates/precompiles/src/path_usd/mod.rs
```diff
@@ -245,6 +245,14 @@ impl<'a, S: PrecompileStorageProvider> PathUSD<'a, S> {
         TIP20Token::<S>::burn_blocked_role()
     }
 
+    /// Returns the BURN_FROM_ROLE constant
+    ///
+    /// This role identifier grants permission to burn tokens from any account.
+    /// The role is computed as `keccak256("BURN_FROM_ROLE")`.
+    pub fn burn_from_role() -> B256 {
+        TIP20Token::<S>::burn_from_role()
+    }
```
0xKitsune: Similar to the comments in `dispatch.rs`, this will never be called since post `allegretto` calls are forwarded to the TIP20 implementation.

crates/precompiles/src/tip20/dispatch.rs
```diff
@@ -64,6 +64,9 @@ impl<'a, S: PrecompileStorageProvider> Precompile for TIP20Token<'a, S> {
             ITIP20::BURN_BLOCKED_ROLECall::SELECTOR => {
                 view::<ITIP20::BURN_BLOCKED_ROLECall>(calldata, |_| Ok(Self::burn_blocked_role()))
             }
+            ITIP20::BURN_FROM_ROLECall::SELECTOR => {
+                view::<ITIP20::BURN_FROM_ROLECall>(calldata, |_| Ok(Self::burn_from_role()))
+            }
```
0xKitsune, line 77: We should gate this behind the allegretto hardfork and return `UnknownSelector`. 


```rust
  if !self.storage.spec().is_allegretto() {
        return unknown_selector(
            selector,
            self.storage.gas_used(),
            self.storage.spec(),
        );
  }
```

crates/precompiles/src/tip20/dispatch.rs
```diff
@@ -158,6 +161,11 @@ impl<'a, S: PrecompileStorageProvider> Precompile for TIP20Token<'a, S> {
                     self.burn_blocked(s, call)
                 })
             }
+            ITIP20::burnFromCall::SELECTOR => {
+                mutate_void::<ITIP20::burnFromCall>(calldata, msg_sender, |s, call| {
+                    self.burn_from(s, call)
+                })
+            }
```
0xKitsune, line 184: Same thing here, we should gate this behind `allegretto`.

crates/precompiles/src/tip20/mod.rs
```diff
@@ -2646,6 +2692,295 @@ pub(crate) mod tests {
...
+        };
+
+        // Verify target address is NOT in blacklist, so it's authorized
+        {
+            let mut registry = TIP403Registry::new(&mut storage);
+            assert!(registry.is_authorized(ITIP403Registry::isAuthorizedCall {
+                policyId: policy_id,
+                user: target,
+            })?);
+        }
```
0xKitsune: Why do we create the registry twice here? Can we just initialize once similar to below:


```rust
let mut registry = TIP403Registry::new(&mut storage);
let policy_id = registry.create_policy(
    admin,
    ITIP403Registry::createPolicyCall {
    admin,
    policyType: ITIP403Registry::PolicyType::BLACKLIST,
    },
)?;

assert!(registry.is_authorized(ITIP403Registry::isAuthorizedCall {
                policyId: policy_id,
                user: target,
            })?);

```

crates/precompiles/src/tip20/mod.rs
```diff
@@ -2646,6 +2692,295 @@ pub(crate) mod tests {
...
+                amount: U256::from(1000),
+            },
+        )?;
+
+        let initial_balance = token.balance_of(ITIP20::balanceOfCall { account: target })?;
+        let initial_supply = token.total_supply()?;
+        assert_eq!(initial_balance, U256::from(1000));
+        assert_eq!(initial_supply, U256::from(1000));
+
+        // burnFrom should work even though target is authorized (unlike burnBlocked)
```
0xKitsune: Can you clarify what you mean by the target is authorized? Is this indicating that they arent blacklisted?

crates/precompiles/src/tip20/mod.rs
```diff
@@ -2646,6 +2692,295 @@ pub(crate) mod tests {
...
+        let burn_amount = U256::from(300);
+        token.burn_from(
+            burner,
+            ITIP20::burnFromCall {
+                from: target,
+                amount: burn_amount,
+            },
+        )?;
```
0xKitsune, line 2751: Can we also update to expect event emission?

crates/precompiles/src/tip20/mod.rs
```diff
@@ -2646,6 +2692,295 @@ pub(crate) mod tests {
...
+
+        // Verify balances and total supply were updated
+        let final_balance = token.balance_of(ITIP20::balanceOfCall { account: target })?;
+        let final_supply = token.total_supply()?;
+        assert_eq!(final_balance, U256::from(700));
+        assert_eq!(final_supply, U256::from(700));
```
0xKitsune: nit: can we assert pre/post balances here instead of hardcoding?

crates/precompiles/src/tip20/mod.rs
```diff
@@ -2646,6 +2692,295 @@ pub(crate) mod tests {
...
+
+        Ok(())
+    }
+
+    #[test]
+    fn test_burn_from_successful_from_blocked_address() -> eyre::Result<()> {
```
0xKitsune: very minor nit: can we update to `from_blacklisted_address` to be explicit?

crates/precompiles/src/tip20/mod.rs
```diff
@@ -2646,6 +2692,295 @@ pub(crate) mod tests {
...
+
+        // burnFrom should work on blocked addresses too
+        let burn_amount = U256::from(300);
+        token.burn_from(
+            burner,
+            ITIP20::burnFromCall {
+                from: target,
+                amount: burn_amount,
+            },
+        )?;
```
0xKitsune, line 2847: Same thing here can we expect event emission?

crates/precompiles/src/tip20/mod.rs
```diff
@@ -2646,6 +2690,314 @@ pub(crate) mod tests {
         Ok(())
     }
 
+    #[test]
+    fn test_burn_from() -> eyre::Result<()> {
+        let mut storage = HashMapStorageProvider::new(1);
```
fgimenez: nit, even though it is not required for the test (it calls the function directly instead of using dispatch where the gating is defined) we could include here `.with_spec(TempoHardfork::Allegretto)` for consistency with the rest of tests related to the new functionality, wdyt?
```suggestion
        let mut storage = HashMapStorageProvider::new(1).with_spec(TempoHardfork::Allegretto);
```

crates/precompiles/src/tip20/mod.rs
```diff
@@ -2646,6 +2690,314 @@ pub(crate) mod tests {
...
+            })
+            .into_log_data()
+        );
+
+        Ok(())
+    }
+
+    #[test]
+    fn test_burn_from_blacklisted_address() -> eyre::Result<()> {
+        let mut storage = HashMapStorageProvider::new(1);
```
fgimenez: same here about adding `.with_spec(TempoHardfork::Allegretto)`
```suggestion
        let mut storage = HashMapStorageProvider::new(1).with_spec(TempoHardfork::Allegretto);
```

crates/precompiles/src/tip20/mod.rs
```diff
@@ -2570,7 +2614,7 @@ pub(crate) mod tests {
     }
 
     #[test]
-    fn test_unable_to_burn_blocked_from_protected_address() -> eyre::Result<()> {
+    fn test_burn_from_protected_address() -> eyre::Result<()> {
```
fgimenez: why changing the test name here? the previous one is more accurate, it tests that it is not possible to use `burn_blocked` from a protected address

fgimenez (approved):
lgm, some nits

crates/precompiles/src/tip20/mod.rs
0xKitsune: Fixed in [c8eff0d](https://github.com/tempoxyz/tempo/pull/1140/commits/c8eff0d8081a42311cf8491eee0c6d96e3d9b7b9)

PR #1147, opened 2025-11-30
chore: more payload builder defaults

klkvr:
Right now we have those configured manually for all our nodes https://github.com/tempoxyz/helm-charts/blob/53a53f7f2c7b789cc22e9704532d770a3a95cf75/charts/tempo-node/templates/statefulset.yaml#L163-L164

This PR would allow to remove that

### DIFF
--- a/crates/alloy/src/rpc/compat.rs
+++ b/crates/alloy/src/rpc/compat.rs
@@ -10,7 +10,10 @@ use reth_rpc_convert::{
 };
 use reth_rpc_eth_types::EthApiError;
 use tempo_evm::TempoBlockEnv;
-use tempo_primitives::{AASignature, SignatureType, TempoHeader, TempoTxEnvelope, TempoTxType};
+use tempo_primitives::{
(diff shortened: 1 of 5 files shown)


bin/tempo/src/defaults.rs
```diff
@@ -21,6 +21,8 @@ fn init_download_urls() {
 fn init_payload_builder_defaults() {
     DefaultPayloadBuilderValues::default()
         .with_interval(Duration::from_millis(100))
+        .with_max_payload_tasks(16)
```
SuperFluffy, line 24: What will the node return if one exceeds this? Or will it just block on await?

klkvr: yeah it will block, that's essentially an input for `Semaphore` aquired before each payload job

https://github.com/paradigmxyz/reth/blob/b3c00ed602f2a8805974f25be714a3bea26901e9/crates/payload/basic/src/lib.rs#L234-L237
https://github.com/paradigmxyz/reth/blob/b3c00ed602f2a8805974f25be714a3bea26901e9/crates/payload/basic/src/lib.rs#L346

PR #1153, opened 2025-12-01
chore:  use `scoped-tls` crate

0xrusowsky:
## Motivation

use the battle tested `scoped-tls` crate

## Solution

- replace manual tls impl with `scoped-tls`'s macro
- drop `StorageGuard` in favor of a closure-based approach
- drop `trait PrecompileStorageContext` and move `fn enter()` into `StorageAccessor` (which has been renamed to `StorageContext` for clarity now that the trait doesn't exist)

### DIFF
--- a/crates/precompiles/src/storage/thread_local.rs
+++ b/crates/precompiles/src/storage/thread_local.rs
@@ -70,7 +70,9 @@ impl StorageContext {
         }
         STORAGE.with(|cell| {
             // SAFETY: `scoped_tls` ensures the pointer is only accessible within the closure scope.
-            f(unsafe { ptr_as_mut(*cell.borrow_mut()) })
+            // Holding the guard prevents re-entrant borrows.
+            let guard = cell.borrow_mut();
+            f(unsafe { ptr_as_mut(*guard) })
         })
     }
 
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/storage/thread_local.rs
```diff
@@ -1,173 +1,151 @@
...
-        // while a guard is active, the reference remains valid.
-        Self::with_storage_call(|s| {
+        // `StorageContext::enter` closure. Since `StorageContext` can only be used
+        // while inside enter(), the reference remains valid.
+        Self::with_storage(|s| {
             let info = s.get_account_info(address)?;
-            // Extend the lifetime to match &'_ self
-            // This is safe because the underlying storage outlives the accessor
+            // SAFETY: Extend the lifetime to match `&'_ self`.
+            // Underlying storage data outlives the accessor.
```
DaniPopes, line 107: comment should mention that this is the only type that exposes access to the `STORAGE` static.

crates/precompiles/src/storage/thread_local.rs
```diff
@@ -1,173 +1,151 @@
...
+impl StorageContext {
+    /// Enter storage context. All storage operations must happen within the closure.
+    ///
+    /// # IMPORTANT
+    ///
+    /// The caller must ensure that:
+    /// 1. Only one `enter` call is active at a time, in the same thread.
+    /// 2. If multiple storage providers are instantiated in parallel threads,
+    ///    they CANNOT point to the same storage addresses.
+    pub fn enter<S, R>(storage: &mut S, f: impl FnOnce() -> R) -> R
```
DaniPopes, line 49: This is mostly fine, but i would like to use `RefCell`, because technically you can create nested `StorageContext`s with different `S` but same `STORAGE`. This can cause UB if used at the same time since the borrow rule cannot be guaranteed at compile time

PR #1156, opened 2025-12-01
fix: add registry check on mint

howydev:
fix: we should do a registry check on mint

### DIFF
--- a/crates/alloy/src/network.rs
+++ b/crates/alloy/src/network.rs
@@ -11,11 +11,16 @@ use alloy_primitives::{Address, Bytes, ChainId, TxKind, U256};
 use alloy_provider::fillers::{
     ChainIdFiller, GasFiller, JoinFill, NonceFiller, RecommendedFillers,
 };
-use alloy_rpc_types_eth::AccessList;
+use alloy_rpc_types_eth::{AccessList, Block, Transaction};
 use alloy_signer_local::PrivateKeySigner;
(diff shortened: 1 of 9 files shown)


crates/precompiles/src/tip20/mod.rs
```diff
@@ -422,6 +413,16 @@ impl<'a, S: PrecompileStorageProvider> TIP20Token<'a, S> {
...
 
+        // Check if the `to` address is authorized to receive tokens
+        let transfer_policy_id = self.transfer_policy_id()?;
+        let mut registry = TIP403Registry::new(self.storage);
+        if !registry.is_authorized(ITIP403Registry::isAuthorizedCall {
+            policyId: transfer_policy_id,
+            user: to,
+        })? {
+            return Err(TIP20Error::policy_forbids().into());
+        }
```
mattsse, line 417: should this happen sooner perhaps?

crates/precompiles/src/tip20/mod.rs
```diff
@@ -422,6 +413,16 @@ impl<'a, S: PrecompileStorageProvider> TIP20Token<'a, S> {
             return Err(TIP20Error::supply_cap_exceeded().into());
         }
 
+        // Check if the `to` address is authorized to receive tokens
```
mattsse, line 407: I also think we need a hardfork gate for this?

PR #1157, opened 2025-12-01
feat!(tip20): add burnFrom to `TIP-20`

0xKitsune:
Ref #1140

PR to reintroduce `burnFrom` for TIP20 tokens. This was initially reverted to align on the implications of this change before landing in `main`.


crates/precompiles/src/tip20/mod.rs
```diff
@@ -99,6 +99,7 @@ pub static PAUSE_ROLE: LazyLock<B256> = LazyLock::new(|| keccak256(b"PAUSE_ROLE"
 pub static UNPAUSE_ROLE: LazyLock<B256> = LazyLock::new(|| keccak256(b"UNPAUSE_ROLE"));
 pub static ISSUER_ROLE: LazyLock<B256> = LazyLock::new(|| keccak256(b"ISSUER_ROLE"));
 pub static BURN_BLOCKED_ROLE: LazyLock<B256> = LazyLock::new(|| keccak256(b"BURN_BLOCKED_ROLE"));
+pub static BURN_FROM_ROLE: LazyLock<B256> = LazyLock::new(|| keccak256(b"BURN_FROM_ROLE"));
```
mattsse, line 102: I think we could techincally replace all of these with constants but the lazylock deref isnt too bad

crates/precompiles/src/tip20/mod.rs
```diff
@@ -518,6 +527,41 @@ impl<'a, S: PrecompileStorageProvider> TIP20Token<'a, S> {
...
+    pub fn burn_from(&mut self, msg_sender: Address, call: ITIP20::burnFromCall) -> Result<()> {
+        self.check_role(msg_sender, *BURN_FROM_ROLE)?;
+
+        // Prevent burning from `FeeManager` and `StablecoinExchange` to protect accounting invariants
+        if matches!(
+            call.from,
+            TIP_FEE_MANAGER_ADDRESS | STABLECOIN_EXCHANGE_ADDRESS
+        ) {
+            return Err(TIP20Error::protected_address().into());
+        }
```
mattsse, line 540: should this happen first? before the role check?

crates/precompiles/src/tip20/mod.rs
```diff
@@ -518,6 +527,41 @@ impl<'a, S: PrecompileStorageProvider> TIP20Token<'a, S> {
...
+        let total_supply = self.total_supply()?;
+        let new_supply =
+            total_supply
+                .checked_sub(call.amount)
+                .ok_or(TIP20Error::insufficient_balance(
+                    total_supply,
+                    call.amount,
+                    self.address,
+                ))?;
+        self.set_total_supply(new_supply)?;
```
mattsse, line 553: we actually have this twice already

https://github.com/tempoxyz/tempo/blob/91efc9f69321c3d802194745f721c9f264942085/crates/precompiles/src/tip20/mod.rs#L526-L535

https://github.com/tempoxyz/tempo/blob/91efc9f69321c3d802194745f721c9f264942085/crates/precompiles/src/tip20/mod.rs#L500-L509

maybe worth unifying in a followup?

mattsse (approved):
this is simple enough

perhaps @fgimenez or @klkvr sanity check

jenpaff:
deprioritised and will be picked up in another fork

PR #1158, opened 2025-12-01
feat(alloy): configure random nonces on provider builder

### LINKED ISSUE #1152: Default nonce filler
I guess we could wrap an default nonce filler and make the new one delegate to it by default

and then we could maybe have a flag on our custom filler that we can flip through an ext trait to enable random nonces

_Originally posted by @klkvr in https://github.com/tempoxyz/tempo/pull/1138#discussion_r2576992567_

shekhirin:
Closes #1152

### DIFF
--- a/crates/alloy/src/fillers/mod.rs
+++ b/crates/alloy/src/fillers/mod.rs
@@ -1,4 +1,4 @@
 //! Transaction fillers for Tempo network.
 
 mod nonce;
-pub use nonce::{Random2DNonceFiller};
+pub use nonce::Random2DNonceFiller;


crates/alloy/src/fillers/nonce.rs
```diff
@@ -58,12 +58,47 @@ impl<N: Network<TransactionRequest = TempoTransactionRequest>> TxFiller<N> for R
     }
 }
 
+pub trait Random2DNoncesProviderExt {
```
mattsse: can we move this and rename TempoProviderBuilderExt then we can use this for additional things later

PR #1163, opened 2025-12-01
fix(alloy): 2D nonce filler fill condition

shekhirin:
Alloy's filler API doesn't guarantee that `fill_sync` for a filler will not be called if the `status` has returned `Finished`. Because of that, we need to check whether we need to fill in two places.

### DIFF
--- a/crates/precompiles/src/account_keychain/dispatch.rs
+++ b/crates/precompiles/src/account_keychain/dispatch.rs
@@ -1,6 +1,7 @@
 use super::{AccountKeychain, IAccountKeychain};
 use crate::{
-    Precompile, input_cost, mutate_void, storage::PrecompileStorageProvider, unknown_selector, view,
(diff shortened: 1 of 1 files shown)


crates/alloy/src/fillers/nonce.rs
```diff
@@ -14,18 +14,28 @@ use tempo_primitives::subblock::has_sub_block_nonce_key_prefix;
...
             return FillerControlFlow::Finished;
         }
         FillerControlFlow::Ready
     }
 
     fn fill_sync(&self, tx: &mut SendableTx<N>) {
-        if let Some(builder) = tx.as_mut_builder() {
+        if let Some(builder) = tx
+            .as_mut_builder()
+            .filter(|builder| Self::is_filled(builder))
```
onbjerg: supposed to be `!Self::is_filled`?

PR #1164, opened 2025-12-01
chore: add test for subblock transactions failing to pay fees

klkvr:


crates/e2e/src/tests/subblocks.rs
```diff
@@ -199,13 +209,185 @@ fn subblocks_are_included_post_allegretto() {
...
+
+            // Assert that all expected transactions are included in the block.
+            for tx in expected_transactions.drain(..) {
+                if !block
+                    .sealed_block()
+                    .body()
+                    .transactions
+                    .iter()
+                    .any(|t| t.tx_hash() == *tx)
+                {
```
mattsse, line 278: there's

https://github.com/paradigmxyz/reth/blob/a62b46a96070df0c6925238439b3c4ed16df5eaf/crates/primitives-traits/src/block/body.rs#L73-L78

maybe we need to add this to sealedblock type as well?

klkvr: yeah that'd be nice

PR #1168, opened 2025-12-01
fix(genesis): create PathUSD through factory with address(0) as quote token

fgimenez:
ref https://tempoxyz.slack.com/archives/C09KCGR4LQ4/p1764244109204229

Updates the genesis generator to create PathUSD through the TIP20 factory post-Allegretto instead of initializing it directly.

  - Configure EVM with Allegretto hardfork so factory starts token counter at 0
  - Create PathUSD as first token (token_id=0) via factory with `address(0)` as quote token
  - Initialize factory once before any token creation
  - Remove direct `PathUSD::initialize()` call that bypassed factory

### DIFF
--- a/xtask/src/genesis_args.rs
+++ b/xtask/src/genesis_args.rs
@@ -473,6 +473,12 @@ fn create_and_mint_token(
 
     let token_id = {
         let mut factory = TIP20Factory::new(&mut provider);
+        assert!(
+            factory
+                .is_initialized()
+                .expect("Could not check factory initialization"),
+            "TIP20Factory must be initialized before creating tokens"
+        );
         let token_address = factory
             .create_token(
                 admin,
--- a/crates/precompiles/src/tip20_factory/mod.rs
+++ b/crates/precompiles/src/tip20_factory/mod.rs
@@ -42,6 +42,12 @@ impl<'a, S: PrecompileStorageProvider> TIP20Factory<'a, S> {
         )
     }
 
+    /// Returns true if the factory has been initialized (has code set).
+    pub fn is_initialized(&mut self) -> Result<bool> {
+        let info = self.storage.get_account_info(TIP20_FACTORY_ADDRESS)?;
+        Ok(info.code.is_some())
+    }
+
     pub fn create_token(
         &mut self,
         sender: Address,


xtask/src/genesis_args.rs
```diff
@@ -377,18 +381,96 @@ impl GenesisArgs {
...
+    token.grant_role_internal(admin, *ISSUER_ROLE)?;
+
+    token
+        .set_supply_cap(
+            admin,
+            ITIP20::setSupplyCapCall {
+                newSupplyCap: U256::from(u128::MAX),
+            },
+        )
+        .expect("Could not set PathUSD supply cap");
```
0xKitsune: nit: we do not need to call this function as the supply cap is set to `u128::MAX` when the token is deployed.

xtask/src/genesis_args.rs
```diff
@@ -377,18 +381,96 @@ impl GenesisArgs {
...
+        token
+            .mint(
+                admin,
+                ITIP20::mintCall {
+                    to: *recipient,
+                    amount: U256::from(u64::MAX),
+                },
+            )
+            .expect("Could not mint pathUSD");
+    }
```
0xKitsune, line 453: Just noting that this is fine for testnet but eventually for mainnet we will have to skip this step. Maybe it makes sense to add a `--testnet` flag or similar. No action needed now, just a note for the future.

xtask/src/genesis_args.rs
```diff
@@ -377,18 +381,96 @@ impl GenesisArgs {
...
+
+    Ok(())
+}
+
+/// Creates a TIP20 token through the factory (factory must already be initialized)
```
0xKitsune, line 458: nit: if the factory must be initialized, maybe we can add an assertion in this function to check this. This avoids any potential footguns of calling this function without the factory initialized. Alternatively we could pass in a reference to the factory instance.

xtask/src/genesis_args.rs
fgimenez: good call, removed

xtask/src/genesis_args.rs
fgimenez: makes sense, added `TIP20Factory::is_initialized` that checks if the factory has code set, and asserting on that here

PR #1173, opened 2025-12-02
fix(consensus): fill holes when forwarding finalized blocks

SuperFluffy:
This patch has the application backfill holes on the execution layer if it attempts to finalize a block (received from the consensus layer) that is not immediately on top of the highest finalized height of the execution layer.

This patch is necessary because `marshal::Mailbox::get_info(Identifier::Latest)` was misunderstood: it did not return the latest finalized block that the marshal actor has forwarded and was acknowledged by the application, but the finalized tip of the overall network.

This led to crash-loops of the node because it was attempting to backfill blocks that the marshal actor did not yet have available (and that should be forwarded by different means).


crates/commonware-node/src/consensus/application/executor.rs
```diff
@@ -456,6 +325,35 @@ where
...
+                )
+            })
+            && execution_height + 1 < block.height()
+        {
+            info!(
+                execution.finalized_height = execution_height,
+                "hole detected; consensus attempts to finalize block with gaps \
+                on the execution layer; filling them in first",
+            );
+            let _ = self.fill_holes(execution_height + 1, block.height()).await;
```
SuperFluffy, line 348: So this is going to potentially stall the execution actor until this hole-plugging is done.

But a) it should be fast (all blocks should be available in marshal), and b) it's no use doing any of the other execution anway.

crates/commonware-node/src/consensus/application/executor.rs
```diff
@@ -456,6 +325,35 @@ where
             );
         }
 
+        if let Ok(execution_height) = self
+            .execution_node
+            .provider
+            .finalized_block_number()
```
SuperFluffy: This seems to be the most reasonable method to get the finalized block number. I understand it is only populated if the execution layer actually gets an FCU for a known block?

crates/commonware-node/src/consensus/application/executor.rs
```diff
@@ -456,6 +325,35 @@ where
...
+            .inspect_err(|error| {
+                warn!(
+                    %error,
+                    "failed getting last finalized block from execution layer, will \
+                    finalize forward block to execution layer without extra checks, \
+                    but it might fail"
+                )
+            })
+            && execution_height + 1 < block.height()
+        {
```
klkvr, line 342: this will return `None` if EL did not yet receive a `finalized` hash that it has a block for https://github.com/paradigmxyz/reth/blob/b3c00ed602f2a8805974f25be714a3bea26901e9/crates/engine/tree/src/tree/mod.rs#L2612-L2626

not sure what's the expected flow on startup but if the sequence is
1. both reth and commonware are at height 3
2. reth is restarted and is now at height 1
3. `forward_finalized` is called with height 4, causing FCU `{finalized: 4, head: 4}`

this would be `None`

SuperFluffy: Good point - I reverted to using `BlockNumreader::last_block_number` - since are only concerned with the case `finalized.height < execution.height`, we don't need to worry about head blocks.

PR #1177, opened 2025-12-02
feat(precompiles)!: migrate to new API

### LINKED ISSUE #860: feat(precompiles): support individual struct field access

0xrusowsky:
## Motivation 

Improve the devex by introducing an API that feels like native Rust and, at the same time, closer to the usual Solidity patterns.

This PR closes #860 and achieves the final goal of #856 

## Solution

The type system changes of this PR have been proposed in its own PR #986 and have already been extensively reviewed. The feedback from the several review rounds has been incorporated and re-reviewed.

However, PR #986 can't be directly merged into main because it uses a new storage access pattern, and it also changes the precompiles API. These breaking changes made it impossible to merge the storage type system changes without a full migration.

### Reviewing this PR

If you haven't been involved in the storage type system PR review, you are more than welcome (even encouraged) to have a look at the files under `crates/precompiles/src/storage/` or at the #986 PR (which is around 8k diff, although most of it are tests).

With that being said, the main focus of this PR review should be the migration of the precompiles to the new API, as well as their integration within the rest of the crates in the monorepo.

Out of the ~10k diff (excluding the storage type system changes), most of the diff are because of tests.

Please review super carefully the implementation changes, as they should preserve the logic untouched.

### Using the new storage API

unlike previously, where we had to hold a mutable reference to the storage provider (type implementing `trait PrecompileStorageProvider`), we now use the [scoped_tls](https://docs.rs/scoped-tls/latest/scoped_tls/) crate to generate a scoped thread-local variable. In practice, this means that our storage interactions will be done using the zero-sized `StorageContext` need to use the following pattern:
```rs
let mut provider = EvmPrecompileStorageProvider::new();

// we must always "enter" to setup the storage provider as a thread-local variable
StorageContext::enter(&mut provider, || {
    // inside the closure, we can access all the `trait PrecompileStorageProvider` fns via `StorageContext`
    // which performs them atomically. Thus, precompiles don't need to hold storage refs, and we won't have
    // mutable ref borrow conflicts anymore.

    let current_time = StorageContext.timestamp();
    let spec = StorageContext.sepc();

    let path_usd = PathUSD::new();
    path_usd.grant_role_internal(admin, *ISSUER_ROLE)?;
        
    let token = TIP20Token::new(1);
    token.grant_role_internal(admin, *ISSUER_ROLE)?;
});
```

### Using the new precompiles API

precompiles now use handlers for interacting with storage, rather than low-level helper functions like before. In pratice, this means that we can consume them with a solidity-like devex:

```rs
#[contract]
pub struct TIP20Token {
    // ...
    total_supply: U256,
    balances: Mapping<Address, U256>,
    allowances: Mapping<Address, Mapping<Address, U256>>,
    // ...
}

impl TIP20Token {
    fn get_allowance(&self, owner: Address, spender: Address) -> Result<U256> {
        self.allowances.at(owner).at(spender).read()
    }

    fn set_allowance(&mut self, owner: Address, spender: Address, amount: U256) -> Result<()> {
        self.allowances.at(owner).at(spender).write(amount)
    }
}
```

handlers have powerful APIs provide us better qol:
```rs
 // Base handler for all storage operations. Other handlers delegate to `Slot`.
 struct Slot<T> {
     pub fn read(&self) -> Result<T>
     pub fn write(&mut self, value: T) -> Result<()>
     pub fn delete(&mut self) -> Result<()>
 }
```
```rs
 // Returns a handler for the given key.
 // For nested mappings (Mapping<K1, Mapping<K2, V>>), returns another Mapping.
 struct Mapping<K, V> {
     pub fn at(&self, key: K) -> <V as StorableType>::Handler
 }
```
```rs
 // Generated by `#[derive(Storable)]`, its fields are handlers.
 // This means that we can load each field individually
 struct MyStructHandler {
     pub foo: Slot<Address>,
     pub bar: Mapping<u64, U256>,
 }
 impl Handler<MyStruct> for MyStructHandler {
     pub fn read(&self) -> Result<MyStruct>
     pub fn write(&mut self, value: MyStruct) -> Result<()>
     pub fn delete(&mut self) -> Result<()>
 }
```
```rs
 // Fixed-size array. Element access is checked (returns an option).
 struct ArrayHandler<T, N> {
     pub fn read(&self) -> Result<[T; N]>
     pub fn write(&mut self, value: [T; N]) -> Result<()>
     pub fn delete(&mut self) -> Result<()>
     pub fn at(&self, index: usize) -> Option<<T as StorableType>::Handler>
     pub const fn len(&self) -> usize
 }
```
```rs
 // Dynamic array with stack-like ops. Supports multi-slot element types.
 struct VecHandler<T> {
     pub fn read(&self) -> Result<Vec<T>>
     pub fn write(&mut self, value: Vec<T>) -> Result<()>
     pub fn delete(&mut self) -> Result<()>
     pub fn at(&self, index: usize) -> <T as StorableType>::Handler
     pub fn len(&self) -> Result<usize>
     pub fn pop(&self) -> Result<Option<T>>
 }
```

## TODO

- [x] migrate precompiles to the new storage API
- [X] update other crates to the new storage API
- [x] ensure all tests pass
- [ ] ensure gas snapshot tests pass

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,10 +3,25 @@
 version = 4
 
 [[package]]
+name = "addr2line"
+version = "0.25.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1b5d307320b3181d6d7954e663bd7c774a838b8220fe0593c86d9fb09f498b4b"
+dependencies = [
+ "gimli",
+]
+
+[[package]]
 name = "adler2"
 version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
+
+[[package]]
+name = "adler32"
+version = "1.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "aae1277d39aeec15cb388266ecc24b11c80469deae6067e17a1a7aa9e5c1f234"
 
 [[package]]
 name = "aead"
@@ -123,9 +138,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-chains"
-version = "0.2.22"
+version = "0.2.23"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ff8c665521d11efbb11d5e5c5d63971426bb63df00d24545baf97e7f3dc91c0c"
+checksum = "35d744058a9daa51a8cf22a3009607498fcf82d3cf4c5444dd8056cdf651f471"
 dependencies = [
  "alloy-primitives",
  "alloy-rlp",
@@ -277,6 +292,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "alloy-eip7928"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "926b2c0d34e641cf8b17bf54ce50fda16715b9f68ad878fa6128bae410c6f890"
+dependencies = [
+ "alloy-primitives",
+ "alloy-rlp",
+ "borsh",
+ "serde",
+]
+
+[[package]]
 name = "alloy-eips"
 version = "1.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -304,9 +331,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-evm"
-version = "0.24.2"
+version = "0.25.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "01be36ba6f5e6e62563b369e03ca529eac46aea50677f84655084b4750816574"
+checksum = "3e7b4fb2418490bca9978e74208215ed5fcb21a10aba7eea487abaa60fd588db"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
@@ -1009,7 +1036,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -1020,7 +1047,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -1366,9 +1393,9 @@ checksum = "155a5a185e42c6b77ac7b88a15143d930a9e9727a5b7b77eed417404ab15c247"
 
 [[package]]
 name = "async-compression"
-version = "0.4.34"
+version = "0.4.36"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0e86f6d3dc9dc4352edeea6b8e499e13e3f5dc3b964d7ca5fd411415a3498473"
+checksum = "98ec5f6c2f8bc326c994cb9e241cc257ddaba9afa8555a43cffbb5dd86efaa37"
 dependencies = [
  "compression-codecs",
  "compression-core",
@@ -1534,6 +1561,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "backtrace"
+version = "0.3.76"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bb531853791a215d7c62a30daf0dde835f381ab5de4589cfe7c649d2cbe92bd6"
+dependencies = [
+ "addr2line",
+ "cfg-if",
+ "libc",
+ "miniz_oxide",
+ "object",
+ "rustc-demangle",
+ "windows-link 0.2.1",
+]
+
+[[package]]
 name = "base-x"
 version = "0.2.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1569,9 +1611,9 @@ checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
 
 [[package]]
 name = "base64ct"
-version = "1.8.0"
+version = "1.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "55248b47b0caf0546f7988906588779981c43bb1bc9d0c44087278f80cdb44ba"
+checksum = "0e050f626429857a27ddccb31e0aca21356bfa709c04041aefddac081a8f068a"
 
 [[package]]
 name = "bech32"
@@ -2595,9 +2637,9 @@ dependencies = [
 
 [[package]]
 name = "compression-codecs"
-version = "0.4.33"
+version = "0.4.35"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "302266479cb963552d11bd042013a58ef1adc56768016c8b82b4199488f2d4ad"
+checksum = "b0f7ac3e5b97fdce45e8922fb05cae2c37f7bbd63d30dd94821dacfd8f3f2bf2"
 dependencies = [
  "brotli",
  "compression-core",
@@ -2743,6 +2785,15 @@ name = "cow-utils"
 version = "0.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "417bef24afe1460300965a25ff4a24b8b45ad011948302ec221e8a0a81eb2c79"
+
+[[package]]
+name = "cpp_demangle"
+version = "0.4.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f2bb79cb74d735044c972aae58ed0aaa9a837e85b01106a54c39e42e97f62253"
+dependencies = [
+ "cfg-if",
+]
 
 [[package]]
 name = "cpufeatures"
@@ -3098,6 +3149,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f578e8e2c440e7297e008bb5486a3a8a194775224bbc23729b0dbdfaeebf162e"
 
 [[package]]
+name = "debugid"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bef552e6f588e446098f6ba40d89ac146c8c7b64aade83c051ee00bb5d2bc18d"
+dependencies = [
+ "uuid",
+]
+
+[[package]]
 name = "delay_map"
 version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3277,7 +3337,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users 0.5.2",
- "windows-sys 0.59.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -3575,7 +3635,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -3742,6 +3802,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "findshlibs"
+version = "0.10.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "40b9e59cd0f7e0806cca4be089683ecb6434e602038df21fe6bf6711b2f07f64"
+dependencies = [
+ "cc",
+ "lazy_static",
+ "libc",
+ "winapi",
+]
+
+[[package]]
 name = "fixed-hash"
 version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3761,9 +3833,9 @@ checksum = "1d674e81391d1e1ab681a28d99df07927c6d4aa5b027d7da16ba32d1d21ecd99"
 
 [[package]]
 name = "flate2"
-version = "1.1.7"
+version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a2152dbcb980c05735e2a651d96011320a949eb31a0c8b38b72645ce97dec676"
+checksum = "bfe33edd8e85a12a67454e37f8c75e730830d83e313556ab9ebf9ee7fbeb3bfb"
 dependencies = [
  "crc32fast",
  "miniz_oxide",
@@ -4030,10 +4102,16 @@ dependencies = [
 ]
 
 [[package]]
-name = "git2"
-version = "0.20.2"
+name = "gimli"
+version = "0.32.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2deb07a133b1520dc1a5690e9bd08950108873d7ed5de38dcc74d3b5ebffa110"
+checksum = "e629b9b98ef3dd8afe6ca2bd0f89306cec16d43d907889945bc5d6687f2f13c7"
+
+[[package]]
+name = "git2"
+version = "0.20.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3e2b37e2f62729cdada11f0e6b3b6fe383c69c29fc619e391223e12856af308c"
 dependencies = [
  "bitflags 2.10.0",
  "libc",
@@ -5204,10 +5282,30 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "37c93d8daa9d8a012fd8ab92f088405fb202ea0b6ab73ee2482ae66af4f42091"
 
 [[package]]
-name = "libgit2-sys"
-version = "0.18.2+1.9.1"
+name = "libflate"
+version = "1.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1c42fe03df2bd3c53a3a9c7317ad91d80c81cd1fb0caec8d7cc4cd2bfa10c222"
+checksum = "5ff4ae71b685bbad2f2f391fe74f6b7659a34871c08b210fdc039e43bee07d18"
+dependencies = [
+ "adler32",
+ "crc32fast",
+ "libflate_lz77",
+]
+
+[[package]]
+name = "libflate_lz77"
+version = "1.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a52d3a8bfc85f250440e4424db7d857e241a3aebbbe301f3eb606ab15c39acbf"
+dependencies = [
+ "rle-decode-fast",
+]
+
+[[package]]
+name = "libgit2-sys"
+version = "0.18.3+1.9.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c9b3acc4b91781bb0b3386669d325163746af5f6e4f73e6d2d630e09a35f3487"
 dependencies = [
  "cc",
  "libc",
@@ -5243,9 +5341,9 @@ dependencies = [
 
 [[package]]
 name = "libp2p-identity"
-version = "0.2.12"
+version = "0.2.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3104e13b51e4711ff5738caa1fb54467c8604c2e94d607e27745bcf709068774"
+checksum = "f0c7892c221730ba55f7196e98b0b8ba5e04b4155651736036628e9f73ed6fc3"
 dependencies = [
  "asn1_der",
  "bs58",
@@ -5735,6 +5833,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "names"
+version = "0.14.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7bddcd3bf5144b6392de80e04c347cd7fab2508f6df16a85fc496ecd5cec39bc"
+dependencies = [
+ "rand 0.8.5",
+]
+
+[[package]]
+name = "nix"
+version = "0.27.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2eb04e9c688eff1c89d72b407f168cf79bb9e867a9d3323ed6c01519eb9cc053"
+dependencies = [
+ "bitflags 2.10.0",
+ "cfg-if",
+ "libc",
+]
+
+[[package]]
 name = "nix"
 version = "0.30.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -5801,7 +5919,7 @@ version = "0.50.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7957b9740744892f114936ab4a57b3f487491bbeafaf8083688b16841a4240e5"
 dependencies = [
- "windows-sys 0.59.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -5961,6 +6079,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "object"
+version = "0.37.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ff76201f031d8863c38aa7f905eca4f53abbfa15f609db4277d44cd8938f33fe"
+dependencies = [
+ "memchr",
+]
+
+[[package]]
 name = "once_cell"
 version = "1.21.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -5984,9 +6111,9 @@ checksum = "d6790f58c7ff633d8771f42965289203411a5e5c68388703c06e14f24770b41e"
 
 [[package]]
 name = "op-alloy"
-version = "0.22.4"
+version = "0.23.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c3b13412d297c1f9341f678b763750b120a73ffe998fa54a94d6eda98449e7ca"
+checksum = "e9b8fee21003dd4f076563de9b9d26f8c97840157ef78593cd7f262c5ca99848"
 dependencies = [
  "op-alloy-consensus",
  "op-alloy-network",
@@ -5997,9 +6124,9 @@ dependencies = [
 
 [[package]]
 name = "op-alloy-consensus"
-version = "0.22.4"
+version = "0.23.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "726da827358a547be9f1e37c2a756b9e3729cb0350f43408164794b370cad8ae"
+checksum = "736381a95471d23e267263cfcee9e1d96d30b9754a94a2819148f83379de8a86"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
@@ -6017,9 +6144,9 @@ dependencies = [
 
 [[package]]
 name = "op-alloy-network"
-version = "0.22.4"
+version = "0.23.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f63f27e65be273ec8fcb0b6af0fd850b550979465ab93423705ceb3dfddbd2ab"
+checksum = "4034183dca6bff6632e7c24c92e75ff5f0eabb58144edb4d8241814851334d47"
 dependencies = [
  "alloy-consensus",
  "alloy-network",
@@ -6033,9 +6160,9 @@ dependencies = [
 
 [[package]]
 name = "op-alloy-provider"
-version = "0.22.4"
+version = "0.23.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a71456699aa256dc20119736422ad9a44da8b9585036117afb936778122093b9"
+checksum = "6753d90efbaa8ea8bcb89c1737408ca85fa60d7adb875049d3f382c063666f86"
 dependencies = [
  "alloy-network",
  "alloy-primitives",
@@ -6048,9 +6175,9 @@ dependencies = [
 
 [[package]]
 name = "op-alloy-rpc-types"
-version = "0.22.4"
+version = "0.23.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "562dd4462562c41f9fdc4d860858c40e14a25df7f983ae82047f15f08fce4d19"
+checksum = "ddd87c6b9e5b6eee8d6b76f41b04368dca0e9f38d83338e5b00e730c282098a4"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
@@ -6067,9 +6194,9 @@ dependencies = [
 
 [[package]]
 name = "op-alloy-rpc-types-engine"
-version = "0.22.4"
+version = "0.23.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d8f24b8cb66e4b33e6c9e508bf46b8ecafc92eadd0b93fedd306c0accb477657"
+checksum = "77727699310a18cdeed32da3928c709e2704043b6584ed416397d5da65694efc"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
@@ -6082,6 +6209,7 @@ dependencies = [
  "ethereum_ssz_derive",
  "op-alloy-consensus",
  "serde",
+ "sha2 0.10.9",
  "snap",
  "thiserror 2.0.17",
 ]
@@ -6147,7 +6275,7 @@ dependencies = [
  "opentelemetry-http",
  "opentelemetry-proto",
  "opentelemetry_sdk",
- "prost",
+ "prost 0.14.1",
  "reqwest",
  "thiserror 2.0.17",
  "tokio",
@@ -6163,7 +6291,7 @@ checksum = "a7175df06de5eaee9909d4805a3d07e28bb752c34cab57fa9cff549da596b30f"
 dependencies = [
  "opentelemetry",
  "opentelemetry_sdk",
- "prost",
+ "prost 0.14.1",
  "tonic",
  "tonic-prost",
 ]
@@ -6482,7 +6610,7 @@ dependencies = [
  "hyper",
  "hyper-util",
  "mime",
- "nix",
+ "nix 0.30.1",
  "parking_lot",
  "percent-encoding",
  "pin-project-lite",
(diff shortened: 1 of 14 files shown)


crates/node/tests/it/eth_call.rs
```diff
@@ -127,7 +125,10 @@ async fn test_eth_trace_call() -> eyre::Result<()> {
 
     let token_storage_diff = token_diff.storage.clone();
     // Assert sender token balance has changed
-    let slot = mapping_slot(caller, tip20::slots::BALANCES);
+    let slot = TIP20Token::from_address(*token.address())
+        .balances
+        .at(caller)
+        .slot();
```
klkvr: not sure if I like this pattern. to me `mapping_slot` feels simpler than initializing the entire contract just to compute a hash

crates/revm/src/handler.rs
```diff
@@ -685,70 +692,71 @@ where
             let internals = EvmInternals::new(journal, block);
             let mut storage_provider = EvmPrecompileStorageProvider::new_max_gas(internals, cfg);
 
-            let mut keychain = AccountKeychain::new(&mut storage_provider);
-
-            let access_key_addr = key_auth.key_id;
+            StorageContext::enter(&mut storage_provider, || {
```
klkvr: some ideas about how to improve this API

```rust
StorageContext::evm(&mut internals, || { ... })

AccountKeychain::new_with_storage(&mut storage, |keychain| { ...  })
```

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -41,13 +37,44 @@ pub struct AuthorizedKey {
     pub is_revoked: bool,
 }
 
+// TODO(rusowsky): remove this and create a read-only wrapper that is callable from read-only ctx with db access
 impl AuthorizedKey {
     /// Decode AuthorizedKey from a storage slot value
     ///
     /// This is useful for read-only contexts (like pool validation) that don't have
     /// access to PrecompileStorageProvider but need to decode the packed struct.
     pub fn decode_from_slot(slot_value: U256) -> Self {
```
klkvr, line 46: I guess this could be achieved with a `PackedSlot(U256)`-like hack?

crates/precompiles/src/nonce/mod.rs
```diff
@@ -34,38 +32,40 @@ pub struct NonceManager {
...
 }
 
-impl<'a, S: PrecompileStorageProvider> NonceManager<'a, S> {
-    pub fn new(storage: &'a mut S) -> Self {
-        Self::_new(NONCE_PRECOMPILE_ADDRESS, storage)
+impl Default for NonceManager {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
klkvr: imo we should either make macro produce it or don't impl at all

klkvr: or was it clippy's idea?

crates/precompiles/src/nonce/mod.rs
```diff
@@ -34,38 +32,40 @@ pub struct NonceManager {
...
     }
 
     /// Initializes the nonce manager contract.
     pub fn initialize(&mut self) -> Result<()> {
-        // must ensure the account is not empty, by setting some code
-        self.storage.set_code(
-            NONCE_PRECOMPILE_ADDRESS,
-            Bytecode::new_legacy(Bytes::from_static(&[0xef])),
-        )
+        self.__initialize()
```
klkvr, line 38: now that I think about it feels like`0xef` logic should not be hidden inside `#[contract]` macro

we were recently discussing changing how precompiles are initialized #1254 and I guess it'd be trickier if this is a proc-macro generated code that needs to be changed

crates/precompiles/src/stablecoin_exchange/dispatch.rs
```diff
@@ -212,506 +210,513 @@ mod tests {
...
-        .unwrap();
-
-        let base_token = base.address();
-        let quote_token = quote.token.address();
+        let quote = TIP20Setup::path_usd(admin)
+            .with_issuer(admin)
+            .with_role(user, *TRANSFER_ROLE)
+            .with_mint(user, U256::from(amount))
+            .with_approval(user, exchange.address, U256::from(amount))
+            .apply()?;
```
klkvr, line 228: oh that's cool

crates/precompiles/src/lib.rs
```diff
@@ -57,8 +58,8 @@ pub use tempo_contracts::precompiles::{
     TIP403_REGISTRY_ADDRESS, VALIDATOR_CONFIG_ADDRESS,
 };
 
-// Re-export storage layout helpers for read-only contexts (e.g., pool validation)
-pub use account_keychain::{AuthorizedKey, compute_keys_slot};
+//Re-export storage layout helpers for read-only contexts (e.g., pool validation)
```
klkvr:
```suggestion
// Re-export storage layout helpers for read-only contexts (e.g., pool validation)
```

crates/precompiles/src/nonce/mod.rs
0xrusowsky: yeah this was because of clippy.

i guess a better solution is to add try avoid the warning by adding a clippy macro in the contract macro expansion

crates/node/tests/it/eth_call.rs
0xrusowsky: mmm ok, i did it for 2 reasons:
1. minimize human errors 
2. have the setup for the follow-up PR where I plan to impl a read-only wrapper of the storage provider for all DB types, so that we can read from storage using precompiles directly. So when that's ready it would only require changing `.slot()` for `.read()`

fine with reverting the changes if u think it's best though

0xrusowsky: also, `fn mapping_slot()` doesn't exist anymore, as now it's a fn impled by `triat StorageKey`.

so i'd use it as a method call from `Address` in this case. Is that fine? or should i re-impl the original fn?

crates/precompiles/src/nonce/mod.rs
0xrusowsky: sure, i just did it as a qol improvement, as it is always the same pattern. We still need to do the `self.__initialize()` call, which i thaught was fine.

i can either rename it or totally drop it and go back to manual impls, whatever u think its best.

crates/precompiles/benches/tempo_precompiles.rs
```diff
@@ -3,131 +3,191 @@ use criterion::{Criterion, criterion_group, criterion_main};
...
+    if !StorageContext.spec().is_allegretto() {
+        let mut path_usd = TIP20Token::from_address(PATH_USD_ADDRESS)?;
+        path_usd.initialize(
+            "PathUSD",
+            "PUSD",
+            "USD",
+            Address::ZERO,
+            admin,
+            Address::ZERO,
+        )
```
0xKitsune: Do we need to run benches on anything but the latest hardfork?

crates/precompiles/src/nonce/mod.rs
0xKitsune: +1 on this, can we include this auatomatically in the contract macro?

crates/precompiles/src/nonce/mod.rs
0xKitsune: Yeah I think we can probably move contract initialization out of the contract macro but we can do this in a separate PR to address #1254 IMO.

crates/precompiles/src/nonce/mod.rs
```diff
@@ -130,244 +124,252 @@ impl<'a, S: PrecompileStorageProvider> NonceManager<'a, S> {
...
+                }),
+                NonceEvent::NonceIncremented(INonce::NonceIncremented {
+                    account,
+                    nonceKey: nonce_key,
+                    newNonce: 1,
+                }),
+                NonceEvent::NonceIncremented(INonce::NonceIncremented {
+                    account,
+                    nonceKey: nonce_key,
+                    newNonce: 2,
```
0xKitsune, line 340: nit: looks like we are asserting events twice when we could just be asserting all events at the end of the last call. This is the case in a few of the tests.

crates/precompiles/src/path_usd/mod.rs
```diff
@@ -265,1066 +273,1136 @@ impl<'a, S: PrecompileStorageProvider> PathUSD<'a, S> {
...
+    //     };
+
+    //     fn transfer_test_setup(admin: Address) -> Result<PathUSD> {
+    //         let mut path_usd = PathUSD::new();
+    //         path_usd.initialize(admin)?;
+    //         path_usd.token.grant_role_internal(admin, *ISSUER_ROLE)?;
+
+    //         Ok(path_usd)
+    //     }
+
```
0xKitsune: Can we uncomment/reintroduce all of these tests?

crates/precompiles/src/stablecoin_exchange/dispatch.rs
0xKitsune: Yeah nice, much cleaner for tests!

crates/precompiles/src/stablecoin_exchange/orderbook.rs
```diff
@@ -155,247 +145,93 @@ impl Orderbook {
...
-    /// Read a `TickLevel` at a specific tick
-    pub fn read_tick_level<S: StorageOps>(
-        storage: &mut S,
-        book_key: B256,
-        is_bid: bool,
-        tick: i16,
-    ) -> Result<TickLevel, TempoPrecompileError> {
-        let orderbook_base_slot = mapping_slot(book_key.as_slice(), super::slots::BOOKS);
+impl OrderbookHandler {
+    pub fn handle_tick_level(&self, tick: i16, is_bid: bool) -> TickLevelHandler {
```
0xKitsune: nit: usually I dont comment about naming but this is a bit ambigious, can we update to `get_tick_level` or similar?

crates/precompiles/src/stablecoin_exchange/orderbook.rs
```diff
@@ -155,247 +145,93 @@ impl Orderbook {
...
-            )
-            .write(storage, tick_level)
-        } else {
-            Orders::at_offset(
-                orderbook_base_slot,
-                __packing_orderbook::ASKS_LOC.offset_slots,
-                tick,
-            )
-            .write(storage, tick_level)
+    fn handle_tick_bit(&self, tick: i16, is_bid: bool) -> Result<Slot<U256>> {
```
0xKitsune: Same thing here, this is ambigious naming, we should update so that the function describes what this is doing so its easy to read/reason about.

crates/precompiles/src/stablecoin_exchange/orderbook.rs
```diff
@@ -155,247 +145,93 @@ impl Orderbook {
...
-        let current_word =
-            BitMaps::at_offset(orderbook_base_slot, bitmap_slot, word_index).read(storage)?;
-
         // Set the bit
-        let new_word = current_word | mask;
-        BitMaps::at_offset(orderbook_base_slot, bitmap_slot, word_index)
-            .write(storage, new_word)?;
-
-        Ok(())
+        bitmap.write(current_word | mask)
```
0xKitsune, line 185: Nice, this is way cleaner than it was previously.

crates/precompiles/src/stablecoin_exchange/orderbook.rs
```diff
@@ -625,246 +457,274 @@ mod tests {
...
 
-        impl TestStorage {
-            fn new(chain_id: u64) -> Self {
-                Self(HashMapStorageProvider::new(chain_id))
-            }
+        #[test]
+        fn test_tick_lifecycle() -> eyre::Result<()> {
+            let mut storage = HashMapStorageProvider::new(1);
+            StorageContext::enter(&mut storage, || {
+                let exchange = StablecoinExchange::new();
```
0xKitsune: Do we need to initialize the exchange in these tests?

crates/precompiles/src/tip20/dispatch.rs
```diff
@@ -260,857 +260,671 @@ mod tests {
...
-
-        // Set up initial balance for sender by minting
-        token
-            .mint(
+        StorageContext::enter(&mut storage, || {
+            initialize_path_usd(admin)?;
+            let mut token = TIP20Token::new(1);
+            token.initialize("Test", "TST", "USD", PATH_USD_ADDRESS, admin, Address::ZERO)?;
+            token.grant_role_internal(admin, *ISSUER_ROLE)?;
+            token.mint(
```
0xKitsune, line 387: Can we also use the TIP20Builder in these tests?

crates/precompiles/src/tip20_factory/mod.rs
```diff
@@ -23,31 +22,34 @@ pub struct TIP20Factory {
...
+    pub fn new() -> Self {
+        Self::__new(TIP20_FACTORY_ADDRESS)
     }
 
     /// Initializes the TIP20 factory contract.
     ///
-    /// Ensures the [`TIP20Factory`] account isn't empty and prevents state clear by setting
-    /// placeholder bytecode.
+    /// Sets the initial token counter to 1, reserving token ID 0 for the LinkingUSD precompile.
+    /// Also ensures the [`TIP20Factory`] account isn't empty and prevents state clear.
```
0xKitsune: This is an outdated comment and is not handling token counter logic.

crates/transaction-pool/src/amm.rs
```diff
@@ -133,8 +136,11 @@ impl AmmLiquidityCache {
         for (slot, value) in storage.iter() {
             if let Some(pool) = inner.slot_to_pool.get(slot).copied() {
                 // Update AMM pools
-                let validator_reserve =
-                    U256::from(Pool::from_slot(value.present_value).reserve_validator_token);
+                let validator_reserve = U256::from(
+                    Pool::decode_from_slot(value.present_value)
+                        .expect("unable to decode 'Pool'")
```
0xKitsune: Can we propagate errors? Is there a reason you are using expect here?

crates/precompiles/src/path_usd/mod.rs
0xrusowsky: ups, my bad

crates/precompiles/src/stablecoin_exchange/orderbook.rs
0xrusowsky: sure, i acknowledge that naming is not one of my strengths lol

crates/precompiles-macros/src/storable.rs
```diff
@@ -178,134 +198,175 @@ fn gen_packing_module_from_ir(fields: &[LayoutField<'_>], mod_ident: &Ident) ->
...
         }
     });
 
     quote! {
-        #(#decode_fields)*
+        #(#dynamic_deletes)*
+
+        for slot_offset in 0..#packing::SLOT_COUNT {
+            // Only zero this slot if at least one non-dynamic field lives here
+            if #(#has_non_dynamic_checks)||* {
```
fgimenez: nit, this would be true if any field is static right? in that case all the slots would be zeroed, including dynamic slots which can already be zeroed by their own `delete()`. very minor in any case, extra unnecessary SSTORE gas spent for some slots

crates/revm/src/handler.rs
0xrusowsky: addressed in [`1313e24` (#1177)](https://github.com/tempoxyz/tempo/pull/1177/commits/1313e248b065e9192d8b29b0c8ca9ea6a4441bce)

crates/precompiles/src/account_keychain/mod.rs
0xrusowsky: addressed in [`4d5b6f4` (#1177)](https://github.com/tempoxyz/tempo/pull/1177/commits/4d5b6f47c133c429b2332ad2d22f60043091357c)

crates/precompiles/src/nonce/mod.rs
0xrusowsky: addressed in [`381478c` (#1177)](https://github.com/tempoxyz/tempo/pull/1177/commits/381478ce609f0ff898b9ecd7a79b8690ad5e7a0c)

crates/precompiles/src/nonce/mod.rs
0xrusowsky: so is it good to leave it there for the time being? or should i move it out again?

crates/precompiles/src/stablecoin_exchange/dispatch.rs
0xrusowsky: after we merge the PR we want to move all tests to this setup for consistency and readibility

crates/precompiles/src/stablecoin_exchange/orderbook.rs
0xrusowsky: i just wanted to showcase that this gives the handler, and it doesn't read.

what about `get_tick_level_handler`?

crates/precompiles/src/nonce/mod.rs
0xrusowsky: as discussed privately, ended up adding an `addr`attribute [`d1baef2` (#1177)](https://github.com/tempoxyz/tempo/pull/1177/commits/d1baef2d0d6a8f48b29f4970ceceef47e45370f2)

crates/transaction-pool/src/amm.rs
0xrusowsky: made the fn infallible in [`4d5b6f4` (#1177)](https://github.com/tempoxyz/tempo/pull/1177/commits/4d5b6f47c133c429b2332ad2d22f60043091357c)

crates/precompiles/src/stablecoin_exchange/orderbook.rs
0xrusowsky: `get_tick_bit_handler`?

crates/precompiles/src/stablecoin_exchange/orderbook.rs
klkvr: I like `tick_level_handler`/`get_tick_level_handler`

crates/precompiles/src/tip20/dispatch.rs
0xrusowsky: will do a followup PR to:
- use the `TIP20Setup` helper
- drop `fn setup_storage`
- use `eyre::Result<()>` to propagate errors and avoid `unwrap()`s

so that all tests are consistent

crates/precompiles/src/stablecoin_exchange/orderbook.rs
0xrusowsky: addressed in [`1315fb7` (#1177)](https://github.com/tempoxyz/tempo/pull/1177/commits/1315fb723241fb89deb18a5d43a8afd2a910c535)

crates/precompiles/src/stablecoin_exchange/orderbook.rs
0xrusowsky: addressed in [`1315fb7` (#1177)](https://github.com/tempoxyz/tempo/pull/1177/commits/1315fb723241fb89deb18a5d43a8afd2a910c535)

crates/precompiles/src/nonce/mod.rs
0xrusowsky: addressed in [`6e2e0eb` (#1177)](https://github.com/tempoxyz/tempo/pull/1177/commits/6e2e0eb2c8b1ffa4f83533e3b223037ccc82db86)

crates/precompiles/src/path_usd/mod.rs
0xrusowsky: addressed in [`8e7202c` (#1177)](https://github.com/tempoxyz/tempo/pull/1177/commits/8e7202c7bc24a450fe7d0501932835d951789b8a)

crates/transaction-pool/src/amm.rs
0xKitsune: Can we just add a comment over this line explaining why this is infallible to be explicit?

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -805,7 +797,18 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
...
         taker: Address,
     ) -> Result<(u128, Option<(TickLevel, Order)>)> {
-        let orderbook = self.sload_books(order.book_key())?;
+        let book_handler = self.books.at(book_key);
+        // Pre-Allegretto: use order.book_key() for gas compatibility with legacy behavior.
+        // This is a patched bug where stale best_tick can cause reading an empty order (order_id=0).
+        // ref: https://github.com/tempoxyz/tempo/issues/1062
+        let mut book_handler_order = if self.storage.spec().is_allegretto() {
+            debug_assert_eq!(order.book_key(), book_key);
+            book_handler.clone()
```
0xKitsune, line 806: Just confirming that this has parity with the spec impl?

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -2997,1335 +2886,818 @@ mod tests {
...
+                .apply()?;
+            let eurc = TIP20Setup::create("EURC", "EURC", admin)
+                .with_issuer(admin)
+                .with_mint(alice, min_order_amount_x10)
+                .with_approval(alice, exchange.address, min_order_amount_x10)
+                .apply()?;
+
+            // Create pairs
+            exchange.create_pair(usdc.address())?;
+            exchange.create_pair(eurc.address())?;
```
0xKitsune: nit: we no longer need to create pairs, this will happen automatically when calling place if the pair does not exist.

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -2997,1335 +2886,818 @@ mod tests {
...
+                .apply()?;
+            let eurc = TIP20Setup::create("EURC", "EURC", admin)
+                .with_issuer(admin)
+                .with_mint(alice, min_order_amount_x10)
+                .with_approval(alice, exchange.address, min_order_amount_x10)
+                .apply()?;
+
+            // Create pairs (same as previous test)
+            exchange.create_pair(usdc.address())?;
+            exchange.create_pair(eurc.address())?;
```
0xKitsune: nit: same thing here, we dont need to make this call.

crates/transaction-pool/src/amm.rs
0xrusowsky: note that what u see here is stale (should be marked as "outdated")

in the new code there is no `.expect()` anymore, as the fn doesn't return a `Result` now (thus why i said infallible).

imo no need for a cmnt, as is enforced by the time system

crates/precompiles/src/stablecoin_exchange/orderbook.rs
```diff
@@ -155,247 +145,93 @@ impl Orderbook {
...
-        is_bid: bool,
-    ) -> Result<(), TempoPrecompileError> {
-        if !(MIN_TICK..=MAX_TICK).contains(&tick) {
-            return Err(StablecoinExchangeError::invalid_tick().into());
-        }
+    pub fn set_tick_bit(&mut self, tick: i16, is_bid: bool) -> Result<()> {
+        let mut bitmap = self.get_tick_bit_handler(tick, is_bid)?;
+
+        // Read current bitmap word
+        let current_word = bitmap.read()?;
```
0xKitsune, line 179: Nice, all of the tick bitmap functions are significantly cleaner with these changes.

crates/precompiles-macros/src/storable.rs
0xrusowsky: addressed in [`1984d76` (#1177)](https://github.com/tempoxyz/tempo/pull/1177/commits/1984d768818324f83a5e12094bf378c0c079934b)

crates/precompiles/src/tip20/mod.rs
```diff
@@ -1078,558 +1001,476 @@ pub(crate) mod tests {
...
-        admin: Address,
-        name: &str,
-        symbol: &str,
-    ) -> u64 {
-        initialize_path_usd(storage, admin).unwrap();
-        let mut factory = TIP20Factory::new(storage);
+    fn setup_factory_with_token(admin: Address, name: &str, symbol: &str) -> Result<u64> {
+        initialize_path_usd(admin)?;
+        let mut factory = TIP20Factory::new();
+        factory.initialize()?;
```
0xKitsune: The factory is already initalized in `initialize_path_usd`. Can we remove the extra `factory.initialize()`?

crates/precompiles/src/tip20/rewards.rs
```diff
@@ -670,771 +643,543 @@ mod tests {
...
+            assert_eq!(
+                token.get_balance(alice)?,
+                alice_balance_before_claim + claimed_amount
+            );
+
+            let alice_info = token.get_user_reward_info(alice)?;
+            assert_eq!(alice_info.reward_balance, U256::ZERO);
+
+            Ok(())
+        })
```
0xKitsune, line 1188: Looks like we removed `test_scheduled_rewards_disabled_post_moderato` and `test_cancel_reward_ensure_tip403_is_not_blacklisted`, can we add these back?

crates/precompiles/src/tip20_factory/mod.rs
```diff
@@ -7,47 +7,33 @@ use tempo_precompiles_macros::contract;
...
         TIP20Error, TIP20Token, address_to_token_id_unchecked, is_tip20_prefix, token_id_to_address,
     },
 };
-use alloy::primitives::{Address, Bytes, IntoLogData, U256};
-use revm::state::Bytecode;
+use alloy::primitives::{Address, U256};
 use tracing::trace;
 
-#[contract]
+#[contract(addr = TIP20_FACTORY_ADDRESS)]
```
0xKitsune, line 18: This is nice.

crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -25,33 +24,47 @@ pub struct PolicyData {
...
-        Self::_new(TIP403_REGISTRY_ADDRESS, storage)
+// NOTE(rusowsky): can be removed once revm uses precompiles rather than directly
+// interacting with storage slots.
+impl PolicyData {
+    pub fn decode_from_slot(slot_value: U256) -> Self {
+        use crate::storage::{LayoutCtx, Storable, packing::PackedSlot};
+
+        Self::load(&PackedSlot(slot_value), U256::ZERO, LayoutCtx::FULL)
+            .expect("unable to decode PoliciData from slot")
     }
```
0xKitsune, line 36: Can we add doc comments detailing why this is infallible? Just want to have explicit comments for any unwraps/expects in the codebase so reviewers/devs can clearly understand why this is not a risk.

This also makes the assumptions clear so that if there is any flaw in assumptions, it is very visible and others can flag this.

crates/precompiles/src/tip_account_registrar/mod.rs
```diff
@@ -139,270 +125,302 @@ fn validate_signature(signature: &[u8]) -> Result<(B512, u8)> {
...
     use alloy::sol_types::{SolCall, SolError};
     use alloy_signer::SignerSync;
     use alloy_signer_local::PrivateKeySigner;
+    use tempo_chainspec::hardfork::TempoHardfork;
     use tempo_contracts::precompiles::{TIPAccountRegistrarError, UnknownFunctionSelector};
 
     #[test]
-    fn test_delegate_to_default_v1_pre_moderato() {
-        // Pre-Moderato: delegateToDefault(bytes32,bytes) should work
+    fn delegate_to_default_v1_pre_moderato() -> eyre::Result<()> {
```
0xKitsune: very minor nit: why did we remove `test_`?

crates/precompiles/src/tip_fee_manager/mod.rs
```diff
@@ -470,489 +439,378 @@ mod tests {
...
+            assert!(is_tip20_prefix(token));
+
+            // Random address is not TIP20
+            assert!(!is_tip20_prefix(random));
 
-        let token = Address::random();
-        assert!(!is_tip20_prefix(token));
+            Ok(())
+        })
     }
```
0xKitsune: Should we move this out of the fee manager test suite? Seems unrelated to fee manager logic.

crates/precompiles/src/tip_fee_manager/mod.rs
```diff
@@ -470,489 +439,378 @@ mod tests {
...
+            // Simulate collected fees > actual balance by setting directly
+            fee_manager
+                .collected_fees
+                .at(validator)
+                .write(collected_fees)?;
+            fee_manager
+                .validator_in_fees_array
+                .at(validator)
+                .write(true)?;
+            fee_manager.validators_with_fees.push(validator)?;
```
0xKitsune: Why dont we call `increment_collected_fees` anymore?

crates/precompiles/src/tip_fee_manager/amm.rs
```diff
@@ -54,6 +48,16 @@ pub struct PoolKey {
     pub validator_token: Address,
 }
 
+// TODO(rusowsky): remove this and create a read-only wrapper that is callable from read-only ctx with db access
+impl Pool {
+    pub fn decode_from_slot(slot_value: U256) -> Self {
+        use crate::storage::{LayoutCtx, Storable, packing::PackedSlot};
+
+        Self::load(&PackedSlot(slot_value), U256::ZERO, LayoutCtx::FULL)
+            .expect("unable to decode Pool from slot")
```
0xKitsune, line 58: Same thing here, can we add explicit comments/docs as to why this cant panic? Just makes this more visible/easier to reason when reviewing or interacting with this rather than assuming this is safe or having to check under the hood.

crates/revm/src/handler.rs
```diff
@@ -682,73 +685,70 @@ where
...
-
-            // Convert signature type to precompile SignatureType enum
-            // Use the key_type field which specifies the type of key being authorized
-            let signature_type = match key_auth.key_type {
-                SignatureType::Secp256k1 => PrecompileSignatureType::Secp256k1,
-                SignatureType::P256 => PrecompileSignatureType::P256,
-                SignatureType::WebAuthn => PrecompileSignatureType::WebAuthn,
-            };
+            StorageCtx::enter_precompile(journal, block, cfg, |mut keychain: AccountKeychain| {
+                let access_key_addr = key_auth.key_id;
```
0xKitsune, line 689: Previously we were using `EvmPrecompileStorageProvider::new_max_gas`, does this approach account for gas limits in the same way?

crates/revm/src/handler.rs
```diff
@@ -1237,8 +1234,11 @@ pub fn get_token_balance<JOURNAL>(
 where
     JOURNAL: JournalTr,
 {
+    // Address has already been validated
+    let token_id = tip20::address_to_token_id_unchecked(token);
+
     journal.load_account(token)?;
-    let balance_slot = mapping_slot(sender, tip20::slots::BALANCES);
+    let balance_slot = TIP20Token::new(token_id).balances.at(sender).slot();
```
0xKitsune, line 1241: Looks like we can use `TIP20Token::from_address` here.

xtask/src/genesis_args.rs
```diff
@@ -557,50 +546,50 @@ fn create_and_mint_token(
...
+        let token_id = address_to_token_id_unchecked(token_address);
 
-    let result = token.set_supply_cap(
-        admin,
-        ITIP20::setSupplyCapCall {
-            newSupplyCap: U256::from(u128::MAX),
-        },
-    );
-    assert!(result.is_ok());
+        let mut token = TIP20Token::new(token_id);
```
0xKitsune, line 551: nit: We can use `TIP20Token::from_address` here.

0xKitsune (approved):

crates/commonware-node/src/dkg/manager/validators.rs
```diff
@@ -89,14 +88,13 @@ pub(super) async fn read_from_contract(
...
         let ctx = evm.ctx_mut();
-        let internals = EvmInternals::new(&mut ctx.journaled_state, &ctx.block);
-        let mut provider = EvmPrecompileStorageProvider::new_max_gas(internals, &ctx.cfg);
-
-        let mut validator_config = ValidatorConfig::new(&mut provider);
-        validator_config
-            .get_validators(IValidatorConfig::getValidatorsCall {})
-            .wrap_err("failed to query contract for validator config")?
-    };
+        StorageCtx::enter_evm(&mut ctx.journaled_state, &ctx.block, &ctx.cfg, || {
```
klkvr, line 91: is taking entire `&mut ctx` problematic in some of the places?

crates/precompiles/src/stablecoin_exchange/mod.rs
0xrusowsky: yes, it is.

an invariant that `book_key == order.book_key()` should always hold. However this wasn't the case pre-allegreto patch.

i decided to gatekeep this under the fork so that we ensure the same gas consumption in old forks, but this is just to avoid doing an extra keccak calculation after allegretto (but we could avoid it if we didn't cache handlers)

crates/precompiles/src/tip_account_registrar/mod.rs
0xrusowsky: my bad, stale from when i removed the `test_precompile!` macro, where i got rid of the `test_` prefixes to avoid seeing "test" twice

crates/revm/src/handler.rs
0xrusowsky: this is just a QoL change that i made today after @klkvr's request

`StorageCtx::enter_precompile` will basically create `EvmPrecompileStorageProvider` and call `StorageCtx::enter()` with that provider, thus the behavior should remain identical

crates/revm/src/handler.rs
0xrusowsky: unfortunately not anymore, cause now `TIP20Token::from_address` relies on `StorageCtx` being initialized, which is not the case here.

i could create `from_address_unchecked` if u want

xtask/src/genesis_args.rs
0xrusowsky: same as above

crates/commonware-node/src/dkg/manager/validators.rs
0xrusowsky: in `fn validate_against_state_and_deduct_caller` i would have to drops some mut refs and reset them again. do you think it's worth it?

klkvr: hmm yeah let's keep this as is

crates/precompiles/src/tip403_registry/mod.rs
0xrusowsky: seems that there is some sort of gh issue and it doesn't show on the diffs, but i actually added cmnts to address ur feedback. see https://github.com/tempoxyz/tempo/blob/67913698f7539d04aef6d06a6127fa351236381b/crates/precompiles/src/tip403_registry/mod.rs#L33

crates/precompiles/src/tip_fee_manager/amm.rs
0xrusowsky: addressed in [`5bb71c4` (#1177)](https://github.com/tempoxyz/tempo/pull/1177/commits/5bb71c451410d3a1fdd1a24f4369219bda9a9e64)

crates/precompiles/src/tip403_registry/mod.rs
0xrusowsky: addressed in [`5bb71c4` (#1177)](https://github.com/tempoxyz/tempo/pull/1177/commits/5bb71c451410d3a1fdd1a24f4369219bda9a9e64)

PR #1181, opened 2025-12-02
fix: await FCU on proposal

klkvr:
When proposing, reth expects the `parent_hash` to be part of the canonical chain. However, right now FCU with the parent races the `send_new_payload` call which results in it failing when engine is not able to process FCU soon enough.

It seems that issue is responsible for majority of nullifications happening on Andantino (there's one every few minutes)

Adds a `oneshot` channel that can be used to await the FCU when calling `canonicalize` and uses it in `handle_propose`

### DIFF
--- a/crates/commonware-node/src/consensus/application/executor.rs
+++ b/crates/commonware-node/src/consensus/application/executor.rs
@@ -202,7 +202,7 @@ where
     /// Canonicalizes `digest` by sending a forkchoice update to the execution layer.
     #[instrument(
         skip_all,
-        follows_from = [cause],
+        parent = &cause,
         fields(
             head.height = height,
             head.digest = %digest,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1667,9 +1667,9 @@ checksum = "5e764a1d40d510daf35e07be9eb06e75770908c27d411ee6c92109c9840eaaf7"
 
 [[package]]
 name = "bitcoin-io"
-version = "0.1.3"
+version = "0.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0b47c4ab7a93edb0c7198c5535ed9b52b63095f4e9b45279c6736cec4b856baf"
+checksum = "2dee39a0ee5b4095224a0cfc6bf4cc1baf0f9624b96b367e53b66d974e51d953"
 
 [[package]]
 name = "bitcoin_hashes"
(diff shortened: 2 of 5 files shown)


crates/commonware-node/src/consensus/application/executor.rs
```diff
@@ -479,7 +499,11 @@ struct Message {
 #[derive(Debug)]
 enum Command {
     /// Requests the agent to set the head of the canonical chain to `digest`.
-    CanonicalizeHead { height: u64, digest: Digest },
+    CanonicalizeHead {
+        height: u64,
+        digest: Digest,
+        tx: oneshot::Sender<()>,
```
SuperFluffy: Better name to communicate purpose:
```suggestion
        ack: oneshot::Sender<()>,
```

crates/commonware-node/src/consensus/application/executor.rs
```diff
@@ -447,13 +460,20 @@ pub(super) struct ExecutorMailbox {
 
 impl ExecutorMailbox {
     /// Requests the agent to update the head of the canonical chain to `digest`.
-    pub(super) fn canonicalize_head(&self, height: u64, digest: Digest) -> eyre::Result<()> {
+    pub(super) fn canonicalize_head(
+        &self,
+        height: u64,
+        digest: Digest,
+    ) -> eyre::Result<oneshot::Receiver<()>> {
+        let (tx, rx) = oneshot::channel();
```
SuperFluffy, line 472: keep the mechanism opaque to the caller, i.e. have this function `rx.await` in this function.

crates/commonware-node/src/consensus/application/executor.rs
```diff
@@ -209,6 +212,7 @@ where
         head_or_finalized: HeadOrFinalized,
         height: u64,
         digest: Digest,
+        ack: oneshot::Sender<()>,
```
SuperFluffy, line 219: IMO make this take an option if there are cases where an ack is not required:

```suggestion
        ack: Option<oneshot::Sender<()>>,
```

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -479,18 +479,11 @@ impl Inner<Init> {
...
-            tracing::warn!(
-                %error,
-                parent.height = parent.height(),
-                parent.digest = %parent.digest(),
-                "failed updating canonical head to parent",
-            );
-        }
+            .canonicalize_head(parent.height(), parent.digest())?
+            .await
+            .wrap_err("failed updating canonical head to parent")?;
```
SuperFluffy: I loathe bare `?`, can we do something like this?
```suggestion
        ready(self.state
            .executor_mailbox
            .canonicalize_head(parent.height(), parent.digest())
        ).and_then(|ack| async move { ack.await.map_err(Report::new) })
        .await
        .wrap_err("failed updating canonical head to parent")?;
```

(using `futures::future::ready` and `futures::future::TryFutureExt`

PR #1182, opened 2025-12-02
fix!(stablecoin-exchange): set balance after `transfer_from`

### LINKED ISSUE #1169: Bug: Flip order completion can incorrectly zero a user’s balance if insufficient funds
### Describe the bug

In an edge case, a user’s on-exchange balance can be incorrectly zeroed when a flip order is filled but the user no longer has sufficient balance or approval to fund the subsequent flip order that gets created. 

When a swap triggers the flip order to be filled, the precompile calls calls `place_flip()` which calls `decrement_balance_or_transfer_from()`. This logic zeroes the user’s balance before attempting a `transferFrom`, and if approval was revoked or the wallet balance is too low, that transfer fails and the user is left with a zero balance even though no tokens were transferred. 

Updating the balance only after a successful transfer should prevent this specific issue.

0xKitsune:
Closes #1169

This PR updates the `decrement_balance_or_transfer_from()` function to set balance after successfully transferring from the user. Note that since `allegretto` is already live on testnet, this logic should be gated behind a subsequent hardfork before merging.


crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -274,11 +274,21 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
...
         } else {
-            self.set_balance(user, token, 0)?;
             let remaining = amount
                 .checked_sub(user_balance)
                 .ok_or(TempoPrecompileError::under_overflow())?;
-            self.transfer_from(token, user, remaining)
+
+            // FIXME: Since allegretto is already live, this should be gated behind a subsequent
+            // hardfork.
+            if self.storage.spec().is_allegretto() {
```
fgimenez: isn't the gating here inverted? the buggy version is behind allegretto+, and the fixed one is set for previous hardforks

PR #1191, opened 2025-12-03
 test(stablecoin-exchange): cover decrement_balance_or_transfer_from hardfork behavior

fgimenez:


crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -4866,4 +4866,118 @@ mod tests {
...
+
+        assert!(result.is_err());
+        assert_eq!(
+            exchange.balance_of(alice, base_address)?,
+            internal_balance,
+            "User balance should be preserved when transfer_from fails"
+        );
+
+        Ok(())
+    }
```
0xKitsune: Is this the testing the same thing as `test_decrement_balance_preserves_balance_post_allegretto`?

fgimenez: indeed, with the post allegretto test this one is redundant, removed

PR #1201, opened 2025-12-03
feat(builder): gauge metrics, less verbose payload built log

shekhirin:
1. More gauge metrics
2. Less verbose "Built payload" logs otherwise it's horrific image

### DIFF
--- a/crates/alloy/README.md
+++ b/crates/alloy/README.md
@@ -1,0 +1,34 @@
+# Tempo Alloy
+
+Tempo types for [Alloy](https://alloy.rs).
+
+## Getting Started
+
+To use `tempo-alloy`, add the crate as a dependency in your `Cargo.toml` file:
+
+```toml
+[dependencies]
+tempo-alloy = { git = "https://github.com/tempoxyz/tempo" }
+```
+
+## Development Status
+
+`tempo-alloy` is currently in development and is not yet ready for production use.
+
+## Usage
+
+To get started, instantiate a provider with [`TempoNetwork`]:
+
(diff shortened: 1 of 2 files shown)


crates/payload/builder/src/lib.rs
```diff
@@ -655,9 +666,14 @@ where
         self.metrics.gas_per_second_last.set(gas_per_second);
 
         info!(
-            sealed_block_header = ?sealed_block.sealed_header(),
-            total_transactions = block.transaction_count(),
-            ?payment_transactions,
+            parent_hash = ?sealed_block.parent_hash(),
+            number = sealed_block.number(),
+            hash = ?sealed_block.hash(),
+            timestamp = sealed_block.timestamp_millis(),
```
mattsse, line 672: this is timestamp_millis, maybe we should include both

crates/payload/builder/src/lib.rs
```diff
@@ -655,9 +666,14 @@ where
         self.metrics.gas_per_second_last.set(gas_per_second);
 
         info!(
```
mattsse, line 668: do we also want extradata?

crates/payload/builder/src/lib.rs
shekhirin: this is `timestamp * 1000 + timestamp_millis_part`, should be ok no?

crates/payload/builder/src/lib.rs
shekhirin: added

PR #1205, opened 2025-12-03
feat: `admin_validatorKey`

klkvr:
Exposes an additional `admin` endpoint that exposes validator's public key for validating nodes.

This will be useful for `tempo-bench` where we need to send subblock transactions but don't know which nodes can actually accept them.

### DIFF
--- a/crates/commonware-node/src/args.rs
+++ b/crates/commonware-node/src/args.rs
@@ -1,9 +1,7 @@
 //! Command line arguments for configuring the consensus layer of a tempo node.
 use std::{net::SocketAddr, path::PathBuf, sync::OnceLock};
 
-use commonware_cryptography::{
-    ed25519::{PublicKey},
-};
+use commonware_cryptography::ed25519::PublicKey;
 use eyre::Context;
 use tempo_commonware_node_config::SigningKey;
 
--- a/crates/commonware-node-config/src/lib.rs
+++ b/crates/commonware-node-config/src/lib.rs
@@ -7,11 +7,13 @@ use std::{fmt::Display, net::SocketAddr, path::Path};
 
 use commonware_codec::{DecodeExt as _, Encode as _, FixedSize, Read};
 use commonware_cryptography::{
-    Signer, bls12381::primitives::{
+    Signer,
+    bls12381::primitives::{
         group::Share,
         poly::Public,
         variant::{MinSig, Variant},
-    }, ed25519::{PrivateKey, PublicKey}
+    },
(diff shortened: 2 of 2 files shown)


crates/commonware-node/src/args.rs
```diff
@@ -122,4 +126,36 @@ pub struct Args {
...
     pub subblock_broadcast_interval: jiff::SignedDuration,
+
+    /// Cache for the signing key loaded from CLI-provided file.
+    #[clap(skip)]
+    loaded_signing_key: OnceLock<Option<PrivateKey>>,
+}
+
+impl Args {
+    /// Returns the signing key loaded from specified file.
+    pub fn signing_key(&self) -> eyre::Result<Option<PrivateKey>> {
```
SuperFluffy: IMO add a `fn public_key` next to this and keep this `pub(crate) fn` so that we know that the signing key stays in this crate.

Maybe there is a way to cache just the public key and not the signing key?

crates/commonware-node/src/args.rs
```diff
@@ -122,4 +126,36 @@ pub struct Args {
     /// once we sent it in the first time.
     #[arg(long = "consensus.subblock-broadcast-interval", default_value = "50ms")]
     pub subblock_broadcast_interval: jiff::SignedDuration,
+
+    /// Cache for the signing key loaded from CLI-provided file.
+    #[clap(skip)]
+    loaded_signing_key: OnceLock<Option<PrivateKey>>,
```
SuperFluffy: Let's a) change this to use our `commonware_node_config::SigningKey` but then b) give it a custom `Debug for SigningKey` impl so that we never accidentally leak the key in a debug log.

(I just realize that I just `#[derive(Debug)]` on it, which is really bad)

bin/tempo/Cargo.toml
```diff
@@ -22,6 +22,7 @@ tempo-evm.workspace = true
 tempo-faucet.workspace = true
 
 commonware-runtime = { workspace = true, features = ["external"] }
+commonware-cryptography.workspace = true
```
SuperFluffy: Commented elsewhere - I don't think this dependency is necessary if you provide an inherent method.

crates/commonware-node/src/args.rs
klkvr: > Maybe there is a way to cache just the public key and not the signing key?

main purpose of the cache was to avoid reading the file twice. so we'd need to cache the full key. idk if this is really that important though

PR #1208, opened 2025-12-04
fix(fee-manager): fix `set_fee_token` post moderato

0xKitsune:
Currently, there is a bug in the latest hardfork logic where if a user changes their fee token, they will not be able to set it back to PathUSD in the future. Not allowing Path/LinkingUSD as the fee token was [expected behavior in Moderato](https://github.com/tempoxyz/docs/blob/8319a0eb1332e64a40646fb151566df5a1e7f324/specs/src/FeeManager.sol#L64) but post moderato this should have been updated.

This PR updates to allow PathUSD to be specified when calling `set_fee_token()`.

### DIFF
--- a/docs/api/index-supply.ts
+++ b/docs/api/index-supply.ts
@@ -1,5 +1,5 @@
 import type { VercelRequest, VercelResponse } from '@vercel/node'
-import { tempoAndantino } from 'tempo.ts/chains'
+import { tempoDevnet } from 'tempo.ts/chains'
 
 type QueryRequest = {
   query: string
@@ -65,7 +65,7 @@ export default async function handler(req: VercelRequest, res: VercelResponse) {
     const signatures =
       body.signatures && body.signatures.length > 0 ? body.signatures : ['']
 
-    const chainId = tempoAndantino.id
+    const chainId = tempoDevnet.id
     const chainCursor = `${chainId}-0`
 
     const response = await fetch(url.toString(), {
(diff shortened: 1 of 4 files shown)


crates/node/tests/it/tip20_gas_fees.rs
```diff
@@ -65,6 +65,79 @@ async fn test_fee_in_stable() -> eyre::Result<()> {
     Ok(())
 }
 
+#[tokio::test(flavor = "multi_thread")]
+async fn test_default_fee_token() -> eyre::Result<()> {
```
fgimenez, line 71: what is the goal of this integration test? it passes without the fix, and is not exercising `set_user_token`

crates/node/tests/assets/test-genesis-allegretto.json
klkvr, line 1: can we use the `allegretto_activated` helper instead of introducing an extra test-genesis json? https://github.com/tempoxyz/tempo/blob/8d87bfe1928a0a8cbfdb54ccd303a6fe5c3cb24f/crates/node/tests/it/utils.rs#L230-L233

crates/node/tests/it/tip20_gas_fees.rs
0xKitsune: Yes this test is not directly testing `set_user_token()`, but adding a test to enure that `pathUSD` works as the default fee token. We already test this in other ways but is noticed that we should have an explicit test to assert this.

Issue #1082, opened 2025-11-26
Bug: handling `valid_after` in tx pool correctly


legion2002:
### Describe the bug

We have a `valid_after` field in AA Txs which allow users to schedule a tx in the future. 
But we don't handle this in the mempool right now, which could lead to spam, and incorrectly invalidating txs.

* We should queue txs which are not valid rn, but could be valid in the future
* We should promote such txs with some maintenance service every block, when they become valid.
* Txs too far in the future should be rejected from the tx pool, based on some constant allowed value, to prevent spam.

### Steps to reproduce

NA

### Logs

```text

```

### Container Type

Other, Not running in a container

### What version/commit are you on?

NA

### Code of Conduct

- [x] I agree to follow the Code of Conduct

PR #1218, opened 2025-12-04
refactor: support protocol nonces in 2d pool

klkvr:
For supporting extra validations like https://github.com/tempoxyz/tempo/issues/1082 ideally we need to extend our custom pool's support to not just 2d transactions but also to arbitrary AA (and potentially non-AA) transactions.

This PR starts with customizing the pool to allow protocol nonces and lifting the requirement of non-zero sequence key.

Changes in the PR:
- `nonce_keys` are removed and instead we simply track the reverse-lookup for 2d nonce slots. instead of relying on transaction validator to insert the on-chain nonce into `NonceKeys` we now simply pass it in the `TransactionValidationOutcome::Valid` and take it as input in `add_transaction`
- 2d pool now processes all state updates, including nonce changes

### DIFF
--- a/crates/transaction-pool/src/tt_2d_pool.rs
+++ b/crates/transaction-pool/src/tt_2d_pool.rs
@@ -91,8 +91,9 @@ impl AA2dPool {
 
     /// Entrypoint for adding a 2d AA transaction.
     ///
-    /// ## Limitations
-    /// * This currently assumes that the account's nonce key is already tracked in [`AA2dNonceKeys`], if not then this transaction is considered pending.
+    /// `on_chain_nonce` is expected to be the nonce of the sender at the time of validation.
+    /// If transaction is using 2D nonces, this is expected to be the nonce corresponding
+    /// to the transaction's nonce key.
     pub(crate) fn add_transaction(
         &mut self,
         transaction: Arc<ValidPoolTransaction<TempoPooledTransaction>>,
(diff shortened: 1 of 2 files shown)


crates/transaction-pool/src/metrics.rs
```diff
@@ -18,9 +18,6 @@ pub struct AA2dPoolMetrics {
     /// Number of queued (non-executable) transactions in the AA2D pool
     pub queued_transactions: Gauge,
 
-    /// Number of unique addresses with tracked nonce keys
-    pub tracked_addresses: Gauge,
-
```
klkvr, line 23: we will need an extra index or a `HashSet` collect for exposing this so removed for now, lmk if this seems unreasonable

crates/transaction-pool/src/tt_2d_pool.rs
```diff
@@ -653,17 +660,63 @@ impl AA2dPool {
...
+                        );
+                    }
+                }
+            }
+            let nonce = state
+                .account_info()
+                .map(|info| info.nonce)
+                .unwrap_or_default();
+            changes.insert(AASequenceId::new(*account, U256::ZERO), nonce);
+        }
```
mattsse, line 716: would it be easier to prep this first, so that we cut down on write lock access here?

crates/transaction-pool/src/tt_2d_pool.rs
```diff
@@ -89,9 +96,10 @@ impl AA2dPool {
     pub(crate) fn add_transaction(
         &mut self,
         transaction: Arc<ValidPoolTransaction<TempoPooledTransaction>>,
+        on_chain_nonce: u64,
```
mattsse, line 100: for 2d nonces this is now the nonce that belongs to the non zero noncekey, right?

can we document this?

crates/transaction-pool/src/tt_2d_pool.rs
klkvr: we would still need at least the read lock here I think for the 2d nonces

PR #1220, opened 2025-12-04
feat: add dkg ceremony binary

joshieDo:
Adds `tempo-ceremony` that allows validators to generate the public outcome and their share for the genesis block. Check `README.md` for instructions and flow.

note to self: add to GH CI


bin/tempo-ceremony/src/protocol.rs
```diff
@@ -0,0 +1,849 @@
...
+            .into_iter()
+            .filter_map(|(dealer_idx, shares)| {
+                shares
+                    .iter()
+                    .find(|s| s.index == my_index as u32)
+                    .cloned()
+                    .map(|s| (dealer_idx, s))
+            })
+            .collect();
+
```
klkvr, line 719: should we return an error if any reveals or disqualifications happen?

bin/tempo-ceremony/src/ceremony.rs
```diff
@@ -0,0 +1,376 @@
...
+
+/// Run the ceremony with a pre-existing runtime context.
+async fn run_with_context(setup: CeremonySetup, runtime_ctx: Context) -> eyre::Result<()> {
+    let mut ctx = setup.init(runtime_ctx).await?;
+
+    ctx.wait_for_connections().await?;
+    ctx.distribute_and_collect_acks().await?;
+    ctx.broadcast_and_collect_dealings().await?;
+    let shares = ctx.compute_shares()?;
+    let public_outcome = ctx.collect_outcomes(&shares).await?;
```
klkvr, line 301: it seems that at least one of `outcomes`/`dealings` steps can be omitted?

i.e if we've received correct dealings from everyone there's no way for outcomes to be mismatched

and if we've received correct outcomes from everyone there's no way someone didn't send the proper shares

especially if we're expecting everyone to verify outputs manually post-DKG

klkvr (commented):
the logic looks good to me, though I think under our constraints we can simplify this?

bin/tempo-ceremony/src/protocol.rs
joshieDo: * added for `disqualifications`
* we check for reveals on `validate_dealings`, but added another check here, just in case

https://github.com/tempoxyz/tempo/pull/1220/commits/9b0e2f42c4d14901cedbede2643a9996480649ef

PR #1231, opened 2025-12-04
feat: add guide for rewards

gorried:
Major changes:
  - Adds interactive "Distribute Rewards" guide with full demo flow for TIP-20 rewards
  - Adds reward step components

Minor changes:
  - Add `flowDependencies` prop to DemoStepProps for per-demo prerequisite validation (this lets us use the same component across demos with different dependencies for when they become active)
  - Add `rewardId` to `DemoContext` and `checkFlowDependencies` helper to validate `flowDependencies`
  - Update `ClaimRewards` to support configurable amount prop

Next:
- Thread `flowDependencies` into other components that have a more brittle prop that does the same thing e.g. `CreateTokenPolicy` and `waitForTransfer`


docs/pages/guide/issuance/distribute-rewards.mdx
```diff
@@ -1,8 +1,274 @@
...
+// @noErrors
+export function ClaimReward() {
+  const tokenAddress = '0x...' // Your token address
+
+  const claim = Hooks.reward.useClaimSync()
+
+  const handleClaim = () => { // [!code hl]
+    claim.mutate({ // [!code hl]
+      token: tokenAddress, // [!code hl]
+      feeToken: '0x20c0000000000000000000000000000000000001', // [!code hl]
```
jxom, line 174: is this needed?

gorried: I decided to put in a fee token because we don't mention fee liquidity in the steps at all. But also happy to remove

Issue #1228, opened 2025-12-04
Enable faucet to use AA nonceKeys


Zygimantass:
### Describe the feature

To ensure faucet reliability, we should use nonceKeys instead of nonces. This would both enable us to use the same signer on all nodes and enable higher % of faucet txs landing on-chain

PR #1232, opened 2025-12-04
feat(faucet): use random 2d nonce filler in faucet

Rjected:
fixes https://github.com/tempoxyz/tempo/issues/1228

Uses the 2d nonce filler to send txs in the faucet and replaces some concrete types with `DynProvider<TempoNetwork>`


crates/faucet/src/faucet.rs
```diff
@@ -19,19 +17,19 @@ pub trait TempoFaucetExtApi {
...
-    provider: DynProvider,
+    provider: DynProvider<TempoNetwork>,
 }
 
 impl TempoFaucetExt {
     pub fn new(
         faucet_token_addresses: Vec<Address>,
         funding_amount: U256,
-        provider: FillProvider<impl TxFiller + 'static, impl Provider + 'static>,
+        provider: DynProvider<TempoNetwork>,
```
klkvr, line 27: we could keep this to make sure that non-filling provider doesn't go here but also fine to simplify

Issue #1229, opened 2025-12-04
Switch faucet to use the mint call instead of transfer


Zygimantass:
### Describe the feature

This would prevent a malicious attacker from draining the faucet funds and us having to replenish them. Hitting `u128::MAX` with 10**6 tokens per transfer will take a longer time than we expect the testnet to be live, so this effectively prevents draining attacks.

PR #1233, opened 2025-12-04
fix(faucet): use mint instead of transfer

Rjected:
fixes https://github.com/tempoxyz/tempo/issues/1229

uses `mint` instead of `transfer`


crates/faucet/src/faucet.rs
```diff
@@ -42,7 +42,7 @@ impl TempoFaucetExtApiServer for TempoFaucetExt {
         let mut tx_hashes = Vec::new();
         for token in &self.faucet_token_addresses {
             let tx_hash = *ITIP20::new(*token, &self.provider)
-                .transfer(address, self.funding_amount)
+                .mint(address, self.funding_amount)
```
shekhirin, line 45: I think this will not work out of the box with current infra configs, because we grant the issuer role only to the first address in genesis https://github.com/tempoxyz/tempo/blob/dead860fe506a6a086505108157699927b10de42/xtask/src/genesis_args.rs#L508-L509

But we're using unique addresses per node in our infra https://github.com/tempoxyz/stg-infra/blob/c27b5257929efb508775ca3595839993fd4870ff/clusters/nae/argocd/apps/tempo-networks/devnet-stable/templates/val-funk-tribu.yml#L42 https://github.com/tempoxyz/stg-infra/blob/c27b5257929efb508775ca3595839993fd4870ff/clusters/nae/argocd/apps/tempo-networks/devnet-stable/templates/val-aphex-twin.yml#L43

So I think we either need to change the infra configs to use the first address (the one with issuer role), or give the issuer role to more addresses.

PR #1234, opened 2025-12-04
fix(tip20-factory): check tokenIdCounter in isTIP20

### LINKED ISSUE #1230: BUG: `is_tip20` in the rust impl, does not check tokenId
This issue was caught in the fuzz test CI, the solidity impl checks that an address is only considered `tip_20` if it's tokenId is less than the `tokenIdCounter` in the factory.
We don't have this check in the rust implementation right now.
```
  Location: crates/precompiles/src/tip20/mod.rs:43-45

  Solidity (docs/specs/src/TIP20Factory.sol:73-76):
  function isTIP20(address token) public view returns
  (bool) {
      return bytes12(bytes20(token)) ==
  0x20c000000000000000000000
          && uint64(uint160(token)) < tokenIdCounter;
  // ← checks tokenIdCounter
  }

  Rust (crates/precompiles/src/tip20/mod.rs:43-45):
  pub fn is_tip20(token: Address) -> bool {
      token.as_slice().starts_with(&TIP20_TOKEN_PREFIX)
      // ← MISSING: does NOT check if token_id <
  tokenIdCounter
  }

  Fix needed: The Rust is_tip20 function needs to also
  check that the token ID is less than tokenIdCounter.
```

fgimenez:
Closes #1230 

Fix `isTIP20` to check that token ID < `tokenIdCounter`, matching the Solidity implementation. Previously only checked the address prefix, allowing non-existent token IDs to pass validation.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,6 @@
+* text=auto
+.gitattributes !filter !diff
+
+.env.example linguist-language=Dotenv
+Dockerfile.* linguist-language=Dockerfile
+biome.json linguist-language=JSON-with-Comments
--- a/crates/faucet/src/faucet.rs
+++ b/crates/faucet/src/faucet.rs
@@ -42,7 +42,7 @@ impl TempoFaucetExtApiServer for TempoFaucetExt {
         let mut tx_hashes = Vec::new();
         for token in &self.faucet_token_addresses {
             let tx_hash = *ITIP20::new(*token, &self.provider)
-                .transfer(address, self.funding_amount)
+                .mint(address, self.funding_amount)
                 .send()
                 .await
                 .map_err(|err| rpc_err(INTERNAL_ERROR_CODE, err.to_string(), None))?
--- a/docs/components/guides/steps/index.ts
+++ b/docs/components/guides/steps/index.ts
@@ -27,6 +27,7 @@ export { SendPaymentWithIssuedToken } from './payments/SendPaymentWithIssuedToke
 export {
   CreateSponsorAccount,
   FundSponsorAccount,
+  SendRelayerSponsoredPayment,
   SendSponsoredPayment,
 } from './payments/SponsorUserFees'
 export { ClaimReward } from './rewards/ClaimReward'
--- a/docs/components/guides/steps/payments/SponsorUserFees.tsx
+++ b/docs/components/guides/steps/payments/SponsorUserFees.tsx
@@ -323,3 +323,158 @@ export function SendSponsoredPayment(props: DemoStepProps) {
     </Step>
   )
 }
+
+export function SendRelayerSponsoredPayment(props: DemoStepProps) {
+  const { stepNumber, last = false } = props
+  const { address } = useAccount()
+  const [recipient, setRecipient] = React.useState<string>(FAKE_RECIPIENT)
+  const [memo, setMemo] = React.useState<string>('')
+  const [expanded, setExpanded] = React.useState(false)
+
+  const { data: userBalance, refetch: userBalanceRefetch } =
+    Hooks.token.useGetBalance({
+      account: address,
+      token: alphaUsd,
+    })
+
+  const sendPayment = Hooks.token.useTransferSync({
+    mutation: {
+      onSettled() {
+        userBalanceRefetch()
+      },
+    },
+  })
+
+  useAccountEffect({
+    onDisconnect() {
+      setExpanded(false)
+      sendPayment.reset()
+    },
+  })
+
+  const isValidRecipient = recipient && isAddress(recipient)
+
+  const handleTransfer = () => {
+    if (!isValidRecipient) return
+
+    sendPayment.mutate({
+      amount: parseUnits('100', 6),
+      to: recipient as `0x${string}`,
+      token: alphaUsd,
+      memo: memo ? pad(stringToHex(memo), { size: 32 }) : undefined,
+      feePayer: true,
+    })
+  }
+
+  const active = React.useMemo(() => {
+    return Boolean(address && userBalance && userBalance > 0n)
+  }, [address, userBalance])
+
+  return (
+    <Step
+      active={active && (last ? true : !sendPayment.isSuccess)}
+      completed={sendPayment.isSuccess}
+      actions={
+        expanded ? (
+          <Button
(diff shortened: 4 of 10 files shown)


crates/precompiles/src/tip20_factory/dispatch.rs
```diff
@@ -33,7 +31,7 @@ impl<'a, S: PrecompileStorageProvider> Precompile for TIP20Factory<'a, S> {
                 })
             }
             ITIP20Factory::isTIP20Call::SELECTOR => {
-                view::<ITIP20Factory::isTIP20Call>(calldata, |call| Ok(is_tip20(call.token)))
+                view::<ITIP20Factory::isTIP20Call>(calldata, |call| self.is_tip20(call.token))
```
legion2002, line 34: Should we need to hardfork gate this?

fgimenez: i believe we don't need it, this is a view function that is not called by any other precompile. also the snapshot reexecution is passing

legion2002: Yup makes sense

0xKitsune: Given that this will not roll out until after public testnet, we should put this behind a hardfork. It is possible that within the time between public testnet launch and the hardfork, someone could deploy a contract that calls this function. Since we have an upcoming hardfork anyway, we can include this unless you think we need to roll this out sooner.

crates/precompiles/src/tip20_factory/mod.rs
```diff
@@ -48,6 +48,19 @@ impl<'a, S: PrecompileStorageProvider> TIP20Factory<'a, S> {
         Ok(info.code.is_some())
     }
 
+    /// Returns true if the address is a valid TIP20 token.
+    ///
+    /// Matches the Solidity implementation which checks both:
+    /// 1. The address has the correct TIP20 prefix
+    /// 2. The token ID (lower 8 bytes) is less than tokenIdCounter
+    pub fn is_tip20(&mut self, token: Address) -> Result<bool> {
+        if !crate::tip20::is_tip20(token) {
```
0xKitsune: since this only checks prefix, can we rename `is_tip20` to `is_tip20_prefix`?

crates/precompiles/src/tip20_factory/mod.rs
```diff
@@ -48,6 +48,19 @@ impl<'a, S: PrecompileStorageProvider> TIP20Factory<'a, S> {
...
+            return Ok(false);
+        }
+        let token_id = U256::from(address_to_token_id_unchecked(token));
```
0xKitsune: One additional note, maybe we should add a function called `address_to_token_id` that checks `is_tip20` internally? I imagine there are some places where we are using `address_to_token_id_unchecked` where we would benefit from checking `is_tip20` first. We can do this in a follow on PR.

crates/precompiles/src/tip20_factory/dispatch.rs
fgimenez: makes sense, gated fix behind allegro moderato and added tests, ptal

crates/precompiles/src/tip20_factory/mod.rs
fgimenez: good call done

crates/precompiles/src/tip20_factory/mod.rs
fgimenez: sure will give it a try

crates/precompiles/src/tip20/mod.rs
```diff
@@ -280,8 +284,26 @@ impl<'a, S: PrecompileStorageProvider> TIP20Token<'a, S> {
         self.check_role(msg_sender, DEFAULT_ADMIN_ROLE)?;
 
         // Verify the new quote token is a valid TIP20 token that has been deployed
-        if !is_tip20(call.newQuoteToken) {
-            return Err(TIP20Error::invalid_quote_token().into());
+        if self.storage.spec().is_allegro_moderato() {
+            // Post-AllegroModerato: use factory's is_tip20 which checks both prefix and counter
```
0xKitsune, line 288: Agreed that we should update the spec implementation and the precompile in a follow-on to check that the specified quote token is a valid deployed TIP20 rather than just checking the prefix.

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -453,7 +454,7 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
 
     pub fn create_pair(&mut self, base: Address) -> Result<B256> {
```
0xKitsune, line 455: Same thing here, just noting that we should update this in the spec and precompile in a follow on to check that this is a deployed TIP20 rather than just the prefix.

crates/precompiles/src/tip20_factory/mod.rs
```diff
@@ -427,4 +456,57 @@ mod tests {
...
     }
+
+    #[test]
+    fn test_is_tip20_post_allegro_moderato() -> eyre::Result<()> {
+        let mut storage = HashMapStorageProvider::new(1).with_spec(TempoHardfork::AllegroModerato);
+        let sender = Address::random();
+        initialize_path_usd(&mut storage, sender)?;
+
+        let mut factory = TIP20Factory::new(&mut storage);
+        factory.initialize()?;
```
0xKitsune, line 467: Can we update this test after [1246](https://github.com/tempoxyz/tempo/pull/1246) is merged to deploy pathUSD from the factory?

crates/precompiles/src/tip_fee_manager/mod.rs
```diff
@@ -114,7 +114,7 @@ impl<'a, S: PrecompileStorageProvider> TipFeeManager<'a, S> {
         call: IFeeManager::setValidatorTokenCall,
         beneficiary: Address,
     ) -> Result<()> {
-        if !is_tip20(call.token) {
+        if !is_tip20_prefix(call.token) {
```
0xKitsune, line 117: Same here should update spec/precompile in a follow-on.

PR #1244, opened 2025-12-05
chore: add tempo-forge script to easily run fuzz tests

legion2002:
This allows anyone to easily run fuzz tests locally using `./tempo-forge test`.

We also add another utility command `./tempo-cast` so people can use tempo transaction features in cast easily

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -340,9 +340,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-hardforks"
-version = "0.4.4"
+version = "0.4.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1e29d7eacf42f89c21d7f089916d0bdb4f36139a31698790e8837d2dbbd4b2c3"
+checksum = "2d9a33550fc21fd77a3f8b63e99969d17660eec8dcc50a95a80f7c9964f7680b"
 dependencies = [
  "alloy-chains",
  "alloy-eip2124",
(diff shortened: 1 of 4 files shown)


docs/specs/README.md
```diff
@@ -0,0 +1,134 @@
...
+The `tempo-forge` and `tempo-cast` scripts require the [tempo-foundry](https://github.com/tempoxyz/tempo-foundry) repository.
+
+### Option 1: Clone as Sibling Directory (Recommended)
+
+Clone tempo-foundry as a sibling to the tempo repository:
+
+```
+Tempo/
+├── tempo/              # This repository
+└── tempo-foundry/      # tempo-foundry repository
```
mattsse, line 77: this makes sense, at least I have it like this as well

docs/specs/README.md
```diff
@@ -0,0 +1,134 @@
...
+```
+
+### Building tempo-foundry
+
+The scripts will automatically build the forge/cast binaries on first run. To build manually:
+
+```bash
+cd /path/to/tempo-foundry
+cargo build -p forge --profile dev
+cargo build -p cast --profile dev
```
mattsse, line 101: if you want faster fuzzing then maybe we should do release here?

docs/specs/README.md
0xrusowsky: +1

docs/specs/README.md
legion2002: Yeah I usually prefer dev, because the bottleneck is compile times when I am testing locally.

PR #1246, opened 2025-12-05
fix(tip20): skip quote token currency check for zero quote token post-AllegroModerato

### LINKED ISSUE #1236: Bug: TIP20 initialization enforces quote token currency for first deployed token
### Describe the bug

Currently, TIP20 [initialization logic ](https://github.com/tempoxyz/tempo/blob/5802d32a4f1873af67d24fa533970d6d1915f38e/crates/precompiles/src/tip20/mod.rs#L746-L753) enforces that if the newly deployed token's currency is `USD`, the quote token specified must also be `USD`.

When deploying the first TIP20, the factory requires that the quote token is `Address::ZERO` and will fail this check if the first token deployed specifies `USD` as the currency.

Tests are currently passing because [initialize_path_usd()](https://github.com/tempoxyz/tempo/blob/5802d32a4f1873af67d24fa533970d6d1915f38e/crates/precompiles/src/tip20/mod.rs#L1009-L1022) sets the currency in storage for the first deployed token. This logic is outdated and we should update `initialize_path_usd` to pass in an initialized factory and deploy the first token instead.

```rust
   /// Takes an initialized TIP20 factory, checks that no tokens have been deployed and deploys pathUSD
    pub(crate) fn deploy_path_usd(
        factory: &mut TIP20Factory,
        admin: Address,
    ) -> Result<()> {
        // --snip-- check that no tokens have been deployed else return an error

       // deploy path usd
       factory.create_token(admin, ITIP20Factory::createTokenCall{"PathUSD", "PathUSD", "USD", Address::ZERO, admin})?;
        )
    }

```

TIP20 initialization logic should be relaxed to skip the quote token currency check for the first deployed token.

```rust
// If the currency is USD, the quote token must also be USD
if currency == USD_CURRENCY {
   let quote_token_currency = TIP20Token::from_address(quote_token).currency()?;

// If the current spec is post allegro moderato and the quote token is zero, skip this check
if self.storage.spec().is_allegro_moderato() && !quote_token.is_zero() {
   if quote_token_currency != USD_CURRENCY {
   }
}
}

```

This is not currently an issue on testnet since PathUSD is already deployed but this should be patched for subsequent deployments.

fgimenez:
Closes #1236

Relaxes TIP20 initialization to allow USD tokens with `Address::ZERO` as quote token when deploying the first token post-AllegroModerato. Previously, the quote token currency check would fail because `Address::ZERO` has no currency stored.

  - Skip quote token currency check in `initialize()` when quote_token is `Address::ZERO` post-AllegroModerato
  - Update `initialize_path_usd` to use factory flow for AllegroModerato+ specs
  - Add `deploy_path_usd` helper for explicit factory-based PathUSD deployment

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ scripts/data
 localnet/**
 
 .claude/
+!docs/.claude/
 
 # Docs
 docs/dist
--- a/docs/.claude/commands/sync-tempo-ts-docs.md
+++ b/docs/.claude/commands/sync-tempo-ts-docs.md
@@ -1,0 +1,751 @@
+# Sync tempo.ts SDK Documentation
+
+This skill documents the process for adding missing documentation for tempo.ts SDK functions. Use this when new functions are added to the SDK or when documentation gaps are discovered.
+
+## Overview
+
+The tempo.ts SDK has three layers that need documentation:
+1. **Viem Actions** - Low-level viem client actions (`pages/sdk/typescript/viem/`)
+2. **Wagmi Actions** - Wagmi-compatible actions (`pages/sdk/typescript/wagmi/actions/`)
+3. **Wagmi Hooks** - React hooks wrapping wagmi actions (`pages/sdk/typescript/wagmi/hooks/`)
+
+Each layer has corresponding source files in `node_modules/tempo.ts/src/`:
+- `viem/Actions/{module}.ts`
+- `wagmi/Actions/{module}.ts`
+- `wagmi/Hooks/{module}.ts`
+
+## Step 1: Identify Missing Documentation
+
+Compare source files against existing documentation:
+
+```bash
+# Source locations
+node_modules/tempo.ts/src/viem/Actions/{module}.ts
+node_modules/tempo.ts/src/wagmi/Actions/{module}.ts
+node_modules/tempo.ts/src/wagmi/Hooks/{module}.ts
+
+# Documentation locations
+pages/sdk/typescript/viem/{module}.{function}.mdx
+pages/sdk/typescript/wagmi/actions/{module}.{function}.mdx
+pages/sdk/typescript/wagmi/hooks/{module}.use{Function}.mdx
+
+# Index files that list all functions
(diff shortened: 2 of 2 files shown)


crates/precompiles/src/tip20/mod.rs
```diff
@@ -1005,19 +1009,44 @@ pub(crate) mod tests {
...
+        factory.create_token(
             admin,
-            Address::ZERO,
+            ITIP20Factory::createTokenCall {
+                name: "PathUSD".to_string(),
+                symbol: "PUSD".to_string(),
+                currency: "USD".to_string(),
+                quoteToken: Address::ZERO,
+                admin,
+            },
```
0xKitsune, line 1066: Can we add a check to ensure that `tokenId == 0` and this is the first token deployed? This will remove any potential footguns when deploying tokens in tests and ensures PathUSD is first. This might already be covered though since `quoteToken: Address::ZERO` but it might be nice to have an easy error to read so we can quickly fix when writing tests rather than having to debug.

crates/precompiles/src/tip20/mod.rs
```diff
@@ -2680,4 +2709,119 @@ pub(crate) mod tests {
...
+            eur_token_address,
+            admin,
+            Address::ZERO,
+        );
+
+        assert!(
+            result.is_err(),
+            "USD token should not be allowed to have a non-USD quote token"
+        );
+    }
```
0xKitsune, line 2816: nit: consider aggregating these into `test_initialize_pre_allegro_moderato` and `test_intialize_post_moderato` and asserting usd/non usd conditions in each.

crates/precompiles/src/tip20/mod.rs
fgimenez: ok, returning now `TIP20Error::invalid_quote_token()` in case token id != 0

crates/precompiles/src/tip20/mod.rs
fgimenez: sure done

PR #1248, opened 2025-12-05
feat: support delaying transactions in 2d pool

### LINKED ISSUE #1082: Bug: handling `valid_after` in tx pool correctly

klkvr:
Closes #1082 

Refactors 2d pool to properly track delayed transactions and make sure that we never put a transaction with `valid_after > now` into `indenependent_transactions` set.

With this PR all AA transactions with `valid_after` are now routed into our custom pool.


crates/transaction-pool/src/maintain.rs
```diff
@@ -109,8 +109,10 @@ where
 {
     let mut events = pool.client().canonical_state_stream();
     while let Some(notification) = events.next().await {
-        pool.notify_aa_pool_on_state_updates(
+        let timestamp = notification.tip().header().timestamp();
+        pool.notify_aa_pool_on_state_and_timestamp(
             notification.committed().execution_outcome().state().state(),
+            timestamp,
```
mattsse: should this use timestamp millis?

crates/transaction-pool/src/tt_2d_pool.rs
```diff
@@ -56,6 +58,8 @@ pub struct AA2dPool {
     ///
     /// This identifies the account and nonce key based on the slot in the `NonceManager`.
     address_slots: HashMap<U256, (Address, U256)>,
+    /// Timestamp of the last block.
+    tip_timestamp: u64,
```
mattsse: can we document whether this is unix or  with millis?

mattsse (commented):
all of this looks right and makes sense

one question re the actual timestamp value

crates/transaction-pool/src/transaction.rs
```diff
@@ -147,6 +154,10 @@ pub enum TempoPoolTransactionError {
         "Insufficient liquidity for fee token: {0}, please see https://docs.tempo.xyz/documentation/protocol/fees for more"
     )]
     InsufficientLiquidity(Address),
+
+    /// Thrown when a new transaction attempts to replace another one, but has a higher `valid_after` timestamp.
+    #[error("Can't replace a transaction with a higher `valid_after` timestamp")]
+    InvalidValidAfterReplacement,
```
mattsse, line 160: this is reasonable imo

crates/transaction-pool/src/tt_2d_pool.rs
```diff
@@ -48,6 +51,8 @@ pub struct AA2dPool {
     by_id: BTreeMap<AA2dTransactionId, AA2dInternalTransaction>,
     /// _All_ transactions by hash.
     by_hash: HashMap<TxHash, Arc<ValidPoolTransaction<TempoPooledTransaction>>>,
+    /// Transactions that are delayed until a configured `valid_after` timestamp.
+    delayed: BTreeMap<u64, HashSet<AA2dTransactionId>>,
```
mattsse, line 55: this should work because the distance from current head is also enforced so we we'll only have 1,2,3 entries? in the btreemap?

mattsse, line 55: we also dont need to do any housekeeping here because if he chain ticks, outdated entries will be removed automatically?

crates/transaction-pool/src/tt_2d_pool.rs
```diff
@@ -617,6 +663,60 @@ impl AA2dPool {
...
+                } else {
+                    // For all descendants, set the flag to false as we ensure no nonce gaps.
+                    tx.has_queued_ancestors_or_nonce_gap = false;
+                }
+
+                if tx.is_independent() {
+                    // Record transactions in independent set if they are independent.
+                    self.independent_transactions
+                        .insert(id.seq_id, tx.inner.clone());
+                }
```
mattsse, line 704: this makes sense

mattsse (approved):
lgtm, all of this looks correct and should be quite efficient from what I can tell

PR #1249, opened 2025-12-05
feat(consensus): active dynamic validators at allegretto-time=0

### LINKED ISSUE #1087: Consensus runs post-allegretto DKG logic if allegretto time is 0
### Describe the feature

This requires writing a) the initial DKG state into the `extraData` field of the genesis header, and b) writing the initial validator state into the genesis block.

### Additional context

_No response_

### LINKED ISSUE #1211: Include validators into the genesis at generation
### Describe the feature

We currently only add validators to the genesis chainspec, instead we should also add them to the validator system contract on generation.

### Additional context

_No response_

SuperFluffy:
This patch ensures that dynamic validator logic is active when `allegretto-time = 0` is set in genesis.

The xtask tool is changed to always write the initial DKG outcome into the genesis header and the initial validator set into the genesis validator config contract. To simulate a hardfork on testnet, the escape hatches `--no-validators-in-genesis` and `--no-dkg-in-genesis` are provided.

The test setup was expanded to also cover the cases where `allegretto-time = 0` to test the specific code paths of the actor.

Closes #1087
Closes #1211


crates/commonware-node/src/dkg/manager/actor/post_allegretto.rs
```diff
@@ -44,6 +46,79 @@ where
...
+                        spec.genesis().extra_data.len(),
+                    )
+                })?;
+
+            ensure!(
+                initial_dkg_outcome.epoch == 0,
+                "at genesis, the epoch must be zero, but genesis reported `{}`",
+                initial_dkg_outcome.epoch
+            );
+            let initial_validators = validators::read_from_contract(
```
joshieDo, line 96: can we  add  `ensure! initial_validators.is_empty().not()`

joshieDo: or to the dkg.participants

crates/commonware-node/src/dkg/manager/actor/post_allegretto.rs
```diff
@@ -44,6 +46,79 @@ where
...
+                .epoch_metadata
+                .put_sync(
+                    CURRENT_EPOCH_KEY,
+                    EpochState {
+                        dkg_outcome: DkgOutcome {
+                            dkg_successful: true,
+                            epoch: 0,
+                            participants: initial_dkg_outcome.participants,
+                            public: initial_dkg_outcome.public,
+                            share: self.config.initial_share.clone(),
```
joshieDo, line 135: we should validate that the `initial_share` is correct here, or the node might [silently be spawned as a verifier](https://github.com/commonwarexyz/monorepo/blob/709d24861c4ebe0091e50337f6d2f61d1547fdb1/consensus/src/simplex/signing_scheme/bls12381_threshold.rs#L103-L117) later on

crates/commonware-node/src/dkg/manager/actor/post_allegretto.rs
SuperFluffy: IMO this is redundant: a few lines down we `ensure!(peers_as_per_contract.keys() == initial_dkg_outcome.participants)`. Since `participants` cannot be empty, the on-chain state cannot be empty.

crates/commonware-node/src/dkg/manager/actor/post_allegretto.rs
SuperFluffy: Good catch. I added that in  38589362

PR #1259, opened 2025-12-05
feat: add sync docs claude command

gorried:
Adds the command mentioned in #1255 but was ignored due to `.gitignore` at root of the repo

Updates the repo `.gitignore` to allow docs claude skills


docs/.claude/commands/sync-tempo-ts-docs.md
tmm, line 1: should we throw in the ole "suggest updates to this file if you notice patterns changing..."

PR #1261, opened 2025-12-05
feat(ci): optimize builds by adding a bake file

Zygimantass:
very esoteric optimizations:
* enabled autoscaling on Depot, 2x builds per machine max
* switched to docker.bake.hcl to trigger everything as one build. Depot automatically splits off after the base image is built to optimize for caching
* there's now a base image with chef, meaning chef cook is triggered only once per build. Depot says they do this automatically, but I couldn't confirm that looking at the builds themselves.
* removed dependencies with `cargo machete` and `cargo udeps`. I think this was majority of the improvement.


Cargo.toml
```diff
@@ -81,8 +81,9 @@ codegen-units = 16
 # e.g. `cargo build --profile profiling`
 [profile.profiling]
 inherits = "release"
-debug = "full"
+debug = "line-tables-only"
 strip = "none"
+incremental = true
```
DaniPopes, line 86: why was this changed? note that incremental can hinder performance and having less debug info is not really what u want here

PR #1271, opened 2025-12-06
fix: use correct createX deployed bytecode

howydev:
Uses createx bytecode with immutables set. value grabbed from etherscan deployed contract at 0xba5ed...

Fixes: https://github.com/tempoxyz/tempo/issues/1274

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7414,7 +7414,7 @@ checksum = "1e061d1b48cb8d38042de4ae0a7a6401009d6143dc80d2e2d6f31f0bdd6470c7"
 [[package]]
 name = "reth-basic-payload-builder"
 version = "1.9.3"
-source = "git+https://github.com/paradigmxyz/reth?rev=bfcd46d#bfcd46d01dd428b184a7003a3255b9741d7c8d0c"
+source = "git+https://github.com/paradigmxyz/reth?rev=ac0f968#ac0f9687bddf5bede680c769acb47c181e4659dd"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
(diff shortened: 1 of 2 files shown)


crates/evm/src/block.rs
```diff
@@ -491,6 +492,35 @@ where
...
+
+            let acc = db
+                .load_cache_account(CREATEX_ADDRESS)
+                .map_err(BlockExecutionError::other)?;
+
+            let mut acc_info = acc.account_info().unwrap_or_default();
+
+            // Only update if bytecode doesn't match the correct (post-fork) version
+            let correct_code = tempo_contracts::contracts::CREATEX_POST_ALLEGRO_MODERATO_BYTECODE;
+            let correct_code_hash = alloy_primitives::keccak256(&correct_code);
```
klkvr: let's have a constant for the correct codehash

this is a bit wasteful right now and will end up doing a keccak on every block

0xKitsune: Also do we need to check this every time? Or can we update the check to only run at the fork block? Something like the following.

```
if self. inner.spec.is_allegro_moderato_fork_block(){
   // Update the account code for createx
   // --snip--
}
```

crates/contracts/src/lib.rs
```diff
@@ -58,6 +58,10 @@ pub mod contracts {
...
+    );
```
0xKitsune, line 64: Is there a way we can add tests to confirm this is now correct? Ideally we run existing foundry test suites against all predeploys. Let me know if you have any thoughts.

howydev: possible, but im pretty confident that this would work

this value is taken from: https://etherscan.io/address/0xba5Ed099633D3B313e4D5F7bdc1305d3c28ba5Ed#code

howydev: ill try to run a test. worried that ill spend too much time wrangling config, will timebox it to 2h

0xKitsune (approved):
Lgtm pending @klkvr

crates/contracts/src/lib.rs
```diff
@@ -58,6 +58,13 @@ pub mod contracts {
...
+
+    pub const CREATEX_POST_ALLEGRO_MODERATO_BYTECODE_HASH: B256 =
+        b256!("bd8a7ea8cfca7b4e5f5041d7d4b17bc317c5ce42cfbc42066a00cf26b43eb53f");
```
klkvr:
```suggestion
        b256!("0xbd8a7ea8cfca7b4e5f5041d7d4b17bc317c5ce42cfbc42066a00cf26b43eb53f");
```

klkvr (approved):

PR #1273, opened 2025-12-06
test: update default `TempoHardfork` and patch tests

### LINKED ISSUE #1100: Testing pre- and post hardfork behaviour
### Describe the feature

We want a test suite that ensures test pass for Moderato and Allegretto, on top of adding tests for 

e.g. https://github.com/tempoxyz/tempo/pull/1057 updates default chain 

### Additional context

_No response_

0xKitsune:
Closes #1100 

This PR updates the default `TempoHardfork` to `Allegretto` and updates tests. Additionally tests were added to assert pre/post hardfork behaviors.


crates/precompiles/src/tip_fee_manager/amm.rs
```diff
@@ -1424,4 +1424,152 @@ mod tests {
             ))
         ));
     }
+
+    ///// Tests the mean calculation in add_liquidity for pre-Moderato hardfork
```
fgimenez: why are these tests commented? should we keep them?

crates/precompiles/src/path_usd/dispatch.rs
```diff
@@ -257,11 +281,13 @@ mod tests {
...
 
         assert_full_coverage([itip20_unsupported, roles_unsupported]);
+
+        Ok(())
     }
 
     #[test]
-    fn test_start_reward_disabled() -> eyre::Result<()> {
-        let mut storage = HashMapStorageProvider::new(1);
+    fn test_start_reward_disabled_pre_allegretto() -> eyre::Result<()> {
```
fgimenez: nit: aiui rewards were enabled in adagio, disabled pre-allegretto can be misleading, wdyt? maybe better disabled post moderato?

0xKitsune: Pushed a commit to update to `_post_moderato`.

crates/precompiles/src/tip_fee_manager/amm.rs
0xKitsune: Updated to fix and reintroduce these tests.

PR #1277, opened 2025-12-07
fix: update `reserve_liquidity` calculation

0xKitsune:
This PR fixes the `reserve_liquidity` calculation in the fee manager.

[Thread for reference](https://tempoxyz.slack.com/archives/C09PGSBNXFG/p1765089790918429?thread_ts=1764913784.505099&cid=C09PGSBNXFG)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -163,7 +163,7 @@ reth-revm = { git = "https://github.com/paradigmxyz/reth", rev = "3830c76", feat
   "std",
   "optional-checks",
 ] }
-revm = "33.1.0"
+revm = { version = "33.1.0", features = ["optional_fee_charge"] }
 
 alloy = { version = "1.1.2", default-features = false }
 alloy-consensus = "1.1.2"
--- a/bin/tempo-bench/README.md
+++ b/bin/tempo-bench/README.md
@@ -36,28 +36,89 @@ Usage: tempo-bench run-max-tps [OPTIONS] --tps <TPS>
 Options:
   -t, --tps <TPS>
           Target transactions per second
+
   -d, --duration <DURATION>
-          Test duration in seconds [default: 30]
+          Test duration in seconds
+
+          [default: 30]
+
   -a, --accounts <ACCOUNTS>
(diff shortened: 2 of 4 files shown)


crates/precompiles/src/tip_fee_manager/amm.rs
```diff
@@ -93,24 +93,44 @@ impl<'a, S: PrecompileStorageProvider> TipFeeManager<'a, S> {
...
                         .map_err(|_| TempoPrecompileError::under_overflow())?,
                 )
-                .ok_or(TempoPrecompileError::under_overflow())?,
-        )?;
+                .ok_or(TempoPrecompileError::under_overflow())?;
+
+            let total_out_needed = compute_amount_out(U256::from(new_total_pending))?;
+
+            let pool = self.sload_pools(pool_id)?;
+            if total_out_needed > U256::from(pool.reserve_validator_token) {
```
fgimenez: we use `pool.reserve_validator_token` post allegro-moderato, and `get_effective_validator_reserve` pre hardfork, would be good to have a smol comment about why

crates/precompiles/src/tip_fee_manager/amm.rs
```diff
@@ -93,24 +93,44 @@ impl<'a, S: PrecompileStorageProvider> TipFeeManager<'a, S> {
...
+                pool_id,
+                current_pending_fee_swap_in
+                    .checked_add(
+                        max_amount
+                            .try_into()
+                            .map_err(|_| TempoPrecompileError::under_overflow())?,
+                    )
+                    .ok_or(TempoPrecompileError::under_overflow())?,
+            )?;
+        }
```
klkvr: we should be able to only keep the new logic because it's unlikely that this was ever reached before

crates/precompiles/src/tip_fee_manager/amm.rs
0xKitsune: This is no longer the case now that we are only using the latest logic with the fix.

PR #1279, opened 2025-12-07
feat(consenus): don't eagerly canonicalize just-proposed blocks

klkvr:
With this patch a proposer no longer eagerly canonicalizes its just proposed block (by sending an `FCU(head_hash = proposed_hash)` to the execution layer). If the block is not notarized by the network, then the proposer will necessarily have to reorg its `head_hash` on the next view. Since blocks not being notarized is standard behavior of the network, and since there is no advantage in the proposer eagerly canonicalizing its proposal, this step should be removed.

In case the proposal `P` of view `V` is accepted by the network, the proposer of view `V` will canonicalize `P` in view `V+1` when processing the next proposal with `P` its parent.


crates/commonware-node/src/dkg/manager/validators.rs
```diff
@@ -38,19 +40,40 @@ pub(super) async fn read_from_contract(
     epoch_length: u64,
 ) -> eyre::Result<OrderedAssociated<PublicKey, DecodedValidator>> {
     let last_height = last_height_before_epoch(for_epoch, epoch_length);
+
+    // Try mapping the block height to a hash tracked by reth.
+    //
+    // Firstly check the canonical chain and fallback to pending block state.
```
SuperFluffy:
```suggestion
    // First check the canonical chain, then fallback to pending block state.
    //
    // Necessary because the DKG and application actors process finalized block concurrently.
```

PR #1282, opened 2025-12-07
chore: allow skips in test-success

gorried:
Added the allowed-skips parameter to the alls-green action

When a PR only modified files under docs/, the workflow's paths-ignore filter caused all test jobs to be skipped. However, the test-success job (which is a required check for PR merges) still ran due to its if: `always()` condition. The all green action didn't know how to handle all jobs being in a `skipped` state, causing the check to fail and block PR merges.

Ran into this on a docs PR:
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -97,9 +97,9 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "alloy"
-version = "1.1.2"
+version = "1.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9e15860af634cad451f598712c24ca7fd9b45d84fff68ab8d4967567fa996c64"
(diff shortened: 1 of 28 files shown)


.github/workflows/test.yml
```diff
@@ -101,3 +101,4 @@ jobs:
         uses: re-actors/alls-green@release/v1
         with:
           jobs: ${{ toJSON(needs) }}
+          allowed-skips: test,msrv,genesis,snapshot-re-execute
```
struong, line 104: is there a way to conditionally allow skip for only changes to docs?

struong: as I understand
- there's a branch protection rule that requires `test success` to be ✅ 
- `test.yml` doesn't run for `docs/**` https://github.com/tempoxyz/tempo/pull/1282/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R9-R10
- gridlock :( 

consulted chat on how we can get around this w/o allowing skips at the job level – there should be an option to "not require the check when there are no code changes, aka Allow skipping". which should allow `docs` changes to go through without this job passing

PR #1289, opened 2025-12-08
fix(precompiles): enforce TIP20 blacklist on internal balance operations in StablecoinExchange

### LINKED ISSUE #1288: Blacklisted users can extract value from the StablecoinExchange by operating entirely on internal balances
### Describe the bug

ref https://tempoxyz.slack.com/archives/C09PGSBNXFG/p1764913784505099

When decrement_balance_or_transfer_from() is called during order placement (line 438) or swaps (line 255), users with sufficient internal balance only have their accounting decremented via sub_balance() at line 197 - no TIP20 transfer occurs, therefore no blacklist check.

### Steps to reproduce

N/A

### Logs

```text

```

### Platform(s)

_No response_

### Container Type

Not running in a container

da49fd8cb0aca41837e1e82249fed437d20e3a8d

_No response_

### Code of Conduct

fgimenez:
Closes #1288 

Adds blacklist authorization check in `decrement_balance_or_transfer_from()` before operating on internal balances. Previously, users with sufficient internal balance bypassed TIP20 transfer policy checks since no actual token transfer occurred.

### DIFF
--- a/docs/components/ConnectWallet.tsx
+++ b/docs/components/ConnectWallet.tsx
@@ -7,6 +7,7 @@ import {
   useSwitchChain,
 } from 'wagmi'
 import { Button, Logout } from './guides/Demo'
+import { filterSupportedInjectedConnectors } from './lib/wallets'
 
 export function ConnectWallet({
   showAddChain = true,
@@ -17,7 +18,7 @@ export function ConnectWallet({
   const connect = useConnect()
   const connectors = useConnectors()
   const injectedConnectors = React.useMemo(
-    () => connectors.filter((connector) => connector.id !== 'webAuthn'),
+    () => filterSupportedInjectedConnectors(connectors),
     [connectors],
   )
   const switchChain = useSwitchChain()
--- a/docs/components/guides/steps/wallet/ConnectWallet.tsx
+++ b/docs/components/guides/steps/wallet/ConnectWallet.tsx
@@ -10,6 +10,7 @@ import {
 } from 'wagmi'
 import LucideCheck from '~icons/lucide/check'
 import LucideWalletCards from '~icons/lucide/wallet-cards'
+import { filterSupportedInjectedConnectors } from '../../../lib/wallets'
 import { Button, Step, StringFormatter, useCopyToClipboard } from '../../Demo'
 import type { DemoStepProps } from '../types'
 
@@ -21,7 +22,7 @@ export function ConnectWallet(props: DemoStepProps) {
   const disconnect = useDisconnect()
   const connectors = useConnectors()
   const injectedConnectors = React.useMemo(
-    () => connectors.filter((connector) => connector.id !== 'webAuthn'),
+    () => filterSupportedInjectedConnectors(connectors),
     [connectors],
   )
   const switchChain = useSwitchChain()
--- a/docs/components/lib/wallets.ts
+++ b/docs/components/lib/wallets.ts
@@ -1,0 +1,13 @@
+import type { Connector } from 'wagmi'
+
+const UNSUPPORTED_WALLET_IDS = new Set(['app.phantom'])
+const UNSUPPORTED_WALLET_NAMES = new Set(['Phantom'])
+
+export function filterSupportedInjectedConnectors(connectors: readonly Connector[]) {
+  return connectors.filter(
+    (connector) =>
+      connector.id !== 'webAuthn' &&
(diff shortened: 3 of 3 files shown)


crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -264,13 +264,29 @@ impl<'a, S: PrecompileStorageProvider> StablecoinExchange<'a, S> {
...
 
+    fn ensure_transfer_authorized(&mut self, token: Address, user: Address) -> Result<()> {
+        if token == PATH_USD_ADDRESS {
+            PathUSD::new(self.storage)
+                .token
+                .ensure_transfer_authorized(user, self.address)?;
+        } else {
+            TIP20Token::from_address(token, self.storage)?
+                .ensure_transfer_authorized(user, self.address)?;
+        }
```
0xKitsune: Now that PathUSD is treated like any other TIP20, we can just use `TIP20Token::from_address(token, self.storage)?.ensure_transfer_authorized(user, self.address)?;`. With this in mind, we probably can also remove the extra `ensure_transfer_authorized` function and jus tuse `TIP20.ensure_transfer_authorized` inline instead.

fgimenez: sure done

howydev (approved):

legion2002 (commented):
@fgimenez Pushed the solidity spec changes and tests for this.
Currently the precompile fuzz tests don't work because allegro moderato is not active on tempo-foundry.

Discussing with the foundry team what's the best workflow solution here.
CC: @grandizzy

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -4284,4 +4289,93 @@ mod tests {
             Ok(())
         })
     }
+
+    #[test]
+    fn test_blacklisted_user_cannot_use_internal_balance() -> eyre::Result<()> {
```
0xKitsune, line 4295: Can we update this test to use the new `StorageCtx` approach? You can check out other tests in this mod for reference.

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -4284,4 +4289,93 @@ mod tests {
...
+        let mut base = TIP20Token::new(1, quote.token.storage());
+        base.initialize("BASE", "BASE", "USD", quote_address, admin, Address::ZERO)?;
+        base.grant_role_internal(admin, *ISSUER_ROLE)?;
+        let base_address = base.address();
+        base.change_transfer_policy_id(
+            admin,
+            ITIP20::changeTransferPolicyIdCall {
+                newPolicyId: policy_id,
+            },
+        )?;
```
0xKitsune: We can use the new [TIP20Setup in the test_util mod](https://github.com/tempoxyz/tempo/blob/main/crates/precompiles/src/test_util.rs#L201-L230) to simplify this.

docs/specs/src/StablecoinExchange.sol
```diff
@@ -220,6 +223,15 @@ contract StablecoinExchange is IStablecoinExchange {
...
             }
 
+            // Check if maker is authorized by the token's transfer policy before operating on internal balance
+            uint64 policyId = ITIP20(escrowToken).transferPolicyId();
+            if (
+                !TIP403_REGISTRY.isAuthorized(policyId, maker)
+                    || !TIP403_REGISTRY.isAuthorized(policyId, address(this))
+            ) {
+                revert ITIP20.PolicyForbids();
+            }
```
0xKitsune, line 233: It might be nice to update the `TIP20` contract  expose an `isAuthorized` function in a subsequent PR. Any thoughts @howydev @legion2002?

crates/precompiles/src/stablecoin_exchange/mod.rs
fgimenez: sure done

crates/precompiles/src/stablecoin_exchange/mod.rs
fgimenez: yep, already done as part of the StorageCtx refactor, ptal

docs/specs/src/StablecoinExchange.sol
howydev: Could see it being useful for defi, but would be a lower priority for now. If we have a tracking list, can add to that

PR #1292, opened 2025-12-08
feat: add feeamm tests

howydev:
This bumps fee amm test coverage of {lines, function, branches} from ~30-40% to 100%

No errors when run with foundry-tempo locally

### DIFF
--- a/docs/specs/test/FeeAMM.t.sol
+++ b/docs/specs/test/FeeAMM.t.sol
@@ -134,6 +134,12 @@ contract FeeAMMTest is BaseTest {
             assertEq(err, abi.encodeWithSelector(IFeeAMM.InvalidCurrency.selector));
         }
 
+        try amm.mintWithValidatorToken(address(validatorToken), address(eurToken), 1e18, alice) {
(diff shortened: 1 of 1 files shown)


docs/specs/test/FeeAMM.t.sol
```diff
@@ -93,11 +96,58 @@ contract FeeAMMTest is BaseTest {
...
+        }
+
+        // ONLY_USD_TOKENS (valid TIP20 but non-USD currency)
+        TIP20 eurToken = TIP20(factory.createToken("Euro", "EUR", "EUR", pathUSD, admin));
+
+        try amm.mintWithValidatorToken(address(eurToken), address(validatorToken), 1e18, alice) {
+            revert CallShouldHaveReverted();
+        } catch (bytes memory err) {
+            assertEq(err, abi.encodeWithSelector(IFeeAMM.InvalidCurrency.selector));
+        }
```
0xKitsune, line 141: We should test both cases when the user token is USD but the validator token is not, as well as when the user token is non-USD but the validator token is.

PR #1301, opened 2025-12-08
feat(consensus): make it very difficult to leak consensus secrets

SuperFluffy:
This makes it very hard to accidentally leak secrets by making the user jump through hoops before a secret is formatted as a `String`.

Related: https://github.com/commonwarexyz/monorepo/pull/2423


joshieDo (approved):
i feel like this is a bit too much .  having Display print xxxx & `Secret::reveal()` should suffice imo 😅

PR #1304, opened 2025-12-08
fix: set 2d nonces on estimateGas

klkvr:


crates/node/src/rpc/mod.rs
```diff
@@ -268,6 +269,31 @@ impl<N: FullNodeTypes<Types = TempoNode>> Call for TempoEthApi<N> {
...
+        evm_env: &EvmEnvFor<Self::Evm>,
+        mut request: TempoTransactionRequest,
+        mut db: impl Database<Error: Into<EthApiError>>,
+    ) -> Result<TxEnvFor<Self::Evm>, Self::Error> {
+        // if nonce key is provided, fetch nonce from nonce manager's storage.
+        if let Some(nonce_key) = request.nonce_key
+            && request.nonce.is_none()
+        {
+            let slot = double_mapping_slot(
+                request.from.unwrap_or_default().as_slice(),
```
mattsse, line 284: this is fine here, if the user specified a noncekey but no address then this will be nonce 0 (I guess we could always assume that, but is very rare so we can keep as is)

Issue #1033, opened 2025-11-24
Replay protection in Access Keys


legion2002:
### Describe the bug

Currently we don't have a nonce in access keys, so the following flow could be vulnerable to replay attacks - 
* Authorize Key A with expiry 100
* Revoke Key A at time 50
* At time 51 someone cold reauthorize the key.

We can't do this right now because of some complexity around how we would handle it in the 2d nonce pool. It is not sufficient to limit key authorizations to just the protocol nonce, because they are async by nature.

### Steps to reproduce

NA

### Logs

```text

```

### Container Type

Not running in a container

### What version/commit are you on?

https://github.com/tempoxyz/tempo/pull/598

### Code of Conduct

- [x] I agree to follow the Code of Conduct

PR #1307, opened 2025-12-08
chore: add fuzz tests & solidity impl for AccountKeychain precompile

legion2002:
The old tests didn't run on the rust implementation properly. 
This PR adds a correct implementation of the AccountKeychain precompile, and adds tests for it.

We need to bump the solidity version to 0.8.24 to be able to use the `transient` modifier to TSTORE/TLOAD the tx key automatically.
@zerosnacks if this is problematic on your end, can also bump this down to 0.8.13, and manage the transaction key slot manually.


docs/specs/src/AccountKeychain.sol
```diff
@@ -0,0 +1,293 @@
...
+        }
+
+        // Check if this key was previously revoked - prevents replay attacks
+        if (existingKey.isRevoked) {
+            revert KeyAlreadyRevoked();
+        }
+
+        // Convert SignatureType enum to u8 for storage
+        uint8 sigType;
+        if (signatureType == SignatureType.Secp256k1) {
```
howydev: hmm, feel like there should be some way to cast this to u8 directly like `uint8(signatureType)` since enums are uint8 under the hood

docs/specs/src/AccountKeychain.sol
```diff
@@ -0,0 +1,293 @@
...
+            revert ZeroPublicKey();
+        }
+
+        // Check if key already exists (key exists if expiry > 0)
+        AuthorizedKey storage existingKey = keys[msg.sender][keyId];
+        if (existingKey.expiry > 0) {
+            revert KeyAlreadyExists();
+        }
+
+        // Check if this key was previously revoked - prevents replay attacks
```
howydev, line 87: Hmm, what does a replay here look like? I'm probably missing something, I'm assuming that this is a the top level tx call to `authorizeKey`, but in that case that tx has a nonce which should prevent replay already

howydev: is there a ERC712-like mechanism (offchain sig) for enabling a key?

docs/specs/src/AccountKeychain.sol
```diff
@@ -0,0 +1,293 @@
...
+        uint64 expiry;
+        bool enforceLimits;
+        bool isRevoked;
+    }
+
+    /// @dev Mapping from account -> keyId -> AuthorizedKey
+    mapping(address => mapping(address => AuthorizedKey)) private keys;
+
+    /// @dev Mapping from keccak256(account || keyId) -> token -> spending limit
+    mapping(bytes32 => mapping(address => uint256)) private spendingLimits;
```
howydev, line 43: Nit: Prefer if we either skip the concat and have `address -> keyId -> token -> limit`, or concat all and do `keccak256(account || keyId || token) -> limit` instead

docs/specs/src/AccountKeychain.sol
```diff
@@ -0,0 +1,293 @@
...
+        _transactionKey = keyId;
+    }
+
+    /// @notice Internal function to verify and update spending for a token transfer
+    /// @dev This would be called by the protocol during TIP20 transfers to enforce spending limits
+    /// @param account The account performing the transfer
+    /// @param keyId The key ID that signed the transaction
+    /// @param token The token being transferred
+    /// @param amount The amount being transferred
+    function _verifyAndUpdateSpending(address account, address keyId, address token, uint256 amount)
```
howydev, line 247: How does this interact with transfer-like fns that are not transfer/approve/transferWithMemo?

E.g. `transferFeePreTx` from fee manager pulls funds from the account without an approval

Similarly, the DEX uses `systemTransferFrom` to pull funds from the account in a way that bypasses approvals

docs/specs/src/AccountKeychain.sol
```diff
@@ -0,0 +1,293 @@
...
+        // Check and update spending limit
+        bytes32 limitKey = _spendingLimitKey(account, keyId);
+        uint256 remaining = spendingLimits[limitKey][token];
+
+        if (amount > remaining) {
+            revert SpendingLimitExceeded();
+        }
+
+        // Update remaining limit
+        spendingLimits[limitKey][token] = remaining - amount;
```
howydev: nit: can wrap in `unchecked`

docs/specs/src/AccountKeychain.sol
legion2002: Yes there is a separate mechanism, it is called the KeyAuthorization, more context here - https://github.com/tempoxyz/tempo/blob/main/crates/primitives/src/transaction/key_authorization.rs

It does not include a nonce, because then we would have had to create a separate nonce space for just key auths.
Also note key auths are like precalls, so they are signed separately from the main tx, we cannot use main tx nonce for key auths.

So we use a mechanism , where the same key with the same expiry cannot be authorized again. And if a key has previously been revoked we cannot re authorize it. 

More context in this issue - 
https://github.com/tempoxyz/tempo/issues/1033

and this thread - 
https://tempoxyz.slack.com/archives/C09KCGR4LQ4/p1763992902077179?thread_ts=1763984972.406259&cid=C09KCGR4LQ4

But it would be worth to get your thoughts on this too from a security perspective, if you can think of any loopholes

docs/specs/src/AccountKeychain.sol
legion2002: We only include this for this `transfer`
https://github.com/tempoxyz/tempo/blob/f8c4c034563b8588d094df0058e3d00341b0099b/crates/precompiles/src/tip20/mod.rs#L582

And this `approve` fn right now
https://github.com/tempoxyz/tempo/blob/f8c4c034563b8588d094df0058e3d00341b0099b/crates/precompiles/src/tip20/mod.rs#L559

But this looks like a good point, we also need to add this for `transfer_with_memo` atleast. And some other functions. 
Can you think of any other functions where we need to add this?

Not sure how we should deal with the dex related entities though. will think more about it.

docs/specs/src/AccountKeychain.sol
legion2002: Think this is a good point, but it would require a hardfork to do this, so maybe not worth doing now.

docs/specs/src/AccountKeychain.sol
```diff
@@ -0,0 +1,286 @@
...
+    ) external {
+        // Check that the transaction key for this transaction is zero (main key)
+        _requireRootKey();
+
+        // Validate inputs
+        if (keyId == address(0)) {
+            revert ZeroPublicKey();
+        }
+
+        // Check if key already exists (key exists if expiry > 0)
```
howydev, line 81: nit: would be nice to check that `expiry` input is not `0`

PR #1321, opened 2025-12-08
fix(docs): add stablecoin exchange case to fee waterfall

danrobinson:
This updates the fee token spec to match the implementation, adding an already-implemented rule that—in certain cases—uses the `tokenIn` argument to swaps on the Stablecoin Exchange as the default fee token.

### DIFF
--- a/docs/pages/protocol/fees/spec-fee.mdx
+++ b/docs/pages/protocol/fees/spec-fee.mdx
@@ -70,12 +70,13 @@ When `feePayerSignature` is present:
 
 ## Fee token preferences
 
-The protocol checks for token preferences in four ways, with this order of precedence:
+The protocol checks for token preferences in five ways, with this order of precedence:
 
 1. Transaction (set by the `fee_token` field of the transaction)
 2. Account (set on the FeeManager contract by the `fee_payer` of the transaction)
-3. TIP-20 contract (if transaction is calling any function on a TIP-20 contract, the transaction uses that token as its fee token)
-4. PathUSD (as a fallback)
+3. TIP-20 contract (if the transaction is calling any function on a TIP-20 contract, the transaction uses that token as its fee token)
+4. Stablecoin Exchange (for certain swap calls, the transaction uses the `tokenIn` argument as its fee token)
+5. PathUSD (as a fallback)
 
 The protocol checks preferences at each of these levels, stopping at the first one at which a preference is specified. At that level, the protocol performs the following checks. If any of the checks fail, the transaction is invalid (without looking at any further levels):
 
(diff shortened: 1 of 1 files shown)


docs/pages/protocol/fees/spec-fee.mdx
```diff
@@ -70,12 +70,13 @@ When `feePayerSignature` is present:
 
 ## Fee token preferences
 
-The protocol checks for token preferences in four ways, with this order of precedence:
+The protocol checks for token preferences in five ways, with this order of precedence:
 
 1. Transaction (set by the `fee_token` field of the transaction)
 2. Account (set on the FeeManager contract by the `fee_payer` of the transaction)
```
0xKitsune, line 76: After checking if the fee token is specified in the tx envelope, we then check if the fee payer is the msg.sender and the transaction is calling FeeManager to set a new preference. If so, the newly set preference is used instead of the previously stored one.

Is this still the expected behavior? This might be a bit non intuitive if someone has set their fee token already and then is specifying a different token. Intuitively it feels if the fee token is set, the user should pay in that token until it changes. Let me know if you have any thoughts.


After this check, then we check if the fee token is set in the userTokens mapping.

docs/pages/protocol/fees/spec-fee.mdx
```diff
@@ -70,12 +70,13 @@ When `feePayerSignature` is present:
...
-3. TIP-20 contract (if transaction is calling any function on a TIP-20 contract, the transaction uses that token as its fee token)
-4. PathUSD (as a fallback)
+3. TIP-20 contract (if the transaction is calling any function on a TIP-20 contract, the transaction uses that token as its fee token)
+4. Stablecoin Exchange (for certain swap calls, the transaction uses the `tokenIn` argument as its fee token)
```
0xKitsune, line 78: Should we specify that this is only the case when users call `swapExactAmountIn` or `swapExactAmountOut`? I see that we specify this later but it might be nice to add this here.

docs/pages/protocol/fees/spec-fee.mdx
danrobinson: Hmm, it seems more natural to me to use the newly specified token. One case where it would come up is if the user runs out of a token, so they want to start using a new one. In your case, they would have to go get the old token in order to change it.

docs/pages/protocol/fees/spec-fee.mdx
danrobinson: Aren't those the only "swap" calls?

docs/pages/protocol/fees/spec-fee.mdx
0xKitsune: > Hmm, it seems more natural to me to use the newly specified token

Sounds good, this is currently the implemented behavior in the node. In any case we should probably update the fee token preference described to include this case.

docs/pages/protocol/fees/spec-fee.mdx
0xKitsune: Yes, just "for certain swap calls" makes it seem ambiguous as to what calls trigger this when we could just specify the actual calls. I dont have a strong opinion here though, happy to leave as is if you prefer.

PR #1338, opened 2025-12-09
feat: use index supply proxy

jxom:
- In dev, we can just use the public index supply endpoint (default `https://api.index-supply.net`)
- In prod, `VITE_INDEXSUPPLY_API_URL` is set to `https://p.tempo.xyz/index-supply`, which has CORS + rate limit restrictions, and obfuscates our API key.


Zygimantass:
@jxom do you want to move this to docs repo?

PR #1356, opened 2025-12-09
fix(xtask): initialize contracts with timestamp 0

klkvr:
Changes genesis generation tool to initialize contracts with timestamp 0.

Before, we were using timestamp 1 which was causing issues for deterministic runtime tests because the tip20 rewards logic was calculating `elapsed = 0 - 1`


joshieDo (approved):
nice

PR #1357, opened 2025-12-09
poc(precompiles): custom hashing strategies for mappings

0xrusowsky:
## Motivation

for performance reasons we want to explore diverging from solidity's default storage layout, and instead explore using custom hashing strategies such as blake3, or even direct byte computation.

## Technical Specification

we partition the 32-byte storage slot space by reserving the first byte as a namespace identifier. This creates 256 non-overlapping "storage spaces", each with 2^248 possible slots.

why partition?
- Standard keccak256 mappings produce uniformly distributed slots across the full 2^256 range
- Direct byte-encoded slots (for performance) would collide with hashed slots
- By masking the first byte, we guarantee collision-free coexistence of different slot computation strategies

### Storage Space Allocation

 | First Byte | Space      | Strategy       | Purpose |
 |------------|------------|----------------|---------|
 | 0x00       | 0 (Default)| hashed         | std and nested mappings; solidity-compatible|
 | 0x01       | 1          | direct address | direct address access (i.e. TIP20 balances)|
 | 0x02       | 2          | direct address | direct address access (i.e. allowances)|
 | ...        | ...        | ...            | direct address access|

each contract can have multiple `AddressMapping<V>` fields, which get uniquely auto-assigned space values at compile time, thanks to the `#[contract]` macro.

### Slot Computation Strategies

**`Keccak256`** (`SPACE = 0`)
```
slot = keccak256(left_pad(key, 32) || base_slot)
slot[0] = 0x00  // override first byte
```
 - Supports any key type (Address, B256, structs)
 - Supports arbitrary nesting (Mapping<K, Mapping<K2, V>>)

**`DirectAddressMap`** (`SPACE ≠ 0`)
```
slot = [SPACE (1 byte)][ADDRESS (20 bytes)][0x00 (11 bytes)]
```
 - Key type restricted to Address
 - No nesting allowed (enforced at compile-time). Prevents slot collision from different spaces.
 - ~3 gas (no hashing, direct byte encoding)

## Solution

this PR introduces a pluggable hashing strategy system for storage mappings, refactoring `Mapping<K, V>` into a generic `MappingInner<S: HashStrategy, V>` that can use different slot computation algorithms.

**New Traits:**
  - `HashStrategy`: strategy for computing mapping storage slots with associated `Key` type and `STORAGE_SPACE` identifier
  - `StorableInMapping<const SPACE: u8>`: marker trait enforcing nesting constraints between storage spaces

**New Types:**
- `MappingInner<S: HashStrategy, V>`: generic mapping parameterized by hash strategy
- `Keccak256<K>`: solidity-like hashing with storage space byte (`STORAGE_SPACE = 0`)
- `DirectBytes<const SPACE: u8>`: direct byte encoding for cheap address lookups

**Type Aliases:**
- `Mapping<K, V> = MappingInner<V, Keccak256<K>>`: old/regular mapping
- `UserMapping<V> = MappingInner<V, DirectBytes<1>>`: cheap user lookups for TIP20

**Storage Space Rules:**
the `trait StorableInMapping<SPACE>` enforces:
  1. Primitives/structs can be stored in any space
  2. `Keccak256` mappings (space 0) can be arbitrarily nested
  3. `DirectBytes` mappings (space ≠ 0) cannot be nested
  4. Different storage spaces cannot interact

the `#[contract]` macro validates that non-zero `STORAGE_SPACE` values are unique per contract, preventing slot collisions between `DirectBytes`-based mappings.


crates/precompiles/src/storage/types/mapping.rs
```diff
@@ -1,51 +1,114 @@
...
+        let key_bytes = key_bytes.as_ref();
+
+        // Pad key to nearest multiple of 32 bytes
+        let padded_len = key_bytes.len().div_ceil(32) * 32;
+        let mut buf = vec![0u8; padded_len + 32];
+
+        // Left-pad the key bytes
+        buf[padded_len - key_bytes.len()..padded_len].copy_from_slice(key_bytes);
+        // Append slot in big-endian
+        buf[padded_len..].copy_from_slice(&base_slot.to_be_bytes::<32>());
```
rakita, line 54: This looks strange as keccak256 function does not require padding, is this something that Solidity does?

We can probably just concatenate base_slot and key bytes

And last question, is it assumed that base_slot is always used?

crates/precompiles/src/storage/types/mapping.rs
```diff
@@ -1,51 +1,114 @@
...
+
+        // First byte is always reserved for the storage space
+        let mut hash = keccak256(&buf).0;
+        hash[0] = Self::STORAGE_SPACE;
```
rakita, line 58: nice!

crates/precompiles/src/storage/types/mapping.rs
```diff
@@ -1,51 +1,114 @@
...
+impl<const SPACE: u8> HashStrategy for DirectBytes<SPACE> {
+    type Key = Address;
+    const STORAGE_SPACE: u8 = SPACE;
+
+    #[inline]
+    fn compute_slot(key: &Address, _base_slot: U256) -> U256 {
+        let mut slot = [0u8; 32];
+        slot[0] = SPACE;
+        slot[1..21].copy_from_slice(key.as_slice());
+        U256::from_be_bytes(slot)
```
rakita: ~~key.into_word()~~ edit: cant use this as space byte needs to be set

crates/precompiles/src/storage/types/mapping.rs
```diff
@@ -1,51 +1,114 @@
...
+/// Slot computation: `[SPACE][address][zeros]`
 ///
-/// Mappings use Solidity's storage layout:
-/// - Base slot: stored in `base_slot` field (never accessed directly)
-/// - Actual slot for key `k`: `keccak256(k || base_slot)`
+/// The const generic `SPACE` parameter allows multiple `DirectBytes`-based mappings
+/// to coexist without slot collisions.
+pub struct DirectBytes<const SPACE: u8>;
+
+impl<const SPACE: u8> HashStrategy for DirectBytes<SPACE> {
```
rakita: Maybe renaming to something like`DirectAddressAccess`, I am bad with names but this mapping works only with address

crates/precompiles/src/tip20/mod.rs
```diff
@@ -77,7 +77,11 @@ pub struct TIP20Token {
 
     // TIP20 Token
     total_supply: U256,
-    balances: Mapping<Address, U256>,
+
+    // Migrated to `users` to leverage `UserMapping` and avoid hashing. Used to be slot 9.
+    // balances: Mapping<Address, U256>,
+    users: UserMapping<UserInfo>,
```
rakita: `address -> UserInfo` is not possible to map directly as UserInfo is an complex type, am I missing something here?

I would assume that we will have:
* `balances`: UserMapping<U256>, SPACE = 1
* `nonces`: UserMapping<U256>, SPACE = 2


Maybe add:
* `scheduled_rate_decrease`: would require direct mapping with u128.
* `role_admins` it is mapped as B256->B256, could evaluate if u128 would work here. Just to make sure if B256 is used for some specific reason or not 

And would evaluate if `user_rewards_info` can be split into multiple mappings so it can be directly indexed. If not, it is fine to skip it.

rakita: `role_admins` looks like it will be a big lift both changes to [solidity](https://github.com/tempoxyz/tempo/blob/main/docs/specs/src/interfaces/ITIP20RolesAuth.sol) and code, so would skip it. Maybe [same](https://github.com/tempoxyz/tempo/blob/f4daf5aa5ecac0e6934c724747e6cd9eb1ad9fcc/docs/specs/src/TIP20.sol#L86) with `user_rewards_info`

crates/precompiles/src/storage/types/mapping.rs
0xrusowsky: i constrained `type Key = Address`, but i agree that ur name suggestion is way better 🤝

PR #1388, opened 2025-12-09
Faucet page: remember wallet reconnection, show faucet receipts

snario:
- Fixed wallet connection to properly reconnect on page reload and show specific wallet connectors instead of generic 'Injected' option
- Enhanced fee token step to show completion state when token is set and display current token as clickable explorer link
- Added transaction receipt links after faucet funding with balance animation and clickable token names in footer

### DIFF
--- a/docs/components/guides/steps/wallet/SetFeeToken.tsx
+++ b/docs/components/guides/steps/wallet/SetFeeToken.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react'
 import { type Address, isAddress } from 'viem'
-import { useChainId, useConfig, useConnection } from 'wagmi'
+import { useConfig, useConnection } from 'wagmi'
 import { Hooks } from 'wagmi/tempo'
 import { Button, ExplorerLink, Step, StringFormatter } from '../../Demo'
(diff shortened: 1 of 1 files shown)


docs/wagmi.config.ts
```diff
@@ -29,6 +30,7 @@ export function getConfig(options: getConfig.Options = {}) {
         grantAccessKey: true,
         keyManager: KeyManager.localStorage(),
       }),
+      ...(multiInjectedProviderDiscovery ? [injected()] : []),
```
tmm, line 35: when `multiInjectedProviderDiscovery: true`, we shouldn't need `injected`. what issue are you seeing?

snario: The issue was that the injected wallet was not remembered from last time. Go to the browser, connect MetaMask, succeed, reload page, it is not still connected.

tmm: Interesting. It should be. Can take a look.

tmm: Fixed reconnect in https://github.com/tempoxyz/tempo/pull/1492

PR #1393, opened 2025-12-10
[WIP] feat: deterministic tempo.ts sdk documentation

gorried:


gorried:
Threw this up to do a comparison against prod docs. As it stands there are too many small discrepancies that are being successfully smoothed over by the LLM based sdk docs generation so going to abandon this in favor of higher priority work. 

Most importantly, these include missing parameters in the ts type extraction, which is a non-starter

Issue #1346, opened 2025-12-09
Arb bot starts spamming `eth_call` when runs out of funds


shekhirin:
### Describe the feature

<img width="1263" height="305" alt="Image" src="https://github.com/user-attachments/assets/70733efc-aa5f-4d7f-99ce-53d510194840" />

<img width="2163" height="999" alt="Image" src="https://github.com/user-attachments/assets/baa6b276-e017-4341-8495-22a47ddb6c8f" />

PR #1408, opened 2025-12-10
fix(arbbot): skip swaps for insufficent balance tokens

xwid138:
* Mutes certain token swaps for configured time (default 5 minutes).
* Adds `--token` cli argument, can be provided multiple times to specify from which tokens build pairs (permutation) and rebalance. If none provided then uses all tokens based on factory contract.

fixes: #1346


Zygimantass:
Hi, we don't accept contributions from non-core contributors that aren't assigned to a specific issue. Next time, please check that there's an issue related to your contribution before submitting it:)

PR #1413, opened 2025-12-10
feat: add gas schedule for access keys 

legion2002:
Currently setting an access key using key authorization and setting spending limits is basically free in the protocol.
Since every authorized key and every spending limit requires 1 SSTORE, this makes the node potentially vuln to DOS.

In this PR, we update the spec, and add extra costs for key authorization. The rationale, and exact gas changes can be seen in the docs. 

We also add better `eth_estimateGas` support for the access key type.

### DIFF
--- a/crates/node/tests/assets/test-genesis.json
+++ b/crates/node/tests/assets/test-genesis.json
@@ -20,6 +20,7 @@
     "terminalTotalDifficulty": 0,
     "terminalTotalDifficultyPassed": true,
     "adagioTime": 0,
+    "allegrettoTime": 0,
     "epochLength": 302400,
     "moderatoTime": 0,
     "depositContractAddress": "0x00000000219ab540356cbb839cbe05303d7705fa"
--- a/crates/revm/src/common.rs
+++ b/crates/revm/src/common.rs
@@ -18,10 +18,23 @@ use tempo_precompiles::{
     error::{Result as TempoResult, TempoPrecompileError},
     storage::{Handler, PrecompileStorageProvider, StorageCtx},
     tip_fee_manager::TipFeeManager,
-    tip20::{self, TIP20Token, is_tip20_prefix},
+    tip20::{self, ITIP20, TIP20Token, is_tip20_prefix},
     tip403_registry::TIP403Registry,
 };
 use tempo_primitives::TempoTxEnvelope;
+
+/// Returns true if the calldata is for a TIP-20 function that should trigger fee token inference.
+/// Only `transfer`, `transferWithMemo`, and `startReward` qualify.
+fn is_tip20_fee_inference_call(input: &[u8]) -> bool {
+    input.first_chunk::<4>().is_some_and(|&s| {
+        matches!(
+            s,
+            ITIP20::transferCall::SELECTOR
+                | ITIP20::transferWithMemoCall::SELECTOR
+                | ITIP20::startRewardCall::SELECTOR
+        )
+    })
+}
 
 /// Helper trait to abstract over different representations of Tempo transactions.
 #[auto_impl::auto_impl(&)]
@@ -140,12 +153,26 @@ pub trait TempoStateAccess<M = ()> {
             return Ok(user_token);
         }
 
-        // If tx.to() is a TIP-20 token, use that token as the fee token
-        if let Some(to) = tx.calls().next().and_then(|(kind, _)| kind.to().copied())
-            && tx.calls().all(|(kind, _)| kind.to() == Some(&to))
-            && self.is_valid_fee_token(to, spec)?
-        {
-            return Ok(to);
+        // Check if the fee can be inferred from the TIP20 token being called
+        if let Some(to) = tx.calls().next().and_then(|(kind, _)| kind.to().copied()) {
+            let can_infer_tip20 = if spec.is_allegro_moderato() {
+                // Post-AllegroModerato: restricted to transfer/transferWithMemo/startReward,
+                // and for AA txs only when fee_payer == tx.origin.
+                if tx.is_aa() && fee_payer != tx.caller() {
+                    false
+                } else {
+                    tx.calls().all(|(kind, input)| {
+                        kind.to() == Some(&to) && is_tip20_fee_inference_call(input)
+                    })
+                }
+            } else {
+                // Pre-AllegroModerato: all calls must target the same address
+                tx.calls().all(|(kind, _)| kind.to() == Some(&to))
+            };
+
+            if can_infer_tip20 && self.is_valid_fee_token(to, spec)? {
+                return Ok(to);
+            }
         }
 
         // If calling swapExactAmountOut() or swapExactAmountIn() on the Stablecoin Exchange,
@@ -408,6 +435,7 @@ where
 #[cfg(test)]
 mod tests {
     use super::*;
+    use alloy_primitives::uint;
     use revm::{context::TxEnv, database::EmptyDB, interpreter::instructions::utility::IntoU256};
     use tempo_precompiles::tip20;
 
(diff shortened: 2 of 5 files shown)


crates/alloy/src/rpc/request.rs
```diff
@@ -52,13 +52,26 @@ pub struct TempoTransactionRequest {
     /// Required when key_type is WebAuthn to calculate calldata gas costs.
     pub key_data: Option<Bytes>,
 
+    /// Whether the signature is a Keychain wrapper (adds 3,000 gas overhead for key validation).
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub is_keychain: Option<bool>,
```
klkvr: this could just be a bool, `None == false` iiuc

crates/alloy/src/rpc/request.rs
```diff
@@ -52,13 +52,26 @@ pub struct TempoTransactionRequest {
...
     pub tempo_authorization_list: Vec<TempoSignedAuthorization>,
+
+    /// Key authorization signature type for gas estimation.
+    /// Required when transaction provisions an access key.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub key_auth_signature_type: Option<SignatureType>,
+
+    /// Number of spending limits in key authorization (for gas estimation).
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub key_auth_num_limits: Option<u64>,
```
klkvr: should this just take a signed key auth item? this is already the case with `TempoSignedAuthorization` and regular 7702 auth lists

crates/revm/src/handler.rs
```diff
@@ -618,8 +665,10 @@ where
         // Note: Transaction parameter validation (priority fee, time window) happens in validate_env()
 
         // If the transaction includes a KeyAuthorization, validate and authorize the key
+        // Skip validation during gas estimation (signature_hash is zero for mock signatures)
         if let Some(tempo_tx_env) = tx.tempo_tx_env.as_ref()
             && let Some(key_auth) = &tempo_tx_env.key_authorization
+            && tempo_tx_env.signature_hash != alloy_primitives::B256::ZERO
```
klkvr: I guess if we require a complete key auth on RPC side this hack would also not be needed?

crates/alloy/src/rpc/request.rs
legion2002: Hmm not sure if we already have the Key auth item, by the time we are validating the sig.
We probably should. @jxom any thoughts?

legion2002: Checked with jake, and converting it to signed key auths now.

crates/node/tests/it/tempo_transaction.rs
```diff
@@ -2662,6 +2701,166 @@ async fn test_aa_estimate_gas_with_key_types() -> eyre::Result<()> {
...
+    println!("  ✓ Keychain P256 adds {keychain_p256_diff} gas (expected ~8,000)");
+
+    // Test 4: KeyAuthorization with secp256k1 (no limits)
+    println!("\nTest 4: KeyAuthorization (secp256k1, no limits)");
+    let key_auth_secp = create_signed_key_authorization(&signer, SignatureType::Secp256k1, 0);
+    let mut tx_key_auth = tx_request.clone();
+    tx_key_auth.as_object_mut().unwrap().insert(
+        "keyAuthorization".to_string(),
+        serde_json::to_value(&key_auth_secp)?,
+    );
```
0xKitsune: Do we need to do all of this manually or can we use the `TempoTransactionRequest` type directly?

crates/alloy/src/rpc/request.rs
```diff
@@ -52,13 +52,22 @@ pub struct TempoTransactionRequest {
     /// Required when key_type is WebAuthn to calculate calldata gas costs.
     pub key_data: Option<Bytes>,
 
+    /// Whether the signature is a Keychain wrapper (adds 3,000 gas overhead for key validation).
+    #[serde(default, skip_serializing_if = "std::ops::Not::not")]
+    pub is_keychain: bool,
```
0xKitsune: Is there any way to infer this rather than having the user set this directly? It seems like it could be better UX if the protcol could infer.

crates/revm/src/handler.rs
```diff
@@ -56,6 +56,18 @@ use crate::{
 /// P256 precompile cost (6900 from EIP-7951) + 1100 for 129 bytes extra signature size - ecrecover savings (3000)
 const P256_VERIFY_GAS: u64 = 5_000;
 
+/// Gas cost for ecrecover signature verification (used by KeyAuthorization)
+const ECRECOVER_GAS: u64 = 3_000;
+
+/// Additional gas for Keychain signatures (key validation overhead: 2,100 SLOAD + 900 processing)
+const KEYCHAIN_VALIDATION_GAS: u64 = 3_000;
```
0xKitsune: nit: consider using gas constants from revm where applicable to be explicit.

For example:
```
const KEYCHAIN_VALIDATION_GAS: u64 = COLD_SLOAD_COST + 900;
```

crates/alloy/src/rpc/request.rs
legion2002: Don't think there is another easy way to infer this right now, because we don't actually get the signature from the user during gas estimation.
We get some details, like key type, key data etc. and then we construct a mock signature internally.
So the user has to explicitly inform us if this key will be used as part of a keychain.

crates/node/tests/it/tempo_transaction.rs
legion2002: yup done

crates/alloy/src/rpc/compat.rs
```diff
@@ -106,14 +114,24 @@ impl TryIntoTxEnv<TempoTxEnv, TempoBlockEnv> for TempoTransactionRequest {
...
+                        create_mock_tempo_signature(kt, key_data.as_ref(), is_keychain, caller_addr)
+                    })
                     .unwrap_or_else(|| {
-                        create_mock_tempo_signature(&SignatureType::Secp256k1, None)
+                        create_mock_tempo_signature(
+                            &SignatureType::Secp256k1,
+                            None,
+                            is_keychain,
+                            caller_addr,
+                        )
```
klkvr: hmm is this just `create_mock_tempo_signature(&key_type.unwrap_or(SignatureType::Secp256k1), ...)`?

crates/revm/src/handler.rs
```diff
@@ -753,8 +801,10 @@ where
 
         // For Keychain signatures, validate that the keychain is authorized in the precompile
         // UNLESS this transaction also includes a KeyAuthorization (same-tx auth+use case)
+        // Skip validation during gas estimation (signature_hash is zero for mock signatures)
         if let Some(tempo_tx_env) = tx.tempo_tx_env.as_ref()
             && let Some(keychain_sig) = tempo_tx_env.signature.as_keychain()
+            && tempo_tx_env.signature_hash != alloy_primitives::B256::ZERO
```
klkvr: hmm so `set_transaction_key` is unreachable in `eth_call`/`eth_estimateGas`? would that mean that logic for spend permissions will never be entered resulting in potentially wrong gas usage?

crates/revm/src/handler.rs
```diff
@@ -1081,20 +1131,23 @@ where
...
+/// - Key authorization costs (if present, post-AllegroModerato): 30k/32k base + 22k per spending limit
 /// - Floor gas calculation (EIP-7623, Prague+)
 fn calculate_aa_batch_intrinsic_gas<'a>(
     calls: &[tempo_primitives::transaction::Call],
     signature: &TempoSignature,
     access_list: Option<impl Iterator<Item = &'a AccessListItem>>,
     authorization_list: &[RecoveredTempoAuthorization],
+    key_authorization: Option<&tempo_primitives::transaction::SignedKeyAuthorization>,
+    spec: tempo_chainspec::hardfork::TempoHardfork,
 ) -> Result<InitialAndFloorGas, TempoInvalidTransaction> {
```
klkvr, line 1218: this could take entire tx env

crates/alloy/src/rpc/compat.rs
```diff
@@ -151,37 +161,61 @@ impl TryIntoTxEnv<TempoTxEnv, TempoBlockEnv> for TempoTransactionRequest {
...
     key_type: &SignatureType,
     key_data: Option<&Bytes>,
+    key_id: Option<Address>,
+    caller_addr: alloy_primitives::Address,
 ) -> TempoSignature {
+    use tempo_primitives::transaction::tt_signature::{KeychainSignature, TempoSignature};
+
+    let inner_sig = create_mock_primitive_signature(key_type, key_data.cloned());
+
+    if let Some(_key_id) = key_id {
```
klkvr:
```suggestion
    if key_id.is_some() {
```

crates/revm/src/handler.rs
```diff
@@ -751,27 +860,42 @@ where
...
+            let access_key_addr = if is_gas_estimation {
+                // For gas estimation, use the key_id provided in the request
+                // This allows proper spending limits simulation
+                tempo_tx_env.key_id.ok_or_else(|| {
+                    EVMError::Transaction(TempoInvalidTransaction::AccessKeyAuthorizationFailed {
+                        reason:
+                            "key_id must be provided for gas estimation with Keychain signatures"
+                                .to_string(),
+                    })
+                })?
```
klkvr: could we just remove the zero signature hash hack and have `is_gas_estimation = key_id.is_some()`? also I'd rather rename `key_id` to something like `override_key_id`/`fake_key_id`/`unchecked_key_id` so that it's obvious that this is not something that should be set usually

crates/revm/src/handler.rs
```diff
@@ -793,9 +918,11 @@ where
 
             // Always need to set the transaction key for Keychain signatures
             StorageCtx::enter_precompile(journal, block, cfg, |mut keychain: AccountKeychain| {
-                if !is_authorizing_this_key {
+                // Skip keychain validation during gas estimation or when authorizing this key
+                if !is_gas_estimation && !is_authorizing_this_key {
```
klkvr: do we need to skip this check in estimation? seems correct to reject estimations when user provided mismatched `keyAuth` and `keyId`

klkvr (approved):
this looks good to me, just final nits

crates/precompiles/src/nonce/mod.rs
```diff
@@ -98,7 +107,7 @@ impl NonceManager {
 
         self.active_key_count.at(account).write(new_count)?;
 
-        // Emit ActiveKeyCountChanged event (only after Moderato hardfork)
+        // Emit ActiveKeyCountChanged event (only between Moderato and AllegroModerato)
         if self.storage.spec().is_moderato() {
```
0xKitsune: nit: This comment mentions that the event is only emitted between moderato and allegro moderato but this only checks if the spec is post moderato.

0xKitsune (approved):

Issue #1329, opened 2025-12-08
Don't do DEX setup in tempo-bench if `--tip20-weight 1`


shekhirin:
### Describe the feature

As per the issue title

PR #1427, opened 2025-12-10
feat(tempo-bench): dex setup not needed in case `tip20_weight` is 1

shubhexists:
Solves #1329

### DIFF
--- a/bin/tempo-bench/src/cmd/max_tps.rs
+++ b/bin/tempo-bench/src/cmd/max_tps.rs
@@ -12,6 +12,7 @@ use tempo_alloy::{
     TempoNetwork, primitives::TempoTxEnvelope, provider::ext::TempoProviderBuilderExt,
 };
 
+use crate::cmd::{max_tps::dex::create_tokens_for_bench, signer_providers::SignerProviderManager};
 use alloy::{
     consensus::BlockHeader,
     eips::Encodable2718,
@@ -70,8 +71,6 @@ use tokio::{
     time::{Sleep, interval, sleep},
 };
 use tokio_util::sync::CancellationToken;
-
-use crate::cmd::signer_providers::SignerProviderManager;
 
 /// Run maximum TPS throughput benchmarking
 #[derive(Parser, Debug)]
@@ -250,8 +249,8 @@ impl MaxTpsArgs {
             }
         }
 
-        // Grab first provider to call some RPC methods
-        let provider = signer_providers[0].1.clone();
+        // Grab first provider and signer to call some RPC methods
+        let (first_signer, provider) = signer_providers[0].clone();
 
         // Fund accounts from faucet if requested
         if self.faucet {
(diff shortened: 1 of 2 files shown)


bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -557,8 +565,9 @@ async fn generate_transactions<F: TxFiller<TempoNetwork> + 'static>(
                     );
 
                     // Swap minimum possible amount
+                    // unwrap is safe here as if a swap txn is needed, there would definitely be a quote token
                     exchange
-                        .quoteSwapExactAmountIn(token, quote_token, 1)
+                        .quoteSwapExactAmountIn(token, quote_token.unwrap(), 1)
```
shekhirin: let's do an `expect()` with a proper message here

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -280,16 +280,24 @@ impl MaxTpsArgs {
...
-        let user_tokens = 2;
-        info!(user_tokens, "Setting up DEX");
-        let (quote_token, user_tokens) = dex::setup(
-            signer_providers,
-            user_tokens,
-            self.max_concurrent_requests,
-            self.max_concurrent_transactions,
-        )
-        .await?;
+        let mut quote_token = None;
```
shekhirin: we can avoid the reassignment and two letter variables here by doing `let (quote_token, user_tokens) = if { ... } else { (None, Vec::new()) }`

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -280,16 +280,24 @@ impl MaxTpsArgs {
...
+        let mut user_tokens: Vec<Address> = Vec::new();
+
+        if self.tip20_weight != 1.0 {
```
shekhirin: thinking more about it, let's check swap and order weights instead, and if any of them is `> 0`, then do the setup

just in case we add another weight argument unrelated to DEX transactions

shekhirin (changes requested):
Changes above

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -280,16 +280,20 @@ impl MaxTpsArgs {
...
+        let (quote_token, user_tokens) = if self.place_order_weight > 0.0 && self.swap_weight > 0.0
+        {
+            info!(user_tokens, "Setting up DEX");
+            dex::setup(
+                signer_providers,
+                user_tokens,
+                self.max_concurrent_requests,
+                self.max_concurrent_transactions,
+            )
+            .await?
```
shekhirin: Oh, this is not really correct to do, because `dex::setup` also sets up the TIP-20 tokens that we use for transfers. What we should do instead, is **always** setup the TIP-20 tokens, but the DEX setup should be done only if place order or swap weights are > 0.

Does this make sense?

shubhexists: Yes makes sense. Missed it. 2 ways to do this - 
1) I can add conditions in the `dex::setup` function and remove the conditional logic completely from the `run_with_manager`. Only issue is the function name might not be aligned with what it does.
2) I can extract the logic in some `create_tokens_for_bench` and then call it directly in the `run_with_manager`. Then pass all the token structs as params to `dex::setup`.

Any suggestions which could be a better approach?

shekhirin: Let's proceed with option 2, we need to set up TIP-20 tokens no matter what, so it makes sense to move it to a separate function!

PR #1432, opened 2025-12-10
feat(tempo-bench): ERC-20 tokens

### LINKED ISSUE #1428: Add ERC-20 transfers in `tempo-bench`
### Describe the feature

Same as TIP-20 transfers, but need to deploy and mint ERC-20 tokens and send them to random addresses

### Additional context

_No response_

shekhirin:
Closes #1428


bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -291,11 +296,26 @@ impl MaxTpsArgs {
...
+        let erc20_tokens = if self.erc20_weight > 0.0 {
+            let num_erc20_tokens = 2;
+            info!(num_erc20_tokens, "Setting up ERC-20 tokens");
+            erc20::setup(
+                signer_providers,
+                num_erc20_tokens,
+                self.max_concurrent_requests,
+                self.max_concurrent_transactions,
+            )
+            .await?
```
Rjected: why are we setting up 2 ERC20 tokens? why not just 1

bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -512,16 +536,17 @@ async fn generate_transactions<F: TxFiller<TempoNetwork> + 'static>(
 
     info!(transactions = total_txs, "Generating transactions");
 
-    const TX_TYPES: usize = 3;
+    const TX_TYPES: usize = 4;
     // Weights for random sampling for each transaction type
-    let tx_weights: [u64; TX_TYPES] = [tip20_weight, swap_weight, place_order_weight];
+    let tx_weights: [u64; TX_TYPES] = [tip20_weight, swap_weight, place_order_weight, erc20_weight];
```
Rjected, line 541: I think a good followup would be to separate out the selection logic, for each type, maybe with an enum? especially if we add more scenarios we would have to keep extending this match, `tx_weights` etc

Rjected (approved):
just one nit about the number of tokens we use in init, otherwise lgtm

bin/tempo-bench/src/cmd/max_tps.rs
shekhirin: yup https://github.com/tempoxyz/tempo/blob/efd8111fdf4e2464988076071e322d7a9f0f30ce/bin/tempo-bench/src/cmd/max_tps.rs#L559

bin/tempo-bench/src/cmd/max_tps.rs
shekhirin: yeah, no reason, changed to 1

PR #1441, opened 2025-12-11
fix: set system tx gas for `inspect_one_system_call_with_caller`

klkvr:


crates/revm/src/exec.rs
```diff
@@ -117,9 +117,9 @@ where
         system_contract_address: Address,
         data: Bytes,
     ) -> Result<Self::ExecutionResult, Self::Error> {
-        self.inner
-            .ctx
-            .set_tx(TxEnv::new_system_tx_with_caller(caller, system_contract_address, data).into());
+        let mut tx = TxEnv::new_system_tx_with_caller(caller, system_contract_address, data);
+        tx.set_gas_limit(250_000_000);
```
mattsse: let's add a constant for this instead, so that we can update all usages of this?

PR #1442, opened 2025-12-11
fix(revm): 2d nonce gas schedule

legion2002:
Adds a gas schedule for 2d nonces as written in the spec.
Without this the node can be susceptible to storage DOS vectors.

We also do a minor update in the gas schedule mentioned in the specs, to start the incrementing nonce costs from `20,000` instead of `0`

### DIFF
--- a/crates/node/tests/it/tempo_transaction.rs
+++ b/crates/node/tests/it/tempo_transaction.rs
@@ -2650,7 +2650,7 @@ async fn test_aa_estimate_gas_with_key_types() -> eyre::Result<()> {
     let baseline_gas: String = provider
         .raw_request(
             "eth_estimateGas".into(),
-            [serde_json::to_value(&base_tx_request())?],
+            [serde_json::to_value(base_tx_request())?],
         )
         .await?;
     let baseline_gas_u64 = u64::from_str_radix(baseline_gas.trim_start_matches("0x"), 16)?;
@@ -2748,7 +2748,7 @@ async fn test_aa_estimate_gas_with_keychain_and_key_auth() -> eyre::Result<()> {
     let baseline_gas: String = provider
         .raw_request(
             "eth_estimateGas".into(),
-            [serde_json::to_value(&base_tx_request())?],
+            [serde_json::to_value(base_tx_request())?],
         )
         .await?;
     let baseline_gas_u64 = u64::from_str_radix(baseline_gas.trim_start_matches("0x"), 16)?;


crates/revm/src/handler.rs
```diff
@@ -69,6 +72,12 @@ const KEY_AUTH_BASE_GAS: u64 = 27_000;
 /// Gas per spending limit in KeyAuthorization
 const KEY_AUTH_PER_LIMIT_GAS: u64 = 22_000;
 
+/// Gas cost for using an existing 2D nonce key (cold SSTORE on non-zero slot: 2,900 base + 2,100 cold access)
+const EXISTING_NONCE_KEY_GAS: u64 = 5_000;
```
klkvr: let's use revm constants explicitly here? it's `COLD_SLOAD_COST + WARM_SSTORE_RESET`

crates/revm/src/handler.rs
```diff
@@ -141,6 +150,43 @@ fn calculate_key_authorization_gas(
...
+
+    if current_nonce > 0 {
+        // Existing key - fixed cost
+        Ok(EXISTING_NONCE_KEY_GAS)
+    } else {
+        // New key - progressive cost based on number of active keys
+        let active_count = nonce_manager
+            .get_active_nonce_key_count(getActiveNonceKeyCountCall { account: caller })?;
+        Ok((active_count.saturating_to::<u64>() + 1) * NEW_NONCE_KEY_MULTIPLIER)
+    }
```
klkvr: hmm that seems super expensive and kinda breaks the usecase of using random nonce keys to not care about your nonces

after all it's just a cold sstore to non-existent slot which costs `22100` usually, why does it have to scale with number of active keys?

crates/revm/src/handler.rs
```diff
@@ -447,14 +493,20 @@ where
         evm: &mut Self::Evm,
         init_and_floor_gas: &InitialAndFloorGas,
     ) -> Result<FrameResult, Self::Error> {
+        // Add 2D nonce gas to the initial gas
+        let adjusted_gas = InitialAndFloorGas::new(
+            init_and_floor_gas.initial_gas + evm.nonce_2d_gas,
+            init_and_floor_gas.floor_gas,
+        );
```
klkvr, line 492: cc @rakita `validate_initial_tx_gas` should take `&mut Evm` so that we don't have to do this hack

crates/revm/src/handler.rs
legion2002: Initially when we designed this dan and georgios were concerned about state bloat. We were discussing adding garbage collection and state rent, but unlike other state 2d nonces can never be garbage collected because it would lead to tx replay attacks.

So we wanted to limit the number of nonce keys a user could create.

legion2002: I still think some of those concerns exist. But since we are relaxing our requirements for state bloat solutions, we can probably also relax them for 2d nonces.
This comes with the added benefit that we don't have to keep an active nonce key count.

legion2002: I have updated this to remove this linear gas cost for now, as I think the perf benefits of not having to keep track of the active nonce key count, alone makes it worth it.
When we update the new state creation gas costs, we can update this accordingly

PR #1451, opened 2025-12-11
perf: avoid cloning txs vec

klkvr:


crates/node/src/engine.rs
```diff
@@ -24,10 +26,19 @@ impl PayloadValidator<TempoPayloadTypes> for TempoEngineValidator {
...
+            header,
+            transactions,
+            withdrawals,
             validator_set: _,
         } = payload;
-        Ok(block)
+        Ok(SealedBlock::from_sealed_parts(
+            header,
+            BlockBody {
+                transactions: transactions.into_iter().map(Arc::unwrap_or_clone).collect(),
```
mattsse: this should be okay here,

but usure about the collect, because isn't this more or less effectively the same as the clone?

mattsse (approved):

PR #1456, opened 2025-12-11
feat(POC): add pathUSD token controller

legion2002:
POC implementation for a token controller, which allows you to multiplex a single stablecoin into multiple stablecoins on the chain using a simple wrapping mechanism.

Will primarily be used by Bridge for Path USD.

We add the wrapped token and the reserve token as immutables for efficiency. Which means you need to deploy a controller for each new token you want to create.
There is also a way to do a singleton controller here, which can be reused for multiple tokens, but comes at the cost of 1 extra storage operation per read. ( this might be totally fine, since mint/burn are not critical path functions for transfers )

### DIFF
--- a/docs/specs/src/ReserveStore.sol
+++ b/docs/specs/src/ReserveStore.sol
@@ -1,0 +1,32 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.13;
+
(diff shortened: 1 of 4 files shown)


legion2002:
PR has been moved to this bridge repo -

https://github.com/withbridge/erc20-stablecoin-template/pull/9

PR #1457, opened 2025-12-11
deps: bump reth

shekhirin:
Want to be triple-checked here, as it's affecting `extra_data` field.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,10 +3,25 @@
 version = 4
 
 [[package]]
+name = "addr2line"
+version = "0.25.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


crates/evm/src/context.rs
```diff
@@ -64,7 +60,6 @@ impl reth_rpc_eth_api::helpers::pending_block::BuildPendingEnv<tempo_primitives:
             general_gas_limit,
             shared_gas_limit,
             timestamp_millis_part: parent.timestamp_millis_part,
-            extra_data: Bytes::default(),
```
shekhirin, line 67: @klkvr is it ok to not set it to default here?

klkvr: yeah it's fine, this is just the pending rpc block and it will be valid anyway

klkvr (approved):
lgtm

PR #1464, opened 2025-12-12
perf(builder): sender recovery cache

### LINKED ISSUE #1458: Sender recovery cache
### Describe the feature

We recover senders for all transactions that get to tx pool, we can reuse these recovered senders in newPayload processing

### Additional context

_No response_

shekhirin:
**NOTE: needs benchmarking**

Closes #1458

This PR introduces a cache for recovered senders.

The flow is the following:
1. Validated transaction is inserted into tx pool. Recovered sender is inserted into the cache.
2. `newPayload` is received, senders cache is used when we're iterating over the transactions in the payload. On miss, recover sender in-place.
3. Mined transactions are evicted from tx pool. Recovered sender is removed from the cache.

If a transaction from tx pool is not included in any of the blocks, it will eventually get evicted by tx pool itself.

The size of the cache is controlled by the maximum size of tx pool.


legion2002 (commented):
Not sure if this happens somewhere already, but should we also cache the precompile storage access that we do during validation. 
Mainly 2d nonces and auth keys.

Issue #881, opened 2025-11-12
improvement(precompiles): event emitted during minting new liquidity should include `to`


howydev:
### Describe the bug

The `mint` event currently includes the relevant token amount but not the recipient of liq tokens: https://github.com/tempoxyz/docs/blob/e184b79cb44642c0bda8cc16ddaf40b803e5ee8a/specs/src/FeeAMM.sol#L139-L141

Credit: @samczsun: https://tempoxyz.slack.com/archives/C09KCGR4LQ4/p1762941597467839?thread_ts=1762896791.409559&cid=C09KCGR4LQ4

### Steps to reproduce

n/a

### Logs

```text

```

### Platform(s)

Mac (Apple Silicon)

### Container Type

Other

### What version/commit are you on?

n/a

### Code of Conduct

- [x] I agree to follow the Code of Conduct

PR #1466, opened 2025-12-12
feat: add new MintWithValidatorToken event with  field

howydev:
Add a new `MintWithValidatorToken` event that includes `to` field, the recipient of the liq tokens

Also removes `amountUserToken` since that's always 0 in the new `Mint` we have

Fixes: https://github.com/tempoxyz/tempo/issues/881

### DIFF
--- a/crates/precompiles/src/tip_fee_manager/amm.rs
+++ b/crates/precompiles/src/tip_fee_manager/amm.rs
@@ -278,16 +278,14 @@ impl TipFeeManager {
         )?;
 
         // Emit Mint event
-        self.emit_event(TIPFeeAMMEvent::Mint(
-            ITIPFeeAMM::Mint {
-                sender: msg_sender,
-                to,
-                userToken: user_token,
-                validatorToken: validator_token,
-                amountValidatorToken: amount_validator_token,
-                liquidity,
-            },
-        ))?;
+        self.emit_event(TIPFeeAMMEvent::Mint(ITIPFeeAMM::Mint {
+            sender: msg_sender,
+            to,
+            userToken: user_token,
+            validatorToken: validator_token,
+            amountValidatorToken: amount_validator_token,
+            liquidity,
+        }))?;
 
         Ok(liquidity)
     }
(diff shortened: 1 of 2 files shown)


docs/specs/src/interfaces/IFeeAMM.sol
```diff
@@ -25,11 +25,11 @@ interface IFeeAMM {
         uint256 amountIn,
         uint256 amountOut
     );
-    event Mint(
-        address indexed sender,
+    event MintWithValidatorToken(
```
0xKitsune: Do we want to introduce a new `Mint` event rather than `MintWithValidatorToken`? 

`mintWithValidatorToken` was introduced to update the minting logic to only supply the validator token in the feeAMM but this is a bit of debt since `mint` is now deprecated. Before mainnet it might make sense update `mintWithValdiatorToken` to `mint` and use a `Mint` event instead. Let me know if you have any thoughts.

0xKitsune (approved):

docs/specs/src/interfaces/IFeeAMM.sol
howydev: Haha I tried with `Mint` first actually, this is fine in solidity but it creates a collision within the `sol!` macro sadly

One possibility is to delay this change till mainnet, so we no longer need to HF gate this, then we can use `Mint`. WDYT?

crates/contracts/src/precompiles/tip_fee_manager.rs
```diff
@@ -99,7 +99,7 @@ sol! {
         function rebalanceSwap(address userToken, address validatorToken, uint256 amountOut, address to) external returns (uint256 amountIn);
 
         // Events
-        event Mint(address indexed sender, address indexed userToken, address indexed validatorToken, uint256 amountUserToken, uint256 amountValidatorToken, uint256 liquidity);
+        event Mint(address sender, address indexed to, address indexed userToken, address indexed validatorToken, uint256 amountValidatorToken, uint256 liquidity);
```
0xKitsune, line 102: Noting that we will want to flag this to front end teams to let them know the indexed field has changed. Since the previously indexed field was also `address` this could go unnoticed.

0xKitsune (approved):

crates/contracts/src/precompiles/tip_fee_manager.rs
howydev: good call, just tagged them on slack

PR #1469, opened 2025-12-12
fix(precompiles): check OOB in `Vec<T>::at(index)`

### LINKED ISSUE #1381: bug(precompiles): make `VecHandler` API memory safe when indexing
### Context

unlike `ArrayHandler`, i initially designed `VecHandler` to expand storage when indexed OOB:
```rs
// ArrayHandler is memory safe
fn at(&self, index: usize) -> Option<T::Handler>

// VecHandler expands storage
fn at(&self, index: usize) -> T::Handler;
```

@jeremyrubin reported on twitter:
>if an attacker controls keys and can set array length arbitrarily, they can pick a K2 and index j such that slot(K2,j) aliases slot(K1,i).
>
>without a bounds check, it's trivial to just compute a big enough index to alias.

## Solution

adopt a consistent API `fn at(&self, index: usize) -> Option<T::Handler>` for both indexable handlers, and if we ever need arbitrary expansion beyond `fn push(value: T)`, we can add a dedicated function

0xrusowsky:
## Motivation

closes #1381 

## Solution

turn the fn into `fn at(&self, index: usize) -> Result<Option<T::Handler>>` and add OOB checks by reading the vec length from storage.

Additionally, introduce `fn at_unchecked(&self, index: usize) -> T::Handler` to have an efficient that avoids the SLOAD when using it internally in a safe manner

### DIFF
--- a/crates/precompiles/src/storage/types/vec.rs
+++ b/crates/precompiles/src/storage/types/vec.rs
@@ -261,14 +261,13 @@ where
         T::handle(base_slot, layout_ctx, self.address)
     }
 
-    /// Exactly like `fn at_unchecked(..)` but with an OOB check.
+    /// Exactly like `fn at_unchecked(..)` but with an out-of-bounds check.
     ///
     /// # Returns
     /// - If the SLOAD to read the length fails, throws an error.
     /// - If the index is OOB returns `Ok(None)`.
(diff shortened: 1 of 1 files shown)


klkvr (approved):

mattsse (approved):
no functional changes lgtm

PR #1471, opened 2025-12-12
feat: impl Iter trait for `ArrayHandler` and `VecHandler`

0xrusowsky:
## Motivation

now that we have the new storage type system, we can extend handlers with useful traits for better devex.

this PR adds impls `trait Iter` for both `ArrayHandler` and `VecHandler`, so that we can easily iterate over their elements in an efficient manner (i.e. the vec impl just uses one SLOAD)

NOTE: despite we don't have many arrays in the codebase, i think it's worth to improve the API to easily interact with storage in a safe and efficient manner


0xKitsune (approved):
Lgtm, pending @klkvr

0xrusowsky:
disclaimer: @klkvr had some potential concerns about users being confused with the expected behavior of the iter trait, so let's hold merging before we also have his approval

crates/precompiles/src/storage/types/vec.rs
```diff
@@ -326,8 +326,46 @@ where
...
+        })
+    }
+}
+
+/// Iterator over `Vec<T>` element handlers. Implements [`ExactSizeIterator`].
+pub struct VecIter<'a, T: Storable> {
+    handler: &'a VecHandler<T>,
+    length: usize,
+    current: usize,
 }
```
klkvr, line 347: yeah I'm not sure if I really like this given that it allows to do something like

```rust
for item in array.iter() {
     array.push(new_item);
}
```
and `new_item` will not be emitted as part of iterator

maybe changing `VecIter` to hold `&mut VecHandler` would make it better, but I guess this does not address this entirely because it's quite easy to get 2 instances of the same handler

0xKitsune:
Closing for now given the above discussion, feel free to reopen if there is alignment on an approach and we still want this feature.

Issue #1459, opened 2025-12-11
Pre-calculate storage slots for transactions


shekhirin:
### Describe the feature

We know beforehand for a transaction in the txpool that it will access certain storage slots when it's executed:
1. If `fee_token` is specified, the balance of that token for the tx sender
2. If it's a TIP-20 transfer, the balances of that token for `from` and `to`

This can be done in the txpool, same as we do here for cached 2D nonce slots https://github.com/tempoxyz/tempo/blob/316e75e3a8f691583bd4a486915fe915f36c236b/crates/transaction-pool/src/transaction.rs#L33-L34

It should be moved to a new struct that contains all pre-calculated or cached slots.

PR #1472, opened 2025-12-12
perf(pool): cache balance slots

xwid138:
Adds `cached_slots` to `TempoPooledTransaction`:

* `nonce_key` - slot pointing to sender's 2D nonce
* `fee_token_balance` - slot pointing to sender's balance of tee token
* `tip20_from_balance` - TIP-20 transfer sender's token balance
* `tip20_to_balance` -  TIP-20 transfer receiver's token balance

fixes: #1459

### DIFF
--- a/crates/revm/src/tx.rs
+++ b/crates/revm/src/tx.rs
@@ -298,7 +298,23 @@ impl IntoTxEnv<Self> for TempoTxEnv {
 
 impl FromRecoveredTx<EthereumTxEnvelope<TxEip4844>> for TempoTxEnv {
     fn from_recovered_tx(tx: &EthereumTxEnvelope<TxEip4844>, sender: Address) -> Self {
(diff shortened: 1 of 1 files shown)


crates/transaction-pool/src/transaction.rs
```diff
@@ -82,14 +97,81 @@ impl TempoPooledTransaction {
...
+        *self.cached_slots.nonce_key.get_or_init(|| {
             let nonce_key = self.nonce_key()?;
             let sender = self.sender();
             let slot = NonceManager::new().nonces.at(sender).at(nonce_key).slot();
             Some(slot)
         })
     }
 
+    /// Returns the storage slot for sender's fee token balance if transaction contains fee token
+    pub fn fee_token_balance_slot(&self) -> Option<U256> {
```
gakonst, line 85: @shekhirin should this also take into account if the tx is a tip20 or fee manager setUserToken call?

PR #1477, opened 2025-12-12
feat(consensus): finalize DKG speculatively

### LINKED ISSUE #924: consensus: finalize DKG ceremony speculatively

SuperFluffy:
Changes DKG finalization to work on notarized blocks.

Tempo wants to finalize its DKG ceremony and include the outcome in the boundary block of each epoch. With this patch, the application actor requests the DKG outcome based on the boundary block's parent digest. The DKG manager then decides whether it can already construct the outcome of the DKG ceremony (because it has already finalized the pre-to-boundary block), or if it has to first fetch notarized blocks and construct the outcome based on those.

Before this patch, the DKG actor would only construct the DKG outcome upon finalization of the pre-to-boundary block. But because simplex gives no guarantees about observing direct finalizations of all blocks, this could lead to chain halts in certain edge cases (network instability comes to mind).

With this patch the node no longer persists the dealings it finds in finalized blocks. Instead, on restart it backfills them from the marshal actor.

Closes #924

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11504,13 +11504,13 @@ name = "tempo-chainspec"
 version = "0.7.5"
 dependencies = [
  "alloy-eips",
+ "alloy-evm",
  "alloy-genesis",
  "alloy-hardforks",
  "alloy-primitives",
  "eyre",
  "reth-chainspec",
  "reth-cli",
- "reth-ethereum",
  "reth-network-peers",
  "serde",
  "serde_json",
@@ -11607,7 +11607,9 @@ dependencies = [
 name = "tempo-contracts"
 version = "0.7.5"
 dependencies = [
- "alloy",
+ "alloy-contract",
+ "alloy-primitives",
+ "alloy-sol-types",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -111,7 +111,7 @@ tempo-payload-builder = { path = "crates/payload/builder", default-features = fa
 tempo-payload-types = { path = "crates/payload/types", default-features = false }
 tempo-precompiles = { path = "crates/precompiles", default-features = false }
 tempo-precompiles-macros = { path = "crates/precompiles-macros" }
-tempo-primitives = { path = "crates/primitives", default-features = false }
+tempo-primitives = { path = "crates/primitives", default-features = false, features = ["reth"] }
 tempo-contracts = { path = "crates/contracts", default-features = false }
 tempo-telemetry-util = { path = "crates/telemetry-util", default-features = false }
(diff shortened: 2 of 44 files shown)


crates/e2e/src/tests/backfill.rs
```diff
@@ -52,16 +54,16 @@ async fn run_validator_late_join_test(
...
             0, actual_runs,
-            "Expected no backfill, got {actual_runs} runs"
+            "expected no backfill, got {actual_runs} runs"
         );
     }
 
     // Verify that the node is still progressing after sync
     let last_block = last.execution_provider().last_block_number().unwrap();
-    context.sleep(Duration::from_secs(5)).await;
+    context.sleep(Duration::from_secs(10)).await;
```
SuperFluffy, line 63: Waiting a bit longer made the tests less flaky.

crates/e2e/src/tests/dkg/dynamic.rs
```diff
@@ -11,16 +11,6 @@ use tracing::info;
 
 use crate::{CONSENSUS_NODE_PREFIX, Setup, execution_runtime::validator, setup_validators};
 
-#[test_traced]
-fn after_hardfork_validator_is_added_to_a_set_of_one() {
```
SuperFluffy, line 15: These single node tests are extremely flaky and slow - because we are on a single thread, the active node runs ahead and it takes the new node a long time to catch up. Looking at logs, the epoch-gap is slowly closing, but the test takes extraordinarily long to actually complete.

The multiple-validator tests equally show test success but are much faster.

crates/commonware-node/src/consensus/engine.rs
```diff
@@ -245,7 +245,7 @@ where
                 execution_node,
                 initial_share: self.share.clone(),
                 mailbox_size: self.mailbox_size,
-                marshal: marshal_mailbox,
+                marshal: marshal_mailbox.clone(),
```
joshieDo: is this necessary?

SuperFluffy: Unnecessary; remnant of a refactor I I didn't end up keeping: e6e1119d

Issue #1083, opened 2025-11-26
TransferWithMemo should come after Transfer


jenpaff:
### Describe the feature

ref https://tempoxyz.slack.com/archives/C09K7873V4N/p1764171629256859?thread_ts=1764119087.606879&cid=C09K7873V4N 

This needs to be gated behind a fork, can pick this up if we get round to it, otherwise postpone to next fork.

PR #1479, opened 2025-12-12
fix: reorder transfer events for mint/burn with memo

howydev:
fixes: https://github.com/tempoxyz/tempo/issues/1083


0xrusowsky:
@howydev i don't have context, but maybe double-check with the UI team to make sure this won't impact reintroduce the explorer bug where mints where double-counted

0xKitsune (approved):

PR #1484, opened 2025-12-12
perf: use secp256k1 for recovery

mattsse:
this always used the default (slow) k256 recovery impl

ref 

https://github.com/tempoxyz/tempo/blob/3336a339bc179d1a351a5aec340a47fbde941b95/crates/revm/src/tx.rs#L286-L286

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -79,7 +79,8 @@ jobs:
       - name: Get latest filename from API
         id: latest
         run: |
-          FILENAME=$(curl -s https://snapshots.tempoxyz.dev/42429/latest.txt)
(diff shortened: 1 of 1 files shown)


legion2002 (approved):
I believe, we also need to update it here - 
https://github.com/tempoxyz/tempo/blob/f8c4c034563b8588d094df0058e3d00341b0099b/crates/revm/src/tx.rs#L298

and here
https://github.com/tempoxyz/tempo/blob/f8c4c034563b8588d094df0058e3d00341b0099b/crates/revm/src/tx.rs#L376C21-L376C22

Other than that, lgtm!
Also noting that this is not just a perf change, but also a security improvement, because the alloy function adds the "high s" check while recovering sigs.

PR #1488, opened 2025-12-12
feat(precompiles): impl `Index` and `IndexMut` traits

0xrusowsky:
## Motivation

this PR tackles 2 goals:

1. **devex:** enable solidity-like index access (`[key]`) for indexable handlers (`Mapping`, `ArrayHandler`, and `VecHandler`)
2. **performance:** cache handlers that require expensive computations (keccak256 hashes)

## Solution

since `fn at()` returns an owned `T::Handler`, we cannot impl the index trait with our current design. Instead, we need to returns refs `&T::Handler`.

Because of that, this PR introduces a new `HandlerCache` struct to store child handlers at the parent (indexable) handler. The cache:
- uses `UnsafeCell<HashMap<K, Box<H>>>` for interior mutability
- lazily computes handlers on first access

## API Changes

| Before                                | After                            |
|---------------------------------------|----------------------------------|
| `self.books.at(book_key)`             | `self.books[book_key]`           |
| `self.nonces.at(account).at(nonce_key)`| `self.nonces[account][nonce_key]`|

**NOTE:** this is exploratory work. There may be better alternatives, but I like this design as it removes the need for manual handler caching we used to do in the stablecoin exchange.

### DIFF
--- a/crates/precompiles/src/storage/evm.rs
+++ b/crates/precompiles/src/storage/evm.rs
@@ -44,6 +44,7 @@ impl<'a> EvmPrecompileStorageProvider<'a> {
         Self::new(internals, u64::MAX, cfg.chain_id, cfg.spec, false)
     }
 
+    /// Ensures that an account is loaded.
     pub fn ensure_loaded_account(&mut self, account: Address) -> Result<(), EvmInternalsError> {
         self.internals.load_account(account)?;
         self.internals.touch_account(account);
@@ -80,14 +81,12 @@ impl<'a> PrecompileStorageProvider for EvmPrecompileStorageProvider<'a> {
         address: Address,
         f: &mut dyn FnMut(&AccountInfo),
     ) -> Result<(), TempoPrecompileError> {
-        self.ensure_loaded_account(address)?;
         let account = self.internals.load_account_code(address)?.map(|a| &a.info);
-        let is_cold = account.is_cold;
 
         // deduct gas
         self.gas_remaining = self
             .gas_remaining
-            .checked_sub(revm::interpreter::gas::warm_cold_cost(is_cold))
+            .checked_sub(revm::interpreter::gas::warm_cold_cost(account.is_cold))
             .ok_or(TempoPrecompileError::OutOfGas)?;
 
         f(account.data);


crates/precompiles/src/storage/types/mapping.rs
```diff
@@ -30,21 +34,22 @@ use crate::storage::{Layout, LayoutCtx, StorableType, StorageKey};
...
 /// When a mapping is a field within a struct stored in another mapping, use the static
 /// `at_offset` method to compute the slot without creating a `Mapping` instance:
 #[derive(Debug, Clone)]
-pub struct Mapping<K, V> {
+pub struct Mapping<K, V: StorableType> {
     base_slot: U256,
     address: Address,
-    _phantom: PhantomData<(K, V)>,
+    cache: HandlerCache<K, V::Handler>,
+    _phantom: PhantomData<V>,
```
klkvr: I don't think we need `PhantomData` here anymore

crates/precompiles/src/storage/types/array.rs
```diff
@@ -47,30 +53,26 @@ tempo_precompiles_macros::storable_nested_arrays!();
...
 /// ```
-pub struct ArrayHandler<T, const N: usize>
-where
-    T: StorableType,
-{
+pub struct ArrayHandler<T: StorableType, const N: usize> {
     base_slot: U256,
     address: Address,
+    cache: HandlerCache<usize, T::Handler>,
     _phantom: PhantomData<T>,
```
klkvr: and here

crates/precompiles/src/storage/types/mapping.rs
```diff
@@ -65,52 +71,93 @@ impl<K, V> Mapping<K, V> {
...
-    pub fn at_offset(
-        struct_base_slot: U256,
-        field_offset_slots: usize,
-        address: Address,
-        key: K,
-    ) -> V::Handler
+    /// Use this when you need to call mutable methods like `write()` or `delete()`.
+    ///
+    /// The handler is computed on first access and cached for subsequent accesses.
+    pub fn at_mut(&self, key: K) -> &mut V::Handler
```
klkvr: hmm what's the usecase for this? iiuc we didn't allow this before and technically it now makes it super simple to obtain 2 `&mut` refs to the same handler which could cause UB

crates/precompiles/src/storage/types/mod.rs
```diff
@@ -310,3 +311,61 @@ pub trait StorageKey {
...
+    pub(super) fn get_or_insert_mut(&self, key: K, f: impl FnOnce() -> H) -> &mut H {
+        // SAFETY:
+        // 1. Single-threaded access (EVM execution model)
+        // 2. Box ensures stable heap address even when HashMap rehashes
+        // 3. Append-only: we only insert, never remove entries
+        // 4. Caller must ensure `f()` doesn't access (reenter) this cache
+        unsafe {
+            let cache = &mut *self.inner.get();
+            cache.entry(key).or_insert_with(|| Box::new(f()))
+        }
```
klkvr: hmm so we assume here that this is a unique refernece to `&self` and thus it's safe to get a `&mut self.inner`

and returning a ref to `Box<T>` location should be fine too

though it does not seem right that this returns `&mut H`

crates/precompiles/src/storage/types/mod.rs
```diff
@@ -310,3 +311,61 @@ pub trait StorageKey {
...
+    }
+
+    /// Returns a mutable reference to a lazily initialized handler for the given key.
+    ///
+    /// # SAFETY
+    ///
+    /// This method is safe to call as long as the closure `f` doesn't re-enter.
+    #[inline]
+    #[allow(clippy::mut_from_ref)]
+    pub(super) fn get_or_insert_mut(&self, key: K, f: impl FnOnce() -> H) -> &mut H {
```
klkvr: can we make this take `&mut self` too? and `get_or_insert` can do the same thing but return `&H`

crates/precompiles/src/storage/types/mod.rs
```diff
@@ -310,3 +311,69 @@ pub trait StorageKey {
...
+        }
+    }
+
+    /// Returns a mutable reference to a lazily initialized handler for the given key.
+    ///
+    /// # SAFETY
+    ///
+    /// This method is safe to call as long as the closure `f` doesn't re-enter.
+    #[inline]
+    #[allow(clippy::mut_from_ref)]
```
klkvr:
```suggestion
```

crates/node/src/rpc/mod.rs
```diff
@@ -288,8 +288,7 @@ impl<N: FullNodeTypes<Types = TempoNode>> Call for TempoEthApi<N> {
         {
             let slot = NonceManager::new()
                 .nonces
-                .at(request.from.unwrap_or_default())
-                .at(nonce_key)
+                .at(request.from.unwrap_or_default())[nonce_key]
```
klkvr: could we do this?

```suggestion
                .[request.from.unwrap_or_default()][nonce_key]
```

crates/precompiles/src/stablecoin_exchange/orderbook.rs
```diff
@@ -150,9 +150,9 @@ impl Orderbook {
 impl OrderbookHandler {
     pub fn get_tick_level_handler(&self, tick: i16, is_bid: bool) -> TickLevelHandler {
         if is_bid {
-            self.bids.at(tick)
+            self.bids[tick].clone()
```
klkvr: could return ref here Ig

crates/precompiles/src/storage/types/mapping.rs
0xrusowsky: we need this cause the `IndexMut` trait requires a mutable ref.

we need `IndexMut` cause `delete()` and `write()` require the handler to be mutable

crates/precompiles/src/storage/types/mod.rs
0xrusowsky: yes, my bad, this should definitely be `&mut`

crates/precompiles/src/stablecoin_exchange/orderbook.rs
0xrusowsky: addressed in [`76bcf7b` (#1488)](https://github.com/tempoxyz/tempo/pull/1488/commits/76bcf7b4b1fd880a9739a0c954e6dc3647738435)

DaniPopes (approved):

PR #1494, opened 2025-12-13
perf(sub-block): avoid double encoding during signature hashing

0x00101010:
Avoid double encoding during signature hashing

### DIFF
--- a/crates/primitives/src/subblock.rs
+++ b/crates/primitives/src/subblock.rs
@@ -132,6 +132,10 @@ impl Encodable for SubBlock {
         self.rlp_header().encode(out);
         self.rlp_encode_fields(out);
     }
+
+    fn length(&self) -> usize {
+        self.rlp_header().length_with_payload()
+    }
 }
 
 /// A subblock with a signature.


crates/primitives/src/subblock.rs
```diff
@@ -86,8 +86,7 @@ pub struct SubBlock {
 impl SubBlock {
     /// Returns the hash for the signature.
     pub fn signature_hash(&self) -> B256 {
-        let mut buf = Vec::with_capacity(self.length() + 1);
```
0x00101010, line 89: `self.length()` calls into default impl (no override), and I think vec reallocation is still gonna be much faster than preallocation by calculating the length of SubBlock through other methods

```rust
fn length(&self) -> usize {
    let mut out = Vec::new();
    self.encode(&mut out);
    out.len()
}
    
```

klkvr (changes requested):
ah good catch, proper fix here would be to define `fn length` as `self.rlp_header().length_with_payload()`

klkvr:
we should definitely override `length` fn

`self.encode` will still compute the length for the header so I guess we could just do something like 
```rust
let header = self.rlp_header();
let mut buf = Vec::with_capacity(header.length_with_payload() + 1);
buf.put_u8(SUBBLOCK_SIGNATURE_HASH_MAGIC_BYTE);
header.encode(&mut buf);
self.rlp_encode_fields(&mut buf);
```

PR #1498, opened 2025-12-13
perf: remove inefficient operations

malik672:
small nits:  remove inefficient operations

https://godbolt.org/z/jE1oYxozn

div_ceil when the divisor is a power of 2

### DIFF
--- a/crates/precompiles/src/storage/packing.rs
+++ b/crates/precompiles/src/storage/packing.rs
@@ -133,8 +133,7 @@ pub fn insert_packed_value<T: Packable>(
 
     // Calculate shift and mask
     let shift_bits = offset * 8;
-
-    let mask =  create_element_mask(bytes);
+    let mask = create_element_mask(bytes);
 
     // Clear the bits for this field in the current slot value
     let clear_mask = !(mask << shift_bits);
@@ -161,7 +160,7 @@ pub fn zero_packed_value(current: U256, offset: usize, bytes: usize) -> Result<U
         )));
     }
 
-    let mask =  create_element_mask(bytes);
+    let mask = create_element_mask(bytes);
     let shifted_mask = mask << (offset * 8);
     Ok(current & !shifted_mask)
 }
@@ -181,19 +180,15 @@ pub const fn calc_element_offset(idx: usize, elem_bytes: usize) -> usize {
 /// Calculate the byte offset within a slot for an array element at index `idx` and which slot an array element at index `idx` starts in .
 #[inline]
 pub const fn calc_element_offset_and_slot(idx: usize, elem_bytes: usize) -> (usize, usize) {
-   let total =  idx * elem_bytes;
-   ((total / 32), (total % 32))
+    let total = idx * elem_bytes;
+    ((total / 32), (total % 32))
 }
 
 /// Calculate the element location within a slot for an array element at index `idx`.
 #[inline]
 pub const fn calc_element_loc(idx: usize, elem_bytes: usize) -> FieldLocation {
     let (slots, offset) = calc_element_offset_and_slot(idx, elem_bytes);
-    FieldLocation::new(
-        slots,
-        offset,
-        elem_bytes,
-    )
+    FieldLocation::new(slots, offset, elem_bytes)
 }
 
 /// Calculate the total number of slots needed for an array.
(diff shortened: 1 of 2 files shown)


crates/precompiles/src/storage/packing.rs
```diff
@@ -133,7 +133,9 @@ pub fn insert_packed_value<T: Packable>(
 
     // Calculate shift and mask
     let shift_bits = offset * 8;
-    let mask = create_element_mask(bytes);
+
+    // we know bytes is less than 32 here because of the invariant above
+    let mask = (U256::ONE << (bytes * 8)) - U256::ONE;
```
0xrusowsky: the invariant above checks > 32, but we could still pass 32 bytes
can we leave this as it was set? `fn create_element_mask` is marked as `inline` anyways

crates/precompiles/src/storage/packing.rs
```diff
@@ -177,20 +180,29 @@ pub const fn calc_element_offset(idx: usize, elem_bytes: usize) -> usize {
...
 pub const fn calc_element_loc(idx: usize, elem_bytes: usize) -> FieldLocation {
+    let (slots, offset) = idx * elem_bytes;
     FieldLocation::new(
-        calc_element_slot(idx, elem_bytes),
-        calc_element_offset(idx, elem_bytes),
+        slots,
+        offset,
         elem_bytes,
     )
 }
```
0xrusowsky, line 192: while this makes sense, `fn calc_element_slot` and `fn calc_element_offset` become deadcode (just used for tests)

please remove them and update the tests accordingly to use `fn calc_element_offset_and_slot`

crates/precompiles/src/storage/packing.rs
malik672: This just made me see a bug offset can be zero

crates/precompiles/src/storage/packing.rs
malik672: had the same thought but Nahh they are not dead code yet.

it's used here

```
/// Generate store implementation for packed arrays
fn gen_packed_array_store(array_size: &usize, elem_byte_count: &usize) -> TokenStream {
    quote! {
        // Determine how many slots we need
        let slot_count = (#array_size * #elem_byte_count).div_ceil(32);

        // Build slots by packing elements
        for slot_idx in 0..slot_count {
            let slot_addr = base_slot + U256::from(slot_idx);
            let mut slot_value = U256::ZERO;

            // Pack all elements that belong to this slot
            for i in 0..#array_size {
                let elem_slot = calc_element_slot(i, #elem_byte_count);
                if elem_slot == slot_idx {
                    let offset = calc_element_offset(i, #elem_byte_count);
                    slot_value = insert_packed_value(slot_value, &self[i], offset, #elem_byte_count)?;
                }
            }

            storage.store(slot_addr, slot_value)?;
        }
        Ok(())
    }
}
```

0xrusowsky: ah, good catch, should have used grep rather than the LSP

0xrusowsky: @malik672 let's replace the individual fns to use `calc_element_offset_and_slot.<0/1>` instead, to centralize the logic in a single fn.

since we use `#[inline]` LLVM will optimize the unused branch away

malik672: You can't optimize a branch away though, the moment you end up hitting it it's already there

but I think it's a concession that can be made not just sure

0xrusowsky: hey @malik672, i had a quick sync with @DaniPopes and agreed that we should only do micro optimizations which are backed by benchmarks or godbolt. Otherwise the base case is to assume that the compiler is smarter than us.

because of that, let's keep this PR with just the `div_ceil` change, and if u want to explore further micro optimizations, i encourage you to do that in followup PRs. Thanks!

malik672: > hey @malik672, i had a quick sync with @DaniPopes and agreed that we should only do micro optimizations which are backed by benchmarks or godbolt. Otherwise the base case is to assume that the compiler is smarter than us.
> 
> because of that, let's keep this PR with just the `div_ceil` change, and if u want to explore further micro optimizations, i encourage you to do that in followup PRs. Thanks!

fascinating just checked without inline the proposed is better with inline they are the same 
but this is godbolt anyway I need to run the code as a whole and generate the ASM then check it but since this is unlikely, I think I should close it

0xrusowsky: good to know. thanks for the effort anyways!

PR #1510, opened 2025-12-15
fix: P256 high s check

legion2002:
The secp256k1 function doesn't allow s values higher than n/2, to prevent 2 tx hashes being valid on the same signature.

We add the same check for our P256 verification function.

The n value for the curve was taken from [RIP 2712](https://eip.tools/rip/7212) 
The division with 2 was checked with wolfram, and python scripts.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -76,10 +76,46 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - uses: rui314/setup-mold@v1
       - uses: mozilla-actions/sccache-action@v0.0.9
-      - name: Download snapshot
-        run: cargo run --release --bin tempo download --datadir snap-rexec
+      - name: Get latest filename from API
+        id: latest
+        run: |
+          # FILENAME=$(curl -s https://snapshots.tempoxyz.dev/42429/latest.txt)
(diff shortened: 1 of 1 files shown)


crates/primitives/src/transaction/tt_signature.rs
```diff
@@ -10,6 +10,40 @@ use p256::{
...
+///
+/// For signatures to be valid, the s value must be less than or equal to n/2
+/// (low-s requirement). This prevents signature malleability where (r, s) and
+/// (r, n-s) are both valid signatures for the same message.
+///
+/// n/2 = 0x7FFFFFFF800000007FFFFFFFFFFFFFFFDE737D56D38BCF4279DCE5617E3192A8
+pub const P256N_HALF: U256 = U256::from_be_bytes([
+    0x7F, 0xFF, 0xFF, 0xFF, 0x80, 0x00, 0x00, 0x00, 0x7F, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
+    0xDE, 0x73, 0x7D, 0x56, 0xD3, 0x8B, 0xCF, 0x42, 0x79, 0xDC, 0xE5, 0x61, 0x7E, 0x31, 0x92, 0xA8,
+]);
```
klkvr: can we use `uint!` macro here?

PR #1515, opened 2025-12-15
fix(stablecoin-dex): propagate underflow/overflow errors

0xKitsune:
This PR updates the stablecoin dex to propagate all underflow/overflow errors.

### DIFF
--- a/crates/precompiles/src/stablecoin_exchange/mod.rs
+++ b/crates/precompiles/src/stablecoin_exchange/mod.rs
@@ -637,7 +637,7 @@ impl StablecoinExchange {
         // Create the flip order
         let order_id = self.increment_pending_order_id()?;
         let order = Order::new_flip(order_id, sender, book_key, amount, tick, is_bid, flip_tick)
-            .expect("Invalid flip tick");
+            .map_err(|_| StablecoinExchangeError::invalid_flip_tick())?;
 
         // Store in pending queue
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -817,12 +817,13 @@ impl StablecoinExchange {
...
                 .checked_mul(price as u128)
                 .and_then(|v| v.checked_div(orderbook::PRICE_SCALE as u128))
-                .expect("Amount out calculation overflow")
+                .ok_or(TempoPrecompileError::under_overflow())?
         } else {
             let quote_amount = fill_amount
                 .checked_mul(price as u128)
                 .and_then(|v| v.checked_div(orderbook::PRICE_SCALE as u128))
-                .expect("Amount out calculation overflow");
+                .ok_or(TempoPrecompileError::under_overflow())?;
```
mattsse, line 825: these are all fine, but technically we only need these for the checked_mul

and could do 

```
.checked_mul(price as u128).ok_or(TempoPrecompileError::under_overflow())? / orderbook::PRICE_SCALE as u128
```

0xKitsune: This makes sense, I can update in a subsequent PR.

PR #1517, opened 2025-12-15
fix: disallow precompile static calls

0xKitsune:
This PR adds `is_static` to `StorageCtx` and rejects static calls when calling `mutate`/`mutate_void`. This PR also bumps `alloy-evm` to `0.25.2`.


klkvr (approved):
lgtm

PR #1532, opened 2025-12-15
feat(cli): `--builder.disable-state-cache`

shekhirin:
Makes it possible to disabled cached reads in payload builder

### DIFF
--- a/crates/payload/builder/src/lib.rs
+++ b/crates/payload/builder/src/lib.rs
@@ -342,7 +342,7 @@ where
             .with_database(if self.disable_state_cache {
                 Box::new(state) as Box<dyn Database<Error = ProviderError>>
             } else {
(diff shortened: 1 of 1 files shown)


crates/payload/builder/src/lib.rs
```diff
@@ -335,7 +339,11 @@ where
         };
         let state = StateProviderDatabase::new(&state_provider);
         let mut db = State::builder()
-            .with_database(cached_reads.as_db_mut(state))
+            .with_database(if self.disable_state_cache {
+                Box::new(state) as Box<dyn Database<Error = ProviderError>>
+            } else {
+                Box::new(cached_reads.as_db_mut(state))
```
mattsse, line 345: this now makes it always operate on box, should be fine here

PR #1537, opened 2025-12-16
feat(fee-manager): remove fee manager system tx

### LINKED ISSUE #1439: Swap immediately and remove system transaction
### Describe the feature

Removes system tx that optimises by doing one swap at bottom of the block

### Additional context

- [ ] spec PR https://github.com/tempoxyz/tempo/pull/1415

0xKitsune:
Closes #1439 

Implements immediate fee swaps during transaction execution, replacing the batched end-of-block approach.

- `collectFeePostTx` now executes swaps immediately via executeFeeSwap and accumulates fees directly
- Skips fee manager executeBlock system transaction
- Adds distributeFees(validator) for validators to claim accumulated fees on-demand
- Pre-hardfork behavior is preserved for backward compatibility

### DIFF
--- a/crates/payload/builder/src/lib.rs
+++ b/crates/payload/builder/src/lib.rs
@@ -26,7 +26,11 @@ use reth_evm::{
 use reth_payload_builder::{EthBuiltPayload, PayloadBuilderError};
 use reth_payload_primitives::PayloadBuilderAttributes;
 use reth_primitives_traits::{Recovered, transaction::error::InvalidTransactionError};
-use reth_revm::{State, context::Block, database::StateProviderDatabase};
+use reth_revm::{
+    State,
+    context::{Block, BlockEnv},
+    database::StateProviderDatabase,
+};
 use reth_storage_api::{StateProvider, StateProviderFactory};
 use reth_transaction_pool::{
     BestTransactions, BestTransactionsAttributes, TransactionPool, ValidPoolTransaction,
@@ -154,13 +158,12 @@ impl<Provider: ChainSpecProvider<ChainSpec = TempoChainSpec>> TempoPayloadBuilde
     /// 3. Subblocks signatures - validates subblock signatures
     fn build_seal_block_txs(
         &self,
-        evm: &TempoEvm<impl Database>,
+        block_env: &BlockEnv,
         subblocks: &[RecoveredSubBlock],
         timestamp: u64,
     ) -> Vec<Recovered<TempoTxEnvelope>> {
         let chain_spec = self.provider.chain_spec();
         let chain_id = Some(chain_spec.chain().id());
-        let block_env = evm.block();
         let mut txs = Vec::with_capacity(3);
 
         // Build fee manager and stablecoin dex system transaction (pre-AllegroModerato only)
@@ -443,7 +446,7 @@ where
         // Prepare system transactions before actual block building and account for their size.
         let prepare_system_txs_start = Instant::now();
         let system_txs =
-            self.build_seal_block_txs(builder.evm(), &subblocks, attributes.timestamp());
+            self.build_seal_block_txs(builder.evm().block(), &subblocks, attributes.timestamp());
         for tx in &system_txs {
             block_size_used += tx.inner().length();
         }


crates/precompiles/src/tip_fee_manager/mod.rs
```diff
@@ -183,17 +185,25 @@ impl TipFeeManager {
...
+            if self.storage.spec().is_allegro_moderato() {
+                // Post-AllegroModerato: ensure there is enough liquidity to swap `max_amount`
+                self.check_sufficient_liquidity(user_token, validator_token, max_amount)?;
+            } else {
+                // Pre-AllegroModerato: reserve liquidity for later swap in execute_block
+                self.reserve_liquidity(user_token, validator_token, max_amount)?;
+            }
+        } else if self.storage.spec().is_allegro_moderato() {
+            // Same token, no swap needed - just accumulate fees
+            self.increment_collected_fees(beneficiary, max_amount)?;
```
fgimenez: later we increment fees too by `actual_spending`, is it correct to collect `max_amount + actual_spending` as fees?

crates/precompiles/src/tip_fee_manager/mod.rs
```diff
@@ -218,12 +228,18 @@ impl TipFeeManager {
...
+            if !self.storage.spec().is_allegro_moderato() {
+                self.release_liquidity(fee_token, validator_token, refund_amount)?;
+            }
 
             // Record the pool if there was a non-zero swap
             if !actual_spending.is_zero() {
-                if !self.storage.spec().is_allegretto() {
+                if self.storage.spec().is_allegro_moderato() {
+                    // Execute fee swap immediately and accumulate fees
+                    self.execute_fee_swap(fee_token, validator_token, actual_spending)?;
```
fgimenez, line 241: nit, here we are discarding the return value of `execute_fee_swap` (`amount_out` calculated with `compute_amount_out`), later we call `compute_amount_out` when fee token != validator token, maybe we could store `amount_out` and avoid calling `compute_amount_out` twice

fgimenez (approved):
lgtm

crates/contracts/src/precompiles/tip_fee_manager.rs
```diff
@@ -34,12 +34,15 @@ sol! {
         // Fee functions
         function getFeeTokenBalance(address sender, address validator) external view returns (address, uint256);
         function executeBlock() external;
-        // NOTE: collectFeePreTx and collectFeePostTx are protocol-internal functions
-        // called directly by the execution handler, not exposed via the dispatch interface.
+        function distributeFees(address validator) external;
```
legion2002: Why do we need this new distributeFees function?

crates/node/tests/it/tip_fee_manager.rs
```diff
@@ -66,27 +66,54 @@ async fn test_set_user_token() -> eyre::Result<()> {
...
-        .await?
-        .get_receipt()
         .await?;
+    let tx_hash = *pending_tx.tx_hash();
+    let receipt = pending_tx.get_receipt().await?;
     assert!(receipt.status());
 
-    let expected_cost = calc_gas_balance_spending(receipt.gas_used, receipt.effective_gas_price);
+    // Post-AllegroModerato: fees are collected based on max_fee (gas_limit * gas_price), not gas_used.
+    // No refunds are issued - users pay for the full gas_limit upfront.
```
legion2002: Is this comment correct?

crates/precompiles/src/tip_fee_manager/dispatch.rs
```diff
@@ -97,9 +97,35 @@ impl Precompile for TipFeeManager {
...
+                    unknown_selector(selector, self.storage.gas_used(), self.storage.spec())
+                } else {
+                    mutate_void::<IFeeManager::executeBlockCall>(
+                        calldata,
+                        msg_sender,
+                        |s, _call| self.execute_block(s, self.storage.beneficiary()),
+                    )
+                }
+            }
+            IFeeManager::distributeFeesCall::SELECTOR => {
```
legion2002, line 110: nit: maybe we don't need to hardfork gate the new function selectors? Because very low chance that anyone ever called these.

crates/precompiles/src/tip_fee_manager/mod.rs
legion2002: Yes, this looks incorrect to me.
Have we removed gas refunds in this PR?
Because it seems like we are charging the max amount upfront, but not refunding it later.

crates/precompiles/src/tip_fee_manager/mod.rs
```diff
@@ -393,11 +410,45 @@ impl TipFeeManager {
...
+            if collected_fees.is_zero() {
+                self.validator_in_fees_array.at(validator).write(true)?;
+                self.validators_with_fees.push(validator)?;
+            }
+        }
+        Ok(())
+    }
+
+    /// Transfers the validator's fee balance to their address.
+    pub fn distribute_fees(&mut self, validator: Address) -> Result<()> {
```
legion2002, line 425: Why do we need this `distribute_fees` mechanism, can't we just transfer the funds directly to the validator post tx?

0xKitsune: Yes, we could send the funds post tx but the thinking here is to avoid an additional transfer on every tx.

crates/precompiles/src/tip_fee_manager/mod.rs
0xKitsune: This has been removed https://github.com/tempoxyz/tempo/pull/1537/commits/3658ffe5640a9ee190d8011c7c831c86c2e80763

crates/node/tests/it/tip_fee_manager.rs
0xKitsune: No this is outdated, will update.

crates/precompiles/src/tip_fee_manager/dispatch.rs
howydev: i think we actually must have this

IMO its not a probability thing, our PRs are open now so any bad actor can intentionally make a bad function call pre allegro-moderato

crates/precompiles/src/tip_fee_manager/mod.rs
legion2002: but the number of storage writes remain the same? 
Because all of this would be warm storage anyways. Whereas now, we have to write to this new collected fees storage slot.

crates/precompiles/src/tip_fee_manager/dispatch.rs
legion2002: fair point.

docs/specs/src/FeeManager.sol
```diff
@@ -57,29 +36,23 @@ contract FeeManager is IFeeManager, FeeAMM {
...
         }
 
-        // If user token is different from validator token, reserve AMM liquidity
+        ITIP20(userToken).transferFeePreTx(user, maxAmount);
+
         if (userToken != validatorToken) {
-            reserveLiquidity(userToken, validatorToken, maxAmount);
+            checkSufficientLiquidity(userToken, validatorToken, maxAmount);
+        } else {
+            collectedFees[block.coinbase] += maxAmount;
```
legion2002: The solidity spec also needs to be updated, with the latest changes you made around refunds.

0xKitsune: Sounds good, I just updated the PR to remove logic incrementing collected fees here.

crates/payload/builder/src/lib.rs
```diff
@@ -153,38 +149,49 @@ impl<Provider: ChainSpecProvider> TempoPayloadBuilder<Provider> {
...
         &self,
-        block_env: &BlockEnv,
+        evm: &TempoEvm<impl Database>,
         subblocks: &[RecoveredSubBlock],
     ) -> Vec<Recovered<TempoTxEnvelope>> {
         let chain_id = Some(self.provider.chain_spec().chain().id());
+        let block_env = evm.block();
+        let spec = &evm.ctx().cfg.spec;
+
+        let mut txs = Vec::with_capacity(3);
```
0xrusowsky, line 167: nit:
```suggestion
        let mut txs = Vec::with_capacity(if spec.is_allegro_moderato() { 2 } else { 3 });
```

note: may require fmt

0xrusowsky: actually, even better imo, you could add a helper fn to `TempoHardfork` (spec) so that it can yield the number of system txs.

This way you can do `spec.system_tx_count()` directly, and get rid off the branching, the constants, and the magic numbers, centralizing the logic in the source of truth which is the hardfork enum

crates/precompiles/src/tip_fee_manager/amm.rs
```diff
@@ -613,6 +630,47 @@ impl TipFeeManager {
...
+        let pool_id = self.pool_id(user_token, validator_token);
+        let mut pool = self.pools.at(pool_id).read()?;
+
+        // Calculate output at fixed price m = 0.9970
+        let amount_out = compute_amount_out(amount_in)?;
+
+        // Check if there's enough validatorToken available
+        if amount_out > U256::from(pool.reserve_validator_token) {
+            return Err(TIPFeeAMMError::insufficient_liquidity().into());
+        }
```
0xrusowsky, line 650: nit: this is literally the same code as in the newly introduced `fn check_sufficient_liquidity`, which makes me think that it is worth changing its sig to:
```rs
/// Converts `userToken` to `validatorToken` at the fixed rate `m = 0.9970` and validates that
/// the pool has enough liquidity.
fn validate_fee_swap(
    &mut self, user_token: Address, validator_token: Address, amount_in: U256
) -> Result<U256>;
```

so that we can reuse it:
- in `fn execute_fee_swap` you first directly validate the swap with the new helper (and get the amount out), to then execute it.
- in `tip_fee_manager/mod.rs` you can replace the `check_sufficient_liquidity` call for:
```rs
let _ = self.validate_fee_swap(user_token, validator_token, max_amount)?;
```

crates/precompiles/src/tip_fee_manager/mod.rs
```diff
@@ -393,11 +407,45 @@ impl TipFeeManager {
...
 
-        // If this is the first fee for the validator, record it in validators with fees
-        if collected_fees.is_zero() {
-            self.validator_in_fees_array.at(validator).write(true)?;
-            self.validators_with_fees.push(validator)?;
+        // Pre Allegro Moderato, add fees to be processed end of block
+        if !self.storage.spec().is_allegro_moderato() {
+            // If this is the first fee for the validator, record it in validators with fees
+            if collected_fees.is_zero() {
+                self.validator_in_fees_array.at(validator).write(true)?;
```
fgimenez, line 417: are we setting `validator_in_fees_array` post AllegroModerato? the guard in `set_validator_token` still relies on it, a validator might accumulate fees in token x, then change to token y, then distribute fees in token y?

crates/precompiles/src/tip_fee_manager/mod.rs
```diff
@@ -393,11 +407,45 @@ impl TipFeeManager {
...
+        if amount.is_zero() {
+            return Ok(());
+        }
+        self.collected_fees.at(validator).write(U256::ZERO)?;
+
+        // Transfer fees to validator
+        let validator_token = self.get_validator_token(validator)?;
+
+        let mut token = TIP20Token::from_address(validator_token)?;
+        token.transfer(
```
fgimenez, line 436: should we check balance before transfer? like we do in `execute_block`

crates/precompiles/src/tip_fee_manager/amm.rs
```diff
@@ -613,6 +630,47 @@ impl TipFeeManager {
...
+            .reserve_user_token
+            .checked_add(amount_in_u128)
+            .ok_or(TempoPrecompileError::under_overflow())?;
+        pool.reserve_validator_token = pool
+            .reserve_validator_token
+            .checked_sub(amount_out_u128)
+            .ok_or(TempoPrecompileError::under_overflow())?;
+
+        self.pools.at(pool_id).write(pool)?;
+
```
fgimenez, line 670: should we emit `FeeSwap` here? like https://github.com/tempoxyz/tempo/blob/main/docs/specs/src/FeeAMM.sol#L234

crates/precompiles/src/tip_fee_manager/mod.rs
0xKitsune: No we are not setting this post hardfork. I have updated to skip this check in `set_validator_token` PTAL: 
https://github.com/tempoxyz/tempo/pull/1537/commits/aa4b962cc0e14543f3e719d3784d1123c7e2206d

crates/precompiles/src/tip_fee_manager/amm.rs
0xKitsune: I think this will get pretty noisy if we emit this on every Tx and IMO we should leave this out, but will defer to @danrobinson if there is a strong preference.

crates/precompiles/src/tip_fee_manager/mod.rs
0xKitsune: Is there any case where the balance would not be sufficient to cover the payout?

fgimenez (approved):

crates/precompiles/src/tip_fee_manager/amm.rs
danrobinson: Strong preference to have no FeeSwap event!

crates/precompiles/src/tip_fee_manager/amm.rs
0xKitsune: `execute_fee_swap` needs access to the pool, if we drop in `valdiate_fee_swap` into this function we would need to pass the pool in, otherwise we would be sloading the pool twice (and would cause a diff in gas).

crates/payload/builder/src/lib.rs
0xKitsune: Sounds good, I think we can do this in a follow on to simplify.

PR #1543, opened 2025-12-16
fix: precompile CI

legion2002:
After the repos were made public, the sed command needs to be updated


0xrusowsky (approved):
good catch, thanks!

PR #1546, opened 2025-12-16
feat: add `ReadOnlyStorageProvider` to use precompiles anywhere

0xrusowsky:
## Motivation

there are still several places in the codebase (like the revm handler, or the tx-pool validator), where we perform manual storage slot checks on the precompiles.

This wasn't initially tackled with the precompile refactor, but is a potential source of human-errors, as we need to manually handle slots and their encoding/decoding.

## Solution

add a new `struct ReadOnlyStorageProvider` which partially implements `PrecompileStorageProvider` which panicks on all non-supported functions (which are no-op and should never be called).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11504,13 +11504,13 @@ name = "tempo-chainspec"
 version = "0.7.5"
 dependencies = [
  "alloy-eips",
+ "alloy-evm",
  "alloy-genesis",
  "alloy-hardforks",
  "alloy-primitives",
  "eyre",
  "reth-chainspec",
  "reth-cli",
- "reth-ethereum",
  "reth-network-peers",
  "serde",
  "serde_json",
@@ -11546,6 +11546,7 @@ dependencies = [
  "indexmap 2.12.1",
  "jiff",
  "parking_lot",
+ "pin-project",
  "prometheus-client",
  "rand 0.8.5",
  "rand_core 0.6.4",
@@ -11565,6 +11566,7 @@ dependencies = [
  "tempo-precompiles",
  "tempo-primitives",
  "tempo-telemetry-util",
+ "thiserror 2.0.17",
  "tokio",
  "tracing",
 ]
@@ -11605,7 +11607,9 @@ dependencies = [
 name = "tempo-contracts"
 version = "0.7.5"
 dependencies = [
- "alloy",
+ "alloy-contract",
+ "alloy-primitives",
+ "alloy-sol-types",
 ]
 
 [[package]]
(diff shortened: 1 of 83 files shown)


crates/revm/src/common.rs
```diff
@@ -84,24 +94,245 @@ impl TempoTx for Recovered<TempoTxEnvelope> {
...
+
+    fn is_static(&self) -> bool {
+        // read-only operations should always be static
+        true
+    }
+
+    fn sload(&mut self, address: Address, key: U256) -> TempoResult<U256> {
+        self.state
+            .sload(address, key)
+            .map_err(IntoTempoError::into_tempo_error)
```
klkvr: it should be fine to just map those to `Fatal` variant without an extra trait

crates/transaction-pool/src/validator.rs
```diff
@@ -262,17 +262,22 @@ where
...
-            {
-                Ok(fee_token) => fee_token,
-                Err(err) => {
-                    return TransactionValidationOutcome::Error(*transaction.hash(), Box::new(err));
-                }
-            };
+
+        let mut provider = ReadOnlyStorageProvider::from_state(&mut state_provider, spec);
+
+        let fee_token = match provider.get_fee_token(transaction.inner(), Address::ZERO, fee_payer)
```
klkvr: can we not preserve the `state_provider.get_fee_token` API here? I guess we could still have `TempoStateAccess` be the main entrypoint and just handle storage provider creation internally?

crates/revm/src/common.rs
```diff
@@ -199,85 +205,52 @@ pub trait TempoStateAccess<T> {
 
         // Ensure the currency is USD
         // load fee token account to ensure that we can load storage for it.
-        self.basic(fee_token)?;
-        Ok(self.sload(fee_token, tip20::slots::CURRENCY)? == USD_CURRENCY_SLOT_VALUE)
+        self.with_read_only_storage_ctx(spec, || {
+            TIP20Token::new(tip20::address_to_token_id_unchecked(fee_token))
+                .currency
+                .read()
+                .map(|currency| &currency == "USD")
```
klkvr: do we have a way to check string len? don't think we want to be tricked into reading a super long currency here from storage

0xrusowsky: addressed in [`a4d8b2e` (#1546)](https://github.com/tempoxyz/tempo/pull/1546/commits/a4d8b2e95bdf97e767c45411d01c11d7281609e8)

klkvr (approved):

PR #1547, opened 2025-12-16
chore: trim dependencies

klkvr:
Changes in the PR:
- In `tempo-chainspec`, `cli` implementations are feature-gated behind `cli` flag, allowing to not pull `reth-cli` without the feature
- In `tempo-contracts`, helper macro is introduced which feature-gates `rpc` implementations and thus `alloy-contract`/`alloy-provider` dependency behind `rpc` feature flag
- In `tempo-evm`, `ConfigureEngineEvm` impl is feature-gated behind `engine` flag, allowing to skip pulling many reth deps through `tempo-payload-types`

Dep count changes (without default features):
- `tempo-chainspec` 455 -> 329
- `tempo-precompiles` 514 -> 366
- `tempo-evm` 773 -> 386


crates/contracts/src/lib.rs
```diff
@@ -3,9 +3,9 @@
 #![cfg_attr(not(test), warn(unused_crate_dependencies))]
 #![cfg_attr(docsrs, feature(doc_cfg))]
 
-use alloy::primitives::{Address, address};
+use alloy_primitives::{Address, address};
 
-pub const MULTICALL_ADDRESS: Address = alloy::providers::MULTICALL3_ADDRESS;
+pub const MULTICALL_ADDRESS: Address = address!("0xcA11bde05977b3631167028862bE2a173976CA11");
```
mattsse, line 9: this is the correct one, can we copy past
```suggestion
/// Default address for the Multicall3 contract on most chains. See: <https://github.com/mds1/multicall>
pub const MULTICALL_ADDRESS: Address = address!("0xcA11bde05977b3631167028862bE2a173976CA11");
```

crates/contracts/src/lib.rs
```diff
@@ -14,43 +14,51 @@ pub const DEFAULT_7702_DELEGATE_ADDRESS: Address =
 pub const ARACHNID_CREATE2_FACTORY_ADDRESS: Address =
     address!("0x4e59b44847b379578588920cA78FbF26c0B4956C");
 
-pub mod contracts {
-    use alloy::{
-        primitives::{B256, Bytes, b256, bytes},
-        sol,
+macro_rules! sol {
```
mattsse, line 19: let's add a doc why this this exists?

Issue #1112, opened 2025-11-28
bug: transaction stalls after deploying a new TIP20 token on a fresh account using `PathUSD` as fee token


zerosnacks:
### Describe the bug

Possibly duplicate root cause of https://github.com/tempoxyz/tempo/issues/1111

Transaction after TIP20 token creation stalls forever on fresh account that has no default fee token configured (uses PathUSD)

```solidity
        vm.startBroadcast();

        ITIP20 token = ITIP20(
            StdPrecompiles.TIP20_FACTORY.createToken("testUSD", "tUSD", "USD", StdPrecompiles.PATH_USD, msg.sender)
        );

       // From here onwards:

        ITIP20RolesAuth(address(token)).grantRole(token.ISSUER_ROLE(), msg.sender);
        token.mint(msg.sender, 1_000_000 * 10 ** token.decimals());

        new Mail(token);

        vm.stopBroadcast();
```

Given the recent changes I am only able to confirm this is an issue on `rpc.devnet.tempo.xyz` but could possibly also be an issue with the testnet as reported in here: https://github.com/tempoxyz/tempo/issues/1111

What I suspect is that because a user does not have a fee token configured and they are using `PathUSD` for the first transaction, the moment they deploy a `TIP20` it will configure itself to use this as a fee token. This fails because the user has not set up the liquidity yet.

### Steps to reproduce

To reproduce: 

Clone: https://github.com/tempoxyz/tempo-foundry -> `zerosnacks/check-simulation` 
Run: `cargo install --path ./crates/forge --profile dev --force --locked`
Run: `cargo install --path ./crates/cast --profile dev --force --locked`
Run: `cast wallet new`
Run: `cast rpc tempo_fundAddress <ADDRESS> --rpc-url https://eng:zealous-mayer@rpc.devnet.tempo.xyz`
Run: `forge init -n tempo repro` && `cd repro`
Run: `export TEMPO_RPC_URL=https://eng:zealous-mayer@rpc.devnet.tempo.xyz`
Run: `forge script script/Mail.s.sol --broadcast --private-key <PRIVATE_KEY>`

### Logs

```text
[⠊] Compiling...
No files changed, compilation skipped
Script ran successfully.
Gas used: 734954

## Setting up 1 EVM.

==========================

Chain 42429

Estimated gas price: 20.000000001 gwei

Estimated total gas used for script: 1112271

Estimated amount required: 0.022245420001112271 ETH

==========================

##### 42429
[...]

PR #1549, opened 2025-12-16
fix(precompiles): limit functions that TIP20 fees apply to

0xrusowsky:
## Motivation

precompile impl to patch #1112 as per the spec change in #1337

## Solution

modify `fn get_fee_token` impl on `TempoStateAccess` to match the new spec

### DIFF
--- a/crates/revm/src/common.rs
+++ b/crates/revm/src/common.rs
@@ -152,18 +152,27 @@ pub trait TempoStateAccess<M = ()> {
         if !user_token.is_zero() {
             return Ok(user_token);
         }
-        // If tx.to() is a TIP-20 token, use that token as the fee token.
-        // Post-AllegroModerato: restricted to transfer/transferWithMemo/startReward,
-        // and for AA txs only when fee_payer == tx.origin.
-        if let Some(to) = tx.calls().next().and_then(|(kind, _)| kind.to().copied())
-            && !(spec.is_allegro_moderato() && tx.is_aa() && fee_payer != tx.caller())
-            && tx.calls().all(|(kind, input)| {
-                kind.to() == Some(&to)
-                    && (!spec.is_allegro_moderato() || is_tip20_fee_inference_call(input))
-            })
-            && self.is_valid_fee_token(to, spec)?
-        {
-            return Ok(to);
+
+        // Check if the fee can be inferred from the TIP20 token being called
+        if let Some(to) = tx.calls().next().and_then(|(kind, _)| kind.to().copied()) {
+            let can_infer_tip20 = if spec.is_allegro_moderato() {
+                // Post-AllegroModerato: restricted to transfer/transferWithMemo/startReward,
+                // and for AA txs only when fee_payer == tx.origin.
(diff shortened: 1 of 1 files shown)


crates/revm/src/common.rs
```diff
@@ -24,6 +24,19 @@ use tempo_primitives::TempoTxEnvelope;
...
 
+/// Returns true if the calldata is for a TIP-20 function that should trigger fee token inference.
+/// Only `transfer`, `transferWithMemo`, and `startReward` qualify.
+fn is_tip20_fee_inference_call(input: &[u8]) -> bool {
+    input.first_chunk::<4>().is_some_and(|&s| {
+        matches!(
+            s,
+            ITIP20::transferCall::SELECTOR
+                | ITIP20::transferWithMemoCall::SELECTOR
+                | ITIP20::startRewardCall::SELECTOR
```
legion2002, line 34: Dt there is a harm in  keeping this. But startRewardCall is currently disabled

crates/revm/src/common.rs
```diff
@@ -130,10 +143,15 @@ pub trait TempoStateAccess<T> {
...
             return Ok(stored_user_token);
         }
-
-        // If tx.to() is a TIP-20 token, use that token as the fee token
+        // If tx.to() is a TIP-20 token, use that token as the fee token.
+        // Post-AllegroModerato: restricted to transfer/transferWithMemo/startReward,
+        // and for AA txs only when fee_payer == tx.origin.
         if let Some(to) = tx.calls().next().and_then(|(kind, _)| kind.to().copied())
-            && tx.calls().all(|(kind, _)| kind.to() == Some(&to))
+            && !(spec.is_allegro_moderato() && tx.is_aa() && fee_payer != tx.caller())
```
legion2002: looks correct according to the spec in the issue

legion2002 (approved):
lgtm

crates/revm/src/common.rs
0xKitsune: `startReward` is only disabled if the `duration` arg is > 0. Otherwise, `startReward` is enabled to distribute rewards immediately.

0xrusowsky: this aligns with the spec changes in #1337

crates/revm/src/common.rs
```diff
@@ -139,10 +152,15 @@ pub trait TempoStateAccess<M = ()> {
...
+            && tx.calls().all(|(kind, input)| {
+                kind.to() == Some(&to)
+                    && (!spec.is_allegro_moderato() || is_tip20_fee_inference_call(input))
+            })
```
0xKitsune: Generally this looks right but can we make this a bit more readable? At the moment, this requires a really close read to know exactly what this is deoing.

0xrusowsky: addressed in [`e72ec82` (#1549)](https://github.com/tempoxyz/tempo/pull/1549/commits/e72ec8231f323e0be35445292d0637c73b7c325f)

PR #1551, opened 2025-12-16
feat(xtask): move generate-signing-key to consensus subcmd

Zygimantass:
This moves the `generate-signing-key` command to the main Tempo binary for operator friendliness


bin/tempo/src/main.rs
```diff
@@ -109,6 +110,10 @@ fn main() -> eyre::Result<()> {
     tempo_node::init_version_metadata();
     defaults::init_defaults();
 
+    if let Some(result) = tempo_cmd::try_run_tempo_subcommand() {
+        return result;
+    }
```
mattsse, line 115: seems fine here imo

bin/tempo/src/tempo_cmd.rs
```diff
@@ -0,0 +1,97 @@
...
+        Ok(cli) => match cli.command {
+            TempoCommand::Consensus(cmd) => match cmd.command {
+                ConsensusSubcommand::GenerateSigningKey(args) => Some(args.run()),
+                ConsensusSubcommand::CalculatePublicKey(args) => Some(args.run()),
+            },
+        },
+        Err(e) => match e.kind() {
+            ErrorKind::InvalidSubcommand => None,
+            _ => {
+                e.print().expect("failed to print error");
```
SuperFluffy: Where does `print` come from?

Also, I recommend not using `expect(<msg>)` as an error message - it panicked, so it's clear printing here failed. `expect` should say why you expected this to work. Otherwise just unwrap

mattsse (commented):
unsure about the command itself but imo we can invoke it like this, maybe we can make this easier on the rethcli side

bin/tempo/src/tempo_cmd.rs
```diff
@@ -0,0 +1,97 @@
...
+    #[command(subcommand)]
+    command: ConsensusSubcommand,
+}
+
+#[derive(Debug, Subcommand)]
+enum ConsensusSubcommand {
+    /// Generates an ed25519 signing key pair to be used in consensus.
+    GenerateSigningKey(GenerateSigningKey),
+    /// Calculates the public key from an ed25519 signing key.
+    CalculatePublicKey(CalculatePublicKey),
```
SuperFluffy, line 32:
```suggestion
    CalculateVerificationKey(CalculateVerificationKey),
```

Should use `signing <> verification` or `private <> public`. Mixing is weird

bin/tempo/src/tempo_cmd.rs
```diff
@@ -0,0 +1,97 @@
...
+    fn run(self) -> eyre::Result<()> {
+        let Self { output } = self;
+        let signing_key = PrivateKey::from_rng(&mut rand::thread_rng());
+        let validating_key = signing_key.public_key();
+        let signing_key = SigningKey::from(signing_key);
+        signing_key
+            .write_to_file(&output)
+            .wrap_err_with(|| format!("failed writing signing key to `{}`", output.display()))?;
+        println!(
+            "wrote signing key to: {}\npublic key: {validating_key}",
```
SuperFluffy: I guess I didn't do a good job at this either. ^.^

```suggestion
            "wrote signing key to: {}\nvalidating key: {validating_key}",
```

bin/tempo/src/tempo_cmd.rs
Zygimantass: ok, will change to private <> public, since thats what most operators will be familiar with. good call

SuperFluffy: Only caveat of course being that are using `--consensus.signing-key`. So the symmetry here makes sense. I let you do decide. :-)

bin/tempo/src/tempo_cmd.rs
Zygimantass: clap error

bin/tempo/Cargo.toml
```diff
@@ -16,12 +16,15 @@ workspace = true
...
 tempo-commonware-node.workspace = true
+tempo-commonware-node-config.workspace = true
 tempo-chainspec = { workspace = true, features = ["default"] }
 tempo-consensus.workspace = true
 tempo-evm.workspace = true
 tempo-faucet.workspace = true
 
+commonware-cryptography.workspace = true
 commonware-runtime = { workspace = true, features = ["external"] }
+rand.workspace = true
```
SuperFluffy: sort this alphabetically (under `futures`)

bin/tempo/src/tempo_cmd.rs
SuperFluffy: expect does not add any valuable info
```suggestion
                e.print().expect("should be able to write to STDOUT");.
```

PR #1553, opened 2025-12-16
feat: add `--pause-after-epoch N` and `--pause-export-share <>` startup flags

joshieDo:
Allows the node to exit at an epoch boundary and export the new ceremony share for the next epoch to disk.


crates/commonware-node/src/dkg/manager/export.rs
```diff
@@ -0,0 +1,44 @@
...
+use tempo_commonware_node_config::SigningShare;
+use tracing::info;
+
+/// Write the signing share to a file in hex format.
+///
+/// The exported file can be used directly as `--consensus.signing-share` for a new node.
+/// If the file already exists, this is a no-op to prevent accidental overwrites.
+pub fn write_share_to_file(share: &Share, path: &Path) -> Result<()> {
+    // Check if file exists to avoid overwriting
+    if path.exists() {
```
SuperFluffy: IMO this check should not exist - we should just blast over the target file. Otherwise you end up in a weird spot where your binary runs for a long time and then you end up not writing. I guess you could re-run the binary, but somehow this does not feel right.

Alternative suggestion: just write out the signing share on every new DKG result?

crates/commonware-node/src/dkg/manager/actor/post_allegretto.rs
```diff
@@ -235,6 +237,19 @@ where
             self.update_and_register_current_epoch_state(tx, our_outcome, block_outcome)
                 .await;
 
+            // Check if we're shutting down at this epoch boundary
+            if self.config.exit.args.exit_after_epoch == Some(block_epoch) {
+                // Stop the consensus engine we just started to prevent it from running
+                self.config
```
SuperFluffy: You don't want to even start this epoch (which you are starting a few lines up in `self.update_and_register_current_epoch_state`) - so I would move this check up there.

crates/commonware-node/src/dkg/manager/export.rs
joshieDo: overwriting it every time rn

crates/commonware-node/src/args.rs
```diff
@@ -136,11 +136,42 @@ pub struct Args {
...
+    pub pause_export_share: Option<PathBuf>,
+}
+
+/// Runtime configuration for pause behavior (includes shutdown token).
+#[derive(Clone, Debug)]
+pub struct PauseConfig {
+    /// CLI arguments for pause behavior.
+    pub args: PauseArgs,
+    /// Shutdown token to signal graceful shutdown.
+    pub shutdown_token: tokio_util::sync::CancellationToken,
```
SuperFluffy: We should remove this. Commonware's shutdown mechanism works by exiting the root context, which triggers an avalanche of shutdowns.

crates/commonware-node/src/args.rs
```diff
@@ -136,11 +136,42 @@ pub struct Args {
...
+/// Used for breaking storage migrations where nodes need to export their DKG state,
+/// shut down cleanly, and restart with a new binary.
+#[derive(Debug, Clone, Default, PartialEq, Eq, clap::Args)]
+pub struct PauseArgs {
+    /// Pause after processing the last block of this epoch. When set, the node
+    /// will exit with code 0 after the specified epoch boundary is finalized.
+    #[arg(long = "consensus.pause-after-epoch")]
+    pub pause_after_epoch: Option<u64>,
+
+    /// Path to export the BLS12-381 signing share before pausing. Requires
```
SuperFluffy, line 158: Mention that this will overwrite whatever is the target.

crates/commonware-node/src/args.rs
```diff
@@ -136,11 +136,42 @@ pub struct Args {
...
 }
 
+/// Configuration for coordinated node pause at epoch boundaries.
+///
+/// Used for breaking storage migrations where nodes need to export their DKG state,
+/// shut down cleanly, and restart with a new binary.
+#[derive(Debug, Clone, Default, PartialEq, Eq, clap::Args)]
+pub struct PauseArgs {
+    /// Pause after processing the last block of this epoch. When set, the node
+    /// will exit with code 0 after the specified epoch boundary is finalized.
```
SuperFluffy: Will it still exit?

PR #1558, opened 2025-12-16
fix: add spending limit checks

legion2002:
This PR adds spending limit checks for access keys in 4 places that were previously
   missing them:

  1. `transfer_with_memo`

  Bypass: Access key could use transferWithMemo() instead of transfer() to bypass
  spending limits since both call _transfer() internally but only transfer() had the
  check.

  2. `start_reward`

  Bypass: Access key could call startReward() to transfer tokens from the account to
  fund rewards without any spending limit deduction.

  3. `system_transfer_from`

  Bypass: Access key could call TIPFeeAMM functions (rebalanceSwap,
  mintWithValidatorToken) which use system_transfer_from() to take tokens from
  msg_sender without approval or spending limit checks.

  4. `transfer_fee_pre_tx`

  Bypass: Access key could submit transactions with very high max_fee_per_gas *
  gas_limit to escrow large amounts for fees. While unused fees are refunded, the
  actual gas consumed bypasses spending limits entirely.


crates/precompiles/src/tip20/mod.rs
```diff
@@ -684,6 +691,13 @@ impl TIP20Token {
         self.check_not_token_address(call.to)?;
         self.ensure_transfer_authorized(msg_sender, call.to)?;
 
+        // Only check access keys after Allegro Moderato hardfork
+        if self.storage.spec().is_allegro_moderato() {
+            // Check and update spending limits for access keys
+            let mut keychain = AccountKeychain::new();
+            keychain.authorize_transfer(msg_sender, self.address, call.amount)?;
```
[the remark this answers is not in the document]
legion2002: lame.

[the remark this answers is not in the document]
legion2002: 👎

crates/precompiles/src/tip20/mod.rs
```diff
@@ -872,6 +886,13 @@ impl TIP20Token {
             );
         }
 
+        // Only check access keys after Allegro Moderato hardfork
+        if self.storage.spec().is_allegro_moderato() {
+            // Check and update spending limits for access keys
+            let mut keychain = AccountKeychain::new();
+            keychain.authorize_transfer(from, self.address, amount)?;
+        }
```
0xKitsune: Note that this uses the entire `amount` but does not account for any refunded fees. Is this the intended behavior?

crates/precompiles/src/tip20/mod.rs
```diff
@@ -641,6 +641,13 @@ impl TIP20Token {
         self.check_not_token_address(to)?;
         self.ensure_transfer_authorized(from, to)?;
 
+        // Only check access keys after Allegro Moderato hardfork
+        if self.storage.spec().is_allegro_moderato() {
+            // Check and update spending limits for access keys
+            let mut keychain = AccountKeychain::new();
+            keychain.authorize_transfer(from, self.address, amount)?;
+        }
```
0xKitsune: nit: we use this same snippet in a few places, it might make sense to encapsulate this into a function like `ensure_transfer_authorized()`.

crates/precompiles/src/tip20/mod.rs
```diff
@@ -641,6 +641,13 @@ impl TIP20Token {
         self.check_not_token_address(to)?;
         self.ensure_transfer_authorized(from, to)?;
 
+        // Only check access keys after Allegro Moderato hardfork
+        if self.storage.spec().is_allegro_moderato() {
+            // Check and update spending limits for access keys
+            let mut keychain = AccountKeychain::new();
+            keychain.authorize_transfer(from, self.address, amount)?;
```
0xKitsune: Do we have any tests to assert that spending limits are enforced? If so can we update these to encorporate these additions? If not we should probably add this.

crates/precompiles/src/tip20/mod.rs
legion2002: Yes, this is intended for now. Doing spending limit gas refunds is just complexity, with no practical benefit at this point. 
Also we might remove gas refunds altogether soon.

We will just document this somewhere

0xKitsune: Seems like this could be a simple change by putting this logic in `transfer_fee_post_tx` instead to authorize_transfer for the exact spent amount while refunds are still present?

crates/precompiles/src/tip20/rewards.rs
```diff
@@ -30,6 +31,13 @@ impl TIP20Token {
             return Err(TIP20Error::invalid_amount().into());
         }
 
+        // Only check access keys after Allegro Moderato hardfork
+        if self.storage.spec().is_allegro_moderato() {
+            // Check and update spending limits for access keys
+            let mut keychain = AccountKeychain::new();
+            keychain.authorize_transfer(msg_sender, token_address, call.amount)?;
+        }
```
0xKitsune: If this is related to transfer authorization, we could consider moving this to into `ensure_transfer_authorized` and passing in the call amount. Then we don't have to repeat this logic everywhere.

crates/precompiles/src/tip20/mod.rs
legion2002: Yes this could work, the consideration there was that this only gets called at the end of the tx.
So we are letting a tx key without enough spending limits, do the whole execution, and failing at the end without paying the validator. So it's a DOS vector.

legion2002: I think to do it properly, we'd have to first run spending limit on the max_fee pre tx. 
Then refund the spending limit somehow post tx for whatever wasn't used.

crates/precompiles/src/tip20/mod.rs
legion2002: We have tests that check spending limits for normal transfers. Might be too many LOC to check it for everything we fixed here.

PR #1562, opened 2025-12-17
chore: introduce `reth` feature for `tempo-primitives`

klkvr:
Adds `reth` feature for `tempo-primitives`, allowing usage of primitive types without importing any reth code


mattsse (approved):
makes sense

PR #1565, opened 2025-12-17
fix(precompiles): spending limits enforcement only for tx.origin spends

legion2002:
## Fix spending limits incorrectly applied to contract transfers

  ### Problem

  When an EOA with an access key calls a contract that transfers tokens, the transfer
   fails. The spending limit check uses `msg_sender` (the contract) but looks up
  limits using the EOA's `transaction_key` - a key the contract never authorized.

```
  Alice (EOA + access key) → SwapContract.swap() → TIP20.transfer()
                                                    ↓
                                            msg_sender = SwapContract
                                            transaction_key = Alice's key
                                                    ↓
                                            keys[SwapContract][Alice's key] → FAILS
```

 ### Solution

  Only enforce spending limits in TIP20 when `msg_sender == tx.origin`

### Implementation Details

  Precompiles only receive msg.sender via PrecompileInput - not tx.origin. We work
  around this by:

  1. Handler sets set_tx_origin(tx.caller()) in the transient storage slot of the precompile before execution.
  2. Precompile reads `tx.origin` from transient storage.

alloy-evm has added [this](https://github.com/alloy-rs/evm/pull/246), we can bump this later to remove this transient storage hack.


crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -432,6 +432,15 @@ impl AccountKeychain {
...
         }
 
+        // Only apply spending limits if the caller is the tx origin.
+        // Gate behind allegro-moderato to avoid breaking existing behavior.
+        if self.storage.spec().is_allegro_moderato() {
+            let tx_origin = self.storage.tx_origin();
+            if account != tx_origin {
+                return Ok(());
+            }
+        }
```
0xKitsune, line 453: Should this return an error or are we purposely returning early? IIUC when calling this function its not immediately obvious to the user if the spending limits are updated.

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -466,6 +475,14 @@ impl AccountKeychain {
             return Ok(());
         }
 
+        // Only apply spending limits if the caller is the tx origin.
+        if self.storage.spec().is_allegro_moderato() {
+            let tx_origin = self.storage.tx_origin();
+            if account != tx_origin {
+                return Ok(());
+            }
+        }
```
0xKitsune, line 496: Same thing here, should this return an error or are we purposely returning early?

crates/precompiles/src/account_keychain/mod.rs
legion2002: We purposefully return early here. We don't want the call flow to continue to verifying spending limits, when the `msg.sender` of a `transfer` or `approve` call is not the tx.origin.

crates/precompiles/src/account_keychain/mod.rs
legion2002: answered above.

PR #1566, opened 2025-12-17
Prevent logging private key during DKG ceremony

### LINKED ISSUE #1561: Stop leaking private keys
### Describe the bug

We are leaking a node's private key every a new ceremony is started: https://github.com/tempoxyz/tempo/blob/ea602bd73d9057e4f166ee96a43587115a0ccfc9/crates/commonware-node/src/dkg/manager/actor/post_allegretto.rs#L442

We should prevent accidental leakage by wrapping the private key in a refinement type that does not implement `std::fmt::Display` (or better, has a display impl that emits no useful information like in https://github.com/commonwarexyz/monorepo/pull/2423 )

### Steps to reproduce

Run a node, observe that this log emits its private ed25519 signing key:
https://github.com/tempoxyz/tempo/blob/ea602bd73d9057e4f166ee96a43587115a0ccfc9/crates/commonware-node/src/dkg/manager/actor/post_allegretto.rs#L442

### Logs

```text

```

### Platform(s)

_No response_

### Container Type

Not running in a container

ea602bd73d9057e4f166ee96a43587115a0ccfc9

_No response_

### Code of Conduct

Vivek23456:
This PR prevents accidental leakage of a node’s private ed25519 key in logs during DKG ceremony startup.

Previously, the log formatted `self.config.me` using `Display`, which resulted in private key material being emitted. This change replaces it with a safe public identifier while preserving useful observability.

Fixes #1561

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -111,7 +111,7 @@ tempo-payload-builder = { path = "crates/payload/builder", default-features = fa
 tempo-payload-types = { path = "crates/payload/types", default-features = false }
 tempo-precompiles = { path = "crates/precompiles", default-features = false }
 tempo-precompiles-macros = { path = "crates/precompiles-macros" }
-tempo-primitives = { path = "crates/primitives", default-features = false }
+tempo-primitives = { path = "crates/primitives", default-features = false, features = ["reth"] }
 tempo-contracts = { path = "crates/contracts", default-features = false }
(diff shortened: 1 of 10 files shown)


SuperFluffy:
Thank you for the contribution but this is surpassed by #1759. Apologies I did not get to it sooner, this got lost while sprinting.

PR #1583, opened 2025-12-17
fix(stablecoin-exchange): consolidate rounding with explicit direction to favor protocol

fgimenez:
Closes TMPO-61
Closes TMPO-24
Closes CHAIN-207

- Added `RoundingDirection` enum and centralized `base_to_quote()` / `quote_to_base()` helper functions in `orderbook.rs`
- Refactored scattered `mulDiv` operations to use the new helpers with explicit rounding direction
- Applied consistent rounding policy (round in favor of protocol) to prevent cumulative losses

### DIFF
--- a/crates/node/tests/assets/test-genesis.json
+++ b/crates/node/tests/assets/test-genesis.json
@@ -21,6 +21,7 @@
     "terminalTotalDifficultyPassed": true,
     "adagioTime": 0,
     "allegrettoTime": 0,
+    "allegroModeratoTime": 0,
     "epochLength": 302400,
     "moderatoTime": 0,
     "depositContractAddress": "0x00000000219ab540356cbb839cbe05303d7705fa"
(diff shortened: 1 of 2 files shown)


crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -35,15 +38,13 @@ pub const TICK_SPACING: i16 = 10;
...
 }
 
 /// Calculate quote amount using ceiling division (rounds up)
-/// Post-Moderato behavior
+/// Post-Moderato behavior - favors protocol when user deposits funds
 fn calculate_quote_amount_ceil(amount: u128, tick: i16) -> Option<u128> {
-    let price = tick_to_price(tick) as u128;
-    Some(amount.checked_mul(price)?.div_ceil(PRICE_SCALE as u128))
+    base_to_quote(amount, tick, RoundingDirection::Up)
 }
```
0xKitsune: Wdyt about having a single `calculate_quote_amount()` and passing in the rounding direction?

fgimenez: good call, done

PR #1615, opened 2025-12-17
fix(revm): validate validator fee token preference to prevent user DoS

fgimenez:
Closes TMPO-37

Pre-Allegretto, when a validator configured `PATH_USD` (LINKING_USD) as their fee token preference, users without explicit fee token preferences would have this invalid token returned by `get_fee_token()`. Since `is_valid_fee_token()` rejects PATH_USD pre-Allegretto, these users' transactions would be rejected in the mempool, causing a denial of service.

Modified `get_fee_token()` in `crates/revm/src/common.rs` to validate the validator's fee token preference using `is_valid_fee_token()` before returning it. If the validator's token is invalid (e.g., PATH_USD pre-Allegretto), the system now falls back to `DEFAULT_FEE_TOKEN_PRE_ALLEGRETTO` instead.

### DIFF
--- a/crates/node/tests/assets/test-genesis.json
+++ b/crates/node/tests/assets/test-genesis.json
@@ -21,6 +21,7 @@
     "terminalTotalDifficultyPassed": true,
     "adagioTime": 0,
     "allegrettoTime": 0,
+    "allegroModeratoTime": 0,
     "epochLength": 302400,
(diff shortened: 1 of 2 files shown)


0xKitsune (approved):
Noting that this will be removed as we remove hardforking logic.

PR #1616, opened 2025-12-17
fix(precompiles): prevent TIP20 transfers to zero addr

### LINKED ISSUE #1572: bug: transfer functions allow sending to zero address
ref: `TMPO-27`
severity: `medium`
impacted files: `crates/precompiles/src/tip20/mod.rs`

## Description

The TIP-20 `transfer()` and `transferFrom()` functions do not explicitly prevent transfers to the zero address. Tokens sent to the zero address are removed from the sender's balance but they are not credited anywhere and the total supply is not decreased, resulting in accounting inconsistencies that differ from the proper burn mechanism.

Standard token implementations typically prevent transfers to the zero address to avoid accidental token loss. The ERC-20 standard does not mandate this behaviour, but it has become an established convention because:

1. Tokens "burned" via transfer maintain inflated total supply metrics, whilst proper burns decrease supply. Moreover, TIP-20 tokens may only be burned by authorised users but transfer to the zero address is otherwise permitted.
2. Users who accidentally enter a zero address as recipient permanently lose tokens with no revert or warning
3. Transfer events to zero address look identical to normal transfers, whilst burns emit distinct Burn events that indexers recognise
4. Burning should be an explicit action through the burn function, not a side effect of transfers

Likelihood is medium. Zero address transfers occur through common user errors such as uninitialised variables, copy-paste mistakes, etc. The absence of validation means nothing prevents these realistic mistakes from causing permanent token loss.

Impact is medium. Users who transfer to the zero address suffer permanent, irrecoverable token loss with no warning or revert. The total supply remains artificially inflated, creating accounting inconsistencies that affect supply metrics, analytics, price calculations, and integration assumptions. Whilst this requires user error rather than attacker exploitation, the combination of permanent loss and supply accounting corruption represents a significant operational impact.

**Recommendations:** Add an explicit check to prevent transfers to the zero address, consistent with common ERC-20 token implementations.

0xrusowsky:
## Motivation

closes #1572

## Solution

modify the recipient check to also consider address zero as invalid

### DIFF
--- a/docs/specs/src/TIP20.sol
+++ b/docs/specs/src/TIP20.sol
@@ -207,10 +207,9 @@ contract TIP20 is ITIP20, TIP20RolesAuth {
 
     modifier validRecipient(address to) {
         // Don't allow sending to the zero address not other precompiled tokens.
-        if (
-            to == address(0)
-                || (uint160(to) >> 64) == 0x20c000000000000000000000
-        ) revert InvalidRecipient();
+        if (to == address(0) || (uint160(to) >> 64) == 0x20c000000000000000000000) {
+            revert InvalidRecipient();
+        }
         _;
     }
 
--- a/crates/precompiles/src/tip20/mod.rs
+++ b/crates/precompiles/src/tip20/mod.rs
@@ -2710,7 +2710,7 @@ pub(crate) mod tests {
                 admin,
                 ITIP20::transferCall {
                     to: Address::ZERO,
-                    amount: amount,
+                    amount,
                 },
             )?;
 
@@ -2719,7 +2719,7 @@ pub(crate) mod tests {
                 ITIP20::transferFromCall {
                     from: admin,
                     to: Address::ZERO,
-                    amount: amount,
+                    amount,
                 },
             )?;
 


docs/specs/src/TIP20.sol
```diff
@@ -205,11 +205,12 @@ contract TIP20 is ITIP20, TIP20RolesAuth {
...
 
-    modifier notTokenAddress(address to) {
-        // Don't allow sending to other precompiled tokens.
-        if ((uint160(to) >> 64) == 0x20c000000000000000000000) {
-            revert InvalidRecipient();
-        }
+    modifier validRecipient(address to) {
+        // Don't allow sending to the zero address not other precompiled tokens.
+        if (
+            to == 0x0000000000000000000000000000000000000000
```
0xKitsune: This can be `address(0)` instead of `0x0000000000000000000000000000000000000000`

docs/specs/src/TIP20.sol
```diff
@@ -205,11 +205,12 @@ contract TIP20 is ITIP20, TIP20RolesAuth {
...
+                || (uint160(to) >> 64) == 0x20c000000000000000000000
+        ) revert InvalidRecipient();
```
0xKitsune: Can update tests to assert that we return `InvalidRecipient` when trying to transfer to `address(0)`?

0xKitsune (commented):
Looks like we also need this check on the `to` address when calling `transfer_from`.

0xrusowsky:
> Looks like we also need this check on the `to` address when calling `transfer_from`.

this is handled internally by `fn _transfer_from`:
```rs
    fn _transfer_from(
        &mut self,
        msg_sender: Address,
        from: Address,
        to: Address,
        amount: U256,
    ) -> Result<bool> {
        self.check_not_paused()?;
        self.check_recipient(to)?;
```

PR #1617, opened 2025-12-17
fix(stablecoin-exchange): add blacklist check to increment_balance to prevent bypass on order fill

fgimenez:
Closes TMPO-20

The previous fix (#1289) addressed `decrement_balance_or_transfer_from` to prevent blacklisted users from spending internal balances. This PR addresses the other attack vector: preventing blacklisted users from receiving tokens to their internal balance when their orders are filled.

- Add `ensure_transfer_authorized` check to `increment_balance()` (post AllegroModerato)
- Add test `test_blacklisted_maker_cannot_receive_tokens_on_order_fill` to verify blacklisted makers cannot receive tokens when orders are filled by a taker


danrobinson:
Closing this in favor of #1805 and #1794 instead

PR #1622, opened 2025-12-17
perf(orderbook): optimize tick bitmap traversal using word-level bit manipulation

fgimenez:
Closes TMPO-53

Optimizes `next_initialized_bid_tick()` and `next_initialized_ask_tick()` to use efficient bitmap word traversal instead of linear tick-by-tick search. Reduces storage reads from O(n) to O(n/256) in worst case, where n is the tick distance traversed.

- Read entire 256-bit bitmap words instead of checking individual ticks
- Use `trailing_zeros()` / `leading_zeros()` bit manipulation to find next set bit in O(1)
- Only read next/previous word when current word is exhausted
- Added tests for `next_initialized_tick` traversal behavior

### DIFF
--- a/crates/precompiles/src/stablecoin_exchange/orderbook.rs
+++ b/crates/precompiles/src/stablecoin_exchange/orderbook.rs
@@ -270,7 +270,7 @@ impl OrderbookHandler {
     }
 
     /// Find next initialized ask tick higher than current tick
-    pub fn next_initialized_tick(&self, tick: i16, is_bid: bool) -> (i16, bool) {
+    pub fn next_initialized_tick(&self, tick: i16, is_bid: bool) -> Result<(i16, bool)> {
         if is_bid {
             self.next_initialized_bid_tick(tick)
         } else {
@@ -282,10 +282,10 @@ impl OrderbookHandler {
     ///
     /// Uses efficient bitmap word traversal: reads entire 256-bit words and uses
     /// bit manipulation to find set bits, minimizing storage reads.
-    fn next_initialized_ask_tick(&self, tick: i16) -> (i16, bool) {
+    fn next_initialized_ask_tick(&self, tick: i16) -> Result<(i16, bool)> {
         // Guard against overflow when tick is at or above MAX_TICK
         if tick >= MAX_TICK {
-            return (MAX_TICK, false);
+            return Ok((MAX_TICK, false));
         }
 
         let mut next_tick = tick + 1;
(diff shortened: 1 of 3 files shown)


crates/precompiles/src/stablecoin_exchange/orderbook.rs
```diff
@@ -278,36 +278,120 @@ impl OrderbookHandler {
...
             return (MAX_TICK, false);
         }
+
         let mut next_tick = tick + 1;
-        while next_tick <= MAX_TICK {
-            if self.is_tick_initialized(next_tick, false).unwrap_or(false) {
-                return (next_tick, true);
+        if next_tick > MAX_TICK {
+            return (next_tick, false);
+        }
```
0xrusowsky: this return is unreachable as above we check `tick >= MAX_TICK`, so the check is safe to be removed

crates/precompiles/src/stablecoin_exchange/orderbook.rs
```diff
@@ -278,36 +278,120 @@ impl OrderbookHandler {
...
         if tick <= MIN_TICK {
             return (MIN_TICK, false);
         }
+
         let mut next_tick = tick - 1;
-        while next_tick >= MIN_TICK {
-            if self.is_tick_initialized(next_tick, true).unwrap_or(false) {
-                return (next_tick, true);
+        if next_tick < MIN_TICK {
+            return (next_tick, false);
```
0xrusowsky: same as above

crates/precompiles/src/stablecoin_exchange/orderbook.rs
```diff
@@ -278,36 +278,120 @@ impl OrderbookHandler {
...
+            let word_index = next_tick >> 8;
+            if word_index < min_word_index {
+                return (next_tick, false);
+            }
+
+            let bit_index = (next_tick & 0xFF) as usize;
+
+            let word = match self.bid_bitmap.at(word_index).read() {
+                Ok(w) => w,
+                Err(_) => return (next_tick, false),
```
0xrusowsky: the old code would keep iterating over all ticks (if storage reads always failed), returning an OOB tick at the very end.

now we would return the next tick, which could have implications. If we want to match the old behavior, i think that on a failed read we should fallback to `word = U256::ZERO`

0xrusowsky: alternatively we could make the fn fallible and propagate errors (which is probably the safest approach)

crates/precompiles/src/stablecoin_exchange/orderbook.rs
fgimenez: good catch, removed

crates/precompiles/src/stablecoin_exchange/orderbook.rs
fgimenez: removed

crates/precompiles/src/stablecoin_exchange/orderbook.rs
fgimenez: makes sense, i've pushed changes following the second approach here https://github.com/tempoxyz/tempo/pull/1622/commits/349b86cb00f2fe0dda7a34ca9089dc4b303eed05 ptal

PR #1623, opened 2025-12-17
feat: bump commonware [do-not-merge]

joshieDo:
* on top of https://github.com/tempoxyz/tempo/pull/1553 until it can be merged

this (or another) will become the base branch of the commonware breaking changes

right now:
* bumps CW to https://github.com/commonwarexyz/monorepo/pull/2271 - adds `set_floor` + Epoch/View data type changes

PR #1626, opened 2025-12-17
feat: allow to start node from an arbitrary epoch boundary 

joshieDo:
* follow-up to #1553
* on top of #1623 

With the node at epoch boundary and with the exported share, this allows the node to proceed after a `rm -rf datadir/consensus`

wip - needs a reth change for the test to work

### DIFF
--- a/crates/commonware-node/src/dkg/manager/actor/mod.rs
+++ b/crates/commonware-node/src/dkg/manager/actor/mod.rs
@@ -38,6 +38,7 @@ use futures::{
 };
 use prometheus_client::metrics::{counter::Counter, gauge::Gauge};
 use rand_core::CryptoRngCore;
+use reth_provider::BlockNumReader as _;
 use tempo_chainspec::hardfork::TempoHardforks as _;
 use tempo_dkg_onchain_artifacts::PublicOutcome;
 use tempo_node::TempoFullNode;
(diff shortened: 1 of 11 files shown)


crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -269,6 +270,21 @@ where
...
+                error!(
+                    "sync_floor enabled but failed to get highest block from execution provider"
+                );
+                return;
+            };
+            info!(
+                floor_height,
+                "sync_floor enabled; setting marshal floor to highest execution block"
+            );
+            self.config.marshal.clone().set_floor(floor_height).await;
```
SuperFluffy, line 285: The DKG actor is not the right place for this I think - the floor should already be set when configuring the marshal actor itself.

crates/commonware-node/src/dkg/manager/mod.rs
```diff
@@ -92,4 +92,7 @@ pub(crate) struct Config<TPeerManager> {
 
     /// Exit configuration for coordinated shutdown at epoch boundaries.
     pub(crate) exit: ExitConfig,
+
+    /// Whether to enable sync floor mode (start from highest execution block instead of genesis).
+    pub(crate) sync_floor: bool,
```
SuperFluffy, line 97: Can we pass in the exact height (or better: epoch) at which the DKG actor should start? I'd like to keep the smarts to a minimum and just have e.g.:
```suggestion
    pub(crate) start_epoch: Epoch,
```

PR #1627, opened 2025-12-17
fix(precompiles): update `changeTransferPolicyId()` to validate policy existence

0xKitsune:
Closes ~CHAIN-177~ CHAIN-175

This PR updates `changeTransferPolicyId()` to check if the newly specified policy is valid.

### DIFF
--- a/crates/node/tests/assets/test-genesis.json
+++ b/crates/node/tests/assets/test-genesis.json
@@ -20,6 +20,7 @@
     "terminalTotalDifficulty": 0,
     "terminalTotalDifficultyPassed": true,
     "adagioTime": 0,
+    "allegrettoTime": 0,
     "epochLength": 302400,
     "moderatoTime": 0,
     "depositContractAddress": "0x00000000219ab540356cbb839cbe05303d7705fa"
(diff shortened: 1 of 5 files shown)


0xrusowsky (approved):

crates/precompiles/src/tip20/mod.rs
```diff
@@ -206,6 +206,17 @@ impl TIP20Token {
         call: ITIP20::changeTransferPolicyIdCall,
     ) -> Result<()> {
         self.check_role(msg_sender, DEFAULT_ADMIN_ROLE)?;
+
+        // Validate that the policy exists (only after Allegretto hardfork)
+        if self.storage.spec().is_allegretto() {
```
0xrusowsky:
```suggestion
        if self.storage.spec().is_allegro_moderato() {
```

crates/precompiles/src/tip20/mod.rs
```diff
@@ -206,6 +206,17 @@ impl TIP20Token {
         call: ITIP20::changeTransferPolicyIdCall,
     ) -> Result<()> {
         self.check_role(msg_sender, DEFAULT_ADMIN_ROLE)?;
+
+        // Validate that the policy exists (only after Allegretto hardfork)
```
0xrusowsky:
```suggestion
        // Validate that the policy exists (only after AllegroModerato hardfork)
```

0xrusowsky (changes requested):

legion2002 (approved):
Makes sense. Will need to think about if there is some other way to set an invalid policy other than changeTransferPolicyId, but this looks good to merge.

PR #1628, opened 2025-12-18
fix(rpc): mitigate subblock tx DoS via early validator filtering and increased channel capacity

fgimenez:
Closes TMPO-15

Attackers could censor users by flooding the RPC with fake sub-block transactions targeting any validator. Since filtering occurred after channel receive, legitimate transactions were evicted from the bounded channel (capacity 100) before processing.

- Subblock transactions are now validated against this node's validator key *before* being added to the channel. Transactions targeting other validators are rejected immediately.
- Buffer size increased from 100 to 10,000 to handle legitimate high-load scenarios.
- `RecvError::Lagged` is now caught and logged with the count of dropped messages, making attacks visible.

### DIFF
--- a/crates/node/src/rpc/mod.rs
+++ b/crates/node/src/rpc/mod.rs
@@ -364,12 +364,12 @@ impl<N: FullNodeTypes<Types = TempoNode>> EthTransactions for TempoEthApi<N> {
                     Ok(tx_hash)
                 }))
             }
-            Some(_) => Either::Left(Either::Right(async move {
-                Err(EthApiError::from(RethError::msg(
+            Some(_) => Either::Left(Either::Right(futures::future::err(
+                EthApiError::from(RethError::msg(
                     "subblock transaction rejected: target validator mismatch",
                 ))
-                .into())
-            })),
+                .into(),
+            ))),
             None => Either::Right(self.inner.send_transaction(tx).map_err(Into::into)),
         }
     }


crates/node/src/rpc/mod.rs
```diff
@@ -89,23 +97,42 @@ pub struct TempoEthApi<N: FullNodeTypes<Types = TempoNode>> {
...
     }
+
+    /// Returns `true` if the given partial validator key matches this node's validator key.
+    ///
+    /// Returns `true` if no validator key is configured (accepts all subblock transactions
+    /// for backwards compatibility).
+    fn matches_validator_key(&self, partial_key: &PartialValidatorKey) -> bool {
+        self.validator_key
+            .map(|key| partial_key.matches(key.as_slice()))
+            .unwrap_or(true)
```
0xKitsune: For clarity, what is the backwards compatibility needed? Do we want nodes that are not validators to accept subblock txs?

fgimenez: yep, not needed at all, non-validators should not accept subblock txs, removed

crates/commonware-node/src/subblocks.rs
```diff
@@ -169,8 +169,23 @@ impl<TContext: Spawner + Metrics + Pacer> Actor<TContext> {
...
+                        }
+                        Err(broadcast::error::RecvError::Lagged(count)) => {
+                            warn!(
+                                lagged_count = count,
+                                "subblock transaction receiver lagged, {} messages dropped",
+                                count
+                            );
+                        }
+                        Err(broadcast::error::RecvError::Closed) => {
+                            warn!("subblock transactions channel closed unexpectedly");
```
0xKitsune: nit: should this be `err!` instead?

crates/node/src/rpc/mod.rs
```diff
@@ -325,22 +351,26 @@ impl<N: FullNodeTypes<Types = TempoNode>> EthTransactions for TempoEthApi<N> {
...
-            Either::Right(self.inner.send_transaction(tx).map_err(Into::into))
+                    Ok(tx_hash)
+                }))
+            }
+            Some(_) => Either::Left(Either::Right(async move {
+                Err(EthApiError::from(RethError::msg(
+                    "subblock transaction rejected: target validator mismatch",
+                ))
+                .into())
+            })),
```
0xKitsune: I believe we can use `future::err` here instead of an async block.

```rust
Some(_) => Either::Left(Either::Right(future::err(
    EthApiError::from(RethError::msg(
        "subblock transaction rejected: target validator mismatch",
    ))
    .into(),
))),
```

0xKitsune (approved):

crates/commonware-node/src/subblocks.rs
fgimenez: yep, done

crates/node/src/rpc/mod.rs
fgimenez: better, done

PR #1629, opened 2025-12-18
fix: multicall3 error, and add sync tests against ethereum mainnet

legion2002:
Currently the `multicall` bytecode is deployed at the `multicall3` address. 

This PR also adds tests which makes sure that the bytecode hash and storage root of these predeployed genesis contracts is always synced with ethereum mainnet.

The tests have also caught an inconsistency between the permit2 codehashes. Investigating what's the reason, and accordingly we might need to update it as well.


crates/evm/src/block.rs
```diff
@@ -532,6 +532,35 @@ where
             }
         }
 
+        // Upgrade Multicall to Multicall3 if AllegroModerato is active and bytecode is outdated
+        if self
+            .inner
+            .spec
+            .is_allegro_moderato_active_at_timestamp(block_timestamp)
+        {
```
klkvr: we could merge this `if` and the one above

PR #1630, opened 2025-12-18
fix(precompiles): reject zero public key in ValidatorConfig to prevent non-existent validators

fgimenez:
Closes TMPO-42
Closes CHAIN-192

Adds input validation to reject zero public keys in `ValidatorConfig` precompile, fixing a sentinel value invariant violation.

The `validator_exists()` function uses zero as a sentinel value to determine if a validator exists. However, `add_validator()` and `update_validator()` accepted zero public keys, creating "ghost validators" that occupy storage but are considered non-existent by existence checks.

- Added `InvalidPublicKey` error variant to rust contract bindings and solidity interface
- Added zero public key validation in both `add_validator()` and `update_validator()`
- Added tests for both cases

### DIFF
--- a/crates/contracts/src/precompiles/tip20.rs
+++ b/crates/contracts/src/precompiles/tip20.rs
@@ -98,6 +98,7 @@ crate::sol! {
         function optedInSupply() external view returns (uint128);
         function globalRewardPerToken() external view returns (uint256);
         function userRewardInfo(address account) external view returns (UserRewardInfo memory);
+        function getPendingRewards(address account) external view returns (uint128);
 
         // Events
         event Transfer(address indexed from, address indexed to, uint256 amount);
--- a/crates/precompiles/src/tip20/dispatch.rs
+++ b/crates/precompiles/src/tip20/dispatch.rs
@@ -174,6 +174,12 @@ impl Precompile for TIP20Token {
                 })
             }
 
+            ITIP20::getPendingRewardsCall::SELECTOR => {
+                view::<ITIP20::getPendingRewardsCall>(calldata, |call| {
+                    self.get_pending_rewards(call.account)
+                })
+            }
+
             // RolesAuth functions
             IRolesAuth::hasRoleCall::SELECTOR => {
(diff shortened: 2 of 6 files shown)


crates/precompiles/src/validator_config/mod.rs
```diff
@@ -124,6 +124,11 @@ impl ValidatorConfig {
         // Only owner can create validators
         self.check_owner(sender)?;
 
+        // Reject zero public key - zero is used as sentinel value for non-existence
+        if call.publicKey.is_zero() {
+            return Err(ValidatorConfigError::invalid_public_key())?;
+        }
+
```
0xrusowsky, line 128: nit: for perf reasons it is better if we perform pure checks before those that require state access (in case that we early reject). can we move it above `check_owner`?

crates/precompiles/src/validator_config/mod.rs
```diff
@@ -182,6 +187,11 @@ impl ValidatorConfig {
             return Err(ValidatorConfigError::validator_not_found())?;
         }
 
+        // Reject zero public key - zero is used as sentinel value for non-existence
+        if call.publicKey.is_zero() {
+            return Err(ValidatorConfigError::invalid_public_key())?;
+        }
+
```
0xrusowsky, line 189: same here, let's move it above `validator_exists`

0xKitsune (approved):
Lgtm, agreed with @0xrusowsky 's comments. Can we also update the foundry test suite for the validator config?

crates/precompiles/src/validator_config/mod.rs
fgimenez: ok done also for the `self.validator_exists` call

crates/precompiles/src/validator_config/mod.rs
fgimenez: yep done

fgimenez:
> Can we also update the foundry test suite for the validator config?

sure done here https://github.com/tempoxyz/tempo/pull/1630/commits/b2ad75d114ee297d99c7caed82d93bf6c236c75c

PR #1631, opened 2025-12-18
fix: prevent index overflow in `VecHandler`

0xrusowsky:
## Motivation

audit issue: `TMPO-35`

## Solution

add a bound to to the max accepted index


crates/precompiles/src/storage/types/vec.rs
```diff
@@ -286,6 +302,9 @@ where
     {
         // Read current length
         let length = self.len()?;
+        if length >= Self::max_index() {
+            return Err(TempoPrecompileError::under_overflow());
```
0xKitsune: nit: is there a better error for this?

0xrusowsky: maybe `Fatal` with custom msg?

crates/precompiles/src/storage/types/vec.rs
```diff
@@ -142,6 +142,11 @@ where
 ///     slot.write(42)?;
 /// }
 /// ```
+///
+/// # Capacity
+///
+/// Vectors have a maximum capacity of `u32::MAX * element_size` to prevent
```
fgimenez:
```suggestion
/// Vectors have a maximum capacity of `u32::MAX / element_size` to prevent
```

crates/precompiles/src/storage/types/vec.rs
```diff
@@ -278,6 +292,8 @@ where
     /// Pushes a new element to the end of the vector.
     ///
     /// Automatically increments the length and handles packing for small types.
+    ///
+    /// Returns `Err` if the vector has reached its maximum capacity.
     #[inline]
     pub fn push(&self, value: T) -> Result<()>
```
fgimenez, line 298: `push` is protected, but other length-mutating operations, like `extend`, `insert` or direct `len_slot`, may still overflow, we could add a check in `Vec<T>::store`, wdyt?

0xrusowsky: @fgimenez since this is the `VecHandler`'s impl, not all the methods in `Vec<T>` are available. I only added `push` as it seemed to be the most useful one for our purpose.

note that the insert pattern would be achieved using `at(index)` and the performing the SSTORE.

i don't see any problems with `len_slot`, but maybe i'm missing something, could u clarify that if my prev explanation doesn't resolve that concern u had?

fgimenez: sgtm, the potential overflows in `Vec<T>::store()` and `Vec<T>::delete()` (called via `VecHandler::write()` and `VecHandler::delete()` respectively) are very unlikely and can be addressed on a separate PR if needed :+1:

0xrusowsky: agreed + on top of unlikeliness, my reasoning is that we probably wouldn't be able to load the `max_capacity` amount of elements in memory

PR #1634, opened 2025-12-18
fix: ensure compiler errors when multi-slot types overflow in slot alloc

0xrusowsky:
## Motivation
audit issue: `TMPO-36`

additionally, it also fixes another bug where the `base_slot` was ignored in the first field of a struct.

## Solution

force the compiler to evaluate the full storage space taken by the type, rather than only computing the initial slot.

### DIFF
--- a/crates/precompiles/tests/storage_tests/layouts.rs
+++ b/crates/precompiles/tests/storage_tests/layouts.rs
@@ -275,3 +275,30 @@ fn test_collision_overlapping_slots_auto() {
     let (_, address) = setup_storage();
     let _layout = Layout::__new(address);
 }
+
+#[test]
+fn test_no_collision_when_using_manual_slot_with_packing() {
+    #[contract]
+    pub struct Layout {
+        a: u128, // assigned to slot 0 with 0 offset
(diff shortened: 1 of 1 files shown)


crates/precompiles-macros/src/packing.rs
```diff
@@ -136,6 +136,10 @@ pub(crate) fn gen_constants_from_ir(fields: &[LayoutField<'_>], gen_location: bo
         let ty = field.ty;
         let consts = PackingConstants::new(field.name);
         let (loc_const, (slot_const, offset_const)) = (consts.location(), consts.into_tuple());
+        let slots_to_end = quote! {
+            ::alloy::primitives::U256::from_limbs([<#ty as crate::storage::StorableType>::SLOTS as u64, 0, 0, 0])
+                .saturating_sub(alloy::primitives::U256::ONE)
```
fgimenez: nit for consistency with other usages in the file
```suggestion
                .saturating_sub(::alloy::primitives::U256::ONE)
```

fgimenez (approved):
lgtm, smol nit

PR #1637, opened 2025-12-18
chore: bump commonware to 4a5fca0

SuperFluffy:
Bumps commonware to https://github.com/commonwarexyz/monorepo/tree/4a5fca0223772b1a1627138d970ce231f876e6ff, which is the version of commonware-main right after merging https://github.com/commonwarexyz/monorepo/pull/2433.

### DIFF
--- a/crates/commonware-node-config/src/lib.rs
+++ b/crates/commonware-node-config/src/lib.rs
@@ -5,17 +5,16 @@
 
 use std::{fmt::Display, net::SocketAddr, path::Path};
 
-use commonware_codec::{DecodeExt as _, Encode as _, FixedSize, Read};
+use commonware_codec::{DecodeExt as _, Encode as _, RangeCfg, Read};
 use commonware_cryptography::{
     Signer,
-    bls12381::primitives::{
-        group::Share,
-        poly::Public,
-        variant::{MinSig, Variant},
-    },
+    bls12381::primitives::{group::Share, poly::Public, variant::MinSig},
     ed25519::{PrivateKey, PublicKey},
 };
-use commonware_utils::set::{Ordered, OrderedAssociated};
+use commonware_utils::{
+    NZU32, TryFromIterator,
+    ordered::{Map, Set},
+};
 use indexmap::IndexMap;
 use serde::{
     Deserialize,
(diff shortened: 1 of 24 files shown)


crates/commonware-node-config/src/lib.rs
```diff
@@ -134,7 +133,7 @@ impl<'de> Deserialize<'de> for Peers {
 
         let peers = deserializer.deserialize_map(PeersVisitor)?;
         Ok(Self {
-            inner: peers.into_iter().collect(),
+            inner: Map::try_from_iter(peers).expect("peers was deduped"),
```
joshieDo, line 136: maybe we should just use `from_iter_dedup` to be on the safe side?

crates/commonware-node/src/dkg/ceremony/persisted.rs
```diff
@@ -57,12 +57,19 @@ impl Read for State {
         _cfg: &Self::Cfg,
     ) -> Result<Self, commonware_codec::Error> {
         let num_players = UInt::read_cfg(buf, &())?.into();
-        let dealing = Option::<Dealing>::read_cfg(buf, &(quorum(num_players as u32) as usize))?;
+        if num_players == 0 {
```
joshieDo, line 60: nice

crates/commonware-node/src/dkg/manager/actor/pre_allegretto.rs
```diff
@@ -416,9 +415,10 @@ impl Read for EpochState {
         _cfg: &Self::Cfg,
     ) -> Result<Self, commonware_codec::Error> {
         let epoch = Epoch::read(buf)?;
-        let participants = Ordered::read_cfg(buf, &(RangeCfg::from(0..=usize::MAX), ()))?;
+        let participants = Set::read_cfg(buf, &(RangeCfg::from(0..=usize::MAX), ()))?;
+        let quorum = quorum(participants.len() as u32);
```
joshieDo: i know this is only internal, but i wonder if we could still err or just log if participants == 0

crates/commonware-node/src/dkg/manager/validators.rs
```diff
@@ -148,43 +148,38 @@ impl ValidatorState {
...
-                    public_key: key.clone(),
-                    inbound: *addr,
-                    outbound: SocketAddr::from(([0, 0, 0, 0], 0)),
-                    index: 0,
-                    address: Address::ZERO,
-                };
-                (key, validator)
-            })
-            .collect();
+    pub(super) fn with_unknown_contract_state(validators: Map<PublicKey, SocketAddr>) -> Self {
```
joshieDo, line 152:
```suggestion
    //// TODO: remove unused on post_allegretto
    pub(super) fn with_unknown_contract_state(validators: Map<PublicKey, SocketAddr>) -> Self {
```

crates/commonware-node/src/dkg/manager/actor/pre_allegretto.rs
SuperFluffy: I changed the line immediately before to ensure there is at least one participant. That way we don't need the extra check: 498745c5

crates/commonware-node/src/dkg/ceremony/persisted.rs
SuperFluffy: Yeah, this is a logic change, but I thought it wouldn't so bad to sneak this in here. :-)

crates/commonware-node-config/src/lib.rs
SuperFluffy: I have contemplated this myself and decided on the expect - it's an invariant that the map is deduped in the visitor. If we do `from_iter_dedup` then this invariant is silently ignored and it would be very confusing in practice. Figured a loud panic to notify of the violation is better.

SuperFluffy:
Superseded by #1678

PR #1651, opened 2025-12-19
fix: std `get_validator_token` to return the default token when unset

0xrusowsky:
## Motivation

audit issue: `TMPO-51`
linear: `CHAIN-198`

## Solution

default token logic to `fn get_validator_token()`

### DIFF
--- a/crates/precompiles/src/tip_fee_manager/mod.rs
+++ b/crates/precompiles/src/tip_fee_manager/mod.rs
@@ -831,7 +831,7 @@ mod tests {
 
         StorageCtx::enter(&mut storage, || {
             let balance = U256::from(1000);
-            let path_usd = TIP20Setup::path_usd(admin)
+            let _path_usd = TIP20Setup::path_usd(admin)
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/tip_fee_manager/mod.rs
```diff
@@ -474,20 +474,10 @@ impl TipFeeManager {
...
-                });
-            } else {
-                token = validator_token;
-            }
+            token = self.get_validator_token(call.validator)?;
         }
 
-        let tip20_token = TIP20Token::from_address(token)?;
-        let token_balance = tip20_token.balance_of(ITIP20::balanceOfCall {
+        let token_balance = TIP20Token::from_address(token)?.balance_of(ITIP20::balanceOfCall {
```
0xKitsune: We should unify this with `get_fee_token()` behavior. We no longer fallback to the validator token, in the case where the user token is not set, we simply fallback to pathUSD.

PR #1671, opened 2025-12-19
feat!(consensus): remove all pre-allegretto logic

SuperFluffy:
Removes all pre-allegretto logic from the consensus layer.

No logic changes were implemented other than branches related to pre- vs post-allegretto handling.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -6,7 +6,8 @@ yanked = "warn"
 ignore = [
   # https://rustsec.org/advisories/RUSTSEC-2024-0436 paste! is unmaintained
   "RUSTSEC-2024-0436",
-
+  # https://rustsec.org/advisories/RUSTSEC-2025-0137 ruint reciprocal_mg10 unsoundness. No safe upgrade available and currently unused
+  "RUSTSEC-2025-0137",
 ]
 
 # This section is considered when running `cargo deny check bans`.


crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -184,12 +177,9 @@ where
 
         let db = MetadataDatabase::new(metadata);
 
-        // Run migration from old metadata stores if needed
         {
+            // TODO: set this when opening the database/metadata for the first time?
```
joshieDo, line 181: ye, although we probably should add the small commit hash as well (another PR)

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -272,11 +238,7 @@ where
         let mut tx = DkgReadWriteTransaction::new(self.db.read_write());
 
         // Emits an error event on return.
-        if self.post_allegretto_init(&mut tx).await.is_err() {
-            return;
-        }
-        // Emits an error event on return.
-        if self.pre_allegretto_init(&mut tx).await.is_err() {
+        if self.initialize_epoch_state(&mut tx).await.is_err() {
```
joshieDo, line 241: i forget, why dont we `?`

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -711,9 +753,236 @@ where
...
+            );
+            // At this point we cannot know if the public outcome was successful
+            // or not so we don't change the our_dkg_outcome.dkg_successful.
+            //
+            // FIXME(janis): it is critical that the next set of validators and
+            // players get pushed into the DKG outcome so that the we get
+            // global agreement on these values.
+            dkg_outcome.public = canonical_dkg_outcome.public;
+            dkg_outcome.participants = canonical_dkg_outcome.participants;
+            dkg_outcome.share.take();
```
joshieDo, line 911:
```suggestion
            // TODO: look into revealed shares to see if we can heal as a last ditch effort
            dkg_outcome.share.take();
```

crates/commonware-node/src/dkg/manager/read_write_transaction.rs
```diff
@@ -14,20 +17,8 @@ fn validators_key(epoch: u64) -> String {
...
-    }
-}
-
-fn previous_epoch_key(regime: HardforkRegime) -> &'static str {
-    match regime {
-        HardforkRegime::PreAllegretto => "pre_allegretto_epoch_previous",
-        HardforkRegime::PostAllegretto => "post_allegretto_epoch_previous",
-    }
-}
+const CURRENT_EPOCH_KEY: &str = "current_epoch";
```
joshieDo, line 20: im sure we'll have forks in the future, maybe we can keep a prefix?

crates/commonware-node/src/dkg/manager/actor/mod.rs
SuperFluffy: The actor (specifically, the `dkg::manager::Actor::run` future) runs inside a long running span-free task (span-free because if we ever migrate to OTEL, then we can't have long-runnin spans). So the question is - where do you want to emit the error event?

The immediately correct choice is to emit it at the end of the span encompassing `initialize_epoch_state` (so that the error event is inside that function's span, you have the correct context, etc).

That leaves the question what to do with the error itself. We could in theory bubble it up all the way to `consensus::Engine::start`, create an ad-hoc "exit" span, and then print the error. But it would not be in the correct span - but you'd be missing context, which you'd need to dig up anyway. Further, just because your actor exited here and its error was caught first in the root of the execution stack, it doesn't mean that this actor exiting was the root cause of the system going down. You always need to look at events for that.

crates/commonware-node/src/dkg/manager/actor/mod.rs
SuperFluffy: I think this won't be a problem with the DKG rework that's coming, but I am going to add the TODO to revisit it while doing the transition.

crates/commonware-node/src/dkg/manager/actor/mod.rs
joshieDo: lets log it at least?

SuperFluffy: But we do! See the definition:
```rust
    #[instrument(skip_all, err)]
    pub(super) async fn initialize_epoch_state(
```

~~I am going to add back the comment.~~
The comment is still above the line. Was confused by the diff.

crates/commonware-node/src/dkg/manager/read_write_transaction.rs
SuperFluffy: discussed offline: we won't add a prefix because the allegretto transition was messy and is not a good template for how to do future transitions. Instead, we should have 2 separate databases, which would make this strategy redundant.

PR #1675, opened 2025-12-19
test(precompiles): add additional policy_exists tests

0xKitsune:
This PR adds additional tests to assert policy_exists functionality.

### DIFF
--- a/crates/precompiles/src/tip20/mod.rs
+++ b/crates/precompiles/src/tip20/mod.rs
@@ -2848,6 +2848,9 @@ pub(crate) mod tests {
 
             assert!(result.is_ok());
 
+            // Verify the policy ID was actually stored
+            assert_eq!(token.transfer_policy_id()?, arbitrary_policy_id);
+
             Ok(())
         })
     }
@@ -2874,11 +2877,13 @@ pub(crate) mod tests {
                 admin,
                 ITIP20::changeTransferPolicyIdCall { newPolicyId: 0 },
             )?;
+            assert_eq!(token.transfer_policy_id()?, 0);
 
             token.change_transfer_policy_id(
                 admin,
                 ITIP20::changeTransferPolicyIdCall { newPolicyId: 1 },
             )?;
+            assert_eq!(token.transfer_policy_id()?, 1);
 
             // Test random invalid policy IDs should fail
             let mut rng = rand::thread_rng();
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/tip20/mod.rs
```diff
@@ -2823,4 +2823,106 @@ pub(crate) mod tests {
...
+
+            assert!(result.is_ok());
+
+            Ok(())
+        })
+    }
+
+    #[test]
+    fn test_change_transfer_policy_id_allegro_moderato() -> eyre::Result<()> {
+        use rand::Rng;
```
fgimenez: can we move the import to the top of the module? like in the other file

crates/precompiles/src/tip20/mod.rs
```diff
@@ -2823,4 +2823,106 @@ pub(crate) mod tests {
...
+            let arbitrary_policy_id = 999u64;
+            let result = token.change_transfer_policy_id(
+                admin,
+                ITIP20::changeTransferPolicyIdCall {
+                    newPolicyId: arbitrary_policy_id,
+                },
+            );
+
+            assert!(result.is_ok());
+
```
fgimenez, line 2850: we could add here assertions for the value to be actually stored, like `assert!(token.transfer_policy_id(), arbitrary_policy_id);` wdyt?

crates/precompiles/src/tip20/mod.rs
```diff
@@ -2823,4 +2823,106 @@ pub(crate) mod tests {
...
+            let mut rng = rand::thread_rng();
+            for _ in 0..20 {
+                let invalid_policy_id = rng.gen_range(2..u64::MAX);
+                let result = token.change_transfer_policy_id(
+                    admin,
+                    ITIP20::changeTransferPolicyIdCall {
+                        newPolicyId: invalid_policy_id,
+                    },
+                );
+                assert!(result.is_err());
```
fgimenez: maybe we could assert here the error type, similar to what we do in `test_change_transfer_policy_id_invalid_policy`, wdyt?

crates/precompiles/src/tip20/mod.rs
0xKitsune: Sounds good, updated to verify `assert_eq!(token.transfer_policy_id()?, arbitrary_policy_id);` below.

PR #1678, opened 2025-12-19
chore: bump commonware to 4a5fca0

SuperFluffy:
This patch bumps commonware to https://github.com/commonwarexyz/monorepo/commit/4a5fca0223772b1a1627138d970ce231f876e6ff, which is the commit after https://github.com/commonwarexyz/monorepo/pull/2433 was merged, but before https://github.com/commonwarexyz/monorepo/pull/2439.

This patch contains no logic changes. The main changes required were:

1. `Epoch` and `View` are now refinement types. Addition/subtraction have to be done in terms of `Delta<Epoch>` and `Delta<View>` types (or their type aliases `EpochDelta` and `ViewDelta`).
2. Because `Epoch`s are refinement types, they now provide their own `commonware_codec` implementations (`Read`, `Write`, `EncodeSize`), and so we no longer have to rely on `UInt`.
3. The finalization cert and block archives are now passed constructed outside the marshal actor and passed into it. These are only genuine new code that enters the code base.
4. Polynomials now take `RangeCfg` arguments in their `commonwre_codec::Read` implementation.
5. Signature construction and verifications now always takes a byte slice argument (no longer an `Option`).
6. `Ordered` and `OrderedAssociated` are replaced by `ordered::Set` and `ordered::Map`.
7. In e2e tests, single node scenarios are currently broken and disabled. These will be addressed upstream in https://github.com/commonwarexyz/monorepo/pull/2583

### DIFF
--- a/crates/dkg-onchain-artifacts/src/lib.rs
+++ b/crates/dkg-onchain-artifacts/src/lib.rs
@@ -203,65 +203,10 @@ impl IntermediateOutcome {
         }
     }
 
-    /// Creates a new intermediate ceremony outcome.
-    ///
-    /// This method constructs a signature without the number players. This is
-    /// incorrect and addressed by [`Self::new`]. [`Self::new_pre_allegretto`]
-    /// exists for compatibility reasons and should only be used for hardforks
-    /// pre allegretto.
-    ///
-    /// This object contains, the number of players, the epoch of the ceremony,
-    /// the dealer's commitment (public polynomial), the acks received by the
-    /// players and the revealed shares for which no acks are received.
-    ///
-    /// Finally, it also includes a signature over
-    /// `(namespace, epoch, commitment, acks, reveals)` signed by the dealer.
-    pub fn new_pre_allegretto(
-        n_players: u16,
-        dealer_signer: &PrivateKey,
-        namespace: &[u8],
-        epoch: Epoch,
-        commitment: Public<MinSig>,
-        acks: Vec<Ack>,
-        reveals: Vec<group::Share>,
-    ) -> Self {
-        // Sign the resharing outcome
-        let payload =
-            Self::signature_payload_from_parts_pre_allegretto(epoch, &commitment, &acks, &reveals);
-        let dealer_signature = dealer_signer.sign(namespace, payload.as_ref());
-
-        Self {
-            n_players,
-            dealer: dealer_signer.public_key(),
-            dealer_signature,
-            epoch,
-            commitment,
-            acks,
-            reveals,
-        }
-    }
-
     /// Verifies the intermediate outcome's signature.
     pub fn verify(&self, namespace: &[u8]) -> bool {
         let payload = Self::signature_payload_from_parts(
             self.n_players,
-            self.epoch,
-            &self.commitment,
-            &self.acks,
-            &self.reveals,
-        );
-        self.dealer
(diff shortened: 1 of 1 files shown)


crates/dkg-onchain-artifacts/src/lib.rs
```diff
@@ -212,7 +251,24 @@ impl IntermediateOutcome {
...
+            .verify(namespace, &payload, &self.dealer_signature)
+    }
+
+    /// Verifies the intermediate outcome's signature.
+    ///
+    /// This method constructs a signature without the number players. This is
+    /// incorrect and addressed by [`Self::new`]. [`Self::new_pre_allegretto`]
+    /// exists for compatibility reasons and should only be used for hardforks
+    /// pre allegretto.
+    pub fn verify_pre_allegretto(&self, namespace: &[u8]) -> bool {
```
joshieDo: do we still need?

crates/dkg-onchain-artifacts/src/lib.rs
```diff
@@ -189,7 +190,45 @@ impl IntermediateOutcome {
...
+    /// exists for compatibility reasons and should only be used for hardforks
+    /// pre allegretto.
+    ///
+    /// This object contains, the number of players, the epoch of the ceremony,
+    /// the dealer's commitment (public polynomial), the acks received by the
+    /// players and the revealed shares for which no acks are received.
+    ///
+    /// Finally, it also includes a signature over
+    /// `(namespace, epoch, commitment, acks, reveals)` signed by the dealer.
+    pub fn new_pre_allegretto(
```
joshieDo: do we still need?

crates/dkg-onchain-artifacts/src/lib.rs
```diff
@@ -225,13 +281,38 @@ impl IntermediateOutcome {
...
+        buf
+    }
+
+    /// Returns the payload that was signed by the dealer, formed from raw parts.
+    ///
+    /// This method constructs a signature without the number players. This is
+    /// incorrect and addressed by [`Self::new`]. [`Self::new_pre_allegretto`]
+    /// exists for compatibility reasons and should only be used for hardforks
+    /// pre allegretto.
+    fn signature_payload_from_parts_pre_allegretto(
```
joshieDo: do we still need?

crates/evm/src/block.rs
```diff
@@ -345,7 +345,8 @@ where
                 ));
             };
 
-            if !validator.verify(None, signature_hash.as_slice(), &signature) {
+            // TODO: Add namespace?
```
joshieDo, line 348: makes sense to me? cc @klkvr

crates/dkg-onchain-artifacts/src/lib.rs
SuperFluffy: Squirreled its way back in when I `git checkout`ed from another branch.

Thank you for noticing. [32dc757](https://github.com/tempoxyz/tempo/pull/1678/commits/32dc7573bd79e7ceb4bf9a1226db000f25675aa3)

crates/dkg-onchain-artifacts/src/lib.rs
SuperFluffy: [32dc757](https://github.com/tempoxyz/tempo/pull/1678/commits/32dc7573bd79e7ceb4bf9a1226db000f25675aa3)

PR #1680, opened 2025-12-19
feat: add re-dkg trigger methods to validator precompile

joshieDo:
```rust
    /// Get the epoch at which a fresh DKG ceremony will be triggered.
    ///
    /// The fresh DKG ceremony runs in epoch N, and epoch N+1 uses the new DKG polynomial.
    pub fn get_next_dkg_ceremony(&self) -> Result<u64> {
```

```rust
    /// Set the epoch at which a fresh DKG ceremony will be triggered (owner only).
    ///
    /// Epoch N runs the ceremony, and epoch N+1 uses the new DKG polynomial.
    pub fn set_next_dkg_ceremony(
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2369,8 +2369,7 @@ dependencies = [
 [[package]]
 name = "commonware-broadcast"
 version = "0.0.64"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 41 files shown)


SuperFluffy (approved):
Looks fine to me, but I don't have approval rights.

PR #1682, opened 2025-12-19
Ability to provide consensus signing key and share through env variables

kamsz:


Zygimantass:
closing in favor of #1721

PR #1718, opened 2025-12-23
feat(tip-fee-manager): remove hardforks from fee manager and fee AMM

0xKitsune:
This PR removes hardforks from the fee manager and fee AMM.

### DIFF
--- a/crates/precompiles/src/test_util.rs
+++ b/crates/precompiles/src/test_util.rs
@@ -142,7 +142,7 @@ pub struct TIP20Setup {
     approvals: Vec<(Address, Address, U256)>,
     reward_opt_ins: Vec<Address>,
     reward_streams: Vec<(U256, u32)>,
-    preserve_events: bool,
+    clear_events: bool,
 }
 
 #[cfg(any(test, feature = "test-utils"))]
@@ -180,8 +180,10 @@ impl TIP20Setup {
     }
 
     /// Do not clear the emitted events of the token.
-    pub fn preserve_events(mut self) -> Self {
-        self.preserve_events = true;
+    ///
+    /// SAFETY: it is the caller's responsibility to ensure the test uses `HashMapStorageProvider`.
+    pub fn clear_events(mut self) -> Self {
+        self.clear_events = true;
         self
     }
 
@@ -355,7 +357,7 @@ impl TIP20Setup {
             token.start_reward(admin, ITIP20::startRewardCall { amount, secs })?;
         }
 
-        if !self.preserve_events {
+        if self.clear_events {
             token.clear_emitted_events();
         }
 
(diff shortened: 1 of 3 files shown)


crates/consensus/src/lib.rs
```diff
@@ -16,18 +16,17 @@ use reth_primitives_traits::{RecoveredBlock, SealedBlock, SealedHeader};
...
     Block, BlockBody, TempoHeader, TempoPrimitives, TempoReceipt, TempoTxEnvelope,
 };
 
 // End-of-block system transactions (required)
-// Pre AllegroModerato: 3 txs (fee manager, stablecoin exchange, subblocks signatures)
+// Pre AllegroModerato: 2 txs (stablecoin exchange, subblocks signatures)
 // Post AllegroModerato: 1 tx (subblocks signatures)
-const SYSTEM_TX_COUNT_PRE_ALLEGRO_MODERATO: usize = 3;
+const SYSTEM_TX_COUNT_PRE_ALLEGRO_MODERATO: usize = 2;
 const SYSTEM_TX_ADDRESSES_PRE_ALLEGRO_MODERATO: [Address; SYSTEM_TX_COUNT_PRE_ALLEGRO_MODERATO] = [
```
0xrusowsky: shouldn't we only keep `POST_ALLEGRO_MODERATO` stuff? also, even if we need to keep some constants, can we rename them without the HF logic?

crates/evm/src/block.rs
```diff
@@ -649,12 +620,11 @@ where
             .spec
             .is_allegro_moderato_active_at_timestamp(block_timestamp);
 
-        // Post AllegroModerato, both fee manager and stablecoin DEX system txs are no longer required
+        // Post AllegroModerato, stablecoin DEX system tx is no longer required
         let expected_seen_system_tx = !is_allegro_moderato;
```
0xrusowsky, line 568: i think i commented this in another PR too, but this logic is still hardfork-dependant

crates/payload/builder/src/lib.rs
```diff
@@ -166,31 +164,8 @@ impl<Provider: ChainSpecProvider<ChainSpec = TempoChainSpec>> TempoPayloadBuilde
         let chain_id = Some(chain_spec.chain().id());
         let mut txs = Vec::with_capacity(3);
 
-        // Build fee manager and stablecoin dex system transaction (pre-AllegroModerato only)
+        // Build stablecoin dex system transaction (pre-AllegroModerato only)
         if !chain_spec.is_allegro_moderato_active_at_timestamp(timestamp) {
```
0xrusowsky, line 124: shouldn't this also be updated?

crates/evm/src/block.rs
0xKitsune: Yeah since there is a separate PR for removing the stable dex hardfork logic, I left this in for now. Once this is merged in, we can remove this altogether.

PR #1719, opened 2025-12-23
feat(precompiles): remove hardforks from TIP20Token

0xrusowsky:
This PR removes the hardfork logic from the `TIP20Token` precompile


crates/precompiles/src/storage/thread_local.rs
```diff
@@ -265,6 +265,10 @@ impl StorageCtx {
         self.as_hashmap().clear_transient()
     }
 
+    pub fn clear_events(&mut self, address: Address) {
+        self.as_hashmap().clear_events(address);
+    }
```
0xKitsune, line 274: nit: can we add doc comments describing when this is used?

crates/precompiles/src/tip20/rewards.rs
```diff
@@ -443,17 +441,14 @@ mod tests {
...
-        let token_id = 1;
+        let amount = U256::random() % U256::from(u128::MAX);
 
         StorageCtx::enter(&mut storage, || {
-            initialize_path_usd(admin)?;
-
-            let mut token = TIP20Token::new(token_id);
-            token.initialize("Test", "TST", "USD", PATH_USD_ADDRESS, admin, Address::ZERO)?;
-            token.grant_role_internal(admin, *ISSUER_ROLE)?;
+            let mut token = TIP20Setup::create("Test", "TST", admin)
```
0xKitsune, line 447: Are we missing path usd inititalization here?

crates/precompiles/src/tip20/rewards.rs
```diff
@@ -483,38 +478,20 @@ mod tests {
...
-            let mint_amount = (U256::random().min(U256::from(u128::MAX)) % token.supply_cap()?)
-                .min(U256::from(10));
-            token.mint(
-                admin,
-                ITIP20::mintCall {
-                    to: alice,
-                    amount: mint_amount,
-                },
-            )?;
+            let mut token = TIP20Setup::create("Test", "TST", admin)
```
0xKitsune, line 485: Same here, I suspect this might be case in a few places. Before deploying an arbitrary token pathUSD needs to be deployed, or does the TIP20Setup handle this under the hood?

crates/precompiles/src/tip20/mod.rs
```diff
@@ -761,17 +680,9 @@ impl TIP20Token {
         self.next_quote_token.write(quote_token)?;
 
         // Set default values
-        if self.storage.spec().is_moderato() {
-            self.supply_cap.write(U256::from(u128::MAX))?;
-        } else {
-            self.supply_cap.write(U256::MAX)?;
-        }
+        self.supply_cap.write(U256::from(u128::MAX))?;
```
0xKitsune, line 683: Just a note for a future PR, should we update the supply cap to u128 instead of U256 since we cap it at u128?

crates/precompiles/src/tip20/mod.rs
```diff
@@ -899,26 +805,23 @@ impl TIP20Token {
...
-                    .ok_or(TempoPrecompileError::under_overflow())?;
-                self.set_opted_in_supply(
-                    opted_in_supply
-                        .try_into()
-                        .map_err(|_| TempoPrecompileError::under_overflow())?,
-                )?;
-            }
+        // Accrue rewards up to current timestamp
+        let current_timestamp = self.storage.timestamp();
+        self.accrue(current_timestamp)?;
```
0xKitsune, line 810: Noting that this should be removed as we remove all storage variables related to streaming rewards. We can do this in a subsequent PR.

crates/precompiles/src/tip20/rewards.rs
0xrusowsky: it's handled under the hood by `TIP20Setup` yes

crates/precompiles/src/tip20/mod.rs
0xrusowsky: i'm in favor of that too, i think that would be helpful

crates/precompiles/src/storage/thread_local.rs
0xrusowsky: addressed in [`b146eb0` (#1719)](https://github.com/tempoxyz/tempo/pull/1719/commits/b146eb03a6568891f635aa8b2d5505fdb887d8f4)

crates/precompiles/src/tip20/rewards.rs
0xrusowsky: as well as the factory initialization

PR #1720, opened 2025-12-23
chore: remove hardfork logic from `revm` + tx-pool

0xrusowsky:
This PR removes hardfork-related logic from the `revm` + `transaction-pool` crates

### DIFF
--- a/crates/precompiles/src/account_keychain/dispatch.rs
+++ b/crates/precompiles/src/account_keychain/dispatch.rs
@@ -58,7 +58,7 @@ impl Precompile for AccountKeychain {
                 })
             }
 
-            _ => unknown_selector(selector, self.storage.gas_used(), self.storage.spec()),
+            _ => unknown_selector(selector, self.storage.gas_used()),
         };
 
         result.map(|res| fill_precompile_output(res, &mut self.storage))
--- a/crates/precompiles/src/lib.rs
+++ b/crates/precompiles/src/lib.rs
@@ -42,7 +42,7 @@ use alloy::{
 use alloy_evm::precompiles::{DynPrecompile, PrecompilesMap};
 use revm::{
     context::CfgEnv,
-    precompile::{PrecompileError, PrecompileId, PrecompileOutput, PrecompileResult},
+    precompile::{PrecompileId, PrecompileOutput, PrecompileResult},
 };
 
 pub use tempo_contracts::precompiles::{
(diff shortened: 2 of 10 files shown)


crates/revm/src/handler.rs
```diff
@@ -1097,29 +1078,16 @@ where
...
-
             if aa_env.subblock_transaction {
-                if !cfg.spec.is_allegretto() {
-                    if tx.max_fee_per_gas() > 0 {
-                        return Err(
-                            TempoInvalidTransaction::SubblockTransactionMustHaveZeroFee.into()
-                        );
-                    }
+                if tx.max_fee_per_gas() > 0 {
+                    return Err(TempoInvalidTransaction::SubblockTransactionMustHaveZeroFee.into());
```
0xKitsune: Looks like this is pre allegretto logic?

The previous logic was:
```rust
if !cfg.spec.is_allegretto() {
                    if tx.max_fee_per_gas() > 0 {
                        return Err(
                            TempoInvalidTransaction::SubblockTransactionMustHaveZeroFee.into()
                        );
                    }
                } else if has_keychain_fields {
                    return Err(TempoInvalidTransaction::KeychainOpInSubblockTransaction.into());
                }
```

The updated logic should just be

```rust
if has_keychain_fields {
    return Err(TempoInvalidTransaction::KeychainOpInSubblockTransaction.into());
}
```

crates/revm/src/common.rs
```diff
@@ -119,7 +119,6 @@ pub trait TempoStateAccess<M = ()> {
     fn get_fee_token(
         &mut self,
         tx: impl TempoTx,
-        validator: Address,
         fee_payer: Address,
         spec: TempoHardfork,
```
0xKitsune: Do we need spec here?

crates/revm/src/handler.rs
0xrusowsky: you are right, i mixed that

0xrusowsky: addressed in [`5710bff` (#1720)](https://github.com/tempoxyz/tempo/pull/1720/commits/5710bffdcee8dac076b2294bf99b94bb643cf7a1)

crates/revm/src/common.rs
0xrusowsky: addressed in [`32da7a8` (#1720)](https://github.com/tempoxyz/tempo/pull/1720/commits/32da7a8c27ca8183cbb7c512faa926524c859659)

PR #1721, opened 2025-12-23
feat: ensure keys are encrypted on disk

### LINKED ISSUE #1713: Encrypt signing share, signing key on disk
### Describe the feature

We don't want to store the signing share and signing key in plain text. Instead, they should be stored encrypted. To decrypt them at runtime, a secret should be passed in through an environment variable.

### Additional context

tl;dr: we will get asked why the secrets are stored unencrypted/in plaintext. Partners will very likely ask us to fix that. And we very much want to do that.<br>Required to upgrade: we need to provide a bridging mechanism, where the unencrypted data is read, encrypted, and deleted. It's in the works.

joshieDo:
* Closes https://github.com/tempoxyz/tempo/issues/1713
* Adds `ChaCha20Poly1305` encryption for signing keys and shares. Keys are encrypted at rest and decrypted at runtime using secrets from environment variables.
  - **Running node**: If the corresponding env var is set, data is decrypted; otherwise read as is. [backwards compatibility + localnet ] 
  - **Generation (`devnet/genesis/signing-key`)**: Encryption is required. **ENV VARS SHOULD BE SET**
  - **Generation (`localnet`)**: Plaintext for local dev convenience.
- `TEMPO_SIGNING_KEY_SECRET` - secret for encrypting/decrypting the ed25519 signing key
- `TEMPO_SIGNING_SHARE_SECRET` - secret for encrypting/decrypting the BLS signing share


kuyziss (approved):

crates/commonware-node-config/src/encryption.rs
```diff
@@ -0,0 +1,88 @@
+//! Encryption for signing keys and shares using ChaCha20Poly1305.
+
+use chacha20poly1305::{
+    ChaCha20Poly1305, KeyInit, Nonce,
+    aead::{Aead, OsRng, rand_core::RngCore},
+};
+use eyre::{bail, eyre};
```
SuperFluffy: Let's not start dragging eyre into this.

crates/commonware-node-config/src/encryption.rs
```diff
@@ -0,0 +1,88 @@
...
+    let cipher = ChaCha20Poly1305::new_from_slice(&key).unwrap();
+    let nonce = Nonce::from_slice(nonce_bytes);
+
+    cipher
+        .decrypt(nonce, ciphertext)
+        .map_err(|e| eyre!("decryption failed: {e:?}"))
+}
+
+/// Get the signing key encryption secret from environment.
+pub fn get_signing_key_secret() -> eyre::Result<String> {
```
SuperFluffy: Just create a custom error kind for this, something like:

```rust
enum SigningKeyErrorKind {
    #[error("could not read secret; env var `{0}` not set or not unicode")]
    EnvVar(&'static str),
}
```

crates/commonware-node-config/src/encryption.rs
```diff
@@ -0,0 +1,88 @@
...
+    std::env::var(SIGNING_KEY_ENV_VAR).map_err(|_| eyre!("{SIGNING_KEY_ENV_VAR} not set"))
+}
+
+/// Get the signing share encryption secret from environment.
+pub fn get_signing_share_secret() -> eyre::Result<String> {
```
SuperFluffy: Same as with the other fn - let's stay clear of eyre. Purity wins :)

crates/commonware-node-config/src/encryption.rs
```diff
@@ -0,0 +1,88 @@
...
+    output.extend_from_slice(&nonce_bytes);
+    output.extend_from_slice(&ciphertext);
+
+    Ok(output)
+}
+
+/// Decrypt encrypted data (nonce + ciphertext).
+pub fn decrypt(data: &[u8], secret: &str) -> eyre::Result<Vec<u8>> {
+    if data.len() < NONCE_SIZE + TAG_SIZE {
+        bail!("ciphertext too short");
```
SuperFluffy: IMO don't even check this and don't report this. If it fails it fails when you try to decrypt.

crates/commonware-node-config/src/encryption.rs
```diff
@@ -0,0 +1,88 @@
...
+    let nonce_bytes = &data[..NONCE_SIZE];
+    let ciphertext = &data[NONCE_SIZE..];
+
+    let key = derive_key(secret);
+    let cipher = ChaCha20Poly1305::new_from_slice(&key).unwrap();
+    let nonce = Nonce::from_slice(nonce_bytes);
+
+    cipher
+        .decrypt(nonce, ciphertext)
+        .map_err(|e| eyre!("decryption failed: {e:?}"))
```
SuperFluffy: error variants for these plz

crates/commonware-node-config/src/encryption.rs
```diff
@@ -0,0 +1,88 @@
...
+    let key = derive_key(secret);
+    let cipher = ChaCha20Poly1305::new_from_slice(&key).unwrap();
+
+    let mut nonce_bytes = [0u8; NONCE_SIZE];
+    OsRng.fill_bytes(&mut nonce_bytes);
+    let nonce = Nonce::from_slice(&nonce_bytes);
+
+    let ciphertext = cipher
+        .encrypt(nonce, plaintext)
+        .map_err(|e| eyre!("encryption failed: {e:?}"))?;
```
SuperFluffy: please create typed error variants for these (no unwrap, no eyre)

crates/commonware-node-config/Cargo.toml
```diff
@@ -11,11 +11,14 @@ publish.workspace = true
 workspace = true
 
 [dependencies]
+blake3.workspace = true
+chacha20poly1305.workspace = true
```
SuperFluffy, line 15: Out of curiosity - any reason to use this?

SuperFluffy (commented):
Looks great! The only thing I'd like to do before merge is to have error variants. :-)

crates/commonware-node-config/src/lib.rs
```diff
@@ -88,21 +121,50 @@ impl SigningShare {
...
+    /// If `TEMPO_SIGNING_SHARE_SECRET` is set, the data will be decrypted.
     pub fn read_from_file<P: AsRef<Path>>(path: P) -> Result<Self, SigningShareError> {
         let hex = std::fs::read_to_string(path).map_err(SigningShareErrorKind::Read)?;
         Self::try_from_hex(&hex)
     }
 
+    /// Parse a signing share from hex string.
+    ///
+    /// If `TEMPO_SIGNING_SHARE_SECRET` is set, the data will be decrypted.
     pub fn try_from_hex(hex: &str) -> Result<Self, SigningShareError> {
```
SuperFluffy, line 135: Should this remain part of the public API? It does not seem to be used outside of this crate and seems pretty weird because it has side effects (reading the env var).

crates/commonware-node-config/src/lib.rs
```diff
@@ -26,22 +28,51 @@ impl SigningKey {
...
     pub fn write_to_file<P: AsRef<Path>>(&self, path: P) -> Result<(), SigningKeyError> {
-        std::fs::write(path, self.to_string()).map_err(SigningKeyErrorKind::Write)?;
+        std::fs::write(path, self.to_encrypted_string()?).map_err(SigningKeyErrorKind::Write)?;
         Ok(())
     }
 
+    /// Convert to encrypted hex string.
+    ///
+    /// Requires `TEMPO_SIGNING_KEY_SECRET` to be set.
+    pub fn to_encrypted_string(&self) -> Result<String, SigningKeyError> {
```
SuperFluffy, line 67: Weird usability in that it has side effects (reading from an env var).

crates/commonware-node-config/src/lib.rs
```diff
@@ -88,21 +121,50 @@ impl SigningShare {
...
         let inner = Share::decode(&bytes[..]).map_err(SigningShareErrorKind::Parse)?;
         Ok(Self { inner })
     }
 
+    /// Write the signing share to file (encrypted).
+    ///
+    /// Requires `TEMPO_SIGNING_SHARE_SECRET` to be set.
     pub fn write_to_file<P: AsRef<Path>>(&self, path: P) -> Result<(), SigningShareError> {
-        std::fs::write(path, self.to_string()).map_err(SigningShareErrorKind::Write)?;
+        std::fs::write(path, self.to_encrypted_string()?).map_err(SigningShareErrorKind::Write)?;
```
SuperFluffy, line 153: I think this does not work with the intended use of this functionality. The idea is that the env var is set at startup, read, and then deleted/unset. If we want to `write_to_file` during the runtime of the program, we need to have the env var be present at all times.

What if we read the env var into memory, and guard it with something like the secrecy crate? https://docs.rs/secrecy/latest/secrecy/

SuperFluffy (commented):
For writing the secrets, I think this impl does not work - see comment there. tl;dr: the env var should be deleted after startup, and so the app could not write the signing share to disk.

PR #1729, opened 2025-12-23
chore: misc drop hardfork logic

0xrusowsky:
This PR performs miscellaneous updates


crates/evm/src/block.rs
```diff
@@ -344,90 +344,76 @@ where
...
-            // Update Multicall to Multicall3 if outdated
-            let acc = db
-                .load_cache_account(MULTICALL_ADDRESS)
-                .map_err(BlockExecutionError::other)?;
+        let correct_code_hash = tempo_contracts::contracts::CREATEX_BYTECODE_HASH;
+        if acc_info.code_hash != correct_code_hash {
+            acc_info.code_hash = correct_code_hash;
+            acc_info.code = Some(Bytecode::new_legacy(
+                tempo_contracts::contracts::CREATEX_BYTECODE,
+            ));
```
0xKitsune: I believe we can remove this entirely? The correct bytecode should be deployed at genesis.

crates/precompiles/src/tip20/dispatch.rs
```diff
@@ -115,15 +115,9 @@ impl Precompile for TIP20Token {
...
-                if !self.storage.spec().is_allegretto() {
-                    return unknown_selector(selector, self.storage.gas_used());
-                }
                 view::<ITIP20::feeRecipientCall>(calldata, |_call| self.fee_recipient.read())
             }
             ITIP20::setFeeRecipientCall::SELECTOR => {
-                if !self.storage.spec().is_allegretto() {
-                    return unknown_selector(selector, self.storage.gas_used());
-                }
                 mutate_void::<ITIP20::setFeeRecipientCall>(calldata, msg_sender, |s, call| {
```
0xKitsune, line 121: Lets remove this entirely. @danrobinson mentioned that we can reintroduce fee recipient logic/variables in a future hardfork.

0xrusowsky: sounds good, will do it together with the other field deletions in the dedicated TIP20 PR https://github.com/tempoxyz/tempo/pull/1730

Issue #1696, opened 2025-12-22
Avoid extra copy in Block::read RLP decoding (zero-copy path + correct buffer advancing)


0xKarl98:
## Problem 
Block::read (block.rs, lines ~74-105) calls buf.copy_to_bytes(header.length_with_payload()) to decode a block. This forces an extra allocation/copy of the entire RLP payload even when the underlying Buf is already contiguous. The comment notes we assume the RLP header is in the first chunk() and that we don’t advance the buf cursor, but it doesn’t address the avoidable copy or payload continuity. If Buf is backed by contiguous bytes (common for Bytes/Cursor), we could decode in-place and advance without cloning the payload.

https://github.com/tempoxyz/tempo/blob/c470dd10710cfcbac0cc5888bb6d4f61a900b8ce/crates/commonware-node/src/consensus/block.rs#L74-L105

## Why it matters
- Large blocks incur a full payload copy and allocation before decoding. 
- Violates the “zero-copy” intent; adds CPU + heap pressure on hot paths.
- Cursor advancement is implicit via copy; a zero-copy variant needs explicit advance to keep the contract clear.

## Where
impl Read for Block::read_cfg (lines ~74-105):[ copy_to_bytes](https://github.com/tempoxyz/tempo/blob/c470dd10710cfcbac0cc5888bb6d4f61a900b8ce/crates/commonware-node/src/consensus/block.rs#L94
)(...) of [header.length_with_payload](https://github.com/tempoxyz/tempo/blob/c470dd10710cfcbac0cc5888bb6d4f61a900b8ce/crates/commonware-node/src/consensus/block.rs#L89
) then [decode.]
https://github.com/tempoxyz/tempo/blob/c470dd10710cfcbac0cc5888bb6d4f61a900b8ce/crates/commonware-node/src/consensus/block.rs#L74-L105

## Suggested direction
- Add a zero-copy fast path when buf.chunk().len() >= len: decode from &chunk[..len], then buf.advance(len).
- Fallback for non-contiguous buffers: retain the current copy-based path (or aggregate safely) when chunk is shorter than len.
- Make the “payload must be contiguous” expectation explicit if that is a contract; otherwise keep the dual-path for robustness.

PR #1740, opened 2025-12-24
feat: avoid extra copy in Block::read RLP decoding

0xKarl98:
This aims to resolve issue https://github.com/tempoxyz/tempo/issues/1696 , which i did :

Optimize Block::read_cfg RLP decoding by adding a zero-copy fast path when the payload is fully contained in the current buffer chunk .

Cache length_with_payload, explicitly advance the buffer on the fast path, and retain the existing copy_to_bytes fallback for non-contiguous buffers

cc @mattsse  @klkvr

### DIFF
--- a/crates/precompiles/src/tip20/mod.rs
+++ b/crates/precompiles/src/tip20/mod.rs
@@ -596,6 +596,16 @@ impl TIP20Token {
         Ok(Self::__new(address))
     }
 
+    /// Creates a TIP20Token without validating the prefix.
+    ///
+    /// # Safety
+    /// Caller must ensure `is_tip20_prefix(address)` returns true.
+    #[inline]
+    pub fn from_address_unchecked(address: Address) -> Self {
+        debug_assert!(is_tip20_prefix(address), "address must have TIP20 prefix");
+        Self::__new(address)
+    }
+
     /// Only called internally from the factory, which won't try to re-initialize a token.
     pub fn initialize(
         &mut self,
--- a/crates/revm/src/common.rs
+++ b/crates/revm/src/common.rs
@@ -194,6 +194,20 @@ pub trait TempoStateAccess<M = ()> {
         Ok(DEFAULT_FEE_TOKEN)
     }
 
+    /// Checks if the given TIP20 token has USD currency.
+    ///
+    /// IMPORTANT: Caller must ensure `fee_token` has a valid TIP20 prefix.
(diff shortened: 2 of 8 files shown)


0xKarl98:
@klkvr The CI workflow here shows Vercel stuff here , believe this isn't anything to do with my pr code ?

crates/commonware-node/src/consensus/block.rs
```diff
@@ -86,19 +86,30 @@ impl Read for Block {
...
+        let decode_block = |input: &mut &[u8]| {
+            // TODO: decode straight to a reth SealedBlock once released:
+            // https://github.com/paradigmxyz/reth/pull/18003
+            // For now relies on `Decodable for alloy_consensus::Block`.
+            alloy_rlp::Decodable::decode(input).map_err(|rlp_err| {
+                commonware_codec::Error::Wrapped("reading RLP encoded block", rlp_err.into())
+            })
+        };
+        let inner = if buf.chunk().len() >= payload_len {
+            let mut slice = &buf.chunk()[..payload_len];
```
mattsse: we should tmp store the chunk here as well

crates/commonware-node/src/consensus/block.rs
```diff
@@ -86,19 +86,30 @@ impl Read for Block {
...
-        // TODO: decode straight to a reth SealedBlock once released:
-        // https://github.com/paradigmxyz/reth/pull/18003
-        // For now relies on `Decodable for alloy_consensus::Block`.
-        let inner = alloy_rlp::Decodable::decode(&mut bytes.as_ref()).map_err(|rlp_err| {
-            commonware_codec::Error::Wrapped("reading RLP encoded block", rlp_err.into())
-        })?;
+        let decode_block = |input: &mut &[u8]| {
+            // TODO: decode straight to a reth SealedBlock once released:
+            // https://github.com/paradigmxyz/reth/pull/18003
+            // For now relies on `Decodable for alloy_consensus::Block`.
```
mattsse: I believe this comment can be removed

0xKarl98: True i was trying to propose that after these changes get approved

SuperFluffy: The referenced PR was actually intended to be this issue I think: https://github.com/paradigmxyz/reth/issues/18002

But it was closed as stale.

It'd be great to fix this upstream.

0xKarl98:
@mattsse  @klkvr  pending

mattsse (approved):
lgtm

~~pending @SuperFluffy ~~

PR #1759, opened 2025-12-25
refactor!(consensus): standardize dkg with upstream commonware

### LINKED ISSUE #1650: Write dealers, players into on chain DKG outcome
### Describe the feature

We are already writing the participants of the next epoch into the DKG outcome. These will be the dealers in the next epoch's DKG ceremony. In addition, we should also write the players into the outcome as a forced synchronization point to ensure all nodes are guaranteed to have the same view of the next epoch.

Related change: populate the p2p peer set by looking up the dealer and player pubkeys in the validator-config smart contract.

### Additional context

_No response_

### LINKED ISSUE #1672: Use ancestors stream instead of hand-rolling
### Describe the feature

Since https://github.com/commonwarexyz/monorepo/pull/2067 commonware provides `marshal::Mailbox::ancestry`. We should use that instead of our hand-rolled mechanism to fetch missing finalized blocks in DKG ceremonies.

Ancestory stream:
https://github.com/commonwarexyz/monorepo/blob/99799d61e2ae3742944d1a6c71017a4cd81cca2a/consensus/src/marshal/ingress/mailbox.rs#L255

Handrolle code:
https://github.com/tempoxyz/tempo/blob/daac535ba660ac17a8a73f7e305d1ade655b493d/crates/commonware-node/src/dkg/manager/actor/mod.rs#L305-L310

### Additional context

_No response_

SuperFluffy:
This patch refactors the DKG actor to bring it in line with the upstream commonware "reshare" example.

This rework was forced by updates commonware to revision 709d24861c4ebe0091e50337f6d2f61d1547fdb1, which is the main branch right after https://github.com/commonwarexyz/monorepo/pull/2439 was merged.

Dealer dealings, player ACKs, and their p2p communications are now no longer defined inside tempo, but employ their upstream versions.

The metadata DB API was removed in favor of journals, as in the commonware example. State management is now capable of dealing with replayed finalized blocks by checking data against an in-memory cache to avoid duplicates and maintain a consistent view. This makes atomic reads/writes unnecessary.

The DKG event loop is now also implemented in terms of a nested loop:

The outer loop iterates over epochs, while the inner loop iterates over finalized blocks.

Closes #1527
Closes #1566 
Closes #1650 
Closes #1672

### DIFF
--- a/crates/commonware-node/src/dkg/manager/actor/mod.rs
+++ b/crates/commonware-node/src/dkg/manager/actor/mod.rs
@@ -265,7 +265,7 @@ where
             .await;
 
         // TODO: emit an event with round info
-        let round = state::Round::from_previous_state(&state, &self.config.namespace);
+        let round = state::Round::from_state(&state, &self.config.namespace);
 
         let mut dealer_state = storage
             .create_dealer_for_round(
--- a/crates/commonware-node/src/dkg/manager/actor/state.rs
+++ b/crates/commonware-node/src/dkg/manager/actor/state.rs
@@ -943,7 +943,7 @@ pub(super) struct Round {
 }
 
 impl Round {
-    pub(super) fn from_previous_state(state: &State, namespace: &[u8]) -> Self {
+    pub(super) fn from_state(state: &State, namespace: &[u8]) -> Self {
         Self {
             epoch: state.epoch,
             dealers: state.dealers.keys().clone(),


xtask/src/genesis_args.rs
```diff
@@ -740,40 +750,98 @@ fn generate_consensus_config(
...
+    // .expect("must not contain duplicate keys");
+
+    // let mut validators = vec![];
+    // for (addr, (signer, share)) in peers.values().iter().zip(signers.into_iter().zip(shares)) {
+    //     validators.push(Validator {
+    //         addr: *addr,
+    //         signing_key: SigningKey::from(signer),
+    //         signing_share: SigningShare::from(share),
+    //     });
+    // }
```
joshieDo: can remove

xtask/src/genesis_args.rs
```diff
@@ -740,40 +750,98 @@ fn generate_consensus_config(
...
+
+    // Some anonymous dealer
+    let dealer_key = PrivateKey::from_rng(&mut rng);
```
joshieDo, line 755: no biggie, but the following code is dupped in other crate(s?), wondering if we can dedup somehow

crates/dkg-onchain-artifacts/src/lib.rs
```diff
@@ -1,500 +1,157 @@
...
-        let max_participants: usize = u16::MAX.into();
-        let participants = Set::read_cfg(buf, &(RangeCfg::from(1..=max_participants), ()))?;
-        let quorum = quorum(participants.len() as u32);
-        let public =
-            Public::<MinSig>::read_cfg(buf, &RangeCfg::from(NZU32!(quorum)..=NZU32!(quorum)))?;
+        let epoch = ReadExt::read(buf)?;
+        let output = Read::read_cfg(buf, &MAX_VALIDATORS)?;
+        let next_players = Read::read_cfg(
+            buf,
+            &(RangeCfg::from(1..=(MAX_VALIDATORS.get() as usize)), ()),
```
joshieDo, line 69: whats the max size that `OnchainDkgOutcome` can take ? if this is the type that goes into the header, lets just add that information to the type doc

crates/commonware-node/src/lib.rs
```diff
@@ -101,7 +100,6 @@ pub async fn run_consensus_stack(
         partition_prefix: "engine".into(),
         signer: signing_key.into_inner(),
         share,
-        delete_signing_share: config.delete_signing_share,
```
joshieDo, line 104: good

crates/dkg-onchain-artifacts/src/lib.rs
```diff
@@ -1,500 +1,157 @@
...
-
-    pub fn player(&self) -> &PublicKey {
-        &self.player
-    }
+/// The outcome of a DKG ceremony as it is written to the chain.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct OnchainDkgOutcome {
+    pub epoch: Epoch,
+    pub output: Output<MinSig, PublicKey>,
+    pub next_players: ordered::Set<PublicKey>,
```
joshieDo, line 36: maybe on write/read we could just set `next_players` to some kind of diff against the `output.players` -p1,+p2 etc

crates/commonware-node/src/dkg/manager/validators.rs
```diff
@@ -19,26 +19,25 @@ use tempo_precompiles::{
...
         attempt = _attempt,
-        for_epoch,
-        from_block = last_height_before_epoch(for_epoch, epoch_length),
+        epoch = epoch.map(tracing::field::display),
+        height = epoch.map_or(0, |epoch| last_block_in_epoch(epoch_length, epoch)),
     ),
     err
 )]
-pub(super) async fn read_from_contract(
+pub(super) async fn read_from_contract_on_epoch_boundary(
```
joshieDo, line 36: can we add to the docs here and `read_from_contract_on_epoch_boundary` that we're returning inactive ones too

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -160,444 +137,353 @@ where
...
+                format!("unable to prune storage before up until epoch `{previous}`",)
+            })?;
+        }
+
+        let all_peers = state.construct_merged_peer_set();
+        self.metrics.peers.set(all_peers.len() as i64);
+        self.config
+            .peer_manager
+            .update(state.epoch.get(), all_peers)
+            .await;
```
joshieDo, line 252: unsure on the implications right now but we're including inactive peers here as well, if a dealer/player was set to inactive in the contract, maybe we shouldnt connect to it ?

crates/commonware-node/src/dkg/manager/ingress.rs
```diff
@@ -29,31 +29,28 @@ impl Mailbox {
     ///
     /// Returns `None` if this node was not a dealer, or if the request is
     /// for a different epoch than the ceremony that's currently running.
-    pub(crate) async fn get_intermediate_dealing(
+    pub(crate) async fn get_dealer_log(
```
joshieDo, line 32: should we stick with `dealer log` instead of `intermediate dealing` nomenclature? in this method doc and other mentions

xtask/src/genesis_args.rs
SuperFluffy: All replaced in #1778 to use upstream commonware utilities.

crates/commonware-node/src/dkg/manager/ingress.rs
```diff
@@ -29,31 +29,28 @@ impl Mailbox {
...
+            .wrap_err("actor dropped channel before responding with signed dealer log")
     }
 
-    pub(crate) async fn get_public_ceremony_outcome(
+    pub(crate) async fn get_dkg_outcome(
         &self,
-        parent: (View, Digest),
-        round: Round,
-    ) -> eyre::Result<PublicOutcome> {
+        digest: Digest,
```
joshieDo, line 46: super nit: can we call these things `block_hash` or `parent_hash` instead of `digest`? i was lost on what this `digest` referred to when first glancing at it

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -864,51 +861,47 @@ async fn verify_header_extra_data(
...
         ensure!(
-            dkg_manager
-                .verify_intermediate_dealings(dealing)
-                .await
-                .wrap_err("failed request to verify DKG dealing")?,
-            "signature of intermediate DKG outcome could not be verified",
+            &dealer == proposer,
+            "proposer `{proposer}` is not the dealer `{dealer}` of the dealing \
+            in the block",
         );
```
joshieDo, line 905: while odd, is it a big deal?

crates/dkg-onchain-artifacts/src/lib.rs
SuperFluffy: That seems like a reasonable optimization - but let's do that once we have a significant number of validators so that we need to start shaving off bytes?

crates/commonware-node/src/dkg/manager/actor/mod.rs
SuperFluffy: So the "syncers" are actually set by filtering all validators read from the contract. Even though peers are toggled off on-chain,  they might still be participants for the purpose of consensus. For example, if the resharing process failed, we fall back to the dealer.

crates/commonware-node/src/consensus/application/actor.rs
SuperFluffy: Well, I think we shouldn't permit a validator that is not the proposer to include dealer logs. We originally introduced this because the old commonware DKG logic would panic if you finalized it against dealings of unknown dealers. So this patch surfaced this check here, but it's technically not new logic.

I am not sure how this is still an attack vector. I think when checking the log against the current round the log would fail (well, unless the proposer acquired another dealer's signing key...).

crates/commonware-node/src/dkg/manager/ingress.rs
SuperFluffy: I feel you. I am following commonware nomenclature, and they are using `digest` instead of `hash`. Right now, we are referring to `hash` when it comes to talking to the EL, and use `digest` at the CL level.

If I do the change, I wonder if it would be equally confusing that I talk about `hash` but sometimes it's a `B256` and sometimes wrapped in a `Digest`. :-/

crates/commonware-node/src/dkg/manager/ingress.rs
SuperFluffy: Yeah, good idea. I did it in some places, but now it's everywhere: e6421ac2

crates/commonware-node/src/dkg/manager/validators.rs
SuperFluffy: Added in e6421ac2

crates/dkg-onchain-artifacts/src/lib.rs
SuperFluffy: Added a comment in e6421ac2 - is that how you mean it?

xtask/src/genesis_args.rs
SuperFluffy: Removed e6421ac2

xtask/src/genesis_args.rs
joshieDo: oh nice

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -160,1035 +137,1023 @@ where
...
-
-        // Emits an error event on return.
-        if self.initialize_epoch_state(&mut tx).await.is_err() {
+        let Ok(initial_state) = read_initial_state_from_genesis(
+            &mut self.context,
+            &self.config.execution_node,
+            self.config.epoch_length,
+            self.config.initial_share.clone(),
+        )
+        .await
```
joshieDo: this won't work with a full node since we can't provide this historical state. we either read directly from genesis or we fetch the state from the EL last finalized block

crates/commonware-node/src/dkg/manager/actor/state.rs
```diff
@@ -1,140 +1,1092 @@
...
+#[derive(Clone, Debug)]
+pub(super) struct Round {
+    epoch: Epoch,
+    info: dkg::Info<MinSig, PublicKey>,
+    dealers: ordered::Set<PublicKey>,
+    players: ordered::Set<PublicKey>,
+}
+
+impl Round {
+    pub(super) fn from_previous_state(state: &State, namespace: &[u8]) -> Self {
```
joshieDo: nit: `from_previous_state` but this is used for current as well? `from_state` ?

crates/commonware-node/src/dkg/manager/actor/state.rs
```diff
@@ -1,140 +1,1092 @@
...
+            share = share.as_ref().map_or("<unset>", |_| "<set>"),
+        )
+        err,
+    )]
+    pub(super) fn create_dealer_for_round(
+        &mut self,
+        me: PrivateKey,
+        round: Round,
+        share: Option<Share>,
+        seed: Summary,
```
joshieDo, line 352: `summary` instead of `seed` ?  (also for `State { .. seed: }`)

joshieDo: it can be a tad confusing either way, so id say lets stick to the types name

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -160,1035 +137,1023 @@ where
...
+                self.config.me.clone(),
+                round.clone(),
+                state.share.clone(),
+                state.seed,
+            )
+            .wrap_err("unable to instantiate dealer state")?;
+
+        if dealer_state.is_some() {
+            self.metrics.how_often_dealer.inc();
         }
```
joshieDo, line 281: should we log if we are a dealer or not for this epoch ?

joshieDo: same for player afterwards

crates/commonware-node/src/dkg/manager/actor/state.rs
```diff
@@ -1,140 +1,1092 @@
...
+            .flat_map(|cache| cache.logs.iter())
+    }
+
+    /// Returns the DKG outcome for the current epoch.
+    pub(super) fn current(&self) -> State {
+        self.current.clone()
+    }
+
+    /// Appends the outcome of a DKG ceremony to state
+    pub(super) async fn append_state(&mut self, state: State) -> eyre::Result<()> {
```
joshieDo, line 98: should we clear out the cache from previous epoch(s) here? i didnt see any clean-up i think

crates/commonware-node/src/dkg/manager/actor/state.rs
```diff
@@ -1,140 +1,1092 @@
...
+                height: UInt::read(buf)?.into(),
+            }),
+            other => Err(commonware_codec::Error::InvalidEnum(other)),
+        }
+    }
+}
+
+/// Internal state for a dealer in the current round.
+pub(super) struct Dealer {
+    dealer: Option<dkg::Dealer<MinSig, PrivateKey>>,
```
joshieDo, line 837: can we add a smol doc that `None` means that this dealer has already finalized

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -160,1035 +137,1023 @@ where
...
+                }
+            }
+        };
+
+        ensure!(
+            local_output == onchain_outcome.output,
+            "the output of the local DKG ceremony does not match what is on chain;
+            something is terribly wrong; cannot recover from that; delete this
+            node and spin up a new identity",
+        );
```
joshieDo: if we're part of the players, we could just take the onchain outcome, be a player-only for this epoch and become a dealer in the next?

crates/commonware-node/src/dkg/manager/actor/mod.rs
SuperFluffy: Thank you for flagging this! In 6e6bd40e I changed `Builder::initial_state` to take a future instead, which will only ever resolve if we indeed need to populate state. This will also be a good entry point for fast sync.

crates/commonware-node/src/dkg/manager/actor/state.rs
SuperFluffy: `seed` makes sense in that it's named after how it's used: `state.seed` is used an RNG source. So it kinda does make sense?

crates/commonware-node/src/dkg/manager/actor/mod.rs
SuperFluffy: Yeah, we had that, but it got lost in the refactor. Added in 24209419

crates/commonware-node/src/dkg/manager/actor/state.rs
SuperFluffy: This logic lives in `Storage::prune`

crates/commonware-node/src/dkg/manager/actor/state.rs
SuperFluffy: Added comments on all fields explaining how they are used: 2048b621

crates/commonware-node/src/dkg/manager/actor/state.rs
```diff
@@ -1,140 +1,1102 @@
...
         Ok(Self {
-            dkg_outcome,
-            validator_state,
+            epoch: ReadExt::read(buf)?,
+            seed: ReadExt::read(buf)?,
+            output: Read::read_cfg(buf, cfg)?,
+            share: ReadExt::read(buf)?,
+            dealers: Read::read_cfg(buf, &(RangeCfg::from(1..=usize::MAX), (), ()))?,
+            players: Read::read_cfg(buf, &(RangeCfg::from(1..=usize::MAX), (), ()))?,
+            syncers: Read::read_cfg(buf, &(RangeCfg::from(1..=usize::MAX), (), ()))?,
```
joshieDo:
```suggestion
            dealers: Read::read_cfg(buf, &(RangeCfg::from(1..=u16::MAX), (), ()))?,
            players: Read::read_cfg(buf, &(RangeCfg::from(1..=u16::MAX), (), ()))?,
            syncers: Read::read_cfg(buf, &(RangeCfg::from(1..=u16::MAX), (), ()))?,
```

crates/commonware-node/src/dkg/manager/actor/state.rs
joshieDo: im blind

crates/commonware-node/src/dkg/manager/actor/mod.rs
SuperFluffy: Why actually not - I was contemplating what to do, but nothing stops us from trying to participate: will now loudly complain but still try to participate: 3289ebd8

crates/commonware-node/src/dkg/manager/actor/state.rs
SuperFluffy: 93eecb2b

crates/commonware-node/src/dkg/manager/actor/state.rs
SuperFluffy: Yeah, confusing: abf4b773

joshieDo (dismissed):

PR #1762, opened 2025-12-26
feat: bump revm v34.0.0

rakita:
Staging commit for upcoming revm bump

### DIFF
--- a/crates/e2e/src/execution_runtime.rs
+++ b/crates/e2e/src/execution_runtime.rs
@@ -726,6 +726,8 @@ pub async fn launch_execution_node<P: AsRef<Path>>(
         })
         .apply(|mut c| {
             c.network.discovery.disable_discovery = true;
+            // TODO: remove once reth state root issues are fixed
+            c.engine.state_root_fallback = true;
             c.network.trusted_peers = config
                 .trusted_peers
                 .into_iter()


crates/precompiles/src/storage/evm.rs
```diff
@@ -11,50 +12,43 @@ use crate::{error::TempoPrecompileError, storage::PrecompileStorageProvider};
...
         }
     }
 
     /// Create a new storage provider with maximum gas limit and non static context
     pub fn new_max_gas(internals: EvmInternals<'a>, cfg: &CfgEnv<TempoHardfork>) -> Self {
-        Self::new(internals, u64::MAX, cfg.chain_id, cfg.spec, false)
-    }
-
-    /// Ensures that an account is loaded.
-    pub fn ensure_loaded_account(&mut self, account: Address) -> Result<(), EvmInternalsError> {
```
rakita, line 48: There is no more need for it

crates/revm/src/handler.rs
```diff
@@ -1256,10 +1325,8 @@ where
         batch_gas.floor_gas = 0u64;
     }
 
-    // For T0, include 2D nonce gas in the validation
-    if spec.is_t0() {
```
rakita, line 1269: Revert back

crates/revm/src/handler.rs
```diff
@@ -1131,19 +1195,29 @@ pub fn calculate_aa_batch_intrinsic_gas<'a>(
...
+    gas.initial_gas += cold_account_cost * calls.len().saturating_sub(1) as u64;
 
     // 4. Authorization list costs (EIP-7702)
-    gas.initial_gas += authorization_list.len() as u64 * eip7702::PER_EMPTY_ACCOUNT_COST;
+    gas.initial_gas +=
+        authorization_list.len() as u64 * gas_params.tx_eip7702_per_empty_account_cost();
     // Add signature verification costs for each authorization
     for auth in authorization_list {
         gas.initial_gas += tempo_signature_verification_gas(auth.signature());
+        // TIP-1000: Storage pricing updates for launch
```
rakita: Remove, does not belong to this PR

crates/revm/src/handler.rs
```diff
@@ -1054,22 +1059,80 @@ where
     /// - P256 (129 bytes): 21k base + 5k for P256 verification
     /// - WebAuthn (>129 bytes): 21k base + 5k + calldata gas for variable data
     #[inline]
-    fn validate_initial_tx_gas(&self, evm: &Self::Evm) -> Result<InitialAndFloorGas, Self::Error> {
+    fn validate_initial_tx_gas(
```
rakita, line 1029: Revert this and dont change the logic from main. Just apply revms gas_prams when needed.

crates/revm/src/handler.rs
```diff
@@ -1270,11 +1337,6 @@ where
         .into());
     }
 
-    // For pre-T0, add 2D nonce gas after validation
-    if !spec.is_t0() {
```
rakita, line 1283: Revert

crates/transaction-pool/src/validator.rs
```diff
@@ -263,21 +260,13 @@ where
         };
 
         // Calculate the intrinsic gas for the AA transaction
-        let mut init_and_floor_gas =
-            calculate_aa_batch_intrinsic_gas(&aa_env, Some(tx.access_list.iter()))
-                .map_err(|_| TempoPoolTransactionError::NonZeroValue)?;
-
-        // Add 2D nonce gas if nonce_key is non-zero
-        // If tx nonce is 0, it's a new key (0 -> 1 transition), otherwise existing key
-        if !tx.nonce_key.is_zero() {
```
rakita, line 272: Revert

crates/chainspec/src/hardfork.rs
```diff
@@ -44,13 +44,20 @@ hardfork!(
         #[default]
         /// T0 hardfork (default)
         T0,
+        /// Tempo Protocol Version 1.
+        T1,
```
rakita: Remove it

.vscode/settings.json
```diff
@@ -36,5 +36,8 @@
     "**/ph.js": "plaintext",
     "**/.env": "dotenv",
     "**/.env.*": "dotenv"
+	},
+	"[rust]": {
+		"editor.defaultFormatter": "rust-lang.rust-analyzer"
 	}
```
0xKitsune, line 42: Can we add `.vscode` to the gitignore?

rakita: Added in https://github.com/tempoxyz/tempo/pull/1762/changes/2ede44e9183e35ebf4df523cf12d560b5930b9bd

crates/revm/src/handler.rs
```diff
@@ -516,71 +517,23 @@ where
...
+            let refunded_gas = apply_auth_list::<_, Self::Error>(
+                chain_id,
+                tempo_tx_env
+                    .tempo_authorization_list
+                    .iter()
+                    // T0 hardfork: skip keychain signatures in auth list processing
+                    .filter(|auth| !(spec.is_t0() && auth.signature().is_keychain())),
+                journal,
+            )
+            .unwrap();
```
0xKitsune: We should not unwrap here, lets handle the error.

crates/revm/src/handler.rs
```diff
@@ -1165,16 +1132,17 @@ pub fn calculate_aa_batch_intrinsic_gas<'a>(
 
     for call in calls {
         // 4a. Calldata gas using revm helper
-        let tokens = get_tokens_in_calldata(&call.input, true);
+        let tokens = get_tokens_in_calldata_istanbul(&call.input);
```
0xKitsune, line 1134: Does this need to be hardfork gated?

crates/evm/src/lib.rs
```diff
@@ -129,7 +129,7 @@ impl ConfigureEvm for TempoEvmConfig {
         let spec = self.chain_spec().tempo_hardfork_at(header.timestamp());
 
         Ok(EvmEnv {
-            cfg_env: cfg_env.with_spec(spec),
+            cfg_env: cfg_env.with_spec_and_mainnet_gas_params(spec),
```
0xKitsune, line 132: Do we need a hardfork here? Or should this be fine?

crates/precompiles/src/storage/evm.rs
```diff
@@ -81,15 +76,25 @@ impl<'a> PrecompileStorageProvider for EvmPrecompileStorageProvider<'a> {
...
-            .checked_sub(revm::interpreter::gas::warm_cold_cost(account.is_cold))
-            .ok_or(TempoPrecompileError::OutOfGas)?;
+        let mut account = self
+            .internals
+            .load_account_mut_skip_cold_load(address, skip_cold_load)?;
 
-        f(account.data);
+        // TODO(rakita) can be moved to the beginning of the function. Requires fork.
+        deduct_gas(&mut self.gas_remaining, WARM_STORAGE_READ_COST)?;
+
```
0xKitsune: Should `skip_cold_load` always be false? If so we should just pass it in directly. In either case we should remove `// let skip_cold_load = 0 < additional_cost;`.

```suggestion
        let mut account = self
            .internals
            .load_account_mut_skip_cold_load(address, false)?;

        // TODO(rakita) can be moved to the beginning of the function. Requires fork.
        deduct_gas(&mut self.gas_remaining, WARM_STORAGE_READ_COST)?;

```

crates/precompiles/src/storage/evm.rs
```diff
@@ -100,20 +105,25 @@ impl<'a> PrecompileStorageProvider for EvmPrecompileStorageProvider<'a> {
...
     ) -> Result<(), TempoPrecompileError> {
-        self.ensure_loaded_account(address)?;
-        let result = self.internals.sstore(address, key, value)?;
+        //let skip_cold_load = 0 < self.gas_params.cold_storage_additional_cost();
+        let skip_cold_load = false;
+
+        let result =
+            self.internals
+                .load_account_mut(address)?
+                .sstore(key, value, skip_cold_load)?;
```
0xKitsune: Same thing here, should this always be false? If so we should just pass this in directly. In either case we should remove `// let skip_cold_load = 0 < self.gas_params.cold_storage_additional_cost();`



```suggestion
        let result =
            self.internals
                .load_account_mut(address)?
                .sstore(key, value, false)?;

```

crates/precompiles/src/lib.rs
```diff
@@ -111,86 +115,72 @@ macro_rules! tempo_precompile {
             let mut storage = crate::storage::evm::EvmPrecompileStorageProvider::new(
                 $input.internals,
                 $input.gas,
-                $chain_id,
-                $spec,
+                spec,
                 $input.is_static,
+                gas_params.clone(),
```
0xrusowsky, line 120: we shouldn't have to clone here as we already do `let gas_params = $cfg.gas_params.clone();` above

crates/precompiles/src/lib.rs
```diff
@@ -61,33 +61,35 @@ pub const INPUT_PER_WORD_COST: u64 = 6;
...
 }
 
 pub trait Precompile {
     fn call(&mut self, calldata: &[u8], msg_sender: Address) -> PrecompileResult;
 }
 
 pub fn extend_tempo_precompiles(precompiles: &mut PrecompilesMap, cfg: &CfgEnv<TempoHardfork>) {
-    let chain_id = cfg.chain_id;
-    let spec = cfg.spec;
+    let cfg = cfg.clone();
```
0xrusowsky, line 74: can't we use the ref? we clone `GasParams` inside the `fn create()`

crates/revm/src/handler.rs
```diff
@@ -1165,16 +1132,17 @@ pub fn calculate_aa_batch_intrinsic_gas<'a>(
...
 
         // 4b. CREATE-specific costs
         if call.to.is_create() {
             // CREATE costs 32000 additional gas
             gas.initial_gas += CREATE; // 32000 gas
 
-            // EIP-3860: Initcode analysis gas using revm helper
-            gas.initial_gas += initcode_cost(call.input.len());
+            // EIP-3860: Initcode analysis gas (2 gas per 32-byte word)
+            let num_words = call.input.len().div_ceil(32);
```
legion2002, line 1143: Was this helper intentionally removed to update the gas costs for CREATEing bytcode?

crates/revm/src/handler.rs
```diff
@@ -753,68 +706,74 @@ where
...
-                let access_key_addr = key_auth.key_id;
-
-                // Convert signature type to precompile SignatureType enum
-                // Use the key_type field which specifies the type of key being authorized
-                let signature_type = match key_auth.key_type {
-                    SignatureType::Secp256k1 => PrecompileSignatureType::Secp256k1,
-                    SignatureType::P256 => PrecompileSignatureType::P256,
-                    SignatureType::WebAuthn => PrecompileSignatureType::WebAuthn,
-                };
+            StorageCtx::enter_precompile(
```
legion2002, line 708: These AA tx related changes, look correct to me, just cleanup with some new helper functions.

0xrusowsky (approved):
besides the cmnts left by @0xKitsune everything else make sense to me.

i've paid special attention to the `precompiles` crate and they look good.

crates/e2e/src/tests/subblocks.rs
```diff
@@ -67,7 +67,7 @@ fn subblocks_are_included() {
             .engine_events
             .new_listener();
 
-        let mut expected_transactions: Vec<TxHash> = Vec::new();
+        let mut pending_transactions: Vec<TxHash> = Vec::new();
```
legion2002: Not sure why these subblock test changes are included in this PR? 
Is this related to the revm changes?

crates/e2e/src/tests/subblocks.rs
mattsse, line 1: all of this seems unrelated
can we undo this?

crates/primitives/Cargo.toml
```diff
@@ -18,6 +18,9 @@ reth-primitives-traits = { workspace = true, optional = true }
 reth-codecs = { workspace = true, optional = true }
 reth-rpc-convert = { workspace = true, optional = true }
 
+# revm
+revm.workspace = true
```
mattsse, line 22: this makes the crate slightly heavier and the trait can be pulled in from another revm dep as well

but i also dont know why we need to impl the trait

crates/e2e/src/tests/subblocks.rs
0xKitsune: +1 on this, I have pushed a commit to remove changes to subblocks tests.

crates/e2e/src/tests/subblocks.rs
0xKitsune: +1, I updated to remove these changes from the PR.

crates/precompiles/src/storage/evm.rs
0xKitsune: Why is this not needed anymore?

crates/precompiles/src/storage/evm.rs
```diff
@@ -125,17 +130,19 @@ impl<'a> PrecompileStorageProvider for EvmPrecompileStorageProvider<'a> {
...
 
     #[inline]
     fn emit_event(&mut self, address: Address, event: LogData) -> Result<(), TempoPrecompileError> {
         self.deduct_gas(
-            revm::interpreter::gas::log_cost(event.topics().len() as u8, event.data.len() as u64)
-                .unwrap_or(u64::MAX),
+            gas::LOG
+                + self
+                    .gas_params
+                    .log_cost(event.topics().len() as u8, event.data.len() as u64),
```
0xKitsune, line 145: Is the calculation here different? Here we are adding `gas::LOG` where as before this was not here.

crates/precompiles/src/storage/evm.rs
```diff
@@ -100,20 +103,22 @@ impl<'a> PrecompileStorageProvider for EvmPrecompileStorageProvider<'a> {
         key: U256,
         value: U256,
     ) -> Result<(), TempoPrecompileError> {
-        self.ensure_loaded_account(address)?;
-        let result = self.internals.sstore(address, key, value)?;
+        let result = self
+            .internals
+            .load_account_mut(address)?
+            .sstore(key, value, false)?;
```
0xKitsune, line 110: Are you sure we do not need `self.ensure_loaded_account(address)?`? Why is this safe to remove? One suspicion as to why the tests are failing is that the account storage is not being persisted.

crates/precompiles/src/storage/evm.rs
rakita: As the account gets loaded when you do the sload/sstore, API got improved.

crates/revm/src/handler.rs
rakita: No this is fine. That `true` is not more visable sa `_instanbul`

crates/evm/src/lib.rs
rakita: This is fine for now, gas_params changes are introduces in T1 so mainnet gas_parmas is fine for this PR

crates/precompiles/src/storage/evm.rs
rakita: @0xKitsune this code would align the behaviour that we can see in mainnet, where skip_cold_gas is used, but this would require a fork in tempol. Would say low prio

crates/precompiles/src/lib.rs
rakita: cfg needs to be moved inside closure so clone i needed her.

crates/precompiles/src/lib.rs
rakita: This is needed as first clone moves the values inside closure while second clone is needed as to not move gas_params from that closure.

We could add `&'a GasParams` in `EvmPrecompileStorageProvider` (As it already has a lifetime). But cloning gas_params is cheap (It is a arc) so it is fine like this.

crates/revm/src/handler.rs
rakita: All helper function were removed or moved to GasParams. It is going to be used inside TIP-1000 PR where gas params should be passed to it, as there is a `gas_params.initcode_cost` function

crates/primitives/Cargo.toml
rakita: `reth-rpc-convert` already includes reth-evm/alloy-evm and with it revm, so it is the same size

crates/precompiles/src/storage/evm.rs
rakita: It is same logic. From evm.codes we can see formula is:
```
static_gas = 375
dynamic_gas = 375 * topic_count + 8 * size + memory_expansion_cost
```

gas::LOG is static part of 375gas.

log_cost is dynamic part and it [does](https://github.com/bluealloy/revm/blob/50405a27f57f80a96ab9d91e93180bca22217a3a/crates/context/interface/src/cfg/gas_params.rs#L510).

crates/precompiles/src/storage/evm.rs
rakita: `ensure_loaded_account` does two things. loads account and does the `touch` (aka account is changed).

loading of account is done in `load_account_mut` and touching of account is always done in sstore [here](https://github.com/bluealloy/revm/blob/50405a27f57f80a96ab9d91e93180bca22217a3a/crates/context/interface/src/journaled_state/account.rs#L237-L238) 

This looks correct.

0xKitsune (approved):

Issue #1764, opened 2025-12-26
Optimize allocation in get_pooled_transaction_elements by pre‑allocating capacity


0xKarl98:
## Problem 
In [get_pooled_transaction_elements](https://github.com/tempoxyz/tempo/blob/7f34c562086cd5789ba65b3241a880472db86f81/crates/transaction-pool/src/tempo_pool.rs#L352-L371), we first build txs via filter_map + collect::<Vec<_>>() and then extend(...) with results from protocol_pool. Because filter_map produces an iterator with an imprecise size_hint, collect can’t pre‑allocate effectively, and the subsequent extend may trigger another reallocation and memcpy.

## Why it matters
- filter_map makes the iterator’s size unknown, so collect often grows Vec incrementally.
- extend may then reallocate and copy existing elements.
- With large tx_hashes, this adds avoidable allocations and memcpy overhead.

## Where 
In [get_pooled_transaction_elements](https://github.com/tempoxyz/tempo/blob/7f34c562086cd5789ba65b3241a880472db86f81/crates/transaction-pool/src/tempo_pool.rs#L352-L371), the AA‑2D branch builds txs via collect, then appends the protocol pool results.

## Suggested direction

- Pre‑allocate using a known upper bound: Vec::with_capacity(tx_hashes.len()).
- Push AA‑2D results directly into txs, then reserve(protocol_txs.len()) (or extend after reserving) before appending protocol‑pool results.

PR #1766, opened 2025-12-26
chore: optimize allocation in get_pooled_transaction_elements by pre‑allocating capacity

0xKarl98:
This aims at resolve issue #1764

cc @mattsse

### DIFF
--- a/crates/precompiles/src/tip20/mod.rs
+++ b/crates/precompiles/src/tip20/mod.rs
@@ -596,6 +596,16 @@ impl TIP20Token {
         Ok(Self::__new(address))
     }
 
+    /// Creates a TIP20Token without validating the prefix.
+    ///
+    /// # Safety
+    /// Caller must ensure `is_tip20_prefix(address)` returns true.
+    #[inline]
+    pub fn from_address_unchecked(address: Address) -> Self {
+        debug_assert!(is_tip20_prefix(address), "address must have TIP20 prefix");
+        Self::__new(address)
+    }
+
     /// Only called internally from the factory, which won't try to re-initialize a token.
     pub fn initialize(
         &mut self,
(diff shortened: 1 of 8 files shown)


crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -354,17 +354,12 @@ where
...
-            .read()
-            .get_all_iter(&tx_hashes)
-            .filter_map(|tx| {
-                tx.transaction
-                    .clone_into_pooled()
-                    .ok()
-                    .map(|tx| tx.into_inner())
-            })
-            .collect::<Vec<_>>();
+        let mut txs = Vec::with_capacity(tx_hashes.len());
```
mattsse: this could over allocate

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -354,17 +354,12 @@ where
...
+        for tx in self.aa_2d_pool.read().get_all_iter(&tx_hashes) {
+            // Safe unwrap: TryFromConsensusError is Infallible for TempoPooledTransaction.
+            let pooled = tx.transaction.clone_into_pooled().unwrap();
+            txs.push(pooled.into_inner());
+        }
```
mattsse: instead of this we could simple flip it and first obtain a vec from 

```
self.protocol_pool
                .get_pooled_transaction_elements
```

and then extend that

if you want you can also submit a pr that adds `fn append_pooled_transaction_elements` variant for this:

https://github.com/paradigmxyz/reth/blob/8ae7a1c8d14025243919d9c05d3505ef414f6937/crates/transaction-pool/src/pool/mod.rs#L402-L411

similar to 

https://github.com/paradigmxyz/reth/blob/8ae7a1c8d14025243919d9c05d3505ef414f6937/crates/transaction-pool/src/pool/mod.rs#L321-L325

PR #1768, opened 2025-12-26
fix(stablecoin-exchange): add MAX_ORDER_ITERATIONS limit to prevent DoS via unbounded order iteration

fgimenez:
Closes CHAIN-322
Closes TMPO-50

Added a `MAX_ORDER_ITERATIONS` constant (100) that limits the number of orders that can be processed in a single swap. When exceeded, the swap fails with a new `MaxOrderIterationsExceeded` error, preventing gas exhaustion attacks.

 - Added `MAX_ORDER_ITERATIONS = 100` constant
- Added `MaxOrderIterationsExceeded(uint32)` error to Solidity interface
- Added iteration counter and limit check to both fill functions
- Added tests verifying the iteration limit behavior
- Updated solidity reference implementation

### DIFF
--- a/crates/precompiles/src/stablecoin_exchange/mod.rs
+++ b/crates/precompiles/src/stablecoin_exchange/mod.rs
@@ -35,6 +35,11 @@ pub const MIN_ORDER_AMOUNT: u128 = 10_000_000;
 /// Allowed tick spacing for order placement
 pub const TICK_SPACING: i16 = 10;
 
+/// Maximum number of orders that can be iterated through in a single swap.
+/// This prevents DoS attacks where an attacker places many small orders to force
+/// expensive iteration through the orderbook.
+pub const MAX_ORDER_ITERATIONS: u32 = 100;
+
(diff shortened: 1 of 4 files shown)


crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -994,8 +999,20 @@ impl StablecoinExchange {
...
+            iterations = iterations
+                .checked_add(1)
+                .ok_or(TempoPrecompileError::under_overflow())?;
+
+            if iterations > MAX_ORDER_ITERATIONS {
+                return Err(StablecoinExchangeError::max_order_iterations_exceeded(
+                    MAX_ORDER_ITERATIONS,
+                )
+                .into());
+            }
```
0xKitsune, line 1014: Doesn't this introduce an attack vector where a user could place 101 very small orders on a frequently used pair, causing swaps to fail due to `MAX_ORDER_ITERATIONS` being exceeded?

PR #1773, opened 2025-12-29
chore: remove DAA mechanism

legion2002:
As decided, DAA is no longer needed after the Tempo Tx features. 
This PR removes DAA from all parts of the codebase, for the new testnet.

Already merged here  #1732


legion2002:
Already merged here https://github.com/tempoxyz/tempo/pull/1732

PR #1774, opened 2025-12-29
fix(executor): propagate fork choice update errors 

fgimenez:
Closes CHAIN-177
Closes TMPO-23

Fork choice update errors in the executor were silently discarded using `let _ = ...`, which could cause consensus-execution layer divergence when the execution layer rejects canonical chain updates. With this changes, when a fork choice update fails, the executor now shuts down cleanly instead of continuing with stale state.

- `handle_message()` now returns `Result<()>` and propagates errors
- `finalize()` now returns `Result<()>` and propagates errors
- `run()` loop breaks on error with a warning log
- `forward_finalized()` propagates canonicalize and fill_holes errors

### DIFF
--- a/crates/commonware-node/src/consensus/application/executor.rs
+++ b/crates/commonware-node/src/consensus/application/executor.rs
@@ -23,7 +23,9 @@ use futures::{
 };
 use reth_provider::BlockNumReader as _;
 use tempo_node::{TempoExecutionData, TempoFullNode};
-use tracing::{Level, Span, debug, error, info, info_span, instrument, warn, warn_span};
+use tracing::{
+    Level, Span, debug, error, error_span, info, info_span, instrument, warn, warn_span,
+};
 
 use crate::consensus::{Digest, block::Block};
 
@@ -243,11 +245,11 @@ where
                     // Backfills will be spawned as tasks and will also send
                     // resolved the blocks to this queue.
                     if let Err(error) = self.handle_message(msg).await {
-                        error!(
+                        error_span!("shutdown").in_scope(|| error!(
                             %error,
                             "executor encountered fatal fork choice update error; \
                             shutting down to prevent consensus-execution divergence"
-                        );
+                        ));
                         break;
                     }
                 },


0xKitsune (commented):
Lgtm pending @SuperFluffy , @joshieDo

crates/commonware-node/src/consensus/application/executor.rs
```diff
@@ -267,27 +274,30 @@ where
         Ok(())
     }
 
-    #[instrument(parent = &cause, skip_all)]
+    #[instrument(parent = &cause, skip_all, err)]
```
SuperFluffy: Remove. We are reporting the error already as part of `forward_finalized` and `canonicalize`

crates/commonware-node/src/consensus/application/executor.rs
```diff
@@ -362,7 +365,9 @@ where
                 "hole detected; consensus attempts to finalize block with gaps \
                 on the execution layer; filling them in first",
             );
-            let _ = self.fill_holes(execution_height + 1, block.height()).await;
+            self.fill_holes(execution_height + 1, block.height())
+                .await
+                .wrap_err("failed to fill holes during block finalization")?;
```
SuperFluffy: Why early return here? `fill_holes` emits error events and I am not sure it should cause the entire node to abort.

crates/commonware-node/src/consensus/application/executor.rs
```diff
@@ -325,22 +335,15 @@ where
...
-        }
+        self.canonicalize(
+            Span::current(),
+            HeadOrFinalized::Finalized,
+            block.height(),
+            block.digest(),
+            oneshot::channel().0,
+        )
+        .await
+        .wrap_err("failed canonicalizing finalized block")?;
```
SuperFluffy, line 422: Probably fine to just exit here considering we can't handle this. Finalizations must go through.

crates/commonware-node/src/consensus/application/executor.rs
```diff
@@ -267,27 +274,30 @@ where
...
-                    )
-                    .await;
+                self.canonicalize(
+                    Span::current(),
+                    HeadOrFinalized::Finalized,
+                    height,
+                    digest,
+                    oneshot::channel().0,
+                )
+                .await?;
```
SuperFluffy: If you propagate errors, provide context.

crates/commonware-node/src/consensus/application/executor.rs
```diff
@@ -267,27 +274,30 @@ where
...
             }
             Update::Block(block, acknowledgment) => {
-                let _: Result<_, _> = self
-                    .forward_finalized(Span::current(), block, acknowledgment)
-                    .await;
+                self.forward_finalized(Span::current(), block, acknowledgment)
+                    .await?;
```
SuperFluffy: Context

crates/commonware-node/src/consensus/application/executor.rs
```diff
@@ -175,28 +175,35 @@ where
...
             Command::CanonicalizeHead {
                 height,
                 digest,
                 ack,
             } => {
-                let _ = self
-                    .canonicalize(cause, HeadOrFinalized::Head, height, digest, ack)
-                    .await;
+                self.canonicalize(cause, HeadOrFinalized::Head, height, digest, ack)
+                    .await?;
```
SuperFluffy: context

crates/commonware-node/src/consensus/application/executor.rs
```diff
@@ -175,28 +175,35 @@ where
...
             }
             Command::Finalize(finalized) => {
-                let _ = self.finalize(cause, *finalized).await;
+                self.finalize(cause, *finalized).await?;
```
SuperFluffy: context

crates/commonware-node/src/consensus/application/executor.rs
```diff
@@ -175,28 +175,35 @@ where
                     //
                     // Backfills will be spawned as tasks and will also send
                     // resolved the blocks to this queue.
-                    self.handle_message(msg).await;
+                    if let Err(error) = self.handle_message(msg).await {
+                        warn!(
```
SuperFluffy: this should be `error!` if we can't handle the message and subsequently shut down.

crates/commonware-node/src/consensus/application/executor.rs
SuperFluffy: On a second thought, this should not kill the node. Failed canonicalization of finalized hashes are one thing, but this is just for the head hash. Shouldn't we be able to recover from this?

crates/commonware-node/src/consensus/application/executor.rs
fgimenez: sure done

crates/commonware-node/src/consensus/application/executor.rs
fgimenez: makes sense, removed

crates/commonware-node/src/consensus/application/executor.rs
fgimenez: sure done

crates/commonware-node/src/consensus/application/executor.rs
fgimenez: ok, made non-fatal

crates/commonware-node/src/consensus/application/executor.rs
fgimenez: done

crates/commonware-node/src/consensus/application/executor.rs
fgimenez: yep done

crates/commonware-node/src/consensus/application/executor.rs
fgimenez: done

joshieDo (approved):

crates/commonware-node/src/consensus/application/executor.rs
```diff
@@ -242,28 +242,41 @@ where
...
                 digest,
                 ack,
             } => {
-                let _ = self
+                if let Err(error) = self
                     .canonicalize(cause, HeadOrFinalized::Head, height, digest, ack)
-                    .await;
+                    .await
+                {
+                    warn!(%error, "failed canonicalizing head; will retry on next block");
```
SuperFluffy: Remove: `canonicalize` already emits an error event (and the event here is both outside of a root span and also does not carry any of the extra information of the `canonicalize` event to get more information).

crates/commonware-node/src/consensus/application/executor.rs
```diff
@@ -242,28 +242,41 @@ where
                     //
                     // Backfills will be spawned as tasks and will also send
                     // resolved the blocks to this queue.
-                    self.handle_message(msg).await;
+                    if let Err(error) = self.handle_message(msg).await {
+                        error!(
```
SuperFluffy: This event is outside of a root span - would cause issues when doing OTEL:

```suggestion
                        error_span!("shutdown").in_scope(||
                            error!(
```

SuperFluffy (approved):
looks good. Please address the last 2 comments before merge

crates/commonware-node/src/consensus/application/executor.rs
fgimenez: sure, also added a comment about why the error is intentionally discarded

crates/commonware-node/src/consensus/application/executor.rs
fgimenez: sure done

PR #1775, opened 2025-12-29
chore: fee manager dead code

legion2002:
Removes dead code and tests, after removing system transactions in fee manager.

### DIFF
--- a/crates/precompiles/src/tip_fee_manager/amm.rs
+++ b/crates/precompiles/src/tip_fee_manager/amm.rs
@@ -499,7 +499,6 @@ pub fn sqrt(x: U256) -> U256 {
 
 #[cfg(test)]
 mod tests {
-    use tempo_chainspec::hardfork::TempoHardfork;
     use tempo_contracts::precompiles::TIP20Error;
 
     use super::*;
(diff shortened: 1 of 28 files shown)


crates/precompiles/src/tip_fee_manager/amm.rs
```diff
@@ -699,242 +588,6 @@ mod tests {
         })
     }
 
-    #[test]
-    fn test_fee_swap() -> eyre::Result<()> {
```
0xKitsune, line 702: Can we rewrite these tests to include the updated fee swap functions?

crates/precompiles/src/tip_fee_manager/mod.rs
```diff
@@ -15,38 +15,17 @@ pub use tempo_contracts::precompiles::{
...
 
 #[contract(addr = TIP_FEE_MANAGER_ADDRESS)]
 pub struct TipFeeManager {
     validator_tokens: Mapping<Address, Address>,
     user_tokens: Mapping<Address, Address>,
     collected_fees: Mapping<Address, U256>,
-    tokens_with_fees: Vec<Address>,
-    token_in_fees_array: Mapping<Address, bool>,
     pools: Mapping<B256, Pool>,
     pending_fee_swap_in: Mapping<B256, u128>,
```
0xKitsune: Also, do we need pending_fee_swap_in anymore?

legion2002: idts, removing.

0xrusowsky:
@legion2002 looks like there are multiple integration tests failing

legion2002:
Fixing the failing tests, it is because of the storage layout changing, the test-genesis.json needs to be updated.

PR #1778, opened 2025-12-29
chore!: bump commonware to v0.0.64

SuperFluffy:
Updates commonware to v0.0.64.

There are no functional changes, but a plethora of new types and refactors due to changing Commonware APIs. Highlights:

1. The static `epoch_length` was replaced with a `FixedEpocher`, which is now passed to all our actors. All epoch utilities were also removed.
2. Rate limiters are now all operating at the p2p channel level and so were removed from application logic (we only used them in the epoch manager actor)
3. `application::Mailbox` also needs to implement `CertifiableAutomaton` (no other code changes though)
4. public polynomials are now called `Sharing`.
5. a few traits are now imported from the new `commonware_math` crate.
6. we can now make use of `dkg::deal` and `dkg::deal_anonymous`, simplifying test code and xtask

NOTE: this is marked as breaking because in between PR #1759 and this patch commonware had a few other changes going into their 0.0.64 release. For example, https://github.com/commonwarexyz/monorepo/pull/2555

### DIFF
--- a/crates/commonware-node/src/dkg/manager/actor/mod.rs
+++ b/crates/commonware-node/src/dkg/manager/actor/mod.rs
@@ -560,20 +560,20 @@ where
         TStorageContext: commonware_runtime::Metrics + commonware_runtime::Storage,
         TSender: Sender<PublicKey = PublicKey>,
     {
-        let bounds = self
+        let epoch_info = self
             .config
             .epoch_strategy
             .containing(block.height())
             .expect("epoch strategy is covering all block heights");
 
-        let block_epoch = bounds.epoch();
+        let block_epoch = epoch_info.epoch();
 
         if block_epoch != round.epoch() {
             info!("block was not for this epoch");
             return Ok(None);
         }
 
-        if block.height() == bounds.first()
+        if block.height() == epoch_info.first()
             && let Some(epoch) = round.epoch().previous()
         {
             self.config
@@ -582,7 +582,7 @@ where
                 .await;
         }
 
-        match bounds.phase() {
+        match epoch_info.phase() {
             EpochPhase::Early => {
                 if let Some(dealer_state) = dealer_state {
                     self.distribute_shares(
@@ -602,7 +602,7 @@ where
             }
         }
 
-        if block.height() != bounds.last() {
+        if block.height() != epoch_info.last() {
             if !block.header().extra_data().is_empty() {
                 'handle_log: {
                     let (dealer, log) =
--- a/crates/commonware-node/src/dkg/manager/actor/state.rs
+++ b/crates/commonware-node/src/dkg/manager/actor/state.rs
@@ -11,7 +11,7 @@ use commonware_cryptography::{
     Signer as _,
     bls12381::{
         dkg::{self, DealerPrivMsg, DealerPubMsg, Info, Output, PlayerAck, SignedDealerLog},
-        primitives::{group::Share, variant::MinSig},
+        primitives::{group::Share, sharing::Mode, variant::MinSig},
     },
     ed25519::{PrivateKey, PublicKey},
     transcript::{Summary, Transcript},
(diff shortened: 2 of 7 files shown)


crates/dkg-onchain-artifacts/src/lib.rs
```diff
@@ -45,7 +45,7 @@ impl OnchainDkgOutcome {
         &self.next_players
     }
 
-    pub fn public(&self) -> &Public<MinSig> {
+    pub fn public(&self) -> &Sharing<MinSig> {
```
joshieDo: should we rename as well to `sharing`

crates/e2e/src/tests/linkage.rs
```diff
@@ -25,6 +25,7 @@ fn only_good_links() {
             // // We should rename this to marshal (the actor, that is).
             if metric.ends_with("_marshal_processed_height") {
                 let value = value.parse::<u64>().unwrap();
+                println!("{metric} {value}");
```
joshieDo:
```suggestion
```

joshieDo: dont mind it, im guessing it was for debugging

xtask/src/genesis_args.rs
```diff
@@ -751,65 +752,30 @@ fn generate_consensus_config(
...
-    player_keys.sort_by_key(|key| key.public_key());
-    let info = Info::new(
-        b"tempo-genesis",
-        0,
-        None,
-        ordered::Set::try_from_iter(std::iter::once(dealer_key.public_key())).unwrap(),
-        ordered::Set::try_from_iter(player_keys.iter().map(|key| key.public_key())).unwrap(),
+    let (output, shares) = dkg::deal(
+        &mut rng,
+        Default::default(),
```
joshieDo: super nit: we could be explicit instead of `Default::default()`, just in case...

crates/commonware-node/src/dkg/manager/actor/state.rs
```diff
@@ -94,6 +95,15 @@ where
         self.current.clone()
     }
 
+    /// Returns the DKG outcome for the current epoch.
```
joshieDo:
```suggestion
    /// Returns the DKG outcome for the previous epoch.
```

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -540,13 +560,20 @@ where
         TStorageContext: commonware_runtime::Metrics + commonware_runtime::Storage,
         TSender: Sender<PublicKey = PublicKey>,
     {
-        let block_epoch = utils::epoch(self.config.epoch_length, block.height());
+        let bounds = self
```
joshieDo:
```suggestion
        let epoch_info = self
```

joshieDo (approved):

crates/commonware-node/src/dkg/manager/actor/mod.rs
SuperFluffy: b8de3949

crates/commonware-node/src/dkg/manager/actor/state.rs
SuperFluffy: b8de3949

xtask/src/genesis_args.rs
SuperFluffy: b8de3949

crates/e2e/src/tests/linkage.rs
SuperFluffy: b8de3949

crates/dkg-onchain-artifacts/src/lib.rs
SuperFluffy: b8de3949

PR #1779, opened 2025-12-29
feat: remove hardforks and cleanup remaining logic

0xKitsune:
This PR removes `TempoHardfork` variants and cleans up misc remaining hardfork logic.

### DIFF
--- a/crates/chainspec/src/spec.rs
+++ b/crates/chainspec/src/spec.rs
@@ -301,6 +301,9 @@ mod tests {
         // Should always return Genesis
         assert_eq!(chainspec.tempo_hardfork_at(0), TempoHardfork::Genesis);
         assert_eq!(chainspec.tempo_hardfork_at(1000), TempoHardfork::Genesis);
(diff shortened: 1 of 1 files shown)


xtask/src/genesis_args.rs
```diff
@@ -23,7 +23,7 @@ use std::{
     net::SocketAddr,
```
fgimenez, line 23: there's still a stale post-allegreto comment at line 178

crates/precompiles/src/stablecoin_exchange/dispatch.rs
```diff
@@ -541,7 +540,7 @@ mod tests {
 
     #[test]
     fn stablecoin_exchange_test_selector_coverage() -> eyre::Result<()> {
-        let mut storage = HashMapStorageProvider::new(1).with_spec(TempoHardfork::AllegroModerato);
```
fgimenez, line 544: do we need to keep the `with_spec` method? i assume it will be used for future hardforks, right?

0xKitsune: Yeah I left this in for future hardforks.

PR #1780, opened 2025-12-29
chore: add keychain precompile to genesis

legion2002:
We used to initialize this in block pre execution, because of hardforking logic. 
We can directly add it to the genesis file now.

### DIFF
--- a/xtask/src/genesis_args.rs
+++ b/xtask/src/genesis_args.rs
@@ -34,7 +34,8 @@ use tempo_contracts::{
 use tempo_dkg_onchain_artifacts::PublicOutcome;
 use tempo_evm::evm::{TempoEvm, TempoEvmFactory};
 use tempo_precompiles::{
-    ACCOUNT_KEYCHAIN_ADDRESS, PATH_USD_ADDRESS,
+    PATH_USD_ADDRESS,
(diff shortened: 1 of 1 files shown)


xtask/src/genesis_args.rs
```diff
@@ -343,6 +343,16 @@ impl GenesisArgs {
...
 
+        // Account Keychain precompile - uses 0xef as marker code
+        genesis_alloc.insert(
+            ACCOUNT_KEYCHAIN_ADDRESS,
+            GenesisAccount {
+                code: Some(Bytes::from_static(&[0xef])),
+                nonce: Some(1),
+                ..Default::default()
+            },
+        );
```
0xKitsune: Can we intialize this like we do with the rest of the precomplies? Eg.

```rust
/// Initializes the TIP20Factory contract (should be called once before creating any tokens)
fn initialize_tip20_factory(evm: &mut TempoEvm<CacheDB<EmptyDB>>) -> eyre::Result<()> {
    let ctx = evm.ctx_mut();
    StorageCtx::enter_evm(&mut ctx.journaled_state, &ctx.block, &ctx.cfg, || {
        TIP20Factory::new().initialize()
    })?;
    Ok(())
}
```

PR #1788, opened 2025-12-30
chore: drop `Multicall` + cleanup stale comments and tests

0xrusowsky:
This PR:
- completely drops `Multicall` so that we only work with `Multicall3` going forward (which requires a new genesis file)
- simplifies tests + re-adds some deleted tests with their checks updated to reflect the new post-hardfork-cleanup logic
- updates stale cmnts


xtask/src/genesis_args.rs
```diff
@@ -26,7 +26,7 @@ use std::{
 use tempo_chainspec::spec::TEMPO_BASE_FEE;
 use tempo_commonware_node_config::{Peers, PublicPolynomial, SigningKey, SigningShare};
 use tempo_contracts::{
-    ARACHNID_CREATE2_FACTORY_ADDRESS, CREATEX_ADDRESS, MULTICALL_ADDRESS, PERMIT2_ADDRESS,
+    ARACHNID_CREATE2_FACTORY_ADDRESS, CREATEX_ADDRESS, MULTICALL3_ADDRESS, PERMIT2_ADDRESS,
     SAFE_DEPLOYER_ADDRESS,
     contracts::{ARACHNID_CREATE2_FACTORY_BYTECODE, CREATEX_BYTECODE},
```
fgimenez: `CREATEX_BYTECODE` has been removed, we should use instead `CreateX::DEPLOYED_BYTECODE`

0xrusowsky: addressed in [`f0769e9` (#1788)](https://github.com/tempoxyz/tempo/pull/1788/commits/f0769e9c00f80c8fe16f5a90f45fa7e51de543d0)

PR #1789, opened 2025-12-30
fix: incorrect permit2 bytecode

howydev:
We currently etch permit2 bytecode on genesis. This works but is not equivalent to a deployment since EIP712.sol has a constructor

We cannot simulate deployment and etch since the constructor caches `block.chainid` which would give different bytecode for tempo testnet and tempo mainnet

This PR switches the deployment to use EVM execution via the genesis EVM instance

### DIFF
--- a/crates/contracts/src/lib.rs
+++ b/crates/contracts/src/lib.rs
@@ -163,21 +163,6 @@ mod tests {
     }
 
     #[tokio::test]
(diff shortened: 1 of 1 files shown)


howydev:
@legion2002 this test cannot pass for a working permit2 impl since the bytecode contains a cached `chainId` that is diff for mainnet vs us

Will delete the test instead

PR #1790, opened 2025-12-30
fix: ensure mutable TIP20Token calls point to initialized addresses

0xrusowsky:
This PR adds a safety check on the `TIP20Token`'s dispatcher (our only dynamic precompile), which prevents mutable calls to uninitiallized addresses (despite they have already been validated as TIP20 addresses).

ref: `CHAIN-251`
note: since we need to check if the account has bytecode, the gas for mutable calls to TIP20 tokens increases slightly

### DIFF
--- a/crates/precompiles/src/tip20/dispatch.rs
+++ b/crates/precompiles/src/tip20/dispatch.rs
@@ -717,7 +717,7 @@ mod tests {
     }
 
     #[test]
-    fn test_mutable_call_uninitialized_token_reverts() -> eyre::Result<()> {
+    fn test_call_uninitialized_token_reverts() -> eyre::Result<()> {
         let (mut storage, _) = setup_storage();
         let caller = Address::random();
 
@@ -734,7 +734,7 @@ mod tests {
             let result = token.call(&calldata, caller)?;
 
             assert!(result.reverted);
-            let expected: Bytes = TIP20Error::invalid_token().selector().into();
+            let expected: Bytes = TIP20Error::uninitialized().selector().into();
             assert_eq!(result.bytes, expected);
 
             TIP20Setup::create("Test", "TST", caller).apply()?;
--- a/crates/precompiles/src/lib.rs
+++ b/crates/precompiles/src/lib.rs
@@ -286,8 +286,8 @@ where
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::tip20::TIP20Token;
-    use alloy::primitives::{Address, Bytes, U256};
+    use crate::tip20::{TIP20Token, token_id_to_address};
+    use alloy::primitives::{Address, Bytes, U256, bytes};
     use alloy_evm::{
         EthEvmFactory, EvmEnv, EvmFactory, EvmInternals,
         precompiles::{Precompile as AlloyEvmPrecompile, PrecompileInput},
@@ -295,6 +295,7 @@ mod tests {
     use revm::{
         context::ContextTr,
         database::{CacheDB, EmptyDB},
+        state::{AccountInfo, Bytecode},
     };
     use tempo_contracts::precompiles::ITIP20;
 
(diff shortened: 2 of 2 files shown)


fgimenez (approved):
lgtm

0xKitsune (approved):

crates/precompiles/src/tip20/dispatch.rs
```diff
@@ -1,18 +1,28 @@
...
 
 impl Precompile for TIP20Token {
     fn call(&mut self, calldata: &[u8], msg_sender: Address) -> PrecompileResult {
         self.storage
             .deduct_gas(input_cost(calldata.len()))
             .map_err(|_| PrecompileError::OutOfGas)?;
 
+        // Ensure that the token is initialized (has bytecode)
+        // Note that if the initialization check fails, this is treated as uninitialized
+        if !self.is_initialized().unwrap_or(false) {
```
fgimenez, line 21: do we need to add a `!self.storage.is_static()` check here? it was removed in https://github.com/tempoxyz/tempo/pull/1790/commits/2a4c8af95fcd7804a9aa41a6a0251c9735b68659 currently this blocks all calls (including view/read-only), should it call mutable calls only?

crates/precompiles/src/tip20/dispatch.rs
```diff
@@ -706,6 +716,35 @@ mod tests {
...
+            // Any mutable call should revert with invalid_token error
+            let calldata = ITIP20::approveCall {
+                spender: Address::random(),
+                amount: U256::random(),
+            }
+            .abi_encode();
+            let result = token.call(&calldata, caller)?;
+
+            assert!(result.reverted);
+            let expected: Bytes = TIP20Error::invalid_token().selector().into();
```
fgimenez: we are returning now `TIP20Error::uninitialized()` here?

0xKitsune: Yeah I updated to return a different error but can revert if preferred. `InvalidToken` didnt seem very descriptive as this can also be thrown when updating quote tokens. Lmk what you think.

crates/precompiles/src/tip20/dispatch.rs
0xKitsune: Yes this was updated to return an uninitialized error to all calls to uninitialized TIP20s rather than just mutable calls. For consistency it seems more intuitive for any interaction with an uninitialized TIP20 to fail in the same way. However if there is a reason to only revert on mutable calls then happy to change this back. One reason could be perf but we should consider benching this to see if there is a measurable difference. Lmk what you think.

crates/precompiles/src/tip20/dispatch.rs
fgimenez: sure, `uninitialized` feels better, but we should update the expectation here right? i see this error on ci:
```
thread 'tip20::dispatch::tests::test_mutable_call_uninitialized_token_reverts' (81747) panicked at crates/precompiles/src/tip20/dispatch.rs:738:13:
    assertion `left == right` failed
      left: 0x1c72fad4
     right: 0xc1ab6dc1
```

crates/precompiles/src/tip20/dispatch.rs
fgimenez: sounds good, maybe we could document it somewhere, also the PR description should be updated

legion2002 (approved):

crates/precompiles/src/tip20/dispatch.rs
0xrusowsky: fixed the 2 failing tests because of the change discussed above
[`7651dc6` (#1790)](https://github.com/tempoxyz/tempo/pull/1790/commits/7651dc6b7118e91d29d3445fb9a52bfcef402ed5)

PR #1791, opened 2025-12-30
docs(specs): change DEX pair storage layout to be order-sensitive

danrobinson:
This changes how DEX pairs are keyed so that a pair that uses token A as a base token and token B as a quote token would be keyed differently from a pair that uses token B as a base token and token A as a quote token. This would be a breaking change that affects storage layout and cannot be implemented in a hard fork without a regenesis.

### DIFF
--- a/docs/specs/src/StablecoinExchange.sol
+++ b/docs/specs/src/StablecoinExchange.sol
@@ -247,18 +247,12 @@ contract StablecoinExchange is IStablecoinExchange {
             uint128 userBalance = balances[maker][escrowToken];
             if (userBalance >= escrowAmount) {
                 balances[maker][escrowToken] -= escrowAmount;
-            } else {
+            } else if (revertOnTransferFail) {
                 balances[maker][escrowToken] = 0;
-                if (revertOnTransferFail) {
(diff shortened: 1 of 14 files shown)


docs/specs/src/StablecoinExchange.sol
```diff
@@ -1202,21 +1202,54 @@ contract StablecoinExchange is IStablecoinExchange {
         bookKeys = new bytes32[](path.length - 1);
         baseForQuote = new bool[](path.length - 1);
 
+        // Track whether we are currently moving "up" the quote tree (child -> parent)
+        // or "down" (parent -> child). We start with the natural direction from
+        // `tokenIn` towards the LCA and flip once when we cross the LCA so that
+        // at most one hop has to check both orientations.
+        bool isBaseForQuote = true;
```
fgimenez, line 1232: nit, looks like `isBaseForQuote` is used for two purposes now:
- loop state (are we still going up)
- per-hop output (is this hop spending base to get quote)

these coincide bc paths are all up, then all down, but maybe we could use two separate variables to make it more clear?

Issue #1550, opened 2025-12-16
Incorrect contractAddress calculations for tempo transactions


ozwaldorf:
### Describe the bug

Currently when deploying a contract via CREATE, the returned contract address is incorrect in the transaction receipt. 

On tempo, the deployed address is computed from the signer and nonce + 1 instead of just the signer and its current nonce. Currently the RPC returns the contractAddress computed from what would be correct on ethereum, however trying to interact with the contract fails due to it being deployed on the next address, instead of the current one

### Steps to reproduce

1. Deploy a contract with a tempo transaction and passkey
2. Observe the contract address in the transaction receipt, it is incorrect on the explorer, as well as calling eth_getCode for the address it says and it's empty
3. Manually compute the next contract address (nonce + 1 at the time of deployment) and observe both the explorer and eth_getCode correctly shows the deployment

Example deployment:
https://explore.tempo.xyz/tx/0x206ab4a6b3c42332246cdfe71b4b7b18d2bab11ae344646da8a173cb2bb26860

This contract funds itself with a transferFrom during deployment. You can see it transferred to a different address that has the contract, compared to the empty address placed inside the receipt

### Notes on the nonce usage

I suspect the nonce is incremented before the contract address is derived, thus producing a difference from the ethereum spec which also is not accounted for when producing the receipts. Another possibility is a minor difference in how deployment addresses are computed compared to ethereum due to tempo's nonceKey logic.

If this is expected behavior, we should have sufficient documentation highlighting the differences (we already have sections like this for EVM differences, 1559 vs tempo transactions, etc), as well as updating the receipt construction behavior for this edge-case.

Alternatively if this is unexpected, the nonce logic should be corrected to allow the contract to deploy to the standard destination

### What version/commit are you on?

Public testnet - assuming latest commit 

### Code of Conduct

- [x] I agree to follow the Code of Conduct

ozwaldorf:
It appears the following logic increments the nonce for all tempo calls, including ones with CREATE:

https://github.com/tempoxyz/tempo/blob/c9edc5a2c9a506decff9c3f8066fc0d01bc98242/crates/revm/src/handler.rs#L605-L612

As noted by the comment, make_create_frame in revm later captures the nonce, bumps it, and then computes the address from the stored old nonce:

https://github.com/bluealloy/revm/blob/0d424ba/crates/handler/src/frame.rs#L288-L303

[...]

PR #1792, opened 2025-12-30
fix: create address calculations + restrictions

legion2002:
Solves #1550   

### Problem

  When deploying a contract via CREATE in a Tempo transaction, the contract address was computed incorrectly. The address was derived using nonce + 1 instead of the current nonce, causing:
  - Receipt contractAddress to not match the actual deployed address
  - Contract interactions to fail since the contract was at a different address than expected

###  Root Cause

  In handler.rs, the nonce was always bumped during validation for AA transactions (tx.tempo_tx_env.is_some()), even for CREATE operations. This caused make_create_frame to use the already-incremented nonce for address derivation.

  ### Fix

  1. Nonce handling: Only bump nonce during validation for CALL operations. Let make_create_frame handle nonce for CREATE (consistent with standard Ethereum behavior).
  2. CREATE validation: Added rules to ensure predictable behavior:
    - Only one CREATE allowed per transaction (receipt only captures one contractAddress)
    - CREATE must be first call in the batch (matches kind() behavior)

### DIFF
--- a/crates/primitives/src/transaction/tempo_transaction.rs
+++ b/crates/primitives/src/transaction/tempo_transaction.rs
@@ -245,31 +245,24 @@ impl TempoTransaction {
             return Err("valid_before must be greater than valid_after");
         }
 
-        // Authorization list validation: Cannot have Create in any call when aa_authorization_list is non-empty
-        // This follows EIP-7702 semantics - when using delegation
-        if !self.tempo_authorization_list.is_empty() {
-            for call in &self.calls {
-                if call.to.is_create() {
-                    return Err(
-                        "calls cannot contain Create when aa_authorization_list is non-empty",
-                    );
-                }
-            }
+        let mut calls = self.calls.iter();
+
+        // Only the first call in the batch can be a CREATE call.
+        if let Some(call) = calls.next()
+            // Authorization list validation: Can NOT have CREATE when `aa_authorization_list` is non-empty
+            // This follows EIP-7702 semantics - when using delegation
+            && !self.tempo_authorization_list.is_empty()
+            && call.to.is_create()
+        {
+            return Err("calls cannot contain CREATE when 'aa_authorization_list' is non-empty");
         }
 
-        // CREATE validation for batch transactions:
-        // 1. Only ONE CREATE is allowed per transaction
-        // 2. If present, CREATE must be the FIRST call in the batch
-        let mut create_found = false;
-        for (index, call) in self.calls.iter().enumerate() {
+        // All subsequent calls must be CALL.
+        for call in calls {
             if call.to.is_create() {
-                if create_found {
-                    return Err("only one CREATE call is allowed per transaction");
-                }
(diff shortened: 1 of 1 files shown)


crates/primitives/src/transaction/tempo_transaction.rs
```diff
@@ -257,6 +257,22 @@ impl TempoTransaction {
...
+            if call.to.is_create() {
+                if create_found {
+                    return Err("only one CREATE call is allowed per transaction");
+                }
+                if index != 0 {
+                    return Err("CREATE call must be at the first position in the calls array");
+                }
+                create_found = true;
+            }
+        }
```
0xKitsune: Since we already loop through calls above when checking if there are any `CREATE`s when an auth list is present, I believe we can simplify this to something like the following that ensures the first call is not `CREATE` when there is an auth list, and that there is only one `CREATE` in subsequent calls.

```rust
        let mut calls = self.calls.iter();

        // Authorization list validation: Cannot have Create in any call when aa_authorization_list is non-empty
        // This follows EIP-7702 semantics - when using delegation
        if let Some(call) = calls.next()
            && !self.tempo_authorization_list.is_empty()
            && call.to.is_create()
        {
            return Err("calls cannot contain Create when aa_authorization_list is non-empty");
        }

        // Ensure that there is only one create call in the batch
        for call in calls {
            if call.to.is_create() {
                return Err("only one CREATE call is allowed per transaction");
            }
        }
```

0xrusowsky (commented):
the changes seem to be functionally correct.
agreed with @0xKitsune's perf-related cmnts

crates/primitives/src/transaction/tempo_transaction.rs
0xrusowsky: addressed in [`309c739` (#1792)](https://github.com/tempoxyz/tempo/pull/1792/commits/309c7395c7957d8273223c5eb547201a067b994c)

crates/revm/src/handler.rs
```diff
@@ -333,6 +333,19 @@ where
...
+                // make_create_frame during execution. Since checkpoint_revert rolled that back,
+                // we need to manually bump the nonce here to ensure it persists even on failure.
+                // This maintains the invariant that nonces always increment regardless of tx outcome.
+                if calls.first().map(|c| c.to.is_create()).unwrap_or(false) {
+                    let caller = evm.ctx().tx().caller();
+                    if let Ok(mut caller_acc) =
+                        evm.ctx().journal_mut().load_account_with_code_mut(caller)
+                    {
+                        caller_acc.data.bump_nonce();
+                    }
```
rakita, line 346: How does this interact with nonce_key manager? Looking at code [here](https://github.com/tempoxyz/tempo/blob/d1f8b815fa8deb85f466a4fdd1d96060675e2960/crates/revm/src/handler.rs#L693-L694) it looks like nonce manager always bump the nonce from storage

PR #1798, opened 2025-12-30
fix: validator fee tracking by token

legion2002:
## Problem

Currently the TipFeeManager::set_validator_token` allows validators to change their preferred token even when they have accumulated but undistributed fees. This leads to cross-token mispayment.

The `collected_fees` mapping was keyed only by validator address (`Mapping<Address, U256>`), without recording which token was used at accumulation time. This allowed validators to:

1. Accumulate fees in TokenA in block N
2. In block N+1 (produced by another validator), call `set_validator_token` to switch to TokenB
3. In block N+2, when `distribute_fees` is called, receive the stored amount but in TokenB instead of TokenA

This creates a shortfall in TokenB and leaves orphaned TokenA in the fee manager.

## Solution

Changed `collected_fees` to a nested mapping keyed by `(validator, token)`:
```rust
// Before
collected_fees: Mapping<Address, U256>

// After  
collected_fees: Mapping<Address, Mapping<Address, U256>>

This allows validators to:

Collect fees for multiple tokens simultaneously
Change their preferred token at any time without affecting already-accumulated fees
Withdraw fees per-token via distributeFees(validator, token)

API Changes

// Before
function distributeFees(address validator) external;
function collectedFeesByValidator(address validator) external view returns (uint256);

// After
function distributeFees(address validator, address token) external;
function collectedFees(address validator, address token) external view returns (uint256);

### DIFF
--- a/docs/specs/test/FeeManager.t.sol
+++ b/docs/specs/test/FeeManager.t.sol
@@ -266,11 +266,11 @@ contract FeeManagerTest is BaseTest {
         vm.coinbase(validator);
 
         try amm.collectFeePreTx(user, address(userToken), maxAmount) {
-            uint256 validatorBalanceBefore = amm.collectedFees(validator, address(validatorToken));
+            uint256 validatorBalanceBefore = amm.collectedFees(validator, _PATH_USD);
 
             amm.collectFeePostTx(user, maxAmount, actualUsed, address(userToken));
 
-            uint256 validatorBalanceAfter = amm.collectedFees(validator, address(validatorToken));
+            uint256 validatorBalanceAfter = amm.collectedFees(validator, _PATH_USD);
             vm.stopPrank();
 
             assertGt(validatorBalanceAfter, validatorBalanceBefore);

### CODE REVIEW

File: crates/node/tests/it/tempo_transaction.rs [hunk 1]
```diff
@@ -5452,3 +5452,95 @@ async fn test_aa_key_authorization_chain_id_validation() -> eyre::Result<()> {
 
     Ok(())
 }
+
+/// Test that contract CREATE in a Tempo transaction computes the correct contract address.
+#[tokio::test(flavor = "multi_thread")]
+async fn test_aa_create_correct_contract_address() -> eyre::Result<()> {
```
0xKitsune: Looks like this change is unrelated to this PR?

crates/precompiles/src/tip20/mod.rs
```diff
@@ -755,6 +755,7 @@ impl TIP20Token {
 
     /// Transfers fee tokens from user to fee manager before transaction execution
     pub fn transfer_fee_pre_tx(&mut self, from: Address, amount: U256) -> Result<()> {
+        self.check_not_paused()?;
```
0xKitsune: Can we separate this change from the validator fee tracking changes?

crates/precompiles/src/tip_fee_manager/mod.rs
```diff
@@ -21,7 +21,7 @@ use tempo_precompiles_macros::contract;
 pub struct TipFeeManager {
     validator_tokens: Mapping<Address, Address>,
     user_tokens: Mapping<Address, Address>,
-    collected_fees: Mapping<Address, U256>,
+    collected_fees: Mapping<Address, Mapping<Address, U256>>,
     pools: Mapping<B256, Pool>,
```
0xKitsune, line 25: This will incur an additional keccak on every tx right? We could consider encoding the key as `valdiator_address | tokenId` to keep the same number of hashes when indexing the mapping. Lmk what you think.

crates/primitives/src/transaction/tempo_transaction.rs
```diff
@@ -256,6 +256,22 @@ impl TempoTransaction {
...
+            if call.to.is_create() {
+                if create_found {
+                    return Err("only one CREATE call is allowed per transaction");
+                }
+                if index != 0 {
+                    return Err("CREATE call must be at the first position in the calls array");
+                }
+                create_found = true;
+            }
+        }
```
0xKitsune: Same thing here, looks unrelated to valdiator fees.

crates/precompiles/src/tip_fee_manager/mod.rs
legion2002: Yes it does add an additional keccak, we could do some optimizations like `validator_address | tokenId`. In general  it would be nicer if we just implement the storage layout optimizations we have been talking about across the board.
Non standard encodings seem hard to maintain for mainnet.

Think we should merge this PR, and then open a perf issue separately to discuss this, because simple tokenId might not work if we make some changes we're discussing in the thread here - https://tempoxyz.slack.com/archives/C09KCGR4LQ4/p1767086851885849?thread_ts=1765883621.693449&cid=C09KCGR4LQ4

0xKitsune: Sounds good to me

PR #1799, opened 2025-12-30
fix: add pause check to transfer_fee_pre_tx

legion2002:
When a TIP20 token is paused, pre-transaction fee transfers should fail to make the transaction invalid. This prevents the fee manager from transferring funds when the token is paused.


crates/precompiles/src/tip20/mod.rs
```diff
@@ -755,6 +755,7 @@ impl TIP20Token {
 
     /// Transfers fee tokens from user to fee manager before transaction execution
     pub fn transfer_fee_pre_tx(&mut self, from: Address, amount: U256) -> Result<()> {
+        self.check_not_paused()?;
```
0xKitsune, line 762: nit: can we add a comment for context as to why we check this here but not in `transfer_fee_post_tx`?

PR #1801, opened 2025-12-30
fix: add revert if policy doesn't exist

legion2002:
Addresses TMPO 29

It reverts when a policy doesn't exist, instead of returning false. This removes ambiguity from errors, for what is causing the issue.

It does incur one extra storage read of the policyId counter per policy fetch. I don't see a way to avoid this. Maybe the perf degradation is not worth the fix, because it primarily affects offchain tooling, and it should be pretty easy to check offchain if a policy exists or not.

CC: @0xKitsune

### DIFF
--- a/docs/specs/test/TIP403Registry.t.sol
+++ b/docs/specs/test/TIP403Registry.t.sol
@@ -702,6 +702,15 @@ contract TIP403RegistryTest is BaseTest {
         assertFalse(registry.isAuthorized(999, alice));
     }
 
+    function test_PolicyData_RevertsForNonExistentPolicy() public {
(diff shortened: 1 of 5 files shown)


docs/specs/test/TIP403Registry.t.sol
```diff
@@ -702,6 +702,15 @@ contract TIP403RegistryTest is BaseTest {
         assertFalse(registry.isAuthorized(999, alice));
     }
 
+    function test_PolicyData_RevertsForNonExistentPolicy() public {
```
0xKitsune, line 705: nit: consider fuzzing this

0xKitsune (approved):

PR #1804, opened 2025-12-30
Bump commonware to d16c7481c72cf6f19b77e1435f1e5dd8e32a1aac

kamsz:
As https://github.com/commonwarexyz/monorepo/pull/2653 is required for devnets to function; backports argument name change from commonware to tempo


Zygimantass (approved):

Cargo.toml
```diff
@@ -184,16 +184,16 @@ alloy-signer-local = "1.1.3"
...
-commonware-codec = "0.0.64"
-commonware-consensus = "0.0.64"
-commonware-cryptography = "0.0.64"
-commonware-macros = "0.0.64"
-commonware-math = "0.0.64"
-commonware-p2p = "0.0.64"
-commonware-runtime = "0.0.64"
-commonware-storage = "0.0.64"
-commonware-utils = "0.0.64"
+commonware-broadcast = { git = "https://github.com/commonwarexyz/monorepo", rev = "d16c7481c72cf6f19b77e1435f1e5dd8e32a1aac" }
```
SuperFluffy: Please provide this in a `[patch.crates-io]` section and keep the version here stable.

Also please note what PR this rev corresponds to so that it's easy to locate (i.e. something like `# Commonware main right after https://github.com/commonwarexyz/monorepo/pull/2653 was merged`)

crates/commonware-node/src/args.rs
```diff
@@ -123,11 +123,8 @@ pub struct Args {
     /// Connections are still authenticated via public key cryptography, but
     /// anyone can attempt handshakes, increasing exposure to DoS attacks.
     /// Only enable in trusted network environments.
-    #[arg(
-        long = "consensus.allow-unregistered-handshakes",
-        default_value_t = false
-    )]
-    pub allow_unregistered_handshakes: bool,
+    #[arg(long = "consensus.bypass-ip-check", default_value_t = false)]
+    pub bypass_ip_check: bool,
```
SuperFluffy, line 127: Not against this change, but want to note that this is not technically necessary - we don't need to break CLI args to follow commonware naming changes.

kamsz: got it, although the reason behind it makes sense

> to avoid confusion, I'm changing the config name to bypass_ip_check (allow_unregistered_handshakes implies we allow people to connect that aren't registered peers, which is never what this did)

Cargo.toml
```diff
@@ -315,3 +315,16 @@ vergen-git2 = "1"
...
+commonware-codec = { git = "https://github.com/commonwarexyz/monorepo", rev = "d16c7481c72cf6f19b77e1435f1e5dd8e32a1aac" }
+commonware-consensus = { git = "https://github.com/commonwarexyz/monorepo", rev = "d16c7481c72cf6f19b77e1435f1e5dd8e32a1aac" }
+commonware-cryptography = { git = "https://github.com/commonwarexyz/monorepo", rev = "d16c7481c72cf6f19b77e1435f1e5dd8e32a1aac" }
+commonware-macros = { git = "https://github.com/commonwarexyz/monorepo", rev = "d16c7481c72cf6f19b77e1435f1e5dd8e32a1aac" }
+commonware-math = { git = "https://github.com/commonwarexyz/monorepo", rev = "d16c7481c72cf6f19b77e1435f1e5dd8e32a1aac" }
+commonware-p2p = { git = "https://github.com/commonwarexyz/monorepo", rev = "d16c7481c72cf6f19b77e1435f1e5dd8e32a1aac" }
+commonware-runtime = { git = "https://github.com/commonwarexyz/monorepo", rev = "d16c7481c72cf6f19b77e1435f1e5dd8e32a1aac" }
+commonware-storage = { git = "https://github.com/commonwarexyz/monorepo", rev = "d16c7481c72cf6f19b77e1435f1e5dd8e32a1aac" }
+commonware-utils = { git = "https://github.com/commonwarexyz/monorepo", rev = "d16c7481c72cf6f19b77e1435f1e5dd8e32a1aac" }
```
SuperFluffy: Just add an EOL here :-)

SuperFluffy (approved):

PR #1805, opened 2025-12-31
fix(specs): check 403 policy on both tokens in stablecoin DEX

danrobinson:
Partially addresses TMPO-20.

Currently, the DEX only checks TIP-403 policies on the token that is being escrowed (i.e. sold by the limit order). This could allow a maker to place an order to buy a token that they are not authorized to transfer. This could have slightly odd results in the case of blacklisting, but particularly bad UX in the case of whitelisting, where anyone would be allowed to buy a token without being on the whitelist for it, resulting in them being stuck with the token.

This PR changes this in the reference implementation by having the placeOrder function check the authorization policies on both tokens. It also checks that the DEX itself is authorized by both tokens.

In the case of a flip order, this will cause the newly placed order to silently fail (ending the cycle of flips).

### DIFF
--- a/crates/precompiles/src/stablecoin_exchange/mod.rs
+++ b/crates/precompiles/src/stablecoin_exchange/mod.rs
@@ -3927,7 +3927,7 @@ mod tests {
             )?;
 
             // Set up base and quote tokens
-            let (base_addr, quote_addr) =
+            let (base_addr, _quote_addr) =
                 setup_test_tokens(admin, alice, exchange.address, MIN_ORDER_AMOUNT * 4)?;
 
             // Get the base token and apply blacklist policy


docs/pages/protocol/exchange/spec.mdx
```diff
@@ -160,6 +160,8 @@ Places a limit order against the pair of `token` and its quote, immediately addi
 Notes:
 
 - `tick` must be within `[MIN_TICK, MAX_TICK]` and divisible by `TICK_SPACING` (10).
+- The maker must be authorized by the TIP-403 transfer policies of both the base and quote tokens. This ensures makers cannot place orders to buy or sell tokens they are not permitted to transfer.
+- Additionally, the DEX contract itself must be authorized by the TIP-20 transfer policies of both the base and quote tokens. This allows token issuers to prevent their tokens from being traded on the DEX.
```
howydev, line 166: (asking out of curiosity) Have we considered having the DEX just always be on the whitelist for all policies? Would there be a case in which a token issuer doesn't want to use our DEX? 

This should reduce a bunch of storage loads for all swap/place operations which is nice

docs/specs/test/StablecoinExchange.t.sol
```diff
@@ -1206,6 +1206,96 @@ contract StablecoinExchangeTest is BaseTest {
...
+        vm.prank(alice);
+        try exchange.place(address(token1), orderAmount, false, 100) {
+            revert CallShouldHaveReverted();
+        } catch (bytes memory err) {
+            assertEq(err, abi.encodeWithSelector(ITIP20.PolicyForbids.selector));
+        }
+    }
+
+    /// @notice Test that a maker blacklisted in either token cannot place a flip order
+    function test_BlacklistedUser_CannotPlaceFlipOrder() public {
```
howydev, line 1398: nit: could we add a test for a flip order silently failing the flip stage?

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -397,16 +397,20 @@ impl StablecoinExchange {
...
                 .ok_or(StablecoinExchangeError::insufficient_balance())?;
-            (quote_token, quote_amount)
+            (quote_token, quote_amount, token)
         } else {
             // For asks, escrow base tokens
-            (token, amount)
+            (token, amount, quote_token)
         };
 
+        // Check policy on non-escrow token (escrow token is checked in decrement_balance_or_transfer_from)
```
howydev, line 408: rly small nit: could use `token` and `quote_token` instead of the extra var

were we not checking for authorization at all before? oof

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -537,16 +541,20 @@ impl StablecoinExchange {
         }
 
         // Calculate escrow amount and token based on order side
-        let (escrow_token, escrow_amount) = if is_bid {
+        let (escrow_token, escrow_amount, non_escrow_token) = if is_bid {
```
howydev, line 544: same nit as above

howydev (approved):
change makes sense (I assume we're treating orders that are already placed and then a blacklist/whitelist happens as a separate case)

just have some small nits, preapproving to save a round trip

docs/pages/protocol/exchange/spec.mdx
danrobinson: The loads will only be on places (including flip places), so it's not as bad as it might look.

I can see reasons why an issuer might want to disable trading on the DEX. But maybe it's enough to just let them blacklist deposits and withdrawals, rather than doing the check on every place...

docs/specs/test/StablecoinExchange.t.sol
danrobinson: I have this test as part of #1806, which also changes the behavior

crates/precompiles/src/stablecoin_exchange/mod.rs
0xKitsune: > were we not checking for authorization at all before? oof



Auth is checked on the escrow token in `decrement_balance_or_transfer_from`. I think the changes in this PR are using an additional var to try to avoid checking auth on the escrowed token twice.

danrobinson: > rly small nit: could use token and quote_token instead of the extra var

Sorry, what do you mean?

> were we not checking for authorization at all before? oof

We were checking on the token that gets escrowed; not the other one

danrobinson: We check on non_escrow_token so that we don't do two checks on the escrow_token (since we check that in the decrement_balance function)

howydev: ah i see now that the check happened in `decrement_balance_or_transfer_from`

0xKitsune (approved):
Can we add precompile tests to assert that new orders are rejected if the non escrowed token is not authorized?

docs/pages/protocol/exchange/spec.mdx
howydev: its also hit on transferring in/out to DEX so basically all swaps too right?

but yeah feeAMM/fee manager is probably worse in terms of number of times it's hit, so we should probably consider whitelisting that first

crates/precompiles/src/stablecoin_exchange/mod.rs
howydev: yep mb, we need to check the other token instead so this is necessary!

docs/pages/protocol/exchange/spec.mdx
danrobinson: Yep that's true, but that's just twice per swap. I think allowing tokens to blacklist the feeAMM seems valuable (for example it won't really make sense if whitelist tokens can always be used to pay fees)

PR #1806, opened 2025-12-31
fix(dex): have flip orders only draw from internal balances

danrobinson:
This change (to the spec and implementation) prevents flip orders from triggering a transferFrom() from the maker if the maker has insufficient balance on the DEX (an edge case that should only happen if the maker withdraws after a flip order has been partially filled). This prevents swappers from paying extra gas, and prevents swaps on the DEX from having side effects on non-DEX state.

### DIFF
--- a/crates/precompiles/src/stablecoin_exchange/mod.rs
+++ b/crates/precompiles/src/stablecoin_exchange/mod.rs
@@ -489,22 +489,8 @@ impl StablecoinExchange {
     /// Flip orders automatically create a new order on the opposite side when completely filled.
     /// For bids: flip_tick must be > tick
     /// For asks: flip_tick must be < tick
-    pub fn place_flip(
-        &mut self,
-        sender: Address,
-        token: Address,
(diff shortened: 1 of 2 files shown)


howydev (approved):
LGTM

crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -497,6 +497,22 @@ impl StablecoinExchange {
...
+    #[allow(clippy::too_many_arguments)]
+    fn place_flip_impl(
+        &mut self,
+        sender: Address,
+        token: Address,
+        amount: u128,
+        is_bid: bool,
+        tick: i16,
+        flip_tick: i16,
+        internal_balance_only: bool,
```
0xKitsune: We can remove the `place_flip_impl` function and just add a `internal_balance` bool to the original function.

0xKitsune (approved):

PR #1811, opened 2025-12-31
feat: update `compute_book_key` and `validate_and_build_route`

0xKitsune:
This PR updates the stablecoin exchange precompile to reflect the spec changes introduced in `dr/dex-order-sensitive-keys`.

### DIFF
--- a/crates/precompiles/src/stablecoin_exchange/mod.rs
+++ b/crates/precompiles/src/stablecoin_exchange/mod.rs
@@ -2318,7 +2318,7 @@ mod tests {
     }
 
     /// Helper to verify a single hop in a route
-    fn verify_hop(hop: (B256, bool), token_in: Address, _token_out: Address) -> eyre::Result<()> {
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/stablecoin_exchange/mod.rs
```diff
@@ -2312,18 +2318,20 @@ mod tests {
     }
 
     /// Helper to verify a single hop in a route
-    fn verify_hop(hop: (B256, bool), token_in: Address, token_out: Address) -> eyre::Result<()> {
-        let (book_key, base_for_quote) = hop;
-        let expected_book_key = compute_book_key(token_in, token_out);
-        assert_eq!(book_key, expected_book_key, "Book key should match");
+    fn verify_hop(hop: (B256, bool), token_in: Address, _token_out: Address) -> eyre::Result<()> {
```
fgimenez: why do we need to keep `_token_out`, could we remove it?

PR #1816, opened 2025-12-31
[1/n] feat: change tip20 deployment scheme 

howydev:
First PR in stack

1. Changes the deployment scheme to be `0x${20c00000000000000000}${keccak256(abi.encode(msg.sender, salt)}` in ref impl
2. Change createToken function to take in salt. Event no longer emits tokenId, but emits salt
3. Change isTIP20 check to check prefix and bytecode length. Length > 0 for test environment, code == "0xEF" for tempo mainnet
4. Add "reservedSize" to reserve the first 1000 TIP20 addresses

### DIFF
--- a/docs/specs/test/StablecoinExchange.t.sol
+++ b/docs/specs/test/StablecoinExchange.t.sol
@@ -117,8 +117,11 @@ contract StablecoinExchangeTest is BaseTest {
     function test_OrderbookPairs_AreOrderSensitive_AcrossQuoteUpdates() public {
         // Create a dedicated base token and two possible quote tokens, all USD-denominated.
         vm.startPrank(admin);
-        ITIP20 base = ITIP20(factory.createToken("OBBase", "OBB", "USD", pathUSD, admin));
-        ITIP20 quote1 = ITIP20(factory.createToken("OBQuote1", "OBQ1", "USD", pathUSD, admin));
+        ITIP20 base =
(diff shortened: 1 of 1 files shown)


legion2002:
@howydev have been thinking, do we need to hash the msg.sender and salt? 
Can we allow the user to pass any salt and just add it directly after the prefix? 

Reason being that it seems like if we have a 12 byte prefix, it should be possible to mine the hashed version of the salt anyways, so if we do just the salt, then we can save on a keccak for each deployment.

But not sure if my assumptions are correct.

docs/specs/src/TIP20Factory.sol
```diff
@@ -11,49 +12,36 @@ contract TIP20Factory is ITIP20Factory {
     // Foundry cheatcode VM for deployCodeTo
     Vm private constant vm = Vm(address(uint160(uint256(keccak256("hevm cheat code")))));
 
-    uint256 public tokenIdCounter = 0;
+    uint256 internal immutable reservedSize = 1000;
```
legion2002, line 15: We should make this 1024 or 256, so that we can reserve a specific number of lower bits in the address.

howydev:
> Can we allow the user to pass any salt and just add it directly after the prefix?

I think we definitely need msg.sender, else an adversary can DoS token creation right?

> We should make this 1024 or 256, so that we can reserve a specific number of lower bits in the address.

1024 makes sense, will change to that

PR #1819, opened 2025-12-31
feat!: change tip20 deployment scheme

howydev:
The token factory today allocates token addresses based on a FCFS basis. This could potentially create phishing style issues in which adversaries front run token deployments (a la USDP).

This PR contains a fix by modifying our token deployment address scheme. We use `0x${12-byte TIP20 prefix}${keccak256(abi.encode(msg.sender, salt))[:8]}` instead to retain properties of vanity TIP-20 addresses and cheap TIP20 checks, while providing token creators a way to avoid front-running attacks while creating tokens

12-bytes is chosen for the prefix for certainty that it's impossible to mine. It's possible to engineer a suffix collision with 8-byte suffix, but there would not be any gain for the attacker so this is fine

### DIFF
--- a/docs/specs/test/TIP20Factory.t.sol
+++ b/docs/specs/test/TIP20Factory.t.sol
@@ -106,7 +106,6 @@ contract TIP20FactoryTest is BaseTest {
     function testIsTIP20Function() public view {
         assertTrue(factory.isTIP20(_PATH_USD));
         assertTrue(factory.isTIP20(address(token1)));
-        assertFalse(factory.isTIP20(0x20C0000000000000000000000000000000000001));
         assertFalse(factory.isTIP20(address(0)));
         assertFalse(factory.isTIP20(address(0x1234)));
         assertFalse(factory.isTIP20(0x21C0000000000000000000000000000000000000));


crates/precompiles/src/test_util.rs
```diff
@@ -204,6 +205,18 @@ impl TIP20Setup {
...
+        self.salt = Some(salt);
+        self
+    }
+
+    /// Set the admin address explicitly. Required for `config()` when using `with_mint()`.
+    pub fn with_admin(mut self, admin: Address) -> Self {
+        self.admin = Some(admin);
+        self
+    }
+
```
0xrusowsky, line 219: the admin is derived automatically when calling `apply()` so this shouldn't be necessary

crates/precompiles/src/tip20_factory/mod.rs
```diff
@@ -1,23 +1,47 @@
...
+};
+use alloy::{
+    primitives::{Address, B256, keccak256},
+    sol_types::SolValue,
 };
-use alloy::primitives::{Address, U256};
 use tracing::trace;
 
+/// Number of reserved addresses (0 to RESERVED_SIZE-1) that cannot be deployed via factory
+const RESERVED_SIZE: u128 = 1000;
```
legion2002: Reserved size should be a factor of 2, so we can reserve a particular number of bits.

xtask/src/genesis_args.rs
```diff
@@ -197,36 +197,39 @@ impl GenesisArgs {
...
-        let (_, alpha_token_address) = create_and_mint_token(
+        let alpha_token_address = create_and_mint_token(
             "AlphaUSD",
             "AlphaUSD",
             "USD",
             PATH_USD_ADDRESS,
             admin,
             &addresses,
             U256::from(u64::MAX),
+            B256::from(U256::from(1)),
```
legion2002: Is this correct? We want to initialize alphaUSD, etc. at `prefix +  0x01`. But with the current create_and_mint, it would assume that 1 is the salt, and then deploy it at a random address?

docs/specs/src/TempoUtilities.sol
```diff
@@ -0,0 +1,27 @@
...
+
+    function isTIP20(address token) internal view returns (bool) {
+        bool isTempo = _TIP403REGISTRY.code.length + _TIP20FACTORY.code.length
+                + _PATH_USD.code.length + _STABLECOIN_DEX.code.length + _NONCE.code.length == 5;
+
+        // If this is tempo mainnet, we use code == 0xEF check
+        // Otherwise we just check that the address has code
+        return bytes10(bytes20(token)) == bytes10(0x20c00000000000000000)
+            && ((isTempo && token.code.length == 1 && token.code[0] == hex"EF")
+                || (!isTempo && token.code.length > 0));
```
legion2002, line 24: why can't we check token.code.length > 0 for both cases?

crates/precompiles/src/tip20/mod.rs
```diff
@@ -33,8 +33,9 @@ const TIP20_DECIMALS: u8 = 6;
 /// USD currency string constant
 pub const USD_CURRENCY: &str = "USD";
 
-/// TIP20 token address prefix (12 bytes for token ID encoding)
-const TIP20_TOKEN_PREFIX: [u8; 12] = hex!("20C000000000000000000000");
+/// TIP20 token address prefix (10 bytes)
+/// The full address is: TIP20_TOKEN_PREFIX (10 bytes) || derived_bytes (10 bytes)
+const TIP20_TOKEN_PREFIX: [u8; 10] = hex!("20C00000000000000000");
```
legion2002: Noting that this reduces our TIP20 prefix mining assumptions by 2 bytes.
It should be fine. But if we're worried about it, we can also do a `12 byte prefix || 8 derived bytes salt`

Prefix mining is a critical issue, because it affects the security properties of the chain. Salt mining is not, because even if you mine the same TIP20 salt, you should not be allowed to deploy to a token which is already deployed.

crates/precompiles/benches/tempo_precompiles.rs
```diff
@@ -42,10 +42,7 @@ fn tip20_metadata(c: &mut Criterion) {
         let mut storage = HashMapStorageProvider::new(1);
         StorageCtx::enter(&mut storage, || {
             initialize_path_usd(admin).unwrap();
-            let mut token = TIP20Token::new(1);
-            token
-                .initialize("TestToken", "T", "USD", PATH_USD_ADDRESS, admin)
-                .unwrap();
+            let mut token = create_test_token(admin, B256::from(U256::from(1))).unwrap();
```
0xrusowsky: nit: rather than using a helper fn, can we use `TIP20Setup::create()` as that's what it was designed for?

this way we can remove the helper and the `initialize_path_usd` call

crates/precompiles/src/tip20_factory/mod.rs
```diff
@@ -34,91 +58,77 @@ impl TIP20Factory {
...
-            {
-                return Err(TIP20Error::invalid_quote_token().into());
-            }
+        // Ensure that the quote token is a valid TIP20 that is currently deployed.
+        if !self.is_tip20(call.quoteToken)? {
+            return Err(TIP20Error::invalid_quote_token().into());
         }
 
-        TIP20Token::new(token_id).initialize(
+        TIP20Token::from_address(token_address)?.initialize(
```
legion2002, line 111: If we increase the prefix bytes to 12, then we can just add a check here which ensures that the token address does not have code already.
Reasoning discussed above.

crates/precompiles/benches/tempo_precompiles.rs
howydev: This and the prev was needed because we can't use the factory function to deploy to PathUSD anymore (we need a sender/salt combo that hashes to 0x0)

But yep I think its cleaner to remove and have an internal function that does these deployments. IIUC we want this eventually, too

xtask/src/genesis_args.rs
howydev: Yep this is a randomish address. Why do we need alphaUSD at `prefix +  0x01`?

docs/specs/src/TempoUtilities.sol
howydev: Just in case hashes become cheap/someone gets lucky, it stops a random smart contract having access to payment lane

crates/precompiles/src/tip20/mod.rs
howydev: I think 10 bytes prefix is sufficiently secure, if not happy to revisit

bin/tempo-sidecar/src/cmd/simple_arb.rs
```diff
@@ -42,12 +43,22 @@ pub struct SimpleArbArgs {
 
 #[instrument(skip(provider))]
 async fn fetch_all_pairs<P: Provider>(provider: P) -> eyre::Result<HashSet<(Address, Address)>> {
-    let tip20_factory = ITIP20Factory::new(TIP20_FACTORY_ADDRESS, provider);
-    let last_token_id = tip20_factory.tokenIdCounter().call().await?.to::<u64>();
-
-    let tokens = (0..last_token_id)
-        .map(token_id_to_address)
-        .collect::<Vec<_>>();
+    use alloy::rpc::types::Filter;
```
0xKitsune: nit: consider moving imports to the top of the file

crates/precompiles/src/tip20/mod.rs
0xKitsune: Agreed we should make this 12 bytes as mentioned in other threads.

crates/precompiles/src/tip20_factory/mod.rs
```diff
@@ -1,23 +1,45 @@
...
+    sol_types::SolValue,
 };
-use alloy::primitives::{Address, U256};
 use tracing::trace;
 
+/// Number of reserved addresses (0 to RESERVED_SIZE-1) that cannot be deployed via factory
+const RESERVED_SIZE: u128 = 1024;
+
+/// TIP20 token address prefix (10 bytes): 0x20C00000000000000000
+const TIP20_PREFIX_BYTES: [u8; 10] = [0x20, 0xC0, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00];
```
0xKitsune: Same thing here, lets update to 12 bytes.

crates/precompiles/src/tip20_factory/mod.rs
```diff
@@ -34,91 +56,107 @@ impl TIP20Factory {
...
-                .ok_or(TempoPrecompileError::under_overflow())?,
-        )?;
-
         Ok(token_address)
     }
 
-    pub fn token_id_counter(&self) -> Result<U256> {
-        self.token_id_counter.read()
+    /// Creates a token at a reserved address. Internal function only, to create pathUSD at genesis.
+    pub fn create_path_usd(&mut self, admin: Address) -> Result<Address> {
```
0xKitsune: This comment says "creates a token at a reserved address" but this function is hardcoded for path usd only. Can we update this function to deploy any reserved token address and update the fn name repsectively?

crates/precompiles/src/tip20_factory/mod.rs
```diff
@@ -34,91 +56,107 @@ impl TIP20Factory {
...
+        use crate::PATH_USD_ADDRESS;
```
0xKitsune: nit: move imports to the top of the file.

crates/precompiles/src/tip20_factory/mod.rs
```diff
@@ -34,91 +56,107 @@ impl TIP20Factory {
...
+
+        let mut path_usd = TIP20Token::from_address(PATH_USD_ADDRESS)?;
+        path_usd.initialize("PathUSD", "PUSD", "USD", Address::ZERO, admin)?;
```
0xKitsune: @danrobinson what should the PathUSD symbol be?

crates/precompiles/src/tip20_factory/mod.rs
```diff
@@ -133,49 +171,38 @@ mod tests {
...
+            // Verify addresses are different
+            assert_ne!(token_addr_1, token_addr_2);
+
+            // Verify addresses have TIP20 prefix
+            assert!(is_tip20_prefix(token_addr_1));
+            assert!(is_tip20_prefix(token_addr_2));
+
+            // Verify tokens are valid TIP20s
+            assert!(factory.is_tip20(token_addr_1)?);
+            assert!(factory.is_tip20(token_addr_2)?);
```
0xKitsune, line 235: Looks like we updated this test to drop event emission assertions, can we add this back?

crates/precompiles/src/tip20_factory/mod.rs
```diff
@@ -231,97 +264,46 @@ mod tests {
...
-            );
-
-            let call = ITIP20Factory::createTokenCall {
-                name: "Test".to_string(),
-                symbol: "Test".to_string(),
-                currency: "USD".to_string(),
-                quoteToken: Address::ZERO,
-                admin: sender,
-            };
+            if let Some(reserved_salt) = found_reserved_salt {
```
0xKitsune: We should update this test so that it always asserts that reserved addresses cant be deployed. Right now its possible that this test passes but really this assertion is not being checked if a reserved salt is not found.

crates/precompiles/src/tip20_factory/mod.rs
```diff
@@ -361,16 +343,38 @@ mod tests {
...
+    fn test_compute_tip20_address_deterministic() {
+        let sender = Address::random();
+        let salt = B256::random();
+
+        // Same sender + salt should produce same address
+        let (addr1, lower1) = compute_tip20_address(sender, salt);
+        let (addr2, lower2) = compute_tip20_address(sender, salt);
+
+        assert_eq!(addr1, addr2);
+        assert_eq!(lower1, lower2);
```
0xKitsune: This assertion does not tell us much other than the same inputs return the same outputs, IMO we should remove this.

A more useful check would be that the same salt but different senders produce different values, as well as the same sender but different salts. We can probably prop test/fuzz this as well.

crates/precompiles/tests/storage_tests/solidity/testdata/tip20_factory.sol
```diff
@@ -3,9 +3,10 @@ pragma solidity ^0.8.0;
 
 /// Test contract for TIP20Factory storage layout.
 /// Factory for creating TIP20 tokens.
+///
+/// NOTE: TIP20Factory no longer has storage fields.
+/// Token addresses are now derived from keccak256(sender, salt) instead of
+/// using a sequential tokenIdCounter.
```
0xKitsune: Lets remove this comment, this mentions a previous version/feature of the factory that will be be removed with this PR and will be confusing to anyone reading without context on previous factory behavior.

crates/precompiles/tests/storage_tests/solidity/precompiles.rs
```diff
@@ -9,18 +9,8 @@ use tempo_precompiles_macros::{
...
-    let sol_path = testdata("tip20_factory.sol");
-    let solc_layout = load_solc_layout(&sol_path);
-    let rust_layout = layout_fields!(token_id_counter);
-
-    if let Err(errors) = compare_layouts(&solc_layout, &rust_layout) {
-        panic_layout_mismatch("Layout", errors, &sol_path);
-    }
-}
+// TIP20Factory no longer has storage fields - address derivation is now
+// based on keccak256(sender, salt) instead of a sequential token_id_counter
```
0xKitsune: Lets remove this comment as well, no need for comments about previous behavior.

crates/precompiles/tests/storage_tests/solidity/precompiles.rs
```diff
@@ -201,17 +191,7 @@ fn export_all_storage_constants() {
...
-        use tempo_precompiles::tip20_factory::slots;
-        let fields = layout_fields!(token_id_counter);
-        all_constants.insert(
-            "tip20_factory".to_string(),
-            json!({
-                "fields": fields.iter().map(field_to_json).collect::<Vec<_>>()
-            }),
-        );
-    }
+    // TIP20 Factory - no storage fields (address derivation is now salt-based)
```
0xKitsune: Same thing here, "address is now salt-based" requires context about past behavior.

crates/primitives/src/transaction/envelope.rs
```diff
@@ -10,9 +10,9 @@ use alloy_consensus::{
 use alloy_primitives::{Address, B256, Bytes, Signature, TxKind, U256, hex};
 use core::fmt;
 
-/// TIP20 payment address prefix (14 bytes for payment classification)
-/// Same as TIP20_TOKEN_PREFIX but extended to 14 bytes for payment classification
-pub const TIP20_PAYMENT_PREFIX: [u8; 14] = hex!("20C0000000000000000000000000");
+/// TIP20 payment address prefix (10 bytes for payment classification)
+/// Same as TIP20_TOKEN_PREFIX but extended to 10 bytes for payment classification
+pub const TIP20_PAYMENT_PREFIX: [u8; 10] = hex!("20C00000000000000000");
```
0xKitsune: Same thing here, lets update to 12 bytes.

crates/revm/src/common.rs
```diff
@@ -208,7 +208,8 @@ pub trait TempoStateAccess<M = ()> {
         // Ensure the currency is USD
         // load fee token account to ensure that we can load storage for it.
         self.with_read_only_storage_ctx(spec, || {
-            let token = TIP20Token::new(tip20::address_to_token_id_unchecked(fee_token));
+            // SAFETY: prefix already checked above
+            let token = TIP20Token::from_address(fee_token)?;
```
0xKitsune, line 212: This incurs a redundant `is_tip20` check as this is also checked in `from_address`.

crates/revm/src/common.rs
```diff
@@ -575,15 +576,17 @@ mod tests {
 
     #[test]
     fn test_read_token_balance_typed_storage() -> eyre::Result<()> {
-        // Create a TIP-20 token address
-        let token_id = 1u64;
-        let token_address = tip20::token_id_to_address(token_id);
+        // Use PATH_USD as the token address
```
0xKitsune: nit: this comment isnt adding any helpful context that the code doesnt already imply, consider removing.

crates/revm/src/handler.rs
```diff
@@ -1317,11 +1317,13 @@ pub fn get_token_balance<JOURNAL>(
...
     JOURNAL: JournalTr,
 {
-    // Address has already been validated
-    let token_id = tip20::address_to_token_id_unchecked(token);
-
+    // Address has already been validated as having TIP20 prefix
     journal.load_account(token)?;
-    let balance_slot = TIP20Token::new(token_id).balances.at(sender).slot();
+    let balance_slot = TIP20Token::from_address(token)
+        .expect("TIP20 prefix already validated")
```
0xKitsune, line 1336: Can we propagate this error here rather than using expect? The inputs should be validated before calling this function but it seems safer to propagate rather than panic here.

crates/transaction-pool/src/amm.rs
```diff
@@ -207,10 +204,10 @@ impl AmmLiquidityCache {
 #[derive(Debug, Default)]
 struct AmmLiquidityCacheInner {
     /// Cache for (user_token, validator_token) -> liquidity
-    cache: HashMap<(u64, u64), U256>,
+    cache: HashMap<(Address, Address), U256>,
```
0xKitsune, line 207: nit: we could still use the lower TIP20 bytes rather than the whole address with the prefix. Not sure if there is a measurable/worthwile difference though.

crates/precompiles/src/tip20_factory/mod.rs
```diff
@@ -34,91 +58,134 @@ impl TIP20Factory {
...
+        // Ensure that the quote token is a valid TIP20 that is currently deployed or the quote
+        // quote_token must be a valid TIP20 or address(0)
+        if !(self.is_tip20(quote_token)? || quote_token.is_zero()) {
+            return Err(TIP20Error::invalid_quote_token().into());
+        }
+
+        let mut token = TIP20Token::from_address(address)?;
+        token.initialize(name, symbol, currency, Address::ZERO, admin)?;
+
+        self.emit_event(TIP20FactoryEvent::TokenCreated(
```
legion2002, line 168: This event might be confusing offchain, because the salt address derivation for reserved addresses won't be correct.

legion2002 (approved):
lgtm

PR #1820, opened 2025-12-31
fix: monitor pool creation by listening events + use token WL

fgimenez:
## Motivation

This PR reduces the computational complexity of the sidecar monitor, which currently performs O(n²) requests per monitoring cycle, where `n` represents the number of registered tokens.

It does so by listening `FeeAMM::Mint` events, and tracking new pools (if we emitted a `PoolCreated` event, we could monitor that instead and slightly simplify).

Additionally, it requires users to define a set of target tokens to be monitored, so that we filter out irrelevant pools

### DIFF
--- a/crates/payload/builder/src/lib.rs
+++ b/crates/payload/builder/src/lib.rs
@@ -255,8 +255,7 @@ where
         let block_gas_limit: u64 = parent_header.gas_limit();
         let shared_gas_limit = block_gas_limit / TEMPO_SHARED_GAS_DIVISOR;
         let non_shared_gas_limit = block_gas_limit - shared_gas_limit;
-        let general_gas_limit =
-            (parent_header.gas_limit() - shared_gas_limit) / TEMPO_GENERAL_GAS_DIVISOR;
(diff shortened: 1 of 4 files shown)


bin/tempo-sidecar/src/monitor/mod.rs
```diff
@@ -178,3 +319,13 @@ pub async fn prometheus_metrics(handle: poem::web::Data<&PrometheusHandle>) -> R
         .header("content-type", "text/plain")
         .body(metrics)
 }
+
+/// Parses user and validator token addresses from a `FeeAMM::Mint` event log.
+///
+/// WARNING: Caller is responsible for ensuring the input is a `FeeAMM::Mint` event.
```
0xKitsune, line 325: We could also simply validate the event signature here.

0xKitsune (approved):
In general this lgtm, this is a pretty simple service and we can probably update this to be more efficient with another follow-on PR but this should work for a small set of tokens.

J1939AZ-code (commented):
Ok

PR #1821, opened 2025-12-31
feat: add full dkg mechanism 

joshieDo:
* on top of https://github.com/tempoxyz/tempo/pull/1680
* uses the precompile to know when to execute a full/fresh dkg ceremony 

How it works
* Outcome in the last block of `epoch_N-1` will specify if the next epoch will execute a full dkg ceremony
* Outcome in the last block of `epoch_N`  will have the new public
* `epoch_N` will be using the new public

### DIFF
--- a/crates/commonware-node/src/dkg/manager/actor/mod.rs
+++ b/crates/commonware-node/src/dkg/manager/actor/mod.rs
@@ -1171,12 +1171,10 @@ where
         // Check if next ceremony should be full.
         // Read from pre-last block of the epoch, but never ahead of the current request.
         let next_epoch = state.epoch.next();
-        let is_next_full_dkg = validators::read_next_full_dkg_ceremony(
-            &self.config.execution_node,
-            request.height,
-        )
-        // in theory it should never fail, but if it does, just stick to reshare.
-        .is_ok_and(|epoch| epoch == next_epoch.get());
+        let is_next_full_dkg =
+            validators::read_next_full_dkg_ceremony(&self.config.execution_node, request.height)
+                // in theory it should never fail, but if it does, just stick to reshare.
+                .is_ok_and(|epoch| epoch == next_epoch.get());
         if is_next_full_dkg {
             info!(%next_epoch, "next DKG will change the network identity and not be a reshare process");
         }


crates/dkg-onchain-artifacts/src/lib.rs
```diff
@@ -34,6 +34,10 @@ pub struct OnchainDkgOutcome {
     /// The next players. These will be the players in the DKG ceremony running
     /// during `epoch`.
     pub next_players: ordered::Set<PublicKey>,
+
+    /// Whether the next DKG ceremony should be a full ceremony (new polynomial)
+    /// instead of a reshare. Set when `nextFullDkgCeremony == epoch`.
+    pub is_next_full_dkg: bool,
```
SuperFluffy, line 40: Should this be an epoch number instead of a boolean?

crates/commonware-node/src/dkg/manager/actor/state.rs
```diff
@@ -360,26 +360,31 @@ where
...
             warn!(
                 "we are a dealer in this round, but we do not have a share, \
                 which means we likely lost it; will not instantiate a dealer \
                 instance and hope to get a new share in the next round if we \
                 are a player"
             );
             return Ok(None);
-        }
+        } else {
+            share
```
SuperFluffy, line 380: Let's emit an event here that we are running a full re-dkg dealer (at info level).

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -726,8 +729,14 @@ where
         // state), then we know the DKG failed.
         if onchain_outcome.output == state.output {
             self.metrics.failures.inc();
+            if state.is_full_dkg {
+                self.metrics.full_failures.inc();
```
SuperFluffy: Let's just have a single counter for full re-DKGs instead of 2 new metrics for failures and successes.

crates/commonware-node/src/dkg/manager/validators.rs
```diff
@@ -206,3 +212,34 @@ impl std::fmt::Display for DecodedValidator {
...
+/// (new polynomial) instead of a reshare.
+#[instrument(
+    skip_all,
+    fields(
+        attempt = _attempt,
+        at_height,
+    ),
+    err
+)]
+pub(super) async fn read_next_full_dkg_ceremony(
```
SuperFluffy: I don't think this is async?

crates/commonware-node/src/dkg/manager/validators.rs
```diff
@@ -206,3 +212,34 @@ impl std::fmt::Display for DecodedValidator {
...
+///
+/// This is used to determine if the next DKG ceremony should be a full ceremony
+/// (new polynomial) instead of a reshare.
+#[instrument(
+    skip_all,
+    fields(
+        attempt = _attempt,
+        at_height,
+    ),
+    err
```
SuperFluffy: Add `ret` here, then you don't need the event in the body.

crates/commonware-node/src/dkg/manager/validators.rs
```diff
@@ -206,3 +212,34 @@ impl std::fmt::Display for DecodedValidator {
...
+    node: &TempoFullNode,
+    at_height: u64,
+) -> eyre::Result<u64> {
+    let next_full_dkg_ceremony = read_validator_config_at_height(node, at_height, |config| {
+        config
+            .get_next_full_dkg_ceremony()
+            .wrap_err("failed to query contract for next full dkg ceremony")
+    })?;
+
+    info!(
```
SuperFluffy: You can remove this if you add `ret` to instrumentation.

crates/commonware-node/src/dkg/manager/validators.rs
```diff
@@ -19,55 +19,32 @@ use tempo_precompiles::{
...
-    fields(
-        attempt = _attempt,
-        epoch = epoch.map(tracing::field::display),
-    ),
-    err
-)]
-pub(super) async fn read_from_contract_on_epoch_boundary(
-    _attempt: u32,
+/// Reads state from the ValidatorConfig precompile at a given block height.
+fn read_validator_config_at_height<T>(
```
SuperFluffy, line 22: `read_from_contract_on_epoch_boundary` essentially becomes this function in #1838


Also - what if this function always returned both the validators and the next re-dkg height? We have constructed the evm state already, why not just run both contract calls? Would make this function simpler because we didn't need the closure.

crates/commonware-node/src/dkg/manager/actor/state.rs
```diff
@@ -360,26 +360,31 @@ where
...
         share: Option<Share>,
         seed: Summary,
+        is_full_dkg: bool,
     ) -> eyre::Result<Option<Dealer>> {
         if round.dealers.position(&me.public_key()).is_none() {
             return Ok(None);
         }
 
-        if share.is_none() {
+        let share = if is_full_dkg {
```
SuperFluffy: Eh, this is a bit ugly/roundabout because of the way I wrote this originally. No idea how to make this cleaner.

crates/commonware-node/src/dkg/manager/actor/state.rs
```diff
@@ -360,26 +360,31 @@ where
         round: Round,
         share: Option<Share>,
         seed: Summary,
+        is_full_dkg: bool,
```
SuperFluffy: You could stuff this into the `round: Round` argument instead.

crates/commonware-node/src/dkg/manager/actor/state.rs
```diff
@@ -954,14 +964,21 @@ pub(super) struct Round {
 
 impl Round {
     pub(super) fn from_state(state: &State, namespace: &[u8]) -> Self {
+        // For full DKG, don't pass the previous output - this creates a new polynomial
+        let previous_output = if state.is_full_dkg {
```
SuperFluffy, line 969: Could put the `is_full_dkg` flag into this `Round` bit here as well.

crates/e2e/src/tests/dkg/full_ceremony.rs
```diff
@@ -0,0 +1,246 @@
...
+            );
+            tracing::info!("Verified: reshare preserved polynomial (full DKG only ran once)");
+        })
+    }
+
+    /// Waits for and reads the DKG outcome from the last block of the given epoch.
+    async fn wait_for_outcome(
+        &self,
+        context: &Context,
+        validators: &[TestingNode<Context>],
```
SuperFluffy, line 127: Maybe just pass in one of them if we only read the first?

crates/e2e/src/tests/dkg/full_ceremony.rs
```diff
@@ -0,0 +1,246 @@
...
+        let mut at_epoch = 0;
+        let mut full_successes = 0;
+
+        for line in metrics.lines() {
+            let Some((metric, value)) = parse_metric_line(line) else {
+                continue;
+            };
+
+            // Assert no DKG failures
+            if metric.ends_with("_dkg_manager_ceremony_failures_total")
```
SuperFluffy: Mentioned when you defined the metrics: IMO remove these. I think it completely sufficient to a) assert tht all DKG ceremonies in this test were a success, and that b) you observed a different polynomial identity. I think having an extra metric that triggers extremely rarely is a bit overkill.

xtask/src/genesis_args.rs
```diff
@@ -128,6 +128,7 @@ impl ConsensusConfig {
                 self.validators.iter().map(Validator::public_key),
             )
             .unwrap(),
+            is_next_full_dkg: false,
```
SuperFluffy, line 134: We could make this configurable to force a full DKG on start.

SuperFluffy (approved):
Looks fine to me. The test seems a bit overkill and I am on the fence about the metric, but other than that it's great!

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -1014,15 +1014,13 @@ where
...
 
-        // Check if next ceremony should be full
+        // Check if next ceremony should be full.
+        // Read from pre-last block of the epoch, but never ahead of the current request.
         let next_epoch = state.epoch.next();
         let is_next_full_dkg = validators::read_next_full_dkg_ceremony(
-            1,
             &self.config.execution_node,
-            request.height.saturating_sub(1),
+            epoch_info.last().saturating_sub(1).min(request.height),
```
SuperFluffy: I think you can safely replace this by `request.height`. I don't think this will ever read the boundary, unless I am missing something.

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -1014,15 +1014,13 @@ where
...
         )
-        .await
-        // in theory it should never fail, but if it does, just stick to reshare.
         .is_ok_and(|epoch| epoch == next_epoch.get());
         if is_next_full_dkg {
             info!(%next_epoch, "next DKG ceremony will be full");
```
SuperFluffy: Can you also add this to the `warn!` line in the application actor? Under `fn verify_header_extra_data` if the local outcome doesn't match the proposer's.

PR #1822, opened 2025-12-31
feat!(consensus): fast sync

SuperFluffy:
Adds a fast-sync mechanism that works by skipping outdated epochs, jumping from epoch boundary to epoch boundary.

Note that this implementation covers the general case that does not require operator intervention.

## Assumptions and drawbacks

1. there still exists at least 1 peer from the validator set known to the node with the same public key and socket address (so either what's available in the genesis block or in the node's state when it shut down).
2. There exists at least 1 peer that is not fast-synced, i.e. a peer that has the boundary certificates.
3. the other nodes are still able to serve finalization certificates and boundary blocks (they have not yet pruned their state).

### Future epoch signaling

1. if on restart a node is running in epoch `E` and it observes a vote from some peer `P` in a future epoch `E' > E`, it hints to its sync system (its "marshal" actor) that it should try and fetch the finalization certificate for the boundary height of epoch `E` from peer `P`.
4. once the marshal actor has verified the certificate for the boundary, it will emit an `Update::Tip` event with the boundary height and digest (= block hash).
5. the node will then read the DKG outcome from this boundary block, register the scheme for the next epoch, and "soft enter" that epoch without spinning up a consensus engine backing it.
6. once this is done, the node goes back to 1. observing futures votes.

### DKG behavior

1. while running a DKG round for epoch `E`, the node constantly reads marshal update events.
2. if it detects an update event for an epoch that is at least 2 ahead of its current epoch (i.e. `E' >= E + 2), it sets marshal's finalized floor to the boundary of `E` and skips processing all blocks except for the boundary block.
3. it then reads the DKG outcome of epoch `E`, giving state to enter `E+1`.
4. it then enters the DKG round for the `E+1` and goes back to 1.

The condition `E' >= E+2` was chosen because there no longer is a point in trying to recover the node's share of the group polynomial, since the historic epochs are over.

The node will attempt to recover its share for epoch `E'` by entering epoch `E'-1`.

### DIFF
--- a/crates/commonware-node/src/dkg/manager/actor/mod.rs
+++ b/crates/commonware-node/src/dkg/manager/actor/mod.rs
@@ -256,14 +256,7 @@ where
             // If there is no finalized block for this epoch, then we must not
             // have observed the first one. Therefore we need to start a
             // consensus engine to ensure we register peers, schemes, etc.
-            self.config
-                .epoch_manager
-                .enter(
-                    previous.epoch,
-                    previous.output.public().clone(),
-                    previous.share.clone(),
-                    previous.dealers.keys().clone(),
-                )
+            self.enter_epoch(&previous)
                 .wrap_err("could not instruct epoch manager to enter the previous epoch")?;
         }
 
@@ -287,14 +280,7 @@ where
             .update(state.epoch.get(), all_peers)
             .await;
 
-        self.config
-            .epoch_manager
-            .enter(
-                state.epoch,
-                state.output.public().clone(),
-                state.share.clone(),
-                state.dealers.keys().clone(),
-            )
+        self.enter_epoch(&state)
             .wrap_err("could not instruct epoch manager to enter a new epoch")?;
 
         // TODO: emit an event with round info
@@ -1189,6 +1175,18 @@ where
         };
 
         None
+    }
+
+    fn enter_epoch(&mut self, state: &state::State) -> eyre::Result<()> {
+        self.config
+            .epoch_manager
+            .enter(
+                state.epoch,
+                state.output.public().clone(),
+                state.share.clone(),
+                state.dealers.keys().clone(),
+            )
+            .wrap_err("could not instruct epoch manager to enter epoch")
     }
 }
 


crates/e2e/src/tests/restart.rs
```diff
@@ -298,6 +332,25 @@ fn validator_catches_up_across_epochs() {
...
+
+#[test_traced("WARN")]
+fn validator_catches_up_with_gap_of_three_epochs() {
+    let _ = tempo_eyre::install();
+
+    let epoch_length = 30;
+    let setup = RestartSetup {
+        node_setup: Setup::new()
+            .epoch_length(epoch_length)
+            .connect_execution_layer_nodes(true),
```
SuperFluffy, line 349: Since this is skipping blocks in the CL, we must be connected at the EL level so that the EL can fetch blocks from its peers.

crates/e2e/src/tests/restart.rs
```diff
@@ -298,6 +332,25 @@ fn validator_catches_up_across_epochs() {
         shutdown_height: epoch_length + 1,
         restart_height: 2 * epoch_length + 1,
         final_height: 3 * epoch_length + 1,
+        assert_skips: false,
+    };
+
+    let _state = run_restart_test(setup);
+}
+
+#[test_traced("WARN")]
```
SuperFluffy, line 341: This is generating too many logs which freaks out nextest/git.

crates/e2e/src/tests/restart.rs
```diff
@@ -298,6 +332,25 @@ fn validator_catches_up_across_epochs() {
...
+fn validator_catches_up_with_gap_of_three_epochs() {
+    let _ = tempo_eyre::install();
+
+    let epoch_length = 30;
+    let setup = RestartSetup {
+        node_setup: Setup::new()
+            .epoch_length(epoch_length)
+            .connect_execution_layer_nodes(true),
+        shutdown_height: epoch_length + 1,
+        restart_height: 4 * epoch_length + 1,
```
SuperFluffy, line 351: Need to wait a sufficiently long time to trigger the fast sync mechanism (>2 epochs, 3 here to be on the safe side). Unfortunately this means the test takes quite a while.

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -746,6 +827,77 @@ where
...
+            &self.config.epoch_strategy,
+            &self.metrics.attempts_to_read_validator_contract,
+        )
+        .await;
+
+        Ok(Some(state::State {
+            epoch: onchain_outcome.epoch,
+            seed: Summary::random(&mut self.context),
+            output: onchain_outcome.output.clone(),
+            share: None,
```
joshieDo, line 888: i wonder if we should support an existing validator who knows its share (or was revealed) to participate. not for this PR

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -746,6 +827,77 @@ where
...
+        info!("reading validators from contract");
+
+        let all_validators = read_validator_config_with_retry(
+            &self.context,
+            &self.config.execution_node,
+            round.epoch(),
+            &self.config.epoch_strategy,
+            &self.metrics.attempts_to_read_validator_contract,
+        )
+        .await;
```
joshieDo, line 882: this will never resolve in a full node, because we dont have historical state for each epoch boundary, at least with current impl  of `read_validator_config_with_retry`

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -1236,13 +1410,20 @@ async fn read_validator_config_with_retry<C: commonware_runtime::Clock>(
...
+                    attempts,
+                    retry_after = %tempo_telemetry_util::display_duration(retry_after),
+                    "reading validator config from contract failed; will retry",
+                );
+            } else {
+                error!(
+                    attempts,
+                    retry_after = %tempo_telemetry_util::display_duration(retry_after),
+                    "reading validator config from contract failed; will retry",
+                );
```
joshieDo, line 1440: is this meant to break?

crates/commonware-node/src/dkg/manager/actor/mod.rs
SuperFluffy: I think this is fine. We need to have some sort of synchronization bound.

But I can make use of the fact that the on-chain validator set is immutable and just try and read the latest state.

I will leave this to a follow-up so we have a separate PR - the behavior in this PR is consistent with what's on main.

crates/commonware-node/src/dkg/manager/actor/mod.rs
SuperFluffy: No - I meant to escalate the log level so that it's visible to the operator. Unfortunately the `event!` macro requires a const argument for the level and so I need to have a separate branch.

If the node is unable to read the contract then there is really no point in continuing and it will just loop here.

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -240,16 +258,13 @@ where
...
-                        public: previous.output.public().clone(),
-                        share: previous.share.clone(),
-                        participants: previous.dealers.keys().clone(),
-                    }
-                    .into(),
+                .enter(
+                    previous.epoch,
+                    previous.output.public().clone(),
+                    previous.share.clone(),
+                    previous.dealers.keys().clone(),
```
joshieDo: enter(&previous)?

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -274,16 +289,13 @@ where
...
             .epoch_manager
-            .report(
-                EpochTransition {
-                    epoch: state.epoch,
-                    public: state.output.public().clone(),
-                    share: state.share.clone(),
-                    participants: state.dealers.keys().clone(),
-                }
-                .into(),
+            .enter(
```
joshieDo: enter(&state)?

crates/commonware-node/src/epoch/manager/actor.rs
```diff
@@ -418,120 +400,124 @@ where
...
-            .await
-        {
-            warn!(
-                reason = ?error,
-                "failed forwarding finalization certificate to requester"
-            );
-            bail!("failed forwarding finalization certificate to requester");
-        }
-        Ok(())
+            .hint_finalized(boundary_height, NonEmptyVec::new(from))
```
joshieDo, line 519: if we have more validators, should we just ask all? there's a chance this is also a fast-synced-node that wouldnt be able to give us what we need, and we'd need to wait for a msg coming from a not-fast-synced-node

SuperFluffy: The idea here is that the node could be lying and causing us to spam the other nodes, asking for the certificate - subsequently leading to us getting blocked.

crates/commonware-node/src/dkg/manager/actor/mod.rs
SuperFluffy: Makes sense, added in [78696ed](https://github.com/tempoxyz/tempo/pull/1822/commits/78696edaf8e34d2fe40961853696cf7f92e47ef6)

crates/commonware-node/src/dkg/manager/actor/mod.rs
SuperFluffy: added in [78696ed](https://github.com/tempoxyz/tempo/pull/1822/commits/78696edaf8e34d2fe40961853696cf7f92e47ef6)

PR #1836, opened 2026-01-02
feat(xtask): add genesis customization options

Zygimantass:
- Add --pathusd-admin and --validator-admin for separate admin addresses
- Add --validator-addresses to manually set validator onchain addresses
- Add --no-extra-tokens to disable Alpha/Beta/ThetaUSD creation
- Add --no-pairwise-liquidity to disable FeeAMM liquidity minting

### DIFF
--- a/xtask/src/genesis_args.rs
+++ b/xtask/src/genesis_args.rs
@@ -279,7 +279,8 @@ impl GenesisArgs {
             validator_admin,
             &mut evm,
             &consensus_config,
-            &addresses,
+            // Skip first address (used as default admin)
+            &addresses[1..],
             validator_onchain_addresses,
             self.no_dkg_in_genesis,
         )?;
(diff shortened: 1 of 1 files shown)


xtask/src/genesis_args.rs
```diff
@@ -238,18 +270,24 @@ impl GenesisArgs {
...
+            Some(&self.validator_addresses[..])
+        };
         initialize_validator_config(
-            admin,
+            validator_admin,
             &mut evm,
             &consensus_config,
-            // Skip admin
-            &addresses[1..],
+            &addresses,
```
SuperFluffy: This is going to set the first address to the admin address. Is that desirable? That's why it was `&addresses[1..]` originally

xtask/src/genesis_args.rs
```diff
@@ -674,19 +723,32 @@ fn initialize_validator_config(
...
             for (i, validator) in consensus_config.validators.iter().enumerate() {
                 #[expect(non_snake_case, reason = "field of a snakeCase smart contract call")]
-                let newValidatorAddress = *addresses.get(i).ok_or_else(|| {
-                    eyre!(
-                        "need `{}` addresses for all validators, but only `{}` were generated",
-                        consensus_config.validators.len(),
-                        addresses.len()
-                    )
-                })?;
+                let newValidatorAddress = if let Some(custom_addrs) = custom_validator_addresses {
```
SuperFluffy: Let's make this block more elegant by binding this before:
```rust
    let addrs = custom_validator_addresses.as_ref().unwrap_or(fallback_addresses);
```

xtask/src/genesis_args.rs
```diff
@@ -674,19 +723,32 @@ fn initialize_validator_config(
...
+                    custom_addrs[i]
+                } else {
+                    *fallback_addresses.get(i).ok_or_else(|| {
+                        eyre!(
+                            "need `{}` addresses for all validators, but only `{}` were generated",
```
SuperFluffy: Check the length of `addrs` before the loop, then we can get rid of this on every iteration.

PR #1838, opened 2026-01-02
feat(consensus): start at the latest finalized height, not at genesis

SuperFluffy:
This patch changes the cold-start behavior of a node: if a node starts without any consensus state but with advanced execution layer state (for example when having been provided with a snapshot), it determines the scheme for its current epoch by reading the header of the last available boundary block.

How this is tested:

Provided a test where the consensus states of a new is deleted before the node is restarted. This simulates a start from a snapshot. The test enforces that the node never starts a consensus engine for the epoch below the one it was restarted at.

### DIFF
--- a/crates/commonware-node/src/dkg/manager/actor/mod.rs
+++ b/crates/commonware-node/src/dkg/manager/actor/mod.rs
@@ -1470,11 +1470,6 @@ impl Metrics {
     }
 }
 
-enum NewState {
-    FromSkip(state::State),
-    Normal(state::State),
-}
-
 /// Attempts to read the validator config from the smart contract until it becomes available.
 async fn read_validator_config_with_retry<C: commonware_runtime::Clock>(
     context: &C,


crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -621,9 +636,20 @@ impl Inner<Init> {
...
+                    .provider
+                    .block_by_number(0)
+                    .map_or_else(
+                        |e| Err(eyre::Report::new(e)),
+                        |block| block.ok_or_eyre("execution layer did not have block"),
+                    )
+                    .wrap_err("execution layer did not have the genesis block")?
+                    .seal(),
+            );
+            Either::Left(always_ready(move || Ok(genesis_block.clone())))
```
joshieDo: dupped from above, move to helper?

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -1179,26 +1192,86 @@ where
...
+            |header| header.ok_or_eyre("execution layer reported it had no header"),
+        )
+        .wrap_err_with(|| {
+            format!("failed to read header for last boundary block number `{last_boundary}`")
+        })?;
+    let onchain_outcome =
+        tempo_dkg_onchain_artifacts::OnchainDkgOutcome::read(&mut header.extra_data().as_ref())
+            .wrap_err("the boundary header did not contain the on-chain DKG outcome")?;
+
+    let all_validators = validators::read_from_contract_at_block(0, node, newest_height)
```
joshieDo, line 1257:
```suggestion
    let all_validators = validators::read_from_contract_at_block(last_boundary, node, newest_height)
```

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -1179,26 +1192,86 @@ where
...
+        .await
+        .wrap_err_with(|| {
+            format!("failed reading validator config from block height `{newest_height}`")
```
joshieDo, line 1260:
```suggestion
            format!("failed reading validator config from block height `{last_boundary}`")
```

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -1179,26 +1192,86 @@ where
...
+        })?;
 
-    let all_validators =
-        validators::read_from_contract_on_epoch_boundary(0, node, None, epoch_strategy)
-            .await
-            .wrap_err("the genesis block did not contain a validator config")?;
+    let share = 'verify_initial_share: {
```
joshieDo, line 1263: if None and newest == 0, maybe we panic? shouldnt happen

crates/commonware-node/src/dkg/manager/actor/mod.rs
SuperFluffy: I have to fix the signature of this method, it's confusing - the first argument is the attempt number. That's because this function was originally only called from inside a loop.

The `newest_height` argument makes use of the fact that the on-chain contract is immutable.

crates/commonware-node/src/dkg/manager/actor/mod.rs
SuperFluffy: `newest_height is actually correct here`. But I see how that can be confusing.

crates/commonware-node/src/dkg/manager/actor/mod.rs
joshieDo: oh yeah my bad

crates/commonware-node/src/dkg/manager/actor/mod.rs
SuperFluffy: Don't think we should - this is the "sync from genesis" functionality that we were using so far. We shouldn't want to sync from genesis, but I wouldn't rule it out entirely.

crates/commonware-node/src/consensus/application/actor.rs
SuperFluffy: Makes sense - added in [c6cb79b](https://github.com/tempoxyz/tempo/pull/1838/commits/c6cb79b3ac067121a2d3349f6e605682620c58a3)

PR #1839, opened 2026-01-02
feat(cli): implement consensus subcommand as ExtSubCmd

Zygimantass:
Use reth's new extensible CLI subcommand mechanism to add the consensus subcommand. This removes the try_run_tempo_subcommand workaround and integrates properly with reth's CLI infrastructure.

- Add TempoSubcommand enum implementing ExtendedCommand trait
- Update Cli generic to include TempoSubcommand as SubCmd parameter

Amp-Thread-ID: https://ampcode.com/threads/T-019b7efe-dd0e-73ed-903f-74a67dfbd7e0

### DIFF
--- a/bin/tempo/src/main.rs
+++ b/bin/tempo/src/main.rs
@@ -27,8 +27,8 @@ use eyre::WrapErr as _;
 use futures::{FutureExt as _, future::FusedFuture as _};
 use reth_ethereum::{chainspec::EthChainSpec as _, cli::Commands, evm::revm::primitives::B256};
 use reth_ethereum_cli::Cli;
-use reth_rpc_server_types::DefaultRpcModuleValidator;
 use reth_node_builder::{NodeHandle, WithLaunchContext};
+use reth_rpc_server_types::DefaultRpcModuleValidator;
 use std::{sync::Arc, thread};
 use tempo_chainspec::spec::{TempoChainSpec, TempoChainSpecParser};
 use tempo_commonware_node::{feed as consensus_feed, run_consensus_stack};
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -241,81 +242,6 @@ rayon = "1.10"
...
-# reth-tasks = { path = "../reth/crates/tasks" }
-# reth-tokio-util = { path = "../reth/crates/tokio-util" }
-# reth-tracing = { path = "../reth/crates/tracing" }
-# reth-transaction-pool = { path = "../reth/crates/transaction-pool" }
-# reth-trie = { path = "../reth/crates/trie/trie" }
-# reth-trie-common = { path = "../reth/crates/trie/common" }
-# reth-trie-db = { path = "../reth/crates/trie/db" }
-# reth-trie-parallel = { path = "../reth/crates/trie/parallel" }
-# reth-trie-sparse = { path = "../reth/crates/trie/sparse" }
-# reth-trie-sparse-parallel = { path = "../reth/crates/trie/sparse-parallel" }
```
shekhirin, line 320: can we revert this patch section removal, it's convenient to have it to test stuff locally

crates/payload/types/src/lib.rs
```diff
@@ -59,6 +59,10 @@ impl ExecutionPayload for TempoExecutionData {
         self.block.timestamp()
     }
 
+    fn transaction_count(&self) -> usize {
+        self.block.body().transaction_count()
```
SuperFluffy, line 63: Eh, this stuff is why I am not a fan of bumping deps while adding features.

PR #1854, opened 2026-01-04
feat: use 12 byte `TIP20_PREFIX`

0xKitsune:
This PR updates the `TIP20_PREFIX` to use 12 bytes instead of 10 and fixes small nits with #1819.

### DIFF
--- a/crates/node/tests/it/tip_fee_manager.rs
+++ b/crates/node/tests/it/tip_fee_manager.rs
@@ -10,14 +10,14 @@ use alloy::{
 };
 use alloy_eips::{BlockId, Encodable2718};
 use alloy_network::{AnyReceiptEnvelope, EthereumWallet};
-use alloy_primitives::{Address, B256, Signature, U256, address};
+use alloy_primitives::{Address, Signature, U256, address};
 use alloy_rpc_types_eth::TransactionRequest;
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/tip20_factory/mod.rs
```diff
@@ -129,26 +131,46 @@ impl TIP20Factory {
...
+    /// Creates a token at a reserved address
+    /// Internal function used to deploy TIP20s at reserved addresses at genesis or hardforks
+    pub fn create_token_reserved_address(
+        &mut self,
+        address: Address,
+        name: &str,
+        symbol: &str,
+        currency: &str,
+        admin: Address,
+    ) -> Result<Address> {
```
0xrusowsky, line 144: nice!

while PathUSD will use `address(0)` as its quote token, should we support custom quote tokens for future reserved token addrs? probably overkill, but flagging just in case

0xKitsune: Ive updated this function to specify a quote token and check if that token is either `address(0)` or a valid TIP20. I think this should be OK since the quote token is only used to route pairs in the stable dex. If the quote token is zero, creating a pair should fail.

PR #1855, opened 2026-01-05
fix: properly check if an account is cold when loading its info

0xrusowsky:
This PR ensures that `with_account_info` properly charges COLD account access cost.

### DIFF
--- a/crates/precompiles-macros/src/storable.rs
+++ b/crates/precompiles-macros/src/storable.rs
@@ -187,13 +187,16 @@ fn gen_packing_module_from_ir(fields: &[LayoutField<'_>], mod_ident: &Ident) -> 
     let last_field = &fields[fields.len() - 1];
     let last_slot_const = PackingConstants::new(last_field.name).slot();
     let packing_constants = packing::gen_constants_from_ir(fields, true);
+    let last_type = &last_field.ty;
 
     quote! {
         pub mod #mod_ident {
             use super::*;
 
             #packing_constants
-            pub const SLOT_COUNT: usize = (#last_slot_const.saturating_add(::alloy::primitives::U256::ONE)).as_limbs()[0] as usize;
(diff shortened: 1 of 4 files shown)


crates/precompiles/src/storage/evm.rs
```diff
@@ -66,7 +75,7 @@ impl<'a> PrecompileStorageProvider for EvmPrecompileStorageProvider<'a> {
 
     #[inline]
     fn set_code(&mut self, address: Address, code: Bytecode) -> Result<(), TempoPrecompileError> {
-        self.ensure_loaded_account(address)?;
+        _ = self.ensure_loaded_account(address)?;
```
rakita: til this compiles

rakita (approved):
lgtm

ensure_loaded_account is going to be removed with upcoming [revm bump](https://github.com/tempoxyz/tempo/pull/1762/changes)

crates/precompiles/src/storage/evm.rs
```diff
@@ -80,9 +89,8 @@ impl<'a> PrecompileStorageProvider for EvmPrecompileStorageProvider<'a> {
         address: Address,
         f: &mut dyn FnMut(&AccountInfo),
     ) -> Result<(), TempoPrecompileError> {
-        self.ensure_loaded_account(address)?;
+        let is_cold = self.ensure_loaded_account(address)?;
         let account = self.internals.load_account_code(address)?.map(|a| &a.info);
-        let is_cold = account.is_cold;
```
rakita, line 85: As this only reads the account and its code, we could remove `self.ensure_loaded_account(address)?;` and just use  `self.internals.load_account_code(address)?.map(|a| &a.info);`. If the account is not inside revm it would get loaded from db.

And as account is not changed, touching it does nothing.

PR #1857, opened 2026-01-05
fix(genesis): make the validator address logic simpler

Zygimantass:
Refactoring the `initialize_validator_config` to remove the dual-path address resolution and removing `fallback_addresses` + `custom_validator_addresses`.

### DIFF
--- a/xtask/src/genesis_args.rs
+++ b/xtask/src/genesis_args.rs
@@ -203,8 +203,6 @@ impl GenesisArgs {
 
         // system contracts/precompiles must be initialized bottom up, if an init function (e.g. mint_pairwise_liquidity) uses another system contract/precompiles internally (tip403 registry), the registry must be initialized first.
 
-        // Deploy TestUSD fee token
-        println!("{:?}", self.pathusd_admin);
         let pathusd_admin = self.pathusd_admin.unwrap_or_else(|| addresses[0]);
         let validator_admin = self.validator_admin.unwrap_or_else(|| addresses[0]);
         let mut evm = setup_tempo_evm(self.chain_id);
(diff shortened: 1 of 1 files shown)


xtask/src/genesis_args.rs
```diff
@@ -204,8 +204,9 @@ impl GenesisArgs {
         // system contracts/precompiles must be initialized bottom up, if an init function (e.g. mint_pairwise_liquidity) uses another system contract/precompiles internally (tip403 registry), the registry must be initialized first.
 
         // Deploy TestUSD fee token
-        let pathusd_admin = self.pathusd_admin.unwrap_or(addresses[0]);
-        let validator_admin = self.validator_admin.unwrap_or(addresses[0]);
+        println!("{:?}", self.pathusd_admin);
+        let pathusd_admin = self.pathusd_admin.unwrap_or_else(|| addresses[0]);
+        let validator_admin = self.validator_admin.unwrap_or_else(|| addresses[0]);
```
0xKitsune, line 207: nit: why update to `unwrap_or_else` rather than `unwrap_or`?

xtask/src/genesis_args.rs
```diff
@@ -275,16 +276,15 @@ impl GenesisArgs {
 
         println!("Initializing validator config");
         let validator_onchain_addresses = if self.validator_addresses.is_empty() {
-            None
+            &addresses[1..self.validators.len() + 1]
```
0xKitsune, line 282: This can cause OOB if a user passes in a number of addresses < validators.

0xKitsune (approved):

xtask/src/genesis_args.rs
Zygimantass: `unwrap_or` immediately tries to grab `addresses[0]`, but addresses is empty if we specify 0 accounts

xtask/src/genesis_args.rs
Zygimantass: fixed

PR #1858, opened 2026-01-05
feat(xtask): add get-dkg-outcome command

kuyziss:
Add `cargo xtask get-dkg-outcome` to dump polynomial commitment info from a block's extra_data as JSON.

Usage:
```cargo xtask get-dkg-outcome --rpc-url <URL> --epoch <N> --epoch-length <N>```

Output includes epoch, boundary_block, constant_term (hex), threshold, and total_participants. Useful for verifying polynomial changes in re-DKG tests.

### CODE REVIEW

File: xtask/Cargo.toml [hunk 1]
```diff
@@ -21,6 +21,8 @@ tempo-revm.workspace = true
 
 alloy = { workspace = true, features = [
   "genesis",
+  "providers",
+  "reqwest",
```
Zygimantass, line 25: use `reqwest-rustls-tls` as well for TLS support

xtask/src/dump_polynomial.rs
```diff
@@ -0,0 +1,84 @@
...
+use alloy::providers::{Provider, ProviderBuilder};
+use commonware_codec::{Encode as _, ReadExt as _};
+use commonware_consensus::types::{Epoch, Epocher as _, FixedEpocher};
+use commonware_utils::NZU64;
+use eyre::{Context as _, eyre};
+use serde::Serialize;
+use tempo_dkg_onchain_artifacts::OnchainDkgOutcome;
+
+#[derive(Debug, clap::Args)]
+pub(crate) struct DumpPolynomial {
```
SuperFluffy: Bit of a misnomer IMO. It's not really dumping a polynomial but a) only its constant term and b) some extra info.

Let's upgrade this to actually pretty-print the full DKG outcome? I think that'd be more useful. I am am thinking dealers, players, next players, the "summary" hash, and if the next DKG would be a full re-dkg.

xtask/src/dump_polynomial.rs
```diff
@@ -0,0 +1,84 @@
...
+    #[arg(long)]
+    rpc_url: String,
+
+    /// Epoch number to query
+    #[arg(long)]
+    epoch: u64,
+
+    /// Epoch length (blocks per epoch)
+    #[arg(long)]
+    epoch_length: u64,
```
SuperFluffy: Probably fine for now, but do note that the moment we adjust the epoch length (which now in theory we can), this will break.

IMO, it'd be more useful (and intuitive?) to give this a height/block number.

xtask/src/dump_polynomial.rs
```diff
@@ -0,0 +1,84 @@
...
+
+impl DumpPolynomial {
+    pub(crate) async fn run(self) -> eyre::Result<()> {
+        let epoch_length = NZU64!(self.epoch_length);
+        let epocher = FixedEpocher::new(epoch_length);
+        let epoch = Epoch::new(self.epoch);
+
+        let boundary_block = epocher
+            .last(epoch)
+            .ok_or_else(|| eyre!("invalid epoch {}", self.epoch))?;
```
SuperFluffy: This should never be the case with the fixed epocher.

xtask/src/dump_polynomial.rs
```diff
@@ -0,0 +1,84 @@
...
+    /// Epoch length (blocks per epoch)
+    #[arg(long)]
+    epoch_length: u64,
+}
+
+#[derive(Serialize)]
+struct PolynomialInfo {
+    epoch: u64,
+    boundary_block: u64,
+    constant_term: String,
```
SuperFluffy: Hm, maybe we call this `network_identity` or `network_public_key`? Unless this leads to confusion with the chain ID? Also so you don't have to do this manually in the method def:

```suggestion
    #[serde(with = "::const_hex::serialize")]
    network_identity: Bytes,
```

xtask/src/dump_polynomial.rs
```diff
@@ -0,0 +1,84 @@
...
+        let provider = ProviderBuilder::new()
+            .connect(&self.rpc_url)
+            .await
+            .wrap_err("failed to connect to RPC")?;
+
+        let block = provider
+            .get_block_by_number(boundary_block.into())
+            .await
+            .wrap_err("failed to fetch block")?
+            .ok_or_else(|| eyre!("block {} not found", boundary_block))?;
```
SuperFluffy:
```suggestion
            map_or_else(
                |err| Err(Report::new(err)),
                || eyre!("block not found"),
            )
            .wrap_err_with(|| format!("failed to fetch block number `{boundary_block}`"))?;
```

xtask/src/main.rs
```diff
@@ -44,11 +46,8 @@ struct Args {
 }
 
 #[derive(Debug, clap::Subcommand)]
-#[expect(
-    clippy::enum_variant_names,
-    reason = "the variant names map to actual cli inputs and are desired"
-)]
 enum Action {
+    DumpPolynomial(DumpPolynomial),
```
SuperFluffy: Rename this to `GetDkgOutcome`?

xtask/src/main.rs
```diff
@@ -2,15 +2,16 @@
...
+    dump_polynomial::DumpPolynomial, generate_devnet::GenerateDevnet,
+    generate_genesis::GenerateGenesis, generate_localnet::GenerateLocalnet,
 };
 
 use alloy::signers::{local::MnemonicBuilder, utils::secret_key_to_address};
 use clap::Parser as _;
 use commonware_codec::DecodeExt;
 use eyre::Context;
 
+mod dump_polynomial;
```
SuperFluffy:
```suggestion
mod get_dkg_outcome;
```

xtask/src/dump_polynomial.rs
SuperFluffy:
```suggestion
pub(crate) struct GetDkgOutcome {
```

xtask/src/dump_polynomial.rs
```diff
@@ -0,0 +1,123 @@
...
+    /// Block number to query directly (use when epoch length varies)
+    #[arg(long, group = "target")]
+    block: Option<u64>,
+
+    /// Epoch number to query (requires --epoch-length)
+    #[arg(long, group = "target", requires = "epoch_length")]
+    epoch: Option<u64>,
+
+    /// Epoch length in blocks (required with --epoch)
+    #[arg(long)]
```
SuperFluffy:
```suggestion
    #[arg(long, requires = "epoch")]

File: xtask/src/dump_polynomial.rs [hunk 10]
```diff
@@ -0,0 +1,123 @@
...
+    /// Dealers in this DKG ceremony (ed25519 public keys)
+    dealers: Vec<String>,
+    /// Players in this DKG ceremony (ed25519 public keys)
+    players: Vec<String>,
+    /// Players for the next DKG ceremony (ed25519 public keys)
+    next_players: Vec<String>,
+    /// Whether the next DKG should be a full ceremony (new polynomial)
+    is_next_full_dkg: bool,
+    /// The shared public polynomial info
+    sharing: SharingInfo,
```
SuperFluffy: IMO flatten all of this into the struct. This being called `sharing` is a CW-ism and IMO confusing.

xtask/src/get_dkg_outcome.rs
```diff
@@ -0,0 +1,123 @@
...
+    epoch_length: Option<u64>,
+}
+
+#[derive(Serialize)]
+struct DkgOutcomeInfo {
+    /// The epoch for which this outcome is used
+    epoch: u64,
+    /// Block number where this outcome was stored
+    boundary_block: u64,
+    /// Dealers in this DKG ceremony (ed25519 public keys)
```
SuperFluffy: These are the dealers that contributed to the outcome of this DKG.

xtask/src/dump_polynomial.rs
```diff
@@ -0,0 +1,123 @@
...
+    #[arg(long)]
+    epoch_length: Option<u64>,
+}
+
+#[derive(Serialize)]
+struct DkgOutcomeInfo {
+    /// The epoch for which this outcome is used
+    epoch: u64,
+    /// Block number where this outcome was stored
+    boundary_block: u64,
```
SuperFluffy: This would be useful:
```suggestion
    block_number: u64,
    block_hash: B256,
```

xtask/src/get_dkg_outcome.rs
```diff
@@ -0,0 +1,123 @@
...
+#[derive(Debug, clap::Args)]
+#[clap(group = clap::ArgGroup::new("target").required(true))]
+pub(crate) struct DumpPolynomial {
+    /// RPC endpoint URL (http://, https://, ws://, or wss://)
+    #[arg(long)]
+    rpc_url: String,
+
+    /// Block number to query directly (use when epoch length varies)
+    #[arg(long, group = "target")]
+    block: Option<u64>,
```
SuperFluffy, line 24: Add the ability to also query by block hash?

xtask/src/dump_polynomial.rs
```diff
@@ -0,0 +1,123 @@
...
+    pub(crate) async fn run(self) -> eyre::Result<()> {
+        let boundary_block = if let Some(block) = self.block {
+            block
+        } else {
+            let epoch = self.epoch.expect("epoch required when block not provided");
+            let epoch_length = self.epoch_length.expect("epoch_length required with epoch");
+            let epocher = FixedEpocher::new(NZU64!(epoch_length));
+            epocher
+                .last(Epoch::new(epoch))
+                .expect("fixed epocher always returns boundary")
```
SuperFluffy:
```suggestion
            .expect("fixed epocher is valid for all epochs")
```

xtask/src/dump_polynomial.rs
```diff
@@ -0,0 +1,123 @@
...
+
+        let block = provider
+            .get_block_by_number(boundary_block.into())
+            .await
+            .wrap_err_with(|| format!("failed to fetch block number `{boundary_block}`"))?
+            .ok_or_else(|| eyre!("block {boundary_block} not found"))?;
+
+        let extra_data = &block.header.inner.extra_data;
+
+        if extra_data.is_empty() {
```
SuperFluffy:
```suggestion
        eyre::ensure!(
            !extra_data.is_empty(),
            "<msg>",
        )
```

xtask/src/dump_polynomial.rs
```diff
@@ -0,0 +1,123 @@
...
+
+#[derive(Serialize)]
+struct DkgOutcomeInfo {
+    /// The epoch for which this outcome is used
+    epoch: u64,
+    /// Block number where this outcome was stored
+    boundary_block: u64,
+    /// Dealers in this DKG ceremony (ed25519 public keys)
+    dealers: Vec<String>,
+    /// Players in this DKG ceremony (ed25519 public keys)
```
SuperFluffy: And these are the players that received a share due to this DKG ceremony (if it was successful that is; we can't actually see this by just looking at the DKG ceremony).

PR #1859, opened 2026-01-05
fix: review fixes

howydev:
1. `create_token_reserved_address` uses the provided input `quoteToken` instead of using address(0)
2. `create_token_reserved_address`  now enforces that deployed token addresses are within the address space
3. `getTokenAddress` reverts if the counterfactual deployment address would fall within the reserve space
4. Sol spec now uses the 12 byte prefix like the rust precompile
5. Added a unit test to test an actual deployment case with fixed sender, salt
6. Moved USD currency checks from tip20 to tip20factory

### DIFF
--- a/crates/precompiles/src/tip20_factory/mod.rs
+++ b/crates/precompiles/src/tip20_factory/mod.rs
@@ -111,7 +111,9 @@ impl TIP20Factory {
         }
 
         // If token is USD, its quote token must also be USD
-        if call.currency == USD_CURRENCY && TIP20Token::from_address(call.quoteToken)?.currency()? != USD_CURRENCY{
+        if call.currency == USD_CURRENCY
+            && TIP20Token::from_address(call.quoteToken)?.currency()? != USD_CURRENCY
+        {
             return Err(TIP20Error::invalid_quote_token().into());
         }
 
@@ -176,7 +178,9 @@ impl TIP20Factory {
                 return Err(TIP20Error::invalid_quote_token().into());
             }
             // If token is USD, its quote token must also be USD
-            if currency == USD_CURRENCY && TIP20Token::from_address(quote_token)?.currency()? != USD_CURRENCY {
+            if currency == USD_CURRENCY
+                && TIP20Token::from_address(quote_token)?.currency()? != USD_CURRENCY
+            {
                 return Err(TIP20Error::invalid_quote_token().into());
             }
         }


crates/precompiles/src/tip20_factory/mod.rs
```diff
@@ -108,6 +110,21 @@ impl TIP20Factory {
             return Err(TIP20Error::invalid_quote_token().into());
         }
 
+        // If token is USD, its quote token must also be USD
+        if call.currency == USD_CURRENCY {
+            let quote_token_currency = TIP20Token::from_address(call.quoteToken)?.currency()?;
+            if quote_token_currency != USD_CURRENCY {
+                return Err(TIP20Error::invalid_quote_token().into());
+            }
+        }
```
0xrusowsky: nit(requires fmt):
```rs
if USD_CURRENCY == call.currency &&
   USD_CURRENCY != TIP20Token::from_address(call.quoteToken)?.currency()?
{
}
```

crates/precompiles/src/tip20_factory/mod.rs
```diff
@@ -156,22 +173,41 @@ impl TIP20Factory {
...
+            if !self.is_tip20(quote_token)? {
+                return Err(TIP20Error::invalid_quote_token().into());
+            }
+            // If token is USD, its quote token must also be USD
+            if currency == USD_CURRENCY {
+                let quote_token_currency = TIP20Token::from_address(quote_token)?.currency()?;
+                if quote_token_currency != USD_CURRENCY {
+                    return Err(TIP20Error::invalid_quote_token().into());
+                }
+            }
```
0xrusowsky, line 185: nit: since we check this twice, maybe we can have a helper method
```rs
fn validate_quote_currency(currency: &str, quote_token: Address) -> Result<()>;
```

crates/precompiles/src/tip20_factory/mod.rs
```diff
@@ -88,13 +97,6 @@ impl TIP20Factory {
         // Compute the deterministic address from sender and salt
         let (token_address, lower_bytes) = compute_tip20_address(sender, call.salt);
 
-        // Check if address is in reserved range
```
legion2002, line 91: Why was this check removed? What stops a user from mining and deploying to a reserved address now?

0xrusowsky (commented):
lgtm, just a couple of minor nits

crates/precompiles/src/tip20_factory/mod.rs
howydev: just shifted this check below to make the ordering match the sol impl

crates/precompiles/src/tip20_factory/mod.rs
howydev: haha, actually we could go further - `create_token_reserved_address` should be a helper for `create_token`

PR #1868, opened 2026-01-06
feat: add `consensus_` namespace to RPC

joshieDo:
```rust
/// Consensus namespace RPC trait.
#[rpc(server, client, namespace = "consensus")]
pub trait TempoConsensusApi {
    /// Get finalization by height query.
    ///
    /// Use `"latest"` to get the most recent finalization, or `{"height": N}` for a specific height.
    #[method(name = "getFinalization")]
    async fn get_finalization(&self, query: Query) -> RpcResult<Option<CertifiedBlock>>;

    /// Get the current consensus state snapshot.
    ///
    /// Returns the latest finalized block and the latest notarized block (if not yet finalized).
    #[method(name = "getLatest")]
    async fn get_latest(&self) -> RpcResult<ConsensusState>;

    /// Subscribe to the following consensus events: Notarized, Finalized, Nullified.
    #[subscription(name = "subscribe" => "event", unsubscribe = "unsubscribe", item = Event)]
    async fn subscribe_events(&self) -> jsonrpsee::core::SubscriptionResult;
}

```


crates/commonware-node/src/feed/state.rs
```diff
@@ -0,0 +1,146 @@
...
+};
+use tokio::sync::broadcast;
+
+const BROADCAST_CHANNEL_SIZE: usize = 1024;
+const MAX_NOTARIZATIONS: u32 = 1024;
+
+/// Internal shared state for the feed.
+pub(super) struct FeedState {
+    /// In-memory notarization cache, keyed by view.
+    pub(super) notarizations: LruMap<u64, CertifiedBlock, ByLength>,
```
joshieDo: i think btreemap would be fine tbh, maybe this is too defensive?

bin/tempo/src/main.rs
```diff
@@ -270,6 +277,11 @@ fn main() -> eyre::Result<()> {
                     ctx.modules.merge_configured(ext.into_rpc())?;
                 }
 
+                if validator_key.is_some() {
+                    ctx.modules
+                        .merge_configured(TempoConsensusRpc::new(cl_feed_state).into_rpc())?;
+                }
+
```
joshieDo, line 284: initially i had it in TempoAddons, but it meant that TempoFullNode would need to hold some kind of additional wrapper type that implements ConsensusFeed , and current changes seemed simpler than the following:

```rust
pub struct ConsensusFeedHandle(Arc<dyn ConsensusFeed>);

impl ConsensusFeedHandle {
    pub fn new<I: ConsensusFeed>(feed: I) -> Self {
        Self(Arc::new(feed))
    }
}

#[async_trait]
impl ConsensusFeed for ConsensusFeedHandle {
    ... 
}
```

crates/node/src/rpc/consensus/types.rs
```diff
@@ -0,0 +1,93 @@
+//! RPC types for the consensus namespace.
+
+use alloy_primitives::B256;
+use async_trait::async_trait;
+use serde::{Deserialize, Serialize};
+use tokio::sync::broadcast;
+
+/// A block with a threshold BLS certificate (notarization or finalization).
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
```
mattsse, line 10: this is currently redundant, but we can keep

bin/tempo/src/main.rs
mattsse: this is fine, we don't have an internal use case for this yet

crates/node/src/rpc/consensus/types.rs
```diff
@@ -0,0 +1,93 @@
...
+pub struct ConsensusState {
+    /// The latest finalized block (if any).
+    pub finalized: Option<CertifiedBlock>,
+    /// All cached notarizations.
+    pub notarized: Vec<CertifiedBlock>,
+}
+
+/// Trait for accessing consensus feed data.
+#[async_trait]
```
mattsse: I believe this isn't necessary for this trait

crates/commonware-node/src/feed/actor.rs
```diff
@@ -0,0 +1,197 @@
...
+                        notarization.proposal.round.epoch().get(),
+                        notarization.proposal.payload,
+                        &notarization.certificate,
+                    )
+                    .await;
+
+                let _ = self.state.events_tx().send(Event::Notarized {
+                    block: block.clone(),
+                    seen,
+                });
```
mattsse, line 121: depending on how efficient this should be and how many subscribers we expect, we can check first if the subscriber_count before cloning the block, unsure how large the block is

crates/commonware-node/src/feed/state.rs
```diff
@@ -0,0 +1,146 @@
...
+            .field("marshal_set", &self.marshal.get().is_some())
+            .field("subscriber_count", &self.events_tx.receiver_count())
+            .finish()
+    }
+}
+
+#[async_trait]
+impl ConsensusFeed for FeedStateHandle {
+    async fn get_notarization(&self, query: Query) -> Result<Option<CertifiedBlock>, QueryError> {
+        let state = self.state.read();
```
mattsse: we don't expect a lot of load on these rpc endpoints I assume?

the lock guards multiple entries, that could be wrapped into their own rwlock/atomic in FeedState, but afaict, just locking state seems fine

mattsse (changes requested):
some nits,

haven't looked too closely at the consensus stack/ feed actor impl

crates/commonware-node/src/feed/actor.rs
joshieDo: validators shouldnt be exposed to outside, so we shouldnt expect many. its more for internal use

crates/commonware-node/src/feed/state.rs
joshieDo: validators shouldnt be exposed to outside, so we shouldnt expect many. its more for internal use

crates/commonware-node/src/feed/state.rs
```diff
@@ -0,0 +1,144 @@
...
+    }
+}
+
+impl ConsensusFeed for FeedStateHandle {
+    async fn get_notarization(&self, query: Query) -> Result<Option<CertifiedBlock>, QueryError> {
+        let state = self.state.read();
+        match query {
+            Query::Latest => Ok(state
+                .latest_notarized_view
+                .and_then(|view| state.notarizations.peek(&view).cloned())),
```
mattsse: yeah this dont use lru ops, so I guess we can do a simple btreemap and pop manually once > certain capacity?

crates/commonware-node/src/feed/state.rs
```diff
@@ -0,0 +1,144 @@
...
+use parking_lot::RwLock;
+use schnellru::{ByLength, LruMap};
+use std::sync::{Arc, OnceLock};
+use tempo_node::rpc::consensus::{
+    CertifiedBlock, ConsensusFeed, ConsensusState, Event, Query, QueryError,
+};
+use tokio::sync::broadcast;
+
+const BROADCAST_CHANNEL_SIZE: usize = 1024;
+const MAX_NOTARIZATIONS: u32 = 1024;
```
SuperFluffy: This could follow the marshal actor. We are using `view_retention_timeout = ViewDelta::new(config.views_to_track * SYNCER_ACTIVITY_TIMEOUT_MULTIPLIER)`

(although I think there is no use caching notarizations....)

crates/commonware-node/src/feed/state.rs
SuperFluffy: TBH, I don't think there is a justification spending so much effort on tracking/caching notarizations. It's interesting to look at them "as they come in", but I would just not cache them.

crates/commonware-node/src/feed/actor.rs
```diff
@@ -0,0 +1,197 @@
...
+        spawn_cell!(self.context, self.run().await)
+    }
+
+    /// Run the actor's main loop.
+    async fn run(&mut self) {
+        loop {
+            select!(
+                activity = self.receiver.next() => {
+                    let Some(activity) = activity else {
+                        debug!("feed actor shutting down");
```
SuperFluffy: Always want a root span. Also info level makes more sense IMO. This is one-time event.
```suggestion
info_span!("shutdown").in_scope(|| info!("actor shutting down"));
```

crates/commonware-node/src/feed/state.rs
```diff
@@ -0,0 +1,144 @@
...
+                latest_notarized_view: None,
+                latest_finalized: None,
+            })),
+            marshal: Arc::new(OnceLock::new()),
+            events_tx,
+        }
+    }
+
+    /// Set the marshal mailbox for historical finalization lookups. Should only be called once.
+    pub(crate) fn set_marshal(&self, marshal: marshal::Mailbox) {
```
SuperFluffy, line 52: Reason for this is that we need `handle -> reth`, but we need `reth -> commonware`, right?

Kinda roundabout. Wish we could configure the consensus engine such it spits out a handle, which we then pass to reth.

crates/commonware-node/src/feed/actor.rs
```diff
@@ -0,0 +1,197 @@
...
+                });
+
+                {
+                    let mut state = self.state.write();
+                    if state
+                        .latest_finalized
+                        .as_ref()
+                        .is_none_or(|f| f.height < block.height)
+                    {
+                        let to_remove: Vec<_> = state
```
SuperFluffy: way too much activity IMO. Let's not worry about notarizations so much.

crates/e2e/src/lib.rs
SuperFluffy, line 1: Can we add a test with a client that sits on the other side and receives a finalization cert this way?

crates/node/src/rpc/consensus/types.rs
```diff
@@ -0,0 +1,98 @@
...
+
+/// A block with a threshold BLS certificate (notarization or finalization).
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct CertifiedBlock {
+    pub epoch: u64,
+    pub view: u64,
+    pub height: u64,
+    pub digest: B256,
+    pub certificate: String,
```
SuperFluffy, line 17: Probably fine for now, but it'd be nice if this was more useful/carried more information.

crates/commonware-node/src/feed/state.rs
```diff
@@ -0,0 +1,144 @@
...
+                    return Ok(None);
+                };
+
+                let Some(finalization) = marshal.get_finalization(height).await else {
+                    return Ok(None);
+                };
+
+                let view = finalization.proposal.round.view().get();
+                let epoch = finalization.proposal.round.epoch().get();
+                let digest = B256::from_slice(finalization.proposal.payload.as_ref());
```
SuperFluffy: This unecessary. Our `proposal` is a `Digest(B256)`. Just unwrap it.

crates/node/src/rpc/consensus/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+
+/// Consensus namespace RPC trait.
+#[rpc(server, namespace = "consensus")]
+pub trait TempoConsensusApi {
+    /// Get notarization by view query.
+    ///
+    /// Use `"latest"` to get the most recent notarization, or `{"view": N}` for a specific view.
+    /// Returns an error if `{"height": N}` is used (height queries only work for finalizations).
+    #[method(name = "getNotarization")]
+    async fn get_notarization(&self, query: Query) -> RpcResult<Option<CertifiedBlock>>;
```
SuperFluffy: IMO remove.

crates/commonware-node/src/feed/state.rs
joshieDo: ok, removed it but still kept the `latest_notarized` that gets returned on `get_latest`

crates/commonware-node/src/feed/state.rs
```diff
@@ -0,0 +1,114 @@
...
+            .finish()
+    }
+}
+
+impl ConsensusFeed for FeedStateHandle {
+    async fn get_finalization(&self, query: Query) -> Option<CertifiedBlock> {
+        match query {
+            Query::Latest => self.state.read().latest_finalized.clone(),
+            Query::Height(height) => {
+                let mut marshal = self.marshal.get().cloned()?;
```
SuperFluffy: If this does not work for whatever reason, it'd be interesting to know that the marshal actor was not yet set. debug log?

crates/commonware-node/src/feed/state.rs
```diff
@@ -0,0 +1,114 @@
...
+                let finalization = marshal.get_finalization(height).await?;
```
SuperFluffy, line 116: same here - would be interesting to know if the block was not available because the mailbox was not set yet vs the actor not having it available at all.

crates/commonware-node/src/feed/actor.rs
```diff
@@ -0,0 +1,197 @@
...
+        digest: Digest,
+        certificate: &impl Encode,
+    ) -> CertifiedBlock {
+        let certificate = hex::encode(certificate.encode());
+        let height = self
+            .marshal
+            .get_block(&digest)
+            .await
+            .map(|b| b.height())
+            .unwrap_or(0);
```
SuperFluffy: IMO don't do this. If you don't know the height, then set this to `None`.

For certain nodes there is a chance that this would end up all being `0` for a very long time, which IMO is more confusing than anything.

I think this should just be `marshal.subscribe` and resolve once the node has the block.

Alternatively, don't publish the height at all?

crates/commonware-node/src/feed/actor.rs
```diff
@@ -0,0 +1,197 @@
...
+                {
+                    let mut state = self.state.write();
+                    if state
+                        .latest_finalized
+                        .as_ref()
+                        .is_none_or(|f| f.height < block.height)
+                        && state
+                            .latest_notarized
+                            .as_ref()
+                            .is_none_or(|n| n.view < view)
```
SuperFluffy, line 132: Technically you need to check for the full round information, not just the view - but `Round` implements `PartialOrd`, so you can just compare `new.round > old.round`.

crates/node/src/rpc/consensus/types.rs
```diff
@@ -0,0 +1,77 @@
...
+use futures::Future;
+use serde::{Deserialize, Serialize};
+use tokio::sync::broadcast;
+
+/// A block with a threshold BLS certificate (notarization or finalization).
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct CertifiedBlock {
+    pub epoch: u64,
+    pub view: u64,
```
SuperFluffy, line 13: Mentioned elsewhere - if we don't have the height, we shouldn't set it (if the block is not yet available through the marshal actor).

crates/commonware-node/src/feed/actor.rs
joshieDo: set height as option

PR #1870, opened 2026-01-06
fix: emit event for tip20 default admin role initialization 

howydev:


docs/specs/src/TIP20.sol
```diff
@@ -49,7 +49,8 @@ contract TIP20 is ITIP20, TIP20RolesAuth {
         string memory _symbol,
         string memory _currency,
         ITIP20 _quoteToken,
-        address admin
+        address admin,
+        address sender
```
zerosnacks, line 53: minor style note, for the spec I feel it is more intuitive to use `msg.sender` rather than an arbitrary user controlled value; feels like this is leaking implementation into spec

crates/precompiles/src/tip20/roles.rs
```diff
@@ -16,8 +16,17 @@ impl TIP20Token {
...
-    pub fn grant_default_admin(&mut self, admin: Address) -> Result<()> {
-        self.grant_role_internal(admin, DEFAULT_ADMIN_ROLE)
+    pub fn grant_default_admin(&mut self, msg_sender: Address, admin: Address) -> Result<()> {
+        self.grant_role_internal(admin, DEFAULT_ADMIN_ROLE)?;
+
+        self.emit_event(RolesAuthEvent::RoleMembershipUpdated(
+            IRolesAuth::RoleMembershipUpdated {
+                role: DEFAULT_ADMIN_ROLE,
+                account: admin,
+                sender: msg_sender,
```
0xKitsune, line 26: Do we need to emit the sender here?

crates/precompiles/src/tip20/roles.rs
```diff
@@ -16,8 +16,17 @@ impl TIP20Token {
...
+                hasRole: true,
+            },
+        ))
     }
```
0xKitsune, line 30: Can we update tests to assert event emission?

docs/specs/src/TIP20.sol
howydev: Msg.sender of this call would be the factory unfortunately

have to use this and bubble the factory’s msg.sender down

crates/precompiles/src/tip20/roles.rs
howydev: I’d leaning towards removing if this was 2 weeks ago

There might be some use - if there are 10+ admins it could be useful to see which admin was the one that did certain calls

PR #1878, opened 2026-01-06
feat!(precompiles): add public getters to `ValidatorConfig` precompile to match spec impl

0xKitsune:
Closes CHAIN-415

This PR adds public getter functions to the `ValidatorConfig` precompile to match the auto generated getters from the Solidity spec.

The following functions were added:
- `validatorCount()`
- `validatorsArray(uint256 index)` 
- `validators(address validator)` 
- `nextDkgCeremony()`

The Solidity spec declares these as public storage variables which auto-generate getters. This adds the equivalent to the Rust precompile for spec parity.


0xrusowsky:
since this is a breaking PR before the new testnet is deployed, before merging, it would be nice to include the changes from https://github.com/tempoxyz/tempo/pull/1896, as it modifies the storage layout

crates/precompiles/src/validator_config/mod.rs
```diff
@@ -75,10 +75,28 @@ impl ValidatorConfig {
...
 
     /// Get the current validator count
-    fn validator_count(&self) -> Result<u64> {
+    pub fn validator_count(&self) -> Result<u64> {
         self.validator_count.read()
     }
 
+    /// Get validator address at a specific index in the validators array
+    pub fn validators_array(&self, index: u64) -> Result<Address> {
+        self.validators_array[index as usize].read()
```
fgimenez: not sure if possible, but don't we need to check here if the index is out of bounds? like:
```suggestion
        let count = self.validator_count()?;
        if index >= count {
            return Err(TempoPrecompileError::Panic(PanicKind::ArrayOutOfBounds));
        } 
        self.validators_array[index as usize].read()
```

crates/precompiles/src/validator_config/mod.rs
```diff
@@ -260,6 +278,11 @@ impl ValidatorConfig {
         self.next_dkg_ceremony.read()
     }
 
+    /// Get the epoch at which a fresh DKG ceremony will be triggered (public getter)
+    pub fn next_dkg_ceremony(&self) -> Result<u64> {
```
fgimenez, line 274: nit, this has the exact same code as `get_next_full_dkg_ceremony` maybe we could make one call the other, to prevent future drift?

0xrusowsky: looking closer into this imo this is completely redundant as we have 2 functions doing the same (one is solidity's autogenerated public getter, and the other is a manually created getter).

should we consolidate into a single one (the auto-generated) instead, and also align the spec?

fgimenez: yep makes sense :+1:

crates/precompiles/src/validator_config/mod.rs
0xrusowsky: addressed in [`ec0926f` (#1878)](https://github.com/tempoxyz/tempo/pull/1878/commits/ec0926ff1657e96f42c6999fd40e21b5ed685dd8)

crates/precompiles/src/validator_config/mod.rs
0xrusowsky: ended up making the spec var `private` so that there is no autogen getter [`a074daa` (#1878)](https://github.com/tempoxyz/tempo/pull/1878/commits/a074daa1297eb9a9f81ae927e0f6509266d8554d)

PR #1879, opened 2026-01-06
chore(precompiles): remove `sqrt` from `FeeAMM.sol` spec, make `sqrt` private test helper, remove unused `SQRT_SCALE` constant

zerosnacks:
Remove unused `sqrt` from FeeAMM.sol specification, confusing in the context of AMMs.

Move sqrt, which is only used for arbitrary liquidity scenario setup into test section as a helper.

Note that `U256::MAX` would overflow `sqrt` due to unchecked `+1`.

Removed unused `SQRT_SCALE` constant.


zerosnacks:
RFR @legion2002 / @howydev

PR #1883, opened 2026-01-06
chore(test): add `validator_can_fast_sync_after_full_dkg` test

joshieDo:
```rust
/// Tests that a late-joining validator can sync and participate after a full DKG ceremony.
///
/// This verifies:
/// 1. A full DKG ceremony completes successfully (new polynomial,/sharing)
/// 2. A validator that joins late (after full DKG) can sync the chain
/// 3. The late validator uses fast-sync to jump epoch boundaries (including the full DKG epoch)
/// 4. The late validator continues progressing after sync
```

### DIFF
--- a/crates/e2e/src/tests/dkg/fast_sync_after_full_dkg.rs
+++ b/crates/e2e/src/tests/dkg/fast_sync_after_full_dkg.rs
@@ -29,8 +29,7 @@ fn validator_can_fast_sync_after_full_dkg() {
     let how_many_signers = 4;
     let epoch_length = 20;
     let full_dkg_epoch = 1;
-    // Epoch 1 (full DKG) runs blocks 20-39, so 65 is well past it
-    let blocks_before_late_join = 65;
+    let blocks_before_late_join = 3 * epoch_length + 1;
 
     let setup = Setup::new()
         .how_many_signers(how_many_signers)


crates/e2e/src/tests/dkg/fast_sync_after_full_dkg.rs
```diff
@@ -0,0 +1,133 @@
...
+
+/// Tests that a late-joining validator can sync and participate after a full DKG ceremony.
+///
+/// This verifies:
+/// 1. A full DKG ceremony completes successfully (new polynomial, different public key)
+/// 2. A validator that joins late (after full DKG) can sync the chain
+/// 3. The late validator uses fast-sync to jump epoch boundaries (including the full DKG epoch)
+/// 4. The late validator continues progressing after sync
+#[test_traced]
+fn validator_can_fast_sync_after_full_dkg() {
```
SuperFluffy, line 26: Maybe we can somehow merge this and the test `restart::validator_catches_up_with_a_gap_of_three_epochs`?

crates/e2e/src/tests/dkg/fast_sync_after_full_dkg.rs
```diff
@@ -0,0 +1,133 @@
...
+    let _ = tempo_eyre::install();
+
+    let how_many_signers = 4;
+    let epoch_length = 20;
+    let full_dkg_epoch = 1;
+    // Epoch 1 (full DKG) runs blocks 20-39, so 65 is well past it
+    let blocks_before_late_join = 65;
```
SuperFluffy: IMO just define this in terms of the epoch length, e.g. `3*x+1` (like in the restar tests).

crates/e2e/src/tests/dkg/fast_sync_after_full_dkg.rs
```diff
@@ -0,0 +1,132 @@
...
+        assert_ne!(
+            pubkey_before,
+            *outcome_after.sharing().public(),
+            "Full DKG must create different public key"
+        );
+
+        // wait for chain to advance
+        while validators[0]
+            .execution_provider()
+            .last_block_number()
```
SuperFluffy, line 85: It's not wrong, but IMO we should try and use the same kind of condition in all tests unless we need to explicitly talk to the execution layer.

In this case, parse the metrics for `_marshal_processed_height`.

Right now it's a bit mix and match in that we use the metric in some instances, but talk to the EL in others.

SuperFluffy (approved):

PR #1888, opened 2026-01-06
fix(subblock): [build] drop txs above gas budget to prevent DOS

0xrusowsky:
This PR drops subblock txs above the max gas budget of a subblock to prevent DOS attacks

audit: `TEMPO-16` follow-up, as the initial patch was incomplete

### DIFF
--- a/crates/commonware-node/src/subblocks.rs
+++ b/crates/commonware-node/src/subblocks.rs
@@ -746,7 +746,7 @@ async fn build_subblock(
                 }
             }
 
-            // If necessary, acquire lock and drop all outpriced txs
+            // If necessary, acquire lock and drop all invalid txs
             if !to_remove.is_empty() {
                 let mut txs = transactions.lock();
                 for hash in to_remove {


crates/commonware-node/src/subblocks.rs
```diff
@@ -708,14 +708,29 @@ async fn build_subblock(
...
+                // Remove transactions over subblock gas budget
+                if tx.gas_limit() > gas_budget {
+                    warn!(
+                        %tx_hash,
+                        tx_gas_limit = tx.gas_limit(),
+                        gas_budget,
+                        "removing transaction with gas limit exceeding maximum subblock gas budget"
+                    );
+                    transactions.lock().swap_remove(&tx_hash);
+                    continue;
```
0xKitsune: Rather than locking the transactions each time here, we could add the affected `tx_hash` to a vec and drop all affected txs at the end of the loop, which would only require locking the mutex a single time.

0xrusowsky: addressed in [`ffc68ab` (#1888)](https://github.com/tempoxyz/tempo/pull/1888/commits/ffc68abbe714b08662f276b784e5df0a239e9080)

PR #1889, opened 2026-01-07
chore(test): Testing coverage for revm handler

rakita:
Testing coverage for revm, testing inspector calls with precompile logs,  first/second create call, webauth sig, 2d nonce aa tx time window.

Some of those are alrady tested in tempo-node as e2e test. Unit testing them allow us to see llvm coverage.

### DIFF
--- a/crates/revm/src/evm.rs
+++ b/crates/revm/src/evm.rs
@@ -943,7 +943,10 @@ mod tests {
                 matches!(
                     err,
                     revm::context::result::EVMError::Transaction(
-                        TempoInvalidTransaction::ValidAfter { current: 100, valid_after: 200 }
+                        TempoInvalidTransaction::ValidAfter {
+                            current: 100,
+                            valid_after: 200
+                        }
                     )
                 ),
                 "Expected ValidAfter error, got: {err:?}"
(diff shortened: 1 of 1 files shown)


crates/revm/src/evm.rs
```diff
@@ -246,4 +592,720 @@ mod tests {
...
+
+        // Test case 1: Transaction fails when block_timestamp < valid_after
+        {
+            let mut evm = create_funded_evm_with_timestamp(caller, 100);
+            let signed_tx = create_signed_tx(Some(200), None)?;
+            let tx_env = TempoTxEnv::from_recovered_tx(&signed_tx, caller);
+
+            let result = evm.transact(tx_env);
+            assert!(result.is_err());
+            assert!(format!("{:?}", result.unwrap_err()).contains("ValidAfter"));
```
fgimenez: nit, we could check here the error type, like:
```suggestion
            assert!(
                matches!(
                    result,
                    Err(revm::context::result::EVMError::Transaction(
                        TempoInvalidTransaction::ValidAfter {
                            current: 100,
                            valid_after: 200
                        }
                    ))
                ),
                "Expected ValidAfter error, got: {result:?}"
            );
```
same for other assertions about `ValidAfter` and `ValidBefore`

0xrusowsky (approved):

PR #1890, opened 2026-01-07
chore(precompiles): match against `SolCall` enum

0xrusowsky:
This PR matches against the `SolCall` enum generated by the `sol!` macro, so that we have compile-time guarantees rather than relying on test validation.

### DIFF
--- a/crates/precompiles/src/tip20/dispatch.rs
+++ b/crates/precompiles/src/tip20/dispatch.rs
@@ -17,12 +17,14 @@ enum TIP20Call {
 
 impl TIP20Call {
     fn decode(calldata: &[u8]) -> Result<Self, alloy::sol_types::Error> {
-        match ITIP20Calls::abi_decode(calldata) {
-            Ok(call) => return Ok(Self::TIP20(call)),
-            Err(alloy::sol_types::Error::UnknownSelector { .. }) => {}
-            Err(e) => return Err(e),
+        // safe to expect as `dispatch_call` pre-validates calldata len
+        let selector: [u8; 4] = calldata[..4].try_into().expect("calldata len >= 4");
+
+        if IRolesAuthCalls::valid_selector(selector) {
+            IRolesAuthCalls::abi_decode(calldata).map(Self::RolesAuth)
+        } else {
+            ITIP20Calls::abi_decode(calldata).map(Self::TIP20)
         }
-        IRolesAuthCalls::abi_decode(calldata).map(Self::RolesAuth)
     }
 }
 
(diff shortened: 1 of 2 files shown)


crates/precompiles/src/nonce/dispatch.rs
```diff
@@ -20,11 +19,23 @@ impl Precompile for NonceManager {
             .try_into()
             .unwrap();
 
-        let result = match selector {
-            INonce::getNonceCall::SELECTOR => {
-                view::<INonce::getNonceCall>(calldata, |call| self.get_nonce(call))
-            }
-            _ => unknown_selector(selector, self.storage.gas_used()),
+        if !INonceCalls::SELECTORS.contains(&selector) {
```
DaniPopes: just match on the result, there is an error that is unknown selector https://docs.rs/alloy-sol-types/latest/alloy_sol_types/enum.Error.html

0xrusowsky: addressed in [`415916f` (#1890)](https://github.com/tempoxyz/tempo/pull/1890/commits/415916fe84aad4d9e54f5a1655c83773bb535c44)

crates/precompiles/src/nonce/dispatch.rs
```diff
@@ -1,30 +1,38 @@
...
+                    .map(|res| fill_precompile_output(res, &mut self.storage));
+            }
+            Err(_) => {
+                return Ok(fill_precompile_output(
+                    PrecompileOutput::new_reverted(0, alloy::primitives::Bytes::new()),
+                    &mut self.storage,
+                ));
             }
-            _ => unknown_selector(selector, self.storage.gas_used()),
+        };
```
DaniPopes: can maybe extract into a util since this is copy pasted in all of them?

0xrusowsky: addressed in [`7803af9` (#1890)](https://github.com/tempoxyz/tempo/pull/1890/commits/7803af956461567fd3e912dc45a4fe9694c01228)

crates/precompiles/src/nonce/dispatch.rs
```diff
@@ -1,33 +1,23 @@
...
-            INonce::getNonceCall::SELECTOR => {
-                view::<INonce::getNonceCall>(calldata, |call| self.get_nonce(call))
-            }
-            _ => unknown_selector(selector, self.storage.gas_used()),
-        };
+        if calldata.len() < 4 {
+            return Err(PrecompileError::Other(
+                "Invalid input: missing function selector".into(),
+            ));
+        }
```
onbjerg: smol nit, could this not be part of dispatch call? e.g. `dispatch_call<T>(calldata, |call| match call { .. })`

onbjerg (approved):

PR #1893, opened 2026-01-07
fix!(precompiles): fix capitalization for pathusd name

danrobinson:
This consistently updates the name and symbol for pathUSD from "PathUSD" (where it was in many comments and some tests) to "pathUSD" (the name in the current testnet, and most of the docs).

This reverts an inadvertent change in #1819, which would have caused the token to be deployed with that name and symbol.

This PR also updates the docs, comments, and tests throughout to consistently use the lowercase name.


docs/specs/test/FeeManager.t.sol
```diff
@@ -261,7 +261,7 @@ contract FeeManagerTest is BaseTest {
         }
     }
 
-    function test_defaultValidatorTokenIsPathUSD() public {
+    function test_defaultValidatorTokenIsPathUsd() public {
```
0xKitsune, line 264: nit: should this stay USD instead of Usd?

PR #1896, opened 2026-01-07
perf!(precompiles): drop useless validator_count state var

0xrusowsky:
This PR drops `validator_count`, as it is useless and redundant.

The variable 1:1 tracks the state of `validators_array`'s length slot. Thus, we can skip extra SLOAD/SSTOREs by simply reading the correct slot.

**IMPORTANT:** this PR changes the storage layout as it gets rid of one var

### DIFF
--- a/crates/precompiles/src/validator_config/mod.rs
+++ b/crates/precompiles/src/validator_config/mod.rs
@@ -103,7 +103,7 @@ impl ValidatorConfig {
     /// Get all validators (view function)
     pub fn get_validators(&self) -> Result<Vec<IValidatorConfig::Validator>> {
         let count = self.validators_array.len()?;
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/validator_config/mod.rs
```diff
@@ -106,12 +102,12 @@ impl ValidatorConfig {
 
     /// Get all validators (view function)
     pub fn get_validators(&self) -> Result<Vec<IValidatorConfig::Validator>> {
-        let count = self.validator_count()?;
+        let count = self.validators_array.len()?;
         let mut validators = Vec::new();
```
fgimenez:
```suggestion
        let mut validators = Vec::with_capacity(count);
```

PR #1899, opened 2026-01-07
fix(subblock): [validation] drop txs above gas budget

0xrusowsky:
This PR gracefully drops invalid subblocks which require a bigger gas limit than the per-validator subblock gas budget

### DIFF
--- a/crates/commonware-node/src/subblocks.rs
+++ b/crates/commonware-node/src/subblocks.rs
@@ -843,7 +843,7 @@ async fn validate_subblock(
         }
     }
 
-    // Ensure all transactions can be commited
+    // Ensure all transactions can be committed
     for tx in subblock.transactions_recovered() {
         if let Err(err) = evm.transact_commit(tx) {
             return Err(eyre::eyre!("transaction failed to execute: {err:?}"));


crates/commonware-node/src/subblocks.rs
```diff
@@ -830,6 +829,20 @@ async fn validate_subblock(
...
+        .transactions_recovered()
+        .map(|tx| tx.gas_limit() as usize)
+        .sum();
+    if total_gas > gas_budget {
+        warn!(
+            total_gas,
+            gas_budget, "subblock exceeds gas budget, skipping"
+        );
+        return Ok(());
+    }
```
fgimenez: could we define `gas_budget` as `u64`? matches `gas_limit` type and avoids potential issues on 32-bit platforms. also, instead of adding all tx we could exit as soon as the budget is exceeded, which would also give overflow protection with `saturating_add`, like:
```suggestion
    let gas_budget: u64 = evm.block().gas_limit / TEMPO_SHARED_GAS_DIVISOR / participants as u64;
    let mut total_gas: u64 = 0;
    for tx in subblock.transactions_recovered() {
        total_gas = total_gas.saturating_add(tx.gas_limit());
        if total_gas > gas_budget {
            warn!(
                total_gas,
                gas_budget, "subblock exceeds gas budget, skipping"
            );
            return Ok(());
        }
    }
```

crates/commonware-node/src/subblocks.rs
```diff
@@ -830,6 +829,20 @@ async fn validate_subblock(
         return Ok(());
     }
 
+    // Bound subblock gas at the per-validator allocation.
+    let gas_budget = (evm.block().gas_limit / TEMPO_SHARED_GAS_DIVISOR) as usize / participants;
+    let total_gas: usize = subblock
+        .transactions_recovered()
+        .map(|tx| tx.gas_limit() as usize)
+        .sum();
```
mattsse: I'd rather have this a dedicated fn like

https://github.com/tempoxyz/tempo/blob/dbc36f32de58e82763bf2650b87daed007fa4a15/crates/primitives/src/subblock.rs#L125-L128

is cleaner

mattsse (commented):
lgtm, smol suggestion

crates/commonware-node/src/subblocks.rs
0xrusowsky: addressed in [`b1b071b` (#1899)](https://github.com/tempoxyz/tempo/pull/1899/commits/b1b071b8844470b29c68d554ecb526ceee075fbe)

crates/commonware-node/src/subblocks.rs
0xrusowsky: ended up following fede's advice, as it felt better to early exit for perf reasons

crates/commonware-node/src/subblocks.rs
fgimenez: awesome thx!

crates/commonware-node/src/subblocks.rs
```diff
@@ -830,6 +829,21 @@ async fn validate_subblock(
...
+        if total_gas > gas_budget {
+            warn!(
+                total_gas,
+                gas_budget, "subblock exceeds gas budget, skipping"
+            );
+            return Ok(());
+        }
+    }
+
+    // Ensure all transactions can be commited
```
fgimenez:
```suggestion
    // Ensure all transactions can be committed
```

PR #1903, opened 2026-01-07
fix(spec): update dex reference implementation with escrow rounding up

danrobinson:
When a quote token is escrowed into the DEX when placing a limit order, the amount that is escrowed should be rounded up to favor the protocol. This was already updated in the Rust implementation (in #1583), but was not changed in the reference implementation. This updates the reference implementation to match.

### DIFF
--- a/docs/specs/foundry.toml
+++ b/docs/specs/foundry.toml
@@ -4,7 +4,6 @@ out = "out"
 libs = ["lib"]
 optimizer = true
 optimizer_runs = 200
-via_ir = true
 
 [profile.ci]
 optimizer = true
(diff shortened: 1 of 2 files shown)


docs/specs/foundry.toml
```diff
@@ -4,6 +4,7 @@ out = "out"
 libs = ["lib"]
 optimizer = true
 optimizer_runs = 200
+via_ir = true
```
fgimenez: is this change intentional, is it required for the PR changes?

danrobinson: I was hitting it (surprised me too) but you're right, let me try to find a different fix

PR #1915, opened 2026-01-07
feat(consensus/p2p): allow switching to local defaults

Zygimantass:


crates/commonware-node/src/args.rs
```diff
@@ -117,10 +117,10 @@ pub struct Args {
     #[arg(long = "consensus.bypass-ip-check", default_value_t = false)]
     pub bypass_ip_check: bool,
 
-    /// Reduces security by enabling connections from private IPs.
-    /// Only enable in trusted network environments.
-    #[arg(long = "consensus.allow-private-ips", default_value_t = false)]
-    pub allow_private_ips: bool,
+    /// Use P2P defaults optimized for local network environments.
+    /// Only enable in non-production network nodes.
+    #[arg(long = "consensus.use-local-defaults", default_value_t = false)]
```
SuperFluffy: IMO a tad more expressive:
```suggestion
    #[arg(long = "consensus.use-local-p2p-defaults", default_value_t = false)]
```

PR #1917, opened 2026-01-07
docs: TIP-1000 & TIP-1010 docs updates

dankrad:
Adding the changes to the documentation:
 * State creation gas cost
 * Basefee
 * Payment lane and main gas limits

As suggested by @danrobinson converted form these: 
 * https://github.com/tempoxyz/docs/pull/402
 * https://github.com/tempoxyz/docs/pull/403

### DIFF
--- a/docs/pages/protocol/blockspace/overview.mdx
+++ b/docs/pages/protocol/blockspace/overview.mdx
@@ -29,6 +29,14 @@ pub struct Header {
 - `general_gas_limit` and `shared_gas_limit` carve up the canonical `gas_limit` for payment and subblock gas (see [payment lane specification](/protocol/blockspace/payment-lane-specification) and [subblock specification](/protocol/blockspace/subblock-specification)).
 - `timestamp_millis_part` stores the sub‑second component; the full timestamp is `inner.timestamp * 1000 + timestamp_millis_part` .
 
+### Mainnet Launch Parameters
+
+At mainnet launch, the following parameters are configured:
+
+- **Base fee**: `2 × 10^10` wei (targets 0.1 cent per 50,000 gas TIP-20 transfer)
+- **Payment lane gas limit**: 500,000,000 gas per block (supports ~20,000 TPS for payments)
+- **Main transaction gas limit** (`general_gas_limit`): 25,000,000 gas per block
+
 ### Block body
 The block body in Tempo retains the canonical Ethereum block body structure, with the addition of system transactions. Transactions are ordered in the following sections:
 
--- a/docs/pages/protocol/blockspace/overview.mdx
+++ b/docs/pages/protocol/blockspace/overview.mdx
@@ -38,7 +38,7 @@ At mainnet launch, the following parameters are configured:
 - **Main transaction gas limit** (`general_gas_limit`): 30,000,000 gas per block
 - **Transaction gas cap**: 30,000,000 gas per transaction
 
-See [TIP-1011](/protocol/tips/tip-1011) for the full specification and rationale.
+See [TIP-1010](/protocol/tips/tip-1010) for the full specification and rationale.
 
 ### Block body
 The block body in Tempo retains the canonical Ethereum block body structure, with the addition of system transactions. Transactions are ordered in the following sections:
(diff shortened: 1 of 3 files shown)


docs/pages/protocol/tips/tip-1000.mdx
```diff
@@ -0,0 +1,316 @@
...
+- This pricing applies to the contract code size (the bytecode being deployed)
+
+**Implementation Details:**
+- The code storage cost is calculated as: `code_size × 1,000`
+- Additional state creation costs: 2 × 250,000 gas = 500,000 gas for keccak hash and codesize fields
+- Total contract creation cost: `(code_size × 1,000) + 500,000` gas
+- This replaces the existing EVM per-byte cost for contract creation (not an additional charge)
+- Applies to both CREATE and CREATE2 operations
+- The account creation cost (250,000 gas) is separate and still applies when the contract account's nonce transitions from 0 to 1
+
```
rakita, line 93: Would be good to add new section for initial gas tx increase.

Something like

## Intrinsic tx gas 

Checks that can make a transaction invalid if it can't be covered by caller's balance. Those checks are done on txpool as ddos prevention and validation/block building.

The transaction gas cap is changed from 16M to 32M to accommodate the deployment of 24kb contract creation.

* Transaction with `nonce == 0` would require an additional 250,000 gas, as it created a new account.
* Tempo AA Transactions with `nonce_key == 0` would require additional 250,000gas to create a nonce key for 2d nonces. 
* Authorisation List with `auth_list.nonce == 0` would require an additional 250,000 gas.
  * Current base cost should be discounted from 25,000 gas to 12,500 gas (12,500 covers state read and signature recovery).
  * Refund added when auth is not creating a new account should be removed, as this is covered by an additional 250,000 gas.
  * Gas accounting is going to diverge from EIP-7702 as tempo can know if an account is going to be created by looking at the nonce (as the tempo account balance is always zero).
* Create tx kind requires an additional 250,000 gas.
* First call in batches, that is an Create kind requires an additional 250,000 gas.

Tempo AA Transaction key authorisation can't determine whether it is going to create new storage or not. Only way to enforce new gas is to check them dynamically in runtime. If there is enough gas to cover creation, the revert of the authorisation key changes should happen and transaction should be marke as revert (Not invalid). `authorizeKeyCall` has 4 storage fields, and a list of limits per token.

docs/pages/protocol/tips/tip-1000.mdx
```diff
@@ -0,0 +1,316 @@
...
+- **Related Specs/TIPs**: N/A
+- **Protocol Version**: TBD
+
+---
+
+# Overview
+
+## Abstract
+
+This TIP increases the gas cost for creating new state elements, accounts, and contract code to provide economic protection against state growth spam attacks. The proposal increases the cost of writing a new state element from 20,000 gas to 250,000 gas, introduces a 250,000 gas charge for account creation (when the account's nonce is first written), and implements a new contract creation cost model: 1,000 gas per byte of contract code plus 500,000 gas for keccak hash and codesize fields.
```
rakita, line 16: Codesize requires an account format change on how to store it in dn, this increases complexity of spec that I didn't account for.

Would propose reducing contract creation to 250k gas and removing codesize for now

docs/pages/protocol/tips/tip-1000.mdx
dankrad: Don't understand these: 

  * Current base cost should be discounted from 25,000 gas to 12,500 gas (12,500 covers state read and signature recovery).
  * Refund added when auth is not creating a new account should be removed, as this is covered by an additional 250,000 gas.
  * Gas accounting is going to diverge from EIP-7702 as tempo can know if an account is going to be created by looking at the nonce (as the tempo account balance is always zero).

dankrad: Also this:

* Create tx kind requires an additional 250,000 gas.
* First call in batches, that is an Create kind requires an additional 250,000 gas.

jenpaff:
superseded by https://github.com/tempoxyz/tempo/pull/2152

PR #1918, opened 2026-01-07
feat: add `consensus_getIdentityTransitionProof`

joshieDo:
Assumes the client has access to the rest of the headers, so we only provide the headers/finalizations of the relevant boundaries.

```rust
    /// Get identity transition proofs (full DKG events).
    ///
    /// Each proof contains the block header with the new DKG outcome, and a BLS certificate from the OLD
    /// network identity that signs the block.
    ///
    /// - `from_epoch`: Optional epoch to start searching from (defaults to latest finalized)
    /// - `full = false` (default): Returns only the most recent transition
    /// - `full = true`: Returns all transitions from the starting epoch back to genesis
    #[method(name = "getIdentityTransitionProof")]
    async fn get_identity_transition_proof(
        &self,
        from_epoch: Option<u64>,
        full: Option<bool>,
    ) -> RpcResult<IdentityTransitionResponse>;
```

### DIFF
--- a/crates/e2e/src/execution_runtime.rs
+++ b/crates/e2e/src/execution_runtime.rs
@@ -19,7 +19,7 @@ use commonware_codec::Encode;
 use commonware_cryptography::ed25519::PublicKey;
 use commonware_utils::ordered;
 use eyre::{OptionExt as _, WrapErr as _};
-use futures::{future::BoxFuture, StreamExt};
+use futures::{StreamExt, future::BoxFuture};
 use reth_db::mdbx::DatabaseEnv;
 use reth_ethereum::{
     evm::{


crates/node/src/rpc/consensus/types.rs
```diff
@@ -64,6 +65,63 @@ pub struct ConsensusState {
...
+/// This proves that the network transitioned from `old_public_key` to
+/// `new_public_key` at the given epoch, with a certificate signed by
+/// the old network identity.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct IdentityTransition {
+    /// Epoch where the full DKG ceremony occurred.
+    pub transition_epoch: u64,
+    /// Hex-encoded BLS public key before the transition.
+    pub old_public_key: String,
```
SuperFluffy:
```suggestion
    pub old_identity: String,
```

crates/node/src/rpc/consensus/types.rs
```diff
@@ -64,6 +65,63 @@ pub struct ConsensusState {
...
+    /// Hex-encoded BLS public key after the transition.
+    pub new_public_key: String,
```
SuperFluffy:
```suggestion
    pub new_identity: String,
```

crates/node/src/rpc/consensus/types.rs
```diff
@@ -64,6 +65,63 @@ pub struct ConsensusState {
...
+}
+
+/// Cryptographic proof data for an identity transition.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct TransitionProofData {
+    /// The block header containing the new DKG outcome in extra_data.
+    pub header: TempoHeaderResponse,
+    /// Hex-encoded finalization.
+    pub finalization: String,
```
SuperFluffy: This is supposed to hold the cert, right?
```suggestion
    pub finalization_certificate: String,
```

joshieDo: https://github.com/tempoxyz/tempo/pull/2079

crates/commonware-node/src/feed/state.rs
```diff
@@ -148,4 +175,101 @@ impl ConsensusFeed for FeedStateHandle {
...
+            get_outcome(&mut marshal, &epocher, start_epoch.saturating_sub(1)).await?;
+        let mut curr_pubkey = *identity_outcome.sharing().public();
+        let identity = hex::encode(curr_pubkey.encode());
+
+        let mut transitions = Vec::new();
+
+        // Walk backwards comparing public keys to detect full DKG transitions.
+        let mut search_epoch = start_epoch.saturating_sub(1);
+        while search_epoch > 0 {
+            let prev_outcome = get_outcome(&mut marshal, &epocher, search_epoch - 1).await?;
```
SuperFluffy: IMO don't just stop. Try to go as far as you can. If you can't continue, return what you found. It should be obvious from the chain if you reached genesis or not.

Also - you can always toss in the genesis keye identity into the path (note that you will never have a cert for it.

crates/commonware-node/src/feed/state.rs
```diff
@@ -148,4 +175,101 @@ impl ConsensusFeed for FeedStateHandle {
...
+            // If keys differ, there was a full DKG at search_epoch
+            if curr_pubkey != prev_pubkey {
+                // Fetch the block and certificate that committed the new identity
+                let proof_height = epocher
+                    .last(Epoch::new(search_epoch))
+                    .expect("fixed epocher is valid for all epochs");
+                let proof_block = marshal
+                    .get_block(proof_height)
+                    .await
+                    .ok_or(IdentityProofError::PrunedData(proof_height.get()))?;
```
SuperFluffy: Same - return what you got. Don't just die.

crates/commonware-node/src/feed/state.rs
```diff
@@ -148,4 +175,101 @@ impl ConsensusFeed for FeedStateHandle {
     async fn subscribe(&self) -> Option<broadcast::Receiver<Event>> {
         Some(self.events_tx.subscribe())
     }
+
+    async fn get_identity_transition_proof(
+        &self,
+        from_epoch: Option<u64>,
```
SuperFluffy, line 245: We can cache the result of this epoch so we don't walk it again.

That also means that if we are asked for a newer proof, we can extend that the latest proof.

And if somebody asks for an older proof, we can just take a subsection of the current proof. That's pretty cool. ^.^

SuperFluffy (changes requested):
This is cool. Smol request to not just die if you it can't find the full chain. Just return what's already collected instead.

gakonst:
@SuperFluffy These review comments appear to be stale - all feedback has been addressed in the recent commits:

1. **"Don't just die"** → ✅ Implemented in commit `3c79a88` - errors now `break` and return partial results instead of using `?`
2. **Genesis identity marker** → ✅ Implemented in same commit - epoch 0 marker with `proof: None` is added when reaching genesis  
3. **Caching** → ✅ Implemented in commit `94aef2f` - `IdentityTransitionCache` stores and reuses the transition chain

Could you re-review? The line references in the comments (e.g. `None`, `240`) don't match the current code due to rebasing.

crates/e2e/src/tests/consensus_rpc.rs
```diff
@@ -131,3 +142,163 @@ async fn wait_for_height(context: &Context, target_height: u64) {
...
     }
 }
+
+/// Test that `get_identity_transition_proof` returns valid proofs after a full DKG ceremony.
+///
+/// This verifies:
+/// 1. After a full DKG, the RPC returns a transition with different old/new public keys
+/// 2. The transition epoch matches where the full DKG occurred
+/// 3. The proof contains a valid header and certificate
+#[tokio::test]
```
SuperFluffy: remove

crates/commonware-node/src/feed/state.rs
```diff
@@ -148,4 +234,204 @@ impl ConsensusFeed for FeedStateHandle {
...
+                    );
+                    break;
+                };
+
+                transitions.push(IdentityTransition {
+                    transition_epoch: search_epoch,
+                    old_identity: hex::encode(prev_pubkey.encode()),
+                    new_identity: hex::encode(curr_pubkey.encode()),
+                    proof: Some(TransitionProofData {
+                        header: TempoHeaderResponse::from_consensus_header(
```
SuperFluffy, line 350: Seems extremely chonky? Should we really dump the full header here?

crates/commonware-node/src/feed/state.rs
```diff
@@ -148,4 +234,204 @@ impl ConsensusFeed for FeedStateHandle {
...
+                        height = proof_height.get(),
+                        start_epoch,
+                        search_epoch,
+                        "stopping identity transition walk early (proof block pruned)"
+                    );
+                    break;
+                };
+
+                let Some(finalization) = marshal.get_finalization(proof_height).await else {
+                    tracing::debug!(
```
SuperFluffy: This happens very rarely - info level?

crates/commonware-node/src/feed/state.rs
```diff
@@ -148,4 +234,204 @@ impl ConsensusFeed for FeedStateHandle {
...
+
+            // If keys differ, there was a full DKG at search_epoch
+            if curr_pubkey != prev_pubkey {
+                // Fetch the block and certificate that committed the new identity
+                let proof_height = epocher
+                    .last(Epoch::new(search_epoch))
+                    .expect("fixed epocher is valid for all epochs");
+
+                let Some(proof_block) = marshal.get_block(proof_height).await else {
+                    tracing::debug!(
```
SuperFluffy: rare - info level?

crates/commonware-node/src/feed/state.rs
```diff
@@ -148,4 +234,204 @@ impl ConsensusFeed for FeedStateHandle {
...
+                transitions.extend(cache.transitions.clone());
+                search_epoch = cache.to_epoch;
+                reached_genesis = cache.to_epoch == 0;
+                break;
+            }
+
+            let prev_outcome = match get_outcome(&mut marshal, &epocher, search_epoch - 1).await {
+                Ok(o) => o,
+                Err(err) => {
+                    tracing::debug!(
```
SuperFluffy: info

crates/e2e/src/tests/consensus_rpc.rs
SuperFluffy: The htt/websocket fetch can be run from inside the execution runtime.

PR #1925, opened 2026-01-08
[DONTMERGE] feat!(consensus): encrypt private share data at rest

SuperFluffy:
With this patch the node requires an encryption key provided in the env var `TEMPO_DKG_ENCRYPTION_KEY`, which is used to encrypt all DKG data at rest.

The main intention was to encrypt dealer shares and the group share acquired as the result of the DKG,  but in this implementation all data gets encrypted. It turned out to be nicer to provide a wrapper `Encrypted<T>` which is written to state, as opposed to having two types `StateInMemory` and `StateAtRest` which mostly share the same fields but where some fields are decrypted/encrypted.

This patch also provides two utilities on the tempo binary, one to generate a new encryption key, and one to encrypt existing data using that key:

```
Generates an encryption/decryption key

Usage: tempo consensus generate-encryption-key --output <FILE>

Options:
  -o, --output <FILE>  Destination of the generated signing key
  -h, --help           Print help

Encrypts an input file with the DKG encryption and writes it to a file in hex format..

---

Usage: tempo consensus encrypt-with-dkg-encryption-key --output <FILE> <INPUT>

Arguments:
  <INPUT>

Options:
  -o, --output <FILE>
  -h, --help           Print help
```


joshieDo (approved):

crates/commonware-node-config/src/lib.rs
```diff
@@ -5,16 +5,123 @@
...
     ed25519::{PrivateKey, PublicKey},
 };
+use crypto_common::{generic_array::typenum::Unsigned as _, rand_core::CryptoRngCore};
 
 #[cfg(test)]
 mod tests;
 
+const SIGNING_SHARE_KEY_ENV: &str = "TEMPO_SIGNING_SHARE_KEY";
+
+pub fn sining_share_key_from_env() -> Result<EncryptionKey, EncryptionKeyError> {
```
kuyziss: lets fix the typo?

kuyziss (approved):

crates/commonware-node/src/dkg/manager/read_write_transaction.rs
SuperFluffy, line 1: forot to delete

crates/commonware-node/src/dkg/manager/actor/state/mod.rs
```diff
@@ -42,70 +53,42 @@ const READ_BUFFER: NonZeroUsize = NZUsize!(1 << 20);
...
 where
-    TContext: commonware_runtime::Storage + Metrics,
+    TContext: commonware_runtime::Storage + Metrics + Clock,
 {
-    states: contiguous::variable::Journal<TContext, State>,
-    events: segmented::variable::Journal<TContext, Event>,
+    // Used as an RNG source.
+    context: TContext,
+
+    state: metadata::Metadata<TContext, U64, Encrypted<State>>,
```
SuperFluffy, line 70: We did not really need a journal for this since we only ever consider the latest state.

So this metadata is only ever written to once (upon completing a DKG round), which means we won't ever run into atomicity issues.

bin/tempo/src/tempo_cmd.rs
```diff
@@ -25,14 +26,65 @@ struct ConsensusCommand {
...
 }
 
+#[expect(
+    clippy::enum_variant_names,
+    reason = "these map to descriptive cli subcommands"
+)]
 #[derive(Debug, Subcommand)]
 enum ConsensusSubcommand {
+    /// Encrypts an input file with the DKG encryption and writes it
+    /// to a file in hex format..
```
joshieDo:
```suggestion
    /// to a file in hex format.
```

joshieDo (approved):

PR #1927, opened 2026-01-08
refactor(evm): use test-only helpers instead of exposing executor internals

fgimenez:
Refactors EVM tests to avoid exposing `TempoBlockExecutor` internal fields as `pub(crate)`.

- Add `#[cfg(test)]` impl block with helper methods for setting executor state
- Extend `TestExecutorBuilder` with `with_section`, `with_seen_subblock`, `with_incentive_gas_used`
- Revert executor fields to private visibility
- Update tests to use builder pattern


0xrusowsky (approved):
that way cleaner, nice!

PR #1929, opened 2026-01-08
moderato docs

jenpaff:
update docs to Moderato network


docs/vocs.config.tsx
```diff
@@ -468,8 +468,741 @@ export default defineConfig({
...
+                          {
+                            text: 'useWatchTransfer',
+                            link: '/sdk/typescript/wagmi/hooks/token.useWatchTransfer',
+                          },
+                        ],
+                      },
+                    ],
+                  },
+                ],
               },
```
struong, line 473: delete 

all of these are gone after wagmi upstreamed tempo! they live at https://wagmi.sh/tempo/getting-started now

struong (changes requested):
other than deleting the pointers to non-existent tempo.ts/wagmi pages lgtm!

PR #1930, opened 2026-01-08
test(transaction-pool): improve test quality and coverage

fgimenez:
Closes CHAIN-421

Improves test quality and coverage for `tempo-transaction-pool`.

## Changes

- `tt_2d_pool.rs`: tests for edge cases (nonce overflow, empty pool), error paths, boundary conditions, state transitions, and 2D nonce scenarios
- `validator.rs`: tests for value validation, fee token rejection, validity windows
- `transaction.rs`: tests for payment classification, fee costs, nonce key slots
- `amm.rs`**: tests for liquidity cache initialization and edge cases
- `best.rs`**: tests for `MergeBestTransactions` trait methods
- `test_utils.rs`**: consolidated shared test helpers

## Test Quality Focus

- All pool tests call `assert_invariants()` after state-changing operations
- Error path tests verify specific error types, not just that an error occurred
- Edge case tests cover boundary conditions (u64::MAX, empty collections)
- Tests are isolated and don't depend on external state

### DIFF
--- a/crates/transaction-pool/src/amm.rs
+++ b/crates/transaction-pool/src/amm.rs
@@ -436,27 +436,6 @@ mod tests {
     }
 
     // ============================================
-    // Cache clone test
-    // ============================================
-
-    #[test]
-    fn test_amm_liquidity_cache_clone() {
-        let cache = AmmLiquidityCache {
-            inner: Arc::new(RwLock::new(AmmLiquidityCacheInner {
-                unique_tokens: vec![address!("1111111111111111111111111111111111111111")],
-                ..Default::default()
-            })),
-        };
-
-        let cloned = cache.clone();
-
-        assert_eq!(
-            cache.inner.read().unique_tokens.len(),
-            cloned.inner.read().unique_tokens.len()
-        );
-    }
-
-    // ============================================
     // AmmLiquidityCacheInner direct manipulation tests
     // ============================================
 


crates/transaction-pool/src/best.rs
0xrusowsky, line 1: is this really needed?

i'm not sure how useful testing the functionality of a mock struct is, although i guess it doesn't hurt either

crates/transaction-pool/src/test_utils.rs
0xrusowsky, line 1: nit: rather than having 4 dedicated AA-tx fns, i think it would be cleaner to use the builder pattern

we could even have a single builder pattern for all the tx-related fns in the file which outputs `TempoPooledTransaction`s and which can be built via eip1559 or AA.

crates/transaction-pool/src/transaction.rs
```diff
@@ -420,17 +432,447 @@ mod tests {
...
+    // InMemorySize tests
+    // ============================================
+
+    #[test]
+    fn test_in_memory_size() {
+        let tx = create_aa_tx(Address::random(), U256::ZERO, 0);
+
+        // Size should be positive
+        assert!(tx.size() > 0);
+    }
```
0xrusowsky: imho tests like this one don't add much value, as the `size` fn is just for heuristics and the test check doesn't assert much

crates/transaction-pool/src/tt_2d_pool.rs
```diff
@@ -2333,4 +2264,1005 @@ mod tests {
...
+
+    // ============================================
+    // AA2dPoolConfig tests
+    // ============================================
+
+    #[test]
+    fn test_aa2d_pool_config_default() {
+        let _config = AA2dPoolConfig::default();
+        // Verifies that default config can be constructed without panicking
+    }
```
0xrusowsky: similar to the previous cmnt, imo we don't need 100% coverage for the sake of it

crates/transaction-pool/src/amm.rs
```diff
@@ -223,3 +223,316 @@ struct AmmLiquidityCacheInner {
...
+    #[test]
+    fn test_compute_amount_out_basic() {
+        use tempo_precompiles::tip_fee_manager::amm::compute_amount_out;
+
+        let fee = U256::from(1000);
+        let result = compute_amount_out(fee);
+        assert!(result.is_ok());
+        let amount = result.unwrap();
+        assert!(amount > U256::ZERO);
+    }
```
0xrusowsky: same here, we are only checking that `compute_amount_out` doesn't err and returns != 0, which doesn't add much value and may give us "false confidence".

personally, i'd either test negative cases or drop the test

crates/transaction-pool/src/amm.rs
```diff
@@ -223,3 +223,316 @@ struct AmmLiquidityCacheInner {
...
+            })),
+        };
+
+        let cloned = cache.clone();
+
+        assert_eq!(
+            cache.inner.read().unique_tokens.len(),
+            cloned.inner.read().unique_tokens.len()
+        );
+    }
```
0xrusowsky: not sure we need to clone derived traits like clone

crates/transaction-pool/src/best.rs
fgimenez: yep, there are 2 tests that are actually testing the mock struct, already removed. but the rest are exercising `MergeBestTransactions`, i believe those are providing real value and covering untested paths, wdyt?

crates/transaction-pool/src/test_utils.rs
fgimenez: good call, done

crates/transaction-pool/src/transaction.rs
fgimenez: agree, removed also `test_encodable_2718`

crates/transaction-pool/src/tt_2d_pool.rs
fgimenez: agree removed also `test_best_transactions_no_updates` and `test_best_transactions_set_skip_blobs`

crates/transaction-pool/src/amm.rs
fgimenez: yep dropped

crates/transaction-pool/src/amm.rs
fgimenez: removed

crates/transaction-pool/src/test_utils.rs
```diff
@@ -0,0 +1,199 @@
...
+/// let tx = TxBuilder::aa(sender)
+///     .nonce_key(U256::from(1))
+///     .nonce(5)
+///     .build();
+///
+/// // Create an EIP-1559 transaction
+/// let tx = TxBuilder::eip1559(to_address).build();
+/// ```
+#[derive(Debug, Clone)]
+pub(crate) struct TxBuilder {
```
0xrusowsky, line 44: nice!

crates/transaction-pool/src/amm.rs
```diff
@@ -223,3 +223,280 @@ struct AmmLiquidityCacheInner {
...
+    use super::*;
+    use crate::test_utils::create_mock_provider;
+    use alloy_primitives::address;
+
+    // ============================================
+    // AmmLiquidityCacheInner tests
+    // ============================================
+
+    #[test]
+    fn test_amm_liquidity_cache_inner_default() {
```
0xKitsune, line 238: nit: do we need this test?

crates/transaction-pool/src/best.rs
```diff
@@ -271,4 +271,93 @@ mod tests {
...
         assert_eq!(merged.next(), None);
     }
+
+    // ============================================
+    // BestTransactions trait method tests
+    // ============================================
+
+    #[test]
+    fn test_merge_best_transactions_mark_invalid() {
+        use reth_primitives_traits::transaction::error::InvalidTransactionError;
```
0xKitsune, line 281: nit: consider moving imports to the top of the tests mod.

crates/transaction-pool/src/best.rs
```diff
@@ -271,4 +271,93 @@ mod tests {
         assert_eq!(merged.next(), Some("tx_d"));
         assert_eq!(merged.next(), None);
     }
+
+    // ============================================
+    // BestTransactions trait method tests
+    // ============================================
```
0xKitsune, line 277: Are we actually asserting anything in these tests or just checking that this doesnt panic? If the latter, I dont think these tests add much value, could we either update to assert specific state changes or remove? Wdyt?

crates/transaction-pool/src/metrics.rs
```diff
@@ -73,3 +73,79 @@ impl AA2dPoolMetrics {
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_aa2d_pool_metrics_default() {
+        // Ensure we can create the metrics struct (macro-generated)
+        let _metrics = AA2dPoolMetrics::default();
+    }
```
0xKitsune, line 85: Lets remove this, I dont think this test is necessary.

crates/transaction-pool/src/metrics.rs
```diff
@@ -73,3 +73,79 @@ impl AA2dPoolMetrics {
...
+
+    #[test]
+    fn test_set_transaction_counts() {
+        let metrics = AA2dPoolMetrics::default();
+
+        // Should not panic
+        metrics.set_transaction_counts(100, 50, 50);
+        metrics.set_transaction_counts(0, 0, 0);
+        metrics.set_transaction_counts(usize::MAX, usize::MAX / 2, usize::MAX / 2);
```
0xKitsune, line 94: Rather than checking if this shouldnt panic can we check that the expected values are updated correclty?

crates/transaction-pool/src/metrics.rs
```diff
@@ -73,3 +73,79 @@ impl AA2dPoolMetrics {
...
+
+    #[test]
+    fn test_inc_nonce_key_count() {
+        let metrics = AA2dPoolMetrics::default();
+
+        // Should not panic
+        metrics.inc_nonce_key_count(0);
+        metrics.inc_nonce_key_count(1);
+        metrics.inc_nonce_key_count(100);
+    }
```
0xKitsune, line 105: Same thing here, although not sure we need this test?

crates/transaction-pool/src/metrics.rs
```diff
@@ -73,3 +73,79 @@ impl AA2dPoolMetrics {
         self.demoted_transactions.increment(count as u64);
     }
 }
+
+#[cfg(test)]
+mod tests {
```
0xKitsune, line 78: Not sure we need these at all as this is module is just incrementing metrics. Many of these tests are just checking that there are no panics, I think we can remove these to reduce bloat as these tests do not provide much value IMO.

crates/transaction-pool/src/validator.rs
```diff
@@ -1004,4 +1004,154 @@ mod tests {
...
         }
     }
+
+    // ============================================
+    // Non-zero value rejection tests
+    // ============================================
+
+    #[tokio::test]
+    async fn test_non_zero_value_in_eip1559_rejected() {
+        let transaction = get_transaction(Some(U256::from(1)));
```
0xKitsune, line 1014: Are we able to use the `TxBuilder` here instead?

crates/transaction-pool/src/validator.rs
```diff
@@ -1004,4 +1004,154 @@ mod tests {
...
+
+    #[tokio::test]
+    async fn test_invalid_fee_token_rejected() {
+        use alloy_primitives::{Signature, TxKind, address};
+        use tempo_primitives::transaction::{
+            TempoTransaction,
+            tempo_transaction::Call,
+            tt_signature::{PrimitiveSignature, TempoSignature},
+            tt_signed::AASigned,
+        };
```
0xKitsune, line 1073: nit: can we move imports to the top of the tests mod?

crates/transaction-pool/src/validator.rs
```diff
@@ -1004,4 +1004,154 @@ mod tests {
...
+            key_authorization: None,
+        };
+
+        let signed_tx = AASigned::new_unhashed(
+            tx_aa,
+            TempoSignature::Primitive(PrimitiveSignature::Secp256k1(Signature::test_signature())),
+        );
+        let envelope: TempoTxEnvelope = signed_tx.into();
+        let recovered = envelope.try_into_recovered().unwrap();
+        let transaction = TempoPooledTransaction::new(recovered);
```
0xKitsune, line 1104: nit: is there a more succinct way to write this?

crates/transaction-pool/src/validator.rs
```diff
@@ -1004,4 +1004,154 @@ mod tests {
...
+            .validate_transaction(TransactionOrigin::External, transaction)
+            .await;
+
+        if let TransactionValidationOutcome::Invalid(_, err) = &outcome {
+            let error_msg = err.to_string();
+            assert!(
+                !error_msg.contains("Native transfers are not supported"),
+                "Should not fail with NonZeroValue error, got: {error_msg}"
+            );
+        }
```
0xKitsune, line 1058: Can we update this test to assert that the valdiation succeeds, rather than allowing other errors?

crates/transaction-pool/src/validator.rs
```diff
@@ -1004,4 +1004,154 @@ mod tests {
...
+            .as_secs();
+        let validator = setup_validator(&transaction, current_time);
+
+        let outcome = validator
+            .validate_transaction(TransactionOrigin::External, transaction)
+            .await;
+
+        match &outcome {
+            TransactionValidationOutcome::Invalid(_, err) => {
+                let error_msg = err.to_string();
```
0xKitsune, line 1028: Can we assert exact errors rather than `error_msg.contains()`?

crates/transaction-pool/src/validator.rs
```diff
@@ -1004,4 +1004,154 @@ mod tests {
...
+                assert!(
+                    error_msg.contains("fee token") || error_msg.contains("InvalidFeeToken"),
+                    "Expected InvalidFeeToken error, got: {error_msg}"
+                );
+            }
+            other => panic!("Expected Invalid outcome for invalid fee token, got: {other:?}"),
+        }
```
0xKitsune, line 1125: We can use `assert!(matches!())` instead. We can also use this in other tests asserting errors.

crates/transaction-pool/src/metrics.rs
fgimenez: sure, will address in a followup

crates/transaction-pool/src/transaction.rs
```diff
@@ -420,17 +432,444 @@ mod tests {
...
+        let non_payment_addr = address!("1234567890123456789012345678901234567890");
+        let pooled_tx = TxBuilder::eip1559(non_payment_addr)
+            .gas_limit(21000)
+            .build_eip1559();
+        assert!(!pooled_tx.is_payment());
+    }
+
+    // ============================================
+    // fee_token_cost tests
+    // ============================================
```
0xKitsune, line 460: nit: This is a sylistic preference but IMO we should remove these where added throughout this PR.

crates/transaction-pool/src/transaction.rs
```diff
@@ -420,17 +432,444 @@ mod tests {
...
 
-        // Create via new() and verify caching
         let pooled_tx = TempoPooledTransaction::new(recovered);
         assert!(pooled_tx.is_payment());
     }
+
+    #[test]
+    fn test_payment_classification_negative() {
+        // Test that non-TIP20 address is NOT classified as payment
+        let non_payment_addr = address!("1234567890123456789012345678901234567890");
```
0xKitsune, line 451: nit: we can use `Address::random()`.

crates/transaction-pool/src/transaction.rs
```diff
@@ -420,17 +432,444 @@ mod tests {
...
+        let tx = TxBuilder::aa(sender)
+            .gas_limit(100_000)
+            .value(value)
+            .build();
+
+        // fee_token_cost should be cost - value
+        let fee_cost = tx.fee_token_cost();
+
+        // The cost includes gas spending + value, so fee_token_cost = cost - value = gas spending
+        assert!(fee_cost > U256::ZERO, "Fee token cost should be positive");
```
0xKitsune, line 475: Can we assert the exact value rather than using `>` here?

crates/transaction-pool/src/transaction.rs
```diff
@@ -420,17 +432,444 @@ mod tests {
...
+        assert_eq!(slot1, slot2);
+    }
+
+    #[test]
+    fn test_nonce_key_slot_returns_none_for_non_aa() {
+        let tx = TxBuilder::eip1559(Address::random())
+            .gas_limit(21000)
+            .build_eip1559();
+        assert!(tx.nonce_key_slot().is_none());
+    }
```
0xKitsune, line 594: Isnt this already covered in `test_non_aa_transaction_helpers`?

crates/transaction-pool/src/transaction.rs
```diff
@@ -420,17 +432,444 @@ mod tests {
...
+    #[test]
+    fn test_into_with_tx_env_without_prepare() {
+        let sender = Address::random();
+        let tx = TxBuilder::aa(sender).build();
+        let hash = *tx.hash();
+
+        // Don't call prepare_tx_env, should still work (compute on demand)
+        let with_env = tx.into_with_tx_env();
+        assert_eq!(with_env.tx.tx_hash(), &hash);
+    }
```
0xKitsune, line 625: I might be overlooking something what are this testing? Do we need this?

crates/transaction-pool/src/transaction.rs
```diff
@@ -420,17 +432,444 @@ mod tests {
...
+                !err.is_bad_transaction(),
+                "Expected is_bad_transaction() == false for: {err}"
+            );
+        }
+    }
+
+    #[test]
+    fn test_error_is_bad_transaction_true_cases() {
+        // These errors ARE "bad transactions" (should not be retried)
+        let true_cases = [
```
0xKitsune, line 665: Can we combine these into a single `test_is_bad_transaction()`?

crates/transaction-pool/src/transaction.rs
```diff
@@ -420,17 +432,444 @@ mod tests {
...
+    fn test_requires_nonce_check_aa_with_zero_nonce_key() {
+        let tx = TxBuilder::aa(Address::random()).build();
+        assert!(
+            tx.requires_nonce_check(),
+            "AA with nonce_key=0 should require nonce check"
+        );
+    }
+
+    #[test]
+    fn test_requires_nonce_check_aa_with_nonzero_nonce_key() {
```
0xKitsune, line 707: Can we consolidate these tests?

crates/transaction-pool/src/transaction.rs
```diff
@@ -420,17 +432,444 @@ mod tests {
...
+        let result = tx.validate_blob(&sidecar, settings);
+
+        assert!(result.is_err());
+        match result {
+            Err(BlobTransactionValidationError::NotBlobTransaction(ty)) => {
+                // EIP-1559 type
+                assert_eq!(ty, tx.ty());
+            }
+            other => panic!("Expected NotBlobTransaction error, got: {other:?}"),
+        }
```
0xKitsune, line 743: nit: we can use `assert!(matches!())` here.

crates/transaction-pool/src/transaction.rs
```diff
@@ -420,17 +432,444 @@ mod tests {
...
+
+    #[test]
+    fn test_inner_returns_correct_reference() {
+        let sender = Address::random();
+        let tx = TxBuilder::aa(sender).nonce(5).build();
+
+        let inner = tx.inner();
+        assert_eq!(inner.signer(), sender);
+        assert_eq!(inner.nonce(), 5);
+    }
```
0xKitsune, line 771: Inner is `EthPooledTransaction` which is already sufficiently covered in Reth. I dont think we need this test.

crates/transaction-pool/src/tt_2d_pool.rs
```diff
@@ -2446,4 +2426,1002 @@ mod tests {
...
+        pool.add_transaction(Arc::new(wrap_valid_tx(tx, TransactionOrigin::Local)), 0)
+            .unwrap();
+
+        let retrieved = pool.get(&tx_hash);
+        assert!(retrieved.is_some());
+        assert_eq!(retrieved.unwrap().hash(), &tx_hash);
+    }
+
+    #[test]
+    fn test_pool_get_all() {
```
0xKitsune, line 2466: Looks like these have largely the same setup, we could consolidate these into a single `test_2d_pool_helpers` test.

crates/transaction-pool/src/tt_2d_pool.rs
```diff
@@ -2446,4 +2426,1002 @@ mod tests {
...
+
+    #[test]
+    fn test_pool_queued_transactions() {
+        let mut pool = AA2dPool::default();
+        let sender = Address::random();
+
+        // Insert tx0 (pending) and tx2 (queued due to gap)
+        let tx0 = TxBuilder::aa(sender).build();
+        let tx2 = TxBuilder::aa(sender).nonce(2).build();
+        let tx2_hash = *tx2.hash();
```
0xKitsune, line 2515: It might be nice to add a few more txs on each side (pending/queued) and assert both pending and queued in a single test.

PR #1936, opened 2026-01-08
fix(feeAMM): handle zero amounts gracefully

zerosnacks:
Zero amounts are currently allowed to flow through the functions which can have unexpected side effects unless specifically handled

### DIFF
--- a/crates/precompiles/src/tip_fee_manager/amm.rs
+++ b/crates/precompiles/src/tip_fee_manager/amm.rs
@@ -1016,9 +1016,9 @@ mod tests {
         })
     }
 
-    /// Test zero liquidity mint
+    /// Test zero amount validator token
     #[test]
-    fn test_mint_zero_liquidity() -> eyre::Result<()> {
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/tip_fee_manager/amm.rs
```diff
@@ -974,4 +986,62 @@ mod tests {
             Ok(())
         })
     }
+
+    /// Test zero liquidity burn
+    #[test]
+    fn test_burn_zero_liquidity() -> eyre::Result<()> {
```
0xKitsune, line 992: Can we also add a test for `rebalance_swap` where the amount out is 0?

zerosnacks: added in https://github.com/tempoxyz/tempo/pull/1936/changes/fd4c555c2b2f000abf65047555dbe0d4531d90db

updated test names in https://github.com/tempoxyz/tempo/pull/1936/changes/e5c28d3a255c1d1e7b70f81c4f637fc481bf0ba4 for clarity

PR #1941, opened 2026-01-08
fix(revm)!: multicall gas calc

0xrusowsky:
This PR patches a bug in the revm handler, where gas calculation for multicalls was inaccurate. The previous impl had a bug were it would over-refund gas.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2407,7 +2407,7 @@ dependencies = [
 [[package]]
 name = "commonware-broadcast"
 version = "0.0.64"
-source = "git+https://github.com/commonwarexyz/monorepo?rev=82318232ea611fd5ff754b00fa617c6c4a63454b#82318232ea611fd5ff754b00fa617c6c4a63454b"
+source = "git+https://github.com/commonwarexyz/monorepo?rev=5ca1fb0bba4f2f04bf27d1da2b68ef034dca93be#5ca1fb0bba4f2f04bf27d1da2b68ef034dca93be"
 dependencies = [
  "bytes",
  "commonware-codec",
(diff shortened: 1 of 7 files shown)


crates/revm/src/handler.rs
```diff
@@ -378,7 +378,7 @@ where
             }
 
             // Call succeeded - accumulate gas usage and refunds
-            let gas_used = frame_result.gas().used();
+            let gas_used = frame_result.gas().spent();
```
legion2002: Nit: We should rename this to gas_spent, to avoid confusion.
(Don't have to block PR for this)

crates/revm/src/handler.rs
```diff
@@ -360,8 +360,8 @@ where
                 }
 
                 // Include gas from all previous successful calls + failed call
-                let gas_used_by_failed_call = frame_result.gas().used();
-                let total_gas_used = (gas_limit - remaining_gas) + gas_used_by_failed_call;
+                let gas_spent_by_failed_call = frame_result.gas().spent();
+                let total_gas_used = (gas_limit - remaining_gas) + gas_spent_by_failed_call;
```
fgimenez: maybe this should be `total_gas_spent` too?
```suggestion
                let total_gas_spent = (gas_limit - remaining_gas) + gas_spent_by_failed_call;
```

0xrusowsky: addressed in [`2a29def` (#1941)](https://github.com/tempoxyz/tempo/pull/1941/commits/2a29def40040707b54d45c2e95f1a9632bc8b957)

crates/revm/src/handler.rs
0xrusowsky: addressed in [`568f0d5` (#1941)](https://github.com/tempoxyz/tempo/pull/1941/commits/568f0d5c7917e2734dc7ab418e00360a9797d57c)

crates/revm/src/handler.rs
```diff
@@ -2170,4 +2170,98 @@ mod tests {
...
+                input: Bytes::new(),
+            },
+            Call {
+                to: TxKind::Call(Address::random()),
+                value: U256::ZERO,
+                input: Bytes::new(),
+            },
+        ];
+
+        let (mut call_idx, calls_gas) = (0, [(SPENT.0, REFUND.0), (SPENT.1, REFUND.1)]);
```
0xKitsune, line 2226: Do we need `(SPENT.1, REFUND.1)` here? It looks like we never use the last index in this array?

0xrusowsky: mmm it is accessed via `calls_gas[call_idx]` in the second iteration of the loop

PR #1943, opened 2026-01-08
chore(precompiles): add `FixedBytes` storage test

zerosnacks:
This was not explicitly tested, additionally shows the mapping of Solidity's `bytes*` fields <> `FixedBytes<*>`

Artifact from an investigation into alignment of various types during packing

cc @0xrusowsky for visibility


0xrusowsky (approved):
great to bring coverage of the storage primitives up, thanks!

PR #1945, opened 2026-01-08
docs: stablecoin dex refs

jxom:
Fixes references to `stablecoin-dex/` path.


legion2002 (commented):
This section also needs to be updated

https://github.com/tempoxyz/tempo/blob/6e6e239dae2773717330f35a7c5b2669bd5035ae/docs/vocs.config.tsx#L410-L447

PR #1948, opened 2026-01-08
docs(node): add --chain testnet flag to RPC node examples

### LINKED ISSUE #1937: [Docs][RPC] Testnet snapshot (chainId 42429) fails with genesis hash mismatch unless --chain testnet is set
## Summary
When deploying a Tempo RPC node using a testnet snapshot (chainId 42429), starting `tempo node` without explicitly setting `--chain testnet` can select a different default chainspec (logs show `tempo-moderato`), leading to an immediate **genesis hash mismatch** against the snapshot DB.

This is easy to hit for new operators following generic "RPC node" instructions, and the resulting error does not clearly explain the fix.

## Environment
- Network / chain: testnet (chainId 42429)
- OS: Ubuntu (systemd)
- Snapshot: `tempo-42429-9007530-1767762022.tar.lz4` from `tempo-node-snapshots.tempoxyz.dev`
- Node mode: RPC node (`--follow`)

## Steps to reproduce
1. Import a testnet snapshot (chainId 42429) into an empty datadir.
2. Start Tempo without chain flag (or with incorrect chain), e.g.:
  
   tempo node --datadir /path/to/data --follow ...
   3. Observe it exits with genesis mismatch.

## Expected
- Docs should clearly state that testnet (chainId 42429) deployments must use:
 
  --chain testnet
  - Error message could be more actionable (e.g., suggest trying `--chain testnet` when DB genesis hash mismatches).

## Actual
Node exits with:
genesis hash in the storage does not match the specified chainspec:
chainspec is 0xba49153e6f3996a340f3fe3d69e7c0826b717ae672cb115cf0527cada0b78e9e,
database is 0x8561f1bc7c4bbada054f2edc616deb0be221d4b2f712dc0ba8f7ef385ec86ce9Also saw logs indicating a different default chain (e.g. `tempo-moderato`) when chain flag is not set.

## Suggested improvements
- Update RPC node docs to explicitly mention:
  - testnet chainId 42429 requires `--chain testnet`
  - snapshot users must match chain + snapshot
- Improve error text to include a hint about wrong `--chain` selection.

### Additional context

_No response_

0xbyt4:
## Summary

- Add the required `--chain testnet` flag to Quick Start and Systemd service examples in the RPC node documentation
- Add the required `--chain testnet` flag to validator node documentation
- Add an info callout explaining when and why this flag is needed

Without this flag, nodes using testnet snapshots fail with a genesis hash mismatch error, which can be confusing for new operators.

Closes #1937

## Test plan

- [ ] Verify docs build successfully with `bun run build`
- [ ] Check that the info callout renders correctly

### DIFF
--- a/docs/pages/guide/node/rpc.mdx
+++ b/docs/pages/guide/node/rpc.mdx
@@ -13,10 +13,15 @@ tempo download
 
 # Run node
 tempo node \
+  --chain testnet \
   --follow \
   --http --http.port 8545 \
   --http.api eth,net,web3,txpool,trace
 ```
+
+:::info
+When using testnet snapshots, you must specify `--chain testnet`. Without this flag, the node defaults to a different chain specification and will fail with a genesis hash mismatch error.
+:::
 
 ## Manually downloading snapshots
 
@@ -39,6 +44,7 @@ Group=$USER
 Environment=RUST_LOG=info
 WorkingDirectory=$HOME/tempo
 ExecStart=/usr/local/bin/tempo node \\
+  --chain testnet \\
   --datadir <datadir> \\
   --follow \\
   --port 30303 \\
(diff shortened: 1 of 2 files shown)


mattsse (commented):
since new testnet is out should we even update these instructions here @Zygimantass ?

PR #1950, opened 2026-01-08
fix(tip20): prevent cycles in quote token updates for `pathUSD`

0xrusowsky:
## Motivation

the cycle detection in complete_quote_token_update used `while current != PATH_USD_ADDRESS`, which silently bypassed detection when `pathUSD` (with `quote_token = Address::ZERO`) attempted to set a token pointing back to it—the loop exited upon encountering PATH_USD without detecting the cycle.

## Solution

change the termination condition to `while current != Address::ZERO`, correctly traversing the chain to the root.

additionally, the PR adds test coverage for `create_token_reserved_address` to verify that the first deployed token can only be deployed with `quote_token = Address::ZERO`.

### DIFF
--- a/crates/precompiles/src/tip20_factory/mod.rs
+++ b/crates/precompiles/src/tip20_factory/mod.rs
@@ -224,9 +224,8 @@ mod tests {
         error::TempoPrecompileError,
         storage::{ContractStorage, StorageCtx, hashmap::HashMapStorageProvider},
         test_util::TIP20Setup,
-        tip20::TIP20Error,
     };
-    use alloy::primitives::{Address, address};
+    use alloy::primitives::Address;
 
     #[test]
     fn test_create_token() -> eyre::Result<()> {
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/tip20_factory/mod.rs
```diff
@@ -462,4 +463,43 @@ mod tests {
         assert!(is_tip20_prefix(addr4));
         assert!(is_tip20_prefix(addr5));
     }
+
+    #[test]
+    fn test_create_token_reserved_address() -> eyre::Result<()> {
```
0xKitsune: Can we update this test name to describe the failure case this is testing?

Also would be good to have tests for other failure cases, as well as success case for this function. We could also do this in a separate PR if preferred.

0xKitsune (approved):

crates/precompiles/src/tip20_factory/mod.rs
0xrusowsky: addressed in [`f99f169` (#1950)](https://github.com/tempoxyz/tempo/pull/1950/commits/f99f169fec471d80297783d15fa8412fc9fcc370)

0xrusowsky: ended up getting rid of this test and will tackle it in another PR with the others, as the changes became bigger than the actual fix

PR #1952, opened 2026-01-09
test(precompiles): increase coverage for TIP20Factory

0xrusowsky:
## Motivation

Increase test coverage for `TIP20Factory`, specifically for `create_token_reserved_address()` error branches and previously untested functions (`is_initialized`, `get_token_address`).

## Solution

- Add helper functions for `TIP20FactoryError` construction
- Add negative tests for all error branches in `create_token_reserved_address()`
- Add tests for `is_initialized` and `get_token_address`

## Bug Fix

Tests surfaced a bug in `is_initialized` where `info.code.is_some()` was used instead of `!info.is_empty_code_hash()`. The former always returns `true` for default `AccountInfo` (revm sets `code: Some(Bytecode::default())`).

Not critical since the factory is initialized at genesis, but now correctly reflects actual initialization state.

Validated that TIP20 initialization checks already use the correct approach.

### DIFF
--- a/docs/pages/index.mdx
+++ b/docs/pages/index.mdx
@@ -3,15 +3,10 @@ import LucideBookOpen from '~icons/lucide/book-open'
 import LucideUsers from '~icons/lucide/users'
 import LucideMail from '~icons/lucide/mail'
 import * as Card from "../components/Card.tsx"
-import { Callout } from 'vocs/components'
-
-<Callout type="warning">
-  **Testnet migration:** We've launched a new testnet. You'll need to update your RPC configuration and redeploy any contracts. The old testnet will be deprecated on March 8th. [See details below](#testnet-migration).
-</Callout>
 
 # Tempo [Documentation, integration guides, and protocol specifications]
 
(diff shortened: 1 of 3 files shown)


crates/precompiles/src/tip20_factory/mod.rs
```diff
@@ -225,7 +221,133 @@ mod tests {
...
+        StorageCtx::enter(&mut storage, || {
+            // PATH_USD has correct prefix
+            assert!(is_tip20_prefix(PATH_USD_ADDRESS));
+
+            // Address with TIP20 prefix (0x20C0...)
+            let tip20_addr = Address::from(alloy::hex!("20C0000000000000000000000000000000001234"));
+            assert!(is_tip20_prefix(tip20_addr));
+
+            // Random address does not have TIP20 prefix
+            let random = Address::random();
```
fgimenez, line 261: nit, could we use here an address that we know that does not have the TIP20 prefix instead of a random one?

PR #1956, opened 2026-01-09
fix!: validate expiry is in the future when authorizing access keys

legion2002:
- Add `expiry > current_timestamp` validation in `authorizeKey` to prevent creating immediately-invalid keys
- Use existing `ExpiryInPast` error (catches both expiry=0 and past timestamps)
- Add tests for expiry validation in both Rust and Solidity

Addresses review comment from @howydev on PR #1307.

This is a non-critical but breaking fix, meant to prevent footguns while authorizing keys.

Closes CHAIN-466

### DIFF
--- a/crates/precompiles/src/account_keychain/mod.rs
+++ b/crates/precompiles/src/account_keychain/mod.rs
@@ -761,8 +761,13 @@ mod tests {
             };
             let result_past = keychain.authorize_key(account, auth_call_past);
             assert!(
-                result_past.is_err(),
-                "Authorizing with past expiry should fail"
+                matches!(
+                    result_past,
+                    Err(TempoPrecompileError::AccountKeychainError(
+                        AccountKeychainError::ExpiryInPast(_)
+                    ))
+                ),
+                "Expected ExpiryInPast error for past expiry, got: {result_past:?}"
             );
 
             Ok(())


crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -141,6 +141,12 @@ impl AccountKeychain {
             return Err(AccountKeychainError::zero_public_key().into());
         }
 
+        // Expiry must be in the future (also catches expiry == 0 which means "key doesn't exist")
+        let current_timestamp = self.storage.timestamp().saturating_to::<u64>();
+        if call.expiry <= current_timestamp {
+            return Err(AccountKeychainError::expiry_in_past().into());
+        }
```
0xKitsune: Looks like this is breaking snapshot re-execution and likely needs to go behind the T0 hardfork.

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -704,6 +710,61 @@ mod tests {
...
+            match result.unwrap_err() {
+                TempoPrecompileError::AccountKeychainError(e) => {
+                    assert!(
+                        matches!(e, AccountKeychainError::ExpiryInPast(_)),
+                        "Expected ExpiryInPast error, got: {e:?}"
+                    );
+                }
+                e => panic!("Expected AccountKeychainError, got: {e:?}"),
+            }
+
```
0xKitsune, line 753: nit: this can be simplified to just `assert!(matches!())` without the match statement.

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -704,6 +710,61 @@ mod tests {
...
+                signatureType: SignatureType::Secp256k1,
+                expiry: 1, // Very old timestamp - should fail
+                enforceLimits: false,
+                limits: vec![],
+            };
+            let result_past = keychain.authorize_key(account, auth_call_past);
+            assert!(
+                result_past.is_err(),
+                "Authorizing with past expiry should fail"
+            );
```
0xKitsune, line 771: Lets assert the exact error.

0xKitsune (approved):

PR #1957, opened 2026-01-09
chore(precompiles): add `fn is_initialized` to `StorageContract`

0xrusowsky:
## Motivation
Prevent future bugs, like the one addressed in https://github.com/tempoxyz/tempo/pull/1952, where we implement a flawed `fn is_initialized()`

## Solution

Add a `fn is_initialized` to the `StorageContract` trait, with a default impl, so that it is available to all precompiles.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11507,7 +11507,7 @@ dependencies = [
 
 [[package]]
 name = "tempo"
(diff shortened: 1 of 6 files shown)


legion2002 (approved):
Yup this is nice, specially for when we add new precompiles in hardforks.

PR #1959, opened 2026-01-09
feat(rpc): add feePayer field to estimate gas for sponsored txs

legion2002:
## Summary
- Adds `feePayer` field to `TempoTransactionRequest` for gas estimation
- When `feePayer` is specified, uses a different address than caller to trigger cold account access simulation
- Fixes gas underestimation for fee-sponsored transactions with access keys

## Problem
Fee-sponsored transactions with access keys were being underestimated because:
- During gas estimation with `fee_payer: None`, the caller's balance is loaded (often warm)
- During execution with a sponsor, the sponsor's balance requires cold account access (2,600 gas)

## Solution
Added `feePayer` field that accepts:
- `true` - use a dummy address to trigger cold access
- `Address` - use the specified address
- `Account` object (viem format) - extract address from object

## Test plan
- [x] Run viem e2e tests: `VITE_TEMPO_ENV=local-direct pnpm test --project tempo e2e`
- [x] All 31 tests pass including `relay.*webAuthn.*with access key`

### DIFF
--- a/crates/alloy/src/rpc/compat.rs
+++ b/crates/alloy/src/rpc/compat.rs
@@ -41,6 +41,7 @@ impl TryIntoSimTx<TempoTxEnvelope> for TempoTransactionRequest {
                     key_id,
                     tempo_authorization_list,
                     key_authorization,
+                    fee_payer,
                 } = self;
                 let envelope = match TryIntoSimTx::<EthereumTxEnvelope<TxEip4844>>::try_into_sim_tx(
                     inner.clone(),
@@ -57,6 +58,7 @@ impl TryIntoSimTx<TempoTxEnvelope> for TempoTransactionRequest {
                             key_id,
                             tempo_authorization_list,
                             key_authorization,
+                            fee_payer,
                         }));
                     }
                 };
@@ -73,6 +75,7 @@ impl TryIntoSimTx<TempoTxEnvelope> for TempoTransactionRequest {
                             key_id,
                             tempo_authorization_list,
                             key_authorization,
+                            fee_payer,
                         })
                     },
                 )?)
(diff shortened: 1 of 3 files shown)


crates/alloy/src/rpc/request.rs
```diff
@@ -72,6 +72,45 @@ pub struct TempoTransactionRequest {
...
+    /// fee_payer_signature, adding ECRECOVER_GAS (3,000) to intrinsic gas calculation.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub fee_payer: Option<FeePayer>,
+}
+
+/// Fee payer specification for gas estimation.
+/// Can be either a boolean (true = sponsored), a specific address, or an account object.
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum FeePayer {
```
legion2002: We can choose to not have this enum, and opt for something simpler like the fee_payer just being a `bool`, but it might require some changes in viem, or some custom deserializer.

crates/alloy/src/rpc/compat.rs
```diff
@@ -98,11 +101,29 @@ impl TryIntoTxEnv<TempoTxEnv, TempoBlockEnv> for TempoTransactionRequest {
...
+        let resolved_fee_payer = fee_payer.as_ref().and_then(|fp| match fp {
+            crate::rpc::FeePayer::Address(addr) => Some(Some(*addr)),
+            crate::rpc::FeePayer::Account(account) => Some(Some(account.address)),
+            crate::rpc::FeePayer::Bool(true) => {
+                // Use a dummy address different from caller to trigger cold account access
+                // Address(1) is a common dummy/precompile address
+                Some(Some(Address::with_last_byte(1)))
+            }
+            crate::rpc::FeePayer::Bool(false) => None,
+        });
```
onbjerg: Can we make this a helper on `FeePayer` instead that returns an `Option<Address>`?

crates/alloy/src/rpc/request.rs
```diff
@@ -72,6 +72,45 @@ pub struct TempoTransactionRequest {
...
+
+/// Fee payer account object with address field.
+/// Used when viem/wagmi sends the full account object as feePayer.
+/// Unknown fields (keyType, publicKey, source, type, etc.) are ignored during deserialization.
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct FeePayerAccount {
+    /// The fee payer address
+    pub address: Address,
+}
```
onbjerg: why is this specialcased/why doesn't viem/wagmi just use an address?

onbjerg (changes requested):

crates/alloy/src/rpc/request.rs
jxom: I think it’s okay to be a bool?

crates/alloy/src/rpc/request.rs
legion2002: Not sure, I was getting serialization errors without it when fixing a viem test. @jxom do you have more context on this?

crates/alloy/src/rpc/request.rs
```diff
@@ -72,6 +72,12 @@ pub struct TempoTransactionRequest {
     /// Provide a signed KeyAuthorization when the transaction provisions an access key.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub key_authorization: Option<SignedKeyAuthorization>,
+
+    /// Optional fee payer indicator for gas estimation of sponsored transactions.
+    /// When true, indicates the transaction will have a fee_payer_signature,
+    /// adding ECRECOVER_GAS (3,000) to intrinsic gas calculation.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub fee_payer: Option<bool>,
```
fgimenez, line 80: nit, maybe it would be more clear to use `has_fee_payer` or `sponsored` here, wdyt?

fgimenez (approved):
lgtm

jenpaff:
solved by PR #2283

PR #1960, opened 2026-01-09
fix(consensus): exit epochs after DKGs are concluded

SuperFluffy:
Addresses the problem of seeing multiple metrics of the same name.

With this patch, epochs and their backing simplex engines are exited after a DKG is concluded. This makes use of the new `marshal::Mailbox::hint_finalized` API, which allows the marshaller to directly try and fetch finalization certificates and corresponding boundary blocks and does not require any simplex engine to be running.

Before this patch, epochs were only exited when the first block of the next epoch was observed. This was to ensure that all nodes witness the finalization of the boundary certificate. However, this lead to multiple metrics of the same name, as every simplex engine registers new metrics and we had more than 1 running.

With this change the `e2e::tests::sync::joins_from_snapshot` test started failing because we made the node byzantine. This is fixed by reworking the test to remove the validator, have a new one join the set without actually starting, and then copying the identity of the new validator to the old/stopped one. This simulates a snapshot sync without making the node byzantine.

### On shared gas validation and deleted schemes 

With this patch, signing schemes are deleted before a node receives the first for a new epoch. This runs afoul of a workaround we implemented in https://github.com/tempoxyz/tempo/pull/1410 to feed (notarized) parent blocks to the execution layer if the parent block is the boundary block of the last epoch (as there is no signing scheme).

This patch sets `TempoExecutionData.validator_set = None` if the scheme is not available. This data is only used in `TempoBlockExecutor::validate_shared_gas` (originally introduced in https://github.com/tempoxyz/tempo/pull/581) and skipped if the value is not set: https://github.com/tempoxyz/tempo/blob/3c678cb25f8168fccd0ca1725b6ca08a940633b6/crates/evm/src/block.rs#L181

Because this is for the edge case of checking an already notarized block, this seems ok.

### Changes to e2e testing nodes

The onchain validator address and network address (used in validator config smart contracts calls) are now stored inside `struct TestingNode`. No more ad-hoc calls to the `fn validator` utility by guessing the correct index. This makes some tests clearer. Also, the execution node name/identity is now set at construction and no longer set ad-hoc from the public key. This is done to make the "identity-change" hack work to simulate launching from snapshots.

### DIFF
--- a/crates/e2e/src/tests/sync.rs
+++ b/crates/e2e/src/tests/sync.rs
@@ -185,7 +185,6 @@ fn joins_from_snapshot() {
                 }
             }
             if validators_at_epoch == 4 {
-                std::fs::write("metrics", metrics).unwrap();
                 break;
             }
         }


crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -781,19 +781,29 @@ async fn verify_block<TContext: Pacer>(
         );
         return Ok(false);
     }
-    let scheme = scheme_provider
-        .scoped(epoch)
-        .ok_or_eyre("cannot determine participants in the current epoch")?;
+
+    // FIXME: in cases where validate_block is called on the boundary block,
```
SuperFluffy, line 785: @joshieDo @klkvr Flagging this. Is it critical that the validator set be available? This leads to problems when feeding it the parent of the first block of an epoch (parent has epoch - 1), since the scheme for that epoch is no longer available.

We can work around this during normal operation by not deleting the schemes (which I tried), but it fails when restarting a node right after it finalized the boundary because nobody provides that anymore.

crates/e2e/src/testing_node.rs
```diff
@@ -51,6 +52,8 @@ where
     pub execution_database: Option<Arc<DatabaseEnv>>,
     /// Last block number in database when stopped (used for restart verification)
     pub last_db_block_on_stop: Option<u64>,
+    pub network_address: SocketAddr,
+    pub chain_address: Address,
```
SuperFluffy, line 63: Stuffed this into the testing-node, because that's where it belongs. It was very confusing adding and removing the nodes in the reworked `e2e::tests::sync::joins_from_snapshot` test.

crates/e2e/src/tests/sync.rs
```diff
@@ -4,45 +4,162 @@
...
     Clock as _, Metrics as _, Runner as _,
     deterministic::{self, Context, Runner},
 };
 use futures::future::join_all;
-use tracing::info;
+use reth_ethereum::provider::BlockNumReader as _;
 
 use crate::{CONSENSUS_NODE_PREFIX, Setup, setup_validators};
 
+#[ignore = "the mechanism to fake an advanced start currently fails with a database inconsistency"]
```
SuperFluffy: @joshieDo flagging this, see the comment I provided inline in the test below.

crates/e2e/src/tests/sync.rs
```diff
@@ -4,45 +4,162 @@
...
+            let provider = receiver.execution_provider();
+            let height = provider.best_block_number().unwrap();
+            FixedEpocher::new(NonZeroU64::new(epoch_length).unwrap())
+                .containing(Height::new(height))
+                .unwrap()
+                .epoch()
+        };
+        receiver.stop().await;
+
+        // FIXME: This panics with, even though the docs suggest that this
```
SuperFluffy, line 122: This is the first failure - can't get the provider after taking down the node. Worked around this by fetching the best block while it's still running.

This here does not cause issues though and can be worked around.

crates/e2e/src/tests/sync.rs
```diff
@@ -4,45 +4,162 @@
...
+        // Now turn the receiver into the donor - except for the database dir and
+        // env. This simulates a start from a snapshot.
+        receiver.uid = donor.uid;
+        receiver.public_key = donor.public_key;
+        receiver.consensus_config = donor.consensus_config;
+        receiver.execution_config = donor.execution_config;
+        receiver.network_address = donor.network_address;
+        receiver.chain_address = donor.chain_address;
+
+        // FIXME: currently fails with a panic:
```
SuperFluffy: @joshieDo this is the fundamental issue. Restarting the node with the changed identities does not work.

While writing this - the problem is likely that I overwrite `public_key`?

crates/e2e/src/tests/sync.rs
SuperFluffy: Fixed it by keeping the execution node name/identity constant. Now only the initial public key is considered.

crates/e2e/src/tests/sync.rs
SuperFluffy: Fixed it by keeping the execution node identity constant.

crates/e2e/src/testing_node.rs
```diff
@@ -49,8 +50,17 @@ where
     pub execution_config: ExecutionNodeConfig,
     /// Database instance for the execution node
     pub execution_database: Option<Arc<DatabaseEnv>>,
+    /// The execution node name assigned at initialization. Important when
+    /// constructing the datadir at which to find the node.
+    pub execution_node_name: String,
```
SuperFluffy, line 55: This is now set on init and ensures that when restarting the node, the execution runtime can find the database dir.

crates/e2e/src/testing_node.rs
```diff
@@ -160,7 +176,7 @@ where
         let execution_node = self
             .execution_runtime
             .spawn_node(
-                &execution_runtime::execution_node_name(&self.public_key),
+                &self.execution_node_name,
```
SuperFluffy, line 181: Related - when calling `spawn_node` further down, this argument is eventually joined with the execution-runtime internal `datadir`. So we need to keep this constant even if we change the node identity.

crates/e2e/src/tests/sync.rs
```diff
@@ -4,47 +4,157 @@
...
+        receiver.chain_address = donor.chain_address;
+        receiver.start().await;
+
+        tracing::info!(
+            uid = %receiver.uid,
+            "started the validator with a changed identity",
+        );
 
         loop {
+            context.sleep(Duration::from_secs(1)).await;
```
SuperFluffy, line 156: Moved this to the start of the loop to give the node some time to actually fire up its first simplex engine. Otherwise its `_epoch_manager_latest_epoch` metric will stay at zero, causing the test to fail.

crates/commonware-node/src/consensus/application/actor.rs
SuperFluffy: I put this in the PR message body, but copying it here for completeness:

## On shared gas validation

With this patch, signing schemes are deleted before a node receives the first for a new epoch. This runs afoul of a workaround we implemented in https://github.com/tempoxyz/tempo/pull/1410 to feed (notarized) parent blocks to the execution layer if the parent block is the boundary block of the last epoch (as there is no signing scheme).

This patch sets `TempoExecutionData.validator_set = None` if the scheme is no available. This data is only used in `TempoBlockExecutor::validate_shared_gas` (originally introduced in https://github.com/tempoxyz/tempo/pull/581) and skipped if the value is not set: https://github.com/tempoxyz/tempo/blob/3c678cb25f8168fccd0ca1725b6ca08a940633b6/crates/evm/src/block.rs#L181

Because this is for the edge case of checking an already notarized block, this seems ok.

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -382,38 +341,54 @@ where
...
+                                                "run_dkg_loop",
+                                                epoch = %state.epoch
+                                            ).in_scope(|| info!(
+                                                "constructed a new epoch state; \
+                                                persisting new state and exiting \
+                                                current epoch",
+                                            ));
+
+                                            if let Err(err) = storage
+                                                .append_state(new_state)
```
SuperFluffy, line 375: This also moves `append_storage` to before the final block is acknowledged. This is actually critical: if anything goes went wrong after the last block was acknowledged but before the state was committed to disk, there would be no way to recover.

joshieDo (approved):

PR #1965, opened 2026-01-09
chore: simplify size functions

DaniPopes:
Use `size_of::<Self>()` instead of manually summing field sizes for in-memory size calculations. This is simpler and more maintainable.

Re-opening #1690
https://github.com/alloy-rs/alloy/pull/3403

### DIFF
--- a/crates/primitives/src/transaction/tt_authorization.rs
+++ b/crates/primitives/src/transaction/tt_authorization.rs
@@ -108,7 +108,7 @@ impl TempoSignedAuthorization {
 
     /// Calculates a heuristic for the in-memory size of this authorization
     pub fn size(&self) -> usize {
-        core::mem::size_of::<Authorization>() + self.signature.size()
+        size_of::<Self>()
     }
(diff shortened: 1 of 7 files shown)


crates/primitives/src/transaction/tt_authorization.rs
```diff
@@ -108,7 +108,7 @@ impl TempoSignedAuthorization {
 
     /// Calculates a heuristic for the in-memory size of this authorization
     pub fn size(&self) -> usize {
-        core::mem::size_of::<Authorization>() + self.signature.size()
+        size_of::<Self>()
```
0xrusowsky, line 111: is this equivalent?

0xrusowsky (approved):
besides the question LGTM

crates/primitives/src/transaction/tt_authorization.rs
DaniPopes: yes

PR #1966, opened 2026-01-09
fix(consensus): prune state segments only after reading them first

SuperFluffy:
Fixes overpruning the DKG state journal because epochs were mapped to segments 1-to-1.

Instead, the previous (never the latest) state is read from the journal. If `previous.epoch >= target_epoch`, all segments below it are pruned. If `previous.epoch < target_epoch`, then all segments below the latest (the last segment in the journal) are pruned.

### DIFF
--- a/crates/commonware-node/src/dkg/manager/actor/state.rs
+++ b/crates/commonware-node/src/dkg/manager/actor/state.rs
@@ -457,10 +457,16 @@ where
             .await
             .wrap_err("unable to prune events journal")?;
 
-        // Figure out the *previous* state and determine whether we can prune.
+        // Cannot map epochs directly to segments like in the events journal.
+        // Need to first check what the epoch of the state is and go from there.
+        //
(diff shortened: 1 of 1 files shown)


crates/commonware-node/src/dkg/manager/actor/state.rs
```diff
@@ -456,10 +456,24 @@ where
...
-        self.states
-            .prune(up_to_epoch.get())
-            .await
-            .wrap_err("unable to prune outcomes journal")?;
+
+        // Figure out the *previous* state and determine whether we can prune.
+        if let Some(previous_segment) = self.states.size().checked_sub(2)
+            && let Ok(previous_state) = self.states.read(previous_segment).await
+        {
+            let to_prune = if previous_state.epoch >= up_to_epoch {
```
joshieDo, line 470: `>` should never happen, i think we could just add a comment or bail like an `expect` ?

PR #1971, opened 2026-01-09
refactor(transaction-pool): consolidate tests and remove low-value test coverage

fgimenez:
Ref CHAIN-421
Ref #1930

Cleans up transaction-pool tests by:
- Consolidating tests with similar setup into single tests
- Removing low-value tests that only verify default/trivial behavior
- Using `Address::random()` instead of hardcoded addresses
- Replacing `assert!(result.is_err()) + match` patterns with `assert!(matches!())`

### DIFF
--- a/crates/transaction-pool/src/validator.rs
+++ b/crates/transaction-pool/src/validator.rs
@@ -1188,7 +1188,7 @@ mod tests {
         #[test]
         fn test_non_aa_transaction_skips_keychain_validation() -> Result<(), ProviderError> {
             // Non-AA transaction should return Ok(Ok(())) immediately
-            let transaction = get_transaction(None);
+            let transaction = TxBuilder::eip1559(Address::random()).build_eip1559();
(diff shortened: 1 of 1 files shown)


crates/transaction-pool/src/validator.rs
```diff
@@ -1049,59 +977,34 @@ mod tests {
...
+                );
+            }
+            TransactionValidationOutcome::Valid { .. } => {
+                // Full validation passed - even better!
+            }
+            TransactionValidationOutcome::Error(_, _) => {
+                // Provider errors are acceptable, just not NonZeroValue rejection
+            }
         }
     }
```
0xKitsune: We should update to always expect success rather than allow the test to pass with other errors.

fgimenez: sure done here https://github.com/tempoxyz/tempo/pull/1971/commits/60e59a004b35a9a50aa7a7d3f548a90ef1195327

PR #1978, opened 2026-01-10
fix: revert 1950, re-enable snapshot execution tests

0xKitsune:
This PR reverts 1950 and re-enables snapshot execution tests to ensure changes do not break syncing.

### DIFF
--- a/crates/precompiles/src/tip20/mod.rs
+++ b/crates/precompiles/src/tip20/mod.rs
@@ -1750,23 +1750,6 @@ pub(crate) mod tests {
     }
 
     #[test]
-    fn test_deploy_path_usd_directly() -> eyre::Result<()> {
-        let mut storage = HashMapStorageProvider::new(1);
-        let admin = Address::random();
-
-        StorageCtx::enter(&mut storage, || {
-            // pathUSD is at a reserved address, so we initialize it directly (not via factory)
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/tip20/mod.rs
```diff
@@ -1785,6 +1749,23 @@ pub(crate) mod tests {
...
+            let mut path_usd = TIP20Token::from_address(PATH_USD_ADDRESS)?;
+            path_usd.initialize(admin, "pathUSD", "pathUSD", "USD", PATH_USD_ADDRESS, admin)?;
+
+            assert_eq!(path_usd.currency()?, "USD");
+            // pathUSD uses itself as quote token
+            assert_eq!(path_usd.quote_token()?, PATH_USD_ADDRESS);
+            Ok(())
+        })
+    }
+
```
0xrusowsky: this test is incorrect though, we should delete it, as it must be created via `create_token_reserved_address`

0xrusowsky (changes requested):

0xrusowsky (approved):

crates/precompiles/src/tip20/mod.rs
0xrusowsky: fixed with [`dc4ab77` (#1978)](https://github.com/tempoxyz/tempo/pull/1978/commits/dc4ab7714caa22b78c8f6c0cb8596b4fcc991794)

PR #1980, opened 2026-01-11
fix(tip20): require root tokens quote token to be `address(0)`

0xKitsune:
Stacks on #1978 

Closes ZELLIC-19

This PR updates `set_next_quote_token()` and `complete_quote_token_update()` to reject non-zero quote tokens for the root token. This prevents cycles that could occur if the root token's quote token were changed to point to another token in the chain.

### DIFF
--- a/crates/precompiles/src/tip20/mod.rs
+++ b/crates/precompiles/src/tip20/mod.rs
@@ -1956,6 +1956,7 @@ pub(crate) mod tests {
                 )))
             ));
 
+            // assert that quote tokens are unchanged
             assert_eq!(token_a.quote_token()?, token_b.address);
             assert_eq!(token_b.quote_token()?, PATH_USD_ADDRESS);
 


crates/precompiles/src/tip20/mod.rs
```diff
@@ -267,20 +274,30 @@ impl TIP20Token {
...
         let next_quote_token = self.next_quote_token()?;
 
-        // Check that this does not create a loop
-        // Loop through quote tokens until we reach the root
-        let mut current = next_quote_token;
-        while current != Address::ZERO {
-            if current == self.address {
+        // If the token address is the root, require that the quote token is `Address::ZERO`
+        if self.address == PATH_USD_ADDRESS {
+            if next_quote_token != Address::ZERO {
```
0xrusowsky: we can collapse the double if into:

```
if self.address == PATH_USD_ADDRESS && next_quote_token != Address::ZERO {
```

as we are anyways blocking quote token updates on pathUSD with this change

crates/precompiles/src/tip20/mod.rs
```diff
@@ -244,19 +244,26 @@ impl TIP20Token {
...
-        }
-
-        // Check if the currency is USD, if so then the quote token's currency MUST also be USD
-        let currency = self.currency()?;
-        if currency == USD_CURRENCY {
-            let quote_token_currency = Self::from_address(call.newQuoteToken)?.currency()?;
-            if quote_token_currency != USD_CURRENCY {
+        // If the token address is the root, require that the quote token is `Address::ZERO`
+        if self.address == PATH_USD_ADDRESS {
+            if call.newQuoteToken != Address::ZERO {
```
0xrusowsky: i think the check is unnecessary here, as we already check it in `fn complete_quote_token_update()`.

imo one check is enough, as this way we minimize LOC + it is more in line with the existing approach

0xKitsune:
@0xrusowsky I have updated the approach to ensure that `pathUSD` sets it's quote token as `adddress(0)` when calling `create_token_reserved_address()`. Additionally, the PR has been updated to reject quote token updates for `pathUSD` when calling `set_next_quote_token()`. Lmk what you think.

crates/precompiles/src/tip20_factory/mod.rs
```diff
@@ -163,6 +163,11 @@ impl TIP20Factory {
             ));
         }
 
+        // pathUSD must set address(0) as the quote token
+        if address == PATH_USD_ADDRESS && !quote_token.is_zero() {
+            return Err(TIP20Error::invalid_quote_token().into());
+        }
+
```
0xrusowsky: nit: we could move this inside the following block, which already validates the quote token

```rs
if !quote_token.is_zero() {
    // pathUSD must set address(0) as the quote token
    if address == PATH_USD_ADDRESS {
    }
    // quote_token must be address(0) or a valid TIP20
    else if !self.is_tip20(quote_token)? {
    }

    // If token is USD, its quote token must also be USD
    if currency == USD_CURRENCY
        && TIP20Token::from_address(quote_token)?.currency()? != USD_CURRENCY
    {
    }
}
```

or if we want it more succinct:

```rs
if !quote_token.is_zero() {
    // quote_token must be address(0) or a valid TIP20. pathUSD must use address(0)
    if address == PATH_USD_ADDRESS || !self.is_tip20(quote_token)? {
    }

    // If token is USD, its quote token must also be USD
    if currency == USD_CURRENCY
        && TIP20Token::from_address(quote_token)?.currency()? != USD_CURRENCY
    {
    }
}
```

0xrusowsky (approved):

crates/precompiles/src/tip20_factory/mod.rs
0xKitsune: Updated in [fbcd7619](https://github.com/tempoxyz/tempo/pull/1980/commits/fbcd7619bf1799f28be0c859ead1207d16bca9e9), PTAL.

PR #1982, opened 2026-01-11
fix(rpc): disable local pending block

mattsse:
we can't build a local pending block because vanilla block building here would fail due to missing system tx

so this will effectively pin "pending" to "latest"

### DIFF
--- a/docs/specs/src/StablecoinDEX.sol
+++ b/docs/specs/src/StablecoinDEX.sol
@@ -588,11 +588,14 @@ contract StablecoinDEX is IStablecoinDEX {
                 );
             }
 
+            bytes32 bookKey = order.bookKey;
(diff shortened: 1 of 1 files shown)


0xKitsune (approved):

crates/node/src/rpc/mod.rs
```diff
@@ -223,7 +223,8 @@ impl<N: FullNodeTypes<Types = TempoNode>> LoadPendingBlock for TempoEthApi<N> {
 
     #[inline]
     fn pending_block_kind(&self) -> PendingBlockKind {
-        self.inner.pending_block_kind()
+        // don't build a local pending block
```
0xKitsune: nit: should update this comment to add why we disable pending blocks?

PR #1984, opened 2026-01-11
feat: configure custom endpoints via cli

klkvr:
Changes rpc code so that custom endpoints are enabled via `--http.api` arg vs being always enabled by default

### DIFF
--- a/bin/tempo/src/main.rs
+++ b/bin/tempo/src/main.rs
@@ -29,10 +29,12 @@ use reth_ethereum::{
     chainspec::EthChainSpec as _,
     cli::{Cli, Commands},
     evm::revm::primitives::B256,
+    rpc::builder::{RethRpcModule, RpcModuleSelection},
 };
 use reth_ethereum_cli as _;
 use reth_node_builder::{NodeHandle, WithLaunchContext};
-use std::{sync::Arc, thread};
+use reth_rpc_server_types::RpcModuleValidator;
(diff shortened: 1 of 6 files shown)


shekhirin (approved):

bin/tempo/src/main.rs
```diff
@@ -49,6 +51,32 @@ use tempo_node::{
...
+    fn parse_selection(s: &str) -> Result<RpcModuleSelection, String> {
+        // First try standard parsing
+        let selection = RpcModuleSelection::from_str(s)
+            .map_err(|e| format!("Failed to parse RPC modules: {e}"))?;
+
+        // Validate each module in the selection
+        if let RpcModuleSelection::Selection(modules) = &selection {
+            for module in modules {
+                if let RethRpcModule::Other(name) = module
+                    && !matches!(name.as_str(), "amm" | "dex" | "token" | "policy")
```
mattsse: should we introduce an enum for this?, like we have for RethModule?

mattsse (changes requested):
smol suggestion

bin/tempo/src/main.rs
klkvr: added in 9cff3ac

0xKitsune (commented):
Do we still need this now with #1991 merged?

mattsse:
no longer needed then

PR #1985, opened 2026-01-11
chore: use sealed api

mattsse:
mostlikely this is "latest" so we can avoid the hashing here by using the sealed_ api


mattsse:
closing because we yeet these fns

PR #1987, opened 2026-01-12
fix(revm): ensure that adjusted gas isn't above the gas limit before

0xrusowsky:
ref: `CHAIN-446`

## Motivation

2D nonce gas costs are calculated after initial gas validation but added to intrinsic gas before execution. This creates a window where a transaction can pass validation with a gas limit that's insufficient to cover the full intrinsic cost (base + 2D nonce).

When execution computes remaining gas via `gas_limit - adjusted_intrinsic_gas`, the subtraction underflows in release mode, giving the execution loop ~2^64 gas units. An attacker can exploit this to force nodes to expend massive computational resources on a single transaction before it eventually fails during fee reimbursement.

## Solution

Prevent wasting computation by ensuring that `adjusted_intrinsic_gas < gas_limit` at the start of AA execution, before any call frames are created.

### DIFF
--- a/crates/revm/src/handler.rs
+++ b/crates/revm/src/handler.rs
@@ -1442,12 +1442,13 @@ pub fn validate_time_window(
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::{TempoBlockEnv, TempoTxEnv};
-    use alloy_primitives::{Address, B256, U256};
+    use crate::{TempoBlockEnv, TempoTxEnv, evm::TempoEvm, tx::TempoBatchCallEnv};
+    use alloy_primitives::{Address, B256, Bytes, TxKind, U256};
     use revm::{
         Context, Journal, MainContext,
         context::CfgEnv,
         database::{CacheDB, EmptyDB},
+        handler::Handler,
         interpreter::instructions::utility::IntoU256,
         primitives::hardfork::SpecId,
     };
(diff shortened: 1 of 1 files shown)


crates/revm/src/handler.rs
```diff
@@ -500,7 +500,18 @@ where
         );
 
         // Check if this is an AA transaction by checking for tempo_tx_env
-        if let Some(tempo_tx_env) = evm.ctx().tx().tempo_tx_env.as_ref() {
+        let evm_ctx = evm.ctx();
+        if let Some(tempo_tx_env) = evm_ctx.tx().tempo_tx_env.as_ref() {
+            // Ensure gas limit covers 2D nonce cost
+            let gas_limit = evm_ctx.tx().gas_limit();
```
0xKitsune, line 506: nit: it looks like we can cache `evm.ctx().tx()` instead of `evm_ctx`.

crates/revm/src/handler.rs
```diff
@@ -2128,6 +2150,77 @@ mod tests {
...
+    fn test_2d_nonce_gas_limit_validation() {
+        use crate::{evm::TempoEvm, tx::TempoBatchCallEnv};
+        use alloy_primitives::{Bytes, TxKind};
+        use revm::{
+            Context, Journal,
+            context::CfgEnv,
+            database::{CacheDB, EmptyDB},
+            handler::Handler,
+        };
+        use tempo_primitives::transaction::Call;
```
0xKitsune: nit: can we move imports to the top of the tests mod?

0xrusowsky: addressed in [`9d37cdd` (#1987)](https://github.com/tempoxyz/tempo/pull/1987/commits/9d37cdd3906ae5f7cd816bc43d86f598ec94e516)

crates/revm/src/handler.rs
0xrusowsky: i need to bind `evm.ctx()` alone apparently, so prob not worth it?

PR #1988, opened 2026-01-12
feat(genesis): add mainnet gas token support

Zygimantass:


xtask/src/genesis_args.rs
```diff
@@ -265,6 +277,41 @@ impl GenesisArgs {
...
+                    "DONOTUSE",
+                    "USD",
+                    PATH_USD_ADDRESS,
+                    self.mainnet_gas_token_admin.expect(
+                        "Mainnet gas token admin is required if you want to deploy the token",
+                    ),
+                    &addresses,
+                    U256::from(u64::MAX),
+                    None,
+                    Some(B256::from(salt_bytes)),
```
0xKitsune: Instead of generating `salt_bytes` like this, we can use `B256::random()`.

xtask/src/genesis_args.rs
```diff
@@ -571,9 +624,36 @@ fn create_and_mint_token(
...
             "TIP20Factory must be initialized before creating tokens"
         );
-        let token_address = factory
-            .create_token_reserved_address(address, name, symbol, currency, quote_token, admin)
-            .expect("Could not create token");
+
+        ensure!(
+            salt.is_some() || address.is_some(),
+            "Either salt or address must be specified"
+        );
```
0xKitsune: Looks like since we use `ensure!(salt.is_some() != address.is_some())` below, we do not need this check as this ensures either salt or address is some.

xtask/src/genesis_args.rs
```diff
@@ -559,7 +611,8 @@ fn create_and_mint_token(
     admin: Address,
     recipients: &[Address],
     mint_amount: U256,
-    address: Address,
+    address: Option<Address>,
+    salt: Option<B256>,
```
0xKitsune: Consider making this a single enum instead (eg. `SaltOrAddress`).

xtask/src/genesis_args.rs
```diff
@@ -135,6 +136,14 @@ pub(crate) struct GenesisArgs {
     #[arg(long)]
     no_extra_tokens: bool,
 
+    /// Enable creating mainnet gas token.
+    #[arg(long)]
+    mainnet_gas_token: bool,
+
+    /// Custom admin address for mainnet gas token.
+    #[arg(long)]
+    mainnet_gas_token_admin: Option<Address>,
```
0xKitsune: nit: consider renaming to `deployment_gas_token` or similar to avoid confusion. Currently this might be misconstrued as the canonical mainnet gas token.

xtask/src/genesis_args.rs
Zygimantass: I want to use the seed specifically, so that's why I pass the rng

xtask/src/genesis_args.rs
Zygimantass: good call

xtask/src/genesis_args.rs
```diff
@@ -265,6 +274,42 @@ impl GenesisArgs {
...
+        }
+
+        let deployment_gas_token = {
+            if self.deployment_gas_token {
+                let mut rng = rand::rngs::StdRng::seed_from_u64(
+                    self.seed.unwrap_or_else(rand::random::<u64>),
+                );
+
+                let mut salt_bytes = [0u8; 32];
+                rand::Rng::fill(&mut rng, &mut salt_bytes);
```
0xKitsune, line 289: We can use `B256::random()` here

xtask/src/genesis_args.rs
```diff
@@ -265,6 +274,42 @@ impl GenesisArgs {
...
+
+                let address = create_and_mint_token(
+                    "DONOTUSE",
+                    "DONOTUSE",
```
0xKitsune, line 293: Confirming that we want the name/symbol to be `DONOTUSE` rather than `deployUSD`?

xtask/src/genesis_args.rs
```diff
@@ -297,7 +342,12 @@ impl GenesisArgs {
...
 
         println!("Initializing fee manager");
-        let default_fee_token = alpha_token_address.unwrap_or(PATH_USD_ADDRESS);
+        let default_fee_token = if let Some(address) = deployment_gas_token {
+            address
+        } else {
+            alpha_token_address.unwrap_or(PATH_USD_ADDRESS)
+        };
+
         initialize_fee_manager(
```
0xKitsune, line 356: Just noting that in `initialize_fee_manager`, `pathUSD` is still used as the gas token for validators specified. Is this the expected behavior?

0xKitsune (approved):
A few additional comments, would also like a second set of eyes from @fgimenez.

xtask/src/genesis_args.rs
Zygimantass: correct

xtask/src/genesis_args.rs
Zygimantass: we need it to be seeded, so we cant use this

PR #1989, opened 2026-01-12
feat: add invariants

howydev:
1. adding some more invariants on linked list 
2. adding a dust invariant for orders

### DIFF
--- a/docs/specs/test/invariants/StablecoinDEX.t.sol
+++ b/docs/specs/test/invariants/StablecoinDEX.t.sol
@@ -968,11 +968,11 @@ contract StablecoinDEXInvariantTest is BaseTest {
                 assertEq(askLiquidity, 0, "TEMPO-DEX7: ask liquidity non-zero but head is zero");
             } else {
                 // TEMPO-DEX17: head.prev should be 0
-                IStablecoinDEX.Order memory headOrder = exchange.getOrder(bidHead);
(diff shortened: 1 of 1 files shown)


docs/specs/test/invariants/StablecoinDEX.t.sol
```diff
@@ -934,23 +940,39 @@ contract StablecoinDEXInvariantTest is BaseTest {
...
                 // TEMPO-DEX7: If liquidity > 0, head should be non-zero
                 assertTrue(bidHead != 0, "TEMPO-DEX7: bid tick has liquidity but no head");
+                assertTrue(bidTail != 0, "TEMPO-DEX7: bid tick has liquidity but no tail");
                 // TEMPO-DEX11: Bitmap correctness verified indirectly via bestBidTick/bestAskTick in _assertBestTickConsistency
             }
             if (bidHead == 0) {
                 // If head is 0, tail should also be 0 and liquidity should be 0
                 assertEq(bidTail, 0, "TEMPO-DEX10: bid tail non-zero but head is zero");
                 assertEq(bidLiquidity, 0, "TEMPO-DEX7: bid liquidity non-zero but head is zero");
+            } else {
```
grandizzy, line 950: I think this should be `} else if (bidHead != 0) {` the test fails if 0, is that expected?

PR #1990, opened 2026-01-12
fix! : access keys from 7702 delegating accounts

legion2002:
Currently authorization list checks for Tempo Txs, allow access keys to 7702 delegate a user's account.
Delegating an account is a critical function, and should only be handled by the main key.

Additionally, we don't have access to the keychain precompile in the `recover_signer` function, the `authorize()` function in revm, uses recover_signer internally to verify the authorization signature. 
This means that currently, you can use a random access key, to delegate a user's account. 

This PR fixes it by completely blocking access keys from doing 7702 delegations.

Additionally we highlight this footgun in the natspec docs, and we should think about if there's a way to remove this, as it could lead to critical issues for people utilizing the `recover_signer` function.

As a follow-up task, I am currently reviewing all other uses of `recover_signer` to double check that they are not affected. This was missed originally, because for 7702 auths, the `recover_signer` is hidden under revm's `authorize()` function.

Fixes [ZELLIC 21](https://linear.app/tempoxyz/issue/ZELLIC-21)


crates/primitives/src/transaction/tt_signature.rs
```diff
@@ -593,6 +593,10 @@ impl TempoSignature {
     ///
     /// For Keychain signatures, this performs full validation of the inner signature.
     /// The access key address is cached in the KeychainSignature for later use.
+    /// Note: This pattern has a big footgun, that someone using recover_signer, cannot assume
+    /// that the signature is valid for the keychain. They also need to check the access key is authorized
+    /// in the keychain precompile.
+    /// We cannot check this here, as we don't have access to the keychain precompile.
```
0xrusowsky, line 599: do we have any ideas on how to protect against this?

crates/node/tests/it/tempo_transaction.rs
```diff
@@ -3108,6 +3108,139 @@ async fn test_tempo_authorization_list() -> eyre::Result<()> {
     Ok(())
 }
 
+/// Test that keychain signatures in tempo_authorization_list are rejected.
+#[tokio::test(flavor = "multi_thread")]
+async fn test_keychain_authorization_in_auth_list_is_skipped() -> eyre::Result<()> {
+    use tempo_primitives::transaction::tt_signature::KeychainSignature;
```
0xKitsune: nit: can we move imports to the top of the `tests` mod?

crates/primitives/src/transaction/tt_signature.rs
0xKitsune: @legion2002 can we open an issue to track/resolve this? Id rather not have comments noting big footguns that arent TODOs with a corresponding issue to track. We can do this in a separate PR if not critical but this should be addressed.

legion2002: Don't have any good thoughts about this yet, because it is used in upstream dependencies, so not easy to change anything about it. (also jammed with amp on this, and couldn't come up with anything good)

Lmk if you have any ideas!

crates/revm/src/handler.rs
```diff
@@ -569,6 +569,11 @@ where
             let mut refunded_accounts = 0;
 
             for authorization in &tempo_tx_env.tempo_authorization_list {
+                // Access keys cannot perform EIP-7702 delegation - skip keychain signatures
+                if authorization.signature().is_keychain() {
+                    continue;
+                }
+
```
klkvr, line 576: can we make such auths unrepresentable? i.e change this to `PrimitiveSignature` https://github.com/tempoxyz/tempo/blob/9cff3ac213054feeda0d519b9d9ef6990b4af94d/crates/primitives/src/transaction/tt_authorization.rs#L28-L29

PR #1992, opened 2026-01-12
chore(test): Test initial gas for simple tx and batch calls

rakita:
Testing expected initial gas cost.


crates/revm/src/evm.rs
```diff
@@ -754,6 +768,115 @@ mod tests {
         Ok(())
     }
 
+    use tempo_precompiles::storage::Handler;
```
0xKitsune: nit: can we move imports to the top of the test mod?

crates/revm/src/evm.rs
```diff
@@ -754,6 +768,115 @@ mod tests {
...
+        let internals = EvmInternals::new(&mut evm.ctx.journaled_state, &block);
+        let mut provider =
+            EvmPrecompileStorageProvider::new_max_gas(internals, &Default::default());
+
+        _ = StorageCtx::enter(&mut provider, || {
+            TIP20Setup::path_usd(caller)
+                .with_issuer(caller)
+                .with_mint(caller, U256::from(100_000))
+                .apply()
+        });
```
0xKitsune: Lets propagate this error rather than ignore.

crates/revm/src/evm.rs
```diff
@@ -754,6 +768,115 @@ mod tests {
...
+        let mut provider =
+            EvmPrecompileStorageProvider::new_max_gas(internals, &Default::default());
+
+        let slot = StorageCtx::enter(&mut provider, || {
+            TIP20Token::from_address(PATH_USD_ADDRESS)?
+                .balances
+                .at(caller)
+                .read()
+        })
+        .unwrap();
```
0xKitsune: Same here lets propagate rather than unwrap().

crates/revm/src/evm.rs
```diff
@@ -754,6 +768,115 @@ mod tests {
...
+        drop(provider);
+
+        assert_eq!(slot, U256::from(100_000));
+
+        let result1 = evm.transact_commit(tx_env1)?;
+        //println!("{result1:#?}");
```
0xKitsune: Can we remove this commented out println?

0xKitsune (approved):

crates/revm/src/evm.rs
```diff
@@ -754,6 +768,107 @@ mod tests {
...
+            .with_max_fee_per_gas(200_000_000_000)
+            .with_max_priority_fee_per_gas(0)
+            .build();
+
+        let signed_tx2 = key_pair.sign_tx(tx2)?;
+        let tx_env2 = TempoTxEnv::from_recovered_tx(&signed_tx2, caller);
+
+        let result2 = evm.transact_commit(tx_env2)?;
+        assert!(result2.is_success());
+        assert_eq!(result2.gas_used(), 31_286);
```
legion2002, line 853: Is there a justification for this 31286 - 28671 (2615) difference between 1 call and 2 calls ?

rakita: Second one has one more call in batch

PR #2002, opened 2026-01-12
feat(chainspec): add T0 hardfork with security fixes

legion2002:
Introduces the T0 hardfork which gates two security fixes:

1. Prevents access keys from performing EIP-7702 delegation
   - Keychain signatures are now skipped in authorization list processing
   - Fixes potential attack vector where access keys could delegate accounts

2. Ensures gas limit covers 2D nonce cost before AA execution
   - Prevents underflow when computing remaining gas
   - Blocks DoS attack exploiting insufficient gas validation

Changes:
- Add T0 variant to TempoHardfork enum
- Add is_t0() method and is_t0_active_at_timestamp() trait method
- Add t0_time field to TempoGenesisInfo for activation timestamp
- Gate both fixes behind T0 hardfork in handler.rs
- Enable T0 at genesis (timestamp 0) in dev.json

Amp-Thread-ID: https://ampcode.com/threads/T-019bb3c0-b3f2-75c5-a129-5c402ff06c5e

### DIFF
--- a/crates/chainspec/src/hardfork.rs
+++ b/crates/chainspec/src/hardfork.rs
@@ -81,8 +81,12 @@ impl From<TempoHardfork> for SpecId {
 }
 
 impl From<SpecId> for TempoHardfork {
-    fn from(_spec: SpecId) -> Self {
-        Self::Genesis
+    fn from(spec: SpecId) -> Self {
+        if spec.is_enabled_in(SpecId::from(Self::T0)) {
+            Self::T0
+        } else {
+            Self::Genesis
+        }
     }
 }
 


crates/chainspec/src/hardfork.rs
```diff
@@ -42,18 +42,38 @@ hardfork!(
         /// The current Tempo hardfork (genesis).
         #[default]
         Genesis,
+        /// T0 hardfork - adds security fixes:
+        /// - Prevents access keys from performing EIP-7702 delegation
+        /// - Ensures gas limit covers 2D nonce cost before AA execution
+        T0,
```
0xKitsune: Can we remove the `Genesis` default hardfork and replace it with T0?

legion2002: we can't remove genesis, because of testnet compatiblity

mattsse (approved):

mattsse (approved):

PR #2003, opened 2026-01-12
chore(revm): enforce checked arithmetic 

legion2002:

PR #2007, opened 2026-01-13
chore: reapply PR 1987 fixes, with hardfork gating and checked math

legion2002:
This is a stacked PR with the revm-checked-math, but reapplies PR #1987 on it.

### DIFF
--- a/crates/revm/src/handler.rs
+++ b/crates/revm/src/handler.rs
@@ -500,18 +500,7 @@ where
         );
 
         // Check if this is an AA transaction by checking for tempo_tx_env
-        let evm_ctx = evm.ctx();
-        if let Some(tempo_tx_env) = evm_ctx.tx().tempo_tx_env.as_ref() {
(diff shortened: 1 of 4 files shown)


0xrusowsky (commented):
while the gas-related safety comment are true in principle, they kinda defeat the goal of having total piece of mind and preventing footguns.

the PR ends up not adding any checked math

crates/revm/src/handler.rs
```diff
@@ -395,6 +431,9 @@ where
         let mut result =
             final_result.ok_or_else(|| EVMError::Custom("No calls executed".into()))?;
 
+        // SAFETY: gas_limit >= remaining_gas because remaining_gas is derived from
+        // gas_limit and only decremented by gas spent during execution
```
0xKitsune, line 435: I think we can remove these, this is the same comment in each place. If useful we could add comments why each unchecked operation is fine but @mattsse checked this in #2003, I think this should be fine to remove these comments.

PR #2008, opened 2026-01-13
fix! : apply 1987 fixes with T0 hardfork gating

legion2002:
Applies fixes from PR #1987 hardfork gated behind the T0 hardfork, so that snapshot execution passes.

Also hardfork gates the other breaking fix in PR #1990 

Note: T0 should only be applied to the testnet at the end of the week.

Closes CHAIN-446

### DIFF
--- a/xtask/src/genesis_args.rs
+++ b/xtask/src/genesis_args.rs
@@ -478,6 +478,9 @@ impl GenesisArgs {
         chain_config
             .extra_fields
             .insert_value("epochLength".to_string(), self.epoch_length)?;
+        chain_config
+            .extra_fields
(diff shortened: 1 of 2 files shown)


0xKitsune (approved):

xtask/src/genesis_args.rs
```diff
@@ -478,6 +478,9 @@ impl GenesisArgs {
         chain_config
             .extra_fields
             .insert_value("epochLength".to_string(), self.epoch_length)?;
+        chain_config
```
legion2002, line 481: @Zygimantass please verify if this changed to the genesis script is correct. 
Had to do it, to make the CI pass.

PR #2009, opened 2026-01-13
feat(txpool): add configurable limit for AA authorization list size

mattsse:
Closes CHAIN-454

## Summary

Adds a configurable limit for the number of authorizations allowed in AA transactions. The validation runs early in the transaction pool validation pipeline, before signature recovery operations, to efficiently reject transactions that exceed the limit.

**Changes:**
- Added `TooManyAuthorizations` error variant to `TempoPoolTransactionError`
- Added `DEFAULT_MAX_TEMPO_AUTHORIZATIONS` constant (default: 16)
- Added `--txpool.max-tempo-authorizations` CLI flag for configuration
- Added `ensure_authorization_list_size()` check in validator before signature recovery
- Added tests for authorization limit validation

### DIFF
--- a/crates/transaction-pool/src/validator.rs
+++ b/crates/transaction-pool/src/validator.rs
@@ -28,6 +28,9 @@ use tempo_revm::{TempoBatchCallEnv, TempoStateAccess, calculate_aa_batch_intrins
 // Reject AA txs where `valid_before` is too close to current time (or already expired) to prevent block invalidation.
 const AA_VALID_BEFORE_MIN_SECS: u64 = 3;
 
+/// Default maximum number of authorizations allowed in an AA transaction's authorization list.
+pub const DEFAULT_MAX_TEMPO_AUTHORIZATIONS: usize = 16;
+
 /// Validator for Tempo transactions.
 #[derive(Debug)]
 pub struct TempoTransactionValidator<Client> {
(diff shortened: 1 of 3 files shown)


crates/transaction-pool/src/validator.rs
```diff
@@ -1154,4 +1198,122 @@ mod tests {
...
+        authorization_count: usize,
+    ) -> TempoPooledTransaction {
+        use alloy_eips::eip7702::Authorization;
+        use alloy_primitives::{Signature, TxKind, address};
+        use tempo_primitives::transaction::{
+            TempoSignedAuthorization, TempoTransaction,
+            tempo_transaction::Call,
+            tt_signature::{PrimitiveSignature, TempoSignature},
+            tt_signed::AASigned,
+        };
```
0xKitsune: nit: consider moving these imports to the top of the tests mod.

crates/transaction-pool/src/validator.rs
```diff
@@ -1154,4 +1198,122 @@ mod tests {
...
+            .await;
+
+        // Should not fail with TooManyAuthorizations (may fail for other reasons)
+        if let TransactionValidationOutcome::Invalid(_, err) = &outcome {
+            let error_msg = err.to_string();
+            assert!(
+                !error_msg.contains("Too many authorizations"),
+                "Should not fail with TooManyAuthorizations at the limit, got: {error_msg}"
+            );
+        }
```
0xKitsune: If we fail for any other error it looks like this assertion will still pass. We should update to use `expect()` or similar to ensure that `validate_transaction` does not fail.

0xKitsune (approved):

PR #2017, opened 2026-01-13
fix(payload-builder): align timestamp validation with handler

0xrusowsky:
This PR aligns the `validate_before` check in the payload builder with the revm handler

Closes CHAIN-457


legion2002 (approved):
lgtm

PR #2025, opened 2026-01-13
feat: FeeAMM & FeeManager invariant testing

zerosnacks:
Uses a single unified invariant entrypoint akin to the StablecoinDEX invariant suite

Proposing this would replace: https://github.com/tempoxyz/tempo/pull/1935 and https://github.com/tempoxyz/tempo/pull/2006

I've made sure it covers all invariants previously specified in other drafts but please review and test

### DIFF
--- a/crates/primitives/src/transaction/tt_signature.rs
+++ b/crates/primitives/src/transaction/tt_signature.rs
@@ -1508,4 +1508,47 @@ mod tests {
             "key_id should return access key address"
         );
     }
+
+    #[test]
+    fn test_webauthn_rejects_challenge_injection() {
+        let (tx_hash, attack_hash) = (B256::from([0xAA; 32]), B256::from([0xFF; 32]));
+        let (challenge, attack_challenge) = (
+            URL_SAFE_NO_PAD.encode(tx_hash.as_slice()),
+            URL_SAFE_NO_PAD.encode(attack_hash.as_slice()),
+        );
+
+        // Ensure that the happy path works
+        let valid_payload = format!(r#"{{"type":"webauthn.get","challenge":"{challenge}"}}"#);
+
+        let mut auth_data = vec![0u8; 37];
+        auth_data[32] = 0x01;
+        let mut webauthn_data = auth_data;
+        webauthn_data.extend_from_slice(valid_payload.as_bytes());
+
+        let result = verify_webauthn_data_internal(&webauthn_data, &tx_hash);
+        assert!(result.is_ok());
+
+        // Ensure that malicious payloads cannot pass validation
+        let attack_variants = [
+            format!(
+                r#"{{"type":"webauthn.get","challenge":"{attack_challenge}","extra":{{"challenge":"{challenge}"}}}}"#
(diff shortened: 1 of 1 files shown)


docs/specs/test/invariants/FeeAMM.t.sol
```diff
@@ -0,0 +1,2150 @@
...
+        // Skip if tokens are identical
+        vm.assume(userToken != validatorToken);
+
+        // Skip if actor is blacklisted for validatorToken (can't mint funds to them)
+        uint64 policyId =
+            validatorToken == address(pathUSD) ? _pathUsdPolicyId : _tokenPolicyIds[validatorToken];
+        vm.assume(registry.isAuthorized(policyId, actor));
+
+        // Bound amount to reasonable range (must be > 2 * MIN_LIQUIDITY for first mint)
+        amount = bound(amount, MIN_LIQUIDITY * 3, 10_000_000_000);
```
legion2002: we shoudln't bound this amount here, instead it should check failure for MIN_LIQUIDITY errors

docs/specs/test/invariants/FeeAMM.t.sol
```diff
@@ -0,0 +1,2150 @@
...
+                pool.reserveValidatorToken > 0,
+                "TEMPO-AMM30: Initialized pool has zero validator reserve"
+            );
+        }
+    }
+
+    /// @notice TEMPO-AMM24: All participants can exit - verify everyone can withdraw
+    /// @dev After all operations, all LPs should be able to burn their positions and
+    ///      all validators should be able to claim their fees. Only dust should remain.
+    function _verifyAllCanExit() internal {
```
legion2002, line 1411: This seems really powerful!

docs/specs/test/invariants/FeeAMM.t.sol
```diff
@@ -0,0 +1,2150 @@
...
+        bytes32 poolSlot = keccak256(abi.encode(poolId, poolsSlot));
+
+        // Pack: lower 128 bits = reserveUserToken, upper 128 bits = reserveValidatorToken
+        bytes32 newPoolValue = bytes32(uint256(reserveUser) | (uint256(reserveValidator) << 128));
+        vm.store(address(amm), poolSlot, newPoolValue);
+    }
+
+    /// @dev Stores/increments collected fees using vm.store
+    function _storeCollectedFees(address validator, address token, uint256 amount) internal {
+        // Storage layout differs between Rust and Solidity implementations:
```
legion2002, line 964: hmm, maybe FeeAMM should not inherit from FeeManager. 
But this shoudln't block this PR, we can do it separately

docs/specs/test/invariants/FeeAMM.t.sol
zerosnacks: good call, fixed in https://github.com/tempoxyz/tempo/pull/2025/commits/3262987440750c26da498a4d527bcc34ff76d3dc in a way we don't waste budget post initial deposit but still catch err

PR #2028, opened 2026-01-13
test(transaction-pool): add unit tests for access key validation

fgimenez:
Ref CHAIN-421

Adds unit tests for access key validation in the transaction pool validator.

- Add `encode_to_slot()` method to `AuthorizedKey` for setting up keychain storage state in tests
- Add comprehensive tests for keychain signature validation including:
  - Valid access key acceptance
  - Expired key rejection
  - Revoked key rejection
  - Non-existent key rejection
  - Key authorization signature validation

### DIFF
--- a/crates/transaction-pool/src/validator.rs
+++ b/crates/transaction-pool/src/validator.rs
@@ -1351,9 +1351,10 @@ mod tests {
             let state_provider = validator.inner.client().latest().unwrap();
 
             let result = validator.validate_against_keychain(&transaction, &state_provider);
-            assert!(
-                matches!(result, Ok(Err(msg)) if msg.contains("revoked")),
-                "Revoked key should be rejected, got: {result:?}"
+            assert_eq!(
+                result.expect("should not be a provider error"),
+                Err("access key has been revoked"),
+                "Revoked key should be rejected"
             );
         }
 
@@ -1383,9 +1384,10 @@ mod tests {
             let state_provider = validator.inner.client().latest().unwrap();
 
             let result = validator.validate_against_keychain(&transaction, &state_provider);
-            assert!(
-                matches!(result, Ok(Err(msg)) if msg.contains("does not exist")),
-                "Non-existent key should be rejected, got: {result:?}"
+            assert_eq!(
+                result.expect("should not be a provider error"),
+                Err("access key does not exist"),
+                "Non-existent key should be rejected"
             );
         }
 
@@ -1407,9 +1409,10 @@ mod tests {
             let state_provider = validator.inner.client().latest().unwrap();
 
             let result = validator.validate_against_keychain(&transaction, &state_provider);
-            assert!(
-                matches!(result, Ok(Err(msg)) if msg.contains("does not exist")),
-                "Missing storage should result in non-existent key error, got: {result:?}"
+            assert_eq!(
+                result.expect("should not be a provider error"),
+                Err("access key does not exist"),
+                "Missing storage should result in non-existent key error"
             );
         }
 
@@ -1493,9 +1496,10 @@ mod tests {
             let state_provider = validator.inner.client().latest().unwrap();
 
             let result = validator.validate_against_keychain(&transaction, &state_provider);
-            assert!(
-                matches!(result, Ok(Err(msg)) if msg.contains("chain_id")),
-                "Wrong chain_id should be rejected, got: {result:?}"
+            assert_eq!(
+                result.expect("should not be a provider error"),
+                Err("KeyAuthorization chain_id does not match current chain"),
+                "Wrong chain_id should be rejected"
             );
         }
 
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -50,6 +50,47 @@ impl AuthorizedKey {
...
             .expect("unable to decode AuthorizedKey from slot")
     }
+
+    /// Encode AuthorizedKey to a storage slot value
+    ///
+    /// This is useful for tests that need to set up storage state directly.
+    pub fn encode_to_slot(&self) -> U256 {
+        use __packing_authorized_key::{
+            ENFORCE_LIMITS_LOC, EXPIRY_LOC, IS_REVOKED_LOC, SIGNATURE_TYPE_LOC,
+        };
```
0xKitsune: nit: can we move imports to the top of the file?

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -50,6 +50,47 @@ impl AuthorizedKey {
...
+
+        let encoded = insert_into_word(
+            U256::ZERO,
+            &self.signature_type,
+            SIGNATURE_TYPE_LOC.offset_bytes,
+            SIGNATURE_TYPE_LOC.size,
+        )
+        .expect("unable to insert 'signature_type'");
```
0xKitsune, line 67: @0xrusowsky confirming that these expects are ok?

crates/transaction-pool/src/validator.rs
```diff
@@ -1154,4 +1154,556 @@ mod tests {
...
+            // Non-AA transaction should return Ok(Ok(())) immediately
+            let transaction = get_transaction(None);
+            let validator = setup_validator(&transaction, 0);
+            let state_provider = validator.inner.client().latest().unwrap();
+
+            let result = validator.validate_against_keychain(&transaction, &state_provider);
+            assert!(
+                matches!(result, Ok(Ok(()))),
+                "Non-AA tx should skip keychain validation"
+            );
```
0xKitsune: nit: consider simplifying assertions throughout these tests to the following:
```suggestion
            let result = validator.validate_against_keychain(&transaction, &state_provider)?;
            assert!(
                result.is_ok(),
                "Non-AA tx should skip keychain validation"
            );
```

crates/transaction-pool/src/validator.rs
```diff
@@ -1154,4 +1154,556 @@ mod tests {
...
+                &transaction,
+                user_address,
+                access_key_address,
+                Some(slot_value),
+            );
+            let state_provider = validator.inner.client().latest().unwrap();
+
+            let result = validator.validate_against_keychain(&transaction, &state_provider);
+            assert!(
+                matches!(result, Ok(Err(msg)) if msg.contains("revoked")),
```
0xKitsune: Can we assert the exact error rather than `msg.contains()`?

crates/transaction-pool/src/validator.rs
```diff
@@ -1154,4 +1154,556 @@ mod tests {
...
+                &transaction,
+                user_address,
+                access_key_address,
+                Some(slot_value),
+            );
+            let state_provider = validator.inner.client().latest().unwrap();
+
+            let result = validator.validate_against_keychain(&transaction, &state_provider);
+            assert!(
+                matches!(result, Ok(Err(msg)) if msg.contains("does not exist")),
```
0xKitsune: Same thing here can we assert the exact error?

crates/transaction-pool/src/validator.rs
```diff
@@ -1154,4 +1154,556 @@ mod tests {
...
+                &transaction,
+                user_address,
+                Address::ZERO,
+                None,
+            );
+            let state_provider = validator.inner.client().latest().unwrap();
+
+            let result = validator.validate_against_keychain(&transaction, &state_provider);
+            assert!(
+                matches!(result, Ok(Err(msg)) if msg.contains("does not exist")),
```
0xKitsune: Can we update all error assertions to expect the exact error?

crates/precompiles/src/account_keychain/mod.rs
fgimenez: sure done

crates/transaction-pool/src/validator.rs
fgimenez: yep done

crates/transaction-pool/src/validator.rs
fgimenez: good point done

crates/precompiles/src/account_keychain/mod.rs
0xrusowsky: yes, this can only fail if `offset_bytes + size > 32`, so in this case it is infallible.

additionally he added a test for it

PR #2030, opened 2026-01-13
test(tempo-revm): add property-based tests for tx, block, and handler modules

fgimenez:
Ref CHAIN-402

Adds property-based tests (proptest) to the `tempo-revm` crate for improved test coverage of core transaction and block environment logic.

- tx.rs Tests for `max_balance_spending`, `effective_balance_spending`, `first_call`, and `calls` methods
- block.rs: Tests for `timestamp_millis` including saturation, monotonicity, and sub-second precision
- handler.rs: Tests for gas calculation functions (`primitive_signature_verification_gas`, `tempo_signature_verification_gas`, `calculate_key_authorization_gas`, `calculate_2d_nonce_gas`)


crates/revm/src/block.rs
```diff
@@ -74,3 +74,101 @@ impl BlockEnvironment for TempoBlockEnv {
...
+            timestamp_millis_part: millis_part,
+        }
+    }
+
+    /// Strategy for random U256 values.
+    fn arb_u256() -> impl Strategy<Value = U256> {
+        any::<[u64; 4]>().prop_map(U256::from_limbs)
+    }
+
+    proptest! {
```
rakita, line 99: All of these tests test millis timestamps. Proptest is really nice!

crates/revm/src/handler.rs
```diff
@@ -2307,4 +2311,172 @@ mod tests {
...
     }
+
+    /// Strategy for optional u64 timestamps.
+    fn arb_opt_timestamp() -> impl Strategy<Value = Option<u64>> {
+        prop_oneof![Just(None), any::<u64>().prop_map(Some)]
+    }
+
+    /// Helper to create a secp256k1 signature for testing.
+    fn secp256k1_sig() -> TempoSignature {
+        TempoSignature::Primitive(PrimitiveSignature::Secp256k1(
```
rakita, line 2387: Should we test both valid and invalid real signatures? Not sure how would this fit inside tests, fine to skip it from prop tests if it does not fit.

fgimenez: good point, aiui for gas calculation testing we need to take into account the signature types, not its validity, which is checked in `recover_signer` is this correct? added a comment about it in any case to make it clearer

crates/revm/src/handler.rs
```diff
@@ -2307,4 +2311,177 @@ mod tests {
...
+
+    /// Helper to compute AA batch gas with no access list.
+    fn compute_aa_gas(env: &TempoBatchCallEnv) -> InitialAndFloorGas {
+        calculate_aa_batch_intrinsic_gas(env, None::<std::iter::Empty<&AccessListItem>>).unwrap()
+    }
+
+    proptest! {
+        #![proptest_config(ProptestConfig::with_cases(500))]
+
+        /// Property: validate_time_window returns Ok iff (after <= ts < before)
```
0xKitsune:
```suggestion
        /// Property: validate_time_window returns Ok if (after <= ts < before)
```

crates/revm/src/handler.rs
```diff
@@ -2307,4 +2311,177 @@ mod tests {
...
+
+        /// Property: validate_time_window with valid_after=0 is equivalent to None
+        #[test]
+        fn proptest_validate_time_window_zero_after_equivalent_to_none(
+            valid_before in arb_opt_timestamp(),
+            block_timestamp in any::<u64>(),
+        ) {
+            let with_zero = validate_time_window(Some(0), valid_before, block_timestamp);
+            let with_none = validate_time_window(None, valid_before, block_timestamp);
+            prop_assert_eq!(with_zero.is_ok(), with_none.is_ok());
```
0xKitsune, line 2470: Do we need to ensure that `valid_before` is < `block_timestamp` in the assertions?

crates/revm/src/handler.rs
```diff
@@ -2307,4 +2311,177 @@ mod tests {
...
+        }
+
+        /// Property: validate_time_window - if before <= after, the window is empty
+        #[test]
+        fn proptest_validate_time_window_empty_window(
+            valid_after in 1u64..=u64::MAX,
+            offset in 0u64..1000u64,
+        ) {
+            let valid_before = valid_after.saturating_sub(offset);
+            let ts_in_range = valid_after;
```
0xKitsune: minor nit: we dont need to bind to a variable here

crates/revm/src/handler.rs
```diff
@@ -2307,4 +2311,177 @@ mod tests {
...
+                    num_calls1, gas1.initial_gas, num_calls2, gas2.initial_gas);
+            }
+        }
+
+        /// Property: base gas is always at least 21000
+        #[test]
+        fn proptest_gas_minimum_base_stipend(num_calls in 1usize..5) {
+            let gas = compute_aa_gas(&make_multi_call_env(num_calls));
+            prop_assert!(gas.initial_gas >= 21_000, "Base gas stipend should be at least 21k");
+        }
```
0xKitsune: Should we have any assertions around the max amount we expect base gas to be?

crates/revm/src/tx.rs
```diff
@@ -732,4 +734,134 @@ mod tests {
...
+            max_fee_per_gas in any::<u128>(),
+            value in arb_u256(),
+        ) {
+            let tx_env = make_tx_env(gas_limit, max_fee_per_gas, value);
+            let result = tx_env.effective_balance_spending(0, 0);
+            prop_assert!(
+                result.is_ok()
+                    || result == Err(InvalidTransaction::OverflowPaymentInTransaction)
+            );
+        }
```
0xKitsune: This test is only checking if there is an error, rather than asserting the actual expected value.

crates/revm/src/tx.rs
```diff
@@ -732,4 +734,134 @@ mod tests {
...
+        /// Property: calls() returns exactly 1 for non-AA transactions
+        #[test]
+        fn proptest_calls_count_non_aa_tx(
+            gas_limit in any::<u64>(),
+            value in arb_u256(),
+        ) {
+            let non_aa_tx = make_tx_env(gas_limit, 0, value);
+            prop_assert_eq!(non_aa_tx.calls().count(), 1);
+        }
+    }
```
0xKitsune: Im not sure we need a prop test for this, a unit test should suffice.

crates/revm/src/handler.rs
```diff
@@ -2307,4 +2311,177 @@ mod tests {
...
+
+        /// Property: gas calculation monotonicity - more calldata means more gas
+        #[test]
+        fn proptest_gas_monotonicity_calldata(
+            calldata_len1 in 0usize..1000,
+            calldata_len2 in 0usize..1000,
+        ) {
+            let gas1 = compute_aa_gas(&make_single_call_env(Bytes::from(vec![1u8; calldata_len1])));
+            let gas2 = compute_aa_gas(&make_single_call_env(Bytes::from(vec![1u8; calldata_len2])));
```
0xKitsune, line 2512: This uses non-zero bytes but we should consider also testing with `vec![0u8; len]` since zero bytes cost different gas (4 vs 16). Also we should mix zero and non-zero calldata.

> Property: gas calculation monotonicity - more calldata means more gas

The property may not hold if mixing zero/non-zero calldata.

crates/revm/src/block.rs
```diff
@@ -74,3 +74,101 @@ impl BlockEnvironment for TempoBlockEnv {
...
+            prop_assert_eq!(block.timestamp_millis(), expected);
+        }
+
+        /// Property: timestamp_millis is monotonic in both inputs
+        #[test]
+        fn proptest_timestamp_millis_monotonicity(
+            ts1 in 0u64..u64::MAX / 1000,
+            ts2 in 0u64..u64::MAX / 1000,
+            mp1 in 0u64..1000,
+            mp2 in 0u64..1000,
```
0xKitsune, line 144: This restricts millis_part to `0..1000` but the actual field is `u64`. What happens when `millis_part >= 1000`? Consider testing overflow behavior.

crates/revm/src/handler.rs
```diff
@@ -2307,4 +2311,177 @@ mod tests {
...
+                    "Fewer calls should mean less gas: calls1={}, gas1={}, calls2={}, gas2={}",
+                    num_calls1, gas1.initial_gas, num_calls2, gas2.initial_gas);
+            }
+        }
+
+        /// Property: base gas is always at least 21000
+        #[test]
+        fn proptest_gas_minimum_base_stipend(num_calls in 1usize..5) {
+            let gas = compute_aa_gas(&make_multi_call_env(num_calls));
+            prop_assert!(gas.initial_gas >= 21_000, "Base gas stipend should be at least 21k");
```
0xKitsune: AA transactions have different base gas. Is 21k actually the right floor for AA batch calls?

0xKitsune (commented):
Left a few comments, also looks like there is no boundary testing for `first_call()` but PR mentions it.

Same thing for `calculate_key_authorization_gas` and `calculate_2d_nonce_gas`.

crates/revm/src/handler.rs
fgimenez: i don't think so, the property being tested is that Some(0) and None are equivalent for valid_after, they should produce the same result in all cases, regardless of wether valid_before causes success of failure. clarified this in the doc comment

crates/revm/src/handler.rs
fgimenez: good catch, removed

crates/revm/src/handler.rs
fgimenez: sure, added

crates/revm/src/tx.rs
fgimenez: done, it asserts expected value now

crates/revm/src/tx.rs
fgimenez: makes sense, done

crates/revm/src/handler.rs
fgimenez: good call, added add calldata gas proptests for zero, nonzero, and mixed bytes proptests

crates/revm/src/block.rs
fgimenez: sure done

crates/revm/src/handler.rs
fgimenez: good to raise this, aiui aa txs use the same base 21k for AA batch calls, plus signature verification (0 for Secp256k1 as in the test) plus `COLD_ACCOUNT_ACCESS_COST * (num_calls - 1)` for batch calls. i've updated the name and the doc to make this clear, and updated the test to check exact expected value instead of bounds ptal

PR #2033, opened 2026-01-13
feat(tips): TIP-1006: burnAt for TIP-20 tokens

danrobinson:
## Summary

- Adds TIP-1006 specification for `burnAt` function on TIP-20 tokens
- New `BURN_AT_ROLE` allows burning tokens from any address
- Unlike `burnBlocked`, does not require the target address to be policy-blocked
- Maintains protected address check (fee manager, DEX)


docs/pages/protocol/tips/tip-1006.mdx
```diff
@@ -0,0 +1,113 @@
+# Burn At for TIP-20 Tokens
```
jenpaff, line 1: @danrobinson filename should be in lowercase to be consistent with the rest, can push the fix on the branch

PR #2035, opened 2026-01-13
fix(txpool): evict AA transactions when signing key is revoked

fgimenez:
Closes CHAIN-439

Evict AA transactions from the mempool when their authorizing keychain key is revoked.

Transactions signed with an access key that is later revoked would remain stuck in the mempool indefinitely—never evicted, never executable. This created a DoS vector where attackers could clog the mempool with transactions using keys they planned to revoke.

This PR consolidates the existing maintenance tasks and adds logic to evict transactions with revoked signing key

/cc @mattsse @klkvr ptal

### DIFF
--- a/crates/transaction-pool/src/maintain.rs
+++ b/crates/transaction-pool/src/maintain.rs
@@ -130,24 +130,25 @@ where
                 // 2. Evict transactions with revoked keychain keys
                 // Key revocations are rare, so we scan the pool on-demand rather than
                 // tracking all keychain-signed transactions.
-                for receipts in tip.execution_outcome().receipts().iter() {
-                    for receipt in receipts {
-                        for log in &receipt.logs {
-                            if log.address != ACCOUNT_KEYCHAIN_ADDRESS {
-                                continue;
-                            }
+                // Collect all revocations first, then scan the pool once per block.
+                let revoked_keys: Vec<_> = tip
+                    .execution_outcome()
+                    .receipts()
+                    .iter()
+                    .flatten()
+                    .flat_map(|receipt| &receipt.logs)
+                    .filter(|log| log.address == ACCOUNT_KEYCHAIN_ADDRESS)
+                    .filter_map(|log| IAccountKeychain::KeyRevoked::decode_log(log).ok())
+                    .map(|event| (event.account, event.publicKey))
+                    .collect();
 
-                            if let Ok(event) = IAccountKeychain::KeyRevoked::decode_log(log) {
-                                debug!(
-                                    target: "txpool",
-                                    account = %event.account,
-                                    key_id = %event.publicKey,
-                                    "Processing keychain key revocation"
-                                );
-                                pool.evict_transactions_by_revoked_key(event.account, event.publicKey);
-                            }
-                        }
-                    }
+                if !revoked_keys.is_empty() {
+                    debug!(
+                        target: "txpool",
+                        count = revoked_keys.len(),
+                        "Processing keychain key revocations"
+                    );
+                    pool.evict_transactions_by_revoked_keys(&revoked_keys);
                 }
 
                 // 3. Update 2D nonce pool
--- a/crates/transaction-pool/src/tempo_pool.rs
+++ b/crates/transaction-pool/src/tempo_pool.rs
@@ -81,17 +81,17 @@ where
             .notify_on_transaction_updates(promoted, Vec::new());
     }
 
-    /// Evicts all transactions signed with a revoked keychain key.
+    /// Evicts all transactions signed with revoked keychain keys.
     ///
-    /// This scans both pools for AA transactions using the specified keychain key
-    /// and removes them. Called when a `KeyRevoked` event is detected on-chain.
+    /// This scans both pools for AA transactions using any of the specified keychain keys
+    /// and removes them. Called once per block when `KeyRevoked` events are detected.
     ///
     /// Key revocations are expected to be rare, so this on-demand scan is more
     /// efficient than tracking all keychain-signed transactions in the pool.
-    pub(crate) fn evict_transactions_by_revoked_key(&self, account: Address, key_id: Address) {
+    pub(crate) fn evict_transactions_by_revoked_keys(&self, revoked_keys: &[(Address, Address)]) {
         let mut to_remove = Vec::new();
 
-        // Scan all transactions in both pools
+        // Scan all transactions in both pools once
         for tx in self.all_transactions().all() {
             let Some(aa_tx) = tx.inner().as_aa() else {
                 continue;
@@ -101,12 +101,15 @@ where
                 continue;
             };
 
-            // Check if this transaction uses the revoked key
-            if keychain_sig.user_address == account
-                && let Ok(tx_key_id) = keychain_sig.key_id(&aa_tx.signature_hash())
-                && tx_key_id == key_id
-            {
-                to_remove.push(*aa_tx.hash());
+            // Check if this transaction uses any of the revoked keys
+            for &(account, key_id) in revoked_keys {
+                if keychain_sig.user_address == account
+                    && let Ok(tx_key_id) = keychain_sig.key_id(&aa_tx.signature_hash())
+                    && tx_key_id == key_id
+                {
+                    to_remove.push(*aa_tx.hash());
+                    break;
+                }
             }
         }
 


crates/transaction-pool/src/maintain.rs
```diff
@@ -138,3 +142,159 @@ where
...
+/// The task:
+/// - Tracks keychain-signed transactions by their key slot (user_address, key_id)
+/// - Monitors `ACCOUNT_KEYCHAIN_ADDRESS` storage changes for revocations
+/// - Evicts affected transactions when `is_revoked` becomes true
+/// - Periodically checks all tracked keys' current revocation status (handles missed events)
+pub async fn evict_revoked_keychain_txs<P, C>(pool: P, client: C)
+where
+    P: TransactionPool<Transaction = TempoPooledTransaction> + 'static,
+    C: CanonStateSubscriptions<Primitives = TempoPrimitives> + StateProviderFactory + 'static,
+{
```
mattsse: I think we should convert this to a general purpose maintain task, because it doesnt seem optimal to have multiple tasks that then compete for updates, and we could also batch different update kinds, so we can do `pool.write().update()` once

crates/transaction-pool/src/maintain.rs
```diff
@@ -138,3 +142,159 @@ where
...
+    // Track keychain-signed transactions by their key slot
+    // key_slot (U256) -> Set of tx hashes using that key
+    let mut keychain_txs: HashMap<U256, HashSet<TxHash>> = HashMap::new();
+    // Reverse mapping: tx_hash -> key_slot (for cleanup when tx is removed)
+    let mut tx_to_slot: HashMap<TxHash, U256> = HashMap::new();
+
+    // Helper to compute the storage slot for a key in the AccountKeychain
+    let compute_key_slot = |user_address: Address, key_id: Address| -> U256 {
+        AccountKeychain::new().keys[user_address][key_id].base_slot()
+    };
```
mattsse: dont we have a cache for this already somewhere?

crates/transaction-pool/src/maintain.rs
```diff
@@ -138,3 +142,159 @@ where
...
+                            if authorized_key.is_revoked {
+                                to_remove.extend(tx_hashes.iter().copied());
+                            }
+                        }
+                    }
+                }
+
+                // Second, for any remaining tracked slots, check current state to catch
+                // revocations that happened before we subscribed to events
+                if to_remove.is_empty() && !keychain_txs.is_empty() {
```
mattsse: this could maybe be expensive

it might be easier to scan for:

https://github.com/tempoxyz/tempo/blob/3d2ce678fc6b31cd4abb0d4e221b143d9462d9c5/crates/precompiles/src/account_keychain/mod.rs#L183-L188

then we dont need to deal with the slot tracking I believe

crates/transaction-pool/src/maintain.rs
fgimenez: makes sense, unified all of them on a single `maintain_tempo_pool` task ptal

crates/transaction-pool/src/maintain.rs
fgimenez: afaik there is no cache for this, there are two keccak calls, which should be ~cheap. i've added a `AccountKeychain::key_slot` that avoids current duplication, lmk if we should add cache here

crates/transaction-pool/src/maintain.rs
fgimenez: much better, done

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -38,6 +42,19 @@ pub struct TempoTransactionPool<Client> {
...
+    aa_2d_last_processed_tip: Arc<AtomicU64>,
+    /// Notifies waiters when 2D pool maintenance task processes a new tip.
+    #[cfg(feature = "test-utils")]
+    aa_2d_processed_notify: Arc<Notify>,
+    /// Tracks the last block tip number processed by the keychain eviction maintenance task.
+    #[cfg(feature = "test-utils")]
+    keychain_last_processed_tip: Arc<AtomicU64>,
+    /// Notifies waiters when keychain eviction task processes a new tip.
+    #[cfg(feature = "test-utils")]
+    keychain_processed_notify: Arc<Notify>,
```
0xKitsune: The fields marked with `test-utils` here are a bit noisy and also affects `new()` and `clone()`. Consider consolidating into a single `Arc<TestSyncState>` that holds all the atomics and notifies, reducing it to one cfg attribute per location. 

IMO we should consider removing this entirely, what is the reason for having this, is this needed?

crates/transaction-pool/src/maintain.rs
```diff
@@ -1,28 +1,49 @@
...
+/// This task consolidates multiple pool maintenance operations into a single event loop
+/// to avoid multiple tasks competing for the same canonical state updates:
+///
+/// - **Expired AA transactions**: Evicts transactions when `valid_before <= tip_timestamp`
+/// - **2D nonce pool**: Updates the AA 2D nonce pool based on `NonceManager` state changes
+/// - **AMM liquidity cache**: Updates the AMM liquidity cache from `FeeManager` state changes
+/// - **Revoked keychain keys**: Evicts transactions signed with revoked keychain keys
+///
+/// By batching these operations, we can process all updates with a single state subscription
+/// and minimize contention on pool locks.
```
0xKitsune:
```suggestion
/// Unified maintenance task for the Tempo transaction pool.
///
/// Handles:
/// - Evicting expired AA transactions (`valid_before <= tip_timestamp`)
/// - Updating the AA 2D nonce pool from `NonceManager` changes
/// - Refreshing the AMM liquidity cache from `FeeManager` updates
/// - Removing transactions signed with revoked keychain keys
```

crates/transaction-pool/src/maintain.rs
```diff
@@ -33,107 +54,175 @@ where
...
+                    }
+                }
+
+                if !revoked_txs.is_empty() {
+                    debug!(
+                        target: "txpool",
+                        count = revoked_txs.len(),
+                        "Evicting AA transactions with revoked keychain keys"
+                    );
+                    remove_from_keychain_tracking(&mut keychain_txs, &mut tx_to_key, &revoked_txs);
```
mattsse: if we only remove on revoke, doesnt this mean that this map grows unbounded?

crates/transaction-pool/src/tempo_pool.rs
mattsse: could we perhaps group these in a helper struct?

I understand why these are useful, but perhaps a simple sleep is also sufficient

crates/transaction-pool/src/maintain.rs
```diff
@@ -1,28 +1,49 @@
...
+        + 'static,
 {
+    // Track valid_before timestamp -> Vec<TxHash>
+    let mut expiry_map: BTreeMap<u64, Vec<TxHash>> = BTreeMap::new();
+
+    // Keychain revocation tracking state:
+    // (account, key_id) -> Set of tx hashes using that key
+    let mut keychain_txs: HashMap<(Address, Address), HashSet<TxHash>> = HashMap::new();
+    // Reverse mapping: tx_hash -> (account, key_id) (for cleanup when tx is removed)
+    let mut tx_to_key: HashMap<TxHash, (Address, Address)> = HashMap::new();
```
mattsse: what if we introduce a helper State type or something for these, then we can convert the closures into regular fns which imo would make this easier to read

0xKitsune: Consider using a type alias for this since it looks like this is used in a few places.

crates/transaction-pool/src/maintain.rs
```diff
@@ -33,107 +54,175 @@ where
...
-    // Track valid_before timestamp -> Vec<TxHash>
-    let mut expiry_map: BTreeMap<u64, Vec<TxHash>> = BTreeMap::new();
+    // Helper to track a keychain-signed transaction
+    let track_keychain_tx = |keychain_txs: &mut HashMap<(Address, Address), HashSet<TxHash>>,
+                             tx_to_key: &mut HashMap<TxHash, (Address, Address)>,
+                             maybe_aa_tx: Option<&AASigned>,
+                             tx_hash: TxHash| {
+        let Some(aa_tx) = maybe_aa_tx else {
+            return;
+        };
```
0xKitsune: Just curious on the rationale why this is a closure vs it's own function?

crates/transaction-pool/src/maintain.rs
fgimenez: good catch, added cleanup for both tracking maps (`keychain_txs`/`tx_to_key` and `expiry_map`/`tx_to_expiry`) by subscribing to `all_transactions_event_listener()`. now when transactions are mined, discarded, invalidated, or replaced, we remove them from the tracking maps, lmk if this is ok

crates/transaction-pool/src/tempo_pool.rs
fgimenez: ok added `MaintenanceTestSync` struct. i tried with sleep but was very flaky, started with the more deterministic approach after fighting for quite some time with CI

crates/transaction-pool/src/maintain.rs
fgimenez: for sure, good call, extracted to a separate `MaintenanceState` type

crates/transaction-pool/src/tempo_pool.rs
fgimenez: yep already addressed see https://github.com/tempoxyz/tempo/pull/2035#discussion_r2695366773

crates/transaction-pool/src/maintain.rs
fgimenez: yep, already moved to a `MaintenanceState` method

crates/transaction-pool/src/maintain.rs
fgimenez: sure, added type alias for `(Address, Address)` (plus doc about what each element represents) here https://github.com/tempoxyz/tempo/pull/2035/commits/9b60ac50def8c38b568640c4505fa14e7a7d8746

crates/node/tests/it/tempo_transaction.rs
```diff
@@ -1438,17 +1459,13 @@ async fn test_aa_2d_nonce_pool_comprehensive() -> eyre::Result<()> {
...
-        }
-        tokio::time::sleep(std::time::Duration::from_millis(10)).await;
-    }
+    // Wait for 2D pool maintenance task to process the block and remove mined transactions
+    setup
+        .node
+        .inner
+        .pool
+        .wait_for_maintenance_processed_tip(payload4.block().inner.number, POOL_PROCESS_TIMEOUT)
+        .await?;
```
0xKitsune: Would it be sufficient to sleep here? IIUC then we could remove the fields marked with `test_utils` from the `TempoTransactionPool`?

fgimenez: nope see https://github.com/tempoxyz/tempo/pull/2035#discussion_r2695366773 it was the initial approach i tried but was very flaky

gakonst (commented):
## High-Impact Review: Unified Maintenance Task

Overall this is a solid consolidation. A few high-impact observations:

### Issue 1: Memory growth in tracking maps without bounds

```rust
struct MaintenanceState {
    expiry_map: BTreeMap<u64, Vec<TxHash>>,
    tx_to_expiry: HashMap<TxHash, u64>,
    keychain_txs: HashMap<KeychainKey, HashSet<TxHash>>,
    tx_to_key: HashMap<TxHash, KeychainKey>,
}
```

These maps grow with every new transaction and shrink when transactions are removed. However, cleanup only happens on `FullTransactionEvent::Mined|Discarded|Invalid|Replaced`.

**Question**: If a transaction is removed via `pool.remove_transactions()` (e.g., by this very maintenance task), does that emit a `FullTransactionEvent`? If not, `state.remove_tx()` is called explicitly for keychain evictions, but I don't see it called for expired transactions:

```rust
let mut to_remove = state.drain_expired(tip_timestamp);
// ... later ...
pool.remove_transactions(to_remove);  // Does this trigger cleanup?
```

The `drain_expired` cleans up expiry maps internally, but if the same tx was ALSO keychain-tracked, `tx_to_key` and `keychain_txs` won't be cleaned.

**Recommendation**: After `pool.remove_transactions(to_remove)`, call `state.remove_tx()` for each hash to ensure all tracking maps are cleaned.

### Issue 2: Startup race condition window

```rust
// Small delay to allow other tasks to initialize (skip in tests)
#[cfg(not(feature = "test-utils"))]
tokio::time::sleep(std::time::Duration::from_secs(1)).await;

// Subscribe to new transactions...
let mut new_txs = pool.new_transactions_listener();
// ...

// Populate tracking maps with existing transactions
pool.all_transactions().all().for_each(|tx| { ... });
```

There's a window between subscribing to `new_txs` and populating from `all_transactions()`. Transactions added during this window could be:
1. Received by `new_txs` listener AND
2. Returned by `all_transactions()`

This would result in duplicate tracking (same tx_hash in HashSets twice - benign due to HashSet, but wasteful).

Consider subscribing BEFORE the sleep, or accepting the minor inefficiency.

### Positive: Event-based approach is correct

Using `KeyRevoked` events instead of polling storage on every block is the right approach for scalability. Well done on the consolidation.

gakonst:
Applied @0xKitsune's doc comment suggestion in 023edbdc - simplified the  documentation to use the cleaner format.

All other review feedback from @mattsse and @0xKitsune appears to have been addressed by Federico in previous commits:
- ✅ Unified into single `maintain_tempo_pool` task
- ✅ Using `KeyRevoked` events instead of expensive slot tracking  
- ✅ Extracted `MaintenanceState` struct
- ✅ Added `KeychainKey` type alias
- ✅ Added cleanup via `all_transactions_event_listener()`
- ✅ Added `MaintenanceTestSync` struct for test-utils

Ready for re-review.

0xKitsune (approved):
Lgtm, we could consider further cleanup/refactoring in a subsequent PR. Pending @mattsse  or @klkvr

crates/transaction-pool/src/maintain.rs
```diff
@@ -1,79 +1,203 @@
...
+        + ChainSpecProvider<ChainSpec = TempoChainSpec>
+        + CanonStateSubscriptions<Primitives = TempoPrimitives>
+        + 'static,
+{
+    let mut state = TempoPoolState::default();
+
+    // Subscribe to new transactions, transaction events, and chain events
     let mut new_txs = pool.new_transactions_listener();
-    let mut chain_events = client.canonical_state_stream();
+    let mut tx_events = pool.all_transactions_event_listener();
```
mattsse: this channel is kinda super inefficient, why do we need this so fine granular, this will effecitvely add additional overhead per tx insertion and state change, which I dont think we really need here

crates/transaction-pool/src/maintain.rs
```diff
@@ -82,58 +206,54 @@ where
...
+                            }
+
+                            if let Ok(event) = IAccountKeychain::KeyRevoked::decode_log(log)
+                                && let Some(tx_hashes) = state.txs_for_key(event.account, event.publicKey)
+                            {
+                                revoked_txs.extend(tx_hashes.iter().copied());
+                            }
+                        }
+                    }
+                }
```
mattsse: how many key revoke do we expect?

my feeling is like we dont have too many so we don't need t cross ref them all the time with what we have in the pool and instead ntifiy the pool about every revoked key instead, over all this should be a lot more efficient then monitoring all aa txs

crates/transaction-pool/src/maintain.rs
fgimenez: makes sense, using use lazy cleanup instead of tx event listener now

crates/transaction-pool/src/maintain.rs
fgimenez: indeed, done ptal

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -81,6 +81,40 @@ where
...
+
+        // Scan all transactions in both pools
+        for tx in self.all_transactions().all() {
+            let Some(aa_tx) = tx.inner().as_aa() else {
+                continue;
+            };
+
+            let Some(keychain_sig) = aa_tx.signature().as_keychain() else {
+                continue;
+            };
```
mattsse: we could potentially optimize this further but imo seems fine to scan this once per block

crates/transaction-pool/src/maintain.rs
```diff
@@ -206,48 +124,36 @@ where
...
-                            {
-                                revoked_txs.extend(tx_hashes.iter().copied());
+                            if let Ok(event) = IAccountKeychain::KeyRevoked::decode_log(log) {
+                                debug!(
+                                    target: "txpool",
+                                    account = %event.account,
+                                    key_id = %event.publicKey,
+                                    "Processing keychain key revocation"
+                                );
+                                pool.evict_transactions_by_revoked_key(event.account, event.publicKey);
```
mattsse: this now runs per keyrevoked event, I think we could batch this so that we run this once per block update

0xKitsune (commented):
Review submitted

crates/transaction-pool/src/maintain.rs
fgimenez: good call, done here https://github.com/tempoxyz/tempo/pull/2035/commits/2dfac7db710bdea063e659b0e7dad560339f0bd1

PR #2041, opened 2026-01-14
fix(txpool): reject paused TIP20 tokens in fee validation to prevent DoS

fgimenez:
Closes CHAIN-440
Ref CHAIN-441

Reject paused TIP20 tokens during fee token validation to prevent mempool DoS attacks.

Added a `paused` check to `is_valid_fee_token()` in the revm layer. Paused tokens are now rejected before reaching the AMM cache lookup, closing the bypass path.


crates/revm/src/common.rs
```diff
@@ -609,4 +621,36 @@ mod tests {
...
+        // paused=false is the default (zero), no need to set
+
+        let is_paused = db.is_fee_token_paused(TempoHardfork::Genesis, token_address)?;
+        assert!(
+            !is_paused,
+            "Unpaused tokens should not be detected as paused"
+        );
+
+        Ok(())
+    }
```
0xKitsune, line 686: Can we consolidate these tests?

crates/transaction-pool/src/amm.rs
```diff
@@ -66,13 +66,6 @@ impl AmmLiquidityCache {
...
-                // If user token matches one of the recently seen validator tokens,
-                // short circuit and return true. We assume that validators are willing to
-                // accept transactions that pay fees in their token directly.
-                if validator_token == &user_token {
-                    return Ok(true);
-                }
-
                 if let Some(validator_reserve) = inner.cache.get(&(user_token, *validator_token)) {
                     if *validator_reserve >= amount_out {
                         return Ok(true);
```
0xKitsune, line 78: Why do we remove `if validator_token == &user_token` here? This check seems unrelated to rejecting paused tokens and still relevant to the liquidity check. Lmk if I am overlooking something though.

crates/transaction-pool/src/validator.rs
```diff
@@ -356,6 +356,23 @@ where
...
+                        InvalidPoolTransactionError::other(
+                            TempoPoolTransactionError::PausedFeeToken(fee_token),
+                        ),
+                    );
+                }
+            }
+            Err(err) => {
+                return TransactionValidationOutcome::Error(*transaction.hash(), Box::new(err));
+            }
+        }
```
0xKitsune, line 511: Do we also need to check/evict transactions that are already in the pool where the fee token is paused?

mattsse: I think we should be a bit more careful with this because the token might be paused for only a short period and this would then evict them entirely, we could perhaps mark them as `paused` so that they are no longer in the pool

like an additional pool that we can maybe use for more things as well

fgimenez: makes sense, given that we will unify pool maintenance tasks in #2035 will wait for it to be merged to add the new pool here

crates/revm/src/common.rs
fgimenez: yep done

crates/transaction-pool/src/amm.rs
fgimenez: agree, this has been finally kept

crates/transaction-pool/src/validator.rs
fgimenez: done here ptal https://github.com/tempoxyz/tempo/pull/2041/commits/62464d70c53cd812ce671246b47d8dcb771f7124

crates/revm/src/common.rs
```diff
@@ -219,7 +219,23 @@ pub trait TempoStateAccess<M = ()> {
...
 
         // Ensure the currency is USD
-        self.is_tip20_usd(spec, fee_token)
+        // load fee token account to ensure that we can load storage for it.
+        self.with_read_only_storage_ctx(spec, || {
+            // SAFETY: prefix already checked above
+            let token = TIP20Token::from_address(fee_token)?;
+            Ok(token.currency.len()? == 3 && token.currency.read()?.as_str() == "USD")
+        })
+    }
```
0xrusowsky, line 223: is there a reason to replace the call with its own code?

fgimenez: good call, the only difference was not using `TIP20Token::from_address_unchecked`, and we actually want to use it, reverted to use `is_tip20_usd`

crates/transaction-pool/src/paused.rs
```diff
@@ -0,0 +1,270 @@
...
+
+/// Entry in the paused pool.
+#[derive(Debug, Clone)]
+pub struct PausedEntry {
+    /// The recovered transaction that was paused.
+    pub tx: Recovered<TempoTxEnvelope>,
+    /// The origin of the transaction (local, external, etc.).
+    pub origin: TransactionOrigin,
+    /// When the transaction was moved to the paused pool.
+    pub paused_at: Instant,
```
mattsse: because these are paused per token, this means that all are paused at the same time and we ould track this on a per token level instead?

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -196,6 +205,134 @@ where
...
+    }
+
+    /// Moves transactions for the specified fee token back from the paused pool.
+    ///
+    /// Called when a `PauseStateUpdate(isPaused=false)` event is detected.
+    /// Transactions are re-added via the normal validation path, so they will
+    /// be dropped if they have become invalid for other reasons.
+    ///
+    /// Returns the number of transactions that were re-added successfully.
+    pub(crate) async fn unpause_fee_token(&self, fee_token: Address) -> usize {
```
mattsse: this will wait until all transactins are added back to the pool, risking delays in maintain in case we have a lot of them

we should spawn this operation

crates/transaction-pool/src/paused.rs
```diff
@@ -0,0 +1,270 @@
...
+}
+
+/// Pool for transactions whose fee token is temporarily paused.
+///
+/// Transactions are indexed by both fee token address and transaction hash
+/// for efficient lookup in both directions.
+#[derive(Debug, Default)]
+pub struct PausedFeeTokenPool {
+    /// Fee token -> (tx_hash -> entry)
+    by_token: HashMap<Address, HashMap<TxHash, PausedEntry>>,
```
mattsse: we can keep this uncapped because after the token is paused we will reject new transactions so this is kinda bounded by the pool's own capacity and we dont expect too many of those

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -38,6 +42,10 @@ pub struct TempoTransactionPool<Client> {
     >,
     /// Minimal pool for 2D nonces (nonce_key > 0)
     aa_2d_pool: Arc<RwLock<AA2dPool>>,
+    /// Pool for transactions whose fee token is temporarily paused.
+    /// Transactions are moved here when their fee token is paused, and moved back
+    /// when the token is unpaused.
+    paused_fee_token_pool: Arc<RwLock<PausedFeeTokenPool>>,
```
mattsse: if we only interact with this from maintain, then we dont necessarily need a lock for this and could keep this in the maintain context, but maybe we want to add reuse this feature for other paused like states

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -196,6 +205,134 @@ where
...
+    /// Called when a `PauseStateUpdate(isPaused=true)` event is detected.
+    /// Transactions are scanned on-demand (pauses are rare).
+    ///
+    /// Returns the number of transactions moved to the paused pool.
+    pub(crate) fn pause_fee_token(&self, fee_token: Address) -> usize {
+        use reth_primitives_traits::transaction::TxHashRef;
+
+        let mut to_pause = Vec::new();
+
+        for tx in self.all_transactions().all() {
```
mattsse: this does a clone into recovered which is kinda expensive here, we can avoid this and track Arc validtransaction instead

crates/transaction-pool/src/maintain.rs
```diff
@@ -154,18 +157,73 @@ where
...
+                        if count > 0 {
+                            debug!(
+                                target: "txpool",
+                                %token,
+                                count,
+                                "Moved transactions to paused pool (fee token paused)"
+                            );
+                        }
+                    } else {
+                        let count = pool.unpause_fee_token(token).await;
```
mattsse: I think we should spawn this because this needs to revalidate the transactions

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -196,6 +205,134 @@ where
...
+    pub(crate) async fn unpause_fee_token(&self, fee_token: Address) -> usize {
+        let entries = self.paused_fee_token_pool.write().drain_token(&fee_token);
+        if entries.is_empty() {
+            return 0;
+        }
+
+        let mut success_count = 0;
+        for entry in entries {
+            let tx = TempoPooledTransaction::new(entry.tx);
+            match self.add_transaction(entry.origin, tx).await {
```
mattsse: we should also use the batch submission api here to submit all at once

crates/transaction-pool/src/paused.rs
mattsse: are we using this rn?

I think we should timeout these after a while as well, like if a token isnt unpaused within 30mins or smth, we clear them all

fgimenez: good catch, pushed changes to track pause timestamp at token level ptal

crates/transaction-pool/src/tempo_pool.rs
fgimenez: already addressed, lmk if anything else is needed

crates/transaction-pool/src/paused.rs
fgimenez: already addressed

crates/transaction-pool/src/tempo_pool.rs
fgimenez: ok moved paused fee token pool ownership to maintain task

crates/transaction-pool/src/tempo_pool.rs
fgimenez: good point, done

crates/transaction-pool/src/maintain.rs
fgimenez: already done

crates/transaction-pool/src/tempo_pool.rs
fgimenez: already done

crates/transaction-pool/src/paused.rs
fgimenez: already addressed

mattsse (approved):
this should work now and still be efficient

PR #2046, opened 2026-01-14
fix(tx-pool): address DoS vulnerabilities in mempool eviction and access-key validation

gakonst:
## Summary

This PR addresses two DoS-related audit findings from the Sigp security review:

### CHAIN-449: AA2dPool DoS via address-based eviction

**Problem:** The `discard()` function in `tt_2d_pool.rs` was evicting transactions based on lexicographic address ordering (highest addresses first). An attacker could mine addresses with leading zeros and fill the pool with unexecutable transactions that would never be evicted.

**Fix:** Changed eviction to use priority-based ordering (lowest priority first). This ensures attackers can't protect their transactions from eviction by using vanity addresses.

### CHAIN-451: Permanent mempool DoS via expired access-key

**Problem:** The mempool validator was NOT checking access-key expiry during validation. It only checked if the key existed and wasn't revoked. This allowed expired access-key transactions to remain in the mempool forever, repeatedly failing at execution time.

**Fixes:**
1. Added expiry validation in `validate_against_keychain()` using mempool-safe timestamps (`tip_timestamp`)
2. Added `evict_expired_keychain_txs()` maintenance task that tracks keychain-signed transactions and evicts them when their access keys expire, are revoked, or no longer exist

## Changes

- `tt_2d_pool.rs`: Changed `discard()` to evict by lowest priority instead of lexicographic address order
- `validator.rs`: Added access-key expiry check using `tip_timestamp`
- `maintain.rs`: Added `evict_expired_keychain_txs()` maintenance task for ongoing expiry monitoring

## Testing

- Updated `test_discard_removes_from_back` → `test_discard_removes_lowest_priority` to verify new eviction behavior
- All 132 tests pass
- Clippy clean

## Related Issues

- Fixes CHAIN-449
- Fixes CHAIN-451

### DIFF
--- a/docs/specs/foundry.lock
+++ b/docs/specs/foundry.lock
@@ -1,0 +1,8 @@
+{
+  "lib/tempo-std": {
+    "branch": {
+      "name": "master",
+      "rev": "91dfcf70289b07ec6409f289917c6d4c9ce7e73e"
+    }
+  }
+}
--- a/docs/specs/test/StablecoinDEX.t.sol
+++ b/docs/specs/test/StablecoinDEX.t.sol
@@ -604,14 +604,13 @@ contract StablecoinDEXTest is BaseTest {
     }
 
     function test_PlaceOrder_SucceedsAbove_MinimumOrderSize(uint128 amount) public {
-        // For bid orders (buying token1 with pathUSD), the escrow amount is:
-        // escrow = amount * tickToPrice(100) / PRICE_SCALE
-        // escrow = amount * (PRICE_SCALE + 100) / PRICE_SCALE
-        // escrow = amount * (1000000 + 100) / 1000000
+        // For bid orders (buying token1 with pathUSD), the escrow amount uses ceiling division:
+        // escrow = ceil(amount * tickToPrice(100) / PRICE_SCALE)
+        // escrow = ceil(amount * (PRICE_SCALE + 100) / PRICE_SCALE)
+        // escrow = ceil(amount * 1000100 / 1000000)
         // We need escrow <= INITIAL_BALANCE
-        // So: amount * 1000100 / 1000000 <= INITIAL_BALANCE
-        // Therefore: amount <= INITIAL_BALANCE * 1000000 / 1000100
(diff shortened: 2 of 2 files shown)


gakonst (commented):
## High-Impact Review: CHAIN-449/CHAIN-451 DoS Fixes

### Issue 1: Duplicate maintenance task - conflicts with PR #2035

This PR adds `evict_expired_keychain_txs()` as a separate maintenance task, but PR #2035 consolidates all maintenance into a unified `maintain_tempo_pool()` task that already handles keychain revocation.

If both PRs merge, there will be **two competing tasks** listening to the same canonical state stream and trying to evict keychain transactions. This creates:
- Race conditions between tasks
- Duplicate eviction attempts
- Wasted resources

**Recommendation**: This PR should be rebased on top of #2035 and integrated into the unified maintenance task, or one of the PRs should be closed.

### Issue 2: evict_expired_keychain_txs does storage reads on every block

```rust
let Ok(state_provider) = pool.client().latest() else { ... };

for tx_info in tracked_txs.drain(..) {
    let storage_slot = AccountKeychain::new().keys[tx_info.user_address][tx_info.key_id].base_slot();
    let slot_value = state_provider.storage(ACCOUNT_KEYCHAIN_ADDRESS, storage_slot.into())?;
    ...
}
```

For N tracked keychain transactions, this does N storage reads per block. With a large mempool, this could be expensive.

The unified task in #2035 uses **event-based** eviction (scanning `KeyRevoked` events), which is O(events) not O(transactions). The event-based approach is more efficient.

### Issue 3: expiry == 0 treated as "doesn't exist" but also triggers eviction

```rust
if authorized_key.is_revoked
    || authorized_key.expiry == 0  // "doesn't exist"
    || authorized_key.expiry <= tip_timestamp  // expired
{
    to_remove.push(tx_info.hash);
}
```

If `expiry == 0` means the key doesn't exist, how did the transaction pass validation? This suggests either:
1. A race between validation and maintenance, OR
2. The key was deleted after validation

The first case is fine (eviction is correct), but for the second - can keys actually be deleted (set to 0) without a `KeyRevoked` event? If so, this is a gap in the event-based approach of #2035.

PR #2053, opened 2026-01-14
refactor(revm): move 2D nonce gas calculation to intrinsic gas

rakita:
## Summary

- Move 2D nonce gas calculation from state validation phase to intrinsic gas calculation
- Remove `nonce_2d_gas` field from `TempoEvm` struct
- Remove `calculate_2d_nonce_gas` function
- Simplify gas logic: if `nonce_key > 0` and `tx.nonce == 0`, charge `NEW_NONCE_KEY_GAS`, otherwise `EXISTING_NONCE_KEY_GAS`
- Update txpool validator to include 2D nonce gas in intrinsic gas validation

## Test plan

- [x] All tempo-revm tests pass (55 tests)
- [x] All tempo-transaction-pool tests pass (132 tests)
- [x] Updated tests to verify 2D nonce gas is correctly included in intrinsic gas

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -10,6 +10,8 @@ ignore = [
   "RUSTSEC-2026-0002",
   # https://rustsec.org/advisories/RUSTSEC-2025-0141 bincode is unmaintained
   "RUSTSEC-2025-0141",
+  # https://rustsec.org/advisories/RUSTSEC-2026-0003 cmov timing on ARM32 - we don't target ARM32
+  "RUSTSEC-2026-0003",
 ]
 
 # This section is considered when running `cargo deny check bans`.


crates/transaction-pool/src/validator.rs
```diff
@@ -232,10 +236,22 @@ where
...
+        // If tx nonce is 0, it's a new key (0 -> 1 transition), otherwise existing key
+        if !tx.nonce_key.is_zero() {
+            if tx.nonce == 0 {
+                // New key - cold SLOAD + SSTORE set (0 -> non-zero)
+                init_and_floor_gas.initial_gas += NEW_NONCE_KEY_GAS;
+            } else {
+                // Existing key - cold SLOAD + warm SSTORE reset
+                init_and_floor_gas.initial_gas += EXISTING_NONCE_KEY_GAS;
+            }
+        }
```
rakita, line 280: This looks like a bug where we didn't check full intrinsic gas in txpool. This means someone can ddos txpool by passing balance check but than fail on block building as additional balance is required. 

Bumping this in txpool is safe thing

crates/revm/src/handler.rs
mattsse, line 1406: hmm unsure if this is still equivalent if we remove the t0 spec check here

rakita: I think you are right, this was a reason to introduce new protocol version. Was working on slightly older tempo version and didn't see it.

rakita: This commit https://github.com/tempoxyz/tempo/pull/2053/changes/721b2ef4728d66040000605d7bcb9098533c277b should be a fix as without `is_t0` we dont do this check at all

rakita: t0 gating was done in https://github.com/tempoxyz/tempo/pull/2008

crates/transaction-pool/src/validator.rs
legion2002: Yes, I think this was because initially 2d nonce gas calculations required some EVM state about how many nonce lanes were open.
But we have since changed the design, so should have added this back. (and probably we should have found some way to read that ethereum state in the validator too)

legion2002: good catch!

0xKitsune (approved):

PR #2054, opened 2026-01-14
feat(precompiles): add getFeeToken() for transaction fee token introspection

gakonst:
## Summary

Add `IFeeManager.getFeeToken()` view function that allows smart contracts to read the fee token being used for the current transaction during execution.

**Requested by:** LayerZero/Stargate for their integration with Tempo's TIP-20 fee system.

**Context:** https://tempoxyz.slack.com/archives/C09GJBN2SEL/p1768415497871839?thread_ts=1767384201.823829&cid=C09GJBN2SEL

## Implementation

Uses the same transient storage pattern already established by `AccountKeychain` for storing `tx_origin`:

1. **Interface change**: Added `getFeeToken()` to `IFeeManager` interface in contracts crate
2. **Transient storage**: Added `tx_fee_token` field to `TipFeeManager` struct (at the end, as transient)
3. **Handler integration**: Set `tx_fee_token` in `TempoEvmHandler::validate_against_state_and_deduct_caller` before transaction execution
4. **Dispatch**: Wired `getFeeToken()` to read from transient storage

## Usage

From Solidity:
```solidity
interface IFeeManager {
    function getFeeToken() external view returns (address);
    // ... existing functions
}

contract MyContract {
    IFeeManager constant FEE_MANAGER = IFeeManager(0xfeec000000000000000000000000000000000000);

    function someFunction() external {
        address feeToken = FEE_MANAGER.getFeeToken();
        // Use feeToken to adjust logic, pricing, etc.
    }
}
```

## Testing

- Added unit tests for `set_tx_fee_token()` and `get_fee_token()` in mod.rs
- Added dispatch tests for `getFeeToken()` call via precompile interface  
- Verified selector coverage test still passes
- All 509 precompile tests pass

## Notes

- Returns `address(0)` if no fee token is set (e.g., during free transactions or if called outside transaction context)
- Safe for `staticcall` since it only reads transient storage
- Transient storage is automatically cleared between transactions

### DIFF
--- a/crates/precompiles/src/tip_fee_manager/dispatch.rs
+++ b/crates/precompiles/src/tip_fee_manager/dispatch.rs
@@ -47,6 +47,9 @@ impl Precompile for TipFeeManager {
             TipFeeManagerCall::FeeManager(IFeeManagerCalls::collectedFees(call)) => {
                 view(call, |c| self.collected_fees[c.validator][c.token].read())
             }
+            TipFeeManagerCall::FeeManager(IFeeManagerCalls::getFeeToken(call)) => {
+                view(call, |_| self.get_fee_token())
+            }
 
             // IFeeManager mutate functions
             TipFeeManagerCall::FeeManager(IFeeManagerCalls::setValidatorToken(call)) => {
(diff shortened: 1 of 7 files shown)


crates/precompiles/src/tip_fee_manager/mod.rs
```diff
@@ -232,6 +239,29 @@ impl TipFeeManager {
             Ok(token)
         }
     }
+
+    // -----------------------------------------------------------------
+    // Transaction context methods
+    // -----------------------------------------------------------------
```
0xKitsune: nit: lets remove these header comments, this is not needed and can be inferred from the function names.
```suggestion
```

crates/precompiles/src/tip_fee_manager/mod.rs
```diff
@@ -232,6 +239,29 @@ impl TipFeeManager {
...
+
+    /// Set the current transaction's fee token (transient storage).
+    ///
+    /// This is called by the execution handler before transaction execution begins.
+    /// It allows smart contracts to introspect which token is paying for gas fees
+    /// by calling `getFeeToken()`.
+    ///
+    /// NOTE: This is a protocol-internal function, not exposed via the dispatch interface.
```
0xKitsune:
```suggestion
    /// Set the current transaction's fee token (transient storage)
    ///
    /// NOTE: Thishis function is only called by the execution handler before 
    /// transaction execution begins and allows smart contracts to introspect which 
    /// token is paying for gas fees by calling `getFeeToken()`.
    ///
    /// This is a protocol-internal function, not exposed via the dispatch interface.
```

0xKitsune (approved):
~~Lgtm, need to decide if we want to put this into T0. Pending @0xrusowsky~~

crates/precompiles/src/tip_fee_manager/dispatch.rs
```diff
@@ -47,6 +47,9 @@ impl Precompile for TipFeeManager {
             TipFeeManagerCall::FeeManager(IFeeManagerCalls::collectedFees(call)) => {
                 view(call, |c| self.collected_fees[c.validator][c.token].read())
             }
+            TipFeeManagerCall::FeeManager(IFeeManagerCalls::getFeeToken(call)) => {
+                view(call, |_| self.get_fee_token())
+            }
```
0xKitsune, line 63: Lets actually put this behind the `T0` hardfork as this is a breaking change.

0xKitsune (changes requested):

crates/revm/src/handler.rs
```diff
@@ -648,6 +648,15 @@ where
...
 
+        // Set the transaction fee token in the fee manager's transient storage
+        // so contracts can introspect it via IFeeManager.getFeeToken().
+        let fee_token = self.fee_token;
+        StorageCtx::enter_evm(journal, block, cfg, || {
+            let mut fee_manager = TipFeeManager::new();
+            fee_manager.set_tx_fee_token(fee_token)
+        })
+        .map_err(|e| EVMError::Custom(e.to_string()))?;
+
```
0xrusowsky, line 659: i believe this is breaking because TLOAD/TSTORE incurs gas, thus it should go behind a HF

0xrusowsky (commented):
besides the cmnt, it LGTM

0xKitsune:
Closing for now as we have decided to include this in a subsequent hardfork. We will open an issue to track this in the meantime.

crates/precompiles/src/tip_fee_manager/dispatch.rs
jenpaff: T0? or T1?

PR #2061, opened 2026-01-14
fix(specs): remove incorrect TEMPO-AMM30 validator reserve assertion

grandizzy:
## Summary

Removes incorrect assertion that initialized pools must have `reserveValidatorToken > 0`.

## Problem

The `test_repro_TEMPO_AMM30_ZeroValidatorReserve` test was failing because the invariant incorrectly assumed that initialized pools (totalSupply > 0) must always have non-zero validator reserves.

## Root Cause

Due to rounding in burn calculations, when `(liquidity * reserve / totalSupply)` rounds down to 0, the burner's LP tokens are burned but they receive 0 tokens. This can drain totalSupply without proportionally draining reserves, eventually leading to `reserves = 0` while `totalSupply >= MIN_LIQUIDITY`.

## Fix

Remove the assertion and document that zero validator reserves in initialized pools is a valid (though suboptimal) state, not a bricked pool. The key invariant that `totalSupply >= MIN_LIQUIDITY` is preserved.

### DIFF
--- a/docs/specs/test/invariants/FeeAMM.t.sol
+++ b/docs/specs/test/invariants/FeeAMM.t.sol
@@ -1397,11 +1397,13 @@ contract FeeAMMInvariantTest is BaseTest {
                 totalSupply >= MIN_LIQUIDITY,
                 "TEMPO-AMM30: Initialized pool has totalSupply < MIN_LIQUIDITY (bricked state)"
             );
-            // At least validator reserve must be > 0 (mints always deposit validator tokens)
-            assertTrue(
-                pool.reserveValidatorToken > 0,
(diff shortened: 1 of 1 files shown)


0001-Take-1.patch
```diff
@@ -0,0 +1,134 @@
+From 0bcd3eeb7b7d640a5ee2537d156cca9bcd70ba19 Mon Sep 17 00:00:00 2001
```
0xKitsune: Can we remove this file?

0xKitsune (approved):

0001-Take-1.patch
grandizzy: oops, removed

PR #2068, opened 2026-01-15
docs: add TIP spec for Multisig precompile

gakonst:
## Summary

Adds a TIP specification for native multisig wallets as precompiles on Tempo.

## Key Design Decisions

- **Multisig Factory** at `0xF15C000000000000000000000000000000000000`
- **Deterministic addressing**: `MULTISIG_PREFIX || keccak256(sorted_owners, threshold)[0:8]`
- **On-chain confirmations**: submit → confirm → execute flow
- **TIP-20 integration**: `value` field encodes `(token, amount)` for transfers
- **Immutable config**: owners and threshold fixed at creation

### Factory
- `createMultisig(owners, threshold)` → deterministic address
- `getMultisigAddress(owners, threshold)` → predict address
- `isMultisig(addr)` → check if deployed

### Multisig
- `submitTransaction(to, value, data)` → create + auto-confirm
- `confirmTransaction(txId)` → add confirmation
- `revokeConfirmation(txId)` → remove confirmation
- `executeTransaction(txId)` → execute when threshold met

## Next Steps

- [ ] Assign TIP number
- [ ] Review and iterate on spec
- [ ] Implement precompiles following TIP-20 patterns

### DIFF
--- a/docs/pages/protocol/tips/tip-multisig.mdx
+++ b/docs/pages/protocol/tips/tip-multisig.mdx
@@ -373,7 +373,55 @@ interface IMultisig {
         bytes calldata value,
         bytes calldata data
     ) external;
+
+    // ============ Owner Management ============
+
+    /// @notice Adds a new owner to the multisig
+    /// @dev Requires threshold confirmations. Reverts if owner already exists or MAX_OWNERS reached.
+    /// @param owner The address to add as owner
+    function addOwner(address owner) external;
+
+    /// @notice Removes an owner from the multisig
+    /// @dev Requires threshold confirmations. Reverts if owner doesn't exist or would make threshold > ownerCount.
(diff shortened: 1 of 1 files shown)


gakonst:
Closing this PR - after deeper review, the SimpleMultisig spec isn't production ready.

**Critical issues:**
- ETH can be force-sent and become permanently stuck (no `receive()`, `value` must be 0)
- No nonce means identical txs can't be submitted twice

**Missing vs production multisigs (Gnosis Safe):**
- No EIP-1271 for dapp integrations
- No batching/multicall
- No timelocks/guards/modules
- No gas params or relayer support

The implementation works fine as a lightweight governance executor, but doesn't meet the bar for a general-purpose wallet spec. Will revisit if there's a specific use case that doesn't require these features.

PR #2069, opened 2026-01-15
fix(revm): validate TIP20 token before balance lookup to prevent panic

gakonst:
## Summary

Fixes *BUG-002: TIP20 Invalid Token Causes EVM Panic*

When an invalid fee token bypasses early validation in `load_fee_fields` (which can happen when `max_balance_spending` is zero and it's not a subblock transaction), the subsequent call to `get_token_balance` would panic with:

```
TIP20 prefix already validated: TIP20(InvalidToken(InvalidToken))
```

This crashes the entire node/test runner rather than gracefully rejecting the transaction.

## Fix

Adds early validation in `validate_against_state_and_deduct_caller` to check `is_tip20_prefix` before calling `get_token_balance`. If the token is invalid, returns `TempoInvalidTransaction::InvalidFeeToken` error instead of panicking.

## Testing

- All 16 handler tests pass
- Invariant tests should now complete without panicking on invalid fee tokens

## Impact

- Prevents DoS vector where attackers could crash nodes with malformed transactions
- Eliminates consensus risk from divergent node crashes
- Invariant tests can now run to completion


crates/revm/src/handler.rs
```diff
@@ -648,6 +648,13 @@ where
         })
         .map_err(|e| EVMError::Custom(e.to_string()))?;
 
+        // Validate fee token has TIP20 prefix before attempting to read balance.
+        // This prevents panic in get_token_balance if an invalid token bypassed earlier validation
+        // (e.g., when max_balance_spending is zero and it's not a subblock transaction).
```
0xKitsune:
```suggestion
        // Validate fee token has TIP20 prefix before attempting to read balance.
        // Note that this is checked during tx validation but this check is also necessary at the protocol level since it is possible for a validator to bypass initial validation and inject the tx directly to the block
    
```

crates/revm/src/handler.rs
```diff
@@ -1483,6 +1490,42 @@ mod tests {
...
+
+        assert!(result.is_err(), "Should reject invalid fee token");
+        let err = result.unwrap_err();
+        match err {
+            EVMError::Transaction(TempoInvalidTransaction::InvalidFeeToken(addr)) => {
+                assert_eq!(addr, invalid_token, "Error should contain the invalid token address");
+            }
+            other => panic!("Expected InvalidFeeToken error, got: {:?}", other),
+        }
+    }
```
0xKitsune: nit: can we update to use `assert!(matches!())`

crates/revm/src/handler.rs
```diff
@@ -648,6 +648,13 @@ where
         })
         .map_err(|e| EVMError::Custom(e.to_string()))?;
 
+        // Validate fee token has TIP20 prefix before attempting to read balance.
+        // Note that this is checked during tx validation but this check is also necessary at the protocol level since it is possible for a validator to bypass initial validation and inject the tx directly to the block
+    
+        if !is_tip20_prefix(self.fee_token) {
+            return Err(TempoInvalidTransaction::InvalidFeeToken(self.fee_token).into());
+        }
```
0xKitsune: It looks like we need this check due to `load_fee_fields()` skipping this if `max_balance_spending` is 0. @rakita @0xrusowsky when would a valid tx result in a max balance spending of `0`?


```rust

impl<DB: alloy_evm::Database, I> TempoEvmHandler<DB, I> {
    fn load_fee_fields(
        &mut self,
        evm: &mut TempoEvm<DB, I>,
    ) -> Result<(), EVMError<DB::Error, TempoInvalidTransaction>> {
        let ctx = evm.ctx_mut();

        self.fee_payer = ctx.tx.fee_payer()?;
        self.fee_token = ctx
            .journaled_state
            .get_fee_token(&ctx.tx, self.fee_payer, ctx.cfg.spec)
            .map_err(|err| EVMError::Custom(err.to_string()))?;

        // Skip fee token validity check for cases when the transaction is free and is not a part of a subblock.
        if (!ctx.tx.max_balance_spending()?.is_zero() || ctx.tx.is_subblock_transaction())
            && !ctx
                .journaled_state
                .is_valid_fee_token(ctx.cfg.spec, self.fee_token)
                .map_err(|err| EVMError::Custom(err.to_string()))?
        {
            return Err(TempoInvalidTransaction::InvalidFeeToken(self.fee_token).into());
        }

        Ok(())
    }

```

fgimenez (approved):
lgtm

crates/revm/src/handler.rs
```diff
@@ -648,6 +648,14 @@ where
         })
         .map_err(|e| EVMError::Custom(e.to_string()))?;
 
+        // Validate fee token has TIP20 prefix before attempting to read balance.
+        // Note that this is checked during tx validation but this check is also
+        // necessary at the protocol level since it is possible for a validator to
+        // bypass initial validation and inject the tx directly to the block
+        if !is_tip20_prefix(self.fee_token) {
```
rakita: Should this check be done in place where `self.fee_token` is set?

0xKitsune: This should be updated now, can you PTAL?

crates/revm/src/handler.rs
```diff
@@ -199,7 +199,13 @@ impl<DB: alloy_evm::Database, I> TempoEvmHandler<DB, I> {
...
-        // Skip fee token validity check for cases when the transaction is free and is not a part of a subblock.
+        // Always validate TIP20 prefix to prevent panics in get_token_balance.
+        // This is a protocol-level check since validators could bypass initial validation.
+        if !is_tip20_prefix(self.fee_token) {
+            return Err(TempoInvalidTransaction::InvalidFeeToken(self.fee_token).into());
+        }
+
+        // Skip additional fee token validity check for cases when the transaction is free and is not a part of a subblock.
         if (!ctx.tx.max_balance_spending()?.is_zero() || ctx.tx.is_subblock_transaction())
             && !ctx
```
0xKitsune, line 211: Now that we are checking the `tip20_prefix` here, there is a redundant check in `is_valid_fee_token()` that also checks for the prefix.

```rust
    /// Checks if the given token can be used as a fee token.
    fn is_valid_fee_token(&mut self, spec: TempoHardfork, fee_token: Address) -> TempoResult<bool>
    where
        Self: Sized,
    {
        // Must have TIP20 prefix to be a valid fee token
        if !is_tip20_prefix(fee_token) {
            return Ok(false);
        }

        // Ensure the currency is USD
            // SAFETY: prefix already checked above
            let token = TIP20Token::from_address(fee_token)?;
            Ok(token.currency.len()? == 3 && token.currency.read()?.as_str() == "USD")
        })
    }

```

We should have a separate function to `is_tip20_usd()` that is then called in `is_valid_fee_token()` and separately called in `load_fee_fields` after checking for tip20 prefix.

rakita (approved):
lgtm

0xrusowsky (approved):

crates/precompiles/src/tip20/mod.rs
```diff
@@ -596,6 +596,16 @@ impl TIP20Token {
         Ok(Self::__new(address))
     }
 
+    /// Creates a TIP20Token without validating the prefix.
+    ///
+    /// # Safety
+    /// Caller must ensure `is_tip20_prefix(address)` returns true.
+    #[inline]
+    pub fn from_address_unchecked(address: Address) -> Self {
```
0xKitsune, line 604: Can we make this `pub(crate)`?

crates/revm/src/common.rs
```diff
@@ -194,6 +194,20 @@ pub trait TempoStateAccess<M = ()> {
...
+    fn is_tip20_usd(&mut self, spec: TempoHardfork, fee_token: Address) -> TempoResult<bool>
+    where
+        Self: Sized,
+    {
+        self.with_read_only_storage_ctx(spec, || {
+            // SAFETY: caller must ensure prefix is already checked
+            let token = TIP20Token::from_address_unchecked(fee_token);
+            Ok(token.currency.len()? == 3 && token.currency.read()?.as_str() == "USD")
+        })
+    }
```
0xKitsune, line 209: Now that we have an isolated function for this, can we add tests?

0xKitsune (approved):

crates/precompiles/src/tip20/mod.rs
fgimenez: we are using it from revm crate, am i missing something?

crates/revm/src/common.rs
fgimenez: sure added here https://github.com/tempoxyz/tempo/pull/2069/commits/4128c229ab5a722277bf7ddb46dd45039c693715

fgimenez:
cargo deny error on CI should be fixed with https://github.com/tempoxyz/tempo/pull/2091

PR #2075, opened 2026-01-15
chore(payload): refactor and cleanup to improve testing

0xrusowsky:
This PR refactors the payload builder to use helper fns to perform checks, so that we can test them in isolation


crates/payload/builder/src/lib.rs
```diff
@@ -68,6 +68,122 @@ fn has_expired_transactions(subblock: &RecoveredSubBlock, timestamp: u64) -> boo
...
+        Self {
+            total_gas_used: 0,
+            non_payment_gas_used: 0,
+            block_size: initial_block_size,
+        }
+    }
+
+    /// Records a transaction's gas usage and size.
+    fn add_tx(&mut self, gas_used: u64, rlp_length: usize, is_payment: bool) {
+        self.total_gas_used += gas_used;
```
fgimenez: could this overflow, should we use `checked_add` here?

crates/payload/builder/src/lib.rs
```diff
@@ -68,6 +68,122 @@ fn has_expired_transactions(subblock: &RecoveredSubBlock, timestamp: u64) -> boo
...
+        self.block_size += size;
+    }
+
+    /// Checks if a transaction fits within the pool gas budget.
+    fn check_gas_limit(
+        &self,
+        tx_gas_limit: u64,
+        gas_limits: &GasLimits,
+    ) -> Result<(), InvalidPoolTransactionError> {
+        if self.total_gas_used + tx_gas_limit > gas_limits.non_shared {
```
fgimenez: same about potential overflow

PR #2079, opened 2026-01-15
feat: encode full finalization/notarization in consensus namespace

gakonst:
## Summary

Previously, the consensus namespace RPC types only encoded the `certificate` field (just the threshold signature). This change updates it to encode the full `Finalization` or `Notarization` object, which includes the proposal alongside the certificate.

This matches the approach taken in #1918 for identity transition proofs, ensuring clients have all the data needed to verify certificates independently.

## Changes

- Renamed `certificate` field to `certification` in `CertifiedBlock`
- Updated encoding to use full notarization/finalization objects in both:
  - `crates/commonware-node/src/feed/actor.rs` - for live events
  - `crates/commonware-node/src/feed/state.rs` - for historical lookups

## Breaking Change

This is a breaking change for RPC clients:
- JSON field name changed from `certificate` to `certification`
- The hex-encoded value now contains the full `Finalization`/`Notarization` object instead of just the certificate

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,4 @@ docs/.sonda
 _
 .vercel
 .env*.local
+.worktrees/


crates/node/src/rpc/consensus/types.rs
```diff
@@ -14,7 +14,8 @@ pub struct CertifiedBlock {
     /// Block height, if known. May be `None` if the block hasn't been stored yet.
     pub height: Option<u64>,
     pub digest: B256,
-    pub certificate: String,
+    /// Hex-encoded full certification (Notarization or Finalization), not just the certificate.
```
joshieDo:
```suggestion
    /// Hex-encoded full notarization or finalization).
```

crates/commonware-node/src/feed/actor.rs
```diff
@@ -79,15 +79,15 @@ impl<TContext: Spawner> Actor<TContext> {
         }
     }
 
-    /// Create a [`CertifiedBlock`] from a proposal and certificate.
+    /// Create a [`CertifiedBlock`] from the full certification (Notarization or Finalization).
```
joshieDo:
```suggestion
    /// Create a [`CertifiedBlock`] from the notarization or finalization.
```

crates/node/src/rpc/consensus/types.rs
```diff
@@ -14,7 +14,8 @@ pub struct CertifiedBlock {
     /// Block height, if known. May be `None` if the block hasn't been stored yet.
     pub height: Option<u64>,
     pub digest: B256,
-    pub certificate: String,
+    /// Hex-encoded full notarization or finalization).
```
joshieDo:
```suggestion
    /// Hex-encoded full notarization or finalization.
```

crates/node/src/rpc/consensus/types.rs
```diff
@@ -14,6 +14,7 @@ pub struct CertifiedBlock {
     /// Block height, if known. May be `None` if the block hasn't been stored yet.
     pub height: Option<u64>,
     pub digest: B256,
+    /// Hex-encoded full notarization or finalization.
     pub certificate: String,
```
SuperFluffy, line 18: Wouldn't it be nicer to not dump just the encoded bytes, but the "full" object so it's a bit more structured? Basically add a custom `struct Notarization {}` and `struct Finalization {}` that matches the commonware types?

PR #2087, opened 2026-01-15
fix(revm): discard state and manually bump nonce in catch_error

gakonst:
## Summary

Rework `catch_error` in `TempoEvmHandler` to properly handle failed subblock transactions.

## Changes

- **Discard state**: Use `finalize()` to clear the inner journal state instead of `commit_tx()`, discarding all changes from the failed transaction
- **Manually bump nonce**: After clearing state, bump the appropriate nonce for replay protection:
  - **Account nonce** (`nonce_key == 0` or no tempo_tx_env): bump via `bump_nonce()`
  - **2D nonce** (`nonce_key != 0`): bump via `NonceManager::increment_nonce()` through the precompile storage context

This ensures that failed `collectFeePreTx` calls in subblock transactions don't leave behind unwanted state changes while still maintaining replay protection through nonce increments.


crates/revm/src/handler.rs
rakita, line 1088: @klkvr have a question here, why do we partially chose errors. Isn't idea for all subblock transaction to never fail?

PR #2091, opened 2026-01-15
chore(deny): ignore RUSTSEC-2026-0003 cmov timing advisory (ARM32 not targeted)

fgimenez:
Ignores RUSTSEC-2026-0003 which reports a timing side-channel in the `cmov` crate on ARM32. We don't target ARM32, so this advisory doesn't affect us.

https://rustsec.org/advisories/RUSTSEC-2026-0003


0xKitsune (approved):

mattsse (approved):
lgtm

PR #2093, opened 2026-01-15
fix(consensus): deduplicate metrics by attaching the epoch

SuperFluffy:
This patch deduplicates simplex metrics by adding a `_{epoch}_` infix.

To ensure that metrics are unique from now on, a simple test was added
that revealed more duplicate dupes coming out of the commonware
dependencies (NOTE: commonware cf349ae38b862f0336e9b33b3559ec846a499e0c enforces that all tests are unique from now, so strictly speaking this test is not necessary, but we will keep it in case of upstream regressions).

To not trigger the upstreamed assertions that metrics are unique, this patch changes a node's context from `{uid}` to `{uid}_{n_starts}`.

Note that metrics starting with `consensus_engine_epoch_manager_consensus_engine_*` are now renamed `consensus_engine_epoch_manager_simplex_epoch_{epoch}` to get rid of the confusing duplicate word parts.

This patch *is* backward compatible since no persisted on-disk storage is affected. However, it *will break* grafana dashboards and alerts. However, since our metrics were duplicate, any dashboards and alerts were subtly broken and unusable anyway, so this does not constitute a regression.

This was reported upstream and fixed in cf349ae38b862f0336e9b33b3559ec846a499e0c, which with this patch tempo is pointing to.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -52,7 +52,7 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - uses: rui314/setup-mold@725a8794d15fc7563f59595bd9556495c0564878 # v1
       - uses: mozilla-actions/sccache-action@7d986dd989559c6ecdb630a3fd2557667be217ad # v0.0.9
-      - uses: taiki-e/install-action@03ef6f57d573ca4522fb02950f326083373b85bf # v2.66.2
+      - uses: taiki-e/install-action@2e9d707ef49c9b094d45955b60c7e5c0dfedeb14 # v2.66.5
         with:
           tool: nextest
       - name: Build and compile tests


crates/commonware-node/src/epoch/manager/actor.rs
```diff
@@ -322,7 +322,8 @@ where
         self.config.scheme_provider.register(epoch, scheme.clone());
 
         let engine = simplex::Engine::new(
-            self.context.with_label("consensus_engine"),
+            // self.context.with_label("consensus_engine"),
```
joshieDo:
```suggestion
```
?

SuperFluffy: [285f649](https://github.com/tempoxyz/tempo/pull/2093/commits/285f649fd2c3fa7390522a403838b19aa1bb670a)

PR #2097, opened 2026-01-15
fix(txpool): evict keychain transactions on spending limit changes

gakonst:
Closes CHAIN-444

## Summary

Extends the pool maintenance task to evict keychain-signed transactions when their spending limits are modified.

## Problem

TOCTOU vulnerability where:
1. Keychain transactions pass validation
2. Spending limit is reduced below transaction value
3. Transactions fail at execution but remain stuck in mempool

## Solution

- Listen for `SpendingLimitUpdated` events in the maintenance task
- Evict all affected transactions when limits change (conservative but safe)
- Added integration test to verify the fix

Stacked on #2035

### DIFF
--- a/crates/transaction-pool/src/paused.rs
+++ b/crates/transaction-pool/src/paused.rs
@@ -5,10 +5,10 @@
 //! When the token is unpaused, transactions are moved back to the main pool
 //! and re-validated.
 
-use crate::{RevokedKey, SpendingLimitUpdates, transaction::TempoPooledTransaction};
+use crate::{RevokedKeys, SpendingLimitUpdates, transaction::TempoPooledTransaction};
 use alloy_primitives::{Address, TxHash, map::HashMap};
 use reth_transaction_pool::ValidPoolTransaction;
-use std::{collections::HashSet, sync::Arc, time::Instant};
+use std::{sync::Arc, time::Instant};
 
 /// Duration after which paused transactions are expired and removed.
 /// If a token isn't unpaused within this time, we clear all pending transactions.
@@ -138,11 +138,11 @@ impl PausedFeeTokenPool {
     /// Removes transactions matching invalidation criteria from the paused pool.
     ///
     /// This handles both revoked keys and spending limit updates in a single pass.
-    /// Uses pre-built sets for O(1) lookup per transaction.
+    /// Uses account-keyed indexes for O(1) account lookup per transaction.
     /// Returns the number of transactions removed.
     pub fn evict_invalidated(
         &mut self,
-        revoked_keys: &HashSet<RevokedKey>,
+        revoked_keys: &RevokedKeys,
         spending_limit_updates: &SpendingLimitUpdates,
     ) -> usize {
         if revoked_keys.is_empty() && spending_limit_updates.is_empty() {
--- a/crates/transaction-pool/src/transaction.rs
+++ b/crates/transaction-pool/src/transaction.rs
@@ -16,7 +16,7 @@ use reth_transaction_pool::{
     error::PoolTransactionError,
 };
 use std::{
-    collections::{HashMap, HashSet},
+    collections::HashMap,
     convert::Infallible,
     fmt::Debug,
     sync::{Arc, OnceLock},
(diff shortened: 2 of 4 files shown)


crates/transaction-pool/src/transaction.rs
```diff
@@ -20,11 +20,51 @@ use std::{
...
+/// revocation and spending limit events.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct KeychainSubject {
+    /// The account that owns the keychain key (from `user_address` in the signature).
+    pub account: Address,
+    /// The key ID recovered from the keychain signature.
+    pub key_id: Address,
+    /// The fee token used by this transaction.
+    pub fee_token: Address,
+}
```
mattsse: these types are defined out of order

this should be moved below the main tx type of this module

crates/transaction-pool/src/paused.rs
```diff
@@ -135,29 +135,28 @@ impl PausedFeeTokenPool {
...
+    /// Removes transactions matching invalidation criteria from the paused pool.
     ///
+    /// This handles both revoked keys and spending limit updates in a single pass.
     /// Returns the number of transactions removed.
-    pub fn evict_by_revoked_keys(&mut self, revoked_keys: &[(Address, Address)]) -> usize {
-        if revoked_keys.is_empty() {
+    pub fn evict_invalidated(
+        &mut self,
+        revoked_keys: &[(Address, Address)],
+        spending_limit_updates: &[(Address, Address, Address)],
```
mattsse: these are a lot of address types, not super obvious what these represent, maybe worth adding a helper struct for these or at least document the order

crates/transaction-pool/src/transaction.rs
```diff
@@ -20,11 +20,51 @@ use std::{
...
+    /// Only matches if the transaction's fee token matches the updated token.
+    pub fn matches_spending_limit_update(
+        &self,
+        spending_limit_updates: &[(Address, Address, Address)],
+    ) -> bool {
+        spending_limit_updates
+            .iter()
+            .any(|&(account, key_id, token)| {
+                self.account == account && self.key_id == key_id && self.fee_token == token
+            })
```
mattsse: we need to apply this on every tx in the pool, right?

hould this be a map instead, because I assume most tx would not be affected by this?

crates/transaction-pool/src/transaction.rs
```diff
@@ -20,11 +20,51 @@ use std::{
...
+}
+
+impl KeychainSubject {
+    /// Returns true if this subject matches any of the revoked keys.
+    ///
+    /// Revoked keys are identified by (account, key_id) pairs.
+    pub fn matches_revoked(&self, revoked_keys: &[(Address, Address)]) -> bool {
+        revoked_keys
+            .iter()
+            .any(|&(account, key_id)| self.account == account && self.key_id == key_id)
```
mattsse: same here, I think we can optimize this when we decode the events so that all of this becomes cheaper over all?

crates/transaction-pool/src/transaction.rs
```diff
@@ -865,3 +897,88 @@ mod tests {
...
+    /// Uses a pre-built set for O(1) lookup instead of linear scan.
+    pub fn matches_spending_limit_update(
+        &self,
+        spending_limit_updates: &HashSet<SpendingLimitUpdate>,
+    ) -> bool {
+        spending_limit_updates.contains(&SpendingLimitUpdate::new(
+            self.account,
+            self.key_id,
+            self.fee_token,
+        ))
```
mattsse: hmm unsure if this is the optimal representation because now we operate on larger keys, and always need to create a new SpendingLimitUpdate, we could consider using keyed by account only and a list for key id and fee token

crates/transaction-pool/src/transaction.rs
```diff
@@ -865,3 +897,88 @@ mod tests {
...
+    /// The fee token used by this transaction.
+    pub fee_token: Address,
+}
+
+impl KeychainSubject {
+    /// Returns true if this subject matches any of the revoked keys.
+    ///
+    /// Uses a pre-built set for O(1) lookup instead of linear scan.
+    pub fn matches_revoked(&self, revoked_keys: &HashSet<RevokedKey>) -> bool {
+        revoked_keys.contains(&RevokedKey::new(self.account, self.key_id))
```
mattsse: same here

crates/transaction-pool/src/transaction.rs
fgimenez: good point, added `SpendingLimitUpdates` for this

crates/transaction-pool/src/transaction.rs
fgimenez: yep, added similar `RevokedKeys`

PR #2098, opened 2026-01-15
fix(txpool): evict transactions when fee payer is blacklisted

gakonst:
Closes CHAIN-443

## Summary

Extends the pool maintenance task to evict transactions when a fee payer is added to a TIP403 blacklist.

## Problem

There's a TOCTOU vulnerability where:
1. Transactions are submitted from an address that passes blacklist validation
2. A subsequent transaction blacklists the sender via `modifyPolicyBlacklist()`
3. Pending transactions fail at execution but remain stuck in the mempool

## Solution

- Track fee payer -> transaction hash mappings in the maintenance task
- Listen for `BlacklistUpdated` events from the TIP403 registry
- When a fee payer is blacklisted, evict all their pending transactions

## Testing

Added integration test `test_tip403_blacklist_evicts_fee_payer_transactions` that verifies transactions are correctly evicted when the fee payer is added to a blacklist.

Stacked on #2035

### DIFF
--- a/crates/commonware-node/src/feed/state.rs
+++ b/crates/commonware-node/src/feed/state.rs
@@ -432,7 +432,7 @@ impl ConsensusFeed for FeedStateHandle {
                                 merged.push(t.clone());
                             }
                         }
-                        merged.sort_by(|a, b| b.transition_epoch.cmp(&a.transition_epoch));
+                        merged.sort_by_key(|t| std::cmp::Reverse(t.transition_epoch));
                         Arc::new(merged)
                     } else {
                         Arc::new(transitions.clone())


crates/commonware-node/src/feed/state.rs
```diff
@@ -432,7 +432,7 @@ impl ConsensusFeed for FeedStateHandle {
                                 merged.push(t.clone());
                             }
                         }
-                        merged.sort_by(|a, b| b.transition_epoch.cmp(&a.transition_epoch));
+                        merged.sort_by_key(|b| std::cmp::Reverse(b.transition_epoch));
```
mattsse: can we do these separately, they dont belong in here and would likely just cause conflicts

crates/transaction-pool/src/maintain.rs
```diff
@@ -71,50 +174,108 @@ impl TempoPoolState {
...
+            Some(event) = tx_events.next() => {
+                let tx_hash = match event {
+                    FullTransactionEvent::Mined { tx_hash, .. }
+                    | FullTransactionEvent::Discarded(tx_hash)
+                    | FullTransactionEvent::Invalid(tx_hash) => tx_hash,
+                    FullTransactionEvent::Replaced { transaction, .. } => *transaction.hash(),
+                    // Pending/Queued/Propagated don't indicate removal
+                    _ => continue,
+                };
+                state.remove_tx(&tx_hash);
```
mattsse: I think we had the same thing at some point, we should not do this like this because tracking individual transactions like this is just too expensive

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -38,6 +85,9 @@ pub struct TempoTransactionPool<Client> {
     >,
     /// Minimal pool for 2D nonces (nonce_key > 0)
     aa_2d_pool: Arc<RwLock<AA2dPool>>,
+    /// Test synchronization for the maintenance task.
+    #[cfg(feature = "test-utils")]
+    maintenance_sync: Arc<MaintenanceTestSync>,
```
mattsse: didnt we remove this just recently?

crates/commonware-node/src/feed/state.rs
fgimenez: sure proposed here https://github.com/tempoxyz/tempo/pull/2244 ptal

PR #2110, opened 2026-01-16
fix(tx-pool): enforce queued txs limit in 2D pool

0xrusowsky:
audit: `CHAIN-471`

## Motivation
queued transactions (with nonce gaps) do not trigger `discard()`, allowing attackers to submit unlimited transactions with far-future nonces to take pool space and exhaust node memory.

## Solution
- split `aa_2d_limit` into separate `pending_limit` and `queued_limit` for independent enforcement
- call `discard()` when adding queued transactions

### DIFF
--- a/crates/transaction-pool/src/tt_2d_pool.rs
+++ b/crates/transaction-pool/src/tt_2d_pool.rs
@@ -781,7 +781,11 @@ impl AA2dPool {
                     next_nonce = next_nonce.saturating_add(1);
                 } else {
                     // Gap detected - mark this and all remaining transactions as non-pending
-                    existing_tx.is_pending = false;
+                    // Update counters if tx was previously pending
+                    if std::mem::replace(&mut existing_tx.is_pending, false) {
+                        self.pending_count -= 1;
+                        self.queued_count += 1;
+                    }
                 }
             }
 


crates/transaction-pool/src/tt_2d_pool.rs
```diff
@@ -165,7 +171,13 @@ impl AA2dPool {
                     ));
                 }
 
-                Some(entry.insert(tx.clone()))
+                let old_tx = entry.insert(tx.clone());
+                if old_tx.is_pending {
+                    self.pending_count = self.pending_count.saturating_sub(1);
+                } else {
+                    self.queued_count = self.queued_count.saturating_sub(1);
+                }
```
mattsse: we can use unchecked math here because the the invariant is that the txs mirror the counter

crates/transaction-pool/src/tt_2d_pool.rs
```diff
@@ -239,6 +261,12 @@ impl AA2dPool {
             }));
         }
 
+        // Increment queued counter for the newly inserted transaction
+        self.queued_count += 1;
+
+        // Call discard for queued transactions too
+        let _ = self.discard();
```
mattsse, line 284: this is fine here because we also do it in the pending branch, but ideally we have a way to bubble them up as well, tho it's not strictly necessary

crates/transaction-pool/src/tt_2d_pool.rs
```diff
@@ -646,16 +689,27 @@ impl AA2dPool {
...
+        }
+
+        // Evict pending if over pending limit
+        while self.pending_count > self.config.pending_limit.max_txs {
+            if let Some(tx) = self.evict_one(true) {
+                removed.push(tx);
+            } else {
+                break;
+            }
         }
```
mattsse, line 841: this is currently appropriate because we call this after each txs, so wont have the case where we exceed all limits

if that were the case we should iterate only once over the set and check both limits but crurrently this is fine

crates/transaction-pool/src/tt_2d_pool.rs
0xrusowsky: addressed in https://github.com/tempoxyz/tempo/pull/2110/commits/fb3377cb8dbbd0a7035685d61957ed0a363d2b60

PR #2121, opened 2026-01-16
feat: expiring nonces

legion2002:
Adds expiring nonces to tempo , to prevent state bloat of storing old unused nonce keys forever.

Complete description of feature added as [TIP 1009](docs/pages/protocol/tips/tip-1009.mdx)

We gate the relevant changes behind the T1 hardfork.

### DIFF
--- a/crates/revm/src/handler.rs
+++ b/crates/revm/src/handler.rs
@@ -880,11 +880,18 @@ where
                     if !is_authorizing_this_key {
                         // Validate that user_address has authorized this access key in the keychain
                         let user_address = &keychain_sig.user_address;
+
+                        // Extract the signature type from the inner signature to validate it matches
+                        // the key_type stored in the keychain. This prevents using a signature of one
+                        // type to authenticate as a key registered with a different type.
+                        let sig_type: u8 = keychain_sig.signature.signature_type().into();
+
                         keychain
                             .validate_keychain_authorization(
                                 *user_address,
                                 access_key_addr,
                                 block.timestamp().to::<u64>(),
+                                sig_type,
                             )
                             .map_err(|e| TempoInvalidTransaction::KeychainValidationFailed {
                                 reason: format!("{e:?}"),
--- a/crates/contracts/src/precompiles/account_keychain.rs
+++ b/crates/contracts/src/precompiles/account_keychain.rs
@@ -103,10 +103,16 @@ crate::sol! {
         error ZeroPublicKey();
         error ExpiryInPast();
         error KeyAlreadyRevoked();
+        error SignatureTypeMismatch(uint8 expected, uint8 actual);
     }
 }
 
 impl AccountKeychainError {
+    /// Creates an error for signature type mismatch.
+    pub const fn signature_type_mismatch(expected: u8, actual: u8) -> Self {
+        Self::SignatureTypeMismatch(IAccountKeychain::SignatureTypeMismatch { expected, actual })
+    }
+
     /// Creates an error for unauthorized caller.
     pub const fn unauthorized_caller() -> Self {
         Self::UnauthorizedCaller(IAccountKeychain::UnauthorizedCaller {})
--- a/crates/precompiles/src/account_keychain/mod.rs
+++ b/crates/precompiles/src/account_keychain/mod.rs
@@ -371,20 +371,36 @@ impl AccountKeychain {
         Ok(key)
     }
 
-    /// Validate keychain authorization (existence, revocation, and expiry)
+    /// Validate keychain authorization (existence, revocation, expiry, and signature type)
     ///
     /// This consolidates all validation checks into one method.
     /// Returns Ok(()) if the key is valid and authorized, Err otherwise.
+    ///
+    /// # Arguments
+    /// * `account` - The account that owns the key
+    /// * `key_id` - The key identifier to validate
+    /// * `current_timestamp` - Current block timestamp for expiry check
+    /// * `expected_sig_type` - The signature type from the actual signature (0=Secp256k1, 1=P256, 2=WebAuthn)
     pub fn validate_keychain_authorization(
         &self,
         account: Address,
         key_id: Address,
         current_timestamp: u64,
+        expected_sig_type: u8,
     ) -> Result<()> {
         let key = self.load_active_key(account, key_id)?;
 
         if current_timestamp >= key.expiry {
             return Err(AccountKeychainError::key_expired().into());
+        }
+
+        // Validate that the signature type matches the key type stored in the keychain
+        if key.signature_type != expected_sig_type {
+            return Err(AccountKeychainError::signature_type_mismatch(
+                key.signature_type,
+                expected_sig_type,
+            )
+            .into());
         }
 
         Ok(())
@@ -1069,5 +1085,61 @@ mod tests {
         let encoded = revoked.encode_to_slot();
         let decoded = AuthorizedKey::decode_from_slot(encoded);
         assert_eq!(decoded, revoked);
+    }
+
+    #[test]
+    fn test_validate_keychain_authorization_checks_signature_type() -> eyre::Result<()> {
+        let mut storage = HashMapStorageProvider::new(1);
+        let account = Address::random();
+        let key_id = Address::random();
+        StorageCtx::enter(&mut storage, || {
+            let mut keychain = AccountKeychain::new();
+            keychain.initialize()?;
+
+            // Use main key for authorization
+            keychain.set_transaction_key(Address::ZERO)?;
+
+            // Authorize a P256 key
+            let auth_call = authorizeKeyCall {
+                keyId: key_id,
+                signatureType: SignatureType::P256,
+                expiry: u64::MAX,
+                enforceLimits: false,
+                limits: vec![],
+            };
+            keychain.authorize_key(account, auth_call)?;
+
+            // Test 1: Validation should succeed with matching signature type (P256 = 1)
+            let result = keychain.validate_keychain_authorization(account, key_id, 0, 1);
+            assert!(
+                result.is_ok(),
+                "Validation should succeed with matching signature type"
+            );
+
+            // Test 2: Validation should fail with mismatched signature type (Secp256k1 = 0)
+            let mismatch_result = keychain.validate_keychain_authorization(account, key_id, 0, 0);
+            assert!(
+                mismatch_result.is_err(),
+                "Validation should fail with mismatched signature type"
+            );
+            match mismatch_result.unwrap_err() {
+                TempoPrecompileError::AccountKeychainError(e) => {
+                    assert!(
+                        matches!(e, AccountKeychainError::SignatureTypeMismatch(_)),
+                        "Expected SignatureTypeMismatch error, got: {e:?}"
+                    );
+                }
+                e => panic!("Expected AccountKeychainError, got: {e:?}"),
+            }
+
+            // Test 3: Validation should fail with WebAuthn (2) when key is P256 (1)
+            let webauthn_mismatch = keychain.validate_keychain_authorization(account, key_id, 0, 2);
+            assert!(
+                webauthn_mismatch.is_err(),
+                "Validation should fail with WebAuthn when key is P256"
+            );
+
+            Ok(())
+        })
     }
 }


docs/specs/src/TIP-EXPIRING-NONCE.md
```diff
@@ -0,0 +1,270 @@
+# TIP: Expiring Nonces
```
jenpaff: please pick the next `TIP-1008` & add the number here

jenpaff: move the TIP to https://github.com/tempoxyz/tempo/blob/464c119fa3b9ad790bbbb3eabd059895fb25ce7f/docs/pages/protocol/tips/tip-1007.mdx

legion2002: Yup done

crates/transaction-pool/src/tt_2d_pool.rs
```diff
@@ -49,6 +49,9 @@ pub struct AA2dPool {
     by_id: BTreeMap<AA2dTransactionId, AA2dInternalTransaction>,
     /// _All_ transactions by hash.
     by_hash: HashMap<TxHash, Arc<ValidPoolTransaction<TempoPooledTransaction>>>,
+    /// Expiring nonce transactions, keyed by tx hash (always pending/independent).
+    /// These use tx hash for replay protection instead of sequential nonces.
+    expiring_nonce_txs: HashMap<TxHash, PendingTransaction<Ordering>>,
```
klkvr, line 54: do we ever remove transactions from here on inclusion?

crates/transaction-pool/src/tt_2d_pool.rs
klkvr, line 1: idk if it really matters but `senders_iter` and `descendant_txs_mut` and probably some other getters don't account for expiring txs

generally it feels like it could've been easier to deal with those txs if they actually had random nonces vs a same one for each tx

crates/transaction-pool/src/tt_2d_pool.rs
legion2002: No we don't do it on inclusion, they just get evicted on expiry (which is 30 sec right now)

Zygimantass (approved):

Zygimantass (approved):

crates/transaction-pool/src/tt_2d_pool.rs
klkvr: we should remove on inclusion as well, otherwise builder will have to always iterate through them and discard which is wasteful

crates/commonware-node/src/epoch/manager/actor.rs
```diff
@@ -322,7 +323,9 @@ where
         self.config.scheme_provider.register(epoch, scheme.clone());
 
         let engine = simplex::Engine::new(
-            self.context.with_label(&format!("simplex_epoch_{epoch}")),
+            self.context
+                .with_label("simplex")
+                .with_attribute("epoch", epoch),
```
klkvr, line 328: need to get rid of unrelated changes, probably same rebase-related issues

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -172,10 +183,23 @@ where
                         authority_ids: authorities
                             .map(|auths| self.protocol_pool.inner().get_sender_ids(auths)),
                     };
+
+                    // Check if T1 hardfork is active for expiring nonce handling
+                    use tempo_chainspec::hardfork::TempoHardforks;
```
klkvr: please move import to the top of file

crates/revm/src/handler.rs
```diff
@@ -624,19 +637,60 @@ where
...
+
+            StorageCtx::enter_evm(journal, block, cfg, || {
+                let mut nonce_manager = NonceManager::new();
+
+                nonce_manager
+                    .check_and_mark_expiring_nonce(
+                        tx_hash,
+                        valid_before,
+                        now,
+                        TEMPO_EXPIRING_NONCE_MAX_EXPIRY_SECS,
```
klkvr: afaict precompile has access to both block timestamp and `TEMPO_EXPIRING_NONCE_MAX_EXPIRY_SECS` so we can simplify this?

klkvr (approved):
looks good to me, just have nits

crates/node/tests/it/tempo_transaction.rs
```diff
@@ -5896,3 +5896,537 @@ async fn test_aa_keychain_revocation_toctou_dos() -> eyre::Result<()> {
...
+        .raw_request("eth_getTransactionReceipt".into(), [tx_hash])
+        .await?;
+
+    assert!(raw_receipt.is_some(), "Transaction receipt should exist");
+    let receipt = raw_receipt.unwrap();
+    let status = receipt["status"]
+        .as_str()
+        .map(|s| s == "0x1")
+        .unwrap_or(false);
+    assert!(status, "Transaction should succeed");
```
0xKitsune, line 5984: There is a helper function for this already called `await_receipts`.

crates/node/tests/it/tempo_transaction.rs
```diff
@@ -5896,3 +5896,537 @@ async fn test_aa_keychain_revocation_toctou_dos() -> eyre::Result<()> {
...
+    );
+    println!("✓ Protocol nonce unchanged (still 0)");
+
+    Ok(())
+}
+
+/// Test expiring nonce replay protection - same tx hash should be rejected
+#[tokio::test(flavor = "multi_thread")]
+async fn test_aa_expiring_nonce_replay_protection() -> eyre::Result<()> {
+    use tempo_primitives::transaction::TEMPO_EXPIRING_NONCE_KEY;
```
0xKitsune: nit: lets move imports to the top of the file.

crates/node/tests/it/tempo_transaction.rs
```diff
@@ -5896,3 +5896,537 @@ async fn test_aa_keychain_revocation_toctou_dos() -> eyre::Result<()> {
...
+
+        // This should be rejected at pool or execution level
+        let inject_result = setup
+            .node
+            .rpc
+            .inject_tx(envelope.encoded_2718().into())
+            .await;
+
+        if inject_result.is_err() {
+            println!("✓ valid_before = now + 31s rejected at pool level");
```
0xKitsune: We should expect the exact error, this might pass as a false positive due to some change in the future.

crates/node/tests/it/tempo_transaction.rs
```diff
@@ -5896,3 +5896,537 @@ async fn test_aa_keychain_revocation_toctou_dos() -> eyre::Result<()> {
...
+            let raw_receipt: Option<serde_json::Value> = provider
+                .raw_request("eth_getTransactionReceipt".into(), [tx_hash])
+                .await?;
+            let status = raw_receipt
+                .as_ref()
+                .and_then(|r| r["status"].as_str())
+                .map(|s| s == "0x1")
+                .unwrap_or(false);
+            if raw_receipt.is_none() || !status {
+                println!("✓ valid_before = now + 31s rejected at execution level");
```
0xKitsune: This could pass as a false postive if not included in the block for a different reason. We should rely on `!status` here unless there is something that might cause the tx not to land.

crates/node/tests/it/tempo_transaction.rs
```diff
@@ -5896,3 +5896,537 @@ async fn test_aa_keychain_revocation_toctou_dos() -> eyre::Result<()> {
...
+    }
+
+    println!("✓ Replay protection working");
+
+    Ok(())
+}
+
+/// Test expiring nonce validity window - reject transactions outside the valid window
+#[tokio::test(flavor = "multi_thread")]
+async fn test_aa_expiring_nonce_validity_window() -> eyre::Result<()> {
```
0xKitsune, line 6085: Does this need to be an integration test or are these same assertions covered via unit tests?

crates/primitives/src/transaction/tempo_transaction.rs
```diff
@@ -270,11 +276,32 @@ impl TempoTransaction {
...
+        // Validate expiring nonce transaction constraints
+        if self.is_expiring_nonce_tx() {
+            // Expiring nonce txs must have nonce == 0
+            if self.nonce != 0 {
+                return Err("expiring nonce transactions must have nonce == 0");
+            }
+            // Expiring nonce txs must have valid_before set
+            if self.valid_before.is_none() {
+                return Err("expiring nonce transactions must have valid_before set");
+            }
```
0xKitsune, line 302: Just to clarify, do we want to accept txs that specify an expiring nonce with a valid before that is greater than the max allowed amount?

crates/transaction-pool/src/maintain.rs
```diff
@@ -154,7 +155,17 @@ where
...
+                // Expiring nonce txs use tx hash for replay protection rather than sequential nonces,
+                // so we need to remove them on inclusion rather than relying on nonce changes.
+                let mined_tx_hashes: Vec<_> = tip
+                    .blocks_iter()
+                    .flat_map(|block| block.body().transactions())
+                    .map(|tx| *tx.tx_hash())
+                    .collect();
+                pool.remove_included_expiring_nonce_txs(mined_tx_hashes.iter());
+
+                // 5. Update AMM liquidity cache
```
0xKitsune: IIUC we also need to remove txs where the `valid_before` has expired but the tx has not been inlcuded in the pool. This may already be covered though.

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -172,10 +185,25 @@ where
...
+                        .tip_timestamp();
+                    let is_t1_active = self
+                        .client()
+                        .chain_spec()
+                        .is_t1_active_at_timestamp(tip_timestamp);
+
+                    let added = self.aa_2d_pool.write().add_transaction(
+                        Arc::new(tx),
+                        state_nonce,
+                        is_t1_active,
```
0xKitsune: Hardfork gating this will become unnecessary once the hardfork is active. If we absolutely need this, we should make sure we are tracking to remove this later. Also lets not pass in a bool for the hardfork into `add_transaction`, but rather pass in the SpecId if necessary to check the fork.

crates/node/tests/it/tempo_transaction.rs
legion2002: I thought this was convention for tests, but will move.

crates/transaction-pool/src/tt_2d_pool.rs
```diff
@@ -98,10 +102,15 @@ impl AA2dPool {
...
     /// to the transaction's nonce key.
+    ///
+    /// `is_t1_active` indicates whether the T1 hardfork is active. If true, expiring nonce
+    /// transactions (nonce_key == U256::MAX) are handled specially. If false, they are
+    /// treated as regular 2D nonce transactions.
     pub(crate) fn add_transaction(
         &mut self,
         transaction: Arc<ValidPoolTransaction<TempoPooledTransaction>>,
         on_chain_nonce: u64,
+        is_t1_active: bool,
```
0xKitsune: We should pass in the SpecId rather than `is_t1_active`. Also note that we should remove this once the hardfork is active. Can we open an issue to track?

crates/transaction-pool/src/validator.rs
```diff
@@ -495,29 +520,55 @@ where
...
-                    state_nonce = match state_provider.storage(
-                        NONCE_PRECOMPILE_ADDRESS,
-                        transaction.transaction().nonce_key_slot().unwrap().into(),
-                    ) {
-                        Ok(nonce) => nonce.unwrap_or_default().saturating_to(),
-                        Err(err) => {
-                            return TransactionValidationOutcome::Error(
-                                *transaction.hash(),
-                                Box::new(err),
+                    if nonce_key == TEMPO_EXPIRING_NONCE_KEY {
```
0xKitsune: We are hardfork gating elsewhere in validation, do we need to do this here as well? Otherwise is the `expiring_seen_slot` active?

crates/node/tests/it/tempo_transaction.rs
legion2002: We can't use `await_receipts` here because Tempo AA transactions aren't supported by alloy's `PendingTransactionBuilder`. We're using raw RPC requests (`eth_getTransactionReceipt`) to handle the custom tx type.

crates/node/tests/it/tempo_transaction.rs
legion2002: This is an integration test because it verifies end-to-end behavior: pool rejection at validation → execution-level revert → block building exclusion. The validator has unit tests for the individual checks, but this confirms the full stack works correctly together.

crates/primitives/src/transaction/tempo_transaction.rs
legion2002: The primitives `validate()` doesn't check max expiry because that's enforced at the pool level (see validator.rs:218-227). Consensus rules only care that `valid_before > valid_after`; the max window is a DoS protection in the mempool.

crates/transaction-pool/src/maintain.rs
legion2002: This is covered by the expiry tracking in `maintain_tempo_pool()` (maintain.rs:115-129). The `TempoPoolState` tracks all AA txs with `valid_before` and evicts them when `tip_timestamp >= valid_before`.

crates/transaction-pool/src/tt_2d_pool.rs
legion2002: Yup added issue here 
https://github.com/tempoxyz/tempo/issues/2187

crates/transaction-pool/src/validator.rs
legion2002: Yes good catch, this should be hardfork gated too.
Updated.

crates/node/tests/it/tempo_transaction.rs
0xKitsune: Sounds good, consider creating a separate helper like `await_aa_tx` or similar as it looks like this is used frequently.

crates/node/tests/it/tempo_transaction.rs
```diff
@@ -5896,3 +5897,516 @@ async fn test_aa_keychain_revocation_toctou_dos() -> eyre::Result<()> {
...
+    if replay_result.is_err() {
+        println!("✓ Replay rejected at transaction pool level");
+    } else {
+        // If it was accepted to pool, mine a block and check it wasn't included
+        setup.node.advance_block().await?;
+
+        // The tx hash is the same, so we can't distinguish - but if we check
+        // that trying to query it again doesn't show a new block number, we're good
+        println!("✓ Replay may have been rejected at block building level");
+    }
```
0xKitsune: Is there a reason why we don't assert one, but allow both here? Why might this pass at the validation level but fail block building? This also applies to other instances in the tests where we assert this.

crates/precompiles/src/nonce/mod.rs
```diff
@@ -9,26 +9,42 @@ use crate::{
     error::Result,
     storage::{Handler, Mapping},
 };
-use alloy::primitives::{Address, U256};
+use alloy::primitives::{Address, B256, U256};
+
+/// Capacity of the expiring nonce seen set (supports 10k TPS for 30 seconds).
+pub const EXPIRING_NONCE_SET_CAPACITY: u32 = 300_000;
```
0xKitsune, line 15: This is quite low, I imagine that someone would be able to DoS this and fill this quite quickly?

crates/precompiles/src/nonce/mod.rs
```diff
@@ -71,6 +87,81 @@ impl NonceManager {
 
         Ok(new_nonce)
     }
+
+    // ========== Expiring Nonce Methods ==========
```
0xKitsune: nit: lets remove this IMO

crates/revm/src/handler.rs
```diff
@@ -73,6 +74,17 @@ pub const EXISTING_NONCE_KEY_GAS: u64 = COLD_SLOAD_COST + WARM_SSTORE_RESET;
 /// Gas cost for using a new 2D nonce key (cold SLOAD + SSTORE set for 0 -> non-zero)
 pub const NEW_NONCE_KEY_GAS: u64 = COLD_SLOAD_COST + SSTORE_SET;
 
+/// Gas cost for expiring nonce transactions (replay check + insert).
+///
+/// Operations in `check_and_mark_expiring_nonce`:
+/// - 2 cold SLOADs: `seen[tx_hash]`, `ring[idx]`
+/// - 3 warm SSTOREs: `seen[old_hash]=0`, `ring[idx]`, `seen[tx_hash]`
```
0xKitsune: is `seen[old_hash]=0` warm? If this hasnt been called within the current context wouldnt this be cold?

crates/transaction-pool/src/validator.rs
```diff
@@ -495,29 +520,61 @@ where
...
+
+                        // If expiry is non-zero and in the future, tx hash is still valid (replay)
+                        if seen_expiry != 0 && seen_expiry > current_time {
+                            return TransactionValidationOutcome::Invalid(
+                                transaction.into_transaction(),
+                                InvalidPoolTransactionError::other(
+                                    TempoPoolTransactionError::ExpiringNonceReplay,
+                                ),
                             );
                         }
```
0xKitsune: Are we also enforcing this at the protocol level or just the validation level? Is it possible to invalidate a temporary nonce if a tx is injected directly into the block without validation?

xtask/src/genesis_args.rs
```diff
@@ -481,6 +481,9 @@ impl GenesisArgs {
         chain_config
             .extra_fields
             .insert_value("t0Time".to_string(), 0u64)?;
+        chain_config
+            .extra_fields
+            .insert_value("t1Time".to_string(), 0u64)?;
```
0xKitsune: We should not hardcode `t1Time` and `t0Time` to 0 here. Rather we should pass these in as cli args and we can set the default to 0.

gakonst (commented):
# PR Review: Expiring Nonces (TIP-1009)

I've completed a detailed review of this PR using both code analysis and cross-referencing against the spec. Below are the **verifiable issues/bugs** and **fundamental implementation problems** found. No stylistic comments are included.

---

## 🔴 Critical Issues

### 1. Ring Buffer Pointer Wrap Bug (Liveness Issue)

**File:** `crates/precompiles/src/nonce/mod.rs` (lines 139-161)

**Problem:** The ring pointer is stored as `u32` and advanced using `wrapping_add(1)`, which wraps at `u32::MAX` (~4.3B), not at `CAPACITY` (300k). When the pointer wraps from `u32::MAX` to `0`, the index calculation `ptr % CAPACITY` suddenly jumps from one position to `0`, causing early slots to be revisited **much sooner than CAPACITY inserts** - potentially before their `valid_before` has expired.

**Impact:** After ~4.3B expiring nonce transactions, the system will start rejecting valid transactions with `ExpiringNonceSetFull` even at normal TPS levels, causing a liveness failure.

**Fix:** Either:
1. Store pointer mod capacity and update to `next = if ptr + 1 == CAPACITY { 0 } else { ptr + 1 }`
2. Or change to `u64` to avoid wrap in any realistic timeframe

---

### 2. Intrinsic Gas Severely Under-estimated (OOG Risk)

**File:** `crates/revm/src/handler.rs` (lines 77-86)

**Current calculation:**
```rust
pub const EXPIRING_NONCE_GAS: u64 = 2 * COLD_SLOAD_COST + 3 * WARM_SSTORE_RESET;
                                  = 2 * 2100 + 3 * 2900 = 12,900 gas
```

**Actual operations in `check_and_mark_expiring_nonce`:**
- `ring_ptr` SLOAD (cold) - **NOT excluded, warmth resets per transaction**
- `seen[tx_hash]` SLOAD (cold)
- `ring[idx]` SLOAD (cold)  
- `seen[old_hash]` SLOAD (cold, when old_hash != 0)
- `seen[old_hash] = 0` SSTORE (warm reset)
- `ring[idx] = tx_hash` SSTORE (warm reset)
- `seen[tx_hash] = valid_before` SSTORE - **This is 0→non-zero = SSTORE_SET (20,000 gas), not WARM_SSTORE_RESET!**
- `ring_ptr = next` SSTORE (warm reset)

**Correct calculation (worst case):**
```rust
// 4 cold SLOADs + 1 SSTORE_SET + 3 WARM_SSTORE_RESET
pub const EXPIRING_NONCE_GAS: u64 = 4 * 2100 + 20000 + 3 * 2900 = 37,100 gas
```

**Impact:** Transactions can be admitted to the pool and included in blocks but then fail with OOG during execution, wasting pool/builder effort and causing failed transactions.

---

## 🟡 Pool Validation Gap

### 3. Missing `nonce == 0` Check in Transaction Pool

**File:** `crates/transaction-pool/src/validator.rs` (lines 193-228)

**Problem:** The pool validator checks for expiring nonce replay and `valid_before` constraints, but does **not** reject expiring nonce transactions with `nonce != 0`. The handler rejects these later (`TempoInvalidTransaction::ExpiringNonceNonceNotZero`), but without pool-level rejection, invalid transactions can enter and propagate through the mempool.

**Fix:** Add to `ensure_valid_conditionals`:
```rust
// Expiring nonce transactions MUST have nonce == 0
if is_expiring_nonce && tx.nonce != 0 {
    return Err(TempoPoolTransactionError::ExpiringNonceNonceNotZero);
}
```

---

## 🟠 Test Coverage Gaps (New Behaviors Not Tested)

The following new behaviors are introduced but lack unit or integration tests:

1. **Ring buffer wrap at u32::MAX** - No test verifies correct behavior when pointer wraps
2. **Buffer wrap at CAPACITY** - No test inserts CAPACITY+N entries to verify eviction across wrap
3. **`ExpiringNonceSetFull` error path** - No test drives conditions where an unexpired entry is encountered
4. **Pool rejection of `nonce != 0`** - Currently not tested because it's not implemented at pool level
5. **Replay after expiry** - Precompile unit tests check `is_expiring_nonce_seen` becomes false after expiry, but don't test that re-submitting the same hash *after expiry* is accepted and correctly updates state
6. **Post-expiry slot reuse** - No test verifies the same tx hash can be used again after its expiry window passes and it's evicted from the buffer

crates/precompiles/src/nonce/mod.rs
```diff
@@ -71,6 +87,81 @@ impl NonceManager {
...
+            // Clear the old entry from seen set
+            self.expiring_nonce_seen[old_hash].write(0)?;
+        }
+
+        // 5. Insert new entry
+        self.expiring_nonce_ring[idx].write(tx_hash)?;
+        self.expiring_nonce_seen[tx_hash].write(valid_before)?;
+
+        // 6. Advance pointer (wraps automatically via modulo on next read)
+        self.expiring_nonce_ring_ptr.write(ptr.wrapping_add(1))?;
```
gakonst: **Bug: Pointer wrap at u32::MAX breaks buffer invariant**

The pointer wraps at `u32::MAX` (~4.3B), not at `CAPACITY` (300k). When it wraps from `u32::MAX` to `0`, the `idx = ptr % CAPACITY` calculation suddenly jumps, causing early slots to be revisited before their `valid_before` has expired.

**Fix:** Store pointer mod capacity:
```rust
let next = if ptr + 1 == EXPIRING_NONCE_SET_CAPACITY { 0 } else { ptr + 1 };
self.expiring_nonce_ring_ptr.write(next)?;
```

crates/revm/src/handler.rs
```diff
@@ -73,6 +74,17 @@ pub const EXISTING_NONCE_KEY_GAS: u64 = COLD_SLOAD_COST + WARM_SSTORE_RESET;
...
+///
+/// Note: ring_ptr SLOAD/SSTORE excluded since it's cached (accessed almost every tx).
+/// Note: `seen[old_hash]` SLOAD excluded - buffer sized so entries are always expired.
+/// After cold SLOADs, slots are warm so SSTOREs use WARM_SSTORE_RESET.
+pub const EXPIRING_NONCE_GAS: u64 = 2 * COLD_SLOAD_COST + 3 * WARM_SSTORE_RESET;
```
gakonst: **Bug: Gas severely underestimated**

1. `ring_ptr` SLOAD/SSTORE are NOT cached across transactions—warmth resets per tx
2. `seen[tx_hash]` is a 0→non-zero write = `SSTORE_SET` (20,000 gas), not `WARM_SSTORE_RESET` (2,900 gas)

**Actual worst-case:**
- 4 cold SLOADs: `ring_ptr`, `seen[tx_hash]`, `ring[idx]`, `seen[old_hash]`
- 1 SSTORE_SET: `seen[tx_hash] = valid_before` (new entry)
- 3 WARM_SSTORE_RESET: `seen[old_hash]=0`, `ring[idx]`, `ring_ptr`

```rust
pub const EXPIRING_NONCE_GAS: u64 = 4 * COLD_SLOAD_COST + SSTORE_SET + 3 * WARM_SSTORE_RESET;
// = 4 * 2100 + 20000 + 3 * 2900 = 37,100 gas
```

crates/transaction-pool/src/validator.rs
```diff
@@ -189,17 +193,38 @@ where
...
     ) -> Result<(), TempoPoolTransactionError> {
+        let current_time = self.inner.fork_tracker().tip_timestamp();
+
+        // Check if T1 is active for expiring nonce specific validations
+        let spec = self.inner.chain_spec().tempo_hardfork_at(current_time);
+        let is_expiring_nonce = tx.is_expiring_nonce_tx() && spec.is_t1();
+
+        // Expiring nonce transactions MUST have valid_before set
+        if is_expiring_nonce && tx.valid_before.is_none() {
+            return Err(TempoPoolTransactionError::ExpiringNonceMissingValidBefore);
```
gakonst, line 210: **Missing validation: `nonce == 0` check**

Expiring nonce transactions MUST have `nonce == 0` (enforced in handler), but this isn't validated at pool level. Invalid txs with `nonce != 0` can enter/propagate through the mempool.

**Add:**
```rust
// Expiring nonce transactions MUST have nonce == 0
if is_expiring_nonce && tx.nonce != 0 {
    return Err(TempoPoolTransactionError::ExpiringNonceNonceNotZero);
}
```

0xKitsune (approved):

0xKitsune:
Left a few more comments, but overall lgtm pending resolving over comments above.

crates/precompiles/src/nonce/mod.rs
legion2002: It is supposed to be a ring buffer. So the efficiency gains are actually realized once it is filled, and then you start replacing the old txHashes.
So we want this to fill in a reasonable time.

crates/revm/src/handler.rs
legion2002: We make some concessions for the buffer, because we know it is guaranteed to have a cache hit in a reasonable time. If this is not guaranteed, we will do this as a perf optimization later on.
Basically the ring pointer will always be used for each expiring tx. And the buffer will ideally always be kept in cache, because it should be touched once every 30 seconds to a minute.

xtask/src/genesis_args.rs
legion2002: updated.

crates/transaction-pool/src/validator.rs
legion2002: Yes, it is enforced at the protocol level. The handler calls check_and_mark_expiring_nonce and checks this during execution to prevent replay.

crates/transaction-pool/src/validator.rs
legion2002: Updated.

crates/revm/src/handler.rs
legion2002: This is not correct. We make some concessions because we know that some storage for ring buffers will always remain hot.
Explained more in the spec, and here - 
https://github.com/tempoxyz/tempo/pull/2121#discussion_r2712107434

crates/precompiles/src/nonce/mod.rs
legion2002: This has been updated, good catch!

crates/node/tests/it/tempo_transaction.rs
rakita: This one fails on tip1000 pR

PR #2150, opened 2026-01-17
docs: recommend reusing nonce keys instead of random values

dankrad:
With TIP-1000 state growth spam protection, creating new nonce keys incurs increasing gas costs.

## Changes
- Updated examples to use sequential nonce keys (`1n`, `2n`) instead of arbitrary values (`567n`, `789n`)
- Added a warning in the concurrent transactions section explaining the cost implications of creating new nonce keys

## Context
Using random nonce keys for each transaction is no longer recommended because TIP-1000 charges progressively higher gas for each new nonce key created. Users should reuse a small set of sequential nonce keys instead.


dankrad:
Outdated as we are now going to recommend using expiring nonces for these use cases.

PR #2166, opened 2026-01-19
feat: tip1000 

0xKitsune:
This PR implements tip1000 gas repricing.

Originally this PR was merged into `devnet/nonce-expiry-gas-repricing` with #2128 which also includes nonce expiry. This PR takes the gas repricing changes to target main.

### DIFF
--- a/crates/evm/src/evm.rs
+++ b/crates/evm/src/evm.rs
@@ -11,15 +11,14 @@ use alloy_primitives::{Address, Bytes, Log, TxKind};
 use reth_revm::{InspectSystemCallEvm, MainContext, context::result::ExecutionResult};
 use std::ops::{Deref, DerefMut};
 use tempo_chainspec::hardfork::TempoHardfork;
-use tempo_revm::{
-    TempoHaltReason, TempoInvalidTransaction, TempoTxEnv, evm::TempoContext,
-    gas_params::tempo_gas_params,
-};
+use tempo_revm::{TempoHaltReason, TempoInvalidTransaction, TempoTxEnv, evm::TempoContext};
 
 use crate::TempoBlockEnv;
 
 /// TIP-1000 sets transaction gas limit cap to 30 million gas.
 ///
+/// Increase of tx gas limit cap to 30M gas is to allow for maximum-sized
+/// contract deployments under TIP-1000 state creation costs.
 pub const TIP1000_TX_GAS_LIMIT_CAP: u64 = 30_000_000;
 
 #[derive(Debug, Default, Clone, Copy)]
@@ -68,9 +67,7 @@ pub struct TempoEvm<DB: Database, I = NoOpInspector> {
 
 impl<DB: Database> TempoEvm<DB> {
     /// Create a new [`TempoEvm`] instance.
-    pub fn new(db: DB, mut input: EvmEnv<TempoHardfork, TempoBlockEnv>) -> Self {
-        input.cfg_env.gas_params = tempo_gas_params(input.cfg_env.spec);
-
+    pub fn new(db: DB, input: EvmEnv<TempoHardfork, TempoBlockEnv>) -> Self {
         let ctx = Context::mainnet()
             .with_db(db)
             .with_block(input.block_env)
@@ -249,6 +246,7 @@ mod tests {
         database::{EmptyDB, in_memory_db::CacheDB},
     };
     use tempo_chainspec::hardfork::TempoHardfork;
+    use tempo_revm::gas_params::tempo_gas_params;
 
     use super::*;
 
--- a/crates/revm/src/evm.rs
+++ b/crates/revm/src/evm.rs
@@ -35,9 +35,9 @@ pub struct TempoEvm<DB: Database, I> {
     pub logs: Vec<Log>,
     /// The fee collected in `collectFeePreTx` call.
     pub(crate) collected_fee: U256,
-    /// Additional initial gas cost is needed for authorization_key setting in pre execution.
-    pub(crate) additional_initial_gas: u64,
     /// Initial gas cost. Used for key_authorization validation in collectFeePreTx.
+    ///
+    /// Additional initial gas cost is added for authorization_key setting in pre execution.
     pub(crate) initial_gas: u64,
 }
 
@@ -72,7 +72,6 @@ impl<DB: Database, I> TempoEvm<DB, I> {
             inner,
             logs: Vec::new(),
             collected_fee: U256::ZERO,
-            additional_initial_gas: 0,
             initial_gas: 0,
         }
     }
@@ -1463,10 +1462,6 @@ mod tests {
         let evm_with_inspector = evm.with_inspector(CountInspector::new());
 
         // Verify fields are still initialized correctly
-        assert_eq!(
-            evm_with_inspector.additional_initial_gas, 0,
-            "additional_initial_gas should be 0 after with_inspector"
-        );
         assert_eq!(
             evm_with_inspector.initial_gas, 0,
             "initial_gas should be 0 after with_inspector"
--- a/crates/revm/src/handler.rs
+++ b/crates/revm/src/handler.rs
@@ -493,12 +493,7 @@ where
     ) -> Result<FrameResult, Self::Error> {
         // Add key_authorization gas to the initial gas (calculated in validate_aa_initial_tx_gas)
         // Use saturating_add to avoid overflow when additional_initial_gas is u64::MAX (OOG case)
-        let adjusted_gas = InitialAndFloorGas::new(
-            init_and_floor_gas
-                .initial_gas
-                .saturating_add(evm.additional_initial_gas),
-            init_and_floor_gas.floor_gas,
-        );
+        let adjusted_gas = InitialAndFloorGas::new(evm.initial_gas, init_and_floor_gas.floor_gas);
 
         let tx = evm.tx();
         if tx.gas_limit() < adjusted_gas.initial_gas {
@@ -865,12 +860,14 @@ where
                 }
             })?;
 
-            // get how much gas was spent by the precompile
-            evm.additional_initial_gas = if out_of_gas {
-                u64::MAX
-            } else {
-                provider.gas_used()
-            };
+            if spec.is_t1() {
+                // Add additional gas that was spent by the precompile
+                if out_of_gas {
+                    evm.initial_gas = u64::MAX
+                } else {
+                    evm.initial_gas += provider.gas_used()
+                };
+            }
 
             drop(provider);
         }
@@ -953,7 +950,6 @@ where
         // Short-circuit if there is no spending for this transaction and `collectFeePreTx`
         // call will not collect any fees.
         if gas_balance_spending.is_zero() {
-            evm.additional_initial_gas = 0;
             return Ok(());
         }
 
@@ -1178,7 +1174,7 @@ where
             }
 
             // TIP-1000: Storage pricing updates for launch
-            // Tempo transactions with any `nonce_key` and `nonce == 0` require an additional 250,000 gas.
+            // Transactions with any `nonce_key` and `nonce == 0` require an additional 250,000 gas.
             if spec.is_t1() && tx.nonce == 0 {
                 init_gas.initial_gas += gas_params.get(GasId::new_account_cost());
             }
@@ -1505,12 +1501,7 @@ where
     ) -> Result<FrameResult, Self::Error> {
         // Add key_authorization gas to the initial gas (calculated in validate_aa_initial_tx_gas)
         // Use saturating_add to avoid overflow when additional_initial_gas is u64::MAX (OOG case)
-        let adjusted_gas = InitialAndFloorGas::new(
-            init_and_floor_gas
-                .initial_gas
-                .saturating_add(evm.additional_initial_gas),
-            init_and_floor_gas.floor_gas,
-        );
+        let adjusted_gas = InitialAndFloorGas::new(evm.initial_gas, init_and_floor_gas.floor_gas);
 
         let tx = evm.tx();
         if tx.gas_limit() < adjusted_gas.initial_gas {
--- a/xtask/src/genesis_args.rs
+++ b/xtask/src/genesis_args.rs
@@ -492,7 +492,6 @@ impl GenesisArgs {
         chain_config
             .extra_fields
             .insert_value("t1Time".to_string(), self.t1_time)?;
-
         let mut extra_data = Bytes::from_static(b"tempo-genesis");
 
         if let Some(consensus_config) = &consensus_config {
--- a/crates/node/tests/it/tempo_transaction.rs
+++ b/crates/node/tests/it/tempo_transaction.rs
@@ -2756,18 +2756,18 @@ async fn test_aa_estimate_gas_with_keychain_and_key_auth() -> eyre::Result<()> {
     let keychain_gas_u64 = u64::from_str_radix(keychain_gas.trim_start_matches("0x"), 16)?;
     println!("  Keychain gas: {keychain_gas_u64}");
 
-    // Keychain with same-tx auth adds:
+    // Keychain with same-tx auth adds ~289,693 gas which includes:
     // - 3,000 for keychain validation
     // - ~30,000 for KeyAuthorization (27,000 base + 3,000 ecrecover)
-    // Total: ~33,000 gas
+    // - storage costs for key authorization precompile
     let keychain_diff = keychain_gas_u64 as i64 - baseline_gas_u64 as i64;
     assert!(
-        (32_500..=34_000).contains(&keychain_diff.unsigned_abs()),
-        "Keychain + KeyAuth should add ~33,000 gas: actual diff {keychain_diff} (expected 33,000 ±500)"
+        (289_000..=291_000).contains(&keychain_diff.unsigned_abs()),
+        "Keychain + KeyAuth should add ~289,693 gas: actual diff {keychain_diff}"
     );
-    println!("  ✓ Keychain + KeyAuth adds {keychain_diff} gas (expected ~33,000)");
+    println!("  ✓ Keychain + KeyAuth adds {keychain_diff} gas (expected ~289,693)");
 
-    // Test 3: Keychain signature with P256 inner - should add 3,000 + 5,000 + ~30,000 = ~38,000 gas
+    // Test 3: Keychain signature with P256 inner
     println!("\nTest 3: Keychain signature (P256 inner)");
     let key_auth_p256_for_keychain =
         create_signed_key_authorization(&signer, SignatureType::P256, 0);
@@ -2789,17 +2789,17 @@ async fn test_aa_estimate_gas_with_keychain_and_key_auth() -> eyre::Result<()> {
         u64::from_str_radix(keychain_p256_gas.trim_start_matches("0x"), 16)?;
     println!("  Keychain P256 gas: {keychain_p256_gas_u64}");
 
-    let keychain_p256_diff = keychain_p256_gas_u64 as i64 - baseline_gas_u64 as i64;
-    // Keychain P256 with same-tx auth adds:
+    // Keychain P256 with same-tx auth adds ~294,733 gas which includes:
     // - 3,000 for keychain validation
     // - 5,000 for P256 signature verification
     // - ~30,000 for KeyAuthorization (27,000 base + 3,000 ecrecover)
-    // Total: ~38,000 gas
+    // - storage costs for key authorization precompile
+    let keychain_p256_diff = keychain_p256_gas_u64 as i64 - baseline_gas_u64 as i64;
     assert!(
-        (37_500..=39_000).contains(&keychain_p256_diff.unsigned_abs()),
-        "Keychain P256 + KeyAuth should add ~38,000 gas: actual diff {keychain_p256_diff} (expected 38,000 ±500)"
+        (294_000..=296_000).contains(&keychain_p256_diff.unsigned_abs()),
+        "Keychain P256 + KeyAuth should add ~294,733 gas: actual diff {keychain_p256_diff}"
     );
-    println!("  ✓ Keychain P256 + KeyAuth adds {keychain_p256_diff} gas (expected ~38,000)");
+    println!("  ✓ Keychain P256 + KeyAuth adds {keychain_p256_diff} gas (expected ~294,733)");
 
     // Test 4: KeyAuthorization with secp256k1 (no limits)
     println!("\nTest 4: KeyAuthorization (secp256k1, no limits)");
@@ -2818,13 +2818,15 @@ async fn test_aa_estimate_gas_with_keychain_and_key_auth() -> eyre::Result<()> {
     let key_auth_gas_u64 = u64::from_str_radix(key_auth_gas.trim_start_matches("0x"), 16)?;
     println!("  KeyAuth gas: {key_auth_gas_u64}");
 
-    // KeyAuth secp256k1 adds ~30,000 gas (27,000 base + 3,000 ecrecover)
+    // KeyAuth secp256k1 adds ~286,669 gas which includes:
+    // - ~30,000 for KeyAuthorization (27,000 base + 3,000 ecrecover)
+    // - storage costs for key authorization precompile
     let key_auth_diff = key_auth_gas_u64 as i64 - baseline_gas_u64 as i64;
     assert!(
-        (29_500..=31_000).contains(&key_auth_diff.unsigned_abs()),
-        "KeyAuth secp256k1 should add ~30,000 gas: actual diff {key_auth_diff} (expected 30,000 ±500)"
+        (286_000..=288_000).contains(&key_auth_diff.unsigned_abs()),
+        "KeyAuth secp256k1 should add ~286,669 gas: actual diff {key_auth_diff}"
     );
-    println!("  ✓ KeyAuth secp256k1 adds {key_auth_diff} gas (expected ~30,000)");
+    println!("  ✓ KeyAuth secp256k1 adds {key_auth_diff} gas (expected ~286,669)");
 
     // Test 5: KeyAuthorization with P256 key type (no limits)
     // Note: The key authorization signature is secp256k1 (signed by root key).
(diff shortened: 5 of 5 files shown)


crates/revm/src/handler.rs
```diff
@@ -469,14 +469,33 @@ where
...
+
+        if evm.ctx().tx().gas_limit() < adjusted_gas.initial_gas {
+            let kind = *evm
+                .ctx()
+                .tx()
+                .first_call()
+                .expect("we already checked that there is at least one call in aa tx")
+                .0;
+            // return 0 as this is case where we can't cover the initial gas.
+            // TODO(rakita) need a test for this!
```
0xKitsune: @rakita Can you address this TODO.

crates/revm/src/handler.rs
```diff
@@ -1038,20 +1082,81 @@ where
...
+            init_gas
+        };
+
+        // TIP-1000: Storage pricing updates for launch
+        // Tempo transactions with any `nonce_key` and `nonce == 0` require an additional 250,000 gas
+        // TODO(rakita): check if this is needed for `validate_aa_initial_tx_gas` function.
+        if spec.is_t1() && tx.nonce == 0 {
+            // TODO(rakita): make simpler function that give only the value without additional checks.
+            init_gas.initial_gas += gas_params.new_account_cost(true, true);
+        }
```
0xrusowsky: should we address the todos before T1?

crates/revm/src/handler.rs
```diff
@@ -1320,6 +1434,25 @@ where
...
+            let kind = *evm
+                .ctx()
+                .tx()
+                .first_call()
+                .expect("we already checked that there is at least one call in aa tx")
+                .0;
+            // return 0 as this is case where we can't cover the initial gas.
+            // TODO(rakita) need a test for this!
+            // check both execution and inspect_execution
+            return Ok(oog_frame_result(kind, 0));
```
0xrusowsky: should we add the test before T1?

rakita: There should already be tests before T1 that cover it

crates/revm/src/handler.rs
rakita: Used direct getter

crates/evm/src/evm.rs
```diff
@@ -61,7 +64,15 @@ pub struct TempoEvm<DB: Database, I = NoOpInspector> {
...
 impl<DB: Database> TempoEvm<DB> {
     /// Create a new [`TempoEvm`] instance.
-    pub fn new(db: DB, input: EvmEnv<TempoHardfork, TempoBlockEnv>) -> Self {
+    pub fn new(db: DB, mut input: EvmEnv<TempoHardfork, TempoBlockEnv>) -> Self {
+        // override the gas params for tempo use case.
+        input.cfg_env.gas_params = tempo_gas_params(input.cfg_env.spec);
+
+        // In TIP-1000 we bumped gas to 30M.
+        if input.cfg_env.spec.is_t1() {
+            input.cfg_env.tx_gas_limit_cap = Some(30_000_000);
```
0xKitsune: Can we use consts here or a const `spec.tx_gas_limit_cap()` function?

crates/revm/src/evm.rs
```diff
@@ -284,6 +290,30 @@ mod tests {
         evm
     }
 
+    /// Create an EVM with T1 hardfork enabled and a funded account.
+    /// This applies TIP-1000 gas params via `tempo_gas_params()`.
+    fn create_funded_evm_t1(address: Address) -> TempoEvm<CacheDB<EmptyDB>, ()> {
+        use crate::gas_params::tempo_gas_params;
+        use revm::context::CfgEnv;
+        use tempo_chainspec::hardfork::TempoHardfork;
```
0xKitsune: nit: can we move imports to the top of the file?

crates/revm/src/evm.rs
```diff
@@ -1424,6 +1454,366 @@ mod tests {
...
+        assert_eq!(
+            evm_with_inspector.initial_gas, 0,
+            "initial_gas should be 0 after with_inspector"
+        );
+    }
+
+    /// Test that take_logs clears logs and returns them.
+    #[test]
+    fn test_tempo_evm_take_logs() {
+        use alloy_primitives::Log;
```
0xKitsune: nit: same here can we move imports to the top of the tests mod?

crates/evm/src/evm.rs
```diff
@@ -61,7 +67,15 @@ pub struct TempoEvm<DB: Database, I = NoOpInspector> {
 
 impl<DB: Database> TempoEvm<DB> {
     /// Create a new [`TempoEvm`] instance.
-    pub fn new(db: DB, input: EvmEnv<TempoHardfork, TempoBlockEnv>) -> Self {
+    pub fn new(db: DB, mut input: EvmEnv<TempoHardfork, TempoBlockEnv>) -> Self {
+        // override the gas params for tempo use case.
+        input.cfg_env.gas_params = tempo_gas_params(input.cfg_env.spec);
+
+        // In TIP-1000 we bumped gas to 30M.
+        if input.cfg_env.spec.is_t1() {
```
legion2002: From @mattsse 
https://github.com/tempoxyz/tempo/pull/2208#discussion_r2714753038

0xKitsune: Addressed in https://github.com/tempoxyz/tempo/pull/2166/commits/d88045e6682dee3817b155645d9c9ecf6f6fe343

crates/e2e/src/tests/subblocks.rs
```diff
@@ -296,12 +296,14 @@ fn subblocks_are_included_with_failing_txs() {
             }
 
             // Send subblock transactions to all nodes.
+            // TIP-1000 charges 250k gas for new account creation, so txs from random signers
+            // need ~300k intrinsic gas. With 600k per-validator budget (5 validators), we fit 2 txs.
             for node in nodes.iter() {
-                for _ in 0..5 {
+                for _ in 0..2 {
```
legion2002: Note: We need to update this in TIP 1010, and should create a github issue for this. With the recent gas changes, only 2 txs will fit in a subblock.

crates/transaction-pool/src/validator.rs
```diff
@@ -300,13 +307,25 @@ where
...
 
-        // Add 2D nonce gas if nonce_key is non-zero
+        // Add nonce gas based on hardfork
         // If tx nonce is 0, it's a new key (0 -> 1 transition), otherwise existing key
-        if !tx.nonce_key.is_zero() {
+        if spec.is_t1() {
+            // Expiring nonce transactions
+            if tx.nonce_key == TEMPO_EXPIRING_NONCE_KEY {
+                init_and_floor_gas.initial_gas += EXPIRING_NONCE_GAS;
+            } else if tx.nonce == 0 {
```
legion2002, line 322: This charges 250,000 gas for a 2d nonce tx, on an existing account which uses a new nonce key.
This looks correct, but still double checking that it is intentional

legion2002 (approved):

crates/transaction-pool/src/validator.rs
```diff
@@ -658,6 +683,48 @@ where
     }
 }
 
+/// Ensures that gas limit of the transaction exceeds the intrinsic gas of the transaction.
+///
+/// Caution: This only checks past the Merge hardfork.
```
0xKitsune: @rakita Can you clarify what you mean by the Merge hardfork here? Seems like we can remove this.

crates/transaction-pool/src/validator.rs
```diff
@@ -658,6 +683,48 @@ where
...
+        if auth.nonce == 0 {
+            gas.initial_gas += gas_params.tx_tip1000_auth_account_creation_cost();
+        }
+    }
+
+    // TIP-1000: Storage pricing updates for launch
+    // Tempo transactions with any `nonce_key` and `nonce == 0` require an additional 250,000 gas.
+    if spec.is_t1() && tx.nonce() == 0 {
+        gas.initial_gas += gas_params.get(GasId::new_account_cost());
+    }
```
0xKitsune, line 716: @legion2002 is this to spec with expiring nonces? If `nonce_key !=0` && `nonce==0` is this accurate?

crates/revm/src/evm.rs
```diff
@@ -35,6 +35,10 @@ pub struct TempoEvm<DB: Database, I> {
     pub logs: Vec<Log>,
     /// The fee collected in `collectFeePreTx` call.
     pub(crate) collected_fee: U256,
+    /// Additional initial gas cost is needed for authorization_key setting in pre execution.
+    pub(crate) additional_initial_gas: u64,
+    /// Initial gas cost function. Used for key_authorization validation in collectFeePreTx.
```
0xKitsune: nit: Should this say function here?

crates/transaction-pool/src/validator.rs
rakita: This was copied from reth function

crates/transaction-pool/src/validator.rs
rakita: This function is only for eth tx, tempo tx has its own intrinsic fn where expiring nonce is taken into account.

crates/e2e/src/tests/subblocks.rs
klkvr: we don't have to use random signer for each tx here, the idea is to just use signers that don't have funds. we can also bump the gas limit for the tests

also the `else` branch below sends tx from the default signer so restricting entire loop to 2 iterations is wrong

crates/evm/src/evm.rs
```diff
@@ -61,7 +67,9 @@ pub struct TempoEvm<DB: Database, I = NoOpInspector> {
 
 impl<DB: Database> TempoEvm<DB> {
     /// Create a new [`TempoEvm`] instance.
-    pub fn new(db: DB, input: EvmEnv<TempoHardfork, TempoBlockEnv>) -> Self {
+    pub fn new(db: DB, mut input: EvmEnv<TempoHardfork, TempoBlockEnv>) -> Self {
+        input.cfg_env.gas_params = tempo_gas_params(input.cfg_env.spec);
```
klkvr: hmm if that's just a field on `CfgEnv` let's configure it in `ConfigureEvm::evm_env` like everything else?

crates/evm/src/evm.rs
```diff
@@ -11,10 +11,16 @@ use alloy_primitives::{Address, Bytes, Log, TxKind};
...
-use tempo_revm::{TempoHaltReason, TempoInvalidTransaction, TempoTxEnv, evm::TempoContext};
+use tempo_revm::{
+    TempoHaltReason, TempoInvalidTransaction, TempoTxEnv, evm::TempoContext,
+    gas_params::tempo_gas_params,
+};
 
 use crate::TempoBlockEnv;
 
+/// TIP-1000 gas limit cap (30 million gas).
+pub const TIP1000_GAS_LIMIT_CAP: u64 = 30_000_000;
```
mattsse: I'd like to be explicit here, I assume this is tx gas limit here?

crates/evm/src/evm.rs
mattsse: @klkvr I think this should also be moved to ConfigureEvm impl because this only operates on the input?

crates/node/src/node.rs
```diff
@@ -433,6 +433,7 @@ where
             .with_local_transactions_config(pool_config.local_transactions_config.clone())
             .set_tx_fee_cap(ctx.config().rpc.rpc_tx_fee_cap)
             .with_max_tx_gas_limit(ctx.config().txpool.max_tx_gas_limit)
+            .set_block_gas_limit(ctx.chain_spec().inner.genesis().gas_limit)
```
klkvr, line 436: hmm why do we need this change? should we set `max_tx_gas_limit` instead?

crates/evm/src/evm.rs
```diff
@@ -421,4 +430,114 @@ mod tests {
...
     }
+
+    // ==================== TIP-1000 EVM Configuration Tests ====================
+
+    /// Helper to create EvmEnv with a specific hardfork spec.
+    fn evm_env_with_spec(
+        spec: tempo_chainspec::hardfork::TempoHardfork,
+    ) -> EvmEnv<tempo_chainspec::hardfork::TempoHardfork, TempoBlockEnv> {
+        let mut env = EvmEnv::<tempo_chainspec::hardfork::TempoHardfork, TempoBlockEnv>::default();
+        env.cfg_env.spec = spec;
```
mattsse: should this also update the gas params?

crates/evm/src/lib.rs
```diff
@@ -128,8 +131,15 @@ impl ConfigureEvm for TempoEvmConfig {
 
         let spec = self.chain_spec().tempo_hardfork_at(header.timestamp());
 
+        // Set TIP-1000 gas limit cap (30M) for T1 hardfork.
```
mattsse: would like this to be explicit: this is tx gas limit

crates/evm/src/lib.rs
```diff
@@ -161,8 +171,15 @@ impl ConfigureEvm for TempoEvmConfig {
 
         let spec = self.chain_spec().tempo_hardfork_at(attributes.timestamp);
 
+        // Set TIP-1000 gas limit cap (30M) for T1 hardfork.
+        // This is set here so RPC layer can override it before constructing the EVM.
```
mattsse: dont think this comment is valuable here?

crates/node/src/node.rs
mattsse, line 436: this doesnt really matter, this is a legacy setting pre osaka, but also doesnt hurt

crates/revm/src/handler.rs
```diff
@@ -1171,19 +1274,31 @@ pub fn calculate_aa_batch_intrinsic_gas<'a>(
     let mut gas = InitialAndFloorGas::default();
 
     // 1. Base stipend (21k, once per transaction)
-    gas.initial_gas += 21_000;
+    gas.initial_gas += gas_params.tx_base_stipend();
```
klkvr, line 1273: ideally we should've done this kind of changes separately (replacing constanst with gas params)

crates/evm/src/evm.rs
rakita: Probably yes, will put it inside Default. edit: Default is insie alloy/evm so would create new CfgEnv here

crates/revm/src/handler.rs
```diff
@@ -762,75 +791,88 @@ where
...
-                            expiry,
-                            current_timestamp,
-                        }
-                        .into());
-                    }
+                tx.gas_limit() - initial_gas,
+            );
+
+            // The core logic of setting up thread-local storage is here.
+            let out_of_gas = StorageCtx::enter(&mut provider, || {
```
klkvr, line 797: it's a bit unfortunate that we can go OOG here in validation. could we do it like with 7702 auth lists where we charge a fixed amount that can be ensured beforehand and then refund the rest?

crates/evm/src/evm.rs
rakita: I see, will change CfgEnv::gas_params there.

crates/node/src/node.rs
rakita: Matt said in this [comment](https://github.com/tempoxyz/tempo/pull/2166#discussion_r2716431513) that `this doesnt really matter, this is a legacy setting pre osaka, but also doesnt hurt`

crates/revm/src/handler.rs
rakita: Unfortunately have skipped it in the initial revm/reth bump

crates/revm/src/handler.rs
rakita: This decision was made as there was consideration to remove fee refund, if the fee refund is removed gas refund wouldn't do anything.

crates/evm/src/evm.rs
rakita: @klkvr would it be enought to change gas_params only in `evm_env` and `next_evm_env`, is it safe to remove this line from `TempoEvm::new`?

crates/transaction-pool/src/validator.rs
rakita: 2d nonce key storage will get created if tx.nonce == 0 this is why we add additional 250k.

crates/evm/src/lib.rs
rakita: Chnged [here](https://github.com/tempoxyz/tempo/pull/2166/changes/ae12dda43c10e3b263352f3aea0c1936187a8254)

crates/evm/src/evm.rs
rakita: I moved it to `evm_env` and `next_evm_env`

rakita: @klkvr commented it [here](https://github.com/tempoxyz/tempo/pull/2166#discussion_r2716415837)

crates/evm/src/evm.rs
```diff
@@ -11,10 +11,17 @@ use alloy_primitives::{Address, Bytes, Log, TxKind};
...
+use tempo_revm::{
+    TempoHaltReason, TempoInvalidTransaction, TempoTxEnv, evm::TempoContext,
+    gas_params::tempo_gas_params,
+};
 
 use crate::TempoBlockEnv;
 
+/// TIP-1000 sets transaction gas limit cap to 30 million gas.
+///
+pub const TIP1000_TX_GAS_LIMIT_CAP: u64 = 30_000_000;
```
klkvr, line 22: can we add motivation for this limit to the comment?

crates/evm/src/evm.rs
klkvr: can we remove this from `TempoEvm::new` now?

xtask/src/genesis_args.rs
```diff
@@ -492,6 +492,7 @@ impl GenesisArgs {
         chain_config
             .extra_fields
             .insert_value("t1Time".to_string(), self.t1_time)?;
+
```
klkvr:
```suggestion
```

crates/revm/src/handler.rs
```diff
@@ -538,36 +557,46 @@ where
...
+        } else {
+            // For standard EIP-7702 transactions, use the default implementation
+            pre_execution::apply_eip7702_auth_list::<_, Self::Error>(evm.ctx())?
+        };
+
+        // TIP-1000: State Creation Cost Increase
+        // Authorization lists: There is no refund if the account already exists
+        if spec.is_t1() {
+            return Ok(0);
         }
```
klkvr, line 580: should refund amounts be configurable on `GasParams` @rakita?

crates/revm/src/handler.rs
```diff
@@ -1096,20 +1139,79 @@ where
...
+                if auth.nonce == 0 {
+                    init_gas.initial_gas += gas_params.tx_tip1000_auth_account_creation_cost();
+                }
+            }
+
+            // TIP-1000: Storage pricing updates for launch
+            // Tempo transactions with any `nonce_key` and `nonce == 0` require an additional 250,000 gas.
+            if spec.is_t1() && tx.nonce == 0 {
+                init_gas.initial_gas += gas_params.get(GasId::new_account_cost());
+            }
```
klkvr, line 1180: comment is odd here as this branch is not for tempo transactions

crates/revm/src/handler.rs
```diff
@@ -1171,19 +1274,31 @@ pub fn calculate_aa_batch_intrinsic_gas<'a>(
...
-    gas.initial_gas += 21_000;
+    gas.initial_gas += gas_params.tx_base_stipend();
 
     // 2. Signature verification gas
     gas.initial_gas += tempo_signature_verification_gas(signature);
 
-    // 3. Per-call overhead: cold account access for additional calls beyond the first.
-    gas.initial_gas += COLD_ACCOUNT_ACCESS_COST * calls.len().saturating_sub(1) as u64;
+    let cold_account_cost =
+        gas_params.warm_storage_read_cost() + gas_params.cold_account_additional_cost();
```
klkvr, line 1279: can `gas_params` expose a helper for this computation? unclear what is storage/addition cost here

crates/revm/src/handler.rs
```diff
@@ -1171,19 +1274,31 @@ pub fn calculate_aa_batch_intrinsic_gas<'a>(
...
+
+    // 3. Per-call overhead: cold account access
+    // if the `to` address has not appeared in the call batch before.
+    gas.initial_gas += cold_account_cost * calls.len().saturating_sub(1) as u64;
```
klkvr, line 1283: hmm do we actually check that `to` account has not appeared before? looks like we just charge for every item as if it was cold

crates/revm/src/handler.rs
klkvr: hmm it seems that we even have configurable `tx_eip7702_per_empty_account_cost` but it's unused in `apply_auth_list`

crates/revm/src/handler.rs
klkvr: now that I look at the code it's just `authorizeKey` that might go OOG. can we not price it statically based on key authorization? i guess it's just some amount of cold sstores

crates/e2e/src/tests/subblocks.rs
rakita: I have locally reverted it to 0..5 and it passes. Would push this change up

crates/revm/src/handler.rs
```diff
@@ -491,14 +491,32 @@ where
         evm: &mut Self::Evm,
         init_and_floor_gas: &InitialAndFloorGas,
     ) -> Result<FrameResult, Self::Error> {
+        // Add key_authorization gas to the initial gas (calculated in validate_aa_initial_tx_gas)
+        // Use saturating_add to avoid overflow when additional_initial_gas is u64::MAX (OOG case)
+        let adjusted_gas = InitialAndFloorGas::new(
+            init_and_floor_gas
+                .initial_gas
+                .saturating_add(evm.additional_initial_gas),
```
```suggestion
            evm.initial_gas
```

this seems cleaner, instead of extra `additional_initial_gas` field we can just increment it on evm

crates/revm/src/handler.rs
```diff
@@ -1096,20 +1139,79 @@ where
...
+                    })
+                    .unwrap_or_default();
+            };
+            let mut init_gas = gas_params.initial_tx_gas(
+                tx.input(),
+                tx.kind().is_create(),
+                acc as u64,
+                storage as u64,
+                tx.authorization_list_len() as u64,
+            );
```
klkvr, line 1166: hmm @rakita can we not have default impl for `validate_initial_tx_gas` take gas params? it's abstracted but not propagated deep enough it seems and we're still forced to duplicate e.g access list checks

crates/revm/src/handler.rs
```diff
@@ -1218,27 +1332,23 @@ pub fn calculate_aa_batch_intrinsic_gas<'a>(
...
-                    storage_count + item.storage_slots().count() as u64,
-                )
-            });
-        gas.initial_gas += accounts * ACCESS_LIST_ADDRESS; // 2400 per account
-        gas.initial_gas += storages * ACCESS_LIST_STORAGE_KEY; // 1900 per storage
+        let (accounts, storages) = access_list.fold((0, 0), |(acc_count, storage_count), item| {
+            (acc_count + 1, storage_count + item.storage_slots().count())
+        });
+        gas.initial_gas += accounts * gas_params.tx_access_list_address_cost(); // 2400 per account
+        gas.initial_gas += storages as u64 * gas_params.tx_access_list_storage_key_cost(); // 1900 per storage
```
klkvr, line 1343: this should be a helper on `gas_params`

crates/revm/src/handler.rs
rakita: `tx_eip7702_per_empty_account_cost` is used in intrinsic gas as base cost of authorization

rakita: Make a task https://github.com/bluealloy/revm/issues/3341

crates/revm/src/handler.rs
rakita: If authorization_key creates even one storage, it will be 250k gas, which is too much.

Yeah authorizeKey is only one that can trigger oog.

crates/revm/src/handler.rs
rakita: cost cost = warm cost + additional cold cost.

Can make a helper

crates/revm/src/handler.rs
rakita: Little bit out of scope of this PR, It was simpler to just charge cold storage for number of calls in tx. In general we could optimize and make it dynamic, and reduce cost of intrinsic gas.

crates/revm/src/handler.rs
rakita: It is cleaner, i do like it better as passing InitialAndFloorGas over function is kinda not flexible.

I am even thinking to have nativly in revm, but this is for later.

crates/revm/src/handler.rs
rakita: https://github.com/bluealloy/revm/issues/3343

crates/revm/src/handler.rs
rakita: Dont fully understand, validate_initial_tx_gas does not take gas params but it creates the default mainnet GasParams variant, this is not most practical thing as you said it is not flexible enough to cover everything.

I imagine we should have a lot more flexibility on initial gas verification and how it is done inside revm.

crates/revm/src/handler.rs
```diff
@@ -953,7 +950,6 @@ where
         // Short-circuit if there is no spending for this transaction and `collectFeePreTx`
         // call will not collect any fees.
         if gas_balance_spending.is_zero() {
-            evm.additional_initial_gas = 0;
```
rakita, line 956: Needed to change the testin tempo_transaction.rs as additional gas would not be zeroed. Setting `additonal_initial_gas` to zero would be a bug as gas limit still needs to be checked.

crates/revm/src/handler.rs
klkvr: yep it should just take gas params from `CfgEnv`. right now it's not possible to have custom gas params without overriding handler functions which is not very convenient

crates/revm/src/handler.rs
klkvr: hmm yeah but we will charge that anyway right? `authorizeKey` will always create storage

crates/revm/src/handler.rs
klkvr: tracking here https://github.com/tempoxyz/tempo/issues/2222

PR #2167, opened 2026-01-19
fix(txpool): evict transactions when validator token preference changes to prevent DoS

fgimenez:
Closes CHAIN-472

Prevents a mempool DoS vector where a validator could change their token preference after users submit transactions, causing those transactions to fail at execution due to insufficient liquidity in the new (user_token, validator_token) AMM pool.

The fix monitors `ValidatorTokenSet` events on block commit and proactively evicts pooled transactions that would fail with the new validator token configuration.

### DIFF
--- a/crates/transaction-pool/src/tempo_pool.rs
+++ b/crates/transaction-pool/src/tempo_pool.rs
@@ -6,7 +6,7 @@ use crate::{
     amm::AmmLiquidityCache, best::MergeBestTransactions, transaction::TempoPooledTransaction,
     tt_2d_pool::AA2dPool, validator::TempoTransactionValidator,
 };
-use alloy_consensus::{Transaction, transaction::TxHashRef};
+use alloy_consensus::Transaction;
 use alloy_primitives::{Address, B256, map::HashMap};
 use parking_lot::RwLock;
 use reth_chainspec::ChainSpecProvider;
@@ -117,10 +117,11 @@ where
         let mut revoked_count = 0;
         let mut liquidity_count = 0;
 
-        for tx in self.all_transactions().all() {
+        let all_txs = self.all_transactions();
+        for tx in all_txs.pending.iter().chain(all_txs.queued.iter()) {
             // Check 1: Revoked keychain keys (only for AA transactions with keychain signatures)
             if !revoked_keys.is_empty()
-                && let Some(aa_tx) = tx.inner().as_aa()
+                && let Some(aa_tx) = tx.transaction.inner().as_aa()
                 && let Some(keychain_sig) = aa_tx.signature().as_keychain()
             {
                 let is_revoked = revoked_keys.iter().any(|&(account, key_id)| {
@@ -130,7 +131,7 @@ where
                             .is_ok_and(|tx_key_id| tx_key_id == key_id)
                 });
                 if is_revoked {
-                    to_remove.push(*tx.tx_hash());
+                    to_remove.push(*tx.hash());
                     revoked_count += 1;
                     continue;
                 }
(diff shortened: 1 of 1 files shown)


crates/transaction-pool/src/maintain.rs
```diff
@@ -154,13 +154,37 @@ where
...
+                    .map(|event| (event.validator, event.token))
+                    .collect();
+
+                if !validator_token_changes.is_empty() {
+                    debug!(
+                        target: "txpool",
+                        count = validator_token_changes.len(),
+                        "Processing validator token preference changes"
+                    );
+                    pool.evict_transactions_by_validator_token_change(&validator_token_changes);
```
mattsse: same suggestion here as on the other one, it would be nice if we can batch those per update then we dont need to call multiple event update per block and can avoid scanning all txs multiple times, but this can be unified in a followup

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -118,6 +118,91 @@ where
...
+        let mut to_remove = Vec::new();
+
+        for tx in self.all_transactions().all() {
+            let user_token = tx
+                .fee_token()
+                .unwrap_or(tempo_precompiles::DEFAULT_FEE_TOKEN);
+            let cost = tempo_primitives::transaction::calc_gas_balance_spending(
+                tx.gas_limit(),
+                tx.max_fee_per_gas(),
+            );
```
mattsse: dnt we have this cached as as cost in the tx itself?

https://github.com/tempoxyz/tempo/blob/4baa4e38f5610927a25fdb050bb5e52a0526dffa/crates/transaction-pool/src/transaction.rs#L48-L52

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -118,6 +118,91 @@ where
...
+                    continue;
+                }
+
+                // Check if there's enough liquidity in pool(user_token, new_validator_token)
+                let pool_key = PoolKey::new(user_token, new_validator_token).get_id();
+                let slot = tempo_precompiles::tip_fee_manager::TipFeeManager::new().pools[pool_key]
+                    .base_slot();
+
+                let pool_value = match state_provider.storage(TIP_FEE_MANAGER_ADDRESS, slot.into())
+                {
```
mattsse: I see, this makes sense

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -118,6 +118,91 @@ where
...
+            tip_fee_manager::amm::{Pool, PoolKey, compute_amount_out},
+        };
+
+        let Ok(state_provider) = self.client().latest() else {
+            return;
+        };
+
+        let mut to_remove = Vec::new();
+
+        for tx in self.all_transactions().all() {
```
mattsse: okay, this makes sense, this doesnt hold the lock

and cloning this is kinda cheap enough

crates/transaction-pool/src/maintain.rs
fgimenez: ok done in this same PR, now invalidation checks are batched into a single pool scan

crates/transaction-pool/src/tempo_pool.rs
fgimenez: indeed, updated to use `tx.transaction.fee_token_cost()` which already caches the gas cost calculation, also switched from `.all()` to iterating `pending` and `queued` directly to access the `ValidPoolTransaction` wrapper

mattsse (approved):
lgtm

PR #2169, opened 2026-01-19
fix: address PR #1762 review comments

gakonst:
This PR addresses the review comments on #1762:

## Changes

- **handler.rs:536** - Handle error properly using `?` operator instead of `.unwrap()` (@0xKitsune)
- **storage/evm.rs** - Remove commented `skip_cold_load` logic and pass `false` directly (@0xKitsune)
  - `with_account_info`: removed commented line and variable
  - `sstore`: removed commented line and variable  
  - `sload`: removed commented line and variable

## Not addressed

- **lib.rs gas_params clone** (@0xrusowsky) - The clone on line 120 is necessary because the closure can be called multiple times. The first clone on line 107 creates the value captured by the `move` closure, and each invocation needs to pass ownership to `EvmPrecompileStorageProvider::new`.
- **hardfork gating questions** (@0xKitsune, @legion2002) - These need domain expertise to determine if hardfork gates are required. Left for the PR author to address.
- **subblock test changes** (@legion2002) - These appear to be timing-related test fixes, not directly related to the revm changes.

Closes comments on #1762

### DIFF
--- a/crates/precompiles/src/storage/evm.rs
+++ b/crates/precompiles/src/storage/evm.rs
@@ -80,7 +80,7 @@ impl<'a> PrecompileStorageProvider for EvmPrecompileStorageProvider<'a> {
 
         let mut account = self
             .internals
-            .load_account_mut_skip_cold_load(address, false)?;
+            .load_account_mut_skip_cold_load(address, additional_cost == 0)?;
 
         // TODO(rakita) can be moved to the beginning of the function. Requires fork.
         deduct_gas(&mut self.gas_remaining, WARM_STORAGE_READ_COST)?;
@@ -103,10 +103,11 @@ impl<'a> PrecompileStorageProvider for EvmPrecompileStorageProvider<'a> {
         key: U256,
         value: U256,
     ) -> Result<(), TempoPrecompileError> {
-        let result = self
-            .internals
-            .load_account_mut(address)?
-            .sstore(key, value, false)?;
+        let skip_cold_load = self.gas_params.cold_storage_additional_cost() == 0;
+        let result =
+            self.internals
+                .load_account_mut(address)?
+                .sstore(key, value, skip_cold_load)?;
 
         // TODO(rakita) can be moved to the beginning of the function. Requires fork.
         self.deduct_gas(self.gas_params.sstore_static_gas())?;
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/storage/evm.rs
```diff
@@ -77,12 +77,10 @@ impl<'a> PrecompileStorageProvider for EvmPrecompileStorageProvider<'a> {
         f: &mut dyn FnMut(&AccountInfo),
     ) -> Result<(), TempoPrecompileError> {
         let additional_cost = self.gas_params.cold_account_additional_cost();
-        // let skip_cold_load = 0 < additional_cost;
-        let skip_cold_load = false;
 
         let mut account = self
             .internals
-            .load_account_mut_skip_cold_load(address, skip_cold_load)?;
+            .load_account_mut_skip_cold_load(address, false)?;
```
0xKitsune: Should this be `additional_cost == 0` instead?

crates/precompiles/src/storage/evm.rs
```diff
@@ -105,13 +103,10 @@ impl<'a> PrecompileStorageProvider for EvmPrecompileStorageProvider<'a> {
...
-        let skip_cold_load = false;
-
-        let result =
-            self.internals
-                .load_account_mut(address)?
-                .sstore(key, value, skip_cold_load)?;
+        let result = self
+            .internals
+            .load_account_mut(address)?
+            .sstore(key, value, false)?;
```
0xKitsune: Same thing here should this be `self.gas_params.cold_storage_additional_cost() == 0`?

crates/precompiles/src/storage/evm.rs
```diff
@@ -162,14 +157,11 @@ impl<'a> PrecompileStorageProvider for EvmPrecompileStorageProvider<'a> {
...
 
-        //let skip_cold_load = 0 < additional_cost;
-        let skip_cold_load = false;
-
         let value;
         let is_cold;
         {
             let mut account = self.internals.load_account_mut(address)?;
-            let val = account.sload(key, skip_cold_load)?;
+            let val = account.sload(key, false)?;
```
0xKitsune: Same here, should this be `additional_cost == 0`?

PR #2172, opened 2026-01-19
feat: --telemetry-url flag for unified telemetry configuration

kamsz:
Adds a single --telemetry-url flag that configures all telemetry: **This is coupled to a VictoriaMetrics implementation**. Supports both prometheus/otlp collection

Metrics:
  - Exporter pushes both reth/commonware metrics together
  - All exported metrics have a `consensus_id` label to identify the node. Possible with `extra_labels` query parameter victoria metrics provides.
  
Logs:
  - Configure --logs-otlp which reth exports. common uses the same tracing package, thus also exported
  - **--telemetry-url and --logs.otlp cannot be used together**

### DIFF
--- a/bin/tempo/src/defaults.rs
+++ b/bin/tempo/src/defaults.rs
@@ -41,7 +41,7 @@ impl TelemetryArgs {
         let password = telemetry_url.password().expect("ensured when parsing args");
 
         // Build auth header for metrics push and OTLP logs
-        let credentials = format!("{}:{}", username, password);
+        let credentials = format!("{username}:{password}");
         let encoded = BASE64_STANDARD.encode(credentials.as_bytes());
         let auth_header = format!("Basic {encoded}");
 


crates/commonware-node/src/args.rs
```diff
@@ -32,6 +32,15 @@ pub struct Args {
     #[arg(long = "consensus.metrics-address", default_value = "127.0.0.1:8001")]
     pub metrics_address: SocketAddr,
 
+    /// The OTLP endpoint URL to push consensus metrics to (e.g., `https://metrics.example.com/v1/metrics`).
+    /// If not set, metrics will only be exposed on the metrics-address endpoint.
+    #[arg(long = "consensus.metrics-otlp")]
+    pub metrics_otlp_url: Option<String>,
```
SuperFluffy: Bikeshed: important to call prefix this with `metrics_`?

crates/commonware-node/src/args.rs
```diff
@@ -32,6 +32,15 @@ pub struct Args {
...
+
+    /// The interval at which to push consensus metrics via OTLP.
+    #[arg(long = "consensus.metrics-otlp.interval", default_value = "10s", value_parser = parse_duration)]
+    pub metrics_otlp_interval: Duration,
```
SuperFluffy: For consistency, please use `jiff::SignedDuration` as is done in the other instances and without `value_parser`. We can replace this by `parse_value`, but not in this PR.

bin/tempo/src/main.rs
```diff
@@ -177,6 +180,35 @@ fn main() -> eyre::Result<()> {
...
+                let mut metrics_otlp = args.consensus.metrics_otlp_url.clone().map(|endpoint| {
+                    // Build labels for identifying this node in metrics
+                    let mut labels = std::collections::HashMap::new();
+
+                    // Add consensus public key as node identifier
+                    if let Ok(Some(public_key)) = args.consensus.public_key() {
+                        let pubkey_hex: String = public_key
+                            .as_ref()
+                            .iter()
+                            .map(|b| format!("{b:02x}"))
```
SuperFluffy: The pub key implements `Display`, so this is not necessary.

bin/tempo/src/main.rs
```diff
@@ -110,12 +110,15 @@ fn main() -> eyre::Result<()> {
     tempo_node::init_version_metadata();
     defaults::init_defaults();
 
+    // Expand --telemetry-url into the equivalent telemetry arguments
+    let args = defaults::expand_telemetry_args(std::env::args().collect())?;
```
SuperFluffy: Why is the clap parser by-passed? Is there a reason for it? Unless there is a good justification for this, it should be done via `TempoArgs` - that's what the tempo-specfic arguments are for.

crates/commonware-node/src/metrics.rs
```diff
@@ -49,3 +55,155 @@ pub fn install(context: Context, listen_addr: SocketAddr) -> Handle<eyre::Result
...
+    pub interval: Duration,
+    /// Labels to add to all metrics (e.g., node_id, consensus_pubkey).
+    pub labels: HashMap<String, String>,
+}
+
+/// Installs an OTLP metrics exporter that periodically pushes consensus metrics.
+///
+/// This creates an OpenTelemetry MeterProvider with an OTLP exporter and periodically
+/// reads metrics from the commonware context, converting them to OpenTelemetry format.
+pub fn install_otlp(context: Context, config: OtlpConfig) -> Handle<eyre::Result<()>> {
```
SuperFluffy: There is no need to construct the periodic exporter in a separate task. The OTEL exporter itself spawning a background task, the object itself is "inert". Pass the context in, but don't call `context.spawn` on it - that's an unnecessary extra step.

This method should instead just return the `PeriodicReader`.

crates/commonware-node/src/metrics.rs
```diff
@@ -49,3 +55,155 @@ pub fn install(context: Context, listen_addr: SocketAddr) -> Handle<eyre::Result
...
+        tracing::info!(
+            endpoint = %config.endpoint,
+            interval = ?config.interval,
+            "OTLP metrics exporter started"
+        );
+
+        // Keep the meter provider alive
+        // The periodic reader will automatically export metrics at the configured interval
+        loop {
+            tokio::time::sleep(Duration::from_secs(3600)).await;
```
SuperFluffy: Unnecessary - just build the exporter outside.

crates/commonware-node/src/metrics.rs
```diff
@@ -49,3 +55,155 @@ pub fn install(context: Context, listen_addr: SocketAddr) -> Handle<eyre::Result
...
+                        observer.observe(value, &attrs_with_name);
+                    }
+                }
+            })
+            .build();
+
+        tracing::info!(
+            endpoint = %config.endpoint,
+            interval = ?config.interval,
+            "OTLP metrics exporter started"
```
SuperFluffy: I think this event is emitted outside a root span (or a root span that is alive for the entire runtime of the program so it would never show up in OTEL).

crates/commonware-node/src/metrics.rs
```diff
@@ -49,3 +55,155 @@ pub fn install(context: Context, listen_addr: SocketAddr) -> Handle<eyre::Result
...
+
+        let meter = meter_provider.meter("tempo-consensus");
+
+        // Create gauges for the metrics we want to export
+        // We'll parse the prometheus text format and update these gauges
+        let metrics_context = Arc::new(context);
+        let meter = Arc::new(meter);
+
+        // Create an observable gauge that reads from the context on each export
+        let ctx_clone = metrics_context.clone();
```
SuperFluffy: All that cloning is not necessary. Just move the context inside the closure.

crates/commonware-node/src/metrics.rs
```diff
@@ -49,3 +55,155 @@ pub fn install(context: Context, listen_addr: SocketAddr) -> Handle<eyre::Result
...
+        let meter_provider = SdkMeterProvider::builder()
+            .with_resource(resource)
+            .with_reader(reader)
+            .build();
+
+        let meter = meter_provider.meter("tempo-consensus");
+
+        // Create gauges for the metrics we want to export
+        // We'll parse the prometheus text format and update these gauges
+        let metrics_context = Arc::new(context);
```
SuperFluffy: Unnecessary - the context is cloneable already.

crates/commonware-node/src/metrics.rs
```diff
@@ -49,3 +55,155 @@ pub fn install(context: Context, listen_addr: SocketAddr) -> Handle<eyre::Result
...
+        // The periodic reader will automatically export metrics at the configured interval
+        loop {
+            tokio::time::sleep(Duration::from_secs(3600)).await;
+        }
+    })
+}
+
+/// Parse a prometheus metric line like `metric_name{label="value"} 123`
+/// Returns (metric_name, vec of (label_name, label_value))
+fn parse_prometheus_metric(name_labels: &str) -> (String, Vec<(String, String)>) {
```
SuperFluffy: Nice; looks easy enough. We should push for replacing `prometheus_client` by the metrics crate to make this nicer.

crates/commonware-node/src/args.rs
kamsz: yep, because reth exposes `--tracing-otlp` and `--logs-otlp`, so for consistency metrics are also named `--metrics-otlp`

bin/tempo/src/main.rs
kamsz: fixed in https://github.com/tempoxyz/tempo/pull/2172/commits/e33bf818db8c8b648e778ab5eb57dc15a4bb768c

crates/commonware-node/src/args.rs
kamsz: fixed in https://github.com/tempoxyz/tempo/pull/2172/commits/e33bf818db8c8b648e778ab5eb57dc15a4bb768c

crates/commonware-node/src/metrics.rs
kamsz: i'm not sure how to do that properly to preserve original metric names - otel sdk doesn't know them so we need to poll them out of commonware periodically and register as separate gauges - otherwise they will end up as a single gauge with original metric name as label

bin/tempo/src/main.rs
kamsz: fixed in https://github.com/tempoxyz/tempo/pull/2172/commits/e12bdc80144a1bfc37f8b85208c26407de1ea069

SuperFluffy: Bringing this back here - is there a reason for this?

IMO this is the completely wrong approach and incredibly brittle.

If some configuration cannot be expressed with clap natively, then a) clap should first be parsed, and then b) the options configured as required, before c) sending them to the application for execution.

bin/tempo/src/main.rs
```diff
@@ -54,6 +54,22 @@ struct TempoArgs {
     #[arg(long, value_name = "URL", default_missing_value = "auto", num_args(0..=1))]
     pub follow: Option<String>,
 
+    /// Unified telemetry URL that expands to configure logs, metrics, and consensus telemetry.
+    ///
+    /// When provided, this URL is expanded and overwrites set values for:
```
SuperFluffy: This is potentially very confusing. If this is the case, then shouldn't this conflict with the other values?

bin/tempo/src/defaults.rs
```diff
@@ -56,3 +61,130 @@ pub(crate) fn init_defaults() {
     init_payload_builder_defaults();
     init_txpool_defaults();
 }
+
+/// Extracts the telemetry URL from CLI args or environment variable.
+///
+/// Looks for `--telemetry-url <url>` or `--telemetry-url=<url>` in the args.
+/// Falls back to the `TELEMETRY_URL` environment variable if not found in args.
```
SuperFluffy: Where does it fall back to the env var? I don't see it set anywhere.

crates/commonware-node/src/metrics.rs
```diff
@@ -49,3 +58,323 @@ pub fn install(context: Context, listen_addr: SocketAddr) -> Handle<eyre::Result
...
+pub struct OtlpConfig {
+    /// The OTLP endpoint URL (e.g., `https://metrics.example.com/v1/metrics`).
+    pub endpoint: String,
+    /// The interval at which to export metrics.
+    pub interval: SignedDuration,
+    /// Labels to add to all metrics as resource attributes (e.g., node_id, consensus_pubkey).
+    pub labels: HashMap<String, String>,
+}
+
+/// Handle to the OTLP metrics exporter that must be held for the lifetime of the export.
```
SuperFluffy: Why must the handle be held if we just spawn in on a runtime? Because of `SdkMeterProvider`? Couldn't we just move it into the task (or construct it there) and call it a day?

bin/tempo/src/main.rs
```diff
@@ -179,6 +196,34 @@ fn main() -> eyre::Result<()> {
...
+                            interval: args.consensus.metrics_otlp_interval,
+                            labels,
+                        };
+
+                        tempo_commonware_node::metrics::install_otlp(
+                            ctx.with_label("metrics_otlp"),
+                            config,
+                        )
+                    })
+                    .transpose()?;
```
SuperFluffy: error context

bin/tempo/src/main.rs
```diff
@@ -179,6 +196,34 @@ fn main() -> eyre::Result<()> {
...
+                    .consensus
+                    .metrics_otlp_url
+                    .clone()
+                    .map(|endpoint| {
+                        // Build labels for identifying this node in metrics
+                        let mut labels = std::collections::HashMap::new();
+
+                        // Add consensus public key as node identifier
+                        if let Ok(Some(public_key)) = args.consensus.public_key() {
+                            labels.insert("consensus_pubkey".to_string(), public_key.to_string());
```
SuperFluffy: Just `consensus_id`?

crates/commonware-node/src/metrics.rs
```diff
@@ -49,3 +58,323 @@ pub fn install(context: Context, listen_addr: SocketAddr) -> Handle<eyre::Result
...
+    let counters: Arc<Mutex<HashMap<String, opentelemetry::metrics::Counter<f64>>>> =
+        Arc::new(Mutex::new(HashMap::new()));
+
+    // Cache for dynamically created gauges (including histograms)
+    let gauges: Arc<Mutex<HashMap<String, opentelemetry::metrics::Gauge<f64>>>> =
+        Arc::new(Mutex::new(HashMap::new()));
+
+    // Poll at half the export interval to ensure fresh data for each export
+    let poll_interval = interval / 2;
+    let task = context.spawn(move |context| async move {
```
SuperFluffy: This is relatively short-lived but compute intensive. Let's use `Spawner::shared` to run this async on a dedicated thread within the same runtime:

https://github.com/commonwarexyz/monorepo/blob/4fbfebb2aa48a482fc844948f6011fb1fe5f8af3/runtime/src/tokio/runtime.rs#L482-L487

crates/commonware-node/src/metrics.rs
```diff
@@ -49,3 +58,323 @@ pub fn install(context: Context, listen_addr: SocketAddr) -> Handle<eyre::Result
...
+        // Track last counter values to compute deltas (OTLP counters expect deltas, not absolutes)
+        let mut last_counter_values: HashMap<(String, Vec<(String, String)>), f64> = HashMap::new();
+
+        loop {
+            context.sleep(poll_interval).await;
+
+            let encoded = context.encode();
+
+            // Parse TYPE comments to determine metric types
+            let metric_types = parse_prometheus_types(&encoded);
```
SuperFluffy: Unless this is a micro-optimization, we can construct this hashmap once before the loop and then only insert new fields as we encounter them.

The metrics that come out of the (rust prometheus client) registry remain there until the end of the program, so this histogram will only ever grow but never shrink.

crates/commonware-node/src/metrics.rs
```diff
@@ -49,3 +58,323 @@ pub fn install(context: Context, listen_addr: SocketAddr) -> Handle<eyre::Result
...
+                        let key = (metric_name, labels);
+                        let last_value = last_counter_values.get(&key).copied().unwrap_or(0.0);
+                        let delta = (value - last_value).max(0.0);
+                        last_counter_values.insert(key, value);
+
+                        if delta > 0.0 {
+                            counter.add(delta, &attributes);
+                        }
+                    } else {
+                        // Gauge - use gauge (histograms are handled as counters above)
```
SuperFluffy: I don't understand this comment.

crates/commonware-node/src/metrics.rs
```diff
@@ -49,3 +58,323 @@ pub fn install(context: Context, listen_addr: SocketAddr) -> Handle<eyre::Result
...
+            && let Some((name, type_str)) = rest.split_once(' ')
+        {
+            types.insert(name.to_string(), type_str);
+        }
+    }
+    types
+}
+
+/// Strip known Prometheus suffixes to find the base metric name for type lookup.
+fn strip_prometheus_suffix(metric_name: &str) -> &str {
```
SuperFluffy: Can we use this to classify metrics into enum variants?

I thinking something like:

```rust
match metric.rsplit_once('_') {
    Some((base, "total")) => MetricType::Total(base),
    Some((base, "count")) => MetricType::Count(base),
    Some((base, "sum")) => MetricType::Sum(base),
    Some((base, "bucket")) => MetricType::Bucket(base),
    _ => MetricType::Other(metric),
}
```

crates/commonware-node/src/metrics.rs
```diff
@@ -49,3 +58,323 @@ pub fn install(context: Context, listen_addr: SocketAddr) -> Handle<eyre::Result
...
+                    if is_counter_type(metric_type) {
+                        let counter = {
+                            let mut cache = counters.lock();
+                            cache
+                                .entry(metric_name.clone())
+                                .or_insert_with(|| meter.f64_counter(metric_name.clone()).build())
+                                .clone()
+                        };
+
+                        // Compute delta from last value (counters are cumulative in Prometheus)
```
SuperFluffy: Interesting, this actually breaks us setting some counters back to 0 in the DKG loop (technically this was never correct). Going to double check.

crates/commonware-node/src/metrics.rs
```diff
@@ -49,3 +58,323 @@ pub fn install(context: Context, listen_addr: SocketAddr) -> Handle<eyre::Result
...
+                        let key = (metric_name, labels);
+                        let last_value = last_counter_values.get(&key).copied().unwrap_or(0.0);
+                        let delta = (value - last_value).max(0.0);
+                        last_counter_values.insert(key, value);
+
+                        if delta > 0.0 {
```
SuperFluffy: Why would `counter.add(0.0, &attrs)` be bad?

crates/commonware-node/src/metrics.rs
```diff
@@ -49,3 +58,323 @@ pub fn install(context: Context, listen_addr: SocketAddr) -> Handle<eyre::Result
...
+                            counter.add(delta, &attributes);
+                        }
+                    } else {
```
SuperFluffy: So everything is a gauge except for counters?

crates/commonware-node/src/metrics.rs
```diff
@@ -49,3 +58,323 @@ pub fn install(context: Context, listen_addr: SocketAddr) -> Handle<eyre::Result
...
+        .build();
+
+    let meter = meter_provider.meter("tempo-consensus");
+
+    // Cache for dynamically created counters
+    let counters: Arc<Mutex<HashMap<String, opentelemetry::metrics::Counter<f64>>>> =
+        Arc::new(Mutex::new(HashMap::new()));
+
+    // Cache for dynamically created gauges (including histograms)
+    let gauges: Arc<Mutex<HashMap<String, opentelemetry::metrics::Gauge<f64>>>> =
```
SuperFluffy: Why do histograms end up as gauges? Do the prometheus ones not map well to the OTEL ones? https://docs.rs/opentelemetry/latest/opentelemetry/metrics/struct.Histogram.html

crates/commonware-node/src/metrics.rs
SuperFluffy: Yeah, would like to understand why that is.

bin/tempo/src/main.rs
hamdiallam: The only conflicting flag is `logs.otlp` that reth exposes. Updated this such that setting both is an error state

bin/tempo/src/defaults.rs
hamdiallam: Removed.

crates/commonware-node/src/metrics.rs
hamdiallam: This is definitely incorrect usage of a counter. Anything that can be reset/decrease should be a gauge

crates/commonware-node/src/metrics.rs
hamdiallam: No reason. updated to be a >=

crates/commonware-node/src/metrics.rs
hamdiallam: Updated the comment but since prometheus exports cumulative counters, we cant record the individual observations for an open telemetry histogram.

There is a hacky way around this by constructing the entire histogram from the cumulative values then exporting that directly
```rust
use opentelemetry_sdk::metrics::data::{Histogram, HistogramDataPoint, ResourceMetrics};

// Parse prometheus histogram -> construct HistogramDataPoint directly
let data_point = HistogramDataPoint {
    bounds: vec![0.1, 0.5, 1.0],           // from le labels
    bucket_counts: vec![10, 15, 5, 5],     // converted from cumulative
    count: 35,
    sum: 42.5,
    // ...
};

// Export directly
exporter.export(&resource_metrics).await;
```

But this feels overkill when grafana/victoriametrics recognizes the labels and can apply any histogram operations

crates/commonware-node/src/metrics.rs
hamdiallam: see above comment

crates/commonware-node/src/metrics.rs
hamdiallam: Good idea! that works

crates/commonware-node/src/metrics.rs
hamdiallam: see below comment!

bin/tempo/src/main.rs
```diff
@@ -179,6 +200,26 @@ fn main() -> eyre::Result<()> {
...
                 )
                 .fuse();
+
+                // Start the unified OTLP metrics exporter if configured
+                if let Some(config) = defaults::parse_telemetry_config(&args.telemetry)? {
+                    let mut labels = std::collections::HashMap::new();
+
+                    // Add consensus public key as node identifier
+                    if let Ok(Some(public_key)) = args.consensus.public_key() {
+                        labels.insert("consensus_id".to_string(), public_key.to_string());
```
hamdiallam: This label is applied to both reth/commonware metrics which I think is fine.

If undesired can change this. The metrics are already prefixed with `reth_` and `consensus_` which delineates the two

scripts/consensus/start-network.sh
SuperFluffy, line 1: Is this being actively used at all?

crates/node/src/telemetry.rs
```diff
@@ -0,0 +1,95 @@
...
+            let mut request = client
+                .post(&endpoint)
+                .header("Content-Type", "text/plain")
+                .body(body);
+
+            if let Some(ref auth) = auth_header {
+                request = request.header("Authorization", auth);
+            }
+
+            match request.send().await {
```
SuperFluffy: This is a long running task. The metrics that are emitted here will never the light of day when doing OTEL. Create an ad-hoc span and report the events under that span, something like this:

```rust
warn_span!("export metrics", %endpoint).in_scope(|| warn!(<msg>));
```

bin/tempo/src/main.rs
```diff
@@ -110,12 +114,31 @@ fn main() -> eyre::Result<()> {
...
-    let cli = Cli::<
+    let mut cli = Cli::<
         TempoChainSpecParser,
         TempoArgs,
         DefaultRpcModuleValidator,
         tempo_cmd::TempoSubcommand,
     >::parse();
+
+    // If telemetry is enabled, set logs OTLP and check for conflicts
+    if let Commands::Node(ref node_cmd) = cli.command
```
SuperFluffy: This is an old rust-ism. Explicit reborrowing fell out of style ca 2017. Prefer
```suggestion
    if let Commands::Node(node_cmd) = &cli.command
```

bin/tempo/src/defaults.rs
```diff
@@ -57,3 +94,61 @@ pub(crate) fn init_defaults() {
...
     init_txpool_defaults();
 }
+
+/// Parses the telemetry args into a [`TelemetryConfig`].
+///
+/// Returns `None` if telemetry is not enabled (no URL provided).
+pub(crate) fn parse_telemetry_config(
+    args: &TelemetryArgs,
+) -> eyre::Result<Option<TelemetryConfig>> {
+    let Some(ref telemetry_url) = args.telemetry_url else {
```
SuperFluffy:
```suggestion
    let Some(telemetry_url) = &args.telemetry_url else {
```

bin/tempo/src/defaults.rs
```diff
@@ -1,10 +1,47 @@
...
+
+/// CLI arguments for telemetry configuration.
+#[derive(Debug, Clone, PartialEq, Eq, clap::Args)]
+pub(crate) struct TelemetryArgs {
+    /// Enables telemetry export (OTLP logs & Prometheus metrics push). Coupled
+    /// to VictoriaMetrics which supports both with different api paths.
+    ///
+    /// The URL must include credentials: `https://user:pass@metrics.example.com`
+    #[arg(long, value_name = "URL")]
+    pub telemetry_url: Option<String>,
```
SuperFluffy: If you make this a `Url`, you don't need to check this again the function body.
```suggestion
    pub telemetry_url: Option<Url>,
```

bin/tempo/src/defaults.rs
```diff
@@ -57,3 +94,61 @@ pub(crate) fn init_defaults() {
...
+        return Ok(None);
+    };
+
+    let mut url =
+        Url::parse(telemetry_url).map_err(|e| eyre::eyre!("--telemetry-url: invalid URL: {e}"))?;
+
+    // Extract credentials - both username and password are required
+    let username = url.username();
+    let password = url.password();
+    if username.is_empty() || password.is_none() {
```
SuperFluffy: Use `eyre::ensure!` for this.

bin/tempo/src/defaults.rs
```diff
@@ -57,3 +94,61 @@ pub(crate) fn init_defaults() {
...
+        return Err(eyre::eyre!(
+            "--telemetry-url must include credentials (username and password).\n\
```
SuperFluffy: Not sure in how far people will like providing plain text here. @kamsz ?

bin/tempo/src/defaults.rs
```diff
@@ -57,3 +94,61 @@ pub(crate) fn init_defaults() {
...
+    url.set_username("").ok();
+    url.set_password(None).ok();
+    let base_url_no_creds = url.as_str().trim_end_matches('/');
+
+    // Build logs OTLP URL (Victoria Metrics OTLP path)
+    let logs_otlp_url = Url::parse(&format!("{base_url_no_creds}/opentelemetry/v1/logs"))
+        .map_err(|e| eyre::eyre!("failed to construct logs OTLP URL: {e}"))?;
+
+    // Build metrics prometheus URL (Victoria Metrics Prometheus import path)
+    let metrics_prometheus_url = format!("{base_url_no_creds}/api/v1/import/prometheus");
```
SuperFluffy: Same here - `Url::join` the path here to get a new URL.

bin/tempo/src/defaults.rs
```diff
@@ -1,10 +1,47 @@
...
+/// Telemetry configuration derived from a unified telemetry URL.
+#[derive(Debug, Clone)]
+pub(crate) struct TelemetryConfig {
+    /// OTLP logs endpoint (without credentials).
+    pub logs_otlp_url: Url,
+    /// OTLP logs filter level.
+    pub logs_otlp_filter: String,
+    /// Prometheus metrics push endpoint (without credentials).
+    /// Used for both consensus and execution metrics.
+    pub metrics_prometheus_url: String,
```
SuperFluffy: Can't we treat this as a `Url`?

bin/tempo/src/defaults.rs
```diff
@@ -57,3 +94,61 @@ pub(crate) fn init_defaults() {
...
+        let header_value = format!("Authorization={auth_header}");
+        unsafe {
+            std::env::set_var("OTEL_EXPORTER_OTLP_HEADERS", header_value);
+        }
+    }
+
+    // Build URL without credentials
+    url.set_username("").ok();
+    url.set_password(None).ok();
+    let base_url_no_creds = url.as_str().trim_end_matches('/');
```
SuperFluffy: Unnecessary, see blow

bin/tempo/src/main.rs
```diff
@@ -110,12 +114,31 @@ fn main() -> eyre::Result<()> {
...
+                "--telemetry-url and --logs.otlp cannot both be specified"
+            ));
+        }
+
+        // Set Reth logs OTLP. Consensus logs are exported as well via the same tracing system.
+        cli.traces.logs_otlp = Some(config.logs_otlp_url);
+        cli.traces.logs_otlp_filter = config
+            .logs_otlp_filter
+            .parse()
+            .expect("invalid default logs filter");
```
SuperFluffy: We are in an error environment - prefer `.wrap_err("failed to parse OTLP filter directives"?;`

bin/tempo/src/main.rs
```diff
@@ -110,12 +114,31 @@ fn main() -> eyre::Result<()> {
...
+    let mut cli = Cli::<
         TempoChainSpecParser,
         TempoArgs,
         DefaultRpcModuleValidator,
         tempo_cmd::TempoSubcommand,
     >::parse();
+
+    // If telemetry is enabled, set logs OTLP and check for conflicts
+    if let Commands::Node(ref node_cmd) = cli.command
+        && let Some(config) = defaults::parse_telemetry_config(&node_cmd.ext.telemetry)?
```
SuperFluffy: Context: `.wrap_err("failed to parse telemtry config")?`

bin/tempo/src/main.rs
```diff
@@ -110,12 +114,31 @@ fn main() -> eyre::Result<()> {
...
+    {
+        if cli.traces.logs_otlp.is_some() {
+            return Err(eyre::eyre!(
+                "--telemetry-url and --logs.otlp cannot both be specified"
```
SuperFluffy: If that's the case, then ensure that this is already checked in the args (see comment there).

bin/tempo/src/defaults.rs
```diff
@@ -57,3 +94,61 @@ pub(crate) fn init_defaults() {
...
+
+    // Build logs OTLP URL (Victoria Metrics OTLP path)
+    let logs_otlp_url = Url::parse(&format!("{base_url_no_creds}/opentelemetry/v1/logs"))
```
SuperFluffy: Don't re-parse `URL`. Instead, use `url.join("opentelemetry/v1/logs")`.

bin/tempo/src/defaults.rs
```diff
@@ -1,10 +1,47 @@
...
+const DEFAULT_LOGS_OTLP_FILTER: &str = "debug";
+
+/// CLI arguments for telemetry configuration.
+#[derive(Debug, Clone, PartialEq, Eq, clap::Args)]
+pub(crate) struct TelemetryArgs {
+    /// Enables telemetry export (OTLP logs & Prometheus metrics push). Coupled
+    /// to VictoriaMetrics which supports both with different api paths.
+    ///
+    /// The URL must include credentials: `https://user:pass@metrics.example.com`
+    #[arg(long, value_name = "URL")]
```
SuperFluffy: Based off of what you did in the function body, I think this should read:
```suggestion
    #[arg(long, value_name = "URL", conflicts_with = "logs.otlp")]
```

crates/node/src/telemetry.rs
hamdiallam: Are there any other events to report here besides the error? Feels noisy to log on each loop. wdyt?

As i understand the span is useful if there are many events

bin/tempo/src/main.rs
hamdiallam: Ah nice, wasn't aware of this clap options. Updated to remove this

scripts/consensus/start-network.sh
hamdiallam: Not sure

hamdiallam: `just localnet` spins up a node in dev mode so no consensus layer. This script spins up validators which is distinctly useful

bin/tempo/src/main.rs
hamdiallam: This seemed to not work.. Reverted the change but testing it again locally

hamdiallam: nvm got it! by id, not flag name

crates/node/src/telemetry.rs
SuperFluffy: In OTEL, the span is needed to emit any events.

In this case it doesn't really matter if we attach to the span or to the event.

Re noisiness - this only matters when something goes wrong. And there it *really* matters.

bin/tempo/src/defaults.rs
```diff
@@ -57,3 +94,61 @@ pub(crate) fn init_defaults() {
...
+    let credentials = format!("{}:{}", username, password.unwrap());
+    let encoded = BASE64_STANDARD.encode(credentials.as_bytes());
+    let auth_header = format!("Basic {encoded}");
+
+    // Set OTEL_EXPORTER_OTLP_HEADERS for OTLP logs authentication
+    // SAFETY: This is called at startup before any other threads are spawned
+    if std::env::var_os("OTEL_EXPORTER_OTLP_HEADERS").is_none() {
+        let header_value = format!("Authorization={auth_header}");
+        unsafe {
+            std::env::set_var("OTEL_EXPORTER_OTLP_HEADERS", header_value);
```
SuperFluffy: This unsafe assumption is immediately violated because this is called in 2 different places from main.rs

PR #2174, opened 2026-01-20
fix(test): use correct PRICE_SCALE in StablecoinDEX fuzz test

gakonst:
## Summary

Fixes the failing fuzz test `test_PlaceOrder_SucceedsAbove_MinimumOrderSize` in [PR #2163](https://github.com/tempoxyz/tempo/pull/2163).

## Root Cause

The test was using **hardcoded** values:
- `priceScale = 1_000_000`
- `price = 1_000_100`

But the actual `StablecoinDEX` contract uses `PRICE_SCALE = 100_000`, making `tickToPrice(100) = 100_100`.

This caused `maxAmount` to be calculated as ~9.999e21 instead of ~9.990e21, allowing fuzz inputs that result in escrow amounts exceeding the test account's `INITIAL_BALANCE`.

## Fix

Changed the test to dynamically read values from the exchange contract:
```solidity
uint256 priceScale = exchange.PRICE_SCALE();
uint256 price = exchange.tickToPrice(100);
```

This ensures the test stays in sync with any future changes to these constants.

## Verification

The counterexample from CI was `9998999999999999999999`:
- With wrong PRICE_SCALE (1M): escrow = 9.999e21 ✓ (incorrectly allowed)
- With correct PRICE_SCALE (100k): escrow = 1.001e22 ✗ (correctly exceeds balance)

Fixes: https://github.com/tempoxyz/tempo/actions/runs/21163045701/job/60861409353

### DIFF
--- a/docs/specs/test/StablecoinDEX.t.sol
+++ b/docs/specs/test/StablecoinDEX.t.sol
@@ -606,11 +606,14 @@ contract StablecoinDEXTest is BaseTest {
     function test_PlaceOrder_SucceedsAbove_MinimumOrderSize(uint128 amount) public {
         // For bid orders (buying token1 with pathUSD), the escrow amount uses ceiling division:
         // escrow = ceil(amount * tickToPrice(100) / PRICE_SCALE)
-        // escrow = ceil(amount * (PRICE_SCALE + 100) / PRICE_SCALE)
-        // escrow = ceil(amount * 1000100 / 1000000)
-        // We need escrow <= INITIAL_BALANCE
-        // Account for ceiling rounding by subtracting 1 from maxAmount
-        uint128 maxAmount = uint128((uint256(INITIAL_BALANCE) * 1_000_000) / 1_000_100) - 1;
+        //        = (amount * price + PRICE_SCALE - 1) / PRICE_SCALE
+        // We need escrow <= INITIAL_BALANCE, so:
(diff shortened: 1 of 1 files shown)


docs/specs/test/StablecoinDEX.t.sol
```diff
@@ -606,11 +606,14 @@ contract StablecoinDEXTest is BaseTest {
...
-        // We need escrow <= INITIAL_BALANCE
-        // Account for ceiling rounding by subtracting 1 from maxAmount
-        uint128 maxAmount = uint128((uint256(INITIAL_BALANCE) * 1_000_000) / 1_000_100) - 1;
+        //        = (amount * price + PRICE_SCALE - 1) / PRICE_SCALE
+        // We need escrow <= INITIAL_BALANCE, so:
+        // amount * price + PRICE_SCALE - 1 <= INITIAL_BALANCE * PRICE_SCALE
+        // amount <= (INITIAL_BALANCE * PRICE_SCALE - PRICE_SCALE + 1) / price
+        uint256 priceScale = exchange.PRICE_SCALE();
+        uint256 price = exchange.tickToPrice(100);
+        uint128 maxAmount =
```
legion2002, line 615: Hmm, this equation is not the same as the one above. 
It should have been
```
 uint128((uint256(INITIAL_BALANCE) * priceScale - price) / price);
```

Is this change expected? Because PR desc says that we are just changing hardcoded constants to dynamic retrieval.

legion2002 (approved):

PR #2180, opened 2026-01-20
fix(primitives): limit TempoTransaction calls to prevent DoS attacks

fgimenez:
Closes CHAIN-488

Adds validation limits during RLP decoding to prevent memory exhaustion attacks on validators. Without these limits, an attacker could craft malicious transactions with many calls or large inputs that consume excessive memory during decoding, causing validator DoS before the transaction is even validated or charged gas.

- Max 32 calls per transaction: prevents attackers from sending transactions with thousands of calls
- Max 128KB input data per call: limits memory consumption from oversized call inputs
- Validation occurs during decode (before memory allocation), not just in `validate()`, ensuring attackers cannot exhaust memory before gas payment


crates/primitives/src/transaction/tempo_transaction.rs
```diff
@@ -21,6 +21,12 @@ pub const SECP256K1_SIGNATURE_LENGTH: usize = 65;
 pub const P256_SIGNATURE_LENGTH: usize = 129;
 pub const MAX_WEBAUTHN_SIGNATURE_LENGTH: usize = 2048; // 2KB max
 
+/// Maximum number of calls per transaction (DoS protection)
+pub const MAX_CALLS: usize = 32;
+
+/// Maximum size of input data per call in bytes (128KB, DoS protection)
+pub const MAX_CALL_INPUT_SIZE: usize = 128 * 1024;
```
mattsse: where are these coming from?

if these are docs settings then this shouldnt be hardcoded in the rlp decoding logic imo because technically these are valid txs?

fgimenez: agree, makes more sense to enforce these limits at the txpool validation level

0xrusowsky: test comment

crates/transaction-pool/src/validator.rs
```diff
@@ -34,6 +34,12 @@ const AA_VALID_BEFORE_MIN_SECS: u64 = 3;
 /// Default maximum number of authorizations allowed in an AA transaction's authorization list.
 pub const DEFAULT_MAX_TEMPO_AUTHORIZATIONS: usize = 16;
 
+/// Maximum number of calls allowed per AA transaction (DoS protection).
+pub const MAX_AA_CALLS: usize = 32;
+
+/// Maximum size of input data per call in bytes (128KB, DoS protection).
+pub const MAX_CALL_INPUT_SIZE: usize = 128 * 1024;
```
mattsse, line 41: I think these are sensible ad we can tune

0xrusowsky (approved):

PR #2181, opened 2026-01-20
TIP-1017 Validator Config V2

SuperFluffy:
Rendered [TIP-1017](https://github.com/tempoxyz/tempo/blob/janis/val-cfg-v2/tips/tip-1017.mdx)

### DIFF
--- a/tips/tip-1017.mdx
+++ b/tips/tip-1017.mdx
@@ -53,12 +53,12 @@ By recording `addedAtHeight` and `deactivatedAtHeight`, nodes can determine DKG 
 
 ```
 players(E+1) = validators.filter(v =>
-    v.addedAtHeight <= boundary(E-1) &&
+    v.addedAtHeight <= boundary(E) &&
     (v.deactivatedAtHeight == 0 || v.deactivatedAtHeight > boundary(E))
 )
 ```
 
-Note the asymmetry: addition has a one-epoch warmup (`boundary(E-1)`), while deactivation takes effect immediately (`boundary(E)`).
+Both addition and deactivation take effect at the next epoch boundary—there is no warmup or cooldown period.
 
 This eliminates the need to retain historical account state for consensus recovery—nodes can derive DKG player sets from the current contract state alone.
 
@@ -146,8 +146,8 @@ interface IValidatorConfigV2 {
 
     /// @notice Get the height at which the contract was initialized
     /// @return the height at which the contract was initialized. Note that this
-    ///         value only makes sense in conjuction with isInitialized()
-    function getInitializedAtHeight() external view turns (uint64);
+    ///         value only makes sense in conjunction with isInitialized()
+    function getInitializedAtHeight() external view returns (uint64);
 
     /// @notice Get the owner of the precompile
     /// @return The owner address
@@ -263,8 +263,8 @@ interface IValidatorConfigV2 {
     /// @notice Migrate a single validator from V1 to V2 (owner only)
     /// @dev Can be called multiple times to migrate validators one at a time.
     ///      On first call, copies owner from V1 if V2 owner is address(0).
-    ///      Active V1 validators get addedAtHeight=0 and deactivatedAtHeight=0.
-    ///      Inactive V1 validators get addedAtHeight=deactivatedAtHeight=block.height at migration time.
+    ///      Active V1 validators get addedAtHeight=block.height and deactivatedAtHeight=0.
+    ///      Inactive V1 validators get addedAtHeight=block.height and deactivatedAtHeight=block.height at migration time.
     ///      Reverts if already initialized or already migrated.
     ///      Reverts if idx != validatorsArray.length.
     ///      Reverts if `V2.isInitialized()` (no migrations after V2 is initialized).
@@ -273,8 +273,8 @@ interface IValidatorConfigV2 {
 
     /// @notice Initialize V2 and enable reads (owner only)
     /// @dev Should only be called after all validators have been migrated via migrateValidator.
-    ///      Sets initialized=true and initializedAtheight=block.height. After this call
-     ///     CL reads from V2 instead of V1.
+    ///      Sets initialized=true and initializedAtHeight=block.height. After this call,
+    ///      CL reads from V2 instead of V1.
     ///      Copies nextDkgCeremony from V1.
     ///      Reverts if V2 validators count < V1 validators count (ensures all validators migrated).
     ///      Reverts if validator activity does not match between contracts:
@@ -367,7 +367,7 @@ activeValidators(E+1) = dkgOutcome(boundary(E)).players.map(pubkey =>
 
 This distinction matters because:
 - The DKG can fail, reverting to the previous DKG outcome (for example, not all
-  eligible players may successfully participate in the the DKG due to network
+  eligible players may successfully participate in the DKG due to network
   issues, crashes, etc.)
 - The DKG outcome is the authoritative record of who holds valid key shares
 - Only validators with valid shares can produce valid signatures in epoch `E+1`
@@ -457,7 +457,7 @@ The following invariants must always hold:
 
 3. **Address update**: `transferValidatorOwnership` updates the address of an
     existing validator without changing any other fields. This can only be called
-    by the contract owner (not the owner of the entry) and is required to for
+    by the contract owner (not the owner of the entry) and is required for
     post-migration fixes: validator contract v1 contains unusable dummy addresses
     that are not usable.
 
@@ -486,14 +486,14 @@ The following invariants must always hold:
    changes.
 
 11. **Owner authorization**: Only the owner can call `addValidator`,
-    `transferOwner`, `migrateValidator`, `initiatializeIfMigraated` and
+    `transferOwner`, `migrateValidator`, `initializeIfMigrated`, and
     `setNextFullDkgCeremony`.
 
 12. **Dual authorization**: `rotateValidator`, `deactivateValidator`,
     `transferValidatorOwnership`, and `setIpAddresses` can be called by either
     the owner or the validator itself.
 
-13. **Initialized once**: The `initializedAtHeight` bit (bit 255 of slot 0) can only
+13. **Initialized once**: The `initialized` bit (bit 255 of slot 0) can only
     transition from 0 to 1, never back to 0.
 
 14. **Atomic rotation**: `rotateValidator` atomically deletes the old validator
@@ -559,8 +559,8 @@ which validators have already been migrated.
 
 The CL checks `v2.isInitialized()` to determine which contract to read:
 
-- On `initialized == false`**: CL reads from V1
-- On `initialized == true`**: CL reads from V2
+- **`initialized == false`**: CL reads from V1
+- **`initialized == true`**: CL reads from V2
 
 This is handled entirely in the CL logic, not in the V2 precompile. The V2
 precompile does NOT proxy reads to V1.
@@ -575,10 +575,10 @@ precompile does NOT proxy reads to V1.
 - Creates a V2 validator entry with:
   - `publicKey`: copied from V1
   - `validatorAddress`: copied from V1
-  - `inboundAddress`: copied from V1
-  - `outboundAddress`: copied from V1
+  - `ingress`: copied from V1 `inboundAddress`
+  - `egress`: copied from V1 `outboundAddress`
   - `index`: set to `idx`
-  - `addedAtHeight`: `block.height`
+  - `addedAtHeight`: `0`
   - `deactivatedAtHeight`: set to `0` if V1 `active == true`, otherwise `block.height`
 - Adds to `validatorsArray`
 - Populates lookup maps (`validators[validatorAddress]`, `pubkeyToIndex[pubkey]`)
@@ -623,8 +623,8 @@ interface IValidatorConfigV1 {
 - **No signatures required**: V1 validators are imported without Ed25519 signatures
   (they were already validated in V1)
 - **All validators imported**: Both active and inactive V1 validators are
-  imported; active ones have `addedAtHeght.block.height`, inactive ones have
-  `addedAtHeight == deactivatedAtHeight == block.height`.
+  imported; active ones have `addedAtHeight == block.height` and `deactivatedAtheight = 0`,
+  inactive ones have `addedAtHeight == deactivatedAtHeight == block.height`.
 - **Initialize validates migration**: `initializeIfMigrated()` reverts if not
   all V1 validators have been migrated
 
@@ -742,39 +742,39 @@ The test suite must cover:
     if `isInitialized() == true`
 36. **migrateValidator owner only**: Non-owner cannot call `migrateValidator`
 37. **All validators imported on migration**: Both V1 active and inactive validators
-    are imported; active ones have `addedAtHeight = block.height`, inactive ones have
-    `addedAtHeight == deactivatedAtHeight == block.height`.
+    are imported; active ones have `addedAtHeight > 0` and `deactivatedAtHeight == 0`
+    inactive ones have `addedAtHeight == deactivatedAtHeight > 0`.
 38. **addedAtHeight set correctly**: Active validators have `addedAtHeight > 0`
     at migration time.
-38. **deletedAtHeight set correctly**: Deleted validators have
-    `addedAtHeight == deletedAtHeight` at migration time.
-39. **initialize sets flag**: After `initializeIfMigrated()`, `isInitialized()` returns true
-40. **migrateValidator copies owner**: V2 `owner()` matches V1 after first `migrateValidator` call
-41. **initialize copies DKG ceremony**: V2 `getNextFullDkgCeremony()` matches V1
+39. **deactivatedAtHeight set correctly**: Inactive validators have
+    `deactivatedAtHeight > 0` at migration time.
+40. **initialize sets flag**: After `initializeIfMigrated()`, `isInitialized()` returns true
+41. **migrateValidator copies owner**: V2 `owner()` matches V1 after first `migrateValidator` call
+42. **initialize copies DKG ceremony**: V2 `getNextFullDkgCeremony()` matches V1
     after `initializeIfMigrated()`
-42. **initialize owner only**: Non-owner cannot call `initialize`
-43. **isInitialized returns correct value**: Returns false before initialize, true after
-44. **Writes blocked before init**: `addValidator`, `deactivateValidator`, `rotateValidator`
-45. **initialize reverts if not all migrated**: `initializeIfMigrated()` reverts if
+43. **initialize owner only**: Non-owner cannot call `initialize`
+44. **isInitialized returns correct value**: Returns false before initialize, true after
+45. **Writes blocked before init**: `addValidator`, `deactivateValidator`, `rotateValidator`
+46. **initialize reverts if not all migrated**: `initializeIfMigrated()` reverts if
     `validatorsArray.length < V1.getValidators().length`
 
 ### transferValidatorOwnership
 
-46. **transferValidatorOwnership by owner**: Owner can transfer any validator to a new address
-47. **transferValidatorOwnership by validator**: Validator can transfer themselves to a new address
-48. **transferValidatorOwnership unauthorized**: Non-owner and non-validator cannot transfer
-49. **transferValidatorOwnership reverts on invalid validator**: Reverts if `validatorAddress` does not exist
-50. **transferValidatorOwnership reverts on duplicate address**: Reverts if `newAddress` already exists
-51. **transferValidatorOwnership updates lookup maps**: Old address is removed, new address is added to lookup
+47. **transferValidatorOwnership by owner**: Owner can transfer any validator to a new address
+48. **transferValidatorOwnership by validator**: Validator can transfer themselves to a new address
+49. **transferValidatorOwnership unauthorized**: Non-owner and non-validator cannot transfer
+50. **transferValidatorOwnership reverts on invalid validator**: Reverts if `validatorAddress` does not exist
+51. **transferValidatorOwnership reverts on duplicate address**: Reverts if `newAddress` already exists
+52. **transferValidatorOwnership updates lookup maps**: Old address is removed, new address is added to lookup
 
 ### setIpAddresses
 
-52. **setIpAddresses by owner**: Owner can update any validator's IP addresses
-53. **setIpAddresses by validator**: Validator can update their own IP addresses
-54. **setIpAddresses unauthorized**: Non-owner and non-validator cannot update IP addresses
-55. **setIpAddresses reverts on invalid validator**: Reverts if validator does not exist
-56. **setIpAddresses reverts on deleted validator**: Reverts if validator is already deleted
-57. **setIpAddresses validates format**: Rejects invalid `<ip>:<port>` format
+53. **setIpAddresses by owner**: Owner can update any validator's IP addresses
+54. **setIpAddresses by validator**: Validator can update their own IP addresses
+55. **setIpAddresses unauthorized**: Non-owner and non-validator cannot update IP addresses
+56. **setIpAddresses reverts on invalid validator**: Reverts if validator does not exist
+57. **setIpAddresses reverts on deactivated validator**: Reverts if validator is already deactivated
+58. **setIpAddresses validates format**: Rejects invalid `<ip>:<port>` format
 
 ---
 


tips/tip-1017.mdx
```diff
@@ -0,0 +1,435 @@
+# ValidatorConfig V2
```
jenpaff, line 1: please name the file TIP-1008

SuperFluffy: Will assign a number on merge time, since there are a couple other TIPs in flight.

tips/tip-1017.mdx
```diff
@@ -0,0 +1,545 @@
...
```
SuperFluffy, line 1: Should this be named `DKGConfig` or `DistributedKeyGeneration Config` instead? The fact that this is called `ValidatorConfig` is confusing, since I can't actually read who the validators *are* but only who the validators *should be* (since the validators at a given height are passed into the DKG ceremony as players).

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -0,0 +1,545 @@
...
+This eliminates the need to retain historical account state for consensus recovery—nodes can derive DKG player sets from the current contract state alone.
+
+---
+
+# Specification
+
+## Precompile Address
+
+```solidity
+address constant VALIDATOR_CONFIG_V2_ADDRESS = 0xCCCCCCCC00000000000000000000000000000002;
```
SuperFluffy: Current address `0xCCCCCCCC00000000000000000000000000000000`, make this `0xCCCCCCCC00000000000000000000000000000001` instead?

tips/tip-1017.mdx
Zygimantass: validatorconfig is less confusing to operators imo

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -0,0 +1,545 @@
...
+   queries
+
+```
+┌─────────────┐     addValidator()     ┌─────────────┐     deleteValidator()     ┌─────────────┐
+│             │ ────────────────────►  │             │ ────────────────────────► │             │
+│  Not Exist  │                        │   Active    │                           │   Deleted   │
+│             │                        │ deleted=0   │                           │ deleted>0   │
+└─────────────┘                        └─────────────┘                           └─────────────┘
+                                              │                                         │
+                                              │◄────────────────────────────────────────┘
```
dankrad:
```suggestion
                                              │◄────────────────────────/───────────────┘
```

tips/tip-1017.mdx
```diff
@@ -0,0 +1,545 @@
...
+                                              │         (No transition back)
+```
+
+### Signature Verification
+
+When adding a validator, the caller must provide an Ed25519 signature proving
+ownership of the public key:
+
+```
+message = keccak256(abi.encodePacked(
```
dankrad, line 323: This message format doesn't have any form of domain separation (so that no other message type can potentially collide with it); here is how the beacon chain does it:
https://github.com/ethereum/consensus-specs/blob/d9f40e4a6d8c8b7d2e13a427208381dfbafb342b/specs/phase0/beacon-chain.md#compute_signing_root

Idk how commonware currently signs validator messages but if it defines a standard we should use that

dankrad: Here is the signing function Commonware uses:

https://github.com/subversive-upstream/commonware-cryptography/blob/main/src/ed25519/scheme.rs#L40-L62

with an example domain here: https://github.com/subversive-upstream/commonware-consensus/blob/main/src/aggregation/types.rs#L82

So I suggest we use the same scheme as Commonware and create a new domain, e.g. b"_TEMPO_VALIDATOR_REGISTRATION"

SuperFluffy: Yeah, we are (implicitly) using the `b"TEMPO"` namespace as a prefix - it used to be much more explicit but now rely on the commonware dkg APIs, so it's more tucked away.

I have added `b"TEMPO"` and `b"_VALIDATOR_CONFIG_V2_ADD_VALIDATOR"` as extensions to the namespace: [cbf2d15](https://github.com/tempoxyz/tempo/pull/2181/commits/cbf2d155ae17fa522c4a66293ba2ced4032742cf)

SuperFluffy: As an aside: I am not sure if we should prescribe the ed25519 signature over keccak here, or simply follow commonware by committing the bytes directly to blake3.

Similarly, not sure if `abi.encodePacked` is an unnecessary extra step.

tips/tip-1017.mdx
```diff
@@ -0,0 +1,547 @@
...
+
+3. **Validator re-registration**: V1 allows deleted validators to be re-added
+   with different parameters, complicating historical queries.
+
+V2 solves these problems with an append-only design where:
+- Validators are immutable after creation (no `updateValidator`)
+- `addedAtHeight` and `deletedAtHeight` fields enable historical reconstruction
+  from current state
+- Ed25519 signature verification proves key ownership at registration time
+- Both address and public key remain reserved forever (even after deletion)
```
legion2002, line 48: This has some state bloat implications so we should make sure to charge the proper 250k gas here for each SSTORE.

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -0,0 +1,547 @@
...
+    /// @param validatorAddress The address of the new validator
+    /// @param publicKey The validator's Ed25519 communication public key
+    /// @param inboundAddress The validator's inbound address `<hostname|ip>:<port>` for incoming connections
+    /// @param outboundAddress The validator's outbound IP address `<ip>:<port>` for firewall whitelisting
+    /// @param signature Ed25519 signature (64 bytes) proving ownership of the public key
+    function addValidator(
+        address validatorAddress,
+        bytes32 publicKey,
+        string calldata inboundAddress,
+        string caaaaalldata outboundAddress,
```
legion2002: lol, this should be calldata

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -0,0 +1,547 @@
...
+1. Validator is added at height 1050 during epoch 10, confirmed at `boundary(10) = 1000`
+2. Epoch 11 (heights 1001–1100): The validator syncs with the network
+3. Epoch 12 (heights 1101–1200): The validator participates as a DKG player and receives a share (on success)
+4. Epoch 13 (heights 1201–1300): The validator becomes a signer using the share from the epoch 12 DKG
+
+**Deletion follows the same pattern**: A deleted validator is excluded from epoch `E+2` if `deletedAtHeight <= boundary(E)`.
+
+The complete filter for DKG players at epoch `E+2`:
+```
+players = validators.filter(v =>
```
legion2002: should we be maintaining a nicer onchain DS so that this information is easily available?

tips/tip-1017.mdx
```diff
@@ -0,0 +1,547 @@
...
+        bytes32 publicKey,
+        string calldata inboundAddress,
+        string caaaaalldata outboundAddress,
+        bytes calldata signature
+    ) external;
+
+    /// @notice Delete a validator (owner only)
+    /// @dev Marks the validator as deleted by setting deletedAtHeight to the current block height.
+    ///      The validator's entry remains in storage for historical queries.
+    ///      The address and public key remain reserved and cannot be reused.
```
legion2002, line 202: It's acceptable that you can't reuse the same validator again once deleted? 
Ig it must be easy to switch your validator's ED25519 to a new keypair?

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -0,0 +1,547 @@
...
+    outboundAddress                       // string: Binds network configuration
+))
+```
+
+The signature must be exactly 64 bytes and must verify against the provided
+public key.
+
+### DKG Player Selection
+
+The DKG manager uses the validator set from this contract to determine participants in each DKG ceremony. To ensure validators have sufficient time to join the network and synchronize, there is a **one-epoch gap** between registration and DKG participation.
```
legion2002: What's our epoch time right now? If the epoch time is days, then do we need this extra epoch gap for syncing?

tips/tip-1017.mdx
```diff
@@ -0,0 +1,547 @@
...
+```
+players = validators.filter(v =>
+    v.addedAtHeight <= boundary(E) &&
+    (v.deletedAtHeight == 0 || v.deletedAtHeight > boundary(E))
+)
+```
+
+### Determining Active Validators
+
+Reading this contract alone is **not sufficient** to determine who the active validators (signers) are during a given epoch. The contract only records which validators are *eligible* to participate in DKG ceremonies—it does not record DKG outcomes.
```
legion2002, line 354: Should we be updating the contract with the real active validator set at the epoch boundary?
Ig it could be useful for bridges

tips/tip-1017.mdx
SuperFluffy: It does indeed incur state bloat, but this is entirely controlled by the ADMIN account. Should we really be charging gas for "maintenance"?

tips/tip-1017.mdx
SuperFluffy: Yes, because the orchestration of this is not trivial and it also has implications for nodes catching up across epoch boundaries.

It's easiest if we require generating a new identity for a new validator.

docs/pages/protocol/tips/tip-xxxx.mdx
SuperFluffy: Current epoch length is 21,600 blocks, so 3h at 0.5s block times.

For epochs that are days long an argument could be made to remove this (although commonware was pushing for this "warmup time" even at epoch lengths of 7 days).

tips/tip-1017.mdx
SuperFluffy: It would be very useful, but it would require validators to keep a wallet so that they can push that data to state.

I am not sure if a system transaction would work here?

tips/tip-1017.mdx
```diff
@@ -0,0 +1,547 @@
...
+   potential key squatting attacks.
+
+3. **Validator re-registration**: V1 allows deleted validators to be re-added
+   with different parameters, complicating historical queries.
+
+V2 solves these problems with an append-only design where:
+- Validators are immutable after creation (no `updateValidator`)
+- `addedAtHeight` and `deletedAtHeight` fields enable historical reconstruction
+  from current state
+- Ed25519 signature verification proves key ownership at registration time
```
Zygimantass, line 47: can you ask amp to try doing this - just want to make sure that we can verify cw ed25519 keys - i assume there should be no reason why not

docs/pages/protocol/tips/tip-xxxx.mdx
SuperFluffy: > should we be maintaining a nicer onchain DS so that this information is easily available?

What does DS stand for?

In general this would be very nice to have. But as far as I know contracts don't have access to headers?

tips/tip-1017.mdx
SuperFluffy: I think we would just do that in the precompile? There anything goes, it's Rust after all.

IIRC we even wanted to remove the solidity smart contracts from the spec?

tips/tip-1017.mdx
legion2002: I think we should be able to do this with system txs

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -0,0 +1,547 @@
...
+2. **Key ownership verification**: V1 does not verify that the caller controls
+   the private key corresponding to the public key being registered, allowing
+   potential key squatting attacks.
+
+3. **Validator re-registration**: V1 allows deleted validators to be re-added
+   with different parameters, complicating historical queries.
+
+V2 solves these problems with an append-only design where:
+- Validators are immutable after creation (no `updateValidator`)
+- `addedAtHeight` and `deletedAtHeight` fields enable historical reconstruction
```
Zygimantass: this means we're just making the "if you want to change IP, you need to rotate your identity" property explicit - right? and bumping rotating IP to v3?

tips/tip-1017.mdx
Zygimantass: right - true

docs/pages/protocol/tips/tip-xxxx.mdx
legion2002: Data Structure. I think we can give them access to headers using a system tx

docs/pages/protocol/tips/tip-xxxx.mdx
Zygimantass: yep, zero indexed versioning works

docs/pages/protocol/tips/tip-xxxx.mdx
Zygimantass: we already expose getValidators() onchain - i think this is enough? DKG outcome is separate from this TIP

tips/tip-1017.mdx
Zygimantass: let's keep this separate from ValidatorConfig for simplicity

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -0,0 +1,547 @@
...
+- The DKG can fail, reverting to the previous DKG outcome (for example, not all
+  eligible players may successfully participate in the the DKG due to network
+  issues, crashes, etc.)
+- The DKG outcome is the authoritative record of who holds valid key shares
+- Only validators with valid shares can produce valid signatures in epoch `E+1`
+
+### Address Validation
+
+- **inboundAddress**: Must be in `<hostname|ip>:<port>` format. Hostnames and IP
+  addresses are both accepted.
```
Zygimantass: if we don't want IPs to change, let's make it IP only

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -0,0 +1,547 @@
...
+3. **Bootstrap V2 from V1 for existing networks**: At the fork, copy V1 state
+   into V2
+
+## Hardfork-Based Switching
+
+The DKG manager determines which contract to read based on the block timestamp
+at the boundary height:
+
+```
+if chainspec.is_allegro_active_at_timestamp(block.timestamp) {
```
Zygimantass:
```suggestion
if chainspec.is_hardfork_active_at_timestamp(contract_upgrade_hardfork, block.timestamp) {
```

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -0,0 +1,547 @@
...
+
+# Migration from V1
+
+This section describes the migration strategy from ValidatorConfig V1 to V2.
+
+## Overview
+
+The migration uses a three-pronged approach:
+
+1. **New hardfork ("Allegro")**: Timestamp-based activation that switches
```
Zygimantass:
```suggestion
1. **New hardfork**: Timestamp-based activation that switches
```

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -0,0 +1,547 @@
...
+mechanism:
+
+```solidity
+/// @notice Bootstrap V2 from V1 state (owner only, one-time)
+/// @dev Can only be called once. After bootstrap, normal signature requirements apply.
+/// @param migrationHeight The block height at which migration occurs
+function bootstrapFromV1(uint64 migrationHeight) external;
+```
+
+**Properties:**
```
Zygimantass: Doing this as a system transaction seems hacky - can we figure out a better migration path?

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -0,0 +1,547 @@
...
+3. **At fork activation**: Execute `bootstrapFromV1(migrationHeight)` as a
+   system transaction
+4. **Post-fork**: DKG reads V2; future validator changes use V2's `addValidator`
+   with signatures
+
+### For New Networks
+
+1. Initialize V2 at genesis with the initial validator set
+2. Optionally populate V1 for backward tooling compatibility
+3. Set `allegroTime = 0` to activate V2 immediately
```
Zygimantass:
```suggestion
3. Set `contractHardforkUpgradeTime = 0` to activate V2 immediately
```

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -0,0 +1,547 @@
...
+
+## Safety Checks
+
+Nodes should fail fast if:
+- Allegro is active but V2 validator set is empty (bootstrap didn't execute)
+- V2 `bootstrapped` is false after the fork activation block
```
Zygimantass: this is why i think the migration path is bad - it could lead to network going down and it'd be hard to recover it (since it'd be the post hardfork block)

docs/pages/protocol/tips/tip-xxxx.mdx
SuperFluffy: Yeah. I think that's the cleanest solution given our other constraints.

Note that we can actually now change the dial address because commonware baked dns resolution into their p2p layer.

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -0,0 +1,547 @@
...
+- Both address and public key remain reserved forever (even after deletion)
+
+### Key Design Principle
+
+By recording `addedAtHeight` and `deletedAtHeight`, nodes can determine DKG players for any epoch using only current state. When preparing for a DKG ceremony in epoch `E+2`, a node reads the contract at `boundary(E)` and filters:
+
+```
+players(E+2) = validators.filter(v =>
+    v.addedAtHeight <= boundary(E) &&
+    (v.deletedAtHeight == 0 || v.deletedAtHeight > boundary(E))
```
howydev: does `== 0` represent 2 things here? `deleted at epoch 0` + `not deleted`?

tips/tip-1017.mdx
```diff
@@ -0,0 +1,547 @@
...
+    /// @param outboundAddress IP address for firewall whitelisting (format: `<ip>:<port>`, no hostnames)
+    /// @param index Position in validators array (assigned at creation, immutable)
+    /// @param addedAtHeight Block height when validator was added (set by builder, immutable)
+    /// @param deletedAtHeight Block height when validator was deleted (0 = active, set by builder)
+    struct Validator {
+        bytes32 publicKey;
+        address validatorAddress;
+        string inboundAddress;
+        string outboundAddress;
+        uint64 index;
```
howydev, line 134: I think if we're never removing validators/doing append only, we could get away with just using an array of structs and not tracking `index`

howydev: Ah, we don't get `validatorByAddress` for this in a cheap way. But if we never intend to call this on-chain, we could also do `getAllValidators.filter(v -> v.address == address)`

I think we can get away with referencing everything by `index` but this would be a larger change, and I think the -1 storage slot savings isn't a big deal in the long run, defer to you on interface

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -0,0 +1,547 @@
...
+        string caaaaalldata outboundAddress,
+        bytes calldata signature
+    ) external;
+
+    /// @notice Delete a validator (owner only)
+    /// @dev Marks the validator as deleted by setting deletedAtHeight to the current block height.
+    ///      The validator's entry remains in storage for historical queries.
+    ///      The address and public key remain reserved and cannot be reused.
+    /// @param validatorAddress The validator address to delete
+    function deleteValidator(address validatorAddress) external;
```
howydev: note: should be permissioned to owner or validator only

docs/pages/protocol/tips/tip-xxxx.mdx
howydev:
```suggestion
        string calldata outboundAddress,
```

SuperFluffy: Haha, fixed in [126563d](https://github.com/tempoxyz/tempo/pull/2181/commits/126563dba90edfab62ef3069d0c263de0e3bee36)

SuperFluffy: [126563d](https://github.com/tempoxyz/tempo/pull/2181/commits/126563dba90edfab62ef3069d0c263de0e3bee36)

tips/tip-1017.mdx
```diff
@@ -257,6 +283,22 @@ message = keccak256(abi.encodePacked(
...
 public key.
 
+When rotating a validator, the signature message format is identical to `addValidator`:
+
+```
+message = keccak256(abi.encodePacked(
+    b"TEMPO",                                  // bytes: global namespace prefix 
+    b"_VALIDATOR_CONFIG_V2_ROTATE_VALIDATOR"  // bytes: contract specific namespace
+    chainId,                                   // uint64: Prevents cross-chain replay
+    contractAddress,                           // address: Prevents cross-contract replay
```
howydev, line 345: Good call on replay

Is replay in the same contract an issue? e.g. addr/key A -> rotate to addr/key B -> revert back to addr/key A b/c of a failing build, and someone replays the A -> B sig

docs/pages/protocol/tips/tip-xxxx.mdx
SuperFluffy: Added `setIpAddresses` to set ingress and egress at the same time.

docs/pages/protocol/tips/tip-xxxx.mdx
SuperFluffy: IP-only now. Just like it is in V1.

docs/pages/protocol/tips/tip-xxxx.mdx
SuperFluffy: Why the extra `contract_upgrade_hardfork` argument?

docs/pages/protocol/tips/tip-xxxx.mdx
SuperFluffy: All manual now.

docs/pages/protocol/tips/tip-xxxx.mdx
SuperFluffy: Using `<hardforkTime>` now.

docs/pages/protocol/tips/tip-xxxx.mdx
SuperFluffy: Migration path is all manual now. `initializeIfMigrated` now checks that the entries in V2 match those in V1 (including their activity status).

docs/pages/protocol/tips/tip-xxxx.mdx
SuperFluffy: No - `deletedAtHeight == 0` means not deleted, and `v.deletedAtHeight > boundary(E)` means deleted in the future. This is for consistency with the current approach.

This condition makes sense since we now want to read the validator set for E from any contract state > E.

tips/tip-1017.mdx
SuperFluffy: Shouldn't this be caught when executing? I.e., the original signed transaction carries a nonce -> can't replay the transaction if reverted?

tips/tip-1017.mdx
```diff
@@ -0,0 +1,800 @@
...
+│             │ ────────────────────►  │             │ ────────────────────────► │             │
+│  Not Exist  │                        │   Active    │                           │   Deleted   │
+│             │                        │ deleted=0   │                           │ deleted>0   │
+└─────────────┘                        └─────────────┘                           └─────────────┘
+                                              │                                         │
+                                              │◄────────────────────────/───────────────┘
+                                              │         (No transition back)
+```
+
+### Signature Verification
```
SuperFluffy, line 317: Decided to stay with abci-encoded and not go with commonware's codecs API. While right now this is all done inside precompiles we should at least theoretically keep the door open to doing this in smart contracts/solidity.

Keeping this as ABCI encoded with keccak256 is less work overall (no need to implement blake3 or CW codecs).

tips/tip-1017.mdx
```diff
@@ -0,0 +1,800 @@
...
+    ///      keccak256(abi.encodePacked(
+    ///          <NAMESPACE>, chainId, contractAddress, validatorAddress, ingress, egress
+    ///      )),
+    ///      where `<NAMESPACE>` is "TEMPO_VALIDATOR_CONFIG_V2_ROTATE_VALIDATOR".
+    /// @param validatorAddress The address of the validator to rotate
+    /// @param publicKey The new validator's Ed25519 communication public key
+    /// @param ingress The new validator's inbound address `<ip>:<port>` for incoming connections
+    /// @param egress The new validator's outbound IP address `<ip>:<port>` for firewall whitelisting
+    /// @param signature Ed25519 signature (64 bytes) proving ownership of the new public key
+    function rotateValidator(
```
howydev, line 225: i think sig replay would be an issue. for a validator that has done 2 migrations (key1 -> key2 -> key3), an observer could replay the rotation tx from (key1 -> key2) to downgrade the validator from (key3 -> key2)

Fixes here would be to add a nonce in the `validator` field that we use, or permission the call to `onlyValidatorAddress` and not have that as a call input instead

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -0,0 +1,800 @@
...
+    /// @param egress The new outbound IP address `<ip>:<port>` for firewall whitelisting
+    function setIpAddresses(
+        address validatorAddress,
+        string calldata ingress,
+        string calldata egress
+    ) external;
+
+    /// @notice Change owner (owner only)
+    /// @param newOwner The new owner address
+    function changeOwner(address newOwner) external;
```
howydev: nit: `transferOwnership`, to align with smart contract ownership standard interfaces

tips/tip-1017.mdx
```diff
@@ -0,0 +1,800 @@
...
+    ) external;
+
+    /// @notice Change owner (owner only)
+    /// @param newOwner The new owner address
+    function changeOwner(address newOwner) external;
+
+    /// @notice Set the epoch at which a fresh DKG ceremony will be triggered (owner only)
+    /// @param epoch The epoch in which to run the fresh DKG ceremony.
+    ///        Epoch N runs the ceremony, and epoch N+1 uses the new DKG polynomial.
+    function setNextFullDkgCeremony(uint64 epoch) external;
```
howydev, line 261: ultra stupid question - what is the flow for how this is updated today? this has to be updated every epoch, yes?

tips/tip-1017.mdx
howydev: Oh this is a dup: https://github.com/tempoxyz/tempo/pull/2181/changes#r2770360110

From my understanding:

The signed top level tx has a sig, that is checked/verified outside EVM, if thats valid then the call happens. The call contains a 2nd sig here. The sig verification rebuilds the digest and checks against these fields

The replay would be to send a new top level tx, but replay the 2nd signature

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -0,0 +1,800 @@
...
+
+    /// @notice Set the epoch at which a fresh DKG ceremony will be triggered (owner only)
+    /// @param epoch The epoch in which to run the fresh DKG ceremony.
+    ///        Epoch N runs the ceremony, and epoch N+1 uses the new DKG polynomial.
+    function setNextFullDkgCeremony(uint64 epoch) external;
+
+    /// @notice Migrate a single validator from V1 to V2 (owner only)
+    /// @dev Can be called multiple times to migrate validators one at a time.
+    ///      On first call, copies owner from V1 if V2 owner is address(0).
```
Zygimantass, line 265: what does "if owner is address(0)"?

tips/tip-1017.mdx
SuperFluffy: Will keep charging gas. Might change the fee to prioritize these transactions. But not in this TIP.

docs/pages/protocol/tips/tip-xxxx.mdx
SuperFluffy: Not in this TIP. But it would be valuable to think about how we can make this less confusing (i.e. difference between activate/deactivated validators, and actual members of the committee).

tips/tip-1017.mdx
SuperFluffy: No - $owner sets this once to whatever epoch. If `next_epoch == $this_epoch`, then a full DKG is run. After that epoch the epoch here is in the past and ignored.

docs/pages/protocol/tips/tip-xxxx.mdx
SuperFluffy: Overlooked from the PR that fleshed out this migration. Fixed in 3b3e3f55 (just copies V1 owner to V2 owner).

docs/pages/protocol/tips/tip-xxxx.mdx
SuperFluffy: TIL, done in f19bf145

tips/tip-1017.mdx
SuperFluffy: By "2nd sig" do you literally mean the signature argument to the function?

We have a check that only $owner or $validatorAddress are allowed to call this function. So this signature should not even be checked if that is not the case.

Am I missing something?

tips/tip-1017.mdx
SuperFluffy: The docs of the function already contains this:

```
>     /// @notice Rotate a validator to a new identity (owner or validator only)
```

The `validatorAddress` only exists to allow $owner to rotate the validator. In case the caller is not $owner, $validatorAddress must exactly match the entry on chain and must be the address of the caller.

tips/tip-1017.mdx
howydev: ok i missed that, permissioned owner would make this fine

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -0,0 +1,800 @@
...
+    /// @param field The field name that failed validation
+    /// @param input The invalid input that was provided
+    /// @param backtrace Additional error context
+    error NotIpPort(string field, string input, string backtrace);
+
+    /// @notice Validator information (V2 - append-only, delete-once)
+    /// @param publicKey Ed25519 communication public key (non-zero, unique across all validators)
+    /// @param validatorAddress Ethereum-style address of the validator (unique across all validators)
+    /// @param ingress Address where other validators can connect (format: `<ip>:<port>`)
+    /// @param egress IP address from which this validator will dial, e.g. for firewall whitelisting (format: `<ip>:<port>`)
```
hamdiallam: note here to remove port from the egress string

tips/tip-1017.mdx
```diff
@@ -0,0 +1,800 @@
...
+    ///      keccak256(abi.encodePacked(<NAMESPACE>, chainId, contractAddress, validatorAddress, ingress, egress)),
+    ///      where `<NAMESPACE>` is "TEMPO_VALIDATOR_CONFIG_V2_ADD_VALIDATOR"
+    ///      This proves the caller controls the private key corresponding to publicKey.
+    ///      Reverts if isInitialized() returns false.
+    /// @param validatorAddress The address of the new validator
+    /// @param publicKey The validator's Ed25519 communication public key
+    /// @param ingress The validator's inbound address `<ip>:<port>` for incoming connections
+    /// @param egress The validator's outbound IP address `<ip>:<port>` for firewall whitelisting
+    /// @param signature Ed25519 signature (64 bytes) proving ownership of the public key
+    function addValidator(
```
hamdiallam, line 191: Should we also take a ecdsa signature from validatorAddress to prove ownership? Mainly because this is called from the owner and not the validator

Implications of a typo here would be not being able to change IP nor rotate the pubkey

docs/pages/protocol/tips/tip-xxxx.mdx
SuperFluffy: 48538026

tips/tip-1017.mdx
SuperFluffy: Desirable, but we can't do much about this because there is no guarantee that all validators have a wallet/address. Since $owner can always change this value after the fact we are safe from somebody breaking this too badly.

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -0,0 +1,800 @@
...
+
+**Example timeline** (assuming epoch length of 100 blocks):
+1. Validator is added at height 1050 during epoch 10, confirmed at `boundary(10) = 1000`
+2. Epoch 11 (heights 1001–1100): The validator syncs with the network
+3. Epoch 12 (heights 1101–1200): The validator participates as a DKG player and receives a share (on success)
+4. Epoch 13 (heights 1201–1300): The validator becomes a signer using the share from the epoch 12 DKG
+
+**Deletion follows the same pattern**: A deleted validator is excluded from epoch `E+2` if `deactivatedAtHeight <= boundary(E)`.
+
+The complete filter for DKG players at epoch `E+2`:
```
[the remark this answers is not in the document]
SuperFluffy: Removed entirely.

howydev (approved):

tips/tip-1017.mdx
```diff
@@ -0,0 +1,786 @@
...
+
+    /// @notice Deactivates a validator (owner or existing validator)
+    /// @dev Marks the validator as deactivated by setting deactivatedAtHeight to the current block height.
+    ///      The validator's entry remains in storage for historical queries.
+    ///      The address and public key remain reserved and cannot be reused.
+    /// @param validatorAddress The validator address to deactivate
+    function deactivateValidator(address validatorAddress) external;
+
+    /// @notice Rotate a validator to a new identity (owner or validator only)
+    /// @dev Atomically deletes the specified validator entry and adds a new one. This is equivalent
```
malleshpai, line 207: If you do it this way, and I rotate a validator mid-epoch, i would be ineligible to participate in the next DKG right? is that intentional?  the fact that you're saying it is equivalent to delete and then add suggests that it is intentional, but given our current thin margins is this safe (we might have too few players available for the next dkg)

malleshpai: nevermind now i understand. it might be worth explaining to validators that rotation involves a one epoch worth of deactivation (if it wasn't explained somewhere that i missed) but i think no change needed here.

tips/tip-1017.mdx
```diff
@@ -0,0 +1,786 @@
...
+
+13. **Atomic rotation**: `rotateValidator` atomically deletes the old validator
+    (setting its `deactivatedAtHeight`), adds a new validator entry, and updates
+    the `addresses[validatorAddress]` mapping to point to the new entry. Both
+    operations occur in the same transaction with the same block height.
+
+14. **Address update**: `transferValidatorOwnership` updates the address of an existing
+    validator without changing any other fields. This is useful for updating
+    addresses post-migration or correcting address assignments.
+
```
malleshpai: this clashes with invariant 2?

tips/tip-1017.mdx
```diff
@@ -0,0 +1,786 @@
...
+32. **migrateValidator imports validator**: Calling `migrateValidator(i)`
+    correctly imports validator at index i from V1 with the same address
+33. **migrateValidator copies address from V1**: The V2 validator uses the address from V1
+34. **migrateValidator reverts on duplicate**: Calling `migrateValidator(i)` reverts
+    if `i != validatorsArray.length`
+35. **migrateValidator reverts if initialized**: Calling `migrateValidator` reverts
+    if `isInitialized() == true`
+36. **migrateValidator owner only**: Non-owner cannot call `migrateValidator`
+37. **All validators imported on migration**: Both V1 active and inactive validators
+    are imported; inactive ones have `addedAtHeight == deactivatedAtHeight == block.timestamp`,
```
malleshpai: minor nit here and a few other places: are you sure you want a xxxHeight to be set to block.timestamp? why not use block height or rename the variable to be consistent? sorry if we litigated this already.

tips/tip-1017.mdx
malleshpai: maybe worth a test case?

tips/tip-1017.mdx
```diff
@@ -0,0 +1,786 @@
...
+
+6. **Non-zero public keys**: All validators must have a non-zero `publicKey`.
+
+7. **Monotonic index**: Validator `index` equals its position in
+   `validatorsArray` and equals `validatorCount - 1` at creation time.
+
+8. **Historical consistency**: For any height H, the active validator set
+   consists of validators where `addedAtHeight <= H && (deactivatedAtHeight == 0 ||
+   deactivatedAtHeight > H)`. Validators with `addedAtHeight == deactivatedAtHeight` are
+   never considered active.
```
malleshpai, line 481: should add !=0 here right? otherwise you rule out migrated vals...

tips/tip-1017.mdx
```diff
@@ -0,0 +1,786 @@
...
+    b"_VALIDATOR_CONFIG_V2_ROTATE_VALIDATOR"  // bytes: contract specific namespace
+    chainId,                                   // uint64: Prevents cross-chain replay
+    contractAddress,                           // address: Prevents cross-contract replay
+    validatorAddress,                          // address: Binds to specific new validator address
+    ingress,                            // string: Binds network configuration
+    egress                            // string: Binds network configuration
+))
+```
+
+The old validator is identified by `msg.sender`, which must be an existing active validator.
```
malleshpai: i think this contradicts invariant 11?

tips/tip-1017.mdx
```diff
@@ -0,0 +1,786 @@
...
+| Deletion | Sets `active = false` | Sets `deactivatedAtHeight = block.number` |
+| Re-registration | Allowed after deletion | Never allowed (address/pubkey reserved forever) |
+| Key ownership | Not verified | Ed25519 signature required |
+| Historical queries | Requires historical state | Filter `getValidators()` by `addedAtHeight`/`deactivatedAtHeight` |
+| Uniqueness | Address only | Address AND public key |
+| Precompile address | `0xCCCC...0000` | `0xCCCC...0001` |
+
+---
+
+# Invariants
```
malleshpai, line 446: Amp suggested this: "For validators added via addValidator, addedAtHeight > 0. addedAtHeight == 0 is reserved for validators migrated from V1

tips/tip-1017.mdx
SuperFluffy: It is briefly explained in the doc, but I agree - we should make this part of the docs so people understand.

But you pointing this out raises a point about how this spec wants to make deactivation immediate while leaving 1 epoch to warmup the new validator:

1. in epoch `E`: call `rotateValidator($old, $new)`
2. in epoch `E+1`: `$old` leaves the set of players (it is still a validator + dealer in this epoch); `$new` is not a player, merely registered as a peer and receives votes and certificates allowing it to sync up
3. in epoch `E+2`: `$old` dropped out of the committee; `$new` is a player and will receives shares from the other dealers to acknowledge
4. in epoch `E+3`: `$new` is in the committee.

This means there is a forced downtime of 1 epoch.

tips/tip-1017.mdx
SuperFluffy: Good catch! I updated invariant 2 and moved 14 into position 3 to clarify this point: `validatorAddress` is only mutable by the contract owner to address post-migration fixes. It cannot be changed by the owner of the entry.

c01f8419

tips/tip-1017.mdx
SuperFluffy: Yupp, good catch again. This was an oversight on my end: f3818090

tips/tip-1017.mdx
SuperFluffy: Thank you once more. This, too, was incorrect. `addedAtHeight = block.height` always: f80ecc4b

tips/tip-1017.mdx
SuperFluffy: Removed that comment - unnecessary and confusing: d57544ca

tips/tip-1017.mdx
SuperFluffy: I changed this in [f80ecc4](https://github.com/tempoxyz/tempo/pull/2181/commits/f80ecc4b154a342ea0802af28730be0d14db7b17): `addedAtHeight` is now always set to `block.height.` While this might be confusing for historical backlook, this is fine from CL's point of view because we only care about looking forward.

tips/tip-1017.mdx
```diff
@@ -0,0 +1,790 @@
...
+enclosed in brackets: `[2001:db8::1]:8080`.
+
+### Consensus Layer Behavior
+
+**IP Address Changes**: When a validator's IP address changes via `setIpAddresses`, the consensus layer is expected to update its peer list on the next finalized block.
+
+**Validator Addition and Deactivation**: When validators are added or deleted (this also applies to rotation),
+there is no warmup period: deactivated validators are immediately removed from the set of players on the next epoch,
+while activated validators are immediately added on the next epoch. This means that compared to validator config V1,
+there is no cooldown and no warmup period.
```
SuperFluffy, line 390: @Zygimantass for confirmation: there is no cooldown and warmup period anymore.

PR #2183, opened 2026-01-20
fix(txpool): enforce field-level limits to prevent DoS

fgimenez:
Closes CHAIN-489

Prevents memory exhaustion from oversized transactions via txpool-level field limits with peer penalization (`is_bad_transaction = true`):

- `MAX_AA_CALLS` (32) - already existed
- `MAX_CALL_INPUT_SIZE` (128KB per call) - already existed
- `MAX_ACCESS_LIST_ACCOUNTS` (256)
- `MAX_STORAGE_KEYS_PER_ACCOUNT` (256)
- `MAX_ACCESS_LIST_STORAGE_KEYS_TOTAL` (2048)
- `MAX_TOKEN_LIMITS` (256)

These limits are enforced at the txpool validator level rather than RLP decoding to:
- Keep the core transaction format flexible
- Allow peer penalization for sending bad transactions

### DIFF
--- a/crates/primitives/src/transaction/tempo_transaction.rs
+++ b/crates/primitives/src/transaction/tempo_transaction.rs
@@ -506,8 +506,8 @@ impl TempoTransaction {
         let max_priority_fee_per_gas = Decodable::decode(buf)?;
         let max_fee_per_gas = Decodable::decode(buf)?;
         let gas_limit = Decodable::decode(buf)?;
-        let calls: Vec<Call> = Decodable::decode(buf)?;
-        let access_list: AccessList = Decodable::decode(buf)?;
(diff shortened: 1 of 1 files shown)


0xKitsune (commented):
Review submitted

crates/primitives/src/transaction/tempo_transaction.rs
mattsse, line 1: can we undo changes to this, because these are left over from previous iteration on this fix

fgimenez: sure, done here https://github.com/tempoxyz/tempo/pull/2183/commits/07006f4ddc29c4441f6bef790063d2fcb57fb357

PR #2188, opened 2026-01-20
fix(validator): use index instead of address in changeValidatorStatus to prevent front-running

fgimenez:
Closes ZELLIC-24

Fixes a front-running vulnerability in `changeValidatorStatus` by using validator index instead of address. A malicious validator could front-run the owner's `changeValidatorStatus(validatorAddress, false)` call by first calling `rotateValidator` to change their address, causing the owner's deactivation to fail.

Changed `changeValidatorStatus` to accept a `uint64 index` parameter instead of `address validator`. Since validator indices are stable positions in the array, the owner can now reliably deactivate validators regardless of address rotations.

### DIFF
--- a/tips/ref-impls/foundry.lock
+++ b/tips/ref-impls/foundry.lock
@@ -1,7 +1,4 @@
 {
-  "lib/forge-std": {
-    "rev": "f61e4dd133379a4536a54ee57a808c9c00019b60"
-  },
   "lib/tempo-std": {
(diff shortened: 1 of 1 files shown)


0xrusowsky (commented):
on the rust impl, shouldn't we add a new function with the new type, and update the dispatcher to gatekeep this under T1?

this would be a breaking change if we have already called this fn

PR #2189, opened 2026-01-20
fix(tip20): remove unused storage fields to align with spec

fgimenez:
Closes ZELLIC-23

Removes three unused storage fields from the TIP20 precompile (`domain_separator`, `nonces`, `salts`) to align with the canonical Solidity specification.

These fields were defined in the Rust precompile but never used, causing a storage layout mismatch with `docs/specs/src/TIP20.sol`. Removing them ensures the precompile storage layout matches the spec exactly.

### DIFF
--- a/crates/precompiles/tests/storage_tests/solidity/testdata/tip20.layout.json
+++ b/crates/precompiles/tests/storage_tests/solidity/testdata/tip20.layout.json
@@ -269,5 +269,5 @@
       }
     }
   },
-  "version": "0.8.30+commit.73712a01.Linux.g++"
+  "version": "0.8.30+commit.73712a01.Darwin.appleclang"
 }


crates/precompiles/src/tip20/mod.rs
0xrusowsky, line 1: we can't simply remove here... we will have to leave empty slots

0xrusowsky (changes requested):
we can't do this. doing would break the storage layout.

we need to preserve the old storage slots in the same position as they currently are

crates/precompiles/src/tip20/mod.rs
fgimenez: makes sense, renamed fields to prepend `_` and added comments making clear they are unused

PR #2152, opened 2026-01-17
docs: add TIP-1010 Mainnet Gas Parameters

dankrad:
Defines initial gas parameters for Tempo mainnet launch including base fee, payment lane capacity, and transaction gas limits.

PR #2194, opened 2026-01-20
feat: implement TIP-1010 mainnet gas parameters

jenpaff:
## Summary

Implements [TIP-1010 Mainnet Gas Parameters](https://github.com/tempoxyz/tempo/pull/2152) specification on top of TIP-1000.

### Gas Parameters (TIP-1010 Spec)

| Parameter | Value | Purpose |
|-----------|-------|---------|
| Base fee | 20 gwei (2×10^10 wei) | Target ~0.1 cent per TIP-20 transfer |
| Total block gas limit | 500M gas | Shared capacity for all transactions |
| General gas limit | 30M gas/block | Cap for non-payment transactions |
| Transaction gas cap | 30M gas | Support max-size contract deployments |

### Code Changes

- **chainspec/spec.rs**: Update `TEMPO_BASE_FEE` from 10 gwei to 20 gwei
- **consensus/lib.rs**: 
  - Replace `TEMPO_GENERAL_GAS_DIVISOR` with fixed `TEMPO_GENERAL_GAS_LIMIT` (30M)
  - Add `TEMPO_TRANSACTION_GAS_CAP` constant (30M)
  - Update header validation to check fixed general_gas_limit
- **evm/context.rs**: Use fixed `TEMPO_GENERAL_GAS_LIMIT` instead of divisor calculation
- **evm/evm.rs**: Use `TEMPO_TRANSACTION_GAS_CAP` constant for tx gas limit cap
- **payload/builder**: Use shared 500M pool for total block capacity check

### Shared Gas Pool Model

The payment lane is non-dedicated: general transactions consume from the same 500M total block gas limit, reducing the remaining capacity available for payment transactions. However, general transactions are capped at 30M (`TEMPO_GENERAL_GAS_LIMIT`), guaranteeing that at least 470M gas remains available for payments even under maximum general transaction load.

## Testing

- All unit tests pass
- Cargo clippy passes

## Related

- Spec PR: #2152
- Depends on: #2166 (TIP-1000)

### DIFF
--- a/crates/node/tests/it/base_fee.rs
+++ b/crates/node/tests/it/base_fee.rs
@@ -43,7 +43,7 @@ async fn test_base_fee() -> eyre::Result<()> {
     for _ in 0..500 {
         let pending_tx = token
             .transfer(Address::random(), U256::ONE)
-            .gas_price(TEMPO_BASE_FEE as u128)
+            .gas_price(TEMPO_T1_BASE_FEE as u128)
             .gas(1_000_000)
             .send()
             .await?;
--- a/crates/node/tests/it/block_building.rs
+++ b/crates/node/tests/it/block_building.rs
@@ -37,7 +37,7 @@ where
             tx_req.nonce = Some(nonce);
             tx_req.chain_id = Some(chain_id);
             tx_req.gas = tx_req.gas.or(Some(5_000_000));
-            tx_req.max_fee_per_gas = tx_req.max_fee_per_gas.or(Some(TEMPO_BASE_FEE as u128));
+            tx_req.max_fee_per_gas = tx_req.max_fee_per_gas.or(Some(TEMPO_T1_BASE_FEE as u128));
             tx_req.max_priority_fee_per_gas = tx_req
                 .max_priority_fee_per_gas
                 .or(Some(TEMPO_T1_BASE_FEE as u128));
@@ -155,8 +155,8 @@ where
         tx_request.nonce = Some(current_nonce + i as u64);
         tx_request.chain_id = Some(chain_id);
         tx_request.gas = Some(1_000_000);
-        tx_request.max_fee_per_gas = Some(TEMPO_BASE_FEE as u128);
-        tx_request.max_priority_fee_per_gas = Some(TEMPO_BASE_FEE as u128);
+        tx_request.max_fee_per_gas = Some(TEMPO_T1_BASE_FEE as u128);
+        tx_request.max_priority_fee_per_gas = Some(TEMPO_T1_BASE_FEE as u128);
 
         let signed_tx =
             <TransactionRequest as TransactionBuilder<Ethereum>>::build(tx_request, &signer)
--- a/crates/node/tests/it/createx.rs
+++ b/crates/node/tests/it/createx.rs
@@ -27,7 +27,7 @@ async fn test_createx() -> eyre::Result<()> {
     // Get deployed address from simulated call
     let deployed_address = createx
         .deployCreate(init_code.clone())
-        .gas_price(TEMPO_BASE_FEE as u128)
+        .gas_price(TEMPO_T1_BASE_FEE as u128)
         .gas(5_000_000)
         .call()
         .await?
@@ -36,7 +36,7 @@ async fn test_createx() -> eyre::Result<()> {
     // Send the actual transaction
     createx
         .deployCreate(init_code)
-        .gas_price(TEMPO_BASE_FEE as u128)
+        .gas_price(TEMPO_T1_BASE_FEE as u128)
         .gas(5_000_000)
         .send()
         .await?
--- a/crates/node/tests/it/eth_call.rs
+++ b/crates/node/tests/it/eth_call.rs
@@ -38,7 +38,7 @@ async fn test_eth_call() -> eyre::Result<()> {
     let mint_amount = U256::from(rand::random::<u128>());
     token
         .mint(caller, mint_amount)
-        .gas_price(TEMPO_BASE_FEE as u128)
+        .gas_price(TEMPO_T1_BASE_FEE as u128)
         .gas(1_000_000)
         .send()
         .await?
@@ -78,7 +78,7 @@ async fn test_eth_trace_call() -> eyre::Result<()> {
     let mint_amount = U256::from(rand::random::<u128>());
     token
         .mint(caller, mint_amount)
-        .gas_price(TEMPO_BASE_FEE as u128)
+        .gas_price(TEMPO_T1_BASE_FEE as u128)
         .gas(1_000_000)
         .send()
         .await?
@@ -170,7 +170,7 @@ async fn test_eth_get_logs() -> eyre::Result<()> {
     let mint_amount = U256::from(rand::random::<u128>());
     let mint_receipt = token
         .mint(caller, mint_amount)
-        .gas_price(TEMPO_BASE_FEE as u128)
+        .gas_price(TEMPO_T1_BASE_FEE as u128)
         .gas(1_000_000)
         .send()
         .await?
@@ -180,7 +180,7 @@ async fn test_eth_get_logs() -> eyre::Result<()> {
     let recipient = Address::random();
     token
         .transfer(recipient, mint_amount)
-        .gas_price(TEMPO_BASE_FEE as u128)
+        .gas_price(TEMPO_T1_BASE_FEE as u128)
         .gas(1_000_000)
         .send()
         .await?
--- a/crates/node/tests/it/eth_transactions.rs
+++ b/crates/node/tests/it/eth_transactions.rs
@@ -26,7 +26,7 @@ async fn test_get_transaction_by_sender_and_nonce() -> eyre::Result<()> {
     let mint_amount = U256::from(1000000u64);
     let pending_tx = token
         .mint(caller, mint_amount)
-        .gas_price(TEMPO_BASE_FEE as u128)
+        .gas_price(TEMPO_T1_BASE_FEE as u128)
         .gas(1_000_000)
         .send()
         .await?;
--- a/crates/node/tests/it/payment_lane.rs
+++ b/crates/node/tests/it/payment_lane.rs
@@ -108,7 +108,7 @@ async fn test_payment_lane_with_mixed_load() -> eyre::Result<()> {
                 let tx = TransactionRequest::default()
                     .from(accounts[i])
                     .to(accounts[i]) // Send to self
-                    .gas_price(TEMPO_BASE_FEE as u128)
+                    .gas_price(TEMPO_T1_BASE_FEE as u128)
                     .gas_limit(2_000_000)
                     .value(U256::ZERO);
 
@@ -222,7 +222,7 @@ async fn test_payment_lane_with_mixed_load() -> eyre::Result<()> {
                 let tx = TransactionRequest::default()
                     .from(accounts[j])
                     .to(accounts[j]) // Send to self
-                    .gas_price(TEMPO_BASE_FEE as u128)
+                    .gas_price(TEMPO_T1_BASE_FEE as u128)
                     .gas_limit(2_000_000)
                     .value(U256::ZERO);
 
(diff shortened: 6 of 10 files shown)


crates/consensus/src/lib.rs
```diff
@@ -61,13 +61,12 @@ impl HeaderValidator<TempoHeader> for TempoConsensus {
...
-        {
-            return Err(ConsensusError::Other(
-                "Non-payment gas limit does not match header gas limit".to_string(),
-            ));
+        // Validate the general (non-payment) gas limit is fixed at TEMPO_GENERAL_GAS_LIMIT
+        if header.general_gas_limit != TEMPO_GENERAL_GAS_LIMIT {
+            return Err(ConsensusError::Other(format!(
+                "General gas limit {} does not match expected {}",
+                header.general_gas_limit, TEMPO_GENERAL_GAS_LIMIT
+            )));
```
0xKitsune: This needs to be hardfork gated behind `T1`.

crates/chainspec/src/spec.rs
```diff
@@ -15,7 +15,9 @@ use reth_network_peers::NodeRecord;
 use std::sync::{Arc, LazyLock};
 use tempo_primitives::TempoHeader;
 
-pub const TEMPO_BASE_FEE: u64 = 10_000_000_000;
+/// Target base fee: 20 gwei (2×10^10 wei)
+/// At this base fee, a standard TIP-20 transfer (~50,000 gas) costs ~0.1 cent
+pub const TEMPO_BASE_FEE: u64 = 20_000_000_000;
```
0xKitsune: We will need a `TEMPO_BASE_FEE_PRE_T1` and `TEMPO_BASEFEE_POST_T1` or similar to properly gate gas fee changes behind a hardfork.

crates/consensus/src/lib.rs
```diff
@@ -375,21 +379,21 @@ mod tests {
...
         let sealed = SealedHeader::seal_slow(header);
 
         let result = consensus.validate_header(&sealed);
         let err = result.unwrap_err();
         assert!(matches!(err, ConsensusError::Other(_)));
         assert!(
-            err.to_string()
-                .contains("Non-payment gas limit does not match header gas limit")
+            err.to_string().contains("General gas limit"),
+            "Expected error about general gas limit, got: {err}",
```
0xKitsune: We should have tests for pre/post hardfork behavior.

crates/consensus/src/lib.rs
```diff
@@ -257,10 +264,7 @@ mod tests {
             let shared_gas_limit = self
                 .shared_gas_limit
                 .unwrap_or(self.gas_limit / TEMPO_SHARED_GAS_DIVISOR);
-            let general_gas_limit = self.general_gas_limit.unwrap_or_else(|| {
-                (self.gas_limit - self.gas_limit / TEMPO_SHARED_GAS_DIVISOR)
-                    / TEMPO_GENERAL_GAS_DIVISOR
-            });
+            let general_gas_limit = self.general_gas_limit.unwrap_or(TEMPO_GENERAL_GAS_LIMIT);
```
0xKitsune: Same thing here, we need to gate this behind `spec.is_t1()`.

crates/evm/src/context.rs
```diff
@@ -52,8 +52,7 @@ impl reth_rpc_eth_api::helpers::pending_block::BuildPendingEnv<tempo_primitives:
         use alloy_consensus::BlockHeader as _;
 
         let shared_gas_limit = parent.gas_limit() / tempo_consensus::TEMPO_SHARED_GAS_DIVISOR;
-        let general_gas_limit =
-            (parent.gas_limit() - shared_gas_limit) / tempo_consensus::TEMPO_GENERAL_GAS_DIVISOR;
+        let general_gas_limit = tempo_consensus::TEMPO_GENERAL_GAS_LIMIT;
```
0xKitsune: Same here, we need to hardfork gate this.

crates/evm/src/context.rs
```diff
@@ -52,8 +52,19 @@ impl reth_rpc_eth_api::helpers::pending_block::BuildPendingEnv<tempo_primitives:
         use alloy_consensus::BlockHeader as _;
 
         let shared_gas_limit = parent.gas_limit() / tempo_consensus::TEMPO_SHARED_GAS_DIVISOR;
+
+        // Determine general gas limit based on parent header.
+        // If parent uses T1+ fixed general gas limit, use that; otherwise calculate using pre-T1 divisor.
+        // Note: This is a heuristic since we don't have chainspec access here.
+        // If the parent's general_gas_limit matches T1's fixed value, assume T1 is active.
```
gakonst: **Bug: Heuristic-based T1 detection is incorrect at fork boundary**

This heuristic will fail on the first T1 block because:
- Parent is pre-T1 (divisor-based general_gas_limit ≠ 30M)
- Child should be T1 (fixed 30M)
- But heuristic sees parent's value and uses pre-T1 calculation

The comment acknowledges this limitation. **Fix:** Accept chainspec as parameter and use `chainspec.is_t1_active_at_timestamp(next_block_timestamp)` for correct fork detection.

0xKitsune: The comment has been updated to describe this edge case. In any case this should be fine given pending blocks are disabled and not served via the `TempoEthApi`.

crates/chainspec/src/spec.rs
```diff
@@ -15,7 +15,12 @@ use reth_network_peers::NodeRecord;
 use std::sync::{Arc, LazyLock};
 use tempo_primitives::TempoHeader;
 
-pub const TEMPO_BASE_FEE: u64 = 10_000_000_000;
+/// Pre-T1 base fee: 10 gwei (1×10^10 wei)
+pub const TEMPO_BASE_FEE_PRE_T1: u64 = 10_000_000_000;
+
+/// Post-T1 base fee: 20 gwei (2×10^10 wei)
+/// At this base fee, a standard TIP-20 transfer (~50,000 gas) costs ~0.1 cent
+pub const TEMPO_BASE_FEE_POST_T1: u64 = 20_000_000_000;
```
0xrusowsky: would be nice to have them as a const fn on `TempoHardfork`, this way we decluter

crates/transaction-pool/src/tt_2d_pool.rs
0xrusowsky, line 1: should this be hardfork aware?

0xKitsune: Addressed in https://github.com/tempoxyz/tempo/pull/2194/commits/11681f50b29c9cc5caa80d2ee5aa4becd7f7a8dd PTAL

crates/chainspec/src/spec.rs
```diff
@@ -15,7 +15,12 @@ use reth_network_peers::NodeRecord;
 use std::sync::{Arc, LazyLock};
 use tempo_primitives::TempoHeader;
 
-pub const TEMPO_BASE_FEE: u64 = 10_000_000_000;
+/// T0 base fee: 10 gwei (1×10^10 wei)
+pub const TEMPO_T0_BASE_FEE: u64 = TempoHardfork::T0.base_fee();
```
mattsse: pendantic style nit, we could use the values for consts that we then use in functions, this usually avoids one hop when looking for a const value, but not a blocker for this pr

crates/consensus/src/lib.rs
mattsse: I'd prefer if we move them to top of file

crates/node/src/node.rs
```diff
@@ -422,7 +422,14 @@ where
...
         let mut pool_config = ctx.pool_config();
-        pool_config.minimal_protocol_basefee = TEMPO_BASE_FEE;
+        // Set pool minimum base fee based on T1 status at startup. If node starts post-T1,
+        // use the higher fee. If pre-T1, use lower fee - after T1 activates, txs with fees
+        // between 10-20 gwei will enter pool but fail at block building (acceptable since
+        // pool minimum is just an optimization filter).
+        pool_config.minimal_protocol_basefee = ctx
+            .chain_spec()
+            .tempo_hardfork_at(ctx.head().timestamp)
+            .base_fee();
```
mattsse: this is approproriate, only an issue until first restart after t1 activiation

mattsse (approved):
some style suggestions

generated/genesis.json
```diff
@@ -1,1897 +0,0 @@
-{
```
rakita, line 1: Should this be deleted?

crates/consensus/src/lib.rs
```diff
@@ -178,10 +186,22 @@ impl FullConsensus<TempoPrimitives> for TempoConsensus {
...
-/// Divisor for calculating shared gas limit.
+/// T1+ general (non-payment) gas limit per block: 30M gas
+/// This caps how much gas general transactions can consume, guaranteeing
+/// at least 470M gas remains available for payment transactions.
+pub const TEMPO_GENERAL_GAS_LIMIT: u64 = 30_000_000;
+
+/// T1+ maximum gas per single transaction: 30M gas
+/// Elevated from 16M to support max-size contract deployments under TIP-1000.
+/// Applications should not rely on >16M gas for normal operations.
+pub const TEMPO_TRANSACTION_GAS_CAP: u64 = 30_000_000;
```
klkvr: given that all of those consts are fork-dependent now i'd rather not have them or rename to have hardfork in the name

crates/evm/src/context.rs
```diff
@@ -52,8 +52,17 @@ impl reth_rpc_eth_api::helpers::pending_block::BuildPendingEnv<tempo_primitives:
...
+        // This has a one-block edge case at T1 boundary (parent is pre-T1 but child should be T1).
+        // However, pending block building is disabled for Tempo (PendingBlockKind::None) since
+        // blocks require consensus data (system transactions) that RPC doesn't have.
         let general_gas_limit =
-            (parent.gas_limit() - shared_gas_limit) / tempo_consensus::TEMPO_GENERAL_GAS_DIVISOR;
+            if parent.general_gas_limit == tempo_consensus::TEMPO_GENERAL_GAS_LIMIT {
+                tempo_consensus::TEMPO_GENERAL_GAS_LIMIT
+            } else {
+                (parent.gas_limit() - shared_gas_limit) / tempo_consensus::TEMPO_GENERAL_GAS_DIVISOR
+            };
```
klkvr: i don't think this is even used and it's purely rpc-facing so we can just use parent's values here

xtask/src/genesis_args.rs
```diff
@@ -82,9 +82,11 @@ pub(crate) struct GenesisArgs {
...
     chain_id: u64,
 
-    /// Base fee
-    #[arg(long, default_value_t = TEMPO_BASE_FEE.into())]
-    base_fee_per_gas: u128,
+    /// Base fee. If not specified, defaults based on T1 activation:
+    /// - With --t1-time: uses post-T1 fee (20 gwei)
+    /// - Without --t1-time: uses pre-T1 fee (10 gwei)
+    #[arg(long)]
+    base_fee_per_gas: Option<u128>,
```
klkvr: should we remove it? given that it's only relevant for genesis and fully determined by the hardfork

generated/genesis.json
0xKitsune: I believe this may have been generated accidentally and committed in a separate PR.

crates/node/src/node.rs
0xKitsune: >  Oh we don't actually need this, this setting is only useful if you want to allow basfee 0
basefee is enforced in the pool based on recent block so we don't need to set this

Sounds good, removed in https://github.com/tempoxyz/tempo/pull/2194/commits/03f1d89d7f0d6e860875920a03b7f48651d42c37

crates/evm/src/context.rs
0xKitsune: Sounds good updated to remove in https://github.com/tempoxyz/tempo/pull/2194/commits/ab878e2fb08501905c7f7068a98f270d77b43fda

xtask/src/genesis_args.rs
0xKitsune: Removed in https://github.com/tempoxyz/tempo/pull/2194/commits/ab878e2fb08501905c7f7068a98f270d77b43fda

crates/consensus/src/lib.rs
0xKitsune: Removed in https://github.com/tempoxyz/tempo/pull/2194/commits/29b77748987ff594f84c965cb7991686912e33cc

PR #2196, opened 2026-01-20
feat(cli): add safety check to prevent overwriting existing private keys

luckyson113:
## Description
This PR improves the safety of the `generate-private-key` command.

Currently, if a user accidentally runs the generation command pointing to an existing key file, the CLI silently overwrites it. 

This creates a risk of **permanent loss of the validator identity and associated funds**.

This change adds a safeguard to ensure the output file does not exist before writing. It also introduces a `--force` flag for users who intentionally want to overwrite the file.

## Changes
- Modified `GeneratePrivateKey` struct to include a `force` flag.
- Added a safety check: `if output.exists() && !force` returns an error.
- Updated user feedback to explicitly suggest `--force` if a conflict is detected.

## Usage
**Default Behavior (Safe):**
```bash
$ tempo consensus generate-private-key --output ./secret.key
Error: File `./secret.key` already exists. Use --force to overwrite it.
```
**Force Overwrite:**
```bash
$ tempo consensus generate-private-key --output ./secret.key --force
wrote private key to: ./secret.key
...
```

## Checklist
- [x] I have performed a self-review of my code.
- [x] The code compiles successfully.
- [x] I have verified the protection logic locally.

### DIFF
--- a/bin/tempo/src/tempo_cmd.rs
+++ b/bin/tempo/src/tempo_cmd.rs
@@ -7,6 +7,7 @@ use eyre::Context;
 use reth_cli_runner::CliRunner;
 use reth_ethereum_cli::ExtendedCommand;
 use tempo_commonware_node_config::SigningKey;
+use std::fs::OpenOptions;
 
 /// Tempo-specific subcommands that extend the reth CLI.
 #[derive(Debug, Subcommand)]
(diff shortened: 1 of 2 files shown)


bin/tempo/src/tempo_cmd.rs
```diff
@@ -46,11 +46,25 @@ pub(crate) struct GeneratePrivateKey {
...
     output: PathBuf,
+    /// Overwrite the file if it already exists.
+    #[arg(long, short)]
+    force: bool,
 }
 
 impl GeneratePrivateKey {
     fn run(self) -> eyre::Result<()> {
-        let Self { output } = self;
+        // 👇 FIX: Extract both 'output' and 'force' from the struct
```
Zygimantass: please remove the comments here and it'll be mergeable

luckyson113: Done. I removed the internal comments. Thanks for the review

bin/tempo/src/tempo_cmd.rs
```diff
@@ -46,11 +46,24 @@ pub(crate) struct GeneratePrivateKey {
...
+    /// Overwrite the file if it already exists.
+    #[arg(long, short)]
+    force: bool,
 }
 
 impl GeneratePrivateKey {
     fn run(self) -> eyre::Result<()> {
-        let Self { output } = self;
+        let Self { output, force } = self;
+        // Safety check: Prevent accidental overwrite of existing key files.
```
SuperFluffy: The entire comment is unnecessary.

bin/tempo/src/tempo_cmd.rs
```diff
@@ -46,11 +46,24 @@ pub(crate) struct GeneratePrivateKey {
...
+        // If the file exists and the user did not explicitly provide the --force flag,
+        // we abort the operation to protect the user's funds/identity.
+        if output.exists() && !force {
```
SuperFluffy: This can run into TOCTOU issues. Please extend the `SigningKey::write_to_file` method by a convenience method `SigningKey::to_writer` which is generic over `W: std::io::Write`. Then this method here can just use `std::fs::OpenOptions` to configure a file with/without `truncate: true`

SuperFluffy (changes requested):

bin/tempo/src/tempo_cmd.rs
luckyson113: Done. I've implemented SigningKey::to_writer and updated the CLI to use OpenOptions with create_new(true) (or truncate(true) with force). This should resolve the TOCTOU issue

PR #2200, opened 2026-01-21
chore(consensus): use descriptive channel names and correct epoch manager docs

SuperFluffy:
Renames all instances of "pending" to "votes" and "recovered" to "certificates".


crates/commonware-node/src/config.rs
```diff
@@ -15,8 +15,8 @@ use governor::Quota;
 
 // Hardcoded values to configure commonware's alto toy chain. These could be made into
 // configuration variables at some point.
-pub const PENDING_CHANNEL_IDENT: commonware_p2p::Channel = 0;
-pub const RECOVERED_CHANNEL_IDENT: commonware_p2p::Channel = 1;
```
joshieDo, line 19: way clearer

PR #2208, opened 2026-01-21
[do not merge] testing tip 1000

legion2002:


crates/evm/src/evm.rs
```diff
@@ -61,7 +67,15 @@ pub struct TempoEvm<DB: Database, I = NoOpInspector> {
...
     /// Create a new [`TempoEvm`] instance.
-    pub fn new(db: DB, input: EvmEnv<TempoHardfork, TempoBlockEnv>) -> Self {
+    pub fn new(db: DB, mut input: EvmEnv<TempoHardfork, TempoBlockEnv>) -> Self {
+        // override the gas params for tempo use case.
+        input.cfg_env.gas_params = tempo_gas_params(input.cfg_env.spec);
+
+        // In TIP-1000 we bumped gas to 30M.
+        if input.cfg_env.spec.is_t1() {
+            input.cfg_env.tx_gas_limit_cap = Some(TIP1000_GAS_LIMIT_CAP);
+        }
```
mattsse, line 77: I dont think this should happen here, and should instead be applied on the input params

because this now makes it problematic to bypass this on rpc level which does this:

https://github.com/paradigmxyz/reth/blob/ec50fd40b3d7568eb17d0d18844e625e267ff124/crates/rpc/rpc-eth-api/src/helpers/call.rs#L801-L802

before creating the evm, so that override wont get applied

https://github.com/tempoxyz/tempo/blob/471c1ce2f654391a20815a6ef5b468260c35aaf5/crates/evm/src/lib.rs#L121-L129

mattsse (changes requested):

PR #2216, opened 2026-01-22
fix(txpool): evict AA2dPool transactions by priority instead of address to prevent DoS

fgimenez:
Closes CHAIN-449

Fixes a DoS vulnerability in `AA2dPool::discard()` where transactions were evicted lexicographically by address instead of by priority. Attackers with vanity addresses (many leading zeroes like `0x0000...`) could persist in the pool indefinitely while legitimate transactions from normal addresses were evicted first. This allowed low-cost DoS attacks on the transaction pool.

Changed `discard()` to sort transactions by `CoinbaseTipOrdering` priority and evict lowest-priority transactions first, regardless of sender address.

### DIFF
--- a/crates/transaction-pool/src/tt_2d_pool.rs
+++ b/crates/transaction-pool/src/tt_2d_pool.rs
@@ -228,13 +228,12 @@ impl AA2dPool {
                         // unaffected by our transaction
                     }
                     std::cmp::Ordering::Equal => {
-                        *existing_tx.is_pending.borrow_mut() = true;
+                        existing_tx.set_pending(true);
                         inserted_as_pending = true;
                     }
                     std::cmp::Ordering::Greater => {
                         // if this was previously not pending we need to promote the transaction
-                        let was_pending =
-                            std::mem::replace(&mut *existing_tx.is_pending.borrow_mut(), true);
+                        let was_pending = existing_tx.set_pending(true);
                         if !was_pending {
                             promoted.push(existing_tx.inner.transaction.clone());
                         }
@@ -328,7 +327,7 @@ impl AA2dPool {
     /// Returns how many pending and queued transactions are in the pool.
     pub(crate) fn pending_and_queued_txn_count(&self) -> (usize, usize) {
         let (pending_2d, queued_2d) = self.by_id.values().fold((0, 0), |mut acc, tx| {
-            if *tx.is_pending.borrow() {
+            if tx.is_pending() {
                 acc.0 += 1;
             } else {
                 acc.1 += 1;
@@ -366,7 +365,7 @@ impl AA2dPool {
         let regular = self
             .by_id
             .values()
-            .filter(move |tx| *tx.is_pending.borrow() && tx.inner.transaction.origin == origin)
+            .filter(move |tx| tx.is_pending() && tx.inner.transaction.origin == origin)
             .map(|tx| tx.inner.transaction.clone());
         // Expiring nonce txs are always pending
         let expiring = self
@@ -406,7 +405,7 @@ impl AA2dPool {
     ) -> impl Iterator<Item = Arc<ValidPoolTransaction<TempoPooledTransaction>>> {
         self.by_id
             .values()
-            .filter(|tx| !*tx.is_pending.borrow())
+            .filter(|tx| !tx.is_pending())
             .map(|tx| tx.inner.transaction.clone())
     }
 
@@ -418,7 +417,7 @@ impl AA2dPool {
         let regular_pending = self
             .by_id
             .values()
-            .filter(|tx| *tx.is_pending.borrow())
+            .filter(|tx| tx.is_pending())
             .map(|tx| tx.inner.transaction.clone());
         let expiring_pending = self
             .expiring_nonce_txs
@@ -441,7 +440,7 @@ impl AA2dPool {
             by_id: self
                 .by_id
                 .iter()
-                .filter(|(_, tx)| *tx.is_pending.borrow())
+                .filter(|(_, tx)| tx.is_pending())
                 .map(|(id, tx)| (*id, tx.inner.clone()))
                 .collect(),
             invalid: Default::default(),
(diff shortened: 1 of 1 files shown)


crates/transaction-pool/src/tt_2d_pool.rs
```diff
@@ -762,16 +762,39 @@ impl AA2dPool {
...
-            let Some(id) = self.by_id.last_key_value().map(|(id, _)| *id) else {
-                return removed;
-            };
-            removed.push(self.remove_transaction_by_id(&id).unwrap());
+        // Collect all transactions with their IDs and sort by priority (lowest first)
+        let mut txs_by_priority: Vec<_> = self
+            .by_id
+            .iter()
+            .map(|(id, tx)| (*id, tx.inner.clone()))
+            .collect();
```
mattsse: this is too expensive because no we'd perform this on every new txs if the pool is at capacity

crates/transaction-pool/src/tt_2d_pool.rs
```diff
@@ -762,16 +762,39 @@ impl AA2dPool {
...
+
+        // Sort by priority ascending (lowest priority first for eviction)
+        // PendingTransaction::Ord orders highest priority first, so we reverse
+        txs_by_priority.sort_by(|a, b| a.1.cmp(&b.1));
```
mattsse: I think this has another flaw, because this doesnt take into account whether the txs is queued or pending so I believe you could technically evict pending txs by sending non executable once with a higher prio

crates/transaction-pool/src/tt_2d_pool.rs
```diff
@@ -26,6 +26,54 @@ use tempo_precompiles::NONCE_PRECOMPILE_ADDRESS;
...
+/// Key for ordering transactions by eviction priority.
+///
+/// Orders by:
+/// 1. Priority ascending (lowest priority evicted first)
+/// 2. Submission ID descending (newer transactions evicted first among same priority)
+///
+/// This is the inverse of the execution order (where highest priority, oldest submission wins).
+/// Newer transactions are evicted first to preserve older transactions that have been waiting longer.
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct EvictionKey {
```
mattsse: amp keeps doing this and I really dislike how it orders types, this does not belong top of file :)

crates/transaction-pool/src/tt_2d_pool.rs
```diff
@@ -26,6 +26,54 @@ use tempo_precompiles::NONCE_PRECOMPILE_ADDRESS;
...
+impl Ord for EvictionKey {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        // Lower priority first (evict lowest priority)
+        self.priority
+            .cmp(&other.priority)
+            // Then newer submission first (evict newer transactions among same priority)
+            // This preserves older transactions that have been waiting longer
+            .then_with(|| other.submission_id.cmp(&self.submission_id))
+            // Finally by tx_id for uniqueness
+            .then_with(|| self.tx_id.cmp(&other.tx_id))
```
mattsse: this is effectively a no-op because submission id is unique

crates/transaction-pool/src/tt_2d_pool.rs
```diff
@@ -1166,24 +1169,91 @@ struct AA2dInternalTransaction {
     /// Whether this transaction is pending/executable.
     ///
     /// If it's not pending, it is queued.
-    is_pending: bool,
+    ///
+    /// Uses `Arc<AtomicBool>` so we can mutate this flag without removing/reinserting
+    /// the transaction from the eviction set. This allows a single eviction set for
+    /// all transactions, with pending/queued filtering done at eviction time.
+    is_pending: Arc<AtomicBool>,
```
mattsse: this doesnt technically need to be an atomic and can be rc refcell because this only ever accessed by the pool via mut anyway

crates/transaction-pool/src/tt_2d_pool.rs
```diff
@@ -1166,24 +1169,91 @@ struct AA2dInternalTransaction {
...
+    /// The transaction's priority (from `CoinbaseTipOrdering`).
+    priority: Priority<u128>,
+    /// The submission ID when the transaction was added to the pool.
+    /// This is unique per transaction, so it's sufficient for ordering.
+    submission_id: u64,
+    /// The transaction's unique identifier (for lookup during eviction).
+    tx_id: AA2dTransactionId,
+    /// Shared pending state for filtering without lookup.
+    is_pending: Arc<AtomicBool>,
+}
```
mattsse, line 1263: we could technically just wrap arc AA2dInternalTransaction for this but this doesnt really matter here

mattsse (commented):
the atomic is technically not needed so we could optimize this away but relaxed also seems fine

crates/transaction-pool/src/tt_2d_pool.rs
```diff
@@ -379,7 +366,9 @@ impl AA2dPool {
         let regular = self
             .by_id
             .values()
-            .filter(move |tx| tx.is_pending && tx.inner.transaction.origin == origin)
+            .filter(move |tx| {
+                tx.is_pending.load(Ordering::Relaxed) && tx.inner.transaction.origin == origin
```
mattsse: last suggestion, would be nicer if this were a fn is_pending() then we can change this more easily later

crates/transaction-pool/src/tt_2d_pool.rs
```diff
@@ -206,35 +201,41 @@ impl AA2dPool {
...
-        for (existing_id, existing_tx) in self.descendant_txs_mut(&on_chain_id) {
+        for (existing_id, existing_tx) in self.descendant_txs(&on_chain_id) {
             if existing_id.nonce == next_nonce {
                 match existing_id.nonce.cmp(&tx_id.nonce) {
                     std::cmp::Ordering::Less => {
                         // unaffected by our transaction
                     }
                     std::cmp::Ordering::Equal => {
-                        existing_tx.is_pending = true;
+                        *existing_tx.is_pending.borrow_mut() = true;
```
mattsse: would be nicer if we had fn is_pending and set_pending, then all of the borrow stuff is hidden and we can change this more easily

crates/transaction-pool/src/tt_2d_pool.rs
fgimenez: good call, done

PR #2218, opened 2026-01-22
fix(tx-pool): enforce AMM liquidity check for mempool validation

fgimenez:
Closes CHAIN-442

Closes DoS vector where attackers could spam transactions with fee tokens lacking AMM liquidity. The liquidity cache bypass (intended for block building when user pays directly in the builder's token) was incorrectly applied during mempool validation.

- Added `builder_token: Option<Address>` parameter to `has_enough_liquidity`
- Mempool validation: passes `None`, enforcing full AMM liquidity checks from state
- Block building: passes `Some(builder_token)`, allowing bypass only when user's fee token matches
- Added integration test verifying transactions with insufficient AMM liquidity are rejected

### DIFF
--- a/crates/transaction-pool/src/amm.rs
+++ b/crates/transaction-pool/src/amm.rs
@@ -84,6 +84,15 @@ impl AmmLiquidityCache {
         {
             let inner = self.inner.read();
             for validator_token in &inner.unique_tokens {
+                // When user_token matches a validator's configured token, no AMM swap is needed
+                // at execution time. The TipFeeManager transfers fees directly without swapping.
+                // This is safe because the transaction will only succeed when processed by a
+                // validator whose token matches - other validators will fail to execute it,
+                // but it won't propagate spam since the user IS paying in a valid validator token.
+                if validator_token == &user_token {
+                    return Ok(true);
+                }
+
                 if let Some(validator_reserve) = inner.cache.get(&(user_token, *validator_token)) {
                     if *validator_reserve >= amount_out {
                         return Ok(true);
@@ -278,6 +287,29 @@ mod tests {
         assert!(
             !result.unwrap(),
             "Should return false when no AMM liquidity exists for user's fee token"
+        );
+    }
+
+    #[test]
+    fn test_has_enough_liquidity_user_token_matches_validator_token() {
+        let token = address!("1111111111111111111111111111111111111111");
+
+        let cache = AmmLiquidityCache {
+            inner: Arc::new(RwLock::new(AmmLiquidityCacheInner {
+                unique_tokens: vec![token],
+                ..Default::default()
+            })),
+        };
+
+        let provider = create_mock_provider();
+        let state = provider.latest().unwrap();
+
+        // When user_token == validator_token, no AMM swap is needed. Return true.
+        let result = cache.has_enough_liquidity(token, U256::from(100), &state);
+        assert!(result.is_ok());
+        assert!(
+            result.unwrap(),
+            "Should return true when user token matches a validator's preferred token"
         );
     }
 
--- a/crates/transaction-pool/src/validator.rs
+++ b/crates/transaction-pool/src/validator.rs
@@ -3201,16 +3201,17 @@ mod tests {
             "Token should be in unique_tokens for this test"
         );
 
-        // With the fix for CHAIN-442, user tokens matching a validator's preferred token
-        // no longer bypass the AMM liquidity check. The old bypass was removed entirely.
+        // When user_token matches a validator's preferred token, has_enough_liquidity returns true
+        // because no AMM swap is needed at execution time. The TipFeeManager handles this directly.
         let liquidity_result =
             amm_cache.has_enough_liquidity(paused_validator_token, U256::from(1000), &state);
         assert!(
-            liquidity_result.is_ok() && !liquidity_result.unwrap(),
-            "User token matching a validator token no longer bypasses liquidity check"
+            liquidity_result.is_ok() && liquidity_result.unwrap(),
+            "User token matching a validator token should bypass AMM liquidity check"
         );
 
-        // The pause check in is_fee_token_paused provides defense-in-depth
+        // However, the pause check in is_fee_token_paused provides defense-in-depth and MUST
+        // catch paused tokens BEFORE the liquidity check is even consulted.
         let is_paused = state.is_fee_token_paused(spec, paused_validator_token);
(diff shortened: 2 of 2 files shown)


crates/transaction-pool/src/validator.rs
```diff
@@ -555,7 +555,7 @@ where
 
         match self
             .amm_liquidity_cache
-            .has_enough_liquidity(fee_token, cost, &state_provider)
+            .has_enough_liquidity(fee_token, cost, &state_provider, None)
```
mattsse: we always call this with None during validation, so unsure if this change does address anything?

crates/transaction-pool/src/amm.rs
```diff
@@ -66,13 +86,6 @@ impl AmmLiquidityCache {
         {
             let inner = self.inner.read();
             for validator_token in &inner.unique_tokens {
-                // If user token matches one of the recently seen validator tokens,
-                // short circuit and return true. We assume that validators are willing to
-                // accept transactions that pay fees in their token directly.
-                if validator_token == &user_token {
-                    return Ok(true);
-                }
```
mattsse, line 74: why are we removing this?

crates/transaction-pool/src/amm.rs
```diff
@@ -51,13 +51,33 @@ impl AmmLiquidityCache {
...
         state_provider: &impl StateProvider,
+        builder_token: Option<Address>,
     ) -> Result<bool, ProviderError> {
+        // If building a block with a known builder token, bypass liquidity check
+        // when user pays directly in the builder's token (no swap needed).
+        if let Some(bt) = builder_token
+            && user_token == bt
+        {
+            return Ok(true);
+        }
```
mattsse: this only makes sense locally, but the builders are likely not where the tx originally enters the network

crates/transaction-pool/src/amm.rs
```diff
@@ -51,13 +51,33 @@ impl AmmLiquidityCache {
...
+
+        // DEFAULT_FEE_TOKEN (PATH_USD) is always accepted by all validators without
+        // requiring an AMM swap. Skip liquidity check for the base token.
+        if user_token == DEFAULT_FEE_TOKEN {
+            return Ok(true);
+        }
```
klkvr: is this right? don't remember us changing this, mind pointing me to the logic for this?

crates/transaction-pool/src/validator.rs
klkvr: yeah if it's never actually used i'd rather not have this arg

crates/transaction-pool/src/amm.rs
0xrusowsky: the auditor's concern is that there could be a DOS attack if:
1. there is not enough liquidity for a validator token
2. the validator (proposer) that accepts that token fails to include the tx in their block
3. other validators (future proposers) which want different tokens cannot include the txs due to lack of liquidity
4. repeat 1-3 to slowly fill the mempool

crates/transaction-pool/src/validator.rs
fgimenez: makes sense, removed

crates/transaction-pool/src/amm.rs
fgimenez: exactly, and additionally a malicious validator could intentionally populate the cache with their token then pause it, causing txs to pass validation but fail at execution

crates/transaction-pool/src/amm.rs
fgimenez: agreed, already removed. all non-base-token transactions now require verified AMM liquidity

crates/transaction-pool/src/amm.rs
fgimenez: i think so, lmk if not, this comes from:
* when `fee_token == validator_token`, no swap happens https://github.com/tempoxyz/tempo/blob/6d1e98744d975d0a8efcb5c6b62fce6908153434/crates/precompiles/src/tip_fee_manager/mod.rs#L152-L164

* `DEFAULT_FEE_TOKEN` definition https://github.com/tempoxyz/tempo/blob/6d1e98744d975d0a8efcb5c6b62fce6908153434/crates/contracts/src/precompiles/mod.rs#L24
* `get_validator_token` defaults to `DEFAULT_FEE_TOKEN` (`PATH_USD` per above) when a validator hasn't set a preference  https://github.com/tempoxyz/tempo/blob/6d1e98744d975d0a8efcb5c6b62fce6908153434/crates/precompiles/src/tip_fee_manager/mod.rs#L43-L51 

is this correct?

legion2002: I don't think this is correct. the code says that validator tokens are set to Path USD by default, but validators have the ability to change this default to something they want. In which case they do have to swap the Path USD to their own validator token

crates/transaction-pool/src/amm.rs
legion2002: but doesn't this mean that if a token doesn't have enough liquidity, no one can send a tx using it, even if the validator is willing to accept it directly?

legion2002 (changes requested):

crates/transaction-pool/src/amm.rs
fgimenez: good call, removed

crates/transaction-pool/src/amm.rs
fgimenez: users can submit via the subblock path (`subblock_proposer` field), which bypasses mempool and AMM liquidity checks entirely. When `user_token == validator_token`, no swap is needed. lmk if this is ok, aiui we can't keep the bypass if we want to avoid the DoS attack vector (if user's fee token is in unique_tokens, skip AMM liquidity check)

klkvr: my assumption when implementing this was that most likely until we introduce permissionless validatorset
1. all validators will be honest
2. all validators will attempt to maximize their profits
3. all validators will use major stablecoins with enough liquidity

I think if we expect 3 to be true then removing this is fine anyway, though imo 1 and 2 are likely enough to keep it

legion2002: I think fine to keep the bypass, and work under honest validato set assumption for now.
This is a low likelihood and low impact vector, so we can reconsider later once we have a permissionless validator set

fgimenez: makes sense to me, closing for now

PR #2223, opened 2026-01-22
fix(txpool): reject expired access keys and KeyAuthorizations in mempool validation

fgimenez:
Closes CHAIN-451

Adds expiry validation for access keys and KeyAuthorizations in mempool validation, preventing expired credentials from entering the transaction pool. Transactions using expired access keys or KeyAuthorizations could enter the mempool and persist until block execution rejected them. This wasted pool resources and delayed rejection feedback to users.

The PR extends `validate_against_keychain()` to check expiry timestamps against current time:
- reject KeyAuthorizations where `expiry <= current_time`
- reject keychain signatures where the access key's `expiry <= current_time`
- added two new error variants: `AccessKeyExpired` and `KeyAuthorizationExpired`.

### DIFF
--- a/crates/node/tests/it/tempo_transaction.rs
+++ b/crates/node/tests/it/tempo_transaction.rs
@@ -4869,46 +4869,26 @@ async fn test_aa_keychain_expiry() -> eyre::Result<()> {
 
     let signed_tx = AASigned::new_unhashed(after_expiry_tx, expired_key_sig);
     let envelope: TempoTxEnvelope = signed_tx.into();
-    let tx_hash = *envelope.tx_hash();
     let mut encoded = Vec::new();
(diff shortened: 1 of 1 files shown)


crates/transaction-pool/src/validator.rs
```diff
@@ -115,15 +119,28 @@ where
                 .recover_signer()
                 .is_ok_and(|signer| signer == transaction.sender())
             {
-                return Ok(Err("Invalid KeyAuthorization signature"));
+                return Ok(Err(TempoPoolTransactionError::Keychain(
+                    "Invalid KeyAuthorization signature",
+                )));
```
mattsse, line 124: nit @gakonst can we do a followup pr here that converts these string based errors into an enum

PR #2225, opened 2026-01-22
feat(rpc): add consensus_epochNumber RPC method

gakonst:
## Summary
Add a new RPC method `consensus_epochNumber` that returns the current consensus epoch number based on the latest finalized block height.

## Changes
- Add `consensus_epochNumber` to `TempoConsensusApi` trait
- Add `epoch_number()` to `ConsensusFeed` trait
- Implement in `FeedStateHandle` using the epocher's `containing(height)` method

## Usage
```bash
curl -X POST -H 'Content-Type: application/json' \
  --data '{"jsonrpc":"2.0","method":"consensus_epochNumber","params":[],"id":1}' \
  http://localhost:8545
```

Returns the epoch number (block_height / epoch_length) or null if consensus is not ready.

### DIFF
--- a/crates/commonware-node/src/feed/state.rs
+++ b/crates/commonware-node/src/feed/state.rs
@@ -459,8 +459,12 @@ impl ConsensusFeed for FeedStateHandle {
         })
     }
 
-    async fn epoch_number(&self) -> Option<u64> {
-        let mut mailbox = self.epoch_manager.read().clone()?;
+    async fn epoch_number(&self) -> eyre::Result<Option<u64>> {
+        let mut mailbox = self
+            .epoch_manager
+            .read()
+            .clone()
+            .ok_or_else(|| eyre::eyre!("epoch manager not yet initialized"))?;
         mailbox.get_current_epoch().await
     }
 }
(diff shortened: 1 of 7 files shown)


crates/commonware-node/src/feed/state.rs
```diff
@@ -451,6 +451,15 @@ impl ConsensusFeed for FeedStateHandle {
             transitions,
         })
     }
+
+    async fn epoch_number(&self) -> Option<u64> {
+        let epocher = self.epocher()?;
+        let state = self.state.read();
+        let height = state.latest_finalized.as_ref()?.height?;
```
SuperFluffy: This isn't guaranteed to return the epoch that the node is in, merely the epoch of the latest block that the EL has finalized.

scripts/get-epoch.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+# Fetch the current epoch from the consensus metrics endpoint
+
+METRICS_URL="${1:-http://127.0.0.1:8001/metrics}"
+
+curl -s "$METRICS_URL" | grep -E "epoch_manager_latest_epoch" | grep -v "^#"
```
SuperFluffy, line 6: IMO we should just be doing this.

Code = liability. If all you want is to get the latest epoch from a node, then this simple script does the job.

PR #2229, opened 2026-01-23
fix: cap pool transactions at non_shared_gas_limit to reserve subblock gas

legion2002:
## Summary

- **Fixes audit issue**: Payload builder was allowing pool transactions to consume up to `block_gas_limit` (500M), but the executor reserves `shared_gas_limit` (50M) for validator subblocks. 
- **Root cause**: TIP-1010 changed the gas cap from `non_shared_gas_limit` to `block_gas_limit` without updating the executor's section transition logic
- **Impact**: When pool txs exceeded 450M gas, executor transitioned to GasIncentive section before subblocks, rejecting them with "subblock section already passed"

Should also fix the devnet chain halt
### Changes

- `crates/payload/builder/src/lib.rs`: Cap pool transactions at `non_shared_gas_limit` (450M) instead of `block_gas_limit` (500M)
- `tips/tip-1010.md`: Clarify the 450M/50M gas split between proposer transactions and validator subblocks, reference Sub-block Specification

### DIFF
--- a/bin/tempo/src/tempo_cmd.rs
+++ b/bin/tempo/src/tempo_cmd.rs
@@ -1,5 +1,6 @@
 use std::{collections::HashMap, fs::OpenOptions, path::PathBuf, sync::Arc};
 
+use alloy_primitives::Address;
 use alloy_provider::Provider;
 
 use alloy_rpc_types_eth::TransactionRequest;
@@ -134,6 +135,8 @@ struct ValidatorInfoOutput {
 /// Individual validator entry
 #[derive(Debug, Serialize)]
 struct ValidatorEntry {
+    /// onchain address of the validator
+    onchain_address: Address,
     /// ed25519 public key (hex)
     public_key: String,
     /// Inbound IP address for p2p connections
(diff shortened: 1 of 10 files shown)


tips/tip-1010.md
```diff
@@ -125,7 +127,7 @@ At full payment lane utilization:
 
 2. **Payment Lane Priority**: Transactions qualifying for the payment lane MUST be able to consume up to the remaining block gas capacity (total gas limit minus gas already consumed by general transactions).
 
-3. **Shared Gas Pool**: The payment lane and general transactions share a common block gas limit (500M). General transactions are additionally constrained by `general_gas_limit` (30M), ensuring payment transactions always have access to at least 470M gas.
+3. **Shared Gas Pool**: Proposer pool transactions (payment and general) share the non-shared gas budget (450M). General transactions are additionally constrained by `general_gas_limit` (30M). The remaining 50M is reserved for validator subblocks.
```
klkvr, line 130: i don't think this is entirely correct. there's also incentive gas available to proposer which is part of shared gas not used by subblocks.

crates/payload/builder/src/lib.rs
```diff
@@ -262,6 +262,9 @@ where
 
         let block_gas_limit: u64 = parent_header.gas_limit();
         let shared_gas_limit = block_gas_limit / TEMPO_SHARED_GAS_DIVISOR;
+        // Non-shared gas limit is the maximum gas available for proposer's pool transactions.
+        // The remaining `shared_gas_limit` is reserved for validator subblocks.
+        let non_shared_gas_limit = block_gas_limit - shared_gas_limit;
```
legion2002, line 267: Tried adding a test, but doesn't seem like there is an easy way to test this.

PR #2236, opened 2026-01-25
TIP-1016: Exempt storage creation from gas limits

dankrad:
## Summary

TIP-1016 proposes exempting storage creation gas from protocol limits (transaction and block gas limits) while still charging users for it. This enables higher contract code pricing (2,500 gas/byte vs 1,000) and better throughput for new account operations.

### 1. Contract Deployment Constraints
TIP-1000's 1,000 gas/byte for contract code forces us to:
- Keep transaction gas cap at 30M (would prefer 16M)
- Keep general gas limit at 30M (would prefer 25M target)
- Can't increase to desired 2,500 gas/byte pricing

**With TIP-1015**: 24KB contracts cost ~64M total gas but only ~7M execution gas. Can deploy with 16M protocol limit.

### 2. New Account Throughput Penalty
- TIP-20 to new address: 271,000 gas (26k execution + 245k storage creation) vs 24,000 to existing
- At 500M payment lane limit: only 1,847 new account transfers/block = 3,694 TPS
- With exemption: ~19,000 new account transfers/block = ~38,000 TPS
- **~10x throughput improvement**

## Design

Storage creation operations split their cost into two components:

**Execution gas** (computational overhead):
- Writing, hashing, memory operations
- Counts toward protocol limits (transaction and block)

**Storage creation gas** (permanent storage burden):
- Does NOT count toward protocol limits
- Still counts toward user's `gas_limit` (prevents surprise costs)

```
User Authorization:
  execution_gas + storage_creation_gas <= transaction.gas_limit

Protocol Limits:
  execution_gas <= max_transaction_gas_limit (EIP-7825, e.g. 16M)
  general_execution_gas <= general_gas_limit (25M target)
  payment_execution_gas <= payment_lane_limit (500M)

Cost:
  total_gas = execution_gas + storage_creation_gas
  cost = total_gas × (base_fee_per_gas + priority_fee)
```

## Gas Split for Storage Creation Operations

| Operation | Execution Gas | Storage Creation Gas | Total | Against Limits |
|-----------|---------------|----------------------|-------|----------------|
| Cold SSTORE (zero → non-zero) | 5,000 | 245,000 | 250,000 | 5,000 |
| Hot SSTORE (non-zero → non-zero) | 2,900 | 0 | 2,900 | 2,900 |
| Account creation | 5,000 | 245,000 | 250,000 | 5,000 |
| Contract code (per byte) | 200 | 2,300 | 2,500 | 200 |
| Contract metadata | 5,000 | 495,000 | 500,000 | 5,000 |

**Note**: Hot SSTORE (updates) has NO storage creation gas component - only new storage is exempted.

## Key Benefits

1. **Higher contract code pricing**: 2,500 gas/byte (vs 1,000) = $50M to fill 1TB vs $20M
2. **Lower protocol transaction limit**: Can use 16M while deploying 24KB contracts
3. **Better new account throughput**: ~19,000 transfers/block vs 1,847 in TIP-1000
4. **Accounts for computational cost**: Operations still charge execution gas (not free)
5. **No surprise costs**: User's `gas_limit` still bounds total cost
6. **No complexity**: Single gas unit, one basefee, existing transaction format

## Benefits vs Brick (PR #2232)

✅ **No complexity**: One gas unit, one basefee, existing transaction format works
✅ **No griefing**: Transaction gas limit still bounds execution (CPU/memory)
✅ **No breaking changes**: RPC methods, gas estimation work as-is
✅ **No unbounded costs**: Users pay for storage creation gas in their authorization

## Example: 24KB Contract Deployment

- Execution gas: ~7M (deployment logic + computational overhead)
- Storage creation gas: ~57M (contract code + metadata + account)
- **Total cost**: ~64M gas (user must authorize)
- **Against protocol limits**: ~7M (can deploy with 16M limit)

---

Compare with PR #2232 (brick approach) to see trade-offs.


tips/tip-yyyy.md
```diff
@@ -0,0 +1,163 @@
...
+---
+
+# Specification
+
+## Gas Accounting
+
+All operations continue to consume gas as specified in TIP-1000. Gas is now categorized into two types for accounting purposes:
+
+- **Execution gas**: Compute, memory, storage reads/updates, calldata
+- **Storage gas**: New state elements, account creation, contract code storage
```
legion2002: Do we have any reason to add a protocol cap for the storage gas to some limit too? 

Could a DOS vector be filling a block with a lot of storage write txs, causing some kind of DB ops bottleneck while syncing?

tips/tip-yyyy.md
```diff
@@ -0,0 +1,163 @@
...
+id: TIP-YYYY
+title: Exempt Storage Creation from Gas Limits
+description: Storage creation gas costs are charged but don't count against transaction or block gas limits, enabling higher contract code pricing and better throughput for new account operations.
+authors: Dankrad Feist @dankrad
+status: Draft
+related: TIP-1000, TIP-1010
+protocolVersion: TBD
+---
+
+# TIP-YYYY: Exempt Storage Creation from Gas Limits
```
legion2002: This direction makes sense to me. Much better than doing multidimensional gas for storage.

tips/tip-yyyy.md
dankrad: That would still incur the normal read/write costs so it's not free. I think they should cover the DB/Merkle root cost already?

tips/tip-yyyy.md
```diff
@@ -0,0 +1,163 @@
...
+
+1. **User Authorization**: `execution_gas + storage_gas` MUST NOT exceed `transaction.gas_limit` (prevents surprise costs)
+2. **Protocol Transaction Limit**: `execution_gas` MUST NOT exceed `max_transaction_gas_limit` (EIP-7825 limit, e.g. 16M)
+3. **Protocol Block Limits**: Block `execution_gas` MUST NOT exceed applicable limit:
+   - General transactions: `general_gas_limit` (25M target, currently 30M)
+   - Payment lane transactions: `payment_lane_limit` (500M)
+4. **Storage Gas Exemption**: Storage gas MUST NOT count toward protocol limits (transaction and block)
+5. **Total Cost**: Transaction cost MUST equal `(execution_gas + storage_gas) × (base_fee_per_gas + priority_fee)`
+6. **Classification**: Every operation MUST be classified as either execution gas or storage gas, not both
+7. **Storage Gas Operations**: Storage creation MUST consume storage gas; compute/memory/reads MUST consume execution gas
```
rakita: Would add storage gas must be less than block gas limit.

dankrad: I don't understand, that's exactly what we don't want?

rakita: There is still a limit check that should be done. Storage gas in the block needs to be bound by something i guess. For example.

The storage gas needs to be less than the block gas limit.

tips/tip-yyyy.md
```diff
@@ -0,0 +1,163 @@
...
+- Storage doesn't reduce block execution capacity or prevent large contracts
+- Note: Tempo has two block limits - general gas limit (~25M) for contracts and payment lane limit (500M) for simple transfers
+
+## Storage Gas Operations
+
+The following operations consume storage gas (not counted against limits):
+
+| Operation | Gas Cost | Counted Against Limits |
+|-----------|----------|------------------------|
+| New state element (SSTORE zero → non-zero) | 250,000 | No |
```
danrobinson: I think clearly you need to count something against the gas limit. This makes it seem like an SSTORE would count not at all when it's a fresh SSTORE but would count when it's an update, which doesn't make sense. It should count at least as much toward the gas limit as a non-fresh SSTORE does.

tips/tip-yyyy.md
dankrad: OK, I think you were right and I missed that in the previous spec creation did not count towards the limit at all, which was wrong. Restored this to a more sane version where only the additional costs introduced by TIP-1000 are exempt, but the previous storage creation costs still count.

PR #2242, opened 2026-01-26
feat(txpool): evict stale pending transactions to mitigate DoS from dynamic state changes

fgimenez:
Ref CHAIN-442
Ref CHAIN-443

Evicts pending transactions that have been stuck for >30 minutes to mitigate DoS from transactions that appear valid but consistently fail during block building (e.g., due to AMM liquidity changes).

- Add `PendingStalenessTracker` to track when transactions first became pending
- Evict stale transactions after 30 minutes

### DIFF
--- a/crates/transaction-pool/src/maintain.rs
+++ b/crates/transaction-pool/src/maintain.rs
@@ -223,19 +223,19 @@ impl PendingStalenessTracker {
         }
     }
 
-    /// Checks for stale transactions and updates the snapshot if the interval has elapsed.
+    /// Returns true if the staleness check interval has elapsed and a snapshot should be taken.
+    fn should_check(&self, now: u64) -> bool {
+        self.last_snapshot_time
+            .is_none_or(|last| now.saturating_sub(last) >= self.interval_secs)
+    }
+
+    /// Checks for stale transactions and updates the snapshot.
     ///
     /// Returns transactions that have been pending across two consecutive snapshots
     /// (i.e., pending for at least one full interval).
+    ///
+    /// Call `should_check` first to avoid collecting the pending set on every block.
     fn check_and_update(&mut self, current_pending: HashSet<TxHash>, now: u64) -> Vec<TxHash> {
-        let should_check = self
-            .last_snapshot_time
-            .is_none_or(|last| now.saturating_sub(last) >= self.interval_secs);
-
-        if !should_check {
-            return Vec::new();
-        }
-
         // Find transactions present in both snapshots (stale)
         let stale: Vec<TxHash> = self
             .previous_pending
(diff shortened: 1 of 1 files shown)


crates/transaction-pool/src/maintain.rs
```diff
@@ -30,6 +30,73 @@ use tempo_precompiles::{
...
+/// Default threshold for pending transaction staleness eviction (30 minutes).
+/// Transactions that have been pending for longer than this will be evicted.
+pub const DEFAULT_PENDING_STALENESS_THRESHOLD: u64 = 30 * 60;
+
+/// Tracks when transactions first became pending to detect stale transactions.
+///
+/// Used to evict transactions that have been pending for too long but can't be included
+/// (e.g., due to dynamic state changes like AMM liquidity).
+#[derive(Debug)]
+pub struct PendingStalenessTracker {
```
mattsse: I'm leaving this comment an almost every pr hehe but I really really dislike how amp orders new types, this should not be top of file

crates/transaction-pool/src/maintain.rs
```diff
@@ -454,6 +523,25 @@ where
                         .record(invalidation_start.elapsed());
                 }
 
+                // 10. Track and evict stale pending transactions
+                // Update the pending set tracking (must happen after AA pool promotions in step 6)
+                let all_txs = pool.all_transactions();
+                let current_pending = all_txs.pending.iter().map(|tx| *tx.hash());
+                state.pending_staleness.update_pending(current_pending, tip_timestamp);
```
mattsse: I think this should work slightly differently, like

every x mins take a snapshot of the pending transactions, compare this what we've already stored, if the tx still exists, clear, track remaining pending txs

mattsse (changes requested):

crates/transaction-pool/src/maintain.rs
fgimenez: done, makes sense :)  will prepare a PR to persist this

crates/transaction-pool/src/maintain.rs
fgimenez: better, done

crates/transaction-pool/src/maintain.rs
```diff
@@ -465,6 +534,25 @@ where
                         .record(invalidation_start.elapsed());
                 }
 
+                // 10. Evict stale pending transactions (must happen after AA pool promotions in step 6)
+                // Takes a snapshot every interval and evicts transactions that remain pending
+                // across two consecutive snapshots.
+                let all_txs = pool.all_transactions();
```
mattsse: there should be pending_transactions as well

mattsse: is this running on every block now?

we only need to run this per interval, right?

fgimenez: indeed, fixed

mattsse (approved):

PR #2243, opened 2026-01-26
docs(validator-config): document chain-halt mitigation for mid-epoch key updates

fgimenez:
Closes CHAIN-458

Documents a potential chain-halt vulnerability in `update_validator()` where validators could update their public key mid-epoch, causing DKG panic at epoch boundary.

Added security comment explaining:
- The vulnerability mechanism (DKG manager panics when original key can't be mapped)
- The mitigation (validator addresses are non-user-controllable in genesis config)

### DIFF
--- a/crates/precompiles/src/validator_config/mod.rs
+++ b/crates/precompiles/src/validator_config/mod.rs
@@ -190,14 +190,13 @@ impl ValidatorConfig {
     ///
     /// # Security Note
     ///
+    /// The field `validator_address` must never be set to a user-controllable address.
+    ///
     /// This function allows validators to update their own public key mid-epoch, which could
     /// cause a chain halt at the next epoch boundary if exploited (the DKG manager panics when
     /// the original key stored in the boundary header cannot be mapped to the modified registry).
-    ///
-    /// This is mitigated by setting validator addresses to non-user-controllable addresses in the
-    /// genesis config, ensuring only the contract owner (admin) can effectively call this function.
-    /// As long as validator addresses are not set to user-controllable EOAs, this vulnerability
-    /// cannot be exploited.
+    /// By setting validator addresses to non-user-controllable addresses in the genesis config,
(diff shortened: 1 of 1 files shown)


crates/commonware-node/src/feed/state.rs
```diff
@@ -432,7 +432,7 @@ impl ConsensusFeed for FeedStateHandle {
                                 merged.push(t.clone());
                             }
                         }
-                        merged.sort_by(|a, b| b.transition_epoch.cmp(&a.transition_epoch));
```
SuperFluffy: Wasn't this change already merged yesterday? https://github.com/tempoxyz/tempo/pull/2244

crates/precompiles/src/validator_config/mod.rs
```diff
@@ -187,6 +187,17 @@ impl ValidatorConfig {
...
 
     /// Update validator information (and optionally rotate to new address)
+    ///
+    /// # Security Note
+    ///
+    /// This function allows validators to update their own public key mid-epoch, which could
+    /// cause a chain halt at the next epoch boundary if exploited (the DKG manager panics when
+    /// the original key stored in the boundary header cannot be mapped to the modified registry).
+    ///
+    /// This is mitigated by setting validator addresses to non-user-controllable addresses in the
```
SuperFluffy: Reverse this, put it right after `# Security note`, something like:

```
The field `validator_address` must never be set to a user controllable address`.
```

And then follow with the reason.

crates/commonware-node/src/feed/state.rs
fgimenez: yes, the branch was not rebased on top of main

crates/precompiles/src/validator_config/mod.rs
fgimenez: ok done

PR #2246, opened 2026-01-26
refactor(txpool): aggregate block-level events into BlockPoolUpdates for single-pass receipt scanning

fgimenez:
Closes CHAIN-443

Consolidates block-level transaction pool updates into a single `BlockPoolUpdates` struct, replacing 4 separate receipt-scanning loops with a single pass over chain receipts.

- Introduce `BlockPoolUpdates` to aggregate all invalidation events (expired txs, revoked keys, validator token changes, blacklist additions, pause events, mined tx hashes)
- Add `from_chain()` for single-pass receipt extraction
- Update `evict_invalidated_transactions()` to accept `&BlockPoolUpdates` instead of 3 separate slice arguments
- Add blacklisted fee payer checks

### DIFF
--- a/crates/transaction-pool/src/maintain.rs
+++ b/crates/transaction-pool/src/maintain.rs
@@ -2,6 +2,7 @@
 
 use crate::{
     TempoTransactionPool,
+    metrics::TempoPoolMaintenanceMetrics,
     paused::{PausedEntry, PausedFeeTokenPool},
     transaction::TempoPooledTransaction,
     tt_2d_pool::AASequenceId,
@@ -18,6 +19,7 @@ use reth_transaction_pool::{PoolTransaction, TransactionOrigin, TransactionPool}
 use std::{
     collections::{BTreeMap, HashMap, HashSet},
     sync::Arc,
+    time::Instant,
 };
 use tempo_chainspec::TempoChainSpec;
 use tempo_contracts::precompiles::{IAccountKeychain, IFeeManager, ITIP20, ITIP403Registry};
@@ -184,6 +186,7 @@ where
         + 'static,
 {
     let mut state = TempoPoolState::default();
+    let metrics = TempoPoolMaintenanceMetrics::default();
 
     // Subscribe to new transactions and chain events
     let mut new_txs = pool.new_transactions_listener();
@@ -264,6 +267,8 @@ where
                     CanonStateNotification::Commit { new } => new,
                 };
 
+                let block_update_start = Instant::now();
+
                 let tip = &new;
                 let bundle_state = tip.execution_outcome().state().state();
                 let tip_timestamp = tip.tip().header().timestamp();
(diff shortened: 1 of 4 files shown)


crates/transaction-pool/src/maintain.rs
```diff
@@ -20,13 +20,115 @@ use std::{
...
 };
 use tempo_primitives::{AASigned, TempoPrimitives};
 use tracing::{debug, error};
 
+/// Aggregated block-level updates for the transaction pool.
+///
+/// Collects all invalidation events from a block into a single structure,
+/// allowing efficient batch processing of pool updates.
+#[derive(Debug, Default)]
+pub struct BlockPoolUpdates {
```
mattsse: let's call these TempoPoolUpdates?

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -235,6 +240,50 @@ where
...
+                        provider
+                            .storage(fee_token, tip20_slots::TRANSFER_POLICY_ID.into())
+                            .ok()
+                            .flatten()
+                            .map(|packed: alloy_primitives::U256| {
+                                let shifted: alloy_primitives::U256 = packed >> 160;
+                                let policy_id: u64 = shifted.to();
+                                policy_cache.insert(fee_token, policy_id);
+                                policy_id
+                            })
```
mattsse: dont we have a helper for this, this?

crates/transaction-pool/src/maintain.rs
```diff
@@ -20,13 +20,115 @@ use std::{
...
+    /// Revoked keychain keys: (account, public_key).
+    pub revoked_keys: Vec<(Address, Address)>,
+    /// Validator token preference changes: (validator, new_token).
+    pub validator_token_changes: Vec<(Address, Address)>,
+    /// TIP403 blacklist additions: (policy_id, account).
+    pub blacklist_additions: Vec<(u64, Address)>,
+    /// Fee token pause state changes: (token, is_paused).
+    pub pause_events: Vec<(Address, bool)>,
+    /// Transaction hashes mined in this block.
+    pub mined_tx_hashes: Vec<TxHash>,
```
mattsse: should we treat them separately?

crates/transaction-pool/src/maintain.rs
fgimenez: sure done

crates/transaction-pool/src/tempo_pool.rs
fgimenez: yep, using `extract_from_word` now

crates/transaction-pool/src/maintain.rs
fgimenez: indeed, separated `mined_tx_hashes` from `TempoPoolUpdates` only invalidation events left

PR #2249, opened 2026-01-26
feat(precompiles): OZ enumerable set

0xrusowsky:
## Motivation

Manually implementing enumerable sets in EVM storage is error-prone—incorrect position indexing or missing cleanup causes silent data corruption. A first-class `Set<T>` type provides safe, OZ-compatible semantics with a familiar API, letting developers focus on logic rather than storage invariants.

## Solution

Implements `Set<T>` and `SetHandler<T>` following [OpenZeppelin's EnumerableSet](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/structs/EnumerableSet.sol).

- **storage:** (takes 2 slots) values array at `keccak256(base_slot)`, 1-indexed positions mapping at `base_slot + 1`
- **operations:** `insert`, `remove`, `contains`, `at`, `read`, `read_range`, `delete`

Additionally optimizes `HandlerCache` to take `&K` references, avoiding unnecessary clones on cache hits.

Finally, the added integration tests mirror [OZ's test suite](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/test/utils/structs/EnumerableSet.behavior.js).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,31 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    labels:
+      - "A-dependencies"
+      - "A-ci"
+    commit-message:
+      prefix: "chore(ci)"
+    open-pull-requests-limit: 1
+    groups:
+      actions-weekly:
+        applies-to: "version-updates"
+        patterns: ["*"]
+        update-types: ["minor", "patch"]
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    labels:
+      - "A-dependencies"
+    commit-message:
+      prefix: "chore(deps)"
+    open-pull-requests-limit: 1
+    groups:
+      cargo-weekly:
+        applies-to: "version-updates"
+        patterns: ["*"]
+        update-types: ["minor", "patch"]
--- a/.github/workflows/amp-review.yml
+++ b/.github/workflows/amp-review.yml
@@ -1,0 +1,30 @@
+name: Amp Code Review
+
+on:
+  pull_request:
+    types: [opened, reopened, ready_for_review, labeled]
+
+jobs:
+  review:
+    # Only run on PRs with "amp" label, skip drafts
+    if: |
+      contains(github.event.pull_request.labels.*.name, 'amp') &&
+      !github.event.pull_request.draft
+    runs-on: ubuntu-latest
+    # Ensure only one review runs per PR
+    concurrency:
+      group: amp-review-pr-${{ github.event.pull_request.number }}
+      cancel-in-progress: true
+    permissions:
+      pull-requests: write
+      checks: write
+      contents: read
+    steps:
+      - name: Run Amp Code Review
+        uses: docker://ghcr.io/sourcegraph/cra-github:latest
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          AMP_SERVER_URL: ${{ vars.AMP_SERVER_URL }}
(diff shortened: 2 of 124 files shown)


howydev (approved):
LGTM!

crates/precompiles/src/storage/types/set.rs
```diff
@@ -0,0 +1,622 @@
...
+    where
+        T: StorageKey + Hash + Eq + Clone,
+        T::Handler: Handler<T>,
+    {
+        // Check if already present
+        if self.contains(&value)? {
+            return Ok(false);
+        }
+
+        // Store position (1-indexed: position N means index N-1)
```
legion2002, line 277: Noting that this is 1-indexed, so that `contains` happens in 1 SLOAD.

crates/precompiles/src/storage/types/set.rs
```diff
@@ -0,0 +1,622 @@
...
+        for i in 0..len {
+            let value = self.values[i].read()?;
+            self.positions.at_mut(&value).delete()?;
+        }
+
+        // Delete the underlying vector (clears length and data slots)
+        self.values.delete()
+    }
+
+    fn t_read(&self) -> Result<Set<T>> {
```
legion2002, line 431: Is there a reason to not support this? Or is it just to minimize complexity in the PR?

crates/precompiles/src/storage/types/set.rs
```diff
@@ -0,0 +1,622 @@
...
+            vec.push(self.values[i].read()?);
+        }
+
+        Ok(Set(vec))
+    }
+
+    /// Replaces the entire set with new contents.
+    ///
+    /// The input Set is deduplicated by the `From<Vec<T>>` conversion.
+    fn write(&mut self, value: Set<T>) -> Result<()> {
```
legion2002, line 388: I think we might be able to optimize this for number of SSTOREs, for cases where some old values will be retained in the new set.
But probably not worth doing this, because write won't be used too much anyways.

crates/precompiles/src/storage/types/set.rs
```diff
@@ -0,0 +1,622 @@
...
+    fn from(set: Set<T>) -> Self {
+        set.0
+    }
+}
+
+impl<T: Eq + Clone> From<Vec<T>> for Set<T> {
+    /// Creates a set from a vector, removing duplicates.
+    ///
+    /// Preserves the order of first occurrences.
+    fn from(vec: Vec<T>) -> Self {
```
legion2002, line 102: Is a silent dedup fine here? or should we use something like `TryFrom` here to prevent footguns

crates/precompiles/src/storage/types/set.rs
0xrusowsky: mmm i didn't think there would be a usecase for trasient storage 🤔
do u think it would be useful? @legion2002

crates/precompiles/src/storage/types/set.rs
0xrusowsky: the conversion is infallible, so i don't think `TryFrom` helps in any way.

while it is true that the dedup is silent that's the whole point of using a `Set`, so imho it is good as is

crates/precompiles/src/storage/types/set.rs
0xrusowsky: i think based on what u mention probably worth skipping unless we end up having huge sets with overlap, but that seems rare?

the current impl is O(1), should be quite efficient. also it mimics OZ which kinda gives me extra peace of mind

crates/precompiles/src/storage/types/set.rs
legion2002: don't have anything in mind right now, if we find a use for it in the future, we can update it.

PR #2251, opened 2026-01-26
feat(consensus): add external consensus certificate indexer

Zygimantass:
This will be used to guard our validator nodes against traffic, and eventually will be used for RPC nodes to follow.

The indexer stores finalizations (no notarizations for now) in SQLite, which is fine, given that resyncing an indexer takes around 15 minutes and we do not expect to lose the data.

The indexer also verifies the finalizations against the network identity (and supports identity transitions).


cryt7 (changes requested):
Cryt

PR #2253, opened 2026-01-26
feat(consensus/rpc): add `consensus_getDkgOutcome` RPC query

Zygimantass:
This will be useful in the explorer, where we'll be able to query what validators are dealers/players, instead of having to decode the DKG outcome in TypeScript.

A lot of the inspiration was taken from the xtask. Testing pending localnet working - if it doesn't, I'll boot this up on a devnet node.

Amp-Thread-Id: https://ampcode.com/threads/T-019bfb6f-96c9-7288-85f4-d6ca902c1c88

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12005,6 +12005,7 @@ dependencies = [
  "tempo-chainspec",
  "tempo-consensus",
  "tempo-contracts",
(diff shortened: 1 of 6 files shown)


cryt7 (changes requested):
Cryt

PR #2254, opened 2026-01-26
fix(rpc): correct gas estimation for TIP-1000 nonce=0 on T1+

gakonst:
## Summary

When `eth_estimateGas` is called with an explicit `nonce=0`, the upstream reth implementation clears the nonce via `take_nonce()` and fills it from state. On Tempo T1+, TIP-1000 adds 250,000 gas for `nonce=0` transactions (new account cost).

## Problem

The upstream reth behavior causes gas estimation to undercount when:
1. A user explicitly specifies `nonce=0` in their request
2. The account's on-chain nonce is `> 0`
3. `take_nonce()` clears the `nonce=0` and replaces it with the state nonce
4. TIP-1000's +250k gas for new account cost is not included

This was discovered during devnet testing where `intrinsic gas too low` errors occurred.

## Solution

Override `estimate_gas_with` in `TempoEthApi` to:
1. Capture the original nonce before delegation to parent implementation
2. Call the parent (reth) implementation which handles the standard estimation
3. If T1+ is active AND the original nonce was `0`, add TIP-1000 gas costs:
   - +250,000 for new account cost
   - +447,000 additional for create transactions (500k - 53k delta)

## Testing

Added regression test `test_eth_estimate_gas_tip1000_nonce_zero_explicit` that:
1. Sends a transaction to increment account nonce > 0
2. Estimates gas with explicit `nonce=0` vs `nonce=current`
3. Asserts the difference is exactly 250,000 gas

## Related

- TIP-1000: Storage pricing updates for launch
- Slack thread: https://tempoxyz.slack.com/archives/C0A87C21805/p1769419607162079


cryt7 (changes requested):
Cryt

grandizzy:
not needed

PR #2045, opened 2026-01-14
feat: additional invariant tests & shared invariant base test helper

zerosnacks:
## Summary

Introduces a shared `InvariantBaseTest` helper and adds comprehensive invariant test coverage for multiple core protocol contracts: TIP20, TIP20Factory, TIP403Registry, ValidatorConfig, Nonce, and AccountKeychain.

## Motivation

The existing invariant tests (`FeeAMM`, `StablecoinDEX`) had duplicated setup code and actor management logic. This PR consolidates shared functionality into a base test helper while significantly expanding invariant coverage across the protocol.

## Changes

- **`InvariantBaseTest.t.sol`**: New shared base contract providing common setup logic (`_setupInvariantBase`), actor building with approvals (`_buildActorsWithApprovals`), and contract balance logging
- **`BaseTest.t.sol`**: Extended to support the invariant base helper
- **Refactored existing tests**: `FeeAMM.t.sol` and `StablecoinDEX.t.sol` now use the shared base, reducing ~170 lines of duplicated code

**New invariant test suites:**
| Contract | Invariants | Description |
|----------|------------|-------------|
| `TIP20.t.sol` | TEMPO-TIP20-* | Token supply, balances, transfer policy enforcement |
| `TIP20Factory.t.sol` | TEMPO-FACT-* | Factory deployment consistency, ownership |
| `TIP403Registry.t.sol` | TEMPO-REG-* | Policy creation, admin management, whitelist/blacklist behavior |
| `ValidatorConfig.t.sol` | TEMPO-VAL-* | Validator registration, updates, status management, DKG ceremonies |
| `Nonce.t.sol` | TEMPO-NONCE-* | Nonce monotonicity and bitmap integrity |
| `AccountKeychain.t.sol` | TEMPO-KEY-* | Keychain link/unlink consistency |

## Documentation

Updated `docs/specs/test/invariants/README.md` with complete invariant specifications for all new test suites.

PR #2256, opened 2026-01-27
feat: add base invariant test logic

0xKitsune:
This PR introduces shared test infrastructure for invariant testing of Tempo precompiles.` InvariantBaseTest.t.sol` provides actor management utilities, token selection and setup helpers, blacklist policy creation, funding utilities, and logging infrastructure.

Ref #2045

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -3,29 +3,7 @@ name: Lint
 on:
   push:
     branches: [main]
-    paths:
-      - "crates/**"
-      - "bin/**"
-      - "xtask/**"
-      - "Cargo.toml"
-      - "Cargo.lock"
-      - "rust-toolchain.toml"
-      - "rustfmt.toml"
-      - "deny.toml"
-      - "typos.toml"
-      - ".github/workflows/lint.yml"
   pull_request:
-    paths:
-      - "crates/**"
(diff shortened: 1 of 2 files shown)


tips/ref-impls/test/invariants/InvariantBaseTest.t.sol
```diff
@@ -0,0 +1,688 @@
...
+            || selector == IStablecoinDEX.InvalidTick.selector || _isKnownTIP20Error(selector);
+    }
+
+    /// @dev Asserts a revert is a known TIP20 error
+    function _assertKnownTIP20Revert(bytes memory reason) internal pure {
+        assertTrue(_isKnownTIP20Error(bytes4(reason)), "Unknown TIP20 error");
+    }
+
+    /// @dev Asserts a revert is a known TIP20Factory error
+    function _assertKnownFactoryRevert(bytes memory reason) internal pure {
```
legion2002, line 447: These one line functions might not be super relevant, and we'll probably never remember to use them.

zerosnacks: If we merge all the other invariant PRs they will be used, ok with keeping them and then I can follow up with a cleanup in case they are unused

PR #2262, opened 2026-01-27
test(validator-config): add invariant tests

0xKitsune:
This PR introduces invariant tests for the `ValidatorConfig` precompile. Tests cover validator addition, status changes, public key updates, owner management, and DKG ceremony tracking using ghost state.

Ref #2045


tips/ref-impls/test/invariants/ValidatorConfig.t.sol
```diff
@@ -0,0 +1,655 @@
...
+            "TEMPO-VAL12: DKG epoch should match ghost state"
+        );
+    }
+
+    /*//////////////////////////////////////////////////////////////
+                              HELPERS
+    //////////////////////////////////////////////////////////////*/
+
+    /// @dev Checks if an error is known/expected for ValidatorConfig
+    function _assertKnownValidatorError(bytes memory reason) internal pure {
```
0xKitsune, line 644: IIRC we should have this as part of the base invariant test harness.

PR #2275, opened 2026-01-27
feat(valconfigV2): add manual migration from V1

howydev:
## Summary

**Manual migration**: Admin migrates validators one at a time via `migrateValidator(valIdx)`, then calls `initialize()` to flip the flag. CL reads from V1 until the flag is set.

| Change | Description |
|--------|-------------|
| **Read before init** | CL reads directly from V1 (no proxying in V2) |
| **migrateValidator(valIdx)** | Migrates single validator; copies owner on first call; reverts if `valIdx != validatorsArray.length` |
| **initialize()** | Copies `nextDkgCeremony` from V1, sets `initialized = true` |
| **Active validators** | `addedAtHeight = 0`, `deletedAtHeight = 0` |
| **Inactive validators** | `addedAtHeight = deletedAtHeight = block.timestamp` |
| **Filter logic** | `v.addedAtHeight <= H && (deletedAtHeight == 0 \|\| deletedAtHeight > H)` |

## Migration Steps (Existing Networks)

1. Release new node software with Allegro hardfork support
2. Schedule the fork by updating chainspec with target `allegroTime`
3. At fork activation: CL reads from V1 (since `isInitialized() == false`)
4. Admin migrates validators: `migrateValidator(0)`, `migrateValidator(1)`, ...
5. Admin calls `initialize()` → sets `initialized = true`, CL reads from V2
6. Post-migration: All reads/writes use V2 state directly

## New Networks

1. Initialize V2 at genesis with the initial validator set (set `initialized = true`)
2. Set `allegroTime = 0` to activate V2 immediately
3. V1 Validator Config contract/precompile is not necessary

### DIFF
--- a/docs/pages/protocol/tips/tip-xxxx.mdx
+++ b/docs/pages/protocol/tips/tip-xxxx.mdx
@@ -733,8 +733,7 @@ The test suite must cover:
     after `initialize()`
 38. **initialize owner only**: Non-owner cannot call `initialize`
 39. **isInitialized returns correct value**: Returns false before initialize, true after
-40. **Writes blocked before init**: `addValidator`, `deleteValidator`, `rotateValidator`,
-    `setValidatorAddress` revert if `isInitialized() == false` (only migration functions allowed)
+40. **Writes blocked before init**: `addValidator`, `deleteValidator`, `rotateValidator`
 41. **initialize reverts if not all migrated**: `initialize()` reverts if
     `validatorsArray.length < V1.getValidators().length`
 


docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -443,33 +454,84 @@ if chainspec.is_allegro_active_at_timestamp(block.timestamp) {
...
+- `validatorByPublicKey(bytes32 publicKey)`
+- `addValidator(...)`
+- `initialize()`
+
+### Initialization Procedure
+
+When any triggering function is called and `initialized == false`:
+
+1. Read all validators from ValidatorConfig V1 at
+   `0xCCCCCCCC00000000000000000000000000000001`
```
SuperFluffy: I think the current address is `0xCCCCCCCC00000000000000000000000000000000` (not `0xC....1`) - but best double check.

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -443,33 +455,84 @@ if chainspec.is_allegro_active_at_timestamp(block.timestamp) {
...
+2. For each V1 validator where `active == true`:
```
SuperFluffy: Why `active == true`? What happens if we have a failed DKG that took place in the epoch in which the hardfork happend? Then a new/restarted node needs to potentially get IP addresses of nodes that are not in v2.

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -443,33 +455,84 @@ if chainspec.is_allegro_active_at_timestamp(block.timestamp) {
...
+     - `index`: assigned sequentially (0, 1, 2, ...)
+     - `addedAtHeight`: set to `block.number`
+     - `deletedAtHeight`: set to 0 (active)
+   - Add to `validatorsArray`
+   - Populate lookup maps (`validators[addr]`, `pubkeyToIndex[pubkey]`)
+3. Copy `nextDkgCeremony` from V1 to V2 (if V1 read succeeds)
+4. Set `initialized = true` (bit 255 of slot 0)
+
+If the V1 read fails at any point, skip steps 2-3 and proceed directly to step 4.
+
```
SuperFluffy: This seems extremely dangerous.

docs/pages/protocol/tips/tip-xxxx.mdx
howydev: ur right, good catch

docs/pages/protocol/tips/tip-xxxx.mdx
howydev: Hmm, I assumed `active == false` is equivalent to a non-zero `deletedAtHeight`

SuperFluffy: Yes, but just because `active = false` that does not mean that the validator is actually inactive. The validator config records the validators that *should* be active (these are always becoming the next players). But if the DKG fails, then the network falls back to the previous state which can include deleted validators that have `active = false`.

docs/pages/protocol/tips/tip-xxxx.mdx
howydev: Since this is a read outside the EVM (no gas limit), it should really only fail in instances when we deprecate v1 entirely and just have v2 right? 

Either way, not having any v1 state should not be a blocker for starting v2 consensus, but if there's a safer pattern/option here would be happy to do that also

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -443,33 +455,84 @@ if chainspec.is_allegro_active_at_timestamp(block.timestamp) {
...
+- `validatorByPublicKey(bytes32 publicKey)`
+- `addValidator(...)`
+- `initialize()`
+
+### Initialization Procedure
+
+When any triggering function is called and `initialized == false`:
+
+1. Read all validators from ValidatorConfig V1 at
+   `0xCCCCCCCC00000000000000000000000000000001`
```
howydev:
```suggestion
   `0xCCCCCCCC00000000000000000000000000000000`
```

docs/pages/protocol/tips/tip-xxxx.mdx
SuperFluffy: Ah, I was just worried that this might be a fallible operation where we end up with a partial migration, which seems bad? Can we enforce somehow that this migration is atomic?

howydev: Hmm one possible pattern to get some guarantees could be to pass `bool revertOnFailedMigration` to `initialize`, but then because this is used lazily across all the functions we'd either need the validator to call `initialize` first, or to have this param for all the functions

howydev: If we can guarantee that a call to `initialize` would always happen on genesis, then we might not need the lazy init for the other functions

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -492,33 +504,84 @@ if chainspec.is_allegro_active_at_timestamp(block.timestamp) {
...
+   `0xCCCCCCCC00000000000000000000000000000000`
+2. For each V1 validator (both active and inactive):
+   - Create a V2 validator entry with:
+     - `publicKey`: copied from V1
+     - `validatorAddress`: copied from V1
+     - `inboundAddress`: copied from V1
+     - `outboundAddress`: copied from V1
+     - `index`: assigned sequentially (0, 1, 2, ...)
+     - `addedAtHeight`: set to `block.number`
+     - `deletedAtHeight`: if V1 `active == true`, set to 0; if V1 `active == false`, set to `block.number`
```
SuperFluffy: Yeah, that should work.

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -492,33 +504,84 @@ if chainspec.is_allegro_active_at_timestamp(block.timestamp) {
...
+- `validatorByAddress(address validatorAddress)`
+- `validatorByPublicKey(bytes32 publicKey)`
+- `addValidator(...)`
+- `initialize()`
+
+### Initialization Procedure
+
+When any triggering function is called and `initialized == false`:
+
+1. Read all validators from ValidatorConfig V1 at
```
legion2002: Are we sure this will fit in 1 tx? Maybe we should only initialise what is required

legion2002 (commented):
If it's a one time initialisation, then do we need to make it lazy? We can just call a tx, but yes there's a risk the tx doesn't go through before the end of the epoch

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -529,25 +592,27 @@ Before Fork                    Fork Block                      After Fork
 
 ### For Existing Networks (testnet, mainnet)
 
-1. **Release new node software** with Allegro hardfork support and V2 bootstrap
-   method, but set `allegroTime` far in future
+1. **Release new node software** with Allegro hardfork support
 2. **Schedule the fork** by updating chainspec with target `allegroTime`
-3. **At fork activation**: Execute `bootstrapFromV1(migrationHeight)` as a
-   system transaction
+3. **At fork activation**: First V2 read automatically triggers initialization
```
SuperFluffy: I don't think this works. CL reads the validator config v2 from the *finalized* boundary block. It has access to the config state, but unless I am mistaken this state is read-only at that point.

https://github.com/tempoxyz/tempo/blob/48d58a5bc9bcc2ec111f451565a36dc929c16c77/crates/commonware-node/src/dkg/manager/validators.rs#L73-L79

Zygimantass: ++, how does the trigger of the init work? is it a transaction?

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -468,14 +487,12 @@ This section describes the migration strategy from ValidatorConfig V1 to V2.
...
-The migration uses a three-pronged approach:
+The migration uses a two-pronged approach:
 
 1. **New hardfork ("Allegro")**: Timestamp-based activation that switches
    consensus from reading V1 to V2
-2. **Genesis initialization for new networks**: New deployments initialize V2 at
-   genesis
-3. **Bootstrap V2 from V1 for existing networks**: At the fork, copy V1 state
-   into V2
+2. **Lazy initialization**: On first access (read or write), V2 automatically
```
howydev:
```suggestion
2. **Lazy initialization**: On first write, V2 automatically
```

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -492,62 +509,142 @@ if chainspec.is_allegro_active_at_timestamp(block.timestamp) {
...
+1. Read all validators from ValidatorConfig V1 at
+   `0xCCCCCCCC00000000000000000000000000000000`
+2. For each V1 validator (both active and inactive):
+   - Create a V2 validator entry with:
+     - `publicKey`: copied from V1
+     - `validatorAddress`: copied from V1
+     - `inboundAddress`: copied from V1
+     - `outboundAddress`: copied from V1
+     - `index`: assigned sequentially (0, 1, 2, ...)
+     - `addedAtHeight`: set to `block.number`
```
howydev: @SuperFluffy any issue with this? or should we do `0` instead?

docs/pages/protocol/tips/tip-xxxx.mdx
howydev: Revisiting this: read should only fail if we accidentally HF gate the v1 contract or if we delete it as a precompile before doing the migration

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -492,62 +509,142 @@ if chainspec.is_allegro_active_at_timestamp(block.timestamp) {
...
-   with signatures
+3. **At fork activation**: V2 reads proxy through to V1 (no state changes needed)
+4. **On first write** (e.g., `addValidator`, `deleteValidator`): V2 automatically
+   copies V1 state, sets `initialized = true`, then executes the write
+5. **Post-init**: All reads/writes use V2 state directly
 
 ### For New Networks
 
-1. Initialize V2 at genesis with the initial validator set
+1. Initialize V2 at genesis with the initial validator set (set `initialized = true`)
```
howydev: @SuperFluffy if we have v1 read fail open, we would not need the v1 val config contract at all. Everything is set up through genesis `initialize`

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -138,13 +140,14 @@ interface IValidatorConfigV2 {
         uint64 index;
         uint64 addedAtHeight;
         uint64 deletedAtHeight;
+        bool deletedBeforeMigration;
```
howydev: note: solidity will automatically pack this with the prev 3 `uint64's` into a single `word` (32 bytes), and storage pricing is discrete per `word`, so adding this bool incurs no additional storage cost

docs/pages/protocol/tips/tip-xxxx.mdx
SuperFluffy: You mean `addedAtHeight = block.number`, right? I think this is good/desirable.

docs/pages/protocol/tips/tip-xxxx.mdx
SuperFluffy: I think we could get away with setting this to `deletedAtHeight = block.number` instead of `deletedBeforeMigration`.

Any check CL would necessarily be something like `is_active = addedAtHeight > deletedAtHeight`

docs/pages/protocol/tips/tip-xxxx.mdx
SuperFluffy: This only works if the hardfork time is `= 0`.

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -49,10 +49,11 @@ V2 solves these problems with an append-only design where:
 
 ### Key Design Principle
 
-By recording `addedAtHeight` and `deletedAtHeight`, nodes can determine DKG players for any epoch using only current state. When preparing for a DKG ceremony in epoch `E+2`, a node reads the contract at `boundary(E)` and filters:
+By recording `addedAtHeight`, `deletedAtHeight`, and `deletedBeforeMigration`, nodes can determine DKG players for any epoch using only current state. When preparing for a DKG ceremony in epoch `E+2`, a node reads the contract at `boundary(E)` and filters:
```
SuperFluffy: I think `deletedBeforeMigration` is not strictly necessary. We can just `deletedAtHeight = block.number`.

docs/pages/protocol/tips/tip-xxxx.mdx
howydev: Ok

docs/pages/protocol/tips/tip-xxxx.mdx
howydev: I'm not seeing how, could u elaborate why?

SuperFluffy: I am saying that if the hardfork timestamp is `!= 0`, that v2 cannot be initialized.

docs/pages/protocol/tips/tip-xxxx.mdx
howydev: Is there some inaccuracy in the validator set depending on how we implement the logic of tracking validator sets there, if we do `block.number`? 

Example case I'm thinking of: Val config v1 contains a bunch of `inactive` validators. At HF, we start using v2 and do not immediately init. When we init v2, the validators would appear to be active from HF().block.number to init().block.number for historical queries

If this is an issue, we should set it to either `0` or pass in `HF().block.number` to this precompile on construction, and set that value instead. No personal preference on what we should use

docs/pages/protocol/tips/tip-xxxx.mdx
howydev: added security section at the bottom

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -468,86 +499,142 @@ This section describes the migration strategy from ValidatorConfig V1 to V2.
...
 
-**Properties:**
-- Callable only if `bootstrapped == false`
-- After execution, sets `bootstrapped = true` permanently
-- Internally reads V1 state and copies active validators
-- Sets `addedAtHeight = migrationHeight` for imported validators
-- Inactive V1 validators are not imported
+### Properties
+
+- **CL reads V1 until flag flipped**: No read proxying in V2 precompile
```
howydev:
```suggestion
```

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -412,7 +439,7 @@ This ensures historical queries always return consistent results.
 | Key ownership | Not verified | Ed25519 signature required |
 | Historical queries | Requires historical state | Filter `getValidators()` by `addedAtHeight`/`deletedAtHeight` |
 | Uniqueness | Address only | Address AND public key |
-| Precompile address | `0xCCCC...0001` | `0xCCCC...0002` |
+| Precompile address | `0xCCCC...0000` | `0xCCCC...0002` |
```
0xrusowsky, line 450: OCD request: can we have 0-indexed versions so that we don't need to skip 1? i.e. `ValidatorConfig` becomes `V0`, thus this new contract becomes `ValidatorConfigV1`?

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -468,86 +499,142 @@ This section describes the migration strategy from ValidatorConfig V1 to V2.
...
 
-V2 requires Ed25519 signatures on `addValidator`, but V1 validators don't have
-pre-signed messages. To solve this, V2 includes a one-time privileged bootstrap
-mechanism:
+This is handled entirely in the CL logic, not in the V2 precompile. The V2
+precompile does NOT proxy reads to V1.
+
+### Migration Functions (Owner Only)
+
+**`migrateValidator(valConfigV1Idx, valAddress)`**:
```
0xrusowsky: i don't like this interface, as it makes the caller's life harder and kinda prone to error.

imo you want to have `function migrateValidators(uint8 numValidators)`, where:
- `numValidators`: the amount of validators to migrate at a time (you can simulate the call to see how many u can fit in a single tx)

how the function would work:
- reverts if `isInitialized() == true`
- reverts if `V2.validatorsArray.length >= V1.validatorsArray.length`
- on first call (when `V2.validatorsArray.length == 0`) copies `owner` from V1 to V2 if owner is `address(0)`
- migrates the V1 validators from `idx = [V2.validatorsArray.length, V2.validatorsArray.length + numValidators)` by reading the validator addresses from `V1.validatorsArray[idx]` and then querying the `V1.validators[addr]` mapping

SuperFluffy: One problem is that the addresses that we have inside v1 right now are garbage and cannot be used. We want to set them to addresses that are actually controllable by the owner.

What if we allow the contract owner (admin) to change the addresses until the v2 config is frozen?

Of course that still leaves room for human error.

0xrusowsky: ok, i wasn't aware of this.

now i understand the motivation behind the proposed `migrateValidator(valConfigIdx, valAddress)`.

i guess we could then update my proposal to `function migrateValidators(address[] memory newValidators)`, where
- `numValidators = newValidators.length`

everything would remain the same, but you would map each address from V1 to the corresponding address in the `newValidators` array

wdyt @SuperFluffy ?

SuperFluffy: What we know:

1. the addresses in val config v1 are currently all unique, which is good.
2. so we can copy them over as they are
3. we just need to provide the owner with the option to `updateValidatorAddress(address newAddress)`.
4. this is desirable because at the time of migration we cannot ensure that all operators have a wallet already in place.

Then either migration path works.

I like `function migrateValidators(uint8 numValidators)` because it is conceptually simpler: just migrate `n` validators at a time. I don't like it because the indexing is weird (on every call, check how many validators have been migrated to v2, then based on that index into v1).

At first glance I am not fond of `function migrateValidators(address[] memory newValidators)`, because we need to provide an explicit list of addresses. That seems error prone. On the other hand, the explicitness is also good and easy to understand. We know exactly what to migrate.

SuperFluffy: Can't we just do `function migrateValidator(address validatorAddress)` and require admin to do that for every single validator?


And then have a `function activateContract()` (or whatever) that checks if all validators have been migrated?

docs/pages/protocol/tips/tip-xxxx.mdx
howydev: defer to @SuperFluffy on this

docs/pages/protocol/tips/tip-xxxx.mdx
```diff
@@ -600,3 +711,50 @@ The test suite must cover:
...
+34. **addedAtHeight set correctly**: Active validators have `addedAtHeight == 0`,
+    inactive validators have `addedAtHeight == block.timestamp` at migration time
+35. **initialize sets flag**: After `initialize()`, `isInitialized()` returns true
+36. **migrateValidator copies owner**: V2 `owner()` matches V1 after first `migrateValidator` call
+37. **initialize copies DKG ceremony**: V2 `getNextFullDkgCeremony()` matches V1
+    after `initialize()`
+38. **initialize owner only**: Non-owner cannot call `initialize`
+39. **isInitialized returns correct value**: Returns false before initialize, true after
+40. **Writes blocked before init**: `addValidator`, `deleteValidator`, `rotateValidator`,
+    `setValidatorAddress` revert if `isInitialized() == false` (only migration functions allowed)
```
0xrusowsky: based on what was discussed, i believe that `setValidatorAddress` should always be callable (only by admin) regardless of `isInitialized()` status

0xrusowsky (approved):

Issue #214, opened 2025-09-14
Introduce Tempo transaction types and envelope type


mattsse:
ref https://github.com/tempoxyz/specs/blob/fee-manager/specs/TokenPreferences.md

## TODO
* define transaction type
* define envelope (L1 - blob + tempo tx)

this should be a standalone crate (primitives)

PR #2283, opened 2026-01-27
feat(tempo-alloy): add fee_payer_signature support to TempoTransactionRequest

gakonst:
## Summary

Adds support for setting the `fee_payer_signature` field on `TempoTransactionRequest`, enabling sponsored (gasless) transactions via the alloy API.

### New Fields & Methods
- `fee_payer_signature: Option<Signature>` field on `TempoTransactionRequest`
- `set_fee_payer_signature(signature)` method
- `with_fee_payer_signature(signature)` builder method

### Updated Behavior
- `build_aa()` now uses `self.fee_payer_signature` instead of hardcoded `None`
- `From<TempoTransaction>` preserves `fee_payer_signature` field

## Usage

```rust
let mut tx_request = TempoTransactionRequest::default()
    .with_nonce(0)
    .with_gas(21000)
    // ... other fields
    .with_fee_payer_signature(sponsor_signature);

// Or using setter
tx_request.set_fee_payer_signature(sponsor_signature);
```

## Tests

Added 4 unit tests:
- `test_set_fee_payer_signature` - verifies setter method
- `test_with_fee_payer_signature` - verifies builder method
- `test_build_aa_with_fee_payer_signature` - verifies signature is preserved in build_aa()
- `test_from_tempo_transaction_preserves_fee_payer_signature` - verifies roundtrip

## Context

This is a dependency for tempo-foundry PR #214 which adds `--sponsor` flag to `cast send` and `cast mktx` for sponsored transactions.

The flow is:
1. Build the transaction with all fields
2. Compute `fee_payer_signature_hash(sender)`
3. Have sponsor sign that hash
4. Set `fee_payer_signature` on the request
5. Sign and send

### DIFF
--- a/crates/alloy/src/rpc/compat.rs
+++ b/crates/alloy/src/rpc/compat.rs
@@ -303,5 +303,3 @@ impl FromConsensusHeader<TempoHeader> for TempoHeaderResponse {
         }
     }
 }
-
-


crates/alloy/src/rpc/request.rs
```diff
@@ -185,6 +190,17 @@ impl TempoTransactionRequest {
             key_authorization: None,
         })
     }
+
+    /// Set the fee payer signature for sponsored transactions.
+    pub fn set_fee_payer_signature(&mut self, signature: alloy_primitives::Signature) {
```
0xrusowsky: nit: we should move the fee payer sig methods above `fn build_aa`

crates/alloy/src/rpc/compat.rs
```diff
@@ -106,6 +109,7 @@ impl TryIntoTxEnv<TempoTxEnv, TempoBlockEnv> for TempoTransactionRequest {
             key_authorization,
             valid_before,
             valid_after,
+            fee_payer_signature: _,
```
0xrusowsky, line 124: should we recover the signer here? otherwise `fee_payer` is always `None`

fgimenez: makes sense, without recovering from it we are relying only on the unverified `fee_payer` hint

grandizzy: @0xrusowsky @fgimenez added with https://github.com/tempoxyz/tempo/pull/2283/changes/3b31bd5a8611aa57917a82b636b80d6447e58207 pls recheck

crates/alloy/src/rpc/compat.rs
```diff
@@ -94,6 +100,19 @@ impl TryIntoTxEnv<TempoTxEnv, TempoBlockEnv> for TempoTransactionRequest {
...
+        let fee_payer = if self.fee_payer_signature.is_some() {
+            self.clone()
+                .build_aa()
+                .ok()
+                .and_then(|tx| tx.recover_fee_payer(caller_addr).ok())
+                .map(Some)
+                .or_else(|| self.fee_payer.map(Some))
+        } else {
+            self.fee_payer.map(Some)
+        };
```
0xrusowsky, line 110: if we recover the fee payer form the signature (as i think we should), why do we need `self.fee_payer`?

imo this one should go, as it is useless

0xrusowsky (approved):

PR #2284, opened 2026-01-27
feat: reintroduce background task for commonware gauges

hamdiallam:
Reverted in this 6baec8770c052439769e57c7973c2846b7363929.

Don't see another way to discovery unless at startup, in which any newly created gauges wont be registered

### DIFF
--- a/crates/commonware-node/src/metrics.rs
+++ b/crates/commonware-node/src/metrics.rs
@@ -180,7 +180,7 @@ pub fn install_otlp(context: Context, config: OtlpConfig) -> eyre::Result<OtlpMe
                             counter.add(delta, &attributes);
                         }
                     } else {
-                        // Gauge or histogram bucket - use gauge
+                        // Gauge - use gauge (histograms are handled as counters above)
                         let gauge = {
                             let mut cache = gauges.lock();
                             cache


crates/commonware-node/src/metrics.rs
```diff
@@ -90,58 +97,77 @@ pub fn install_otlp(context: Context, config: OtlpConfig) -> eyre::Result<SdkMet
...
             let encoded = context.encode();
-            // Parse prometheus format and report individual metrics
+
             for line in encoded.lines() {
                 if line.starts_with('#') || line.is_empty() {
                     continue;
                 }
+
                 if let Some((name_labels, value_str)) = line.rsplit_once(' ')
                     && let Ok(value) = value_str.parse::<f64>()
```
SuperFluffy, line 153: With very few exceptions all our gauges and counters are actually integer type (from the top of my head, I'm not sure which ones are float).

I don't know of this matters, just flagging.

crates/commonware-node/src/metrics.rs
```diff
@@ -90,58 +97,77 @@ pub fn install_otlp(context: Context, config: OtlpConfig) -> eyre::Result<SdkMet
...
-        .with_description("Consensus layer metrics from commonware runtime")
-        .with_callback(move |observer| {
+    // Cache for dynamically created gauges - created on first encounter
+    let gauges: Arc<Mutex<HashMap<String, opentelemetry::metrics::Gauge<f64>>>> =
+        Arc::new(Mutex::new(HashMap::new()));
+
+    // Poll at half the export interval to ensure fresh data for each export
+    let poll_interval = interval / 2;
+    let task = tokio::spawn(async move {
+        let mut ticker = tokio::time::interval(poll_interval);
```
SuperFluffy: Let's use commonware-specific ways to handle this in commonware contexts. They provide a sleep method on their runtimes.

crates/commonware-node/src/metrics.rs
```diff
@@ -90,58 +97,77 @@ pub fn install_otlp(context: Context, config: OtlpConfig) -> eyre::Result<SdkMet
...
-        .f64_observable_gauge("consensus_metrics")
-        .with_description("Consensus layer metrics from commonware runtime")
-        .with_callback(move |observer| {
+    // Cache for dynamically created gauges - created on first encounter
+    let gauges: Arc<Mutex<HashMap<String, opentelemetry::metrics::Gauge<f64>>>> =
+        Arc::new(Mutex::new(HashMap::new()));
+
+    // Poll at half the export interval to ensure fresh data for each export
+    let poll_interval = interval / 2;
+    let task = tokio::spawn(async move {
```
SuperFluffy: Unless required, spawn a commonware task from the context.

crates/commonware-node/src/metrics.rs
```diff
@@ -90,58 +97,77 @@ pub fn install_otlp(context: Context, config: OtlpConfig) -> eyre::Result<SdkMet
...
 
     let meter = meter_provider.meter("tempo-consensus");
 
-    // Create an observable gauge that reads from the context on each export
-    let _gauge = meter
-        .f64_observable_gauge("consensus_metrics")
-        .with_description("Consensus layer metrics from commonware runtime")
-        .with_callback(move |observer| {
+    // Cache for dynamically created gauges - created on first encounter
+    let gauges: Arc<Mutex<HashMap<String, opentelemetry::metrics::Gauge<f64>>>> =
```
SuperFluffy, line 127: This basically conflates all metrics into float gauges.

Prometheus requires that e.g. counters are suffixed `_total`. Similar for the other types of metrics.

Can we use that to create map Prometheus metrics you their otel counterparts?

hamdiallam: we could also switch on the # type comment and default to floats for gauges/histograms. Treat counters specially.

Histogram translation feels messy - using a gauge with a bucket label is probably good enough

crates/commonware-node/src/metrics.rs
hamdiallam: I think it might not matter but feels straightforward enough to at least handle counters as integer types and leave the gauges as is.

crates/commonware-node/src/metrics.rs
hamdiallam: Updated to detect all counter types and register them as such

PR #2288, opened 2026-01-28
feat(tempo): user override for consensus storage

SuperFluffy:
Allows users to override the consensus data directory by providing `--consensus.datadir <PATH>`.

### DIFF
--- a/crates/commonware-node/src/args.rs
+++ b/crates/commonware-node/src/args.rs
@@ -140,6 +140,11 @@ pub struct Args {
     /// Cache for the signing key loaded from CLI-provided file.
     #[clap(skip)]
     loaded_signing_key: OnceLock<Option<SigningKey>>,
+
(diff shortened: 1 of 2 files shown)


crates/commonware-node/src/args.rs
```diff
@@ -140,6 +140,11 @@ pub struct Args {
     /// Cache for the signing key loaded from CLI-provided file.
     #[clap(skip)]
     loaded_signing_key: OnceLock<Option<SigningKey>>,
+
+    /// Where to store consensus data. If not set, this will be derived from
+    /// `--datadir`.
+    #[arg(long = "consensus.datadir", value_name = "PATH")]
+    pub storage_dir: Option<PathBuf>,
```
SuperFluffy, line 147: Calling this `storage_dir` because we can't also use `datadir` here. But using `--consensus.datadir` so that it can be found more easily when grepping for the help message for `datadir`.

PR #2290, opened 2026-01-28
fix(txpool): separate intrinsic gas validation for AA and non-AA transactions

rakita:
## Summary
- Gate AA intrinsic gas check to only run for AA transactions (previously ran for all)
- Move non-AA intrinsic gas validation earlier and return `Invalid` outcome instead of `Error` outcome
- Add new `InsufficientGasForIntrinsicCost` error with descriptive gas limit vs intrinsic gas details

## Test plan
- [x] Existing transaction pool tests pass
- [x] Verify AA transactions still get proper intrinsic gas validation
- [x] Verify non-AA transactions return `Invalid` outcome on intrinsic gas failure

### DIFF
--- a/tips/ref-impls/test/invariants/InvariantBaseTest.t.sol
+++ b/tips/ref-impls/test/invariants/InvariantBaseTest.t.sol
@@ -343,6 +343,28 @@ abstract contract InvariantBaseTest is BaseTest {
         vm.writeLine(_logFile, message);
     }
 
+    /// @dev Logs a handler entry to the log file
+    function _logHandlerEntry(string memory handler) internal {
+        vm.writeLine(_logFile, string.concat("CALL: ", handler));
+    }
+
+    /// @dev Logs a skip reason to the log file
+    function _logSkip(string memory reason) internal {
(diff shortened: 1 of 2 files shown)


crates/transaction-pool/src/validator.rs
```diff
@@ -847,7 +852,10 @@ pub fn ensure_intrinsic_gas_tempo_tx(
 
     let gas_limit = tx.gas_limit();
     if gas_limit < gas.initial_gas || gas_limit < gas.floor_gas {
-        Err(InvalidPoolTransactionError::IntrinsicGasTooLow)
+        Err(TempoPoolTransactionError::InsufficientGasForIntrinsicCost {
+            gas_limit,
+            intrinsic_gas: gas.initial_gas,
+        })
```
mattsse: do we need this additional error variant?

this is pretty much the same as `InvalidPoolTransactionError::IntrinsicGasTooLow`

rakita: I guess not, will revert and use InvalidPoolTransactionError::IntrinsicGasTooLow

0xKitsune:
Closes CHAIN-524

Issue #1865, opened 2026-01-06
feat: add periodic spending limits to Access Keys


legion2002:
## Summary

Add support for **periodic spending limits** in key authorization, enabling subscription-based access patterns where spending limits automatically reset after a configurable time period.

## Motivation

Currently, `TokenLimit` in key authorization specifies a one-time spending cap that depletes permanently as tokens are spent. This model works well for single-use or capped authorizations, but doesn't support common subscription-based patterns where:

- A service needs recurring access to a fixed amount per billing cycle
- Users want to authorize "up to X tokens per month" without re-authorizing
- dApps implement subscription models (e.g., streaming services, SaaS payments)

**Real-world use cases:**
1. **Subscription services**: Authorize a streaming service to charge 10 USDC/month
2. **Recurring donations**: Allow an NPO to withdraw up to 5 USDC weekly
3. **Payroll systems**: Enable payroll contracts to transfer salaries monthly
4. **Rate-limited APIs**: Authorize API access keys with per-period token budgets

ERC-20 style approvals cannot fulfill these semantics because they lack periodic limit support. Key authorization with periodic limits provides a native solution.

### Extend `TokenLimit` Structure

**Current:**
```rust
pub struct TokenLimit {
    pub token: Address,
    pub limit: U256,
}
```

**Proposed:**
```rust
pub struct TokenLimit {
    pub token: Address,
    pub limit: U256,
    pub period: Option<u64>,  // Period duration in seconds (None = one-time limit)
}
```

### Update Spending Limit Enforcement

Modify `verify_and_update_spending()` to:
1. Check if the limit is periodic
2. If current timestamp >= `period_end`, reset the period_end:
   - Set `period_end = current_timestamp + period`
   - Set `limit = limit_per_period`
3. Verify and deduct from `limit`

### Period Reset Logic

```
On spending attempt:
  if current_timestamp >= period_end:
      period_end = current_timestamp + period 
      limit = limit_per_period

  if amount > limit:
      revert SpendingLimitExceeded()

  limit -= amount
```

### Interaction with Existing Limits

- Keys can have a mix of one-time and periodic limits for different tokens
- `getRemainingLimit()` returns current remaining (period-aware)
- `updateSpendingLimit()` updates the per-period amount (resets current period)

Issue #1491, opened 2025-12-12
feat: Add destination address scoping to Access Keys


legion2002:
## Summary

Add the ability to restrict Access Keys to specific destination addresses, enabling more granular permission scoping.

## Context

Currently, Access Keys support two configuration options:
- **Spending limits** - per-TIP20 token spending amounts
- **Expiry** - Block timestamp when the key becomes invalid

Users have requested the ability to bind Access Keys to specific destinations (e.g., "only allow transactions to Uniswap"). This would bring Access Keys closer to Solana's delegate primitive while offering more flexibility.

### Phase 1: Address Scoping (target: next hardfork)

Add an **allowed addresses** list to `authorizeKey()` - restrict the key to only call specific contract addresses.

This is relatively straightforward to implement and provides significant value for common use cases.

### Phase 2: Function Selector Scoping (future consideration)

Allow restricting keys to specific function selectors (4-byte selectors). This adds more complexity and should be deferred until we have stronger user validation for the use case.

## Use Cases

- **DeFi integrations**: Allow a trading bot key to only interact with specific DEX contracts
- **Gaming**: Scope a session key to only interact with a game contract  
- **Subscription services**: Allow a key to only call a specific payment contract

## Current Workaround

Deploy a proxy contract that enforces the destination restrictions. This adds gas overhead and complexity.

## References

- [AccountKeychain docs](https://docs.tempo.xyz/protocol/transactions/AccountKeychain)
- [`IAccountKeychain.sol`](docs/specs/src/interfaces/IAccountKeychain.sol)

AmberGOLDExpo:
This fels like a very practical adition, especially for real-world usage of Access Keys.

Address-levl scoping alone already covers a lot of common cases (DEX bots, game session keys, subscriptions), and keeping it simple for the next hardfork makes sense.

From a user perspective, this also improves safety a lot compared to relying only on spend limits + expiry.
Function selector scoping sounds useful too, but I agree its probably better as a later step once address scoping is battle-tested.

PR #2299, opened 2026-01-28
feat(keychain): add periodic limits and call scoping to access keys

legion2002:
## Summary

Implements TIP-1011 (Enhanced Access Key Permissions) which extends Access Keys with two new permission features:

1. **Periodic spending limits** - Automatically reset after a configurable time period, enabling subscription-based access patterns (e.g., "10 USDC/month")
2. **Destination address scoping** - Restricts keys to only interact with specific contract addresses

### Precompile Storage (`account_keychain/mod.rs`)
- Added `spending_limit_max`, `spending_limit_period`, `spending_limit_period_end` mappings for periodic limit data
- Added `allowed_destinations_len` and `allowed_destinations` mappings for destination scoping
- Extended `verify_and_update_spending()` with automatic period reset logic
- Added `set_periodic_limit()`, `set_allowed_destinations()`, `get_allowed_destinations()`, `validate_destination()`, `get_remaining_limit_with_period()` methods

### Handler (`handler.rs`)
- Store periodic limit data when authorizing keys via KeyAuthorization
- Store destination restrictions when authorizing keys
- Validate transaction destinations against allowedDestinations before execution
- Added `DestinationNotAllowed` error type

### Solidity Interface (`IAccountKeychain.sol`)
- Extended `TokenLimit` struct with `remainingInPeriod`, `period`, `periodEnd` fields
- Added `getAllowedDestinations()` view function
- Updated `getRemainingLimit()` to return `(remaining, periodEnd)`
- Added `DestinationNotAllowed` error

### Tests (7 new tests)
- `test_periodic_limit_reset_on_period_expiry`
- `test_periodic_limit_partial_usage_no_rollover`
- `test_destination_scoping_allow_and_deny`
- `test_empty_destinations_allows_any`
- `test_get_allowed_destinations`
- `test_one_time_limit_backwards_compatible`
- `test_main_key_bypasses_destination_restrictions`

## Backward Compatibility

- **No migration required**: Storage is additive - legacy keys (period=0) continue working as one-time limits
- **RLP encoding**: Version-tolerant decoding accepts both V1 (2-field) and V2 (5-field) TokenLimit
- **Empty destinations**: Empty `allowedDestinations` array means unrestricted (can call any address)

## Stacked on
- `tanishk/access-key-upgrade` (TIP-1011 spec + TokenLimit/KeyAuthorization struct changes)

## Related
- TIP-1011 spec: `tips/tip-1011.md`
- #1865 (Periodic spending limits)
- #1491 (Destination address scoping)


crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -430,16 +456,196 @@ impl AccountKeychain {
...
+
+        let mut remaining = if period > 0 {
+            // Periodic limit - check if period has expired and reset if needed
+            let period_end = self.spending_limit_period_end[limit_key][token].read()?;
+            let current_timestamp = self.storage.timestamp().saturating_to::<u64>();
+
+            if current_timestamp >= period_end {
+                // Period expired - reset the remaining amount and update period_end
+                let limit_max = self.spending_limit_max[limit_key][token].read()?;
+                let new_period_end = current_timestamp.saturating_add(period);
```
danrobinson: If the goal is to use this for subscriptions, I think it should either advance by one period, or advance to the next possible periodEnd that is `periodStart + k * period` for some integer k. Otherwise if your transaction is included late (even by a few seconds), the subscription period will begin to drift

tips/tip-1011.md
```diff
@@ -137,38 +162,48 @@ function verifyAndUpdateSpending(key, token, amount):
...
+**Special cases:**
+- Empty calldata (ETH transfer): selector is treated as `bytes4(0)`. A scope with `selector=bytes4(0)` matches ETH transfers.
+- Contract creation (`to=null`): Always rejected if `allowedCalls` is non-empty (cannot whitelist contract creation).
+
 ### Interaction Rules
 
 1. **Mixed limits**: Keys can have a mix of one-time and periodic limits for different tokens
-2. **Destination + limits**: Both constraints are evaluated independently; both must pass
+2. **Call scope + limits**: Both constraints are evaluated independently; both must pass
 3. **Period updates**: Calling `updateSpendingLimit()` updates the per-period amount and resets the current period
```
malleshpai, line 199: i don't see this defined anywhere?

legion2002: `updateSpendingLimit` is defined in the Account Keychain spec [here](https://docs.tempo.xyz/protocol/transactions/AccountKeychain#spending-limit-update). 

It is an artifact of the old spending limit code.

crates/precompiles/src/account_keychain/mod.rs
legion2002: Ah yes, this should be discrete repetitive periods. good catch, will update!

tips/tip-1011.md
```diff
@@ -191,6 +229,11 @@ TokenLimit := RLP([
     period: uint64,
     periodEnd: uint64
```
jxom, line 230: don't think you need to sign over this

```suggestion
```

legion2002 (commented):
🐺 **Ralph Review — PR #2299 (TIP-1011)**

7 findings (5 🚨 Security, 1 ⚠️ Issue, 1 🛡️ Defense-in-Depth). 6 of 7 independently verified with 0 false positives.

Inline comments below cover 6 findings on lines in the diff. One additional finding on pre-existing code:

---

🚨 **[SECURITY] Signed `remainingInPeriod` Ignored — Over-Issues Initial Limit**
**File:** `crates/revm/src/handler.rs:894`
**Severity:** High

The `KeyAuthorization` struct allows users to sign a `remaining_in_period` value for prorated first-period limits (e.g., 50 of 100 tokens). However, handler.rs maps `amount: limit.limit` (the full per-period cap) instead of `amount: limit.remaining_in_period`, granting the full limit and ignoring the user's signed intent.

**Recommended Fix:** Change `amount: limit.limit` → `amount: limit.remaining_in_period` at handler.rs:894.

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -434,16 +460,200 @@ impl AccountKeychain {
...
+                // Period expired - reset the remaining amount and update period_end
+                // Advance period_end by whole multiples of `period` so it lands in
+                // the future without drifting from the original schedule.
+                let limit_max = self.spending_limit_max[limit_key][token].read()?;
+                let elapsed = current_timestamp.saturating_sub(period_end);
+                let periods_elapsed = elapsed / period + 1;
+                let new_period_end = period_end.saturating_add(periods_elapsed.saturating_mul(period));
+
+                self.spending_limit_period_end[limit_key][token].write(new_period_end)?;
+                self.spending_limits[limit_key][token].write(limit_max)?;
```
legion2002: ⚠️ **[ISSUE] `update_spending_limit` Silently Ignored After Periodic Reset**
**Severity:** High

When the period resets here, the remaining limit is overwritten with `spending_limit_max`. However, `update_spending_limit` (line 267) only modifies `spending_limits` without updating `spending_limit_max`. If an account owner calls `updateSpendingLimit` to restrict a compromised periodic key (e.g., from 100 → 50), the restriction is silently reverted to 100 at the next period boundary.

Verified with test: a user updating limit to 50 can still spend 60 after period reset.

**Recommended Fix:** In `update_spending_limit`, also write to `spending_limit_max` when `spending_limit_period > 0`.

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -434,16 +460,200 @@ impl AccountKeychain {
...
+        }
+
+        Ok(())
+    }
+
+    /// TIP-1011: Get allowed call scopes for a key
+    ///
+    /// Returns the list of allowed call scopes for a key.
+    /// Empty array means unrestricted (can call any function on any address).
+    pub fn get_allowed_calls(
```
legion2002: 🛡️ **[DEFENSE-IN-DEPTH] `get_allowed_calls` Returns Data for Revoked/Expired Keys**
**Severity:** Low

This function reads `allowed_calls_len` and returns call scopes without checking if the key has been revoked or expired (same pattern as prior finding SIGP-TMPO2-19 for `getRemainingLimit`). Returns active-looking call scopes for dead keys, creating semantic inconsistency.

**Recommended Fix:** Check key status via `load_active_key` before returning; return empty array for revoked/expired keys.

legion2002:
## 🐺 Ralph Security Review

| Instance | Mode | Status |
|----------|------|--------|
| `pr-2299-w1` | pi/deep+ctx (gemini-3.1-pro-preview) [pr-review/audit-ripple.md] | 🟢 started |
| `pr-2299-w2` | amp/deep+ctx [pr-review/audit-focused.md] | 🟢 started |

**Commit:** `e14b1639ab39`
🔍 **Iteration 1/3** — `pr-2299-w1` (deep) [audit-ripple.md] @ `e14b1639ab39` | [Thread](file:///Users/tanishk/Work/Tempo/ralph-audit/logs/pr-2299/conversations/pi-1771803500-44378.json)
🔍 **Iteration 1/3** — `pr-2299-w2` (deep) [audit-focused.md] @ `e14b1639ab39` | [Thread](https://ampcode.com/threads/T-019c87b7-f2a5-77a9-b066-8806b2aac1f6)

crates/revm/src/handler.rs
```diff
@@ -1011,7 +1052,44 @@ where
...
                     keychain
                         .set_transaction_key(access_key_addr)
-                        .map_err(|e| EVMError::Custom(e.to_string()))
+                        .map_err(|e| TempoInvalidTransaction::KeychainValidationFailed {
+                            reason: e.to_string(),
+                        })?;
+
+                    // TIP-1011: Validate call scopes (address + selector) for all calls
+                    // Skip if authorizing this key in the same tx (call scopes not stored yet)
+                    if !is_authorizing_this_key {
```
legion2002, line 1061: 🚨 **[SECURITY]** Same-Tx Auth+Use Skips TIP-1011 Call Scope Validation

When `is_authorizing_this_key == true`, this gate skips **all** call-scope validation. The access key (which chooses the transaction's calls) can execute arbitrary calls in its bootstrap transaction — even if the root-signed `KeyAuthorization` specified restrictive `allowed_calls`.

The `allowed_calls` are already persisted via `set_allowed_calls()` earlier in this same function, so the inline rationale ("call scopes not stored yet") is inconsistent with the actual code ordering. This breaks TIP-1011's invariant that each top-level call must be validated against `allowedCalls`.

**Recommended Fix:** Do not skip call-scope validation here. Either validate against the `KeyAuthorization.allowed_calls` carried in the transaction directly, or validate against storage (already written above).

crates/revm/src/handler.rs
```diff
@@ -906,18 +906,59 @@ where
...
+                if let Some(limits) = &key_auth.limits {
+                    for limit in limits {
+                        if limit.is_periodic() {
+                            if let Err(err) = keychain.set_periodic_limit(
+                                *root_account,
+                                access_key_addr,
+                                limit.token,
+                                limit.limit,
+                                limit.period,
+                                limit.period_end,
```
legion2002, line 935: 🚨 **[SECURITY]** Arbitrary `period_end` Allows Limit Theft

The user-signed `limit.period_end` is passed directly to `set_periodic_limit()` without validation. A malicious dApp can set `period_end = current_timestamp + 10`, causing the limit to reset in 10 seconds.

Additionally, the handler maps `limit.limit` → `amount` (line 934), ignoring `limit.remaining_in_period`. This grants the key full spending power immediately, regardless of what the user signed. Combined: spend full limit → wait 10s → spend again = 2x extraction.

**Recommended Fix:** For new key authorizations, validate `period_end == 0` (set dynamically on first use) and `remaining_in_period == limit`. Reject deviations.

crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -1,36 +1,271 @@
...
+        if header.payload_length > start_len {
+            return Err(alloy_rlp::Error::InputTooShort);
+        }
+
+        // Decode required fields
+        let token = Address::decode(buf)?;
+        let limit = U256::decode(buf)?;
+
+        let consumed = start_len - buf.len();
+        let remaining_payload = header.payload_length - consumed;
```
legion2002, line 124: ⚠️ **[ISSUE]** Integer Underflow in `TokenLimit` RLP Decoding

`header.payload_length - consumed` underflows if a crafted RLP payload has a list header smaller than the bytes actually consumed. In debug builds this panics the node; in release builds the wrapping value causes the subsequent length check to fail safely.

**Recommended Fix:**
```rust
if consumed > header.payload_length {
    return Err(alloy_rlp::Error::UnexpectedLength);
}
let remaining_payload = header.payload_length - consumed;
```

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -434,16 +460,200 @@ impl AccountKeychain {
...
+        let mut remaining = if period > 0 {
+            // Periodic limit - check if period has expired and reset if needed
+            let period_end = self.spending_limit_period_end[limit_key][token].read()?;
+            let current_timestamp = self.storage.timestamp().saturating_to::<u64>();
+
+            if current_timestamp >= period_end {
+                // Period expired - reset the remaining amount and update period_end
+                // Advance period_end by whole multiples of `period` so it lands in
+                // the future without drifting from the original schedule.
+                let limit_max = self.spending_limit_max[limit_key][token].read()?;
```
legion2002, line 477: ⚠️ **[ISSUE]** Periodic Spending Limits Soft-Brick Keychain Txs in the Pool

When a periodic key's allowance is depleted, `spending_limits` stays at `0` until the next execution-time spend triggers a reset. But the txpool reads `spending_limits` raw and rejects if `fee_cost > remaining`, so the transaction never reaches execution and the reset never fires. This permanently bricks the access key for fee-token payments.

**Recommended Fix:** Update `validate_against_keychain()` in the txpool to emulate the periodic reset: if `period > 0 && current_time >= period_end`, treat effective remaining as `limit_max`.

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -434,16 +460,200 @@ impl AccountKeychain {
...
+        account: Address,
+        key_id: Address,
+        token: Address,
+        limit_max: U256,
+        period: u64,
+        period_end: u64,
+    ) -> Result<()> {
+        let limit_key = Self::spending_limit_key(account, key_id);
+
+        self.spending_limit_max[limit_key][token].write(limit_max)?;
```
legion2002, line 527: ⚠️ **[ISSUE]** `update_spending_limit` Bypassed After Periodic Reset

`update_spending_limit` modifies `spending_limits` (current remaining) but does not update `spending_limit_max`. On the next period reset, `verify_and_update_spending` overwrites remaining with the stale `spending_limit_max`, completely reverting any manual restriction. An account owner lowering a compromised key's allowance has it silently restored at the next period boundary.

**Recommended Fix:** Update `update_spending_limit` to also write to `spending_limit_max` when `spending_limit_period > 0`.

tips/ref-impls/src/interfaces/IAccountKeychain.sol
```diff
@@ -145,7 +159,7 @@ interface IAccountKeychain {
     )
         external
         view
-        returns (uint256);
+        returns (uint256 remaining, uint64 periodEnd);
```
legion2002, line 162: 🛡️ **[DEFENSE-IN-DEPTH]** ABI Mismatch in `getRemainingLimit` Breaks Smart Contract Callers

This Solidity interface now returns `(uint256 remaining, uint64 periodEnd)` (64-byte ABI return), but the Rust precompile's `get_remaining_limit` still returns a single `U256` (32 bytes). Smart contracts compiled against this interface will unconditionally revert due to ABI decode failure.

**Recommended Fix:** Update the `sol!` macro in `account_keychain.rs` and `get_remaining_limit` in `mod.rs` to return `(U256, u64)`.

crates/precompiles/src/account_keychain/mod.rs
legion2002, line 564: 💡 **[SUGGESTION]** `get_allowed_calls` Returns Data for Revoked/Expired Keys

This view function reads `allowed_calls_len` and returns call scopes without checking if the key is revoked or expired — same pattern as prior finding SIGP-TMPO2-19 for spending limits.

**Recommended Fix:** Add a `load_active_key` check before returning; return empty array for inactive keys.

legion2002 (commented):
🐺 **Ralph Review — PR #2299 (TIP-1011: Periodic Limits & Call Scoping)**

This PR introduces significant new surface area for periodic spending limits and call scoping. After consolidating reports from 4 audit workers (10 reports, 8 verifications), **7 actionable findings survived** deduplication and noise filtering.

| # | Tier | Finding | File |
|---|------|---------|------|
| 1 | 🚨 SECURITY | Same-tx auth+use skips call scope validation | `handler.rs` |
| 2 | 🚨 SECURITY | Arbitrary `period_end` allows limit theft | `handler.rs` |
| 3 | ⚠️ ISSUE | Integer underflow in `TokenLimit` RLP decoding | `key_authorization.rs` |
| 4 | ⚠️ ISSUE | Periodic limits soft-brick keychain txs in pool | `account_keychain/mod.rs` |
| 5 | ⚠️ ISSUE | `update_spending_limit` bypassed after periodic reset | `account_keychain/mod.rs` |
| 6 | 🛡️ DEFENSE | ABI mismatch in `getRemainingLimit` breaks callers | `IAccountKeychain.sol` |
| 7 | 💡 SUGGESTION | `get_allowed_calls` returns data for revoked keys | `account_keychain/mod.rs` |

**Reviewer Callouts:**
- ⚡ **Intrinsic Gas**: `calculate_key_authorization_gas` charges 0 gas for `allowed_calls`. Large payloads enter the chain at ~98% discount.
- ⚡ **Pool/Execution Parity**: Txpool does not enforce TIP-1011 call scopes — nodes gossip deterministically invalid txs.

PR #2301, opened 2026-01-28
test(invariants): improve `TempoTransactionInvariants`

0xrusowsky:
This PR fixes T1 hardfork gas repricing for CREATE handlers:
- Add gas constants to TxBuilder (TIP-1000: ACCOUNT_CREATION_COST, CREATE_BASE_COST, etc.)
- Add gas calculation helpers: calldataGas(), initcodeGas(), createGas()
- Fix handler_createGasScaling and other CREATE handlers to use correct T1 gas
- Fix ghost state sync in catch blocks (3 handlers were missing sync)
- Add GAS_LIMIT_BUFFER constant for execution safety margin
- Remove hardcoded numbers from comments, use constant names

It also migrates invariant test suite to use ghost state as source of truth for
nonce values, with catch-block sync for divergence recovery.

### DIFF
--- a/tips/ref-impls/test/helpers/TxBuilder.sol
+++ b/tips/ref-impls/test/helpers/TxBuilder.sol
@@ -94,13 +94,17 @@ library TxBuilder {
     }
 
     /// @notice Estimate gas for multicall (TIP-1000)
-    /// @param numCalls Number of calls in the batch
+    /// @param calls The calls in the batch
     /// @param nonce The sender's current nonce (0 = first tx, account creation cost applies)
-    function multicallGas(uint64 numCalls, uint64 nonce) internal pure returns (uint64) {
-        uint64 gas = DEFAULT_GAS_LIMIT * numCalls;
+    function multicallGas(TempoCall[] memory calls, uint64 nonce) internal pure returns (uint64) {
+        uint64 gas = TX_BASE_COST;
 
         if (nonce == 0) {
             gas += ACCOUNT_CREATION_COST;
+        }
+
+        for (uint256 i = 0; i < calls.length; i++) {
+            gas += calldataGas(calls[i].data);
         }
 
         return gas;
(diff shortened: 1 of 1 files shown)


tips/ref-impls/test/helpers/GhostState.sol
```diff
@@ -1,10 +1,45 @@
...
 /// @title GhostState - Ghost Variable Tracking for Invariant Tests
 /// @dev Ghost variables mirror what we expect on-chain state to be
 abstract contract GhostState {
 
+    // ============ File-based Logging (persists across Foundry state resets) ============
+
+    string internal constant TEMPO_TX_LOG = "tempo_tx_invariant.log";
+
+    /// @dev Foundry VM interface for file operations (uses standard cheatcode address)
+    Vm private constant _ghostVm = Vm(address(uint160(uint256(keccak256("hevm cheat code")))));
```
grandizzy: not sure why this is needed and not using vm. directly?

tips/ref-impls/test/helpers/GhostState.sol
```diff
@@ -1,10 +1,45 @@
...
+
+    /// @dev Initialize the log file (call in setUp)
+    function _initTxLog() internal {
+        try _ghostVm.removeFile(TEMPO_TX_LOG) { } catch { }
+        _ghostVm.writeLine(TEMPO_TX_LOG, "=== TempoTransaction Invariant Log ===");
+    }
+
+    /// @dev Log a handler skip with reason
+    function _logSkip(string memory handler, string memory reason) internal {
```
grandizzy: if I am not wrong these were used to instrument and to figure out ways of making invariants more efficient, do we still need them?

tips/ref-impls/test/helpers/InvariantChecker.sol
```diff
@@ -201,6 +202,36 @@ abstract contract InvariantChecker is HandlerBase {
         }
     }
 
+    // ============ Skip Stats Reporting ============
+
+    /// @notice Report handler skip statistics for debugging slow invariant runs
+    /// @dev Call this in afterInvariant() to see why handlers are being skipped
```
grandizzy: same as above, can be removed?

tips/ref-impls/test/helpers/GhostState.sol
zerosnacks: +1, yeah that should not be necessary

PR #2302, opened 2026-01-28
fix(txpool): validate AA transaction calls

rakita:
## Summary
- Reject AA transactions with no calls (`NoCalls` error)
- Reject AA transactions where CREATE calls are not the first call (`CreateCallNotFirst` error)
- Both errors are marked as "bad transactions" (permanently invalid)

## Test plan
- [x] Added `test_aa_no_calls_rejected` - verifies transactions with no calls are rejected
- [x] Added `test_aa_create_call_not_first_rejected` - verifies CREATE calls not in first position are rejected
- [x] Added `test_aa_create_call_first_accepted` - verifies CREATE calls in first position are accepted
- [x] All 156 transaction-pool tests pass

### DIFF
--- a/crates/transaction-pool/src/validator.rs
+++ b/crates/transaction-pool/src/validator.rs
@@ -446,7 +446,7 @@ where
 
         let tx = aa_tx.tx();
 
-        if tx.calls.len() == 0 {
+        if tx.calls.is_empty() {
             return Err(TempoPoolTransactionError::NoCalls);
         }
 


crates/transaction-pool/src/validator.rs
```diff
@@ -2828,6 +2836,133 @@ mod tests {
...
+                );
+            }
+            _ => panic!("Expected Invalid outcome with NoCalls error, got: {outcome:?}"),
+        }
+    }
+
+    /// CREATE calls (contract deployments) must be the first call in an AA transaction.
+    #[tokio::test]
+    async fn test_aa_create_call_not_first_rejected() {
+        use tempo_primitives::transaction::tempo_transaction::Call;
```
0xKitsune: nit: this is imported in multiple functions, can we move this import to the top of the tests mod?

0xKitsune (approved):

PR #2312, opened 2026-01-28
tip: TIP-XXXX Currency Spending Limits for Access Keys

howydev:
Adds TIP-XXXX documenting the currency spending limits feature for AccountKeychain access keys.

## Summary

This TIP extends AccountKeychain to support per-currency spending limits in addition to per-token limits. Currency limits allow a single spending cap (e.g., 1000 USD) to apply across all TIP-20 tokens that share the same currency code.

## Key Changes

- New `CurrencyLimit` struct with `amount` + `currency` (string like "USD", "EUR")
- `authorizeKey()` signature updated: `limits` → `tokenLimits` + `currencyLimits`
- New `updateCurrencyLimit()` and `getRemainingCurrencyLimit()` functions
- `KeyInfo` includes `hasTokenLimits` + `hasCurrencyLimits` flags
- Currency limits apply across ALL tokens sharing the same currency

## Related

- Implementation: #2117

### DIFF
--- a/tips/TIP-XXXX-currency-limits.md
+++ b/tips/TIP-XXXX-currency-limits.md
@@ -1,0 +1,215 @@
+---
+id: TIP-XXXX
+title: Currency Spending Limits for Access Keys
+description: Extends AccountKeychain to support per-currency spending limits that apply across multiple tokens sharing the same currency
+authors: howydev
+status: Draft
+related: TIP-403
+protocolVersion: TBD
+searchable: false
+---
+
+# TIP-XXXX: Currency Spending Limits for Access Keys
+
+## Abstract
+
+This TIP extends the AccountKeychain precompile to support per-currency spending limits in addition to per-token limits. Currency limits allow a single spending cap (e.g., 1000 USD) to apply across all TIP-20 tokens that share the same currency code, enabling more flexible and intuitive access key permissions.
+
+## Motivation
+
+The current AccountKeychain implementation only supports per-token spending limits. This creates friction for users who want to limit total spending in a currency (e.g., "max $1000 USD per day") without having to enumerate every possible stablecoin.
+
+For example, a user provisioning an access key for a mobile wallet may want to limit USD spending to $500, regardless of whether the app uses USDC, USDT, or any other USD-denominated token. Without currency limits, the user must set individual limits on each token, and the app must predict which USD tokens the user will hold—any tokens not explicitly listed would be unspendable by default.
+
+Currency limits solve this by leveraging the TIP-20 `currency()` method to group tokens. When a transfer occurs, the system checks the token's currency and deducts from the corresponding currency limit.
+
+**Design considerations:**
+- Currency limits are checked *in addition to* token limits (both must pass if set)
+- A key can have token limits only, currency limits only, or both
+- Missing currency limits (zero value) are treated as unlimited for that currency
+
+---
+
+# Specification
+
+## Data Structures
+
+### CurrencyLimit
+
+```solidity
+/// @notice Currency spending limit structure
+struct CurrencyLimit {
+    uint256 amount;   // Spending limit amount
+    string currency;  // Currency code (e.g., "USD", "EUR", "GBP")
+}
+```
+
+### Updated KeyInfo
+
+```solidity
+/// @notice Key information structure
+struct KeyInfo {
+    SignatureType signatureType;  // Signature type of the key
+    address keyId;                // The key identifier (address)
+    uint64 expiry;                // Unix timestamp when key expires
+    bool enforceLimits;           // Master switch for spending limits
+    bool hasTokenLimits;          // Whether per-token limits exist
+    bool hasCurrencyLimits;       // Whether per-currency limits exist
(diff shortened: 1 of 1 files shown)


tips/TIP-XXXX-currency-limits.md
```diff
@@ -0,0 +1,217 @@
...
+
+```solidity
+contract AccountKeychain {
+    // slot 0: existing keys mapping
+    mapping(address => mapping(address => AuthorizedKey)) private keys;
+    
+    // slot 1: existing token spending limits
+    mapping(bytes32 => mapping(address => uint256)) private _spendingLimits;
+    
+    // slot 2: NEW currency spending limits
```
howydev:
```suggestion
    // NEW - slot 2: currency spending limits
```

tips/TIP-XXXX-currency-limits.md
```diff
@@ -0,0 +1,217 @@
...
+
+When `authorizeTransfer` is called, the following checks occur:
+
+1. **Root key bypass:** If `keyId == address(0)`, no limits are checked (root key has unlimited access)
+
+2. **Passthrough keys:** If `enforceLimits == false`, no limits are checked
+
+3. **Token limit check:** If `hasTokenLimits == true`:
+   - Read remaining limit for the specific token
+   - If `amount > remaining`, revert with `SpendingLimitExceeded`
```
howydev: instead: zero out `remaining` and pass `amount - remaining` to be checked by currency spend limit

tips/TIP-XXXX-currency-limits.md
```diff
@@ -0,0 +1,217 @@
...
+}
+```
+
+# Invariants
+
+1. **Currency limits are additive with token limits:** If both `hasTokenLimits` and `hasCurrencyLimits` are true, both checks must pass for a transfer to succeed.
+
+2. **Currency deduction is shared across tokens:** If USDC and USDT both return `"USD"` from `currency()`, spending either token deducts from the same USD limit.
+
+3. **Zero remaining limit blocks spending:** A currency limit of 0 remaining means no further spending is allowed for tokens of that currency.
```
howydev: no, because it can have a separate spend limit. suggest a better invariant

tips/TIP-XXXX-currency-limits.md
```diff
@@ -0,0 +1,217 @@
...
+
+4. **Missing currency = unlimited:** If a currency limit was never set for a currency code, `getRemainingCurrencyLimit` returns 0, which is treated as "no limit set" (unlimited).
```
howydev: no, if a currency limit was never set the `hasCurrencyLimit` flag would be false

if the flag is true then theres a limit, and 0 is a valid remaining limit

tips/TIP-XXXX-currency-limits.md
```diff
@@ -0,0 +1,217 @@
...
+
+5. **Updating limits enables enforcement:** Calling `updateCurrencyLimit` on a key with `enforceLimits=false` automatically sets `enforceLimits=true` and `hasCurrencyLimits=true`.
```
howydev: it should only set this to true if that specific spend limit is being updated

tips/TIP-XXXX-currency-limits.md
```diff
@@ -0,0 +1,215 @@
...
+
+## Abstract
+
+This TIP extends the AccountKeychain precompile to support per-currency spending limits in addition to per-token limits. Currency limits allow a single spending cap (e.g., 1000 USD) to apply across all TIP-20 tokens that share the same currency code, enabling more flexible and intuitive access key permissions.
+
+## Motivation
+
+The current AccountKeychain implementation only supports per-token spending limits. This creates friction for users who want to limit total spending in a currency (e.g., "max $1000 USD per day") without having to enumerate every possible stablecoin.
+
+For example, a user provisioning an access key for a mobile wallet may want to limit USD spending to $500, regardless of whether the app uses USDC, USDT, or any other USD-denominated token. Without currency limits, the user must set individual limits on each token, and the app must predict which USD tokens the user will hold—any tokens not explicitly listed would be unlimited by default.
```
howydev: *would be unspendable by default

tips/TIP-XXXX-currency-limits.md
```diff
@@ -0,0 +1,215 @@
...
+}
+```
+
+# Invariants
+
+1. **Token limits are checked first, overflow goes to currency:** If both `hasTokenLimits` and `hasCurrencyLimits` are true, token limits are deducted first. Any amount exceeding the token limit overflows to the currency limit check.
+
+2. **Currency deduction is shared across tokens:** If USDC and USDT both return `"USD"` from `currency()`, spending either token deducts from the same USD limit.
+
+3. **Limit flags determine enforcement, not storage values:** If `hasCurrencyLimits == false`, currency limits are not checked regardless of storage values. If `hasCurrencyLimits == true`, a remaining value of 0 means the limit is exhausted (not unlimited).
```
howydev, line 200: suggest something higher level, this is a bit too concrete

compare that against this suggestion: "if a key has x token spend limit and y currency spend limit for that token's currency, the key can only spend x+y tokens max"

PR #2316, opened 2026-01-29
tip: add TIP-1015 Compound Transfer Policies

gakonst:
## Summary

Adds TIP-1015, which extends TIP-403 with compound policies that specify different authorization rules for senders, recipients, and mint recipients.

## Motivation

Real-world requirements often differ between sending, receiving, and minting tokens. For example:

- **Vendor credits**: Credits that can be minted to anyone and spent to a specific vendor, but cannot be transferred peer-to-peer
- **Sender restrictions**: Block sanctioned addresses from sending, while allowing anyone to receive (for refunds or seizure)
- **Recipient restrictions**: Require KYC for recipients, while allowing any holder to send

## Changes

A compound policy references three simple policies:
- **senderPolicyId**: checked for transfer senders
- **recipientPolicyId**: checked for transfer recipients
- **mintRecipientPolicyId**: checked for mint recipients

### TIP-403 Registry

- New `PolicyType.COMPOUND` enum value
- `createCompoundPolicy(senderPolicyId, recipientPolicyId, mintRecipientPolicyId)`
- `isAuthorizedSender(policyId, user)`
- `isAuthorizedRecipient(policyId, user)`
- `isAuthorizedMintRecipient(policyId, user)`
- `isAuthorized()` returns `isAuthorizedSender() && isAuthorizedRecipient()`
- Compound policies are **immutable** once created

### Required Code Changes

6 direct replacements of `isAuthorized` calls:

| Location | Current | Replace With |
|----------|---------|--------------|
| TIP-20 `_mint` | `isAuthorized(to)` | `isAuthorizedMintRecipient(to)` |
| TIP-20 `burnBlocked` | `isAuthorized(from)` | `isAuthorizedSender(from)` |
| TIP-20 `isTransferAuthorized` | `isAuthorized(from)` | `isAuthorizedSender(from)` |
| TIP-20 `isTransferAuthorized` | `isAuthorized(to)` | `isAuthorizedRecipient(to)` |
| DEX `cancelStaleOrder` | `isAuthorized(maker)` | `isAuthorizedSender(maker)` |
| Fee payer `can_fee_payer_transfer` | `isAuthorized(fee_payer)` | `isAuthorizedSender(fee_payer)` |

All other call sites use `ensureTransferAuthorized(from, to)` which delegates to `isTransferAuthorized`, so they automatically inherit the correct behavior.

### DIFF
--- a/tips/tip-1015.md
+++ b/tips/tip-1015.md
@@ -182,50 +182,25 @@ interface ITIP403Registry {
     /// @notice The referenced policy is not a simple policy
     error PolicyNotSimple();
 
-    /// @notice The referenced policy does not exist (reuses TIP-403 error)
+    /// @notice The referenced policy does not exist
     error PolicyNotFound();
 }
 ```
 
 ## Authorization Logic
 
-### Internal Helper
-
-The existing `isAuthorized` logic for simple policies is preserved as an internal helper:
-
-```solidity
-function _isAuthorizedSimple(uint64 policyId, address user) internal view returns (bool) {
-    // Handle built-in policies
-    if (policyId < 2) {
-        return policyId == 1; // 0 = always-reject, 1 = always-allow
-    }
-    
-    PolicyData memory data = policyData[policyId];
-    
-    if (data.policyType == PolicyType.WHITELIST) {
-        return policySet[policyId][user];
-    } else {
-        return !policySet[policyId][user];
-    }
-}
-```
-
 ### isAuthorizedSender
 
 ```solidity
 function isAuthorizedSender(uint64 policyId, address user) external view returns (bool) {
-    // Handle built-in policies
-    if (policyId < 2) {
-        return policyId == 1;
-    }
-    
     PolicyRecord storage record = policyRecords[policyId];
-    
+
     if (record.base.policyType == PolicyType.COMPOUND) {
-        return _isAuthorizedSimple(record.compound.senderPolicyId, user);
+        return isAuthorized(record.compound.senderPolicyId, user);
     }
-    
-    return _isAuthorizedSimple(policyId, user);
+
+    // For simple policies, sender authorization equals general authorization
+    return isAuthorized(policyId, user);
 }
 ```
 
@@ -233,18 +208,14 @@ function isAuthorizedSender(uint64 policyId, address user) external view returns
 
 ```solidity
 function isAuthorizedRecipient(uint64 policyId, address user) external view returns (bool) {
-    // Handle built-in policies
-    if (policyId < 2) {
-        return policyId == 1;
-    }
-    
     PolicyRecord storage record = policyRecords[policyId];
-    
+
     if (record.base.policyType == PolicyType.COMPOUND) {
-        return _isAuthorizedSimple(record.compound.recipientPolicyId, user);
+        return isAuthorized(record.compound.recipientPolicyId, user);
     }
-    
-    return _isAuthorizedSimple(policyId, user);
+
+    // For simple policies, recipient authorization equals general authorization
+    return isAuthorized(policyId, user);
 }
 ```
 
@@ -252,18 +223,14 @@ function isAuthorizedRecipient(uint64 policyId, address user) external view retu
 
 ```solidity
 function isAuthorizedMintRecipient(uint64 policyId, address user) external view returns (bool) {
-    // Handle built-in policies
-    if (policyId < 2) {
-        return policyId == 1;
-    }
-    
     PolicyRecord storage record = policyRecords[policyId];
-    
+
     if (record.base.policyType == PolicyType.COMPOUND) {
-        return _isAuthorizedSimple(record.compound.mintRecipientPolicyId, user);
+        return isAuthorized(record.compound.mintRecipientPolicyId, user);
     }
-    
-    return _isAuthorizedSimple(policyId, user);
+
+    // For simple policies, mint recipient authorization equals general authorization
+    return isAuthorized(policyId, user);
 }
 ```
 


tips/tip-XXXX.md
```diff
@@ -0,0 +1,350 @@
+---
+id: TIP-XXXX
+title: Compound Transfer Policies
+description: Extends TIP-403 with compound policies that specify different authorization rules for senders and recipients.
+authors: Dan Robinson @danrobinson
```
danrobinson:
```suggestion
authors: Dan Robinson
```

tips/tip-XXXX.md
```diff
@@ -0,0 +1,334 @@
...
+    }
+    // ... burn logic
+}
+```
+
+## Stablecoin DEX Integration
+
+### Cancel Stale Order
+
+The `cancelStaleOrder` function checks sender authorization since the maker escrowed (sent) the token:
```
danrobinson:
```suggestion
The `cancelStaleOrder` function checks sender authorization on the token escrowed by the maker, since if the order is filled, the maker will have to send that token:
```

tips/tip-XXXX.md
```diff
@@ -0,0 +1,334 @@
...
+        revert OrderNotStale();
+    }
+    
+    _cancelOrder(order);
+}
+```
+
+## Immutability
+
+Compound policies are immutable once created. There is no `modifyCompoundPolicy` function. To change policy behavior, token issuers must:
```
danrobinson:
```suggestion
Compound policies are immutable once created. To change policy behavior, token issuers must:
```

tips/tip-XXXX.md
```diff
@@ -0,0 +1,334 @@
...
+
+1. Create a new compound policy with the desired configuration
+2. Update the token's `transferPolicyId` to the new policy
+
+This design:
```
danrobinson: I think just remove this paragraph

tips/tip-XXXX.md
```diff
@@ -0,0 +1,334 @@
...
+
+4. **Sender/recipient differentiation**: Verify that a compound policy with different sender/recipient policies correctly authorizes asymmetric transfers.
+
+5. **isAuthorized behavior**: Verify that `isAuthorized` on a compound policy returns `isAuthorizedSender() && isAuthorizedRecipient()`.
+
+6. **TIP-20 mint**: Verify that mints only check recipient authorization.
+
+7. **TIP-20 burnBlocked**: Verify that burnBlocked checks sender authorization (and allows burning from blocked senders).
+
+8. **Immutability**: Verify that there is no way to modify a compound policy after creation.
```
danrobinson: This isn’t a real check; can remove it

tips/tip-1015.md
```diff
@@ -0,0 +1,360 @@
...
+    /// For compound policies: checks against the recipientPolicyId
+    function isAuthorizedRecipient(uint64 policyId, address user) external view returns (bool);
+
+    /// @notice Checks if a user is authorized as a mint recipient under the given policy
+    /// @param policyId Policy ID to check against
+    /// @param user Address to check
+    /// @return True if authorized to receive mints, false otherwise
+    /// @dev For simple policies: equivalent to isAuthorized()
+    /// For compound policies: checks against the mintRecipientPolicyId
+    function isAuthorizedMintRecipient(uint64 policyId, address user) external view returns (bool);
```
0xKitsune, line 142: Is there an instance where a user would be authorized as a mint recipient but not a transfer recipient? Otherwise would it be possible to consolidate this into a single  function?

gakonst:
Re: @0xKitsune's question about consolidating mintRecipient and recipient:

Yes! The vendor credits use case in the Motivation section is exactly this scenario:

- `mintRecipientPolicyId = 1` (always-allow) - anyone can receive mints
- `recipientPolicyId = vendor_whitelist` - only vendors can receive transfers

This allows a business to mint credits to any customer, but customers can only spend those credits at the vendor (no peer-to-peer transfers).

See `test_vendor_credits_use_case` in PR #2344 for a concrete example.

tips/tip-1015.md
```diff
@@ -0,0 +1,360 @@
...
+
+    // =========================================================================
+    //                      Errors
+    // =========================================================================
+
+    /// @notice The referenced policy is not a simple policy
+    error PolicyNotSimple();
+
+    /// @notice The referenced policy does not exist
+    error PolicyDoesNotExist();
```
0xrusowsky: isn't this error code equivalent to the existing `error PolicyNotFound()`?

tips/tip-1015.md
```diff
@@ -0,0 +1,360 @@
...
+
+### isAuthorized (updated)
+
+The existing `isAuthorized` function is updated to check both sender and recipient authorization:
+
+```solidity
+function isAuthorized(uint64 policyId, address user) external view returns (bool) {
+    return isAuthorizedSender(policyId, user) && isAuthorizedRecipient(policyId, user);
+}
+```
```
0xrusowsky, line 245: although i understand the design and it is sound. unless i am missing something, the code proposal would create an infinite loop because:

- the spec redefines `isAuthorized()` to delegate to the new functions:
```
  function isAuthorized(uint64 policyId, address user) external view returns (bool) {
      return isAuthorizedSender(policyId, user) && isAuthorizedRecipient(policyId, user);
  }
```

- these functions delegate back to `isAuthorized()` for simple policies:
```
  // For simple policies, sender authorization equals general authorization
  return isAuthorized(policyId, user);
```

suggested fix: the existing `isAuthorized()` implementation (which handles whitelist/blacklist/built-in logic) should be preserved and simply renamed to an internal helper like `_isAuthorizedSimple()`.

`isAuthorizedSender()`, `isAuthorizedRecipient()`, `isAuthorizedMintRecipient()` functions should call the internal helper instead.

gakonst: Addressed - the spec now uses `_isAuthorizedSimple()` as an internal helper exactly as you suggested. The authorization functions call `_isAuthorizedSimple()` instead of `isAuthorized()`, breaking the loop.

tips/tip-1015.md
gakonst: Fixed - renamed to `PolicyNotFound()` for consistency with TIP-403.

PR #2325, opened 2026-01-29
fix: only apply TIP-1000 gas metering for T1 hardfork

legion2002:
## Summary

Fixes snapshot re-execution failures on the moderato chain (Genesis hardfork) caused by TIP-1000 gas repricing changes.

## Root Causes

Three related issues were identified:

### 1. Gas-metered authorize_key on pre-T1
TIP-1000 introduced `EvmPrecompileStorageProvider::new_with_gas_limit()` for `authorize_key` calls, but this was applied unconditionally. For pre-T1 chains, transactions would run out of gas during authorization, leaving keys unwritten.

**Fix:** Use unlimited gas (`new_max_gas`) for pre-T1 chains.

### 2. Double-validation for AA transactions
The outer `validate_initial_tx_gas` was re-validating with inflated gas after `validate_aa_initial_tx_gas` had added `nonce_2d_gas` post-validation.

**Fix:** Skip outer validation for AA transactions since they validate internally.

### 3. Wrong gas source and OOG check for execution on pre-T0
- TIP-1000 changed `execution()` to use `evm.initial_gas` instead of `init_and_floor_gas`
- For Genesis, `nonce_2d_gas` is added to `init_and_floor_gas` after validation but NOT to `evm.initial_gas`
- The OOG check (`gas_limit < initial_gas`) was unconditional, but the original code only checked this for T0+

**Fix:** Use `init_and_floor_gas` for pre-T1 execution and only do the OOG check for T0+.

## Verification

Re-execution workflow `re-execute-gsf6r` on moderato chain completed successfully with image `sha-e39ef4a`.

## Test Plan
- [x] All tempo-revm tests pass (98 tests)
- [x] Snapshot re-execution passes on moderato (Genesis hardfork)

### DIFF
--- a/crates/revm/src/handler.rs
+++ b/crates/revm/src/handler.rs
@@ -1166,11 +1166,10 @@ where
         let gas_params = evm.ctx_ref().cfg().gas_params();
         let gas_limit = tx.gas_limit();
 
-        // Route to appropriate gas calculation based on transaction type
-        // AA transactions do their own validation inside validate_aa_initial_tx_gas
-        let (mut init_gas, is_aa_tx) = if tx.tempo_tx_env.is_some() {
+        // Route to appropriate gas calculation and validation based on transaction type
+        let init_gas = if tx.tempo_tx_env.is_some() {
             // AA transaction - use batch gas calculation (includes validation)
-            (validate_aa_initial_tx_gas(evm)?, true)
+            validate_aa_initial_tx_gas(evm)?
         } else {
             let mut acc = 0;
             let mut storage = 0;
@@ -1207,12 +1206,6 @@ where
                 init_gas.initial_gas += gas_params.get(GasId::new_account_cost());
             }
 
-            (init_gas, false)
-        };
-
-        // Skip validation for AA transactions - they already validated in validate_aa_initial_tx_gas
-        // (AA validation happens before adding post-validation gas like 2D nonce costs)
-        if !is_aa_tx {
             if evm.ctx.cfg.is_eip7623_disabled() {
                 init_gas.floor_gas = 0u64;
             }
@@ -1234,7 +1227,9 @@ where
                 }
                 .into());
             }
-        }
+
+            init_gas
+        };
 
         // used to calculate key_authorization gas spending limit.
         evm.initial_gas = init_gas.initial_gas;


crates/revm/src/handler.rs
```diff
@@ -491,12 +491,22 @@ where
...
+        // For T1+: Use evm.initial_gas which includes key_authorization gas tracking
+        // For pre-T1: Use init_and_floor_gas.initial_gas directly to maintain backward compatibility
+        // Pre-T1 doesn't have key_authorization gas tracking, and Genesis has special handling
+        // where nonce_2d_gas is added to init_and_floor_gas but not to evm.initial_gas
+        let spec = evm.ctx_ref().cfg().spec();
+        let adjusted_gas = if spec.is_t1() {
+            InitialAndFloorGas::new(evm.initial_gas, init_and_floor_gas.floor_gas)
+        } else {
+            *init_and_floor_gas
+        };
```
0xrusowsky: we have this pattern twice, i feel like it would be cleaner if we added a `fn adjusted()` to `InitialAndFloorGas` which hides the complexity

crates/revm/src/handler.rs
```diff
@@ -491,12 +491,22 @@ where
...
 
         let tx = evm.tx();
-        if tx.gas_limit() < adjusted_gas.initial_gas {
+        // For T0+, check gas limit covers intrinsic gas (including 2D nonce for AA txs)
+        // For pre-T0 (Genesis), skip this check to maintain backward compatibility -
+        // the original behavior didn't validate gas limit for 2D nonce cost upfront
+        if spec.is_t0() && tx.gas_limit() < adjusted_gas.initial_gas {
```
0xrusowsky, line 520: maybe i'm confused, but shouldn't this one be T1?

crates/revm/src/handler.rs
```diff
@@ -787,11 +797,17 @@ where
...
+            // For pre-T1 chains, use unlimited gas to maintain backward compatibility.
+            let mut provider = if spec.is_t1() {
+                EvmPrecompileStorageProvider::new_with_gas_limit(
+                    internals,
+                    cfg,
+                    tx.gas_limit() - initial_gas,
+                )
+            } else {
+                EvmPrecompileStorageProvider::new_max_gas(internals, cfg)
+            };
```
0xrusowsky, line 821: it is fine like this, but i want to do a follow-up commit to move this logic into `EvmPrecompileStorageProvider::enter_evm()`, as otherwise we still have dangling usages of `EvmPrecompileStorageProvider::new_max_gas` in T1, which is weird

crates/revm/src/handler.rs
```diff
@@ -1179,29 +1196,33 @@ where
...
-            return Err(TempoInvalidTransaction::InsufficientGasForIntrinsicCost {
-                gas_limit,
-                intrinsic_gas: init_gas.floor_gas,
+            // Validate floor gas (Prague+)
+            if !evm.ctx.cfg.is_eip7623_disabled() && gas_limit < init_gas.floor_gas {
+                return Err(TempoInvalidTransaction::InsufficientGasForIntrinsicCost {
+                    gas_limit,
+                    intrinsic_gas: init_gas.floor_gas,
+                }
+                .into());
```
0xrusowsky: nit: this could be moved inside the `else` block

crates/revm/src/handler.rs
legion2002: Response from amp
```
The current code is correct with `is_t0()`. These are two separate concerns:

1. **`adjusted_gas` selection (uses `is_t1()`)**: Decides which gas source to use:
   - T1+: `evm.initial_gas` (includes key_authorization gas tracking)
   - Pre-T1: `init_and_floor_gas.initial_gas` directly

2. **Gas limit validation (uses `is_t0()`)**: Decides whether to do upfront gas limit check for 2D nonce cost:
   - Genesis (pre-T0): Skips this check (original behavior didn't validate gas limit for 2D nonce cost upfront)
   - T0+: Performs the check

Since T1 is after T0, `is_t0()` returns `true` for T1 as well, so T1 transactions still get this validation.
```

crates/revm/src/handler.rs
```diff
@@ -1499,12 +1530,13 @@ where
...
-        // Use saturating_add to avoid overflow when additional_initial_gas is u64::MAX (OOG case)
-        let adjusted_gas = InitialAndFloorGas::new(evm.initial_gas, init_and_floor_gas.floor_gas);
+        let spec = evm.ctx_ref().cfg().spec();
+        let adjusted_gas = adjusted_initial_gas(*spec, evm.initial_gas, init_and_floor_gas);
 
         let tx = evm.tx();
-        if tx.gas_limit() < adjusted_gas.initial_gas {
+        // For T0+, check gas limit covers intrinsic gas (including 2D nonce for AA txs)
+        // For pre-T0 (Genesis), skip this check to maintain backward compatibility
+        if spec.is_t0() && tx.gas_limit() < adjusted_gas.initial_gas {
```
0xKitsune, line 1539: This is the exact same logic in the `execution()` fn. We should consider refactoring so that we do not need to remember to change this in two places. This can be done in a follow up.

crates/revm/src/handler.rs
rakita: This looks like a main bug. 

Would clean it up as 


```suggestion
 
            // TIP-1000: Only apply gas metering for T1 hardfork.
            // For pre-T1 chains, use unlimited gas to maintain backward compatibility.
            let provider_gas_limit = if spec.is_t1() {
                    tx.gas_limit() - initial_gas
            } else {
                u64::MAX
            };
            let mut provider = EvmPrecompileStorageProvider::new_with_gas_limit(
                    internals,
                    cfg,
                    provider_gas_limit,
                );
```

crates/revm/src/handler.rs
```diff
@@ -1503,12 +1530,13 @@ where
...
```
rakita, line 1539: Yeah t0 check was missing here

PR #2329, opened 2026-01-29
fix(t1): checkpoint revert on key_authorization oog

rakita:
Closes CHAIN-540

## Summary
- Fix bug where `key_authorization` storage changes were not reverted on out-of-gas (OOG)
- Add checkpoint/revert mechanism for T1 hardfork key authorization processing
- Add comprehensive test for key authorization behavior with insufficient/sufficient gas

## Problem
When a transaction ran out of gas during T1 hardfork processing, the `key_authorization` precompile storage changes were persisted even though the transaction failed. This led to storage inconsistency where keys could be authorized despite the transaction failing.

## Solution
Create a journal checkpoint before key_authorization precompile execution and:
- **On OOG**: Revert the checkpoint to undo storage changes
- **On success**: Commit the checkpoint to persist changes

## Test plan
- [x] Added `test_key_authorization_t1` test that verifies:
  - Keys are NOT authorized when transaction fails due to insufficient gas
  - Keys ARE authorized when transaction succeeds with sufficient gas

### DIFF
--- a/crates/revm/src/evm.rs
+++ b/crates/revm/src/evm.rs
@@ -1930,12 +1930,12 @@ mod tests {
 
         let signed_auth = key_pair.create_signed_authorization(Address::repeat_byte(0x42))?;
 
-        // Use very low gas_limit - this should cause OOG
+        // Insufficient gas - will cause OOG during key_authorization processing
         let tx_low_gas = TxBuilder::new()
             .call_identity(&[0x01])
             .authorization(signed_auth)
             .key_authorization(signed_key_auth)
-            .gas_limit(600_000) // Very low - definitely not enough to key authorization
+            .gas_limit(589_000)
             .build();
 
         let signed_tx_low = key_pair.sign_tx(tx_low_gas)?;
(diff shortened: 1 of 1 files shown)


crates/revm/src/evm.rs
```diff
@@ -1863,4 +1863,174 @@ mod tests {
...
+                assert_eq!(result.gas_used(), 600_000, "Gas used should be gas limit");
+                assert!(
+                    !result.is_success(),
+                    "Transaction with insufficient gas should fail"
+                );
+            }
+            Err(_) => {
+                // Transaction rejected during validation - also expected
+            }
+        }
```
0xKitsune, line 1968: Can we update to assert the exact error here?

crates/revm/src/evm.rs
```diff
@@ -1863,4 +1863,174 @@ mod tests {
...
+
+        let signed_auth = key_pair.create_signed_authorization(Address::repeat_byte(0x42))?;
+
+        // Use very low gas_limit - this should cause OOG
+        let tx_low_gas = TxBuilder::new()
+            .call_identity(&[0x01])
+            .authorization(signed_auth)
+            .key_authorization(signed_key_auth)
+            .gas_limit(600_000) // Very low - definitely not enough to key authorization
```
0xKitsune: nit: The comment says `// Very low - definitely not enough` but this is actually above the successful case's `590914`. The test still works because the transaction has additional overhead, but the comment is slightly misleading.

PR #2334, opened 2026-01-29
feat(xtask): replay DKG for an epoch and inspect block for DKG data

SuperFluffy:
Adds 2 tools debug the DKG of an epoch and to inspect blocks for DKG data.

This is useful, for example, to figure out which players received and acknowledged shares from which dealers.

Usage:

```
cargo x replay-dkg \
    --rpc-url https://<credentials>@rpc.moderato.tempo.xyz \
    --chain-spec moderato \
    --epoch 134

cargo x inspect-block \
    --rpc-url https://<credentials>@rpc.moderato.tempo.xyz \
    --chain-spec moderato \
    --block 323451
```


xtask/src/replay_dkg.rs
```diff
@@ -0,0 +1,375 @@
...
+                .wrap_err("failed to add batch calls")?;
+
+            batch.send().await.wrap_err("failed to send batch")?;
+
+            for (height, waiter) in heights.into_iter().zip(waiters) {
+                progress.inc(1);
+
+                let block = match waiter.await {
+                    Ok(Some(block)) => block,
+                    Ok(None) => {
```
hamdiallam, line 237: When would we enter the None case? If it's  the current epoch then there will be no dkg outcome for awhile.

If this is an error, should we retry or at least the missing data in the result for context?

hamdiallam (approved):
LGTM, one comment re dkg replay

xtask/src/replay_dkg.rs
SuperFluffy: I don't actually know the error conditions for this. I assume the block is missing for whatever reason (note that this case is not for DKG missing but for `eth_getBlockByNumber` returning nothing, which I assume means that the RPC does not have that).

PR #2340, opened 2026-01-29
refactor(revm): extract gas limit check into helper function

jenpaff:
## Summary

Extracts the duplicated gas limit checking logic from `execution()` and `inspect_execution()` into a shared `check_gas_limit()` helper function.

## Context

Follow-up from PR #2325 (TIP-1000 gas metering fix). The gas limit checking logic was duplicated between `execution()` and `inspect_execution()` methods, requiring changes in two places.

Closes CHAIN-548

### DIFF
--- a/crates/revm/src/handler.rs
+++ b/crates/revm/src/handler.rs
@@ -1,6 +1,10 @@
 //! Tempo EVM Handler implementation.
 
-use std::{cmp::Ordering, fmt::Debug};
+use std::{
+    cmp::Ordering,
+    fmt::Debug,
+    sync::{Arc, OnceLock},
+};
 
 use alloy_primitives::{Address, TxKind, U256};
 use reth_evm::{EvmError, EvmInternals};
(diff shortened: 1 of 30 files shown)


crates/revm/src/handler.rs
```diff
@@ -172,6 +172,26 @@ fn adjusted_initial_gas(
     }
 }
 
+/// Checks if gas limit is sufficient and returns OOG frame result if not.
+///
+/// For T0+, validates gas limit covers intrinsic gas. For pre-T0, skips check
+/// to maintain backward compatibility.
+#[inline]
+fn check_gas_limit(
```
mattsse: hate to be pedantic here but amp's default locations are driving me insane, this does not belong top of file, this is a private fn and belongs end of file

jenpaff: yeah good point- thanks for flagging

mattsse (approved):
lgtm, pending @rakita or @legion2002

PR #2344, opened 2026-01-30
feat(tip-1015): implement compound transfer policies for T1

gakonst:
## Summary

Implements TIP-1015 (Compound Transfer Policies) targeting the T1 hardfork.

## Motivation

Real-world requirements often differ between sending, receiving, and minting tokens. This TIP enables:
- **Vendor credits**: Credits that can be minted to anyone and spent to specific vendors
- **Sender restrictions**: Block sanctioned addresses from sending while allowing anyone to receive
- **Recipient restrictions**: Require KYC for recipients while allowing any holder to send

### TIP-403 Registry
- Add `PolicyType::COMPOUND` enum variant
- Add `createCompoundPolicy(senderPolicyId, recipientPolicyId, mintRecipientPolicyId)`
- Add `isAuthorizedSender(policyId, user)`
- Add `isAuthorizedRecipient(policyId, user)`
- Add `isAuthorizedMintRecipient(policyId, user)`
- Add `compoundPolicyData(policyId)` view function
- Add `CompoundPolicyCreated` event
- Add `PolicyNotSimple` and `PolicyDoesNotExist` errors

### TIP-20 (T1-gated)
- `_mint`: Use `isAuthorizedMintRecipient` instead of `isAuthorized`
- `is_transfer_authorized`: Use `isAuthorizedSender`/`isAuthorizedRecipient`
- `burn_blocked`: Use `isAuthorizedSender` for blocked check

### Stablecoin DEX (T1-gated)
- `cancel_stale_order`: Use `isAuthorizedSender` for stale detection

### Other
- Fee payer authorization uses `isAuthorizedSender` (T1-gated)
- Updated TIP spec `protocolVersion` from T2 to T1

## Testing
- All existing TIP-403 tests pass
- Clippy clean

## Related
- Implements spec from PR #2372
- Thread: https://tempoxyz.slack.com/archives/C0A99L9RLHZ/p1769645072489649

### DIFF
--- a/crates/precompiles/src/tip20/mod.rs
+++ b/crates/precompiles/src/tip20/mod.rs
@@ -680,7 +680,10 @@ impl TIP20Token {
         let policy_id = self.transfer_policy_id()?;
         let registry = TIP403Registry::new();
 
-        if !registry.is_authorized_as(policy_id, from, AuthRole::sender())? {
+        // (spec: +T1) short-circuit and skip recipient check if sender fails
+        if !registry.is_authorized_as(policy_id, from, AuthRole::sender())?
+            && self.storage.spec().is_t1()
+        {
             return Ok(false);
         }
         registry.is_authorized_as(policy_id, to, AuthRole::recipient())
--- a/crates/precompiles/src/tip403_registry/mod.rs
+++ b/crates/precompiles/src/tip403_registry/mod.rs
@@ -1,7 +1,10 @@
 pub mod dispatch;
 
 use crate::StorageCtx;
-pub use tempo_contracts::precompiles::{ITIP403Registry, TIP403RegistryError, TIP403RegistryEvent};
+pub use tempo_contracts::precompiles::{
+    ITIP403Registry::{self, PolicyType},
+    TIP403RegistryError, TIP403RegistryEvent,
+};
 use tempo_precompiles_macros::{Storable, contract};
 
 use crate::{
@@ -85,9 +88,28 @@ impl PolicyData {
             .expect("unable to insert 'admin'")
     }
 
+    /// Decodes the raw `policy_type` u8 to a `PolicyType` enum.
+    ///
+    /// - **T1+**: fails on invalid values (early revert)
+    /// - **Pre-T1**: returns `__Invalid` for unknown values (defer error to caller)
+    fn policy_type(&self) -> Result<PolicyType> {
+        let is_t1 = StorageCtx.spec().is_t1();
+        match self.policy_type.try_into() {
+            Err(_) => {
+                if is_t1 {
+                    Err(TempoPrecompileError::under_overflow())
+                } else {
+                    Ok(PolicyType::__Invalid)
+                }
+            }
+            Ok(PolicyType::COMPOUND) if !is_t1 => Ok(PolicyType::__Invalid),
+            Ok(policy_type) => Ok(policy_type),
+        }
+    }
+
     /// Returns `true` if the policy data indicates a compound policy
     fn is_compound(&self) -> bool {
-        self.policy_type == ITIP403Registry::PolicyType::COMPOUND as u8
+        self.policy_type == PolicyType::COMPOUND as u8
     }
 }
 
@@ -126,18 +148,13 @@ impl TIP403Registry {
         }
 
         let data = self.get_policy_data(call.policyId)?;
+        let policy_type = data.policy_type()?;
 
-        // Pre-T1: any policy_type >= 2 should revert to match original try_into behavior
-        // where unknown enum values (which became 255/__Invalid) would fail conversion.
-        if data.policy_type >= 2 && !self.storage.spec().is_t1() {
-            return Err(TempoPrecompileError::under_overflow());
-        }
+        // Ensure invalid policy types revert in pre-T1
+        policy_type.ensure_is_valid()?;
 
         Ok(ITIP403Registry::policyDataReturn {
-            policyType: data
-                .policy_type
-                .try_into()
-                .map_err(|_| TempoPrecompileError::under_overflow())?,
+            policyType: policy_type,
             admin: data.admin,
         })
     }
@@ -170,29 +187,12 @@ impl TIP403Registry {
         })
     }
 
-    /// Validates and returns the policy type to store, handling backward compatibility.
-    ///
-    /// Pre-T1: Converts any value >= 2 to 255 (__Invalid) to match original ABI decoding
-    /// behavior where unknown enum values became __Invalid(255).
-    /// T1+: Only allows WHITELIST and BLACKLIST; COMPOUND must use createCompoundPolicy.
-    fn validate_simple_policy_type(&self, policy_type: ITIP403Registry::PolicyType) -> Result<u8> {
-        let policy_type = policy_type as u8;
-        if self.storage.spec().is_t1() {
-            if policy_type >= 2 {
-                return Err(TIP403RegistryError::incompatible_policy_type().into());
-            }
-            Ok(policy_type)
-        } else {
-            Ok(if policy_type >= 2 { 255u8 } else { policy_type })
-        }
-    }
-
     pub fn create_policy(
         &mut self,
         msg_sender: Address,
         call: ITIP403Registry::createPolicyCall,
     ) -> Result<u64> {
-        let policy_type = self.validate_simple_policy_type(call.policyType)?;
+        let policy_type = call.policyType.ensure_is_simple()?;
 
         let new_policy_id = self.policy_id_counter()?;
 
@@ -213,9 +213,7 @@ impl TIP403Registry {
             ITIP403Registry::PolicyCreated {
                 policyId: new_policy_id,
                 updater: msg_sender,
-                policyType: policy_type
-                    .try_into()
-                    .unwrap_or(ITIP403Registry::PolicyType::__Invalid),
+                policyType: policy_type.try_into().unwrap_or(PolicyType::__Invalid),
             },
         ))?;
 
@@ -236,7 +234,7 @@ impl TIP403Registry {
         call: ITIP403Registry::createPolicyWithAccountsCall,
     ) -> Result<u64> {
         let admin = call.admin;
-        let policy_type = self.validate_simple_policy_type(call.policyType)?;
+        let policy_type = call.policyType.ensure_is_simple()?;
 
         let new_policy_id = self.policy_id_counter()?;
 
@@ -256,7 +254,7 @@ impl TIP403Registry {
             self.set_policy_set(new_policy_id, *account, true)?;
 
             match call.policyType {
-                ITIP403Registry::PolicyType::WHITELIST => {
+                PolicyType::WHITELIST => {
                     self.emit_event(TIP403RegistryEvent::WhitelistUpdated(
                         ITIP403Registry::WhitelistUpdated {
                             policyId: new_policy_id,
@@ -266,7 +264,7 @@ impl TIP403Registry {
                         },
                     ))?;
                 }
-                ITIP403Registry::PolicyType::BLACKLIST => {
+                PolicyType::BLACKLIST => {
                     self.emit_event(TIP403RegistryEvent::BlacklistUpdated(
                         ITIP403Registry::BlacklistUpdated {
                             policyId: new_policy_id,
@@ -287,9 +285,7 @@ impl TIP403Registry {
             ITIP403Registry::PolicyCreated {
                 policyId: new_policy_id,
                 updater: msg_sender,
-                policyType: policy_type
-                    .try_into()
-                    .unwrap_or(ITIP403Registry::PolicyType::__Invalid),
+                policyType: policy_type.try_into().unwrap_or(PolicyType::__Invalid),
             },
         ))?;
 
@@ -347,7 +343,7 @@ impl TIP403Registry {
         }
 
         // Check policy type
-        if data.policy_type != ITIP403Registry::PolicyType::WHITELIST as u8 {
+        if !matches!(data.policy_type()?, PolicyType::WHITELIST) {
             return Err(TIP403RegistryError::incompatible_policy_type().into());
         }
 
@@ -376,7 +372,7 @@ impl TIP403Registry {
         }
 
         // Check policy type
-        if data.policy_type != ITIP403Registry::PolicyType::BLACKLIST as u8 {
+        if !matches!(data.policy_type()?, PolicyType::BLACKLIST) {
             return Err(TIP403RegistryError::incompatible_policy_type().into());
         }
 
@@ -415,7 +411,7 @@ impl TIP403Registry {
         // Store policy record with COMPOUND type and compound data
         self.policy_records[new_policy_id].write(PolicyRecord {
             base: PolicyData {
-                policy_type: ITIP403Registry::PolicyType::COMPOUND as u8,
+                policy_type: PolicyType::COMPOUND as u8,
                 admin: Address::ZERO,
             },
             compound: CompoundPolicyData {
@@ -460,8 +456,10 @@ impl TIP403Registry {
                     self.is_authorized_simple_policy(compound.mint_recipient_policy_id, user)
                 }
                 AuthRole::Transfer => {
-                    if !self.is_authorized_simple_policy(compound.sender_policy_id, user)? {
-                        // Short-circuit: if sender fails, skip recipient check
+                    // (spec: +T1) short-circuit and skip recipient check if sender fails
+                    if !self.is_authorized_simple_policy(compound.sender_policy_id, user)?
+                        && self.storage.spec().is_t1()
+                    {
                         return Ok(false);
                     }
                     self.is_authorized_simple_policy(compound.recipient_policy_id, user)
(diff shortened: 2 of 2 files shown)


crates/precompiles/src/tip403_registry/dispatch.rs
```diff
@@ -22,6 +22,18 @@ impl Precompile for TIP403Registry {
...
+                }
+                ITIP403RegistryCalls::isAuthorizedRecipient(call) => {
+                    view(call, |c| self.is_authorized_recipient(c))
+                }
+                ITIP403RegistryCalls::isAuthorizedMintRecipient(call) => {
+                    view(call, |c| self.is_authorized_mint_recipient(c))
+                }
+                ITIP403RegistryCalls::compoundPolicyData(call) => {
+                    view(call, |c| self.compound_policy_data(c))
+                }
```
0xKitsune, line 76: These need to be hardfork gated.

crates/precompiles/src/tip403_registry/dispatch.rs
```diff
@@ -39,6 +51,9 @@ impl Precompile for TIP403Registry {
                 ITIP403RegistryCalls::modifyPolicyBlacklist(call) => {
                     mutate_void(call, msg_sender, |s, c| self.modify_policy_blacklist(s, c))
                 }
+                ITIP403RegistryCalls::createCompoundPolicy(call) => {
+                    mutate(call, msg_sender, |s, c| self.create_compound_policy(s, c))
```
0xKitsune, line 102: Same thing here, this needs to be hardfork gated.

crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -100,6 +110,50 @@ impl TIP403Registry {
...
+        self.is_authorized_recipient_internal(call.policyId, call.user)
+    }
+
+    /// Checks if a user is authorized as a mint recipient under the given policy (TIP-1011)
+    pub fn is_authorized_mint_recipient(
+        &self,
+        call: ITIP403Registry::isAuthorizedMintRecipientCall,
+    ) -> Result<bool> {
+        self.is_authorized_mint_recipient_internal(call.policyId, call.user)
+    }
```
0xKitsune: We dont need to implement a separate `is_authorized_xyz_internal()` function for these. Lets just add the logic directly in the function body.

crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -190,7 +244,8 @@ impl TIP403Registry {
                         },
                     ))?;
                 }
-                ITIP403Registry::PolicyType::__Invalid => {
+                ITIP403Registry::PolicyType::COMPOUND | ITIP403Registry::PolicyType::__Invalid => {
```
0xKitsune, line 277: This also needs to be hardfork gated.

crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -335,11 +437,117 @@ impl TIP403Registry {
...
+            return Ok(policy_id == 1);
+        }
+
+        let data = self.get_policy_data(policy_id)?;
+
+        // For compound policies, delegate to the sender policy
+        if data.policy_type == ITIP403Registry::PolicyType::COMPOUND as u8 {
+            let compound = self.compound_policy_data[policy_id].read()?;
+            return self.is_authorized_internal(compound.sender_policy_id, user);
+        }
```
0xKitsune: This is repeated for every auth check function. Can we encapsulate this into a separate function?

0xKitsune (changes requested):

crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -335,11 +437,117 @@ impl TIP403Registry {
...
+    fn is_authorized_mint_recipient_internal(&self, policy_id: u64, user: Address) -> Result<bool> {
+        // Special case for always-allow and always-reject policies
+        if policy_id < 2 {
+            return Ok(policy_id == 1);
+        }
+
+        let data = self.get_policy_data(policy_id)?;
+
+        // For compound policies, delegate to the mint recipient policy
+        if data.policy_type == ITIP403Registry::PolicyType::COMPOUND as u8 {
```
0xKitsune: It might be nice to have an `is_compound()` function for this instead.

crates/precompiles/src/tip403_registry/dispatch.rs
```diff
@@ -22,6 +28,43 @@ impl Precompile for TIP403Registry {
...
+                    }
+                    view(call, |c| self.is_authorized_recipient(c))
+                }
+                ITIP403RegistryCalls::isAuthorizedMintRecipient(call) => {
+                    if !self.storage.spec().is_t1() {
+                        return unknown_selector(
+                            isAuthorizedMintRecipientCall::SELECTOR,
+                            self.storage.gas_used(),
+                        );
+                    }
```
0xKitsune, line 63: Consider creating a function for this to reduce repetitive code. This is a pattern that we use in various precompile dispatch logic, but it might make sense to create a macro for this or similar.

0xrusowsky: i think this is fine for now, as it will be gone once we have macro-generated dispatcher impls

crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -318,14 +462,48 @@ impl TIP403Registry {
...
+        }
+
+        self.is_authorized_simple(policy_id, user, &data)
+    }
+
+    /// Returns true if the policy ID is a built-in policy (0 = always-reject, 1 = always-allow)
+    #[inline]
+    fn is_builtin_policy(&self, policy_id: u64) -> bool {
+        policy_id < 2
+    }
```
0xrusowsky: this is better suited as:
```rs
fn builtin_authorization(&self, policy_id: u64) -> Option<bool> {
    (policy_id < 2).then(|| policy_id == 1)
}
```

so that it already encapsulates the decision. 
then we simply use it as:
```rs
if let Some(auth) = self.builtin_authorization(policy_id) {
   return Ok(auth);
}
```

crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -100,6 +110,92 @@ impl TIP403Registry {
...
+        }
+
+        let data = self.get_policy_data(policy_id)?;
+
+        if self.is_compound(&data) {
+            let compound = self.compound_policy_data[policy_id].read()?;
+            return self.is_authorized_internal(compound.sender_policy_id, user);
+        }
+
+        self.is_authorized_simple(policy_id, user, &data)
```
0xrusowsky: this logic is identical for `fn is_authorized_<recipient/sender/mint>`.
let's create a common internal `fn is_authorized_for_role` instead.

crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -318,14 +462,48 @@ impl TIP403Registry {
...
+                    ITIP403Registry::isAuthorizedRecipientCall {
+                        policyId: policy_id,
+                        user,
+                    },
+                )?,
+            );
+        }
+
+        self.is_authorized_simple(policy_id, user, &data)
+    }
```
0xrusowsky: if we use the single `fn is_authorized_for_role` we don't need call this fn there anymore.
thus this fn can disappear and its logic be directly inlined in `fn is_authorized`

crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -15,6 +15,16 @@ pub struct TIP403Registry {
...
+    compound_policy_data: Mapping<u64, CompoundPolicyData>,
+}
+
+/// Data for compound policies (TIP-1015)
+#[derive(Debug, Clone, Storable)]
+pub struct CompoundPolicyData {
+    pub sender_policy_id: u64,
+    pub recipient_policy_id: u64,
+    pub mint_recipient_policy_id: u64,
 }
```
0xrusowsky: perf note:

the proposed approach uses two separate mappings (`policy_data` and `compound_policy_data`), which requires 2 keccak operations per compound policy authorization check.
instead, we could merge them into a single mapping with a nested struct:

```rs
struct PolicyData {
    pub policy_type: u8,
    pub admin: Address,
}

struct CompoundPolicyData {
    pub sender_policy_id: u64,
    pub recipient_policy_id: u64,
    pub mint_recipient_policy_id: u64,
}

struct PolicyRecord {
    base: PolicyData,             // slot 0 (21 bytes)
    compound: CompoundPolicyData, // slot 1 (24 bytes)
}

policy_records: Mapping<u64, PolicyRecord>
```

with a single mapping, one keccak computes the base slot, and both base and compound are accessible at known offsets (base+0, base+1) without additional hashing.

storage costs: our macro system (thanks to the autogenerated `PolicyRecordHandler`) supports partial struct read/writes.
- simple policies: self.policy_records[id].base.read() → 1 keccak, 1 SLOAD
- compound policies: read base, then self.policy_records[id].compound.read() → 1 keccak, 2 SLOADs

0xrusowsky (changes requested):

crates/contracts/src/precompiles/tip403_registry.rs
```diff
@@ -9,32 +9,41 @@ crate::sol! {
...
         event WhitelistUpdated(uint64 indexed policyId, address indexed updater, address indexed account, bool allowed);
         event BlacklistUpdated(uint64 indexed policyId, address indexed updater, address indexed account, bool restricted);
+        event CompoundPolicyCreated(uint64 indexed policyId, address indexed creator, uint64 senderPolicyId, uint64 recipientPolicyId, uint64 mintRecipientPolicyId);
 
         // Errors
         error Unauthorized();
         error IncompatiblePolicyType();
         error PolicyNotFound();
+        error PolicyNotSimple();
+        error PolicyDoesNotExist();
```
0xrusowsky: this is redundant, and imo we should use `PolicyNotFound()` instead

0xrusowsky: addressed in https://github.com/tempoxyz/tempo/pull/2344/commits/8bc656084054707131db4e9b65e1b8c1086fb759

crates/precompiles/src/tip403_registry/mod.rs
0xrusowsky: addressed in https://github.com/tempoxyz/tempo/pull/2344/commits/87e4f3a6aa14039c13ffa934d5848d209c7fb000

crates/precompiles/src/tip403_registry/mod.rs
0xrusowsky: addressed in https://github.com/tempoxyz/tempo/pull/2344/commits/87e4f3a6aa14039c13ffa934d5848d209c7fb000

crates/precompiles/src/tip403_registry/mod.rs
0xrusowsky: addressed in https://github.com/tempoxyz/tempo/pull/2344/commits/87e4f3a6aa14039c13ffa934d5848d209c7fb000

crates/precompiles/src/tip403_registry/mod.rs
0xrusowsky: addressed in https://github.com/tempoxyz/tempo/pull/2344/commits/87e4f3a6aa14039c13ffa934d5848d209c7fb000

0xrusowsky (approved):

crates/contracts/src/precompiles/tip403_registry.rs
```diff
@@ -9,32 +9,40 @@ crate::sol! {
...
+            BLACKLIST,
+            COMPOUND
         }
 
         // View Functions
         function policyIdCounter() external view returns (uint64);
         function policyExists(uint64 policyId) external view returns (bool);
         function policyData(uint64 policyId) external view returns (PolicyType policyType, address admin);
         function isAuthorized(uint64 policyId, address user) external view returns (bool);
+        function isAuthorizedSender(uint64 policyId, address user) external view returns (bool);
```
legion2002, line 21: Do we need all of these view functions, or should we just do a single one using the new AuthRoles

legion2002: similar to is_authorized_as()

crates/precompiles/src/tip20/mod.rs
```diff
@@ -414,11 +418,15 @@ impl TIP20Token {
...
         }
 
-        // Check if the address is blocked from transferring
-        if TIP403Registry::new().is_authorized(ITIP403Registry::isAuthorizedCall {
-            policyId: self.transfer_policy_id()?,
-            user: call.from,
-        })? {
+        // Check if the address is blocked from transferring (sender authorization)
+        // TIP-1015: Use `isAuthorizedSender` for T1+, `isAuthorized` for pre-T1
+        let policy_id = self.transfer_policy_id()?;
```
legion2002, line 416: we can add a helper function for this pattern, where you pass in the new role for T1, and if it is not T1, then it just returns Transfer.

crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -335,19 +483,57 @@ impl TIP403Registry {
...
+        let data = self.get_policy_data(policy_id)?;
+        if data.is_compound() {
+            return Err(TIP403RegistryError::policy_not_simple().into());
+        }
+
+        Ok(())
+    }
+
+    // Internal helper functions
+    fn get_policy_data(&self, policy_id: u64) -> Result<PolicyData> {
```
legion2002, line 528: Didn't we have to add an assertion here for the T1 hardfork to prevent the snapshot reexec error with policyType decoding @0xrusowsky ? 
Or was that somewhere else

crates/precompiles/tests/storage_tests/solidity/testdata/tip403_registry.sol
```diff
@@ -11,13 +11,24 @@ contract TIP403Registry {
...
+
     // ========== Storage ==========
 
     /// Counter for policy IDs
     uint64 public policyIdCounter;
 
-    /// Mapping of policy ID to policy data
-    mapping(uint64 => PolicyData) public policyData;
+    /// Mapping of policy ID to policy record (internal, not exposed in ABI)
+    mapping(uint64 => PolicyRecord) internal policyRecords;
```
legion2002, line 31: we also need to update TIP20 and Dex solidity implementations

crates/precompiles/src/tip403_registry/mod.rs
0xrusowsky: only in the `fn policy_data()` called by the dispatcher

crates/precompiles/tests/storage_tests/solidity/testdata/tip403_registry.sol
0xrusowsky: wdym?

0xrusowsky: wdym?

0xrusowsky: wdym?

crates/contracts/src/precompiles/tip403_registry.rs
0xrusowsky: yes,  agree that it is better

however I didn't want to be intrusive with the spec changes.

crates/precompiles/src/tip20/mod.rs
0xrusowsky: addressed in https://github.com/tempoxyz/tempo/pull/2344/commits/7fea212af2605de6e860366c8244c08d128877d0

crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -3810,10 +3803,7 @@ mod tests {
                     restricted: true,
                 },
             )?;
-            assert!(!registry.is_authorized(ITIP403Registry::isAuthorizedCall {
-                policyId: policy_id,
-                user: alice,
-            })?);
+            assert!(!registry.is_authorized_as(policy_id, alice, AuthRole::Transfer)?);
```
0xrusowsky: is this supposed to be a symmetrical check @danrobinson?

crates/precompiles/tests/storage_tests/solidity/testdata/tip403_registry.sol
legion2002: I meant that we need to update the reference implementation solidity contracts

crates/contracts/src/precompiles/tip403_registry.rs
danrobinson: We decided we should leave the functions; more self-documenting and mistake-proof

crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -106,6 +166,16 @@ impl TIP403Registry {
...
     ) -> Result<u64> {
+        if self.storage.spec().is_t1()
+            && matches!(
+                call.policyType,
+                ITIP403Registry::PolicyType::COMPOUND | ITIP403Registry::PolicyType::__Invalid
+            )
+        {
+            // COMPOUND policies are created via createCompoundPolicy
+            return Err(TIP403RegistryError::incompatible_policy_type().into());
+        }
```
0xKitsune: Since we are updating the enum to include `COMPOUND`, if a user calls this function and specifies a compound policy pre t1, wont this cause an issue if this code path is already hit on existing networks?

crates/contracts/src/precompiles/tip403_registry.rs
```diff
@@ -9,32 +9,40 @@ crate::sol! {
...
         event PolicyCreated(uint64 indexed policyId, address indexed updater, PolicyType policyType);
         event WhitelistUpdated(uint64 indexed policyId, address indexed updater, address indexed account, bool allowed);
         event BlacklistUpdated(uint64 indexed policyId, address indexed updater, address indexed account, bool restricted);
+        event CompoundPolicyCreated(uint64 indexed policyId, address indexed creator, uint64 senderPolicyId, uint64 recipientPolicyId, uint64 mintRecipientPolicyId);
 
         // Errors
         error Unauthorized();
         error IncompatiblePolicyType();
-        error PolicyNotFound();
+        error PolicyNotFound(uint64 policyId);
```
0xrusowsky: this is a breaking change. if we do this we need to add a new error and make the builder `fn policy_not_found()` auto route to the correct fork

legion2002: Is there a specific reason we want this error change? don't think it is worth doing this minor update as a hardfork

crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -83,10 +122,17 @@ impl TIP403Registry {
...
         })? {
-            return Err(TIP403RegistryError::policy_not_found().into());
+            return Err(TIP403RegistryError::policy_not_found(call.policyId).into());
         }
 
         let data = self.get_policy_data(call.policyId)?;
+
+        // Pre-T1: any policy_type >= 2 should revert to match original try_into behavior
+        // where unknown enum values (which became 255/__Invalid) would fail conversion.
+        if data.policy_type >= 2 && !self.storage.spec().is_t1() {
```
legion2002: Noting that we have sent some invalid policy type txs already on testnet. We should also send some txs, which view the policy data, so that if snapshot reexecution passes, we can be confident that this is correct

crates/precompiles/src/stablecoin_dex/mod.rs
0xrusowsky: addressed in https://github.com/tempoxyz/tempo/pull/2344/commits/515148fd32447d4b19747ecac9c72e43629d57fd

crates/contracts/src/precompiles/tip403_registry.rs
0xrusowsky: ended up reverting the changes in https://github.com/tempoxyz/tempo/pull/2344/commits/ee2007701850916c133ad47bbe6e0b9e62149c81

PR #2345, opened 2026-01-30
tip: TIP-XXXX TIP-20 Address Prefix Protection for Contract Creation

howydev:
Adds TIP-XXXX documenting the TIP-20 address prefix protection feature for contract creation operations.

## Summary

This TIP enforces TIP-20 address prefix exclusivity by rejecting contract creation at addresses starting with `0x20C0...`.

## Problem

There is no explicit protection preventing contracts from being deployed at addresses with the TIP-20 prefix. While computationally infeasible today, it's still technically possible to deploy non-TIP20 contracts at addresses that will pass protocol `isTIP20` checks.

## Solution

Block contract creation (`CREATE`, `CREATE2`, EIP-7702 delegation) when the target address starts with the reserved TIP-20 address prefix (`0x20C000000000000000000000`).

## Related

- Linear: CHAIN-418
- Requires hard fork

### DIFF
--- a/tips/TIP-XXXX-tip20-prefix-protection.md
+++ b/tips/TIP-XXXX-tip20-prefix-protection.md
@@ -1,0 +1,115 @@
+---
+id: TIP-XXXX
+title: TIP-20 Address Prefix Protection for Contract Creation
+description: Prevents contracts from being deployed at addresses with the reserved TIP-20 address prefix via CREATE, CREATE2, or EIP-7702 delegation.
+authors: howydev
+status: Draft
+related: TIP-20
+protocolVersion: TBD
+---
+
+# TIP-XXXX: TIP-20 Address Prefix Protection for Contract Creation
+
+## Abstract
+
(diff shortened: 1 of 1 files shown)


tips/TIP-XXXX-tip20-prefix-protection.md
```diff
@@ -0,0 +1,115 @@
...
+```solidity
+function isTIP20(address token) internal view returns (bool) {
+    return bytes12(bytes20(token)) == bytes12(0x20c000000000000000000000) 
+        && token.code.length > 0;
+}
+```
+
+Currently, there is no explicit protection preventing a contract from being deployed at an address that happens to have this prefix. With sufficient computing power, an attacker could find a `CREATE2` salt that produces an address with the TIP-20 prefix, allowing arbitrary code to be deployed at an address that passes `isTIP20` checks.
+
+One of the core invariants of the TIP-20 system is that `isTIP20` reliably identifies tokens deployed through the official factory. This proposal enforces that invariant at the protocol level.
```
howydev:
```suggestion
One of the core invariants relied on by the Tempo precompiles is that `isTIP20` reliably identifies tokens deployed through the official factory. This proposal enforces that invariant at the protocol level.
```

PR #2351, opened 2026-01-30
fix(txpool): reject CREATE calls with non-empty authorization list

rakita:
## Summary
- Add validation to reject AA transactions that have CREATE calls with a non-empty authorization list
- Add `CreateCallWithAuthorizationList` error variant to `TempoPoolTransactionError`
- Add test `test_aa_create_call_with_authorization_list_rejected` to verify the validation

## Test plan
- [x] Added unit test that verifies CREATE calls with authorization list are rejected
- [x] Test passes: `cargo test -p tempo-transaction-pool test_aa_create_call_with_authorization_list_rejected`

### DIFF
--- a/crates/transaction-pool/src/transaction.rs
+++ b/crates/transaction-pool/src/transaction.rs
@@ -265,9 +265,8 @@ pub enum TempoPoolTransactionError {
     #[error("CREATE calls must be the first call in an AA transaction")]
     CreateCallNotFirst,
 
-    /// Thrown when an AA transaction contains a CREATE call and has a non-empty authorization list.
-    /// CREATE calls and authorization lists are mutually exclusive.
-    #[error("AA transactions with CREATE calls cannot have an authorization list")]
+    /// Thrown when an AA transaction contains both a CREATE call and an authorization list.
+    #[error("CREATE calls are not allowed in the same transaction that has an authorization list")]
     CreateCallWithAuthorizationList,
 
     /// Thrown when a call in an AA transaction has input data exceeding the maximum allowed size.


crates/transaction-pool/src/validator.rs
```diff
@@ -460,8 +460,15 @@ where
...
         for (idx, call) in tx.calls.iter().enumerate() {
-            if call.to.is_create() && idx != 0 {
-                return Err(TempoPoolTransactionError::CreateCallNotFirst);
+            if call.to.is_create() {
+                // CREATE call must be the first call in the transaction.
+                if idx != 0 {
+                    return Err(TempoPoolTransactionError::CreateCallNotFirst);
+                }
+                // CREATE call must not be present in the authorization list.
+                if !tx.tempo_authorization_list.is_empty() {
```
mattsse: i find this comment a bit confusing can you clarify what 

> not be present in the authorization list.

means

rakita: Reworded

crates/transaction-pool/src/validator.rs
```diff
@@ -460,8 +460,15 @@ where
...
-            if call.to.is_create() && idx != 0 {
-                return Err(TempoPoolTransactionError::CreateCallNotFirst);
+            if call.to.is_create() {
+                // CREATE call must be the first call in the transaction.
+                if idx != 0 {
+                    return Err(TempoPoolTransactionError::CreateCallNotFirst);
+                }
+                // CREATE calls are not allowed in transactions with an authorization list.
+                if !tx.tempo_authorization_list.is_empty() {
+                    return Err(TempoPoolTransactionError::CreateCallWithAuthorizationList);
```
mattsse, line 470: this doesnt match the error's description which is

> Thrown when a call in an AA transaction is a CREATE and is present in the authorization list.

does this mean create calls mandate empty auth list?
not clear from the error's description, maybe the error should be revised so that, "and the transactions has authorizations"

rakita: Have changed the comment in error.

PR #2353, opened 2026-01-30
feat(txpool): evict transactions when keychain keys expire

fgimenez:
Closes CHAIN-551

Evict transactions from the pool when their keychain key expires.

Previously, we only validated key expiry at mempool admission (CHAIN-451). This left a gap: keys could expire while transactions were already in the pool, leaving stale transactions that would fail execution.

This PR adds `KeyExpiryTracker` to track keychain key expiry timestamps. When a keychain-signed transaction enters the pool, we read the key's expiry from AccountKeychain storage and track it. On each block, transactions using expired keys (expiry <= tip_timestamp) are evicted alongside expired `valid_before` transactions.

### DIFF
--- a/crates/transaction-pool/src/maintain.rs
+++ b/crates/transaction-pool/src/maintain.rs
@@ -14,7 +14,7 @@ use futures::StreamExt;
 use reth_chainspec::ChainSpecProvider;
 use reth_primitives_traits::AlloyBlockHeader;
 use reth_provider::{CanonStateNotification, CanonStateSubscriptions, Chain};
-use reth_storage_api::{StateProvider, StateProviderFactory};
+use reth_storage_api::StateProviderFactory;
 use reth_transaction_pool::{PoolTransaction, TransactionOrigin, TransactionPool};
 use std::{
     collections::{BTreeMap, HashMap, HashSet},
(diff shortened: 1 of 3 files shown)


crates/transaction-pool/src/maintain.rs
```diff
@@ -191,6 +194,54 @@ impl TempoPoolState {
...
+    /// Track a keychain transaction's key expiry for eviction.
+    ///
+    /// Reads the key's expiry from state and tracks it. Skips if:
+    /// - Transaction is not a keychain-signed AA transaction
+    /// - Transaction has a key_authorization (key is being provisioned, not yet in state)
+    /// - Key expiry cannot be read from state
+    fn track_key_expiry(&mut self, tx: &TempoPooledTransaction, state: &impl StateProvider) {
+        let Some(aa_tx) = tx.inner().as_aa() else {
+            return;
+        };
```
mattsse: arent we also computing all of this in the validation section already?

it would be better if we add this to the TempoPooledTransaction itself, then we dont need to redo all of this here?

mattsse (changes requested):

crates/transaction-pool/src/maintain.rs
fgimenez: makes sense, added `key_expiry` here https://github.com/tempoxyz/tempo/pull/2353/commits/a8ce588897375e401c073712438481a8e49836e3

crates/transaction-pool/src/maintain.rs
```diff
@@ -264,10 +283,79 @@ impl Default for PendingStalenessTracker {
...
+///
+/// Note: We don't query state here - the expiry is read once when the transaction
+/// enters the pool (during validation) and passed to the tracker.
+#[derive(Debug, Default)]
+struct KeyExpiryTracker {
+    /// Maps expiry timestamp -> set of keys that expire at that time.
+    expiry_map: BTreeMap<u64, HashSet<KeyId>>,
+    /// Maps KeyId -> (expiry timestamp, set of transaction hashes using this key).
+    key_to_txs: HashMap<KeyId, (u64, HashSet<TxHash>)>,
+}
```
mattsse, line 307: this tracker has a flaw, that will eventually always try to evict transactions, and will keep them around until they are expired

mattsse (approved):

PR #2354, opened 2026-01-30
chore(TIP-1015): simpligy impl + minimize SLOADs when possible

0xrusowsky:
This PR addresses the feedback in #2344 regarding code simplification for leaner and easier to maintain impl

additionally it incorposates the storage layout improvements made in #2356

### DIFF
--- a/crates/precompiles/src/tip403_registry/mod.rs
+++ b/crates/precompiles/src/tip403_registry/mod.rs
@@ -527,8 +527,10 @@ impl TIP403Registry {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::error::TempoPrecompileError;
-    use crate::storage::{StorageCtx, hashmap::HashMapStorageProvider};
+    use crate::{
+        error::TempoPrecompileError,
+        storage::{StorageCtx, hashmap::HashMapStorageProvider},
+    };
     use alloy::primitives::Address;
     use rand::Rng;
     use tempo_chainspec::hardfork::TempoHardfork;


crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -448,62 +399,59 @@ impl TIP403Registry {
...
 
-    /// Returns true if the policy data indicates a compound policy
-    #[inline]
-    fn is_compound(&self, data: &PolicyData) -> bool {
-        data.policy_type == ITIP403Registry::PolicyType::COMPOUND as u8
+    /// Authorization for simple (non-compound) policies only.
+    ///
+    /// **WARNING:** skips compound check - caller must guarantee policy is simple.
+    fn is_authorized_simple_policy(&self, policy_id: u64, user: Address) -> Result<bool> {
+        if let Some(auth) = self.builtin_authorization(policy_id) {
```
0xKitsune, line 468: Noting that this is checked twice when calling `is_authorized_as()`.

crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -212,7 +170,7 @@ impl TIP403Registry {
         )?;
 
         // Store policy data
-        self.policy_data[new_policy_id].write(PolicyData {
+        self.policy_records[new_policy_id].base.write(PolicyData {
             policy_type: call.policyType as u8,
             admin: call.admin,
         })?;
```
0xKitsune, line 191: Can we add a test to ensure that Compound can not be set as a policy option pre T1? Otherwise, there may be STF breaking changes.

crates/revm/src/common.rs
```diff
@@ -246,21 +246,15 @@ pub trait TempoStateAccess<M = ()> {
...
+                AuthRole::Sender
             } else {
-                registry.is_authorized(ITIP403Registry::isAuthorizedCall {
-                    policyId: transfer_policy_id,
-                    user: fee_payer,
-                })
-            }
+                AuthRole::Transfer
+            };
+            TIP403Registry::new().is_authorized_as(policy_id, fee_payer, role)
```
0xKitsune, line 257: Just double checking that this does not create a diff in gas?

0xrusowsky: good call, we may need to gate the optimization under T1. will verify

crates/precompiles/src/tip403_registry/mod.rs
0xrusowsky: addressed in https://github.com/tempoxyz/tempo/pull/2354/commits/a7ca5491bfedf6382c57278eb5248f9780dd18ea

crates/precompiles/src/tip403_registry/mod.rs
0xrusowsky: it is checked once for the main policy and once again for teh compound policy

also it's an inexpensive check (just `policy_id < 2` and `policy_id == 1`)

crates/revm/src/common.rs
0xrusowsky: this should be equivalent for `AuthRole::Transfer` hence why i gated that under T1

PR #2356, opened 2026-01-30
perf(TIP-1015)!: optimize storage layout

0xrusowsky:
### Motivation

The original TIP-1015 implementation uses 2 separate mappings for policy data:

```rust
policy_data: Mapping<u64, PolicyData>                   // slot 1
compound_policy_data: Mapping<u64, CompoundPolicyData>  // slot 3
```

This requires **2 keccak computations** per compound policy authorization check—one to locate `PolicyData`, another to locate `CompoundPolicyData`.

This PR merges them into a single `PolicyRecord` mapping:

```rust
policy_records: Mapping<u64, PolicyRecord>  // slot 1
// PolicyRecord { base: PolicyData (base_slot+0), compound: CompoundPolicyData (base_slot+1) }
```

Now compound authorization requires **1 keccak + 2 SLOADs** (base and compound at known offsets). This reduces CPU overhead by eliminating one keccak computation per compound policy check. Additionally, it arguably makes the design more comprehensive.

### Breaking Change Analysis

> **NOTE:** i marked the PR as breaking so that i have the reviewer's full attention. However, it is (please correct me if i'm worng) only breaking vs `the spec`, and not vs `main`.

**❌ Breaking vs `proposed TIP-1015 spec`:**

| Slot | rus/2344 | This PR |
|------|----------|---------|
| 0 | `policyIdCounter` | `policyIdCounter` ✅ |
| 1 | `policy_data` | `policy_records` (contains `PolicyData` at offset 0) ✅ |
| 2 | `policy_set` | `policy_set` ✅ |
| 3 | `compound_policy_data` | removed and moved over `policy_records.compound` (offset 1) ❌ |

The `compound_policy_data` mapping moved from slot 3 to being embedded at offset +1 within `policy_records`. Any compound policies created on `rus/2344` would have their data at `keccak(id, 3)`, but this PR expects it at `keccak(id, 1) + 1`.

**✅ Non-breaking vs `main`:**

| Slot | main | This PR |
|------|------|---------|
| 0 | `policyIdCounter` | `policyIdCounter` ✅ |
| 1 | `policyData` → `PolicyData` at offset 0 | `policy_records.base` → `PolicyData` at offset 0 ✅ |
| 2 | `policySet` | `policy_set` ✅ |

The `PolicyData` struct is stored at the **exact same location**: `keccak(id, 1)` with `policyType` at byte 0 and `admin` at byte 1. The new `CompoundPolicyData` is purely additive at slot+1 (zeroed for existing policies).

### DIFF
--- a/crates/precompiles/tests/storage_tests/solidity/testdata/tip403_registry.sol
+++ b/crates/precompiles/tests/storage_tests/solidity/testdata/tip403_registry.sol
@@ -27,8 +27,8 @@ contract TIP403Registry {
     /// Counter for policy IDs
     uint64 public policyIdCounter;
 
-    /// Mapping of policy ID to policy data
-    mapping(uint64 => PolicyRecord) public policyRecords;
+    /// Mapping of policy ID to policy record (internal, not exposed in ABI)
+    mapping(uint64 => PolicyRecord) internal policyRecords;
 
     /// Nested mapping for policy sets: policy_id -> address -> is_in_set
     /// Used for whitelist/blacklist entries
--- a/tips/tip-1015.md
+++ b/tips/tip-1015.md
@@ -76,8 +76,10 @@ The TIP403Registry storage layout:
 | Slot | Field | Description |
 |------|-------|-------------|
 | 0 | `policyIdCounter` | Counter for generating unique policy IDs |
-| 1 | `policyRecords` | `mapping(uint64 => PolicyRecord)` - Policy ID to policy record |
+| 1 | `policyRecords` (private) | `mapping(uint64 => PolicyRecord)` - Policy ID to policy record |
 | 2 | `policySet` | `mapping(uint64 => mapping(address => bool))` - Whitelist/blacklist membership |
+
+The `policyRecords` mapping is private (not exposed in the ABI). The existing `policyData(uint64 policyId)` view function provides backwards-compatible access to `PolicyData`.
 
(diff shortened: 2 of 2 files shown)


crates/precompiles/tests/storage_tests/solidity/testdata/tip403_registry.sol
```diff
@@ -11,13 +11,24 @@ contract TIP403Registry {
...
+    }
+
     // ========== Storage ==========
 
     /// Counter for policy IDs
     uint64 public policyIdCounter;
 
     /// Mapping of policy ID to policy data
-    mapping(uint64 => PolicyData) public policyData;
+    mapping(uint64 => PolicyRecord) public policyRecords;
```
0xKitsune: Updating the mapping name will afect the public getter's function signature.

0xrusowsky: good point, we should manually expose it in the specs too.
my idea is to not change the proposed API at all, so i'll make `policyRecords` private.

0xrusowsky: addressed in https://github.com/tempoxyz/tempo/pull/2356/commits/b91504c500eac5e2cc393922a0f7c82e6107b21a

PR #2358, opened 2026-01-30
fix(txpool): prevent DoS via permissionless setValidatorToken eviction attacks

fgimenez:
Closes CHAIN-553

Fixes a DoS vulnerability where an attacker could call `setValidatorToken` with an obscure token (one with no AMM pools) to trigger mass eviction of pending transactions, even if the attacker never produces blocks.

The AMM liquidity cache tracks `unique_tokens`, fee tokens used by validators who actually produced recent blocks. When processing `setValidatorToken` events, we now skip any token change where the new token isn't in this set. This prevents permissionless callers from triggering eviction since their token won't match any active validator's preference.

- Add `is_active_validator_token()` to `AmmLiquidityCache` to check if a token is used by recent block producers
- Filter validator token changes in `evict_invalidated_transactions()` to only process tokens already in the active validator set
- Update integration test to verify the fix prevents the attack

### DIFF
--- a/crates/transaction-pool/src/amm.rs
+++ b/crates/transaction-pool/src/amm.rs
@@ -21,8 +21,8 @@ use tempo_precompiles::{
 use tempo_primitives::TempoReceipt;
 use tempo_revm::IntoAddress;
 
-/// Number of recent validator tokens to track.
-const LAST_SEEN_TOKENS_WINDOW: usize = 10;
+/// Number of recent validators/tokens to track.
+const LAST_SEEN_WINDOW: usize = 10;
 
 #[derive(Debug, Clone)]
 pub struct AmmLiquidityCache {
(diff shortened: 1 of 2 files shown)


crates/transaction-pool/src/amm.rs
```diff
@@ -21,8 +21,8 @@ use tempo_precompiles::{
 use tempo_primitives::TempoReceipt;
 use tempo_revm::IntoAddress;
 
-/// Number of recent validator tokens to track.
-const LAST_SEEN_TOKENS_WINDOW: usize = 10;
+/// Number of recent validators/tokens to track.
+const LAST_SEEN_WINDOW: usize = 10;
```
0xrusowsky, line 25: since this is a static number, we could consider increasing it to have a buffer for when the validator set grows.

PR #2360, opened 2026-01-30
fix(revm): use custom gas params for key authorization in T1

rakita:
## Summary
- Use custom `GasParams` with only SSTORE gas enabled (250,000) for key authorization storage operations in T1 hardfork
- Previously was using the full `cfg.gas_params` which included unnecessary gas charges

## Test plan
- [x] Updated gas expectations in `crates/revm/src/evm.rs` (590,914 → 587,177)
- [x] Updated gas expectations in `crates/node/tests/it/tempo_transaction.rs`:
  - Keychain + KeyAuth: ~289,693 → ~285,926
  - Keychain P256 + KeyAuth: ~294,733 → ~290,966
  - KeyAuth secp256k1: ~286,669 → ~282,903
  - KeyAuth with 3 limits: ~355,612 → ~349,426

### DIFF
--- a/crates/revm/src/handler.rs
+++ b/crates/revm/src/handler.rs
@@ -834,8 +834,8 @@ where
                 TABLE
                     .get_or_init(|| {
                         let mut table = [0u64; 256];
-table[GasId::sstore_set_without_load_cost().as_usize()] = 
-    cfg.gas_params.get(GasId::sstore_set_without_load_cost());
+                        table[GasId::sstore_set_without_load_cost().as_usize()] =
+                            cfg.gas_params.get(GasId::sstore_set_without_load_cost());
                         GasParams::new(Arc::new(table))
                     })
                     .clone()


crates/revm/src/handler.rs
```diff
@@ -814,18 +818,34 @@ where
...
             };
 
+            // Create gas_params with only sstore increase for key authorization
+            let gas_params = if spec.is_t1() {
+                static TABLE: OnceLock<GasParams> = OnceLock::new();
+                // only enabled SSTORE gas param for T1 fork in keychain.
+                TABLE
+                    .get_or_init(|| {
+                        let mut table = [0u64; 256];
+                        table[GasId::sstore_set_without_load_cost().as_usize()] = 250_000;
```
legion2002: We should take this constant directly from gas params, without hardcoding it here.
Think we can do something like this

```suggestion
table[GasId::sstore_set_without_load_cost().as_usize()] = 
    cfg.gas_params.get(GasId::sstore_set_without_load_cost());
```

rakita: Looks resonable

legion2002 (approved):

gakonst (dismissed):
## Review: Potential Bug in GasParams Initialization

### Issue

The custom `GasParams` table is initialized with all zeros except `sstore_set_without_load_cost`. This causes **other gas costs to be incorrectly charged as 0** during key authorization storage operations.

```rust
let mut table = [0u64; 256];  // All values start at 0
table[GasId::sstore_set_without_load_cost().as_usize()] =
    cfg.gas_params.get(GasId::sstore_set_without_load_cost());  // Only this one is set
```

**Affected operations in `EvmPrecompileStorageProvider`:**

1. **`sstore()` (line 127)**: `sstore_static_gas()` returns 0 (uses `GasId::sstore_static()`)
2. **`sload()` (line 173)**: `cold_storage_additional_cost()` returns 0 (uses `GasId::cold_storage_additional_cost()`)
3. **`emit_event()` (line 160)**: `log_cost()` returns 0 (uses `GasId::logdata()` and `GasId::logtopic()`)
4. **`sstore()` refunds (line 136)**: `sstore_refund()` uses `sstore_clearing_slot_refund()` which returns 0

The `authorize_key` function calls `read()`, `write()`, and `emit_event()`, so all these paths are affected.

### Fix

Either:
1. Start from the full gas params and only override the ones you want to change (safer):
```rust
let gas_params = if spec.is_t1() {
    // Use the configured gas params as-is for key authorization
    // If only SSTORE cost is needed, make that explicit
    cfg.gas_params.clone()
} else {
    cfg.gas_params.clone()
};
```

2. Or explicitly set all required GasIds (if zeroing others is intentional):
```rust
let mut table = [0u64; 256];
table[GasId::sstore_set_without_load_cost().as_usize()] = cfg.gas_params.get(GasId::sstore_set_without_load_cost());
table[GasId::sstore_static().as_usize()] = cfg.gas_params.get(GasId::sstore_static());
// ... other required GasIds
```

### Question

Is the intent to have all other gas costs be 0 for key authorization? If so, please add a comment explaining this design decision. The current comment says "only enabled SSTORE gas param" but the practical effect is that event emission and reads are free, which may not be intended.

0xKitsune:
Cross posting @rakita's comment from a separate thread

> Is the intent to have all other gas costs be 0 for key authorization? If so, please add a comment explaining this design decision.

Yes, this is intentional. The zeroed gas params fix a **double-pricing bug**:

1. **Reads (`sload`, `t_read`)**: Already charged in the outer EVM context before entering key authorization. Charging again here would double-count.

2. **Event emission (`emit_event`)**: Similarly already accounted for in the overall transaction gas accounting.

3. **SSTORE**: Only the `sstore_set_without_load_cost` (250k) is needed because key authorization **only performs 0 → non-zero SSTOREs** (new key creation). The `key_already_exists` check prevents overwriting existing keys, and revoked keys cannot be re-authorized. So `sstore_static_gas`, refunds, and other SSTORE variants don't apply.

The gas reduction (~3,700-6,200 gas) represents the removal of these duplicate charges, not undercharging.

PR #2370, opened 2026-01-30
feat(txpool): add per-sender transaction limits to AA 2D pool

fgimenez:
Closes CHAIN-563

Adds per-sender transaction limits to the AA 2D pool to prevent a single sender from monopolizing pool capacity.

The global pool limits only cap total transactions, not per-sender. Without this, an attacker could fill the entire AA pool with their transactions, pushing out other users via priority-based eviction.

- Add `max_txs_per_sender` config to `AA2dPoolConfig` (value `pool_config.max_account_slots`, default: 16, matching reth)
- Track per-sender counts and reject transactions exceeding the limit
- Apply limit to both regular 2D nonce and expiring nonce transactions
- Replacements are allowed even at the limit

### DIFF
--- a/crates/transaction-pool/src/tt_2d_pool.rs
+++ b/crates/transaction-pool/src/tt_2d_pool.rs
@@ -608,7 +608,7 @@ impl AA2dPool {
     fn decrement_sender_count(&mut self, sender: Address) {
         if let hash_map::Entry::Occupied(mut entry) = self.txs_by_sender.entry(sender) {
             let count = entry.get_mut();
-            *count = count.saturating_sub(1);
+            *count -= 1;
             if *count == 0 {
                 entry.remove();
             }


crates/transaction-pool/src/tt_2d_pool.rs
```diff
@@ -161,6 +167,18 @@ impl AA2dPool {
             ));
         }
 
+        // Check per-sender limit before proceeding.
+        // Only reject if this is a new tx (not a replacement) and sender is at capacity.
+        let sender = transaction.sender();
+        let sender_count = self.txs_by_sender.get(&sender).copied().unwrap_or(0);
+        let is_replacement = self.by_id.contains_key(&tx_id);
```
mattsse: this is an extra lookup that we then always do once more

crates/transaction-pool/src/tt_2d_pool.rs
```diff
@@ -192,6 +210,8 @@ impl AA2dPool {
             }
             Entry::Vacant(entry) => {
                 entry.insert(Arc::clone(&tx));
```
mattsse, line 211: I think we can move this check here before we alter the internals?

crates/transaction-pool/src/tt_2d_pool.rs
```diff
@@ -563,8 +596,24 @@ impl AA2dPool {
...
+        self.decrement_sender_count(removed_tx.sender());
+
+        Some(removed_tx)
+    }
+
+    /// Decrements the transaction count for a sender, removing the entry if it reaches zero.
+    fn decrement_sender_count(&mut self, sender: Address) {
+        if let hash_map::Entry::Occupied(mut entry) = self.txs_by_sender.entry(sender) {
+            let count = entry.get_mut();
+            *count = count.saturating_sub(1);
```
mattsse: this can technically be -= 1 because we always check for 0 next and thats the only place where we do this

J1939AZ-code: -0ETH

PR #2372, opened 2026-01-30
tip: add TIP-1015 Compound Transfer Policies

gakonst:
## Summary

Adds TIP-1015, which extends TIP-403 with compound policies that specify different authorization rules for senders, recipients, and mint recipients.

## Motivation

Real-world requirements often differ between sending, receiving, and minting tokens. For example:

- **Vendor credits**: Credits that can be minted to anyone and spent to a specific vendor, but cannot be transferred peer-to-peer
- **Sender restrictions**: Block sanctioned addresses from sending, while allowing anyone to receive (for refunds or seizure)
- **Recipient restrictions**: Require KYC for recipients, while allowing any holder to send

## Changes

A compound policy references three simple policies:
- **senderPolicyId**: checked for transfer senders
- **recipientPolicyId**: checked for transfer recipients
- **mintRecipientPolicyId**: checked for mint recipients

### TIP-403 Registry

- New `PolicyType.COMPOUND` enum value
- `createCompoundPolicy(senderPolicyId, recipientPolicyId, mintRecipientPolicyId)`
- `isAuthorizedSender(policyId, user)`
- `isAuthorizedRecipient(policyId, user)`
- `isAuthorizedMintRecipient(policyId, user)`
- `isAuthorized()` returns `isAuthorizedSender() && isAuthorizedRecipient()`
- Compound policies are **immutable** once created

### Required Code Changes

6 direct replacements of `isAuthorized` calls:

| Location | Current | Replace With |
|----------|---------|--------------|
| TIP-20 `_mint` | `isAuthorized(to)` | `isAuthorizedMintRecipient(to)` |
| TIP-20 `burnBlocked` | `isAuthorized(from)` | `isAuthorizedSender(from)` |
| TIP-20 `isTransferAuthorized` | `isAuthorized(from)` | `isAuthorizedSender(from)` |
| TIP-20 `isTransferAuthorized` | `isAuthorized(to)` | `isAuthorizedRecipient(to)` |
| DEX `cancelStaleOrder` | `isAuthorized(maker)` | `isAuthorizedSender(maker)` |
| Fee payer `can_fee_payer_transfer` | `isAuthorized(fee_payer)` | `isAuthorizedSender(fee_payer)` |

All other call sites use `ensureTransferAuthorized(from, to)` which delegates to `isTransferAuthorized`, so they automatically inherit the correct behavior.

---
*Replaces #2316 after branch rename from `dan/tip-1011-complex-policies` to `tip-1015`*

### DIFF
--- a/crates/transaction-pool/src/tt_2d_pool.rs
+++ b/crates/transaction-pool/src/tt_2d_pool.rs
@@ -1488,7 +1488,10 @@ impl BestAA2dTransactions {
                 continue;
             }
             // Advance transaction that just got unlocked, if any.
-            if let Some(unlocked) = self.by_id.get(&id.unlocks()) {
+            // Skip for expiring nonce transactions as they are always independent.
+            if !id.seq_id.is_expiring_nonce()
+                && let Some(unlocked) = self.by_id.get(&id.unlocks())
+            {
                 self.independent.insert(unlocked.clone());
             }
             return Some((best.transaction, best.priority));
@@ -1537,6 +1540,14 @@ impl AASequenceId {
     /// Creates a new instance with the address and nonce key.
     pub const fn new(address: Address, nonce_key: U256) -> Self {
         Self { address, nonce_key }
+    }
+
+    /// Returns `true` if this sequence ID represents an expiring nonce transaction.
+    ///
+    /// Expiring nonce transactions use `nonce_key == U256::MAX` and are always independent,
+    /// meaning they don't have sequential nonce dependencies.
+    pub(crate) fn is_expiring_nonce(&self) -> bool {
+        self.nonce_key == U256::MAX
     }
 
     const fn start_bound(self) -> std::ops::Bound<AA2dTransactionId> {
@@ -3495,6 +3506,32 @@ mod tests {
 
         // The sequence should be in the invalid set, so next tx from same sender should be skipped
         // But since we already consumed tx0, we'd get tx1 next - but the sequence is now invalid
+    }
+
+    #[test]
+    fn test_best_transactions_expiring_nonce_independent() {
+        // Expiring nonce transactions (nonce_key == U256::MAX) are always independent
+        // and should not trigger unlock logic for dependent transactions
+        let mut pool = AA2dPool::default();
+        let sender = Address::random();
+
+        // Add expiring nonce transaction
+        let tx = TxBuilder::aa(sender).nonce_key(U256::MAX).nonce(0).build();
+        pool.add_transaction(
+            Arc::new(wrap_valid_tx(tx, TransactionOrigin::Local)),
+            0,
+            TempoHardfork::T1,
+        )
+        .unwrap();
+
+        let mut best = pool.best_transactions();
+
+        // Should return the transaction
+        let first = best.next();
+        assert!(first.is_some());
+
+        // No more transactions
+        assert!(best.next().is_none());
     }
 
     // ============================================


tips/ref-impls/src/interfaces/ITIP403Registry.sol
```diff
@@ -28,7 +28,8 @@ interface ITIP403Registry {
     error IncompatiblePolicyType();
 
     /// @notice Error when querying a policy that does not exist
-    error PolicyNotFound();
+    /// @param policyId The policy ID that was not found
+    error PolicyNotFound(uint64 policyId);
```
0xKitsune: Lets remove this change and leave the `PolicyNotFound` error as is.

crates/precompiles/src/tip20/mod.rs
```diff
@@ -679,23 +675,18 @@ impl TIP20Token {
...
-
-        // Check if 'to' address is authorized
-        let to_authorized = registry.is_authorized(ITIP403Registry::isAuthorizedCall {
-            policyId: transfer_policy_id,
-            user: to,
-        })?;
-
-        Ok(from_authorized && to_authorized)
+        // (spec: +T1) short-circuit and skip recipient check if sender fails
+        if !registry.is_authorized_as(policy_id, from, AuthRole::sender())?
```
legion2002: Isn't this a bug, where pre T1 the sender check is performed, but the result is ignored.
The correct behaviour would be that pre-T1 both sender and recipient checks are required to return true

crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -304,50 +388,234 @@ impl TIP403Registry {
...
+    fn is_simple(&self, policy_id: u64, user: Address, data: &PolicyData) -> Result<bool> {
+        // NOTE: read `policy_set` BEFORE checking policy type to match original gas consumption.
+        // Pre-T1: the old code read policy_set first, then failed on invalid policy types.
+        // This order must be preserved for block re-execution compatibility.
+        let is_in_set = self.policy_set[policy_id][user].read()?;
+
+        match data.policy_type()? {
+            PolicyType::WHITELIST => Ok(is_in_set),
+            PolicyType::BLACKLIST => Ok(!is_in_set),
+            PolicyType::COMPOUND => Ok(false),
```
legion2002: Is there a reason we don't return an error here? My understanding is that `is_simple` shuld never be called for a compound policy

crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -304,50 +388,234 @@ impl TIP403Registry {
...
+                }
+                AuthRole::Recipient => {
+                    self.is_authorized_simple_policy(compound.recipient_policy_id, user)
+                }
+                AuthRole::MintRecipient => {
+                    self.is_authorized_simple_policy(compound.mint_recipient_policy_id, user)
+                }
+                AuthRole::Transfer => {
+                    // (spec: +T1) short-circuit and skip recipient check if sender fails
+                    if !self.is_authorized_simple_policy(compound.sender_policy_id, user)?
```
legion2002: @0xrusowsky Same fix as before is needed here.

legion2002 (approved):
lgtm! assuming the rest of the comments are resolved.

crates/precompiles/src/tip403_registry/mod.rs
0xrusowsky: it is techinically unreachable, but it wouldn't hurt to "return" the error instead

crates/precompiles/src/tip403_registry/mod.rs
0xrusowsky: addressed in https://github.com/tempoxyz/tempo/pull/2372/commits/e70c9a730e4d87e736d2843fb638a9602bc068e2

crates/precompiles/src/tip403_registry/mod.rs
0xrusowsky: addressed in https://github.com/tempoxyz/tempo/pull/2372/commits/e70c9a730e4d87e736d2843fb638a9602bc068e2

crates/precompiles/src/tip20/mod.rs
0xrusowsky: addressed in https://github.com/tempoxyz/tempo/pull/2372/commits/e70c9a730e4d87e736d2843fb638a9602bc068e2

PR #2381, opened 2026-02-02
tip: add TIP-1020 Signature Verification Precompile

jxom:
Adds a new TIP for a signature verification precompile that enables contracts to verify Tempo signature types (secp256k1, P256, WebAuthn).


tips/tip-xxxx.md
```diff
@@ -0,0 +1,175 @@
...
+    error InvalidSignature();
+    error SignatureNotSupported();
+
+    function verify(address signer, bytes32 hash, bytes calldata signature) external view returns (bool);
+}
+```
+
+### Signature Encoding
+
+The `signature` parameter MUST be encoded as follows:
```
legion2002: Noting that this encoding is exactly the same as what we use for tempo transactions. Which means that you can also verify tx hashes onchain. 
This is desirable, and we want to maintain this 1:1 sig compatibility going forward.

tips/tip-xxxx.md
```diff
@@ -0,0 +1,175 @@
...
+This precompile exposes Tempo's native signature verification to contracts.
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119.
+
+### Precompile Address
+
+```sh
+0x5165300000000000000000000000000000000000 # SIG(5[natur])E
```
brendanjryan: nice

tips/tip-1020.md
```diff
@@ -0,0 +1,112 @@
...
+### Gas Costs
+
+Gas costs follow the [Tempo Transaction Signature Gas Schedule](https://docs.tempo.xyz/protocol/transactions/spec-tempo-transaction#signature-verification-gas-schedule):
+
+| Type | Gas Cost |
+|------|----------|
+| secp256k1 | 3,000 |
+| P256 | 8,000 |
+| WebAuthn | 8,000 |
+| Keychain | Inner signature cost + 3,000 |
```
0xKitsune, line 83: Can we get an ack from @danrobinson or @dankrad on these values?

0xKitsune (approved):

tips/tip-1020.md
danrobinson: Benchmarks indicate that P256 and WebAuthn are just under 1.5x as much time as secp256k1, so I think the right gas cost for each of those is 4,500.

legion2002: The gas values have been taken to match the tempo transaction spec. The justification for those initial values is provided [here](https://docs.tempo.xyz/protocol/transactions/spec-tempo-transaction#signature-verification-gas-schedule)

tips/tip-1020.md
```diff
@@ -0,0 +1,112 @@
...
+```solidity
+interface ISignatureVerifier {
+    error InvalidSignature();
+
+    /// @notice Verifies a Tempo signature
+    /// @param signer The expected signer address
+    /// @param hash The message hash that was signed
+    /// @param signature The encoded signature (see Tempo Transaction spec for formats)
+    /// @return True if valid, reverts otherwise
+    function verify(address signer, bytes32 hash, bytes calldata signature) external view returns (bool);
```
howydev, line 44: If we're covering keychain cases, I think we'd want more granularity/info in the return to separate between keychain vs owner cases. Maybe `address` instead?  Usage for smart contracts that allow owners but not access keys to be something like:

```
address key = precompile.verify(owner, ...);
if (key != owner) {
  revert onlyOwner();
}
```

PR #2384, opened 2026-02-02
TIP-1017: Subblocks v2 - Builder Delegation

malleshpai:
## Summary

**TIP-1017** redesigns the subblock system to decouple builder identity from validator identity, based on customer feedback and internal review.

### Key Changes

1. **Many-to-one builder delegation**: Validators call `delegateToBuilder(builder)` to delegate their gas budget. Multiple validators can share the same builder, who accumulates their combined gas allocation.

2. **ECDSA signing**: Builders sign subblocks with their Ethereum key (standard ECDSA) instead of ed25519, since subblock signing is not part of consensus.

3. **Builder-controlled config**: Builders call `setBuilderConfig(feeRecipient, feeToken)` to configure where fees go and their preferred fee token.

4. **New nonce key format**: `0x5b | builder_address (20 bytes) | nonce (11 bytes)` — builder identity is embedded directly in the transaction.

5. **Expiring nonces in subblocks**: When nonce = `type(uint88).max`, the transaction uses TIP-1009 expiring nonce semantics for replay protection.

### What's NOT included (simplified from earlier draft)

- **No fallback 0x5b transactions**: Dropped for simplicity. If builder is offline, txs wait.
- **No fee_recipient uniqueness constraint**: Multiple builders can share a fee recipient.

### Design Decisions

| Decision | Rationale |
|----------|-----------|
| ECDSA not ed25519 | Subblock signing is not consensus-critical |
| Many-to-one delegation | Allows shared builder infrastructure |
| gasWeight tracking | Explicit counter for efficient gas budget computation |
| Root-only tx signing | Prevents race condition with keychain revocation |
| Builder sets config | Decouples fee routing from validator identity |

See [tips/tip-1017.md](https://github.com/tempoxyz/tempo/blob/TIP-1017/tips/tip-1017.md) for full spec.

### DIFF
--- a/tips/tip-1011.md
+++ b/tips/tip-1011.md
@@ -617,6 +617,31 @@ Validators deduplicate received subblocks by `(builder, parent_hash)`. Only the 
 
 ---
 
+# Design Rationale
+
+## Root-Only Signing for Subblock Transactions
+
+Subblock transactions (0x5b prefix) require root EOA signatures and reject keychain (access key) signatures. This is intentional due to a race condition with key revocation:
+
+1. User signs a 0x5b tx with an access key, submits to builder
+2. Builder validates signature against current keychain state, includes in subblock
+3. In the same block, a main block tx revokes that access key
+4. Block execution order: main txs execute first, then subblock txs
+5. When the subblock tx executes, the key is already revoked
+
+The problem: signature validation happens at subblock validation time (before block building), not during execution. The subblock tx would still execute even though the key was revoked in the same block.
+
+This violates user expectations. When a user revokes an access key, they expect immediate effect. Allowing keychain signatures for subblock txs would create a window where revoked keys can still authorize transactions.
+
+**Alternative considered**: Re-validate keychain signatures during execution. This was rejected because:
+- Adds execution-time overhead for every subblock tx
+- Creates griefing vector: users could intentionally revoke keys to cause builder's subblock to fail
+- Complicates the execution model
+
+Root-only signing is the simplest solution that maintains user expectations around key revocation.
+
+---
+
 # Implementation Notes
 
 ## Relationship to Current Implementation


tips/tip-1011.md
```diff
@@ -0,0 +1,585 @@
...
+status: Draft
+related: Subblock Specification
+protocolVersion: TBD
+---
+
+# TIP-xxxx: Subblocks v2 - Builder Delegation and Fallback Inclusion
+
+## Abstract
+
+This TIP introduces three major changes to the subblock system:
```
danrobinson: three changes

tips/tip-1011.md
```diff
@@ -0,0 +1,585 @@
...
+## 1. Subblock Registry
+
+A new Subblock Registry precompile stores builder configurations designated by validators.
+
+### 1.1 Data Structures
+
+```solidity
+struct SubblockBuilder {
+    bytes32 signingKey;      // ed25519 public key for signing subblocks
+    address feeRecipient;    // receives fees, must be unique across all builders
```
danrobinson: I think we should allow multiple builders to delegate to the same subblock builder, and it tracks their weight

tips/tip-1011.md
```diff
@@ -0,0 +1,585 @@
...
+
+## 1. Subblock Registry
+
+A new Subblock Registry precompile stores builder configurations designated by validators.
+
+### 1.1 Data Structures
+
+```solidity
+struct SubblockBuilder {
+    bytes32 signingKey;      // ed25519 public key for signing subblocks
```
danrobinson: Why is this a bytes32 rather than an address?

tips/tip-1011.md
```diff
@@ -0,0 +1,585 @@
...
+2. Extract `feeRecipient` from nonce key
+3. Look up `builder = SubblockRegistry.builderOf(validatorFor(feeRecipient))`
+4. If registered and `builder.feeToken` differs, attempt to swap via Fee AMM if liquidity available
+5. If no liquidity or unregistered, fees are paid directly in the user's resolved fee token
+
+### 1.2 Storage
+
+```solidity
+// Validator → their designated builder configuration
+mapping(address => SubblockBuilder) public builders;
```
danrobinson: Why is this `address`?

tips/tip-1011.md
```diff
@@ -0,0 +1,585 @@
...
+```solidity
+interface ISubblockRegistry {
+    /// @notice Returns the builder configuration for a validator.
+    /// @param validator The validator address.
+    /// @return The SubblockBuilder struct, or zero values if not set.
+    function builderOf(address validator) external view returns (SubblockBuilder memory);
+
+    /// @notice Returns the ed25519 signing key for a given fee recipient.
+    /// @param feeRecipient The fee recipient address.
+    /// @return The 32-byte ed25519 public key.
```
danrobinson: Definitely confused why feeRecipient would be an ed25519—"fee recipient" sounds very much like an EVM address

tips/tip-1011.md
```diff
@@ -0,0 +1,585 @@
...
+protocolVersion: TBD
+---
+
+# TIP-xxxx: Subblocks v2 - Builder Delegation and Fallback Inclusion
+
+## Abstract
+
+This TIP introduces three major changes to the subblock system:
+
+1. **Subblock Registry**: Validators can delegate subblock production to a separate builder identity with its own signing key, enabling operational separation between consensus participation and subblock building.
```
danrobinson: As discussed—try to have the validator delegate just to a subblock builder (not to a "fee recipient"), and allow many-to-one in that mapping. And probably for now no fee recipient mapping from there

tips/tip-1011.md
```diff
@@ -0,0 +1,585 @@
...
+
+2. **Fee Recipient in Nonce Key**: The nonce key format for subblock transactions changes from `0x5b | validator_pubkey_prefix | nonce` to `0x5b | fee_recipient | nonce`, embedding the fee recipient directly in the transaction.
+
+3. **Fallback 0x5b Transactions**: Allows 0x5b transactions to be included outside signed subblocks when a builder is unavailable or subblocks are full, with fees routed using the embedded fee recipient.
+
```
danrobinson, line 24: Add one more change: allow expiring-nonce transactions into subblocks by changing it to a check that the last 11 bytes of the nonce is 0xFFFFFFF... (thus allowing bytes 0-21 to be 0x5b[subblock_builder])

tips/tip-1011.md
```diff
@@ -0,0 +1,585 @@
...
+## Abstract
+
+This TIP introduces three major changes to the subblock system:
+
+1. **Subblock Registry**: Validators can delegate subblock production to a separate builder identity with its own signing key, enabling operational separation between consensus participation and subblock building.
+
+2. **Fee Recipient in Nonce Key**: The nonce key format for subblock transactions changes from `0x5b | validator_pubkey_prefix | nonce` to `0x5b | fee_recipient | nonce`, embedding the fee recipient directly in the transaction.
+
+3. **Fallback 0x5b Transactions**: Allows 0x5b transactions to be included outside signed subblocks when a builder is unavailable or subblocks are full, with fees routed using the embedded fee recipient.
```
danrobinson: Consider dropping fallback transactions

PR #2389, opened 2026-02-02
fix(TIP-1015): allow cancelling DEX orders with invalid policy ids

0xrusowsky:
ref: `CHAIN-586`

### DIFF
--- a/crates/precompiles/src/stablecoin_dex/mod.rs
+++ b/crates/precompiles/src/stablecoin_dex/mod.rs
@@ -1061,11 +1061,12 @@ impl StablecoinDEX {
         };
 
         let policy_id = TIP20Token::from_address(token)?.transfer_policy_id()?;
-        if TIP403Registry::new().is_authorized_as(policy_id, order.maker(), AuthRole::sender())? {
-            return Err(StablecoinDEXError::order_not_stale().into());
+        // Invalid policy ids throw under_overflow. Treat as unauthorized to clear the orders.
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -1061,11 +1061,12 @@ impl StablecoinDEX {
...
 
         let policy_id = TIP20Token::from_address(token)?.transfer_policy_id()?;
-        if TIP403Registry::new().is_authorized_as(policy_id, order.maker(), AuthRole::sender())? {
-            return Err(StablecoinDEXError::order_not_stale().into());
+        // Invalid policy ids throw under_overflow. Treat as unauthorized to clear the orders.
+        match TIP403Registry::new().is_authorized_as(policy_id, order.maker(), AuthRole::sender()) {
+            Ok(true) => Err(StablecoinDEXError::order_not_stale().into()),
+            Err(e) if e != TempoPrecompileError::under_overflow() => Err(e),
+            _ => self.cancel_active_order(order),
         }
```
0xKitsune, line 1069: In general the change looks ok, as mentioned in a separate PR the `under_overflow()` seems like incorrect semantics and we should consider updating this. Would it be incorrect to return `false` from `is_authorized_as()` instead in this case?

Maybe we could add some comments to make this a bit easier to reason about, its not immediately clear when looking at `cancel_stale_order()` when `under_overflow()` would be hit and why we ignore this.

PR #2390, opened 2026-02-02
fix(tip1015): inexistent compound policy error clarity

0xrusowsky:
ref: `CHAIN-580`

### DIFF
--- a/crates/precompiles/src/tip403_registry/mod.rs
+++ b/crates/precompiles/src/tip403_registry/mod.rs
@@ -1791,9 +1791,8 @@ mod tests {
             let mut registry = TIP403Registry::new();
 
             // Non-existent policy should return PolicyNotFound
-            let result = registry.compound_policy_data(ITIP403Registry::compoundPolicyDataCall {
-                policyId: 999,
-            });
+            let result = registry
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -151,7 +151,15 @@ impl TIP403Registry {
 
         // Only compound policies have compound data
         if !data.is_compound() {
-            return Err(TIP403RegistryError::incompatible_policy_type().into());
+            // Check if the policy exists for error clarity
+            let err = if self.policy_exists(ITIP403Registry::policyExistsCall {
```
legion2002, line 155: should we just make this a generic policy error, instead of checking for more granularity onchain? It should be pretty easy to figure out offchain what the actual error is

0xrusowsky: i'd either:
- leave it as is (on-chain check out of the happy path to only do the extra SLOAD when we know we error)
- ACK the inconsistency `policyData -> PolicyNotFound` vs `compoundPolicyData -> IncompatiblePolicyType` for perf reasons

PR #2392, opened 2026-02-02
fix(txpool): skip unlock check for expiring nonce txs in best iterator

fgimenez:
Closes CHAIN-573

This PR skips unnecessary unlock checks for expiring nonce transactions in the best transaction iterator.

Expiring nonce transactions (nonce_key == U256::MAX) use tx hash for replay protection instead of sequential nonces, making them always independent and immediately executable. The best transaction iterator was incorrectly performing unlock logic for these transactions, which is wasteful since they never have nonce dependencies.

- Add `is_expiring_nonce()` helper to `AASequenceId`
- Skip unlock check in `next_tx_and_priority` when the transaction uses expiring nonces

### DIFF
--- a/crates/transaction-pool/src/tt_2d_pool.rs
+++ b/crates/transaction-pool/src/tt_2d_pool.rs
@@ -1488,7 +1488,10 @@ impl BestAA2dTransactions {
                 continue;
             }
             // Advance transaction that just got unlocked, if any.
-            if let Some(unlocked) = self.by_id.get(&id.unlocks()) {
+            // Skip for expiring nonce transactions as they are always independent.
+            if !id.seq_id.is_expiring_nonce()
+                && let Some(unlocked) = self.by_id.get(&id.unlocks())
+            {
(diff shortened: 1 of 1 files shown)


crates/transaction-pool/src/tt_2d_pool.rs
```diff
@@ -1539,6 +1542,14 @@ impl AASequenceId {
         Self { address, nonce_key }
     }
 
+    /// Returns `true` if this sequence ID represents an expiring nonce transaction.
+    ///
+    /// Expiring nonce transactions use `nonce_key == U256::MAX` and are always independent,
+    /// meaning they don't have sequential nonce dependencies.
+    pub(crate) fn is_expiring_nonce(&self) -> bool {
+        self.nonce_key == U256::MAX
+    }
```
mattsse, line 1551: this makes sense, but since this is static, we should just make this a bool in tempopooledtx instead

PR #2396, opened 2026-02-02
fix(tip-1015): ensure SLOAD number is preserved pre-T1

0xrusowsky:
ref: `CHAIN-583`

### DIFF
--- a/crates/precompiles/src/tip403_registry/mod.rs
+++ b/crates/precompiles/src/tip403_registry/mod.rs
@@ -266,9 +266,9 @@ impl TIP403Registry {
                         },
                     ))?;
                 }
-                PolicyType::COMPOUND | PolicyType::__Invalid => {
-                    // Pre-T1: no events emitted for invalid types, accounts still added
-                    // T1+: unreachable since `ensure_is_simple()` already rejected
+                ITIP403Registry::PolicyType::COMPOUND | ITIP403Registry::PolicyType::__Invalid => {
+                    // T1+: unreachable since `validate_simple_policy_type` already rejected
+                    return Err(TIP403RegistryError::incompatible_policy_type().into());
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -85,9 +88,28 @@ impl PolicyData {
...
+    /// Decodes the raw `policy_type` u8 to a `PolicyType` enum.
+    ///
+    /// - **T1+**: fails on invalid values (early revert)
+    /// - **Pre-T1**: returns `__Invalid` for unknown values (defer error to caller)
+    fn policy_type(&self) -> Result<PolicyType> {
+        let is_t1 = StorageCtx.spec().is_t1();
+        match self.policy_type.try_into() {
+            Err(_) => {
+                if is_t1 {
+                    Err(TempoPrecompileError::under_overflow())
```
0xKitsune, line 100: Is there a better error for this? Technically this is not under/overflow, but rather `InvalidPolicy` or similar. IIUC if the policy type does not map to `PolicyType` this should be handled similar to how the sol macro turns this into `__Invalid`. In this case should we return a corresponding `Invalid` error?

0xrusowsky: we could do that yes, but it would have to be gated under T1. should i go ahead and do it?

0xKitsune: I think this would make sense, then for `policy_type()` we could make this function infallible and rewrite to:

```rust

fn policy_type(&self) -> PolicyType {
        match self.policy_type.try_into() {
            Err(_) => PolicyType::__Invalid,
            Ok(PolicyType::COMPOUND) if !is_t1 => Ok(PolicyType::__Invalid),
            Ok(policy_type) => Ok(policy_type),
        }
    }
```

Then we could handle any `__Invalid` post t1 with an error. Lmk what you thnk, also open to keeping as is or going with an alternative.

0xrusowsky: the point of making it fallible was being able to short-circuit invalid policies, but i guess it is so unlikely that it is better to optimize for clean code rather than efficient code.

will do

legion2002 (approved):
lgtm, nice abstraction to fix it.

PR #2397, opened 2026-02-02
fix(revm): charge new_account_cost for CREATE with 2D nonce when account.nonce=0

rakita:
Closes CHAIN-578
Closes CHAIN-577

## Summary
Fix gas calculation for CREATE transactions with 2D nonce (nonce_key != 0) when the sender's account nonce is 0. This edge case requires an additional 250k gas for account creation that was previously not charged.

## Changes
- **handler.rs**: Added logic to charge additional `new_account_cost` (250k gas) when transaction has 2D nonce, is CREATE, and caller's account nonce is 0. Re-validates gas limit after adding the additional cost.
- **validator.rs**: Updated `ensure_aa_intrinsic_gas` to check the same condition during pool validation using state provider.
- **evm.rs**: Added test verifying the 250k gas difference between first and second CREATE with 2D nonce.

## Test plan
- [x] Run existing tests: `cargo test -p tempo-revm`
- [x] Run pool validator tests: `cargo test -p tempo-transaction-pool`
- [x] New tests verify gas difference is exactly 250,000

### DIFF
--- a/crates/transaction-pool/src/tt_2d_pool.rs
+++ b/crates/transaction-pool/src/tt_2d_pool.rs
@@ -1488,7 +1488,10 @@ impl BestAA2dTransactions {
                 continue;
             }
             // Advance transaction that just got unlocked, if any.
-            if let Some(unlocked) = self.by_id.get(&id.unlocks()) {
+            // Skip for expiring nonce transactions as they are always independent.
+            if !id.seq_id.is_expiring_nonce()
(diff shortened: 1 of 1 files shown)


legion2002 (approved):
This lgtm, also good note from @0xKitsune 's commit, that the same 250k limit also applies to expiring nonces

0xKitsune (approved):

PR #2399, opened 2026-02-02
fix(tip-1015): DEX transfer auth checks

0xrusowsky:
ref: `CHAIN-585`

### DIFF
--- a/crates/precompiles/src/stablecoin_dex/mod.rs
+++ b/crates/precompiles/src/stablecoin_dex/mod.rs
@@ -1369,6 +1369,7 @@ impl StablecoinDEX {
 #[cfg(test)]
 mod tests {
     use alloy::primitives::IntoLogData;
+    use tempo_chainspec::hardfork::TempoHardfork;
     use tempo_contracts::precompiles::TIP20Error;
 
     use crate::{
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -402,8 +402,9 @@ impl StablecoinDEX {
         };
 
         // Check policy on non-escrow token (escrow token is checked in decrement_balance_or_transfer_from)
+        // Direction: DEX → sender (order placer receives non-escrow token when filled)
         TIP20Token::from_address(non_escrow_token)?
-            .ensure_transfer_authorized(sender, self.address)?;
+            .ensure_transfer_authorized(self.address, sender)?;
```
0xKitsune, line 407: On order placement, tokens are being sent from the sender to the DEX. Why is this order being flipped?

0xrusowsky: iiuic the auditor's point is about non-escrow tokens.

- the escrow token (which is sent from sender to DEX at placement) is checked in `decrement_balance_or_transfer_from`.
- this check is for the non-escrow token, which is the token the maker will receive when their order is filled.

PR #2401, opened 2026-02-02
fix(evm): charge TIP-1000 new account cost for expiring nonce CREATE when protocol nonce is 0

fgimenez:
Closes CHAIN-577

When an AA transaction uses expiring nonce for a CREATE operation, the CREATE execution bumps the protocol nonce from 0 to 1. However, the gas calculation wasn't accounting for this—it only checked the nonce key type, not the underlying protocol nonce state. This meant these transactions avoided the 250k gas charge intended to prevent state bloat.

This PR adds a check in the expiring nonce gas calculation branch: if it's a CREATE transaction AND the caller's protocol nonce is 0, charge the `new_account_cost` (250k gas) in addition to the expiring nonce gas.

Applied to:
- `crates/revm/src/handler.rs` (execution)
- `crates/transaction-pool/src/validator.rs` (pool validation)


rakita:
This is a duplicate of https://github.com/tempoxyz/tempo/pull/2397 that covers expiring nonce

fgimenez:
closing in favor of https://github.com/tempoxyz/tempo/pull/2397

PR #2402, opened 2026-02-02
fix(tip-1015): don't short-circuit errors pre-T1

0xrusowsky:
This PR resolves re-execution.

fixes 2 code instances where an SLOAD was omitted (pre-T1) due to an optimization caused by an early return on execution error.

### DIFF
--- a/crates/precompiles/src/tip20/mod.rs
+++ b/crates/precompiles/src/tip20/mod.rs
@@ -680,7 +680,10 @@ impl TIP20Token {
         let policy_id = self.transfer_policy_id()?;
         let registry = TIP403Registry::new();
 
-        if !registry.is_authorized_as(policy_id, from, AuthRole::sender())? {
+        // (spec: +T1) short-circuit and skip recipient check if sender fails
+        if !registry.is_authorized_as(policy_id, from, AuthRole::sender())?
+            && self.storage.spec().is_t1()
+        {
             return Ok(false);
         }
         registry.is_authorized_as(policy_id, to, AuthRole::recipient())
--- a/crates/precompiles/src/tip403_registry/mod.rs
+++ b/crates/precompiles/src/tip403_registry/mod.rs
@@ -448,10 +448,10 @@ impl TIP403Registry {
                     self.is_authorized_simple_policy(compound.mint_recipient_policy_id, user)
                 }
                 AuthRole::Transfer => {
+                    // (spec: +T1) short-circuit and skip recipient check if sender fails
                     if !self.is_authorized_simple_policy(compound.sender_policy_id, user)?
                         && self.storage.spec().is_t1()
                     {
-                        // (spec: +T1) short-circuit and skip recipient check if sender fails
                         return Ok(false);
                     }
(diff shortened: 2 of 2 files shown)


crates/precompiles/src/tip20/mod.rs
```diff
@@ -680,7 +680,10 @@ impl TIP20Token {
         let policy_id = self.transfer_policy_id()?;
         let registry = TIP403Registry::new();
 
-        if !registry.is_authorized_as(policy_id, from, AuthRole::sender())? {
+        // (spec: +T1) short-circuit and skip recipient check if sender fails
+        if !registry.is_authorized_as(policy_id, from, AuthRole::sender())?
+            && self.storage.spec().is_t1()
+        {
```
0xKitsune, line 686: nit: consider flipping these conditions, it is less expensive to check `spec().is_t1()` first.

crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -448,8 +448,10 @@ impl TIP403Registry {
                     self.is_authorized_simple_policy(compound.mint_recipient_policy_id, user)
                 }
                 AuthRole::Transfer => {
-                    if !self.is_authorized_simple_policy(compound.sender_policy_id, user)? {
-                        // Short-circuit: if sender fails, skip recipient check
+                    // (spec: +T1) short-circuit and skip recipient check if sender fails
+                    if !self.is_authorized_simple_policy(compound.sender_policy_id, user)?
+                        && self.storage.spec().is_t1()
```
0xKitsune, line 453: Same thing here, consider flipping these to check `is_t1()` first.

crates/precompiles/src/tip20/mod.rs
0xrusowsky: if i place it first it wouldn't run the check pre-T1

so it's either what we have now or:
```suggestion
        let sender_auth = registry.is_authorized_as(policy_id, from, AuthRole::sender())?;
        // (spec: +T1) short-circuit and skip recipient check if sender fails
        if self.storage.spec().is_t1() && !sender_auth {
```

feel free to commit this cmnt suggestion if u prefer this

crates/precompiles/src/tip403_registry/mod.rs
0xrusowsky: same here

either what we have now or:
```suggestion
                    let simple_auth = self.is_authorized_simple_policy(compound.sender_policy_id, user)?;
                    // (spec: +T1) short-circuit and skip recipient check if sender fails
                    if self.storage.spec().is_t1() && !simple_auth
```
feel free to commit this cmnt suggestion if u prefer this

crates/precompiles/src/tip20/mod.rs
0xKitsune: Ahh sounds good, disregard.

0xKitsune (approved):

Dustinturner44 (commented):
#2344

PR #2404, opened 2026-02-02
chore(deps): bump commonware to a1b60388c0bbc9126aa9934ec6ebade6ec344552

SuperFluffy:
Upgrades commonware dependencies to https://github.com/commonwarexyz/monorepo/tree/a1b60388c0bbc9126aa9934ec6ebade6ec344552. This is to be closer to the version that commonware tags as 0.1.0 and to get access to https://github.com/commonwarexyz/monorepo/pull/2992 and https://github.com/commonwarexyz/monorepo/pull/2987

### DIFF
--- a/crates/precompiles/src/storage/evm.rs
+++ b/crates/precompiles/src/storage/evm.rs
@@ -3,7 +3,6 @@ use alloy_evm::{EvmInternals, EvmInternalsError};
 use revm::{
     context::{Block, CfgEnv},
     context_interface::cfg::{GasParams, gas},
-    interpreter::instructions::WARM_STORAGE_READ_COST,
     state::{AccountInfo, Bytecode},
 };
 use tempo_chainspec::hardfork::TempoHardfork;
@@ -98,7 +97,10 @@ impl<'a> PrecompileStorageProvider for EvmPrecompileStorageProvider<'a> {
             .load_account_mut_skip_cold_load(address, false)?;
 
         // TODO(rakita) can be moved to the beginning of the function. Requires fork.
-        deduct_gas(&mut self.gas_remaining, WARM_STORAGE_READ_COST)?;
+        deduct_gas(
+            &mut self.gas_remaining,
+            self.gas_params.warm_storage_read_cost(),
+        )?;
 
         // dynamic gas
         if account.is_cold {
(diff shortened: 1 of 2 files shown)


crates/commonware-node/src/subblocks.rs
```diff
@@ -315,17 +315,10 @@ impl<TContext: Spawner + Metrics + Pacer> Actor<TContext> {
             Round::new(epoch_of_next_block, View::new(1))
         };
 
-        let seed_signature = if next_round.view() == View::new(1) {
```
SuperFluffy, line 318: This is not necessary: certificates don't have a seed on view = 1, and the case of `seed_signature = None` is handled by `Random::select_leader`.

crates/e2e/src/tests/consensus_rpc.rs
```diff
@@ -261,16 +258,14 @@ fn get_identity_transition_proof_after_full_dkg() {
         )
         .expect("valid finalization");
 
-        // Verify the certificate was signed by the old network identity
-        let namespace = b"TEMPO_FINALIZE";
-        let message = finalization.proposal.encode();
-        verify_message::<MinSig>(
```
SuperFluffy, line 267: This was a bit too low level. We should go through the verifier so we don't need to worry about how the signature namespace is constructed.

crates/commonware-node-config/src/lib.rs
```diff
@@ -15,9 +15,8 @@ use commonware_cryptography::{
 #[cfg(test)]
 mod tests;
 
-#[derive(Clone, PartialEq, Eq, derive_more::Debug)]
+#[derive(Clone, Debug)]
 pub struct SigningKey {
-    #[debug(skip)]
```
joshieDo, line 20: why rm, is it skipped on cw side

crates/commonware-node/src/subblocks.rs
joshieDo: ack

crates/commonware-node-config/src/lib.rs
SuperFluffy: For cleanup. CW has fixed their debug logging and is no longer revealing secrets. That was the only reason we had this.

PR #2405, opened 2026-02-02
feat: parameterized network config

### LINKED ISSUE #2403: Allow parameterizing consensus p2p settings
### Describe the feature

We should have reasonable default settings for our current small peer world (similar to local or recommended), but also allow overriding these settings.

Right now we only have consensus.use-local-p2p-defaults, which are more aggressive p2p settings intended for running a local p2p stack [1]. For the default, we use commonware's recommended settings, which however assume a network of 1000 nodes (so they don't end up spamming each other with connection requests).

All of this ends up here [2], where we just toggle between commonware's recommended settings, or this aggressive local default.

But as they pointed out during our call, for small networks we probably want to be more aggressive and a) define tighter defaults, but also b) allow setting them through config args.

This also extends to rate-limiting when the marshal actor is trying to backfill blocks. [3]

1: https://github.com/tempoxyz/tempo/blob/aa62ce2adbab749dc3d0bef559865a886dbb9c8e/crates/commonware-node/src/args.rs#L120-L124
2: https://github.com/tempoxyz/tempo/blob/aa62ce2adbab749dc3d0bef559865a886dbb9c8e/crates/commonware-node/src/lib.rs#L193-L197
3: https://github.com/tempoxyz/tempo/blob/aa62ce2adbab749dc3d0bef559865a886dbb9c8e/crates/commonware-node/src/config.rs#L33

### Additional context

_No response_

hamdiallam:
Closes #2403

Added flag parameters for consensus p2p arguments that are most critical between large/small validator sets. These settings are most tuned for rate of communication between peers

```
--consensus.allow-private-ips
--consensus.allow-dns
--consensus.synchrony-bound
--consensus.wait-before-peers-redial
--consensus.wait-before-peers-reping
--consensus.wait-before-peers-discovery
--consensus.connection-per-per-min-period
--consensus.handshake-stale-after (renamed from max-handshake-age)
--consensus.max-concurrent-handshakes
--consensus.handshake-per-ip-min-period
--consensus.handshake-per-subnet-min-period
--consensus.handshake-timeout
--consensus.backfill_frequency
--consensus.time-to-unblock-byzantine-peer (renamed from block-duration)
```

### DIFF
--- a/crates/commonware-node/src/args.rs
+++ b/crates/commonware-node/src/args.rs
@@ -139,7 +139,8 @@ pub struct Args {
     #[arg(long = "consensus.synchrony-bound", default_value = "5s")]
     pub synchrony_bound: PositiveDuration,
 
-    /// How long to wait before attempting to dial any peers without a connection..
+    /// How long to wait before attempting to dial peers. Run across all peers
+    /// including the newly discovered ones.
     #[arg(
         long = "consensus.wait-before-peers-redial",
         default_value = "1s",
@@ -155,7 +156,7 @@ pub struct Args {
     )]
     pub wait_before_peers_reping: PositiveDuration,
 
-    /// How often to query for new dialable peers. Also limits re-dial rate per peer.
+    /// How often to query for new dialable peers.
     #[arg(
         long = "consensus.wait-before-peers-discovery",
         default_value = "60s",
@@ -163,7 +164,8 @@ pub struct Args {
     )]
     pub wait_before_peers_discovery: PositiveDuration,
 
-    /// Minimum time between connection attempts to the same peer.
+    /// Minimum time between connection attempts to the same peer. A rate-limit
+    /// on connection attempts.
     #[arg(
         long = "consensus.connection-per-peer-min-period",
         default_value = "60s",
@@ -171,7 +173,8 @@ pub struct Args {
     )]
     pub connection_per_peer_min_period: PositiveDuration,
 
-    /// Minimum time between handshake attempts from a single IP address.
+    /// Minimum time between handshake attempts from a single IP address. A rate-limit
+    /// on attempts.
     #[arg(
         long = "consensus.handshake-per-ip-min-period",
         default_value = "5s",
@@ -179,7 +182,8 @@ pub struct Args {
     )]
     pub handshake_per_ip_min_period: PositiveDuration,
 
-    /// Minimum time between handshake attempts from a single subnet.
+    /// Minimum time between handshake attempts from a single subnet. A rate-limit
+    /// on attempts.
     #[arg(
         long = "consensus.handshake-per-subnet-min-period",
         default_value = "15ms",


crates/commonware-node/src/args.rs
SuperFluffy, line 1: All of the values here should not be `Option` but get as defaults the values that node would use prior to this change.

crates/commonware-node/src/args.rs
```diff
@@ -120,7 +120,37 @@ pub struct Args {
     /// Use P2P defaults optimized for local network environments.
     /// Only enable in non-production network nodes.
     #[arg(long = "consensus.use-local-p2p-defaults", default_value_t = false)]
-    pub use_local_defaults: bool,
+    pub use_local_p2p_defaults: bool,
+
+    /// How often to attempt dialing peers. Lower values mean faster peer discovery.
+    /// Recommended: 1s, Local: 500ms
+    #[arg(long = "consensus.p2p-dial-frequency")]
+    pub p2p_dial_frequency: Option<jiff::SignedDuration>,
```
SuperFluffy: For consistency with other arguments, don't use `frequency` but `interval`. This applies to all fields here.

(as a sidenote, I am not a fan of commonware nomenclature here, because frequency has unit of 1/<time>, but consistency is a stronger argument).

crates/commonware-node/src/args.rs
```diff
@@ -120,7 +120,37 @@ pub struct Args {
...
+    pub p2p_connection_min_period: Option<jiff::SignedDuration>,
+
+    /// Minimum time between handshake attempts from a single IP address.
+    /// Recommended: 5s (1 per 5s), Local: 1s (1/sec)
+    #[arg(long = "consensus.p2p-handshake-per-ip-min-period")]
+    pub p2p_handshake_per_ip_min_period: Option<jiff::SignedDuration>,
+
+    /// Rate limit for the marshal actor when backfilling (messages per second).
+    /// Default: 8
+    #[arg(long = "consensus.marshal-rate-per-sec")]
```
SuperFluffy: Let's not expose `marshal` here. Users should not care about implementation deals.

Let's just call this `backfill_frequency` (here, frequency actually makes sense :-)).

crates/commonware-node/src/args.rs
```diff
@@ -120,7 +120,37 @@ pub struct Args {
...
+    pub marshal_rate_per_sec: Option<u32>,
```
SuperFluffy:
```suggestion
    backfill_frequency: NonZeroU32,
```

crates/commonware-node/src/lib.rs
```diff
@@ -174,14 +224,31 @@ pub async fn run_consensus_stack(
...
+    pub dial_frequency: Option<Duration>,
+    pub query_frequency: Option<Duration>,
+    pub ping_frequency: Option<Duration>,
+    pub connection_min_period: Option<Duration>,
+    pub handshake_per_ip_min_period: Option<Duration>,
+}
+
+/// Configuration for instantiating the P2P network.
+#[derive(Debug, Clone)]
+pub struct NetworkConfig {
```
SuperFluffy: IMO unnecessary. You can fold most of these into the consensus args above and then just `#[clap(flatten)]` this into `ConsensusArgs`.

crates/commonware-node/src/lib.rs
```diff
@@ -190,16 +257,54 @@ async fn instantiate_network(
...
+        .connection_min_period
+        .and_then(commonware_runtime::Quota::with_period)
+        .unwrap_or(default_config.allowed_connection_rate_per_peer);
+    let handshake_per_ip_min_period_quota = config
+        .p2p_overrides
+        .handshake_per_ip_min_period
+        .and_then(commonware_runtime::Quota::with_period)
+        .unwrap_or(default_config.allowed_handshake_rate_per_ip);
+
     let p2p_cfg = lookup::Config {
```
SuperFluffy: While we are at it, let's make all of these configurable?

crates/commonware-node/src/args.rs
```diff
@@ -120,7 +120,37 @@ pub struct Args {
     /// Use P2P defaults optimized for local network environments.
     /// Only enable in non-production network nodes.
     #[arg(long = "consensus.use-local-p2p-defaults", default_value_t = false)]
-    pub use_local_defaults: bool,
+    pub use_local_p2p_defaults: bool,
```
SuperFluffy: Move the logic for setting this into Clap: if this value is set, all other values can take the local values. Use https://docs.rs/clap/latest/clap/struct.Arg.html#method.default_value_if

hamdiallam: This is an awesome feature

crates/commonware-node/src/args.rs
```diff
@@ -35,35 +35,34 @@ pub struct Args {
     #[arg(long = "consensus.max-message-size-bytes", default_value_t = DEFAULT_MAX_MESSAGE_SIZE_BYTES)]
     pub max_message_size_bytes: u32,
 
-    // pub storage_directory: camino::Utf8PathBuf,
     /// The number of worker threads assigned to consensus.
     #[arg(long = "consensus.worker-threads", default_value_t = 3)]
     pub worker_threads: usize,
 
-    /// The maximum number of messages that can be cute on the various consensus
```
SuperFluffy, line 43: That is a hilarious typo.

crates/commonware-node/src/args.rs
```diff
@@ -110,17 +109,128 @@ pub struct Args {
...
+    /// [default: false, local: true]
+    #[arg(
+        long = "consensus.allow-private-ips",
+        default_value_t = false,
+        default_value_if("use_local_defaults", "true", "true")
+    )]
+    pub allow_private_ips: bool,
+
+    /// Whether to allow DNS-based ingress addresses.
+    /// [default: true]
```
SuperFluffy: This applies here and in all other instances: don't need to document the default values because clap will do it itself.

(I think this does not apply to `default_value_if` though, unfortunately)

crates/commonware-node/src/args.rs
```diff
@@ -110,17 +109,128 @@ pub struct Args {
...
+    #[arg(
+        long = "consensus.handshake-max-concurrent",
+        default_value = "512",
+        default_value_if("use_local_defaults", "true", "1024")
+    )]
+    pub handshake_max_concurrent: std::num::NonZeroU32,
+
+    /// Number of epoch peer sets to maintain connections for.
+    /// [default: 4]
+    #[arg(long = "consensus.peer-set-epoch-depth", default_value_t = 4)]
```
SuperFluffy: Very realistic chance that we will decouple this from the epoch to make it possible to update peers on every change of the on-chain validator set. Let's remove this for now until we figure out how we want to do that.

crates/commonware-node/src/args.rs
```diff
@@ -110,17 +109,128 @@ pub struct Args {
...
+
+    /// Whether to allow DNS-based ingress addresses.
+    /// [default: true]
+    #[arg(long = "consensus.allow-dns", default_value_t = true)]
+    pub allow_dns: bool,
+
+    /// Time into the future that a timestamp can be and still be considered valid.
+    /// [default: 5s]
+    #[arg(long = "consensus.synchrony-bound", default_value = "5s")]
+    pub synchrony_bound: jiff::SignedDuration,
```
SuperFluffy: These are becoming unmanageable because we need to manually convert them to Durations all the time.

Can you provide a manual new-type wrapper `Duration(jiff::SignedDuration)` that impls `FromStr`? Then we don't need to do the convert-report-error-dance all the time (and also give better feedback to the user that they supplied a wrong value).

crates/commonware-node/src/args.rs
```diff
@@ -110,17 +109,128 @@ pub struct Args {
...
+
+    /// Number of epoch peer sets to maintain connections for.
+    /// [default: 4]
+    #[arg(long = "consensus.peer-set-epoch-depth", default_value_t = 4)]
+    pub peer_set_epoch_depth: usize,
+
+    /// Duration after which a blocked peer is allowed to reconnect.
+    /// [default: 4h, local: 1h]
+    #[arg(
+        long = "consensus.block-duration",
```
SuperFluffy: Can be confused with a block time. Let's make this more explicit:

```suggestion
        long = "consensus.time-to-unblock-byzantine-peer",
```

crates/commonware-node/src/args.rs
```diff
@@ -110,17 +109,128 @@ pub struct Args {
...
+
+    /// Time into the future that a timestamp can be and still be considered valid.
+    /// [default: 5s]
+    #[arg(long = "consensus.synchrony-bound", default_value = "5s")]
+    pub synchrony_bound: jiff::SignedDuration,
+
+    /// How often to attempt dialing peers. Lower values mean faster peer discovery.
+    /// [default: 1s, local: 500ms]
+    #[arg(
+        long = "consensus.dial-interval",
```
SuperFluffy: What do you think about `--consensus.wait-before-redial` or `--consensus.time-between-dial`?

I am trying to figure out how to make these maximally self-explanatory.

crates/commonware-node/src/args.rs
```diff
@@ -110,17 +109,128 @@ pub struct Args {
...
+        default_value = "1s",
+        default_value_if("use_local_defaults", "true", "500ms")
+    )]
+    pub dial_interval: jiff::SignedDuration,
+
+    /// How often to query for new dialable peers. Also limits re-dial rate per peer.
+    /// [default: 60s, local: 30s]
+    #[arg(
+        long = "consensus.query-interval",
```
SuperFluffy: `--consensus.time-between-queries`?

Trying to come up with names that are self-explanatory just by looking at a k8s yml without having to consult `--help` (without being obviously wrong).

crates/commonware-node/src/args.rs
```diff
@@ -110,17 +109,128 @@ pub struct Args {
...
+        long = "consensus.ping-interval",
+        default_value = "50s",
+        default_value_if("use_local_defaults", "true", "5s")
+    )]
+    pub ping_interval: jiff::SignedDuration,
+
+    /// Minimum time between connection attempts to the same peer.
+    /// [default: 60s, local: 1s]
+    #[arg(
+        long = "consensus.connection-min-period",
```
SuperFluffy: The immediate question that springs to mind is how this is different from `--consensus.dial-inverval`. For comparison (the help text doesn't seem entirely accurate either - "how often" feels like it should take a number, but this is taking a time).

```
--consensus.dial-interval
/// How often to attempt dialing peers. Lower values mean faster peer discovery.
```

crates/commonware-node/src/args.rs
hamdiallam: Added a `PositiveDuration` struct. This has to be re-exported due to the visibility of `Args`. Can introduce a types module if we want to generalize this later.

re-exporting seemed simplest

crates/commonware-node/src/args.rs
hamdiallam: I like the `--wait-before-peers-` prefix!

crates/commonware-node/src/args.rs
hamdiallam: This is a per-peer rate limit. --dial-interval is a global loop if theres'a new peer to dial _or_ a peer with no connection.

Updated the names which I think is much clearer
```
--consensus.wait-before-peers-dial
--consensus.connection-per-peer-min-period
```

crates/commonware-node/src/args.rs
hamdiallam: Changed `query` to discovery to make it clear this is a period were the node is looking for new peers

crates/commonware-node/src/args.rs
hamdiallam: --wait-before-peers is good

"wait-before" for the clarity and "peers" for it being an activity across all peers

crates/commonware-node/src/args.rs
```diff
@@ -117,25 +123,112 @@ pub struct Args {
...
 
-    /// Use P2P defaults optimized for local network environments.
-    /// Only enable in non-production network nodes.
-    #[arg(long = "consensus.use-local-p2p-defaults", default_value_t = false)]
-    pub use_local_defaults: bool,
+    /// Whether to allow connections with private IP addresses.
+    #[arg(
+        long = "consensus.allow-private-ips",
+        default_value_t = false,
+        default_value_if("use_local_defaults", "true", "true")
```
SuperFluffy, line 130: not for this PR: wonder if this can be streamlined so that it's not so repetitive. Probably not. :-/

crates/commonware-node/src/args.rs
```diff
@@ -117,25 +123,112 @@ pub struct Args {
...
+    #[arg(
+        long = "consensus.wait-before-peers-ping",
+        default_value = "50s",
+        default_value_if("use_local_defaults", "true", "5s")
+    )]
+    pub wait_before_peers_ping: PositiveDuration,
+
+    /// How often to query for new dialable peers. Also limits re-dial rate per peer.
+    #[arg(
+        long = "consensus.wait-before-peers-discovery",
```
SuperFluffy, line 161: Still super confused what the difference between this and `--consensus.wait-before-peers-dial` is.

crates/commonware-node/src/args.rs
```diff
@@ -117,25 +123,112 @@ pub struct Args {
...
+    #[arg(long = "consensus.allow-dns", default_value_t = true)]
+    pub allow_dns: bool,
+
+    /// Time into the future that a timestamp can be and still be considered valid.
+    #[arg(long = "consensus.synchrony-bound", default_value = "5s")]
+    pub synchrony_bound: PositiveDuration,
+
+    /// How long to wait before attempting to dial any peers without a connection..
+    #[arg(
+        long = "consensus.wait-before-peers-dial",
```
SuperFluffy: Maybe?
```suggestion
        long = "consensus.wait-before-peers-redial",
```

crates/commonware-node/src/args.rs
```diff
@@ -117,25 +123,112 @@ pub struct Args {
...
+        default_value = "1s",
+        default_value_if("use_local_defaults", "true", "500ms")
+    )]
+    pub wait_before_peers_dial: PositiveDuration,
+
+    /// How long to wait before sending a ping message to peers for liveness detection.
+    #[arg(
+        long = "consensus.wait-before-peers-ping",
```
SuperFluffy: Maybe? Although `reping` sounds really bloody whack.
```suggestion
--consensus.wait-before-peers-reping
```

crates/commonware-node/src/args.rs
```diff
@@ -117,25 +123,112 @@ pub struct Args {
...
+        default_value = "60s",
+        default_value_if("use_local_defaults", "true", "1s")
+    )]
+    pub connection_per_peer_min_period: PositiveDuration,
+
+    /// Minimum time between handshake attempts from a single IP address.
+    #[arg(
+        long = "consensus.handshake-per-ip-min-period",
+        default_value = "5s",
+        default_value_if("use_local_defaults", "true", "62ms")
```
SuperFluffy, line 181: 62ms is an amazingly specific value.

crates/commonware-node/src/args.rs
```diff
@@ -117,25 +123,112 @@ pub struct Args {
...
+    /// Minimum time between handshake attempts from a single subnet.
+    #[arg(
+        long = "consensus.handshake-per-subnet-min-period",
+        default_value = "15ms",
+        default_value_if("use_local_defaults", "true", "7ms")
+    )]
+    pub handshake_per_subnet_min_period: PositiveDuration,
+
+    /// Duration after which a handshake message is considered stale.
+    #[arg(long = "consensus.handshake-max-age", default_value = "10s")]
```
SuperFluffy: Maybe
```
--consensus.handshake-stale-after
```

crates/commonware-node/src/args.rs
```diff
@@ -117,25 +123,112 @@ pub struct Args {
...
+    pub handshake_max_age: PositiveDuration,
+
+    /// Timeout for the handshake process.
+    #[arg(long = "consensus.handshake-timeout", default_value = "5s")]
+    pub handshake_timeout: PositiveDuration,
+
+    /// Maximum number of concurrent handshake attempts allowed.
+    #[arg(
+        long = "consensus.handshake-max-concurrent",
```
SuperFluffy: Reads nicer
```
--consensus.max-concurrent-handshakes
```

crates/commonware-node/src/args.rs
hamdiallam: So I think when commonware discovers new peers it doesn't automatically dial it. This loop controls when all new peers are dialed

crates/commonware-node/src/args.rs
hamdiallam: I like this!

crates/commonware-node/src/args.rs
hamdiallam: Lol yea amp conversion for the local default since it's not a clean fraction 😂

hamdiallam:
> Left some ideas for name changes. For some others I don't have better ideas (I don't like the `-period` suffix, but can't come up with anything better).
> 
> Some arguments are still really confusing to me - maybe have amp try to come up with a better explanation? Not a blocker, we can always come back to this.
> 
> The `PositiveDuration` wrapper is very nice.

Agreed "-period" isn't ideal but the best I could come up with

PR #2415, opened 2026-02-03
feat(txpool): evict transactions on UserTokenSet events to prevent fee token TOCTOU attacks

fgimenez:
Closes CHAIN-605

When a user submits a transaction without an explicit `fee_token`, the pool resolves their fee token preference from storage at validation time. If the user then calls `setUserToken()` to change their preference, the transaction could fail at execution time due to insufficient balance/liquidity in the new token—a classic TOCTOU vulnerability.

This PR adds code to monitor `IFeeManager::UserTokenSet` events and immediately evict transactions from affected users that don't have an explicit `fee_token` set (i.e., those relying on storage preference resolution).

- `maintain.rs`: Added `user_token_changes` to `TempoPoolUpdates`, parse `UserTokenSet` events in `from_chain()`
- `tempo_pool.rs`: Extended `evict_invalidated_transactions()` to evict transactions from users who changed their fee token preference

Follows the same eviction pattern as validator token / blacklist / whitelist changes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2054,9 +2054,9 @@ checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
 name = "bytes"
-version = "1.11.0"
+version = "1.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b35204fbdc0b3f4446b89fc1ac2cf84a8a68971995d0bf2e925ec7cd960f9cb3"
+checksum = "1e748733b7cbc798e1434b6ac524f0c1ff2ab456fe201501e6497c8417a4fc33"
 dependencies = [
  "serde",
 ]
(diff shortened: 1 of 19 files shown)


crates/transaction-pool/src/maintain.rs
```diff
@@ -47,6 +47,11 @@ pub struct TempoPoolUpdates {
     pub spending_limit_changes: SpendingLimitUpdates,
     /// Validator token preference changes: (validator, new_token).
     pub validator_token_changes: Vec<(Address, Address)>,
+    /// User token preference changes.
+    /// When a user changes their fee token preference via `setUserToken()`, pending
+    /// transactions from that user that don't have an explicit fee_token set may now
+    /// resolve to a different token at execution time, causing fee payment failures.
+    pub user_token_changes: Vec<Address>,
```
mattsse: why is this not a set, and how to we clean this?

fgimenez: good point, changed to set. aiui there is no need to cleanup given that this is per-block data, lmk if we should change anything about this

PR #2423, opened 2026-02-03
bump(deps): update dependencies to latest versions

zerosnacks:
## Summary

Bump dependencies across the workspace.

## Changes

- **Dependency updates**: Bump various dependencies in `Cargo.toml` and update `Cargo.lock`
- **Add `rand_08`** symlink for `rand` 0.8.5: used compatibility where needed
- **Minor fixes**: Small nits and adjustments for updated APIs

## Files Changed

- Root `Cargo.toml` and `Cargo.lock`
- Binary crates: `tempo`, `tempo-bench`, `tempo-sidecar`
- Library crates: `node`, `primitives`, `precompiles`, `revm`, `e2e`, `eyre`
- Commonware crates: `commonware-node`, `commonware-node-config`
- Tooling: `xtask`, `dkg-onchain-artifacts`
- Config: `deny.toml`

### DIFF
--- a/crates/precompiles/src/tip20/mod.rs
+++ b/crates/precompiles/src/tip20/mod.rs
@@ -675,14 +675,14 @@ impl TIP20Token {
     }
 
     /// Checks if the transfer is authorized.
-    /// TIP-1015: For T1+, uses directional sender/recipient checks.
+    /// TIP-1015: For T2+, uses directional sender/recipient checks.
     pub fn is_transfer_authorized(&self, from: Address, to: Address) -> Result<bool> {
         let policy_id = self.transfer_policy_id()?;
         let registry = TIP403Registry::new();
 
-        // (spec: +T1) short-circuit and skip recipient check if sender fails
+        // (spec: +T2) short-circuit and skip recipient check if sender fails
         let sender_auth = registry.is_authorized_as(policy_id, from, AuthRole::sender())?;
-        if self.storage.spec().is_t1() && !sender_auth {
+        if self.storage.spec().is_t2() && !sender_auth {
             return Ok(false);
         }
         let recipient_auth = registry.is_authorized_as(policy_id, to, AuthRole::recipient())?;
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -967,9 +967,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-trie"
-version = "0.9.3"
+version = "0.9.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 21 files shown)


deny.toml
```diff
@@ -8,8 +8,8 @@ ignore = [
   "RUSTSEC-2024-0436",
   # https://rustsec.org/advisories/RUSTSEC-2025-0141 bincode is unmaintained
   "RUSTSEC-2025-0141",
-  # https://rustsec.org/advisories/RUSTSEC-2026-0003 cmov timing on ARM32 - we don't target ARM32
-  "RUSTSEC-2026-0003",
+  # https://rustsec.org/advisories/RUSTSEC-2026-0002 lru unsound (waiting for discv5 update)
+  "RUSTSEC-2026-0002",
```
zerosnacks, line 12: `lru` is used by `ratatui` 0.29: resolved by upgrading to more recent version of Reth
`lru` is used by `discv5` 0.10: pending resolve

crates/precompiles/src/tip20/mod.rs
```diff
@@ -846,7 +846,7 @@ pub(crate) mod tests {
         storage::{StorageCtx, hashmap::HashMapStorageProvider},
         test_util::{TIP20Setup, setup_storage},
     };
-    use rand::{Rng, distributions::Alphanumeric, thread_rng};
+    use rand_08::{Rng, distributions::Alphanumeric, thread_rng};
```
0xKitsune, line 849: Why are we using `rand_08` here instead of `rand`?

crates/precompiles/src/tip20/mod.rs
```diff
@@ -1850,7 +1850,7 @@ pub(crate) mod tests {
             assert_eq!(token.transfer_policy_id()?, 1);
 
             // Test random invalid policy IDs should fail
-            let mut rng = rand::thread_rng();
+            let mut rng = rand_08::thread_rng();
```
0xKitsune, line 1853: Same here, looks like there are a few places where `rand_08` is used. Can we revert this to use `rand` instead?

crates/precompiles/src/tip20/mod.rs
zerosnacks: These can be replaced quite easily but need to be done with care - outside of the scope of this PR I feel, added a follow-up ticket here: https://github.com/tempoxyz/tempo/issues/2424

crates/precompiles/src/tip20/mod.rs
zerosnacks: https://github.com/tempoxyz/tempo/pull/2423#discussion_r2764334760

PR #2427, opened 2026-02-03
deps(ci): bump pinned CI hashes, add `dependabot` (grouped updates) and tightly scope permissions

zerosnacks:
## Summary

Routine CI maintenance: updates pinned action versions, adds Dependabot for automated updates, and tightens workflow permissions to follow principle of least permissions.

## Changes

- Updated third-party action pins to latest versions with commit SHA + version comment format
- Some very common and trusted action providers (like `actions/*` / `dtolnay/rust-toolchain`) are intentionally unpinned
- Added `persist-credentials: false` to all `actions/checkout` calls
- Added explicit job-level permissions following principle of least privilege
- Fixed a few template injection patterns by using environment variables
- Added Dependabot config for grouped GitHub Actions updates

## Tools

Tools used: https://github.com/zizmorcore/zizmor

## Note

The permission scoping is done with the best effort as I cannot test these locally (only from docs / reasoning / help of AI), it may be possible you run into issues when running a particular workflow; if so: widen the scope.

No functional changes to CI behavior.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -112,9 +112,9 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "alloy"
-version = "1.4.3"
+version = "1.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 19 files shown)


0xrusowsky (approved):
not an expert, but this makes sense to me (as it is align with what we do in foundry)

jenpaff (approved):

PR #2431, opened 2026-02-03
bench: make expiring nonces the default in tempo-bench

legion2002:
## Summary

Makes expiring nonces (TIP-1009) the default nonce strategy in `tempo-bench`, with 2D nonces available via an opt-in flag.

## Changes

- **Default behavior**: `tempo-bench run-max-tps` now uses expiring nonces by default
- **New flags**:
  - `--use-2d-nonces`: Opt-in to use 2D nonces instead of expiring nonces
  - `--use-standard-nonces`: Use traditional sequential nonces (replaces `--disable-2d-nonces`)
- **Updated README**: Documentation reflects the new defaults and flag names

## Rationale

Expiring nonces use a circular buffer for replay protection, avoiding state bloat from unbounded 2D nonce storage growth. This makes them the more appropriate default for benchmarking.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,31 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    labels:
+      - "A-dependencies"
+      - "A-ci"
+    commit-message:
+      prefix: "chore(ci)"
+    open-pull-requests-limit: 1
+    groups:
+      actions-weekly:
+        applies-to: "version-updates"
+        patterns: ["*"]
+        update-types: ["minor", "patch"]
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    labels:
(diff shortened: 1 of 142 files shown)


bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -168,25 +168,27 @@ pub struct MaxTpsArgs {
...
-    /// Use expiring nonces (TIP-1009) instead of 2D nonces.
+    /// Use standard sequential nonces instead of expiring nonces.
     ///
-    /// Expiring nonces use a circular buffer for replay protection, avoiding state bloat
-    /// from unbounded 2D nonce storage growth. Each transaction has a valid_before timestamp.
+    /// This disables both expiring nonces and 2D nonces, using traditional sequential
+    /// nonce management.
     #[arg(long)]
-    expiring_nonces: bool,
+    use_standard_nonces: bool,
```
0xrusowsky, line 184: should we name them "protocol" nonces for consistency?

0xrusowsky (approved):
makes sense to me

PR #2434, opened 2026-02-04
TIP-1019: Subblocks v2 - Builder Delegation

malleshpai:
## TIP-1017: Subblocks v2 - Builder Delegation

Spec for decoupling builder identity from validator identity.

- Many-to-one delegation via `delegateToBuilder(builder)`
- Builder address = fee recipient (no separate field)
- ed25519 p2p keys for both authentication and subblock signing (no ECDSA/KMS)
- Nonce key: `0x5b | builder_address | nonce`
- Expiring nonce support (TIP-1009)
- HA via multiple machines per builder

Full spec: [tips/tip-1019.md](https://github.com/tempoxyz/tempo/blob/TIP-1017/tips/tip-1019.md)

### DIFF
--- a/tips/tip-1017.md
+++ b/tips/tip-1017.md
@@ -1,7 +1,7 @@
 ---
 id: TIP-1017
 title: Subblocks v2 - Builder Delegation
-description: Separates subblock builder identity from validator identity with many-to-one delegation, ECDSA signing, builder p2p key registration, and expiring nonce support.
+description: Separates subblock builder identity from validator identity with many-to-one delegation, ed25519 subblock signing via registered p2p keys, and expiring nonce support.
 authors: Mallesh Pai
 status: Draft
 related: Subblock Specification, TIP-1009
@@ -14,15 +14,13 @@ protocolVersion: TBD
 
 This TIP introduces major changes to the subblock system:
 
-1. **Builder Delegation**: Validators delegate subblock production to a builder identified by an ECDSA address (secp256k1). Multiple validators can delegate to the same builder, who accumulates their combined gas budget.
+1. **Builder Delegation**: Validators delegate subblock production to a builder identified by an address. Multiple validators can delegate to the same builder, who accumulates their combined gas budget. The builder address is also the fee recipient for subblock fees.
 
-2. **Separate Signing Identity**: Builders sign subblocks with their own ECDSA key, separate from the validator's consensus key. Multiple physical machines share this ECDSA key (via HSM/KMS) for high availability.
+2. **Builder P2P Keys**: Each physical builder machine has its own ed25519 key registered under the builder address in the SubblockRegistry (max 4 per builder). These keys serve dual purpose: commonware p2p authentication and subblock signing.
 
-3. **Builder P2P Keys**: Each physical builder machine has its own ed25519 key used exclusively for commonware p2p authentication. Builder p2p keys are registered under a builder in the SubblockRegistry (max 4 per builder).
+3. **Nonce Key Format**: The nonce key format changes to `0x5b | builder_address | nonce`, embedding the builder's 20-byte address directly in the transaction.
 
-4. **Nonce Key Format**: The nonce key format changes to `0x5b | builder_address | nonce`, embedding the builder's 20-byte ECDSA address directly in the transaction.
-
-5. **Expiring Nonces in Subblocks**: Subblock transactions can use expiring nonces by setting the nonce portion to all 1s (`0xFFFF...`).
+4. **Expiring Nonces in Subblocks**: Subblock transactions can use expiring nonces by setting the nonce portion to all 1s (`0xFFFF...`).
 
 ## Motivation
 
@@ -38,10 +36,9 @@ The current subblock design has several limitations:
 
 This TIP addresses these issues by:
 
-- Allowing validators to delegate to builders identified by ECDSA address
+- Allowing validators to delegate to builders identified by address
 - Supporting many-to-one delegation where multiple validators share a builder
-- Using the builder's own ECDSA key for subblock signing (separate from consensus key)
-- Supporting multiple physical machines per builder via registered p2p keys (ed25519 for p2p auth, shared ECDSA key for signing)
+- Each builder machine has its own ed25519 key for both p2p authentication and subblock signing — no shared key material (KMS/HSM) required
 - Enabling expiring nonces within subblock transactions
 
 ---
@@ -50,60 +47,43 @@ This TIP addresses these issues by:
 
 ## 1. Subblock Registry
 
-A new Subblock Registry precompile stores builder configurations, validator delegations, and builder p2p key registrations.
+A new Subblock Registry precompile stores validator delegations, builder p2p key registrations, and fee token preferences.
 
-### 1.1 Data Structures
-
-```solidity
-struct BuilderConfig {
-    address feeRecipient;  // where subblock fees are sent
-    address feeToken;      // preferred TIP-20 token for fee payment
-}
-```
-
-**Note on fee token**: The builder's fee token is stored in `SubblockRegistry`, separate from `TipFeeManager.validatorTokens`. This allows:
-- **Subblock fees** (for txs in builder's subblocks): `SubblockRegistry.configOf(builder).feeToken`
-- **Proposer fees** (when validator is block proposer): `TipFeeManager.validatorTokens[validator]`
-
-When processing fees for a subblock transaction:
-1. Resolve the user's fee token using standard Tempo resolution (tx preference → user preference → inference → PathUSD)
-2. Extract `builder_address` from nonce key (bytes 1-20)
-3. Look up `config = SubblockRegistry.configOf(builder)`
-4. If registered and `config.feeToken` differs, attempt to swap via Fee AMM if liquidity available
-5. If no liquidity or unregistered, fees are paid directly in the user's resolved fee token
-6. Fees are sent to `config.feeRecipient`
-
-### 1.2 Storage
+### 1.1 Storage
 
 ```solidity
-// Validator → builder they delegate to (zero = not delegated)
+// Validator → builder address they delegate to (zero = not delegated)
 mapping(address => address) public delegatedBuilder;
-
-// Builder address → their configuration
-mapping(address => BuilderConfig) public builderConfig;
 
 // Builder address → number of validators delegating to them (gas weight)
 mapping(address => uint256) public gasWeight;
 
-// Builder address → registered p2p ed25519 pubkeys (max 4)
+// Builder address → preferred TIP-20 fee token (zero = no swap, use user's token)
+mapping(address => address) public feeToken;
+
+// Builder address → registered ed25519 p2p keys (max 4)
 mapping(address => bytes32[]) public p2pKeys;
 
-// P2P ed25519 pubkey → owning builder address (reverse lookup)
+// ed25519 p2p key → owning builder address (reverse lookup)
 mapping(bytes32 => address) public p2pKeyOwner;
 ```
 
-### 1.3 Interface
+**Note on fee token**: The builder's fee token is stored in `SubblockRegistry`, separate from `TipFeeManager.validatorTokens`. This allows:
+- **Subblock fees** (for txs in builder's subblocks): `SubblockRegistry.feeToken[builder]`, paid to `builder` address
+- **Proposer fees** (when validator is block proposer): `TipFeeManager.validatorTokens[validator]`
+
+### 1.2 Interface
 
 ```solidity
 interface ISubblockRegistry {
     // ========== Validator Functions ==========
 
     /// @notice Delegates the caller's subblock production to a builder.
-    /// @param builder The builder's ECDSA address.
+    /// @param builder The builder's address (also the fee recipient).
     /// @dev Only callable by validators in the ValidatorConfig set.
     /// @dev Clears any existing delegation first (atomically updates gasWeight).
     /// @dev Increments the builder's gasWeight.
-    /// @dev Requires builderConfig[builder].feeRecipient != address(0).
+    /// @dev Requires p2pKeys[builder].length > 0 (builder has at least one registered key).
     function delegateToBuilder(address builder) external;
 
     /// @notice Clears the caller's delegation.
@@ -113,95 +93,61 @@ interface ISubblockRegistry {
 
     // ========== Builder Functions ==========
 
-    /// @notice Sets the caller's builder configuration.
-    /// @param feeRecipient Where subblock fees are sent. Must be non-zero.
-    /// @param feeToken Preferred TIP-20 token for fee payment. Zero means no swap.
-    /// @dev The caller (msg.sender) is the builder. No signature needed.
-    function setBuilderConfig(address feeRecipient, address feeToken) external;
-
-    /// @notice Clears the caller's builder configuration.
-    /// @dev Force-clears ALL delegations pointing to this builder.
-    /// @dev Decrements gasWeight to zero.
-    /// @dev Removes all registered p2p keys.
-    function clearBuilderConfig() external;
-
-    /// @notice Registers an ed25519 p2p key under the caller's builder identity.
-    /// @param p2pPubkey The ed25519 public key for commonware p2p authentication.
+    /// @notice Registers an ed25519 p2p key under the caller's builder address.
+    /// @param p2pPubkey The ed25519 public key for p2p authentication and subblock signing.
     /// @dev Max 4 p2p keys per builder. Reverts if limit exceeded.
     /// @dev Reverts if p2pPubkey is already registered to any builder.
-    /// @dev Requires caller to have an active BuilderConfig.
     function registerP2pKey(bytes32 p2pPubkey) external;
 
-    /// @notice Removes an ed25519 p2p key from the caller's builder identity.
+    /// @notice Removes an ed25519 p2p key from the caller's builder address.
     /// @param p2pPubkey The ed25519 public key to remove.
     /// @dev Reverts if p2pPubkey is not registered to the caller.
     function removeP2pKey(bytes32 p2pPubkey) external;
 
+    /// @notice Sets the caller's preferred fee token for subblock fees.
+    /// @param token The TIP-20 token address. Zero means no swap (use user's token).
+    function setFeeToken(address token) external;
+
     // ========== View Functions ==========
 
     /// @notice Returns the builder a validator has delegated to.
-    /// @param validator The validator address.
-    /// @return The builder's ECDSA address, or zero if not delegated.
     function builderOf(address validator) external view returns (address);
 
-    /// @notice Returns a builder's configuration.
-    /// @param builder The builder's ECDSA address.
-    /// @return The BuilderConfig struct (zero values if not configured).
-    function configOf(address builder) external view returns (BuilderConfig memory);
-
-    /// @notice Returns a builder's fee recipient.
-    /// @param builder The builder's ECDSA address.
-    /// @return The fee recipient, or zero if not configured.
-    function feeRecipientFor(address builder) external view returns (address);
-
-    /// @notice Returns a builder's preferred fee token.
-    /// @param builder The builder's ECDSA address.
-    /// @return The TIP-20 token address, or zero if not configured.
-    function feeTokenFor(address builder) external view returns (address);
-
     /// @notice Returns a builder's gas weight (number of delegating validators).
-    /// @param builder The builder's ECDSA address.
-    /// @return The gas weight.
     function gasWeightOf(address builder) external view returns (uint256);
 
+    /// @notice Returns a builder's preferred fee token.
+    function feeTokenOf(address builder) external view returns (address);
+
     /// @notice Returns the p2p keys registered to a builder.
-    /// @param builder The builder's ECDSA address.
-    /// @return Array of ed25519 public keys.
     function p2pKeysOf(address builder) external view returns (bytes32[] memory);
 
     /// @notice Returns the builder that owns a p2p key.
-    /// @param p2pPubkey The ed25519 public key.
-    /// @return The builder's ECDSA address, or zero if not registered.
     function builderOfP2pKey(bytes32 p2pPubkey) external view returns (address);
 }
 ```
 
-### 1.4 Behavior
+### 1.3 Behavior
 
-**Registry state snapshot:** All registry lookups (gasWeight, feeRecipient, p2p key ownership) during subblock validation and execution use the **post-state of the parent block**. Mutations within the current block do not affect that block's subblock validity or fee routing.
+**Registry state snapshot:** All registry lookups (gasWeight, feeToken, p2p key ownership) during subblock validation and execution use the **post-state of the parent block**. Mutations within the current block do not affect that block's subblock validity or fee routing.
 
 **Delegation:**
 - Only addresses in the `ValidatorConfig` validator set can call `delegateToBuilder()`
 - The SubblockRegistry verifies the caller is a validator by querying `ValidatorConfig`: a non-zero `publicKey` for the caller's address confirms membership
 - When delegating, any existing delegation is cleared first (atomically updates gasWeight)
+- Delegation requires the builder to have at least one registered p2p key
 - On-chain delegation changes take effect immediately in the next block
-
-**Builder configuration:**
-- `setBuilderConfig()` is called by the builder directly (`msg.sender` is the builder's ECDSA address). No signature argument is needed since the transaction itself proves ownership.
-- `feeRecipient` MUST be non-zero before validators can delegate to a builder
-- If `feeToken` is zero, no swap is attempted (user's token is used)
-
-**Clearing builder config:**
-- `clearBuilderConfig()` force-clears ALL delegations pointing to the calling builder
-- `gasWeight` is decremented to zero
-- All registered p2p keys are removed
-- This prevents orphaned delegations pointing to a non-existent builder
 
 **P2P keys:**
 - Each builder can register up to 4 ed25519 p2p keys (one per physical machine)
+- These keys serve dual purpose: commonware p2p authentication and subblock signing
 - `registerP2pKey()` reverts if the key is already registered to any builder
-- `registerP2pKey()` requires the caller to have an active `BuilderConfig`
 - `removeP2pKey()` reverts if the key is not owned by the caller
+
+**Fee token:**
+- `setFeeToken()` sets the builder's preferred fee token for receiving subblock fees
+- If zero, no swap is attempted (fees paid in the user's resolved token)
+- The builder address itself is always the fee recipient — no separate `feeRecipient` field
 
 **Gas weight:**
 - `gasWeight[builder]` is the count of validators currently delegating to that builder
@@ -218,13 +164,13 @@ interface ISubblockRegistry {
   - This can be done lazily (during next delegation change) or eagerly (via callback from ValidatorConfig)
 - This ensures gas budgets accurately reflect the current validator set
 
-### 1.5 Precompile Address
+### 1.4 Precompile Address
 
 ```
 SUBBLOCK_REGISTRY_ADDRESS = 0x0b10c0000000000000000000000000000000000X  // TBD
 ```
 
-### 1.6 Relationship to ValidatorConfig
+### 1.5 Relationship to ValidatorConfig
 
 The SubblockRegistry is a **separate precompile** from ValidatorConfig but **depends on it** for validator set membership:
 
@@ -233,8 +179,8 @@ The SubblockRegistry is a **separate precompile** from ValidatorConfig but **dep
 │   ValidatorConfig   │◄────────│   SubblockRegistry  │
 │                     │  reads  │                     │
 │  (validator set)    │         │ - delegatedBuilder  │
-│                     │         │ - builderConfig     │
 │                     │         │ - gasWeight         │
+│                     │         │ - feeToken          │
 │                     │         │ - p2pKeys           │
 └─────────────────────┘         └─────────────────────┘
      Source of truth              Builder delegation
@@ -244,9 +190,9 @@ The SubblockRegistry is a **separate precompile** from ValidatorConfig but **dep
 - ValidatorConfig defines **who is in the validator set** (used for consensus)
 - SubblockRegistry tracks **builder delegations and p2p keys** (separate from consensus)
 
-### 1.7 P2P Peer Set
+### 1.6 P2P Peer Set
 
-Builder p2p keys whose owning builder has `gasWeight[builderOfP2pKey(p2pPubkey)] > 0` are included in the epoch-authorized peer set alongside validators. The DKG Manager reads active builder p2p keys from SubblockRegistry and adds them to the commonware p2p oracle, allowing builder machines to join the authenticated network.
+Builder p2p keys whose owning builder has `gasWeight > 0` are included in the epoch-authorized peer set alongside validators. The DKG Manager reads active builder p2p keys from SubblockRegistry and adds them to the commonware p2p oracle, allowing builder machines to join the authenticated network.
 
 **Key distinction:**
 - Builder p2p keys (ed25519) are authorized for **subblock delivery only**, not consensus participation
@@ -271,7 +217,7 @@ Subblock transactions use the following nonce key format (big-endian, byte 0 = M
 ```
 
 - **Byte 0 (MSB)**: `0x5b` prefix identifying this as a subblock transaction
-- **Bytes 1-20**: Builder's 20-byte ECDSA address
+- **Bytes 1-20**: Builder's 20-byte address (also the fee recipient)
 - **Bytes 21-31**: 11-byte nonce value (2^88 possible values)
 
 ### 2.2 V1 vs V2 Nonce Key Discrimination
@@ -374,7 +320,7 @@ expiringNonceKey = (0x5b << 248) | (builder_address << 88) | EXPIRING_NONCE_MARK
 struct SubBlock {
     version: SubBlockVersion,    // V2 for this spec
     parent_hash: B256,           // Block hash this subblock builds on
-    builder: Address,            // Builder's ECDSA address
+    builder: Address,            // Builder's address (= fee recipient)
     chain_id: u64,               // Chain ID for cross-chain replay protection
     transactions: Vec<Transaction>,
 }
@@ -385,23 +331,28 @@ struct SubBlock {
 ```rust
 struct SignedSubBlock {
     subblock: SubBlock,
-    signature: Bytes,  // ECDSA signature (65 bytes: r, s, v)
+    signer: B256,      // ed25519 public key that signed this subblock
+    signature: Bytes,  // ed25519 signature
 }
 ```
+
+The `signer` is one of the builder's registered p2p keys. Verification:
+1. `ed25519_verify(signer, signature_hash, signature)` — valid signature
+2. `p2pKeyOwner[signer] == subblock.builder` — signer is registered to this builder
 
 ### 4.3 Signature Computation
 
 ```
 signature_hash = keccak256(0x78 || RLP(subblock))
-signature = ecdsa_sign(builder_private_key, signature_hash)
+signature = ed25519_sign(p2p_private_key, signature_hash)
 ```
 
 ### 4.4 Signature Verification
 
 ```
 hash = keccak256(0x78 || RLP(subblock))
-recovered_address = ecrecover(hash, signature)
-valid = (recovered_address == subblock.builder)
+valid = ed25519_verify(signed.signer, hash, signed.signature)
+       && p2pKeyOwner[signed.signer] == signed.subblock.builder
 ```
 
 ### 4.5 Version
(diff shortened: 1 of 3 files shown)


tips/tip-1017.md
```diff
@@ -0,0 +1,700 @@
...
+- Decremented when a validator calls `clearDelegation()` or delegates to a different builder
+- Used to compute the builder's total gas budget
+
+**Validator set changes:**
+- The subblock builder is separate infrastructure from the validator node. A validator being temporarily "down" or offline does NOT affect their builder's ability to produce subblocks.
+- `gasWeight[builder]` always reflects the number of validators **in the consensus set** delegating to that builder.
+- When a validator is **removed from ValidatorConfig** (exits the validator set):
+  - Their delegation MUST be cleared
+  - `gasWeight` for their builder is decremented
+  - This can be done lazily (during next delegation change) or eagerly (via callback from ValidatorConfig)
```
rkrasiuk: lazy approach violates S2 invariant on gas weight consistency, which in turn leads to incorrect gas budget calculation

tips/tip-1017.md
```diff
@@ -0,0 +1,700 @@
...
+### 10.1 Subblock Validity
+
+A signed subblock is valid only if:
+- `parent_hash` matches the current chain tip
+- The signature recovers to `subblock.builder`
+- `gasWeight[builder] > 0` (at least one validator in the set delegates to this builder)
+- All transactions have nonce key prefix `0x5b` with matching `builder`
+- Total gas does not exceed the builder's computed gas budget
+
+**Fee failure**: If a subblock transaction cannot pay fees (e.g., user emptied their account in the main block), the transaction becomes a noop but the nonce is still incremented. The block remains valid.
```
rkrasiuk: requires some clarification for handling of transactions with expiring nonces - does the new entry get inserted into the expiry ring buffer?

tips/tip-1017.md
```diff
@@ -0,0 +1,700 @@
...
+
+### 7.2 Subblock Submission Endpoint
+
+All validators MUST expose an endpoint for receiving signed subblocks from builders:
+
+```
+POST /subblocks/submit
+Content-Type: application/octet-stream
+Body: RLP-encoded SignedSubBlock
+```
```
rkrasiuk: consider an authenticated channel or cheaper validation of sender's identity (e.g. including signature in http header w/o having to decode the request body) to prevent ddos

tips/tip-1017.md
```diff
@@ -0,0 +1,700 @@
...
+
+**Fee failure**: If a subblock transaction cannot pay fees (e.g., user emptied their account in the main block), the transaction becomes a noop but the nonce is still incremented. The block remains valid.
+
+### 10.2 Subblock Expiry
+
+Subblocks expire immediately when `parent_hash` becomes stale. Subblocks have a 1-block lifetime.
+
+### 10.3 Conflicting Subblocks
+
+If a builder submits multiple subblocks for the same `parent_hash`, the proposer selects the one with higher total gas used (maximizing fees).
```
rkrasiuk: gas used is a poor proxy for fee revenue anyway - the subblock could be stuffed with noop high gas transactions. perhaps, it's better to consider the latest submitted builder subblock under the assumption that builder has most up-to-date information regarding the transactions it included

tips/tip-1017.md
```diff
@@ -0,0 +1,700 @@
...
+4. Builder creates SubBlock with parent_hash = current chain tip
+5. Builder signs with ECDSA → SignedSubBlock
+6. Builder submits to validators' subblock endpoint
+```
+
+### 8.5 Block Building (Proposer)
+
+```
+1. Proposer receives SignedSubBlocks from validators' internal propagation
+2. For each subblock:
```
rkrasiuk: as clearly stated by the proposal - the subblocks are executed after the main block body. however, it's not immediately clear how subblocks are ordered among themselves

rkrasiuk (commented):
leaving some comments / suggestions. forgive my ignorance where applicable, granted i'm not well versed in tempo protocol spec

tips/tip-1019.md
```diff
@@ -0,0 +1,823 @@
...
+    /// @notice Clears the caller's delegation.
+    /// @dev Only callable by validators with an existing delegation.
+    /// @dev Decrements the previous builder's gasWeight.
+    function clearDelegation() external;
+
+    // ========== Builder Functions ==========
+
+    /// @notice Registers an ed25519 p2p key under the caller's builder address.
+    /// @param p2pPubkey The ed25519 public key for p2p authentication and subblock signing.
+    /// @dev Max 4 p2p keys per builder. Reverts if limit exceeded.
```
Zygimantass, line 121: registerBuilder instead of registerp2pkey?

tips/tip-1017.md
```diff
@@ -0,0 +1,823 @@
...
+A new Subblock Registry precompile stores validator delegations, builder p2p key registrations, and fee token preferences.
+
+### 1.1 Storage
+
+```solidity
+// Validator → builder address they delegate to (zero = not delegated)
+mapping(address => address) public delegatedBuilder;
+
+// Builder address → number of validators delegating to them (gas weight)
+mapping(address => uint256) public gasWeight;
```
Zygimantass: instead of storing, lets compute on the fly?

tips/tip-1017.md
```diff
@@ -0,0 +1,823 @@
...
+
+// Builder address → preferred TIP-20 fee token (zero = no swap, use user's token)
+mapping(address => address) public feeToken;
```
Zygimantass: let's reuse feemanager imo

tips/tip-1019.md
```diff
@@ -0,0 +1,823 @@
...
+```
+
+**Note on fee token**: The builder's fee token is stored in `SubblockRegistry`, separate from `TipFeeManager.validatorTokens`. This allows:
+- **Subblock fees** (for txs in builder's subblocks): `SubblockRegistry.feeToken[builder]`, paid to `builder` address
+- **Proposer fees** (when validator is block proposer): `TipFeeManager.validatorTokens[validator]`
+
+### 1.2 Interface
+
+```solidity
+interface ISubblockRegistry {
```
Zygimantass, line 99: I wonder if its cleaner to extend validatorconfig instead of doing the subblockregistry? might be annoying for parallelizing the work across the two TIPs, but one surface area might be cleaner

tips/tip-1017.md
```diff
@@ -0,0 +1,823 @@
...
+- ValidatorConfig calls `SubblockRegistry.onValidatorRemoved()` when a validator exits the set
+- SubblockRegistry reads ValidatorConfig for membership checks and tracks **builder delegations and p2p keys**
+
+### 1.6 P2P Peer Set
+
+Builder p2p keys whose owning builder has `gasWeight > 0` are included in the epoch-authorized peer set alongside validators. The DKG Manager reads active builder p2p keys from SubblockRegistry and adds them to the commonware p2p oracle, allowing builder machines to join the authenticated network.
+
+**Key distinction:**
+- Builder p2p keys (ed25519) are authorized for **subblock delivery only**, not consensus participation
+- P2P peer set updates occur at **epoch boundaries**, not per-block
```
Zygimantass: this should be per-block

Zygimantass: cc @SuperFluffy

tips/tip-1019.md
malleshpai: happy to go either way. p2pkey seemed more clear about what this was.

tips/tip-1017.md
malleshpai: can do but why?

tips/tip-1017.md
SuperFluffy: This occurs per block and on a best-case effort.

tips/tip-1017.md
```diff
@@ -0,0 +1,823 @@
+---
+id: TIP-1017
```
SuperFluffy: Should be TIP-1018

malleshpai: i had tip-1017 a while ago, race condition, will take final number now after approvals and right before merging :)

tips/tip-1017.md
malleshpai: can do, maybe @klkvr should weigh in on what's best?

tips/tip-1019.md
malleshpai: mostly didn't want to touch a different in-progress TIP + separation of concerns, this has nothing to do with consensus.

tips/tip-1017.md
```diff
@@ -0,0 +1,823 @@
...
+
+---
+
+# Specification
+
+## Definitions
+
+- **Parent post-state**: The world state after executing the parent block — equivalently, the state at the very start of the current block, before any of its transactions execute.
+- **Nonce key** (`nonceKey`): The full `U256` routing key embedded in a transaction, structured as `0x5b | builder_address | nonce88`.
+- **Nonce portion** (`nonce88`): The low 11 bytes (`uint88`) of the nonce key, representing the sequential nonce value within an independent `(account, nonceKey)` sequence in the NonceManager.
```
klkvr: "sequential" here is a bit confusing, this is just unreserved portion of u256 nonce key right?

tips/tip-1017.md
```diff
@@ -0,0 +1,823 @@
...
+mapping(address => uint256) public gasWeight;
+
+// Builder address → preferred TIP-20 fee token (zero = no swap, use user's token)
+mapping(address => address) public feeToken;
+
+// Builder address → registered ed25519 p2p keys (max 4)
+mapping(address => bytes32[]) public p2pKeys;
+
+// ed25519 p2p key → owning builder address (reverse lookup)
+mapping(bytes32 => address) public p2pKeyOwner;
```
klkvr: i guess given that there will be very few builders that we care about (only those at least one validator delegated to), we shouldn't need this onchain as all builders will be in memory

tips/tip-1017.md
klkvr: one motivation could be that we will keep all builders in memory (just like we keep all validators right now), and thus storing this in contract is not very beneficial because recomputing this on the fly is cheap

i think we have an opportunity to simplify the builder registry to just 2 mappings — `delegatedBuilder` and `builderConfig` (`feeToken` + `p2pKeys` mappings combined)

tips/tip-1017.md
```diff
@@ -0,0 +1,823 @@
...
+```
+
+**Note on fee token**: The builder's fee token is stored in `SubblockRegistry`, separate from `TipFeeManager.validatorTokens`. This allows:
+- **Subblock fees** (for txs in builder's subblocks): `SubblockRegistry.feeToken[builder]`, paid to `builder` address
+- **Proposer fees** (when validator is block proposer): `TipFeeManager.validatorTokens[validator]`
```
klkvr: the way subblocks work right now, the subblock builder/proposer is effectively treated as temporary beneficiary for some duration of a block, so I'm wonderting if for consistency we should just use `FeeManager.validatorTokens[subblockBuilder]`? 

do we consider it likely that some addresses would be used as both validator's beneficiary and a subblock builder, and would want to use different fee tokens for both?

tips/tip-1017.md
```diff
@@ -0,0 +1,823 @@
...
+## 2. Nonce Key Format
+
+### 2.1 Subblock Nonce Key
+
+Subblock transactions use the following nonce key format (big-endian, byte 0 = MSB, matching `U256::to_be_bytes()`):
+
+```
+32 bytes total:
+┌──────────────────┬──────────────────────────────────┬─────────────────────┐
+│ 0x5b (1 byte)    │ builder_address (20 bytes)        │ nonce (11 bytes)    │
```
klkvr: here `nonce` is also slightly confusing, it's just remaining part of the nonce key

tips/tip-1019.md
```diff
@@ -0,0 +1,823 @@
...
+### 5.1 SubBlockMetadata
+
+The proposer includes metadata about included subblocks in a system transaction:
+
+```rust
+struct SubBlockMetadataEntry {
+    version: SubBlockVersion,   // V2
+    builder: Address,           // Builder's address (= fee recipient)
+    signer: B256,               // ed25519 public key that signed the subblock
+    tx_start_index: u32,        // Index of first tx in block
```
klkvr: might be worth clarifying what this should be set to if subblock is empty

tips/tip-1019.md
```diff
@@ -0,0 +1,823 @@
...
+
+### 5.1 SubBlockMetadata
+
+The proposer includes metadata about included subblocks in a system transaction:
+
+```rust
+struct SubBlockMetadataEntry {
+    version: SubBlockVersion,   // V2
+    builder: Address,           // Builder's address (= fee recipient)
+    signer: B256,               // ed25519 public key that signed the subblock
```
klkvr, line 401: `builder` seems redundant given that it can be determined by the `signer`?

tips/tip-1017.md
```diff
@@ -0,0 +1,823 @@
...
+    tx_count: u32,              // Number of transactions
+    signature: Bytes,           // ed25519 signature
+}
+
+struct SubBlockMetadata {
+    entries: Vec<SubBlockMetadataEntry>,
+}
+```
+
+The `builder` address is both the on-chain identity and the fee recipient. No separate `fee_recipient` field is needed.
```
klkvr: should we consider changing this btw? there could be usecases for separating "coinbase" from admin address of the subblock registry

tips/tip-1019.md
```diff
@@ -0,0 +1,823 @@
...
+struct SubBlockMetadata {
+    entries: Vec<SubBlockMetadataEntry>,
+}
+```
+
+The `builder` address is both the on-chain identity and the fee recipient. No separate `fee_recipient` field is needed.
+
+### 5.2 Metadata Validation
+
+During block execution, the SubBlockMetadata system transaction is processed:
```
klkvr, line 551: should we consider moving it to the beginning of the block? impl-wise it might make it slightly easier because we will be able to first see the metadata tx, remember expected structure, and validate it while executing the block vs right now we need to validate it at the end

tips/tip-1017.md
```diff
@@ -0,0 +1,823 @@
...
+
+```
+per_validator_gas = (block_gas_limit / TEMPO_SHARED_GAS_DIVISOR) / num_validators
+builder_gas_budget = per_validator_gas × gasWeight[builder]
+```
+
+The metadata includes the `builder` address. During validation:
+1. Look up `gasWeight[builder]` from SubblockRegistry at parent block post-state
+2. Compute builder's gas budget as `per_validator_gas × gasWeight`
+3. Verify subblock's total gas usage does not exceed budget
```
klkvr: might be worth clarifying that by including builder's subblock its entire gas budged is getting unlocked to validator in the incentive section

tips/tip-1017.md
```diff
@@ -0,0 +1,823 @@
...
+1. Transaction's explicit `fee_token` field (if set)
+2. User's default fee token preference (from user config)
+3. Inference from transaction context
+4. PathUSD (default)
+
+### 6.3 Fee Token Swap
+
+If `user_fee_token != builder_fee_token` and the builder has a token preference (`feeToken[builder] != 0`):
+- If sufficient FeeAMM liquidity exists, swap to `feeToken[builder]` using the same swap mechanism as proposer fee swaps
+- If insufficient liquidity, the builder receives `user_fee_token` directly
```
klkvr: this is not how subblock nor regular transactions work right now, right? quite important addition and is a bit deep in the TIP

i believe right now if there's not enough liquidity we will not invalidate tx but won't charge the user either

klkvr (approved):

tips/tip-1019.md
```diff
@@ -0,0 +1,823 @@
...
+protocolVersion: TBD
+---
+
+# TIP-1017: Subblocks v2 - Builder Delegation
+
+## Abstract
+
+This TIP introduces changes to the subblock system:
+
+1. **Builder Delegation**: Validators delegate subblock production to a builder identified by an address. Multiple validators can delegate to the same builder, who accumulates their combined gas budget. The builder address is also the fee recipient for subblock fees.
```
danrobinson, line 17: Can you say the address corresponds to an ECDSA key

tips/tip-1019.md
```diff
@@ -0,0 +1,823 @@
...
+
+The current subblock design has several limitations:
+
+1. **Validator coupling**: The validator's consensus key must also sign subblocks, meaning validators cannot delegate subblock production to separate infrastructure without sharing their consensus key.
+
+2. **One-to-one only**: Each validator must run their own subblock builder; there is no way to delegate to a shared builder.
+
+3. **No expiring nonces**: Subblock transactions require sequential nonces within their nonce key, preventing use of expiring nonces (TIP-1009).
+
+4. **No high availability**: A single builder machine failure takes down subblock production for that validator. There is no mechanism for multiple machines to operate under a shared identity.
```
danrobinson: Clarify the one IP rule

tips/tip-1019.md
danrobinson: I guess it doesn't necessarily; nevermind

tips/tip-1017.md
```diff
@@ -0,0 +1,823 @@
...
+---
+
+# Specification
+
+## Definitions
+
+- **Parent post-state**: The world state after executing the parent block — equivalently, the state at the very start of the current block, before any of its transactions execute.
+- **Nonce key** (`nonceKey`): The full `U256` routing key embedded in a transaction, structured as `0x5b | builder_address | nonce88`.
+- **Nonce portion** (`nonce88`): The low 11 bytes (`uint88`) of the nonce key, representing the sequential nonce value within an independent `(account, nonceKey)` sequence in the NonceManager.
+- **Builder address**: A 20-byte address that identifies the builder and serves as the fee recipient. Must be capable of sending transactions (i.e., an EOA or an account that can call `registerP2pKey`, `removeP2pKey`, `setFeeToken`).
```
danrobinson: Must be capable of making calls, I think? Not transactions

tips/tip-1017.md
danrobinson: There's already a fee token mapping in the fee manager, yeah—I think could reuse it. I think it actually allows anyone to set their fee token preference anyway

tips/tip-1017.md
danrobinson: I think it feels cleaner to have it in state. Updating it is gonna be pretty rare anyway so the optimization doesn't save you much. But yield to @klkvr

tips/tip-1017.md
danrobinson: This one makes sense to me, unless you hit a roadblock during implementation

tips/tip-1017.md
danrobinson: Agree; I think should just use FeeManager.validatorTokens

danrobinson: It probably should be called builderTokens now or something, but not that important

tips/tip-1019.md
```diff
@@ -0,0 +1,823 @@
...
+interface ISubblockRegistry {
+    // ========== Validator Functions ==========
+
+    /// @notice Delegates the caller's subblock production to a builder.
+    /// @param builder The builder's address (also the fee recipient).
+    /// @dev Only callable by validators in the ValidatorConfig set.
+    /// @dev Clears any existing delegation first (atomically updates gasWeight).
+    /// @dev Increments the builder's gasWeight.
+    /// @dev Requires p2pKeys[builder].length > 0 (builder has at least one registered key).
+    function delegateToBuilder(address builder) external;
```
danrobinson, line 108: Can you just add to the comment that it sets delegatedBuilder[msg.sender] = builder

tips/tip-1017.md
```diff
@@ -0,0 +1,823 @@
...
+
+### 1.2 Interface
+
+```solidity
+interface ISubblockRegistry {
+    // ========== Validator Functions ==========
+
+    /// @notice Delegates the caller's subblock production to a builder.
+    /// @param builder The builder's address (also the fee recipient).
+    /// @dev Only callable by validators in the ValidatorConfig set.
```
danrobinson: Is this needed? Or should we just allow anyone to set it, and we only look it up for current vaidators?

tips/tip-1019.md
danrobinson: I think registerP2pKey because a builder can call it multiple times to register multiple keys; that isn't the builder registering themselves multiple times

tips/tip-1019.md
```diff
@@ -0,0 +1,823 @@
...
+    /// @dev Only callable by validators with an existing delegation.
+    /// @dev Decrements the previous builder's gasWeight.
+    function clearDelegation() external;
+
+    // ========== Builder Functions ==========
+
+    /// @notice Registers an ed25519 p2p key under the caller's builder address.
+    /// @param p2pPubkey The ed25519 public key for p2p authentication and subblock signing.
+    /// @dev Max 4 p2p keys per builder. Reverts if limit exceeded.
+    /// @dev Reverts if p2pPubkey is already registered to any builder.
```
danrobinson, line 122: I think maybe have to provide a signature from the p2pPubKey on the builder address (with domain separation), just to prevent someone from sniping it to grief a builder.

tips/tip-1017.md
```diff
@@ -0,0 +1,823 @@
...
+    function registerP2pKey(bytes32 p2pPubkey) external;
+
+    /// @notice Removes an ed25519 p2p key from the caller's builder address.
+    /// @param p2pPubkey The ed25519 public key to remove.
+    /// @dev Reverts if p2pPubkey is not registered to the caller.
+    function removeP2pKey(bytes32 p2pPubkey) external;
+
+    /// @notice Sets the caller's preferred fee token for subblock fees.
+    /// @param token The TIP-20 token address. Zero means no swap (use user's token).
+    function setFeeToken(address token) external;
```
danrobinson: Can remove

tips/tip-1017.md
```diff
@@ -0,0 +1,823 @@
...
+
+    // ========== ValidatorConfig Callback ==========
+
+    /// @notice Called by ValidatorConfig when a validator is removed from the set.
+    /// @param validator The validator being removed.
+    /// @dev Clears delegatedBuilder[validator] and decrements the builder's gasWeight.
+    /// @dev Only callable by ValidatorConfig.
+    function onValidatorRemoved(address validator) external;
```
danrobinson: This may not be needed if we compute it on the fly

tips/tip-1017.md
```diff
@@ -0,0 +1,823 @@
...
+
+    // ========== View Functions ==========
+
+    /// @notice Returns the builder a validator has delegated to.
+    function builderOf(address validator) external view returns (address);
+
+    /// @notice Returns a builder's gas weight (number of delegating validators).
+    function gasWeightOf(address builder) external view returns (uint256);
```
danrobinson: Remove this if we compute on the fly

tips/tip-1017.md
```diff
@@ -0,0 +1,823 @@
...
+
+    /// @notice Returns a builder's preferred fee token.
+    function feeTokenOf(address builder) external view returns (address);
```
danrobinson: Remove if we use the FeeManager

tips/tip-1017.md
```diff
@@ -0,0 +1,823 @@
...
+}
+```
+
+### 1.3 Behavior
+
+**Registry state snapshot:** All registry lookups (gasWeight, feeToken, p2p key ownership) during subblock validation and execution use the **post-state of the parent block**. Mutations within the current block do not affect that block's subblock validity or fee routing.
+
+**Delegation:**
+- Only addresses in the `ValidatorConfig` validator set can call `delegateToBuilder()`
+- The SubblockRegistry verifies the caller is a validator by querying `ValidatorConfig`: a non-zero `publicKey` for the caller's address confirms membership
```
danrobinson: May not need this anymore if we are computing on the fly

tips/tip-1019.md
```diff
@@ -0,0 +1,823 @@
...
+- Delegation requires the builder to have at least one registered p2p key
+- On-chain delegation changes take effect immediately in the next block
+
+**P2P keys:**
+- Each builder can register up to 4 ed25519 p2p keys (one per physical machine)
+- These keys serve dual purpose: commonware p2p authentication and subblock signing
+- `registerP2pKey()` reverts if the key is already registered to any builder
+- `removeP2pKey()` reverts if the key is not owned by the caller
+
+**Fee token:**
```
danrobinson, line 163: Can remove this and the next two sections if we do the simplifications

tips/tip-1019.md
```diff
@@ -0,0 +1,823 @@
...
+**P2P key removal:**
+- A builder MAY remove all their p2p keys even while `gasWeight > 0`. This effectively disables their subblock production (no key can sign subblocks, no machine can authenticate to the p2p network). The builder must re-register a key before they can produce subblocks again.
+
+### 1.4 Precompile Address
+
+```
+SUBBLOCK_REGISTRY_ADDRESS = 0x0b10c0000000000000000000000000000000000X  // TBD
+```
+
+### 1.5 Relationship to ValidatorConfig
```
danrobinson, line 176: Maybe can remove these dependencies if we compute on the fly

tips/tip-1017.md
danrobinson: Yeah should find a name for this thing. Like nonce remainder

tips/tip-1017.md
```diff
@@ -0,0 +1,823 @@
...
+struct SignedSubBlock {
+    subblock: SubBlock,
+    signer: B256,      // ed25519 public key that signed this subblock
+    signature: Bytes,  // ed25519 signature
+}
+```
+
+The `signer` is one of the builder's registered p2p keys. Verification:
+1. `ed25519_verify(signer, signature_hash, signature)` — valid signature
+2. `p2pKeyOwner[signer] == subblock.builder` — signer is registered to this builder
```
danrobinson: If we remove this mapping then have to search through the p2p keys for the builder to check that this is one of them

tips/tip-1019.md
danrobinson: I think depends if we have the reverse mapping, although I guess you could compute that on the fly in the node

tips/tip-1017.md
danrobinson: I think could probably do that with a smart contract address + offchain. Let's see if we decide we need it

tips/tip-1019.md
danrobinson: IIRC it's at the bottom because that made it easier to build; I think @malleshpai and I are fine with either

tips/tip-1017.md
danrobinson: Yes should announce it up top

tips/tip-1017.md
```diff
@@ -0,0 +1,823 @@
...
+
+All other transaction execution outcomes (revert, out-of-gas) follow standard Tempo execution semantics.
+
+---
+
+## 7. Transaction Routing
+
+### 7.1 Builder Endpoint
+
+Builders MUST expose an RPC endpoint for receiving subblock transactions from users. Users submit transactions to the builder whose address matches their nonce key.
```
danrobinson: Subblock builders MUST

tips/tip-1017.md
```diff
@@ -0,0 +1,823 @@
...
+
+### 7.3 Builder P2P Authorization
+
+Builder p2p keys whose owning builder has `gasWeight > 0` are added to the epoch-authorized peer set. When a builder's `gasWeight` drops to zero (all validators undelegated), their p2p keys are removed from the authorized set at the next epoch boundary.
+
+Builder p2p keys are authorized for **subblock delivery only**, not consensus participation.
+
+### 7.4 High Availability Architecture
+
+Multiple physical machines operate under a single builder address:
```
danrobinson: "can operate"

tips/tip-1017.md
```diff
@@ -0,0 +1,823 @@
...
+
+The proposer stores received subblocks and selects them for inclusion during block building.
+
+### 7.3 Builder P2P Authorization
+
+Builder p2p keys whose owning builder has `gasWeight > 0` are added to the epoch-authorized peer set. When a builder's `gasWeight` drops to zero (all validators undelegated), their p2p keys are removed from the authorized set at the next epoch boundary.
+
+Builder p2p keys are authorized for **subblock delivery only**, not consensus participation.
+
+### 7.4 High Availability Architecture
```
danrobinson: This is maybe needlessly specific/prescriptive; could either say that one benefit of this spec is that builders COULD operate this way, or just drop this section

tips/tip-1017.md
```diff
@@ -0,0 +1,823 @@
...
+1. Validator selects a builder to delegate to (by address)
+2. Validator calls SubblockRegistry.delegateToBuilder(builderAddress)
+3. Registry increments builder's gasWeight
+```
+
+### 8.3 Transaction Creation (User)
+
+```
+1. User selects a builder by address
+2. User constructs nonce key: 0x5b | builder_address | nonce
```
danrobinson: nonce remainder

tips/tip-1017.md
```diff
@@ -0,0 +1,823 @@
...
+2. For each 0x5b transaction:
+   a. Extract builder_address from nonce key (bytes 1-20)
+   b. Determine if expiring nonce (nonce portion = max uint88)
+   c. If expiring: use TIP-1009 replay protection
+   d. Else: use sequential 2D nonce semantics — the nonce key identifies an
+      independent nonce sequence per `(account, nonceKey)` pair, stored in the
+      NonceManager precompile. The nonce value must exactly match the on-chain
+      state and is unconditionally incremented.
+   e. Fees are paid to builder_address
+   f. If builder has feeToken preference and AMM liquidity exists, swap via FeeAMM
```
danrobinson: Maybe we force the builder to specify a preference? And then gotta mention the fallback flow here

tips/tip-1017.md
```diff
@@ -0,0 +1,823 @@
...
+
+Each builder machine has its own ed25519 key that serves dual purpose:
+- **P2P authentication**: The key is used for commonware p2p network authentication, allowing the machine to connect to the network and deliver subblocks.
+- **Subblock signing**: The same key signs subblocks. The SubBlockMetadata includes the `signer` pubkey, which nodes verify via `p2pKeyOwner[signer] == builder`.
+
+This eliminates the need for shared key material (KMS/HSM) across machines. Each machine operates independently with its own key. The on-chain `p2pKeyOwner` mapping is the source of truth for which keys are authorized to sign on behalf of a builder.
+
+## Builder Address as Fee Recipient
+
+The builder address directly receives subblock fees. This eliminates the need for a separate `feeRecipient` field and the `BuilderConfig` struct. A builder's on-chain state is just: which p2p keys are registered to it, how many validators delegate to it, and optionally what fee token it prefers.
```
danrobinson: I think probably mandate that it specifies a fee token, like we do for validators

tips/tip-1019.md
```diff
@@ -0,0 +1,803 @@
...
+
+- **Parent post-state**: The world state after executing the parent block — equivalently, the state at the very start of the current block, before any of its transactions execute.
+- **Nonce key** (`nonceKey`): The full `U256` routing key embedded in a transaction, structured as `0x5b | builder_address | nonce_remainder`.
+- **Nonce remainder** (`nonce_remainder`): The low 11 bytes (`uint88`) of the nonce key. Each unique nonce remainder selects an independent nonce lane: the NonceManager tracks a sequential counter per `(account, nonceKey)` pair, where `nonceKey` is the full 32-byte value including the remainder. The transaction's `nonce` field must match that lane's on-chain counter and is unconditionally incremented. For expiring nonces, the nonce remainder is set to max uint88 (see Section 3).
+- **Builder address**: A 20-byte address that identifies the builder and serves as the fee recipient. Must be capable of calling `registerP2pKey`, `removeP2pKey`, and `FeeManager.setValidatorToken`.
+- **Validator in the set**: An address that is an active validator in `ValidatorConfig`. (Currently: non-zero `publicKey`; after ValidatorConfig V2: `deactivatedAtHeight == 0`.)
```
SuperFluffy: Important note: the validator configs store which validators *should* be active, not the ones that *are* active (active meaning, part of the active committee).

The single source of truth of which validators are active in a given epoch `E` are the `players` written stored in the DKG outcome and written to the chain in `boundary(E-1)`

tips/tip-1019.md
```diff
@@ -0,0 +1,803 @@
...
+
+```
+┌─────────────────────┐  reads   ┌─────────────────────┐
+│   ValidatorConfig   │◄─────────│   SubblockRegistry  │
+│                     │          │                     │
+│  (validator set)    │          │ - delegatedBuilder  │
+│                     │          │ - p2pKeys           │
+└─────────────────────┘          └─────────────────────┘
+     Source of truth               Builder delegation
+     for validator set             and p2p key config
```
SuperFluffy: Mentioned above: these the entries that *should* be validators, not the ones that *are* validators.

On DKG failure, the network always reverts to the previous state.

tips/tip-1019.md
```diff
@@ -0,0 +1,803 @@
...
+```
+
+- ValidatorConfig defines **who is in the validator set** (used for consensus)
```
SuperFluffy: who **should** in the validator set, **not** who is!!

tips/tip-1019.md
```diff
@@ -0,0 +1,803 @@
...
+
+## 7. Transaction Routing
+
+### 7.1 Builder Endpoint
+
+Subblock builders MUST expose an RPC endpoint for receiving subblock transactions from users. Users submit transactions to the builder whose address matches their nonce key. Builders SHOULD authenticate incoming requests (e.g., verify sender signature before full transaction decode) to mitigate DDoS.
+
+### 7.2 Subblock Delivery via P2P
+
+Builder machines join the authenticated commonware p2p network using their ed25519 p2p keys. They send signed subblocks directly to the next proposer via p2p, the same mechanism used in V1.
```
SuperFluffy, line 687: Are builder machines validators in their own right or not?

tips/tip-1019.md
```diff
@@ -0,0 +1,803 @@
...
+
+The proposer stores received subblocks and selects them for inclusion during block building.
+
+### 7.3 Builder P2P Authorization
+
+Builder p2p keys whose owning builder has gas weight > 0 are added to the per-block peer set. When a builder's gas weight drops to zero (all validators undelegated or removed from the set), their p2p keys are removed from the authorized set.
+
+Builder p2p keys are authorized for **subblock delivery only**, not consensus participation.
```
SuperFluffy, line 695: OK, so they are part of the network but never validators nor signers.

SuperFluffy (commented):
Note that P2P communication behaves like N^2 because nodes send most messages all-to-all.

This means we must enforce a fairly strict upper boundary on the maximum amount of builders we want in the network so that we don't affect its performance and eat into (e.g.) consensus budgets.

tips/tip-1019.md
```diff
@@ -0,0 +1,803 @@
...
+    /// @notice Clears the caller's delegation.
+    /// @dev Sets delegatedBuilder[msg.sender] = address(0).
+    function clearDelegation() external;
+
+    // ========== Builder Functions ==========
+
+    /// @notice Registers an ed25519 p2p key under the caller's builder address.
+    /// @param p2pPubkey The ed25519 public key for p2p authentication and subblock signing.
+    /// @param signature Ed25519 signature from p2pPubkey over
+    ///        keccak256(abi.encodePacked("TEMPO_SUBBLOCK_REGISTER_P2P_KEY", chainId, address(this), msg.sender))
```
SuperFluffy, line 119: `abi.encodePacked` probably not nececessary - can write the bytes straight into keccak without encoding first.

~~Should we include sentinel values?~~ Sentinels not necessary. All values are fixed length.

tips/tip-1019.md
```diff
@@ -0,0 +1,1126 @@
...
+
+## Abstract
+
+This TIP introduces major changes to the subblock system:
+
+1. **Builder Delegation**: Validators delegate subblock production to a builder identified by an address. Multiple validators can delegate to the same builder, who accumulates their combined gas budget. The builder address is also the fee recipient for subblock fees.
+
+2. **Builder P2P Keys**: Each physical builder machine has its own ed25519 key registered under the builder address in the SubblockRegistry (max 4 per builder). These keys serve dual purpose: commonware p2p authentication and subblock signing.
+
+3. **Bundle Proposal Model**: The builder signs a SubBlock containing **all proposed transactions** (of any format). The proposer processes the bundle in order — executing valid transactions, **dropping** precheck-invalid ones, and optionally **truncating** overflow beyond the gas budget. Dropped and unprocessed transactions are placed in the SubBlockMetadata system transaction so validators can reconstruct the original signed SubBlock and verify the builder's signature.
```
danrobinson, line 21: Maybe hard limit on gas limit for the subblock for now—optional overflow is a complex rule

tips/tip-1019.md
```diff
@@ -0,0 +1,1126 @@
...
+
+4. **Any Transaction Format**: Subblock transactions are not restricted to the `0x5b` nonce key format. Standard Ethereum transactions (EIP-1559, etc.) and Tempo AA transactions are allowed. The `0x5b` nonce key remains available as an **optional** mechanism for nonce isolation.
```
danrobinson, line 23:
```suggestion
4. **Any Transaction Format**: Subblock transactions are not restricted to the `0x5b` nonce key format. Standard Ethereum transactions (EIP-1559, etc.) and Tempo transactions are allowed. The `0x5b` nonce key remains available as an **optional** mechanism for a user for nonce isolation and to ensure that a transaction can only be included in that validator's subblock.
```

PR #2435, opened 2026-02-04
perf(invariants): optimize TIP20FactoryInvariantTest for phase 1 corpus

gakonst:
## Problem

`invariant_globalInvariants()` was running O(n) loops over all created tokens AND O(actors × salts) loops after every handler call, causing quadratic slowdown during phase 1 corpus generation/replay.

### Root Cause
1. **`invariant_globalInvariants()`** loops over ALL `_createdTokens` (O(n)) after EVERY handler call
2. **`_invariantSaltToTokenConsistency()`** loops over ALL actors × salts (O(actors*salts))
3. Each iteration makes expensive precompile calls: `factory.isTIP20()`, `token.currency()`, `token.quoteToken()`
4. As corpus sequences create tokens, n grows → quadratic behavior

## Solution

1. **Move FAC1/FAC11 verification inline** to `_recordCreatedToken()` - verified once at creation time instead of after every handler call
2. **Replace O(n) global invariant loop with sampling** (3 tokens/call using block.number for variation)
3. **Remove `_invariantSaltToTokenConsistency()`** - redundant with inline checks

## Why Sampling 3 Tokens is Safe

| Invariant | Why Safe |
|-----------|----------|
| FAC1 | **Exhaustively verified inline** in `_recordCreatedToken()` at creation time - 100% coverage guaranteed |
| FAC11 | **Exhaustively verified inline** in `_recordCreatedToken()` at creation time - 100% coverage guaranteed |
| FAC2 | **Stateless & immutable** - once a token is recognized by factory, it can't regress. Inline check in `createToken` handler + sampling provides defense-in-depth |
| FAC12 | **Stateless & immutable** - quote token currency cannot change after creation. Sampling across 256 runs with varying `block.number` ensures all tokens get checked eventually |

The global invariant function is now a **sanity check / defense-in-depth** layer, not the primary verification. Primary verification happens inline at the point of state change.

## Invariant Coverage

| Invariant | Before | After |
|-----------|--------|-------|
| FAC1 | O(actors×salts) per call | ✅ Inline in `_recordCreatedToken()` (exhaustive) |
| FAC2 | O(n) per call | ✅ Inline in handler + sampled |
| FAC8 | O(1) | O(1) in `setUp()` (immutable) |
| FAC11 | O(n) per call | ✅ Inline in `_recordCreatedToken()` (exhaustive) |
| FAC12 | O(n) per call | Sampled (stateless check) |

## Benchmark

| Metric | Before | After | Improvement |
|--------|--------|-------|-------------|
| 100 fuzz runs | >90s timeout | ~49s | ✅ Completes |
| 256 runs, 128k calls | N/A | 49s | ✅ |
| Complexity | O(n²) | O(1) amortized | ~n× faster |

## Changes

- `test/invariants/TIP20Factory.t.sol`: Refactored `invariant_globalInvariants()` and `_recordCreatedToken()`

### DIFF
--- a/tips/ref-impls/test/invariants/TIP20Factory.t.sol
+++ b/tips/ref-impls/test/invariants/TIP20Factory.t.sol
@@ -561,9 +561,7 @@ contract TIP20FactoryInvariantTest is InvariantBaseTest {
 
         // TEMPO-FAC1: Verify salt-to-token mapping consistency immediately
         address factoryAddr = factory.getTokenAddress(actor, salt);
-        assertEq(
-            tokenAddr, factoryAddr, "TEMPO-FAC1: Created address inconsistent with factory"
-        );
+        assertEq(tokenAddr, factoryAddr, "TEMPO-FAC1: Created address inconsistent with factory");
 
         // TEMPO-FAC11: Verify token address has correct prefix
         uint160 addrValue = uint160(tokenAddr);


legion2002 (approved):
Good one!

PR #2447, opened 2026-02-04
chore: enable consensus metrics by default on 127.0.0.1:8002

gakonst:
Changes the default consensus metrics address from 127.0.0.1:8001 to 127.0.0.1:8002.

Requested by @zygis.


SuperFluffy (approved):
Some conflict somewhere? Maybe with reth? I seem to vaguely recall that but don't know where.

PR #2449, opened 2026-02-04
poc(ci-execution-tests):  setup test harness

0xrusowsky:
## Execution Layer Regression Testing Framework

Adds `tempo-execution-tests`, a differential testing framework for the Tempo execution layer. Detects behavioral regressions by comparing execution fingerprints between branches—prevents silent changes like the recent incident where refactoring accidentally altered execution results.

### What it does

- Executes JSON test vectors against the EVM across all hardforks (Genesis, T0, T1)
- Generates fingerprints (hashes of gas, output, logs, state) for each test
- Compares fingerprints against `main` to catch unintended behavioral changes
- Validates expected transaction outcomes (success/revert with specific errors)

### Test Vector Schema

```json
{
  "extends": "setup.json",
  "check_regression": true,
  "transactions": [
    {
      "tx_type": "tempo",
      "from": "0x...",
      "calls": [
        {
          "to": "0x20FC000000000000000000000000000000000000",
          "function": "createToken(string,string,string,address,address,bytes32)",
          "args": ["Test Token", "TEST", "USD", "0x20C0...", "0x1111...", "0x0001..."]
        }
      ],
      "outcome": { "success": true }
    }
  ]
}
```

**Call input formats** (mutually exclusive):
- **Human-readable (preferred):** `"function": "transfer(address,uint256)"` + `"args": ["0x...", "1000"]`
- **Raw calldata:** `"calldata": "0xa9059cbb..."`

**Key fields:**
- `extends` — inherit prestate from another vector
- `check_regression` — when `true`, CI builds both your branch and `main`, runs the vector on both, and compares fingerprints. Any difference in gas, output, logs, or state is flagged as a regression.
- `outcome` — assert success or specific revert (`error: "TokenAlreadyExists(address)"`)
- `outcomes` — per-hardfork assertions when behavior intentionally differs across hardforks

### Per-Hardfork Behavior

All vectors run against **every hardfork** (Genesis, T0, T1). When behavior intentionally differs—for example, a new function added in T1 that should revert with `UnknownSelector` (on-chain becomes an empty revert, so it's just `0x`) on older hardforks—use `outcomes` to define expected results per hardfork:

```json
"outcomes": [
  { "hardforks": ["Genesis", "T0"], "outcome": { "success": false, "error": "0x" } },
  { "hardforks": ["T1"], "outcome": { "success": true } }
]
```

### State Capture & Fingerprinting

The `checks` section specifies which state to **capture into the fingerprint**—not assert against expected values. This includes:
- Precompile fields (`checks.precompiles[].fields`)
- Storage slots (`checks.storage`)
- Account nonces (`checks.nonces`)

Gas usage is also captured. Regression detection works by hashing all captured values: if anything changes between branches, the fingerprint changes, and CI flags it. This catches unintended gas changes automatically without needing explicit gas assertions.

### CLI Commands

| Command | Description |
|---------|-------------|
| `run` | Execute all vectors in ./vectors (default) |
| `run -f <file>` | Execute a single vector |
| `run -d <dir>` | Execute all vectors in directory |
| `diff --baseline-binary <path>` | Compare against baseline binary |
| `compare <a.json> <b.json>` | Compare two fingerprint files |

### CI Integration

New workflow `Execution Regression Tests` runs on PRs touching execution-related crates. Builds both PR and `main` binaries, then compares fingerprints for vectors with `check_regression: true`.

### Full documentation
- [README.md](crates/execution-tests/README.md)


crates/execution-tests/src/database.rs
```diff
@@ -0,0 +1,221 @@
...
+    /// Get a reference to the underlying database.
+    pub fn inner(&self) -> &CacheDB<EmptyDB> {
+        &self.db
+    }
+}
+
+/// Compute keccak256 hash of bytes
+fn hash_bytes(data: &Bytes) -> B256 {
+    alloy_primitives::keccak256(data)
+}
```
onbjerg: just inline this

crates/execution-tests/src/database.rs
```diff
@@ -0,0 +1,221 @@
...
+    /// precompiles (TIP403Registry, TIP20Factory, PATH_USD, etc.).
+    pub fn from_prestate(prestate: &Prestate) -> eyre::Result<Self> {
+        let mut db = CacheDB::new(EmptyDB::default());
+
+        // Always run genesis initialization first
+        let admin = prestate
+            .genesis_path_usd
+            .as_ref()
+            .and_then(|g| g.admin)
+            .unwrap_or(genesis::GENESIS_ADMIN);
```
onbjerg, line 34: should we support multiple fee token types?

crates/execution-tests/vectors/tip20_factory/create_token.json
```diff
@@ -0,0 +1,30 @@
...
+    {
+      "tx_type": "tempo",
+      "from": "0x1111111111111111111111111111111111111111",
+      "gas_limit": 1000000,
+      "max_fee_per_gas": "100",
+      "nonce_key": "1",
+      "calls": [
+        {
+          "to": "0x20FC000000000000000000000000000000000000",
+          "input": "0x6813044500000000000000000000000000000000000000000000000000000000000000c00000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000014000000000000000000000000020c000000000000000000000000000000000000000000000000000000000000011111111111111111111111111111111111111110000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000a5465737420546f6b656e000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000004544553540000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000035553440000000000000000000000000000000000000000000000000000000000"
```
onbjerg: small readability ask: maybe we can support something like selector + inputs? alloy should support you doing this. e.g. instead of this blob, it could be the fn name + parameter types, then the parameters

0xrusowsky: great idea 🙌

0xrusowsky: addressed in https://github.com/tempoxyz/tempo/pull/2449/commits/02f8e07946bd161b19d5eb4319f55750278f6cbe

crates/execution-tests/src/database.rs
0xrusowsky: addressed in https://github.com/tempoxyz/tempo/pull/2449/commits/02f8e07946bd161b19d5eb4319f55750278f6cbe

crates/execution-tests/src/database.rs
0xrusowsky: i want to create some setup templates where we deploy tokens and so on.
this way we can preserve a fairly minimal genesis and test more stuff via vector calls

crates/execution-tests/vectors/tip20_factory/create_token.json
```diff
@@ -0,0 +1,38 @@
+{
+  "extends": "setup.json",
+  "description": "Create a TIP20 token via factory. Tests FAC1, FAC2, FAC6, FAC11, FAC12",
+  "check_regression": true,
+  "transactions": [
```
legion2002, line 5: What do you think about generating these txs using the existing invariant handlers, instead of creating them manually? 
I think you should be able to fetch a variety of different legacy and tempo txs in RLP encoded format.

PR #2492, opened 2026-02-05
faucet: switch to expiring nonces (TIP-1009)

gakonst:
Switches the faucet from random 2D nonces to expiring nonces to help control state growth.

## Changes

- **crates/faucet**: Use `with_expiring_nonces()` instead of `with_random_2d_nonces()`
- **crates/alloy**: Add `with_expiring_nonces()` extension method to `TempoProviderBuilderExt`

## Why

Faucet transactions previously used random 2D nonces which contributed to state bloat as each nonce key persists. With expiring nonces (`nonce_key = U256::MAX`, `valid_before = 25s`), the chain can efficiently prune spent nonces after they expire.

Since faucet was identified as the biggest contributor to nonce-based state growth, this should provide immediate relief after the hardfork.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -53,6 +53,7 @@ jobs:
             type=semver,pattern={{version}}
             type=semver,pattern={{major}}.{{minor}}
             type=semver,pattern={{major}}
+            type=raw,value={{version}},enable=${{ github.ref_type == 'tag' }}
             type=raw,value=stable,enable=${{ github.ref_type == 'tag' }}
             type=raw,value=latest,enable={{is_default_branch}}
             type=edge,branch=main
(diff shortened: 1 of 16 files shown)


gakonst:
Fixed both review comments in 14fddcb:

1. **Simplified filler construction** - Now uses `TempoFillers::default()` to match the pattern in `with_random_2d_nonces()`
2. **Removed unused imports** - Cleaned up `ChainIdFiller`, `GasFiller`, and `TxFiller` imports that were only needed for the manual construction

Note: Both methods still create a new `ProviderBuilder::default()` - this is consistent with `with_random_2d_nonces()` and appears to be the intended pattern (the trait constrains the input type to ensure the network is already set correctly).

legion2002 (approved):

PR #2494, opened 2026-02-05
chore(consensus): read EL state with a backoff and emit useful syncing info on failure

SuperFluffy:
This patch implements a constant backoff from 1s to 30s when reading the validator config from execution layer state at the end of an epoch. Spamming this every second was excessive, especially while a node was syncing.

The event level was also lowered to WARN - after all, the node is actively trying to recover.

Finally, the event provides extra information about the execution layer, which potentially helps understand why it currently fails to read state.


crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -1484,20 +1486,24 @@ async fn read_validator_config_with_retry<C: commonware_runtime::Clock>(
         {
             break validators;
         }
+        let retry_after = MIN_RETRY.saturating_mul(attempts).min(MAX_RETRY);
```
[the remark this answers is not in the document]
SuperFluffy: Linear backoffs are desired.

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -1484,20 +1486,24 @@ async fn read_validator_config_with_retry<C: commonware_runtime::Clock>(
...
-                );
-            }
+            warn!(
+                attempts,
+                retry_after = %tempo_telemetry_util::display_duration(retry_after),
+                is_syncing,
+                best_block = %tempo_telemetry_util::display_result(&best_block),
+                target_block,
+                blocks_behind = %tempo_telemetry_util::display_option(&blocks_behind),
+                "reading validator config from contract failed; will retry",
```
joshieDo, line 1505: i feel like this might scare operators, maybe we can add to the msg that this is expected  when we're behind?

joshieDo: even though is_syncing is in there, i feel like it might be missed

hamdiallam: agreed, or separating the two cases into different logs

SuperFluffy: Well, right now it emits errors for hours, so this is an improvement.

`WARN` is appropriate because it is being handled, yet does constitute a hickup in normal operation and should be looked into.

We already have a log for this (from reth), and it is easily missed. That's why this info is added here.

Separating this into separate events feels incorrect because this is just one event.

PR #2495, opened 2026-02-05
prototype: follow mode also tracks consensus

hamdiallam:
`--follow` trusts blocks that are received blindly. Sync is driven in reth `DebugConsensusClient` which polls the execution engine for blocks via `subscribe_blocks` and `get_block`.

Full nodes should also validate blocks against finalization certificates. In this prototype, we let Reth `DebugConsensusClient` drive sync but add a layer of validation by fetching certs and validating them against the dkg keys read from the chain.

This also means that follow nodes can serve consensus rpcs, very useful for developers that want to verify integrity of the chains (i.e bridges).

## Notes

- It is not required that Reth drives sync. It is likely cleaner if we introduce a lighter commonware runtime in verifier mode that stores finalization certs and pushes blocks to reth.


hamdiallam:
Closing in favor of #2551

PR #2496, opened 2026-02-05
fix: add reserved liquidity tstore/check

0xKitsune:
Closes CHAIN-624

### DIFF
--- a/crates/node/tests/it/tip_fee_amm.rs
+++ b/crates/node/tests/it/tip_fee_amm.rs
@@ -641,7 +641,7 @@ async fn test_cant_burn_required_liquidity() -> eyre::Result<()> {
         .await?
         .get_receipt()
         .await?;
-    assert!(burn_receipt.status());
+    assert!(!burn_receipt.status());
 
     Ok(())
 }


crates/precompiles/src/tip_fee_manager/mod.rs
```diff
@@ -25,6 +25,9 @@ pub struct TipFeeManager {
     pools: Mapping<B256, Pool>,
     total_supply: Mapping<B256, U256>,
     liquidity_balances: Mapping<B256, Mapping<Address, U256>>,
+    /// T2+ specific slot to ensure reserved liquidity remains available for pending fee swaps.
+    /// This is subtracted from available reserves when checking liquidity for burn operations.
+    reserved_liquidity: Mapping<B256, u128>,
```
0xrusowsky: can we add some separation between this field and the others + a warning about the transient storage limitations like in `crates/precompiles/src/account_keychain/mod.rs`?

```suggestion

    // WARNING(rusowsky): transient storage slots must always be placed at the very end until the `contract`
    // macro is refactored and has 2 independent layouts (persistent and transient).
    // If new (persistent) storage fields need to be added to the precompile, they must go above this one.

    /// T2+ specific slot to ensure reserved liquidity remains available for pending fee swaps.
    /// This is subtracted from available reserves when checking liquidity for burn operations.
    reserved_liquidity: Mapping<B256, u128>,
```

PR #2499, opened 2026-02-05
fix(tx-pool): update txs after moderato T1 release

0xrusowsky:
## Motivation

Four txpool validator tests broke after T1 activated on MODERATO — they use `SystemTime::now()` and their hardcoded inputs (10 gwei fee, missing `valid_before`, default 1M gas limit) are invalid under T1 rules.

## Solution

- Fixed test inputs to work under the active hardfork: derive `max_fee` from `base_fee()`, add `valid_before` for expiring nonces, use tx gas limit cap for gas limit.
- Added `TempoHardfork::tx_gas_limit_cap()` with `TEMPO_T1_TX_GAS_LIMIT_CAP` and `TEMPO_T1_GENERAL_GAS_LIMIT` constants in `spec.rs`, replacing the standalone `TIP1000_TX_GAS_LIMIT_CAP` and simplifying EVM config.

### DIFF
--- a/crates/evm/src/lib.rs
+++ b/crates/evm/src/lib.rs
@@ -333,7 +333,7 @@ mod tests {
         // Verify TIP-1000 gas limit cap is set
         assert_eq!(
             evm_env.cfg_env.tx_gas_limit_cap,
-            Some(TIP1000_TX_GAS_LIMIT_CAP),
+            Some(tempo_chainspec::spec::TEMPO_T1_TX_GAS_LIMIT_CAP),
             "TIP-1000 requires 30M gas limit cap for T1 hardfork"
         );
     }


crates/evm/src/lib.rs
```diff
@@ -133,10 +130,7 @@ impl ConfigureEvm for TempoEvmConfig {
 
         // Apply TIP-1000 gas params for T1 hardfork.
         let mut cfg_env = cfg_env.with_spec_and_gas_params(spec, tempo_gas_params(spec));
-        // Set TIP-1000 Transaction gas limit cap (30M) for T1 hardfork.
-        if spec.is_t1() {
-            cfg_env.tx_gas_limit_cap = Some(TIP1000_TX_GAS_LIMIT_CAP);
-        }
+        cfg_env.tx_gas_limit_cap = spec.tx_gas_limit_cap();
```
0xKitsune, line 133: Nice

crates/chainspec/src/hardfork.rs
```diff
@@ -77,11 +77,21 @@ impl TempoHardfork {
     }
 
     /// Returns the fixed general gas limit for T1+, or None for pre-T1.
-    /// - T1+: 30M gas (fixed)
     /// - Pre-T1: None
+    /// - T1+: 30M gas (fixed)
     pub const fn general_gas_limit(&self) -> Option<u64> {
         match self {
-            Self::T1 | Self::T2 => Some(30_000_000),
+            Self::T1 | Self::T2 => Some(crate::spec::TEMPO_T1_GENERAL_GAS_LIMIT),
```
0xKitsune, line 84: nit: It looks like this is imported twice, consider importing once at the top of the file and using `TEMPO_T1_GENERAL_GAS_LIMIT` directly.

PR #2512, opened 2026-02-06
fix(evm): add clear panic message for empty validator set division

howydev:
## Summary

Fixes CHAIN-681 (TMPO2-36): Panic on empty validator set.

Two functions divide by the number of validators without guarding for zero. An empty validator set is a fatal invariant violation (only possible during live consensus, never during sync), so panicking is the correct behavior — but the bare division-by-zero gives an unhelpful panic message.

This replaces the bare divisions with `.checked_div().expect("validator set must not be empty")` for a clear panic message at both sites:

1. **`validate_shared_gas`** in `crates/evm/src/block.rs`
2. **`build_subblock`** in `crates/commonware-node/src/subblocks.rs`

### DIFF
--- a/crates/evm/src/block.rs
+++ b/crates/evm/src/block.rs
@@ -186,7 +186,10 @@ where
         let Some(validator_set) = &self.validator_set else {
             return Ok(());
         };
-        let gas_per_subblock = self.shared_gas_limit / validator_set.len() as u64;
+        let gas_per_subblock = self
+            .shared_gas_limit
+            .checked_div(validator_set.len() as u64)
+            .expect("validator set must not be empty");
 
         let mut incentive_gas = 0;
         let mut seen = HashSet::new();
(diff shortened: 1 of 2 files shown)


SuperFluffy (commented):
I'm not sure that this shouldn't panic.

If this is an invariant violation, then this should panic instead of silently breaking a feature.

AFAIK this value is set by the node and never externally. Can this be attacked in any way?

crates/evm/src/block.rs
```diff
@@ -186,6 +186,9 @@ where
         let Some(validator_set) = &self.validator_set else {
             return Ok(());
         };
+        if validator_set.is_empty() {
+            return Err(BlockValidationError::msg("empty validator set"));
+        }
         let gas_per_subblock = self.shared_gas_limit / validator_set.len() as u64;
```
SuperFluffy: IMO here the division should be `.checked_sub` with an explicit expect message.

SuperFluffy (changes requested):
Blocking this PR until the DoS vector is described.

I thing the panic here is appropriate and desirable (maybe should have a better panic msg).

PR #2513, opened 2026-02-06
fix(stablecoin-dex): add tick spacing checks on price conversion functions

0xKitsune:
Closes CHAIN-674

This PR adds tick spacing checks on `tick_to_price` and `price_to_tick` functions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12340,6 +12340,7 @@ dependencies = [
  "alloy-signer-local",
  "alloy-sol-types",
  "futures",
+ "itertools 0.14.0",
  "metrics",
  "parking_lot",
  "reth-chainspec",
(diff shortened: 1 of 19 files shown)


tips/ref-impls/src/StablecoinDEX.sol
```diff
@@ -88,7 +89,9 @@ contract StablecoinDEX is IStablecoinDEX {
             tick = int16(int32(price) - int32(PRICE_SCALE));
             revert IStablecoinDEX.TickOutOfBounds(tick);
         }
-        return int16(int32(price) - int32(PRICE_SCALE));
+        tick = int16(int32(price) - int32(PRICE_SCALE));
```
fgimenez: nit, this is also calculated at line 89, we could refactor and move it before the if

PR #2521, opened 2026-02-06
fix(precompile-macros): ensure multi-slot mapping keys are abi-encoded

0xrusowsky:
ref: `CHAIN-680`, `SIGP-42`

this PR limits types which can impl `StorageKey` to `sealed::OnlyPrimitives` (single-word primitive types).

this covers our existing and foreseeable needs, and allows a small perf gain by not allocating capacity for a vec

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -136,3 +136,20 @@ jobs:
         env:
           VERGEN_GIT_SHA: ${{ github.sha }}
           VERGEN_GIT_SHA_SHORT: ${{ steps.shortsha.outputs.shortsha }}
+
+      - name: Install cosign
+        uses: sigstore/cosign-installer@d7d6bc7722e3daa8354c50bcb52f4837da5e9b6a # v3.8.1
+
+      - name: Sign Docker images
+        shell: bash
+        run: |
+          set -euo pipefail
+          images=(
+            "${{ steps.meta-tempo.outputs.tags }}"
+            "${{ steps.meta-tempo-bench.outputs.tags }}"
+            "${{ steps.meta-tempo-sidecar.outputs.tags }}"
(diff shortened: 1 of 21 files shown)


crates/precompiles/src/storage/types/mod.rs
```diff
@@ -304,17 +304,13 @@ pub trait StorageKey {
     fn mapping_slot(&self, slot: U256) -> U256 {
```
0xKitsune, line 327: nit: 
Doc comments for this function still say "Left-pads the key to the nearest 32-byte multiple". It now pads to exactly one 32-byte word. Should read something like: "Left-pads the key to 32 bytes (single word primitives only), concatenates with the slot, and hashes."

crates/precompiles/src/storage/types/mod.rs
```diff
@@ -293,8 +293,8 @@ impl<T: Packable> Storable for T {
 /// Keys are hashed using keccak256 along with the mapping's base slot
 /// to determine the final storage location. This trait provides the
 /// byte representation used in that hash.
-pub trait StorageKey {
-    /// Returns a byte slice for this type.
+pub trait StorageKey: sealed::OnlyPrimitives {
```
0xKitsune, line 320: For clarity, why do we need to add `OnlyPrimitives` here?

0xKitsune (approved):

crates/precompiles/src/storage/types/mod.rs
0xrusowsky: this is how we address the audit concern.

by limiting `StorageKey` to types which impl `OnlyPrimitives`, we become we prevent dynamic/fixed-size arrays, strings, bytes, and structs from being mapping keys.

will make sure to document it too, for extra clarity

PR #2522, opened 2026-02-06
fix(txpool): clear and repopulate AmmLiquidityCache on reorg

fgimenez:
Closes SIGP-20

### DIFF
--- a/crates/transaction-pool/src/amm.rs
+++ b/crates/transaction-pool/src/amm.rs
@@ -4,6 +4,7 @@ use alloy_primitives::{
     Address, U256,
     map::{AddressMap, HashMap, U256Map},
 };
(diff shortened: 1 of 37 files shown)


crates/transaction-pool/src/amm.rs
```diff
@@ -34,7 +34,9 @@ impl AmmLiquidityCache {
     /// validator fee tokens of the latest blocks.
     pub fn new<Client>(client: Client) -> ProviderResult<Self>
     where
-        Client: StateProviderFactory + BlockReader + ChainSpecProvider<ChainSpec = TempoChainSpec>,
```
fgimenez, line 38: `BlockReader` bound was overkill here, `HeaderProvider` is enough

PR #2528, opened 2026-02-06
fix(precompiles): refund spending limit for unused gas fees

legion2002:
## Summary

Fixes **SIGP-8 / TMPO2-17**: Access key spending limits were permanently reduced by the full max fee amount even when a portion was refunded after execution.

## Problem

In `transfer_fee_pre_tx()`, `check_and_update_spending_limit()` deducts the full max fee from the access key's spending limit. After execution, `transfer_fee_post_tx()` refunds unused tokens back to the user but never restores the spending limit.

This means every transaction permanently consumes the max gas fee from the spending limit, regardless of actual gas used.

## Fix

- Added `refund_spending()` and `refund_spending_limit()` to `AccountKeychain` (mirrors `verify_and_update_spending` / `authorize_transfer` with the same guards for main key and tx_origin checks)
- Added `refund_spending_limit()` wrapper to `TIP20Token`
- Called it in `transfer_fee_post_tx()` to restore the spending limit by the refunded amount

## Tests

- `test_refund_spending_limit_restores_limit` — verifies limit is correctly restored after partial refund
- `test_refund_spending_limit_noop_for_main_key` — confirms no-op for main key (no spending limits)

### DIFF
--- a/crates/precompiles/src/account_keychain/mod.rs
+++ b/crates/precompiles/src/account_keychain/mod.rs
@@ -476,6 +476,8 @@ impl AccountKeychain {
             return Ok(());
         }
 
+        // Silently skip refund if the key was revoked or expired — the fee was already
+        // collected and the key is no longer active, so there is nothing to restore.
         let key = match self.load_active_key(account, transaction_key) {
             Ok(key) => key,
             Err(_) => return Ok(()),


crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -446,6 +446,57 @@ impl AccountKeychain {
...
+    pub fn refund_spending(
+        &mut self,
+        account: Address,
+        key_id: Address,
+        token: Address,
+        amount: U256,
+    ) -> Result<()> {
+        if key_id == Address::ZERO {
+            return Ok(());
+        }
```
0xKitsune: We are checking if `key_id` is zero twice. Once in `refund_spending` and once in `refund_spending_limits`.

crates/precompiles/src/tip20/mod.rs
```diff
@@ -799,6 +804,10 @@ impl TIP20Token {
             return Ok(());
         }
 
+        if self.storage.spec().is_t2() {
+            self.refund_spending_limit(to, refund)?;
+        }
```
0xKitsune, line 804: Do we need to update invariants to cover this post T2?

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -446,6 +446,57 @@ impl AccountKeychain {
         self.spending_limits[limit_key][token].write(remaining - amount)
     }
 
+    /// Internal: Refund spending limit for a token transfer (e.g., unused gas fees)
```
0xKitsune: nit: Internal functions are a concept in solidity not Rust. Can we update this comment. 


```suggestion
    /// Refund spending limit for a token transfer (e.g., unused gas fees)
```

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -446,6 +446,57 @@ impl AccountKeychain {
...
+        if !key.enforce_limits {
+            return Ok(());
+        }
+
+        let limit_key = Self::spending_limit_key(account, key_id);
+        let remaining = self.spending_limits[limit_key][token].read()?;
+
+        let new_remaining = remaining
+            .checked_add(amount)
+            .ok_or(TempoPrecompileError::under_overflow())?;
```
0xKitsune: If `refund_spending` is ever called with an incorrect or excessive amount (bug in caller, etc.), the remaining limit can exceed the original authorized limit. `verify_and_update_spending` deducts and enforces a floor of 0, but the refund path has no corresponding ceiling.

Wdyt about updating to `new_remaining = min(remaining + amount, original_limit)`?

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -446,6 +446,57 @@ impl AccountKeychain {
...
+            return Ok(());
+        }
+
+        let key = self.keys[account][key_id].read()?;
+        if !key.enforce_limits {
+            return Ok(());
+        }
+
+        let limit_key = Self::spending_limit_key(account, key_id);
+        let remaining = self.spending_limits[limit_key][token].read()?;
```
0xKitsune: Is there a reason we are not using `load_active_key`? Is loading from storage preferred here for some reason?

crates/precompiles/src/tip20/mod.rs
```diff
@@ -703,6 +703,11 @@ impl TIP20Token {
         AccountKeychain::new().authorize_transfer(from, self.address, amount)
     }
 
+    /// Refunds spending limits for access keys after a fee refund.
+    pub fn refund_spending_limit(&mut self, from: Address, amount: U256) -> Result<()> {
```
0xKitsune: This function should be private. 
```suggestion
fn refund_spending_limit(&mut self, from: Address, amount: U256) -> Result<()> {
```

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -1158,4 +1209,74 @@ mod tests {
             Ok(())
         })
     }
+
+    #[test]
+    fn test_refund_spending_limit_restores_limit() -> eyre::Result<()> {
```
0xKitsune, line 1232: Can we add tests to assert refund after key revocation and that refund that would exceed original limit is clamped?

crates/precompiles/src/tip20/mod.rs
0xrusowsky: i'd even say we don't even need this fn and we should directly inline it, as it is just 1 line

crates/precompiles/src/account_keychain/mod.rs
0xrusowsky: agreed. also it is weird to create 2 fns (one of which is internal to the other) for this.
we should inline this logic in `refund_sepnding_limits`

crates/precompiles/src/account_keychain/mod.rs
0xrusowsky: i like this proposal, we should be defensive and enforce the invariant directly in the code (specially if we can do it without extra SLOADs)

legion2002: The problem is that we don't have the `original_limit` stored anywhere. To do this suggestion we would need to add an extra storage mapping to retain the original limit. 

Feel like it should be fine to trust `refund_spending` inputs here, because if they are incorrect, then we have other bigger problems to worry about probably.

crates/precompiles/src/tip20/mod.rs
legion2002: no update to old invariants, but added new ones to conver this.

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -494,7 +468,21 @@ impl AccountKeychain {
             return Ok(());
         }
 
-        self.refund_spending(account, transaction_key, token, amount)
+        let key = match self.load_active_key(account, transaction_key) {
+            Ok(key) => key,
+            Err(_) => return Ok(()),
```
0xrusowsky, line 483: should we add a comment for clarity on why we not propagate the error?

legion2002: yup done

PR #2529, opened 2026-02-06
fix(tx-pool): add missing EXISTING_NONCE_KEY_GAS in 2D nonce validation

rakita:
## Summary
- Add the missing `else if !tx.nonce_key.is_zero()` branch in the txpool validator's T1 nonce gas accounting, matching the logic already present in `handler.rs`
- Without this, AA transactions with an existing 2D nonce key (`nonce_key != 0 && nonce > 0`) could pass pool admission with a `gas_limit` that's 5,000 gas too low, enabling cheaper spam that fails at execution time

## Test plan
- [x] `cargo build -p tempo-transaction-pool` — compiles cleanly
- [x] `cargo test -p tempo-transaction-pool` — all 188 tests pass
- [ ] `cargo test -p tempo-node --test it` — integration tests covering pool validation

### DIFF
--- a/crates/transaction-pool/src/validator.rs
+++ b/crates/transaction-pool/src/validator.rs
@@ -410,7 +410,7 @@ where
             } else if !tx.nonce_key.is_zero() {
                 // Existing 2D nonce key (nonce > 0): cold SLOAD + warm SSTORE reset
                 // TIP-1000 Invariant 3: existing state updates charge 5,000 gas
-                init_and_floor_gas.initial_gas += EXISTING_NONCE_KEY_GAS;
+                init_and_floor_gas.initial_gas += spec.gas_existing_nonce_key();
             }
             // In CREATE tx with 2d nonce, check if account.nonce is 0, if so, add 250,000 gas.
             // This covers caller creation of account.


legion2002 (approved):

crates/transaction-pool/src/validator.rs
```diff
@@ -401,6 +401,10 @@ where
                 // TIP-1000: Storage pricing updates for launch
                 // Tempo transactions with any `nonce_key` and `nonce == 0` require an additional 250,000 gas
                 init_and_floor_gas.initial_gas += gas_params.get(GasId::new_account_cost());
+            } else if !tx.nonce_key.is_zero() {
+                // Existing 2D nonce key (nonce > 0): cold SLOAD + warm SSTORE reset
+                // TIP-1000 Invariant 3: existing state updates charge 5,000 gas
+                init_and_floor_gas.initial_gas += EXISTING_NONCE_KEY_GAS;
```
0xKitsune: We should plan to use `spec.gas_existing_nonce_key()` instead of the hardcoded `EXISTING_NONCE_KEY_GAS` constant since T2 updates this value.

0xrusowsky: agreed

PR #2533, opened 2026-02-07
fix(handler)!: nonce key gas repricing

0xrusowsky:
ref: `SIGP-7`

### DIFF
--- a/crates/primitives/src/transaction/envelope.rs
+++ b/crates/primitives/src/transaction/envelope.rs
@@ -36,6 +36,7 @@ pub const TEMPO_SYSTEM_TX_SENDER: Address = Address::ZERO;
     serde_cfg(feature = "serde")
 )]
 #[cfg_attr(test, reth_codecs::add_arbitrary_tests(compact, rlp))]
+#[allow(clippy::large_enum_variant)]
 pub enum TempoTxEnvelope {
     /// Legacy transaction (type 0x00)
     #[envelope(ty = 0)]


crates/chainspec/src/spec.rs
```diff
@@ -34,6 +40,14 @@ pub const TEMPO_T1_TX_GAS_LIMIT_CAP: u64 = 30_000_000;
 pub const SYSTEM_TX_COUNT: usize = 1;
 pub const SYSTEM_TX_ADDRESSES: [Address; SYSTEM_TX_COUNT] = [Address::ZERO];
 
+/// Gas cost for using an existing 2D nonce key (cold SLOAD + 2*warm SLOAD + warm SSTORE reset)
+pub const TEMPO_T1_EXISTING_NONCE_KEY_GAS: u64 = COLD_SLOAD + WARM_SSTORE_RESET;
+pub const TEMPO_T2_EXISTING_NONCE_KEY_GAS: u64 = COLD_SLOAD + 2 * WARM_SLOAD + WARM_SSTORE_RESET;
```
0xKitsune: Can we update to `TEMPO_T1_EXISTING_NONCE_KEY_GAS + (2 * WARM_SLOAD)`?

crates/chainspec/src/spec.rs
```diff
@@ -34,6 +40,14 @@ pub const TEMPO_T1_TX_GAS_LIMIT_CAP: u64 = 30_000_000;
...
+
+/// Gas cost for using a new 2D nonce key (cold SLOAD + 2*warm SLOAD + SSTORE set for 0 -> non-zero)
+pub const TEMPO_T1_NEW_NONCE_KEY_GAS: u64 = COLD_SLOAD + SSTORE_SET;
+pub const TEMPO_T2_NEW_NONCE_KEY_GAS: u64 = COLD_SLOAD + 2 * WARM_SLOAD + SSTORE_SET;
```
0xKitsune: Same here can we update to Can we update to `TEMPO_T1_NEW_NONCE_KEY_GAS + (2 * WARM_SLOAD)?`

crates/revm/src/handler.rs
```diff
@@ -2398,12 +2390,16 @@ mod tests {
 
         // Base intrinsic gas for AA tx with one call (no 2D nonce)
         const BASE_INTRINSIC_GAS: u64 = 21_000;
+        const SPEC: TempoHardfork = TempoHardfork::T0;
+        const NEW_NONCE_KEY_GAS: u64 = SPEC.gas_new_nonce_key();
+        const EXISTING_NONCE_KEY_GAS: u64 = SPEC.gas_existing_nonce_key();
```
0xKitsune: nit: why not use `spec.gas_existing_nonce_key()` where this is needed. Why are we creating new consts here?

crates/revm/src/handler.rs
```diff
@@ -2398,12 +2390,16 @@ mod tests {
 
         // Base intrinsic gas for AA tx with one call (no 2D nonce)
         const BASE_INTRINSIC_GAS: u64 = 21_000;
+        const SPEC: TempoHardfork = TempoHardfork::T0;
```
0xKitsune: Looks like we are only testing `T0`? Can we update to assert the new behaviors?

crates/revm/src/handler.rs
0xrusowsky: just so that i could leave tests impls untouched, but we can definitely do what u suggest

PR #2538, opened 2026-02-07
perf: Optimize TIP20 approve/transfer storage access

malik672:
This change reduces storage slot recomputation in TIP20 approve and transfer flows by caching mapping handlers within a call. It also adds a lazy load path for approvals so we don’t read the current allowance unless spending limits actually apply.


malik672:
> hey @malik672 first of all thanks for trying to help with perf. some comments for your proposals:
> 
> * i don't think "cache the handler for this allowance slot" helps at all, as our mappings already use `HandlerCache` to avoid recomputing the slots (prevents us from computing keccak multiple times)
> * i like the `authorize_approve_lazy` optimization, but i kinda don't like the closure pattern as it could be confusing. Instead i'd prefer to split `fn authorize_approve` into smth like:
>   ```rust
>   /// Returns the transaction key if spending limits should be enforced.
>   fn spending_limit_tx_key(&self, account: Address) -> Result<Option<Address>>;
>   
>   /// Authorize a token approval given a transaction key.
>   fn authorize_approve(
>      &mut self,
>      account: Address,
>      token: Address,
>      old_approval: U256,
>      new_approval: U256,
>      tx_key: Address,
>   ) -> Result<()>;
>   ```
> 
> also note that TIP20's approve would need to gate keep the optimization under `TempoHardfork::T2` as changing gas is a breaking change

fixed

0xrusowsky:
@malik672 the reason the benchmarks show improvement on `tip20_approve` is because that codepath goes through `authorize_approve` — which the PR optimizes by skipping an SLOAD when no spending limits apply (which is great). That's the win. You're seeing the effect of the `spending_limit_tx_key` split, not the handler reuse.

as prev said, our `Mapping` already uses `HandlerCache` internally avoiding the keccak recomputation. Your optimization pattern - which comes at the expense of readibility - only saves a `HashMap::get` per reuse which is negligible compared to an extra SLOAD.

drop the manual handler reuse changes in `_transfer` and `transfer_from`, and the PR will be g2g.

0xrusowsky (approved):
g2g from my side.

pending reviews from @0xKitsune and @legion2002 before we can merge

0xrusowsky:
@malik672 can you run the formatter to clear CI?

PR #2541, opened 2026-02-08
feat(ref-impls): add TempoStreamChannel streaming payment escrow

brendanjryan:
## Summary

Add the TempoStreamChannel contract to `tips/ref-impls/` — a unidirectional payment channel escrow for streaming payments using EIP-712 signed cumulative vouchers.

**Deployed on Moderato (testnet) at:** [`0x9d136eEa063eDE5418A6BC7bEafF009bBb6CFa70`](https://explore.testnet.tempo.xyz/address/0x9d136eEa063eDE5418A6BC7bEafF009bBb6CFa70?tab=contract&live=false)

**Spec:** https://paymentauth.tempo.xyz/draft-tempo-stream-00

## What's included

- `src/TempoStreamChannel.sol` — exact copy of deployed contract
- `src/interfaces/ITempoStreamChannel.sol` — interface matching the implementation
- `test/TempoStreamChannel.t.sol` — 21 unit/fuzz tests + 2 malicious token tests (all passing)
- `solady` submodule added for ECDSA, EIP712, ReentrancyGuard

### DIFF
--- a/tips/ref-impls/src/TempoStreamChannel.sol
+++ b/tips/ref-impls/src/TempoStreamChannel.sol
@@ -308,8 +308,9 @@ contract TempoStreamChannel is EIP712 {
         }
 
         uint128 settledAmount = channel.settled;
+        uint128 delta = 0;
 
-        // If cumulativeAmount > settled, validate and settle the voucher
+        // If cumulativeAmount > settled, validate the voucher
         if (cumulativeAmount > settledAmount) {
             if (cumulativeAmount > channel.deposit) {
                 revert AmountExceedsDeposit();
@@ -327,19 +328,22 @@ contract TempoStreamChannel is EIP712 {
                 revert InvalidSignature();
             }
 
-            uint128 delta = cumulativeAmount - settledAmount;
+            delta = cumulativeAmount - settledAmount;
             settledAmount = cumulativeAmount;
             channel.settled = cumulativeAmount;
+        }
 
+        // Effects before interactions
+        uint128 refund = channel.deposit - settledAmount;
+        channel.finalized = true;
+
+        // Interactions
+        if (delta > 0) {
             bool success = ITIP20(channel.token).transfer(channel.payee, delta);
             if (!success) {
                 revert TransferFailed();
             }
         }
-
-        // Refund remainder to payer
-        uint128 refund = channel.deposit - settledAmount;
-        channel.finalized = true;
 
         if (refund > 0) {
             bool success = ITIP20(channel.token).transfer(channel.payer, refund);


crates/primitives/src/transaction/envelope.rs
```diff
@@ -36,7 +36,6 @@ pub const TEMPO_SYSTEM_TX_SENDER: Address = Address::ZERO;
     serde_cfg(feature = "serde")
 )]
 #[cfg_attr(test, reth_codecs::add_arbitrary_tests(compact, rlp))]
-#[expect(clippy::large_enum_variant)]
```
legion2002: is this accidental?

tips/ref-impls/test/TempoStreamChannel.t.sol
```diff
@@ -0,0 +1,919 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.20;
+
+import { TempoStreamChannel } from "../src/TempoStreamChannel.sol";
+import { Test, console } from "forge-std/Test.sol";
+
+contract MockTIP20 {
```
legion2002: we don't need to create this mockTIP20 contract, we can just use the factory to deploy a new TIP 20 in the tests

tips/ref-impls/test/TempoStreamChannel.t.sol
```diff
@@ -0,0 +1,919 @@
...
+        uint256 net = amount - fee;
+        balanceOf[from] -= amount;
+        balanceOf[to] += net;
+        allowance[from][msg.sender] -= amount;
+        return true;
+    }
+
+}
+
+contract NonTransferringToken {
```
legion2002: is the TempoStreamChannel only supposed to work with TIP20s? In this case, we don't need all these negative tests, because the TIP 20 bytecode is guaranteed to be consistent

tips/ref-impls/src/TempoStreamChannel.sol
```diff
@@ -0,0 +1,478 @@
...
+
+/**
+ * @title TempoStreamChannel
+ * @notice Unidirectional payment channel escrow for streaming payments.
+ * @dev Users deposit TIP-20 tokens, sign cumulative vouchers, and servers
+ *      can settle or close at any time. Channels have no expiry - they are
+ *      closed either cooperatively by the server or after a grace period
+ *      following a user's close request.
+ */
+contract TempoStreamChannel is EIP712, ReentrancyGuard {
```
legion2002: Since this contract does not have an owner, we can deploy it anyway we want, without having to worry about admin keys, create2 etc.

If you want to set this to a nice vanity address, which we convert to a precompile later, we can also do that

tips/ref-impls/src/TempoStreamChannel.sol
```diff
@@ -0,0 +1,478 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.20;
+
+import { ECDSA } from "solady/utils/ECDSA.sol";
+import { EIP712 } from "solady/utils/EIP712.sol";
+import { ReentrancyGuard } from "solady/utils/ReentrancyGuard.sol";
+
+interface ITIP20 {
```
legion2002: This interface is already available under `/tips/ref-impls/src/interfaces/ITIP20.sol`

tips/ref-impls/src/TempoStreamChannel.sol
```diff
@@ -0,0 +1,478 @@
...
+            payee: payee,
+            token: token,
+            authorizedSigner: authorizedSigner,
+            deposit: deposit,
+            settled: 0,
+            closeRequestedAt: 0,
+            finalized: false
+        });
+
+        emit ChannelOpened(channelId, msg.sender, payee, token, authorizedSigner, deposit);
```
legion2002: is skipping salt here intentional?

tips/ref-impls/src/TempoStreamChannel.sol
```diff
@@ -0,0 +1,478 @@
...
+        returns (bytes32 channelId)
+    {
+        channelId = computeChannelId(msg.sender, payee, token, salt, authorizedSigner);
+
+        if (channels[channelId].payer != address(0)) {
+            revert ChannelAlreadyExists();
+        }
+
+        // Transfer tokens to this contract
+        bool success = ITIP20(token).transferFrom(msg.sender, address(this), deposit);
```
legion2002: we can put this at the bottom, to follow the checks-effects-interaction pattern, and remove the reentrancy guard to save an SLOAD

tips/ref-impls/src/TempoStreamChannel.sol
```diff
@@ -0,0 +1,478 @@
...
+        }
+
+        emit ChannelClosed(channelId, channel.payer, channel.payee, settledAmount, refund);
+    }
+
+    /**
+     * @notice Withdraw remaining funds after close grace period.
+     * @param channelId The channel to withdraw from
+     */
+    function withdraw(bytes32 channelId) external nonReentrant {
```
legion2002: nonReentrant is not needed here, because of checks-effects-interaction pattern. Also TIP 20s are guaranteed to not make external calls

tips/ref-impls/src/TempoStreamChannel.sol
```diff
@@ -0,0 +1,478 @@
...
+            channelId, channel.payer, channel.payee, cumulativeAmount, delta, channel.settled
+        );
+    }
+
+    /**
+     * @notice Add more funds to a channel.
+     * @param channelId The channel to top up
+     * @param additionalDeposit Amount to add
+     */
+    function topUp(bytes32 channelId, uint128 additionalDeposit) external nonReentrant {
```
legion2002: Is there a reason to make these `additionalDeposit` uint128? 
On the solidity side, having this as u256 is slightly better.

tips/ref-impls/src/TempoStreamChannel.sol
brendanjryan: good check -- lets keep salt.

PR #2551, opened 2026-02-09
feat: consensus follower engine

hamdiallam:
Follower nodes should also do verify and drive sync through CL FCU. 

`--follow && --follow.experimental.certify=true` spins up a lightweight consensus engine to drive fcu to the EL
* On startup, if there's snapshotted marshal state, the identity scheme picks up where the snapshot left off. Otherwise bootstraps itself from the upstream provider
* The driver ensures identity schemes are updated exclusively through parsed boundary blocks
* The same archive is used for finalizations and finalizations-by-height, meaning validator snapshots can be used for follower nodes. And the `consensus_` rpcs are available to follower nodes

## Migration
* `follow.experimental.certify=false` by default. `--follow` preserves existing behavior (unsafe)
* Our url that external users use with `--follow` need to first be updated to serve consensus rpcs. Once snapshots also include this consensus state we can flip this to be the default

## E2E Tests
* Follower can bootstrap itself from either a validator or another follower. First identity scheme is trusted
* If a snapshotted/archive state is used, the follower continues with the last finalized height


crates/commonware-node/src/executor/actor.rs
```diff
@@ -449,23 +449,17 @@ where
...
-            HeadOrFinalized::Finalized,
-            block.height(),
-            block.digest(),
-            oneshot::channel().0,
-        )
-        .await
-        .wrap_err("failed canonicalizing finalized block")?;
-
-        let block = block.into_inner();
+        // Send new_payload first to insert the block into the EL.
```
SuperFluffy: @joshieDo Would like your input on this because this reverses what I thought so far was an intentional order of operation (FCU first, then NewPayload).

crates/commonware-node/src/consensus/block.rs
SuperFluffy, line 1: Probably shouldn't all be pub.

crates/commonware-node/src/consensus/engine.rs
```diff
@@ -166,101 +154,24 @@ where
         };
         let scheme_provider = SchemeProvider::new();
 
-        const FINALIZATIONS_BY_HEIGHT: &str = "finalizations-by-height";
+        // Initialize archives using shared format (compatible with follow engine)
```
SuperFluffy: Remove comment

crates/commonware-node/src/consensus/engine.rs
```diff
@@ -166,101 +154,24 @@ where
...
-                freezer_key_write_buffer: WRITE_BUFFER,
-                freezer_value_write_buffer: WRITE_BUFFER,
-                ordinal_write_buffer: WRITE_BUFFER,
-            },
+            &self.partition_prefix,
+            page_cache_ref.clone(),
         )
-        .await
-        .wrap_err("failed to initialize finalizations by height archive")?;
+        .await?;
```
SuperFluffy: Bring back context

bin/tempo/src/main.rs
```diff
@@ -161,13 +163,83 @@ fn main() -> eyre::Result<()> {
...
+            });
+
+            let runtime_config = commonware_runtime::tokio::Config::default()
+                .with_tcp_nodelay(Some(true))
+                .with_worker_threads(args.consensus.worker_threads)
+                .with_storage_directory(follow_storage)
+                .with_catch_panics(true);
+
+            let runner = commonware_runtime::tokio::Runner::new(runtime_config);
+            runner.start(async move |ctx| {
```
SuperFluffy: Merge runtime initialization and execution between the follow-stack and the validator-stack.

crates/commonware-node/src/lib.rs
```diff
@@ -143,48 +177,78 @@ pub async fn run_consensus_stack(
...
+        partition_prefix: "engine".into(),
+
+        epoch_length: config.epoch_length,
+        mailbox_size: config.mailbox_size,
+        fcu_heartbeat_interval: config.fcu_heartbeat_interval,
+        driver_config: follow::FollowDriverConfig::default(),
+    };
+
+    // The shutdown flag is managed by the caller via cancellation
+    let shutdown = Arc::new(AtomicBool::new(false));
```
SuperFluffy: This should be done via the techniques commonware intended (we need to actually spend some effort nailing that down though).

crates/commonware-node/src/follow/engine.rs
```diff
@@ -0,0 +1,203 @@
...
+    let executor_handle = executor_actor.start();
+    let feed_handle = feed_actor.start();
+
+    info_span!("follower").in_scope(|| info!("starting follow mode driver"));
+    let driver_result = driver.run(shutdown).await;
+
+    // Clean up
+    drop(marshal_handle);
+    drop(feed_handle);
+    drop(executor_handle);
```
SuperFluffy: All these drops are unnecessary - the handles should be awaited instead to ensure the actors don't go down without noticing.

crates/commonware-node/src/follow/driver.rs
```diff
@@ -0,0 +1,351 @@
...
+        Ok(())
+    }
+
+    /// Sync a single height.
+    async fn sync_next_height(&mut self) -> eyre::Result<SyncResult> {
+        let height = self.current_height;
+        let mut retries = 0;
+
+        loop {
+            match self.try_sync_height(height).await {
```
SuperFluffy: IMO this should be push not pull. Follower subscribes to node -> nod sends updates.

crates/commonware-node/src/follow/engine.rs
```diff
@@ -0,0 +1,203 @@
...
+    info_span!("follower").in_scope(|| {
+        info!(
+            last_finalized_height = last_finalized_height.get(),
+            "initialized marshal for follower"
+        )
+    });
+
+    // Initialize executor (handles forwarding blocks to EL and FCU heartbeats)
+    let (executor_actor, executor_mailbox) = crate::executor::init(
+        context.clone(),
```
SuperFluffy: `context.with_label`

crates/commonware-node/src/follow/engine.rs
```diff
@@ -0,0 +1,203 @@
...
+    let epoch_strategy = FixedEpocher::new(NZU64!(config.epoch_length));
+
+    // Initialize marshal (use type alias to satisfy type inference)
+    #[allow(unused_variables)]
+    let (marshal_actor, marshal_mailbox, last_finalized_height): (
+        marshal_alias::Actor<TContext>,
+        marshal_alias::Mailbox,
+        _,
+    ) = marshal::Actor::init(
+        context.clone(),
```
SuperFluffy: `context.with_label`

crates/commonware-node/src/follow/engine.rs
```diff
@@ -0,0 +1,203 @@
...
+
+    // Start components
+    // Marshal is started with:
+    // - executor_mailbox as reporter (receives Update<Block> for forwarding to Reth)
+    // - null_broadcast (never used - driver provides blocks via verified())
+    // - null_resolver (never used - no P2P in follow mode)
+    let marshal_handle = marshal_actor.start(
+        executor_mailbox,
+        null_broadcast,
+        (resolver_rx, null_resolver),
```
SuperFluffy: I think for both (resolver and buffer), we should plug into the marshal actor's model of receiving and forwarding blocks so that the flow looks the same as in the validator engine.

SuperFluffy (commented):
I like this, but I think we should have the "fetcher" plug into the exististing marshal logic so that we have the same flow as for a normal validator:

validator -> marshal -> executor -> execution engine

crates/commonware-node/src/executor/actor.rs
hamdiallam: Let me double check change is still needed. Might be leftover from a first attempt at making this work

crates/commonware-node/src/follow/engine.rs
hamdiallam: Yeah that would be pretty nice! Let me integrating this flow

crates/commonware-node/src/executor/actor.rs
hamdiallam: Looping back i'm curious why it was reversed to begin with? Doing FCU first delays finalization from progressing?

SuperFluffy: IIRC sending unknown payload to reth (unknown = not part of the locally known canonical chain) means that you are not guaranteed that it will be retained. Instead, it will be temporarily cached. If you now send the FCU reth should be able to pick up the payload and attach it to the canonical chain. This is extremely likely to happen of course, but not guaranteed.

On the other hand, if you send the FCU first and the payload after, you are guaranteed that the payload is attached to the canonical chain immediately.

Either way, this is probably worth digging into to understand in detail, but I would like that to happen outside this PR.

hamdiallam: Gotcha that makes sense. Since we are doing the FCU right after, the block should be available. 

I'm going to revert this change tho and we should explore separetely form this pr

crates/commonware-node/src/follow/engine.rs
```diff
@@ -0,0 +1,318 @@
...
+use commonware_utils::channel::mpsc;
+use eyre::{OptionExt as _, WrapErr as _};
+use rand_08::{CryptoRng, Rng};
+use tempo_dkg_onchain_artifacts::OnchainDkgOutcome;
+use tempo_node::TempoFullNode;
+use tracing::{info, info_span};
+
+use super::{driver, resolver::RpcResolver, stubs};
+use crate::{
+    alias::marshal as marshal_alias,
```
SuperFluffy: plz no random aliases

bin/tempo/src/main.rs
```diff
@@ -180,13 +186,82 @@ fn main() -> eyre::Result<()> {
                 and a handle to the execution node could be received",
         )?;
 
-        let ret = if node.config.dev.dev || args.follow.is_some() {
-            // When --follow is used (with or without a URL), skip consensus stack
+        let ret = if node.config.dev.dev {
+            // Dev mode: just wait for shutdown
```
SuperFluffy: remove

bin/tempo/src/main.rs
```diff
@@ -180,13 +186,82 @@ fn main() -> eyre::Result<()> {
...
+
+            info_span!("follow").in_scope(|| {
+                info!(
+                    path = %follow_storage.display(),
+                    upstream = %follow_url,
+                    "starting follow engine",
+                )
+            });
+
+            let runtime_config = commonware_runtime::tokio::Config::default()
```
SuperFluffy: Let's merge init of consensus and follow stacks. There's a lot of duplication here.

bin/tempo/src/main.rs
```diff
@@ -180,13 +186,82 @@ fn main() -> eyre::Result<()> {
...
 
-        let ret = if node.config.dev.dev || args.follow.is_some() {
-            // When --follow is used (with or without a URL), skip consensus stack
+        let ret = if node.config.dev.dev {
+            // Dev mode: just wait for shutdown
             futures::executor::block_on(async move {
                 shutdown_token_clone.cancelled().await;
                 Ok(())
             })
+        } else if let Some(ref follow_url) = args.follow_node {
```
SuperFluffy: amp 😢 
```suggestion
        } else if let Some(follow_url) = &args.follow_node {
```

crates/commonware-node/src/follow/engine.rs
```diff
@@ -0,0 +1,318 @@
...
+    executor, feed,
+    feed::FeedStateHandle,
+    storage::{self, REPLAY_BUFFER, WRITE_BUFFER},
+};
+
+/// Builder for the follow engine.
+#[derive(Clone)]
+pub(crate) struct Builder {
+    /// The execution node to drive.
+    pub execution_node: TempoFullNode,
```
SuperFluffy: `pub(crate)` all of this.

crates/commonware-node/src/follow/engine.rs
```diff
@@ -0,0 +1,318 @@
...
+    consensus::block::Block,
+    epoch::SchemeProvider,
+    executor, feed,
+    feed::FeedStateHandle,
+    storage::{self, REPLAY_BUFFER, WRITE_BUFFER},
+};
+
+/// Builder for the follow engine.
+#[derive(Clone)]
+pub(crate) struct Builder {
```
SuperFluffy:
```suggestion
pub(crate) struct Config {
```

crates/commonware-node/src/storage.rs
```diff
@@ -0,0 +1,135 @@
...
+            items_per_section: IMMUTABLE_ITEMS_PER_SECTION,
+            codec_config: (),
+            replay_buffer: REPLAY_BUFFER,
+            freezer_key_write_buffer: WRITE_BUFFER,
+            freezer_value_write_buffer: WRITE_BUFFER,
+            ordinal_write_buffer: WRITE_BUFFER,
+        },
+    )
+    .await
+    .wrap_err("failed to initialize finalized blocks archive")
```
SuperFluffy: Unnecessary - this context should be provided by the caller (you're just wrapping 1 call here).

crates/commonware-node/src/storage.rs
```diff
@@ -0,0 +1,135 @@
...
+            items_per_section: IMMUTABLE_ITEMS_PER_SECTION,
+            codec_config: Scheme::<PublicKey, MinSig>::certificate_codec_config_unbounded(),
+            replay_buffer: REPLAY_BUFFER,
+            freezer_key_write_buffer: WRITE_BUFFER,
+            freezer_value_write_buffer: WRITE_BUFFER,
+            ordinal_write_buffer: WRITE_BUFFER,
+        },
+    )
+    .await
+    .wrap_err("failed to initialize finalizations archive")
```
SuperFluffy: Unnecessary - this context should be provided by the caller (you're just wrapping 1 call here).

crates/commonware-node/src/storage.rs
```diff
@@ -0,0 +1,135 @@
...
+const FREEZER_VALUE_COMPRESSION: Option<u8> = Some(3);
+
+// Marshal configuration
+pub(crate) const REPLAY_BUFFER: NonZeroUsize =
+    NonZeroUsize::new(8 * 1024 * 1024).expect("value is not zero"); // 8MB
+pub(crate) const WRITE_BUFFER: NonZeroUsize =
+    NonZeroUsize::new(1024 * 1024).expect("value is not zero"); // 1MB
+pub(crate) const PRUNABLE_ITEMS_PER_SECTION: NonZeroU64 =
+    NonZeroU64::new(4_096).expect("value is not zero");
+pub(crate) const MAX_REPAIR: NonZeroUsize = NonZeroUsize::new(20).expect("value is not zero");
```
SuperFluffy: Check if pub(crate) visibility is necessary for all of these.

crates/commonware-node/src/storage.rs
```diff
@@ -0,0 +1,135 @@
...
+
+const BUFFER_POOL_PAGE_SIZE: NonZeroU16 = NonZeroU16::new(4_096).expect("value is not zero"); // 4KB
```
SuperFluffy: Can move to the commonware macros for this (`NZU16!` and friends).

Applies to all `NonZero*` types in this module.

crates/commonware-node/src/consensus/engine.rs
```diff
@@ -147,7 +129,7 @@ where
             },
         );
 
-        let page_cache_ref = CacheRef::new(BUFFER_POOL_PAGE_SIZE, BUFFER_POOL_CAPACITY);
+        let page_cache_ref = storage::create_page_cache();
```
SuperFluffy: Yeah, good that this is all done in one point now (same here and elsewhere).

bin/tempo/src/main.rs
```diff
@@ -180,13 +186,82 @@ fn main() -> eyre::Result<()> {
...
+                    path = %follow_storage.display(),
+                    upstream = %follow_url,
+                    "starting follow engine",
+                )
+            });
+
+            let runtime_config = commonware_runtime::tokio::Config::default()
+                .with_tcp_nodelay(Some(true))
+                .with_worker_threads(args.consensus.worker_threads)
+                .with_storage_directory(follow_storage)
```
SuperFluffy: If `follow_storage` is different from `--consensus.datadir` (or the place derived from `--datadir`), then we will have a hard time explaining how to move from a follow node to a consensus node.

IMO they should be the same?


EDIT: Ah, they are. I didn't see that immediately. Best to merge initialization so that there is only 1 runtime constructed for both consensus and follow stacks.

crates/commonware-node/src/follow/driver.rs
```diff
@@ -0,0 +1,305 @@
...
+            }
+        }
+    }
+
+    async fn run_subscription(&mut self) -> eyre::Result<()> {
+        let mut sub = self
+            .resolver
+            .subscribe_events()
+            .await
+            .map_err(|e| eyre::eyre!("{e}"))?;
```
SuperFluffy: `wrap_err`, not eyre + display formatting

crates/commonware-node/src/follow/driver.rs
```diff
@@ -0,0 +1,305 @@
...
+    ) -> eyre::Result<Finalization<Scheme<PublicKey, MinSig>, crate::consensus::Digest>> {
+        let cert_bytes = alloy_primitives::hex::decode(&certified.certificate)
+            .map_err(|e| eyre::eyre!("failed to decode certificate hex: {e}"))?;
+        Finalization::read(&mut &cert_bytes[..])
+            .map_err(|e| eyre::eyre!("failed to decode finalization: {e:?}"))
+    }
+
+    fn extract_scheme(&self, block: &Block, height: Height) -> eyre::Result<()> {
+        let extra_data = block.header().inner.extra_data.as_ref();
+        if extra_data.is_empty() {
```
SuperFluffy: Use `eyre::ensure!` here

crates/commonware-node/src/follow/resolver.rs
```diff
@@ -0,0 +1,346 @@
...
+        Ok(Some(Block::from_execution_block(sealed)))
+    }
+
+    pub(crate) async fn fetch_block_by_hash(
+        &self,
+        hash: alloy_primitives::B256,
+    ) -> eyre::Result<Option<Block>> {
+        let client = self.client().await?;
+        let rpc_block: Option<TempoRpcBlock> = client
+            .request("eth_getBlockByHash", rpc_params![hash, true])
```
SuperFluffy: Why is this necessary? Can't we use the full node to get this and deliver it to marshal?

SuperFluffy (changes requested):
Quick first pass. There is room for some code deduplication.

I was caught by us using explicit `eth_getBlock` RPCs - shouldn't we be going through our own local execution layer instead of calling this RPCs explicitly?

Big final ask: I would like this be e2e tested (in the `tempo-e2e` sense). This is probably the biggest lift. The easiest test would be just starting up a network of 1 and slaving the follower node to it.

You should be able to observe that the follower is able to progress.

A next step would be to force full re-dkg in the single network (change of network identity). The follower should be able to follow through that as well.

Finally (and that's for a follow-up), it would be amazing if we could use this mechanism in the e2e tests to replace or at least augment the existing (and whacky) sync-from-snapshot tests.

crates/commonware-node/src/follow/engine.rs
hamdiallam: some of this is used in the e2e tests

crates/commonware-node/src/follow/resolver.rs
hamdiallam: Going to address this later. Chicken/egg problems with the marshaller and finalized blocks

crates/commonware-node/src/follow/engine.rs
hamdiallam: amp man 😭

crates/commonware-node/src/follow/upstream.rs
```diff
@@ -0,0 +1,265 @@
...
+/// Upstream backed by direct access to state.
+///
+/// Avoids WebSocket I/O, allowing the follow engine to run in a
+/// deterministic runtime for testing.
+pub struct LocalUpstream {
+    feed: FeedStateHandle,
+    execution_node: TempoFullNode,
+}
+
+impl LocalUpstream {
```
hamdiallam: used in E2E tests so the deterministic runtime could be used

crates/commonware-node/src/follow/engine.rs
```diff
@@ -0,0 +1,355 @@
...
+    consensus::{Digest, block::Block},
+    epoch::SchemeProvider,
+    executor, feed,
+    feed::FeedStateHandle,
+    storage::{self, MAX_REPAIR, REPLAY_BUFFER, WRITE_BUFFER},
+};
+
+/// Builder for the follow engine.
+#[derive(Clone)]
+pub struct Builder<U: UpstreamNode> {
```
SuperFluffy: Let's call this
```suggestion
pub struct Config<U: UpstreamNode> {
```

Works better with `Config` in the rest of the crate (`consensus::Builder` is badly named/confusing).

crates/commonware-node/src/follow/resolver.rs
```diff
@@ -0,0 +1,229 @@
...
+use reth_provider::{BlockReader as _, BlockSource};
+use tempo_node::{TempoFullNode, rpc::consensus::Query};
+use tokio::sync::Mutex;
+use tracing::{debug, warn, warn_span};
+
+use super::upstream::UpstreamNode;
+use crate::consensus::{Digest, block::Block};
+
+type Request = handler::Request<Block>;
+type Message = handler::Message<Block>;
```
SuperFluffy: Plz avoid single use aliases.

crates/commonware-node/src/storage.rs
```diff
@@ -0,0 +1,127 @@
...
+    TContext,
+    Digest,
+    commonware_consensus::simplex::types::Finalization<Scheme<PublicKey, MinSig>, Digest>,
+>;
+
+/// Type alias for the finalized blocks archive.
+pub(crate) type FinalizedBlocksArchive<TContext> = immutable::Archive<TContext, Digest, Block>;
+
+/// Create a shared page cache for archives.
+pub(crate) fn create_page_cache() -> CacheRef {
```
SuperFluffy: Need to migrate to the Commonware way of doing this.

crates/commonware-node/src/storage.rs
```diff
@@ -0,0 +1,127 @@
...
+const FREEZER_VALUE_COMPRESSION: Option<u8> = Some(3);
+
+// Marshal configuration
+pub(crate) const REPLAY_BUFFER: std::num::NonZeroUsize = NZUsize!(8 * 1024 * 1024); // 8MB
+pub(crate) const WRITE_BUFFER: std::num::NonZeroUsize = NZUsize!(1024 * 1024); // 1MB
+pub(crate) const PRUNABLE_ITEMS_PER_SECTION: std::num::NonZeroU64 = NZU64!(4_096);
+pub(crate) const MAX_REPAIR: std::num::NonZeroUsize = NZUsize!(20);
+
+const BUFFER_POOL_PAGE_SIZE: std::num::NonZeroU16 = NZU16!(4_096); // 4KB
+const BUFFER_POOL_CAPACITY: std::num::NonZeroUsize = NZUsize!(8_192); // 32MB
```
SuperFluffy: Confusing comment (probably just copied from what was before): `PAGE_SIZE * POOL_CAPACITY = 32MB` with the above values

bin/tempo/src/main.rs
```diff
@@ -169,103 +177,123 @@ fn main() -> eyre::Result<()> {
...
-                let ctx = ctx.with_label("consensus");
+                install_prometheus_metrics(ctx.with_label("telemetry_metrics"), prometheus_config)
+                    .wrap_err("failed to start Prometheus metrics exporter")?;
+            }
 
-                let mut metrics_server = tempo_commonware_node::metrics::install(
-                    ctx.with_label("metrics"),
-                    args.consensus.metrics_address,
+            let follow_ctx = ctx.with_label("follow");
+            let consensus_ctx = ctx.with_label("consensus");
```
SuperFluffy: Nit: don't bind these contexts here, do it in the either-left-right blocks.

crates/commonware-node/src/consensus/engine.rs
```diff
@@ -34,32 +27,22 @@ use tempo_node::TempoFullNode;
...
+    peer_manager,
+    storage::{self, MAX_REPAIR, REPLAY_BUFFER, WRITE_BUFFER},
+    subblocks,
 };
 
 use super::block::Block;
 
-// A bunch of constants to configure commonwarexyz singletons and copied over form alto.
+// A bunch of constants to configure commonwarexyz singletons and copied over from alto.
+// Archive format constants are in crate::storage.
```
SuperFluffy: Let's remove this comment.

crates/commonware-node/src/consensus/engine.rs
```diff
@@ -170,103 +153,31 @@ where
         };
         let scheme_provider = SchemeProvider::new();
 
-        const FINALIZATIONS_BY_HEIGHT: &str = "finalizations-by-height";
+        // Initialize archives using shared format (compatible with follow engine)
```
SuperFluffy: Redundant comment.

bin/tempo/src/main.rs
```diff
@@ -268,110 +189,123 @@ fn main() -> eyre::Result<()> {
...
-            let runner = commonware_runtime::tokio::Runner::new(runtime_config);
+                install_prometheus_metrics(ctx.with_label("telemetry_metrics"), prometheus_config)
+                    .wrap_err("failed to start Prometheus metrics exporter")?;
+            }
 
-            runner.start(async move |ctx| {
-                // Ensure all consensus metrics are prefixed. Shadow `ctx` to
-                // not forget.
-                let ctx = ctx.with_label("consensus");
+            let follow_ctx = ctx.with_label("follow");
```
SuperFluffy: I almost feel like we should leave this as `consensus`, not `follow`. That way we can reuse most of the metric dashboards.

Not totally sure baout this.

bin/tempo/src/main.rs
```diff
@@ -59,17 +62,19 @@ use tempo_node::{
...
+    /// Run in follow mode, syncing consensus and execution state from an upstream node.
     /// If provided without a value, defaults to the RPC URL for the selected chain.
-    #[arg(long, value_name = "URL", default_missing_value = "auto", num_args(0..=1), env = "TEMPO_FOLLOW")]
+    #[arg(long, value_name = "WS_URL", default_missing_value = "auto", num_args(0..=1), env = "TEMPO_FOLLOW")]
     pub follow: Option<String>,
 
+    /// Run in follow mode, without validating any execution state. Do not use in any
+    /// full node production settings.
+    #[arg(long, value_name = "WS_URL", default_missing_value = "auto", conflicts_with = "follow", num_args(0..=1), env = "TEMPO_FOLLOW_UNCERTIFIED")]
+    pub follow_uncertified: Option<String>,
```
SuperFluffy: Thinking to just keep `--follow` but modify it with `--no-validate-certificates` so that you would do:

```sh
tempo node --follow --no-validate-certificates
```

bin/tempo/src/main.rs
```diff
@@ -59,17 +62,19 @@ use tempo_node::{
...
-    Cli<TempoChainSpecParser, TempoArgs, DefaultRpcModuleValidator, tempo_cmd::TempoSubcommand>;
-
 // TODO: migrate this to tempo_node eventually.
 #[derive(Debug, Clone, clap::Args)]
 struct TempoArgs {
-    /// Follow this specific RPC node for block hashes.
+    /// Run in follow mode, syncing consensus and execution state from an upstream node.
     /// If provided without a value, defaults to the RPC URL for the selected chain.
-    #[arg(long, value_name = "URL", default_missing_value = "auto", num_args(0..=1), env = "TEMPO_FOLLOW")]
+    #[arg(long, value_name = "WS_URL", default_missing_value = "auto", num_args(0..=1), env = "TEMPO_FOLLOW")]
```
SuperFluffy: Let's write this out:
```suggestion
    #[arg(long, value_name = "WEBSOCKET_URL", default_missing_value = "auto", num_args(0..=1), env = "TEMPO_FOLLOW")]
```

bin/tempo/src/main.rs
```diff
@@ -268,110 +189,123 @@ fn main() -> eyre::Result<()> {
 
     let shutdown_token_clone = shutdown_token.clone();
     let cl_feed_state_clone = cl_feed_state.clone();
+
     let consensus_handle = thread::spawn(move || {
         // Exit early if we are not executing `tempo node` command.
         if !is_node {
             return Ok(());
-        }
+        };
```
SuperFluffy: Remove semicolon :)

bin/tempo/src/main.rs
```diff
@@ -268,110 +189,123 @@ fn main() -> eyre::Result<()> {
...
+                .data_dir()
+                .join("consensus")
+        });
 
-            let runtime_config = commonware_runtime::tokio::Config::default()
-                .with_tcp_nodelay(Some(true))
-                .with_worker_threads(args.consensus.worker_threads)
-                .with_storage_directory(consensus_storage)
-                .with_catch_panics(true);
+        info_span!("consensus").in_scope(|| {
```
SuperFluffy: Change back to `prepare_consensus` plz. Best to keep things like span names stable unless we are refactoring.

bin/tempo/src/main.rs
```diff
@@ -268,110 +189,123 @@ fn main() -> eyre::Result<()> {
...
             });
+        }
 
-            info_span!("prepare_consensus").in_scope(|| {
-                info!(
-                    path = %consensus_storage.display(),
-                    "determined directory for consensus data",
-                )
-            });
+        let storage = args.consensus.storage_dir.clone().unwrap_or_else(|| {
```
SuperFluffy: The follower engine is also going to store consensus stuff in that dir, so let's keep the name (i.e. change back):

```suggestion
        let consensus_storage = args.consensus.storage_dir.clone().unwrap_or_else(|| {
```

bin/tempo/src/main.rs
```diff
@@ -268,110 +189,123 @@ fn main() -> eyre::Result<()> {
...
+                        )
+                        .wrap_err("consensus stack failed");
+                    }
+
+                    ret = &mut metrics_server, if !metrics_server.is_terminated() => {
+                        let reason = match ret.wrap_err("task_panicked") {
+                            Ok(Ok(())) => "unexpected regular exit".to_string(),
+                            Ok(Err(err)) | Err(err) => format!("{err}"),
+                        };
+                        tracing::warn!(reason, "the metrics server exited");
```
SuperFluffy: Let's toss this into a `warn_span!` if we touch these lines (so that the warnings have a root span).

crates/commonware-node/src/consensus/engine.rs
```diff
@@ -143,106 +127,36 @@ where
...
+        );
 
         let scheme_provider = SchemeProvider::new();
 
-        const FINALIZATIONS_BY_HEIGHT: &str = "finalizations-by-height";
+        // Initialize archives using shared format (compatible with follow engine)
         let start = Instant::now();
-        let finalizations_by_height = immutable::Archive::init(
+        let finalizations_by_height = storage::init_finalizations_archive(
             context.with_label("finalizations_by_height"),
```
SuperFluffy: Provide the context by reference and let the helper function add the label itself. We usually pass in context by ownership if we want the callee to use it as is. But in this case it's better to let the helper do it.

crates/commonware-node/src/consensus/engine.rs
```diff
@@ -143,106 +127,36 @@ where
...
         .await
         .wrap_err("failed to initialize finalizations by height archive")?;
+
         info!(elapsed = ?start.elapsed(), "restored finalizations by height archive");
 
-        const FINALIZED_BLOCKS: &str = "finalized_blocks";
         let start = Instant::now();
-        let finalized_blocks = immutable::Archive::init(
+        let finalized_blocks = storage::init_finalized_blocks_archive(
             context.with_label("finalized_blocks"),
```
SuperFluffy: Same as above re context + label.

crates/commonware-node/src/consensus/engine.rs
```diff
@@ -143,106 +127,36 @@ where
...
-            },
+            &self.partition_prefix,
+            page_cache_ref.clone(),
         )
         .await
-        .wrap_err("failed to initialize finalizations by height archive")?;
-        info!(elapsed = ?start.elapsed(), "restored finalizations by height archive");
+        .wrap_err("failed to initialize finalized blocks archive")?;
+
+        info!(elapsed = ?start.elapsed(), "restored finalized blocks archive");
```
SuperFluffy: Let's use `tempo_tracing_utils::display_duration` if we touch this line.

crates/commonware-node/src/consensus/engine.rs
```diff
@@ -143,106 +127,36 @@ where
...
-                freezer_value_write_buffer: WRITE_BUFFER,
-                ordinal_write_buffer: WRITE_BUFFER,
-            },
+            &self.partition_prefix,
+            page_cache_ref.clone(),
         )
         .await
         .wrap_err("failed to initialize finalizations by height archive")?;
+
         info!(elapsed = ?start.elapsed(), "restored finalizations by height archive");
```
SuperFluffy: Because we touch a similar line a few lines down.

Let's use `tempo_tracing_utils::display_duration`.

crates/commonware-node/src/storage.rs
```diff
@@ -0,0 +1,121 @@
...
+use commonware_runtime::{BufferPooler, Clock, Metrics, Spawner, Storage, buffer::paged::CacheRef};
+use commonware_storage::archive::immutable;
+use commonware_utils::{NZU16, NZU64, NZUsize};
+
+use crate::{
+    config::BLOCKS_FREEZER_TABLE_INITIAL_SIZE_BYTES,
+    consensus::{Digest, block::Block},
+};
+
+/// Archive partition names - shared between consensus and follow engines.
```
SuperFluffy: IMO redundant comment.

crates/commonware-node/src/storage.rs
```diff
@@ -0,0 +1,121 @@
+//! Shared storage configuration for consensus and follow engines.
```
SuperFluffy: IMO overexplained - this module does not need to care about consensus or follower.

crates/commonware-node/src/follow/mod.rs
```diff
@@ -0,0 +1,60 @@
+//! CL-driven follow mode for Tempo nodes.
+//!
+//! This module provides a follow implementation that syncs from an upstream
+//! node (validator or another follower).
+//!
+//! # Architecture
+//!
+//! ```text
```
SuperFluffy: Imo remove this ASCII graph. Too high chance that this gets outdated quickly, and painful to maintain.

If we do this, then we should do mermaid (and outside of rust doc comments).

crates/commonware-node/src/follow/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+//! ```
+
+mod driver;
+pub mod engine;
+pub(crate) mod resolver;
+mod stubs;
+pub mod upstream;
+
+pub use engine::Builder;
+pub use upstream::{LocalUpstream, UpstreamNode, WsUpstream};
```
SuperFluffy: Does all of this need to be pub? Can we lock some of it down?

crates/commonware-node/src/lib.rs
```diff
@@ -11,16 +11,20 @@ pub(crate) mod dkg;
 pub(crate) mod epoch;
 pub(crate) mod executor;
 pub mod feed;
+pub mod follow;
```
SuperFluffy, line 14: Can we lock some of the follow stuff down (make it private)?

bin/tempo/src/main.rs
```diff
@@ -70,61 +66,28 @@ use tempo_node::{
...
-    /// Set to "none" to disable.
+    /// Disable consensus certification in follow mode. The follower syncs execution
+    /// state from the upstream node without validating consensus state.
+    /// DO NOT USE IN PRODUCTION.
     #[arg(
-        long = "tempo.bootnodes-endpoint",
-        value_name = "URL",
-        default_value = "https://peers.tempo.xyz",
-        env = "TEMPO_BOOTNODES_ENDPOINT"
+        long = "follow.uncertified",
```
SuperFluffy: Personal preference because it's more unix-y:
```suggestion
    long = "follow.nocertify"
```

bin/tempo/src/main.rs
```diff
@@ -70,61 +66,28 @@ use tempo_node::{
...
+        requires = "follow",
+        default_missing_value = "true",
```
SuperFluffy: I think we should remove this. In the light of recent events, we can get away with this because it's a security feature.
```suggestion
```

bin/tempo/src/main.rs
```diff
@@ -70,61 +66,28 @@ use tempo_node::{
...
     )]
-    pub bootnodes_endpoint: String,
+    pub follow_uncertified: Option<bool>,
```
SuperFluffy: Unnecessary for booleans:
```suggestion
    pub follow_nocertify: bool,
```

bin/tempo/src/main.rs
```diff
@@ -391,82 +311,96 @@ fn main() -> eyre::Result<()> {
...
                 and a handle to the execution node could be received",
         )?;
 
-        let ret = if node.config.dev.dev || args.follow.is_some() {
-            // When --follow is used (with or without a URL), skip consensus stack
-            futures::executor::block_on(async move {
+        if node.config.dev.dev || args.follow_uncertified.unwrap_or(false) {
+            return futures::executor::block_on(async move {
                 shutdown_token_clone.cancelled().await;
+                let _ = consensus_dead_tx.send(());
```
SuperFluffy: Just curious what made this necessary?

bin/tempo/src/main.rs
```diff
@@ -391,82 +311,96 @@ fn main() -> eyre::Result<()> {
...
-                        prometheus_config,
-                    )
+                install_prometheus_metrics(ctx.with_label("telemetry_metrics"), prometheus_config)
                     .wrap_err("failed to start Prometheus metrics exporter")?;
-                }
+            }
+
+            let follow_ctx = ctx.with_label("follow");
+            let consensus_ctx = ctx.with_label("consensus");
+            let stack = if let Some(follow) = args.follow {
```
SuperFluffy: Super non-descript what a "stack" is. Let's keep `let consensus_stack`?

bin/tempo/src/main.rs
```diff
@@ -569,100 +520,50 @@ fn main() -> eyre::Result<()> {
...
-                        builder
-                            .config()
-                            .chain
-                            .default_follow_url()
-                            .map(|s| s.to_string())
-                    } else {
-                        Some(follow.clone())
-                    };
+                // Uncertified follower mode: set debug RPC when certification is off
+                if args.follow_uncertified.unwrap_or(false) {
```
SuperFluffy: I feel like this warrants a big fat warning emitted at startup.

bin/tempo/src/main.rs
```diff
@@ -569,100 +520,50 @@ fn main() -> eyre::Result<()> {
...
-            })
-            .extend_rpc_modules(move |ctx| {
-                if faucet_args.enabled {
-                    let ext = TempoFaucetExt::new(
-                        faucet_args.addresses(),
-                        faucet_args.amount(),
-                        faucet_args.provider(),
-                    );
+                let has_consensus_engine =
+                    !builder.config().dev.dev && !args.follow_uncertified.unwrap_or(false);
```
SuperFluffy: We used something like this when we entered this branch in the first place, `        if node.config.dev.dev || args.follow_uncertified.unwrap_or(false) {`

Move to some helper fn?

bin/tempo/src/main.rs
```diff
@@ -569,100 +520,50 @@ fn main() -> eyre::Result<()> {
...
 
-                    ctx.modules.merge_configured(ext.into_rpc())?;
-                }
+                // Additional RPC Modules
```
SuperFluffy: redundant comment.

bin/tempo/src/main.rs
```diff
@@ -569,100 +520,50 @@ fn main() -> eyre::Result<()> {
...
+                            faucet_args.addresses(),
+                            faucet_args.amount(),
+                            faucet_args.provider(),
+                        );
 
-                if validator_key.is_some() {
-                    ctx.modules
-                        .merge_configured(TempoConsensusRpc::new(cl_feed_state).into_rpc())?;
-                }
+                        ctx.modules.merge_configured(faucet_ext.into_rpc())?;
```
SuperFluffy: add context

bin/tempo/src/main.rs
```diff
@@ -569,100 +520,50 @@ fn main() -> eyre::Result<()> {
...
+                    }
 
-                Ok(())
+                    if has_consensus_engine {
+                        let consensus_rpc = TempoConsensusRpc::new(cl_feed_state);
+                        ctx.modules.merge_configured(consensus_rpc.into_rpc())?;
```
SuperFluffy: add context

SuperFluffy (commented):
quick fly by comments, going to continue

crates/commonware-node/src/follow/engine.rs
```diff
@@ -0,0 +1,393 @@
...
+                execution_node: self.execution_node,
+                last_finalized_height,
+                marshal: marshal_mailbox.clone(),
+                fcu_heartbeat_interval: self.fcu_heartbeat_interval,
+            },
+        )
+        .wrap_err("failed to initialize executor")?;
+
+        // No broadcast is needed in follow mode.
+        let broadcast = stubs::null_broadcast(context.with_label("broadcast"), self.mailbox_size);
```
SuperFluffy, line 176: I wonder if we can change this my setting up a seconary p2p network, where the follower does broadcast.

crates/commonware-node/src/follow/engine.rs
```diff
@@ -0,0 +1,393 @@
...
+
+        let bootstrap = self.last_finalized_height == Height::zero();
+        let boundary_block = if bootstrap {
+            // If we're bootstrapping, find the latest boundary block and
+            // use it as the trusted floor for the identity scheme.
+            let latest_finalization = self
+                .upstream
+                .get_finalization(Query::Latest)
+                .await?
+                .ok_or_eyre("failed to get latest finalization")?;
```
SuperFluffy: Two comments:

1. Not sure if that's what None means ... but what happens if the other node does not yet have finalizations? Shouldn't we wait for it to provide them? Say you connect you start a fresh node and immediately connect the follower, before the node had time to join the network and collect a finalization.

2. Bit verbose, but maps all of this into one error:
```suggestion
                .get_finalization(Query::Latest)
                .await
                .map_err(eyre::Report::new)
                .and_then(|res| res.ok_or_eyre("validator returned it had no finalizations"))
                .wrap_err("failed to get latest finalization")?;
```

crates/commonware-node/src/follow/engine.rs
```diff
@@ -0,0 +1,393 @@
...
+        + Metrics
+        + Pacer
+        + Spawner
+        + Storage
+        + BufferPooler
+        + Clone
+        + Send
+        + 'static,
+{
+    pub async fn start(mut self) -> eyre::Result<()> {
```
SuperFluffy: Let's follow the rest of the codebase and spawn this on its own task and context.

crates/commonware-node/src/follow/engine.rs
```diff
@@ -0,0 +1,393 @@
...
+        let _executor_handle = self.executor_actor.start();
+        let _feed_handle = self.feed_actor.start();
+        let _marshal_handle = self.marshal_actor.start(
+            self.executor_mailbox.clone(),
+            self.broadcast,
+            (self.resolver_rx, self.resolver),
+        );
```
SuperFluffy: Don't just ignore these but `join_all` them to detect if any of these actors go down.

crates/commonware-node/src/follow/engine.rs
```diff
@@ -0,0 +1,393 @@
...
+
+            let boundary_height = if latest_finalization_height == epoch_info.last() {
+                latest_finalization_height
+            } else {
+                match epoch_info.epoch().previous() {
+                    None => Height::zero(),
+                    Some(prev_epoch) => self
+                        .epoch_strategy
+                        .last(prev_epoch)
+                        .ok_or_eyre("failed to determine previous epoch boundary")?,
```
SuperFluffy: It's ok to `expect` here - the epoch strat is for all heights and epochs (the error msg here is not really helpful either).

crates/commonware-node/src/follow/engine.rs
```diff
@@ -0,0 +1,393 @@
...
+            };
+
+            let block = self
+                .upstream
+                .get_block_by_number(boundary_height.get())
+                .await?
+                .ok_or_eyre(format!(
+                    "block at height {} not found on upstream",
+                    boundary_height.get()
+                ))?;
```
SuperFluffy: Should we me more lenient here? What if the validator is getting aggressively pruned and doesn't have the block anymore? Technically we only need a header to check that `certified.digest == block_hash`.

From then on, we just use whatever the latest block is that the validator gives us.

Maybe this requirement is ok, just wondering if we can be more lax.

crates/commonware-node/src/follow/engine.rs
```diff
@@ -0,0 +1,393 @@
...
+
+            // The genesis block just needs the outcome registered. No need to process it with the
+            // marshal as there is no finalization certificate. `last_finalized_height` will remain zero.
+            if boundary_height > Height::zero() {
+                let certified = self
```
SuperFluffy: Let's not use `certified` nomenclature here as it clashes with `Activity::Certified`.

crates/commonware-node/src/follow/engine.rs
```diff
@@ -0,0 +1,393 @@
...
+
+                let sealed = SealedBlock::seal_slow(block.clone());
+                let consensus_block = Block::from_execution_block(sealed);
+                eyre::ensure!(certified.digest == consensus_block.block_hash());
+
+                let cert_bytes = alloy_primitives::hex::decode(&certified.certificate)?;
+                let finalization: Finalization<Scheme<PublicKey, MinSig>, Digest> =
+                    Finalization::read(&mut &cert_bytes[..])?;
+
+                // Process the boundary block & finalization
```
SuperFluffy: kinda vacuous - it's clear that this is being processed. However, what's the point of valling `marshal.verified` on it?

crates/commonware-node/src/follow/engine.rs
```diff
@@ -0,0 +1,393 @@
...
+                        .epoch_strategy
+                        .last(prev_epoch)
+                        .ok_or_eyre("failed to determine previous epoch boundary")?,
+                }
+            };
+
+            self.marshal_mailbox
+                .get_block(boundary_height)
+                .await
+                .ok_or_eyre(format!(
```
SuperFluffy:
```suggestion
                .ok_or_else(|| eyre!(

File: crates/commonware-node/src/follow/engine.rs [hunk 69]
```diff
@@ -0,0 +1,393 @@
...
+                    "block at height {} not found in local archive",
+                    boundary_height.get()
+                ))?
+                .into_inner()
+                .into_block()
+        };
+
+        let extra_data = boundary_block.extra_data();
+        let outcome = OnchainDkgOutcome::read(&mut &extra_data[..])
+            .wrap_err("block did not contain a DKG outcome")?;
```
SuperFluffy: Could be malformed...
```suggestion
            .wrap_err("could not read DKG outcome from block")?;
```

crates/commonware-node/src/lib.rs
```diff
@@ -147,6 +155,47 @@ pub async fn run_consensus_stack(
...
+    let epoch_length = execution_node
+        .chain_spec()
+        .info
+        .epoch_length()
+        .ok_or_eyre("chainspec did not contain epochLength")?;
+
+    let upstream = std::sync::Arc::new(follow::WsUpstream::new(
+        context.with_label("upstream"),
+        upstream_url,
+    ));
```
SuperFluffy: Can we have the node connect to this URL also for tests?

In e2e tests we are running against actual an actual EL node (not a simulated environment like commonware). That means we should be able to have it connect over WS.

crates/commonware-node/src/follow/resolver.rs
```diff
@@ -0,0 +1,214 @@
...
+use tokio::sync::Mutex;
+use tracing::{debug, warn, warn_span};
+
+use super::upstream::UpstreamNode;
+use crate::consensus::{Digest, block::Block};
+
+type Request = handler::Request<Digest>;
+type Message = handler::Message<Digest>;
+
+pub(crate) struct FollowResolver<TContext: Spawner + Clone + Send + 'static, U: UpstreamNode> {
```
SuperFluffy:
```suggestion
pub(crate) struct Resolver<TContext: Spawner, TUpstreamNode: UpstreamNode> {
```

We already have `Spawner: Clone + Send + Sync + 'static`. (also IMO just name this `FollowResolver` so we have `follow::Resolver`).

Note on the type-bounds: try to avoid them if you can and rely on having them in the impl blocks only. I.e. if you can get away with this, prefer:

```rust
pub(crate) struct FollowResolver<TContext, U> {
```

crates/commonware-node/src/follow/resolver.rs
```diff
@@ -0,0 +1,214 @@
...
+    context: TContext,
+    upstream: Arc<U>,
+    sender: mpsc::Sender<Message>,
+    pending: Arc<Mutex<HashSet<Request>>>,
+    execution_node: TempoFullNode,
+}
+
+impl<TContext: Spawner + Clone + Send + 'static, U: UpstreamNode> Clone
```
SuperFluffy: Derive this. All values are cloneable alreaday.

crates/commonware-node/src/follow/resolver.rs
```diff
@@ -0,0 +1,214 @@
...
+        else {
+            return Ok(None);
+        };
+
+        let cert_bytes = alloy_primitives::hex::decode(&certified.certificate)
+            .wrap_err("failed to decode certificate hex")?;
+
+        let finalization: Finalization<Scheme<PublicKey, MinSig>, Digest> =
+            Finalization::read(&mut &cert_bytes[..])
+                .map_err(|e| eyre::eyre!("failed to decode finalization: {e:?}"))?;
```
SuperFluffy: Don't use `map_err` with `eyre!`. Use `wrap_err` or `wrap_err_with`

crates/commonware-node/src/follow/resolver.rs
```diff
@@ -0,0 +1,214 @@
...
+            pending: Arc::new(Mutex::new(HashSet::new())),
+            execution_node,
+        }
+    }
+
+    // ── Gap-repair internals ────────────────────────────────────────────
+
+    fn spawn_fetch(&self, key: Request) {
+        let resolver = self.clone();
+        let key_clone = key;
```
SuperFluffy: Seems unnecessary to rename this here.

crates/commonware-node/src/follow/resolver.rs
```diff
@@ -0,0 +1,214 @@
...
+                        && let Ok(true) = response_rx.await
+                    {
+                        resolver.pending.lock().await.remove(&key_clone);
+                    }
+                }
+                Ok(None) => {
+                    debug!(?key_clone, "data not yet available from upstream");
+                }
+                Err(e) => {
+                    warn!(?key_clone, error = %e, "failed to fetch from upstream");
```
SuperFluffy:
```suggestion
                Err(error) => {
                    warn!(?key, %error, "failed to fetch from upstream");
                }
```

crates/commonware-node/src/follow/engine.rs
SuperFluffy: Comment on the general flow in this actor - if the validator is not yet up or if there is some hickup, the follower will not start and just die.

I think we should make it more robust at startup by waiting until a connection can be established. Would require changing the pattern here a bit.

bin/tempo/src/main.rs
hamdiallam: Updated the flag so that it's not default true in clap. Instead the default is in `is_following_uncertified` and we can warn if --follow.uncertified is explicitly used

crates/commonware-node/src/follow/resolver.rs
hamdiallam: yep 100%, some old amp code. revisiting

crates/commonware-node/src/follow/engine.rs
hamdiallam: hmm yea good point, this is redundant

hamdiallam: oh I think the point was if we've registered the scheme, we have the block for it as well. Otherwise the scheme would be registered but the block would be skipped. Seems like a good assurance to hold that we have this block

crates/commonware-node/src/follow/engine.rs
hamdiallam: agreed, we should probably rename our `CertifiedBlock` trait

crates/commonware-node/src/follow/engine.rs
hamdiallam: For the bootstrap case, it's requesting information at tip. If a validator is pruning it's last boundary block that seems problematic?

Alternatively we would wait for the first valid boundary block to come in?
 - (1) Try request the last boundary
 - (2) If unavailable, wait till the next boundary?

crates/commonware-node/src/lib.rs
hamdiallam: Ah the reason for this is because we dont have access to the tokio runtime. Trying to connect over WS in the deterministic runtime will panic

hamdiallam: would've been great if commonware had ws library we could use instead

bin/tempo/src/main.rs
hamdiallam: The reason for the option right now is to avoid the log during the migration if unset. If we made it a boolean now, the warning log would be emitted since this should be true by default.

With an option we only emit the warning log if --follow.nocertify=true was explicitly set. Once the default is false post-migration will clean this up.

BTW like `nocertify` over `uncertified`. updating this

bin/tempo/src/main.rs
hamdiallam: ah this is not necessary. maybe some earlier process structure required it. Removing this

crates/commonware-node/src/follow/mod.rs
hamdiallam: yep we can. only the engine and LocalUpstream needs to be public for the tests

crates/commonware-node/src/lib.rs
hamdiallam: Yep, most of follow made private except what is needed in e2e

crates/commonware-node/src/storage.rs
hamdiallam: Right yea we are allocating ~8k pages. I'll update this comment

SuperFluffy (approved):

bin/tempo/src/main.rs
```diff
@@ -70,61 +66,24 @@ use tempo_node::{
...
-        long = "tempo.bootnodes-endpoint",
-        value_name = "URL",
-        default_value = "https://peers.tempo.xyz",
-        env = "TEMPO_BOOTNODES_ENDPOINT"
-    )]
-    pub bootnodes_endpoint: String,
+    /// Disable consensus certification in follow mode. The follower syncs execution
+    /// state from the upstream node without validating consensus state.
+    /// DO NOT USE IN PRODUCTION.
+    #[arg(long = "follow.uncertified", requires = "follow")]
```
hamdiallam: ah the `long` here is wrong

bin/tempo/src/main.rs
```diff
@@ -402,71 +426,81 @@ fn main() -> eyre::Result<()> {
                 and a handle to the execution node could be received",
         )?;
 
-        let ret = if node.config.dev.dev || args.follow.is_some() {
-            // When --follow is used (with or without a URL), skip consensus stack
-            futures::executor::block_on(async move {
+        if node.config.dev.dev || args.is_following_uncertified() {
```
hamdiallam: note: switch to `!has_consensus_engine(node.config.dev.dev)`

crates/commonware-node/src/follow/driver.rs
```diff
@@ -324,12 +324,18 @@ where
...
         let consensus_block = Block::from_execution_block(SealedBlock::seal_slow(certified.block));
 
         // Store the Finalized Block
         let round = Round::new(Epoch::new(certified.epoch), View::new(certified.view));
         let activity = Activity::Finalization(finalization);
-        self.config.marshal.verified(round, consensus_block).await;
+        let verified = self.config.marshal.verified(round, consensus_block).await;
+        if !verified {
+            warn_span!("follow_driver")
+                .in_scope(|| warn!(?round, %height, "failed to verify incoming block"))
```
SuperFluffy: This is not about verification but about persistence I think. Marshal informs us that it couldn't persists the block locally. (plz double check)

hamdiallam: yea it's about persistence, let me update the warning

PR #2554, opened 2026-02-09
test(node): matrix-driven AA integration tests with full key type × fee payer × access key coverage

0xrusowsky:
Replaces ad-hoc standalone AA integration tests with matrix-driven runners that exhaustively cover the key type × fee payer × access key × batch calls parameter space, and adds an end-to-end fill → sign → send matrix across nonce modes.

## Matrix coverage

| Matrix | Dimensions |
|---|---|
| `test_eth_send_raw_transaction_matrix` | key type (secp256k1/P256/WebAuthn) × fee payer × access key |
| `test_eth_send_transaction_matrix` | key type × fee payer; P256/WebAuthn also × access key × batch calls |
| `test_eth_fill_transaction_matrix` | nonce mode (protocol/2D/expiring) × fee token × fee payer |
| `test_e2e_fill_sign_send_matrix` | nonce mode × key type, including expiring nonce boundary/rejection |
| `test_aa_estimate_gas_matrix` | gas estimation across key types and configurations |

Each matrix case is a declarative struct built via `raw_case!`, `send_case!`, or `fill_case!` macros (in `test_macros.rs`) with auto-derived names for clear test output.

## Shared runner hardening

Assertions that previously lived in individual standalone tests are now applied to every matrix case:

- **Encoding roundtrip + signer recovery** on every raw send case.
- **`verify_tx_in_block_via_rpc`** validates type, chainId, nonce, calls, and feeToken fields.
- **Receipt status** checked via `assert_receipt_status` (not just receipt existence).
- **Unfunded sender balance** verified before and after fee-payer transactions.
- **Batch recipient balances** verified after batch_calls cases.
- **Fee payer signature hash recovery** verified in fill matrix fee_payer case.

### DIFF
--- a/crates/node/tests/it/tempo_transaction.rs
+++ b/crates/node/tests/it/tempo_transaction.rs
@@ -7554,20 +7554,29 @@ async fn run_fill_sign_send_test_p256(test_case: &FillTestCase) -> eyre::Result<
     let envelope: TempoTxEnvelope = tx.into_signed(signature).into();
     let tx_hash = *envelope.tx_hash();
 
-    let _ = provider
+    let send_result = provider
         .send_raw_transaction(&envelope.encoded_2718())
-        .await?;
-    setup.node.advance_block().await?;
+        .await;
 
-    let raw_receipt: Option<serde_json::Value> = provider
-        .raw_request("eth_getTransactionReceipt".into(), [tx_hash])
-        .await?;
-    assert!(raw_receipt.is_some(), "Transaction should be mined");
-    let status = raw_receipt.unwrap()["status"]
-        .as_str()
-        .map(|s| s == "0x1")
-        .unwrap_or(false);
-    assert!(status, "Transaction should succeed");
+    match test_case.expected {
+        ExpectedOutcome::Success => {
+            let _ = send_result?;
+            setup.node.advance_block().await?;
+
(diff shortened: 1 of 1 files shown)


crates/node/tests/it/tempo_transaction.rs
```diff
@@ -6912,70 +7532,73 @@ async fn test_eth_fill_transaction_expiring_nonce_gas_is_sufficient() -> eyre::R
...
-    );
-    let envelope: TempoTxEnvelope = signed.into();
+    let envelope: TempoTxEnvelope = tx.into_signed(signature).into();
+    let tx_hash = *envelope.tx_hash();
 
-    setup
-        .node
-        .rpc
-        .inject_tx(envelope.encoded_2718().into())
+    let _ = provider
```
fgimenez: nit, would be nice to allow to include expected rejections here too, similar to what we do in `run_fill_sign_send_test_secp256k1`

fgimenez (approved):

crates/node/tests/it/tempo_transaction.rs
0xrusowsky: addressed in [`2da6f61` (this PR)](https://github.com/tempoxyz/tempo/pull/2554/commits/2da6f6198550892a1284d63faf904cf646238be9)

PR #2561, opened 2026-02-09
fix(consensus): show all validators, not only ones in dkg outcome

Zygimantass:
Changes `tempo consensus validators-info` to contain all validators, even if they are not currently in the committee but are supposed to become members in a future epoch.

### DIFF
--- a/bin/tempo/src/tempo_cmd.rs
+++ b/bin/tempo/src/tempo_cmd.rs
@@ -1,4 +1,4 @@
-use std::{collections::HashMap, fs::OpenOptions, path::PathBuf, sync::Arc};
+use std::{fs::OpenOptions, path::PathBuf, sync::Arc};
 
 use alloy_primitives::Address;
 use alloy_provider::Provider;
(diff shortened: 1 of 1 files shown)


hamdiallam (approved):

bin/tempo/src/tempo_cmd.rs
```diff
@@ -257,38 +257,32 @@ impl ValidatorsInfo {
...
+            .iter()
+            .map(|player| Ok((player.as_ref().try_into()?, ())))
+            .collect::<Result<_, eyre::Error>>()
+            .wrap_err("invalid next player pubkey")?;
+
+        let mut validator_entries = Vec::with_capacity(decoded_validators.len());
+        for validator in decoded_validators.into_iter() {
+            let pubkey_bytes = validator.publicKey.0;
 
             validator_entries.push(ValidatorEntry {
```
SuperFluffy: By default, this should not contain historic validators (validators that are inactive and are not dealers in the next epoch due to failed DKG).

bin/tempo/src/tempo_cmd.rs
```diff
@@ -257,38 +257,32 @@ impl ValidatorsInfo {
...
-                    (
-                        v.validatorAddress,
-                        v.active,
-                        v.inboundAddress.clone(),
-                        v.outboundAddress.clone(),
-                    )
-                } else {
-                    (Address::ZERO, false, String::new(), String::new())
-                };
+        let players: HashMap<[u8; 32], ()> = dkg_outcome
```
SuperFluffy: `HashMap` not `HashSet` - but these already are (ordered) sets that you can cheaply look into.

bin/tempo/src/tempo_cmd.rs
```diff
@@ -257,38 +257,32 @@ impl ValidatorsInfo {
...
-                inbound_address: inbound,
-                outbound_address: outbound,
-                active,
-                is_dealer: dkg_outcome.players().position(&player).is_some(),
-                is_player: dkg_outcome.next_players().position(&player).is_some(),
+                inbound_address: validator.inboundAddress,
+                outbound_address: validator.outboundAddress,
+                active: validator.active,
+                is_dealer: players.contains_key(&pubkey_bytes),
+                is_player: next_players.contains_key(&pubkey_bytes),
```
SuperFluffy: OK, this does contain enough information to figure out who an "active" validator is (as in, participating in consensuns), but you need to know implementation details. If we change this, let's at least change the meaning of `active` away from the onchain flag.

PR #2565, opened 2026-02-09
fix(txpool): check transaction size limit early in validate_one

gakonst:
## Summary
Moves the encoded-length size check to the top of `validate_one`, before any expensive validation (keychain, fee token, intrinsic gas, state lookups).

## Motivation
Addresses [SIGP-36](https://linear.app/tempoxyz/issue/SIGP-36) — RLP deserialization occurs before the transaction size limit is enforced, so oversized transactions trigger full validation before being rejected.

The bytes are already in memory by the time we reach `validate_one` (they've been received and deserialized over the wire), so there's no way to avoid the initial deserialization allocation. However, we can reject oversized txs immediately rather than doing keychain lookups, fee token validation, intrinsic gas calculation, and state provider queries first.

As the audit notes: low impact (temporary memory pressure, no funds at risk) and low likelihood (rate-limited by P2P session buffer).

## Changes
- Added early `encoded_length()` check against `max_tx_input_bytes` at the top of `validate_one`, right after the system-tx rejection
- Updated `test_aa_call_input_too_large_rejected` to accept either `OversizedData` or `CallInputTooLarge` since the early check now fires first for txs that exceed both limits

## Testing
```
cargo test -p tempo-transaction-pool
# 202 passed; 0 failed
```

Thread: https://tempoxyz.slack.com/archives/C0A87C21805/p1770671210505949

### DIFF
--- a/crates/commonware-node/src/subblocks.rs
+++ b/crates/commonware-node/src/subblocks.rs
@@ -701,8 +701,9 @@ async fn build_subblock(
     let (transactions, senders) = match evm_at_block(&node, parent_hash) {
         Ok(mut evm) => {
             let (mut selected, mut senders, mut to_remove) = (Vec::new(), Vec::new(), Vec::new());
-            let gas_budget =
-                evm.block().gas_limit / TEMPO_SHARED_GAS_DIVISOR / num_validators as u64;
+            let gas_budget = (evm.block().gas_limit / TEMPO_SHARED_GAS_DIVISOR)
+                .checked_div(num_validators as u64)
+                .expect("validator set must not be empty");
 
             let mut gas_left = gas_budget;
             let txs = transactions.lock().clone();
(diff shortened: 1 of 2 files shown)


crates/transaction-pool/src/validator.rs
```diff
@@ -592,6 +592,19 @@ where
...
+        let max_size = self.inner.max_tx_input_bytes();
+        if tx_size > max_size {
+            return TransactionValidationOutcome::Invalid(
+                transaction,
+                InvalidPoolTransactionError::OversizedData {
+                    size: tx_size,
+                    limit: max_size,
+                },
+            );
+        }
```
mattsse, line 606: this makes sense,

this field is kinda misnamed on reth.


I dont really like that we check this now twice effectively.

but this seems easier than breakign up internals and this check is cheap enough to do as soon as possible

but this is also only checked after we obtained a stateprovider here,
so we kinda dont have a separation of stateless vs stateful checks rn, imo then adding this here is a bit pointless because this doesnt really exit early

PR #2567, opened 2026-02-09
feat(ref-impls): ValidatorConfigV2 contract, interface, and tests

howydev:
## Summary

Adds the ValidatorConfigV2 reference implementation with Ed25519 signature verification and per-validator migration from V1.

### Contract (`ValidatorConfigV2.sol`)
- Ed25519 signature verification for `addValidator` and `rotateValidator`
- Per-validator migration via `migrateValidator(idx)` with owner copy on first call
- `initializeIfMigrated()` finalizes migration and copies `nextDkgCeremony` from V1
- Modifier order: `onlyInitialized` before `onlyOwner`/`onlyOwnerOrValidator`
- Renamed `transferOwner` → `transferOwnership`

### Interface (`IValidatorConfigV2.sol`)
- Full interface with all errors, structs, and function signatures

### Tests (`ValidatorConfigV2.t.sol`)
- 30 passing tests using real Ed25519 signatures
- Condensed pattern: `test_fn_pass` for success, `test_fn_fail` for grouped reverts via try/catch
- setUp adds V1 validators for natural migration (no direct storage writes)
- Cascading error order in fail tests to avoid resets

### Spec
- Updated `tip-1017.mdx` with `transferOwnership` rename

### DIFF
--- a/tips/ref-impls/src/ValidatorConfigV2.sol
+++ b/tips/ref-impls/src/ValidatorConfigV2.sol
@@ -30,10 +30,10 @@ contract ValidatorConfigV2 is IValidatorConfigV2 {
     Validator[] internal validatorsArray;
 
     /// @dev 1-indexed: 0 means not found. Stored value is arrayIndex + 1.
-    mapping(address => uint256) internal addressToIndex;
+    mapping(address => uint64) internal addressToIndex;
 
     /// @dev 1-indexed: 0 means not found. Stored value is arrayIndex + 1.
-    mapping(bytes32 => uint256) internal pubkeyToIndex;
+    mapping(bytes32 => uint64) internal pubkeyToIndex;
 
     uint64 internal nextDkgCeremony;
 
@@ -99,7 +99,7 @@ contract ValidatorConfigV2 is IValidatorConfigV2 {
             revert Unauthorized();
         }
 
-        uint256 idx = addressToIndex[validatorAddress];
+        uint64 idx = addressToIndex[validatorAddress];
         if (idx == 0) {
             revert ValidatorNotFound();
         }
@@ -141,7 +141,7 @@ contract ValidatorConfigV2 is IValidatorConfigV2 {
         onlyInitialized
         onlyOwnerOrValidator(validatorAddress)
     {
-        uint256 idx = addressToIndex[validatorAddress];
+        uint64 idx = addressToIndex[validatorAddress];
         if (idx == 0) {
             revert ValidatorNotFound();
         }
@@ -179,7 +179,7 @@ contract ValidatorConfigV2 is IValidatorConfigV2 {
             revert Unauthorized();
         }
 
-        uint256 idx = addressToIndex[validatorAddress];
+        uint64 idx = addressToIndex[validatorAddress];
         if (idx == 0) {
             revert ValidatorNotFound();
         }
@@ -210,7 +210,7 @@ contract ValidatorConfigV2 is IValidatorConfigV2 {
             revert InvalidValidatorAddress();
         }
 
-        uint256 idx = addressToIndex[currentAddress];
+        uint64 idx = addressToIndex[currentAddress];
         if (idx == 0) {
             revert ValidatorNotFound();
         }
@@ -239,10 +239,10 @@ contract ValidatorConfigV2 is IValidatorConfigV2 {
 
     /// @inheritdoc IValidatorConfigV2
     function getActiveValidators() external view returns (Validator[] memory validators) {
-        uint256 len = validatorsArray.length;
+        uint64 len = uint64(validatorsArray.length);
         validators = new Validator[](len);
-        uint256 idx = 0;
-        for (uint256 i = 0; i < len; i++) {
+        uint64 idx = 0;
+        for (uint64 i = 0; i < len; i++) {
             Validator storage v = validatorsArray[i];
             if (v.deactivatedAtHeight == 0) {
                 validators[idx] = v;
@@ -268,7 +268,7 @@ contract ValidatorConfigV2 is IValidatorConfigV2 {
     }
 
     /// @inheritdoc IValidatorConfigV2
-    function validatorByIndex(uint256 index) external view returns (Validator memory) {
+    function validatorByIndex(uint64 index) external view returns (Validator memory) {
         if (index >= validatorsArray.length) {
             revert ValidatorNotFound();
         }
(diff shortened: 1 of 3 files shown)


tips/ref-impls/src/interfaces/IValidatorConfigV2.sol
SuperFluffy, line 1: Where is this from? I'd rather we don't handroll crypto.

tips/ref-impls/test/ValidatorConfigV2.t.sol
```diff
@@ -0,0 +1,806 @@
...
+
+    function _getSignature(Signatures memory sigs) internal view returns (bytes memory) {
+        if (block.chainid == 42_431) return sigs.testnet;
+        if (block.chainid == 62_321) return sigs.mainnet;
+        return "";
+    }
+
+    function setUp() public override {
+        super.setUp();
+        vm.skip(isTempo);
```
howydev, line 81: TODO: remove after implemented in rust

tips/tip-1017.mdx
```diff
@@ -623,7 +620,7 @@ interface IValidatorConfigV1 {
 - **No signatures required**: V1 validators are imported without Ed25519 signatures
   (they were already validated in V1)
 - **All validators imported**: Both active and inactive V1 validators are
-  imported; active ones have `addedAtHeight == block.height` and `deactivatedAtheight = 0`,
+  imported; active ones have `addedAtHeight == block.height` and `deactivatedAtHeight = 0`,
```
SuperFluffy:
```suggestion
  imported; active ones have `addedAtHeight == block.height` and `deactivatedAtHeight == 0`,
```

tips/tip-1017.mdx
```diff
@@ -742,12 +739,11 @@ The test suite must cover:
...
-    are imported; active ones have `addedAtHeight > 0` and `deactivatedAtHeight == 0`
+    are imported; active ones have `addedAtHeight > 0` and `deactivatedAtHeight == 0`,
     inactive ones have `addedAtHeight == deactivatedAtHeight > 0`.
-38. **addedAtHeight set correctly**: Active validators have `addedAtHeight > 0`
-    at migration time.
-39. **deactivatedAtHeight set correctly**: Inactive validators have
-    `deactivatedAtHeight > 0` at migration time.
+38. **addedAtHeight set correctly**: All migrated validators have `addedAtHeight > 0` (block.height at migration time).
+39. **deactivatedAtHeight set correctly**: Active validators have `deactivatedAtHeight == 0`.
+    Inactive validators have `deactivatedAtHeight > 0` at migration time, equal to `addedAtHeight`.
```
SuperFluffy:
```suggestion
    Inactive validators have `addedAtHeight == deactivatedAtHeight > 0` at migration time.
```

tips/ref-impls/src/ValidatorConfigV2.sol
```diff
@@ -0,0 +1,726 @@
...
+
+    /// @dev Slot 0: bit 255 = initialized flag, bits 191-254 = initializedAtHeight, bits 0-159 = owner address
+    address private _owner;
+    bool private _initialized;
+    uint64 private _initializedAtHeight;
+
+    IValidatorConfig public immutable v1 =
+        IValidatorConfig(0xCccCcCCC00000000000000000000000000000000);
+
+    Validator[] internal validatorsArray;
```
SuperFluffy, line 30: Should we use `enumerableSet` as implemented in https://github.com/tempoxyz/tempo/pull/2249 ? IIRC valConfigV2 was the reason it was proposed.

tips/ref-impls/src/ValidatorConfigV2.sol
```diff
@@ -0,0 +1,726 @@
...
+    // Storage
+    // =========================================================================
+
+    /// @dev Slot 0: bit 255 = initialized flag, bits 191-254 = initializedAtHeight, bits 0-159 = owner address
+    address private _owner;
+    bool private _initialized;
+    uint64 private _initializedAtHeight;
+
+    IValidatorConfig public immutable v1 =
+        IValidatorConfig(0xCccCcCCC00000000000000000000000000000000);
```
SuperFluffy, line 28: Dangerous magic values. Is there any way we can have this be shared between the Rust precompile and the solidity code here? Dump it some file that includes this somehow? (kind of like `include_str!` in Rust). Probably not, just flagging.

tips/ref-impls/src/interfaces/IValidatorConfigV2.sol
```diff
@@ -0,0 +1,243 @@
...
+    ///        Epoch N runs the ceremony, and epoch N+1 uses the new DKG polynomial.
+    function setNextFullDkgCeremony(uint64 epoch) external;
+
+    // =========================================================================
+    // View Functions
+    // =========================================================================
+
+    /// @notice Get all validators (including deleted ones) in array order
+    /// @return validators Array of all validators with their information
+    function getValidators() external view returns (Validator[] memory validators);
```
SuperFluffy: Maybe `getAllValidators` to contrast with `getActiveValidators`? Just a suggestion; would be breaking with spec (/ spec needs updating).
```suggestion
    function getAllValidators() external view returns (Validator[] memory validators);
```

tips/ref-impls/src/ValidatorConfigV2.sol
howydev: This is basically an implementation of enumerableset, array + map

There arent any solidity libs we could use here since the libs cover base types and not structs

howydev: Will use that for the rust precompile

tips/ref-impls/src/ValidatorConfigV2.sol
howydev: hmm, dangerous in what way? 

we hard code a lot of these addresses, including within our tests since we need to know the address to test against

if addresses are wrong, the foundry-tempo test would fail

SuperFluffy: In the sense that these must match exactly between the sol and the rust precompile (aware that this is done everywhere; would be best if there was a single source of truth, but ok that there is not).

howydev: dont think thats possible unfortunately.. but we'd know if its incorrect b/c our foundry precompile tests will fail in ci, main, on cutting a release

PR #2575, opened 2026-02-10
fix(txpool): enforce exact calldata length for TIP-20 payment lane

### LINKED ISSUE #1379: Enforce exact calldata length for TIP-20 payment lane
### Describe the feature

The `is_payment()` validation/classification logic should be updated to only classify transactions as TIP-20 payments when calldata length exactly matches the expected ABI size for `transfer` or `transferWithMemo` (with memo capped at 32 bytes). Transactions with excess calldata should not be treated as TIP-20 payments and should not be included in dedicated payment blockspace.

### Additional context

_No response_

gakonst:
## Summary
Tightens `is_payment()` to validate exact calldata length, preventing transactions with excess/missing/malformed calldata from being classified as TIP-20 payments.

## Motivation
Closes #1379 — without this check, any transaction targeting a TIP-20 address was classified as a payment regardless of calldata, allowing non-payment calls to occupy dedicated payment blockspace.

## Changes
- Added `is_valid_payment_calldata()` helper in `crates/primitives/src/transaction/envelope.rs` that checks for `transfer` (68 bytes) or `transferWithMemo` (100 bytes) selectors with exact length match
- Updated all `is_payment()` match arms to require valid calldata in addition to the TIP-20 address prefix
- AA variant now rejects empty `calls` lists (previously vacuously true via `.all()`)
- Updated stale comment in `crates/transaction-pool/src/transaction.rs`
- Added tests for empty calldata, excess calldata, unknown selectors, and empty AA calls

## Testing
```
cargo test -p tempo-primitives -p tempo-transaction-pool -- test_payment   # 13 passed
cargo test -p tempo-primitives -- is_payment                              # 1 passed
cargo clippy -p tempo-primitives -p tempo-transaction-pool -- -D warnings # clean
cargo fmt --all --check                                                   # clean
```

Thread: https://tempoxyz.slack.com/archives/C0A87C21805/p1770688749512469


crates/primitives/src/transaction/envelope.rs
```diff
@@ -14,6 +14,30 @@ use core::fmt;
...
+/// matches the expected ABI-encoded size for that function.
+fn is_valid_payment_calldata(input: &[u8]) -> bool {
+    match input.first_chunk::<4>() {
+        Some(&s) if s == TRANSFER_SELECTOR => input.len() == TRANSFER_CALLDATA_LEN,
+        Some(&s) if s == TRANSFER_WITH_MEMO_SELECTOR => {
+            input.len() == TRANSFER_WITH_MEMO_CALLDATA_LEN
+        }
+        _ => false,
+    }
+}
```
0xrusowsky, line 38: nit, let's simplify:

```suggestion
fn is_valid_payment_calldata(input: &[u8]) -> bool {
    match input.first_chunk::<4>() {
        Some(&TRANSFER_SELECTOR) => input.len() == TRANSFER_CALLDATA_LEN,
        Some(&TRANSFER_WITH_MEMO_SELECTOR) => input.len() == TRANSFER_WITH_MEMO_CALLDATA_LEN,
        _ => false,
    }
}
```

0xrusowsky (commented):
why don't we also consider `transferFrom` as payments?

crates/primitives/src/transaction/envelope.rs
```diff
@@ -156,30 +180,46 @@ impl TempoTxEnvelope {
...
-                .is_some_and(|to| to.starts_with(&TIP20_PAYMENT_PREFIX)),
-            Self::Eip7702(tx) => tx.tx().to.starts_with(&TIP20_PAYMENT_PREFIX),
-            Self::AA(tx) => tx.tx().calls.iter().all(|call| {
-                call.to
+            Self::Legacy(tx) => {
+                tx.tx()
+                    .to
+                    .to()
+                    .is_some_and(|to| to.starts_with(&TIP20_PAYMENT_PREFIX))
+                    && is_valid_payment_calldata(&tx.tx().input)
```
fgimenez: nit, this is repeated a lot, could we extract the logic to a helper?

fgimenez (approved):

PR #2587, opened 2026-02-10
fix: default snapshot download URL to mainnet (4217)

gakonst:
`tempo download --chain mainnet` was downloading the moderato (42431) snapshot instead of mainnet (4217).

### DIFF
--- a/bin/tempo/src/defaults.rs
+++ b/bin/tempo/src/defaults.rs
@@ -118,7 +118,7 @@ pub(crate) struct TelemetryConfig {
 fn init_download_urls() {
     let download_defaults = DownloadDefaults {
         available_snapshots: vec![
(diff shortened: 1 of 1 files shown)


SuperFluffy:
Please reduce the amount of text in the PR body before we merge this. This is more text than worth this diff.

bin/tempo/src/defaults.rs
```diff
@@ -118,6 +118,7 @@ pub(crate) struct TelemetryConfig {
 fn init_download_urls() {
     let download_defaults = DownloadDefaults {
         available_snapshots: vec![
+            Cow::Borrowed("https://snapshots.tempoxyz.dev/4217 (mainnet)"),
```
SuperFluffy: This should be constructed from the default download url so (single source of truth).

PR #2632, opened 2026-02-11
tip: TIP-1011 Enhanced Access Key Permissions (call scoping + periodic limits)

legion2002:
## Summary

Closes #2299

This updates TIP-1011 to the current spec model for enhanced access-key permissions.

- Defines call scoping with `CallScope { target, selectors? }`, including `None` / `Some([])` / `Some([...])` semantics.
- Clarifies selector rules: selector entries are exactly 4 bytes, and call validation rejects calls that do not provide exactly 4 selector bytes when scoped selector matching is required.
- Specifies map-based call-scope validation model and per-target create-or-replace behavior for `setAllowedCalls`.
- Documents intrinsic gas delta for call scopes with explicit bounds: `MAX_SCOPES = 8`, `MAX_SELECTORS_PER_SCOPE = 16`.
- Tightens invariants/test cases for bounds, `allowAnySelector` precedence, create-or-replace behavior, and contract-creation rejection under access keys.
- Fixes/updates references (Tempo Transactions docs + local `IAccountKeychain.sol` path).
- Adds explicit batch-validation gas charging statement (normal EVM semantics up to failure).

Spec-only PR; no runtime code changes.

### DIFF
--- a/tips/tip-1011.md
+++ b/tips/tip-1011.md
@@ -1,47 +1,56 @@
 ---
 id: TIP-1011
 title: Enhanced Access Key Permissions
-description: Extends Access Keys with periodic spending limits and destination address scoping for subscription-based and restricted access patterns.
-authors: Tanishk Goyal 
+description: Extends Access Keys with periodic spending limits, destination/function scoping, and limited calldata recipient scoping.
+authors: Tanishk Goyal
 status: Draft
-related: TIP-1000, AccountKeychain, IAccountKeychain.sol
-protocolVersion: TBD (requires hardfork)
+related: Tempo Transaction Spec
+protocolVersion: T3
 ---
 
 # TIP-1011: Enhanced Access Key Permissions
 
 ## Abstract
 
-This TIP extends Access Keys with two new permission features: (1) **periodic spending limits** that automatically reset after a configurable time period, enabling subscription-based access patterns, and (2) **destination address scoping** that restricts keys to only interact with specific contract addresses.
+This TIP extends Access Keys with three permission features:
+
+1. **Periodic spending limits** that reset on fixed intervals.
+2. **Call scoping** that limits what addresses a key can call and which selectors it can use.
+3. **Limited calldata recipient scoping** for token transfer/approval selectors.
+
 
 ## Motivation
 
-Currently, Access Keys support spending limits (per-TIP-20 token caps) and expiry timestamps. However, these primitives are insufficient for common real-world patterns:
+Current Access Keys support per-token limits and expiry, but miss two practical controls.
 
 ### Periodic Spending Limits
 
-The existing `TokenLimit` specifies a one-time spending cap that depletes permanently. This model doesn't support subscription-based patterns where:
-
-- A service needs recurring access to a fixed amount per billing cycle
-- Users want to authorize "up to X tokens per month" without re-authorizing
-- dApps implement subscription models (e.g., streaming services, SaaS payments)
+One-time limits cannot express recurring allowances.
 
 **Use cases:**
-1. **Subscription services**: Authorize a streaming service to charge 10 USDC/month
-2. **Recurring donations**: Allow an NPO to withdraw up to 5 USDC weekly
-3. **Payroll systems**: Enable payroll contracts to transfer salaries monthly
-4. **Rate-limited APIs**: Authorize API access keys with per-period token budgets
 
-### Destination Address Scoping
+1. Subscription billing (`10 USDC / month`).
+2. Payroll schedules (monthly budgeted payouts).
+3. Rate-limited agent/API budgets.
 
-Users have requested the ability to bind Access Keys to specific destinations (e.g., "only allow transactions to Uniswap"). This provides more granular permission scoping similar to Solana's delegate primitive.
+### Call Scoping (Target + Selector Set)
 
-**Use cases:**
-1. **DeFi integrations**: Allow a trading bot key to only interact with specific DEX contracts
-2. **Gaming**: Scope a session key to only interact with a game contract
-3. **Subscription services**: Allow a key to only call a specific payment contract
+Users need finer controls than "any call". They want keys like:
 
-**Current workaround**: Deploy a proxy contract that enforces destination restrictions, adding gas overhead and complexity.
+1. "Only call `swap()` and `exactInput()` on DEX X."
+2. "Only call gameplay methods on contract Y."
+3. "Only perform plain transfers, not token extension methods."
+4. "Only vote() on governance contracts."
+
+**Current workaround**: Deploy a proxy contract that enforces destination/function restrictions, adding gas overhead and complexity.
+
+### Recipient-Bound Token Calls
+
+Target + selector scoping still allows an access key to move funds to arbitrary recipients for token methods like `transfer` and `approve`.
+
+Users need a narrower policy: the key may call transfer/approve selectors, but only when the recipient/spender matches a configured address.
+
+This TIP intentionally adds a narrow calldata rule (first ABI `address` argument equality) instead of a generic calldata policy language.
 
 ---
 
@@ -49,9 +58,15 @@ Users have requested the ability to bind Access Keys to specific destinations (e
 
 ## Extended Data Structures
 
+Conventions used in this section:
+
+1. Protocol/RLP structs are written with Rust-like `Option<...>` notation.
+2. Solidity ABI structs are listed separately where ABI cannot directly represent protocol `Option` semantics.
+
 ### TokenLimit
 
 **Current:**
+
 ```solidity
 struct TokenLimit {
     address token;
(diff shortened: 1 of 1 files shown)


tips/tip-1011.md
```diff
@@ -70,21 +76,40 @@ struct TokenLimit {
...
+    address target;     // Contract address (address(0) = wildcard for any address)
+    bytes4 selector;    // Function selector (bytes4(0) = wildcard for any function)
+}
+```
+
+**Semantics:**
+- `target=0x123, selector=0xaabbccdd` → Only call function `0xaabbccdd` on address `0x123`
+- `target=0x123, selector=0x00000000` → Call any function on address `0x123` (address-only scoping)
+- `target=0x0, selector=0xaabbccdd` → Call function `0xaabbccdd` on any address (selector-only scoping)
+- Empty `allowedCalls` array → Unrestricted (can call any function on any address)
```
howydev: how are we thinking about custom selectors <4 bytes (fallbacks)? block all?

tips/tip-1011.md
```diff
@@ -70,21 +76,40 @@ struct TokenLimit {
...
+```solidity
+struct CallScope {
+    address target;     // Contract address (address(0) = wildcard for any address)
+    bytes4 selector;    // Function selector (bytes4(0) = wildcard for any function)
+}
+```
+
+**Semantics:**
+- `target=0x123, selector=0xaabbccdd` → Only call function `0xaabbccdd` on address `0x123`
+- `target=0x123, selector=0x00000000` → Call any function on address `0x123` (address-only scoping)
```
howydev: recommend separating this into a separate field since bytes4(0) is a valid selector

theres still 8 bytes remaining in storage, maybe use 1 for a `bool allowAllSelectors` instead

tips/tip-1011.md
```diff
@@ -137,38 +165,48 @@ function verifyAndUpdateSpending(key, token, amount):
...
 
 ## Gas Costs
 
 | Operation | Additional Gas |
 |-----------|----------------|
-| `authorizeKey` with N destinations | ~20,000 + 5,000 × N |
+| `authorizeKey` with N call scopes | ~20,000 + 7,000 × N |
 | `authorizeKey` with periodic limit | ~3,000 per periodic token |
-| Destination check (per tx) | ~2,100 + 200 × N |
+| Call scope check (per call) | ~2,100 + 300 × N |
```
howydev: Does the number of storage reads increase with `n`? I'm not following why `300`

jxom (approved):

snario:
Also an `AccessKeySpend` event.

`Transfer(from, to, amount)` only has the account address as from, it doesn't tell you which key signed the transaction. So you'd need to:

- Get the Transfer event
- Look up the transaction that emitted it
- Decode the transaction's authorization field to extract the signing key
- Cross-reference that against known Access Keys


An `AccessKeySpend(account, keyId, token, amount, remainingInPeriod)` event from the precompile would make it a single indexed log query.

tips/tip-1011.md
```diff
@@ -191,6 +229,11 @@ TokenLimit := RLP([
     period: uint64,
     periodEnd: uint64
 ])
+
+CallScope := RLP([
+    target: address,
+    selector: bytes4
+])
 ```
```
jxom, line 498: Can we add the sign format here too?

```
key_auth_digest = rlp(chain_id || key_type || key_id || expiry? || limits? || allowed_destinations?)
limits = rlp(token || limit || period?)
```

onbjerg (commented):
how do we handle fallback? do we intentionally leave this unsupported/only supported if you allow all?

tips/tip-1011.md
```diff
@@ -70,42 +76,81 @@ struct TokenLimit {
...
+struct CallScope {
+    address target;     // Contract address (address(0) = wildcard for any address)
+    bytes4 selector;    // Function selector (bytes4(0) = wildcard for any function)
+}
+```
+
+**Semantics:**
+- `target=0x123, selector=0xaabbccdd` → Only call function `0xaabbccdd` on address `0x123`
+- `target=0x123, selector=0x00000000` → Call any function on address `0x123` (address-only scoping)
+- `target=0x0, selector=0xaabbccdd` → Call function `0xaabbccdd` on any address (selector-only scoping)
```
horsefacts: is there a concrete use case for this? hard for me to think of one.

tips/tip-1011.md
```diff
@@ -137,38 +185,48 @@ function verifyAndUpdateSpending(key, token, amount):
...
+
 ### Interaction Rules
 
 1. **Mixed limits**: Keys can have a mix of one-time and periodic limits for different tokens
-2. **Destination + limits**: Both constraints are evaluated independently; both must pass
+2. **Call scope + limits**: Both constraints are evaluated independently; both must pass
 3. **Period updates**: Calling `updateSpendingLimit()` updates the per-period amount and resets the current period
-4. **Empty destinations**: An empty `allowedDestinations` array means the key is unrestricted (can call any address)
+4. **Empty calls**: An empty `allowedCalls` array means the key is unrestricted (can call any function on any address)
+5. **Batch transactions**: For AA transactions with multiple calls, each call is validated independently against the allowed scopes
```
horsefacts: we should specify explicitly what happens if one is invalid. I'm assuming the batch reverts.

tips/tip-1011.md
horsefacts: does this belong in here?

tips/tip-1011.md
```diff
@@ -137,38 +185,48 @@ function verifyAndUpdateSpending(key, token, amount):
...
+        
+        if targetMatch && selectorMatch:
             return true
     
-    revert DestinationNotAllowed()
+    revert CallNotAllowed()
 ```
 
+**Special cases:**
+- Empty calldata (ETH transfer): selector is treated as `bytes4(0)`. A scope with `selector=bytes4(0)` matches ETH transfers.
```
horsefacts: we should probably not call this an "ETH transfer," should we?

tips/tip-1011.md
legion2002: Yup, we'll just block them. Will update the TIP

tips/tip-1011.md
legion2002: hmm we should identify `allowAllSelectors` by null probably. But it's a good point to separate this, because we don't want to duplicate the address every time, if we want to permit multiple selectors for an address.
Thinking about a better structure here.

tips/tip-1011.md
legion2002: yeah good q, maybe there are multiple pools in a DEX, and you want to allow "swap" selectors for all the addresses.
Or you want to allow transfers on all TIP20s, but want to ensure that it cannot call any other admin functions.

We didn't add this feature specifically, think it flows naturally from the other semantics of the spec, but we can consider blocking it.

Although the TIP 20 transfer use case seems nice to have

legion2002: I removed this behaviour in the latest version

onbjerg (commented):
reiterating my previous comment: even with the new changes, this does not address how you would disallow or allow the fallback selector without also allowing or disallowing every other selector, unless I am missing something

tips/tip-1011.md
```diff
@@ -60,254 +61,424 @@ struct TokenLimit {
...
+2. `Some([])` => allow no selectors (matches nothing for that scope).
+3. `Some([s1, s2, ...])` => allow exactly the listed selectors.
+
+Examples:
+
+1. `{ target: 0x123, selectors: Some([0xaabbccdd, 0xeeff0011]) }` => allow two selectors on one target.
+2. `{ target: 0x123, selectors: None }` => address-only scoping (any selector on `0x123`).
+3. `{ target: 0x123, selectors: Some([]) }` => matches no selector on `0x123`.
+4. `allowedCalls = None` => unrestricted key.
+5. `allowedCalls = Some([])` => key is authorized but cannot make scoped calls.
```
samczsun: What is the point of this?

tips/tip-1011.md
```diff
@@ -60,254 +61,424 @@ struct TokenLimit {
...
+    target: address,
+    selectors: Option<Vec<bytes4>>,   // None => any selector on this target
+}
+```
+
+`selectors` behavior intentionally mirrors `limits` tri-state semantics:
+
+1. `None` => allow any selector.
+2. `Some([])` => allow no selectors (matches nothing for that scope).
+3. `Some([s1, s2, ...])` => allow exactly the listed selectors.
```
samczsun: There's no way to say "I only want to allow selectorless (0-3 bytes) calls"? What is the point of `Some([])` if lack of a scope for an address just reverts?

tips/tip-1011.md
```diff
@@ -60,254 +61,424 @@ struct TokenLimit {
...
+
 ### IAccountKeychain.sol
 
 ```solidity
 /// @notice Authorizes a key with enhanced permissions
 /// @param key The public key to authorize
 /// @param expiry Block timestamp when key expires
 /// @param spendingLimits Token spending limits (may include periodic limits)
-/// @param allowedDestinations Addresses the key may call (empty = unrestricted)
+/// @param allowedCalls Call scopes. Omitted at tx auth layer => unrestricted.
```
samczsun: What does this even mean

tips/tip-1011.md
samczsun, line 207: What happens if `allowAnySelector = true` and `selectors = [...]`? Why is `allowAnySelector = false` and `selectors = []` not just a delete instead of "delete-equivalent"?

tips/tip-1011.md
```diff
@@ -60,254 +61,424 @@ struct TokenLimit {
...
-function validateDestination(key, destination):
-    allowed = getAllowedDestinations(key)
-    
-    if allowed.length == 0:
-        return true  // Unrestricted
-    
-    for addr in allowed:
-        if addr == destination:
+
+Calldata smaller than 4 bytes is treated as selectorless (`None`) and MUST NOT panic.
```
samczsun: Why is panicking in the spec

tips/tip-1011.md
```diff
@@ -60,254 +61,424 @@ struct TokenLimit {
...
+    allowed = getAllowedCalls(account, key)  // Option<Vec<CallScope>>
+
+    if key != Address::ZERO && isContractCreation:
+        revert CallNotAllowed()  // global contract-creation ban for access keys
+
+    if allowed is None:
+        return true  // unrestricted
+
+    if isContractCreation:
+        revert CallNotAllowed()  // root-key create cannot be whitelisted when call scopes are set
```
samczsun: So even the root account is not allowed to deploy contracts?

tips/tip-1011.md
```diff
@@ -60,254 +61,424 @@ struct TokenLimit {
...
             return true
-    
-    revert DestinationNotAllowed()
+
+    revert CallNotAllowed()
 ```
 
+#### Batch Validation
+
+For AA transactions with multiple calls, each call MUST be validated independently. If any call fails scope validation, the entire batch MUST revert.
```
samczsun: Does the user get charged for only the gas used to validate up to the failing call, or all calls?

tips/tip-1011.md
```diff
@@ -60,254 +61,424 @@ struct TokenLimit {
...
+
+### Root-Controlled Scope Updates
```
samczsun, line 347: This just duplicates what the natspec says

tips/tip-1011.md
```diff
@@ -60,254 +61,424 @@ struct TokenLimit {
...
+7. Per-target updates are create-or-replace and duplicate target scopes are not allowed.
 
-## Gas Costs
+## Gas And Complexity Bounds
+
+To prevent unbounded validation/storage growth, implementations MUST enforce:
+
+1. `MAX_CALL_SCOPES` (recommended: 64).
+2. `MAX_SELECTORS_PER_SCOPE` (recommended: 64).
+3. `MAX_TOTAL_SELECTORS` (recommended: 256).
```
samczsun: What does recommend mean in a spec where it says "MUST" enforce

tips/tip-1011.md
```diff
@@ -60,254 +61,424 @@ struct TokenLimit {
...
 |-----------|----------------|
-| `authorizeKey` with N destinations | ~20,000 + 5,000 × N |
+| `authorizeKey` with `S` scopes and `K` total selectors | ~20,000 + 7,000 × S + 300 × K |
+| `setAllowedCalls` (single target) with `K` selectors | ~5,000 + 300 × K |
 | `authorizeKey` with periodic limit | ~3,000 per periodic token |
-| Destination check (per tx) | ~2,100 + 200 × N |
+| Scope validation (per runtime call) | ~2,100 + 300 × K' |
 | Period reset (when triggered) | ~5,000 |
 
+`K'` is selectors examined for that destination during matching.
```
samczsun: Where did these numbers come from?

tips/tip-1011.md
```diff
@@ -60,254 +61,424 @@ struct TokenLimit {
...
+
 ## Encoding
 
-### Transaction Authorization
+### Signing Format
```
samczsun, line 449: What is this section for?

tips/tip-1011.md
```diff
@@ -60,254 +61,424 @@ struct TokenLimit {
...
+
+```text
 KeyAuthorization := RLP([
-    spendingLimits: [TokenLimit, ...],
-    expiry: uint64,
-    allowedDestinations: [address, ...]
+    chain_id: u64,
+    key_type: u8,
+    key_id: address,
+    expiry?: uint64,
```
samczsun, line 478: Why did `expiry` become optional? What does it mean if it's `None`?

tips/tip-1011.md
```diff
@@ -60,254 +61,424 @@ struct TokenLimit {
...
 
-### KeyAuthorization (trailing field addition)
+### KeyAuthorization Trailing Field
 
-`KeyAuthorization` uses `#[rlp(trailing)]` which allows appending optional fields. Adding `allowedDestinations: Option<Vec<Address>>` as the last field is compatible with this pattern:
+`KeyAuthorization` uses trailing optional fields. `allowed_calls` is appended at the end to preserve decode for old encodings:
 
-- Old encodings (without `allowedDestinations`) decode as `allowedDestinations = None` (unrestricted)
-- New encodings with `allowedDestinations` will be rejected by old nodes
+1. Old tx encodings decode as `allowed_calls = None` (unrestricted).
```
samczsun: How?

tips/tip-1011.md
```diff
@@ -60,254 +61,424 @@ struct TokenLimit {
...
-  └─ key_authorization: Option<SignedKeyAuthorization>
-       └─ SignedKeyAuthorization (custom Compact impl → uses RLP internally)
-            └─ KeyAuthorization (RLP encoded)
-                 └─ limits: Option<Vec<TokenLimit>> (RLP encoded)
-```
-
-`SignedKeyAuthorization` implements a custom `Compact` that wraps RLP encoding. Therefore, `TokenLimit` is always serialized as RLP bytes in the database, not via its derived Compact layout.
-
-**Implication**: If we implement version-tolerant RLP decoding for `TokenLimit` (accept 2-field or 5-field lists), existing database entries will decode correctly. **No DB rebuild required** for this change.
+`SignedKeyAuthorization` uses custom Compact encoding backed by RLP. `TokenLimit` is therefore stored as RLP in production paths; version-tolerant decode avoids DB rebuild requirements.
```
samczsun: Why is this in the spec?

tips/tip-1011.md
```diff
@@ -60,254 +61,424 @@ struct TokenLimit {
...
-The current storage layout:
-- `keys[account][keyId] → AuthorizedKey` (packed slot)
-- `spending_limits[key][token] → U256` (remaining amount)
+Current layout:
 
-This TIP requires additional per-token state for periodic limits. **Additive storage approach** (no migration required):
+1. `keys[account][keyId] -> AuthorizedKey`
+2. `spending_limits[(account,keyId)][token] -> U256`
+
+Additive periodic-limit layout:
```
samczsun, line 519: Why isn't this two structs?

tips/tip-1011.md
```diff
@@ -60,254 +61,424 @@ struct TokenLimit {
...
-3. **Destination scoping enforcement**: Validate transaction `to` against `allowedDestinations`
-4. **New precompile storage writes**: Write to new storage slots for period/destination data
-5. **New precompile interface methods**: `getAllowedDestinations()`, updated `getRemainingLimit()` return type
+1. New `TokenLimit` decode/encode behavior.
+2. `allowed_calls` decode/encode behavior.
+3. Periodic reset logic.
+4. Call-scope validation logic.
+5. New precompile storage writes/reads for periodic + call-scope data.
+6. Updated precompile read APIs (`getAllowedCalls(account,key)`, richer `getRemainingLimit`).
+7. New root-control mutator (`setAllowedCalls`) semantics.
```
samczsun: what does point 7 mean?

tips/tip-1011.md
```diff
@@ -60,254 +61,424 @@ struct TokenLimit {
...
 /// @param expiry Block timestamp when key expires
 /// @param spendingLimits Token spending limits (may include periodic limits)
-/// @param allowedDestinations Addresses the key may call (empty = unrestricted)
+/// @param allowedCalls Call scopes. Omitted at tx auth layer => unrestricted.
 function authorizeKey(
     bytes calldata key,
     uint64 expiry,
     TokenLimit[] calldata spendingLimits,
-    address[] calldata allowedDestinations
+    CallScope[] calldata allowedCalls
```
samczsun, line 259: Where did `signatureType` and `enforceLimits` go? Why is `key` a `bytes` now? This is a breaking change

tips/tip-1011.md
```diff
@@ -60,254 +61,424 @@ struct TokenLimit {
...
+    address account,
+    bytes calldata key
+) external view returns (CallScope[] memory calls);
+
+/// @notice Returns remaining limit for a token, accounting for period resets
+function getRemainingLimit(
+    address account,
+    bytes calldata key,
+    address token
+) external view returns (uint256 remaining, uint64 periodEnd);
```
samczsun, line 286: This is also a breaking change

tips/tip-1011.md
```diff
@@ -60,254 +61,424 @@ struct TokenLimit {
...
-
-5. **Expiry precedence**: Key expiry MUST be checked before spending limits or destination restrictions.
+1. `periodEnd` is monotonic and never set to the past.
+2. `remainingInPeriod <= limit` after any operation.
+3. If `allowed_calls` is `Some`, calls not matching any scope MUST revert.
+4. `allowed_calls = None` preserves unrestricted behavior for non-create calls.
+5. If `key != Address::ZERO`, any contract-creation call MUST revert regardless of `allowed_calls`.
+6. Expiry check runs before spending/call-scope checks.
+7. For each `(account, key)`, there is at most one scope per target.
+8. `setAllowedCalls(..., allowAnySelector = false, selectors = [])` disables that target scope.
```
samczsun, line 580: Why is this in the spec?

tips/tip-1011.md
```diff
@@ -1,8 +1,8 @@
 ---
 id: TIP-1011
 title: Enhanced Access Key Permissions
-description: Extends Access Keys with periodic spending limits and destination address scoping for subscription-based and restricted access patterns.
-authors: Tanishk Goyal 
+description: Extends Access Keys with periodic spending limits and destination/function scoping for subscription-based and restricted access patterns.
+authors: Tanishk Goyal
 status: Draft
 related: TIP-1000, AccountKeychain, IAccountKeychain.sol
```
samczsun: Why are the contract files in the frontmatter?

samczsun (changes requested):

tips/tip-1011.md
legion2002: This is maintain consistency with how other fields like "limits" are treated for access keys. There's no real practical utility, but mentioned in the spec to cover all cases.

tips/tip-1011.md
legion2002: yes it is mentioned under hardfork-gated features

tips/tip-1011.md
legion2002: Updated to - "New mutator function `setAllowedCalls`,which can only be called by root key. "

tips/tip-1011.md
legion2002: according to the spec template, each TIP needs to specify a set of invariants and test cases that should hold

tips/tip-1011.md
legion2002: fixed

tips/tip-1011.md
legion2002: we've faced problems before when trying to make changes to the Tx RLP fields, because there are problems with hardfork gating compact encoded txs in the DB. 
It is explicitly specified here, to show that we can make these changes in a backwards compatible way, without doing complicated DB operations.

tips/tip-1011.md
legion2002: outdated change, fixed.

tips/tip-1011.md
legion2002: It was always optional. All optional fields in key authorizations follow similar semantics, mentioned above in the spec.

1. `None` => unlimited expiry/limits/selectors
2. `Some([])` => no limits/selectors
3. `Some([s1, s2, ...])` => allow exactly the listed expiry/limits/selectors.

tips/tip-1011.md
legion2002: This specifies how to calculate the message digest for signing key auths. Needed by the chain, and also for the wallet libs.

tips/tip-1011.md
```diff
@@ -60,254 +61,422 @@ struct TokenLimit {
...
+4. Address-only allow (`selectors=None`).
+5. Deny when no scope matches.
+6. `allowed_calls=None` allows all non-create calls.
+7. `allowed_calls=Some([])` denies all calls.
+8. Mixed one-time and periodic token limits.
+9. Existing pre-fork keys still function post-fork.
+10. Batch validation reverts entire batch on one invalid call.
+11. Calldata lengths `0..=3` never panic and follow selectorless semantics.
+12. Shared key IDs across accounts cannot overwrite each other’s scopes.
+13. Enforced bounds reject too many scopes/selectors.
```
malleshpai: but i think we removed the actual bounds

tips/tip-1011.md
```diff
@@ -60,254 +61,422 @@ struct TokenLimit {
...
+15. `setAllowedCalls(..., allowAnySelector = true, selectors = [])` allows any selector on that target.
+16. `setAllowedCalls(..., allowAnySelector = false, selectors = [])` blocks all calls to that target (unless unrestricted mode is active).
+17. Access-key transactions with CREATE as first call are rejected.
+18. Access-key transactions with any CREATE in a batch are rejected.
 
 ## References
 
 - [AccountKeychain docs](https://docs.tempo.xyz/protocol/transactions/AccountKeychain)
-- [IAccountKeychain.sol](tips/ref-impls/src/interfaces/IAccountKeychain.sol)
+- [IAccountKeychain.sol](docs/specs/src/interfaces/IAccountKeychain.sol)
```
malleshpai: deadlink

tips/tip-1011.md
legion2002: I added cleaner bounds, and added some invariants back.

tips/tip-1011.md
legion2002: refactored the spec, removed.

tips/tip-1011.md
legion2002: because to actually do the delete we'll have to iterate and write to an array of slots, which could lead to the tx overflowing, and the revocation being blocked.

This approach does litter the state a little. Wondering if it is worth using an enumerable map here instead.

tips/tip-1011.md
legion2002: fixed

tips/tip-1011.md
legion2002: we've completely blocked selectorless calls now. 

Some([]) is just there to retain semantic compatibility with other things like limits. I don't think it has any practical use

tips/tip-1011.md
```diff
@@ -60,254 +61,382 @@ struct TokenLimit {
...
-3. **Destination enforcement**: If `allowedDestinations` is non-empty, transactions to addresses not in the list MUST revert.
-
-4. **Backward compatibility**: Keys authorized without the new fields MUST behave as unrestricted (`allowedDestinations = []`) with one-time limits (`period = 0`).
-
-5. **Expiry precedence**: Key expiry MUST be checked before spending limits or destination restrictions.
+1. `periodEnd` is monotonic and never set to the past.
+2. `remainingInPeriod <= limit` after any operation.
+3. If `allowed_calls` is `Some`, calls not matching any scope MUST revert.
+4. `allowed_calls = None` preserves unrestricted behavior for non-create calls.
+5. If `key != Address::ZERO`, any contract-creation call MUST revert regardless of `allowed_calls`.
```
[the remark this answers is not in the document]
legion2002: invalid. We will add this in T3.

tips/tip-1011.md
```diff
@@ -60,254 +61,382 @@ struct TokenLimit {
...
+    address token
+) external view returns (uint256 remaining, uint64 periodEnd);
 ```
 
 ## Semantic Behavior
 
 ### Periodic Limit Reset Logic
 
-When a spending attempt occurs:
+On each spend attempt for `(account, key, token)`:
```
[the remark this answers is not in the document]
legion2002: These are known limitations. We cannot fix them without major changes. Scoped keys actually help with burn/burnWithMemo.

tips/tip-1011.md
```diff
@@ -60,254 +61,382 @@ struct TokenLimit {
...
+
 ```solidity
 struct TokenLimit {
     address token;
-    uint256 limit;      // Per-period limit when period > 0, one-time limit otherwise
-    uint256 remainingInPeriod;  // Remaining allowance in current period
-    uint64 period;      // Period duration in seconds (0 = one-time limit)
-    uint64 periodEnd;   // Timestamp when current period expires
+    uint256 limit;  // One-time cap when period == 0, per-period cap when period > 0
+    uint64 period;  // Period duration in seconds (0 = one-time limit)
```
howydev, line 83: what if we pack `period` with `token` above? this saves a storage slot

we'd want to securely test this to make sure that we don't brick or modify existing token limits

tips/tip-1011.md
```diff
@@ -60,254 +61,382 @@ struct TokenLimit {
...
+CallScope {
+    target: address,
+    selectors: Option<Vec<bytes4>>,   // None => any selector on this target
 }
 ```
 
+`selectors` behavior intentionally mirrors `limits` tri-state semantics:
+
+1. `None` => allow any selector.
+2. `Some([])` => allow no selectors (matches nothing for that scope).
```
howydev: nit: `None => no selector` is more obvious. also in general, default-value = no permissions is a safer pattern

tips/tip-1011.md
```diff
@@ -60,254 +61,382 @@ struct TokenLimit {
...
+3. `Some([s1, s2, ...])` => allow exactly the listed selectors.
+
+Selector width is fixed at 4 bytes.
+
+1. Each selector in `Some([s1, s2, ...])` MUST be exactly 4 bytes.
+2. Implementations MUST revert when decoding or accepting any selector whose length is not exactly 4 bytes.
+3. . We do not explicitly support fallback selectors, but root users can implicitly allow them by approving a 4 byte selector not in the ABI of the target contract.
```
howydev: nit typo
```suggestion
3. We do not explicitly support fallback selectors, but root users can implicitly allow them by approving a 4 byte selector not in the ABI of the target contract.
```

howydev: random selector might not be equivalent to fallback though

for upgradeable contracts, theres a risk that a new selector is added to the contract and the user loses access to fallback/gains access to that selector. could be dangerous

tips/tip-1011.md
```diff
@@ -60,254 +61,382 @@ struct TokenLimit {
...
-    uint64 expiry;
-    address[] allowedDestinations; // Empty array = unrestricted
+```text
+KeyAuthorization {
+    chain_id: u64,
+    key_type: SignatureType,
+    key_id: address,
+    expiry: Option<u64>,
+    limits: Option<Vec<TokenLimit>>,
+    allowed_calls: Option<Vec<CallScope>>,  // New trailing field
```
howydev, line 217: Just to check - if a user wants to allow a key to swap stables on the stablecoin dex and swap unstable tokens on uniswap, this is possible without giving perms to swap stable tokens on uniswap right?

tips/tip-1011.md
```diff
@@ -60,254 +61,382 @@ struct TokenLimit {
...
-/// @return periodEnd The timestamp when the current period ends (0 if one-time limit)
-function getRemainingLimit(bytes calldata key, address token) external view returns (uint256 remaining, uint64 periodEnd);
+/// @notice Returns allowed call scopes for an account+key
+function getAllowedCalls(
+    address account,
+    address keyId
+) external view returns (CallScope[] memory calls);
+
+/// @notice Returns remaining limit for a token, accounting for period resets
+function getRemainingLimit(
```
howydev, line 282: IIUC a key could have a periodic limit + one-time limit, what should this return in that case?

tips/tip-1011.md
```diff
@@ -60,254 +61,382 @@ struct TokenLimit {
...
-/// @return remaining The remaining spending limit for the current period
-/// @return periodEnd The timestamp when the current period ends (0 if one-time limit)
-function getRemainingLimit(bytes calldata key, address token) external view returns (uint256 remaining, uint64 periodEnd);
+/// @notice Returns allowed call scopes for an account+key
+function getAllowedCalls(
+    address account,
+    address keyId
+) external view returns (CallScope[] memory calls);
+
+/// @notice Returns remaining limit for a token, accounting for period resets
```
howydev, line 281: nice

tips/tip-1011.md
```diff
@@ -60,254 +61,382 @@ struct TokenLimit {
...
+Authorization digest format:
 
+```text
+key_auth_digest = keccak256(rlp([
+  chain_id,
+  key_type,
+  key_id,
+  expiry?,
+  limits?,
+  allowed_calls?
```
howydev, line 460: not too familiar with RLP - any chance of preimage collision here?

howydev:
Nice, the spec is definitely more polished and decisions thought out a little better!

Did half a pass and nth major so far, will continue tmr

tips/tip-1011.md
```diff
@@ -1,47 +1,47 @@
...
+This TIP extends Access Keys with two permission features:
+
+1. **Periodic spending limits** that reset on fixed intervals.
+2. **Call scoping** that limits where a key can call and which selectors it can use.
+
 
 ## Motivation
 
-Currently, Access Keys support spending limits (per-TIP-20 token caps) and expiry timestamps. However, these primitives are insufficient for common real-world patterns:
+Current Access Keys support per-token limits and expiry, but miss two practical controls.
```
onbjerg:
```suggestion
Currently Access Keys support per-token limits and expiry, but miss two practical controls.
```

tips/tip-1011.md
```diff
@@ -1,47 +1,47 @@
...
 
 # TIP-1011: Enhanced Access Key Permissions
 
 ## Abstract
 
-This TIP extends Access Keys with two new permission features: (1) **periodic spending limits** that automatically reset after a configurable time period, enabling subscription-based access patterns, and (2) **destination address scoping** that restricts keys to only interact with specific contract addresses.
+This TIP extends Access Keys with two permission features:
+
+1. **Periodic spending limits** that reset on fixed intervals.
+2. **Call scoping** that limits where a key can call and which selectors it can use.
```
onbjerg:
```suggestion
2. **Call scoping** that limits what accounts a key can call and which selectors it can use.
```

tips/tip-1011.md
```diff
@@ -60,254 +61,382 @@ struct TokenLimit {
...
+```
+
+Runtime state is derived and stored by the precompile (not signed):
+
+```text
+TokenLimitState {
+    remainingInPeriod: uint256,
+    periodEnd: uint64,
+}
+```
```
onbjerg, line 96: any reason we use solidity in some places, pseudo rust in others? can we unify?

tips/tip-1011.md
```diff
@@ -60,254 +61,382 @@ struct TokenLimit {
...
+
+### CallScope
+
+Call scoping is defined with optional fields (protocol-level / RLP model):
+
+```text
+CallScope {
+    target: address,
+    selectors: Option<Vec<bytes4>>,   // None => any selector on this target
```
onbjerg: same comment as above

tips/tip-1011.md
```diff
@@ -60,254 +61,382 @@ struct TokenLimit {
...
 }
 ```
 
+`selectors` behavior intentionally mirrors `limits` tri-state semantics:
+
+1. `None` => allow any selector.
```
onbjerg:
```suggestion
1. `None`: allow any selector.
```

tips/tip-1011.md
onbjerg:
```suggestion
2. `Some([])`: allow no selectors (matches nothing for that scope).
```

tips/tip-1011.md
```diff
@@ -60,254 +61,382 @@ struct TokenLimit {
...
+2. `Some([])` => allow no selectors (matches nothing for that scope).
+3. `Some([s1, s2, ...])` => allow exactly the listed selectors.
```
onbjerg:
```suggestion
3. `Some([s1, s2, ...])`: allow exactly the listed selectors.
```

tips/tip-1011.md
```diff
@@ -60,254 +61,382 @@ struct TokenLimit {
...
+3. . We do not explicitly support fallback selectors, but root users can implicitly allow them by approving a 4 byte selector not in the ABI of the target contract.
+4. Contracts with non-standard selector parsing are NOT supported.
+
+Examples:
+
+1. `{ target: 0x123, selectors: Some([0xaabbccdd, 0xeeff0011]) }` => allow two selectors on one target.
+2. `{ target: 0x123, selectors: None }` => address-only scoping (any selector on `0x123`).
+3. `{ target: 0x123, selectors: Some([]) }` => matches no selector on `0x123`.
+4. `allowedCalls = None` => unrestricted key.
+5. `allowedCalls = Some([])` => key is authorized but cannot make scoped calls.
```
onbjerg:
```suggestion
1. `{ target: 0x123, selectors: Some([0xaabbccdd, 0xeeff0011]) }`: allow two selectors on one target.
2. `{ target: 0x123, selectors: None }`: address-only scoping (any selector on `0x123`).
3. `{ target: 0x123, selectors: Some([]) }`: matches no selector on `0x123`.
4. `allowedCalls = None`: unrestricted key.
5. `allowedCalls = Some([])`: key is authorized but cannot make scoped calls.
```

tips/tip-1011.md
```diff
@@ -60,254 +61,382 @@ struct TokenLimit {
...
+Existing fields remain, with a trailing optional call-scope field:
 
-**New field:**
-```solidity
-struct KeyAuthorization {
-    TokenLimit[] spendingLimits;
-    uint64 expiry;
-    address[] allowedDestinations; // Empty array = unrestricted
+```text
+KeyAuthorization {
```
onbjerg, line 211: same as before, this is pseudo rust

onbjerg (approved):
I am approving the PR as I only have copy nits now, which are not blocking.

tips/tip-1011.md
```diff
@@ -60,254 +61,382 @@ struct TokenLimit {
...
-    uint256 limit;      // Per-period limit when period > 0, one-time limit otherwise
-    uint256 remainingInPeriod;  // Remaining allowance in current period
-    uint64 period;      // Period duration in seconds (0 = one-time limit)
-    uint64 periodEnd;   // Timestamp when current period expires
+    uint256 limit;  // One-time cap when period == 0, per-period cap when period > 0
+    uint64 period;  // Period duration in seconds (0 = one-time limit)
+}
+```
+
+Runtime state is derived and stored by the precompile (not signed):
```
horsefacts, line 89: When is this first stored? At key creation time or time of first spend?

tips/tip-1011.md
```diff
@@ -60,254 +61,382 @@ struct TokenLimit {
...
+2. `allowAnySelector = false` with `selectors = []` disables that target scope.
+3. Implementations MUST enforce at most one scope per target for each `(account, key)`.
 
-### TokenLimit (2 fields → 5 fields)
+### Interaction Rules
 
-The current `TokenLimit` struct encodes as `[token, limit]`. This TIP extends it to `[token, limit, remainingInPeriod, period, periodEnd]`.
+1. Keys may mix one-time and periodic token limits.
+2. Spending limits and call scopes are independent checks; both must pass.
+3. `updateSpendingLimit()` resets `remainingInPeriod` and updates periodic metadata.
```
horsefacts: I think we need to more clearly specify how this works:
- does it change `periodEnd`?
- what does it set `remainingInPeriod` to?

horsefacts: And it is not possible to change a key's period, right? Only its limit?

tips/tip-1011.md
horsefacts: don't we need an extra field to distinguish `None` from `Some([])` in the ABI?

horsefacts (changes requested):

tips/tip-1011.md
horsefacts: I think we should strongly discourage (maybe even ban) unscoped access keys, so the ergonomics should probably make this restricted by default.

PR #2438, opened 2026-02-04
tip: add TIP-1020 Signature Verification Precompile

legion2002:
## Summary

Adds TIP-1020: Signature Verification Precompile, which enables contracts to verify Tempo signature types (secp256k1, P256, WebAuthn, Keychain) onchain.

### New Precompile: `ISignatureVerification`

**Address:** `0x5165300000000000000000000000000000000000`

```solidity
interface ISignatureVerification {
    function verify(address signer, bytes32 hash, bytes calldata signature) external view returns (bool);
    error InvalidSignature();
    error SignerMismatch(address expected, address recovered);
    error UnauthorizedKeychainKey();
}
```

### Implementation Details

- Reuses existing `TempoSignature::recover_signer()` from `tempo_primitives`
- For Keychain signatures, reads AccountKeychain storage to validate access key authorization
- Uses same gas schedule as transaction signature verification:
  - secp256k1: 3,000 gas
  - P256: 8,000 gas
  - WebAuthn: 8,000 + calldata gas for webauthn_data
  - Keychain: inner signature gas + 3,000

### Hardfork Gating

- Precompile is only available after T2 hardfork

## Related

- Supersedes #2381 (TIP-XXXX draft)
- Related to payment channels spec: https://github.com/tempoxyz/payment-auth-spec/pull/92

PR #2633, opened 2026-02-11
tip: TIP-1020 Signature Verification Precompile

legion2002:
## Summary

Adds the TIP-1020 spec for a signature verification precompile at `0x5165300000000000000000000000000000000000`.

Enables contracts to verify Tempo signature types (secp256k1, P256, WebAuthn) onchain by reusing the existing audited verification logic from Tempo transaction processing.

Key design decisions:
- **Same gas schedule** as transaction signatures (3k secp256k1, 8k P256/WebAuthn)
- **Revert on invalid** rather than returning false (fail-fast)
- **No Keychain in precompile** — instead, the spec documents how contracts can verify Keychain signatures themselves by combining this precompile (inner signature verification) with the AccountKeychain precompile (access key validation)
- **No double-charge** for WebAuthn calldata gas (EVM already charges it)

Spec-only change — no code changes.

Implementation: https://github.com/tempoxyz/tempo/tree/tip/1020-impl

### DIFF
--- a/tips/tip-1020.md
+++ b/tips/tip-1020.md
@@ -110,6 +110,10 @@ This TIP does not modify `ecrecover` or any existing Ethereum precompile. `ecrec
 
 Solidity developers commonly use `ecrecover(hash, v, r, s)` to recover an address and then compare it to an expected signer. The Tempo signature verification precompile follows the same recover-and-return pattern but differs in one key way: `ecrecover` returns `address(0)` on invalid input or failed recovery, but the TIP-1020 precompile **reverts** on invalid signatures. Contracts that want non-reverting behavior SHOULD wrap calls using `try/catch` (high-level) or `staticcall` (low-level) and treat failure as "invalid signature".
 
+#### `v` Value Normalization
+
+For secp256k1 signatures, the precompile normalizes the recovery identifier `v`: both Ethereum-style values (`27`, `28`) and raw values (`0`, `1`) are accepted. This is intentional — `recover()` is designed to be as close to a drop-in replacement for `ecrecover` as possible, so it accepts the same `v` values. This differs from TIP-1004 (`permit()`), which requires `v ∈ {27, 28}` and reverts on `0` or `1`.
+
 #### Existing secp256k1 Signature Payloads
 
 For backwards compatibility, secp256k1 signatures are encoded as **65 bytes `r || s || v` with no type prefix**. Callers who already produce 65-byte secp256k1 signatures can reuse them directly as the `signature` argument to this precompile.


tips/tip-1020.md
```diff
@@ -0,0 +1,112 @@
...
+The precompile MUST use the same verification logic as Tempo transaction signature validation. See the [Tempo Transaction Signature Validation spec](https://docs.tempo.xyz/protocol/transactions/spec-tempo-transaction#signature-validation) for details.
+
+#### Keychain Signatures
+
+Keychain verification validates that an access key is authorized, not revoked, and not expired. It does NOT enforce spending limits because:
+
+- Spending limits require value/token amounts from transaction context
+- Limit enforcement requires state updates, inappropriate for a `view` function
+
+Contracts needing limit checks should query `getRemainingLimit()` on the AccountKeychain precompile separately.
```
howydev: copying from [this](https://github.com/tempoxyz/tempo/pull/2438) - how would the contract know what `keyId` to use?

tips/tip-1020.md
```diff
@@ -0,0 +1,112 @@
+---
+id: TIP-1020
+title: Signature Verification Precompile
+description: A precompile for verifying Tempo signatures onchain.
+authors: Jake Moxey (@jxom)
+status: Draft
+related: TIP-1003
```
jenpaff: How is TIP-1003 related?

legion2002: Don't think it is related. Was an issue in the original spec, will update.

tips/tip-1020.md
```diff
@@ -0,0 +1,174 @@
...
+### Gas Costs
+
+Gas costs follow the [Tempo Transaction Signature Gas Schedule](https://docs.tempo.xyz/protocol/transactions/spec-tempo-transaction#signature-verification-gas-schedule):
+
+| Type | Gas Cost |
+|------|----------|
+| secp256k1 | 3,000 |
+| P256 | 8,000 |
+| WebAuthn | 8,000 |
+| Keychain | inner signature cost + 2,600 (cold `CALL` to AccountKeychain) |
```
howydev: nit: precompiles are generally considered warm addresses with a cost of 100 even on first access, could we do that instead?

tips/tip-1020.md
```diff
@@ -0,0 +1,174 @@
...
+| secp256k1 | `r \|\| s \|\| v` | 65 bytes |
+| P256 | `0x01 \|\| r \|\| s \|\| x \|\| y \|\| prehash` | 130 bytes |
+| WebAuthn | `0x02 \|\| webauthn_data \|\| r \|\| s \|\| x \|\| y` | 129–2049 bytes |
+| Keychain | `0x03 \|\| user_address \|\| inner_signature` | variable |
+
+For backwards compatibility, secp256k1 signatures have no type prefix. All other signature types are prefixed with a type identifier byte.
+
+### Verification Logic
+
+The precompile MUST use the same verification logic as Tempo transaction signature validation. See the [Tempo Transaction Signature Validation spec](https://docs.tempo.xyz/protocol/transactions/spec-tempo-transaction#signature-validation) for details.
```
howydev, line 67: nit: don't do a docs link since that could break over time

tips/tip-1020.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Invariants
+
+| ID | Invariant | Description |
+|----|-----------|-------------|
+| **SV1** | Transaction-equivalent verification | For any signature type supported by a given function, the precompile MUST use the same cryptographic verification rules as Tempo transaction signature validation. |
+| **SV2** | P256 and ECDSA signature malleability resistance | P256 and ECDSA signatures MUST satisfy the low-s requirement (`s <= n/2`). Signatures with high-s values MUST be rejected. |
+| **SV3** | WebAuthn bounds enforcement | WebAuthn signature parsing MUST enforce the length bounds and structural validity checks used by Tempo transaction verification, preventing out-of-bounds reads and pathological resource usage. |
+| **SV4** | Revert on failure | On any invalid signature, invalid encoding, unsupported type, or signer mismatch, the precompile MUST revert. |
+| **SV5** | Gas schedule consistency | Gas charged MUST follow the Tempo Transaction Signature Gas Schedule. |
```
howydev:
```suggestion
| **SV5** | Gas schedule consistency | Gas charged MUST follow the gas schedule listed above. |
```

tips/tip-1020.md
```diff
@@ -0,0 +1,108 @@
...
+
+```solidity
+interface ISignatureVerifier {
+    error InvalidSignature();
+
+    /// @notice Verifies a Tempo signature (secp256k1, P256, WebAuthn).
+    /// @param signer The expected signer address
+    /// @param hash The message hash that was signed
+    /// @param signature The encoded signature (see Tempo Transaction spec for formats)
+    /// @return True if valid, reverts otherwise
```
samczsun: why not just no-op / revert or true/false?

tips/tip-1020.md
```diff
@@ -0,0 +1,108 @@
...
+
+The precompile MUST charge gas and verify sufficient gas is available **before** performing any cryptographic verification. The precompile MUST revert with out-of-gas if the call has insufficient gas for the signature type. Gas costs per signature type:
+
+| Type | Gas Cost |
+|------|----------|
+| secp256k1 | 3,000 |
+| P256 | 8,000 |
+| WebAuthn | 8,000 |
+
+**Note**: Gas costs are in-line with the [Tempo Transaction Signature Gas Schedule](https://docs.tempo.xyz/protocol/transactions/spec-tempo-transaction#signature-verification-gas-schedule). WebAuthn signatures do not incur additional calldata gas in the precompile since the caller pays for the additional size within the EVM via calldata or memory expansion costs.
```
samczsun: worth noting there's no calldata cost, and its not clear if the 2049 byte limit is enforced here

legion2002: Yeah no calldata cost here, because it is already charged by the EVM. We don't need to enforce a byte limit here for the sig, because the user is metered for all the execution/calldata costs and pays for it like standard EVM execution.

We needed to have this for the tempo tx, because that sig validation happened before tx validity checks, and could have become a DOS vector if the sig was invalid, the protocol wouldn't be paid for it.

tips/tip-1020.md
legion2002: I think true/false is a bit of a footgun sometimes, because people forget to do `true` checks all the time for such functions.

true / revert seems like the best of both worlds for devX, because users can hotswap ecrecover or signature checker libs with this function. 

But don't have a strong opinion on this.

tips/tip-1020.md
samczsun: no, I mean that the call opcode doesn't charge gas for bytes of input data. if I expand memory to 1MB of JSON and then call the precompile in a tight loop, it'll cost me ~2M gas for the memory expansion and then 8100*N gas for all the calls

tips/tip-1020.md
samczsun: mild preference towards consistency, i think the true/revert pattern is a carry-over from erc20 before everyone realized reverting just saved a lot of hassle

tips/tip-1020.md
howydev: Huh interesting, I thought I checked this exact case.. Even in the case of a call with offset + data going past fmp, that still incurs expansion memory cost?

samczsun: You would write this in assembly, something like

```
contract C {
    receive() external payable {
        assembly {
            mstore(0, selector)
            mstore(4, signer)
            mstore(36, hash)
            // prep rest of payload
            for {} 1 {} {
                pop(staticcall(8100, 0x5165300000000000000000000000000000000000, 0, 1000132, 0, 0))
            }
        }
    }
}
```

tips/tip-1020.md
```diff
@@ -0,0 +1,108 @@
...
+1. **Explicit signer parameter**: `ecrecover` returns a recovered address (or `address(0)`), and the caller compares it. TIP-1020 takes an `address signer` and returns success only if the recovered signer matches the provided `signer`.
+2. **Failure behavior (revert vs. sentinel)**: `ecrecover` returns `address(0)` on invalid input or failed recovery. TIP-1020 functions **revert** on invalid signatures and return `true` only on success. Contracts that want non-reverting behavior SHOULD wrap calls using `try/catch` (high-level) or `staticcall` (low-level) and treat failure as "invalid signature".
+
+#### Existing secp256k1 Signature Payloads
+
+For backwards compatibility, secp256k1 signatures are encoded as **65 bytes `r || s || v` with no type prefix**. Callers who already produce 65-byte secp256k1 signatures can reuse them directly as the `signature` argument to this precompile.
+
+### Forward Compatibility
+
+It is expected that this precompile will be updated when other account types are introduced to maintain forward compatibility with Tempo accounts.
```
horsefacts, line 123: what future types might we anticipate? are we comfortable constraining verification to this interface? for example, we implicitly assume future types always have a `bytes32 digest`.

I think probably fine, since we can always design such a format ourselves, but worth considering.

horsefacts (approved):

tips/tip-1020.md
howydev: `bytes32 digest` is the stuff thats signed over, should be fine to leave it

For the interface - within AA, a `context` field is sometimes required besides the signature for different sig modes. The pattern that was established is to pack that with signature and verifiers parse the (realSig, context). We can borrow this pattern - any additional fields thats required for an account type in the future can go into `bytes sig`. This is smth that I can see being mirrored at the tx level also

howydev: This q has stirred up my thinking in this direction (great q). It would be nice to make it non-view (e.g. one-time sigs) and this is the time to do it before everyone implements calls to this precompile with `staticcall`s. Doing this directly is unsafe though, because the precompile has no access control, an attacker can front run a call to the precompile to invalidate a sig

tips/tip-1020.md
howydev: No-ops in [7560](https://github.com/ethereum/RIPs/blob/master/RIPS/rip-7560.md) wasn't great for sc devs - since eoas would also return no op, when deploying crosschain verification contracts, verification flows had to check if the precompile was deployed

As long as theres a non-zero chance that another evm chain would implement this, we shouldn't do no op IMO

tips/tip-1020.md
howydev: We can make it safe + stateful by binding sig to a certain `caller` address/contract. Today we bind it via the 712 domain, but that's app-defined/app-enforced, once it hits the precompile its an opaque `digest`

howydev: a `verifyTypedData` would be interesting! Build the 712 digest in the precompile and additionally enforce that `domain.verifyingContract == msg.sender`

tips/tip-1020.md
```diff
@@ -0,0 +1,108 @@
...
+```solidity
+interface ISignatureVerifier {
+    error InvalidSignature();
+
+    /// @notice Verifies a Tempo signature (secp256k1, P256, WebAuthn).
+    /// @param signer The expected signer address
+    /// @param hash The message hash that was signed
+    /// @param signature The encoded signature (see Tempo Transaction spec for formats)
+    /// @return True if valid, reverts otherwise
+    function verify(address signer, bytes32 hash, bytes calldata signature) external view returns (bool);
```
0xrusowsky, line 51: wouldn't something like:
```solidity
/// Recover the signer address from a signature. Reverts if invalid.
function recover(bytes32 hash, bytes calldata signature) external view returns (address signer);
```
be useful for contracts that need to check vs multiple potential signers?

it would be the equivalent of ecrecover but with support for P256 or webauthn

howydev: good call

e.g. for gnosis safe, they do this: https://github.com/safe-fndn/safe-smart-account/blob/main/contracts/Safe.sol#L403

tips/tip-1020.md
howydev: Making a decision to punt on this thread for 2 reasons:
1. One time sigs don't seem to be widely used in prod, and we'll likely not want to use a stateful sig if possible
2. Adding `verifyTypedData` in the future to the precompile is always an option

howydev (approved):

tips/tip-1020.md
```diff
@@ -0,0 +1,116 @@
...
+
+Signatures MUST be encoded using the same format as [Tempo Transaction signatures](https://docs.tempo.xyz/protocol/transactions/spec-tempo-transaction#signature-types):
+
+| Type | Format | Length |
+|------|--------|--------|
+| secp256k1 | `r \|\| s \|\| v` | 65 bytes |
+| P256 | `0x01 \|\| r \|\| s \|\| x \|\| y \|\| prehash` | 130 bytes |
+| WebAuthn | `0x02 \|\| webauthn_data \|\| r \|\| s \|\| x \|\| y` | 129–2049 bytes |
+
+### Verification Logic
```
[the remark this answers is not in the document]
0xrusowsky: > Add a normative statement that the precompile MUST reject Keychain signatures (prefix 0x03 and 0x04) at the first byte before any parsing.

implicit from the table, but better to explicitly add it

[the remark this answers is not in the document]
0xrusowsky: addressed in https://github.com/tempoxyz/tempo/pull/3306

tips/tip-1020.md
0xrusowsky: addressed in https://github.com/tempoxyz/tempo/pull/3306

tips/tip-1020.md
```diff
@@ -0,0 +1,131 @@
+---
+id: TIP-1020
+title: Signature Verification Precompile
+description: A precompile for verifying Tempo signatures onchain.
+authors: Jake Moxey (@jxom), Tanishk Goyal (@legion2002), Howy (@howydev)
+status: Draft
```
jenpaff:
```suggestion
status: Approved
```

PR #2635, opened 2026-02-11
feat(precompiles): add ValidatorConfigV2 precompile

howydev:
Index-canonical storage model with Mapping<u64, ValidatorV2> as source of truth, plus address_to_index and pubkey_to_index lookup mappings (1-indexed). Gated behind T2 hardfork — pre-T2 calls succeed with empty output (empty contract behavior).

Implements all IValidatorConfigV2 functions:
- addValidator (owner only)
- deactivateValidator (owner or validator)
- rotateValidator (owner or validator) — deactivates old entry, appends new with same address
- setIpAddresses, transferValidatorOwnership (owner or validator)
- transferOwnership, setNextFullDkgCeremony (owner only)
- migrateValidator, initializeIfMigrated (owner only)

23 tests covering dispatch, hardfork gating, auth, storage invariants.

Amp-Thread-ID: https://ampcode.com/threads/T-019c4944-6b37-71fc-bce1-22ca8de69d61


crates/precompiles/src/storage/mod.rs
```diff
@@ -38,6 +38,11 @@ pub trait PrecompileStorageProvider {
     /// Returns the current block beneficiary (coinbase).
     fn beneficiary(&self) -> Address;
 
+    /// Returns the current block number.
+    fn block_number(&self) -> u64 {
+        0
+    }
```
0xrusowsky: let's not add a default impl and make implementors do it themselves
```suggestion
    /// Returns the current block number.
    fn block_number(&self) -> u64;
```

crates/precompiles/src/validator_config_v2/dispatch.rs
```diff
@@ -0,0 +1,316 @@
...
+        // Pre-T2: behave like an empty contract (call succeeds, no execution)
+        if !self.storage.spec().is_t2() {
+            return Ok(PrecompileOutput::new(
+                self.storage.gas_used(),
+                Default::default(),
+            ));
+        }
+
+        // Get the current block number from the EVM context via storage provider
+        let block_height = self.storage.block_number();
```
0xrusowsky: since `block_height` is only used in a handful of fns, and the block number is always accessible, let's not cache it here, and do it inside the relevant fns instead

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1493 @@
...
+    fn check_initialized(&self) -> Result<()> {
+        if !self.initialized.read()? {
+            return Err(ValidatorConfigV2Error::not_initialized())?;
+        }
+        Ok(())
+    }
+
+    pub fn is_initialized(&self) -> Result<bool> {
+        self.initialized.read()
+    }
```
0xrusowsky: this fn signature collides with `trait ContractStorage`'s default impl for `fn is_initialized()`.

to prevent footguns, it is worth writing an explicit trait override to align them.

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1493 @@
...
+#[contract(addr = VALIDATOR_CONFIG_V2_ADDRESS)]
+pub struct ValidatorConfigV2 {
+    owner: Address,
+    validators: Vec<ValidatorV2>,
+    address_to_index: Mapping<Address, u64>,
+    pubkey_to_index: Mapping<B256, u64>,
+    next_dkg_ceremony: u64,
+    initialized: bool,
+    initialized_at_height: u64,
+}
```
0xrusowsky, line 70: we can do several optimizations to improve performance here by using a smart layout and slot packing

- we can drop `initialized` as `initialized = !initialized_at_height.is_zero()`
- most mutable fns check that the contract is initialized + the owner, thus we can save 1 SLOAD if we pack them

proposal:
```rs
pub struct ValidatorConfigV2 {
    config: Config,
    validators: Vec<ValidatorV2>,
    next_dkg_ceremony: u64,
}

pub struct Config {
   owner: Address,             // 20 bytes
   initialized_at_height: u64  // 8 bytes
}
```

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1493 @@
...
+            return Err(ValidatorConfigV2Error::validator_already_deleted())?;
+        }
+        Ok((idx, v))
+    }
+
+    fn read_validator_at(&self, index: u64) -> Result<IValidatorConfigV2::Validator> {
+        // Check bounds first
+        if index >= self.validator_count()? {
+            return Err(ValidatorConfigV2Error::validator_not_found())?;
+        }
```
0xrusowsky: `read_validator_at` is an internal fn, guaranteed to be called with a correct idx, which means that every call to it does an extra SLOAD.

let's drop this check, and instead add a debug assertion for peace of mind (although unnecessary by design):
```suggestion
        debug_assert!(index < self.validator_count()?, "index OOB");
```

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1493 @@
...
+            addedAtHeight: v.added_at_height,
+            deactivatedAtHeight: v.deactivated_at_height,
+        })
+    }
+
+    pub fn validator_by_index(&self, index: U256) -> Result<IValidatorConfigV2::Validator> {
+        // Check if index fits in u64
+        if index > U256::from(u64::MAX) {
+            return Err(ValidatorConfigV2Error::validator_not_found())?;
+        }
```
0xrusowsky: let's simply enforce the type in the fn sig, no reason to use `U256` if index is guaranteed to be a `u64`

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1493 @@
...
+
+    fn validate_egress(egress: &str) -> Result<()> {
+        ensure_address_is_ip(egress).map_err(|err| {
+            TempoPrecompileError::from(ValidatorConfigV2Error::not_ip_port(
+                "egress".to_string(),
+                egress.to_string(),
+                format!("{err:?}"),
+            ))
+        })
+    }
```
0xrusowsky: refactor, they do the exact same

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1493 @@
...
+
+    fn validate_rotate_params(&self, pubkey: B256) -> Result<()> {
+        if pubkey.is_zero() {
+            return Err(ValidatorConfigV2Error::invalid_public_key())?;
+        }
+        if self.pubkey_to_index[pubkey].read()? != 0 {
+            return Err(ValidatorConfigV2Error::public_key_already_exists())?;
+        }
+        Ok(())
+    }
```
0xrusowsky: easier to use individual validation fns

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1493 @@
...
+
+        // Ensure validators are migrated in order (idx must equal current count)
+        let current_count = self.validator_count()?;
+        if call.idx != current_count {
+            return Err(ValidatorConfigV2Error::invalid_migration_index())?;
+        }
+
+        // Get V1 validators
+        let v1 = ValidatorConfig::new();
+        let v1_validators = v1.get_validators()?;
```
0xrusowsky: we shouldn't read all validators, that view fn is not designed to be called on-chain

onbjerg (changes requested):
pls split out the storage provider changes into a standalone pr we can sync into foundry so we don't have to bump continually in tempo-foundry as this pr changes

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1493 @@
...
+        let public_key = PublicKey::decode(pubkey.as_slice())
+            .map_err(|_| ValidatorConfigV2Error::invalid_public_key())?;
+
+        // Decode the signature from bytes
+        let sig = Signature::decode(signature)
+            .map_err(|_| ValidatorConfigV2Error::invalid_signature())?;
+
+        // Verify the signature
+        // namespace is empty for this use case
+        if !public_key.verify(&[], message, &sig) {
```
SuperFluffy: Noted above: we should break the namespace out of the keccak because that's how the verifier functions will be used.

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1493 @@
...
+    ///
+    /// The signature verification uses the commonware-cryptography Ed25519 implementation.
+    fn verify_ed25519_signature(pubkey: &B256, message: &[u8], signature: &[u8]) -> Result<()> {
+        // Decode the public key from bytes
+        let public_key = PublicKey::decode(pubkey.as_slice())
+            .map_err(|_| ValidatorConfigV2Error::invalid_public_key())?;
+
+        // Decode the signature from bytes
+        let sig = Signature::decode(signature)
+            .map_err(|_| ValidatorConfigV2Error::invalid_signature())?;
```
SuperFluffy: Differentiate between incorrectly formatted signatures and signatures that could not be validated.

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1493 @@
+pub mod dispatch;
+
+use tempo_contracts::precompiles::VALIDATOR_CONFIG_V2_ADDRESS;
+pub use tempo_contracts::precompiles::{IValidatorConfigV2, ValidatorConfigV2Error};
+use tempo_precompiles_macros::{Storable, contract};
+
+use crate::{
+    error::{Result, TempoPrecompileError},
+    storage::{Handler, Mapping},
+    validator_config::{ValidatorConfig, ensure_address_is_ip_port},
```
SuperFluffy: Move `ensure_address_is_ip_port` to a common location.

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1493 @@
...
+        &self,
+        is_add: bool,
+        validator_address: Address,
+        ingress: &str,
+        egress: &str,
+    ) -> B256 {
+        let mut data = Vec::new();
+        data.extend_from_slice(b"TEMPO");
+        data.extend_from_slice(if is_add {
+            b"_VALIDATOR_CONFIG_V2_ADD_VALIDATOR"
```
SuperFluffy: Yeah, this is according to spec but unfortunate: we should move the namespace outside the keccak-payload so that the verification happens as `key.verify(<namespace>, <keccak-encoded-payload>)`.

This requires a change to the spec too.

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1493 @@
...
+        } else {
+            b"_VALIDATOR_CONFIG_V2_ROTATE_VALIDATOR"
+        });
+
+        // Add chain ID
+        data.extend_from_slice(&self.storage.chain_id().to_be_bytes());
```
SuperFluffy: Should specify in the spec that the chain ID is hashed as big endian bytes.

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1493 @@
...
+    ) -> Result<()> {
+        self.check_initialized()?;
+        self.check_owner(sender)?;
+        self.validate_add_params(call.validatorAddress, call.publicKey)?;
+        Self::validate_ingress(&call.ingress)?;
+        Self::validate_egress(&call.egress)?;
+
+        // Construct message for signature verification
+        let message = self.construct_validator_message(
+            true, // is_add
```
SuperFluffy: Please make that explicit not via flags.

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1493 @@
...
+        )
+    }
+
+    pub fn deactivate_validator(
+        &mut self,
+        sender: Address,
+        call: IValidatorConfigV2::deactivateValidatorCall,
+        block_height: u64,
+    ) -> Result<()> {
+        self.check_initialized()?;
```
SuperFluffy: This *cannot* check if the contract is initialized: if for whatever reason a validator was migrated from V1 to V2 but gets deactivated before `initializeIfMigrated` is called, then this check would break CL.

We have to be able to decativate validators before migration is complete.

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1493 @@
...
+        v.egress = call.egress;
+        self.validators[idx].write(v)
+    }
+
+    pub fn transfer_validator_ownership(
+        &mut self,
+        sender: Address,
+        call: IValidatorConfigV2::transferValidatorOwnershipCall,
+    ) -> Result<()> {
+        self.check_initialized()?;
```
SuperFluffy: This should work without this check.

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1493 @@
...
+                "ingress".to_string(),
+                ingress.to_string(),
+                format!("{err:?}"),
+            ))
+        })
+    }
+
+    fn validate_egress(egress: &str) -> Result<()> {
+        ensure_address_is_ip(egress).map_err(|err| {
+            TempoPrecompileError::from(ValidatorConfigV2Error::not_ip_port(
```
SuperFluffy: This error should be `not_ip`.

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1493 @@
...
+    }
+
+    fn validate_add_params(&self, addr: Address, pubkey: B256) -> Result<()> {
+        if addr.is_zero() {
+            return Err(ValidatorConfigV2Error::invalid_validator_address())?;
+        }
+        if pubkey.is_zero() {
+            return Err(ValidatorConfigV2Error::invalid_public_key())?;
+        }
+        if self.address_to_index[addr].read()? != 0 {
```
SuperFluffy: We don't want to burn addresses forever. The only invariant we want to maintain is that a single `address` has at most 1 validator with `active = true`.

If all entries with `address` have `active = false`, then it is ok to add a new validator.

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1493 @@
...
+        addr: Address,
+        pubkey: B256,
+        ingress: String,
+        egress: String,
+        block_height: u64,
+    ) -> Result<()> {
+        self.append_validator_raw(addr, pubkey, ingress, egress, block_height, 0)
+    }
+
+    fn validate_add_params(&self, addr: Address, pubkey: B256) -> Result<()> {
```
SuperFluffy: What about ingress and egress? We don't want this to be clashing with any other currently active validators.

The more important of the 2 is certainly `ingress` - if `append_validator` (or `rotate_validator`) sets this to the same value as a currently active other validator, then we'd end up with 2 nodes having the same destination.

howydev: makes sense, this would require a spec change

crates/precompiles/src/validator_config_v2/mod.rs
howydev: ok. this would require a spec change

crates/precompiles/src/validator_config_v2/mod.rs
howydev: this is dangerous, since we deactivate the existing entry and create a new one, and migration/initialized checks happens based on indexes + val config lengths

howydev: it should be fine to finalize migration, then perform rotation

crates/precompiles/src/validator_config_v2/mod.rs
howydev: changed to allow deactivation from either the validator or owner. in practice would just be the owner since no one owns the val addresses

we can assume owner migration happened already, since to deactivate a validator one would need to be migrated over

crates/precompiles/src/validator_config_v2/dispatch.rs
```diff
@@ -0,0 +1,302 @@
...
+use revm::precompile::{PrecompileError, PrecompileOutput, PrecompileResult};
+use tempo_contracts::precompiles::IValidatorConfigV2::IValidatorConfigV2Calls;
+
+impl Precompile for ValidatorConfigV2 {
+    fn call(&mut self, calldata: &[u8], msg_sender: Address) -> PrecompileResult {
+        self.storage
+            .deduct_gas(input_cost(calldata.len()))
+            .map_err(|_| PrecompileError::OutOfGas)?;
+
+        // Pre-T2: behave like an empty contract (call succeeds, no execution)
```
howydev, line 13:
```suggestion
```

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1929 @@
...
+
+        let egress_hash = keccak256(egress.as_bytes());
+        if self.active_egress[egress_hash].read()? {
+            Err(ValidatorConfigV2Error::egress_already_exists(
+                egress.to_string(),
+            ))?
+        }
+
+        Ok(())
+    }
```
0xrusowsky: hashing is expensive.

let's turn this fn into `require_unique_ips(&self, ingress: &str, egress: &str) -> Result<(B256, B256)>;` so that we don't have to compute hashes twice.

0xrusowsky (changes requested):
just one pending improvement from my side, then g2g

tips/tip-1017.mdx
```diff
@@ -317,38 +324,26 @@ Unlike V1, validators in V2 follow a strict lifecycle:
 
 ### Signature Verification
 
-When adding a validator, the caller must provide an Ed25519 signature proving
-ownership of the public key:
+When adding a validator, the caller must provide an Ed25519 signature proving ownership of the public key. The
+signature is checked over the a full message containing: the length of the namespace in bytes, the namespace, and a
```
SuperFluffy, line 324: > the length of the namespace in bytes

Is that correct?

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1929 @@
...
+}
+
+impl ValidatorOperation {
+    fn namespace(&self) -> &'static [u8] {
+        match self {
+            Self::Add => b"TEMPO_VALIDATOR_CONFIG_V2_ADD_VALIDATOR",
+            Self::Rotate => b"TEMPO_VALIDATOR_CONFIG_V2_ROTATE_VALIDATOR",
+        }
+    }
+}
```
SuperFluffy: This enum seems a bit heavy-handed. Why not just?

```rust
const ADD_NS: &[u8] = b"TEMPO_VALIDATOR_CONFIG_V2_ADD_VALIDATOR";
const ROTATE_NS: &[u8] = b"TEMPO_VALIDATOR_CONFIG_V2_ROTATE_VALIDATOR";
```

Also, let's make these `pub` so I can access them from elsewhere (to generate these signatures, for example).

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1929 @@
...
+
+        ensure_address_is_ip(egress).map_err(|err| {
+            TempoPrecompileError::from(ValidatorConfigV2Error::not_ip(
+                egress.to_string(),
+                format!("{err:?}"),
+            ))
+        })
+    }
+
+    fn require_unique_ips(&self, ingress: &str, egress: &str) -> Result<()> {
```
SuperFluffy:
```suggestion
    fn require_unique_active_ips(&self, ingress: &str, egress: &str) -> Result<()> {
```

crates/precompiles/src/validator_config_v2/mod.rs
0xrusowsky: addressed in [`9fa049f`](https://github.com/tempoxyz/tempo/pull/2635/commits/9fa049f7a26227e34a4f462e9741104c17b7a18a)

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1929 @@
...
+    ///
+    /// Constructs the message according to the validator config v2 specification
+    /// and verifies the Ed25519 signature using the appropriate namespace.
+    ///
+    /// **FORMAT**:
+    /// - Namespace: `b"TEMPO_VALIDATOR_CONFIG_V2_ADD_VALIDATOR"` or `b"TEMPO_VALIDATOR_CONFIG_V2_ROTATE_VALIDATOR"`
+    /// - Message: `keccak256(abi.encodePacked(chainId, contractAddr, validatorAddr, ingress, egress))`
+    fn verify_validator_signature(
+        &self,
+        operation: ValidatorOperation,
```
SuperFluffy: I am personally fine if this just takes a `namespace: &[u8]` and is passed in by the caller.

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1929 @@
...
+        pubkey: &B256,
+        signature: &[u8],
+        validator_address: Address,
+        ingress: &str,
+        egress: &str,
+    ) -> Result<()> {
+        let namespace = operation.namespace();
+
+        let mut data = Vec::new();
```
SuperFluffy: Nit: extending this vector a bunch of times is unnecessary if you use `alloy::primitives::Keccak256` directly:

```rust
let mut hasher = Keccak256::new();
hasher.update(VALIDATOR_CONFIG_V2_ADDRESS.as_slice());
hasher.update(validator_address.as_slice());
hasher.update(ingress.as_bytes());
hasher.update(egress.as_bytes());
let message = hasher.finalize()
```

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1929 @@
...
+        &mut self,
+        sender: Address,
+        call: IValidatorConfigV2::deactivateValidatorCall,
+    ) -> Result<()> {
+        if sender != call.validatorAddress && !self.config.read()?.is_owner(sender) {
+            Err(ValidatorConfigV2Error::unauthorized())?
+        }
+        let block_height = self.storage.block_number();
+
+        let (idx, mut v) = self.get_active_validator(call.validatorAddress)?;
```
SuperFluffy, line 426: I see, if `deactivated_at_height > 0` this will reject the call.

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1929 @@
...
+        &mut self,
+        sender: Address,
+        call: IValidatorConfigV2::rotateValidatorCall,
+    ) -> Result<()> {
+        self.require_initialized_owner_or_validator(sender, call.validatorAddress)?;
+        self.require_new_pubkey(call.publicKey)?;
+        Self::validate_endpoints(&call.ingress, &call.egress)?;
+
+        self.verify_validator_signature(
+            ValidatorOperation::Rotate,
```
SuperFluffy: noted elsewhere: fine if this passes in a const.

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1929 @@
...
+        call: IValidatorConfigV2::setIpAddressesCall,
+    ) -> Result<()> {
+        if sender != call.validatorAddress && !self.config.read()?.is_owner(sender) {
+            Err(ValidatorConfigV2Error::unauthorized())?
+        }
+
+        let (idx, mut v) = self.get_active_validator(call.validatorAddress)?;
+        Self::validate_endpoints(&call.ingress, &call.egress)?;
+
+        self.update_active_ips(&v.ingress, &v.egress, &call.ingress, &call.egress)?;
```
SuperFluffy: Is this going to conflict with itself? If I want to change only my egress but keep my ingress the same, that should work.

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1929 @@
...
+        let v1 = v1();
+
+        if self.validator_count()? < v1.validator_count()? {
+            Err(ValidatorConfigV2Error::migration_not_complete())?
+        }
+
+        let v1_next_dkg = v1.get_next_full_dkg_ceremony()?;
+        self.next_dkg_ceremony.write(v1_next_dkg)?;
+
+        config.init_at_height = block_height.max(1);
```
SuperFluffy: Why would the maximum be `1` and not `self.storage.block_number()`?

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1929 @@
...
+    pub fn owner(&self) -> Result<Address> {
+        self.config.owner.read()
+    }
+
+    pub fn get_initialized_at_height(&self) -> Result<u64> {
+        self.config.init_at_height.read()
+    }
+
+    pub fn is_initialized(&self) -> Result<bool> {
+        self.config.read().map(|c| c.is_initialized())
```
SuperFluffy: This is not sufficient. As per the spec, there must be an `init` flag because if this contract is set at genesis then `initializedAtHeight = 0` and `initialized = true`.

SuperFluffy (changes requested):
I believe I found two issues:

1. updating IPs can conflict with itself (need tests for that)
2. there must be an `init` flag, not just `initializedAtHeight`.

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1927 @@
...
+        self.config.write(config)
+    }
+}
+
+fn v1() -> ValidatorConfig {
+    ValidatorConfig::new()
+}
+
+#[cfg(test)]
+mod tests {
```
SuperFluffy, line 641: Can we move this to its own module?

crates/precompiles/src/validator_config_v2/mod.rs
0xrusowsky: that's inaccurate, but i'll put the init flag back in for clarity

crates/precompiles/src/validator_config_v2/mod.rs
0xrusowsky: that's the opposite, it ensures that at least it is 1

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1927 @@
...
+
+        Ok(())
+    }
+
+    pub fn initialize_if_migrated(&mut self, sender: Address) -> Result<()> {
+        let mut config = self.require_migration_owner(sender)?;
+        let block_height = self.storage.block_number();
+        let v1 = v1();
+
+        if self.validator_count()? < v1.validator_count()? {
```
SuperFluffy, line 620: Make a note here: this check only works because `fn add_validator` and `fn rotate_validator` are blocked unless the contract is initialized.

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1927 @@
...
+
+            let v3 = vc.validator_by_public_key(pubkey)?;
+            assert_eq!(v3.validatorAddress, validator);
+
+            Ok(())
+        })
+    }
+
+    #[test]
+    fn test_add_validator_rejects_unauthorized() -> eyre::Result<()> {
```
SuperFluffy, line 775: This checks if `sender != owner`. We should also have a test that rejects `non_owner` if it is neither `owner` nor `validator_owner` for an actual entry.

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1927 @@
...
+    validators: Vec<ValidatorV2>,
+    address_to_index: Mapping<Address, u64>,
+    pubkey_to_index: Mapping<B256, u64>,
+    next_dkg_ceremony: u64,
+    active_ingress: Mapping<B256, bool>,
+    active_egress: Mapping<B256, bool>,
+}
+
+impl ValidatorConfigV2 {
+    pub fn initialize(&mut self, owner: Address) -> Result<()> {
```
SuperFluffy: IMO rename because migration from V1 is so critical.

```suggestion
    pub fn initialize_without_migration(&mut self, owner: Address) -> Result<()> {
```

or
```suggestion
    pub fn initialize_empty(&mut self, owner: Address) -> Result<()> {
```

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1927 @@
...
+            )?;
+
+            vc.storage.set_block_number(201);
+            let result = vc.add_validator(
+                owner,
+                make_valid_add_call(validator, "192.168.1.2:8000", "192.168.1.2"),
+            );
+            assert_eq!(
+                result,
+                Err(ValidatorConfigV2Error::validator_already_exists().into())
```
SuperFluffy: IMO confusing error, since it's not clear if this is about the specific validator entry or the owner. Maybe `ValidatorConfigV2Error::address_already_has_active_validataor`?

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1927 @@
...
+                owner,
+                IValidatorConfigV2::deactivateValidatorCall {
+                    validatorAddress: validator,
+                },
+            )?;
+
+            let v = vc.validator_by_index(0)?;
+            assert_eq!(v.deactivatedAtHeight, 300);
+
+            // Double deactivation fails
```
SuperFluffy, line 919: nice

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1927 @@
...
+                },
+            )?;
+            assert_eq!(vc.validator_by_index(1)?.deactivatedAtHeight, 301);
+
+            Ok(())
+        })
+    }
+
+    #[test]
+    fn test_rotate_validator() -> eyre::Result<()> {
```
SuperFluffy, line 991: We should have a test that rotating to a new validator with the same ingress/egress as the old validator works.

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1927 @@
...
+            // Old pubkey still resolves to the old entry
+            let by_old_pk = vc.validator_by_public_key(old_pubkey)?;
+            assert_eq!(by_old_pk.deactivatedAtHeight, 300);
+
+            Ok(())
+        })
+    }
+
+    #[test]
+    fn test_get_active_validators() -> eyre::Result<()> {
```
SuperFluffy, line 1065: Test for `get_all_validators` too

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1927 @@
...
+            assert_eq!(active[0].validatorAddress, v2);
+
+            assert_eq!(vc.get_validators()?.len(), 2);
+
+            Ok(())
+        })
+    }
+
+    #[test]
+    fn test_set_ip_addresses() -> eyre::Result<()> {
```
SuperFluffy, line 1106: Mentioned elsewhere: need a test that a validator can update egress-only and ingress-only (while keeping the other part the same).

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1927 @@
...
+                },
+            );
+            assert_eq!(result, Err(ValidatorConfigV2Error::unauthorized().into()));
+
+            Ok(())
+        })
+    }
+
+    #[test]
+    fn test_transfer_validator_ownership() -> eyre::Result<()> {
```
SuperFluffy, line 1180: Test that a validator can transfer its own ownership.

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1927 @@
...
+                result,
+                Err(ValidatorConfigV2Error::validator_already_exists().into())
+            );
+
+            Ok(())
+        })
+    }
+
+    #[test]
+    fn test_add_validator_rejects_duplicate_pubkey() -> eyre::Result<()> {
```
SuperFluffy, line 850: Add a test that also covers duplicate-pubkey rejection for already deleted validators.

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,1927 @@
...
+                result,
+                Err(ValidatorConfigV2Error::not_initialized().into())
+            );
+
+            Ok(())
+        })
+    }
+
+    #[test]
+    fn test_egress_validates_ip_without_port() -> eyre::Result<()> {
```
SuperFluffy, line 1314: Same test but for ingress?

crates/precompiles/src/validator_config_v2/mod.rs
SuperFluffy: It was in the spec 3 days ago, now it's gone. :(

crates/precompiles/src/validator_config_v2/mod.rs
SuperFluffy: Indeed, my bad

crates/precompiles/src/validator_config_v2/mod.rs
0xrusowsky: addressed with [`9fa049f`](https://github.com/tempoxyz/tempo/pull/2635/commits/9fa049f7a26227e34a4f462e9741104c17b7a18a)

tips/tip-1017.mdx
0xrusowsky: confirmed — it is correct. added ref to commonware in [`5aa3f5e`](https://github.com/tempoxyz/tempo/pull/2635/commits/5aa3f5e2bff73b4cc4d40ac65a51ff810ffd4fe6)

crates/precompiles/src/validator_config_v2/mod.rs
0xrusowsky: addressed with [`dfc3673`](https://github.com/tempoxyz/tempo/pull/2635/commits/dfc3673934d0582c11fb0efdbce4fe13f88b7552)

crates/precompiles/src/validator_config_v2/mod.rs
0xrusowsky: addressed with [`dfc3673`](https://github.com/tempoxyz/tempo/pull/2635/commits/dfc3673934d0582c11fb0efdbce4fe13f88b7552)

crates/precompiles/src/validator_config_v2/mod.rs
0xrusowsky: addressed with [`dfc3673`](https://github.com/tempoxyz/tempo/pull/2635/commits/dfc3673934d0582c11fb0efdbce4fe13f88b7552)

crates/precompiles/src/validator_config_v2/mod.rs
0xrusowsky: addressed with [`dfc3673`](https://github.com/tempoxyz/tempo/pull/2635/commits/dfc3673934d0582c11fb0efdbce4fe13f88b7552)

crates/precompiles/src/validator_config_v2/mod.rs
0xrusowsky: 👍

crates/precompiles/src/validator_config_v2/mod.rs
0xrusowsky: addressed with [`dfc3673`](https://github.com/tempoxyz/tempo/pull/2635/commits/dfc3673934d0582c11fb0efdbce4fe13f88b7552)

crates/precompiles/src/validator_config_v2/mod.rs
0xrusowsky: no — `update_active_ips` skips the check when `old == new`. added tests in [`3b33e9e`](https://github.com/tempoxyz/tempo/pull/2635/commits/3b33e9eb56756a468af299fc18b9e3fc3f52e2ae)

crates/precompiles/src/validator_config_v2/mod.rs
0xrusowsky: will do in a follow-up

crates/precompiles/src/validator_config_v2/mod.rs
0xrusowsky: addressed with [`dfc3673`](https://github.com/tempoxyz/tempo/pull/2635/commits/dfc3673934d0582c11fb0efdbce4fe13f88b7552)

crates/precompiles/src/validator_config_v2/mod.rs
0xrusowsky: addressed with [`3b33e9e`](https://github.com/tempoxyz/tempo/pull/2635/commits/3b33e9eb56756a468af299fc18b9e3fc3f52e2ae)

crates/precompiles/src/validator_config_v2/mod.rs
0xrusowsky: addressed with [`0649126`](https://github.com/tempoxyz/tempo/pull/2635/commits/0649126712b4e7c42c7f44df7308b7f18818c87b)

crates/precompiles/src/validator_config_v2/mod.rs
0xrusowsky: addressed with [`dfc3673`](https://github.com/tempoxyz/tempo/pull/2635/commits/dfc3673934d0582c11fb0efdbce4fe13f88b7552)

crates/precompiles/src/validator_config_v2/mod.rs
0xrusowsky: 🫡

crates/precompiles/src/validator_config_v2/mod.rs
0xrusowsky: addressed with [`3b33e9e`](https://github.com/tempoxyz/tempo/pull/2635/commits/3b33e9eb56756a468af299fc18b9e3fc3f52e2ae)

crates/precompiles/src/validator_config_v2/mod.rs
0xrusowsky: already covered in `test_get_active_validators` — it calls both `get_active_validators()` and `get_validators()` with assertions

crates/precompiles/src/validator_config_v2/mod.rs
0xrusowsky: addressed with [`3b33e9e`](https://github.com/tempoxyz/tempo/pull/2635/commits/3b33e9eb56756a468af299fc18b9e3fc3f52e2ae)

crates/precompiles/src/validator_config_v2/mod.rs
0xrusowsky: addressed with [`3b33e9e`](https://github.com/tempoxyz/tempo/pull/2635/commits/3b33e9eb56756a468af299fc18b9e3fc3f52e2ae)

crates/precompiles/src/validator_config_v2/mod.rs
0xrusowsky: addressed with [`3b33e9e`](https://github.com/tempoxyz/tempo/pull/2635/commits/3b33e9eb56756a468af299fc18b9e3fc3f52e2ae)

crates/precompiles/src/validator_config_v2/mod.rs
0xrusowsky: already exists — `test_add_validator_rejects_duplicate_ingress`

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -240,54 +239,34 @@ impl ValidatorConfigV2 {
         })
     }
 
-    fn require_unique_active_ips(&self, ingress: &str, egress: &str) -> Result<(B256, B256)> {
-        let ingress_hash = keccak256(ingress.as_bytes());
-        if self.active_ingress[ingress_hash].read()? {
+    fn ingress_ip_hash(ingress: &str) -> B256 {
+        let ingress_ip = ingress.split(':').next().unwrap_or(ingress);
```
SuperFluffy: This does not work: IPv6 IPs can contain `:`

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -0,0 +1,2187 @@
...
+        ensure_address_is_ip(egress).map_err(|err| {
+            TempoPrecompileError::from(ValidatorConfigV2Error::not_ip(
+                egress.to_string(),
+                format!("{err:?}"),
+            ))
+        })
+    }
+
+    fn ingress_ip_hash(ingress: &str) -> B256 {
+        let ingress_ip = ingress.split(':').next().unwrap_or(ingress);
```
SuperFluffy: This is fundamentally broken because ipv6 uses colons as segment separators, not to speak of ipv6 shorthands like `[::1]:8000` for localhost with port 8000.

IMO don't hash the strings, but the binary IP representation for both (i.e. 8 and 16 bytes, respctively). For Rust, this would be for ipv4 and ipv6, respectively:

https://doc.rust-lang.org/beta/std/net/struct.Ipv4Addr.html#method.octets
https://doc.rust-lang.org/beta/std/net/struct.Ipv6Addr.html#method.octets

crates/precompiles/src/ip_validation.rs
```diff
@@ -0,0 +1,109 @@
...
+    input.parse::<std::net::IpAddr>()?;
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_ensure_address_is_ip_port_rejects_invalid() {
```
SuperFluffy: What about ipv6?

crates/precompiles/src/ip_validation.rs
```diff
@@ -0,0 +1,109 @@
...
+    source: std::net::AddrParseError,
+}
+
+pub(crate) fn ensure_address_is_ip(input: &str) -> core::result::Result<(), IpParseError> {
+    input.parse::<std::net::IpAddr>()?;
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
```
SuperFluffy: All of these are fundamentally just testing if the Rust standard library correctly parses IP and/or socket addresses.

This is IMO not necessary if you parsed the input as `IpAddr` and/or `SocketAddr` and then either wrote their binary or `std::fmt::Display` formatted representation into the contract.

SuperFluffy (changes requested):
The ingress parsing is broken for ipv6.

crates/precompiles/src/validator_config_v2/mod.rs
howydev: good call, opted to use this

just in case, i added a prefix to each (to differentiate between 8 byte v4 key vs v6 key that starts with 8 0's)

crates/precompiles/src/ip_validation.rs
howydev: added

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -239,13 +239,28 @@ impl ValidatorConfigV2 {
...
+                        bytes[1..5].copy_from_slice(&ip.octets());
+                    }
+                    std::net::IpAddr::V6(ip) => {
+                        bytes[0] = 6; // version discriminator
+                        bytes[1..17].copy_from_slice(&ip.octets());
+                    }
+                }
+                B256::from(bytes)
+            })
+            .unwrap_or_else(|_| keccak256(ingress.as_bytes()))
```
SuperFluffy: So if parsing fails this still ends up as being parsed? IMO if parsing here fails, then this should return a result (or panic, depending on the context).

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -239,13 +239,28 @@ impl ValidatorConfigV2 {
...
     }
 
-    fn ingress_ip_hash(ingress: &str) -> B256 {
-        let ingress_ip = ingress.split(':').next().unwrap_or(ingress);
-        keccak256(ingress_ip.as_bytes())
+    fn ingress_ip_key(ingress: &str) -> B256 {
+        ingress
+            .parse::<std::net::SocketAddr>()
+            .map(|sa| {
+                let mut bytes = [0u8; 32];
```
SuperFluffy: Why don't you just `keccak256` the the octet arrays?

SuperFluffy: Done in [a0bc2fe](https://github.com/tempoxyz/tempo/pull/2635/commits/a0bc2fe8a414743ceb698147aa185b1fc90d1dd8)

crates/precompiles/src/validator_config_v2/mod.rs
SuperFluffy: Fixed in [a0bc2fe](https://github.com/tempoxyz/tempo/pull/2635/commits/a0bc2fe8a414743ceb698147aa185b1fc90d1dd8)

SuperFluffy (approved):
Approving. The one thing I am unsure about is the binary representation (and hence keccak256 result) of ipv4 and ipv6 addresses.

My impression is that using `Ipv4::octets` and `Ipv6::octets` takes care of portability across platforms (since we don't have to worry about endianess of u32 and u128). Rust parses IP addresses octet by octet and stores that in big endian format in `[u8; 4]` and `[u8; 16]` arrays, so I think we are good.

Can we still fuzz this somehow?

PR #2638, opened 2026-02-11
refactor(consensus): migrate DKG to use metadata, remove IPs from DKG state 

SuperFluffy:
Migrates the DKG actor's state from a continuous journal to a `Metadata` object. The journal had unnecessary extra complexity that was not needed. The `Metadata` object is simpler to reason about.

This migration is non-breaking: the journal is migrated to metadata and deleted. If a metadata object is already present it is not overwritten.

As part of the migration, the DKG actor no longer tracks the socket addresses of dealers and players. This information is available in the execution layer state and can be read from there.

These changes are in preparation for the Validator Config V2 where address tracking will be moved out of the DKG actor.

Related #2181, #2617

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11815,7 +11815,7 @@ dependencies = [
 
 [[package]]
 name = "tempo"
-version = "1.1.3"
+version = "1.1.4"
 dependencies = [
  "alloy-consensus",
  "alloy-primitives",
@@ -11865,7 +11865,7 @@ dependencies = [
 
 [[package]]
 name = "tempo-alloy"
-version = "1.1.3"
+version = "1.1.4"
 dependencies = [
  "alloy",
  "alloy-consensus",
@@ -11898,7 +11898,7 @@ dependencies = [
 
 [[package]]
 name = "tempo-bench"
-version = "1.1.3"
+version = "1.1.4"
 dependencies = [
  "alloy",
  "alloy-consensus",
@@ -11925,7 +11925,7 @@ dependencies = [
 
 [[package]]
 name = "tempo-chainspec"
-version = "1.1.3"
+version = "1.1.4"
 dependencies = [
  "alloy-eips",
  "alloy-evm",
@@ -11943,7 +11943,7 @@ dependencies = [
 
 [[package]]
 name = "tempo-commonware-node"
-version = "1.1.3"
+version = "1.1.4"
 dependencies = [
  "alloy-consensus",
  "alloy-primitives",
@@ -11996,7 +11996,7 @@ dependencies = [
 
 [[package]]
 name = "tempo-commonware-node-config"
-version = "1.1.3"
+version = "1.1.4"
 dependencies = [
  "commonware-codec",
  "commonware-cryptography",
@@ -12008,7 +12008,7 @@ dependencies = [
 
 [[package]]
 name = "tempo-consensus"
-version = "1.1.3"
+version = "1.1.4"
 dependencies = [
  "alloy-consensus",
  "alloy-evm",
@@ -12026,7 +12026,7 @@ dependencies = [
 
 [[package]]
 name = "tempo-contracts"
-version = "1.1.3"
+version = "1.1.4"
 dependencies = [
  "alloy-contract",
  "alloy-primitives",
@@ -12037,7 +12037,7 @@ dependencies = [
 
 [[package]]
 name = "tempo-dkg-onchain-artifacts"
-version = "1.1.3"
+version = "1.1.4"
 dependencies = [
  "bytes",
  "commonware-codec",
@@ -12051,7 +12051,7 @@ dependencies = [
 
 [[package]]
 name = "tempo-e2e"
-version = "1.1.3"
+version = "1.1.4"
 dependencies = [
  "alloy",
  "alloy-evm",
@@ -12095,7 +12095,7 @@ dependencies = [
 
 [[package]]
 name = "tempo-evm"
-version = "1.1.3"
+version = "1.1.4"
 dependencies = [
  "alloy-consensus",
  "alloy-evm",
@@ -12116,6 +12116,7 @@ dependencies = [
  "revm",
  "tempo-chainspec",
  "tempo-consensus",
+ "tempo-contracts",
  "tempo-payload-types",
  "tempo-primitives",
  "tempo-revm",
@@ -12125,7 +12126,7 @@ dependencies = [
 
 [[package]]
 name = "tempo-eyre"
-version = "1.1.3"
+version = "1.1.4"
 dependencies = [
  "eyre",
  "indenter",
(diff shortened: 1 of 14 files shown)


crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -1463,43 +1481,62 @@ impl Metrics {
...
 ) -> ordered::Map<PublicKey, DecodedValidator> {
     let mut attempts = 0;
     const MIN_RETRY: Duration = Duration::from_secs(1);
     const MAX_RETRY: Duration = Duration::from_secs(30);
 
-    let last = epoch_strategy
-        .last(epoch)
-        .expect("epoch strategy is valid for all epochs");
-    loop {
+    'read_contract: loop {
```
hamdiallam, line 1505: TIL some rust syntax

crates/commonware-node/src/dkg/manager/actor/state.rs
```diff
@@ -611,32 +596,99 @@ impl Builder {
...
-        ordered::Map::from_iter_dedup(
-            self.dealers
-                .iter_pairs()
-                .chain(self.players.iter_pairs())
-                .chain(self.syncers.iter_pairs())
-                .map(|(key, val)| (key.clone(), Address::Symmetric(*val))),
-        )
+    /// Returns the dealers active in the DKG round tracked by this state.
+    pub(super) fn dealers(&self) -> &ordered::Set<PublicKey> {
+        self.output.players()
```
hamdiallam, line 729: Nice, these are are the players in the previous dkg ceremony that participated and now dealing?

crates/commonware-node/src/dkg/manager/actor/state.rs
```diff
@@ -50,7 +52,7 @@ pub(super) struct Storage<TContext>
 where
     TContext: commonware_runtime::Storage + Clock + Metrics,
 {
-    states: contiguous::variable::Journal<TContext, State>,
```
hamdiallam, line 53: Curious why before we used an append-only journal instead of keep track of just the pubkey<->socket mapping for the current epoch?

hamdiallam: Metadata approach with epoch keys feels much better

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -787,13 +804,12 @@ where
             seed: Summary::random(&mut self.context),
             output: onchain_outcome.output.clone(),
             share,
-            dealers: pubkeys_to_addrs(onchain_outcome.players().clone(), &all_validators),
-            players: pubkeys_to_addrs(onchain_outcome.next_players, &all_validators),
-            syncers: ordered::Map::from_iter_dedup(
+            players: onchain_outcome.next_players,
+            syncers: ordered::Set::from_iter_dedup(
                 all_validators
```
hamdiallam, line 809: Seems like `all_validators` is only used to get peers to sync with now. Should we update `read_validator_config_with_retry` so that it does this filtering instead of doing it twice?

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -1508,23 +1545,25 @@ async fn read_validator_config_with_retry<C: commonware_runtime::Clock>(
...
-            key.clone(),
-            validators
-                .get_value(&key)
-                .expect(
-                    "all DKG participants must have an entry in the \
-                        unfiltered, contract validator set; if one does not, \
-                        then it was wrongly included in the ceremony or the \
-                        contract was bad",
-                )
-                .outbound,
```
hamdiallam, line 1526: Why was this outbound before?

Is the peer manager also concerned about incoming connections. If so should we account for this?

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -1508,23 +1545,25 @@ async fn read_validator_config_with_retry<C: commonware_runtime::Clock>(
...
+        .chain(state.syncers.iter());
+
+    ordered::Map::from_iter_dedup(all_keys.map(|key| {
+        let addr = validators
+            .get_value(key)
+            .expect(
+                "all DKG participants must have an entry in the \
+                 unfiltered, contract validator set",
+            )
+            .inbound;
```
hamdiallam: related comment here re inbound/outbound. Makes sense to me for this to be `inbound`

hamdiallam (approved):
Looks good. Few comments

crates/commonware-node/src/dkg/manager/actor/state.rs
SuperFluffy: Indeed! Even the AI was confused here though and initially called `output.dealers()`. For a given Output, `output.dealers()` and `output.players()` are who contributed to the `output`. So `output.players()` then become the dealers in the next epoch.

crates/commonware-node/src/dkg/manager/actor/state.rs
SuperFluffy: The reason is simple: we simply took commonware's upstream example and adapted it to our needs. But the journal can be easily used incorrectly (which we did), and we realized that there is no advantage using the journal over metadata. Commonware in the meantime also changed their example to use metadata.

crates/commonware-node/src/dkg/manager/actor/mod.rs
SuperFluffy: This was me not paying attention. Thank you for noticing! 91dc1981

`validator.outbound` is intended to be the IP address from which a validator connects. To be honest - right now there is no difference. When we first implemented this, commonware did not support asymmetric connections: inbound == outbound. So we just took outbound and ran with it.

But this PR is definitely the wrong place to change this behavior.

crates/commonware-node/src/dkg/manager/actor/mod.rs
SuperFluffy: See comment above: we had to decide which of the 2 we should take, outbound or inbound, since commonware didn't support asymmetric connections. But we shouldn't change it here: 91dc1981

crates/commonware-node/src/dkg/manager/actor/mod.rs
SuperFluffy: I think at this point we can't: at the start of `run_dkg_loop`, we use `all_validators` to map `∀ peer ∈ all_peers -> (peer, all_validators(peer).address)`

crates/commonware-node/src/dkg/manager/actor/mod.rs
hamdiallam: That begs the question on the intention of the peerset. Does the validator initiate handshakes? in which case it would be the `inbound` address of it's peer and not the `outbound`?

PR #2645, opened 2026-02-11
fix: log chain id when downloading snapshot

gakonst:
## Summary

Adds an info log line when running `tempo download` that shows the chain ID and snapshot base URL being used.

## Changes

- Log `chain_id` and `url` before snapshot download starts in `main.rs`

## Testing

`cargo check -p tempo` passes.

Thread: https://tempoxyz.slack.com/archives/C09KCGR4LQ4/p1770722789479339

Prompted by: zygis

### DIFF
--- a/bin/tempo/src/main.rs
+++ b/bin/tempo/src/main.rs
@@ -140,6 +140,11 @@ fn main() -> eyre::Result<()> {
         tempo_cmd::TempoSubcommand,
     >::parse();
 
+    if let Commands::Download(ref cmd) = cli.command {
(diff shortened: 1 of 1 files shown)


bin/tempo/src/main.rs
```diff
@@ -140,6 +140,11 @@ fn main() -> eyre::Result<()> {
         tempo_cmd::TempoSubcommand,
     >::parse();
 
+    if let Commands::Download(ref cmd) = cli.command {
+        let chain_id = cmd.chain_spec().map(|cs| cs.chain().id()).unwrap_or(4217);
```
SuperFluffy, line 144: Can't we set a default value for `chain_spec`? Probably not because reth doesn't have a default?

PR #2653, opened 2026-02-12
chore(val-config-v2): perf optimizations + hardening + cleanup

0xrusowsky:
cleanup and hardening pass on the `ValidatorConfigV2` precompile (stacked on #2635).

- pack `owner` + `init_at_height` into a single `Config` struct (1 SLOAD per guard instead of 2)
- change `validatorByIndex` type to `uint64` to be consistent with storage
- strip port from V1 `outboundAddress` during migration — V1 stores `ip:port` but V2 validates egress as plain IP
- add defense-in-depth check in `migrate_validator` to rejects corrupt V1 data instead of silently overwriting lookups
- consolidate auth guards + DRY helper fns

### DIFF
--- a/crates/precompiles/src/validator_config_v2/mod.rs
+++ b/crates/precompiles/src/validator_config_v2/mod.rs
@@ -497,13 +497,20 @@ impl ValidatorConfigV2 {
         self.require_new_address(v1_val.validatorAddress)?;
         self.require_new_pubkey(v1_val.publicKey)?;
 
+        // V1 outboundAddress is ip:port, V2 egress is plain IP — strip the port
+        let egress = v1_val
+            .outboundAddress
+            .parse::<std::net::SocketAddr>()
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -28,61 +28,88 @@ struct ValidatorV2 {
...
             return Err(ValidatorConfigV2Error::not_initialized())?;
         }
-        Ok(())
+        Ok(config)
     }
 
-    pub fn is_initialized(&self) -> Result<bool> {
-        self.initialized.read()
+    /// Requires initialized + caller is owner. Returns the config.
+    fn require_owner(&self, caller: Address) -> Result<Config> {
```
howydev, line 98: nice simplification! wdyt for clarity we do `require_initialized_owner` instead?

PR #2658, opened 2026-02-12
test(invariants): test newly minted pools

horsefacts:
`_selectInitializedPoolPair` filtered on `reserveUserToken > 0`, which excluded freshly minted pools since minting only deposits validator tokens. Redefined it to use `totalSupply > 0` for true initialized pool selection, and extracted `_selectRebalanceablePoolPair` with the original `reserveUserToken > 0` filter for the three rebalance swap handlers that require it.

Resolves CHAIN-772


zerosnacks (dismissed):
Makes sense

tips/ref-impls/test/invariants/FeeAMM.t.sol
```diff
@@ -166,6 +166,44 @@ contract FeeAMMInvariantTest is InvariantBaseTest {
...
+                    count++;
+                }
+            }
+        }
+
+        vm.assume(count > 0);
+        uint256 idx = bound(seed, 0, count - 1);
+        userToken = validUserTokens[idx];
+        validatorToken = validValidatorTokens[idx];
+    }
```
zerosnacks, line 189: after the introduction  of `_selectRebalanceablePoolPair` this function is no longer used anywhere and should likely be removed

zerosnacks (changes requested):

PR #2684, opened 2026-02-13
tip1016: revm state-gas integration

rakita:
## Summary
- Update revm sub-crate revs to `e3223d5b38f7affb901601ed917fa7cdc466be74` (rakita/state-gas branch)
- Update revm-inspectors rev to `8b26ede9215b5cc7e4aa63b4634d5e9480511a9e`
- Update alloy-evm/alloy-op-evm rev to `ab97815eedb5152011c8add8c8d98555338d9990`
- Update reth deps to `f1f29cc96d207f86d3080786fd87db8c6da642cd`
- Fix `InitialAndFloorGas.initial_gas` → `.initial_total_gas` rename across handler.rs, evm.rs, and validator.rs
- Add 6th param (`state_gas_spent: 0`) to `ResultGas::new()` calls

## Test plan
- [x] `cargo check --workspace --exclude tempo-e2e` compiles clean
- [x] `cargo +nightly clippy --workspace --all-targets --all-features --exclude tempo-e2e` passes
- [x] `cargo nextest run --workspace --exclude tempo-e2e` — all 1311 tests pass


jenpaff (dismissed):
It looks like this is currently gated behind t2, can we change this?

crates/evm/src/block.rs
```diff
@@ -122,6 +122,13 @@ pub(crate) struct TempoBlockExecutor<'a, DB: Database, I> {
     non_shared_gas_left: u64,
     non_payment_gas_left: u64,
     incentive_gas_used: u64,
+
+    /// Tracks total cumulative gas used (execution + storage creation) for receipts.
+    /// This differs from block gas limit accounting which only counts execution gas.
+    cumulative_total_gas_used: u64,
+    /// Tracks cumulative storage creation gas used (TIP-1016).
+    /// Used to derive execution-only gas for the block header.
+    cumulative_storage_creation_gas: u64,
```
rakita: There is no need to track state gas aka storage creation gas as block gas limit takes into account just exec/regular gas.

rakita: I see, it is used to derive exec gas, but it feels more logical to have a `cumulative_exec_gas` variable than state gas.

crates/evm/src/block.rs
```diff
@@ -443,7 +452,14 @@ where
...
         let gas_output = self.inner.commit_transaction(inner)?;
-        let gas_used = gas_output.tx_gas_used();
+
+        // Track cumulative totals for receipts (TIP-1016).
+        self.cumulative_total_gas_used += gas_output.tx_gas_used();
+        self.cumulative_storage_creation_gas += gas_output.state_gas_used();
+
+        // Execution gas excludes storage creation gas (TIP-1016).
+        // Only execution gas counts toward protocol limits (block gas limit).
+        let gas_used = gas_output.tx_gas_used() - gas_output.state_gas_used();
```
rakita: To make things simpler, can we use `gas_output.block_regular_gas_used()` which removes the refund from block gas limit and gives us more accurate reading. I would say lets use it, will communicate in slack about it.

Alternative with a refund accounted would be more complex. `tx_gas_used` does `max(tx_gas_refunded,floor)` [here](https://github.com/bluealloy/revm/blob/81d3b17faba8ff999fcaa8124ab09382710c6242/crates/context/interface/src/result.rs#L266-L267) so deducting state gas after it is wrong as deduction needs to be done before max.

crates/evm/src/block.rs
```diff
@@ -498,8 +514,24 @@ where
...
+        {
+            return Err(
+                BlockValidationError::msg("end-of-block system transactions not seen").into(),
+            );
+        }
+
+        // The inner executor's gas_used tracks total gas (execution + storage)
+        // because that's what goes into receipt cumulative_gas_used. For the
+        // block header, we need execution gas only (TIP-1016).
+        self.inner.block_regular_gas_used -= self.cumulative_storage_creation_gas;
```
rakita: This is not exactly correct as i think self.inner have a floor inside. If we have one variable for used gas it is easier to navigate imo

crates/evm/src/block.rs
```diff
@@ -498,8 +514,24 @@ where
...
 
     fn finish(
-        self,
+        mut self,
     ) -> Result<(Self::Evm, BlockExecutionResult<Self::Receipt>), BlockExecutionError> {
+        // Check that we ended in the System section with all end-of-block system txs seen
+        if self.section
+            != (BlockSection::System {
+                seen_subblocks_signatures: true,
+            })
```
rakita: Is this okay, idk why this logic is needed?

crates/precompiles/src/lib.rs
```diff
@@ -114,6 +114,15 @@ pub fn tempo_precompiles(cfg: &CfgEnv<TempoHardfork>) -> PrecompilesMap {
...
 pub fn extend_tempo_precompiles(precompiles: &mut PrecompilesMap, cfg: &CfgEnv<TempoHardfork>) {
+    // Override MODEXP (0x05) to use Berlin pricing instead of Osaka (EIP-7883).
+    // Tempo maps all hardforks to SpecId::OSAKA, which would use the Osaka MODEXP
+    // with a higher minimum gas (500 vs 200). This override keeps Berlin pricing
+    // for all current Tempo hardforks.
+    let berlin_modexp = revm::precompile::modexp::BERLIN;
+    let modexp_dyn: DynPrecompile =
+        (berlin_modexp.id().clone(), *berlin_modexp.precompile()).into();
+    precompiles.extend_precompiles([(*berlin_modexp.address(), modexp_dyn)]);
+
```
rakita:
```suggestion
```

crates/transaction-pool/src/transaction.rs
```diff
@@ -220,6 +233,27 @@ impl TempoPooledTransaction {
...
+    /// Returns the intrinsic state gas, or 0 if not set (pre-T3 or not yet validated).
+    pub fn intrinsic_state_gas(&self) -> u64 {
+        self.intrinsic_state_gas.get().copied().unwrap_or(0)
+    }
+
+    /// Returns the execution gas limit (gas_limit minus state gas).
+    ///
+    /// For T3+ this is the portion that counts against protocol limits.
+    /// For pre-T3, state gas is 0, so this equals `gas_limit()`.
+    pub fn execution_gas_limit(&self) -> u64 {
```
rakita: Didn't think about this, nice!

crates/evm/src/block.rs
```diff
@@ -122,6 +122,13 @@ pub(crate) struct TempoBlockExecutor<'a, DB: Database, I> {
     non_shared_gas_left: u64,
     non_payment_gas_left: u64,
     incentive_gas_used: u64,
+
+    /// Tracks total cumulative gas used (execution + storage creation) for receipts.
+    /// This differs from block gas limit accounting which only counts execution gas.
+    cumulative_total_gas_used: u64,
```
rakita: Would rename this to `cumulative_tx_gas_used` as it represents gas used by tx found in receipt

crates/evm/src/block.rs
rakita: EIP-8037 names it as `block_regular_gas_used`

fgimenez: indeed, fixed ptal

crates/evm/src/block.rs
fgimenez: good catch, fixed

crates/evm/src/block.rs
fgimenez: yep, already fixed

crates/evm/src/block.rs
fgimenez: yep, this was already removed but was accidentally reintroduced, fixed

crates/evm/src/block.rs
fgimenez: yep, already addressed, this field is not needed

crates/consensus/src/lib.rs
```diff
@@ -203,10 +203,39 @@ impl FullConsensus<TempoPrimitives> for TempoConsensus {
...
+                    &result.receipts,
+                ),
+            });
+        }
+
+        // Delegate receipt root, logs bloom, and requests hash validation to the
+        // inner Ethereum consensus. We construct a temporary result with gas_used
+        // matching the header so the inner gas check passes, while the actual
+        // TIP-1016 gas invariant (header <= receipts) is checked above.
+        let mut patched_result = result.clone();
```
rakita: Wouldn't this clone all receipts? Not sure what but there should be better solution to this

crates/evm/src/block.rs
```diff
@@ -446,8 +446,12 @@ where
             tx,
         } = output;
 
+        // Snapshot block_regular_gas_used before commit to derive per-tx execution gas.
+        // block_regular_gas_used (from ResultGas::block_regular_gas_used()) correctly
+        // excludes state gas before applying floor, unlike tx_gas_used() - state_gas_used().
+        let regular_gas_before = self.inner.block_regular_gas_used;
         let gas_output = self.inner.commit_transaction(inner)?;
-        let gas_used = gas_output.tx_gas_used();
+        let gas_used = self.inner.block_regular_gas_used - regular_gas_before;
```
rakita:
```suggestion
        let gas_output = self.inner.commit_transaction(inner)?;
        // block_regular_gas_used excludes state gas and only accounts for regular gas.
        let gas_used = gas_output.gas().block_regular_gas_used();
```

crates/evm/src/block.rs
```diff
@@ -504,14 +508,15 @@ where
...
+            .cumulative_tx_gas_used
+            .saturating_sub(self.inner.block_state_gas_used);
         let (evm, mut result) = self.inner.finish()?;
-        // The upstream `EthBlockExecutor::finish()` sets `gas_used` to
-        // `max_block_gas_used()` (the pre-refund `block_regular_gas_used`),
-        // but receipts accumulate the post-refund `tx_gas_used`. Override so
-        // `header.gas_used` matches `cumulative_gas_used` in the last receipt,
-        // which is required by post-execution validation.
-        result.gas_used = tx_gas_used;
+        result.gas_used = header_gas_used;
```
rakita: There is `self.inner.block_regular_gas_used` that we can use now. It contains regular gas used only.

crates/transaction-pool/src/validator.rs
```diff
@@ -1145,10 +1168,36 @@ pub fn ensure_intrinsic_gas_tempo_tx(
...
+        // nonce == 0 with non-expiring nonce: potential new account creation
+        if tx.nonce() == 0 && tx.nonce_key() != Some(TEMPO_EXPIRING_NONCE_KEY) {
+            gas.initial_state_gas += gas_params.get(GasId::new_account_state_gas());
+        }
+        // EIP-7702 auth list account creation state gas
+        for auth in tx.authorization_list().unwrap_or_default() {
+            if auth.nonce == 0 {
+                gas.initial_state_gas += gas_params.tx_tip1000_auth_account_creation_state_gas();
+            }
+        }
```
rakita: State gas for auth key is not added here

crates/revm/src/lib.rs
```diff
@@ -2,6 +2,7 @@
 
 #![cfg_attr(not(test), warn(unused_crate_dependencies))]
 #![cfg_attr(docsrs, feature(doc_cfg))]
+#![allow(clippy::result_large_err)]
```
rakita: What result is large enought to trigger this?

crates/revm/src/handler.rs
```diff
@@ -1795,6 +1879,10 @@ pub fn calculate_aa_batch_intrinsic_gas<'a>(
 
             // EIP-3860: Initcode analysis gas using revm helper
             gas.initial_total_gas += gas_params.tx_initcode_cost(call.input.len());
+
+            // TIP-1016: Track predictable state gas for CREATE calls
+            gas.initial_state_gas +=
+                gas_params.new_account_state_gas() + gas_params.create_state_gas();
```
fgimenez: addressed here https://github.com/tempoxyz/tempo/pull/2684/commits/23577b4ca31a2d051c918e47426414ec18263dc8

crates/revm/src/evm.rs
```diff
@@ -50,6 +50,9 @@ pub struct TempoEvm<DB: Database, I> {
     /// The transaction pool sets this because it performs its own liquidity
     /// validation against a cached view of the AMM state.
     pub skip_liquidity_check: bool,
+    /// Initial state gas accumulated during validate_against_state_and_deduct_caller.
+    /// Tracks state gas from runtime checks (e.g. 2D nonce account creation, CREATE state gas).
+    pub(crate) initial_state_gas: u64,
```
fgimenez: already fixed

crates/evm/src/block.rs
```diff
@@ -538,7 +538,16 @@ where
...
         self,
     ) -> Result<(Self::Evm, BlockExecutionResult<Self::Receipt>), BlockExecutionError> {
-        self.inner.finish()
+        // TIP-1016: block header `gas_used` = cumulative_tx_gas_used - block_state_gas_used.
+        // State gas (storage creation, account creation, code deposit) is charged
+        // to users via receipts but exempted from block capacity accounting.
+        let header_gas_used = self
+            .inner
+            .cumulative_tx_gas_used
+            .saturating_sub(self.inner.block_state_gas_used);
```
fgimenez: already fixed

crates/precompiles/src/lib.rs
```diff
@@ -306,20 +307,28 @@ fn mutate_void<T: SolCall>(
...
 /// Fills gas accounting fields on a [`PrecompileOutput`] from the storage context.
+///
+/// State gas is only set for T4+ where the state gas / reservoir model is active.
+/// Pre-T4, `state_gas_used` must remain 0 to avoid leaking into revm's reservoir
+/// accounting and corrupting `tx_gas_used()` via `handle_reservoir_remaining_gas`.
 #[inline]
 fn fill_precompile_output(mut output: PrecompileOutput, storage: &StorageCtx) -> PrecompileOutput {
     output.gas_used = storage.gas_used();
+    if storage.spec().is_t4() {
+        output.state_gas_used = storage.gas_used();
```
fgimenez: addressed here https://github.com/tempoxyz/tempo/pull/2684/commits/26d1d0f98653a925597b6fee85fe55350905bdd3

crates/precompiles/src/error.rs
```diff
@@ -206,7 +208,7 @@ impl TempoPrecompileError {
                 return Err(PrecompileError::Fatal(msg));
             }
         };
-        Ok(PrecompileOutput::revert(gas, bytes, 0))
+        Ok(PrecompileOutput::revert(0, bytes, 0))
```
fgimenez: already fixed

bin/tempo/src/defaults.rs
```diff
@@ -197,33 +194,8 @@ fn init_txpool_defaults() {
...
-        // Another case when this might happen is if we optimistically canonicalize a valid block that is later getting nullified (reorged).
-        // In that case, if we are asked to propose, we would have to build a payload on top of its parent, which is an ancestor of the canonical tip as well.
-        //
-        // This setting allows reth to process payload attributes even if `headBlockHash` is an ancestor of the canonical tip.
-        .with_always_process_payload_attributes_on_canonical_head(true)
-        .try_init()
-        .expect("failed to initialize engine defaults");
-}
-
 pub(crate) fn init_defaults() {
```
fgimenez, line 248: fixed here https://github.com/tempoxyz/tempo/pull/2684/commits/e0bb33849ee8718fd908e193ee486aaf490c819c

bin/tempo/src/defaults.rs
```diff
@@ -156,7 +153,7 @@ fn init_download_urls() {
         ],
         default_base_url: Cow::Borrowed(DEFAULT_DOWNLOAD_URL),
         default_chain_aware_base_url: None,
-        snapshot_api_url: Cow::Borrowed(SNAPSHOT_API_URL),
+        snapshot_api_url: Cow::Borrowed("https://snapshots.reth.rs/api/snapshots"),
```
fgimenez: addressed here https://github.com/tempoxyz/tempo/pull/2684/commits/5ed0ee35d1093c7318fd206418b77dd5b83da6cf

crates/precompiles/src/lib.rs
```diff
@@ -306,20 +307,34 @@ fn mutate_void<T: SolCall>(
...
 
 /// Fills gas accounting fields on a [`PrecompileOutput`] from the storage context.
+///
+/// State gas is only set for T4+ where the state gas / reservoir model is active.
+/// Pre-T4, `state_gas_used` must remain 0 to avoid leaking into revm's reservoir
+/// accounting and corrupting `tx_gas_used()` via `handle_reservoir_remaining_gas`.
 #[inline]
 fn fill_precompile_output(mut output: PrecompileOutput, storage: &StorageCtx) -> PrecompileOutput {
     output.gas_used = storage.gas_used();
+    if storage.spec().is_t4() {
```
fgimenez, line 327: not valid, we can't change gas accounting for blocks already part of the chain

crates/evm/src/block.rs
```diff
@@ -538,7 +538,16 @@ where
     fn finish(
         self,
     ) -> Result<(Self::Evm, BlockExecutionResult<Self::Receipt>), BlockExecutionError> {
-        self.inner.finish()
+        // TIP-1016: block header `gas_used` = cumulative_tx_gas_used - block_state_gas_used.
+        // State gas (storage creation, account creation, code deposit) is charged
+        // to users via receipts but exempted from block capacity accounting.
+        let header_gas_used = self
```
fgimenez: fixed here https://github.com/tempoxyz/tempo/pull/2684/commits/c0b5479d27e0f2db04376dd3d2b67e8432d27b70, also T1C not affected, pre-T4 `state_gas_spent` is always 0

crates/revm/src/handler.rs
```diff
@@ -511,8 +517,15 @@ where
...
 
         let gas_limit = evm.ctx().tx().gas_limit();
-        let mut remaining_gas = gas_limit - init_and_floor_gas.initial_total_gas;
+        // Deduct only the regular (non-state) portion of initial gas, matching
+        // revm's convention: regular_initial_gas = initial_total_gas - initial_state_gas.
+        // State gas goes to the reservoir and is tracked separately.
+        let regular_initial_gas = init_and_floor_gas
+            .initial_total_gas
+            .saturating_sub(init_and_floor_gas.initial_state_gas);
+        let mut remaining_gas = gas_limit.saturating_sub(regular_initial_gas);
```
fgimenez: addressed here https://github.com/tempoxyz/tempo/pull/2684/commits/3f03eb0a1df4e0f9f6d18a073c442ffb05f90585

crates/revm/src/handler.rs
```diff
@@ -586,15 +599,47 @@ where
...
                 }
 
-                normalize_failed_batch_result_gas(&mut frame_result, gas_limit, remaining_gas);
+                // Include gas from all previous successful calls + failed call
+                let gas_spent_by_failed_call = frame_result.gas().total_gas_spent();
+                let total_gas_spent = (gas_limit - remaining_gas) + gas_spent_by_failed_call;
+                // State gas only applies to successful calls that create state.
+                // On revert/halt no new state is created, so the failed call's
+                // state gas is not charged.
+                let total_state_gas = accumulated_state_gas_spent;
```
fgimenez: fixed here https://github.com/tempoxyz/tempo/pull/2684/commits/dcddec70632342e1a4ec035981f1600ad9195e7b

crates/evm/src/block.rs
```diff
@@ -538,7 +538,16 @@ where
     fn finish(
         self,
     ) -> Result<(Self::Evm, BlockExecutionResult<Self::Receipt>), BlockExecutionError> {
-        self.inner.finish()
+        // TIP-1016: block header `gas_used` = cumulative_tx_gas_used - block_state_gas_used.
```
fgimenez: already fixed in https://github.com/tempoxyz/tempo/pull/2684/commits/bf6c6a530b8e12203785dae633bb48617a88b394

crates/revm/src/handler.rs
```diff
@@ -274,23 +313,17 @@ fn calculate_key_authorization_gas(
...
+    _spec: tempo_chainspec::hardfork::TempoHardfork,
     init_and_floor_gas: &InitialAndFloorGas,
 ) -> InitialAndFloorGas {
-    if spec.is_t1() {
-        InitialAndFloorGas::new(evm_initial_gas, init_and_floor_gas.floor_gas)
-    } else {
-        *init_and_floor_gas
-    }
+    *init_and_floor_gas
 }
```
rakita: This fn does nothing rn, it just returns `init_and_floor_gas` back

`is_t1()` is not used

crates/revm/src/handler.rs
```diff
@@ -453,7 +486,19 @@ where
...
     {
-        let gas_limit = evm.ctx().tx().gas_limit() - init_and_floor_gas.initial_total_gas;
+        // Deduct only the regular (non-state) portion of initial gas.
+        // State gas is handled separately by first_frame_input's reservoir logic.
+        // This matches revm's Handler::execution() which computes:
+        //   regular_initial_gas = initial_total_gas - initial_state_gas
+        //   gas_limit = tx.gas_limit - regular_initial_gas
+        let regular_initial_gas = init_and_floor_gas
+            .initial_total_gas
+            .saturating_sub(init_and_floor_gas.initial_state_gas);
```
```suggestion
        let regular_initial_gas = init_and_floor_gas.initial_regular_gas();
```

nit

crates/revm/src/gas_params.rs
```diff
@@ -18,33 +22,274 @@ impl TempoGasParams for GasParams {
...
+            (
+                GasId::new_account_cost_for_selfdestruct(),
+                T4_NEW_ACCOUNT_REGULAR,
+            ),
+            // Code deposit: 200 regular + 2,300 state per byte
+            (GasId::code_deposit_cost(), T4_CODE_DEPOSIT_REGULAR),
+            (GasId::code_deposit_state_gas(), T4_CODE_DEPOSIT_STATE),
+            // EIP-7702 delegation: 25k regular + 225k state = 250k per auth
+            (
+                GasId::tx_eip7702_per_empty_account_cost(),
```
fgimenez, line 85: fixed in https://github.com/tempoxyz/tempo/pull/2684/commits/66371856e417fc1cbbb0a3ad7ab913068c59746a

crates/revm/src/handler.rs
```diff
@@ -890,6 +981,10 @@ where
             && caller_account.nonce() == 0
         {
             evm.initial_gas += cfg.gas_params().get(GasId::new_account_cost());
+            // TIP-1016: Track state gas for new account creation (T4+ only)
+            if spec.is_t4() {
+                evm.initial_state_gas += cfg.gas_params().new_account_state_gas();
+            }
 
             // do the gas limit check again.
             if tx.gas_limit() < evm.initial_gas {
```
fgimenez: addressed in https://github.com/tempoxyz/tempo/pull/2684/commits/9e7a4062ce8ceb6e3b2ef67436d9b18b490ba400

crates/revm/src/handler.rs
```diff
@@ -511,8 +556,26 @@ where
...
+        // State gas accounting is not active pre-T4, but initial_state_gas may be
+        // non-zero due to pre_execution synchronization; deducting it would change
+        // live chain gas accounting.
+        let initial_gas_to_deduct = if spec.is_t4() {
+            init_and_floor_gas.initial_total_gas
+        } else {
+            init_and_floor_gas
+                .initial_total_gas
+                .saturating_sub(init_and_floor_gas.initial_state_gas)
+        };
```
rakita: Does this make sense? Else of `is_t4` uses `initial_state_gas`

crates/revm/src/handler.rs
fgimenez: good call, fixed here https://github.com/tempoxyz/tempo/pull/2684/commits/15b7b4fb3fcfa16ff584cf1fb8199793ee1ce9fa

crates/revm/src/handler.rs
```diff
@@ -766,6 +855,43 @@ where
...
+        // evm.initial_state_gas includes state gas additions (account creation state gas).
+        // Both must be reflected in init_and_floor_gas so that post_execution's
+        // build_result_gas produces the same ResultGas as the execution path used.
+        let spec = evm.ctx_ref().cfg().spec();
+        if spec.is_t1() && init_and_floor_gas.initial_total_gas > 0 {
+            // Propagate evm.initial_state_gas into init_and_floor_gas.
+            // evm.initial_gas already contains init_and_floor_gas.initial_total_gas as base,
+            // plus regular gas additions. Adding evm.initial_state_gas gives the full total.
+            init_and_floor_gas.initial_state_gas += evm.initial_state_gas;
+            init_and_floor_gas.initial_total_gas = evm.initial_gas + evm.initial_state_gas;
```
rakita: i see, `adjusted_initial_gas` logic is moved here. It makes sense.

crates/revm/src/handler.rs
fgimenez: good point, it would be always subtracting 0, fixed here https://github.com/tempoxyz/tempo/pull/2684/commits/e784815ef6bee59a9d91adcf059fa6e934834356

crates/revm/src/evm.rs
```diff
@@ -5,16 +5,89 @@ use revm::{
...
+///
+/// For T4+, Tempo precompiles encode the SSTORE gas refund in the `reservoir` field of
+/// `PrecompileOutput`. This wrapper extracts it after `PrecompilesMap::run()`, applies it
+/// via `record_refund`, and zeros the reservoir. Pre-T4 blocks were executed without
+/// refund propagation from precompiles, so we cannot change their gas accounting.
+#[derive(Debug)]
+pub struct TempoPrecompileProvider {
+    inner: PrecompilesMap,
+    spec: TempoHardfork,
+}
```
klkvr: does this only exist because `refund` was removed from `PrecompileOutput`? I'd prefer us to avoid this

bin/tempo/src/defaults.rs
```diff
@@ -208,14 +207,18 @@ fn init_storage_defaults() {
...
-        // Another case when this might happen is if we optimistically canonicalize a valid block that is later getting nullified (reorged).
-        // In that case, if we are asked to propose, we would have to build a payload on top of its parent, which is an ancestor of the canonical tip as well.
+        // Another case when this might happen is if we optimistically canonicalize a valid block
+        // that is later getting nullified (reorged). In that case, if we are asked to propose, we
+        // would have to build a payload on top of its parent, which is an ancestor of the canonical
+        // tip as well.
         //
-        // This setting allows reth to process payload attributes even if `headBlockHash` is an ancestor of the canonical tip.
+        // This setting allows reth to process payload attributes even if `headBlockHash` is an
+        // ancestor of the canonical tip.
```
klkvr: why is this in the diff?

bin/tempo/src/defaults.rs
```diff
@@ -157,7 +156,7 @@ fn init_download_urls() {
         ],
         default_base_url: Cow::Borrowed(DEFAULT_DOWNLOAD_URL),
         default_chain_aware_base_url: None,
-        snapshot_api_url: Cow::Borrowed(SNAPSHOT_API_URL),
+        snapshot_api_url: Cow::Borrowed("https://snapshots.tempoxyz.dev/api/snapshots"),
```
klkvr: same here, seems unrelated?

crates/consensus/src/lib.rs
```diff
@@ -203,15 +205,83 @@ impl FullConsensus<TempoPrimitives> for TempoConsensus {
...
+                    &result.receipts,
+                ),
+            });
+        }
+
+        // Validate receipt root and logs bloom.
+        // We can't delegate to reth's inner validator because it also checks
+        // gas_used against receipts.last().cumulative_gas_used, which differs
+        // under TIP-1016.
+        validate_receipts_and_requests(block, result, receipt_root_bloom)
```
klkvr: a bit unfortunate that we now won't inherit other post-execution checks, can we encapsulate  `validate_receipts_and_requests` in an upstream helper we can reuse from reth?

crates/node/tests/it/tempo_transaction/mod.rs
```diff
@@ -75,6 +75,9 @@ async fn test_matrices_local(schedule: ForkSchedule) -> eyre::Result<()> {
     run_all_matrices(&mut local::Localnet::with_schedule(schedule).await?).await
 }
 
+// TODO: eth_estimateGas binary search lower bound doesn't account for TIP-1016 state gas,
+// causing "gas_limit < intrinsic_gas" errors for TempoTransactions on T4.
+#[ignore = "blocked on eth_estimateGas TIP-1016 state gas fix"]
```
klkvr: any idea on how to fix this?

crates/payload/types/src/attrs.rs
```diff
@@ -161,12 +161,6 @@ impl From<EthPayloadAttributes> for TempoPayloadAttributes {
 
 impl PayloadAttributes for TempoPayloadAttributes {
     fn payload_id(&self, parent_hash: &B256) -> PayloadId {
-        // XXX: derives the payload ID from the parent so that
-        // overlong payload builds will eventually succeed on the
-        // next iteration: if all other nodes take equally as long,
-        // the consensus engine will kill the proposal task. Then eventually
-        // consensus will circle back to an earlier node, which then
-        // has the chance of picking up the old payload.
```
klkvr, line 169: why was this removed?

crates/precompiles/src/storage/evm.rs
```diff
@@ -87,7 +89,11 @@ impl<'a> PrecompileStorageProvider for EvmPrecompileStorageProvider<'a> {
 
     #[inline]
     fn set_code(&mut self, address: Address, code: Bytecode) -> Result<(), TempoPrecompileError> {
-        self.deduct_gas(self.gas_params.code_deposit_cost(code.len()))?;
+        let code_len = code.len();
+        self.deduct_gas(self.gas_params.code_deposit_cost(code_len))?;
+
+        // Track state gas for code deposit
+        self.state_gas_spent += self.gas_params.code_deposit_state_gas(code_len);
```
klkvr: shouldn't this handle state gas OOG?

crates/primitives/src/reth_compat/transaction/envelope.rs
```diff
@@ -189,8 +189,8 @@ mod codec {
         }
     }
 
-    impl reth_db_api::table::Decompress for TempoTxEnvelope {
-        fn decompress(value: &[u8]) -> Result<Self, DecompressError> {
+    impl reth_codecs::Decompress for TempoTxEnvelope {
```
klkvr: this change seems unrelated as well

crates/revm/src/handler.rs
```diff
@@ -250,8 +289,8 @@ fn calculate_key_authorization_gas(
         let sload_cost =
             gas_params.warm_storage_read_cost() + gas_params.cold_storage_additional_cost();
 
-        let limit_slots = if spec.is_t3() {
-            // T3 periodic limits write 2 storage slots per token:
+        let limit_slots = if spec.is_t4() {
+            // T4 periodic limits write 2 storage slots per token:
```
klkvr: why is this changed to t4?

crates/revm/src/handler.rs
klkvr: why did we have to move this here? i'd prefer to avoid overriding new revm handlers where possible so imo this logic should stay in `fn execution`

crates/revm/src/handler.rs
```diff
@@ -1330,11 +1432,45 @@ where
...
+
+                        StorageCtx.gas_used().saturating_sub(gas_before)
+                    } else {
+                        0
+                    };
+
+                    Ok::<_, EVMError<_, TempoInvalidTransaction>>((
+                        scope_validation_gas,
+                        key_expiry,
+                    ))
```
klkvr: this was removed on main, we don't want this validation to happen here

crates/revm/src/evm.rs
fgimenez: yep exactly

bin/tempo/src/defaults.rs
fgimenez: reverted

bin/tempo/src/defaults.rs
fgimenez: yep this is a leftover from some testing, restored

crates/consensus/src/lib.rs
fgimenez: makes sense, proposed here https://github.com/paradigmxyz/reth/pull/23562

crates/payload/types/src/attrs.rs
fgimenez: was removed as part of the rebase, recovered

crates/precompiles/src/storage/evm.rs
fgimenez: aiui state gas is intentionally left aside from gas limits in TIP-1016, the regular gas portion is already deducted via `deduct_gas` (which returns OOG). state gas is only tracked for block-level accounting, is this correct @rakita ?

crates/primitives/src/reth_compat/transaction/envelope.rs
fgimenez: reverted

crates/revm/src/handler.rs
fgimenez: yep, this was wrongly added as part of the last bump of target hardfork, fixed

crates/revm/src/handler.rs
fgimenez: `execution` receives `&InitialAndFloorGas` (immutable), so we can't modify it there. `pre_execution` is the last point where `init_and_floor_gas` is `&mut`, and the synchronization must happen after `validate_against_state_and_deduct_caller` updates `evm.initial_gas`/`evm.initial_state_gas` but before `execution` and `post_execution` consume `init_and_floor_gas`, does that make sense, do you see any other way to do it?

crates/revm/src/handler.rs
fgimenez: removed

crates/consensus/src/lib.rs
klkvr: proper fix for this is to actually perform the check in the same way in reth https://github.com/paradigmxyz/reth/pull/23569 d5c7bff

crates/evm/src/block.rs
```diff
@@ -1226,7 +1400,7 @@ mod tests {
         let info = acc.account_info();
         assert!(
             info.is_none() || info.unwrap().is_empty_code_hash(),
-            "SignatureVerifier code should not be deployed before T3"
+            "SignatureVerifier code should not be deployed before T4"
         );
     }
```
0xrusowsky, line 1525: nit: this test is unrelated and should remain T3

crates/node/tests/it/tempo_transaction/mod.rs
0xrusowsky: i guess it requires upstream changes?

crates/evm/src/block.rs
fgimenez: sure done here https://github.com/tempoxyz/tempo/pull/2684/commits/65ec7a662ea5b78af66874144d8280164f88ac79

crates/node/tests/it/tempo_transaction/mod.rs
klkvr: yeah but it's not immediately obvious what exactly is the root cause

crates/node/tests/it/eth_call.rs
```diff
@@ -293,7 +293,10 @@ async fn test_eth_estimate_gas(schedule: ForkSchedule) -> eyre::Result<()> {
     let gas = provider.estimate_gas(tx.clone()).await?;
     // gas estimation is calldata dependent, but should be consistent with same calldata
     // TIP-1000 (T1): gas includes 250k new account cost when nonce=0
-    let expected_gas = if schedule.is_active(TempoHardfork::T3) {
+    // T4: state gas is split out, so estimate_gas returns only execution gas
+    let expected_gas = if schedule.is_active(TempoHardfork::T4) {
+        86763
+    } else if schedule.is_active(TempoHardfork::T3) {
```
klkvr: @fgimenez i've reverted this change because it makes no sense for an estimate to return this value. this is an `estimateGas` bug that should be investigated and addressed.

crates/node/tests/it/key_authorization.rs
```diff
@@ -284,11 +284,13 @@ async fn test_post_t1b_keyauth_oog_fixed() -> eyre::Result<()> {
...
+    // TIP-1016 (T4): intrinsic includes state gas for CREATE + KeyAuth + sender
+    // new account (~103k regular + ~918k state = ~1,021k total minimum).
     let encoded = build_create_key_auth_tx(
         &signer,
         chain_id,
         nonce,
-        1_050_000,
+        2_000_000,
         TEMPO_T1_BASE_FEE as u128,
     )?;
```
klkvr, line 293: this is also not immediately obvious to me, which repricings are causing tx to become twice more expensive?

crates/precompiles/src/storage/evm.rs
klkvr: this sounds strange. if tx has a gas limit of 1M it shouldn't be able to spend 2M on state gas

klkvr: also i'd imagine that this should be deducting state gas from regular `gas_remaining` based on `reservoir` value?

crates/precompiles/src/lib.rs
```diff
@@ -322,6 +323,24 @@ pub(crate) fn charge_input_cost(
...
+#[inline]
+fn fill_state_gas(output: &mut PrecompileOutput, storage: &StorageCtx) {
+    if storage.spec().is_t4() {
+        output.state_gas_used = storage.state_gas_used();
+        // T4+: propagate SSTORE refunds via reservoir so the TempoPrecompileProvider
+        // wrapper can apply them with record_refund. Pre-T4 blocks were executed
+        // without refund propagation, so we cannot change their gas accounting.
+        if output.is_success() && storage.gas_refunded() != 0 {
+            output.reservoir = storage.gas_refunded() as u64;
+        }
```
klkvr: aren't we supposed to add refunds to the existing reservoir? right now if there are no refunds we'd always override the reservoir to 0

legion2002 (commented):
:eye: **Cyclops Review** — PR #2684

This PR integrates TIP-1016 / EIP-8037 state-gas and reservoir support. T1C behavior is correctly gated and unaffected. Three T4-specific gas-accounting bugs were identified that will cause user overcharging and a validation gap once T4 activates.

---

### :rotating_light: [SECURITY] Unused EIP-8037 reservoir is double-charged to user — validator overpaid, refund omitted
**Severity:** High (T4 activation) · **File:** `crates/revm/src/handler.rs:1422` (not in diff; enabled by `crates/evm/src/lib.rs:148-149` in this PR)

`TempoEvmHandler::reimburse_caller` computes fee spending from `gas.used()` which does **not** subtract the reservoir. Upstream revm's `reimburse_caller` / `reward_beneficiary` explicitly subtract `gas.reservoir()`. Because this PR enables EIP-8037 on T4, every T4 tx with `gas_limit > cap` will overcharge the user by the full unused reservoir.

**Example:** `gas_limit = 40M`, cap = 30M, execution = 121k → reservoir = 10M. Receipt says 121k, user is charged 10.1M × gas_price.

**Fix:** `let effective_used = gas.used().saturating_sub(gas.reservoir());` in `reimburse_caller` before computing `actual_spending`.

---

See inline comments for the remaining two findings.

<details>

- :zap: **Block lane accounting** (`crates/evm/src/block.rs:497-508`): `non_shared_gas_left` subtracts `tx_gas_used()` (includes state gas), reducing T4 lane capacity below TIP-1016 spec. Should use `block_regular_gas_used()` on T4.
- :zap: **Payload builder** (`crates/payload/builder/src/lib.rs:448-458`): Uses state-inclusive gas limit for cumulative check, also reducing T4 block fill below spec.
- :zap: **Receipt vs fee-payer divergence**: Receipt `tx_gas_used` differs from fee token amount moved by `collect_fee_post_tx` when reservoir > 0. Add an end-to-end assertion.
- :zap: **Inspector path**: `inspect_execute_multi_call` routes through same `execute_multi_call_with`, so the accumulated-state-gas-drop bug affects debug/tracing too.
- :zap: **T4 activation boundary**: Header `gas_used` semantics are fork-gated in `block.rs:541-557` — consensus-critical at activation, worth dedicated replay tests.
- :zap: **Reservoir round-tripping**: `last_frame_result` uses `evm.ctx().tx().gas_limit()` which is temporarily overwritten to per-call `remaining_gas` in multi-call. Confirm frame's `gas.limit()` does not bias refund cap.
- :zap: **Missing regression coverage**: New T4 tests cover "raw `gas_limit > cap` is allowed" but not the separate rule that `initial_regular_gas` must fit inside the cap.

</details>

crates/revm/src/handler.rs
```diff
@@ -214,23 +214,20 @@ fn call_scope_extra_gas(auth: &tempo_primitives::transaction::KeyAuthorization)
...
+        corrected_gas.erase_cost(frame_result.gas().remaining());
     }
-    corrected_gas.set_refund(0); // No refunds when batch fails and all state is reverted.
+    // No refunds when batch fails and all state is reverted.
+    corrected_gas.set_refund(0);
+    // No state gas spending for failed calls
+    corrected_gas.set_state_gas_spent(0);
+    // Reservoir and state gas are refunded on failure
+    corrected_gas
+        .set_reservoir(frame_result.gas().reservoir() + frame_result.gas().state_gas_spent());
```
legion2002: :warning: **[ISSUE]** AA batch failure drops accumulated state gas — user overcharged

`normalize_failed_batch_result_gas` only restores the **failed frame's** `reservoir + state_gas_spent` here. The `accumulated_state_gas_spent` from prior successful sub-calls (whose state is rolled back by `checkpoint_revert`) is silently dropped. This violates TIP-1016 invariant 14: rolled-back state gas must be restored to the reservoir.

**Example:** AA batch — Call 0 does cold SSTORE (230k state gas), Call 1 reverts. User should pay ~130k but pays ~360k.

**Recommended Fix:**
```rust
corrected_gas.set_reservoir(
    frame_result.gas().reservoir()
        + frame_result.gas().state_gas_spent()
        + accumulated_state_gas_spent, // pass as new parameter
);
```
Both call sites (lines 571 and 630) must pass `accumulated_state_gas_spent`.

crates/revm/src/handler.rs
```diff
@@ -1952,6 +1942,9 @@ where
         batch_gas.initial_total_gas += nonce_2d_gas;
     }
 
+    // Validate gas limit is sufficient for initial gas.
```
legion2002: :shield: **[DEFENSE-IN-DEPTH]** Missing EIP-8037 regular-gas cap check

This validation only checks `gas_limit >= initial_total_gas` and `gas_limit >= floor_gas`. On T4, upstream EIP-8037 also requires `max(initial_regular_gas, floor_gas) <= tx_gas_limit_cap` when `tx.gas_limit > cap`. Without this, a transaction whose intrinsic regular gas exceeds 30M can pass validation. The same gap exists in the standard-tx path (~line 1635).

**Recommended Fix:** After computing `init_gas`/`batch_gas`, add:
```rust
if cfg.is_amsterdam_eip8037_enabled() && tx.gas_limit() > cfg.tx_gas_limit_cap() {
    let min_regular = batch_gas.initial_regular_gas().max(batch_gas.floor_gas);
    if min_regular > cfg.tx_gas_limit_cap() {
        return Err(InvalidTransaction::...);
    }
}
```

crates/precompiles/src/storage/evm.rs
```diff
@@ -99,7 +102,11 @@ impl<'a> PrecompileStorageProvider for EvmPrecompileStorageProvider<'a> {
 
     #[inline]
     fn set_code(&mut self, address: Address, code: Bytecode) -> Result<(), TempoPrecompileError> {
-        self.deduct_gas(self.gas_params.code_deposit_cost(code.len()))?;
+        let code_len = code.len();
+        self.deduct_gas(self.gas_params.code_deposit_cost(code_len))?;
+
+        // Track state gas for code deposit
+        self.deduct_state_gas(self.gas_params.code_deposit_state_gas(code_len))?;
```
legion2002: 🚨 **[SECURITY]** TIP20Factory deployments skip the fresh-account surcharge

`set_code` charges only `code_deposit_cost` + `code_deposit_state_gas` but never checks whether the target is an empty account and never charges `new_account_cost` (250k on T1C; 25k+225k on T4). The public `TIP20Factory::create_token` → `__initialize()` → `set_code(fresh_addr, 0xef)` path materializes a permanent new account for only code-deposit gas. Verified: `set_code(fresh)` and `set_code(existing)` consume identical gas.

**Recommended Fix:** Detect empty account before installing code. If empty, charge `new_account_cost()` (and `new_account_state_gas()` on T4+). Audit all `StorageCtx::set_code` callers for the same gap.

crates/payload/builder/src/lib.rs
```diff
@@ -568,6 +575,7 @@ where
                 warn!("no resolved fee token for a pool transaction")
             }
             cumulative_gas_used += gas_used;
```
legion2002: 🚨 **[SECURITY]** State gas leaks into block-level capacity counters (T4+)

`cumulative_gas_used` and `non_payment_gas_used` accumulate state-inclusive `gas_used` on T4, counting state gas against block capacity. TIP-1016 Invariant 4: "State gas MUST NOT count toward protocol limits." `tx_state_gas` is already captured (line 522) but never subtracted.

**Recommended Fix:** `cumulative_gas_used += gas_used - tx_state_gas;` (and similarly for `non_payment_gas_used`).

legion2002 (commented):
:eye: **Cyclops Review** — PR #2684

**4 actionable findings** (2 security, 2 defense-in-depth). Two earlier findings (prevalidate revert overcharge + accumulated state gas loss) were fixed between `910e626` and the current head.

### Findings not in the diff (body-only)

**🚨 [SECURITY] TIP-1016 State Gas Leaks Into Block-Level Capacity Counters (T4+)**
`crates/evm/src/block.rs:501-503` / `crates/evm/src/block.rs:384-385`
T4 enables non-zero `state_gas_spent`. TIP-1016 Invariant 4 requires state gas exempt from all protocol limits. However, `commit_transaction` subtracts `gas_used` (state-inclusive `tx_gas_used()`) from `non_shared_gas_left`, `non_payment_gas_left`, and `incentive_gas_used`. On T4, a CREATE tx (32k regular + 468k state) consumes 500k from block capacity instead of 32k, filling blocks ~15× faster than spec intends.
**Fix:** Use `tx_gas_used - state_gas_used` for lane counters on T4+.

**🛡️ [DEFENSE-IN-DEPTH] Pool Validator Caches Pre-T4 `evm_env` Across T4 Activation Boundary**
`crates/transaction-pool/src/validator.rs:85,640-644`
Pool caches `evm_env(parent_header)`. At T4 boundary, pool validates with T3 gas params (`eip7702_per_empty_account_cost=12.5k`) while execution uses T4 (`250k`). Creates a window of poisoned pool transactions.
**Fix:** Revalidate pool on hardfork boundary, or use `next_evm_env` for validation.

<details>

- ⚡ **`commit_transaction` / `validate_tx` block-capacity accounting**: Outside the diff, fine pre-T4. Once T4 is live, `gas_used` carries state gas into lane counters violating TIP-1016 Invariant 4.
- ⚡ **Payload builder captures `tx_state_gas` but doesn't subtract it**: Closure at `lib.rs:522-527` captures it but capacity checks still use total gas.
- ⚡ **Fresh-account paths beyond TIP20Factory**: `__initialize()` exists for multiple precompile types. Audit all `StorageCtx::set_code` callers.
- ⚡ **`sstore_set_without_load_cost = 20,000` vs comment "= 19,900"**: Total SSTORE = 250,100 instead of spec 250,000. Not exploitable, worth fixing.
- ⚡ **Inspector path shares `execute_multi_call_with`**: Gas fixes must also cover debug/tracing path.
</details>

**Stale findings** (fixed on current head): prevalidate revert charging full `gas_limit` ✅, accumulated state gas loss on batch failure ✅

legion2002 (commented):
:eye: **Cyclops Review** — PR #2684 (TIP-1016 / EIP-8037 State Gas Integration)

**Head drift:** Findings audited at `390dc1bf33f1`, current head `f873f0e`. Commit `0311db0bd` ("fix: charge CREATE cost on set_code") partially fixes the TIP20Factory finding for T4 only; T1C remains unfixed. All other findings confirmed still present.

**6 findings** survived dedup + verification across 3 workers × 3 iterations:

| # | Tier | Title | Severity | Status |
|---|------|-------|----------|--------|
| 1 | 🚨 SECURITY | State gas leaks into block-level capacity counters | Medium | Open (T4) |
| 2 | 🚨 SECURITY | Upstream block precheck reserves Osaka gas cap (16.7M) | High | Open (T4) |
| 3 | 🚨 SECURITY | Txpool/RPC still enforce legacy raw gas cap | High | Open (T4) |
| 4 | 🚨 SECURITY | Prevalidate failure drops initial reservoir → overcharge | Medium | Open (T4) |
| 5 | 🚨 SECURITY | TIP20Factory skips fresh-account surcharge on T1C | Medium | Open (T1C) |
| 6 | 🛡️ DEFENSE | Pool validator caches pre-T4 evm_env across boundary | Low | Open |

**Finding #5 is the only one affecting the current active fork (T1C).** All others are latent T4 bugs.

---

### 🚨 State Gas Leaks Into Block-Level Capacity Counters (T4)
`crates/evm/src/block.rs:497-528` (outside diff) — `commit_transaction` and `validate_tx` use `gas_output.tx_gas_used()` (state-inclusive) for `non_shared_gas_left`, `non_payment_gas_left`, `incentive_gas_used`. Violates TIP-1016 Invariant 4. Payload builder has the same issue. Fix: use `tx_gas_used - state_gas_used` for capacity counters on T4+.

### 🚨 Upstream Block Precheck Reserves Osaka Gas Cap (T4)
`alloy-evm-0.32.0/src/eth/block.rs:165-181` — delegated `EthBlockExecutor` reserves `min(tx.gas_limit, TX_GAS_LIMIT_CAP=16,777,216)` before execution. Once >13.2M regular gas is consumed, valid high-gas-limit T4 txs are rejected. Consensus-split risk.

### 🚨 Txpool/RPC Still Enforce Legacy Raw Gas Cap (T4)
Upstream stateless validation rejects raw `gas_limit > 30M`. RPC estimate clamps to `min(cap, block_limit)`. Valid T4 txs become unrelayable.

### 🛡️ Pool Validator Cross-Fork Boundary
`crates/transaction-pool/src/validator.rs:640-644` — Pool caches `evm_env(parent_header)`. At T4 boundary, EIP-7702 txs admitted under T3 rules fail T4 execution.

### 💡 GasPricing Test Missing T4 Config
`tips/ref-impls/test/invariants/GasPricing.t.sol` — No `forge-config` T4 override, so TIP-1016 invariants run under T3.

<details>

- ⚡ **Revert-restore vs halt-consume on T4 multi-call**: Spilled state gas refunding on failure matches TIP-1016 Note 150 but is counterintuitive — worth documenting.
- ⚡ **AA CREATE + 2D-nonce edge**: `validate_against_state_and_deduct_caller` bumps `initial_total_gas` without re-checking T4 cap. User at exactly 30M gets post-bump OOG. Footgun, not security bug.
- ⚡ **`sstore_set_without_load_cost = 20,000` vs comment "19,900"**: 100-gas overcharge on SSTORE total (250,100 vs spec 250,000). Enshrined in tests.
- ⚡ **FeeHistory rewards dimensional mismatch**: T4 header exposes regular gas, but upstream fee-history uses `header.gas_used()` + receipt cumulative gas in same calculator.
- ⚡ **All `StorageCtx::set_code` callers**: Fresh-account gap applies to every precompile using `__initialize()`, not just TIP20Factory.
</details>

crates/precompiles/src/storage/evm.rs
```diff
@@ -99,11 +102,24 @@ impl<'a> PrecompileStorageProvider for EvmPrecompileStorageProvider<'a> {
...
+
+        let was_empty = {
+            let mut account = self.internals.load_account_mut(address)?;
+            let was_empty = account.data.account().info.is_empty();
+            account.set_code_and_hash_slow(code);
+            was_empty
+        };
+
+        // T4: charge TIP20 creations as CREATE
+        if self.spec.is_t4() && was_empty {
```
klkvr, line 119: we can't change existing gas accounting

crates/evm/src/evm.rs
```diff
@@ -65,10 +65,20 @@ pub struct TempoEvm<DB: Database, I = NoOpInspector> {
...
     /// Create a new [`TempoEvm`] instance.
     pub fn new(db: DB, input: EvmEnv<TempoHardfork, TempoBlockEnv>) -> Self {
+        let mut cfg_env = input.cfg_env;
+
+        // TIP-1016 (T4): Enable EIP-8037 state gas charging automatically.
+        // This ensures all consumers (including foundry's executeTransaction cheatcode)
+        // get the correct gas validation behavior where tx.gas_limit > cap is allowed
+        // when the excess is state gas.
+        if cfg_env.spec.is_t4() {
+            cfg_env.enable_amsterdam_eip8037 = true;
```
klkvr, line 78: 1 addressed by https://github.com/paradigmxyz/reth/pull/23612
2 addressed by https://github.com/alloy-rs/evm/pull/343

legion2002:
## :eye: Cyclops Security Review

[`f5bdbdf`](https://github.com/tempoxyz/tempo/commit/f5bdbdf5de3b)

🧭 Auditing · mode=`normal` · workers 1/3 done (2 left) · verify pending 1

| Worker | Engine | Progress | Status |
|--------|--------|:--------:|--------|
| `pr-2684-w1` | claude-opus-4-7 | 🔍 [thread-1](file:///home/ubuntu/cyclops/logs/pr-2684/conversations/pi-1776437789-2934844.json) · · | Running |
| `pr-2684-w2` | amp/deep | ✅ [thread-1](https://ampcode.com/threads/T-019d9bf1-8c30-74be-9815-ec640435a098) ✅ [thread-2](https://ampcode.com/threads/T-019d9bf5-468b-7487-9a76-9607c6144a77) ✅ [thread-3](https://ampcode.com/threads/T-019d9c02-2894-73ab-9371-35f64c810c3c) | Done |
| `pr-2684-w3` | gpt-5.4 | 🚨 [thread-1](https://united-prana-xjws.here.now/) 🔍 [thread-2](file:///home/ubuntu/cyclops/logs/pr-2684/conversations/pi-1776440289-2935393.json) · | Running |

#### Findings

| # | Finding | Severity | Verification | Threads |
|:-:|---------|:--------:|:------------:|---------|
| 1 | Cached precompile outputs replay stale reservoirs during payload validation | `Critical` | ⏳ Pending | [audit](https://united-prana-xjws.here.now/) |

<details><summary>⚙️ Controls</summary>

- 🚀 Keep only 1 remaining iteration per worker after the current work finishes.
- 👀 Keep only 2 remaining iterations per worker after the current work finishes.
- ❤️ Let only worker 1 continue; other workers skip queued iterations.
- 😄 Let only worker 2 continue; other workers skip queued iterations.
- 🎉 End faster by skipping queued iterations and moving toward consolidation.
- 😕 Stop active workers/verifiers now and start consolidation immediately.

</details>

---
<details><summary>📜 13 events</summary>

🔍 `pr-2684-w1` iter 1/3 [audit-general.md]
🔍 `pr-2684-w2` iter 1/3 [audit-focused.md]
🔍 `pr-2684-w3` iter 1/3 [audit-ripple.md]
✅ `pr-2684-w2` iter 1 — clear | [Thread](https://ampcode.com/threads/T-019d9bf1-8c30-74be-9815-ec640435a098)
🔍 `pr-2684-w2` iter 2/3 [audit-ripple.md]
✅ `pr-2684-w2` iter 2 — clear | [Thread](https://ampcode.com/threads/T-019d9bf5-468b-7487-9a76-9607c6144a77)
🔍 `pr-2684-w2` iter 3/3 [audit-invariants.md]
✅ `pr-2684-w2` iter 3 — clear | [Thread](https://ampcode.com/threads/T-019d9c02-2894-73ab-9371-35f64c810c3c)
🏁 `pr-2684-w2` done
🚨 `pr-2684-w3` iter 1 — finding | [Thread](https://united-prana-xjws.here.now/)
🚨 Finding: **Cached precompile outputs replay stale reservoirs during payload validation** (Critical) | [Thread](https://united-prana-xjws.here.now/)
🔍 `pr-2684-w3` iter 2/3 [audit-invariants.md]
🔬 Verifying: **Cached precompile outputs replay stale reservoirs during payload validation** | [Thread](https://ampcode.com/threads/T-019d9c17-ad80-7288-a102-a37b97c8407a)

</details>

crates/precompiles/src/storage/evm.rs
```diff
@@ -99,11 +102,24 @@ impl<'a> PrecompileStorageProvider for EvmPrecompileStorageProvider<'a> {
...
     fn set_code(&mut self, address: Address, code: Bytecode) -> Result<(), TempoPrecompileError> {
-        self.deduct_gas(self.gas_params.code_deposit_cost(code.len()))?;
+        let code_len = code.len();
+        self.deduct_gas(self.gas_params.code_deposit_cost(code_len))?;
 
-        self.internals
-            .load_account_mut(address)?
-            .set_code_and_hash_slow(code);
+        // Track state gas for code deposit
+        self.deduct_state_gas(self.gas_params.code_deposit_state_gas(code_len))?;
```
legion2002: ⚠️ **[ISSUE] Pre-T4 precompile storage paths charge EIP-8037 state gas (active on T1C)**

`deduct_state_gas` runs unconditionally here and at line 178 (`sstore`). Pre-T4 `tempo_gas_params` does not zero state-gas `GasId`s, so upstream OSAKA defaults leak non-zero values (e.g., `code_deposit_state_gas = 2,300`). On T1C, `record_state_cost` spills into regular remaining gas when reservoir is zero, increasing effective gas consumption.

**Recommended Fix:** Gate with `if self.spec.is_t4()` or zero state-gas `GasId`s for pre-T4 in `tempo_gas_params`.

legion2002 (commented):
:eye: **Cyclops Review**

This PR integrates TIP-1016 / EIP-8037 two-dimensional state gas into Tempo's handler, precompile storage, block executor, and payload builder. Code quality is high with extensive test coverage, but several cross-layer integration gaps remain.

**8 findings** (2 🚨 Security, 3 ⚠️ Issue, 2 🛡️ Defense-in-Depth — plus 1 🚨 Security body-only below).

### Findings not in diff (body-only)

🚨 **[SECURITY] Default engine precompile cache replays stale reservoir (T4)**
Upstream `CachedPrecompile` (`reth engine/tree/src/tree/precompile_cache.rs`) caches full `PrecompileOutput` keyed by `(calldata, spec)`. After T4 enables EIP-8037, cached outputs carry `reservoir`/`state_gas_used`/`gas_refunded`. On cache hit, revm restores the cached reservoir into the parent frame — replaying an earlier larger reservoir value after intervening state-gas consumption, effectively minting state gas. **Fix:** Key cache on gas-state fields, strip reservoir from cached outputs, or disable caching for T4.

🚨 **[SECURITY] Upstream block-fit precheck rejects valid late T4 transactions**
Upstream `alloy-evm` `execute_transaction_without_commit` reserves `min(tx.gas_limit, TX_GAS_LIMIT_CAP)` against block capacity. On T4, valid txs can exceed the cap when the excess is state gas, but the precheck uses clamped total. Late-block T4 transactions are rejected even when regular gas fits. **Fix:** Override upstream precheck in `TempoBlockExecutor` to use regular gas on T4.

⚠️ **[ISSUE] Payload builder counts state gas against block capacity (T4)**
`cumulative_gas_used += gas_used` and `non_payment_gas_used += gas_used` (`lib.rs:570-572`) use `tx_gas_used()` (state-inclusive). The new `cumulative_state_gas_used` is only for metrics. Violates TIP-1016 Invariant 4. **Fix:** Use `gas_used - tx_state_gas` for capacity accounting.

🛡️ **[DEFENSE-IN-DEPTH] Pool validator caches pre-T4 evm_env across activation boundary**
`TempoTransactionValidator` refreshes cached env from current tip header, not next-block. At T4 activation, pool validates with pre-T4 intrinsic costs, admitting txs that T4 execution rejects.

<details>

- ⚡ **Subblock budgeting parity (T4)**: `subblocks.rs:716,728,738,856` and `block.rs:287` sum raw `tx.gas_limit()` for per-subblock reservations. Under-utilises T4 subblock capacity.
- ⚡ **EXPIRING_NONCE_GAS under-accounting**: Known ZELLIC-50. On T4, precompile SSTOREs hit fresh ring-buffer slots; intrinsic 13k does not include state-gas reservation.
- ⚡ **`sstore_set_without_load_cost = 20,000` vs spec 19,900**: Total SSTORE cost 250,100 vs spec 250,000. Test `test_t4_totals_match_spec` enshrines the 100-gas overcharge.
- ⚡ **Upstream dependency drift**: Correctness depends on upstream revm reservoir semantics. Any upstream change should trigger Tempo regression tests.
- ⚡ **Constructor vs env split fragility**: Future hardfork flags added only in `TempoEvm::new()` will silently miss `evm_env()` consumers.

</details>

crates/revm/src/evm.rs
```diff
@@ -81,13 +77,31 @@ impl<DB: Database, I> TempoEvm<DB, I> {
...
+    /// Computes initial gas limit and reservoir for a transaction given its initial gas spending.
+    pub(crate) fn initial_gas_and_reservoir(
+        &self,
+        init_and_floor_gas: &InitialAndFloorGas,
+    ) -> (u64, u64) {
+        // Pre-T0 it could happen that the initial gas spending is greater than the gas limit due to faulty validation.
+        //
+        // Before that it would overflow, so we are reproducing this behavior here by setting the gas limit to u64::MAX and the reservoir to 0.
+        if !self.cfg.spec.is_t0() && init_and_floor_gas.initial_total_gas > self.tx.gas_limit {
+            (u64::MAX, 0)
```
klkvr, line 96: Pre-T0 gas accounting is completly messed up and works purely by coincidents. It panics in debug builds and wraps in release builds. Re-execution passed with the current logic so I don't think we can follow it any more closely

crates/precompiles/src/storage/evm.rs
klkvr, line 109: that's not true, Osaka gas params don't include state gas values

crates/precompiles/src/storage/evm.rs
```diff
@@ -99,11 +102,24 @@ impl<'a> PrecompileStorageProvider for EvmPrecompileStorageProvider<'a> {
...
+        let was_empty = {
+            let mut account = self.internals.load_account_mut(address)?;
+            let was_empty = account.data.account().info.is_empty();
+            account.set_code_and_hash_slow(code);
+            was_empty
+        };
+
+        // T4: charge TIP20 creations as CREATE
+        if self.spec.is_t4() && was_empty {
+            self.deduct_gas(self.gas_params.create_cost())?;
```
klkvr, line 120: we can't do this without `is_t4` gate as it's gonna be breaking. it's not possible to transfer value so it's fine to not charge because no accounts can hold value

crates/revm/src/evm.rs
klkvr, line 96: T0 has already passed so it should be ok, see https://github.com/tempoxyz/tempo/pull/2684#discussion_r3102107210

legion2002:
## :eye: Cyclops Security Review

[`48365cf`](https://github.com/tempoxyz/tempo/commit/48365cf665f1)

🧭 Auditing · mode=`normal` · workers 0/3 done (3 left) · verify pending 0

| Worker | Engine | Progress | Status |
|--------|--------|:--------:|--------|
| `pr-2684-w1` | claude-opus-4-7 | ❌ [thread-1](https://calm-savoy-ahcn.here.now/) 🔍 [thread-2](file:///home/ubuntu/cyclops/logs/pr-2684/conversations/pi-1776453584-3228123.json) · | Running |
| `pr-2684-w2` | amp/deep | 🔍 [thread-1](https://ampcode.com/threads/T-019d9ce0-eb92-77fb-b359-68b69252b34a) · · | Running |
| `pr-2684-w3` | gpt-5.4 | 🔍 [thread-1](file:///home/ubuntu/cyclops/logs/pr-2684/conversations/pi-1776453480-3228676.json) · · | Running |

<details><summary>⚙️ Controls</summary>

- 🚀 Keep only 1 remaining iteration per worker after the current work finishes.
- 👀 Keep only 2 remaining iterations per worker after the current work finishes.
- ❤️ Let only worker 1 continue; other workers skip queued iterations.
- 😄 Let only worker 2 continue; other workers skip queued iterations.
- 🎉 End faster by skipping queued iterations and moving toward consolidation.
- 😕 Stop active workers/verifiers now and start consolidation immediately.

</details>

---
<details><summary>📜 5 events</summary>

🔍 `pr-2684-w1` iter 1/3 [audit-general.md]
🔍 `pr-2684-w2` iter 1/3 [audit-focused.md]
🔍 `pr-2684-w3` iter 1/3 [audit-ripple.md]
❌ `pr-2684-w1` iter 1 — failed | [Thread](https://calm-savoy-ahcn.here.now/)
🔍 `pr-2684-w1` iter 2/3 [audit-focused.md]

</details>

legion2002 (commented):
:eye: **Cyclops Review** — PR #2684 (EIP-8037 state gas reservoir + TIP-1016 block-capacity model)

Core state-gas plumbing (reservoir propagation, precompile `fill_state_gas`, block executor section accounting, AA batch normalization) is internally consistent. **4 actionable findings** survive dedup: 1 High (lane-admission monopolization), 2 Medium (payload fee scoring, pre-T4 state-gas overcharge), 1 Low (test config gap). One additional Low-severity finding on pool validator boundary caching is in a file outside the diff.

_Audited at `48365cf665f1`, re-checked against current head `841de2cd`. All findings confirmed still present._

### Findings not on changed lines

🛡️ **[DEFENSE-IN-DEPTH]** Pool validator caches pre-T4 `evm_env` across activation boundary (`crates/transaction-pool/src/validator.rs`, not in diff). At T3→T4 boundary, pool uses pre-T4 intrinsic gas rules while T4 execution applies 250k per-auth costs—txs sized for old costs enter the pool but fail at execution. No automatic revalidation on fork activation.

<details>

- ⚡ **Subblock budgeting variant**: `subblocks.rs:715-741,859-867` and `block.rs:287-295` use the same `min(tx.gas_limit, cap)` surrogate. On multi-validator networks, cheap over-cap T4 txs become unplaceable in subblocks.
- ⚡ **Double-charge for 2D-nonce CREATE with caller.nonce=0**: Both `validate_aa_initial_tx_gas` and `validate_against_state_and_deduct_caller` add `account_cost + account_state_gas` for a T4 AA tx. Over-charges by 250k total / 225k state. Not exploitable.
- ⚡ **`sstore_set_without_load_cost = 20,000` vs spec 19,900**: Total SSTORE cost is 250,100 vs spec 250,000 (100-gas warm-read overcharge).
- ⚡ **eth_feeHistory threshold drift**: On T4, reward percentile threshold uses `block_regular_gas_used` but receipt deltas use `cumulative_tx_gas_used`.
- ⚡ **Upstream late-block precheck**: `alloy-evm` uses `is_amsterdam_active_at_timestamp`; Tempo maps to OSAKA, so check still uses state-inclusive gas.
- ⚡ **Engine precompile cache**: Cached `PrecompileOutput` replays stale reservoir/state_gas_used on hit.
</details>

crates/precompiles/src/storage/evm.rs
klkvr, line 109: that is wrong, pre-T4 gas params do not contain any state gas values because they are inherited from Osaka params

crates/payload/builder/src/lib.rs
```diff
@@ -451,10 +453,15 @@ where
                 break;
             };
 
+            let max_regular_gas_used = core::cmp::min(
```
klkvr, line 455: this is an interesting find but is definitely not a high severify is mostly a UX issue

legion2002:
## :eye: Cyclops Security Review

[`04edc0a`](https://github.com/tempoxyz/tempo/commit/04edc0a3242b)

🧭 Auditing · mode=`normal` · workers 2/3 done (1 left) · verify pending 0

| Worker | Engine | Progress | Status |
|--------|--------|:--------:|--------|
| `pr-2684-w1` | claude-opus-4-7 | ❌ [thread-1](file:///home/ubuntu/cyclops/logs/pr-2684/conversations/pi-1776496116-3640963.json) ✅ [thread-2](https://starry-nimbus-hhxm.here.now/) ❌ [thread-3](https://vast-signal-jheg.here.now/) | Done |
| `pr-2684-w2` | amp/deep | ✅ [thread-1](https://ampcode.com/threads/T-019d9f6b-8c7c-7208-9f76-a78d7152f74f) ✅ [thread-2](https://ampcode.com/threads/T-019d9f74-c332-7668-9cf3-839f387dbbd8) ✅ [thread-3](https://ampcode.com/threads/T-019d9f87-21db-704e-a014-92f35a22de98) | Done |
| `pr-2684-w3` | gpt-5.4 | ✅ [thread-1](https://onyx-forge-7ter.here.now/) 🚨 [thread-2](https://bright-vision-mkqs.here.now/) 🔍 [thread-3](file:///home/ubuntu/cyclops/logs/pr-2684/conversations/pi-1776500371-3641514.json) | Running |

#### Findings

| # | Finding | Severity | Verification | Threads |
|:-:|---------|:--------:|:------------:|---------|
| 1 | PR 2684 / commit `04edc0a3242b` — invariant review | `Unknown` | ✅ Verified | [audit](https://bright-vision-mkqs.here.now/) · [verify](https://ampcode.com/threads/T-019d9fac-7265-73db-b3b3-950aa173f052) |

<details><summary>⚙️ Controls</summary>

- 🚀 Keep only 1 remaining iteration per worker after the current work finishes.
- 👀 Keep only 2 remaining iterations per worker after the current work finishes.
- ❤️ Let only worker 1 continue; other workers skip queued iterations.
- 😄 Let only worker 2 continue; other workers skip queued iterations.
- 🎉 End faster by skipping queued iterations and moving toward consolidation.
- 😕 Stop active workers/verifiers now and start consolidation immediately.

</details>

---
<details><summary>📜 22 events</summary>

🔍 `pr-2684-w1` iter 1/3 [audit-general.md]
🔍 `pr-2684-w2` iter 1/3 [audit-focused.md]
🔍 `pr-2684-w3` iter 1/3 [audit-ripple.md]
✅ `pr-2684-w2` iter 1 — clear | [Thread](https://ampcode.com/threads/T-019d9f6b-8c7c-7208-9f76-a78d7152f74f)
🔍 `pr-2684-w2` iter 2/3 [audit-ripple.md]
❌ `pr-2684-w1` iter 1 — failed | [Thread](file:///home/ubuntu/cyclops/logs/pr-2684/conversations/pi-1776496116-3640963.json)
🔍 `pr-2684-w1` iter 2/3 [audit-focused.md]
✅ `pr-2684-w2` iter 2 — clear | [Thread](https://ampcode.com/threads/T-019d9f74-c332-7668-9cf3-839f387dbbd8)
🔍 `pr-2684-w2` iter 3/3 [audit-invariants.md]
✅ `pr-2684-w3` iter 1 — clear | [Thread](https://onyx-forge-7ter.here.now/)
🔍 `pr-2684-w3` iter 2/3 [audit-invariants.md]
✅ `pr-2684-w2` iter 3 — clear | [Thread](https://ampcode.com/threads/T-019d9f87-21db-704e-a014-92f35a22de98)
🏁 `pr-2684-w2` done
✅ `pr-2684-w1` iter 2 — clear | [Thread](https://starry-nimbus-hhxm.here.now/)
🔍 `pr-2684-w1` iter 3/3 [audit-ripple.md]
❌ `pr-2684-w1` iter 3 — failed | [Thread](https://vast-signal-jheg.here.now/)
🏁 `pr-2684-w1` done
🚨 `pr-2684-w3` iter 2 — finding | [Thread](https://bright-vision-mkqs.here.now/)
🚨 Finding: **PR 2684 / commit `04edc0a3242b` — invariant review** (Unknown) | [Thread](https://bright-vision-mkqs.here.now/)
🔍 `pr-2684-w3` iter 3/3 [audit-general.md]
🔬 Verifying: **PR 2684 / commit `04edc0a3242b` — invariant review** | [Thread](https://ampcode.com/threads/T-019d9fac-7265-73db-b3b3-950aa173f052)
📋 Verify: **PR 2684 / commit `04edc0a3242b` — invariant review** → ✅ Verified | [Thread](https://ampcode.com/threads/T-019d9fac-7265-73db-b3b3-950aa173f052)

</details>

crates/evm/src/lib.rs
```diff
@@ -141,6 +141,13 @@ impl ConfigureEvm for TempoEvmConfig {
         let mut cfg_env = cfg_env.with_spec_and_gas_params(spec, tempo_gas_params(spec));
         cfg_env.tx_gas_limit_cap = spec.tx_gas_limit_cap();
 
+        // TIP-1016 (T4): Enable EIP-8037 state gas charging. All Tempo hardforks map to
+        // SpecId::OSAKA, but revm's AMSTERDAM (which gates record_state_cost in SSTORE/CREATE)
+        // comes after OSAKA in the SpecId enum. Without this flag, state gas overrides
+        // (sstore_set_state_gas, new_account_state_gas, etc.) are configured but never charged.
+        if spec.is_t4() {
+            cfg_env.enable_amsterdam_eip8037 = true;
```
klkvr, line 149: fixed by https://github.com/alloy-rs/evm/pull/344

crates/payload/builder/src/lib.rs
```diff
@@ -451,10 +454,15 @@ where
...
```
klkvr, line 455: if we don't assume that full cap might get consumed it would open a DOS vector

fgimenez (approved):

crates/payload/builder/src/lib.rs
0xrusowsky: i agree that using `tx.gas_limit()` undermines the benefits of the reservoir model by treating state-gas as if it consumed block capacity. IMO we should limit based on regular gas (or a proxy for it).

i guess it's fine to take it in a follow-up PR since this is just at the builder-level

0xrusowsky (approved):
just pending the resolution of the merge conflicts

klkvr (approved):

PR #2693, opened 2026-02-13
fix(invariants): track nonces for CREATE verification

horsefacts:
The C5 invariant checker iterated sequential indices 0..createCount to verify CREATE addresses, but addresses were recorded by actual protocol nonce. Any non-CREATE transaction before a CREATE caused a key mismatch, and the `if (recorded != address(0))` guard silently skipped every entry.

Add ghost_createNonces array to track actual nonces used during CREATE operations and iterate over those in both verifier functions. Assert that recorded addresses are non-zero instead of silently skipping.

Resolves CHAIN-781


zerosnacks (approved):
Makes sense

PR #2696, opened 2026-02-13
chore(consensus): move all peer handling to peer manager

SuperFluffy:
Moves peer handling from the DKG actor and into the peer manager.

In preparation for the val config V2 changes introduced in TIP 1017.

Related: #2181 #2617

### DIFF
--- a/crates/commonware-node/src/peer_manager/actor.rs
+++ b/crates/commonware-node/src/peer_manager/actor.rs
@@ -1,25 +1,27 @@
 use alloy_consensus::BlockHeader as _;
 use commonware_codec::ReadExt as _;
-use commonware_consensus::Heightable as _;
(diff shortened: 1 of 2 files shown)


crates/commonware-node/src/validators.rs
```diff
@@ -0,0 +1,270 @@
...
+
+    'read_contract: loop {
+        total_attempts.inc();
+        attempts += 1;
+
+        let target_height = match target {
+            ReadTarget::Exact { height } => height,
+            ReadTarget::AtLeast { height } => node
+                .provider
+                .best_block_number()
```
joshieDo, line 52: good, since `node.provider` is a `BlockchainProvider` - this will always return the very tip (which might be in-memory and not persisted yet)

PR #2711, opened 2026-02-16
fix(precompiles): gate DEX `place_flip` order under `T1A`

0xrusowsky:
This PR gates the fix in #2654 under the T1A hardfork

### DIFF
--- a/crates/precompiles/src/stablecoin_dex/mod.rs
+++ b/crates/precompiles/src/stablecoin_dex/mod.rs
@@ -4278,33 +4278,73 @@ mod tests {
         })
     }
 
+    /// Common state produced by [`setup_flip_order_test`].
+    struct FlipOrderTestCtx {
+        exchange: StablecoinDEX,
+        alice: Address,
+        bob: Address,
+        admin: Address,
+        base_token: Address,
+        quote_token: Address,
+        book_key: B256,
+        amount: u128,
+        flip_tick: i16,
+    }
+
+    /// Sets up a [`StablecoinDEX`] with a flip bid order ready to be filled.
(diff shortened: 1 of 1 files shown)


klkvr (approved):

crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -4278,58 +4279,156 @@ mod tests {
...
+                let expected_escrow = (amount * price as u128) / orderbook::PRICE_SCALE as u128;
+
+                let (base_token, quote_token) =
+                    setup_test_tokens(admin, alice, exchange.address, expected_escrow * 2)?;
+                exchange.create_pair(base_token)?;
+
+                let book_key = compute_book_key(base_token, quote_token);
+
+                // Place a flip bid order: when filled, it should flip to an ask at flip_tick
+                exchange.place_flip(alice, base_token, amount, true, tick, flip_tick, false)?;
```
0xKitsune: Looks like this setup is exactly the same for `test_flip_order_fill_reverts_on_system_error_post_t2`. Can we add a helper to consolidate?

crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -4278,58 +4279,156 @@ mod tests {
...
+                if spec.is_t2() {
+                    // T2: system errors propagate — swap must revert
+                    assert!(
+                        result.is_err(),
+                        "Swap should revert when flip hits a system error"
+                    );
+                    assert!(
+                        result.unwrap_err().is_system_error(),
+                        "Error must be classified as a system error",
+                    );
```
0xKitsune, line 4446: nit: can we assert the expected error?

0xKitsune (approved):

PR #2713, opened 2026-02-16
feat: add `t1a` hardfork

0xKitsune:
Introduces the `T1A` hardfork to gate the per-transaction gas limit fix (removal of EIP-7825's ~16.7M cap) behind its own activation timestamp.

### DIFF
--- a/crates/commonware-node/src/executor/actor.rs
+++ b/crates/commonware-node/src/executor/actor.rs
@@ -172,16 +172,16 @@ where
     }
 
     async fn run(mut self) {
-        let mut backfill_on_start = {
-            if self.last_consensus_finalized_height > self.last_execution_finalized_height {
-                info!(
-                    last_finalized_consensus_height = %self.last_consensus_finalized_height,
-                    last_finalized_execution_height = %self.last_execution_finalized_height,
-                    "the last finalized height according to the consensus layer \
-                    is ahead of the execution layer, will backfill blocks",
-                );
-            }
+        info_span!("start").in_scope(|| {
+            info!(
+                last_finalized_consensus_height = %self.last_consensus_finalized_height,
+                last_finalized_execution_height = %self.last_execution_finalized_height,
+                "consensus and execution layers reported last finalized heights; \
+                backfilling blocks from consensus to execution if necessary",
+            );
+        });
 
+        let mut backfill_on_start = {
             let marshal = self.marshal.clone();
             std::pin::pin!(
                 futures::stream::iter(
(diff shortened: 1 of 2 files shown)


crates/chainspec/src/hardfork.rs
```diff
@@ -85,34 +92,38 @@ impl TempoHardfork {
...
-    /// - T1+: 30M gas (allows maximum-sized contract deployments under TIP-1000 state creation)
+    /// - Pre-T1A: u64::MAX (no intentional cap — the EIP-7825 Osaka limit was a bug, not a
+    ///   feature. Must be `Some` to prevent revm from falling back to EIP-7825.)
+    /// - T1A+: 30M gas (allows maximum-sized contract deployments under TIP-1000 state creation)
     pub const fn tx_gas_limit_cap(&self) -> Option<u64> {
         match self {
-            Self::T1 | Self::T2 => Some(crate::spec::TEMPO_T1_TX_GAS_LIMIT_CAP),
-            Self::T0 | Self::Genesis => Some(u64::MAX),
+            Self::T1A | Self::T2 => Some(crate::spec::TEMPO_T1_TX_GAS_LIMIT_CAP),
+            Self::T0 | Self::Genesis | Self::T1 => Some(u64::MAX),
```
0xrusowsky: why are we changing the T1 behavior? shouldn't we add T1A together with T1 adn T2?

klkvr: hmm is this right? it seems that between `T1` and `T1A` there will be no cap on tx gas limit on Testnet?

crates/chainspec/src/hardfork.rs
```diff
@@ -96,23 +103,27 @@ impl TempoHardfork {
     /// - T1+: 30M gas (allows maximum-sized contract deployments under TIP-1000 state creation)
     pub const fn tx_gas_limit_cap(&self) -> Option<u64> {
         match self {
-            Self::T1 | Self::T2 => Some(crate::spec::TEMPO_T1_TX_GAS_LIMIT_CAP),
+            Self::T1 | Self::T1A | Self::T2 => Some(crate::spec::TEMPO_T1_TX_GAS_LIMIT_CAP),
```
klkvr: let's map `T1` to Osaka gas limit? as that's technically the current behavior

0xrusowsky (approved):
makes sense given that testnet is effectively running on `MAX_TX_GAS_LIMIT_OSAKA` due to the cap enforced by its reth version

PR #2714, opened 2026-02-16
feat!(evm): split intrinsic gas between regular and state gas for T2+ (TIP-1016)

fgimenez:
Closes CHAIN-789

Introduces state gas tracking in the EVM handler for the T2 hardfork. State gas separates the cost of state-touching operations (account creation, contract deployment) from execution gas, enabling the reservoir model and tx cap bypass for state-heavy transactions.

`crates/evm/src/lib.rs`:
- Set `cfg_env.enable_state_gas = spec.is_t2()` in both `evm_env` and `next_evm_env`

`crates/revm/src/handler.rs`
- `adjusted_initial_gas`: propagate `initial_state_gas` through `InitialAndFloorGas::new_with_state_gas`
- `calculate_aa_batch_intrinsic_gas`: track `new_account_state_gas + create_state_gas` for AA CREATE calls
- Multi-call gas reconstruction: use flattened `Gas::new_spent` + `erase_cost` pattern (robust under EIP-8037 reservoir model) and accumulate `state_gas_spent` across calls
- Tx gas limit cap bypass: upstream revm already skips the cap check when `enable_state_gas` is true
- `reimburse_caller`: no changes needed — `gas.used()` already includes state gas through the reservoir flattening in `last_frame_result`

### DIFF
--- a/crates/revm/src/handler.rs
+++ b/crates/revm/src/handler.rs
@@ -1283,10 +1283,18 @@ where
             }
 
             // Validate gas limit is sufficient for initial gas
-            if gas_limit < init_gas.initial_total_gas {
+            // State gas is only included in the intrinsic check for T2+, since pre-T2
+            // transactions were never validated against state gas.
+            let total_intrinsic = init_gas.initial_total_gas
+                + if spec.is_t2() {
+                    init_gas.initial_state_gas
+                } else {
+                    0
+                };
+            if gas_limit < total_intrinsic {
                 return Err(TempoInvalidTransaction::InsufficientGasForIntrinsicCost {
                     gas_limit,
-                    intrinsic_gas: init_gas.initial_total_gas,
+                    intrinsic_gas: total_intrinsic,
                 }
                 .into());
             }
@@ -1536,10 +1544,18 @@ where
     }
 
     // Validate gas limit is sufficient for initial gas
-    if gas_limit < batch_gas.initial_total_gas {
+    // State gas is only included in the intrinsic check for T2+, since pre-T2
+    // transactions were never validated against state gas.
+    let total_intrinsic = batch_gas.initial_total_gas
+        + if spec.is_t2() {
+            batch_gas.initial_state_gas
+        } else {
+            0
+        };
+    if gas_limit < total_intrinsic {
         return Err(TempoInvalidTransaction::InsufficientGasForIntrinsicCost {
             gas_limit,
-            intrinsic_gas: batch_gas.initial_total_gas,
+            intrinsic_gas: total_intrinsic,
         }
         .into());
     }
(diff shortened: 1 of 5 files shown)


crates/evm/src/lib.rs
```diff
@@ -131,6 +131,8 @@ impl ConfigureEvm for TempoEvmConfig {
         // Apply TIP-1000 gas params for T1 hardfork.
         let mut cfg_env = cfg_env.with_spec_and_gas_params(spec, tempo_gas_params(spec));
         cfg_env.tx_gas_limit_cap = spec.tx_gas_limit_cap();
+        // TIP-1016: Enable state gas tracking for T2+
+        cfg_env.enable_state_gas = spec.is_t2();
```
rakita, line 135: Additionally to enable_state_gas, we should change variable in `tempo_gas_params`, there we need to remove bumps to 250k for t2 and bump [new gas params variables](https://github.com/bluealloy/revm/pull/3406/changes#r2813461787)

crates/revm/src/handler.rs
```diff
@@ -376,27 +381,32 @@ where
 
                 // Include gas from all previous successful calls + failed call
                 let gas_spent_by_failed_call = frame_result.gas().spent();
+                let failed_call_state_gas = frame_result.gas().state_gas_spent();
```
rakita: On revert or on halt, we should not use state gas. State gas is a cost for creating new state, on revert/halt this will not happen and we dont need to introduce cost for it.

crates/revm/src/handler.rs
```diff
@@ -1384,6 +1395,10 @@ pub fn calculate_aa_batch_intrinsic_gas<'a>(
 
             // EIP-3860: Initcode analysis gas using revm helper
             gas.initial_total_gas += gas_params.tx_initcode_cost(call.input.len());
+
+            // TIP-1016: Track predictable state gas for CREATE calls
+            gas.initial_state_gas +=
+                gas_params.new_account_state_gas() + gas_params.create_state_gas();
```
rakita, line 1432: There is few more places where this is present mostly on AA tx. Auth list, nonces etc.

crates/evm/src/lib.rs
fgimenez: yep done, added constants to define values param variables pre and post T2 ptal

crates/revm/src/handler.rs
fgimenez: good catch, done

crates/revm/src/handler.rs
fgimenez: ok done for:
- auth list entries with nonce==0 (account creation via GasId 254/255)
- bew account creation when nonce==0 (new_account_state_gas)
- 2D nonce + CREATE + caller nonce==0 (via evm.initial_state_gas)

ptal

PR #2715, opened 2026-02-16
feat(consensus): report latest finalized heights on start

SuperFluffy:
With this patch the last finalized height according to consensus and execution is always reported. Before: only if consensus is ahead of execution.

### DIFF
--- a/tips/ref-impls/test/invariants/AccountKeychain.t.sol
+++ b/tips/ref-impls/test/invariants/AccountKeychain.t.sol
@@ -54,8 +54,6 @@ contract AccountKeychainInvariantTest is InvariantBaseTest {
     uint256 private _totalKeysRevoked;
     uint256 private _totalLimitUpdates;
 
-    /// @dev Coverage log file path
-
     /*//////////////////////////////////////////////////////////////
                                SETUP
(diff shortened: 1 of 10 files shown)


crates/commonware-node/src/executor/actor.rs
```diff
@@ -172,16 +172,22 @@ where
...
-                    is ahead of the execution layer, will backfill blocks",
-                );
-            }
+        info!(
+            consensus = %self.last_consensus_finalized_height,
+            execution = %self.last_execution_finalized_height,
+            "latest finalized blocks on startup",
+        );
 
+        info_span!("start").in_scope(|| {
```
hamdiallam, line 175: are logs redundant with the above change? They both log the last consensus/execution heights

hamdiallam (approved):

crates/commonware-node/src/executor/actor.rs
SuperFluffy: I forgot to delete those lines - I moved them into the span (because OTEL etc), but forgot to delete.

Ty for noticing! [e516d60](https://github.com/tempoxyz/tempo/pull/2715/commits/e516d608d18ae98abdd9c20e31a776141358de9a)

PR #2716, opened 2026-02-16
feat(consensus): implement TIP 1017, ValConfigV2

SuperFluffy:
Implements Validator Config V2 at the Consensus Level.

For a given block hash, validators are read from the Validator Config V2 contract if hardfork T2 was active at the block's timestamp and if the Validator Config V2 was initialized (meaning, if the migration from V1 to V2 was completed by the operator).

Overall changes:

All execution layer state is now read at explicit block hashes, not at heights. Either the EL state is requested at a block hash or its parent (for example when constructing or verifying proposals), or the highest available finalized block hash is requested from the EL and its state read at that.

Changes to the individual actors:

## Peer manager

Post T2, on every finalized block received, reads the highest finalized block available in the execution layer. Updates the tracked peer set in the p2p layer if changes were made. Post T2 peers are always updated. The source of peer information depends on whether V2 is initialized.

Because on every finalized block there always a chance for a race between block inclusion in the EL and the peer manager reading it, the peer manager waits for the executor actor to acknowledge the block.

Similarly, peers are refreshed on a timer using the latest available finalized execution layer state. The timer and the update on every finalized block replace the explicit retry mechanism that still exists for pre-T2 behavior.

## Executor actor

Emits an event after having processed a finalized block (always emitted if the block if newPayload returns VALID or SYNCING).

## DKG manager

There are 3 points of interaction between the DKG manager actor and the Validator Config V2 contract:

1. when finalizing the overall DKG outcome, the state at the boundary block's parent is read to see if the next epoch rotates its network identity. The parent block's timestamp is used to determine if the V1 or V2 contracts should be read (taking into account initialization, see above).
2. also when finalizing the DKG outcome, the players of the next epoch are determined: if T2 is active and V2 is initialized, the players are read from the V2 contract. Otherwise the players are taken the *previous* epoch's boundary (see point 3. below; this is pre-T2 behavior).
3. when finalizing the boundary block, the players of the epoch after the next (so not in E+1 but E+2) are read from the boundary block if T2 was not active and V2 initialized at the boundary block's parent(!) -- this is important so that behavior consistent with point 2. above is maintained.

Related https://github.com/tempoxyz/tempo/pull/2635, https://github.com/tempoxyz/tempo/pull/2910

Closes #1465

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5762,9 +5762,9 @@ dependencies = [
 
 [[package]]
 name = "lz4_flex"
-version = "0.12.0"
+version = "0.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ab6473172471198271ff72e9379150e9dfd70d8e533e0752a27e515b48dd375e"
+checksum = "98c23545df7ecf1b16c303910a69b079e8e251d60f7dd2cc9b4177f2afaf1746"
 
 [[package]]
 name = "mach2"
--- a/crates/ext/src/launcher.rs
+++ b/crates/ext/src/launcher.rs
@@ -174,14 +174,14 @@ fn run_tempoup(bin_dir: &Path) -> Result<bool, LauncherError> {
             let install_status = Command::new("sh")
                 .arg("-c")
                 .arg("curl -fsSL https://tempo.xyz/install | bash")
+                .env("TEMPO_BIN_DIR", bin_dir)
                 .status()?;
             if !install_status.success() {
                 tracing::error!("failed to install tempoup");
                 return Ok(false);
             }
-            Command::new("tempoup")
-                .env("TEMPO_BIN_DIR", bin_dir)
-                .status()?
+
+            return Ok(true);
         }
         Err(err) => return Err(LauncherError::Io(err)),
     };


crates/commonware-node/src/validators.rs
```diff
@@ -132,54 +117,221 @@ pub(crate) fn read_validator_config_at_height<T>(
...
+/// 1. if `header.timestamp` is active at the hardfork timestamp.
+/// 2. if `header.number` is equal or greater than the contract initialization height.
+/// 3. if the contract initialization flag is set.
+pub(crate) fn can_use_v2(node: &TempoFullNode, header: &TempoHeader) -> eyre::Result<bool> {
+    Ok(node
+        .chain_spec()
+        .is_t2_active_at_timestamp(header.timestamp())
+        && is_v2_initialized(node, header.number())
+            .wrap_err("failed reading validator config v2 initialization flag")?
+        && v2_initialization_height(node, header.number())
```
joshieDo: couldnt these two be unified somehow? we're instantiating the precompile twice for the same height i think?

crates/e2e/src/tests/backfill.rs
```diff
@@ -84,3 +84,23 @@ fn validator_can_join_later_with_pipeline_sync() {
...
+
+#[track_caller]
+fn assert_no_new_epoch(context: &impl Metrics, max_epoch: u64) {
+    let metrics = context.encode();
+    for line in metrics.lines() {
+        let mut parts = line.split_whitespace();
+        let metric = parts.next().unwrap();
+        let value = parts.next().unwrap();
+
+        if metrics.ends_with("_peers_blocked") {
```
joshieDo:
```suggestion
        if metric.ends_with("_peers_blocked") {
```

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -515,6 +515,10 @@ impl Inner<Init> {
             );
             info!(
                 %outcome.epoch,
+                outcome.network_identity = %outcome.network_identity(),
+                outcome.dealers = ?outcome.dealers(),
+                outcome.dealers = ?outcome.players(),
+                outcome.dealers = ?outcome.next_players(),
```
joshieDo:
```suggestion
                outcome.dealers = ?outcome.dealers(),
                outcome.players = ?outcome.players(),
                outcome.next_players = ?outcome.next_players(),
```

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -1422,6 +1444,34 @@ impl Metrics {
...
+        context.register(
+            "read_players_from_v1_contract",
+            "the number of times the players (or syncers) were read from the validator config v1 contract",
+            read_players_from_v1_contract.clone(),
+        );
+
+        let read_players_from_v2_contract = Counter::default();
+        context.register(
+            "read_players_from_v2_contract",
+            "the number of times the players were read from the validator config v1 contract",
```
joshieDo:
```suggestion
            "the number of times the players were read from the validator config v2 contract",
```

crates/commonware-node/src/peer_manager/actor.rs
```diff
@@ -170,31 +171,41 @@ where
...
             epoch = %onchain_outcome.epoch,
             %last_boundary,
             is_syncing,
-            best_block = %tempo_telemetry_util::display_result(&best_block),
-            peers = peers.len(),
-            "bootstrapped initial peer set from boundary block",
+            read_height,
+            %read_hash,
+            ?peers,
+            "bootstrapped initial peer set from last boundary block and best execution layer bock",
```
joshieDo:
```suggestion
            "bootstrapped initial peer set from last boundary block and best execution layer block",
```

crates/commonware-node/src/peer_manager/actor.rs
```diff
@@ -170,31 +171,41 @@ where
...
+            height: header.number(),
+            peers,
+        };
+        self.oracle
+            .track(
+                last_tracked_peer_seet.height,
+                last_tracked_peer_seet.peers.clone(),
+            )
+            .await;
+        self.last_tracked_peer_set = Some(last_tracked_peer_seet);
```
joshieDo:
```suggestion
        let last_tracked_peer_set = LastTrackedPeerSet {
            height: header.number(),
            peers,
        };
        self.oracle
            .track(
                last_tracked_peer_set.height,
                last_tracked_peer_set.peers.clone(),
            )
            .await;
        self.last_tracked_peer_set = Some(last_tracked_peer_set);
```

crates/commonware-node/src/peer_manager/actor.rs
```diff
@@ -303,21 +352,69 @@ fn construct_peer_set(
...
+) -> eyre::Result<(u64, B256, Validators)> {
+    // TODO: is this too harsh? Should we wait until there is a best block
+    // available at a later point?
+    let best_block_number = node
+        .provider
+        .best_block_number()
+        .wrap_err("provider does not have best block available")?;
+
+    ensure!(
+        best_block_number >= reference_header.number(),
```
joshieDo: are there weird scenarios here?

eg. imagine that validator is observing the network (not catching up), and EL is ahead (non finalized block), we construct the peer set from this block, which then ends up reorged (and lets say without the new validator state). Does it "reset" in the next finalized block?

SuperFluffy: Good point. In #2750 I am actually running a `read_validator_config_at_hash` to prevent this scenario. `read_validator_config` is only used for the best block - but I should rename it to make it clearer.

crates/commonware-node/src/validators.rs
SuperFluffy: Yeah, that's true.

crates/commonware-node/src/peer_manager/actor.rs
```diff
@@ -170,31 +171,41 @@ where
...
+            .header(read_hash)
+            .expect("must be access execution layer to get header - just read validator config for")
+            .expect("execution layer must have the header - just read validator config for it");
+        let last_tracked_peer_set = LastTrackedPeerSet {
+            height: header.number(),
+            peers,
+        };
+        self.oracle
+            .track(
+                last_tracked_peer_set.height,
```
hamdiallam: We need to revisit the `PEERSETS_TO_TRACK` constant as I believe this change will now exhibit different behavior.

Previously, maintained peers 3/4 epochs back.

Now will only maintain peers 3/4 blocks back

SuperFluffy: That is true - although I have to admit that I am not sure what the value of tracking several peersets is at all anymore, to be honest.

Note that it only pushes a new peer set if a new entry was added (usually `addValidator` or `rotateValidator`) or if an entry was removed (`deleteValidator` followed by a validator dropping out of the peer set). If the peer set remains stable but the IPs change, then the latest set is just overwritten.

hamdiallam: For additions, makes sense to not need to track backwards.

For removals, `active=false` and will be removed from the latest peerset. However this validator still needs to participate in DKG until being fully removed? If there are a couple of updates (add/removes) then this validator would prematurely get evicted?

SuperFluffy: Oh no, we *always* take into account those peers that are mentioned in the last DKG outcome (copied this from the PR because linking to github diffs is miserable):

```rust
    let all_keys = outcome
        .dealers()
        .iter()
        .chain(outcome.next_players().iter())
        .chain(match validators {
            Validators::V1(validators) => Either::Left(
                validators
                    .iter_pairs()
                    .filter_map(|(k, v)| v.is_active().then_some(k)),
            ),
            Validators::V2(validators) => Either::Right(
                validators
                    .iter_pairs()
                    .filter_map(|(k, v)| v.is_active().then_some(k)),
            ),
        });
```
^ This ensures that we *always* track the validators mentioned in the DKG outcome, as well as whichever validators are active as per the on-chain state.

With that said, there is a bug - I should be chaining `outcome.players()` not `outcome.dealers()` (these are the dealers that helped construct the outcome, giving new shares to the players).

so good thing we had a second look at that.

SuperFluffy: See [378622c](https://github.com/tempoxyz/tempo/pull/2716/commits/378622c2fda98755e3d2d10eba9e4ddeb726fb6c)

hamdiallam: Gotcha! makes sense

crates/commonware-node/src/peer_manager/actor.rs
hamdiallam: Should we pull those changes into this PR?

SuperFluffy: I can. I decided to not push into this PR so there is no churn while reviewing.

hamdiallam: Then we can remove all non-blockhash based reads?

SuperFluffy: They are still valuable for startup.

But ye - probably makes sense to go over all reads and determine if reading heights is ok

hamdiallam (approved):
LGTM!

If we add some metrics on peerset address updates. This is different than the number of peers, then we should also be able to e2e test address updates being processed when a validator updates their IP onchain

crates/commonware-node/src/validators.rs
```diff
@@ -1,133 +1,169 @@
...
-                });
+        }
+        for member in known {
+            if !all.contains_key(member) {
+                let decoded = config
+                    .validator_by_public_key(public_key_to_b256(member))
+                    .map_err(eyre::Report::new)
+                    .and_then(DecodedValidatorV2::decode_from_contract)
+                    .expect(
+                        "invariant: known peers must have an entry in the \
```
decofe, line 126: 🚨 **[SECURITY] Chain Halt via Panicking on Missing V2 Validators**

`read_active_and_known_peers_at_block_hash_v2` iterates `known` peers (DKG participants from the boundary outcome) and hard-panics here if any peer is missing from V2. However, `ValidatorConfigV2::migrate_validator` explicitly skips some active V1 validators for duplicate IPs or invalid formatting. If a skipped validator was a DKG participant, all upgrading nodes will panic simultaneously on first post-T2 peer refresh, permanently halting the network.

**Recommended Fix:** Replace `.expect()` with graceful error handling — log a warning and omit the missing validator instead of panicking:
```rust
match config
    .validator_by_public_key(public_key_to_b256(member))
    .map_err(eyre::Report::new)
    .and_then(DecodedValidatorV2::decode_from_contract)
{
    Ok(decoded) => { all.insert(decoded.public_key.clone(), decoded.to_address()); }
    Err(e) => { warn!(validator = %member, error = %e, "known peer missing or malformed in V2"); }
}
```

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -1168,30 +1159,32 @@ where
...
+            .is_ok_and(|epoch| epoch == next_epoch.get());
+        info!(
+            will_be_re_dkg,
+            %next_epoch,
+            "determined if the next epoch will be a reshare or full re-dkg process",
+        );
 
-        if request
+        let next_players =
+            determine_next_players_at_hash(state, &self.config.execution_node, request.digest.0)
```
decofe, line 1173: 🚨 **[SECURITY] Exact-hash DKG Reads Make Boundary Acceptance Depend on EL Materialization Timing**

`determine_next_players_at_hash()` unconditionally calls `can_use_v2_at_block_hash()`, which does `provider.header(hash)` and errors if the hash is unknown — even on pre-T2/T1C paths. Boundary verification calls `verify_header_extra_data()` before canonicalizing the parent into EL. If the parent block is not yet materialized, this error prevents reaching the pre-T2 `state.syncers` fallback, the response channel is dropped, and the block is failed.

Two honest nodes processing the same valid boundary block can disagree based solely on EL materialization timing. This is a regression: the old pre-T2 path returned `state.syncers` directly without any EL exact-hash lookup.

**Recommended Fix:** Don't gate the pre-T2 fallback on an exact-hash EL lookup. `determine_next_players_at_hash` should return `state.syncers` on pre-T2/T1C without first calling `can_use_v2_at_block_hash`. Restore a pending-aware fallback for boundary-time reads.

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -1248,40 +1241,55 @@ where
...
+                latest_boundary.saturating_sub(1)
+            )
         })?;
 
+    let onchain_outcome = tempo_dkg_onchain_artifacts::OnchainDkgOutcome::read(
+        &mut boundary_header.extra_data().as_ref(),
+    )
+    .wrap_err("the boundary header did not contain the on-chain DKG outcome")?;
+
+    let syncers = read_syncers_if_v2_not_initialized(
```
decofe, line 1293: ⚠️ **[ISSUE] DKG Bootstrap Can Die Silently When Exact Historical State Is Not Queryable**

On empty DKG storage, `read_initial_state_and_set_floor()` uses the non-retrying `read_syncers_if_v2_not_initialized()` here. That calls `state_by_block_hash()`, which only succeeds if the exact hash is available as historical or pending state in reth. On fresh join/snapshot bootstrap, the DKG actor can exit during initialization before starting its mailbox service loop, while the application layer still expects it to be alive.

**Recommended Fix:** Use `read_syncers_if_v2_not_initialized_with_retry()` during initial bootstrap (same as the live boundary handling path), or fail fast with a clear fatal error instead of silently dropping the DKG actor.

crates/commonware-node/src/peer_manager/actor.rs
```diff
@@ -207,102 +148,304 @@ where
...
+            onchain_outcome
+                .players()
+                .iter()
+                .cloned()
+                .chain(onchain_outcome.next_players().iter().cloned()),
+        );
+        let peers = read_active_and_known_peers_at_block_hash(
+            &self.execution_node,
+            &peers_as_per_dkg,
+            highest_finalized_header.hash_slow(),
```
decofe, line 301: ⚠️ **[ISSUE] Bootstrap Peer Refresh Reads Stale Finalized State Instead of Boundary State**

`refresh_peers()` derives `known` members from boundary DKG (`players` + `next_players`) but reads validator contract state at `highest_finalized_header.hash_slow()`, not the boundary hash. In V1 mode, the reader only retains entries currently in `get_validators()` with no backfill. Since V1 `update_validator` permits mid-epoch key/address mutation, a boundary-era DKG participant can be absent from the newer finalized view, causing the tracked peer set to diverge from boundary membership on restart.

**Recommended Fix:** Read validator state at the boundary reference hash, or explicitly merge missing `known` members from boundary-consistent state.

crates/commonware-node/src/peer_manager/actor.rs
```diff
@@ -207,102 +148,304 @@ where
...
-            .expect(
-                "all DKG participants must have an entry in the \
-                 unfiltered, contract validator set",
+#[instrument(skip_all, fields(height), err)]
+fn read_highest_finalized_header(execution_node: &TempoFullNode) -> eyre::Result<TempoHeader> {
+    let highest_finalized = execution_node
+        .provider
+        .finalized_block_hash()
+        .map_err(eyre::Report::new)
+        .and_then(|h| h.ok_or_eyre("execution layer has no record of any finalization hashes"))?;
```
decofe: 🛡️ **[DEFENSE-IN-DEPTH] Missing Finalized Block Hash Returns Error During Bootstrap**

Converting `None` into an error here causes `refresh_peers()` to fail silently on the timer path during bootstrap. The DKG bootstrap path already handles this correctly via `unwrap_or_else(|| BlockNumHash::new(0, genesis_hash()))`.

**Recommended Fix:** Fall back to the genesis block hash when `finalized_block_hash()` returns `None`, consistent with the DKG pattern.

crates/commonware-node/src/validators.rs
SuperFluffy: This is an invariant that needs to be upheld. Before finishing the migration, we must make sure that all validators have an entry in V2.

crates/commonware-node/src/dkg/manager/actor/mod.rs
SuperFluffy: `determine_next_players_at_hash` is called on the *request* digest. This request is sent by the application actor (either when proposing or verifying) and the digest is that of the proposal's parent. Therefore, in order to verify or propose, the execution layer *must* have header available (otherwise `sendNewPayload` when proposing or `newPayload` when verifying would fail).

So:

1. get header (must work both post- and pre-T2).
2. check if header is active at T2 (works both post- and pre-T2)0
3. `&&` in Rust is lazily evaluated: V2 is not read if T2 is not active.

```rust
pub(crate) fn can_use_v2_at_block_hash(
    node: &TempoFullNode,
    hash: B256,
    latest: Option<B256>,
) -> eyre::Result<bool> {
    let header = node
        .provider
        .header(hash)
        .map_err(eyre::Report::new)
        .and_then(|maybe| maybe.ok_or_eyre("hash not known"))
        .wrap_err_with(|| {
            format!("failed reading header for block hash `{hash}` from execution layer")
        })?;
    let state_hash = latest.unwrap_or(hash);
    Ok(node
        .chain_spec()
        .is_t2_active_at_timestamp(header.timestamp())
        && is_v2_initialized_at_block_hash(node, state_hash)
            .wrap_err("failed reading validator config v2 initialization flag")?
        && v2_initialization_height_at_block_hash(node, state_hash)
            .wrap_err("failed reading validator config v2 initialization height")?
            <= header.number())
}
```

crates/commonware-node/src/peer_manager/actor.rs
SuperFluffy: This is correct but corrected by us not providing operators the ability to change their public key, and us (the contract owner) never changing the public key ourselves.

This design error is what required the introduction of the V2 contract.

crates/commonware-node/src/peer_manager/actor.rs
SuperFluffy: That is valid, fixed in 7d2cf59e

crates/commonware-node/src/dkg/manager/actor/mod.rs
SuperFluffy: Clarified in 183e06ca why this is a fast fail.

crates/commonware-node/src/validators.rs
SuperFluffy: Should be unified, but I will leave that as a follow-up.

crates/commonware-node/src/dkg/manager/actor/mod.rs
SuperFluffy: Removed this metric; seemed unnecessary.

crates/commonware-node/src/peer_manager/actor.rs
SuperFluffy: There are no more non-blockhash based reads.

PR #2729, opened 2026-02-16
feat(cmd): add consensus check-peers connectivity diagnostic

### LINKED ISSUE #2400: Tool to test connectivity to other peers
### Describe the feature

We need a tool to debug connectivity issues between a node and its peers. Given the on-chain data (last DKG at an epoch boundary + onchain validator set) we can check if all validators from a node can be reached when run on a machine.

Even if firewalls are configured correctly such that outgoing and incoming connections are permitted, if a node operates behind a proxy it can happen that an IP check on the TCP socket of the incoming connection will be that of the proxy, not that of the dialing peer.

In addition, supplying the tool with the ed25519 consensus signing key would mean that a p2p handshake with the other (running nodes) could be made, ensuring that we don't just check the connection with the proxy.

### Additional context

_No response_

HedgeQuantX:
## Summary

Adds `tempo consensus check-peers` — a standalone diagnostic tool that verifies P2P connectivity and **validator identity** for every peer in the active set. Performs the full `commonware-stream` encrypted handshake (Syn → SynAck → Ack) against each validator, proving the correct ed25519 key is behind the endpoint.

Directly addresses #2400.

## Motivation

Validators behind proxies, NATs, or with misconfigured firewalls need to verify which peers are reachable and — critically — whether the correct validator is actually responding. A TCP probe alone cannot distinguish a working validator from a dumb proxy that forwards bytes. The encrypted handshake proves the remote holds the correct ed25519 private key via the signed SynAck.

## Usage

```bash
# Check all active validators on moderato
tempo consensus check-peers \
  --chain moderato \
  --rpc-url https://rpc.moderato.tempo.xyz \
  --signing-key /path/to/signing.key

# Check a specific validator
tempo consensus check-peers \
  --chain moderato \
  --signing-key ./key.hex \
  --validator <pubkey_hex>

# Custom timeout
tempo consensus check-peers \
  --chain moderato \
  --signing-key ./key.hex \
  --timeout-secs 15
```

## Output (JSON)

```json
{
  "chain": "Moderato Testnet",
  "current_epoch": 42,
  "total_validators": 8,
  "reachable": 7,
  "unreachable": 1,
  "verified": 6,
  "peers": [
    {
      "public_key": "abcd...",
      "inbound_address": "1.2.3.4:8000",
      "active": true,
      "in_committee": true,
      "reachable": true,
      "identity_verified": true,
      "handshake_latency_ms": 24
    },
    {
      "public_key": "ef01...",
      "inbound_address": "5.6.7.8:8000",
      "active": true,
      "in_committee": true,
      "reachable": true,
      "identity_verified": false,
      "handshake_latency_ms": null,
      "error": "handshake failed: handshake timed out"
    }
  ]
}
```

## Implementation

1. Loads the operator's ed25519 signing key from `--signing-key`
2. Connects to RPC, queries `getValidators()` on the `ValidatorConfig` precompile
3. Reads the DKG outcome from the epoch boundary block for committee membership
4. Bootstraps a `commonware_runtime::tokio` context (provides `BufferPooler + CryptoRngCore + Clock + Network`)
5. For each validator, spawns a concurrent task that:
   - Opens a raw TCP connection via `context.dial()`
   - Upgrades to an encrypted channel via `commonware_stream::encrypted::dial()` using the operator's signing key and the exact stream namespace (`union(union_unique(b"TEMPO", b"_P2P"), b"_STREAM")`)
   - The remote's bouncer accepts the dialer's key (since it's in the validator set), and the full handshake verifies the remote's identity via the signed SynAck
6. Collects results into structured JSON with `reachable` (TCP) and `identity_verified` (handshake) fields

The `--signing-key` must belong to an active validator — the remote bouncer rejects unknown public keys before sending SynAck (encrypted.rs L246), so non-validator keys cannot probe identity.

## Changes

- `Cargo.toml`: Add `commonware-stream` workspace dependency
- `bin/tempo/Cargo.toml`: Add `commonware-stream` to tempo binary
- `bin/tempo/src/tempo_cmd.rs`: Rewrite `check-peers` from TCP probe to encrypted handshake

Closes #2400

### DIFF
--- a/bin/tempo/src/tempo_cmd.rs
+++ b/bin/tempo/src/tempo_cmd.rs
@@ -650,3 +650,236 @@ async fn check_handshake(
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use commonware_codec::Encode as _;
+    use commonware_runtime::{Listener as _, Network as _, Runner as _, Spawner as _};
+    use commonware_stream::encrypted;
+
+    /// Spin up a real TCP listener that performs `encrypted::listen` with
+    /// `listener_key`, then run `check_handshake` against it using
+    /// `dialer_key`. Returns the `PeerCheckResult`.
+    fn run_handshake_test(
+        listener_key: PrivateKey,
+        dialer_key: PrivateKey,
+        expected_peer_key: &PublicKey,
+        accept_dialer: bool,
+    ) -> PeerCheckResult {
+        let namespace = stream_namespace();
+        let key_hex = alloy_primitives::hex::encode(expected_peer_key.encode());
+        let ns = namespace.clone();
+
+        commonware_runtime::tokio::Runner::default().start(|ctx| async move {
+            // Bind a TCP listener on an ephemeral port.
+            let mut tcp_listener = ctx
+                .bind(std::net::SocketAddr::from(([127, 0, 0, 1], 0)))
+                .await
+                .expect("bind failed");
+            let addr = tcp_listener.local_addr().expect("local_addr failed");
+            let addr_str = addr.to_string();
+
+            let listener_ns = ns.clone();
+            let dialer_pub = dialer_key.public_key();
+
+            // Spawn the listener side: accept one TCP connection, then run
+            // the encrypted handshake as if we were a validator node.
+            let listener_handle = ctx.clone().spawn(move |inner_ctx| async move {
+                let (_peer_addr, sink, stream) = tcp_listener.accept().await.unwrap();
+                let cfg = encrypted::Config {
+                    signing_key: listener_key,
+                    namespace: listener_ns,
+                    max_message_size: 16 * 1024 * 1024,
+                    synchrony_bound: Duration::from_secs(5),
+                    max_handshake_age: Duration::from_secs(10),
+                    handshake_timeout: Duration::from_secs(5),
+                };
+                let accept = accept_dialer;
+                let expected = dialer_pub;
+                encrypted::listen(
+                    inner_ctx,
+                    move |peer: PublicKey| {
+                        let expected = expected.clone();
+                        async move { accept && peer == expected }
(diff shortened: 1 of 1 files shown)


SuperFluffy:
Now write a test to demonstrate that this works.

HedgeQuantX:
Added four integration tests in `54bb04e` covering the handshake behavior end-to-end.

Each test spins up a real TCP listener running `encrypted::listen` with the Tempo stream namespace, then exercises `check_handshake` against it as if probing a live validator. No mocks for the network layer — actual TCP bind/connect/accept through `commonware_runtime::tokio`:

- **`test_handshake_success`** — correct key behind the endpoint, bouncer accepts. Verifies `identity_verified=true` and latency is recorded.
- **`test_handshake_rejected_by_bouncer`** — correct key but bouncer rejects us (simulates our key not being in the peer set). Reports `reachable=true`, `identity_verified=false`.
- **`test_handshake_unreachable`** — nothing on the port. Reports `reachable=false` with dial failure.
- **`test_handshake_wrong_key`** — listener holds a different ed25519 key than what the dialer expects. Handshake fails at the crypto level, proving the check catches proxy/key mismatch scenarios.

```
running 4 tests
test tempo_cmd::tests::test_handshake_rejected_by_bouncer ... ok
test tempo_cmd::tests::test_handshake_wrong_key ... ok
test tempo_cmd::tests::test_handshake_success ... ok
test tempo_cmd::tests::test_handshake_unreachable ... ok

test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.06s
```

The wrong-key test is the important one — it proves that if someone puts a proxy or the wrong validator behind an IP, the SynAck signature check fails and `identity_verified` comes back false. That's the whole point of using `encrypted::dial` over raw TCP.

PR #2733, opened 2026-02-17
feat: update val config v2 genesis

howydev:
Genesis file is updated with val config v2 initialization

When `t2_time == 0`, we use v2 validator config from genesis and skip v1 entirely. We call `initialize_v2` in this case 

Otherwise, v1 validator config is used. We initialize v1 val config with some values, and monitor block_time to check for t2_time. at t2_time we etch 0xEF into v2 config's address

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -3,7 +3,37 @@ fail-fast = false
 retries = 2
 test-threads = "num-cpus"
 
+# tempo-e2e tests spawn multiple execution nodes, each creating 4 rayon thread
+# pools (cpu, rpc, trie, storage) via reth's Runtime. Even with test_with_handle()
+# (2 threads per pool), running many in parallel causes OOM on CI.
+# threads-required = 8 means nextest counts each test as needing 8 of the -j slots,
+# so on a 32-vCPU runner (-j 32) at most 4 e2e tests run concurrently.
 [[profile.ci.overrides]]
 filter = "package(tempo-e2e)"
 failure-output = "never"
-threads-required = 2
+threads-required = 8
+
+# tempo-node integration tests (pool, block_building, etc.) also launch full
+# execution nodes with rayon pools. Limit to 4 slots so at most 8 run in parallel.
+[[profile.ci.overrides]]
+filter = "package(tempo-node) & binary(it)"
+threads-required = 4
+
+[[profile.ci.overrides]]
+filter = "test(sync::can_restart_after_joining_from_snapshot)"
+slow-timeout = { period = "60s", terminate-after = 4, on-timeout = "pass" }
+success-output = "final"
+
+# Local development defaults (same constraints apply)
+[[profile.default.overrides]]
+filter = "package(tempo-e2e)"
+threads-required = 8
+
+[[profile.default.overrides]]
+filter = "package(tempo-node) & binary(it)"
+threads-required = 4
+
+[[profile.default.overrides]]
+filter = "test(sync::can_restart_after_joining_from_snapshot)"
+slow-timeout = { period = "60s", terminate-after = 4, on-timeout = "pass" }
+success-output = "final"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -83,7 +83,7 @@ jobs:
   test:
     name: test
     needs: check-precompiles
-    runs-on: depot-ubuntu-latest-16
+    runs-on: depot-ubuntu-latest-32
     timeout-minutes: 30
     permissions:
       contents: read
@@ -169,7 +169,7 @@ jobs:
           persist-credentials: false
       - uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: "1.91" # MSRV
+          toolchain: "1.93" # MSRV
       - uses: rui314/setup-mold@725a8794d15fc7563f59595bd9556495c0564878 # v1
       - uses: mozilla-actions/sccache-action@7d986dd989559c6ecdb630a3fd2557667be217ad # v0.0.9
       - name: Build with MSRV
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -121,9 +121,9 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "alloy"
-version = "1.6.1"
+version = "1.6.3"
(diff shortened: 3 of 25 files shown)


crates/chainspec/src/genesis/moderato.json
```diff
@@ -23,6 +23,7 @@
     "t0Time": 1770303600,
     "t1Time": 1770303600,
     "t1aTime": 1771513200,
+    "t2Time": 1771513200,
```
howydev: these vals are from `main` branch i believe

xtask/src/genesis_args.rs
```diff
@@ -923,13 +941,115 @@ fn initialize_validator_config(
...
+                );
+            }
+            Ok(())
+        },
+    )
+}
+
+/// Deploys the `0xef` marker bytecode to the V2 precompile address so the EVM
+/// recognises it as a precompile, without initialising any storage.
+fn deploy_validator_config_v2_bytecode(evm: &mut TempoEvm<CacheDB<EmptyDB>>) {
```
0xrusowsky: why do we need this fn? this is already handled by calling `fn initialize()`

if no migration is needed u pass `needs_migration = false` which sets the bytecode and initialize storage.

if migration is needed later on u pass `needs_migration = true`  which only sets the bytecode, but doesn't init storage and will require a later migration

crates/node/tests/assets/test-genesis.json
```diff
@@ -1766,12 +1766,12 @@
...
     },
-    "0xcccccccc00000000000000000000000000000000": {
+    "0xcccccccc00000000000000000000000000000001": {
       "nonce": "0x0",
       "balance": "0x0",
       "code": "0xef",
       "storage": {
-        "0x0000000000000000000000000000000000000000000000000000000000000000": "0x000000000000000000000000f39fd6e51aad88f6f4ce6ab8827279cfffb92266"
+        "0x0000000000000000000000000000000000000000000000000000000000000000": "0x000000000000000000000001f39fd6e51aad88f6f4ce6ab8827279cfffb92266"
```
0xrusowsky, line 1775: what are we doing here exactly?

crates/chainspec/src/spec.rs
```diff
@@ -446,11 +446,28 @@ mod tests {
...
         );
 
+        // At and after T2 activation
+        assert!(mainnet_chainspec.is_t2_active_at_timestamp(1771858800));
+        assert_eq!(
+            mainnet_chainspec.tempo_hardfork_at(1771858800),
+            TempoHardfork::T2
+        );
+        assert_eq!(
+            mainnet_chainspec.tempo_hardfork_at(1771858801),
```
0xrusowsky: this one is redundant

crates/node/tests/assets/test-genesis.json
howydev: this must be the initialized bit packed with owner

crates/chainspec/src/genesis/moderato.json
SuperFluffy: I think we need to actually change this to T3 now? CC @jenpaff

xtask/src/genesis_args.rs
```diff
@@ -923,13 +941,115 @@ fn initialize_validator_config(
...
+                return Err(eyre!(
+                    "need {} addresses for all validators, but only {} were provided",
+                    num_validators,
+                    onchain_validator_addresses.len()
+                ));
+            }
+
+            println!("writing {num_validators} validators into v2 contract");
+            for (i, validator) in consensus_config.validators.iter().enumerate() {
+                #[expect(non_snake_case, reason = "field of a snakeCase smart contract call")]
```
SuperFluffy: Don't do this. Way too much noise for questionable gain. Just use rust convention and assign the name in the struct below.

xtask/src/genesis_args.rs
```diff
@@ -923,13 +941,115 @@ fn initialize_validator_config(
...
+) -> eyre::Result<()> {
+    let ctx = evm.ctx_mut();
+    StorageCtx::enter_evm(
+        &mut ctx.journaled_state,
+        &ctx.block,
+        &ctx.cfg,
+        &ctx.tx,
+        || {
+            let mut v2 = ValidatorConfigV2::new();
+            v2.initialize(admin, false)
```
SuperFluffy: Confusing argument. :-)

howydev: Zooming out a little - if even in this case of v2 on genesis we're using `false`, then it's meaningless to keep the `true` branch. I'll update the init function and remove that

SuperFluffy: Just curious - what is the intention of having the initializer in the first place? It's for assigning an owner, right?

What's unclear to me is how `fn initialize` and `fn initialize_if_migrated` interact. Is the latter misnamed? Do we always have to `initialize` to assign an owner, and then `initialize_if_migrated` to finish migration?

howydev: `initialize` is not intended to be callable outside of genesis, just makes it easy to set up validators in v2 without going through v1 if we start the chain without v1

howydev: (which is what xtask is doing, since `t2_time == 0` there)

xtask/src/genesis_args.rs
```diff
@@ -930,13 +948,115 @@ fn initialize_validator_config(
...
+            for (i, validator) in consensus_config.validators.iter().enumerate() {
+                #[expect(non_snake_case, reason = "field of a snakeCase smart contract call")]
+                let validatorAddress = onchain_validator_addresses[i];
+                let public_key = validator.public_key();
+                let pubkey: B256 = public_key.encode().as_ref().try_into().unwrap();
+                let addr = validator.addr;
+                let ingress = addr.to_string();
+                let egress = addr.ip().to_string();
+
+                // message: keccak256(chainId || contractAddr || validatorAddr || ingress || egress)
```
SuperFluffy, line 1000: This is not the same as what's written in the spec, cf:

```
keccak256(abi.encodePacked(bytes8(chainId), contractAddress, validatorAddress, ingress, egress))
```

I am fine with changing this, but then we should adjust the spec.

howydev: packed mode encoding just does concat so this should be equivalent

howydev: for peace of mind - `add_validator` signs a digest generated this same way, and that test passes

PR #2734, opened 2026-02-17
feat(evm): split gas tracking into execution vs storage creation gas (TIP-1016)

fgimenez:
Closes CHAIN-791

Adds block-level plumbing to track execution gas and storage creation gas separately in `TempoBlockExecutor`. 

- Added `cumulative_full_gas_used` and `cumulative_storage_creation_gas` fields to `TempoBlockExecutor`
- `commit_transaction` now computes `execution_gas = full_gas - storage_creation_gas` and only execution gas counts toward block gas limits (`non_shared_gas_left`, `non_payment_gas_left`)
- Receipts and `cumulative_full_gas_used` continue to track full gas (execution + storage creation)
- `finish()` returns execution gas (not full gas) for the block header gas field
- `storage_creation_gas` is hardcoded to `0` with a TODO — it will be populated once handler-level tracking lands (CHAIN-789 / PR #2714)

Note: hardfork gating is not required. With `storage_creation_gas = 0`, all gas accounting is identical to the current behavior. The hardfork gate (T2) lives in the handler layer, not here. This is pure plumbing that becomes active only when upstream EVM tracking is enabled.

### DIFF
--- a/crates/consensus/src/lib.rs
+++ b/crates/consensus/src/lib.rs
@@ -3,7 +3,7 @@
 #![cfg_attr(not(test), warn(unused_crate_dependencies))]
 #![cfg_attr(docsrs, feature(doc_cfg))]
 
-use alloy_consensus::{BlockHeader, Transaction, transaction::TxHashRef};
+use alloy_consensus::{BlockHeader, Transaction, TxReceipt, transaction::TxHashRef};
 use alloy_evm::block::BlockExecutionResult;
 use reth_chainspec::EthChainSpec;
 use reth_consensus::{Consensus, ConsensusError, FullConsensus, HeaderValidator};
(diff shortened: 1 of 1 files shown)


crates/evm/src/block.rs
```diff
@@ -114,6 +114,13 @@ pub(crate) struct TempoBlockExecutor<'a, DB: Database, I> {
     non_shared_gas_left: u64,
     non_payment_gas_left: u64,
     incentive_gas_used: u64,
+
+    /// Tracks full cumulative gas used (execution + storage creation) for receipts.
+    /// This differs from block gas limit accounting which only counts execution gas.
+    cumulative_full_gas_used: u64,
```
rakita: `cumulative_total_gas_used` total naming is used in tip1016

crates/evm/src/block.rs
```diff
@@ -402,7 +411,17 @@ where
         let TempoTxResult { inner, tx } = output;
         let next_section = self.validate_tx(&tx, inner.result.result.gas_used())?;
 
-        let gas_used = self.inner.commit_transaction(inner)?;
+        // TODO(TIP-1016): extract storage creation gas from EVM context once
+        // the EVM-level tracking is implemented.
+        let storage_creation_gas: u64 = 0;
+
+        let full_gas_used = self.inner.commit_transaction(inner)?;
```
rakita: I see, we need to change alloy/evm traits

crates/evm/src/block.rs
fgimenez: ok changed `total_gas` -> `full_gas`

PR #2736, opened 2026-02-17
fix(evm): remove system tx presence check from executor finish()

gakonst:
## Summary

Remove the `BlockSection` validation from `TempoBlockExecutor::finish()` that unconditionally required blocks to end with system transactions. This fixes `eth_simulateV1` returning `-32603: end-of-block system transactions not seen` on every call.

## Motivation

`eth_simulateV1` runs user-provided transactions through `TempoBlockExecutor` and calls `finish()`, but simulated blocks never include Tempo's end-of-block system transactions (subblock signatures/metadata), so the check always failed.

System transaction presence is already validated by `TempoConsensus::validate_block_pre_execution()` for real blocks, making the executor check redundant.

## Changes

- Removed `BlockSection::System { seen_subblocks_signatures: true }` guard from `TempoBlockExecutor::finish()` — now delegates directly to `self.inner.finish()`
- Removed `test_finish_system_tx_not_seen` test (no longer applicable)
- Simplified `test_finish` to not require pre-setting the section

## Testing

```
cargo test -p tempo-evm --lib block  # 23 passed
cargo clippy -p tempo-evm -p tempo-consensus -- -D warnings  # clean
```

Prompted by: arsenii


crates/evm/src/block.rs
```diff
@@ -466,15 +466,6 @@ where
     fn finish(
         self,
     ) -> Result<(Self::Evm, BlockExecutionResult<Self::Receipt>), BlockExecutionError> {
-        if self.section
```
mattsse, line 502: if we remove this, is this field then even useful?

this isnt documented so i dont really know how this is used

klkvr: yeah this field is used in `validate_tx` to track section transitions:

https://github.com/tempoxyz/tempo/blob/174ae3244e2dd370e8f9107c40731a885e95a6c3/crates/evm/src/block.rs#L289-L349

so the idea for this PR is that we already check on `TempoConsensus` side that system tx is present:
https://github.com/tempoxyz/tempo/blob/174ae3244e2dd370e8f9107c40731a885e95a6c3/crates/consensus/src/lib.rs#L167-L171

and thus executor only needs to enforce that this system tx contents are correct. given that consenus validation is skipped in `simulateV1`, this should unblock it

PR #2740, opened 2026-02-17
feat!(txpool,payload): track intrinsic state gas in pool and exclude it from protocol gas limits (TIP-1016)

fgimenez:
Closes CHAIN-790
Builds on top of #2714 

Computes and caches intrinsic state gas during transaction validation so the payload builder can exclude it from protocol gas limits, ensuring storage creation gas only counts against the user's gas_limit.

### Transaction pool 

- Add `intrinsic_state_gas: OnceLock<u64>` to `TempoPooledTransaction` and expose `set_intrinsic_state_gas()`, `intrinsic_state_gas()`, and `execution_gas_limit()` (= `gas_limit - state_gas`).
- During validation, compute and cache intrinsic state gas for T2+ transactions (both AA and non-AA paths). Sources include new-account creation, CREATE calls, 2D nonce keys, and EIP-7702 auth list entries.
- For pre-T2, `intrinsic_state_gas` stays 0 so `execution_gas_limit() == gas_limit()`.

### Payload builder 

- Use `execution_gas_limit()` instead of `gas_limit()` for protocol limit checks (`non_shared_gas_limit`, `general_gas_limit`). State gas only counts against the user's `gas_limit`, not the block's execution gas capacity.

### Test utilities 

- Fix `TxBuilder` to use `self.chain_id` instead of hardcoded `1` when building AA transactions.
- Add `chain_id()` setter method.

### DIFF
--- a/crates/revm/src/handler.rs
+++ b/crates/revm/src/handler.rs
@@ -1282,8 +1282,15 @@ where
                 init_gas.floor_gas = 0u64;
             }
 
-            // Validate gas limit is sufficient for initial gas (execution + state)
-            let total_intrinsic = init_gas.initial_total_gas + init_gas.initial_state_gas;
+            // Validate gas limit is sufficient for initial gas
+            // State gas is only included in the intrinsic check for T2+, since pre-T2
+            // transactions were never validated against state gas.
+            let total_intrinsic = init_gas.initial_total_gas
+                + if spec.is_t2() {
+                    init_gas.initial_state_gas
+                } else {
+                    0
+                };
             if gas_limit < total_intrinsic {
                 return Err(TempoInvalidTransaction::InsufficientGasForIntrinsicCost {
                     gas_limit,
@@ -1536,8 +1543,15 @@ where
         batch_gas.initial_total_gas += nonce_2d_gas;
     }
 
-    // Validate gas limit is sufficient for initial gas (execution + state)
-    let total_intrinsic = batch_gas.initial_total_gas + batch_gas.initial_state_gas;
+    // Validate gas limit is sufficient for initial gas
+    // State gas is only included in the intrinsic check for T2+, since pre-T2
+    // transactions were never validated against state gas.
+    let total_intrinsic = batch_gas.initial_total_gas
+        + if spec.is_t2() {
+            batch_gas.initial_state_gas
+        } else {
+            0
+        };
     if gas_limit < total_intrinsic {
         return Err(TempoInvalidTransaction::InsufficientGasForIntrinsicCost {
             gas_limit,


crates/transaction-pool/src/validator.rs
```diff
@@ -683,6 +699,30 @@ where
...
+                    non_aa_state_gas += gas_params.get(GasId::new_account_state_gas())
+                        + gas_params.get(GasId::create_state_gas());
+                }
+                // EIP-7702 auth list account creation state gas
+                for auth in transaction.authorization_list().unwrap_or_default() {
+                    if auth.nonce == 0 {
+                        non_aa_state_gas += gas_params.tx_tip1000_auth_account_creation_state_gas();
+                    }
+                }
+                transaction.set_intrinsic_state_gas(non_aa_state_gas);
```
rakita: `state_gas+regular_gas <= gas_limit` do we have this check

crates/payload/builder/src/lib.rs
```diff
@@ -369,16 +369,21 @@ where
...
+            // TIP-1016: For T2+ transactions, use execution_gas_limit (gas_limit minus state gas)
+            // for protocol limit checks. State gas does NOT count against protocol limits.
+            // For pre-T2, execution_gas_limit() == gas_limit() (state gas is 0).
+            let tx_execution_gas = pool_tx.transaction.execution_gas_limit();
+
             // Ensure we still have capacity for this transaction within the non-shared gas limit.
             // The remaining `shared_gas_limit` is reserved for validator subblocks and must not
             // be consumed by proposer's pool transactions.
-            if cumulative_gas_used + pool_tx.gas_limit() > non_shared_gas_limit {
+            if cumulative_gas_used + tx_execution_gas > non_shared_gas_limit {
```
mattsse: can we use consistent terminology here, gas vs gaslimit is always needlessly confusing

crates/payload/builder/src/lib.rs
```diff
@@ -369,16 +369,21 @@ where
 
         let execution_start = Instant::now();
         while let Some(pool_tx) = best_txs.next() {
+            // TIP-1016: For T2+ transactions, use execution_gas_limit (gas_limit minus state gas)
+            // for protocol limit checks. State gas does NOT count against protocol limits.
+            // For pre-T2, execution_gas_limit() == gas_limit() (state gas is 0).
```
mattsse: this is the builder, but I dont see a protocol check that's enforcing this

crates/transaction-pool/src/transaction.rs
```diff
@@ -44,6 +44,12 @@ pub struct TempoPooledTransaction {
     /// `Some(expiry)` for keychain transactions where expiry < u64::MAX (finite expiry).
     /// `None` for non-keychain transactions or keys that never expire.
     key_expiry: OnceLock<Option<u64>>,
+    /// Intrinsic state gas (TIP-1016 storage creation gas component).
+    ///
+    /// Set during validation for T2+ transactions. The state gas portion does NOT count
+    /// against protocol limits (block/transaction execution gas caps), only against
+    /// the user's `gas_limit`.
+    intrinsic_state_gas: OnceLock<u64>,
```
mattsse, line 58: isnt this problematic, because the intrinsic gas is hardfork specific, right?

so it's possible that a tx survives a hardfork boundary, making this incorrect fpr future blocks?

mattsse (commented):
@rakita upcoming revm now returns intrinsic gas as part of the evm output right? which means tracking this in the pooled tx is no longer necessary after we bump?

crates/transaction-pool/src/validator.rs
fgimenez: good catch, added check ptal

crates/payload/builder/src/lib.rs
fgimenez: indeed, updated comment

crates/payload/builder/src/lib.rs
fgimenez: yep, changed to `tx_execution_gas_limit`

crates/transaction-pool/src/transaction.rs
fgimenez: good catch, added comment about it, lmk if that is enough

PR #2744, opened 2026-02-17
fix: scoped epoch metrics

hamdiallam:
1. Bump CW to pickup commonwarexyz/monorepo#3127
2. CW now leverages a shared buffered pool amongst the caches
3. Reduce `POOL_CAPACITY` used in the DKG actor's state from 2^20 to 2^13. This reduces the overall memory consumption from 4GiB to 32MiB, fixing tests failing with OOM panics.

Explanation: Commonware, instead of requesting one large contiguous chunk of memory from the kernel, now requests many small chunks. This means that now CW's allocation is below a threshold triggering lazy memory allocation, instead eagerly allocating. This causes issues in tests were several actors are spun up.

### DIFF
--- a/crates/commonware-node/src/dkg/manager/actor/state.rs
+++ b/crates/commonware-node/src/dkg/manager/actor/state.rs
@@ -33,7 +33,7 @@ use tracing::{debug, info, instrument, warn};
 use crate::consensus::{Digest, block::Block};
 
 const PAGE_SIZE: NonZeroU16 = NZU16!(1 << 12);
-const POOL_CAPACITY: NonZeroUsize = NZUsize!(1 << 20);
+const POOL_CAPACITY: NonZeroUsize = NZUsize!(1 << 13);
 const WRITE_BUFFER: NonZeroUsize = NZUsize!(1 << 12);
 const READ_BUFFER: NonZeroUsize = NZUsize!(1 << 20);
 


crates/commonware-node/src/epoch/manager/actor.rs
```diff
@@ -322,7 +323,8 @@ where
         let engine = simplex::Engine::new(
             self.context
                 .with_label("simplex")
-                .with_attribute("epoch", epoch),
+                .with_attribute("epoch", epoch)
+                .with_scope(),
```
hamdiallam: Scoped metrics per epoch

Cargo.toml
```diff
@@ -325,16 +327,15 @@ vergen-git2 = "9.1.0"
 # reth-trie-sparse = { path = "../reth/crates/trie/sparse" }
 # reth-trie-sparse-parallel = { path = "../reth/crates/trie/sparse-parallel" }
 
-# [patch.crates-io]
-# # Commonware right after after PR #3018 was merged
-# commonware-broadcast = { git = "https://github.com/commonwarexyz/monorepo", rev = "a1b60388c0bbc9126aa9934ec6ebade6ec344552" }
-# commonware-codec = { git = "https://github.com/commonwarexyz/monorepo", rev = "a1b60388c0bbc9126aa9934ec6ebade6ec344552" }
```
SuperFluffy, line 331: Please add the note so that it's easier to locate/understand which commit this refers to.

crates/commonware-node/src/epoch/manager/actor.rs
SuperFluffy: In their upstream PR, commonware keeps the context around so that it does not go out of scope and the registry is kept alive.

Is that not necessary here?

hamdiallam: The simplex engine takes ownership of the context so it shouldn't get dropped at the end of the epoch being entered.

Looking over upstream, the archive, [example](https://github.com/commonwarexyz/monorepo/blob/main/consensus/src/marshal/cache.rs#L231), also holds reference to the copied context so it seems like an explicit choice to manage the clone.

We can do the same here but I don't think it's necessary since the `ScopeGuard` ensures that all clones have to be dropped in order for the cleanup logic to fire.

SuperFluffy: But in their reshare example (which we use) they hold on to it: https://github.com/commonwarexyz/monorepo/blob/6addcbf5ed3d048a6736990542134827724b088b/examples/reshare/src/orchestrator/actor.rs#L342

that's why I am asking.

But you are right, the simplex engine itself actually holds on to the context. Hmmm:

https://github.com/commonwarexyz/monorepo/blob/6addcbf5ed3d048a6736990542134827724b088b/consensus/src/simplex/engine.rs#L113

Cargo.toml
hamdiallam: Whoops! Good catch

PR #2748, opened 2026-02-17
feat(tip-1004): add EIP-2612 permit to TIP-20 precompile

howydev:
Implements TIP-1004: adds `permit()`, `nonces()`, and `DOMAIN_SEPARATOR()` to TIP-20 precompile (EIP-2612).

- Gated behind T2 hardfork
- Includes Solidity reference implementation
- Comprehensive test coverage

### DIFF
--- a/crates/precompiles/src/tip20/mod.rs
+++ b/crates/precompiles/src/tip20/mod.rs
@@ -551,6 +551,7 @@ impl TIP20Token {
         );
 
         // 4. Validate ECDSA signature
+        // Only v=27/28 is accepted; v=0/1 is intentionally NOT normalized (see TIP-1004 spec).
         if call.v != 27 && call.v != 28 {
             return Err(TIP20Error::invalid_signature().into());
         }
@@ -558,7 +559,7 @@ impl TIP20Token {
         let sig = Signature::from_scalars_and_parity(call.r, call.s, parity);
         let recovered = alloy::consensus::crypto::secp256k1::recover_signer(&sig, digest)
             .map_err(|_| TIP20Error::invalid_signature())?;
-        if recovered != call.owner {
+        if recovered.is_zero() || recovered != call.owner {
             return Err(TIP20Error::invalid_signature().into());
         }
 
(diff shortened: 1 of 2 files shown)


crates/precompiles/src/tip20/mod.rs
```diff
@@ -477,6 +490,97 @@ impl TIP20Token {
...
+            [
+                &[0x19, 0x01],
+                domain_separator.as_slice(),
+                struct_hash.as_slice(),
+            ]
+            .concat(),
+        );
+
+        // 5. Validate signature (EOA only, no EIP-1271)
+        let parity = call.v.wrapping_sub(27) != 0;
```
howydev: hmm, does this allow shenanigans like setting this byte to `0` to save 12 gas when `v` is `27`?

tips/ref-impls/src/TIP20.sol
```diff
@@ -330,6 +335,51 @@ contract TIP20 is ITIP20, TIP20RolesAuth {
...
+        uint8 v,
+        bytes32 r,
+        bytes32 s
+    )
+        external
+    {
+        if (block.timestamp > deadline) revert PermitExpired();
+
+        bytes32 structHash = keccak256(
+            abi.encode(PERMIT_TYPEHASH, owner, spender, value, nonces[owner]++, deadline)
```
fgimenez, line 370: very nit, minor inconsistency with the rust impl, this uses unchecked ++ versus `checked_add` in rust

crates/precompiles/src/tip20/dispatch.rs
```diff
@@ -741,4 +768,50 @@ mod tests {
             Ok(())
         })
     }
+
+    #[test]
+    fn test_permit_selectors_gated_behind_t2() -> eyre::Result<()> {
+        use alloy::sol_types::SolError;
+        use tempo_contracts::precompiles::UnknownFunctionSelector;
```
fgimenez: can we move the `use` statements to the enclosing module?

tips/ref-impls/src/TIP20.sol
howydev: ah in solidity 0.8+ all math ops are checked by default

crates/precompiles/src/tip20/mod.rs
```diff
@@ -477,6 +490,97 @@ impl TIP20Token {
...
+            ]
+            .concat(),
+        );
+
+        // 4. Validate ECDSA signature
+        if call.v != 27 && call.v != 28 {
+            return Err(TIP20Error::invalid_signature().into());
+        }
+        let parity = call.v == 28;
+        let sig = Signature::from_scalars_and_parity(call.r, call.s, parity);
```
fgimenez, line 559: this does not enforce low-s signatures, in other parts of the codebase we use either explicit checks https://github.com/tempoxyz/tempo/blob/main/crates/primitives/src/transaction/tt_signature.rs#L709-L713 or rely on `alloy_consensus::crypto::secp256k1::recover_signer` which has a low-s guard built-in https://github.com/tempoxyz/tempo/blob/main/crates/primitives/src/transaction/tt_signature.rs#L251-L254

howydev: good catch, fixed

howydev: hmm on 2nd thought this should be ok - permit's `nonce` prevents sig replay, so even if we allowed malleability, i think an attacker wouldn't be able to do anything. but better to fix just to be sure

tips/ref-impls/test/invariants/TIP20.t.sol
```diff
@@ -1227,6 +1233,86 @@ contract TIP20InvariantTest is InvariantBaseTest {
...
+
+            // **TEMPO-TIP34**: A permit with a deadline in the past must always revert.
+            assertGe(
+                deadline, block.timestamp, "TEMPO-TIP34: Permit should revert if deadline is past"
+            );
+
+            // **TEMPO-TIP35**: The recovered signer from a valid permit signature must exactly match the `owner` parameter.
+            assertEq(
+                ecrecover(digest, v, r, s),
+                signer,
```
howydev:
```suggestion
                actor,
```

PR #2750, opened 2026-02-17
feat(e2e): test static V1 -> V2 migrations, add full test suite for V2 val config at genesis

SuperFluffy:
Adds tests for DKG scenarios that involve a static set of validators and have a transition from the V1 to V2 validator configs.

Also adds a mechanism to wait for the executor to send a finalized block at a given height to the execution layer - this is to fix the problem of the peer manager attempting to read EL state before the block was submitted to the EL.

### DIFF
--- a/crates/e2e/src/lib.rs
+++ b/crates/e2e/src/lib.rs
@@ -92,7 +92,7 @@ fn generate_consensus_node_config(
                     ingress: SocketAddr::from(([127, 0, 0, (i + 1) as u8], 8000)),
                     egress: SocketAddr::from(([127, 0, 0, (i + 1) as u8], 0)),
                     private_key,
-                    share: shares.get_value(&&public_key).cloned(),
+                    share: shares.get_value(&public_key).cloned(),
                 };
                 (public_key, config)
             }),
--- a/crates/e2e/src/execution_runtime.rs
+++ b/crates/e2e/src/execution_runtime.rs
@@ -1225,7 +1225,7 @@ impl From<AddValidatorV2> for Message {
 
 impl From<ChangeValidatorStatus> for Message {
     fn from(value: ChangeValidatorStatus) -> Self {
-        Self::ChangeValidatorStatus(value.into())
+        Self::ChangeValidatorStatus(value)
     }
 }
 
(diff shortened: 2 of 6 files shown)


crates/commonware-node/src/executor/ingress.rs
```diff
@@ -31,6 +31,19 @@ impl Mailbox {
 
         Ok(rx)
     }
+
+    pub(crate) async fn subscribe_finalized(&self, height: Height) -> eyre::Result<()> {
```
SuperFluffy, line 35: Added this to fix the issue of the peer actor wanting to read the finalized block before it was processed by the executor.

crates/e2e/src/lib.rs
```diff
@@ -179,7 +179,7 @@ pub async fn setup_validators(
     // The port here does not matter because it will be ignored in simulated p2p.
     // Still nice, because sometimes nodes can be better identified in logs.
     let network_addresses = (1..)
-        .map(|port| SocketAddr::from(([127, 0, 0, 1], port)))
+        .map(|i| SocketAddr::from(([127, 0, 0, i], 8000)))
```
SuperFluffy: Addresses are now uniquely owned by validators (port doesn't matter).

crates/commonware-node/src/dkg/manager/actor/state.rs
```diff
@@ -766,14 +766,13 @@ impl Read for State {
...
         cfg: &Self::Cfg,
     ) -> Result<Self, commonware_codec::Error> {
-        let range_cfg = RangeCfg::from(1..=(u16::MAX as usize));
         Ok(Self {
             epoch: ReadExt::read(buf)?,
             seed: ReadExt::read(buf)?,
             output: Read::read_cfg(buf, cfg)?,
             share: ReadExt::read(buf)?,
-            players: Read::read_cfg(buf, &(range_cfg, ()))?,
-            syncers: Read::read_cfg(buf, &(range_cfg, ()))?,
```
SuperFluffy, line 776: Once transitioned to V2, syncers can be 0.

This is not a format breaking change! The `RangeCfg` acts as a check for the encoded length.

Of course once a node has transitioned to V2 it can no longer run in V1. But that's ok.

hamdiallam (approved):

PR #2760, opened 2026-02-18
fix(stablecoin-dex): move `T2` gate to `T1A`

zerosnacks:
Ref: https://github.com/tempoxyz/tempo/pull/2654/
Ref: https://github.com/tempoxyz/tempo/pull/2711/

This PR moves the gate in https://github.com/tempoxyz/tempo/pull/2711 under the `T1A` hardfork


0xrusowsky (approved):
let's triple check with @0xKitsune just in case before merging

crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -4434,8 +4434,8 @@ mod tests {
 
                 let result = exchange.swap_exact_amount_in(bob, base_token, quote_token, amount, 0);
 
-                if spec.is_t2() {
-                    // T2: system errors propagate — swap must revert
+                if spec.is_t1a() {
```
0xKitsune, line 4437: This needs to be gated behind T1B

0xKitsune (approved):

PR #2763, opened 2026-02-18
feat: add EIP-2935 historical block hashes contract to generated genesis

mattsse:
Adds the [EIP-2935](https://eips.ethereum.org/EIPS/eip-2935) historical block hashes contract (`0x0000F90827F1C53a10cb7A02335B175320002935`) to the genesis allocation in `GenesisArgs::generate_genesis()`.

This affects all newly generated genesis files (both `generate-genesis` and `generate-devnet` xtask commands). Existing deployed networks are not affected.


xtask/src/genesis_args.rs
```diff
@@ -491,6 +460,16 @@ impl GenesisArgs {
             },
         );
 
+        // EIP-2935: Historical block hashes contract
+        genesis_alloc.insert(
+            address!("0x0000F90827F1C53a10cb7A02335B175320002935"),
+            GenesisAccount {
+                code: Some("0x3373fffffffffffffffffffffffffffffffffffffffe14604657602036036042575f35600143038111604257611fff81430311604257611fff9006545f5260205ff35b5f5ffd5b5f35611fff60014303065500".parse().unwrap()),
```
SuperFluffy: Can you provide an `expect` msg instead why this should never fail? Unclear why this should work for people without context.

I see the address in the EIP you linked, but not the code bytes.

PR #2764, opened 2026-02-18
feat(txpool): add global cap to paused fee token pool to prevent unbounded growth

fgimenez:
Ref CHAIN-605

### DIFF
--- a/crates/transaction-pool/src/paused.rs
+++ b/crates/transaction-pool/src/paused.rs
@@ -119,7 +119,7 @@ impl PausedFeeTokenPool {
             .iter()
             .map(|(addr, meta)| (*addr, meta.paused_at))
             .collect();
-        tokens_by_age.sort_by_key(|(_, paused_at)| *paused_at);
+        tokens_by_age.sort_unstable_by_key(|(_, paused_at)| *paused_at);
 
         let mut evicted = 0;
         for (token, _) in tokens_by_age {


crates/transaction-pool/src/paused.rs
```diff
@@ -63,18 +70,69 @@ impl PausedFeeTokenPool {
...
+        }
+
+        let remaining_capacity = PAUSED_POOL_GLOBAL_CAP.saturating_sub(self.len());
+        let to_insert = if incoming > remaining_capacity {
+            entries.into_iter().take(remaining_capacity).collect()
+        } else {
+            entries
+        };
+
+        if !to_insert.is_empty() {
```
mattsse: this checks seems a bit redundant

crates/transaction-pool/src/paused.rs
```diff
@@ -63,18 +70,69 @@ impl PausedFeeTokenPool {
...
+    /// Evicts at least `need` entries from the oldest-paused tokens.
+    ///
+    /// Returns the total number of entries evicted.
+    fn evict_oldest(&mut self, need: usize) -> usize {
+        let mut tokens_by_age: Vec<_> = self
+            .by_token
+            .iter()
+            .map(|(addr, meta)| (*addr, meta.paused_at))
+            .collect();
+        tokens_by_age.sort_by_key(|(_, paused_at)| *paused_at);
```
mattsse: this can sort unstable I believe

crates/transaction-pool/src/paused.rs
fgimenez: not needed at this point, removed

crates/transaction-pool/src/paused.rs
fgimenez: yep, no need to preserve relative order, using `sort_unstable_by_key` now

PR #2771, opened 2026-02-19
feat: sharpen tip-1020 standard

howydev:
I think a large reason to do this precompile is to allow smart contracts to forward compatibility with future Tempo account types

Changes:
1. Remove `verifyStateful` - this makes more sense as an addition in the access key precompile
2. Improved the gas model of the precompile and added clarification around cost
3. Improved invariants quality

### DIFF
--- a/tips/tip-1020.md
+++ b/tips/tip-1020.md
@@ -36,24 +36,12 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 interface ISignatureVerifier {
     error InvalidSignature();
 
-    /// @notice Verifies a stateless Tempo signature (secp256k1, P256, WebAuthn).
-    ///         Does NOT support Keychain signatures.
+    /// @notice Verifies a Tempo signature (secp256k1, P256, WebAuthn).
     /// @param signer The expected signer address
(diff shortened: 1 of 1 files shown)


tips/tip-1020.md
```diff
@@ -18,7 +18,7 @@ This TIP introduces a signature verification precompile that enables contracts t
 
 Tempo supports multiple signature schemes beyond standard secp256k1. Currently, contracts cannot verify Tempo signatures onchain without implementing custom verification logic for each signature type.
 
-This precompile exposes Tempo's native signature verification to contracts, reusing the existing audited verification logic from Tempo transaction processing.
+Additionally, since smart contracts have to statically bind their verification logic at deployment time, they cannot natively support future Tempo account signature schemes introduced after deployment. This precompile serves as a stable interface that smart contracts can use to maintain forward compatibility with future Tempo account types and signature schemes.
```
howydev:
```suggestion
Additionally, since smart contracts have to statically bind their verification logic at deployment time, developers cannot maintain forward compatibility with future Tempo account signature schemes introduced after deployment without making their contracts upgradeable. This precompile serves as a stable interface that smart contracts can use to maintain forward compatibility with future Tempo account types and signature schemes.
```

PR #2778, opened 2026-02-19
fix: disable T2 hardfork by default in dev/devnet genesis

gakonst:
## Summary

Disables T2 hardfork by default in dev mode and devnet genesis generation, so nightly devnets are on par with testnet.

## Motivation

T2 was enabled at genesis (`t2Time=0`) in `dev.json` and as the default for `cargo x generate-devnet`. This caused T2 to be active on nightly devnets even though there's no T2 timestamp for testnet/mainnet yet. Testing T1A/T1B changes on devnets with T2 accidentally enabled can cause confusion.

## Changes

- `crates/chainspec/src/genesis/dev.json`: Set `t2Time` to `4102444800` (2100-01-01)
- `xtask/src/genesis_args.rs`: Change `--t2-time` default from `0` to `4102444800`
- `crates/chainspec/src/spec.rs`: Update dev chainspec test to expect T1B at genesis

To explicitly enable T2 on a devnet, pass `--t2-time 0`.

## Testing

`cargo test -p tempo-chainspec` — all 15 tests pass.

Prompted by: kuyziss

### DIFF
--- a/xtask/src/genesis_args.rs
+++ b/xtask/src/genesis_args.rs
@@ -163,8 +163,9 @@ pub(crate) struct GenesisArgs {
     #[arg(long, default_value = "0")]
     t1b_time: u64,
 
-    /// T2 hardfork activation time.
-    #[arg(long, default_value = "0")]
+    /// T2 hardfork activation time. Defaults to far future (disabled).
+    /// Set to 0 to enable at genesis.
+    #[arg(long, default_value = "4102444800")]
     t2_time: u64,
 
     /// Whether to skip initializing validator config v1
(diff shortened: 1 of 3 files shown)


xtask/src/genesis_args.rs
```diff
@@ -163,8 +163,9 @@ pub(crate) struct GenesisArgs {
     #[arg(long, default_value = "0")]
     t1b_time: u64,
 
-    /// T2 hardfork activation time.
-    #[arg(long, default_value = "0")]
+    /// T2 hardfork activation time. Defaults to far future (disabled).
+    /// Set to 0 to enable at genesis.
+    #[arg(long, default_value = "4102444800")]
```
SuperFluffy, line 168: Maybe just `u64::MAX`?

SuperFluffy (approved):

jenpaff (approved):

PR #2780, opened 2026-02-19
feat: add p2p-proxy subcommand to tempo binary

mattsse:
## Summary

Add a new `tempo p2p-proxy` subcommand that runs a proxy P2P node serving cached block data fetched from an RPC endpoint. This replaces the previous sidecar-based implementation by integrating directly into the main tempo binary.

## Features

- **Block caching**: Caches up to 1 day of blocks (172,800 at 500ms block time) with LRU eviction
- **Request serving**: Handles `GetBlockHeaders` and `GetBlockBodies` with automatic RPC fallback for cache misses
- **Transport flexibility**: Supports HTTP and WebSocket RPC transports via `connect()`
- **CLI args**: `--rpc-url`, `--chain`, `--port`, `--discovery-port`, `--max-inbound`, `--max-concurrent-inbound`
- **Monitoring**: Periodic stats logging (peer count, headers/bodies served every 30s)
- **Robustness**: Always responds to peers even on internal failures (never drops response channel)
- **Inbound-only**: Zero outbound connections, tx gossip disabled

## Usage

```bash
tempo p2p-proxy --rpc-url https://rpc.moderato.tempo.xyz --chain moderato
tempo p2p-proxy --rpc-url ws://localhost:8546 --port 9999 --max-inbound 50
```

## Tests

- CLI parser tests for default args, all args, and missing required args
- E2E tests against live Moderato RPC (headers, bodies, hash-based fetch)


bin/tempo/src/p2p_proxy.rs
```diff
@@ -0,0 +1,600 @@
...
+    // Resolve start block number
+    let start_num = match request.start_block {
+        BlockHashOrNumber::Number(n) => Some(n),
+        BlockHashOrNumber::Hash(h) => cache.get_by_hash(&h).map(|block| block.header.number()),
+    };
+
+    let Some(mut current) = start_num else {
+        return headers;
+    };
+
```
klkvr: is this right? if we get `BlockHashOrNumber::Hash` and we don't have a block with this hash cached we'd just return an empty vec

mattsse: yeah this is wrong

bin/tempo/src/p2p_proxy.rs
```diff
@@ -0,0 +1,600 @@
...
+    for _ in 0..request.limit {
+        let block = if let Some(b) = cache.get_by_number(current) {
+            Some(b.clone())
+        } else {
+            if fetch_and_cache_block(provider, cache, current)
+                .await
```
klkvr: `limit` here would usually be ~10k which'd mean 10k rpc requests sent one by one? should we have a helper endpoint for this that allows fetching batches of headers?

klkvr (approved):

PR #2782, opened 2026-02-19
chore: move `init_defaults` to `tempo-node` crate for easier re-use in other binaries

zerosnacks:
Move `init_download_urls`, `init_payload_builder_defaults`, `init_txpool_defaults` and the public `init_defaults` function from `bin/tempo` into `crates/node/src/defaults.rs`.

Re-export `init_defaults` from `tempo_node` lib.rs alongside `init_version_metadata`.

Telemetry CLI types (`TelemetryArgs`, `UrlWithAuth`, `TelemetryConfig`) remain in the tempo binary as they are CLI-specific.


0xrusowsky (approved):
i don't see any problem with this, but let's wait for more reviewers

PR #2786, opened 2026-02-19
fix(tip-1004): reject zero-address ecrecover in permit, add edge-case tests

howydev:
## Summary

Addresses [ZELLIC-53](https://linear.app/tempoxyz/issue/ZELLIC-53): audit findings on TIP-1004 permit implementation.

### Changes

**Bug fix — zero-address recovery rejection**

`permit()` now explicitly rejects `recovered == address(0)` before comparing against `owner`, matching the Solidity reference `ecrecover` behavior. Previously, a crafted signature that somehow recovered to `address(0)` could have been accepted if `owner` was also `address(0)`.

**New tests**

- `test_permit_zero_address_recovery_reverts` — verifies `InvalidSignature` is returned when ecrecover yields the zero address
- `test_permit_domain_separator_changes_with_chain_id` — verifies the EIP-712 domain separator differs across chain IDs (fork safety)

### DIFF
--- a/crates/precompiles/src/tip20/mod.rs
+++ b/crates/precompiles/src/tip20/mod.rs
@@ -2603,6 +2603,42 @@ pub(crate) mod tests {
         }
 
         #[test]
+        fn test_permit_invalid_v_values() -> eyre::Result<()> {
+            let PermitFixture {
+                mut storage,
+                admin,
+                spender,
+                ..
+            } = PermitFixture::new();
+
+            StorageCtx::enter(&mut storage, || {
+                let mut token = TIP20Setup::create("Test", "TST", admin).apply()?;
+
+                for v in [0u8, 1] {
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/tip20/mod.rs
```diff
@@ -551,14 +551,15 @@ impl TIP20Token {
         );
 
         // 4. Validate ECDSA signature
+        // Only v=27/28 is accepted; v=0/1 is intentionally NOT normalized (see TIP-1004 spec).
```
fgimenez, line 554: maybe good to add a test that covers the v=0 and v=1 path calling `permit()`, something like `test_permit_invalid_v_values` wdyt?

howydev: added!

PR #2787, opened 2026-02-19
feat: lower default time-to-build-proposal from 500ms to 200ms

gakonst:
Lower the default `--consensus.time-to-build-proposal` from 500ms to 200ms.

Steady <1s block time, can fit 9-10k transactions into a block.

image

Prompted by: alexey

### DIFF
--- a/crates/commonware-node/src/args.rs
+++ b/crates/commonware-node/src/args.rs
@@ -101,7 +101,7 @@ pub struct Args {
     /// This value should be well below `consensus.wait-for-proposal` to account
     /// for the leader to enter the view, build and broadcast the proposal, and
     /// have the other peers receive the proposal.
-    #[arg(long = "consensus.time-to-build-proposal", default_value = "500ms")]
+    #[arg(long = "consensus.time-to-build-proposal", default_value = "200ms")]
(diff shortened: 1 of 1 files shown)


crates/commonware-node/src/args.rs
```diff
@@ -101,7 +101,7 @@ pub struct Args {
     /// This value should be well below `consensus.wait-for-proposal` to account
     /// for the leader to enter the view, build and broadcast the proposal, and
     /// have the other peers receive the proposal.
-    #[arg(long = "consensus.time-to-build-proposal", default_value = "500ms")]
+    #[arg(long = "consensus.time-to-build-proposal", default_value = "200ms")]
```
SuperFluffy: What worries me is the effect that this might have on subblocks (see below).

We only give it 100ms. While 100ms < 200ms this probably means that subblocks would only land under optimal network conditions and with virtually no clock drift.

SuperFluffy (approved):
Approving but would like @klkvr to leave a comment on how they expect this to interact with subblocks.

shekhirin:
Superseded by https://github.com/tempoxyz/tempo/pull/2939

PR #2789, opened 2026-02-20
perf: zero-copy RLP decode fast path for block deserialization

gakonst:
## Summary
Avoids a full memcpy of the RLP-encoded block during network deserialization on the hot livesync/backfill path.

## Motivation
During livesync, every block received from the network goes through `Block::read_cfg` which calls `buf.copy_to_bytes(len)` to copy the entire RLP payload into a new allocation before decoding. In the common case where the buffer is already contiguous (typical for network frames), this copy is unnecessary.

## Changes
- Added a fast path in `Block::read_cfg` (`crates/commonware-node/src/consensus/block.rs`) that checks if `buf.chunk()` already contains the full RLP item contiguously
- When contiguous, decodes directly from the slice and advances the buffer manually — no allocation or copy
- Falls back to existing `copy_to_bytes` behavior for fragmented buffers
- Added strict consumption validation (`slice.is_empty()` after decode) in both paths

## Testing
- `cargo clippy -p tempo-commonware-node --all-targets -- -D warnings` — clean
- `cargo fmt --all --check` — clean

Prompted by: yk

### DIFF
--- a/crates/commonware-node/src/consensus/block.rs
+++ b/crates/commonware-node/src/consensus/block.rs
@@ -80,12 +80,42 @@ impl Read for Block {
             commonware_codec::Error::Wrapped("reading RLP header", rlp_err.into())
         })?;
 
-        if header.length_with_payload() > buf.remaining() {
+        let total_len = header.length_with_payload();
+        if total_len > buf.remaining() {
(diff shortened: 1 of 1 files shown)


gakonst:
Closing — meant to investigate this in upstream reth, not Tempo.

PR #2806, opened 2026-02-22
feat: tempo download --chain moderato

jenpaff:
## Summary

`tempo download --chain moderato` was downloading the mainnet snapshot because `DownloadDefaults` had `default_chain_aware_base_url: None`, so every chain fell back to the hardcoded mainnet URL.

Setting default_chain_aware_base_url, so the chain ID from --chain gets appended to the base URL dynamically instead of always falling through to the hardcoded mainnet default.

Tested and verified on my dev machine: 
- `tempo download` → `snapshots.tempoxyz.dev/4217` (mainnet default)
- `tempo download --chain moderato` → `snapshots.tempoxyz.dev/42431`
- `tempo download --chain mainnet` → `snapshots.tempoxyz.dev/4217`


jenpaff:
> lets do testnet as moderato as well

should we do before andantino deprecation? i think rn `--chain testnet` 

> lets do testnet as moderato as well

I think rn testnet still points to andantino, right? Might make sense to do this once we've deprecated andantino

klkvr (approved):

PR #2807, opened 2026-02-23
fix: avoid re-walking epochs when generating transition proofs

hamdiallam:
- Always cache even if full=false. Data can be used on both queries.
- Avoid walking if identity hasn't changed. Just forward the marker
- Unrelated bug - When finding the identity of the requested epoch, we should used the oldest transition. The transitions are newest->oldest so [this](https://github.com/tempoxyz/tempo/blob/185025ab3487b05d478cde9b56bf664a310d1b15/crates/commonware-node/src/feed/state.rs#L160) returns the wrong identity (newest)

### DIFF
--- a/crates/commonware-node/src/feed/state.rs
+++ b/crates/commonware-node/src/feed/state.rs
@@ -9,16 +9,22 @@ use commonware_consensus::{
     marshal::ingress::mailbox::Identifier,
     types::{Epoch, Epocher as _, FixedEpocher, Height},
 };
+use commonware_cryptography::bls12381::primitives::variant::{MinSig, Variant};
 use parking_lot::RwLock;
+use reth_provider::HeaderProvider as _;
 use reth_rpc_convert::transaction::FromConsensusHeader;
 use std::sync::{Arc, OnceLock};
 use tempo_alloy::rpc::TempoHeaderResponse;
 use tempo_dkg_onchain_artifacts::OnchainDkgOutcome;
-use tempo_node::rpc::consensus::{
-    CertifiedBlock, ConsensusFeed, ConsensusState, Event, IdentityProofError, IdentityTransition,
-    IdentityTransitionResponse, Query, TransitionProofData,
+use tempo_node::{
+    TempoFullNode,
+    rpc::consensus::{
+        CertifiedBlock, ConsensusFeed, ConsensusState, Event, IdentityProofError,
+        IdentityTransition, IdentityTransitionResponse, Query, TransitionProofData,
+    },
 };
 use tokio::sync::broadcast;
+use tracing::instrument;
 
 const BROADCAST_CHANNEL_SIZE: usize = 1024;
 
@@ -56,6 +62,7 @@ pub struct FeedStateHandle {
     state: Arc<RwLock<FeedState>>,
     marshal: Arc<OnceLock<marshal::Mailbox>>,
     epocher: Arc<OnceLock<FixedEpocher>>,
+    execution_node: Arc<OnceLock<TempoFullNode>>,
     events_tx: broadcast::Sender<Event>,
     /// Cache for identity transition proofs to avoid re-walking the chain.
     identity_cache: Arc<RwLock<Option<IdentityTransitionCache>>>,
@@ -75,6 +82,7 @@ impl FeedStateHandle {
             })),
             marshal: Arc::new(OnceLock::new()),
             epocher: Arc::new(OnceLock::new()),
+            execution_node: Arc::new(OnceLock::new()),
             events_tx,
             identity_cache: Arc::new(RwLock::new(None)),
         }
@@ -88,6 +96,11 @@ impl FeedStateHandle {
     /// Set the epocher for epoch boundary calculations. Should only be called once.
     pub(crate) fn set_epocher(&self, epocher: FixedEpocher) {
         let _ = self.epocher.set(epocher);
+    }
+
+    /// Set the execution node for header lookups. Should only be called once.
+    pub(crate) fn set_execution_node(&self, execution_node: TempoFullNode) {
+        let _ = self.execution_node.set(execution_node);
     }
 
     /// Get the broadcast sender for events.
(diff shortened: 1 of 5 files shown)


crates/commonware-node/src/feed/state.rs
```diff
@@ -130,49 +130,156 @@ impl FeedStateHandle {
...
-        // Filter transitions to only include those at or before start_epoch
-        let transitions: Vec<_> = cache
-            .transitions
-            .iter()
-            .filter(|t| t.transition_epoch <= start_epoch)
-            .cloned()
-            .collect();
+    ) -> Result<(), IdentityProofError> {
+        // Check if cache already covers this epoch
+        let cached = self.identity_cache.read().clone();
```
SuperFluffy, line 158: Why clone?

crates/commonware-node/src/feed/state.rs
```diff
@@ -130,49 +130,156 @@ impl FeedStateHandle {
...
+        if let Some(ref cache) = cached
```
SuperFluffy: Please no `ref`. Very ai code:
```suggestion
    if let Some(cache) = &cached
```

crates/commonware-node/src/feed/state.rs
```diff
@@ -130,49 +130,156 @@ impl FeedStateHandle {
...
+            && start_epoch <= cache.from_epoch
```
SuperFluffy: Use `(a..=b).contains(c)`

crates/commonware-node/src/feed/state.rs
```diff
@@ -130,49 +130,156 @@ impl FeedStateHandle {
...
-        };
+        // Identity active at epoch N is set by the last block of epoch N-1
+        // (epoch 0 uses its own last block - genesis identity)
+        let epoch_outcome = get_outcome(marshal, epocher, start_epoch.saturating_sub(1)).await?;
+        let epoch_pubkey = *epoch_outcome.sharing().public();
+        let epoch_identity = hex::encode(epoch_pubkey.encode());
+
+        // Fast path: if the identity matches the cached one, no new transitions
+        // occurred — just extend the cache's from_epoch forward.
+        if let Some(ref cache) = cached
```
SuperFluffy: Ref

crates/commonware-node/src/feed/state.rs
```diff
@@ -130,49 +130,156 @@ impl FeedStateHandle {
         }
     }
 
-    /// Serve identity transition proof from cache, returning a subsection if needed.
-    fn serve_from_cache(
+    /// Ensure the identity cache covers `start_epoch` by walking backwards
+    /// if needed. After this returns, the cache is guaranteed to contain
+    /// transition data covering `start_epoch` (as far back as available data allows).
+    async fn try_fill_transitions(
```
SuperFluffy, line 150: Slap an `#[instrument(skip_all, fields(start_epoch), err)]` on this

crates/commonware-node/src/feed/state.rs
```diff
@@ -130,49 +130,156 @@ impl FeedStateHandle {
...
-                .iter()
-                .find(|t| t.transition_epoch > start_epoch)
-                .map(|t| t.old_identity.clone())
-                .unwrap_or_else(|| cache.identity.clone())
-        };
+        // Identity active at epoch N is set by the last block of epoch N-1
+        // (epoch 0 uses its own last block - genesis identity)
+        let epoch_outcome = get_outcome(marshal, epocher, start_epoch.saturating_sub(1)).await?;
+        let epoch_pubkey = *epoch_outcome.sharing().public();
+        let epoch_identity = hex::encode(epoch_pubkey.encode());
```
SuperFluffy: Would prefer if this check was on the `epoch_outcome.sharing().public()`, not on the hex-encoded wire format.

crates/commonware-node/src/feed/state.rs
```diff
@@ -130,49 +130,156 @@ impl FeedStateHandle {
...
-            // Find first transition that happened AFTER start_epoch
-            cache
-                .transitions
-                .iter()
-                .find(|t| t.transition_epoch > start_epoch)
-                .map(|t| t.old_identity.clone())
-                .unwrap_or_else(|| cache.identity.clone())
-        };
+        // Identity active at epoch N is set by the last block of epoch N-1
+        // (epoch 0 uses its own last block - genesis identity)
```
SuperFluffy: Fair, but IMO remove the comment for epoch 0 because nothing is done with that info.

Noting that - should we cover that case somehow?

crates/commonware-node/src/feed/state.rs
```diff
@@ -130,49 +130,156 @@ impl FeedStateHandle {
...
+        let epoch_outcome = get_outcome(marshal, epocher, start_epoch.saturating_sub(1)).await?;
+        let epoch_pubkey = *epoch_outcome.sharing().public();
+        let epoch_identity = hex::encode(epoch_pubkey.encode());
+
+        // Fast path: if the identity matches the cached one, no new transitions
+        // occurred — just extend the cache's from_epoch forward.
```
SuperFluffy: Do you extend it forward or backward...?

crates/commonware-node/src/feed/state.rs
```diff
@@ -130,49 +130,156 @@ impl FeedStateHandle {
...
+            return Ok(());
+        }
+
+        // Walk backwards to find all identity transitions
+        let mut transitions = Vec::new();
+        let mut pubkey = epoch_pubkey;
+        let mut search_epoch = start_epoch.saturating_sub(1);
+        while search_epoch > 0 {
+            // Absorb cached transitions and stop.
+            if let Some(ref cache) = cached
```
SuperFluffy: ref

joshieDo (approved):

crates/commonware-node/src/feed/state.rs
hamdiallam: Once we read, we dont need to hold the lock anymore. `clone` releases the lock. Just convenience

crates/commonware-node/src/feed/state.rs
hamdiallam: Makes it a little more verbose but worth it

PR #2810, opened 2026-02-23
test(it): refactor tempo-tx integration tests into modular TestEnv architecture

0xrusowsky:
## Summary

Refactors the monolithic `tempo_transaction.rs` into a modular test architecture with a `TestEnv` trait abstraction, enabling the same matrix tests and scenario runners to execute against both localnet and testnet.

## What changed

**Architecture** — Split into 6 focused modules:
- `types.rs` — `TestEnv` trait + test case types (`RawSendTestCase`, `SendTestCase`, `FillTestCase`, etc.)
- `helpers.rs` — `Localnet` impl, signing helpers, assertion helpers, tx construction
- `testnet.rs` — `Testnet` impl (RPC URL configurable via `TEMPO_TESTNET_RPC_URL` env var)
- `runners.rs` — All matrix runners and scenario runners, generic over `TestEnv`
- `localnet.rs` — Pool-introspection tests that cannot run on testnet
- `mod.rs` — Wiring + isolated debug test entries

**New coverage (5 scenarios):**
- Spending limit exact boundary (amount == limit)
- Fee payer signature validation negatives (wrong signer, placeholder sig)
- Gas/fee boundary rejections (gas too low, max_fee < base_fee, priority > max_fee)
- Nonce replay rejection (protocol nonce + 2D nonce)
- Negative `eth_sendTransaction` cases (empty calls, WebAuthn without key_data)

**Testnet compat:**
- Pre-T1C networks: gracefully skip tests that require T1C features
- Expiring nonce boundary uses 1s buffer to absorb clock/block-timestamp drift
- Expiring nonce window derived from `TEMPO_EXPIRING_NONCE_MAX_EXPIRY_SECS` constant

## Test counts

| Matrix/Scenario | Cases |
|---|---|
| `eth_sendRawTransaction` | 30 |
| `eth_sendTransaction` | 10 |
| `eth_fillTransaction` | 7 |
| `eth_estimateGas` | 7 |
| E2E fill→sign→send | 15 |
| Scenario runners | 9 |
| Localnet-only | 11 |
| **Total** | **89** |

## How to run

```bash
# All local tests
cargo test -p tempo-node --test it test_matrices_local

# Single matrix (fast iteration)
cargo test -p tempo-node --test it test_debug_iso_raw_send -- --ignored

# Testnet (requires network)
TEMPO_TESTNET_RPC_URL=https://rpc.moderato.tempo.xyz cargo test -p tempo-node --test it test_matrices_testnet -- --ignored
```


crates/node/tests/it/tempo_transaction/testnet.rs
```diff
@@ -0,0 +1,239 @@
...
+    fn provider(&self) -> &Self::P {
+        &self.provider
+    }
+
+    fn chain_id(&self) -> u64 {
+        self.chain_id
+    }
+
+    fn hardfork(&self) -> TempoHardfork {
+        TempoHardfork::T1B
```
fgimenez: having `Testnet.hardfork()` hardcoded to `T1B` will make `skip_pre_t1c` in `run_all_matrices` to always return  `true` for testnet runs. iiuc once the testnet upgrades to T1C, regressions in T1C features will still be silently swallowed by the `check()` wrapper. 

to make it more futureproof we could derive the hardfork from chain config/RPC so the skip logic becomes self-correcting, wdyt?

fgimenez: proposed here https://github.com/tempoxyz/tempo/pull/2965

0xrusowsky: yes, this is a way better idea!

crates/node/tests/it/tempo_transaction/types.rs
```diff
@@ -0,0 +1,737 @@
...
+    /// Returns the receipt JSON.
+    async fn submit_tx(
+        &mut self,
+        encoded: Vec<u8>,
+        tx_hash: B256,
+    ) -> eyre::Result<serde_json::Value>;
+
+    /// Submit a transaction that is expected to be rejected by the RPC.
+    /// If `expected_reason` is provided, the error message must contain it.
+    async fn submit_tx_expecting_rejection(
```
legion2002: we can either modify this fn or add a new one `submit_tx_expecting_rejection_handler`/`submit_tx_expecting_rejection_rpc_handler` which checks that the tx gets rejected at both handler and RPC level.

It will enforce that anything that gets rejected at RPC level, also gets rejected in the handler.

crates/node/tests/it/tempo_transaction/testnet.rs
```diff
@@ -0,0 +1,261 @@
...
+use tempo_chainspec::{
+    hardfork::{TempoHardfork, TempoHardforks},
+    spec::{ANDANTINO, DEV, MODERATO, PRESTO},
+};
+use tempo_primitives::{TempoTxEnvelope, transaction::tempo_transaction::Call};
+
+use super::helpers::*;
+
+/// Testnet RPC url (unpermissioned).
+const TESTNET_RPC_URL: &str = "https://rpc.moderato.tempo.xyz";
```
legion2002, line 22: For the CI should we fetch the auth RPC URL from the env?

crates/node/tests/it/tempo_transaction/types.rs
legion2002, line 43: For localnet, can we override this function to check the tx gets rejected at both handler and RPC level

crates/node/tests/it/tempo_transaction/testnet.rs
0xrusowsky: this is used as a fallback: `let rpc_url = std::env::var("TEMPO_TESTNET_RPC_URL").unwrap_or(TESTNET_RPC_URL.to_string());`

crates/node/tests/it/tempo_transaction/types.rs
0xrusowsky: addressed in [`0b8f20d` (this PR)](https://github.com/tempoxyz/tempo/pull/2810/commits/0b8f20de803a5765cbf081118c79a29609274f0e)

PR #2815, opened 2026-02-23
Sync main

sds:
Sync


crates/transaction-pool/src/validator.rs
```diff
@@ -842,20 +849,38 @@ where
...
-                        // Expiring nonce transaction - check if tx hash is already seen
-                        let tx_hash = *transaction.hash();
-                        let seen_slot = NonceManager::new().expiring_seen_slot(tx_hash);
+                        // Expiring nonce transaction - check if the replay hash is already seen.
+                        //
+                        // Pre-T1B: use tx_hash to match handler behavior (handler writes seen[tx_hash]).
+                        // T1B+: use expiring_nonce_hash (invariant to fee payer changes) to match
+                        //        the updated handler replay protection.
+                        //
+                        // TODO: Remove the tx_hash path after T1B is active on mainnet.
```
0xKitsune, line 858: This should be cleaned up, we can do this in a subsequent PR.

Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace.package]
-version = "1.2.0"
+version = "1.3.0"
```
0xKitsune, line 2: Looks like we will need to update reth deps and bump this to `1.3.1`?

0xKitsune (approved):

Cargo.toml
0xKitsune: We can do this in a follow-on PR.

PR #2825, opened 2026-02-24
Fix: ip bugs

howydev:
Fixes 2 bugs found

Problem

1. Unstable Debug formatting in error return data — NotIpPort/NotHostPort/NotIp errors used format!("{err:?}") to populate the backtrace field. Rust's Debug output is not stable across compiler versions, so upgrading rustc could change the return data length → different gas charged → different block hash → chain fails to sync.
2. IPv6 zone IDs accepted — ensure_address_is_ip_port and ensure_address_is_ip delegate to std::net parsing, which accepts IPv6 zones (e.g. fe80::1%eth0). Zone IDs are semantically meaningless on-chain.
Panic in deactivate_validator — ingress_ip_key was called with .expect(), which would crash the node instead of reverting the transaction.

Fix

1. Validator Config V1 (gated behind T2 hardfork for consensus compatibility):
Pre-T2: preserves exact legacy behavior (format!("{err:?}"), zones accepted)
T2+: uses stable err.to_string() and rejects IPv6 zones

2. Validator Config V2 (already T2-only at dispatch, no gating needed):
format!("{err:?}") → err.to_string()
.expect() → ? in deactivate_validator
IPv6 zones rejected via ensure_address_is_ip_port / ensure_address_is_ip

### DIFF
--- a/crates/precompiles/src/ip_validation.rs
+++ b/crates/precompiles/src/ip_validation.rs
@@ -7,29 +7,13 @@
 pub(crate) enum IpWithPortParseError {
     #[error("input was not of the form `<ip>:<port>`")]
     Parse(#[from] std::net::AddrParseError),
-    #[error("IPv6 zone IDs are not allowed")]
-    Ipv6Zone,
 }
 
 /// Validates that `input` is of the form `<ip>:<port>`.
 pub(crate) fn ensure_address_is_ip_port(
     input: &str,
 ) -> core::result::Result<(), IpWithPortParseError> {
-    if input.contains('%') {
-        return Err(IpWithPortParseError::Ipv6Zone);
-    }
     input.parse::<std::net::SocketAddr>()?;
-    Ok(())
-}
-
-/// Like [`ensure_address_is_ip_port`] but without the IPv6 zone check.
-/// Used for pre-T2 consensus compatibility where zones were previously accepted.
-pub(crate) fn ensure_address_is_ip_port_legacy(
-    input: &str,
-) -> core::result::Result<(), IpWithPortParseError> {
-    input
-        .parse::<std::net::SocketAddr>()
-        .map_err(IpWithPortParseError::Parse)?;
     Ok(())
 }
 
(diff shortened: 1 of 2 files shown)


crates/precompiles/src/ip_validation.rs
```diff
@@ -4,27 +4,47 @@
...
+    Parse(#[from] std::net::AddrParseError),
+    #[error("IPv6 zone IDs are not allowed")]
+    Ipv6Zone,
 }
 
+/// Validates that `input` is of the form `<ip>:<port>`.
 pub(crate) fn ensure_address_is_ip_port(
     input: &str,
 ) -> core::result::Result<(), IpWithPortParseError> {
+    if input.contains('%') {
```
SuperFluffy: Don't parse the input yourself; parse into a `SocketAddr` and check if it contains a zone.

crates/precompiles/src/ip_validation.rs
```diff
@@ -4,27 +4,47 @@
...
+pub(crate) enum IpParseError {
+    #[error("input was not a valid IP address")]
+    Parse(#[from] std::net::AddrParseError),
+    #[error("IPv6 zone IDs are not allowed")]
+    Ipv6Zone,
 }
 
 pub(crate) fn ensure_address_is_ip(input: &str) -> core::result::Result<(), IpParseError> {
+    if input.contains('%') {
+        return Err(IpParseError::Ipv6Zone);
```
SuperFluffy: Rust IPv6 addresses do not allow zones.

SuperFluffy (changes requested):
Rust `IpAddr::V6` does not permit zones/scope IDs. Only `SocketAddr::V6` has that.

If we want to guard against this ever changing we should add a unit test:

```rust
#[should_panic]
fn std_ipv6_rejects_scope_ids() {
    "::1%1".parse::<IpAddr>().unwrap();
}
```

crates/precompiles/src/ip_validation.rs
howydev: Is there a function to check if theres a zone?

From ai, the best check seems to be `scope_id != 0` but this would fail the check for `...%0:...`

SuperFluffy: Follow up to this - I don't consider the scope IDs as something we should bother with.

We don't do that for ipv4 either, and accept private, local, and link-local IP ranges (in addition to others). Scope IDs fulfill the same need.

howydev: ok, thanks for elaborating

PR #2826, opened 2026-02-24
feat(revm): expose `TempoEvmHandler` API for foundry consumption

0xrusowsky:
## Motivation

upstream foundry intercepts `CREATE2` opcodes during `vm.startBroadcast()` and routes them through the `Arachnid CREATE2` factory. This happens in `FoundryHandler::inspect_run_exec_loop`.

`tempo-foundry` wraps `TempoEvmHandler` inside `FoundryHandler` for Tempo-specific execution (fee tokens, gas adjustment, AA dispatch). But the `InspectorHandler` impl delegates `inspect_run` and `inspect_execution` directly to `TempoEvmHandler`, so the execution chain never reaches `FoundryHandler::inspect_run_exec_loop` — CREATE2 factory routing is dead code.

this causes `CREATE2` deployments inside broadcast to produce wrong addresses on-chain (nonce-based instead of salt-based). Fixing this bug requires `FoundryHandler` to override `inspect_execution` and intercept `CREATE2` calls.

## Solution

add `TempoEvmHandler::inspect_execution_with` — accepts a custom exec loop closure for standard transactions while keeping all Tempo logic as the single source of truth:

```rust
// TempoEvmHandler trait impl — one-liner
fn inspect_execution(&mut self, evm, gas) -> ... {
    self.inspect_execution_with(evm, gas, Self::inspect_run_exec_loop)
}

// FoundryHandler — injects CREATE2 routing
fn inspect_execution(&mut self, evm, gas) -> ... {
    let overrides = &mut self.create2_overrides;
    self.inner.inspect_execution_with(evm, gas, |_handler, evm, init| {
        create2_exec_loop(overrides, evm, init)
    })
}
```

additionally `load_fee_fields` needs to be public for `FoundryHandler::inspect_run`.

### DIFF
--- a/crates/revm/src/handler.rs
+++ b/crates/revm/src/handler.rs
@@ -216,10 +216,15 @@ impl<DB, I> TempoEvmHandler<DB, I> {
 impl<DB: alloy_evm::Database, I> TempoEvmHandler<DB, I> {
     /// Loads the fee token and fee payer from the transaction environment.
     ///
-    /// Must be called before `validate_against_state_and_deduct_caller`, which uses the loaded
-    /// fee fields for balance checks. Called by `TempoEvmHandler::inspect_run` and
-    /// `Handler::run`; exposed for downstream `InspectorHandler` impls (e.g. tempo-foundry)
-    /// that override `inspect_run` but still need Tempo fee setup.
+    /// Resolves and validates the fee fields used by Tempo's fee system:
+    /// - Fee payer: determined from the transaction
+    /// - Fee token: resolved via the journaled state and validated (TIP20 prefix + USD currency)
+    ///
+    /// Must be called before `validate_against_state_and_deduct_caller`, which uses the
+    /// loaded fee fields for balance checks.
+    ///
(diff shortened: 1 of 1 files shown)


crates/revm/src/handler.rs
```diff
@@ -214,7 +214,7 @@ impl<DB, I> TempoEvmHandler<DB, I> {
 }
 
 impl<DB: alloy_evm::Database, I> TempoEvmHandler<DB, I> {
-    fn load_fee_fields(
+    pub fn load_fee_fields(
```
mattsse, line 228: needs docs, is now pub

mattsse (changes requested):

crates/revm/src/handler.rs
0xrusowsky: addressed in [`2d7700b` (this PR)](https://github.com/tempoxyz/tempo/pull/2826/commits/2d7700b45f677f70cf1b503a47c0bc22ec5654e9)

rakita (approved):

PR #2829, opened 2026-02-24
feat(precompiles): add getFeeToken() view to IFeeManager (TIP-1007)

fgimenez:
Closes CHAIN-467

Adds `IFeeManager.getFeeToken()` so smart contracts can read the fee token address used for the current transaction during execution. 

The fee token address is written to transient storage (EIP-1153) in the handler's `validate_against_state_and_deduct_caller` step, before transaction execution begins. `getFeeToken()` reads from that transient slot, which is automatically cleared between transactions.

- Add `getFeeToken()` to the `IFeeManager` Solidity and Rust interfaces
- Add a `tx_fee_token` transient storage field to `TipFeeManager` and wire its dispatch
- Set the fee token in the handler during fee deduction, before execution begins
- Add two integration tests covering `eth_call` safety and runtime behavior

### DIFF
--- a/.github/workflows/pr-audit.yml
+++ b/.github/workflows/pr-audit.yml
@@ -7,14 +7,12 @@ on:
 jobs:
   publish:
     runs-on: ubuntu-latest
-    if: github.event.pull_request.draft == false && contains(github.event.pull_request.labels.*.name, 'agentic-audit')
+    if: github.event.label.name == 'agentic-audit'
     steps:
       - name: Publish event
         run: |
           set -euo pipefail
-          curl -sf -o /dev/null -X POST ${{ secrets.EVENTS_ARGS }} \
-            -H "Content-Type: application/json" \
-            -d '{
+          curl -sf -o /dev/null -X POST ${{ secrets.EVENTS_ARGS }}             -H "Content-Type: application/json"             -d '{
               "repository": "${{ github.repository }}",
               "event": "pr_audit",
               "data": {
--- a/bin/tempo-bench/src/cmd/max_tps.rs
+++ b/bin/tempo-bench/src/cmd/max_tps.rs
@@ -1180,7 +1180,7 @@ async fn assert_receipts<R: ReceiptResponse, F: Future<Output = eyre::Result<R>>
     receipts: impl IntoIterator<Item = F>,
     max_concurrent_requests: usize,
 ) -> eyre::Result<()> {
-    stream::iter(receipts.into_iter())
+    stream::iter(receipts)
         .buffer_unordered(max_concurrent_requests)
         .try_for_each(|receipt| assert_receipt(receipt))
         .await
--- a/crates/chainspec/src/genesis/dev.json
+++ b/crates/chainspec/src/genesis/dev.json
@@ -24,6 +24,7 @@
     "t1Time": 0,
     "t1aTime": 0,
     "t1bTime": 0,
+    "t1cTime": 0,
     "t2Time": 0,
     "depositContractAddress": "0x0000000000000000000000000000000000000000"
   },
--- a/crates/chainspec/src/hardfork.rs
+++ b/crates/chainspec/src/hardfork.rs
@@ -50,6 +50,8 @@ hardfork!(
         T1A,
         /// T1.B hardfork
         T1B,
+        /// T1.C hardfork
+        T1C,
         /// T2 hardfork - adds compound transfer policies (TIP-1015)
         T2,
     }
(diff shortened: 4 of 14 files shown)


crates/precompiles/src/tip_fee_manager/mod.rs
```diff
@@ -32,6 +32,9 @@ pub struct TipFeeManager {
     /// T2+: Tracks liquidity reserved for a pending fee swap during `collect_fee_pre_tx`.
     /// Checked by `burn` and `rebalance_swap` to prevent withdrawals that would violate the reservation.
     pending_fee_swap_reservation: Mapping<B256, u128>,
+    /// T2+: The fee token used for the current transaction (TIP-1007).
+    /// Set by the handler before execution, read via `getFeeToken()`.
+    tx_fee_token: Address,
```
0xrusowsky, line 41: can we add this warning please?


```suggestion

    // WARNING(rusowsky): transient storage slots must always be placed at the very end until the `contract`
    // macro is refactored and has 2 independent layouts (persistent and transient).
    // If new (persistent) storage fields need to be added to the precompile, they must go above this one.
    /// T2+: The fee token used for the current transaction (TIP-1007).
    /// Set by the handler before execution, read via `getFeeToken()`.
    tx_fee_token: Address,
```

0xrusowsky (changes requested):
lgtm, requesting changes so that we properly add the transient storage disclaimer

crates/precompiles/src/tip_fee_manager/mod.rs
fgimenez: sure good point

crates/precompiles/src/tip_fee_manager/dispatch.rs
```diff
@@ -47,6 +47,9 @@ impl Precompile for TipFeeManager {
             TipFeeManagerCall::FeeManager(IFeeManagerCalls::collectedFees(call)) => {
                 view(call, |c| self.collected_fees[c.validator][c.token].read())
             }
+            TipFeeManagerCall::FeeManager(IFeeManagerCalls::getFeeToken(call)) => {
+                view(call, |_| self.get_fee_token())
+            }
```
0xrusowsky, line 64: if we make it available in the dispatcher, this technically should go behind t2 hardfork too, and if called before it should return `UnknownSelector`

fgimenez: makes sense, done ptal

0xrusowsky (approved):
makes sense!

let's hold merge until we have another reviewer though

tips/ref-impls/src/FeeManager.sol
```diff
@@ -86,6 +86,17 @@ contract FeeManager is IFeeManager, FeeAMM {
         }
     }
 
+    function getFeeToken() external view returns (address) {
+        address token;
+        // NOTE: The slot used here (0) is specific to this reference implementation.
+        // The production Rust precompile uses a different transient slot determined
+        // by the `contract` macro's layout (currently slot 7).
```
legion2002, line 93: Is this comment correct? My understanding was that even though the transient storage variables are placed at location 7 in the struct, they would still start from transient slot 0.

CC: @0xrusowsky

fgimenez: thx for raising this, the comment follows this warning https://github.com/tempoxyz/tempo/blob/main/crates/precompiles/src/tip_fee_manager/mod.rs#L29-L31 aiui this concrete transient slot should be put at the very end, is this correct?

0xrusowsky: yes, this is correct. in practical terms it doesn't matter as this is just transient storage which leaves in memory.

with that being said, i want the macros to eventually have the 2 actual storage spaces properly represented

crates/node/tests/it/tip_fee_manager.rs
```diff
@@ -403,3 +464,166 @@ async fn test_fee_payer_tx() -> eyre::Result<()> {
...
+    let setup = TestNodeBuilder::new().build_http_only().await?;
+    let wallet = MnemonicBuilder::from_phrase(crate::utils::TEST_MNEMONIC).build()?;
+    let user_address = wallet.address();
+    let provider = ProviderBuilder::new()
+        .wallet(wallet)
+        .connect_http(setup.http_url);
+
+    let fee_manager = IFeeManager::new(TIP_FEE_MANAGER_ADDRESS, provider.clone());
+
+    // The handler runs during eth_call, so getFeeToken() returns the user's resolved
```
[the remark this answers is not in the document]
legion2002: @fgimenez think this is a valid issue, and we should update either the TIP or the behaviour.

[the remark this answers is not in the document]
fgimenez: > think this is a valid issue, and we should update either the TIP or the behaviour.

absolutely, pushed changes here https://github.com/tempoxyz/tempo/pull/2829/commits/4a1aa458335821838f0ea210a21456f29468b011 to align rust impl to spec, ptal @legion2002 @0xrusowsky

PR #2835, opened 2026-02-24
test: cross-account key auth replay

onbjerg:
Add `handler_keyAuthCrossAccountReplay` to the invariant fuzz suite. Account A authorizes key K, then account B tries to use key K to sign a transaction claiming `userAddress=B`. The keychain must reject because key K is only authorized for A, not B.

### DIFF
--- a/tips/ref-impls/test/TempoTransactionInvariant.t.sol
+++ b/tips/ref-impls/test/TempoTransactionInvariant.t.sol
@@ -58,7 +58,7 @@ contract TempoTransactionInvariantTest is InvariantChecker {
         targetContract(address(this));
 
         // Define which handlers the fuzzer should call
-        bytes4[] memory selectors = new bytes4[](73);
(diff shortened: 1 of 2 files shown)


tips/ref-impls/test/TempoTransactionInvariant.t.sol
```diff
@@ -4559,4 +4568,93 @@ contract TempoTransactionInvariantTest is InvariantChecker {
...
+                uint256[] memory amounts = new uint256[](1);
+                amounts[0] = 1000e6;
+                _authorizeKey(ownerA, keyId, expiry, true, tokens, amounts);
+            } catch {
+                return;
+            }
+        }
+
+        // Skip if key is also authorized for B — not a meaningful replay test
+        if (ghost_keyAuthorized[ownerB][keyId]) return;
```
0xrusowsky, line 4636: nit: prob better to check and the beginning and early exit

PR #2840, opened 2026-02-24
feat(consensus): add structured fields to verification and build logs

gakonst:
Add `block_height`, `block_digest`, `epoch`, `view` to verification result and cancellation log lines. Add `epoch`, `view` to payload builder timeout log.

Enables precise log correlation by ID instead of timestamp proximity.

Prompted by: joshie

### DIFF
--- a/crates/commonware-node/src/consensus/application/actor.rs
+++ b/crates/commonware-node/src/consensus/application/actor.rs
@@ -414,7 +414,7 @@ impl Inner<Init> {
 
         // Respond with the verification result ASAP. Also generates
         // the event reporting the result of the verification.
-        let _ = report_verification_result(response, &result);
+        let _ = report_verification_result(response, &result, round, payload);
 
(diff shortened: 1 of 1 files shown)


decofe (approved):

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -435,6 +435,13 @@ impl Inner<Init> {
         }
     }
 
+    #[instrument(
+        skip_all,
+        fields(
+            epoch = %round.epoch(),
```
SuperFluffy, line 441: These should be inherited from the parent scope, where these values are already injected.

Do they not propagate further?

I would like to not make log lines/events even more noisy by duplicating all that information.

Before we make this change, can you double check if the fields are indeed not propagated?

SuperFluffy (changes requested):

PR #2841, opened 2026-02-24
feat: tempo consensus validator cli commands

hamdiallam:
* Sanity checks on the signature
* Submit to rpc. Since the signature is dependent on the chain id, incorrect rpc will fail as the validation step

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12435,7 +12435,7 @@ dependencies = [
 
 [[package]]
 name = "tempo-validator-config"
-version = "1.4.0"
+version = "1.4.2"
 dependencies = [
  "alloy-primitives",
  "commonware-codec",


bin/tempo/src/tempo_cmd.rs
```diff
@@ -21,9 +27,13 @@ use serde::Serialize;
 use tempo_alloy::TempoNetwork;
 use tempo_chainspec::spec::TempoChainSpec;
 use tempo_commonware_node_config::SigningKey;
-use tempo_contracts::precompiles::{IValidatorConfig, VALIDATOR_CONFIG_ADDRESS};
+use tempo_contracts::precompiles::{
+    IValidatorConfig, IValidatorConfigV2, VALIDATOR_CONFIG_ADDRESS, VALIDATOR_CONFIG_V2_ADDRESS,
+};
 use tempo_dkg_onchain_artifacts::OnchainDkgOutcome;
 
+const ADD_VALIDATOR_NAMESPACE: &[u8] = b"TEMPO_VALIDATOR_CONFIG_V2_ADD_VALIDATOR";
```
SuperFluffy: Imporat this from the precompile

bin/tempo/src/tempo_cmd.rs
```diff
@@ -53,13 +67,247 @@ pub(crate) enum ConsensusSubcommand {
...
+
+        let validator_config = ValidatorConfig::new(
+            self.validator_address,
+            self.public_key,
+            self.ingress,
+            self.egress,
+        );
+
+        // Sanity check the signature
+        validator_config.check_signature(chain_id, self.signature.clone())?;
```
SuperFluffy: This will fail. Rotation has a different namespace.

hamdiallam: yep, I just fixed this

bin/tempo/src/tempo_cmd.rs
hamdiallam: config redefines them to avoid the dependency on the precompiles which is heavier

crates/commonware-node-config/src/validator.rs
```diff
@@ -0,0 +1,85 @@
...
+use alloy_primitives::{Address, B256, Keccak256};
+use commonware_codec::DecodeExt as _;
+use commonware_cryptography::{Verifier as _, ed25519};
+use tempo_contracts::precompiles::VALIDATOR_CONFIG_V2_ADDRESS;
+
+/// Namespace used for ed25519 signatures when adding a validator.
+pub const ADD_VALIDATOR_NAMESPACE: &[u8] = b"TEMPO_VALIDATOR_CONFIG_V2_ADD_VALIDATOR";
+
+/// Namespace used for ed25519 signatures when rotating a validator.
+pub const ROTATE_VALIDATOR_NAMESPACE: &[u8] = b"TEMPO_VALIDATOR_CONFIG_V2_ROTATE_VALIDATOR";
```
SuperFluffy: Should use the const defined in the precompiles.

crates/validator-config/src/lib.rs
SuperFluffy, line 1: Makes sense to have this, but let's move this to its own crate.

PR #2852, opened 2026-02-25
TIP 1022: Virtual forwarding addresses

malleshpai:
- Adds TIP-1022, introducing virtual TIP-20 deposit addresses that auto-forward to a registered master wallet.
- Eliminates sweep transactions and avoids per-deposit-address state creation/state bloat.
- Defines registry, forwarding behavior, and invariants for safe implementation.


tips/tip-1022.md
```diff
@@ -0,0 +1,454 @@
...
+```
+[8-byte PREFIX] [4-byte masterId] [8-byte userTag]
+= 20 bytes total
+```
+
+| Field | Bytes | Description |
+|-------|-------|-------------|
+| **PREFIX** | 8 | Fixed prefix `0xFDFDFDFDFDFDFDFD`. Identifies virtual addresses with collision probability ~1/2⁶⁴. |
+| **masterId** | 4 | Deterministic identifier derived from the master address. Supports up to ~4 billion masters. |
+| **userTag** | 8 | Opaque per-user identifier, derived off-chain. 64-bit space supports >10¹⁹ unique deposit addresses per master. |
```
danrobinson: Why have 2^64 deposit-addresses-per-master and only 2^32 masters?

tips/tip-1022.md
```diff
@@ -0,0 +1,454 @@
...
+
+Parsing is position-based and endian-neutral:
+
+- `addr[0:8]` = `VIRTUAL_PREFIX`
+- `addr[8:12]` = `masterId` (the raw 4-byte sequence as it appears in the address)
+- `addr[12:20]` = `userTag` (the raw 8-byte sequence as it appears in the address)
+
+Implementations MUST NOT reinterpret `masterId` or `userTag` using host endianness.
```
danrobinson: Probably unnecessary sentence

tips/tip-1022.md
malleshpai: I did it for 8-4-8 symmetry, 8-6-6 would also be fine imo. What's your preferred split?

tips/tip-1022.md
```diff
@@ -0,0 +1,454 @@
...
+
+TIP-1022 applies only to TIP-20 precompile recipient resolution for the entrypoints listed in **Transfer Path Modification**.
+
+TIP-1022 does **not** alter TIP-20 methods that do not carry a recipient (e.g., `approve`, `burn`, `permit`) and does not alter non-TIP-20 protocol behavior.
+
+Before TIP-1022 activation (`protocolVersion` hardfork), all TIP-20 behavior remains unchanged.
+
+### Prefix Selection: `0xFDFDFDFDFDFDFDFD`
+
+An 8-byte prefix provides collision resistance of ~1/2⁶⁴. The probability that any randomly-generated EOA or CREATE2-deployed contract matches this prefix is negligible by any measure and infeasible to grind even with dedicated hardware.
```
danrobinson: I don't think that's a safe assumption for 8 bytes; ChatGPT estimates you could do it for <$1m on cloud hardware. Either we should explicitly tolerate this attack or choose 10 or 12 bytes. (TIP-20 uses a 12-byte prefix).

Maybe either 10/5/5 or 12/4/4?

Should probably describe the fallback behavior anyway.

malleshpai: i'm doing 4-8-8 right now. another option is to do 10-5-5 with commit-reveal used to prevent grinding attacks on the master id. we can discuss tmw?

tips/tip-1022.md
```diff
@@ -0,0 +1,657 @@
...
+Virtual addresses are standard 20-byte EVM addresses with a reserved prefix, interpreted as a byte array where byte `0` is the leftmost byte in the hex string:
+
+```
+[5-byte PREFIX] [10-byte masterId] [5-byte userTag]
+= 20 bytes total
+```
+
+| Field | Bytes | Description |
+|-------|-------|-------------|
+| **PREFIX** | 5 | Fixed prefix `0xFDFDFDFDFD`. Identifies virtual addresses with collision probability ~1/2⁴⁰. |
```
danrobinson: false positive probability

tips/tip-1022.md
```diff
@@ -0,0 +1,657 @@
...
+
+## Conformance and Scope
+
+TIP-1022 applies only to TIP-20 precompile recipient resolution for the entrypoints listed in **Transfer Path Modification**.
+
+TIP-1022 does **not** alter TIP-20 methods that do not carry a recipient (e.g., `approve`, `burn`, `permit`) and does not alter non-TIP-20 protocol behavior.
+
+Before TIP-1022 activation (`protocolVersion` hardfork), all TIP-20 behavior remains unchanged.
+
+Native token (ETH/gas token) transfers to virtual addresses are **not** forwarded and behave as transfers to any other address. TIP-1022 forwarding applies exclusively to TIP-20 precompile operations.
```
danrobinson:
```suggestion
Behavior of non-TIP-20 tokens (such as contracts that implement the ERC-20 interface) are not affected. TIP-1022 forwarding applies exclusively to TIP-20 precompile operations. Any non-TIP-20 tokens or NFTs would be lost if sent to a virtual address.
```

tips/tip-1022.md
```diff
@@ -0,0 +1,657 @@
...
+
+### Prefix Selection: `0xFDFDFDFDFD`
+
+A 5-byte prefix provides collision resistance of ~1/2⁴⁰ (~1 in a trillion). This is an explicit layout tradeoff to maximize `masterId` space (10 bytes) while retaining a practical `userTag` (5 bytes, >10¹² addresses per master).
```
danrobinson: Should explicitly point out that intentionally mining EOAs or contract addresses that match this prefix is feasible.

tips/tip-1022.md
```diff
@@ -0,0 +1,657 @@
...
+
+### Reserved Address Space
+
+Any address whose first 5 bytes match `VIRTUAL_PREFIX` is **unconditionally reserved** by the protocol. If a user happens to control a private key whose corresponding address begins with `0xFDFDFDFDFD` on another chain, that address **cannot be used on Tempo**: TIP-20 transfers to it will either forward to a registered master (if the embedded `masterId` is registered) or revert (if it is not). The address cannot receive TIP-20 tokens directly, cannot accumulate a TIP-20 balance, and cannot be registered as a master address. Users in this situation must use a different address on Tempo.
```
danrobinson: I think let's change it to have the address just receive the tokens if there is no master ID registered.

tips/tip-1022.md
```diff
@@ -0,0 +1,657 @@
...
+
+The `masterId` is **deterministic**, derived from the address that calls `registerVirtualMaster()` and a caller-supplied salt:
+
+```
+masterId = bytes10(keccak256(abi.encodePacked(msg.sender, salt)))
+```
+
+The salt is a `bytes32` value chosen by the caller. Callers SHOULD use a cryptographically random salt rather than a predictable value (e.g., `bytes32(0)`) to prevent attackers from pre-grinding `masterId` collisions before the registration transaction lands. This eliminates dependence on transaction ordering and is stable under reorgs. The `masterId` is permanently bound to the **registration address** and does not change if the master is later rotated to a new address via `proposeUpdateMaster` / `finalizeUpdateMaster`. This is what makes deposit addresses stable: they embed `masterId`, so they continue to work regardless of master rotation.
+
+In the unlikely event of a masterId collision (two `(address, salt)` pairs mapping to the same 10-byte hash), the second registration reverts with `MasterIdCollision`. The caller can retry with a different salt.
```
danrobinson: Note about security—to front-run this master ID, an attacker would have to mine a specific 10-byte target between the transaction becoming publicly visible and it being included onchain. This should be infeasible. Nevertheless, users should check that their master IDs were set successfully before using those addresses.

tips/tip-1022.md
```diff
@@ -0,0 +1,657 @@
...
+
+### User Tag Derivation (Off-Chain)
+
+The `userTag` is an opaque 5-byte value generated off-chain by the exchange/processor. The protocol does not interpret or validate it. It exists solely so the exchange can attribute deposits to specific users via the emitted `VirtualForward` event.
+
+Exchanges maintain their own internal mapping `{internalUserId → virtualAddress}`. No on-chain transaction is needed to create a new deposit address.
+
+Operators SHOULD derive userTags using a keyed hash (e.g., HMAC) to reduce third-party enumeration of deposit addresses.
```
danrobinson:
```suggestion
Operators SHOULD derive userTags using a keyed hash (e.g., HMAC), rather than sequentially starting from 0, to reduce information leakage about the order of address creation.
```

dankrad: Shouldn't it just be random rather than a hash? 5 bytes provides way too little collision resistance in practice so even if you generate random numbers by means of a hash function you still have to store them in a database to resolve possible collisions.

danrobinson: Good point; I think we should either recommend random addresses with checks for collisions, or sequential (or otherwise organized-to-avoid-collisions) addresses

tips/tip-1022.md
```diff
@@ -0,0 +1,640 @@
+---
+id: TIP-1022
+title: Virtual Addresses for TIP-20 Deposit Forwarding
```
danrobinson, line 3: Possible change discussed—don't allow virtual addresses to be blocked in a 403 policy (since they don't get checked anyway; only the master does)

tips/tip-1022.md
```diff
@@ -0,0 +1,640 @@
...
+= 20 bytes total
+```
+
+| Field | Bytes | Description |
+|-------|-------|-------------|
+| **PREFIX** | 10 | Fixed prefix `0xFDFDFDFDFDFDFDFDFDFD`. Identifies virtual addresses with collision probability ~1/2⁸⁰. |
+| **masterId** | 6 | Deterministic identifier derived from the master address. Supports up to ~2.8×10¹⁴ masters. |
+| **userTag** | 4 | Opaque per-user identifier, derived off-chain. 32-bit space supports >4 billion (2³²) unique deposit addresses per master. |
+
+Parsing is position-based and endian-neutral:
```
samczsun: we established endianess above, this sentence doesn't really make sense

tips/tip-1022.md
```diff
@@ -0,0 +1,640 @@
...
+## Interaction with TIP-403 Policies
+
+Virtual address resolution happens **before** TIP-403 transfer authorization checks. Policy evaluation uses a two-level model combining the master account's policy with an optional per-masterId policy.
+
+### Per-MasterId Policies (requires TIP-1023)
+
+A master MAY set a TIP-403 account-level policy scoped to a specific `masterId` by setting a policy on the **masterId policy address**, a synthetic address constructed as:
+
+```
+VIRTUAL_PREFIX | masterId | 00000000
```
samczsun: broken art

tips/tip-1022.md
```diff
@@ -0,0 +1,640 @@
...
+
+### Prefix Selection: `0xFDFDFDFDFDFDFDFDFDFD`
+
+A 10-byte prefix provides collision resistance of ~1/2⁸⁰. This is an explicit layout tradeoff that prioritizes address space safety over `masterId` and `userTag` space. The 6-byte `masterId` supports ~2.8×10¹⁴ masters, and the 4-byte `userTag` supports >4 billion (2³²) deposit addresses per master.
+
+### Reserved Address Space
+
+Any address whose first 10 bytes match `VIRTUAL_PREFIX` is **unconditionally reserved** by the protocol. If a user happens to control a private key whose corresponding address begins with `0xFDFDFDFDFDFDFDFDFDFD` on another chain, that address **cannot be used on Tempo**: TIP-20 transfers to it will either forward to a registered master (if the embedded `masterId` is registered) or revert (if it is not). The address cannot receive TIP-20 tokens directly, cannot accumulate a TIP-20 balance, and cannot be registered as a master address. Users in this situation must use a different address on Tempo.
```
samczsun: very strong statement that is not enforced by invariants, either restrict private keys that resolve to a virtual address or loosen restrictions here

tips/tip-1022.md
```diff
@@ -0,0 +1,640 @@
...
+```
+
+The first 4 bytes of `registrationHash` MUST be zero:
+
+```
+require(bytes4(registrationHash[0:4]) == 0x00000000)  // 32-bit PoW
+masterId = bytes6(registrationHash[4:10])
+```
+
+This requires the caller to grind ~2³² salt values to find a valid registration. Combined with the 10-byte prefix, an attacker attempting to deploy a contract (via CREATE2) at a virtual address must match both the prefix (2⁸⁰) and the PoW constraint (2³²), requiring ~2¹¹² total work.
```
samczsun: shouldn't this be 16 bytes of collision resistance, not 14?

tips/tip-1022.md
```diff
@@ -0,0 +1,640 @@
...
+
+A master MAY set a TIP-403 account-level policy scoped to a specific `masterId` by setting a policy on the **masterId policy address**, a synthetic address constructed as:
+
+```
+VIRTUAL_PREFIX | masterId | 00000000
+^^^^^^^^^^^^^^^^^^^^      ^^^^^^^^^^^^  ^^^^^^^^
+prefix (10)               masterId (6)  zero-padded userTag (4)
+```
+
+This is the virtual address with `userTag = 0x00000000`. It is not a real deposit address; it serves only as a policy anchor for TIP-403. The current master for a `masterId` MAY call TIP-1023 account-level policy functions using this address to set token-level restrictions (e.g., "virtual addresses under this masterId may only receive USDC") or other TIP-403 constraints.
```
samczsun: no restriction on the zero user tag enforced

tips/tip-1022.md
```diff
@@ -0,0 +1,640 @@
...
+
+This enables a single wallet holding multiple masterIds to enforce different policies per masterId — for example, one masterId restricted to USDC for US users and another masterId allowing both USDC and USDT for non-US users — while all funds consolidate into the same master wallet.
+
+### Policy Evaluation Order
```
samczsun: the steps for a transfer involving virtual addresses are scattered throughout the spec, consider consolidating

tips/tip-1022.md
```diff
@@ -0,0 +1,640 @@
...
+
+11. **Rotation consent binding**: `updateMaster` MUST require valid new-master consent bound to `(chainId, registryAddress, masterId, currentMaster, newMaster, deadline)`.
+
+12. **View/execution symmetry**: `effectiveRecipient`, `balanceOfEffective`, and `isTransferAuthorizedEffective` MUST match execution-time recipient resolution semantics.
+
+13. **Two-level policy enforcement**: For virtual-forwarded transfers, TIP-403 authorization MUST check both the masterId policy address (`VIRTUAL_PREFIX | masterId | 00..0`) and the resolved master account. The transfer MUST be rejected if either policy disallows it.
+
+14. **Virtual address policy ignorance**: TIP-403 policies set on individual virtual addresses (non-zero `userTag`) MUST be ignored during transfer authorization. Only the masterId policy address and the master account address are checked.
+
+## Critical Test Cases
```
samczsun: I'm skeptical this belongs in the spec

malleshpai: we should discuss this in the upgrade call, happy to drop this in a separate doc going forward but currently i think (?) it's part of the spec template

tips/tip-1022.md
malleshpai: opening an issue on this for later, we should do this as part of a TIP-403 v2 spec rather than put it in here?

malleshpai: https://github.com/tempoxyz/tempo/issues/3049

tips/tip-1022.md
```diff
@@ -0,0 +1,601 @@
...
+- MUST NOT match `VIRTUAL_PREFIX`
+- MUST NOT be a TIP-20 token address (i.e., MUST pass TIP-20 recipient validation for transfer paths)
+
+### Registry Storage Layout
+
+Each `masterId` maps to a single 32-byte storage slot:
+
+```
+slot = keccak256(masterId, REGISTRY_BASE)
+value = masterAddress | masterType | reserved
```
danrobinson: I think should put the masterType at the end, since not all master IDs will start with a 20-byte address (so it might interfere with the layout of others)

tips/tip-1022.md
```diff
@@ -0,0 +1,601 @@
...
+
+```
+Virtual address = 0xFDFDFDFDFDFDFDFDFDFD  07A3B1C2D4E5  a7c3f19e
+                  ^^^^^^^^^^^^^^^^^^^^    ^^^^^^^^^^^^  ^^^^^^^^
+                  prefix (10)             masterId (6)  userTag (4)
+```
+
+## Registry Precompile
+
+Virtual address resolution requires a registry that maps `masterId → masterAddress`. This is managed through a new precompile.
```
danrobinson: Should define the precompile address here. Maybe `0xfdc0000000000...`

tips/tip-1022.md
```diff
@@ -0,0 +1,601 @@
...
+```
+require(bytes4(registrationHash[0:4]) == 0x00000000)  // 32-bit PoW
+masterId = bytes6(registrationHash[4:10])
+```
+
+This requires the caller to grind ~2³² salt values to find a valid registration. If the first 4 bytes are not zero, the call reverts with `ProofOfWorkFailed`. This Proof-Of-Work renders certain attacks computationally infeasible while still being straighforward for a real user to register a masterId.
+
+### User Tag Derivation (Off-Chain)
+
+The `userTag` is an opaque 4-byte value generated off-chain by the exchange/processor. The protocol does not interpret or validate it. It exists solely so the exchange can attribute deposits to specific users via the two-hop `Transfer` events (see **Event Emission**).
```
danrobinson: I think could generalize the references to "exchange" in most of the body of the TIP to "recipient"

tips/tip-1022.md
```diff
@@ -0,0 +1,601 @@
...
+
+Exchanges maintain their own internal mapping `{internalUserId → virtualAddress}`. No on-chain transaction is needed to create a new deposit address.
+
+Operators SHOULD assign userTags randomly with collision checks to reduce third-party enumeration of deposit addresses.
```
danrobinson: I think let's remove this recommendation—it really depends on the use case

tips/tip-1022.md
```diff
@@ -0,0 +1,601 @@
...
+
+### Worked Example
+
+An exchange with master address `0xABCD...1234` registers with a salt that satisfies the 32-bit PoW:
```
danrobinson, line 126: Fine for this example to be an "exchange", was just referring to the stuff above

tips/tip-1022.md
```diff
@@ -0,0 +1,601 @@
...
+}
+```
+
+### Security Considerations for Master Addresses
+
+The master address has full control over forwarding: it can rotate the master to a new address immediately. If the master key is compromised, an attacker can redirect future deposits instantly. TIP-1022 intentionally does not impose protocol-level timelocks or pause mechanisms — these concerns are best managed by the master address itself (e.g., via multisig wallets, hardware security modules, or governance contracts that enforce their own cooldown periods).
+
+Operators registering as virtual address masters SHOULD use appropriately hardened key management commensurate with the value flowing through their deposit addresses. A multisig or threshold signature scheme is strongly recommended, as it provides both access control and implicit cooldown via multi-party coordination.
+
+### Master Rotation Consent
```
danrobinson: Hmm, this whole feature with signed consents feels like overkill and also overoptimizing for ECDSA. Should we just allow forwarding addresses to send to non-consenting addresses?

danrobinson: For one thing, I think it is probably a nice feature to allow forwarding addresses to forward to contracts. Indeed, the previous section explicitly assumes master addresses can be multisigs!

tips/tip-1022.md
```diff
@@ -0,0 +1,601 @@
...
+
+### Constants
+
+| Name | Value | Description |
+|------|-------|-------------|
+| `VIRTUAL_PREFIX` | `0xFDFDFDFDFDFDFDFDFDFD` | 10-byte prefix identifying virtual addresses |
+
+## Transfer Path Modification
+
+The following existing TIP-20 entrypoints are modified to detect virtual addresses and resolve them before crediting:
```
danrobinson: Should clarify here that the only updates to these is how the `to` address is resolved, not the `from` address on `transferFrom`, `transferFromWithMemo`, or `systemTransferFrom`

tips/tip-1022.md
```diff
@@ -0,0 +1,601 @@
...
+
+#### Execution Flow for `transferFromVirtual` / `transferFromVirtualWithMemo`
+
+1. **Validate source**: `virtualFrom` MUST match `VIRTUAL_PREFIX` (`InvalidVirtualSender` on failure).
+2. **Resolve source master**: resolve `virtualFrom` to its current master; revert with `VirtualAddressUnregistered` if not registered.
+3. **Verify ownership**: resolved master MUST equal `msg.sender` (`NotVirtualSenderOwner` otherwise).
+4. **Resolve recipient**: compute `effectiveRecipient = resolveRecipient(to)`.
+5. **TIP-403 checks**:
+   - **Sender check**: check `msg.sender` (the resolved master) per standard TIP-403 sender rules.
+   - **Recipient check (two-level if virtual)**: if `to` is virtual, check the destination masterId policy address, then the destination master account. If `to` is non-virtual, check `effectiveRecipient` only.
```
danrobinson: should "policy address" here just say "policy"?

tips/tip-1022.md
```diff
@@ -0,0 +1,601 @@
...
+
+If the current master sends tokens to one of its own virtual addresses, the transfer resolves back to the master, effectively a transfer to self. The standard TIP-20 self-transfer semantics apply (no net balance change). The two-hop `Transfer` events are still emitted per invariant #4: `Transfer(master, virtualAddress, amount)` followed by `Transfer(virtualAddress, master, amount)`. Indexers SHOULD NOT interpret this as net inflow when `from == masterAddress` in the first hop.
+
+## Interaction with TIP-403 Policies
+
+Virtual address resolution happens **before** TIP-403 transfer authorization checks. Policy evaluation uses a two-level model combining the master account's policy with an optional per-masterId policy.
+
+### Per-MasterId Policies (requires TIP-1023)
+
+A master MAY set a TIP-403 account-level policy scoped to a specific `masterId` by setting a policy on the **masterId policy address**, a synthetic address constructed as:
```
danrobinson: IMO we should drop this concept of master ID policies and only check the 403 receive policies on the master address. Transfers to virtual addresses are already expensive enough and this ups the worst-case gas cost of transfers even more. If someone wants to have different master IDs with different 403 receive policies, they should use different master addresses for them.

tips/tip-1022.md
```diff
@@ -0,0 +1,601 @@
...
+    ///      Reverts with InvalidNewMaster if newMaster equals current master (no-op).
+    ///      Reverts with InvalidMasterConsent / MasterConsentExpired if consent is invalid.
+    ///      newMaster MAY already be the active master for other masterIds (many-to-one).
+    ///      Consent MUST be EIP-712 over (chainId, registryAddress, masterId, currentMaster,
+    ///      newMaster, deadline), signed by newMaster.
+    /// @param masterId The master identifier to rotate.
+    /// @param newMaster The new master address.
+    /// @param deadline Signature expiration timestamp.
+    /// @param newMasterSig EIP-712 signature by newMaster consenting to this rotation.
+    function updateMaster(bytes6 masterId, address newMaster, uint256 deadline, bytes calldata newMasterSig)
```
danrobinson: Hmm, I wonder if we shouldn't even have `updateMaster`. Adds a lot of complexity and the use case isn't that clear to me. It also seems weird from a 403 perspective that if a token blacklists a master address, they can just rotate their key and then their master ID will start working again. So if we allow updating master, we should probably allow token-level policies to apply to master IDs. But I might rather just make master IDs immutable

tips/tip-1022.md
```diff
@@ -0,0 +1,601 @@
...
+- `transferFrom`
+- `transferWithMemo`
+- `transferFromWithMemo`
+- `mint`
+- `mintWithMemo`
+- `systemTransferFrom`
+
+Additionally, TIP-1022 introduces the following **new** TIP-20 entrypoints:
+
+- `transferFromVirtual` (see **Master-Controlled Virtual Source Transfers**)
```
danrobinson: Should we consider calling these `transferThroughVirtual`? Since they don't use the approve pattern

tips/tip-1022.md
```diff
@@ -0,0 +1,601 @@
...
+  3. `TransferWithMemo(address(0), masterAddress, amount, memo)`
+  4. `Mint(masterAddress, amount)`
+
+- `transferFromVirtual`:
+  - when `to` is non-virtual:
+    1. `Transfer(masterAddress, virtualFrom, amount)`
+    2. `Transfer(virtualFrom, to, amount)`
+  - when `to` is virtual (both source and destination are virtual):
+    1. `Transfer(masterAddress, virtualFrom, amount)`
+    2. `Transfer(virtualFrom, virtualTo, amount)`
```
danrobinson: Should consider the special case where `virtualFrom` and `virtualTo` are the same address

danrobinson: Probably this should just be a transfer to itself in this case

tips/tip-1022.md
```diff
@@ -0,0 +1,601 @@
...
+
+- `mint`:
+  1. `Transfer(address(0), virtualAddress, amount)`
+  2. `Transfer(virtualAddress, masterAddress, amount)`
+  3. `Mint(masterAddress, amount)`
+
+- `mintWithMemo` (matches TIP-20 ordering: Transfer → TransferWithMemo → Mint):
+  1. `Transfer(address(0), virtualAddress, amount)`
+  2. `Transfer(virtualAddress, masterAddress, amount)`
+  3. `TransferWithMemo(address(0), masterAddress, amount, memo)`
```
danrobinson: Hmm, I think the TransferWithMemo address should use the virtual address, not the master address. That carries more information

tips/tip-1022.md
```diff
@@ -0,0 +1,601 @@
...
+- **`balanceOf(virtualAddress)`**: Always returns 0. Virtual addresses do not hold balances.
+- **Nonce**: Virtual addresses cannot sign transactions (no private key controls the forwarding). They have no nonce.
+- **Composability helpers**: Integrators SHOULD use `effectiveRecipient`, `balanceOfEffective`, and `isTransferAuthorizedEffective` for preflight checks that match execution semantics.
+- **`transferFromVirtual` accounting**: Debits always come from `masterAddress`, never from `virtualFrom`.
+
+---
+
+## Risks and Limitations
+
+### Address Poisoning
```
danrobinson: I wonder—if we put the master ID as the last 6 bytes rather than putting the userTag at the end, maybe that makes address poisoning slightly harder?

malleshpai: i was thinking about this but in some cases it might make it even easier (e.g., i get an address at the same exchange/ recipient, trick you into sending to that instead of your deposit addresss)

tips/tip-1022.md
```diff
@@ -0,0 +1,505 @@
...
+    /// @param salt Caller-chosen salt for masterId derivation. Must satisfy 32-bit PoW.
+    /// @return masterId The derived master identifier.
+    function registerVirtualMaster(bytes32 salt) external returns (bytes6 masterId);
+
+    // ────────────────── Queries ─────────────────────
+
+    /// @notice Returns the current master address for a given masterId, or address(0) if unregistered.
+    function getMaster(bytes6 masterId) external view returns (address);
+
+    /// @notice Returns all active masterIds for a given master address.
```
danrobinson: Should we consider having a view function to get just one? In case someone has way too many and someone wants to loop over them. Idk

tips/tip-1022.md
```diff
@@ -0,0 +1,505 @@
...
+    /// @return masterId The 6-byte master identifier (zero if not virtual).
+    /// @return userTag The 4-byte user tag (zero if not virtual).
+    function decodeVirtualAddress(address addr)
+        external pure returns (bool isVirtual, bytes6 masterId, bytes4 userTag);
+}
+```
+
+### Security Considerations for Master Addresses
+
+The master address is permanently bound to its masterId at registration. If the master key is compromised, deposits to virtual addresses under that masterId cannot be redirected. Operators registering as virtual address masters SHOULD use appropriately hardened key management commensurate with the value flowing through their deposit addresses. A multisig or threshold signature scheme is strongly recommended.
```
danrobinson: > A multisig or threshold signature scheme is strongly recommended.

I would just say that for high-value use cases, users could use multisig contracts for their master address.

tips/tip-1022.md
```diff
@@ -0,0 +1,505 @@
...
+
+### View/Execution Symmetry (Composability Helpers)
+
+To avoid view/execution mismatches for routers, vaults, and integrators, TIP-20 implementation post adoption of TIP-1022 MUST expose helper views that mirror execution-time recipient resolution:
+
+```solidity
+interface ITIP20VirtualViews {
+    /// @notice Returns execution-equivalent recipient resolution for `to`.
+    /// @dev Returns `to` for non-virtual addresses; resolves/reverts exactly like transfer paths for virtual addresses.
+    function effectiveRecipient(address to) external view returns (address);
```
danrobinson: Don't we already have resolveRecipient? Not sure it makes sense to add this to TIP-20 if we don't have to https://github.com/tempoxyz/tempo/pull/2852/changes#diff-65b4385445b1d186fb6caad53ec6f82496fc80c68d97bf282c29baa4d358c591R204?

tips/tip-1022.md
```diff
@@ -0,0 +1,505 @@
...
+
+    /// @notice Returns balance using execution-equivalent recipient resolution.
+    /// @dev Equivalent to balanceOf(effectiveRecipient(account)).
+    function balanceOfEffective(address account) external view returns (uint256);
```
danrobinson: I don't really think this makes sense—there's no sense in which the balance of a virtual address is the balance of the master address

tips/tip-1022.md
```diff
@@ -0,0 +1,505 @@
...
+
+    /// @notice Returns TIP-403 transfer authorization using execution-equivalent recipient resolution.
+    /// @dev Equivalent to isTransferAuthorized(from, effectiveRecipient(to)).
+    function isTransferAuthorizedEffective(address from, address to) external view returns (bool);
```
danrobinson: I think the TIP-403 isTransferAuthorized should just account for virtual address resolution?

tips/tip-1022.md
```diff
@@ -0,0 +1,505 @@
...
+```
+
+The execution flow for `transferThroughVirtual` and `transferThroughVirtualWithMemo` is specified in **Execution Flow for `transferThroughVirtual` / `transferThroughVirtualWithMemo`** above.
+
+### Gas Overhead
+
+| Path | Additional cost |
+|------|----------------|
+| Non-virtual transfer | One 10-byte prefix comparison; no registry lookup |
+| Virtual transfer (cold) | ~2,200 gas (one cold registry metadata lookup) |
```
danrobinson: There's an extra event too. I think I would just get rid of the gas section and we derive that from what actually is implemented?

tips/tip-1022.md
malleshpai: done.

tips/tip-1022.md
```diff
@@ -0,0 +1,505 @@
...
+#### Entrypoint-Specific Event Ordering
+
+- `transfer`, `transferFrom`, `systemTransferFrom`:
+  1. `Transfer(sender, virtualAddress, amount)`
+  2. `Transfer(virtualAddress, masterAddress, amount)`
+
+- `transferWithMemo`, `transferFromWithMemo`:
+  1. `Transfer(sender, virtualAddress, amount)`
+  2. `Transfer(virtualAddress, masterAddress, amount)`
+  3. `TransferWithMemo(sender, virtualAddress, amount, memo)`
```
danrobinson: I think maybe put the TransferWithMemo before the second Transfer; that enforces the invariant that it is always immediately after the matching Transfer

tips/tip-1022.md
```diff
@@ -0,0 +1,505 @@
...
+
+- `mint`:
+  1. `Transfer(address(0), virtualAddress, amount)`
+  2. `Transfer(virtualAddress, masterAddress, amount)`
+  3. `Mint(masterAddress, amount)`
+
+- `mintWithMemo` (matches TIP-20 ordering: Transfer → TransferWithMemo → Mint):
+  1. `Transfer(address(0), virtualAddress, amount)`
+  2. `Transfer(virtualAddress, masterAddress, amount)`
+  3. `TransferWithMemo(address(0), virtualAddress, amount, memo)`
```
danrobinson: Same here

tips/tip-1022.md
```diff
@@ -0,0 +1,505 @@
...
+  - when `to` is virtual (both source and destination are virtual):
+    1. `Transfer(masterAddress, virtualFrom, amount)`
+    2. `Transfer(virtualFrom, virtualTo, amount)`
+    3. `Transfer(virtualTo, recipientMaster, amount)`
+
+- `transferThroughVirtualWithMemo`:
+  - when `to` is non-virtual:
+    1. `Transfer(masterAddress, virtualFrom, amount)`
+    2. `Transfer(virtualFrom, to, amount)`
+    3. `TransferWithMemo(virtualFrom, to, amount, memo)`
```
danrobinson: Same here

tips/tip-1022.md
```diff
@@ -0,0 +1,505 @@
...
+
+## Self-Forwarding
+
+If the current master sends tokens to one of its own virtual addresses, the transfer resolves back to the master, effectively a transfer to self. The standard TIP-20 self-transfer semantics apply (no net balance change). The two-hop `Transfer` events are still emitted per invariant #4: `Transfer(master, virtualAddress, amount)` followed by `Transfer(virtualAddress, master, amount)`. Indexers SHOULD NOT interpret this as net inflow when `from == masterAddress` in the first hop.
+
+## Interaction with TIP-403 Policies
+
+Virtual address resolution happens **before** TIP-403 transfer authorization checks. Policy evaluation uses the resolved `masterAddress` — not the virtual address.
+
+- If the **master account** is sanctioned/blocked, transfers to any of its virtual addresses revert.
```
danrobinson:
```suggestion
- If the **master address** is not authorized to receive the token, transfers to any of its virtual addresses revert.
```

tips/tip-1022.md
```diff
@@ -0,0 +1,505 @@
...
+- If the **sender** is sanctioned/blocked, the transfer reverts (existing TIP-403 behavior).
```
danrobinson:
```suggestion
- If the **sender** is not authorized to send the token, the transfer reverts (existing TIP-403 behavior).
```

tips/tip-1022.md
```diff
@@ -0,0 +1,505 @@
...
+- TIP-403 policies set on individual virtual addresses are **ignored**. Virtual addresses have no independent on-chain identity.
+- For `transferThroughVirtual` / `transferThroughVirtualWithMemo`, sender policy checks apply to `msg.sender` (the master), not `virtualFrom`.
+
+## Interaction with Account-Level Features
+
+- **`balanceOf(virtualAddress)`**: Always returns 0. Virtual addresses do not hold balances.
+- **Nonce**: Virtual addresses cannot sign transactions (no private key controls the forwarding). They have no nonce.
```
danrobinson: I think we should specify what happens if, extremely improbably, someone were to send an EOA transaction from a virtual address

tips/tip-1022.md
malleshpai: i couldn't see a use case for this given that most "real" users will probably have some small number (2-3). so skipped. can add in if you insist.

tips/tip-1022.md
```diff
@@ -0,0 +1,470 @@
...
+- `mint`:
+  1. `Transfer(address(0), virtualAddress, amount)`
+  2. `Transfer(virtualAddress, masterAddress, amount)`
+  3. `Mint(masterAddress, amount)`
+
+- `mintWithMemo`:
+  1. `Transfer(address(0), virtualAddress, amount)`
+  2. `TransferWithMemo(address(0), virtualAddress, amount, memo)`
+  3. `Transfer(virtualAddress, masterAddress, amount)`
+  4. `Mint(masterAddress, amount)`
```
danrobinson: Hmm, I think this should probably go before the second transfer, and it should be a mint to the virtual address?

tips/tip-1022.md
```diff
@@ -0,0 +1,470 @@
...
+
+- `transferWithMemo`, `transferFromWithMemo`:
+  1. `Transfer(sender, virtualAddress, amount)`
+  2. `TransferWithMemo(sender, virtualAddress, amount, memo)`
+  3. `Transfer(virtualAddress, masterAddress, amount)`
+
+- `mint`:
+  1. `Transfer(address(0), virtualAddress, amount)`
+  2. `Transfer(virtualAddress, masterAddress, amount)`
+  3. `Mint(masterAddress, amount)`
```
danrobinson: Same as below

tips/tip-1022.md
```diff
@@ -0,0 +1,470 @@
...
+        external
+        returns (bool);
+}
+```
+
+The execution flow for `transferThroughVirtual` and `transferThroughVirtualWithMemo` is specified in **Execution Flow for `transferThroughVirtual` / `transferThroughVirtualWithMemo`** above.
+
+### Event Emission
+
+TIP-1022 does **not** introduce any new events. Instead, virtual address forwarding is represented using **two-hop standard `Transfer` events**: one hop showing funds arriving at the virtual address, and a second hop showing funds moving from the virtual address to the resolved master (or vice versa for `transferThroughVirtual`). This two-hop pattern ensures that existing indexers, block explorers, and analytics tools that track standard ERC-20/TIP-20 `Transfer` events will correctly display the full flow without requiring custom event support.
```
decofe: **Inconsistency: "no new events" vs registry interface.**

This line says TIP-1022 does **not** introduce any new events, but the registry interface above defines `event MasterRegistered(bytes6 indexed masterId, address indexed masterAddress)`.

Suggest scoping this to TIP-20 transfer-path events specifically, e.g.:

> "TIP-1022 forwarding does not introduce new TIP-20 transfer-path events; forwarding is represented via standard `Transfer` (and existing TIP-20 events). The registry precompile emits `MasterRegistered`."

tips/tip-1022.md
```diff
@@ -0,0 +1,470 @@
...
+
+## Core Invariants
+
+1. **No fund loss**: A TIP-20 transfer to a virtual address MUST either credit the registered master's balance by exactly the transfer amount, or revert. Funds MUST NOT be credited to the virtual address itself or lost.
+
+2. **Revert on unregistered**: A transfer to an address matching `VIRTUAL_PREFIX` whose `masterId` is not registered MUST revert. It MUST NOT credit any account.
+
+3. **Balance consistency**: After a successful virtual-forwarded transfer of amount `X`, `balanceOf(master)` MUST have increased by exactly `X`. `balanceOf(virtualAddress)` MUST remain 0.
+
+4. **Event consistency**: Every virtual-forwarded operation MUST emit exactly two `Transfer` events: `Transfer(sender, virtualAddress, amount)` followed by `Transfer(virtualAddress, masterAddress, amount)`. `TransferWithMemo` events MUST use `virtualAddress` as the recipient (preserving deposit attribution); `Mint` events MUST use `masterAddress`.
```
decofe: **Inconsistency: invariant #4 vs `transferThroughVirtual` event specs.**

This says every virtual-forwarded op emits **exactly two** `Transfer` events (`sender→virtual→master`). But the `transferThroughVirtual` entrypoint-specific event ordering above shows:
- When `to` is non-virtual: the second hop is `virtualFrom→to` (not `virtualFrom→masterAddress`)
- When `to` is virtual: **three** `Transfer` events are emitted

This invariant is only correct for inbound deposit forwarding on standard entrypoints. Suggest either scoping it to recipient-forwarding on standard entrypoints, or splitting into separate invariants for inbound forwarding vs virtual-source transfers.

tips/tip-1022.md
```diff
@@ -0,0 +1,470 @@
...
+
+```
+[10-byte PREFIX] [6-byte masterId] [4-byte userTag]
+= 20 bytes total
+```
+
+| Field | Bytes | Description |
+|-------|-------|-------------|
+| **PREFIX** | 10 | Fixed prefix `0xFDFDFDFDFDFDFDFDFDFD`. Identifies virtual addresses. |
+| **masterId** | 6 | Deterministic identifier derived from the master address. Supports up to ~2.8×10¹⁴ masters. |
```
decofe: **Inconsistency: `masterId` derivation description.**

"Derived from the master address" implies a function of the address alone (1:1). But the spec derives it from `(msg.sender, salt)` via `keccak256(abi.encodePacked(msg.sender, salt))[4:10]`, and explicitly allows multiple masterIds per address via different salts.

Suggest: "Deterministic identifier derived from `(masterAddress, salt)` via registration PoW hash."

tips/tip-1022.md
```diff
@@ -0,0 +1,470 @@
...
+                  prefix (10)             masterId (6)  userTag (4)
+```
+
+## Registry Precompile
+
+Virtual address resolution requires a registry that maps `masterId → masterAddress`. This is managed through a new precompile deployed at  `0xFDC0000000000000000000000000000000000000`.
+
+The registry MUST maintain the following mapping constraints:
+
+- each `masterId` maps to at most one active master address (one-to-one from masterId)
```
decofe: **Consistency note: "active" implies lifecycle transitions that do not exist.**

This and several other places use "active master address" / "current master" / "active masterIds" — but the spec provides no update or unregister function, and the Security Considerations section says the master is "permanently bound" and "cannot be redirected."

If immutability is intended, suggest replacing "active" / "current" with "registered" throughout (lines 122, 146, 195, 198, 200, 206, 210, 288, 399, 459, 463) to avoid implying a mutable lifecycle.

decofe (commented):
A few internal inconsistencies flagged below.

tips/tip-1022.md
```diff
@@ -0,0 +1,470 @@
...
+
+- If the **master address** is not authorized to receive the token, transfers to any of its virtual addresses revert.
+- If the **sender** is not authorized to send the token, the transfer reverts (existing TIP-403 behavior).
+- TIP-403 policies set on individual virtual addresses are **ignored**. Virtual addresses have no independent on-chain identity.
+- For `transferThroughVirtual` / `transferThroughVirtualWithMemo`, sender policy checks apply to `msg.sender` (the master), not `virtualFrom`.
+
+## Interaction with Account-Level Features
+
+- **`balanceOf(virtualAddress)`**: Always returns 0. Virtual addresses do not hold balances.
+- **Nonce**: Virtual addresses are not expected to originate transactions. However, if someone were to control a private key whose address matches `VIRTUAL_PREFIX` (requiring ~2⁸⁰ work), EOA transactions sent *from* that address would execute normally — TIP-1022 resolution applies only to the `to` field in TIP-20 precompile calls, not to transaction senders.
```
decofe: Consider dropping this bullet entirely. Virtual addresses are just regular 20-byte addresses — if someone controls a matching key, it's a normal EOA with a nonce. TIP-1022 only affects TIP-20 recipient resolution, not account-level properties. Stating anything about nonces here risks being misleading, and the qualifying caveat makes it read like an exception to a rule that isn't really a rule.

tips/tip-1022.md
```diff
@@ -0,0 +1,478 @@
...
+
+### Prefix Selection: `0xFDFDFDFDFDFDFDFDFDFD`
+
+A 10-byte prefix provides collision resistance of ~1/2⁸⁰. This is an explicit layout tradeoff that prioritizes address space safety over `masterId` and `userTag` space. The 6-byte `masterId` supports ~2.8×10¹⁴ masters, and the 4-byte `userTag` supports >4 billion (2³²) deposit addresses per master.
+
+### Reserved Address Space
+
+Any address whose first 10 bytes match `VIRTUAL_PREFIX` is treated as a virtual address by the TIP-20 precompile. If a user happens to control a private key whose corresponding address begins with `0xFDFDFDFDFDFDFDFDFDFD`, TIP-20 transfers to that address will be subject to TIP-1022 resolution: they will forward to the registered master (if the embedded `masterId` is registered) or revert (if it is not). The address will not accumulate a TIP-20 balance through normal transfers. Users who have such an address SHOULD not use it on Tempo — see **Risks and Limitations**.
```
samczsun: nit: should be "SHOULD NOT"

tips/tip-1022.md
```diff
@@ -0,0 +1,478 @@
...
+
+The first 4 bytes of `registrationHash` are consumed by the proof-of-work check (see **Registration Proof of Work**); the `masterId` is extracted from bytes `[4:10]` of the same hash.
+
+The salt is a `bytes32` value chosen by the caller. Callers MUST grind the salt to satisfy the 32-bit proof-of-work requirement (see below). The `masterId` is permanently bound to the **registration address**.
+
+In the event of a masterId collision (two `(address, salt)` pairs mapping to the same 6-byte masterID), the second registration reverts with `MasterIdCollision`. The caller can retry with a different salt.
+
+### Registration Proof of Work
+
+Registration requires a 32-bit proof of work. This makes it computationally expensive for an attacker to frontrun a registration transaction, without lengtheing the masterID.
```
samczsun: nit: this is not strictly true, the expensive part comes from colliding the `masterId`

tips/tip-1022.md
```diff
@@ -0,0 +1,478 @@
...
+```
+
+The first 4 bytes of `registrationHash` MUST be zero:
+
+```
+require(bytes4(registrationHash[0:4]) == 0x00000000)  // 32-bit PoW
+masterId = bytes6(registrationHash[4:10])
+```
+
+This requires the caller to grind ~2³² salt values to find a valid registration. If the first 4 bytes are not zero, the call reverts with `ProofOfWorkFailed`. This Proof-Of-Work renders certain attacks computationally infeasible while still being straightforward for a real user to register a masterId.
```
samczsun: nit: repetitive, and also vague on which certain attacks

tips/tip-1022.md
```diff
@@ -0,0 +1,478 @@
...
+# Specification
+
+## Address Layout
+
+Virtual addresses are standard 20-byte EVM addresses with a reserved prefix: 
+
+```
+[10-byte PREFIX] [6-byte masterId] [4-byte userTag]
+= 20 bytes total
+```
```
samczsun, line 36: What if we lay it out as: `[6 byte masterId][8 byte magic][6 byte userId]`? The rationale is that the magic bytes are visual noise, and wastes valuable ui space for address comparisons. 8 bytes of magic is more than enough to mitigate accidentally generating a virtual address, and even if an attacker grinded one out, it's useless

tips/tip-1022.md
```diff
@@ -0,0 +1,478 @@
...
+- MUST NOT be `address(0)`
+- MUST NOT match `VIRTUAL_PREFIX`
+- MUST NOT be a TIP-20 token address (i.e., MUST pass TIP-20 recipient validation for transfer paths)
+
+### Registry Storage Layout
+
+Each `masterId` maps to a single 32-byte storage slot:
+
+```
+slot = keccak256(masterId, REGISTRY_BASE)
```
samczsun: What is `REGISTRY_BASE`?

tips/tip-1022.md
```diff
@@ -0,0 +1,478 @@
...
+```
+[10-byte PREFIX] [6-byte masterId] [4-byte userTag]
+= 20 bytes total
+```
+
+| Field | Bytes | Description |
+|-------|-------|-------------|
+| **PREFIX** | 10 | Fixed prefix `0xFDFDFDFDFDFDFDFDFDFD`. Identifies virtual addresses. |
+| **masterId** | 6 | Deterministic identifier derived from `(masterAddress, salt)` via registration PoW hash. Supports up to ~2.8×10¹⁴ masters. |
+| **userTag** | 4 | Opaque per-user identifier, derived off-chain. 32-bit space supports >4 billion (2³²) unique deposit addresses per master. |
```
samczsun: nit: maybe this should be like `masterTag` and `userTag` for symmetry?

tips/tip-1022.md
```diff
@@ -0,0 +1,478 @@
...
+
+### Contracts and EOAs at virtual-prefixed addresses 
+
+It is theoretically possible to deploy a contract or control an EOA whose address matches VIRTUAL_PREFIX, including by grinding CREATE2 salts or private keys. Such addresses can still exist and originate ordinary EVM transactions, but we consider this unlikely in practice because targeting the 10-byte prefix requires roughly 2^80 work, with additional cost for targeted collisions against registered virtual namespaces.
+
+---
+
+# Invariants
+
+## Core Invariants
```
legion2002, line 453: One nice invariant to add here, is that all virtual addresses should have 0 TIP 20 balance.

tips/tip-1022.md
```diff
@@ -0,0 +1,478 @@
...
+If the registered master sends tokens to one of its own virtual addresses, the transfer resolves back to the master, effectively a transfer to self. The standard TIP-20 self-transfer semantics apply (no net balance change). The two-hop `Transfer` events are still emitted per invariant #4: `Transfer(master, virtualAddress, amount)` followed by `Transfer(virtualAddress, master, amount)`. Indexers SHOULD NOT interpret this as net inflow when `from == masterAddress` in the first hop.
+
+## Interaction with TIP-403 Policies
+
+Virtual address resolution happens **before** TIP-403 transfer authorization checks. Policy evaluation uses the resolved `masterAddress` — not the virtual address.
+
+- If the **master address** is not authorized to receive the token, transfers to any of its virtual addresses revert.
+- If the **sender** is not authorized to send the token, the transfer reverts (existing TIP-403 behavior).
+- TIP-403 policies set on individual virtual addresses are **ignored**. Virtual addresses have no independent on-chain identity.
+- For `transferThroughVirtual` / `transferThroughVirtualWithMemo`, sender policy checks apply to `msg.sender` (the master), not `virtualFrom`.
```
legion2002: This makes sense. There's a small footgun that offchain entities who are not aware of this behaviour, blacklist/whitelist virtual addresses instead of master addresses, because that's what they see on the explorer.

Maybe we should revert when someone tries to update a policy with a virtual address.

tips/tip-1022.md
```diff
@@ -0,0 +1,478 @@
...
+### Constants
+
+| Name | Value | Description |
+|------|-------|-------------|
+| `VIRTUAL_PREFIX` | `0xFDFDFDFDFDFDFDFDFDFD` | 10-byte prefix identifying virtual addresses |
+| `REGISTRY_ADDRESS` | `0xFDC0000000000000000000000000000000000000` | Precompile address for the virtual address registry |
+
+## Transfer Path Modification
+
+The following existing TIP-20 entrypoints are modified to resolve the `to` (recipient) address before crediting. The `from` address on `transferFrom`, `transferFromWithMemo`, and `systemTransferFrom` is **not** affected by TIP-1022 resolution.
```
samczsun: what about `setRewardRecipient`?

tips/tip-1022.md
```diff
@@ -0,0 +1,478 @@
...
+        revert VirtualAddressUnregistered()
+
+    return master
+```
+
+#### Execution Flow for Standard Entrypoints
+
+For `transfer`, `transferFrom`, `transferWithMemo`, `transferFromWithMemo`, `mint`, `mintWithMemo`, `systemTransferFrom`:
+
+1. **Resolve recipient**: compute `effectiveRecipient = resolveRecipient(to)`. If `to` is not virtual, `effectiveRecipient = to` (fast path).
```
samczsun: is there a slow path somewhere?

tips/tip-1022.md
```diff
@@ -0,0 +1,478 @@
...
+
+    /// @notice Like `transferThroughVirtual`, but with an attached memo for reconciliation.
+    /// @dev Same validation as `transferThroughVirtual`. Emits additional `TransferWithMemo`.
+    function transferThroughVirtualWithMemo(address virtualFrom, address to, uint256 amount, bytes32 memo)
+        external
+        returns (bool);
+}
+```
+
+The execution flow for `transferThroughVirtual` and `transferThroughVirtualWithMemo` is specified in **Execution Flow for `transferThroughVirtual` / `transferThroughVirtualWithMemo`** above.
```
samczsun: nit: feels like this section should just come first so the reader doesn't jump back up

tips/tip-1022.md
```diff
@@ -0,0 +1,478 @@
...
+    2. `Transfer(virtualFrom, to, amount)`
+  - when `to` is virtual (both source and destination are virtual):
+    1. `Transfer(masterAddress, virtualFrom, amount)`
+    2. `Transfer(virtualFrom, virtualTo, amount)`
+    3. `Transfer(virtualTo, recipientMaster, amount)`
+
+- `transferThroughVirtualWithMemo`:
+  - when `virtualFrom == to`: self-transfer. No net balance change. Events:
+    1. `Transfer(masterAddress, virtualFrom, amount)`
+    2. `TransferWithMemo(masterAddress, virtualFrom, amount, memo)`
```
samczsun: maybe there should be a third `Transfer(virtualFrom, to, amount)` and corresponding `TransferWithMemo(virtualFrom, virtualFrom, amount, memo)`? in which case this becomes the "to is virtual" case rather than special casing it for some reason

tips/tip-1022.md
```diff
@@ -0,0 +1,478 @@
...
+status: Draft
+related: TIP-20, TIP-403, TIP-1028
+protocolVersion: TBD (requires hardfork)
+---
+
+# TIP-1022: Virtual Addresses for TIP-20 Deposit Forwarding
+
+## Abstract
+
+This TIP introduces **virtual addresses**: a reserved address prefix that, when detected in TIP-20 recipient-bearing operations, causes the precompile to auto-credit a registered master wallet instead of the target address. This eliminates sweep transactions entirely for entities such as exchanges, ramps, and payment processors that generate per-user deposit addresses. Master registration is a one-time on-chain call; deposit address derivation is fully off-chain.
```
horsefacts:
```suggestion
This TIP introduces **virtual addresses**: a reserved address prefix that, when detected in TIP-20 recipient-bearing operations, causes the precompile to auto-credit a registered master address instead of the target address. This eliminates sweep transactions entirely for entities such as exchanges, ramps, and payment processors that generate per-user deposit addresses. Master registration is a one-time on-chain call; deposit address derivation is fully off-chain.
```

tips/tip-1022.md
horsefacts: agree, and the attack is not really frontrunning, but precomputing, right? what this does is make it more computationally expensive to compute collisions.

horsefacts: ah, I see what is meant by "frontrunning" (compute it first), but the term is slightly misleading in this context.

malleshpai: sorry just to clarify--- the attacke we're concenred about is this. you have a compute a valid salt and try to get a masterID. i see your tx in the mempool, and work out a salt that gets me (with my account address) the same masterID, then put in a higher tip etc than you. in *principle* this could lead to the same issue as with policies where you think you own a policy but someone else does.

tips/tip-1022.md
malleshpai: well if the address is a virtual address you have to do one more sload to resolve it. will delete :)

tips/tip-1022.md
malleshpai: done. see new invariant #4.

tips/tip-1022.md
malleshpai: updated to resolve this  in subsection "Rejection of Virtual Addresses in Policy Operations"

tips/tip-1022.md
malleshpai: heh yes fixed (but obv the whole para was reworded given the change in spec)

tips/tip-1022.md
malleshpai: reworked

tips/tip-1022.md
malleshpai: 4-10-6 was where we landed, like we discussed.

tips/tip-1022.md
malleshpai: edited. maybe too much detail for this spec?

tips/tip-1022.md
malleshpai: added this language: 
`setRewardRecipient` is **not** a TIP-20 transfer-path operation and is therefore not subject to TIP-1022 recipient resolution. Implementations SHOULD reject virtual addresses when setting reward recipients so that rewards remain tied to canonical accounts rather than aliases.

tips/tip-1022.md
malleshpai: done.

tips/tip-1022.md
```diff
@@ -0,0 +1,561 @@
...
+
+TIP-403 / TIP-1028 operations that accept addresses as policy members MUST reject virtual addresses rather than accepting them silently.
+
+Implementations SHOULD use a clear, informative error indicating that virtual addresses are aliases for TIP-20 forwarding and are not valid literal policy subjects.
+
+Rejecting these operations avoids the footgun where an operator configures policy on the virtual alias they see in logs or explorers instead of on the resolved master address that actually holds the funds.
+
+## Interaction with Account-Level Features
+
+- **`balanceOf(virtualAddress)`**: Always returns 0. Virtual addresses do not hold balances.
```
0xrusowsky, line 383: based on the spec, i understand that this section aims to explicitly list an invariant rather than something we should explicitly enforce, right?

tips/tip-1022.md
```diff
@@ -0,0 +1,561 @@
...
+- Policies configured on individual virtual addresses are ignored by the TIP-20 transfer path because virtual addresses have no independent canonical TIP-20 balance.
+- For `transferThroughVirtual` / `transferThroughVirtualWithMemo`, sender policy checks apply to `msg.sender` (the master), not `virtualFrom`.
+
+### Rejection of Virtual Addresses in Policy Operations
+
+TIP-403 / TIP-1028 operations that accept addresses as policy members MUST reject virtual addresses rather than accepting them silently.
+
+Implementations SHOULD use a clear, informative error indicating that virtual addresses are aliases for TIP-20 forwarding and are not valid literal policy subjects.
+
+Rejecting these operations avoids the footgun where an operator configures policy on the virtual alias they see in logs or explorers instead of on the resolved master address that actually holds the funds.
```
0xrusowsky, line 379: should the error resolve the master address for better UX?

tips/tip-1022.md
```diff
@@ -0,0 +1,561 @@
...
+
+### Why the Magic Bytes Are in the Middle
+
+The middle `VIRTUAL_MAGIC` layout is a deliberate usability tradeoff:
+
+- it leaves the first 4 bytes available for `masterId`
+- it leaves the last 6 bytes available for `userTag`
+- it avoids spending the most visually important bytes of the address on static marker data
+
+This improves address comparison in UIs while still keeping a large reserved pattern that is highly unlikely to appear accidentally. We believe this layout is superior to the other permutations in terrms of the prospect of address poisoning attacks (see below).
```
klkvr:
```suggestion
This improves address comparison in UIs while still keeping a large reserved pattern that is highly unlikely to appear accidentally. We believe this layout is superior to the other permutations in terms of the prospect of address poisoning attacks (see below).
```

tips/tip-1022.md
```diff
@@ -0,0 +1,561 @@
...
+### Address Poisoning and UI Confusion
+
+TIP-1022 still introduces a recognizable structured address format. Wallets, block explorers, and operational tooling that truncate addresses SHOULD display enough of the address to distinguish both the `masterId` and the `userTag`; ideally they SHOULD show the full address.
+
+
+### Non-TIP-20 Token Loss
+
+TIP-1022 forwarding applies exclusively to TIP-20 precompile operations. Non-TIP-20 tokens (e.g. ERC-20 contracts deployed on Tempo) transferred to a virtual address are credited to the literal virtual address by the ERC-20 contract and are irrecoverable: no recovery mechanism is defined here.
+
+This risk is mitigated by the strong incentives for token issuers to use TIP-20 on Tempo (gas-payment eligibility, access to the payment lane, and policy support), but it remains a limitation of this design.
```
klkvr, line 443: do we encourage even non-stablecoin issuers to use TIP20s? e.g if WETH is gonna be deployed on Tempo

malleshpai: yes. ideally everything will be a TIP-20, where nonstablecoins don't use most of the features (e.g., 403)

tips/tip-1022.md
malleshpai: yes, i think that's what the invariant sections are for (e.g. to generate tests)

tips/tip-1022.md
malleshpai: sure maybe the error can say "this is a virtual address that is owned by this master address, if you wish to block/ whitelist the address please block the underlying master." should i add that to the spec or is this an implementation detail?

0xrusowsky: impl detail, no worries

tips/tip-1022.md
malleshpai: going to punt on this for now, Dan and I are discussing other names, but i'm in the "what's in a name" camp.

tips/tip-1022.md
malleshpai: no strong opinions. this basically mirrors the convention of TransferThrough above with a single memo event. I can "merge" to is virtual and self-transfer cases (across both to make it consistent), but I think we should keep it a single TransferWithMemo event.

samczsun: ah yeah I think there should be a single memo event I just think we should replace the one in spec with the two proposed

malleshpai: yeah makes sense. cc: @danrobinson

tips/tip-1022.md
```diff
@@ -0,0 +1,557 @@
...
+
+- it leaves the first 4 bytes available for `masterId`
+- it leaves the last 6 bytes available for `userTag`
+- it avoids spending the most visually important bytes of the address on static marker data
+
+This improves address comparison in UIs while still keeping a large reserved pattern that is highly unlikely to appear accidentally. We believe this layout is superior to the other permutations in terms of the prospect of address poisoning attacks (see below).
+
+### Contracts and EOAs Matching the Virtual Format
+
+A sufficiently resourced adversary could, in principle, grind a CREATE2 deployment or private key so that a contract or EOA lands at an address matching the virtual-address format in a `masterID` controlled by the adversary. We view this as unlikely in practice because the address must match a 10-byte fixed magic value in the middle of the address, while targeted theft of a specific registered namespace also requires colliding the 4-byte `masterId` under the registration proof-of-work design (i.e., 18 bytes totally).
```
samczsun: 14 bytes

tips/tip-1022.md
```diff
@@ -0,0 +1,557 @@
...
+TIP-1022 still introduces a recognizable structured address format. Wallets, block explorers, and operational tooling that truncate addresses SHOULD display enough of the address to distinguish both the `masterId` and the `userTag`; ideally they SHOULD show the full address.
+
+
+### Non-TIP-20 Token Loss
+
+TIP-1022 forwarding applies exclusively to TIP-20 precompile operations. Non-TIP-20 tokens (e.g. ERC-20 contracts deployed on Tempo) transferred to a virtual address are credited to the literal virtual address by the ERC-20 contract and are irrecoverable: no recovery mechanism is defined here.
+
+This risk is mitigated by the strong incentives for token issuers to use TIP-20 on Tempo (gas-payment eligibility, access to the payment lane, and policy support), but it remains a limitation of this design.
+
+### Contracts and EOAs at virtual-prefixed addresses
```
samczsun: this section still references prefixes

tips/tip-1022.md
```diff
@@ -0,0 +1,557 @@
+---
+id: TIP-1022
+title: Virtual Addresses for TIP-20 Deposit Forwarding
+description: Precompile-native virtual addresses that auto-forward TIP-20 deposits to a registered master wallet, eliminating sweep transactions.
+authors: Dankrad Feist, Liam Horne, Mallesh Pai, Dan Robinson
+status: Draft
+related: TIP-20, TIP-403
```
samczsun: missing TIP-1015

malleshpai: 🤦

tips/tip-1022.md
malleshpai: 🤦

samczsun (approved):

tips/tip-1022.md
```diff
@@ -0,0 +1,549 @@
...
+
+### Standard Transfer Entrypoints
+
+For `transfer`, `transferFrom`, `transferWithMemo`, `transferFromWithMemo`, and `systemTransferFrom`:
+
+1. **Resolve recipient**: compute `effectiveRecipient = resolveRecipient(to)`. If `to` is not virtual, `effectiveRecipient = to`.
+2. **Token-level sender check**: apply the standard TIP-403 / TIP-1015 sender authorization rules.
+3. **Token-level recipient check**: apply the standard TIP-403 / TIP-1015 recipient authorization rules to `effectiveRecipient`.
+4. **Apply balance changes**: debit sender, credit `effectiveRecipient`.
+5. **Emit events**: per **Event Emission** (two-hop `Transfer` if virtual, single `Transfer` otherwise).
```
0xKitsune, line 292: What is the rationale for two `Transfer` events? IIUC a single `Transfer(sender, masterAddress, amount)` wouldn't break indexers as this is where the balance actually lands. The two hop pattern actually makes it harder to infer from with logs what is actually happening. If someone just looks at two `Transfer` events, it could either be a virtual deposit, or it could be two back to back normal transfers. The only way to know if the deposit is virtual in this case is to read storage to see if the virtual address is registered (rather than being able to infer from an explicit log). 

An alternative to consider could be pairing a single `Transfer` with a dedicated `VirtualDeposit(sender, virtualAddress, masterAddress, amount)` attribution event. Exchanges/indexers get the same deposit tracking data (filter on VirtualDeposit to extract the userTag), but every other consumer sees correct balances by default without needing virtual address awareness. 

Feel free to clarify if this approach (or other alternative approaches) were already considered and rejected.

tips/tip-1022.md
```diff
@@ -0,0 +1,549 @@
...
+
+## Conformance and Scope
+
+TIP-1022 applies only to TIP-20 precompile recipient resolution for the entrypoints listed in **Transfer Path Modification**.
+
+TIP-1022 does **not** alter TIP-20 methods that do not carry a recipient in the TIP-20 transfer path (e.g. `approve`, `burn`, `permit`) and does not alter non-TIP-20 protocol behavior.
+
+Non-TIP-20 token transfers (e.g. ERC-20 contracts deployed on Tempo) to virtual addresses are **not** subject to TIP-1022 forwarding. Such transfers behave as standard EVM transfers to the literal address. Tokens sent this way may be irrecoverable — see **Risks and Limitations**.
+
+`setRewardRecipient` is **not** a TIP-20 transfer-path operation and is therefore not subject to TIP-1022 recipient resolution. Implementations SHOULD reject virtual addresses when setting reward recipients so that rewards remain tied to canonical accounts rather than aliases.
```
0xKitsune: > SHOULD

nit: should this be `MUST`? In general it seems like TIPs typically should specify `MUST` or `MUST NOT` for any protocol behavior.

tips/tip-1022.md
```diff
@@ -0,0 +1,549 @@
...
+
+## Reserved Virtual Address Format
+
+Any address whose bytes `[4:14]` equal `VIRTUAL_MAGIC` is treated as a virtual address by the TIP-20 precompile.
+
+If a TIP-20 transfer targets such an address:
+- the precompile extracts the `masterId` from bytes `[0:4]`
```
0xKitsune, line 65: nit: not sure if useful, but it might be worth detailing the probability of collisions and when a user can reasonably expect that they will have to mine for prefix with multiple attempts. I am under the impression this is probably very small (napkin math needs a sanity check but somewhere near < 0.2% chance that user will need to grind a new value at ~10M virtual addresses registered) so feel free to ignore this comment if this is not useful to add to the spec.

tips/tip-1022.md
```diff
@@ -0,0 +1,549 @@
...
+- credits the resolved master if registered
+- otherwise reverts with `VirtualAddressUnregistered`
+
+The literal virtual address never accumulates TIP-20 balance through standard TIP-20 transfer paths.
+
+### Reserved Address Space
+
+Addresses matching the virtual-address format occupy a reserved TIP-20 recipient namespace. A user who happens to control an EOA or contract whose address matches this format can still exist on Tempo and can still originate ordinary EVM transactions. However, TIP-20 transfers to such an address will follow TIP-1022 recipient resolution semantics rather than crediting the literal address.
+
+Users who control such an address SHOULD NOT use it as a normal account on Tempo.
```
0xKitsune, line 76: Do we want to have custom errors in the case that a user tries to use a virtual account as a normal account. For example, if a user sets the tx `from` address as a virtual address, do we want to throw an error at the RPC or protocol level? Or should we just let it fail normally with existing codepaths. The former might make it easier for a user to know why the tx failed, however this might not be worth adding more code to catch special cases.

tips/tip-1022.md
```diff
@@ -0,0 +1,549 @@
...
+```
+
+## Registry Precompile
+
+Virtual address resolution requires a registry that maps `masterId -> masterAddress`. This is managed through a new precompile deployed at `0xFDC0000000000000000000000000000000000000`.
+
+The registry MUST maintain the following mapping constraints:
+
+- each `masterId` maps to at most one registered master address (one-to-one from `masterId`)
+- multiple `masterId`s MAY map to the same master address (many-to-one)
```
0xKitsune, line 146: Is there any way to update to the master address for a given master Id? Is this something that was considered but decided against? If so maybe it makes sense to specify why this is not allowed. Otherwise we should consider adding this if this is desired behavior.  

One example this might be useful is if the master address key gets compromised.

tips/tip-1022.md
```diff
@@ -0,0 +1,549 @@
...
+    /// @dev Same validation as `transferThroughVirtual`. Emits additional `TransferWithMemo`.
+    function transferThroughVirtualWithMemo(address virtualFrom, address to, uint256 amount, bytes32 memo)
+        external
+        returns (bool);
+}
+```
+
+### Execution Flow
+
+For `transferThroughVirtual` and `transferThroughVirtualWithMemo`:
```
0xKitsune: What is the rationale for allowing users to transfer from virtual addresses. Since these addresses cant actually hold balances, it seems a bit unintuitive to allow a user to send from this account. It seems simpler to just treat virtual addresses as a forwarding address only. This also simplifies logic around 403 checks, event emission etc.

tips/tip-1022.md
malleshpai: If the masterAddress is tracking deposits by looking at events, then a single Transfer(sender, masterAddress, amount) won't tell me what virtualAddress (and therefore, userId) this transfer is intended for. The two hop event pattern means the master can just look at events and keep their internal attribution (which userId has what effective balance up to date), while indexers who are using transfer events to track balances also accurate without changing.

tips/tip-1022.md
malleshpai: yup done. sorry!

tips/tip-1022.md
malleshpai: For now, we decided so far not to do this. this is in keeping with, e.g., TIP-20 prefix matching addresses, where we do not currently block it. The 10 byte magic was selected to make this secure enough.

My sense is that we should have some "reserved address policy" and do something like what you suggest at some point.

tips/tip-1022.md
malleshpai: We had it and took it out to simplify the protocol for now. Adding the possibility to update can be added later if we get user demands. Another reason is that it just starts a game of whack-a-mole with 403. for example if a token blacklists a master address, rotation would let that same masterId start working again under a new key unless policies also applied to master IDs.

tips/tip-1022.md
malleshpai: Yup, i get something like that too. Adding a line.

tips/tip-1022.md
malleshpai: (Especially if we add more reserved spaces beyond this and TIP-20)

tips/tip-1022.md
malleshpai: The main reason potentially is refunds (it might be useful to make it look like the money came back from the virtual accounts). I don't have *massive* conviction on demand for this (relative to the core feature which I have a lot of conviction on demand for) so we can pause this / leave it as a phase 2 in a future TIP if you think the complexity should be avoided right now.

tips/tip-1022.md
malleshpai: Agreed after live discussion with @0xKitsune that we should do nothing for now, so I'm resolving this.

tips/tip-1022.md
malleshpai: Line 93 now includes a brief summary of our discussion and the reasons for this choice, for posterity.

tips/tip-1022.md
malleshpai: removed per our discussion.

PR #2858, opened 2026-02-25
fix(validator-config-v2): ed25519 decode check on migration, digest preimage collision fix

howydev:
## Summary

Two fixes to ValidatorConfigV2 (TIP-1017):

### 1. Ed25519 decode check during migration

`migrate_validator` now validates that V1 public keys are decodable Ed25519 curve points, reverting with `InvalidPublicKey` if not. V1 did not enforce this, so invalid keys could have been stored.

### 2. Digest preimage collision fix

The signature digest for `addValidator` and `rotateValidator` used `abi.encodePacked(..., ingress, egress)` which is ambiguous — e.g. `("192.168.0.1:800", "12.3.4.5")` and `("192.168.0.1:8001", "2.3.4.5")` produce the same byte sequence. Fixed by adding a `uint8` length prefix before ingress to disambiguate the boundary.

### Files changed

- **Spec**: `tips/tip-1017.md`
- **Solidity**: `ValidatorConfigV2.sol`, `IValidatorConfigV2.sol`
- **Solidity tests**: `ValidatorConfigV2.t.sol`, `invariants/ValidatorConfigV2.t.sol`
- **Rust**: `crates/precompiles/src/validator_config_v2/mod.rs`, `dispatch.rs`
- **Genesis**: `xtask/src/genesis_args.rs`

### DIFF
--- a/crates/precompiles/src/validator_config_v2/mod.rs
+++ b/crates/precompiles/src/validator_config_v2/mod.rs
@@ -1587,8 +1587,8 @@ mod tests {
             let mut v1 = v1();
             v1.initialize(owner)?;
 
-            // Zero is not a valid Ed25519 curve point.
-            let bad_key = [0u8; 32];
+            // [0x02; 32] is not a valid Ed25519 curve point but is non-zero so V1 accepts it.
+            let bad_key = [0x02u8; 32];
             v1.add_validator(
                 owner,
                 tempo_contracts::precompiles::IValidatorConfig::addValidatorCall {
--- a/crates/precompiles/src/validator_config_v2/mod.rs
+++ b/crates/precompiles/src/validator_config_v2/mod.rs
@@ -344,7 +344,7 @@ impl ValidatorConfigV2 {
     ///
     /// **FORMAT**:
     /// - Namespace: [`VALIDATOR_NS_ADD`] or [`VALIDATOR_NS_ROTATE`]
-    /// - Message: `keccak256(chainId || contractAddr || validatorAddr || uint8(ingress.len) || ingress || egress))`
+    /// - Message: `keccak256(chainId || contractAddr || validatorAddr || uint8(ingress.len) || ingress || uint8(egress.len) || egress))`
     fn verify_validator_signature(
         &self,
         namespace: &[u8],
@@ -360,6 +360,7 @@ impl ValidatorConfigV2 {
         hasher.update(validator_address.as_slice());
         hasher.update([ingress.len() as u8]);
         hasher.update(ingress.as_bytes());
+        hasher.update([egress.len() as u8]);
         hasher.update(egress.as_bytes());
         let message = hasher.finalize();
 
@@ -666,6 +667,7 @@ mod tests {
         data.extend_from_slice(validator_address.as_slice());
         data.push(ingress.len() as u8);
         data.extend_from_slice(ingress.as_bytes());
+        data.push(egress.len() as u8);
         data.extend_from_slice(egress.as_bytes());
         let message = keccak256(&data);
 
--- a/crates/precompiles/src/validator_config_v2/mod.rs
+++ b/crates/precompiles/src/validator_config_v2/mod.rs
@@ -584,8 +584,13 @@ impl ValidatorConfigV2 {
         self.require_new_address(v1_val.validatorAddress)?;
         self.require_new_pubkey(v1_val.publicKey)?;
 
-        PublicKey::decode(v1_val.publicKey.as_slice())
-            .map_err(|_| ValidatorConfigV2Error::invalid_public_key())?;
+        let deactivated_at_height = if v1_val.active {
+            PublicKey::decode(v1_val.publicKey.as_slice())
+                .map_err(|_| ValidatorConfigV2Error::invalid_public_key())?;
+            0
+        } else {
+            block_height
+        };
 
         let egress = v1_val
             .outboundAddress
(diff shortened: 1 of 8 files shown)


crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -353,6 +353,7 @@ impl ValidatorConfigV2 {
         hasher.update(self.storage.chain_id().to_be_bytes());
         hasher.update(VALIDATOR_CONFIG_V2_ADDRESS.as_slice());
         hasher.update(validator_address.as_slice());
+        hasher.update([ingress.len() as u8]);
         hasher.update(ingress.as_bytes());
         hasher.update(egress.as_bytes());
```
SuperFluffy, line 364: Technically but necessary, but let's also add a len prefix to egress

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -339,7 +339,7 @@ impl ValidatorConfigV2 {
     ///
     /// **FORMAT**:
     /// - Namespace: [`VALIDATOR_NS_ADD`] or [`VALIDATOR_NS_ROTATE`]
-    /// - Message: `keccak256(abi.encodePacked(chainId, contractAddr, validatorAddr, ingress, egress))`
+    /// - Message: `keccak256(abi.encodePacked(chainId, contractAddr, validatorAddr, uint8(ingress.len), ingress, egress))`
```
SuperFluffy: Should we remove all mention of `abi.encodePacked`? It is not clear to me that what we write into the hasher actually is `abi.encodePacked`. Assuming it is, we should unit test it. If it is not, we should remove (and update the spec accordingly).

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -1563,6 +1573,44 @@ mod tests {
...
+    fn test_migration_rejects_invalid_ed25519_pubkey() -> eyre::Result<()> {
+        let mut storage = HashMapStorageProvider::new(1);
+        let owner = Address::random();
+
+        StorageCtx::enter(&mut storage, || {
+            let mut v1 = v1();
+            v1.initialize(owner)?;
+
+            // Find a 32-byte value that is not a valid Ed25519 curve point.
+            let bad_key = (0u8..)
```
SuperFluffy: Let's just write the invalid key out (or store it in a function).

xtask/src/genesis_args.rs
```diff
@@ -1004,11 +1004,12 @@ fn initialize_validator_config_v2(
                 let ingress = addr.to_string();
                 let egress = addr.ip().to_string();
 
-                // message: keccak256(chainId || contractAddr || validatorAddr || ingress || egress)
+                // message: keccak256(chainId || contractAddr || validatorAddr || uint8(ingress.len) || ingress || egress)
```
SuperFluffy: Yeah, here there is no mention of `abi.encodePacked`

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -579,6 +580,9 @@ impl ValidatorConfigV2 {
         self.require_new_address(v1_val.validatorAddress)?;
         self.require_new_pubkey(v1_val.publicKey)?;
 
+        PublicKey::decode(v1_val.publicKey.as_slice())
```
SuperFluffy: Yeah, this makes sense. We checked ingress/agress, but don't check pubkeys.

Note: we should also check pubkey uniqueness - or is that done in `append_validator`?

crates/precompiles/src/validator_config_v2/mod.rs
howydev:
```suggestion
    /// - Message: `keccak256(chainId || contractAddr || validatorAddr || uint8(ingress.len) || ingress || egress))`
```

tips/tip-1017.md
```diff
@@ -183,7 +183,7 @@ interface IValidatorConfigV2 {
 
     /// @notice Add a new validator (owner only)
     /// @dev The signature must be an Ed25519 signature over:
-    ///      keccak256(abi.encodePacked(bytes8(chainId), contractAddress, validatorAddress, ingress, egress))
+    ///      keccak256(abi.encodePacked(bytes8(chainId), contractAddress, validatorAddress, uint8(bytes(ingress).length), ingress, egress))
```
howydev:
```suggestion
    ///      keccak256(bytes8(chainId) || contractAddress || validatorAddress || uint8(bytes(ingress).length || ingress || egress))
```

tips/tip-1017.md
```diff
@@ -219,7 +219,7 @@ interface IValidatorConfigV2 {
     ///      - Egress must be parseable as <ip>.
     ///      - The signature must prove ownership of the new public key
     ///      The signature must be an Ed25519 signature over:
-    ///      keccak256(abi.encodePacked(bytes8(chainId), contractAddress, validatorAddress, ingress, egress))
+    ///      keccak256(abi.encodePacked(bytes8(chainId), contractAddress, validatorAddress, uint8(bytes(ingress).length), ingress, egress))
```
howydev:
```suggestion
    ///      keccak256(bytes8(chainId) || contractAddress || validatorAddress || uint8(bytes(ingress).length) || ingress || egress))
```

crates/precompiles/src/validator_config_v2/mod.rs
howydev: yep, thats already done

SuperFluffy (approved):

joshieDo (approved):

PR #2862, opened 2026-02-25
feat: add validation checks for keychain txs

0xKitsune:
This PR adds additional validation checks for keychain txs and updates tx classification for keychain errors.


crates/node/tests/it/tempo_transaction.rs
fgimenez:
```suggestion
    // Each block advances timestamp by ~1 second, so 5 seconds should be enough for:
```

PR #2863, opened 2026-02-25
fix: feed actor must account for epoch when accounting for ordering

hamdiallam:


crates/commonware-node/src/feed/actor.rs
```diff
@@ -129,14 +130,17 @@ impl<TContext: Spawner> Actor<TContext> {
...
                     if state
                         .latest_finalized
                         .as_ref()
                         .is_none_or(|f| f.height < block.height)
+                        // Notarization must be strictly newer (view is reset to 1 on epoch change)
                         && state
                             .latest_notarized
                             .as_ref()
-                            .is_none_or(|n| n.view < view)
+                            .is_none_or(|n| (n.epoch, n.view) <= (epoch, view))
```
SuperFluffy: This can be simplified by using the `notarization.proposal.round` information, which has a `PartialOrd` impl that is the same as this line, see https://github.com/commonwarexyz/monorepo/blob/92647ea3648d631426ed8e9b03735be4ec0eb22e/consensus/src/types.rs#L412

crates/commonware-node/src/feed/actor.rs
```diff
@@ -162,19 +167,23 @@ impl<TContext: Spawner> Actor<TContext> {
...
                         .is_none_or(|f| f.height < block.height)
                     {
+                        state.latest_finalized = Some(block);
+
+                        // Clear any older notarization
                         if state
                             .latest_notarized
                             .as_ref()
-                            .is_some_and(|n| n.view <= view)
+                            .is_some_and(|n| (n.epoch, n.view) <= (epoch, view))
```
SuperFluffy: Similar to above, this is the same as the derived `PartialOrd` impl on: https://github.com/commonwarexyz/monorepo/blob/92647ea3648d631426ed8e9b03735be4ec0eb22e/consensus/src/types.rs#L412

SuperFluffy (approved):
Approved because the general direction is correct, but please simplify this (see comment).

PR #2864, opened 2026-02-25
fix(precompiles): validate V1/V2 activity state in initialize_if_migrated (ZELLIC-64)

howydev:
## Summary

**ZELLIC-64**: `initialize_if_migrated` only checked that the validator count matched between V1 and V2, but did not verify each validator's activity state was consistent. A validator could be deactivated in V2 during migration while still active in V1 (or vice versa), leading to inconsistent state after initialization.

## Changes

- Add per-validator activity state validation in `initialize_if_migrated` / `initializeIfMigrated`:
  - V1 `active == true` → V2 `deactivatedAtHeight` must be `0`
  - V1 `active == false` → V2 `deactivatedAtHeight` must be `> 0`
- Add new `MigrationStateMismatch(uint64 idx)` error to identify which validator has mismatched state
- Updated in both Solidity ref-impl and Rust precompile

## Files changed

- `tips/ref-impls/src/interfaces/IValidatorConfigV2.sol` — new error definition
- `tips/ref-impls/src/ValidatorConfigV2.sol` — activity state validation loop
- `crates/contracts/src/precompiles/validator_config_v2.rs` — new error in sol! macro + constructor
- `crates/precompiles/src/validator_config_v2/mod.rs` — activity state validation loop

### DIFF
--- a/crates/precompiles/src/validator_config_v2/mod.rs
+++ b/crates/precompiles/src/validator_config_v2/mod.rs
@@ -1587,8 +1587,8 @@ mod tests {
             let mut v1 = v1();
             v1.initialize(owner)?;
 
-            // Zero is not a valid Ed25519 curve point.
-            let bad_key = [0u8; 32];
+            // [0x02; 32] is not a valid Ed25519 curve point but is non-zero so V1 accepts it.
+            let bad_key = [0x02u8; 32];
             v1.add_validator(
                 owner,
                 tempo_contracts::precompiles::IValidatorConfig::addValidatorCall {
--- a/crates/precompiles/src/validator_config_v2/mod.rs
+++ b/crates/precompiles/src/validator_config_v2/mod.rs
@@ -344,7 +344,7 @@ impl ValidatorConfigV2 {
     ///
     /// **FORMAT**:
     /// - Namespace: [`VALIDATOR_NS_ADD`] or [`VALIDATOR_NS_ROTATE`]
-    /// - Message: `keccak256(chainId || contractAddr || validatorAddr || uint8(ingress.len) || ingress || egress))`
+    /// - Message: `keccak256(chainId || contractAddr || validatorAddr || uint8(ingress.len) || ingress || uint8(egress.len) || egress))`
     fn verify_validator_signature(
         &self,
         namespace: &[u8],
@@ -360,6 +360,7 @@ impl ValidatorConfigV2 {
         hasher.update(validator_address.as_slice());
         hasher.update([ingress.len() as u8]);
         hasher.update(ingress.as_bytes());
+        hasher.update([egress.len() as u8]);
         hasher.update(egress.as_bytes());
         let message = hasher.finalize();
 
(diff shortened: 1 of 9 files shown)


0xrusowsky (approved):

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -611,10 +611,23 @@ impl ValidatorConfigV2 {
...
         }
 
+        for i in 0..v1_count {
+            let v1_val = v1.validators(v1.validators_array(i)?)?;
+            let v2_val = self.validators[i as usize].read()?;
+            if v1_val.active && v2_val.deactivated_at_height != 0 {
+                Err(ValidatorConfigV2Error::migration_state_mismatch(i))?
+            }
+            if !v1_val.active && v2_val.deactivated_at_height == 0 {
+                Err(ValidatorConfigV2Error::migration_state_mismatch(i))?
```
SuperFluffy, line 638: Let's simplify this:

```rust
if v1_val.active != (v2_val.deactivated == 0) {
    return Err(ValidatorConfigV2Error::migration_state_mismatch(i));
}
```

tips/ref-impls/src/interfaces/IValidatorConfigV2.sol
```diff
@@ -51,6 +51,10 @@ interface IValidatorConfigV2 {
     /// @notice Thrown when migration is not complete (not all V1 validators migrated)
     error MigrationNotComplete();
 
+    /// @notice Thrown when a migrated validator's active/deactivated state does not match V1
+    /// @param idx The index of the validator with mismatched state
+    error MigrationStateMismatch(uint64 idx);
```
SuperFluffy, line 56: Should this be added to the spec?

tips/ref-impls/src/ValidatorConfigV2.sol
```diff
@@ -356,6 +356,17 @@ contract ValidatorConfigV2 is IValidatorConfigV2 {
             revert MigrationNotComplete();
         }
 
+        for (uint64 i = 0; i < v1Validators.length; i++) {
+            bool v1Active = v1Validators[i].active;
+            uint64 v2Deactivated = validatorsArray[i].deactivatedAtHeight;
+            if (v1Active && v2Deactivated != 0) {
```
SuperFluffy, line 378: Can probably express this in simpler terms similar to the rust code.

SuperFluffy (approved):
Looks fine, but can be simplified.

Does the spec need to be updated to cover the error?

Add a test to check this activity mismatsch?

Can you double check that this matches the spec?

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -611,10 +611,23 @@ impl ValidatorConfigV2 {
...
+        let v1_count = v1.validator_count()?;
+        if v2_count < v1_count {
             Err(ValidatorConfigV2Error::migration_not_complete())?
         }
 
+        for i in 0..v1_count {
+            let v1_val = v1.validators(v1.validators_array(i)?)?;
+            let v2_val = self.validators[i as usize].read()?;
+            if v1_val.active && v2_val.deactivated_at_height != 0 {
+                Err(ValidatorConfigV2Error::migration_state_mismatch(i))?
```
joshieDo, line 635: some smol unit test would be nice

howydev:
closing in favor because of new migration strategy: https://github.com/tempoxyz/tempo/pull/2918

PR #2867, opened 2026-02-26
chore: add keychain signature type V2

legion2002:
AUD 3

### DIFF
--- a/crates/transaction-pool/src/test_utils.rs
+++ b/crates/transaction-pool/src/test_utils.rs
@@ -77,7 +77,7 @@ impl Default for TxBuilder {
             fee_token: None,
             valid_after: None,
             valid_before: None,
-            chain_id: 1337, // DEV chain_id
+            chain_id: 42431, // MODERATO chain_id
             calls: None,
             authorization_list: None,
             access_list: Default::default(),
@@ -348,7 +348,7 @@ pub(crate) fn wrap_valid_tx(
     }
 }
 
-/// Creates a mock provider configured with the DEV chain spec (all hardforks active).
+/// Creates a mock provider with the DEV chain spec (all hardforks active at genesis).
 pub(crate) fn create_mock_provider()
 -> MockEthProvider<reth_ethereum_primitives::EthPrimitives, TempoChainSpec> {
     MockEthProvider::default().with_chain_spec(std::sync::Arc::unwrap_or_clone(DEV.clone()))


crates/revm/src/handler.rs
```diff
@@ -1266,6 +1266,19 @@ where
...
+            }
+
+            // Pre-T1C: Reject V2 keychain signatures to prevent chain splits
+            if !cfg.spec().is_t1c()
+                && aa_env.signature.is_keychain()
+                && !aa_env.signature.is_legacy_keychain()
+            {
+                return Err(TempoInvalidTransaction::V2KeychainBeforeActivation.into());
+            }
+
```
klkvr: are we ok with impact of this on existing tooling? if you've signed/sent a transaction before T1C it will not land after transition

crates/revm/src/error.rs
```diff
@@ -199,6 +199,21 @@ pub enum TempoInvalidTransaction {
...
+    #[error("legacy V1 keychain signature is no longer accepted, use V2 (type 0x04)")]
+    LegacyKeychainSignature,
+
+    /// V2 keychain signature used before T1C activation.
+    ///
+    /// V2 signatures (type 0x04) are only valid after the T1C hardfork activates.
+    /// Rejecting them before activation prevents chain splits between upgraded and
+    /// non-upgraded nodes.
+    #[error("V2 keychain signature (type 0x04) is not valid before T1C activation")]
+    V2KeychainBeforeActivation,
```
0xKitsune, line 218: Noting that this can be removed after hardfork activation.

crates/transaction-pool/src/test_utils.rs
```diff
@@ -257,12 +257,12 @@ impl TxBuilder {
...
         let sig_hash = unsigned.signature_hash();
 
-        // Sign with the access key
+        // V1: sign raw sig_hash (test chain uses MODERATO which is pre-T1C)
         let signature = access_key_signer
             .sign_hash_sync(&sig_hash)
             .expect("signing failed");
 
-        let keychain_sig = TempoSignature::Keychain(KeychainSignature::new(
+        let keychain_sig = TempoSignature::Keychain(KeychainSignature::new_v1(
```
0xKitsune: Should we instead pass in the keychain signature version to this function rather than always assuming we should use v1? This seems useful if these functions are used post t1c.

crates/revm/src/handler.rs
legion2002: Yeah I think it should be fine, because it is happening pre mainnet launch. so very low impact

crates/revm/src/handler.rs
```diff
@@ -1266,6 +1266,19 @@ where
...
 
+            // T1C+: Reject legacy V1 keychain signatures (deprecated)
+            if cfg.spec().is_t1c() && aa_env.signature.is_legacy_keychain() {
+                return Err(TempoInvalidTransaction::LegacyKeychainSignature.into());
+            }
+
+            // Pre-T1C: Reject V2 keychain signatures to prevent chain splits
+            if !cfg.spec().is_t1c()
+                && aa_env.signature.is_keychain()
+                && !aa_env.signature.is_legacy_keychain()
```
0xKitsune: nit: do we need both of these or should we only check `is_legacy_keychain()`?

crates/node/tests/it/tempo_transaction.rs
```diff
@@ -550,15 +572,18 @@ fn sign_aa_tx_with_p256_access_key(
     access_pub_key_y: &B256,
     root_key_addr: Address,
 ) -> eyre::Result<TempoSignature> {
+    let sig_hash = tx.signature_hash();
+    let effective_hash = KeychainSignature::signing_hash(sig_hash, root_key_addr);
```
0xKitsune: Can we add an e2e test showing that the cross account replay is blocked with the v2 signature? We should do this with both inner signature types.

crates/transaction-pool/src/validator.rs
```diff
@@ -2236,13 +2252,12 @@ mod tests {
...
 
-            // Sign with the access key
+            // V1: sign raw sig_hash (test chain uses MODERATO which is pre-T1C)
             let signature = access_key_signer
                 .sign_hash_sync(&sig_hash)
                 .expect("signing failed");
 
-            // Create keychain signature
-            let keychain_sig = TempoSignature::Keychain(KeychainSignature::new(
+            let keychain_sig = TempoSignature::Keychain(KeychainSignature::new_v1(
```
0xKitsune: Can we add tests asserting the expected behavior for when keychain signatures are rejected at the pool level pre/post t1c?

crates/primitives/src/transaction/tt_signature.rs
```diff
@@ -386,15 +407,47 @@ pub struct KeychainSignature {
...
+    /// - V2: returns `keccak256(sig_hash || user_address)`
+    fn effective_sig_hash(&self, sig_hash: &B256) -> B256 {
+        match self.version {
+            KeychainVersion::V1 => *sig_hash,
+            KeychainVersion::V2 => {
+                let mut buf = [0u8; 52]; // 32 + 20
+                buf[..32].copy_from_slice(sig_hash.as_slice());
+                buf[32..].copy_from_slice(self.user_address.as_slice());
+                keccak256(buf)
+            }
```
0xrusowsky: this should call `fn signing_hash` so that we have a single source of truth

crates/primitives/src/transaction/tt_signature.rs
```diff
@@ -433,39 +505,14 @@ impl core::hash::Hash for KeychainSignature {
     fn hash<H: core::hash::Hasher>(&self, state: &mut H) {
         self.user_address.hash(state);
         self.signature.hash(state);
+        self.version.hash(state);
     }
 }
 
-// Manual Compact implementation that excludes cached_key_id (cache field)
-#[cfg(feature = "reth-codec")]
-impl reth_codecs::Compact for KeychainSignature {
```
mattsse, line 441: was this unused then?

crates/primitives/src/transaction/tt_signature.rs
```diff
@@ -433,39 +505,14 @@ impl core::hash::Hash for KeychainSignature {
...
-            Self {
-                user_address,
-                signature,
-                cached_key_id: OnceLock::new(),
-            },
-            rest,
-        )
-    }
-}
+// Note: No Compact impl for KeychainSignature. It is never stored directly in the DB —
```
mattsse: this note will be a bit confusing here without the pr's contex

could we move this to struct level docs

crates/node/tests/it/tempo_transaction.rs
```diff
@@ -619,13 +646,11 @@ fn sign_aa_tx_with_webauthn_access_key(
...
-        tx.signature_hash(),
-        signing_key,
-        pub_key_x,
-        pub_key_y,
-        origin,
-    )?;
+    // V2: sign keccak256(sig_hash || user_address) instead of raw sig_hash
+    let sig_hash = tx.signature_hash();
+    let effective_hash =
+        alloy_primitives::keccak256([sig_hash.as_slice(), root_key_addr.as_slice()].concat());
```
0xrusowsky: let's use `KeychainSignature::signing_hash()`

crates/revm/src/handler.rs
```diff
@@ -1278,6 +1278,28 @@ where
...
+            // Apply the same version gating to keychain sigs in the authorization list
+            for auth in &aa_env.tempo_authorization_list {
+                if cfg.spec().is_t1c() && auth.signature().is_legacy_keychain() {
+                    return Err(TempoInvalidTransaction::LegacyKeychainSignature.into());
+                }
+                // TODO(tanishk): This check can be removed after T1C activation.
+                if !cfg.spec().is_t1c() && auth.signature().is_v2_keychain() {
+                    return Err(TempoInvalidTransaction::V2KeychainBeforeActivation.into());
+                }
+            }
```
0xrusowsky: this check is repeated identically on the mempool validation.
i think it's better to have a single helper `fn validate_keychain_version()` that acts as the source of truth

crates/transaction-pool/src/validator.rs
```diff
@@ -130,6 +130,28 @@ where
...
+
+        // Apply the same version gating to keychain sigs in the authorization list
+        for auth_sig in &tx.tx().tempo_authorization_list {
+            if spec.is_t1c() && auth_sig.signature().is_legacy_keychain() {
+                return Ok(Err(TempoPoolTransactionError::LegacyKeychainPostT1C));
+            }
+            if !spec.is_t1c() && auth_sig.signature().is_v2_keychain() {
+                return Ok(Err(TempoPoolTransactionError::V2KeychainPreT1C));
+            }
+        }
```
0xrusowsky, line 143: should use the unified helper i mentioned above

crates/transaction-pool/src/validator.rs
```diff
@@ -2724,8 +2802,12 @@ mod tests {
                 )),
             );
             let sig_hash = unsigned.signature_hash();
+
+            // V2: sign keccak256(sig_hash || user_address) instead of raw sig_hash
+            let effective_hash =
+                alloy_primitives::keccak256([sig_hash.as_slice(), real_user.as_slice()].concat());
```
0xrusowsky: should use `KeychainSignature::signing_hash()`

crates/transaction-pool/src/validator.rs
```diff
@@ -2236,16 +2301,29 @@ mod tests {
...
+                        .expect("signing failed");
+                    TempoSignature::Keychain(KeychainSignature::new_v1(
+                        user_address,
+                        PrimitiveSignature::Secp256k1(signature),
+                    ))
+                }
+                KeychainVersion::V2 => {
+                    let effective_hash = alloy_primitives::keccak256(
+                        [sig_hash.as_slice(), user_address.as_slice()].concat(),
+                    );
```
0xrusowsky: should use `KeychainSignature::signing_hash()`

crates/transaction-pool/src/validator.rs
0xrusowsky: weird comment, should be "This should never happen..."

crates/transaction-pool/src/test_utils.rs
```diff
@@ -257,15 +268,38 @@ impl TxBuilder {
...
+                        user_address,
+                        PrimitiveSignature::Secp256k1(signature),
+                    )),
+                )
+            }
+            KeychainVersion::V2 => {
+                // V2: sign keccak256(sig_hash || user_address)
+                let hash = alloy_primitives::keccak256(
+                    [sig_hash.as_slice(), user_address.as_slice()].concat(),
+                );
```
0xrusowsky: should use `KeychainSignature::signing_hash()`

crates/primitives/src/transaction/tt_signature.rs
```diff
@@ -624,6 +682,16 @@ impl TempoSignature {
...
     }
 
+    /// Check if this is a legacy V1 Keychain signature (deprecated at T1C).
+    pub fn is_legacy_keychain(&self) -> bool {
+        matches!(self, Self::Keychain(k) if k.is_legacy())
+    }
+
+    /// Check if this is a V2 Keychain signature.
+    pub fn is_v2_keychain(&self) -> bool {
+        matches!(self, Self::Keychain(k) if !k.is_legacy())
```
fgimenez: nit, if we ever add a `V3` keychain variant this would return true, is this a concern? if so we should match `KeychainVersion::V2` here explicitly

crates/transaction-pool/src/test_utils.rs
0xrusowsky: addressed in PR https://github.com/tempoxyz/tempo/pull/2898

crates/transaction-pool/src/validator.rs
0xrusowsky: addressed in PR https://github.com/tempoxyz/tempo/pull/2898

crates/transaction-pool/src/validator.rs
0xrusowsky: addressed in PR https://github.com/tempoxyz/tempo/pull/2898

crates/transaction-pool/src/validator.rs
0xrusowsky: addressed in PR https://github.com/tempoxyz/tempo/pull/2898

crates/transaction-pool/src/validator.rs
0xrusowsky: addressed in PR https://github.com/tempoxyz/tempo/pull/2898

crates/revm/src/handler.rs
0xrusowsky: addressed in PR https://github.com/tempoxyz/tempo/pull/2898

crates/node/tests/it/tempo_transaction.rs
0xrusowsky: addressed in PR https://github.com/tempoxyz/tempo/pull/2898

crates/primitives/src/transaction/tt_signature.rs
0xrusowsky: addressed in PR https://github.com/tempoxyz/tempo/pull/2898

crates/primitives/src/transaction/tt_signature.rs
0xrusowsky: addressed in PR https://github.com/tempoxyz/tempo/pull/2898

crates/primitives/src/transaction/tt_signature.rs
0xrusowsky: addressed in PR https://github.com/tempoxyz/tempo/pull/2898

crates/primitives/src/transaction/tt_signature.rs
legion2002: Yes it was unused before.

crates/primitives/src/transaction/tt_signature.rs
```diff
@@ -412,18 +476,37 @@ impl KeychainSignature {
...
+    /// Compute the hash that an access key should sign for a V2 keychain transaction.
+    ///
+    /// Returns `keccak256(sig_hash || user_address)` which binds the signature to the
+    /// specific user account.
+    pub fn signing_hash(sig_hash: B256, user_address: Address) -> B256 {
+        let mut buf = [0u8; 52]; // 32 + 20
+        buf[..32].copy_from_slice(sig_hash.as_slice());
+        buf[32..].copy_from_slice(user_address.as_slice());
+        keccak256(buf)
+    }
```
0xKitsune, line 504: For new signature schemes, including a domain separator is usually seen (EIP-712, EIP-7702's 0x05 magic byte, Tempo Tx signing with 0x76). Is there no domain separator on Keychain txs by design?

crates/transaction-pool/src/validator.rs
```diff
@@ -131,6 +131,17 @@ where
...
         let spec = self.inner.chain_spec().tempo_hardfork_at(current_time);
+
+        // Validate keychain signature version (outer + authorization list).
+        if let Err(e) = tx.signature().validate_version(spec.is_t1c()) {
+            return Ok(Err(e.into()));
+        }
+        for auth_sig in &tx.tx().tempo_authorization_list {
+            if let Err(e) = auth_sig.signature().validate_version(spec.is_t1c()) {
+                return Ok(Err(e.into()));
+            }
```
0xKitsune, line 142: Noting that Keychain V1 txs in the mempool before T1C activates will not be evicted. Eventually these will be flushed from the pool or fail within the `handler.rs` logic, just flagging for visibility.

crates/primitives/src/transaction/tt_signature.rs
```diff
@@ -353,23 +354,59 @@ impl reth_codecs::Compact for PrimitiveSignature {
...
-/// Format: 0x03 || user_address (20 bytes) || inner_signature
+/// Determines how the signature hash is computed for the inner signature.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
+#[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
+pub enum KeychainVersion {
+    /// Legacy (V1): inner signature signs `sig_hash` directly.
+    /// Deprecated at T1C.
+    #[default]
```
0xKitsune, line 368: Should V2 be the default post t1c?

crates/revm/src/evm.rs
```diff
@@ -290,16 +290,14 @@ mod tests {
...
+    /// This applies TIP-1000 gas params via `tempo_gas_params()`.
     fn create_funded_evm_t1(address: Address) -> TempoEvm<CacheDB<EmptyDB>, ()> {
         let db = CacheDB::new(EmptyDB::new());
         let mut cfg = CfgEnv::<TempoHardfork>::default();
-        cfg.spec = TempoHardfork::T1;
-        // Apply TIP-1000 gas params for T1 hardfork
-        cfg.gas_params = tempo_gas_params(TempoHardfork::T1);
+        cfg.spec = TempoHardfork::T1C;
+        // Apply TIP-1000 gas params for T1C hardfork
+        cfg.gas_params = tempo_gas_params(TempoHardfork::T1C);
```
0xKitsune, line 300: Is there a reason we switched this to T1C?

crates/transaction-pool/src/test_utils.rs
```diff
@@ -317,8 +349,12 @@ pub(crate) fn wrap_valid_tx(
...
 
-/// Creates a mock provider configured with the MODERATO chain spec.
+/// Creates a mock provider configured with a MODERATO-based chain spec with T1C enabled.
 pub(crate) fn create_mock_provider()
 -> MockEthProvider<reth_ethereum_primitives::EthPrimitives, TempoChainSpec> {
-    MockEthProvider::default().with_chain_spec(std::sync::Arc::unwrap_or_clone(MODERATO.clone()))
+    let mut spec = std::sync::Arc::unwrap_or_clone(MODERATO.clone());
+    spec.inner
+        .hardforks
+        .extend([(TempoHardfork::T1C, ForkCondition::Timestamp(0))]);
```
0xKitsune: Should we use `DEV` here or similar? If we hardcode to T1C we will likely overlook this in the future when new hardforks are added.

crates/primitives/src/transaction/tt_signature.rs
```diff
@@ -624,6 +688,36 @@ impl TempoSignature {
...
+                ..
+            })
+        )
+    }
+
+    /// Validates keychain signature version compatibility with the current hardfork.
+    ///
+    /// - Post-T1C: legacy V1 keychain signatures are rejected.
+    /// - Pre-T1C: V2 keychain signatures are rejected to prevent chain splits.
+    pub fn validate_version(&self, is_t1c: bool) -> Result<(), KeychainVersionError> {
```
0xKitsune, line 715: nit: consider passing in a reference to the chainspec instead of an `is_t1c` bool. Feel free to ignore as well if you prefer this approach.

crates/transaction-pool/src/validator.rs
```diff
@@ -3556,6 +3622,188 @@ mod tests {
...
+            Ok(())
+        }
+
+        #[test]
+        fn test_legacy_keychain_post_t1c_is_bad_transaction() {
+            assert!(
+                TempoPoolTransactionError::LegacyKeychainPostT1C.is_bad_transaction(),
+                "Post-T1C V1 rejection should be a bad transaction (permanent)"
+            );
+        }
```
0xKitsune, line 3838: Note that if there are V1 keychain txs in the mempool at the T1C hardfork boundary, and these are peered to other nodes post t1c, this could cause peers to get banned (since there is not logic to evict these txs from the pool). Just flagging for visibility. Lmk if you have concerns.

crates/revm/src/handler.rs
```diff
@@ -1276,6 +1276,21 @@ where
             )
             .map_err(TempoInvalidTransaction::from)?;
 
+            // Validate keychain signature version (outer + authorization list).
+            // Skip during gas estimation — the mock signature version doesn't affect gas.
+            // TODO(tanishk): Pre-T1C V2 rejection can be removed after T1C activation.
+            if !cfg.is_balance_check_disabled() {
+                aa_env
```
klkvr, line 1286: let's document why do we need to skip this during gas estimation at all

klkvr (approved):

crates/primitives/src/transaction/tt_signature.rs
legion2002: yes will add a todo comment

crates/revm/src/evm.rs
legion2002: yeah we need this for the test_tempo_tx below, because they use V2 sigs

PR #2869, opened 2026-02-26
fix(tip-1017): remove field param from NotIpPort and NotIp errors

howydev:
## Summary

Aligns TIP-1017 spec and Solidity ref-impl with the Rust V2 implementation.

**Before (spec/Solidity):**
```solidity
error NotIpPort(string field, string input, string backtrace);
error NotIp(string field, string input, string backtrace);
```

**After (matches Rust):**
```solidity
error NotIpPort(string input, string backtrace);
error NotIp(string input, string backtrace);
```

### Changes
- `tips/tip-1017.md` — Removed `field` param from `NotIpPort` spec
- `IValidatorConfigV2.sol` — Removed `field` from `NotIpPort` and `NotIp` declarations
- `ValidatorConfigV2.sol` — Removed `field` from all 7 internal validation function signatures, call sites, and ~35 revert statements

V1 files untouched (V1 Rust and Solidity both have `field` and are consistent).

Resolves ZELLIC-65

### DIFF
--- a/tips/tip-1017.md
+++ b/tips/tip-1017.md
@@ -106,7 +106,7 @@ interface IValidatorConfigV2 {
     /// @notice Thrown when trying to delete a validator that is already deleted
     error ValidatorAlreadyDeleted();
 
-    /// @notice Thrown when public key is invalid (zero)
+    /// @notice Thrown when public key is invalid (zero or not a valid Ed25519 curve point)
     error InvalidPublicKey();
 
     /// @notice Thrown when the Ed25519 signature verification fails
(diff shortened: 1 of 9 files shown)


tips/tip-1017.md
```diff
@@ -113,10 +113,9 @@ interface IValidatorConfigV2 {
     error InvalidSignature();
 
     /// @notice Thrown when address is not in valid ip:port format
-    /// @param field The field name that failed validation
     /// @param input The invalid input that was provided
     /// @param backtrace Additional error context
-    error NotIpPort(string field, string input, string backtrace);
+    error NotIpPort(string input, string backtrace);
```
SuperFluffy, line 119: Could call this `error IngressNotIpPort` - or just `error InvalidIngress`?

SuperFluffy (approved):
Change makes sense to match impl and spec.

Should we rename the error to make it ingress-specific (similarly, make `NotIp` egress-specific)?

PR #2870, opened 2026-02-26
fix: fee payer transfer checks

0xKitsune:
This PR updates fee payer auth checks in `t1c`.

### DIFF
--- a/crates/node/tests/it/tip_fee_manager.rs
+++ b/crates/node/tests/it/tip_fee_manager.rs
@@ -573,12 +573,7 @@ async fn test_fee_payer_transfer_whitelist_pre_t1c() -> eyre::Result<()> {
     let tx = TransactionRequest::default()
         .to(Address::ZERO)
         .value(U256::ZERO);
-    let receipt = fee_payer_provider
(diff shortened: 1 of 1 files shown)


crates/revm/src/common.rs
```diff
@@ -244,11 +244,13 @@ pub trait TempoStateAccess<M = ()> {
...
     {
         self.with_read_only_storage_ctx(spec, || {
-            // Ensure the fee payer is not blacklisted (sender authorization)
-            let policy_id = TIP20Token::from_address(fee_token)?
-                .transfer_policy_id
-                .read()?;
-            TIP403Registry::new().is_authorized_as(policy_id, fee_payer, AuthRole::sender())
+            let token = TIP20Token::from_address(fee_token)?;
+            if spec.is_t1c() {
+                token.is_transfer_authorized(fee_payer, TIP_FEE_MANAGER_ADDRESS)
```
legion2002, line 250: Noting that this works because is_transfer_authorized checks for both sender and receiver.

0xKitsune: Added a comment to clarify this.

PR #2871, opened 2026-02-26
chore: reject keychain authorization wildcard after T1C

legion2002:
AUD 5

### DIFF
--- a/crates/node/tests/it/tempo_transaction.rs
+++ b/crates/node/tests/it/tempo_transaction.rs
@@ -6129,7 +6129,7 @@ async fn test_aa_key_authorization_chain_id_validation() -> eyre::Result<()> {
 
     let sig_hash = tx_matching.signature_hash();
     let signature = root_signer.sign_hash_sync(&sig_hash)?;
-    let tx_hash = submit_and_mine_aa_tx(
+    submit_and_mine_aa_tx(
         &mut setup,
         tx_matching,
         TempoSignature::Primitive(PrimitiveSignature::Secp256k1(signature)),


crates/revm/src/handler.rs
```diff
@@ -900,10 +900,19 @@ where
...
-            if key_auth.chain_id != 0 && key_auth.chain_id != expected_chain_id {
+            if spec.is_t1c() {
+                if key_auth.chain_id != expected_chain_id {
+                    return Err(TempoInvalidTransaction::KeyAuthorizationChainIdMismatch {
+                        expected: expected_chain_id,
+                        got: key_auth.chain_id,
+                    }
+                    .into());
+                }
+            } else if key_auth.chain_id != 0 && key_auth.chain_id != expected_chain_id {
```
0xrusowsky: this check is repeated in both the handler and the mempool validator.

imo we should apply the same validation pattern to `KeyAuthorization`'s chain_id as we do in https://github.com/tempoxyz/tempo/pull/2898

crates/transaction-pool/src/validator.rs
```diff
@@ -2502,93 +2511,223 @@ mod tests {
             Ok(())
         }
 
-        #[test]
-        fn test_key_authorization_wrong_chain_id_rejected() {
-            let (access_key_signer, access_key_address) = generate_keypair();
-            let (user_signer, user_address) = generate_keypair();
+        /// Setup a validator using the DEV chain spec (T1C active at genesis).
+        fn setup_validator_with_keychain_storage_t1c(
```
fgimenez, line 2530: this new helper fn is very similar to the existing `setup_validator_with_keychain_storage`, only chainspec changes, could we refactor?

klkvr (approved):

crates/node/tests/it/tempo_transaction.rs
```diff
@@ -6079,13 +6079,67 @@ async fn test_aa_key_authorization_chain_id_validation() -> eyre::Result<()> {
...
+    );
+    let error_msg = inject_result.unwrap_err().to_string();
+    assert!(
+        error_msg.contains("chain_id does not match"),
+        "Error must mention chain_id mismatch. Got: {error_msg}"
+    );
+    println!("  ✓ Wildcard chain_id=0 rejected post-T1C as expected");
+
+    // Test 3: Matching chain_id should be accepted
+    println!("\nTest 3: KeyAuthorization with matching chain_id should be accepted");
```
0xKitsune: minor nit: these printlns are unnecessary

crates/node/tests/it/tempo_transaction.rs
```diff
@@ -6079,13 +6079,67 @@ async fn test_aa_key_authorization_chain_id_validation() -> eyre::Result<()> {
...
+    let signature = root_signer.sign_hash_sync(&sig_hash)?;
     let tx_hash = submit_and_mine_aa_tx(
         &mut setup,
-        tx_wildcard,
+        tx_matching,
         TempoSignature::Primitive(PrimitiveSignature::Secp256k1(signature)),
     )
     .await?;
-    println!("  ✓ Wildcard chain_id KeyAuthorization accepted (tx: {tx_hash})");
+    println!("  ✓ Matching chain_id KeyAuthorization accepted (tx: {tx_hash})");
```
0xKitsune: minor nit: same thing here, this println is not needed

crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -87,6 +88,31 @@ impl KeyAuthorization {
...
+                });
+            }
+        } else if self.chain_id != 0 && self.chain_id != expected_chain_id {
+            return Err(KeyAuthorizationChainIdError {
+                expected: expected_chain_id,
+                got: self.chain_id,
+            });
+        }
+        Ok(())
+    }
```
0xKitsune, line 114: Can we add unit tests covering this function/error cases both for pre/post t1c?

PR #2872, opened 2026-02-26
fix(consensus): rework validators reads from contract

SuperFluffy:
Reworks how peers are from the validator V1 and V2 contracts.

Peer manager actor:

+ bootstrap now takes the highest finalized hash available (from execution layer or consensus layer, whichever is highest).
+ `handle_finalized_block` was renamed to `update_peer_set` and made infallible. On failure, the peer actor will simply retry to update the peer set on the next block.k

DKG manager actor:

+ on fresh start, the best available finalized block is used as a reference point to determine the latest epoch and read its DKG outcome; before: the best boundary block was used, which was not guaranteed to be finalized leading to a potentially wrong epoch.
+ also on fresh restart, the parent of the boundary is taken as the reference point to determine whether to read syncers from the V1 contract; before: the boundary block itself was used, which is not consistent with how the DKG loop itself determines syncers.

Validator read from contract:

Made decoding validators from the contract more resilient. Malformed entries are now simply ignored and won't make it into the peer set. This is consistent with how it is currently handled on main. https://github.com/tempoxyz/tempo/pull/2716 overtightened the requirements.

### DIFF
--- a/crates/commonware-node/src/validators.rs
+++ b/crates/commonware-node/src/validators.rs
@@ -203,7 +203,7 @@ where
 /// Validators can be read from the V2 contract if the following conditions hold:
 ///
 /// 1. `timestamp(hash) >= T2`.
-/// 2. `intialization_height(<state>) <= number(hash)`.
+/// 2. `initialization_height(<state>) <= number(hash)`.
 /// 3. `is_init(<state>) == true`.
 ///
 /// `<state>` is read at either `hash` or `latest` if set.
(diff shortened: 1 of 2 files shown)


crates/commonware-node/src/peer_manager/actor.rs
```diff
@@ -257,158 +255,78 @@ where
...
+            read_peers_from_highest_finalized(
+                &self.execution_node,
+                &self.epoch_strategy,
+                self.last_tracked_peer_set
+                    .as_ref()
+                    .expect("must be set after bootstrap")
+                    .from_height,
+            )
+            .ok()
+            .flatten()
```
joshieDo: we're dropping wtv err might be thrown i think

SuperFluffy: Emits an error now. [b15180f](https://github.com/tempoxyz/tempo/pull/2872/commits/b15180f9550315f39cf3cd93cd3c9286fab18019)

PR #2873, opened 2026-02-26
docs: update TIP-1017 spec doc and simplify

howydev:
1. Update spec with the latest design changes 
2. Simplified wording and reduced density across the spec

[Rendered](https://github.com/tempoxyz/tempo/blob/howy/simplify-tip1017-spec/tips/tip-1017.md)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11796,7 +11796,7 @@ dependencies = [
 
 [[package]]
 name = "tempo"
-version = "1.4.0-rc1"
+version = "1.4.0"
 dependencies = [
  "alloy-consensus",
  "alloy-primitives",
@@ -11852,7 +11852,7 @@ dependencies = [
 
 [[package]]
 name = "tempo-alloy"
-version = "1.4.0-rc1"
+version = "1.4.0"
 dependencies = [
  "alloy",
  "alloy-consensus",
@@ -11886,7 +11886,7 @@ dependencies = [
 
 [[package]]
 name = "tempo-bench"
-version = "1.4.0-rc1"
+version = "1.4.0"
 dependencies = [
  "alloy",
  "alloy-consensus",
@@ -11913,7 +11913,7 @@ dependencies = [
 
 [[package]]
 name = "tempo-chainspec"
-version = "1.4.0-rc1"
+version = "1.4.0"
 dependencies = [
  "alloy-eips",
  "alloy-evm",
@@ -11931,7 +11931,7 @@ dependencies = [
 
 [[package]]
 name = "tempo-commonware-node"
-version = "1.4.0-rc1"
+version = "1.4.0"
 dependencies = [
  "alloy-consensus",
  "alloy-primitives",
@@ -11984,7 +11984,7 @@ dependencies = [
 
 [[package]]
 name = "tempo-commonware-node-config"
-version = "1.4.0-rc1"
+version = "1.4.0"
 dependencies = [
  "commonware-codec",
  "commonware-cryptography",
@@ -11996,7 +11996,7 @@ dependencies = [
 
 [[package]]
 name = "tempo-consensus"
-version = "1.4.0-rc1"
+version = "1.4.0"
 dependencies = [
  "alloy-consensus",
  "alloy-evm",
@@ -12014,7 +12014,7 @@ dependencies = [
 
 [[package]]
 name = "tempo-contracts"
-version = "1.4.0-rc1"
+version = "1.4.0"
 dependencies = [
  "alloy-contract",
  "alloy-primitives",
@@ -12025,7 +12025,7 @@ dependencies = [
 
 [[package]]
 name = "tempo-dkg-onchain-artifacts"
-version = "1.4.0-rc1"
+version = "1.4.0"
 dependencies = [
  "bytes",
  "commonware-codec",
@@ -12039,7 +12039,7 @@ dependencies = [
 
 [[package]]
 name = "tempo-e2e"
-version = "1.4.0-rc1"
+version = "1.4.0"
 dependencies = [
  "alloy",
  "alloy-evm",
@@ -12083,7 +12083,7 @@ dependencies = [
 
 [[package]]
 name = "tempo-evm"
-version = "1.4.0-rc1"
+version = "1.4.0"
 dependencies = [
  "alloy-consensus",
  "alloy-evm",
@@ -12114,7 +12114,7 @@ dependencies = [
 
 [[package]]
 name = "tempo-eyre"
-version = "1.4.0-rc1"
+version = "1.4.0"
 dependencies = [
  "eyre",
  "indenter",
@@ -12122,7 +12122,7 @@ dependencies = [
 
 [[package]]
 name = "tempo-faucet"
-version = "1.4.0-rc1"
+version = "1.4.0"
 dependencies = [
  "alloy",
  "async-trait",
@@ -12135,7 +12135,7 @@ dependencies = [
 
 [[package]]
 name = "tempo-node"
-version = "1.4.0-rc1"
+version = "1.4.0"
 dependencies = [
  "alloy",
  "alloy-eips",
@@ -12198,7 +12198,7 @@ dependencies = [
 
 [[package]]
 name = "tempo-payload-builder"
-version = "1.4.0-rc1"
+version = "1.4.0"
 dependencies = [
  "alloy-consensus",
  "alloy-primitives",
@@ -12230,7 +12230,7 @@ dependencies = [
 
 [[package]]
 name = "tempo-payload-types"
-version = "1.4.0-rc1"
+version = "1.4.0"
 dependencies = [
  "alloy-eips",
  "alloy-primitives",
@@ -12249,7 +12249,7 @@ dependencies = [
 
 [[package]]
 name = "tempo-precompiles"
-version = "1.4.0-rc1"
+version = "1.4.0"
 dependencies = [
  "alloy",
  "alloy-evm",
@@ -12277,7 +12277,7 @@ dependencies = [
 
 [[package]]
 name = "tempo-precompiles-macros"
-version = "1.4.0-rc1"
+version = "1.4.0"
 dependencies = [
  "alloy",
  "proc-macro2",
@@ -12287,7 +12287,7 @@ dependencies = [
 
 [[package]]
 name = "tempo-primitives"
-version = "1.4.0-rc1"
+version = "1.4.0"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
@@ -12321,7 +12321,7 @@ dependencies = [
 
 [[package]]
 name = "tempo-revm"
-version = "1.4.0-rc1"
+version = "1.4.0"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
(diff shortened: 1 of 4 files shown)


tips/tip-1017.md
```diff
@@ -1,73 +1,24 @@
...
-```
-
-Both addition and deactivation take effect at the next epoch boundary—there is no warmup or cooldown period.
-
-This eliminates the need to retain historical account state for consensus recovery—nodes can derive DKG player sets from the current contract state alone.
-
----
+V1 allowed validators to switch between `active` and `inactive`, which forced nodes to retain historical account state for each epoch. Limited input safety checks also prevented safe permissionless rotation. Tempo addressed this by disabling rotation and setting validator addresses to unowned addresses. This TIP adds stricter input validation to reduce operational risk.
 
 # Specification
```
jenpaff, line 68: Please add an assumptions section here. Everything where auditors lacked context, should be clarified here @SuperFluffy

tips/tip-1017.md
```diff
@@ -1,73 +1,24 @@
...
-TIP-1017 introduces ValidatorConfig V2, a new precompile for managing consensus
-validators with append-only semantics. Unlike the original ValidatorConfig, V2
-replaces the mutable `active` boolean with `addedAtHeight` (set when adding an
-entry) and `deactivatedAtHeight` fields (set when deactivating), enabling nodes to
-reconstruct the validator set for any historical epoch using only current state.
-The new design also requires Ed25519 signature verification when adding
-validators to prove key ownership.
+TIP-1017 defines ValidatorConfig V2, a new precompile for managing consensus participants. V2 improves lifecycle tracking so validator sets can be reconstructed for any epoch, and adds stricter input validation. It is designed to safely support permissionless validator rotation, and additionally allows separation of fee custody from day-to-day validator operations.
 
 ## Motivation
```
jenpaff, line 39: Could you add problems that were faced with V1? It usually helps to understand the solution if the problem statement is clear

tips/tip-1017.md
```diff
@@ -1,73 +1,24 @@
...
 ## Abstract
 
-TIP-1017 introduces ValidatorConfig V2, a new precompile for managing consensus
-validators with append-only semantics. Unlike the original ValidatorConfig, V2
-replaces the mutable `active` boolean with `addedAtHeight` (set when adding an
-entry) and `deactivatedAtHeight` fields (set when deactivating), enabling nodes to
-reconstruct the validator set for any historical epoch using only current state.
-The new design also requires Ed25519 signature verification when adding
-validators to prove key ownership.
+TIP-1017 defines ValidatorConfig V2, a new precompile for managing consensus participants. V2 improves lifecycle tracking so validator sets can be reconstructed for any epoch, and adds stricter input validation. It is designed to safely support permissionless validator rotation, and additionally allows separation of fee custody from day-to-day validator operations.
```
jenpaff, line 13: >It is designed to safely support permissionless validator rotation

I think this is slightly misleading, permissionless validator rotation if permissioned validator has been onboarded, right?

tips/tip-1017.md
```diff
@@ -1,73 +1,24 @@
...
-    (v.deactivatedAtHeight == 0 || v.deactivatedAtHeight > boundary(E))
-)
-```
-
-Both addition and deactivation take effect at the next epoch boundary—there is no warmup or cooldown period.
-
-This eliminates the need to retain historical account state for consensus recovery—nodes can derive DKG player sets from the current contract state alone.
-
----
+V1 allowed validators to switch between `active` and `inactive`, which forced nodes to retain historical account state for each epoch. Limited input safety checks also prevented safe permissionless rotation. Tempo addressed this by disabling rotation and setting validator addresses to unowned addresses. This TIP adds stricter input validation to reduce operational risk.
```
jenpaff: I would re-phrase this paragraph its not easy to follow 

>V1 allowed validators to switch between `active` and `inactive`, which forced nodes to retain historical account state for each epoch. 

What exactly does this mean to allow switching between active & inactive? (from an operator perspective. 

>and setting validator addresses to unowned addresses
This sounds pretty bad without context, would say "required to implement a workaround"

tips/tip-1017.md
SuperFluffy: I added a section for necessary background information and expanded the motivational section again.

tips/tip-1017.md
SuperFluffy: Extended the motivational section again that now contains the issues (existed in a previous version of the spec).

tips/tip-1017.md
SuperFluffy: I don't think this is confusing - rotating an entry implies a previous addition.

tips/tip-1017.md
SuperFluffy: > What exactly does this mean to allow switching between active & inactive? (from an operator perspective.


Removed that section because it was misleading.

> and setting validator addresses to unowned addresses
This sounds pretty bad without context, would say "required to implement a workaround"

Changed the wording to `anonymous` (also, this is just how we handled it; no need to go into this further IMO).

tips/tip-1017.md
```diff
@@ -138,671 +149,327 @@ interface IValidatorConfigV2 {
...
-                                              │◄────────────────────────/───────────────┘
-                                              │         (No transition back)
+    /// @notice Get initialization block height.
+    /// @return height Initialization height (`0` if not initialized).
+    function getInitializedAtHeight() external view returns (uint64);
+}
 ```
 
-### Signature Verification
+## Overview
```
SuperFluffy, line 273: This section is odd. I am not sure what it's supposed to give an overview of.

tips/tip-1017.md
```diff
@@ -138,671 +149,327 @@ interface IValidatorConfigV2 {
...
-This means migration of a single validator incurs at least 1 million gas cost, only
-leaving enough room to migrate less than 30 validator entries at a time.
+### Full Migration Steps
 
-This runs the risk of potentially not leaving enough space to migrate all validators
-in one go and would require logic to run several migrations. This would require
-manual intervention anyway, and require extra logic in the precompile to check
-which validators have already been migrated.
+1. At fork activation, the V2 precompile goes live. However, CL continues reading from the V1 precompile.
+2. The owner calls `migrateValidator(n-1)` with `n` being the validator count in the V1 precompile.
```
SuperFluffy, line 406: That's new - we explicitly set how many validators we need to migrate?

tips/tip-1017.md
```diff
@@ -138,671 +149,327 @@ interface IValidatorConfigV2 {
...
 
-### CL Read Behavior
+# Security
 
-The CL checks `v2.isInitialized()` to determine which contract to read:
+## Considerations
 
-- **`initialized == false`**: CL reads from V1
-- **`initialized == true`**: CL reads from V2
+- **Migration timing**: migration and `initializeIfMigrated()` should complete before an epoch boundary to avoid DKG disruption.
```
SuperFluffy, line 428: Why? That seems unnecessary?

tips/tip-1017.md
```diff
@@ -138,671 +149,327 @@ interface IValidatorConfigV2 {
...
+- **Pre-migration validation**: admins should run a validation script against V1 state to detect entries that would fail V2 checks.
```
SuperFluffy, line 429: Why is this necessary? Why isn't this covered by `initializeIfMigrated` and/or `migrateValidator`? Migration should follow the same pattern that CL is currently using, which rejects all entries that are malformed, and prefers newer entries over older as a tiebreaker in case of duplicates.

tips/tip-1017.md
howydev: its still n calls - the new part is that we iterate from n-1...0 instead of 0...n-1

tips/tip-1017.md
```diff
@@ -148,671 +149,327 @@ interface IValidatorConfigV2 {
...
+- `addedAtHeight`: block height where the entry becomes visible to CL epoch filtering.
+- `deactivatedAtHeight`: `0` means active; non-zero marks irreversible deactivation.
+- `index`: immutable array position assigned at creation.
+- `initialized`: one-way migration flag toggled by `initializeIfMigrated()`.
 
-### Determining Active Validators
+### Fee Recipient Separation
 
-Reading this contract alone is **not sufficient** to determine who the active validators (signers) are during a given epoch. The contract only records which validators are *eligible* to participate in DKG ceremonies—it does not record DKG outcomes.
+Each validator entry includes a `feeRecipient` that can differ from the validator's control address. This enables operators to route protocol fees to a dedicated treasury wallet, while retaining a separate validator or treasury-ops multisig for operational calls. This separation reduces blast radius during key compromise: operational key exposure does not cause historically collected fees held by the custody wallet to be lost.
```
legion2002, line 293: Clarifying, this is just the setup for a behaviour that will be enabled in future hardforks? 
Because to do this we'll need to make changes in the fee manager

tips/tip-1017.md
```diff
@@ -148,671 +149,327 @@ interface IValidatorConfigV2 {
...
-   - One transaction per validator
-   - Example: `migrateValidator(0)`, `migrateValidator(1)`, `migrateValidator(2)`, etc.
-   - Validator addresses are copied from V1
-5. **Admin calls `initializeIfMigrated()`**
-   - Sets `initialized = true`
-   - CL now reads from V2
-6. **Post-migration**: All reads/writes use V2 state directly
-
-**Important**: Complete migration before an epoch boundary to avoid disrupting DKG.
+1. **Unique active addresses**: No two active validators share the same `validatorAddress`. Deactivated addresses may be reused.
```
legion2002, line 446: slightly confusing that we say "deactivated addresses may be resused" here. 
But in the motivation for val config V2 we say that we want to stop validator re-registration. 

But point 2 clarifies this, because only the same public key cannot be re registered, the validator address can be kept the same.

tips/tip-1017.md
```diff
@@ -148,671 +149,327 @@ interface IValidatorConfigV2 {
...
-## Migration Timing
+1. **Full-set reconstruction by index**: Reading `validatorByIndex(i)` for all `i` in `0..validatorCount()-1` must reconstruct exactly the ordered validator set.
+2. **Validator activity consistency**: Filtering the reconstructed validator set by `deactivatedAtHeight == 0` must produce exactly `getActiveValidators()` (same members, order not important).
+3. **Address round-trip for active entries**: For any `i` where `validatorByIndex(i).deactivatedAtHeight == 0`, `validatorByAddress(validatorByIndex(i).validatorAddress).index == i`.
+4. **Public-key round-trip for all entries**: For any `i < validatorCount()`, `validatorByPublicKey(validatorByIndex(i).publicKey).index == i`.
+5. **Index round-trip for all entries**: For any `i < validatorCount()`, `validatorByIndex(i).index == i`.
 
-The migration must be completed (including `initializeIfMigrated()`) before an epoch
-boundary to avoid disrupting DKG. The admin should:
+### Migration and Initialization
```
legion2002, line 468: This migration design makes sense.

tips/tip-1017.md
howydev: @Zygimantass has more info on the changes but that is out of scope for t2, we just want to batch the contract changes in rn

tips/tip-1017.md
howydev: Address dups are fine, I believe ed25519 dups are what we need to avoid since that messes with CL

PR #2874, opened 2026-02-26
fix(pool): clean up expiry tracking on tx removal

0xKitsune:
Adds `untrack_expiry()` to remove expiry and key-expiry metadata when transactions are mined, moved to the paused pool, or evicted as stale. Adds a `tx_to_key` reverse map to `KeyExpiryTracker` for O(1) cleanup.

Tests updated to use `valid_before: None` where an explicit expiry isn't needed.

### DIFF
--- a/crates/transaction-pool/src/maintain.rs
+++ b/crates/transaction-pool/src/maintain.rs
@@ -403,9 +403,13 @@ impl KeyExpiryTracker {
 
     /// Removes a single transaction hash from key expiry tracking.
     fn untrack(&mut self, hash: &TxHash) {
-        let Some(key) = self.tx_to_key.remove(hash) else { return };
+        let Some(key) = self.tx_to_key.remove(hash) else {
+            return;
+        };
 
-        let Some((expiry, txs)) = self.key_to_txs.get_mut(&key) else { return };
+        let Some((expiry, txs)) = self.key_to_txs.get_mut(&key) else {
+            return;
+        };
         txs.remove(hash);
 
         if txs.is_empty() {


crates/transaction-pool/src/maintain.rs
```diff
@@ -548,6 +571,9 @@ where
                     .map(|tx| *tx.tx_hash())
                     .collect();
 
+                // Clean up AA tx expiry tracking from expiry_map and tx_to_expiry
+                state.remove_mined(&mined_tx_hashes);
```
0xrusowsky: `remove_mined()` cleans up `expiry_map`/`tx_to_expiry` for mined txs, but `KeyExpiryTracker` has the same issue (mined keychain tx hashes stay in `key_to_txs` until key expiry).

should we add an equivalent `key_expiry.remove_mined(&mined_tx_hashes)` here?

0xrusowsky: nit: when txs are moved to paused_pool on fee-token pause, their entries remain in `expiry_map`/`tx_to_expiry`/`key_to_txs`. 

with the time cap the impact is bounded, but for should we also clean up expiry tracking there for correctness?

0xrusowsky: addressed in https://github.com/tempoxyz/tempo/pull/2878 (needs reviews)

crates/node/tests/it/tempo_transaction.rs
```diff
@@ -722,7 +722,7 @@ fn create_basic_aa_tx(
         // Use AlphaUSD to match fund_address_with_fee_tokens
         fee_token: Some(DEFAULT_FEE_TOKEN),
         fee_payer_signature: None,
-        valid_before: Some(u64::MAX),
+        valid_before: None,
```
legion2002, line 725: nit: maybe we should let randomly assign some of these to remain Some(u64::max) to cover both cases?

PR #2875, opened 2026-02-26
fix: fee payer whitelist integration tests

0xrusowsky:
Fixes the integration tests from #2870.

### Problems
- Both tests called `changeTransferPolicyId` on PATH_USD, but the test mnemonic doesn't have `DEFAULT_ADMIN_ROLE` on PATH_USD (genesis admin is the coinbase `0xbba20...`). The call silently reverted, leaving PATH_USD with `policy_id=1` (always-allow) — making both tests false positives.
- No AMM liquidity was provisioned for the custom fee token.
- After a pool-rejected tx, alloy's internal nonce tracker desynchronized, causing the subsequent tx to hang forever.

### Fixes
- Use `admin_token` (via `setup_test_token`) where the test account actually has `DEFAULT_ADMIN_ROLE`
- Provision AMM liquidity (`fee_amm.mint`) for `admin_token ↔ PATH_USD`
- Have `fee_payer` set preference to `admin_token` via `setUserToken` before the restrictive policy is applied
- Assert `receipt.status()` on `changeTransferPolicyId` to catch silent reverts
- Re-fetch nonce from chain after the rejected tx in the post-T1C test
- Fix error assertion to match actual pool error (`blacklisted` not `TokenPolicyForbids`)

### DIFF
--- a/crates/node/tests/it/tip_fee_manager.rs
+++ b/crates/node/tests/it/tip_fee_manager.rs
@@ -473,7 +473,7 @@ async fn test_fee_payer_transfer_whitelist_pre_t1c() -> eyre::Result<()> {
         .get_receipt()
         .await?;
 
-    // Create whitelist policy: whitelist fee_payer AND FeeManager
+    // Create whitelist policy: whitelist fee_payer only (FeeManager not whitelisted pre-T1C)
     let registry = ITIP403Registry::new(TIP403_REGISTRY_ADDRESS, provider.clone());
     let policy_receipt = registry
         .createPolicy(admin_addr, ITIP403Registry::PolicyType::WHITELIST)
@@ -498,13 +498,6 @@ async fn test_fee_payer_transfer_whitelist_pre_t1c() -> eyre::Result<()> {
         .await?
         .get_receipt()
         .await?;
-    registry
-        .modifyPolicyWhitelist(policy_id, TIP_FEE_MANAGER_ADDRESS, true)
-        .gas(1_000_000)
-        .send()
-        .await?
-        .get_receipt()
-        .await?;
 
     let admin_token_contract = ITIP20::new(token_addr, &provider);
     let receipt = admin_token_contract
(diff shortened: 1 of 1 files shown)


crates/node/tests/it/tip_fee_manager.rs
```diff
@@ -445,26 +491,40 @@ async fn test_fee_payer_transfer_whitelist_pre_t1c() -> eyre::Result<()> {
...
+        .modifyPolicyWhitelist(policy_id, fee_payer_addr, true)
         .gas(1_000_000)
         .send()
         .await?
         .get_receipt()
         .await?;
-
     registry
-        .modifyPolicyWhitelist(policy_id, fee_payer_addr, true)
+        .modifyPolicyWhitelist(policy_id, TIP_FEE_MANAGER_ADDRESS, true)
```
0xKitsune: Pre t1c we want to assert that the tx goes through without whitelisting the fee manager.

legion2002: yup in it's current form it doesn't really test anything meaningful

legion2002 (approved):
lgtm, only the 1 comment needs to be resolved. The pre-t1c test isn't meaninful atm.

crates/node/tests/it/tip_fee_manager.rs
```diff
@@ -522,37 +602,37 @@ async fn test_fee_payer_transfer_whitelist_post_t1c() -> eyre::Result<()> {
...
+    // but FeeManager is not, so the tx should be rejected at the pool level.
     let tx = TransactionRequest::default()
         .to(Address::ZERO)
         .value(U256::ZERO);
     let result = fee_payer_provider.send_transaction(tx).await;
     let err = result.unwrap_err().to_string();
     assert!(
-        err.contains("TokenPolicyForbids"),
-        "expected TokenPolicyForbids, got: {err}"
+        err.contains("blacklisted"),
```
0xKitsune, line 631: Looks like this error may be misleading. In the test we are not whitelisting the fee manager, however the error returned contains "blacklisted". We should update this to reflect the actual error the tx encounters. We can also do this in a separate PR to not block on this.

PR #2878, opened 2026-02-26
fix: clean up `KeyExpiryTracker` + paused-pool expiry tracking

0xrusowsky:
- Add `KeyExpiryTracker::remove_mined()` to clean up `key_to_txs` when txs are mined, preventing unbounded memory growth
- Clean up both expiry trackers when txs are moved to `paused_pool` on fee-token pause events

### DIFF
--- a/crates/transaction-pool/src/maintain.rs
+++ b/crates/transaction-pool/src/maintain.rs
@@ -10,7 +10,7 @@ use crate::{
 use alloy_consensus::transaction::TxHashRef;
 use alloy_primitives::{
     Address, TxHash,
-    map::{self as alloy_map, AddressMap, B256Set, HashMap, HashSet},
+    map::{AddressMap, B256Set, HashMap, HashSet},
 };
 use alloy_sol_types::SolEvent;
 use futures::StreamExt;
@@ -387,7 +387,7 @@ impl KeyExpiryTracker {
         let key = KeyId { account, key_id };
 
         match self.key_to_txs.entry(key) {
-            alloy_map::Entry::Occupied(mut entry) => {
+            alloy_primitives::map::Entry::Occupied(mut entry) => {
                 let (existing_expiry, txs) = entry.get_mut();
                 debug_assert_eq!(
                     *existing_expiry, expiry,
@@ -395,7 +395,7 @@ impl KeyExpiryTracker {
                 );
                 txs.insert(tx_hash);
             }
-            alloy_map::Entry::Vacant(entry) => {
+            alloy_primitives::map::Entry::Vacant(entry) => {
                 entry.insert((expiry, [tx_hash].into_iter().collect()));
                 self.expiry_map.entry(expiry).or_default().insert(key);
             }
@@ -409,7 +409,8 @@ impl KeyExpiryTracker {
             return;
         };
 
-        let alloy_map::Entry::Occupied(mut key_entry) = self.key_to_txs.entry(key) else {
+        let alloy_primitives::map::Entry::Occupied(mut key_entry) = self.key_to_txs.entry(key)
+        else {
             return;
         };
 


crates/transaction-pool/src/maintain.rs
```diff
@@ -403,6 +403,34 @@ impl KeyExpiryTracker {
         }
     }
 
+    /// Removes transaction hashes from key expiry tracking.
+    ///
+    /// When transactions are mined or otherwise removed from the pool, their hashes
+    /// should be cleaned up to avoid unbounded memory growth in `key_to_txs`.
+    fn remove_mined(&mut self, mined_hashes: &[TxHash]) {
```
mattsse: should this accept an iter isntead?

crates/transaction-pool/src/maintain.rs
```diff
@@ -648,6 +677,12 @@ where
                         let count = removed_txs.len();
 
                         if count > 0 {
+                            // Clean up expiry tracking for paused txs
+                            let paused_hashes: Vec<TxHash> =
+                                removed_txs.iter().map(|tx| *tx.hash()).collect();
+                            state.remove_mined(&paused_hashes);
+                            state.key_expiry.remove_mined(&paused_hashes);
```
mattsse: if key_expiry is inside state, and both have remove_mined, I'd assume that the outer also calls the inner

0xrusowsky: [`4564a0e` (this PR)](https://github.com/tempoxyz/tempo/pull/2878/commits/4564a0ed0953adbe33e7572df183a12f85360cf9)

crates/transaction-pool/src/maintain.rs
0xrusowsky: [`4564a0e` (this PR)](https://github.com/tempoxyz/tempo/pull/2878/commits/4564a0ed0953adbe33e7572df183a12f85360cf9)

crates/transaction-pool/src/maintain.rs
```diff
@@ -10,7 +10,7 @@ use crate::{
 use alloy_consensus::transaction::TxHashRef;
 use alloy_primitives::{
     Address, TxHash,
-    map::{AddressMap, B256Set, HashMap, HashSet},
+    map::{AddressMap, B256Set, Entry as AlloyEntry, HashMap, HashSet},
```
mattsse: not a fan of this alias, would rather use full paths

crates/transaction-pool/src/maintain.rs
```diff
@@ -239,19 +239,22 @@ impl TempoPoolState {
...
     ///
     /// Reth removes mined transactions from the pool, but the additional tracking in
-    /// `expiry_map` and `tx_to_expiry` is not cleaned up. This method removes the
-    /// entries from both maps.
-    fn remove_mined(&mut self, mined_hashes: &[TxHash]) {
-        for hash in mined_hashes {
-            if let Some(valid_before) = self.tx_to_expiry.remove(hash)
+    /// `expiry_map`, `tx_to_expiry`, and `key_expiry` is not cleaned up. This method
+    /// removes the entries from all tracking maps.
+    fn remove_mined(&mut self, mined: &[(TxHash, bool)]) {
```
mattsse: what is this extra bool?

crates/transaction-pool/src/maintain.rs
```diff
@@ -648,6 +685,13 @@ where
                         let count = removed_txs.len();
 
                         if count > 0 {
+                            // Clean up expiry tracking for paused txs
+                            let paused: Vec<(TxHash, bool)> = removed_txs
+                                .iter()
+                                .map(|tx| (*tx.hash(), tx.transaction.is_keychain()))
+                                .collect();
+                            state.remove_mined(&paused);
```
mattsse: same here

crates/transaction-pool/src/maintain.rs
```diff
@@ -239,19 +239,22 @@ impl TempoPoolState {
...
-                if hashes.is_empty() {
+                entry.get_mut().retain(|h| *h != hash);
+                if entry.get().is_empty() {
                     entry.remove();
                 }
             }
+
+            if is_keychain {
+                self.key_expiry.remove_tx(&hash);
+            }
```
0xKitsune: Is there another way to do this without passing in the `is_keychain` bool. Do we have access to the tx where we can call the `is_keychain()` method?

crates/transaction-pool/src/maintain.rs
0xKitsune: +1

crates/transaction-pool/src/maintain.rs
```diff
@@ -10,7 +10,7 @@ use crate::{
 use alloy_consensus::transaction::TxHashRef;
 use alloy_primitives::{
     Address, TxHash,
-    map::{AddressMap, B256Set, HashMap, HashSet},
+    map::{self as alloy_map, AddressMap, B256Set, HashMap, HashSet},
```
0xKitsune: Do we need this alias? Can we use fully qualified path instead?

crates/transaction-pool/src/maintain.rs
```diff
@@ -236,23 +236,20 @@ impl TempoPoolState {
...
+        {
+            entry.get_mut().retain(|h| *h != *hash);
+            if entry.get().is_empty() {
+                entry.remove();
             }
         }
+
+        if is_keychain {
+            self.key_expiry.untrack(hash);
+        }
```
0xKitsune, line 252: Can we call `key_expiry.remove_tx()` unconditionally, it looks like this function checks if the entry exists or exists early.

PR #2879, opened 2026-02-26
feat(precompiles): `CheckpointGuard` for batched state-changing ops

0xrusowsky:
introduces `CheckpointGuard`, a RAII guard that wraps EVM journal checkpoints for atomic state-mutation operation batching inside precompiles.

### Motivation

when a precompile caller swallows errors instead of propagating them (for business reasons), multi-step state mutations can fail mid-way and leak partial state.

### Solution

`CheckpointGuard` is returned by `StorageCtx::checkpoint()` and provides two guarantees:

- **auto-revert on drop**: if the guard is dropped without calling `.commit()`, all mutations since the checkpoint are rolled back.
- **hardfork-aware**: pre-T1C the guard is a no-op (no checkpoint created), so callers don't need conditional logic.

### DIFF
--- a/crates/precompiles/src/storage/thread_local.rs
+++ b/crates/precompiles/src/storage/thread_local.rs
@@ -344,12 +344,8 @@ impl StorageCtx {
     }
 
     /// Checks if a contract at the given address has bytecode deployed.
-    pub fn has_bytecode(&self, address: Address) -> bool {
-        if let Some(account_info) = self.get_account_info(address) {
-            !account_info.is_empty_code_hash()
-        } else {
-            false
-        }
+    pub fn has_bytecode(&self, address: Address) -> Result<bool> {
+        self.with_account_info(address, |info| Ok(!info.is_empty_code_hash()))
     }
 }
 
--- a/crates/precompiles/src/test_util.rs
+++ b/crates/precompiles/src/test_util.rs
@@ -256,7 +256,7 @@ impl TIP20Setup {
 
     /// Initialize pathUSD if needed and return it.
     fn path_usd_inner(&self) -> Result<TIP20Token> {
-        if is_initialized(PATH_USD_ADDRESS) {
+        if is_initialized(PATH_USD_ADDRESS)? {
             return TIP20Token::from_address(PATH_USD_ADDRESS);
         }
 
@@ -279,7 +279,7 @@ impl TIP20Setup {
     /// Initialize the TIP20 factory if needed.
     pub fn factory() -> Result<TIP20Factory> {
         let mut factory = TIP20Factory::new();
-        if !is_initialized(TIP20_FACTORY_ADDRESS) {
+        if !is_initialized(TIP20_FACTORY_ADDRESS)? {
             factory.initialize()?;
         }
         Ok(factory)
@@ -315,7 +315,7 @@ impl TIP20Setup {
             }
             Action::ConfigureToken { address } => {
                 assert!(
-                    is_initialized(address),
+                    is_initialized(address)?,
                     "token not initialized, use `fn create` instead"
                 );
                 TIP20Token::from_address(address)?
@@ -375,7 +375,7 @@ impl TIP20Setup {
 
 /// Checks if a contract at the given address has bytecode deployed.
 #[cfg(any(test, feature = "test-utils"))]
-fn is_initialized(address: Address) -> bool {
+fn is_initialized(address: Address) -> Result<bool> {
     crate::storage::StorageCtx.has_bytecode(address)
 }
 


crates/precompiles/src/storage/hashmap.rs
```diff
@@ -1,10 +1,18 @@
...
+};
 use std::collections::HashMap;
 use tempo_chainspec::hardfork::TempoHardfork;
 
 use crate::{error::TempoPrecompileError, storage::PrecompileStorageProvider};
 
+/// Snapshot of mutable state for checkpoint/revert support.
+///
+/// PERF: naive cloning strategy due to its limited usage.
+struct Snapshot(HashMap<(Address, U256), U256>);
```
rakita: HashMap access piles up, any ordinary warm CALL is 100gas that is same price as TSTORE (That just does hashmap insert/set), so any additional HashMap access makes it even more expensive.

rakita: Ah this is only for tests

0xrusowsky: yes, this is only for tests. It a naive snapshot strategy for the DEX unit test

the `EvmPrecompileStorageProvider` uses the real journal, which is what matters

crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -583,6 +583,9 @@ impl StablecoinDEX {
         TIP20Token::from_address(non_escrow_token)?
             .ensure_transfer_authorized(self.address, sender)?;
 
+        // CHECKPOINT START: atomically batch all state-changing operations
+        let batch = self.storage.checkpoint();
```
rakita: Should we move it at the top of the function, just to be safe?

0xrusowsky: addressed in [`61d3fa6` (this PR)](https://github.com/tempoxyz/tempo/pull/2879/commits/61d3fa6f594bebed9ba7ff85be14f598f7b1384c)

crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -599,10 +602,16 @@ impl StablecoinDEX {
...
         // Create the flip order
         let order_id = self.next_order_id()?;
-        self.increment_next_order_id()?;
         let order = Order::new_flip(order_id, sender, book_key, amount, tick, is_bid, flip_tick)
             .map_err(|_| StablecoinDEXError::invalid_flip_tick())?;
 
+        // Commit the flip order
+        if self.storage.spec().is_t1c() {
+            // PERF: skip 1 redundant SLOAD
+            self.next_order_id.write(order_id + 1)?;
```
legion2002, line 612: nice

crates/precompiles/src/storage/evm.rs
```diff
@@ -240,6 +240,21 @@ impl<'a> PrecompileStorageProvider for EvmPrecompileStorageProvider<'a> {
     fn is_static(&self) -> bool {
         self.is_static
     }
+
+    #[inline]
+    fn checkpoint(&mut self) -> revm::context::journaled_state::JournalCheckpoint {
```
0xKitsune: nit: Consider importing this once rather than fully qualifying the path multiple times.
```suggestion
    fn checkpoint(&mut self) -> JournalCheckpoint {
```

crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -4711,4 +4723,70 @@ mod tests {
...
+                // 1. `fill_order` credited alice `amount` base before `place_flip`
+                // 2. `sub_balance` debited it back
+                // 3. `commit_order_to_book` failed
+                let alice_base = exchange.balance_of(alice, base_token)?;
+                let next_id_after = exchange.next_order_id()?;
+
+                if spec.is_t1c() {
+                    // Checkpoint reverts both sub_balance and order_id
+                    assert_eq!(alice_base, amount,);
+                    assert_eq!(next_id_after, next_id_before,);
```
0xKitsune: nit:
```suggestion
                    assert_eq!(alice_base, amount);
                    assert_eq!(next_id_after, next_id_before);
```

crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -4711,4 +4723,70 @@ mod tests {
...
+                } else {
+                    // No checkpoint — partial state leaks
+                    assert_eq!(alice_base, 0,);
+                    assert_eq!(next_id_after, next_id_before + 1,);
```
0xKitsune: nit: 
```suggestion
                    assert_eq!(alice_base, 0);
                    assert_eq!(next_id_after, next_id_before + 1);
```

crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -4711,4 +4723,70 @@ mod tests {
...
+                }
+
+                // verify that `OrderPlaced` event was never emitted due to poisoned tick's revert
+                assert!(
+                    exchange.emitted_events().last().is_some_and(
+                        |e| e.topics()[0] == IStablecoinDEX::OrderFilled::SIGNATURE_HASH
```
0xKitsune: If we are asserting that the last event wasnt `OrderPlaced` should we udpate this to `|e| e.topics()[0] =! IStablecoinDEX::OrderPlaced::SIGNATURE_HASH`?

crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -534,6 +534,9 @@ impl StablecoinDEX {
         // Compute book_key from token pair
         let book_key = compute_book_key(token, quote_token);
 
+        // CHECKPOINT START: atomically batch all state-changing operations
+        let batch = self.storage.checkpoint();
```
0xKitsune: Can we add additional comments why we do this here? When reading this without context its not clear why we create a checkpoint in place flip and nowhere else.

crates/precompiles/src/storage/thread_local.rs
```diff
@@ -174,6 +174,65 @@ impl StorageCtx {
...
+///
+/// ```ignore
+/// let guard = self.storage.checkpoint();
+/// self.sstore(addr, key, value)?;  // reverted on drop (T1C+)
+/// self.emit_event(...)?;
+/// guard.commit();  // finalizes all mutations
+/// ```
+pub struct CheckpointGuard {
+    checkpoint: Option<JournalCheckpoint>,
+}
```
0xKitsune, line 219: Why do we use `Option<JournalCheckpoint>` instead of `JournalCheckpoint` here? Could we simplify `fn checkpoint` to return `Option<CheckpointGuard>` instead?

crates/precompiles/src/storage/hashmap.rs
```diff
@@ -151,6 +161,25 @@ impl PrecompileStorageProvider for HashMapStorageProvider {
...
     }
+
+    fn checkpoint(&mut self) -> JournalCheckpoint {
+        let idx = self.snapshots.len();
+        self.snapshots.push(Snapshot(self.internals.clone()));
+        JournalCheckpoint {
+            log_i: 0,
+            journal_i: idx,
+        }
+    }
```
0xKitsune, line 178: Will this also checkpoint events that have been emitted? It looks like this will not be covered and any tests that assert events will contain the non reverted state? Lmk if I am overlooking something though.

crates/precompiles/src/storage/thread_local.rs
0xrusowsky: cause  we need to impl `Drop` to auto-revert

0xrusowsky: cause 
1. we need to impl `Drop`
2. pre T1C it is always None (no-op)

crates/precompiles/src/storage/hashmap.rs
0xrusowsky: that's correct, this will only track state changes.

right now the only usecase is the DEX and it only needs to checkpoint state changes.

if we ever need to checkpoint/test other stuff we can expand it (i just wanted to keep Snapshots minimal as cloning is inefficient)

PR #2883, opened 2026-02-26
Fix TIP-1000 and TIP-1010 per review feedback

dankrad:
## Summary

Addresses review comments on TIP-1000 and TIP-1010:

**TIP-1000:**
- Clarify that the 250k SSTORE state creation cost excludes EIP-2929 cold/warm access costs (total is 252,100 cold or 250,100 warm)
- Fix all "keccak + codesize" references to "keccak + nonce" — codesize is not a separate account field
- Clarify that the 500k CREATE cost includes the contract's nonce write (0 → 1); no separate 250k account creation charge for contracts
- Document 7702 delegation keccak write cost (250k for no-delegation → delegation)
- Assume initialized sender in all example computations

**TIP-1010:**
- Rename "Payment Lane Gas Limit" section to "Block Gas Limit" (500M is total block capacity, not payment-specific) with explicit gas budget breakdown
- Fix test case 3 from stale 25M to correct 30M general gas limit
- Replace unenforceable Invariant 1 with concrete base fee invariant
- Fix Invariant 5 to match the actual shared capacity model (non-shared 450M, general 30M, shared 50M)
- Remove sender account creation from tx gas cap rationale and cost table

## Test plan

- [ ] Verify TIP-1000 and TIP-1010 content is internally consistent
- [ ] Verify gas figures match implementation constants
- [ ] Existing tests pass (no logic changes, only docs + comments)


tips/ref-impls/test/helpers/TxBuilder.sol
```diff
@@ -46,7 +46,7 @@ library TxBuilder {
     uint64 constant STATE_CREATION_COST = 250_000; // SSTORE zero→non-zero
     uint64 constant CREATE_BASE_COST = 500_000; // CREATE/CREATE2 base
     uint64 constant CODE_DEPOSIT_COST = 1000; // per byte
-    uint64 constant CREATE_FIELDS_COST = 500_000; // keccak + codesize (2 × 250,000)
+    uint64 constant CREATE_FIELDS_COST = 500_000; // keccak + nonce (2 × 250,000)
```
rakita:
```suggestion
    uint64 constant CREATE_FIELDS_COST = 500_000; // nonce (Account creation) + bytecode insertion (2 × 250,000)
```

We create the account with nonce bump, and we insert the bytecode to saparate table. So is it two creations that is covered by base cost.

It is not exactly the keccak that is covered. There is an additional price of 6gas per word of data that is hashed (keccak).

Edit: maybe comment from here https://github.com/tempoxyz/tempo/pull/2883#discussion_r2859047948

tips/tip-1000.md
```diff
@@ -12,7 +12,7 @@ status: Approved
 
 ## Abstract
 
-This TIP increases the gas cost for creating new state elements, accounts, and contract code to provide economic protection against state growth spam attacks. The proposal increases the cost of writing a new state element from 20,000 gas to 250,000 gas, introduces a 250,000 gas charge for account creation (when the account's nonce is first written), and implements a new contract creation cost model: 1,000 gas per byte of contract code plus 500,000 gas for keccak hash and codesize fields.
+This TIP increases the gas cost for creating new state elements, accounts, and contract code to provide economic protection against state growth spam attacks. The proposal increases the cost of writing a new state element from 20,000 gas to 250,000 gas, introduces a 250,000 gas charge for account creation (when the account's nonce is first written), and implements a new contract creation cost model: 1,000 gas per byte of contract code plus 500,000 gas for keccak hash and nonce fields.
```
rakita:
```suggestion
This TIP increases the gas cost for creating new state elements, accounts, and contract code to provide economic protection against state growth spam attacks. The proposal increases the cost of writing a new state element from 20,000 gas to 250,000 gas, introduces a 250,000 gas charge for account creation (when the account's nonce is first written), and implements a new contract creation cost model: 1,000 gas per byte of contract code plus 500,000 base gas for nonce bump (account creation) and bytecode insertion.
```

~~Keccak is priced differently as 6ga sper word and this is only for CREATE2, for CREATE and create tx you dont have keccak~~

Did you mean keccak hash, as in merkle trie hash? I would propose rephrasing it as account creation and bytecode insertion.

tips/tip-1000.md
```diff
@@ -109,8 +112,9 @@ A transaction is invalid if the minimal costs of a (reverting) transaction can't
 * Transaction with `nonce == 0` require an additional 250,000 gas
 * Tempo transactions with any `nonce_key` and `nonce == 0` require an additional 250,000 gas
 * Changes to EIP-7702 authorization lists:
-   * EIP-7702 authorisation list entries with `auth_list.nonce == 0` require an additional 250,000 gas.
    * The base cost per authorization is reduced to 12,500 gas
+   * EIP-7702 authorisation list entries with `auth_list.nonce == 0` require an additional 250,000 gas (account creation for the nonce field)
+   * EIP-7702 authorisation list entries going from no delegation to delegation require an additional 250,000 gas (state creation for the keccak/code hash field)
```
rakita, line 119: We are not doing this in the code, we just did the previous line. We could fix it by adding more gas if condition
`if !authorization.address().is_zero()` is true

tips/tip-1000.md
dankrad: I mean the keccak of the bytecode

dankrad: Maybe we should just frame this as "base cost" and not establish any link to particular storage slots created?

tips/tip-1000.md
dankrad: Should we do this? It does create new state, right?

tips/tip-1000.md
dankrad: Made this change now, wdyt?

tips/tip-1000.md
dankrad: @rakita

rakita: Yeah, we should add it. It is an additional cost of adding new bytecode to the bytecode table.

rakita (approved):

PR #2890, opened 2026-02-26
feat: val config v2 - separate vals into active and inactive

howydev:
Changes
1. Separate validator arrays into active and inactive vals so active validator array reads are ungriefable 
2. Add pagination to inactive validator reads for safety
3. State changing functions for validators now operate on `index`
4. `_byIndex` view functions now return 0 if the validator associated with that value is deactivated

**Note: On correctness of migration, there's a future PR that will overhaul migration. So any security issues around migration should be checked in that PR**

Callouts:
1. There's a tiny footgun in which the owner accidentally front runs a validator state change call with a deactivate validator call. In this case, there's a non-zero chance that the swap-and-pop affects the index of the validator. This is a extremely low chance though - the validator has to be the most recent validator, the call has to happen in the same block and the owner's call has to happen first. There's no danger in these cases - the validator's call reverts and they'll need to retry the call - so we'll leave this as a note within the security section of the 1017 spec. 
2. Invariant changes - pubkeys can now be duplicated between active and inactive sets. But because a signature is required to add a key, any duplications must be from the same entity, so we dont need to consider malicious duplications

### DIFF
--- a/crates/contracts/src/precompiles/validator_config_v2.rs
+++ b/crates/contracts/src/precompiles/validator_config_v2.rs
@@ -32,11 +32,17 @@ crate::sol! {
         // View functions
         // =====================================================================
 
-        /// Get the complete set of validators (including deleted)
-        function getAllValidators() external view returns (Validator[] memory validators);
-
         /// Get only active validators (deactivatedAtHeight == 0)
         function getActiveValidators() external view returns (Validator[] memory validators);
+
+        /// Get inactive (deactivated) validators with pagination
+        function getInactiveValidators(uint64 startIndex) external view returns (Validator[] memory validators);
+
+        /// Get the number of inactive (deactivated) validators
+        function inactiveValidatorCount() external view returns (uint64);
+
+        /// Get the number of currently active validators
+        function activeValidatorCount() external view returns (uint64);
 
         /// Get the block height at which the contract was initialized
         function getInitializedAtHeight() external view returns (uint64);
@@ -76,11 +82,11 @@ crate::sol! {
         ) external;
 
         /// Deactivate a validator (owner or validator)
-        function deactivateValidator(address validatorAddress) external;
+        function deactivateValidator(uint64 idx) external;
 
         /// Rotate a validator to new identity (owner or validator)
         function rotateValidator(
-            address validatorAddress,
+            uint64 idx,
             bytes32 publicKey,
             string calldata ingress,
             string calldata egress,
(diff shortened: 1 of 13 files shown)


crates/contracts/src/precompiles/validator_config_v2.rs
```diff
@@ -76,11 +82,11 @@ crate::sol! {
         ) external;
 
         /// Deactivate a validator (owner or validator)
-        function deactivateValidator(address validatorAddress) external;
+        function deactivateValidator(uint64 idx) external;
 
         /// Rotate a validator to new identity (owner or validator)
         function rotateValidator(
-            address validatorAddress,
+            uint64 idx,
```
SuperFluffy, line 89: I don't understand the reasoning behind this. Why does this need an index now?

Is the index stable? IIRC, we had a mapping address -> index.

tips/ref-impls/src/ValidatorConfigV2.sol
```diff
@@ -414,7 +439,8 @@ contract ValidatorConfigV2 is IValidatorConfigV2 {
     )
         internal
     {
-        uint64 idx = uint64(validatorsArray.length);
+        uint64 idx = deactivatedAtHeight == 0 ? uint64(activeValidatorsArray.length) : 0;
```
SuperFluffy, line 442: These indices are unstable. That's a very dangerous design.

SuperFluffy (commented):
I'll do a second pass, but the unstable indices raise concern.

If your data structure gives out indices as part of its public API, we better make sure they don't become invalid.

Instead of splitting the the vector into two, why don't we maintain a single vector of all validators (as before), and have a second array of active validators that only contains indices into the global array?

That way your indices are stable and cannot be invalidated.

SuperFluffy:
I propose this alternative: https://github.com/tempoxyz/tempo/pull/2900

tips/ref-impls/src/ValidatorConfigV2.sol
```diff
@@ -286,7 +311,7 @@ contract ValidatorConfigV2 is IValidatorConfigV2 {
         if (idx == 0) {
             revert ValidatorNotFound();
         }
-        return validatorsArray[idx - 1];
+        return activeValidatorsArray[idx - 1];
```
SuperFluffy, line 314: On DKG failure CL reverts to the previous set (to use DKG nomenclature - the committee of epoch E+1 will be the dealers of E because the resharing process failed and the players of E do not have valid shares). This means that for a given epoch validators can be in the committee but no longer active.

This design breaks CL and will cause an unrecoverable chainhalt.

SuperFluffy (changes requested):
This design breaks CL, see specifically this PR https://github.com/tempoxyz/tempo/pull/2872 how `validatorByPublicKey` will be used.

PR #2892, opened 2026-02-26
fix(primitives): return actual primitive `Signature` from AA envelope

mablr:
If AA tx is signed with `alloy_primitives::Signature`, return its ref directly. Otherwise, return zeroed signature `TEMPO_SYSTEM_TX_SIGNATURE`.

Added unit tests

### DIFF
--- a/crates/primitives/src/transaction/envelope.rs
+++ b/crates/primitives/src/transaction/envelope.rs
@@ -446,7 +446,7 @@ impl reth_primitives_traits::serde_bincode_compat::RlpBincode for TempoTxEnvelop
 
 #[cfg(feature = "reth-codec")]
 mod codec {
-    use crate::{TempoSignature, TempoTransaction};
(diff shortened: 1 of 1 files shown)


crates/primitives/src/transaction/envelope.rs
```diff
@@ -533,9 +533,16 @@ mod codec {
...
+                Self::AA(tx) => {
+                    // If signed with [`alloy_primitives::Signature`], return it directly.
+                    // Otherwise, return the fake system zeroed signature.
+                    if let TempoSignature::Primitive(PrimitiveSignature::Secp256k1(sig)) =
+                        tx.signature()
+                    {
+                        sig
+                    } else {
+                        &TEMPO_SYSTEM_TX_SIGNATURE
+                    }
```
mattsse, line 545: this will alter what we write to disk,

this is basically just a hack to reuse existing codec code

mattsse (commented):
see also https://github.com/tempoxyz/tempo/pull/2894

PR #2910, opened 2026-02-27
feat: validator config v2 all changes

howydev:
TLDR:
1. Uses the interfaces from https://github.com/tempoxyz/tempo/pull/2890 but the data structures from https://github.com/tempoxyz/tempo/pull/2900
2. WIP new implementation of `rotateValidator` that makes other state changing validator calls safe

This creates an invariant that each validator has a static `index` field, and all state changing validator operations target `index`. This removes race conditions between state changing calls

Full changes:
1. Add `ValidatorStorage` field that has extra `uint64 activeIndices` field
2. Add `activeIndices` array
3. `deactivateValidator`, `rotateValidator`, `setIpAddresses`, `transferValidatorOwnership`, take in `idx` instead of `address`
4. Add `_toValidator` that parses from `ValidatorStorage` to remove the `activeIndices` field. All validator getters (returns `Validator`) now
    1. Should not remove `index` from the struct returned by getters because that’s useful in identifying the validator entry, which is used for state changing calls 
5. Move unique ingress IP check to `addValidator` away from `_validateRotateParams`. `_updateIngressIp` now does the uniqueness check + updating unique fields
6. `addValidator` returns `uint64 index` of which location that validator was added into

### DIFF
--- a/crates/precompiles/src/validator_config_v2/mod.rs
+++ b/crates/precompiles/src/validator_config_v2/mod.rs
@@ -3162,6 +3162,183 @@ mod tests {
     }
 
     #[test]
+    fn test_add_validator_rejects_wrong_key_signature() -> eyre::Result<()> {
+        let mut storage = HashMapStorageProvider::new(1);
+        let owner = Address::random();
+        let validator = Address::random();
+        let fee_recipient = Address::random();
+        StorageCtx::enter(&mut storage, || {
+            let mut vc = ValidatorConfigV2::new();
+            vc.initialize(owner)?;
+
+            // Generate a valid keypair for a different key
+            let (pubkey, _) = make_test_keypair_and_signature(
+                validator,
+                "192.168.1.1:8000",
+                "192.168.1.1",
+                SignatureKind::Add { fee_recipient },
+            );
+
+            // Generate signature from a completely different key
+            let (_, wrong_sig) = make_test_keypair_and_signature(
+                validator,
+                "192.168.1.1:8000",
+                "192.168.1.1",
+                SignatureKind::Add { fee_recipient },
+            );
+
+            vc.storage.set_block_number(200);
+            let result = vc.add_validator(
+                owner,
+                make_add_call(
+                    validator,
+                    pubkey,
+                    "192.168.1.1:8000",
+                    "192.168.1.1",
+                    fee_recipient,
+                    wrong_sig,
+                ),
+            );
+            assert_eq!(
+                result,
+                Err(ValidatorConfigV2Error::invalid_signature().into())
+            );
+
+            Ok(())
+        })
+    }
+
+    #[test]
+    fn test_add_validator_rejects_wrong_namespace_signature() -> eyre::Result<()> {
+        let mut storage = HashMapStorageProvider::new(1);
+        let owner = Address::random();
+        let validator = Address::random();
+        let fee_recipient = Address::random();
+        StorageCtx::enter(&mut storage, || {
+            let mut vc = ValidatorConfigV2::new();
+            vc.initialize(owner)?;
+
+            // Sign with ROTATE namespace, but try to ADD
+            let (pubkey, sig) = make_test_keypair_and_signature(
+                validator,
+                "192.168.1.1:8000",
+                "192.168.1.1",
+                SignatureKind::Rotate,
+            );
+
+            vc.storage.set_block_number(200);
+            let result = vc.add_validator(
+                owner,
+                make_add_call(
+                    validator,
+                    pubkey,
+                    "192.168.1.1:8000",
+                    "192.168.1.1",
+                    fee_recipient,
+                    sig,
+                ),
+            );
+            assert_eq!(
+                result,
+                Err(ValidatorConfigV2Error::invalid_signature().into())
+            );
+
+            Ok(())
+        })
+    }
+
+    #[test]
+    fn test_rotate_validator_rejects_wrong_key_signature() -> eyre::Result<()> {
+        let mut storage = HashMapStorageProvider::new(1);
+        let owner = Address::random();
+        let validator = Address::random();
+        let fee_recipient = Address::random();
+        StorageCtx::enter(&mut storage, || {
+            let mut vc = ValidatorConfigV2::new();
+            vc.initialize(owner)?;
+
+            // Add a valid validator first
+            vc.storage.set_block_number(200);
+            vc.add_validator(
+                owner,
+                make_valid_add_call(validator, "192.168.1.1:8000", "192.168.1.1", fee_recipient),
+            )?;
+
+            // Generate a new pubkey for rotation
+            let (new_pubkey, _) = make_test_keypair_and_signature(
+                validator,
+                "10.0.0.1:8000",
+                "10.0.0.1",
+                SignatureKind::Rotate,
+            );
+
+            // Sign with a different key
+            let (_, wrong_sig) = make_test_keypair_and_signature(
+                validator,
+                "10.0.0.1:8000",
+                "10.0.0.1",
+                SignatureKind::Rotate,
+            );
+
+            vc.storage.set_block_number(300);
+            let result = vc.rotate_validator(
+                owner,
+                IValidatorConfigV2::rotateValidatorCall {
+                    idx: 0,
+                    publicKey: new_pubkey,
+                    ingress: "10.0.0.1:8000".to_string(),
+                    egress: "10.0.0.1".to_string(),
+                    signature: wrong_sig.into(),
+                },
+            );
+            assert_eq!(
+                result,
+                Err(ValidatorConfigV2Error::invalid_signature().into())
+            );
+
+            Ok(())
+        })
+    }
+
+    #[test]
+    fn test_add_validator_rejects_malformed_signature() -> eyre::Result<()> {
+        let mut storage = HashMapStorageProvider::new(1);
+        let owner = Address::random();
+        let validator = Address::random();
+        let fee_recipient = Address::random();
+        StorageCtx::enter(&mut storage, || {
+            let mut vc = ValidatorConfigV2::new();
+            vc.initialize(owner)?;
+
+            let (pubkey, _) = make_test_keypair_and_signature(
+                validator,
+                "192.168.1.1:8000",
+                "192.168.1.1",
+                SignatureKind::Add { fee_recipient },
+            );
+
+            vc.storage.set_block_number(200);
+            let result = vc.add_validator(
+                owner,
+                make_add_call(
+                    validator,
+                    pubkey,
+                    "192.168.1.1:8000",
+                    "192.168.1.1",
+                    fee_recipient,
+                    vec![0xde, 0xad],
+                ),
+            );
+            assert_eq!(
+                result,
+                Err(ValidatorConfigV2Error::invalid_signature_format().into())
+            );
+
+            Ok(())
+        })
+    }
+
+    #[test]
     fn test_ipv4_ipv6_different_ips() -> eyre::Result<()> {
         let mut storage = HashMapStorageProvider::new(1);
         let owner = Address::random();


tips/ref-impls/src/ValidatorConfigV2.sol
```diff
@@ -27,15 +30,11 @@ contract ValidatorConfigV2 is IValidatorConfigV2 {
     IValidatorConfig public immutable v1 =
         IValidatorConfig(0xCccCcCCC00000000000000000000000000000000);
 
-    Validator[] internal validatorsArray;
+    ValidatorStorage[] internal validatorsArray;
 
     /// @notice Indices into the validatorsArray. This array does not preserve order.
     /// @dev 1-indexed: 0 means not found. Stored value is arrayIndex + 1.
-    uint64[] internal activeValidatorsByIndex;
```
howydev: moved into a u64 field in `validatorsArray`. We can pack with the other 3 u64's to save a storage slot

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -291,23 +308,33 @@ impl ValidatorConfigV2 {
...
             index: count,
+            active_idx,
             added_at_height,
             deactivated_at_height,
         };
 
         self.validators.push(v)?;
 
+        self.pubkey_to_index[pubkey].write(count + 1)?;
+        // if there are duplicated vals with addresses, the latter validator entry is correct
```
howydev: TODO: 
```suggestion
        // for any dups the prev entries must be deactivated since we check above
```

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -417,17 +444,38 @@ impl ValidatorConfigV2 {
...
-
         self.active_ingress_ips[Self::ingress_ip_key(&v.ingress)?].delete()?;
 
-        v.deactivated_at_height = block_height;
-        self.validators[idx].write(v)
+        self.validators[call.idx as usize]
+            .deactivated_at_height
+            .write(block_height)?;
+
+        // Pop-and-swap for active_indices
```
howydev:
```suggestion
        // Swap-and-pop for active_indices
```

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -458,35 +506,51 @@ impl ValidatorConfigV2 {
...
 
         let block_height = self.storage.block_number();
-        let (idx, mut old) = self.get_active_validator(call.validatorAddress)?;
 
-        self.update_ingress_ip_tracking(&old.ingress, &call.ingress)?;
+        self.update_ingress_ip_tracking(&v.ingress, &call.ingress)?;
 
-        old.deactivated_at_height = block_height;
-        self.validators[idx].write(old)?;
+        // Append deactivated snapshot of the old validator
```
howydev, line 724: note: we don't use the `append_validator` helper since that does `self.address_to_index[addr].write(count + 1)?;` which we don't want to update

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -1962,8 +2004,8 @@ mod tests {
             )?;
 
             assert_eq!(vc.validator_count()?, 2);
-            assert_eq!(vc.validator_by_index(0)?.deactivatedAtHeight, 300);
-            assert_eq!(vc.validator_by_index(1)?.deactivatedAtHeight, 0);
+            assert_eq!(vc.validator_by_index(0)?.deactivatedAtHeight, 0);
```
howydev, line 2648: note: these changed because we update the existing idx and append the to-be-deactivated validator entry instead

crates/contracts/src/precompiles/validator_config_v2.rs
```diff
@@ -32,8 +32,8 @@ crate::sol! {
         // View functions
         // =====================================================================
 
-        /// Get the complete set of validators (including deleted)
-        function getAllValidators() external view returns (Validator[] memory validators);
+        /// Get validators starting from a given index (including deactivated)
+        function getValidators(uint64 startIndex) external view returns (Validator[] memory validators);
```
joshieDo: why is this useful btw?

tips/ref-impls/src/ValidatorConfigV2.sol
```diff
@@ -141,27 +147,27 @@ contract ValidatorConfigV2 is IValidatorConfigV2 {
 
     /// @inheritdoc IValidatorConfigV2
     function rotateValidator(
-        address validatorAddress,
+        uint64 idx,
```
joshieDo, line 165: why? imo we keep the `addressToIndex[..] ` ? is it because address is not unique?

crates/contracts/src/precompiles/validator_config_v2.rs
SuperFluffy: I think we should completely remove this. This is a footgun.

tips/ref-impls/src/ValidatorConfigV2.sol
```diff
@@ -88,6 +86,7 @@ contract ValidatorConfigV2 is IValidatorConfigV2 {
         external
         onlyInitialized
         onlyOwner
+        returns (uint64 index)
```
SuperFluffy, line 89: Yeah, I think this makes sense. What's important is that we communicate that this objects does not correspond to any kind of index into an array, but is an actual ID for the validator "slot".

tips/ref-impls/src/interfaces/IValidatorConfigV2.sol
```diff
@@ -74,12 +74,24 @@ interface IValidatorConfigV2 {
...
+    }
+
+    /// @notice Validator information returned by view functions
     /// @param publicKey Ed25519 communication public key (non-zero, unique across all validators)
-    /// @param validatorAddress Ethereum-style address of the validator (unique across all validators)
+    /// @param validatorAddress Ethereum-style address of the validator (unique across active validators)
     /// @param ingress Address where other validators can connect (format: `<ip>:<port>`)
     /// @param egress IP address from which this validator will dial, e.g. for firewall whitelisting (format: `<ip>`)
-    /// @param index Position in validators array (assigned at creation, immutable)
+    /// @param index Position in validators array (stable across rotations for the same slot)
```
SuperFluffy, line 134: I think we should remove any mention of `position in array` or anything of that kind. That's an implementation detail and the returned value is an index corresponding to a validator "slot".

tips/ref-impls/src/ValidatorConfigV2.sol
```diff
@@ -98,31 +97,38 @@ contract ValidatorConfigV2 is IValidatorConfigV2 {
...
 
-        _deleteActive(idx - 1);
+        // do a pop-and-swap for validatorsArray
+        uint64 toDeactivateIndex = v.activeIdx - 1;
+        uint256 lastPos = activeIndices.length - 1;
+
+        if (toDeactivateIndex != lastPos) {
+            uint64 movedVal = activeIndices[lastPos];
+            activeIndices[toDeactivateIndex] = movedVal;
+            validatorsArray[movedVal - 1].activeIdx = toDeactivateIndex + 1;
```
SuperFluffy, line 139: This I don't understand - is this intended as a reverse lookup from all validators to active validators?

tips/ref-impls/src/interfaces/IValidatorConfigV2.sol
```diff
@@ -74,12 +74,24 @@ interface IValidatorConfigV2 {
...
 
-    /// @notice Validator information (V2 - append-only, delete-once)
+    /// @dev Internal storage struct with bookkeeping fields not exposed in the public Validator struct.
+    struct ValidatorStorage {
+        bytes32 publicKey;
+        address validatorAddress;
+        string ingress;
+        string egress;
+        uint64 index;
+        uint64 activeIdx;
```
SuperFluffy, line 124: OK, so this is a reverse lookup, yes?

tips/ref-impls/src/interfaces/IValidatorConfigV2.sol
```diff
@@ -113,34 +125,34 @@ interface IValidatorConfigV2 {
...
     ///      keccak256(abi.encodePacked("TEMPO", "_VALIDATOR_CONFIG_V2_ROTATE_VALIDATOR", chainId, contractAddress, validatorAddress, ingress, egress))
-    /// @param validatorAddress The address of the validator to rotate
+    /// @param idx The global index of the validator to rotate
     /// @param publicKey The new validator's Ed25519 communication public key
     /// @param ingress The new validator's inbound address `<ip>:<port>` for incoming connections
     /// @param egress The new validator's outbound IP address `<ip>` for firewall whitelisting
     /// @param signature Ed25519 signature (64 bytes) proving ownership of the new public key
     function rotateValidator(
-        address validatorAddress,
+        uint64 idx,
```
SuperFluffy, line 199: We should make it explicit that this is not an index into any kind of array, but an identifier that uniquely identifies a validator slot.

tips/ref-impls/src/interfaces/IValidatorConfigV2.sol
```diff
@@ -152,23 +164,18 @@ interface IValidatorConfigV2 {
     /// @dev Can be called by the contract owner or by the validator's own address.
     ///      This allows validators to update their network addresses without requiring
     ///      a full rotation.
-    /// @param validatorAddress The address of the validator to update
+    /// @param idx The global index of the validator to update
```
SuperFluffy, line 211: An implementation detail that should not be part of any public API - we should change this to the ID that identifies the validator slot and is kept stable across rotations.

crates/contracts/src/precompiles/validator_config_v2.rs
```diff
@@ -89,14 +89,14 @@ crate::sol! {
 
         /// Update IP addresses (owner or validator)
         function setIpAddresses(
-            address validatorAddress,
+            uint64 idx,
```
SuperFluffy, line 100: This applies here and anywhere else - calling this `index` is inconsistent if what we want is to refer to a stable index that is kept constant for across rotations.

tips/ref-impls/src/ValidatorConfigV2.sol
SuperFluffy: This is to prevent front running the contract owner. If $owner wants to disable $validator, $validatorOwner can call transferValidatorOwnership preventing $owner from disabling it.

tips/ref-impls/src/ValidatorConfigV2.sol
```diff
@@ -173,32 +179,47 @@ contract ValidatorConfigV2 is IValidatorConfigV2 {
...
+                egress: oldValidator.egress,
+                index: appendedIdx,
+                activeIdx: 0,
+                addedAtHeight: oldValidator.addedAtHeight,
+                deactivatedAtHeight: uint64(block.number)
+            })
+        );
+
+        // Update pubkeyToIndex: old pubkey → deactivated copy
+        pubkeyToIndex[oldValidator.publicKey] = appendedIdx + 1;
```
SuperFluffy, line 222: This seems unnecessary. Why don't we just leave the old entry?

tips/ref-impls/src/ValidatorConfigV2.sol
```diff
@@ -173,32 +179,47 @@ contract ValidatorConfigV2 is IValidatorConfigV2 {
...
 
-        _deleteActive(idx - 1);
+        // Modify slot in-place with new identity
+        oldValidator.publicKey = publicKey;
+        oldValidator.ingress = ingress;
+        oldValidator.egress = egress;
+        oldValidator.addedAtHeight = uint64(block.number);
```
SuperFluffy, line 229: Likewise - so when we add a validator, we overwrite the old slot in-place and push the old one to the end of the queue.

That seems complicated.

Why don't we just modify it in place push the new one? All we need is to associate the "validator slot ID" with the new validator.

tips/ref-impls/src/ValidatorConfigV2.sol
howydev: yes, necessary to support o(1) deactivations

in the parent branch this was a separate map(address -> uint), but we can save a storage slot per val by adding it to the val struct

tips/ref-impls/src/ValidatorConfigV2.sol
howydev: resolved during sync

tips/ref-impls/src/ValidatorConfigV2.sol
howydev: resolved during sync

crates/contracts/src/precompiles/validator_config_v2.rs
howydev: the pagination is useful because validators b/c in theory, a validator can spam rotations that inflate the size of validators array

but given that its not used in CL any more, agree that we can remove it. An equivalent getter here for other clients to use if necessary is: `for i in validatorCount, array.push(getValidatorByIndex(i))`

crates/contracts/src/precompiles/validator_config_v2.rs
howydev: imo `id` is an abstraction that has some leaks based on the current impl - it will be hard to grok why rotation creates a new validator entry with a different id associated with details from the old entry. IMO, although `index` reveals implementation details, it saves more time in the long run by preventing confusion

Feel free to push back if u disagree

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -458,35 +486,51 @@ impl ValidatorConfigV2 {
...
+        // Update pubkey_to_index: old pubkey → appended_idx + 1
+        self.pubkey_to_index[v.public_key].write(appended_idx + 1)?;
+
+        // Modify in-place at the original index
+        let mut updated = self.validators[call.idx as usize].read()?;
+        updated.public_key = call.publicKey;
+        updated.ingress = call.ingress;
+        updated.egress = call.egress;
+        updated.added_at_height = block_height;
+        self.validators[call.idx as usize].write(updated)?;
```
howydev, line 748: spec is updated in a separate PR

decofe:
## :eye: Cyclops Security Review

[`9a2724f`](https://github.com/tempoxyz/tempo/commit/9a2724f968cb)

🧭 Auditing · mode=`normal` · workers 0/2 done (2 left) · verify pending 0

| Worker | Engine | Progress | Status |
|--------|--------|:--------:|--------|
| `pr-2910-w1` | gemini-3.1-pro-preview | 🔍 [thread-1](file:///cyclops-audit/logs/pr-2910/conversations/pi-1772954510-589.json) · · | Running |
| `pr-2910-w2` | amp/deep | ✅ [thread-1](https://ampcode.com/threads/T-019ccc52-ef3b-707a-b4a8-c033103aa584) 🔍 [thread-2](https://ampcode.com/threads/T-019ccc57-0ea6-7385-8efe-8346ebd5237f) · | Running |

<details><summary>⚙️ Controls</summary>

- 🚀 Keep only 1 remaining iteration per worker after the current work finishes.
- 👀 Keep only 2 remaining iterations per worker after the current work finishes.
- ❤️ Let only worker 1 continue; other workers skip queued iterations.
- 😄 Let only worker 2 continue; other workers skip queued iterations.
- 🎉 End faster by skipping queued iterations and moving toward consolidation.
- 😕 Stop active workers/verifiers now and start consolidation immediately.

</details>

---
<details><summary>📜 4 events</summary>

🔍 `pr-2910-w1` iter 1/3 [audit-ripple.md]
🔍 `pr-2910-w2` iter 1/3 [audit-focused.md]
✅ `pr-2910-w2` iter 1 — clear | [Thread](https://ampcode.com/threads/T-019ccc52-ef3b-707a-b4a8-c033103aa584)
🔍 `pr-2910-w2` iter 2/3 [audit-ripple.md]

</details>

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -71,7 +73,9 @@ pub struct ValidatorConfigV2 {
     address_to_index: Mapping<Address, u64>,
     pubkey_to_index: Mapping<B256, u64>,
     next_dkg_ceremony: u64,
-    active_ingress_ips: Mapping<B256, bool>,
+    migration_skipped_count: u16,
```
decofe: ⚠️ **[ISSUE] Saturating `u16` skip counter can permanently block migration finalization**

`migration_skipped_count` is `u16` with `saturating_add(1)`. V1 validator cardinality is `u64` and V1 accepts any non-zero `bytes32` pubkey (no Ed25519 decode validation). If skipped entries exceed 65,535, the counter saturates and `total_processed` undercounts real progress. The migration index gate (`call.idx + total_processed + 1 == v1_count`) and initialization check (`validator_count + skipped >= v1_count`) both become unsatisfiable, permanently blocking migration.

**Recommended Fix:** Widen to `u64` (or `u32` with explicit bounds checks). Ensure migration arithmetic is exact and non-saturating.

howydev: not an issue - we wont have 65535 validator entries in v1!

tips/ref-impls/src/ValidatorConfigV2.sol
```diff
@@ -75,53 +80,79 @@ contract ValidatorConfigV2 is IValidatorConfigV2 {
...
         onlyInitialized
         onlyOwner
+        returns (uint64 index)
     {
         _validateAddParams(validatorAddress, publicKey, ingress, egress);
 
         bytes32 message = keccak256(
-            abi.encodePacked(block.chainid, address(this), validatorAddress, ingress, egress)
+            abi.encodePacked(
+                block.chainid,
```
howydev:
```suggestion
                uint64(block.chainid),
```

PR #2914, opened 2026-02-28
feat(keychain): add AccessKeySpend event gated behind T2

decofe:
Emits `AccessKeySpend(account, publicKey, token, amount, remainingLimit)` from `verify_and_update_spending` whenever an access key's spending limit is deducted. Gated behind `is_t2()`.

Only fires for actual access-key spends — root key txs (`key_id == Address::ZERO`) return early before reaching the emission point.

Prompted by: tanishk


jenpaff (changes requested):
lets add this with TIP-1011 in T3

PR #2918, opened 2026-03-02
feat: skip invalid V1 entries during ValidatorConfigV2 migration

howydev:
1. During migration, we skip if the entry has invalid pubkeys or duplicate ingress IPs. We also skip on egress IP parsing as a defensive check
2. Duplicate pubkeys overwrite the existing entry in place
3. A `migration_skipped_count` field tracks skipped indices and duplications so finalization is not blocked

Implementation: We enumerate over the validators in reverse, skipping on dups. This is equiv to enumerating forwards and overwriting on dups, but much simpler

### DIFF
--- a/crates/precompiles/src/validator_config_v2/mod.rs
+++ b/crates/precompiles/src/validator_config_v2/mod.rs
@@ -618,7 +618,7 @@ impl ValidatorConfigV2 {
         let skipped = self.migration_skipped_count.read()?;
 
         let total_processed = migrated + u64::from(skipped);
-        if call.idx >= v1_count || call.idx + total_processed + 1 != v1_count {
+        if call.idx + total_processed + 1 != v1_count {
             Err(ValidatorConfigV2Error::invalid_migration_index())?
         }
 


crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -76,6 +76,7 @@ pub struct ValidatorConfigV2 {
     address_to_index: Mapping<Address, u64>,
     pubkey_to_index: Mapping<B256, u64>,
     next_dkg_ceremony: u64,
+    migration_skipped_count: u64,
```
0xrusowsky: nit: this could be turned into `u16` and packed into `Config` since it is a var that is only relevant during migration like `init_at_height`.

using a u16 is safe as we know we won't have 2**16 validators in v1

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -631,40 +632,68 @@ impl ValidatorConfigV2 {
...
-        let v1 = v1();
-        if call.idx >= v1.validator_count()? {
-            Err(ValidatorConfigV2Error::validator_not_found())?
-        }
         let v1_val = v1.validators(v1.validators_array(call.idx)?)?;
 
-        self.require_new_address(v1_val.validatorAddress)?;
-        self.require_new_pubkey(v1_val.publicKey)?;
+        if PublicKey::decode(v1_val.publicKey.as_slice()).is_err()
+            || v1_val.validatorAddress.is_zero()
```
0xrusowsky: is this needed?

V1 config already validates that `publickKey` is `B256` and `!= B256::ZERO` on `fn add_validator()` and `fn update_validator()`

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -631,40 +632,68 @@ impl ValidatorConfigV2 {
...
+                    .write(self.migration_skipped_count.read()? + 1)?;
+                return Ok(());
+            }
+        };
+
+        if self.pubkey_to_index[v1_val.publicKey].read()? != 0 {
+            self.migration_skipped_count
+                .write(self.migration_skipped_count.read()? + 1)?;
+            return Ok(());
+        }
```
0xrusowsky, line 644: can we add cmnts for clarity

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -631,40 +632,68 @@ impl ValidatorConfigV2 {
...
+        {
+            Err(ValidatorConfigV2Error::address_already_has_validator())?
+        }
 
-        let ingress_hash = self.require_unique_ingress_ip(&v1_val.inboundAddress)?;
+        let now_active = v1_val.active;
+        let ingress_hash = Self::ingress_ip_key(&v1_val.inboundAddress)?;
 
-        let deactivated_at_height = if v1_val.active { 0 } else { block_height };
+        if now_active && self.active_ingress_ips[ingress_hash].read()? {
```
0xrusowsky, line 661: can we add cmnts for clarity?

howydev:
> For testing, the only thing I would propose is to ensure that the migration leads to the same result as the consensus layer's read-and-filter

Can add this as an invariant!

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -611,40 +632,72 @@ impl ValidatorConfigV2 {
         self.require_migration_owner(sender)?;
         let block_height = self.storage.block_number();
 
-        let current_count = self.validator_count()?;
-        if call.idx != current_count {
+        let v1 = v1();
+        let v1_count = v1.validator_count()?;
+        let skipped = self.migration_skipped_count.read()?;
+        let total_processed = self.validator_count()? + u64::from(skipped);
```
0xrusowsky: this should use `v1_count` otherwise we do a useless SLOAD

howydev: i made the change and just realized its a diff validator_count - v1 vs v2

0xrusowsky: ohhh ok, my bad

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -611,40 +612,73 @@ impl ValidatorConfigV2 {
...
+                .deactivated_at_height
+                .read()?
+                == 0
+        {
+            Err(ValidatorConfigV2Error::address_already_has_validator())?
+        }
 
-        let ingress_hash = self.require_unique_ingress_ip(&v1_val.inboundAddress)?;
+        let now_active = v1_val.active;
+        let ingress_hash = Self::ingress_ip_key(&v1_val.inboundAddress)?;
```
0xrusowsky, line 658: should this skip rather than reverting?

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -611,40 +612,66 @@ impl ValidatorConfigV2 {
...
+        // Skip if address is a duplicate of an existing entry
+        let addr_idx = self.address_to_index[v1_val.validatorAddress].read()?;
+        if addr_idx != 0
+            && self.validators[(addr_idx - 1) as usize]
+                .deactivated_at_height
+                .read()?
+                == 0
+        {
+            Err(ValidatorConfigV2Error::address_already_has_validator())?
+        }
```
0xrusowsky, line 655: this reverts rather than skipping, is this correct?

crates/precompiles/src/validator_config_v2/mod.rs
0xrusowsky: looks like this could cause a security issue (idk if accounted for in the design): https://github.com/tempoxyz/tempo/pull/3010#discussion_r2895523582

crates/precompiles/src/validator_config_v2/mod.rs
0xrusowsky: looks like this could cause a security issue (idk if accounted for in the design): https://github.com/tempoxyz/tempo/pull/3010#discussion_r2895523582

howydev: for both addrs and IPs - we're good since vals dont control addrs in v1 => they dont control the val entry in v1 or v2 till post migration

so the only drops would be validators that were already ignored from the POV of CL

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -611,40 +612,73 @@ impl ValidatorConfigV2 {
...
 
-        let current_count = self.validator_count()?;
-        if call.idx != current_count {
+        let v1 = v1();
+        let v1_count = v1.validator_count()?;
+        let migrated = self.validator_count()?;
+        let skipped = self.migration_skipped_count.read()?;
+
+        let total_processed = migrated + u64::from(skipped);
+        if call.idx >= v1_count || call.idx + total_processed + 1 != v1_count {
```
howydev: the nittiest of nits: not a fan of this, it adds extra compute to the hot path (valid migration call) to simplify the cold path (invalid index used)

PR #2920, opened 2026-03-02
feat: feed actor resolves blocks with finalized & notarized events

hamdiallam:
There is a small race between the feed actor getting the activity update and the marshaller persisting the block in storage. Rather than surface this race to the caller via an `Option` and out-of-band fills on queries, the feed actor will subscribe to the activity digest

* Update `CertifiedBlock` with block in question
* `ConsensusFeed::get_latest` excludes older notarizations.

### DIFF
--- a/crates/commonware-node/src/feed/actor.rs
+++ b/crates/commonware-node/src/feed/actor.rs
@@ -23,6 +23,7 @@ use commonware_cryptography::{bls12381::primitives::variant::MinSig, ed25519::Pu
 use commonware_macros::select;
 use commonware_runtime::{ContextCell, Handle, Spawner, spawn_cell};
 use commonware_utils::channel::oneshot;
+use eyre::eyre;
 use futures::{FutureExt, StreamExt};
 use std::{
     collections::BTreeMap,
@@ -35,7 +36,7 @@ use tempo_node::{
     TempoFullNode,
     rpc::consensus::{CertifiedBlock, Event},
 };
-use tracing::{info, info_span, instrument, warn, warn_span};
+use tracing::{error, info_span, instrument, warn, warn_span};
 
 use super::state::FeedStateHandle;
 use crate::{
@@ -133,7 +134,7 @@ impl<TContext: Spawner> Actor<TContext> {
     /// The loop races the oldest pending block subscription
     /// against incoming activity so events are emitted in order.
     async fn run(&mut self) {
-        loop {
+        let reason = loop {
             // We need a mutable reference to poll pending subscription. Thus if a new activity arrives,
             // we also need to re-insert this popped subscription.
             let mut oldest = OptionFuture::from(self.pending.pop_first().map(|(_, p)| p));
@@ -142,15 +143,15 @@ impl<TContext: Spawner> Actor<TContext> {
                 result = &mut oldest => {
                     match result {
                         Ok((_, activity, block)) => self.handle_activity(activity, block),
-                        Err(err) => warn_span!("feed_actor").in_scope(||
-                            warn!(%err, "pending block subscription cancelled")
+                        Err(error) => warn_span!("feed_actor").in_scope(||
+                            warn!(%error, "did not get pending block")
                         ),
                     }
                 },
 
                 activity = self.receiver.next() => {
                     let Some(activity) = activity else {
-                        break;
+                        break eyre!("mailbox closed");
                     };
 
                     if let Some(p) = oldest.take() {
@@ -159,9 +160,9 @@ impl<TContext: Spawner> Actor<TContext> {
                     self.subscribe(activity).await;
                 },
             );
-        }
+        };
 
-        info_span!("feed_actor").in_scope(|| info!("shutting down"));
+        info_span!("feed_actor").in_scope(|| error!(%reason, "shutting down"));
     }
 
     async fn subscribe(&mut self, activity: FeedActivity) {


crates/commonware-node/src/feed/actor.rs
```diff
@@ -68,127 +75,126 @@ impl<TContext: Spawner> Actor<TContext> {
...
+                    Activity::Notarization(n) => (n.proposal.round, n.proposal.payload),
+                    Activity::Finalization(f) => (f.proposal.round, f.proposal.payload),
+                    _ => continue 'subscribe,
+                };
+
+                select!(
+                    // We should expect to see the block immediately as the block was processed by the
+                    // consensus pipeline. We allow interrupts to avoid a syncing node from blocking this
+                    // actor from processing.
+                    block = self.marshal.subscribe(Some(round), payload).await => {
```
hamdiallam: Main logic here

crates/node/src/rpc/consensus/types.rs
```diff
@@ -12,11 +13,17 @@ use tokio::sync::broadcast;
...
-    /// Block height, if known. May be `None` if the block hasn't been stored yet.
-    pub height: Option<u64>,
     pub digest: B256,
+
     /// Hex-encoded full notarization or finalization.
     pub certificate: String,
+
+    // The Tempo block. Option wrapper will be removed after one release so that
+    // any caller reading from an older node can still parse this response type.
+    pub block: Option<Block>,
```
SuperFluffy: I am wondering if this is necessary - do we have any critical services that depend on this? I think we have not publicly communicated this anywhere.

crates/commonware-node/src/feed/actor.rs
```diff
@@ -68,127 +75,126 @@ impl<TContext: Spawner> Actor<TContext> {
...
-        let view = round.view().get();
-        let epoch = round.epoch().get();
-
-        CertifiedBlock {
-            epoch,
-            view,
-            height,
-            digest: digest.0,
-            certificate,
+        'subscribe: loop {
```
SuperFluffy: the nested loops are very hard to reason about - especially as you await the same channel both in the outside scope and the inside loop.

Can you flatten these?

crates/commonware-node/src/feed/actor.rs
```diff
@@ -68,127 +75,126 @@ impl<TContext: Spawner> Actor<TContext> {
...
             }
-            Activity::Nullification(nullification) => {
-                let _ = self.state.events_tx().send(Event::Nullified {
-                    epoch: nullification.round.epoch().get(),
-                    view: nullification.round.view().get(),
-                    seen: now_millis(),
-                });
-            }
-            _ => {}
+            _ => unreachable!("only notarizations and finalizations are inflight"),
```
SuperFluffy: IMO too harsh; just `_ => {}`.

crates/commonware-node/src/feed/actor.rs
```diff
@@ -68,127 +75,126 @@ impl<TContext: Spawner> Actor<TContext> {
...
-                            .is_some_and(|r| r <= round)
-                        {
-                            state.latest_notarized = None;
-                        }
+                        state.latest_notarized = None;
                     }
+                    state.latest_finalized = Some(certified);
                 }
             }
-            Activity::Nullification(nullification) => {
```
SuperFluffy, line 187: Makes sense - no point sending out nullifications for our purpose. Although it might be interesting to an outside observer.

crates/node/src/rpc/consensus/types.rs
```diff
@@ -12,11 +13,17 @@ use tokio::sync::broadcast;
...
+
+    /// Deprecated: To be removed after one release in favor of `block`.
```
SuperFluffy: Likewise, is anyone using these APIs? I was not considering them to be stable in any way.

If nobody except for us is using them, then IMO it's fine to remove (but double check).

crates/commonware-node/src/feed/actor.rs
```diff
@@ -68,127 +75,126 @@ impl<TContext: Spawner> Actor<TContext> {
...
+
+            'process: loop {
+                let (round, payload) = match &activity {
+                    Activity::Notarization(n) => (n.proposal.round, n.proposal.payload),
+                    Activity::Finalization(f) => (f.proposal.round, f.proposal.payload),
+                    _ => continue 'subscribe,
+                };
+
+                select!(
+                    // We should expect to see the block immediately as the block was processed by the
```
SuperFluffy: I think that's too optimistic. You are not in fact guaranteed to have observed all blocks, since finalizations and notarizations will come in independent of whether or not your node has seen the block.

crates/node/src/rpc/consensus/types.rs
hamdiallam: Just the follower engine but imo fine to break. Just requires a release before we can roll out the follower engine (or invert by subscribing to the EL then fetching certs)

crates/commonware-node/src/feed/actor.rs
hamdiallam: Yeah it was being sent out before but didn't make sense. Totally fine to add back in once there's an ask for it

crates/commonware-node/src/feed/actor.rs
hamdiallam: Okay with relaxing this. I will double check but I believe the marshal needs to first process the block prior to sending activity updates. But the feed actor is not downstream of the marshal so there's a small race

hamdiallam: Updated to process all pending subscriptions in order. A notarization that _could_ be blocking is pruned when a newer finalization comes in

crates/commonware-node/src/feed/actor.rs
```diff
@@ -68,126 +92,149 @@ impl<TContext: Spawner> Actor<TContext> {
...
+                            break; // exit
+                        };
+
+                        self.pending.insert(round, pending);
+                        self.subscribe(activity).await;
+                    },
+                );
+            } else {
+                let Some(activity) = self.receiver.next().await else {
+                    break; // exit
```
SuperFluffy: rm comment; maybr `break 'event_loop'` to make it clearer

crates/commonware-node/src/feed/actor.rs
```diff
@@ -1,34 +1,54 @@
...
+};
 
 /// Type alias for the activity type used by the feed actor.
 pub(super) type FeedActivity = Activity<Scheme<PublicKey, MinSig>, Digest>;
 
 /// Receiver for activity messages.
 pub(super) type Receiver = futures::channel::mpsc::UnboundedReceiver<FeedActivity>;
 
+/// A pending block subscription paired with its originating activity.
+struct PendingSubscription {
```
SuperFluffy, line 57: `impl Future for PendingSubscription` that returns the block on success (and the activity on both error and success, if you want).

crates/commonware-node/src/feed/actor.rs
```diff
@@ -93,38 +125,35 @@ impl<TContext: Spawner> Actor<TContext> {
...
+                        Ok((_, activity, block)) => self.handle_activity(activity, block),
+                        Err(err) => warn_span!("feed_actor").in_scope(||
+                            warn!(%err, "pending block subscription cancelled")
+                        ),
+                    }
+                },
+
+                activity = self.receiver.next() => {
+                    let Some(activity) = activity else {
+                        break;
```
SuperFluffy: Nice pattern: you could do something like:
```rust
let reason = loop { select!(
    msg = mailbox.next() => {
        let Some(msg) = msg {
            break Err(eyre!("mailbox closed"));
        }
    }
};

info_span!("feed_actor").in_scope(|| {
    match reason {
        Ok(reason) => info!(%reason, "shutting down"),
        Err(reason) => error!(%reason, "shutting down"),
    }
};
```

crates/commonware-node/src/feed/actor.rs
```diff
@@ -93,38 +125,35 @@ impl<TContext: Spawner> Actor<TContext> {
...
+            // We need a mutable reference to poll pending subscription. Thus if a new activity arrives,
+            // we also need to re-insert this popped subscription.
+            let mut oldest = OptionFuture::from(self.pending.pop_first().map(|(_, p)| p));
+
+            select!(
+                result = &mut oldest => {
+                    match result {
+                        Ok((_, activity, block)) => self.handle_activity(activity, block),
+                        Err(err) => warn_span!("feed_actor").in_scope(||
+                            warn!(%err, "pending block subscription cancelled")
```
SuperFluffy: Nit: Consistency (this crate uses `error` as a field name; also easier to grep)
```suggestion
                            warn!(%error, "did not get pending block")
```

Don't know how I feel about the repeated message.

crates/commonware-node/src/feed/actor.rs
```diff
@@ -93,38 +125,35 @@ impl<TContext: Spawner> Actor<TContext> {
...
-            } else {
-                let Some(activity) = self.receiver.next().await else {
-                    break; // exit
-                };
-
-                self.subscribe(activity).await;
-            }
+            // We need a mutable reference to poll pending subscription. Thus if a new activity arrives,
+            // we also need to re-insert this popped subscription.
+            let mut oldest = OptionFuture::from(self.pending.pop_first().map(|(_, p)| p));
```
SuperFluffy, line 140: Don't think the `OptionFuture` would work on just `self.pending.first_mut()`, right? Probably note because that'd borrow all of `self` mutably, which then doesn't work with the other arm. Bummer.

crates/commonware-node/src/feed/actor.rs
hamdiallam: Ah this is much nicer. TIL you can return values from the loop lol

crates/commonware-node/src/feed/actor.rs
hamdiallam: Yeah exactly

PR #2922, opened 2026-03-02
feat: add t3 to chainspec

fgimenez:
This PR adds the t3 fork the chainspec and genesis generator.


crates/chainspec/src/genesis/dev.json
```diff
@@ -26,6 +26,7 @@
     "t1bTime": 0,
     "t1cTime": 0,
     "t2Time": 0,
+    "t3Time": 0,
```
jenpaff: pls add to future time for now . This should be overwritten in a devnet.

fgimenez: done

PR #2925, opened 2026-03-02
docs: add TIP-1023 Tempo Address Format

dankrad:
## Summary

- Adds TIP-1023 defining a human-readable address format for Tempo using bech32m encoding (BIP-350)
- Addresses use the `tempo1` prefix (46 chars total), a version byte for future extensibility, and guaranteed detection of up to 4 character substitution errors
- Includes complete Python reference implementation and test vectors

## Test plan

- [ ] Verify the test vector: raw address `0x742d35Cc6634C0532925a3b844Bc9e7595f2bD28` encodes to `tempo1qp6z6dwvvc6vq5efyk3ms39une6etu4a9qtj2kk0`
- [ ] Verify round-trip: decode(encode(addr)) == addr for sample addresses
- [ ] Review bech32m checksum correctness against BIP-350 spec


tips/tip-1023.md
```diff
@@ -0,0 +1,264 @@
...
+│   HRP   │ 1 │           Base32-encoded data + checksum             │
+│ 5 chars │sep│                                                      │
+└─────────┴───┴──────────────────────────────────────────────────────┘
+                              │
+          ┌───────────────────┴──────────────────────────┐
+          │               Data (base32)                  │
+          ├─────────┬──────────────┬─────────────────────┤
+          │ Version │ Raw Address  │ Bech32m Checksum     │
+          │ 1 byte  │ 20 bytes     │ 6 chars              │
+          └─────────┴──────────────┴─────────────────────┘
```
dankrad:
```suggestion
          │ Version │ Raw Address  │ Bech32m Checksum    │
          │ 1 byte  │ 20 bytes     │ 6 chars             │
          └─────────┴──────────────┴─────────────────────┘
```

decofe (commented):
**Mixed-case handling does not match BIP-350**

BIP-350 (and BIP-173) require decoders to **reject mixed-case strings** — accepting either all-lowercase or all-uppercase, but never both in the same address. The current reference implementation just lowercases the input (`addr = addr.lower()`), which silently accepts mixed-case like `TeMpO1qp6z6...`.

The spec text in "Bech32m Encoding" and Invariant 7 says:

> Addresses MUST be valid regardless of case

This reads as "mixed case is fine," which contradicts BIP-350.

**Suggested fix:**

Spec text (Invariant 7 + §Bech32m Encoding item 4):
```
Decoders MUST accept either all-lowercase or all-uppercase encodings.
Decoders MUST reject mixed-case strings. Encoders SHOULD produce lowercase.
```

Reference `bech32m_decode`:
```python
def bech32m_decode(addr):
    has_lower = any("a" <= c <= "z" for c in addr)
    has_upper = any("A" <= c <= "Z" for c in addr)
    if has_lower and has_upper:
        raise ValueError("Mixed case")
    addr = addr.lower()
    ...
```

Everything else (test vectors, encoding, round-trips, checksums, substitution detection) verified correct ✅

PR #2928, opened 2026-03-02
Update TIP spec status

jenpaff:
fixed: status for each TIP , tip-1017 wasn't showing up in the docs

### DIFF
--- a/.github/workflows/pr-audit.yml
+++ b/.github/workflows/pr-audit.yml
@@ -1,30 +1,10 @@
-name: Pull request audit
+name: PR Audit
 
 on:
   pull_request:
     types: [labeled]
 
 jobs:
-  publish:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


tips/tip-1017.md
```diff
@@ -1,14 +1,24 @@
+---
+id: TIP-1017
+title: ValidatorConfig V2
+description: Second-generation validator configuration precompile with append-only semantics for historical validator set reconstruction.
+authors: Janis, Howy
```
malleshpai, line 5: I don't want to change this in case they don't want be doxxed or such, but ideally everyone puts their full names (or alias, i guess) as the author. having just first name looks a little informal?

malleshpai (approved):

tips/tip-1017.md
jenpaff: i'll clarify with everyone & would make this consistent across all specs, thanks for flagging

PR #2931, opened 2026-03-02
fix(consensus): race canonical stream against retry in read_validator_config_with_retry

### LINKED ISSUE #2929: perf(consensus): adaptive retry interval in read_validator_config_with_retry
When caught up to the tip, the 1s minimum retry in `read_validator_config_with_retry` (`crates/commonware-node/src/validators.rs:41`) causes unnecessary delays for validators. A block that arrives 37ms after the first attempt still costs a full 1s wait, which can cause view-2 proposers to miss their slot.

The 1s floor exists to avoid spamming during historical sync — that is still valid. But near the tip we should poll much more aggressively.

## Proposals

1. **Use `is_syncing` to pick the interval.** The node already calls `node.network.is_syncing()` in the retry log. Use it to branch:
   - Syncing → keep current backoff (`1s * attempt`, capped at 30s`)
   - At tip → poll at ~50ms, no exponential backoff (block is imminent)

2. **Subscribe to new-block notifications instead of polling.** When at tip, `tokio::select!` over a new-block channel + a safety timeout (~500ms). This gets the read within one event-loop tick of the block landing and eliminates busy-wait entirely.

3. **Hybrid.** Start with a short poll (50ms) and double on each attempt up to 1s while syncing, but cap at 50–100ms when at tip. Simple and no new plumbing.

Option 1 is the smallest change. Option 2 is the cleanest long-term but requires threading a block notification channel into the retry loop.

decofe:
Closes #2929

Subscribes to `canonical_state_stream()` before the retry loop and `tokio::select!`s it against the existing sleep. Whichever fires first wins — at the tip the stream wakes within one tick of the block landing instead of waiting the full 1s minimum retry. During historical sync the stream is silent (pipeline doesn't emit canonical notifications) so the existing backoff kicks in unchanged.

**Why:** At epoch boundaries, `read_validator_config_with_retry` blocks until the boundary block is available to read the next validator set. The 1s minimum retry means a validator assigned to propose on view 2 of the new epoch can miss its proposal window — the boundary block may land milliseconds after a failed read, but the node sleeps the full 1s before retrying. This is especially problematic because view 2 is the first proposal slot after the epoch transition, so any unnecessary delay directly risks a skipped view.

Prompted by: joshie

### DIFF
--- a/crates/commonware-node/src/validators.rs
+++ b/crates/commonware-node/src/validators.rs
@@ -22,7 +22,7 @@ use tempo_precompiles::{
     validator_config::{IValidatorConfig, ValidatorConfig},
 };
 
-use tracing::{Level, info, instrument, warn};
+use tracing::{Level, debug, info, instrument, warn};
 
 pub(crate) enum ReadTarget {
     AtLeast { height: Height },
@@ -81,8 +81,16 @@ pub(crate) async fn read_validator_config_with_retry(
             );
         });
         tokio::select! {
-            _ = canon_events.next() => {}
-            _ = context.sleep(retry_after) => {}
+            _ = canon_events.next() => {
+                tracing::info_span!("read_validator_config_with_retry").in_scope(|| {
+                    debug!("woke from canonical state notification");
+                });
+            }
+            _ = context.sleep(retry_after) => {
+                tracing::info_span!("read_validator_config_with_retry").in_scope(|| {
+                    debug!("woke from retry timeout");
+                });
+            }
         }
     }
 }


crates/commonware-node/src/validators.rs
```diff
@@ -1,5 +1,7 @@
 use std::{collections::HashMap, net::SocketAddr, time::Duration};
 
+use futures::StreamExt as _;
+
```
joshieDo:
```suggestion
```

crates/commonware-node/src/validators.rs
```diff
@@ -1,5 +1,6 @@
 use std::{collections::HashMap, net::SocketAddr, time::Duration};
```
joshieDo:
```suggestion
```

crates/commonware-node/src/validators.rs
```diff
@@ -78,7 +80,10 @@ pub(crate) async fn read_validator_config_with_retry(
                 "reading validator config from contract failed; will retry",
             );
         });
-        context.sleep(retry_after).await;
+        tokio::select! {
+            _ = canon_events.next() => {}
```
SuperFluffy: Maybe a debug log why this fired? If it's too noisy we can mute.

Ah, and please do it like above with an `info_span!` using the same name.

PR #2937, opened 2026-03-02
refactor: make --consensus.signing-key and --follow mutually exclusive

decofe:
Adds `conflicts_with = "follow"` to `--consensus.signing-key` and `--consensus.fee-recipient` so clap rejects passing both at parse time instead of silently ignoring the signing key.

Prompted by: zygis

### DIFF
--- a/crates/commonware-node/src/args.rs
+++ b/crates/commonware-node/src/args.rs
@@ -17,6 +17,7 @@ pub struct Args {
     #[arg(
         long = "consensus.signing-key",
         required_unless_present_any = ["follow", "dev"],
+        conflicts_with = "follow",
     )]
     signing_key: Option<PathBuf>,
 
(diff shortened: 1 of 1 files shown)


shekhirin (approved):

hamdiallam (approved):

crates/commonware-node/src/args.rs
```diff
@@ -17,6 +17,7 @@ pub struct Args {
     #[arg(
         long = "consensus.signing-key",
         required_unless_present_any = ["follow", "dev"],
+        conflicts_with = "follow",
```
SuperFluffy, line 20: Why is that necessary I wonder? We have `required_unless_present_any`, which implies `conflicts_with`

SuperFluffy: Yeah, it's a negative vs positive filter. Nevermind me.

PR #2938, opened 2026-03-02
chore(precompiles): validate policy type on T1C

0xrusowsky:
this PR validates policy types on T1C rather than waiting on T2

### DIFF
--- a/crates/precompiles/src/tip403_registry/mod.rs
+++ b/crates/precompiles/src/tip403_registry/mod.rs
@@ -611,6 +611,15 @@ impl AuthRole {
     }
 }
 
+/// Returns `true` if the error indicates a failed policy lookup — the policy type is invalid
+/// or the policy doesn't exist. Covers the legacy `Panic(UnderOverflow)` sentinel (pre-T1C)
+/// and the typed TIP403 errors (T1C+/T2+).
+pub fn is_policy_lookup_error(e: &TempoPrecompileError) -> bool {
+    *e == TempoPrecompileError::under_overflow()
+        || *e == TIP403RegistryError::invalid_policy_type().into()
+        || *e == TIP403RegistryError::policy_not_found().into()
+}
+
 trait PolicyTypeExt {
     fn ensure_is_simple(&self) -> Result<u8>;
 }
--- a/crates/precompiles/src/tip403_registry/mod.rs
+++ b/crates/precompiles/src/tip403_registry/mod.rs
@@ -111,18 +111,28 @@ impl PolicyData {
     /// Decodes the raw `policy_type` u8 into a [`PolicyType`] enum.
     ///
     /// * `WHITELIST` and `BLACKLIST` are valid on all specs.
-    /// * `COMPOUND` is only valid (spec T2+).
-    /// * invalid values return `TIP403RegistryError::invalid_policy_type` (spec T1C+), otherwise `TempoPrecompileError::under_overflow`
+    /// * `COMPOUND` is only valid post T2+.
+    /// * invalid types throw `TIP403RegistryError::invalid_policy_type` (T1C+), otherwise `TempoPrecompileError::under_overflow`
     fn policy_type(&self) -> Result<PolicyType> {
+        let res = self.policy_type.try_into();
         let spec = StorageCtx.spec();
 
-        match self.policy_type.try_into() {
-            Ok(PolicyType::COMPOUND) if spec.is_t2() => Ok(PolicyType::COMPOUND),
-            Ok(PolicyType::COMPOUND) | Err(_) if spec.is_t1c() => {
-                Err(TIP403RegistryError::invalid_policy_type().into())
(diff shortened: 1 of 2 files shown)


crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -607,20 +611,29 @@ impl AuthRole {
...
 }
 
+/// Returns `true` if the error indicates a failed policy lookup — the policy type is invalid
+/// or the policy doesn't exist. Covers the legacy `Panic(UnderOverflow)` sentinel (pre-T1C)
+/// and the typed TIP403 errors (T1C+/T2+).
+pub fn is_policy_lookup_error(e: &TempoPrecompileError) -> bool {
+    *e == TempoPrecompileError::under_overflow()
+        || *e == TIP403RegistryError::invalid_policy_type().into()
+        || *e == TIP403RegistryError::policy_not_found().into()
+}
```
0xKitsune, line 634: Do we need this helper? Post T1C, under_overflow will not indicate a policy lookup error. If this function is ever misused and an actual under_overflow error is passed in, this will incorrectly return true. I would prefer that we explictly handle cases rather than have this helper function.

crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -108,17 +108,21 @@ impl PolicyData {
...
-                TIP403RegistryError::invalid_policy_type().into()
-            } else {
-                TempoPrecompileError::under_overflow()
-            }),
+            Ok(PolicyType::COMPOUND) if spec.is_t2() => Ok(PolicyType::COMPOUND),
+            Ok(PolicyType::COMPOUND) | Err(_) if spec.is_t1c() => {
+                Err(TIP403RegistryError::invalid_policy_type().into())
+            }
+            Ok(PolicyType::COMPOUND) | Err(_) => Err(TempoPrecompileError::under_overflow()),
+            Ok(ty) => Ok(ty),
```
0xKitsune: Can we update this to be a bit more readable? Right now you have to read closely to know what this is doing across hardfork boundaries. Would prefer to simplify to something like the following which is a bit easier to reason about across forks (note the following may not be exact and needs a check).
```rust
        // T2+: WHITELIST, BLACKLIST, COMPOUND are valid
        if spec.is_t2() {
            ty.map_err(|_| TIP403RegistryError::invalid_policy_type().into())
        } else {
            match ty {
                // Pre T2: WHITELIST, BLACKLIST, are valid
                Ok(ty @ (PolicyType::WHITELIST | PolicyType::BLACKLIST)) => Ok(ty),
                _ => {
                    if spec.is_t1c() {
                        Err(TIP403RegistryError::invalid_policy_type().into())
                    } else {
                        Err(TempoPrecompileError::under_overflow())
                    }
                }
            }
        }

```

crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -108,17 +108,21 @@ impl PolicyData {
...
-            Ok(ty) if is_t2 || ty != PolicyType::COMPOUND => Ok(ty),
-            _ => Err(if is_t2 {
-                TIP403RegistryError::invalid_policy_type().into()
-            } else {
-                TempoPrecompileError::under_overflow()
-            }),
+            Ok(PolicyType::COMPOUND) if spec.is_t2() => Ok(PolicyType::COMPOUND),
+            Ok(PolicyType::COMPOUND) | Err(_) if spec.is_t1c() => {
+                Err(TIP403RegistryError::invalid_policy_type().into())
+            }
```
0xKitsune, line 135: The last arm in the match statement causes T2 errors to return `under_overflow()`.

0xKitsune (commented):
Left a few comments but in general the changes look straightforward. Where we need to be very precise is checking for any other instances where `under_overflow` is checked to indicate an invalid policy.

This PR fixes `cancel_stale_order`, but **any other code that matches on `under_overflow` to detect invalid policies will silently break at T1C.**

crates/precompiles/src/tip403_registry/mod.rs
0xrusowsky: addressed in [`2162844` (this PR)](https://github.com/tempoxyz/tempo/pull/2938/commits/21628444eb6e5cb93979c6238ac51873fe5e9a32)

crates/precompiles/src/tip403_registry/mod.rs
0xrusowsky: addressed in [`2162844` (this PR)](https://github.com/tempoxyz/tempo/pull/2938/commits/21628444eb6e5cb93979c6238ac51873fe5e9a32)

crates/precompiles/src/tip403_registry/mod.rs
0xrusowsky: addressed in [`2162844` (this PR)](https://github.com/tempoxyz/tempo/pull/2938/commits/21628444eb6e5cb93979c6238ac51873fe5e9a32)

PR #2939, opened 2026-03-02
feat(payload): separate builder interrupt and wait deadlines

shekhirin:
Supersedes https://github.com/tempoxyz/tempo/pull/2787

## Problem

Payload building is currently limited by a 500ms deadline. Builder can only enforce the time it spends on transaction execution, but there's always an additional overhead of state root calculation.

## Solution

Lower the time we give payload builder to execute transactions to 200ms, but wait until 450ms before returning the built block back to consensus to ensure the ~ stable block time at 500ms.

image

## Changes

* Interrupt payload building at T+200ms (this gives us a budget of additional 250ms for state root calc) https://github.com/tempoxyz/tempo/blob/f7a56d6c17da92f3870a83d1fc6b8c2f8a8b05db/crates/commonware-node/src/consensus/application/actor.rs#L583-L585
* Wait until T+450ms to return the block to consensus (this + some propagation overhead determines the block time) https://github.com/tempoxyz/tempo/blob/f7a56d6c17da92f3870a83d1fc6b8c2f8a8b05db/crates/commonware-node/src/consensus/application/actor.rs#L582 https://github.com/tempoxyz/tempo/blob/f7a56d6c17da92f3870a83d1fc6b8c2f8a8b05db/crates/commonware-node/src/consensus/application/actor.rs#L600


crates/commonware-node/src/args.rs
```diff
@@ -97,11 +97,18 @@ pub struct Args {
     )]
     pub inactive_views_until_leader_skip: u64,
 
+    /// The amount of time this node will wait before interrupting the payload builder.
```
SuperFluffy: Users will not necessarily know what a payload builder is.
```suggestion
    /// The amount of time to spend on executing transactions when preparing a proposal as a leader.
```

crates/commonware-node/src/args.rs
```diff
@@ -97,11 +97,18 @@ pub struct Args {
...
+    ///
+    /// NOTE: This only limits the time the builder spends on transaction execution, and does not
+    /// include the state root calculation time. For this reason, we keep it well below `consensus.time-to-build-proposal`.
+    #[arg(long = "consensus.time-to-build-interrupt", default_value = "200ms")]
```
SuperFluffy: "time to build interrupt" sounds weird - naming is hard.

Pretty verbose, but easier to understand?

```suggestion
    #[arg(long = "consensus.time-to-prepare-proposal-transactions", default_value = "200ms")]
```

crates/commonware-node/src/args.rs
```diff
@@ -97,11 +97,18 @@ pub struct Args {
...
+    pub time_to_build_interrupt: PositiveDuration,
+
     /// The amount of time this node will use to construct a block as a proposal.
     /// This value should be well below `consensus.wait-for-proposal` to account
     /// for the leader to enter the view, build and broadcast the proposal, and
     /// have the other peers receive the proposal.
-    #[arg(long = "consensus.time-to-build-proposal", default_value = "500ms")]
+    #[arg(long = "consensus.time-to-build-proposal", default_value = "450ms")]
```
SuperFluffy: Need to leave this now, but maybe we can alias this?

```suggestion
    #[arg(
        long = "consensus.minimum-time-before-propose",
        alias = "consensus.time-to-build-proposal",
        default_value = "450ms")]
```

Wonder if we can set some kind of deprecation notice?

crates/commonware-node/src/args.rs
```diff
@@ -97,11 +97,18 @@ pub struct Args {
...
     pub inactive_views_until_leader_skip: u64,
 
+    /// The amount of time this node will wait before interrupting the payload builder.
+    ///
+    /// NOTE: This only limits the time the builder spends on transaction execution, and does not
+    /// include the state root calculation time. For this reason, we keep it well below `consensus.time-to-build-proposal`.
+    #[arg(long = "consensus.time-to-build-interrupt", default_value = "200ms")]
+    pub time_to_build_interrupt: PositiveDuration,
+
     /// The amount of time this node will use to construct a block as a proposal.
```
SuperFluffy:
```suggestion
    /// The minimum amount of time this node waits before sending a proposal
    ///
    /// The intention is to keep block times stable even if there is low load on the network.
```

crates/commonware-node/src/consensus/application/mod.rs
```diff
@@ -55,8 +55,11 @@ pub(super) struct Config<TContext> {
     /// A handle to the subblocks service to get subblocks for proposals.
     pub(crate) subblocks: subblocks::Mailbox,
 
-    /// The minimum amount of time to wait before resolving a new payload from the builder
-    pub(super) new_payload_wait_time: Duration,
+    /// The minimum amount of time to wait before resolving a new payload from the builder.
+    pub(super) payload_resolve_time: Duration,
+
+    /// The minimum amount of time to wait before returning the built payload back to consensus for proposal.
+    pub(super) payload_return_time: Duration,
```
mattsse, line 62: why would we delay this?

shekhirin: We want to keep block time stable at around 500ms

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -593,6 +608,9 @@ impl Inner<Init> {
             .and_then(|rsp| rsp.map_err(Into::<eyre::Report>::into))
             .wrap_err_with(|| format!("failed getting payload for payload ID `{payload_id}`"))?;
 
+        // Keep waiting for `self.payload_return_time`, if there's anything left after building the block.
+        context.sleep_until(payload_return_time).await;
```
SuperFluffy, line 612: This reads like a proper sentence, which is nice.

PR #2940, opened 2026-03-02
feat: handle nodes losing DKG player state (and bump Commonware)

hamdiallam:
- Bumps Commonware to `240e0207ee10c3c37a42867ce4de97b581c06b32`
- Handle DKG finalization on invalid player state: Commonware now returns an error if player state is inconsistent (instead of panicking)
- CW now supports incompatible features. The right mode must be specified
- Broadcast engine subscribes to peer updates. Must be created after starting the peer manager actor
- `max_pending_acks` in the marshal being set to 1 preserves sequential delivery

### DIFF
--- a/crates/commonware-node/src/subblocks.rs
+++ b/crates/commonware-node/src/subblocks.rs
@@ -304,6 +304,9 @@ impl<TContext: Spawner + Metrics + Pacer> Actor<TContext> {
             .epoch();
 
         // Can't proceed without knowing a validator set for the current epoch.
+        //
+        // TODO(hamdi): When finalizing a boundary block, the scheme for the next epoch is not yet registered meaning
+        // we skip the subblock building task. This issue is scoped to the boundary and will be fixed.
         let Some(scheme) = self.scheme_provider.scoped(epoch_of_next_block) else {
             debug!(%epoch_of_next_block, "scheme not found for epoch");
             return;
--- a/crates/e2e/src/tests/subblocks.rs
+++ b/crates/e2e/src/tests/subblocks.rs
@@ -42,7 +42,7 @@ fn subblocks_are_included() {
 
         let setup = Setup::new()
             .how_many_signers(how_many_signers)
-            .epoch_length(10);
+            .epoch_length(100);
 
         // Setup and start all nodes.
         let (mut nodes, _execution_runtime) = setup_validators(&mut context, setup.clone()).await;
@@ -148,7 +148,7 @@ fn subblocks_are_included_with_failing_txs() {
 
         let setup = Setup::new()
             .how_many_signers(how_many_signers)
-            .epoch_length(10);
+            .epoch_length(100);
 
         // Setup and start all nodes.
         let (mut nodes, _execution_runtime) = setup_validators(&mut context, setup.clone()).await;
@@ -328,7 +328,7 @@ fn oversized_subblock_txs_are_removed() {
 
         let setup = Setup::new()
             .how_many_signers(how_many_signers)
-            .epoch_length(10);
+            .epoch_length(100);
 
         let (mut nodes, _execution_runtime) = setup_validators(&mut context, setup.clone()).await;
 


crates/commonware-node/src/consensus/application/actor.rs
SuperFluffy, line 1: undid some unnecessary changes

crates/commonware-node/src/consensus/engine.rs
```diff
@@ -41,25 +41,25 @@ use crate::{
...
-
 // A bunch of constants to configure commonwarexyz singletons and copied over form alto.
 
 /// To better support peers near tip during network instability, we multiply
 /// the consensus activity timeout by this factor.
 const SYNCER_ACTIVITY_TIMEOUT_MULTIPLIER: u64 = 10;
-const PRUNABLE_ITEMS_PER_SECTION: NonZeroU64 = NonZeroU64::new(4_096).expect("value is not zero");
-const IMMUTABLE_ITEMS_PER_SECTION: NonZeroU64 =
-    NonZeroU64::new(262_144).expect("value is not zero");
+const PRUNABLE_ITEMS_PER_SECTION: NonZeroU64 = NZU64!(4_096);
```
SuperFluffy, line 51: Not really necessary, but welcome cleanup.

crates/commonware-node/src/consensus/engine.rs
```diff
@@ -574,7 +566,23 @@ where
...
+        // for peer set updates. Thus we create the engine here rather than try_init().
+        let (broadcast_engine, broadcast_mailbox) = buffered::Engine::new(
+            self.context.with_label("broadcast"),
+            buffered::Config {
+                public_key: self.broadcast_public_key,
+                mailbox_size: self.broadcast_mailbox_size,
+                deque_size: self.broadcast_deque_size,
+                priority: true,
+                codec_config: (),
+                peer_set_subscription: self.peer_manager_mailbox.clone().subscribe().await,
```
SuperFluffy: Annoying; `subscribe()` definitely suggests an interaction with a running actor, so it would make sense to move this to `broadcast_engine.start`, not config.

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -721,27 +720,42 @@ where
...
-                match player_state.finalize(logs, &Sequential) {
+                match player.finalize(logs.clone(), &Sequential) {
                     Ok((new_output, new_share)) => {
                         info!("local DKG ceremony was a success");
-                        (new_output, state::ShareState::Plaintext(Some(new_share)))
+                        Some((new_output, state::ShareState::Plaintext(Some(new_share))))
+                    }
+                    Err(
+                        reason
+                        @ commonware_cryptography::bls12381::dkg::Error::MissingPlayerDealing,
```
SuperFluffy, line 732: This is the new error commonware introduced so that we can fallback to finalizing the DKG round as an observer.

This also forces us to fallback to finalizing the DKG round as an observer because otherwise we would end the round with the wrong result (this should technically still be handled, but it's better to do this correctly).

crates/commonware-node/src/consensus/engine.rs
SuperFluffy: This hangs the node - because the peer manager is not yet ready, this subscription never returns.

crates/commonware-node/src/consensus/engine.rs
hamdiallam: Yeah exactly, `MAX_PENDING_ACKS` is an added constant and took the opportunity to clean the rest up

hamdiallam (commented):
Looks good! Ty for the fix

crates/commonware-node/src/peer_manager/actor.rs
```diff
@@ -115,7 +136,7 @@ where
     /// than the execution layer's best block, because the execution layer may
     /// be behind and missing boundary blocks.
     #[instrument(skip_all, err)]
```
SuperFluffy: Let's add the `attempt` into instrument:

```suggestion
    #[instrument(skip_all, fields(attempt), err)]
```

crates/commonware-node/src/consensus/engine.rs
```diff
@@ -291,12 +283,24 @@ where
             },
         );
 
+        let (broadcast, broadcast_mailbox) = buffered::Engine::new(
+            context.with_label("broadcast"),
+            buffered::Config {
+                public_key: self.signer.public_key(),
+                mailbox_size: self.mailbox_size,
+                deque_size: self.deque_size,
+                peer_provider: peer_manager_mailbox.clone(),
```
SuperFluffy, line 292: So much better.

crates/commonware-node/src/peer_manager/actor.rs
```diff
@@ -81,25 +83,44 @@ where
     }
 
     async fn run(mut self) {
```
SuperFluffy, line 85: Technically none of this is necessary any longer, but it's a better pattern that we should follow everywhere (i.e. no work before entering the event loop).

hamdiallam: Initially I actually undid these changes to reduce the number of changes and the tests started failing again

SuperFluffy: good to know (and makes sense - the buffer/broadcast actor is probably trying to get a subscription before entering its event loop, meaning we just moved the deadlock to another place)

PR #2944, opened 2026-03-03
test: add gas snapshots for tempo txs

0xrusowsky:
Closes: `CHAIN-615`

this PR builds on top of #2810 to create snapshot test for gas estimations, while extending the matrix to more test cases

### DIFF
--- a/crates/node/tests/it/tempo_transaction/mod.rs
+++ b/crates/node/tests/it/tempo_transaction/mod.rs
@@ -103,7 +103,7 @@ async fn test_gas_estimation_snapshots() -> eyre::Result<()> {
         }
     }
 
-    let mut localnet = helpers::Localnet::new().await?;
+    let mut localnet = local::Localnet::new().await?;
     let results = localnet.run_estimate_gas_matrix().await?;
     let gas_estimation: indexmap::IndexMap<String, u64> = results.into_iter().collect();
 


crates/node/tests/it/tempo_transaction/mod.rs
```diff
@@ -83,7 +82,22 @@ async fn test_matrices_local() -> eyre::Result<()> {
     run_all_matrices(&mut helpers::Localnet::new().await?).await
 }
 
+#[tokio::test(flavor = "multi_thread")]
+async fn test_gas_estimation_snapshots() -> eyre::Result<()> {
+    let mut localnet = helpers::Localnet::new().await?;
+    let results = localnet.run_estimate_gas_matrix().await?;
+    let mut gas_estimation: indexmap::IndexMap<String, u64> = results.into_iter().collect();
+    gas_estimation.sort_by(|_, a, _, b| a.cmp(b));
```
rakita: Nit, should it be sorted by name, it is now sorted by gas amount? Ignore if this is okay

0xrusowsky: the idea was to easily compare gas costs for each scenario

0xrusowsky: do u think it would be more comprehensive sorting by name? @rakita

crates/node/tests/it/tempo_transaction/types.rs
```diff
@@ -710,28 +707,47 @@ pub(crate) fn key_type_to_signature_type(key_type: KeyType) -> SignatureType {
...
 
+#[derive(Clone)]
+pub(crate) enum GasPayload {
+    /// Single no-op call to a random address (default).
+    NoOp,
+    /// TIP-20 transfer call.
+    Transfer,
+    /// Contract creation with fixed initcode.
+    ContractCreation,
+    /// N no-op calls.
```
legion2002: Can we have transfers instead of noops for batch calls too

0xrusowsky: done in [`ab542b5` (this PR)](https://github.com/tempoxyz/tempo/pull/2944/commits/ab542b517178e4336f8a3920516f1eff97f4bfe2)

crates/node/tests/it/tempo_transaction/mod.rs
0xrusowsky: did a hybrid so that tests are sorted by group and gas costs, so that its even easier to review

PR #2947, opened 2026-03-03
docs(tip-1028): Address-Level Receive Policies

malleshpai:
Extends TIP-403 with token sets and address-level receive policies. Addresses review feedback: renamed to TIP-1025, Account→Address-Level, added bit flag for zero-ID semantics, orchestrator use cases, built-in policy handling, gas cost analysis.

### DIFF
--- a/tips/tip-1028.md
+++ b/tips/tip-1028.md
@@ -1,5 +1,5 @@
 ---
-id: TIP-1025
+id: TIP-1028
 title: Address-Level Receive Policies
 description: Extends TIP-403 with token sets and address-level receive policies, allowing individual addresses to control who can send to them and which tokens they can receive.
 authors: Mallesh Pai
@@ -8,7 +8,7 @@ related: TIP-403, TIP-1015, TIP-20
 protocolVersion: TBD
 ---
 
-# TIP-1025: Address-Level Receive Policies
+# TIP-1028: Address-Level Receive Policies
 
 ## Abstract
 


tips/tip-1023.md
```diff
@@ -0,0 +1,643 @@
...
+protocolVersion: TBD
+---
+
+# TIP-1023: Account-Level Receive Policies
+
+## Abstract
+
+This TIP extends the TIP-403 transfer policy system in two ways:
+
+1. **Token Sets** — a new TIP-403 primitive for managing sets of token addresses, distinct from policies (which manage sets of counterparty addresses).
```
danrobinson:
```suggestion
1. **Token Sets** — a new TIP-403 primitive for managing sets of token addresses, distinct from policies (which manage sets of addresses).
```

tips/tip-1023.md
```diff
@@ -0,0 +1,643 @@
...
+status: Draft
+related: TIP-403, TIP-1015, TIP-20
+protocolVersion: TBD
+---
+
+# TIP-1023: Account-Level Receive Policies
+
+## Abstract
+
+This TIP extends the TIP-403 transfer policy system in two ways:
```
danrobinson: I think three ways—token sets, address policies for what addresses they can receive from, and address policies for what tokens they can receive

tips/tip-1023.md
```diff
@@ -0,0 +1,643 @@
+---
+id: TIP-1023
+title: Account-Level Receive Policies
```
danrobinson: I think it's "Address-Level"—it's keyed by addresses, and contracts can set it too so it's not just EOAs

tips/tip-1023.md
```diff
@@ -0,0 +1,643 @@
...
+
+1. **Token Sets** — a new TIP-403 primitive for managing sets of token addresses, distinct from policies (which manage sets of counterparty addresses).
+2. **Account-Level Receive Policies** — the ability for individual accounts to control who can send to them and which tokens they can receive.
+
+Currently, TIP-403 policies are set at the token level — each TIP-20 token has a single `transferPolicyId` that governs all transfers. This proposal adds the ability for individual accounts to set their own controls across two dimensions:
+
+1. **Receive policy** (counterparty filter): Control who can send to this account
+2. **Token set** (token filter): Control which tokens this account can receive
+
+This enables regulated entities (banks, exchanges, etc.) to enforce inbound compliance controls at the account level, independent of and in addition to token-level policies.
```
danrobinson: Compliance is one reason, but it also can be easier for orchestration (for example if someone hasn't built in the capability to handle unknown tokens)

tips/tip-1023.md
```diff
@@ -0,0 +1,643 @@
...
+
+Outbound send restrictions are explicitly out of scope. Controlling what an account can send is an account-level choice best handled by mechanisms like multisigs, offchain policy engines, or smart contract wallets — not protocol-level transfer hooks.
+
+## Motivation
+
+Regulated entities operating on Tempo need the ability to control *inbound* transactions, regardless of what policies the token issuer has set. For example:
```
danrobinson: and orchestrators

tips/tip-1028.md
```diff
@@ -0,0 +1,643 @@
...
+Outbound send restrictions are explicitly out of scope. Controlling what an account can send is an account-level choice best handled by mechanisms like multisigs, offchain policy engines, or smart contract wallets — not protocol-level transfer hooks.
+
+## Motivation
+
+Regulated entities operating on Tempo need the ability to control *inbound* transactions, regardless of what policies the token issuer has set. For example:
+
+- A bank may want to only receive funds from KYC'd addresses (whitelist on receives)
+- A custodian may require that all incoming transfers come from approved counterparties
+- **A regulated entity may only want to hold approved stablecoins** (whitelist on tokens, e.g. a 'MiCA'-only list)
+- **An institution may need to block specific tokens** known to be associated with illicit activity or memecoins (blacklist on tokens)
```
danrobinson, line 37: Add example for orchestrators

tips/tip-1028.md
```diff
@@ -0,0 +1,643 @@
...
+# Specification
+
+## Part 1: Token Sets (TIP-403 Extension)
+
+Token sets are a new primitive in the TIP-403 Registry for managing sets of token addresses.
+
+### Storage
+
+```solidity
+uint64 public tokenSetIdCounter = 2; // Skip 0 (always-reject) and 1 (always-allow)
```
danrobinson, line 56: Possibly start with 1 rather than 0

danrobinson: Maybe not needed if the slot has other stuff in it

danrobinson: Well if 0 is always-reject, then setting no policy would mean always-reject—maybe should swap these?

tips/tip-1028.md
```diff
@@ -0,0 +1,643 @@
...
+/// @param setType The type of token set (WHITELIST or BLACKLIST, not COMPOUND)
+/// @return newTokenSetId The ID of the newly created token set
+function createTokenSet(address admin, PolicyType setType) 
+    external returns (uint64 newTokenSetId);
+
+/// @notice Creates a new token set with initial token members
+/// @param admin The address authorized to modify this token set
+/// @param setType The type of token set (WHITELIST or BLACKLIST, not COMPOUND)
+/// @param tokens The initial token addresses to add
+/// @return newTokenSetId The ID of the newly created token set
```
danrobinson, line 85: One thing that will come up with this is that there'll be a race condition—if you create one of these in a forge script, it might get the wrong ID when it simulates it. We hit this with tokens

tips/tip-1028.md
```diff
@@ -0,0 +1,643 @@
...
+function createTokenSetWithTokens(
+    address admin, 
+    PolicyType setType, 
+    address[] calldata tokens
+) external returns (uint64 newTokenSetId);
+
+/// @notice Updates the admin address for a token set
```
danrobinson, line 92: Can only be called by current admin

tips/tip-1028.md
danrobinson: Nevermind—must keep it consistent with address policies

tips/tip-1023.md
```diff
@@ -0,0 +1,643 @@
...
+
+| Bits (inclusive) | Size | Field | Description |
+|------------------|------|-------|-------------|
+| 0–63 | 64 bits | `receivePolicyId` | TIP-403 policy checked when this account is the receiver |
+| 64–71 | 8 bits | `receivePolicyType` | Type of receive policy (0 = whitelist, 1 = blacklist) |
+| 72–135 | 64 bits | `tokenSetId` | TIP-403 token set checked on tokens being received |
+| 136–143 | 8 bits | `tokenSetType` | Type of token set (0 = whitelist, 1 = blacklist) |
+| 144–255 | 112 bits | Reserved | For future use (must be 0) |
+
+A policy/token set ID of `0` means "no control" (always authorized at the account level). This differs from the token-level TIP-403 semantics where ID 0 is "always-reject"; for account-level controls, ID 0 means "no account-level control set" (defer to token policy only). When an ID is 0, the corresponding type field is ignored and SHOULD be set to 0.
```
danrobinson: This contradicts what it says above.

I think instead, offset everything by 1 and add a bit flag for whether there are address-level policies set on this address.

tips/tip-1028.md
malleshpai: done

tips/tip-1028.md
malleshpai: added a comment on this above

tips/tip-1028.md
malleshpai: added

tips/tip-1025.md
```diff
@@ -0,0 +1,724 @@
...
+This TIP extends the TIP-403 transfer policy system in three ways:
+
+1. **Token Sets** — a new TIP-403 primitive for managing sets of token addresses, distinct from policies (which manage sets of addresses).
+2. **Address-level receive policies** (counterparty filter) — the ability for individual addresses to control who can send to them.
+3. **Address-level token filtering** (token filter) — the ability for individual addresses to control which tokens they can receive.
+
+Currently, TIP-403 policies are set at the token level — each TIP-20 token has a single `transferPolicyId` that governs all transfers. This proposal adds the ability for individual addresses (EOAs or contracts) to set their own inbound controls across two dimensions:
+
+1. **Receive policy** (counterparty filter): Control who can send to this address
+2. **Token set** (token filter): Control which tokens this address can receive
```
samczsun: would be nice to mirror the language above ("sent to address")

tips/tip-1028.md
```diff
@@ -0,0 +1,724 @@
...
+
+All functions that call `ensureTransferAuthorized(from, to)` automatically inherit address-level checks:
+
+- `transfer(address to, uint256 amount)`
+- `transferFrom(address from, address to, uint256 amount)`
+- `transferWithMemo(address to, uint256 amount, bytes32 memo)`
+- `transferFromWithMemo(address from, address to, uint256 amount, bytes32 memo)`
+- `systemTransferFrom(address from, address to, uint256 amount)`
+
+### Mint Behavior
```
samczsun, line 517: this should check the issuer against the policy

PR #2948, opened 2026-03-03
fix(commonware-node): avoid panic when executor mailbox is closed

dev3Nigerian:
Stops panicking in the executor ingress report path when the mailbox receiver is closed. Failed sends are now logged at warn level and the activity is dropped, which keeps shutdown and teardown paths non-fatal.

### DIFF
--- a/crates/commonware-node/src/executor/ingress.rs
+++ b/crates/commonware-node/src/executor/ingress.rs
@@ -4,7 +4,7 @@ use futures::{
     SinkExt as _,
     channel::{mpsc, oneshot},
 };
-use tracing::Span;
+use tracing::{Span, warn};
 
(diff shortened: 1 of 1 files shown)


crates/commonware-node/src/executor/ingress.rs
```diff
@@ -79,9 +79,8 @@ impl Reporter for Mailbox {
     type Activity = Update<Block>;
 
     async fn report(&mut self, update: Self::Activity) {
```
SuperFluffy, line 81: This will panic anyway after we wail to ACK the finalized block.

Also, there is no mechanism to communicate that the mailbox is closed given commonware's current implementation, and they themselves rely on panics to shutdown the consensus stack.

We will stick to the same pattern for now.

PR #2952, opened 2026-03-03
fix(consensus): account for elapsed time in builder timeout

decofe:
The payload builder timeout sleeps for a fixed duration after `send_new_payload`, regardless of how much time was already spent in the proposal flow (getting parent, verifying, canonicalizing head, querying DKG). When the proposer starts late — e.g. because it was still processing a previous block — all that pre-build work eats into the view's time budget but the builder sleep doesn't account for it.

Tracks elapsed time from the start of `propose()` and subtracts it from the configured builder timeout via `saturating_sub`, so the total proposal time stays within the intended budget. Adds `elapsed_ms` and `remaining_ms` to the existing log line for observability.

Co-Authored-By: joshieDo <93316087+joshieDo@users.noreply.github.com>

Prompted by: joshie


crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -574,13 +576,19 @@ impl Inner<Init> {
...
 
+        let elapsed = propose_start.elapsed();
+        let remaining_resolve = self.payload_resolve_time.saturating_sub(elapsed);
+        let remaining_return = self.payload_return_time.saturating_sub(elapsed);
         debug!(
             resolve_time_ms = self.payload_resolve_time.as_millis(),
             return_time_ms = self.payload_return_time.as_millis(),
+            elapsed_ms = elapsed.as_millis(),
+            remaining_resolve_ms = remaining_resolve.as_millis(),
+            remaining_return_ms = remaining_return.as_millis(),
```
joshieDo: just set resolve_time_ms to remaining and the other too

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -574,24 +579,29 @@ impl Inner<Init> {
...
 
+        let elapsed = propose_start.elapsed();
+        let remaining_resolve = self.payload_resolve_time.saturating_sub(elapsed);
+        let remaining_return = self.payload_return_time.saturating_sub(elapsed);
         debug!(
-            resolve_time_ms = self.payload_resolve_time.as_millis(),
-            return_time_ms = self.payload_return_time.as_millis(),
+            elapsed_ms = elapsed.as_millis(),
+            resolve_time_ms = remaining_resolve.as_millis(),
+            return_time_ms = remaining_return.as_millis(),
```
SuperFluffy: Can use jiff / `tempo-telemetry-util::display_duration` to format these.

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -574,24 +579,29 @@ impl Inner<Init> {
...
             "sleeping before payload builder resolving"
         );
 
         // Start the timer for `self.payload_return_time`
```
SuperFluffy: Comment no longer up to date.

crates/commonware-node/src/consensus/application/actor.rs
decofe: Done — switched to `tempo_telemetry_util::display_duration` for all three fields.

crates/commonware-node/src/consensus/application/actor.rs
decofe: Updated — comment now references `remaining_return` instead of `self.payload_return_time`.

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -574,24 +579,29 @@ impl Inner<Init> {
...
 
+        let elapsed = propose_start.elapsed();
+        let remaining_resolve = self.payload_resolve_time.saturating_sub(elapsed);
+        let remaining_return = self.payload_return_time.saturating_sub(elapsed);
         debug!(
-            resolve_time_ms = self.payload_resolve_time.as_millis(),
-            return_time_ms = self.payload_return_time.as_millis(),
+            elapsed = %tempo_telemetry_util::display_duration(elapsed),
+            resolve_time = %tempo_telemetry_util::display_duration(remaining_resolve),
+            return_time = %tempo_telemetry_util::display_duration(remaining_return),
```
SuperFluffy: Could just import `display_duration` into the scope by now. ^.^

PR #2954, opened 2026-03-03
feat: add feeRecipient to tip, ref impl

SuperFluffy:
Adds the `feeRecipient` field to val config v2 `Validator` entries. Also introduces the `setFeeRecipient` function to change the fee recipient, and `FeeRecipientUpdated` event.

Unclear how this should be added to the spec:

1. this is not expected to be activated until T3. This is in preparation for T3 but won't be implemented in T2.
2. during V1 -> V2 migration, the value is expected to be set to `0x0`.

### DIFF
--- a/tips/ref-impls/src/ValidatorConfigV2.sol
+++ b/tips/ref-impls/src/ValidatorConfigV2.sol
@@ -94,7 +94,9 @@ contract ValidatorConfigV2 is IValidatorConfigV2 {
         _validateAddParams(validatorAddress, publicKey, ingress, egress);
 
         bytes32 message = keccak256(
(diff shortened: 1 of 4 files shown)


tips/tip-1017.md
```diff
@@ -310,12 +323,12 @@ This read switch is implemented in CL logic. V2 does not proxy reads to V1.
 
 ### DKG Player Selection
 
-The consensus layer determines DKG players for epoch `E+1` by reading state at `boundary(E)` and filtering:
```
SuperFluffy, line 313: You can move this over to the simplify PR

howydev (approved):

PR #2960, opened 2026-03-04
fix(rpc): create hardfork-aware mock keychain signatures for gas estimation

decofe:
`create_mock_tempo_signature` always created V2 keychain signatures (`KeychainSignature::new` defaults to V2), but the handler rejects V2 pre-T1C. This broke `eth_estimateGas` for any keychain transaction before T1C activation:

```
V2 keychain signature (type 0x04) is not valid before T1C activation
```

Now creates V1 pre-T1C and V2 post-T1C by passing the active hardfork spec to the mock signature builder. Gas cost is identical for both versions.

Prompted by: daniel


mattsse (commented):
Unclear why this needs eip3607 check here

PR #2966, opened 2026-03-04
test(evm): add block executor unit tests for boundary conditions

decofe:
Adds unit tests to `crates/evm/src/block.rs`:

**validate_system_tx:**
- Rejects input shorter than 32 bytes
- Rejects input of exactly 32 bytes (no metadata prefix)
- Rejects wrong block number in input suffix

**validate_shared_gas:**
- Accepts reserved gas exactly equal to gas-per-subblock limit
- Accepts incentive gas exactly equal to used amount
- Rejects incentive gas exceeding used by 1
- Verifies incentive gas accumulation across multiple validators

**validate_tx:**
- Non-shared gas boundary: at limit → NonShared, at limit+1 → GasIncentive
- Non-payment gas boundary: at limit → NonShared, at limit+1 → GasIncentive

**commit_transaction:**
- Verifies non_shared_gas_left decremented after commit
- Verifies payment txs skip non_payment_gas_left decrement
- Verifies incentive gas section tracking

Prompted by: George


crates/evm/src/block.rs
```diff
@@ -1133,6 +1133,233 @@ mod tests {
...
+        let mut db = State::builder().with_bundle_update().build();
+        let executor = TestExecutorBuilder::default().build(&mut db, &chainspec);
+        let result = executor.validate_system_tx(&create_system_tx(
+            chain_id,
+            Bytes::copy_from_slice(&[0u8; 31]),
+        ));
+        assert_eq!(
+            result.unwrap_err().to_string(),
+            "invalid subblocks metadata system transaction"
+        );
```
0xrusowsky: in all error asserts of the PR we should match vs the types rather than using strings

crates/evm/src/block.rs
```diff
@@ -1133,6 +1133,233 @@ mod tests {
...
+            .unwrap();
+        // non_shared decremented, non_payment unchanged (30M > 19_979_000 so non_shared is the limit)
+        assert_eq!(
+            executor.validate_tx(&tx, 19_979_000).unwrap(),
+            BlockSection::NonShared
+        );
+        assert_eq!(
+            executor.validate_tx(&tx, 19_979_001).unwrap(),
+            BlockSection::GasIncentive
+        );
```
0xrusowsky: this asserts the same boundaries as the non-payment case above (`non_shared_gas_left` is the binding constraint in both, so the payment flag has no observable effect)

to verify that payment txs preserve `non_payment_gas_left`, u'd need to exhaust `non_shared` to show that payment txs still execute while non-payment txs don't

0xrusowsky (changes requested):

crates/evm/src/block.rs
```diff
@@ -1146,10 +1162,10 @@ mod tests {
...
             Bytes::copy_from_slice(&[0u8; 31]),
         ));
-        assert_eq!(
-            result.unwrap_err().to_string(),
-            "invalid subblocks metadata system transaction"
-        );
+        assert!(matches!(
+            result.unwrap_err(),
+            BlockValidationError::Other(_)
+        ));
```
0xrusowsky: let's use a helper to match against both, enum variant + msg

```rs
fn assert_block_validation_err(result: Result<_, BlockValidationError>, expected: &str) {
    match result {
        Err(BlockValidationError::Other(msg)) => assert!(
            msg.contains(expected),
            "expected error containing '{expected}', got: {msg}"
        ),
        other => panic!("expected BlockValidationError::Other, got: {other:?}"),
    }
}
```

0xrusowsky (approved):

PR #2967, opened 2026-03-04
test(ci): add pre-hardfork test job to catch regressions in new-binary-old-fork window

fgimenez:
Adds a CI job that runs integration tests with the latest hardforks (T1C, T2) deactivated, catching regressions in the "new binary, old fork" window before a hardfork activates on the network.

- `TEMPO_PRE_HARDFORK_TEST=1` env var sets `t1cTime`/`t2Time` to `u64::MAX` in the test genesis, simulating a pre-hardfork chain
- `skip_pre_hardfork!()` macro skips 19 tests that exercise features introduced by T1C/T2 (V2 keychain signatures, access keys, spending limits, key expiry, TIP-1007 getFeeToken, TIP-403 whitelist enforcement, etc.) — these are expected to fail when those forks are deactivated
- New `test-pre-hardfork` CI job runs the remaining integration tests against the old fork rules

### DIFF
--- a/crates/node/tests/it/tip_fee_amm.rs
+++ b/crates/node/tests/it/tip_fee_amm.rs
@@ -577,6 +577,7 @@ async fn test_burn_liquidity_partial() -> eyre::Result<()> {
 
 #[tokio::test(flavor = "multi_thread")]
 async fn test_cant_burn_required_liquidity() -> eyre::Result<()> {
+    skip_pre_hardfork!();
     reth_tracing::init_test_tracing();
 
     let setup = TestNodeBuilder::new().build_http_only().await?;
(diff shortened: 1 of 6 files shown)


crates/node/tests/it/tip_fee_amm.rs
```diff
@@ -577,6 +577,7 @@ async fn test_burn_liquidity_partial() -> eyre::Result<()> {
 
 #[tokio::test(flavor = "multi_thread")]
 async fn test_cant_burn_required_liquidity() -> eyre::Result<()> {
+    skip_pre_hardfork!();
```
mattsse: pre which hardfork?

I find this a bit confusing
a) why are we skipping this 
b) which hardfork? what happens if there's another one?

fgimenez: > a) why are we skipping this

yep, these are skipped bc they test new functionality

> b) which hardfork? what happens if there's another one?

added changes to specify which hardfork to skip here https://github.com/tempoxyz/tempo/pull/2967/commits/c7a7460887421a3dce88b28a38eaf2a8c2248d7e good call!

PR #2969, opened 2026-03-04
fix: set jemalloc malloc_conf at compile time when jemalloc-prof is enabled

decofe:
Exports `_rjem_malloc_conf` with `prof:true,prof_active:true,lg_prof_sample:19` when `jemalloc-prof` feature is enabled, matching upstream reth. Without this, the `/debug/pprof/heap` endpoint requires `_RJEM_MALLOC_CONF=prof:true` env var at runtime because tikv-jemallocator uses prefixed symbols.

We have the same in Reth: https://github.com/paradigmxyz/reth/blob/dbded15aed9a113ae0a6ea6c6475da047ed1aa9f/bin/reth/src/main.rs#L6-L8

Prompted by: alexey

### DIFF
--- a/bin/tempo/src/main.rs
+++ b/bin/tempo/src/main.rs
@@ -18,6 +18,13 @@
 #[global_allocator]
 static ALLOC: reth_cli_util::allocator::Allocator = reth_cli_util::allocator::new_allocator();
 
+/// Compile-time jemalloc configuration for heap profiling.
+///
+/// tikv-jemallocator uses prefixed symbols, so the runtime `MALLOC_CONF` env var is ignored.
+/// This exported symbol is read by jemalloc at init time to enable profiling unconditionally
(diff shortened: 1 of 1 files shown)


bin/tempo/src/main.rs
```diff
@@ -18,6 +18,10 @@
 #[global_allocator]
 static ALLOC: reth_cli_util::allocator::Allocator = reth_cli_util::allocator::new_allocator();
 
+#[cfg(all(feature = "jemalloc-prof", unix))]
+#[unsafe(export_name = "_rjem_malloc_conf")]
+static MALLOC_CONF: &[u8] = b"prof:true,prof_active:true,lg_prof_sample:19\0";
```
mattsse, line 30: this is a private static, can we add a note here why this exists?

mattsse (approved):

PR #2973, opened 2026-03-04
fix(node): make `eth_estimateGas` hardfork-aware

0xrusowsky:
supersedes https://github.com/tempoxyz/tempo/pull/2960 + builds on top of https://github.com/tempoxyz/tempo/pull/2971

## Motivation

`eth_estimateGas` fails for keychain (access-key) txs when `T1C` is not yet active on the target network because `create_mock_tempo_signature` always produces a keychain V2 sig —which are rejected pre-T1C.

## Solution

make `create_mock_tempo_sig` fork-aware by extract the active `TempoHardfork` from the `CfgEnv` spec.

### DIFF
--- a/crates/alloy/src/rpc/compat.rs
+++ b/crates/alloy/src/rpc/compat.rs
@@ -395,6 +395,20 @@ mod tests {
     }
 
     #[test]
+    fn test_is_t1c_active() {
+        // pre-T1C (false)
+        assert!(!is_t1c_active(&TempoHardfork::Genesis));
+        assert!(!is_t1c_active(&TempoHardfork::T0));
+        assert!(!is_t1c_active(&TempoHardfork::T1));
+        assert!(!is_t1c_active(&TempoHardfork::T1A));
+        assert!(!is_t1c_active(&TempoHardfork::T1B));
+
+        // T1C and later (true)
+        assert!(is_t1c_active(&TempoHardfork::T1C));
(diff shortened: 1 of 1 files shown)


crates/alloy/src/rpc/compat.rs
```diff
@@ -184,25 +190,49 @@ impl TryIntoTxEnv<TempoTxEnv, TempoBlockEnv> for TempoTransactionRequest {
...
+/// Defaults to `true` (latest behavior) if the type doesn't match.
+///
+/// NOTE: the `unsafe` block will be removed with the reth release of: <https://github.com/alloy-rs/evm/pull/306>
+fn is_t1c_active<Spec>(spec: &Spec) -> bool {
+    if std::mem::size_of::<Spec>() != std::mem::size_of::<TempoHardfork>() {
+        return true;
+    }
+    // SAFETY: reading a single u8 is always valid for any type with size >= 1.
+    let discriminant = unsafe { std::ptr::read(spec as *const Spec as *const u8) };
+    discriminant >= TempoHardfork::T1C as u8
```
mattsse, line 207: this works as a workaround

0xrusowsky: added test in [`54606b3` (this PR)](https://github.com/tempoxyz/tempo/pull/2973/commits/54606b3364827dffa6a3584db8db98369105ec09)

PR #2982, opened 2026-03-04
docs(tip-1024): unify quote and swap fill logic

danrobinson:
Routes quote through the same fill logic as swap in a read-only simulation path to eliminate quote/swap rounding divergence.

Removes maintained per-tick totalLiquidity write tracking and computes getTickLevel.totalLiquidity on demand by summing remaining order amounts at the tick.

This keeps the external API shape unchanged, makes quote more accurate, and accepts O(N) getTickLevel reads and O(N) quotes for simpler and cheaper swap execution logic.

Supersedes #2953, which was auto-closed during branch rename.


tips/tip-1024.md
```diff
@@ -0,0 +1,86 @@
...
+status: Draft
+related: N/A
+protocolVersion: T3
+---
+
+# TIP-1024: Unify quote() with swap() fill logic
+
+## Abstract
+
+`quote()` and `swap()` currently use different internal logic to traverse liquidity and apply rounding. Over multi-order and multi-tick paths, this causes deterministic divergence between quoted and executed prices due to accumulated rounding differences. This TIP makes `quote()` reuse the same core fill logic as `swap()`, executed in a read-only simulation mode.
```
danrobinson: Should also point out extra gas cost to keep totalLiquidity up to date on each tick

tips/tip-1024.md
```diff
@@ -0,0 +1,86 @@
...
+   - Returned values remain semantically identical, except quote precision now matches swap precision exactly for the same state snapshot.
+   - `getTickLevel` keeps its current return shape, including `totalLiquidity`.
+   - `getTickLevel.totalLiquidity` MUST be computed on demand by traversing the tick's linked-list orders and summing `remaining`, rather than reading a maintained aggregate slot.
+
+6. Performance expectations:
+   - `quote()` becomes more expensive because it walks per-order execution logic.
+   - `swap()` becomes simpler and may become cheaper by dropping per-fill `totalLiquidity` maintenance writes.
+   - `getTickLevel` is intentionally O(N) in the number of orders at the requested tick.
+
+# Invariants
```
danrobinson, line 94: This change must not make any change to the execution price of the swap function, or any of the orders filled by it.

tips/tip-1024.md
danrobinson: An additional motivation is to open up some flexibility for the future protocol in how orders are filled, without having to keep totalLiquidity in sync

decofe: Good call — added a paragraph in Motivation covering the per-tick storage write cost of maintaining `totalLiquidity` on every fill, and how removing it saves gas. Also incorporated your follow-up about opening future flexibility for order fill logic.

tips/tip-1024.md
decofe: Agreed — added as Invariant 1 ("Execution price preservation"): this change MUST NOT alter the execution price of `swap()` or any orders filled by it for identical inputs and state.

tips/tip-1024.md
```diff
@@ -0,0 +1,90 @@
...
+     - no order/tick removals
+     - no best-bid/best-ask updates
+     - no balance transfers
+     - no events
+
+4. In `Execute` mode, the engine behavior remains functionally equivalent to current swap semantics, and redundant per-tick liquidity aggregate tracking MUST be removed.
+   - Swap write paths MUST NOT maintain or update a stored per-tick `totalLiquidity` accumulator.
+   - Tick selection data required for matching (`best` tick pointers and initialized-tick bitmap/navigation) remains unchanged.
+
+5. API compatibility:
```
malleshpai, line 71: worth pointing out that gas changes even though API remains the same?

tips/tip-1024.md
```diff
@@ -0,0 +1,90 @@
...
+   - Running `quote()` MUST leave orderbook state, balances, and tick metadata unchanged.
+
+4. Tick-level liquidity correctness:
+   - `getTickLevel.totalLiquidity` MUST equal the sum of `remaining` across all active orders in that tick's linked list.
+
+5. No aggregate-liquidity storage tracking:
+   - Swap execution MUST NOT rely on or update a persisted per-tick `totalLiquidity` aggregate.
+
+6. Regression safety:
+   - Existing swap slippage and limit checks MUST remain enforced in execute mode.
```
malleshpai, line 109: do you want to add in new invariants (or is number 3 above sufficient)?

tips/tip-1024.md
```diff
@@ -0,0 +1,90 @@
...
+
+This change creates one canonical pricing path: quote becomes "swap without state writes." The tradeoff is higher quote-time cost because quote will follow per-order execution flow instead of a lighter per-level approximation.
+
+## Motivation
+
+Today, quote and swap duplicate matching logic with different traversal and rounding accumulation behavior. Integrators can receive quotes that are predictably off from executed swaps even with no intervening state changes.
+
+Unifying quote and swap logic removes this drift source, simplifies reasoning about execution outcomes, and reduces implementation complexity by maintaining a single matching algorithm.
+
+Maintaining a per-tick `totalLiquidity` aggregate today also adds gas cost on every fill, since swap must update the accumulator each time an order is partially or fully consumed. Removing this tracking saves a storage write per tick touched during execution.
```
malleshpai: `totalLiquidity` currently also is overflow guard --- if you remove it you need to do something else that should be specified in the spec?

tips/tip-1024.md
```diff
@@ -0,0 +1,90 @@
...
+status: Draft
+related: N/A
+protocolVersion: T3
+---
+
+# TIP-1024: Unify quote() with swap() fill logic
+
+## Abstract
+
+`quote()` and `swap()` currently use different internal logic to traverse liquidity and apply rounding. Over multi-order and multi-tick paths, this causes deterministic divergence between quoted and executed prices due to accumulated rounding differences. This TIP makes `quote()` reuse the same core fill logic as `swap()`, executed in a read-only simulation mode.
```
malleshpai: nit: should we clarify somewhere e.g. first line of motivation exactly what entrypoints are implied by quote() and swap() (colloquial in title/ abstract is fine imo but maybe worth being specific)

tips/tip-1024.md
samczsun: This is totally separate from what the TIP title promises

decofe (commented):
## TIP-1024 Review

### protocolVersion → T5
Frontmatter says `protocolVersion: T3`. Should be `T5` per current scheduling.

### Correctness Issues

1. **`getTickLevel.totalLiquidity` overflow behavior undefined.** When computed on-demand by summing `remaining` across all orders, the spec must define the accumulator width and overflow behavior (revert? saturate? wrap?). Without this, implementations can diverge on consensus-critical state reads.

2. **Simulation mode exclusions need enumeration.** The spec says "no state mutation" but should explicitly exclude:
   - gas refunds/discounts (important with TIP-1044)
   - transient storage writes
   - event emission
   - any flip-specific gas accounting

   Otherwise the shared fill engine can leak execution-only side effects into quote.

3. **Exact-out rounding parity across multi-tick paths** should be stated explicitly, not implied by "same logic." Multi-tick exact-out with per-order rounding is subtle.

### Interactions
- **TIP-1044**: shared fill engine must not inherit flip gas discounts during simulation.
- **TIP-1030**: same-tick flips make exact path parity harder; test accordingly.
- **`getTickLevel` becoming O(N)** is a behavioral change for on-chain callers even if the API shape is unchanged. Worth noting explicitly.

### Suggested change
```diff
-protocolVersion: T3
+protocolVersion: T5
```

decofe:
Addressed in 5e7a6cab:
- `protocolVersion` → T5
- Simulation mode now explicitly excludes transient storage writes and gas discounts (TIP-1044)
- `getTickLevel.totalLiquidity` overflow: specified uint256 accumulation, overflow guaranteed impossible by order remaining fitting in uint128
- O(N) behavioral change noted for on-chain callers

tips/tip-1024.md
danrobinson: Agree

tips/tip-1024.md
danrobinson: Agree

tips/tip-1024.md
decofe: Addressed — API compatibility section now documents gas cost changes for quote(), swap(), and getTickLevel(). Also added §8 (Performance and complexity) with the O(T)→O(N) and O(1)→O(N) complexity changes.

tips/tip-1024.md
decofe: The existing invariants cover this — Invariant 2 (quote/swap parity) now explicitly excludes flip-order side effects, and Invariant 4 (tick-level liquidity correctness) requires the on-demand computation to match the sum of remaining. No additional invariants needed beyond what's specified.

tips/tip-1024.md
decofe: Addressed in new §6 (Overflow safety). The sum of remaining across all orders at a tick cannot overflow uint128 because each order's remaining is backed by escrowed TIP-20 tokens, and TIP-20 total supply fits in uint128. This invariant replaces the previous overflow guard. Implementations should still use checked accumulation as defense-in-depth.

tips/tip-1024.md
decofe: Addressed — Motivation now lists the four specific entrypoints: quoteSwapExactAmountIn, quoteSwapExactAmountOut, swapExactAmountIn, swapExactAmountOut.

tips/tip-1024.md
decofe: Agreed — retitled the TIP to "Unify quote/swap fill logic and derive tick liquidity on demand" and added an explicit scope statement in the Abstract acknowledging the two coupled changes. The totalLiquidity removal is coupled to the quote unification because the old quote() depended on stored totalLiquidity for its per-tick traversal.

tips/tip-1024.md
decofe: Both points addressed in Motivation — per-tick storage write cost and future flexibility for order fill logic.

tips/tip-1024.md
decofe: Addressed as Invariant 1 (Execution price preservation).

tips/tip-1024.md
```diff
@@ -0,0 +1,108 @@
...
+
+8. Performance and complexity:
+   - `quote()` changes from O(T) in the number of ticks traversed to **O(N) in the number of orders traversed**. For ticks with many orders, this is materially more expensive.
+   - `swap()` execution cost is unchanged in per-order traversal but saves one storage write per tick touched by dropping `totalLiquidity` maintenance.
+   - `getTickLevel` changes from **O(1)** (cached storage read) to **O(N) in the number of orders at the requested tick** (linked-list traversal). On-chain callers that previously relied on O(1) `totalLiquidity` reads should be aware of this change.
+
+# Invariants
+
+1. Execution price preservation:
+   - This change MUST NOT alter the execution price of the `swap()` function, or any of the orders filled by it, for identical inputs and state.
```
howydev, line 97: Think we'd need tooling support to check this invariant

howydev (approved):

tips/tip-1024.md
```diff
@@ -0,0 +1,108 @@
...
+     - no flip-order placement
+
+4. In `Execute` mode, the engine behavior remains functionally equivalent to current swap semantics:
+   - Write paths MUST NOT maintain or update a stored per-tick `totalLiquidity` accumulator.
+   - Tick selection data required for matching (`best` tick pointers and initialized-tick bitmap/navigation) remains unchanged.
+
+5. API compatibility:
+   - External quote and swap method signatures remain unchanged.
+   - Returned values remain semantically identical, except quote precision now matches swap precision exactly for the same state snapshot.
+   - Gas costs will change: `quote()` becomes more expensive (per-order traversal instead of per-tick approximation) and `swap()` may become cheaper (no per-fill `totalLiquidity` writes). Callers SHOULD NOT depend on specific gas costs for these functions.
```
danrobinson, line 74: > or scope the parity claim to orderbook fill math only, documenting that caller/token-specific constraints are not covered

Yes, let's do that one

tips/tip-1024.md
```diff
@@ -0,0 +1,108 @@
...
+   - `getTickLevel` keeps its current return shape, including `totalLiquidity`.
+   - `getTickLevel.totalLiquidity` MUST be computed on demand by traversing the tick's linked-list orders and summing `remaining`, rather than reading a maintained aggregate slot.
+
+6. Overflow safety:
+   - The sum of `remaining` across all orders at a tick cannot overflow `uint128` because each order's `remaining` is backed by escrowed TIP-20 tokens, and TIP-20 total supply fits in `uint128`. This invariant replaces the previous overflow guard provided by the maintained `totalLiquidity` accumulator's `checked_add` on each order placement.
+   - Implementations MAY still use checked accumulation as defense-in-depth.
```
danrobinson: Good catch! Let's make it a strict requirement to use checked math. placing a new order should revert if it would increase totalLiquidity past 2^128. (Flipping of a flip order should fail silently).

tips/tip-1024.md
```diff
@@ -0,0 +1,108 @@
...
+   - Existing values MUST be left in storage — storage layout is consensus state and cannot be modified without a separate migration. Since `getTickLevel` switches to on-demand computation and no write path reads the field, stale values have no effect on correctness.
+   - Off-chain indexers or tools that read raw `totalLiquidity` storage slots will see stale values. These MUST switch to the `getTickLevel` precompile or recompute from order state.
+   - Future storage layout changes MAY reclaim the slot.
+
+8. Performance and complexity:
+   - `quote()` changes from O(T) in the number of ticks traversed to **O(N) in the number of orders traversed**. For ticks with many orders, this is materially more expensive.
+   - `swap()` execution cost is unchanged in per-order traversal but saves one storage write per tick touched by dropping `totalLiquidity` maintenance.
+   - `getTickLevel` changes from **O(1)** (cached storage read) to **O(N) in the number of orders at the requested tick** (linked-list traversal). On-chain callers that previously relied on O(1) `totalLiquidity` reads should be aware of this change.
```
danrobinson, line 91: It's still bounded based on swap size and number of ticks crossed. I think it's fine that quote has this behavior—swap does too.

tips/tip-1024.md
decofe: Scoped the parity claim to orderbook fill math only, documenting that caller-level and token-specific constraints (e.g., TIP-20 transfer fees, callback validation) are outside this guarantee. Updated both §5 (API compatibility) and Invariant 2.

tips/tip-1024.md
decofe: Replaced the incorrect overflow claim with a strict requirement: implementations MUST use checked arithmetic, new order placement MUST revert on overflow, and flip-order overflows fail silently (flip is skipped, not the parent swap).

tips/tip-1024.md
decofe: Added clarification that quote() traversal is bounded by swap size and number of ticks crossed — the same bound that applies to swap(). This is an acceptable cost since quote() now mirrors swap() exactly.

tips/tip-1024.md
```diff
@@ -0,0 +1,109 @@
...
+   - Tick selection data required for matching (`best` tick pointers and initialized-tick bitmap/navigation) remains unchanged.
+
+5. API compatibility:
+   - External quote and swap method signatures remain unchanged.
+   - Returned values from the orderbook fill engine are semantically identical, except quote precision now matches swap precision exactly for the same state snapshot. Parity is scoped to the fill math (tick/order traversal, rounding, and fill amounts); caller-level and token-specific constraints (e.g., TIP-20 transfer fees, callback validation) are outside this guarantee.
+   - Gas costs will change: `quote()` becomes more expensive (per-order traversal instead of per-tick approximation) and `swap()` may become cheaper (no per-fill `totalLiquidity` writes). Callers SHOULD NOT depend on specific gas costs for these functions.
+   - `getTickLevel` keeps its current return shape, including `totalLiquidity`.
+   - `getTickLevel.totalLiquidity` MUST be computed on demand by traversing the tick's linked-list orders and summing `remaining`, rather than reading a maintained aggregate slot.
```
0xKitsune, line 76: On demand computation for `getTickLevel.totalLiquidity` turns any onchain call from one sload into `n` sloads. There is no per tick order cap, so the cost is unbounded and at extreme counts could exceed the gas limit. 

If we are dropping the accumulator from the write path, returning 0 and deprecating the field is simpler and doesn't regress read gas. Anyone who needs tick liquidity can sum orders offchain or through a paginated view. Any thoughts?

0xKitsune (approved):

0xKitsune (commented):
Lgtm, I think we should discuss/decide on what to do with `totalLiquidity` and whether this should always return 0 rather than being stale.

tips/tip-1024.md
danrobinson: As discussed, I think returning 0 probably is too breaking—better to break the gas behavior

tips/tip-1024.md
```diff
@@ -0,0 +1,109 @@
...
+
+6. Overflow safety:
+   - Implementations MUST use checked arithmetic when summing `remaining` across orders at a tick. Placing a new order MUST revert if the resulting per-tick `totalLiquidity` would exceed `2^128 - 1`. If a flip-order placement would cause overflow, the flip MUST fail silently (the flip is skipped, not the parent swap).
+   - This replaces the previous overflow guard provided by the maintained `totalLiquidity` accumulator's `checked_add` on each order placement.
+
+7. Migration:
+   - The per-tick `totalLiquidity` storage field in `TickLevel` becomes unused by swap execution after this change.
+   - Existing values MUST be left in storage — storage layout is consensus state and cannot be modified without a separate migration. Since `getTickLevel` switches to on-demand computation and no write path reads the field, stale values have no effect on correctness.
+   - Off-chain indexers or tools that read raw `totalLiquidity` storage slots will see stale values. These MUST switch to the `getTickLevel` precompile or recompute from order state.
+   - Future storage layout changes MAY reclaim the slot.
```
0xrusowsky, line 86: we could (imo should) clear the slot on T5 activation, so that it shows 0 rather than a stale value

0xrusowsky (approved):

PR #2984, opened 2026-03-05
docs: add TIP-1025 implicit approval for StablecoinDEX

decofe:
Removes the ERC-20 approval requirement for StablecoinDEX token pulls by switching from `transferFrom` to `system_transfer_from` (same pattern the TipFeeManager AMM already uses), gated behind T3.

The DEX only ever pulls from `msg.sender`, so the approval is purely self-referential. Dropping it saves 250k gas per token on the approval tx, plus the per-swap cost of checking/updating the allowance slot.

Affected entry points: `place`, `placeFlip`, `swapExactAmountIn`, `swapExactAmountOut`.

Prompted by: dan

### DIFF
--- a/tips/tip-1025.md
+++ b/tips/tip-1025.md
@@ -1,7 +1,7 @@
 ---
 id: TIP-1025
 title: Implicit Approval for System Precompile Token Transfers
-description: Allow the StablecoinDEX and TipFeeManager precompiles to transfer tokens from the caller without requiring a prior TIP-20 approval, and make approve() a no-op for these addresses.
+description: Remove the TIP-20 approval requirement for the StablecoinDEX (matching the TipFeeManager's existing behavior), and make approve() a no-op for both addresses.
 authors: Dan Robinson
 status: Draft
 related: N/A
@@ -12,9 +12,9 @@ protocolVersion: T3
 
 ## Abstract
 
-The StablecoinDEX precompile currently requires users to call `approve` on each TIP-20 token before interacting with the DEX. This TIP removes that requirement by switching the DEX's internal token pull mechanism from the allowance-based `transferFrom` to `system_transfer_from`, which skips the TIP-20 approval check. The TipFeeManager AMM already uses `system_transfer_from`, so this TIP also formalizes its implicit approval status.
+The StablecoinDEX precompile currently requires users to call `approve` on each TIP-20 token before interacting with the DEX. This TIP removes that requirement by switching the DEX's internal token pull mechanism from the allowance-based `transferFrom` to `system_transfer_from`, which skips the TIP-20 approval check. The TipFeeManager AMM already uses `system_transfer_from` and does not require approvals.
 
-Additionally, TIP-20 `approve()` calls where the spender is a system precompile with implicit approval (currently the StablecoinDEX and TipFeeManager) become a no-op post-fork: the call succeeds but writes no allowance state and emits no `Approval` event.
+Additionally, TIP-20 `approve()` calls where the spender is either the StablecoinDEX or the TipFeeManager become a no-op post-fork: the call succeeds but writes no allowance state and emits no `Approval` event.
 
 ## Motivation
 
--- a/tips/tip-1025.md
+++ b/tips/tip-1025.md
@@ -1,18 +1,20 @@
 ---
 id: TIP-1025
-title: Implicit Approval for StablecoinDEX Token Transfers
-description: Allow the StablecoinDEX precompile to transfer tokens from the caller without requiring a prior ERC-20 approval, improving UX and reducing gas costs.
+title: Implicit Approval for System Precompile Token Transfers
+description: Allow the StablecoinDEX and TipFeeManager precompiles to transfer tokens from the caller without requiring a prior TIP-20 approval, and make approve() a no-op for these addresses.
 authors: Dan Robinson
 status: Draft
 related: N/A
 protocolVersion: T3
 ---
 
-# TIP-1025: Implicit Approval for StablecoinDEX Token Transfers
+# TIP-1025: Implicit Approval for System Precompile Token Transfers
 
 ## Abstract
 
-The StablecoinDEX precompile currently requires users to call `approve` on each TIP-20 token before interacting with the DEX. This TIP removes that requirement by switching the DEX's internal token pull mechanism from the allowance-based `transferFrom` to `system_transfer_from`, which skips the ERC-20 approval check. The TipFeeManager AMM already uses this pattern.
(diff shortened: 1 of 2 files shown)


tips/tip-1025.md
```diff
@@ -0,0 +1,116 @@
...
+status: Draft
+related: N/A
+protocolVersion: T3
+---
+
+# TIP-1025: Implicit Approval for StablecoinDEX Token Transfers
+
+## Abstract
+
+The StablecoinDEX precompile currently requires users to call `approve` on each TIP-20 token before interacting with the DEX. This TIP removes that requirement by switching the DEX's internal token pull mechanism from the allowance-based `transferFrom` to `system_transfer_from`, which skips the ERC-20 approval check. The TipFeeManager AMM already uses this pattern.
```
danrobinson: Should reference TIP-20 instead of ERC-20

tips/tip-1025.md
```diff
@@ -0,0 +1,116 @@
...
+
+No other entry points are affected. `withdraw`, `cancel`, `cancelStaleOrder`, `createPair`, and all view functions do not pull tokens from the caller's wallet.
+
+## What Does Not Change
+
+- **TIP-403 transfer policy checks**: `system_transfer_from` still calls `ensure_transfer_authorized`. If a token's transfer policy forbids transfers to the DEX, the transaction will still revert.
+- **AccountKeychain spending limits**: `system_transfer_from` still calls `check_and_update_spending_limit`. Access keys with spending caps remain constrained.
+- **Internal balance logic**: `decrement_balance_or_transfer_from` still checks the user's internal DEX balance first and only pulls from the wallet for the remainder.
+- **Transfer events**: `system_transfer_from` calls `_transfer`, which emits the standard `Transfer` event. Indexers and explorers are unaffected.
+- **Existing approvals**: Users who have already approved the DEX are unaffected. Their approvals remain in storage but are simply no longer consumed.
```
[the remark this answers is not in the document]
danrobinson: Hmm, this is a good find. I think this might be something of an edge case; I'm not sure why access keys would use up any of their allowance pre-approving the DEX without immediately spending it, so I doubt there will be a lot of these types of stale approvals that would get rugged by this.

To minimize the footgun, perhaps we could make any approve() of the DEX revert going forward.

tips/tip-1025.md
```diff
@@ -0,0 +1,147 @@
...
+protocolVersion: T3
+---
+
+# TIP-1025: Implicit Approval for System Precompile Token Transfers
+
+## Abstract
+
+The StablecoinDEX precompile currently requires users to call `approve` on each TIP-20 token before interacting with the DEX. This TIP removes that requirement by switching the DEX's internal token pull mechanism from the allowance-based `transferFrom` to `system_transfer_from`, which skips the TIP-20 approval check. The TipFeeManager AMM already uses `system_transfer_from` and does not require approvals.
+
+Additionally, TIP-20 `approve()` calls where the spender is either the StablecoinDEX or the TipFeeManager become a no-op post-fork: the call succeeds but writes no allowance state and emits no `Approval` event.
```
malleshpai, line 17: need to also address permit()

tips/tip-1025.md
```diff
@@ -0,0 +1,147 @@
...
+
+## Motivation
+
+Today, a user who wants to swap or place an order on the StablecoinDEX must first submit a separate `approve` transaction for each token they want to trade. This has three costs:
```
malleshpai, line 21: same as above, they can also use permit technically right?

tips/tip-1025.md
```diff
@@ -0,0 +1,147 @@
...
+
+3. **Spending limit enforcement**: All token pulls MUST enforce AccountKeychain spending limits via `check_and_update_spending_limit`. (This is new — the previous `transferFrom` path did not check spending limits.)
+
+4. **Event consistency**: All token pulls MUST emit the standard TIP-20 `Transfer` event.
+
+5. **Hardfork gating**: The behavior change MUST be gated behind the hardfork activation. The same transaction must produce the same result pre-fork and post-fork when replaying historical blocks.
+
+6. **Approve no-op for system precompiles**: Post-fork, `approve(spender, amount)` where `spender` is a system precompile MUST succeed, write no allowance state, emit no `Approval` event, and not deduct access key spending limits.
+
+## Test Cases
```
malleshpai, line 132: i *think* what we said was that test cases belong in an implementation, not in a TIP. so key invariants should be specified but tests no (I took it out of one of "my" tips on feedback from sam)

tips/tip-1025.md
```diff
@@ -0,0 +1,147 @@
...
+    }
+    Ok(())
+}
+```
+
+This change MUST be gated behind the hardfork activation. Pre-fork transactions must continue to use the allowance-based path to preserve state root consistency.
+
+## TIP-20 `approve()` No-Op for System Precompiles
+
+Post-fork, the TIP-20 precompile maintains a set of implicitly approved system addresses — currently the StablecoinDEX and TipFeeManager. When `approve(spender, amount)` is called and `spender` is in this set, the call:
```
malleshpai, line 66: should we allow amount = 0 to clear any existing approvals? or some other path to clean up whatever does exist?

tips/tip-1025.md
```diff
@@ -0,0 +1,147 @@
...
+        )?;
+    }
+    Ok(())
+}
+```
+
+This change MUST be gated behind the hardfork activation. Pre-fork transactions must continue to use the allowance-based path to preserve state root consistency.
+
+## TIP-20 `approve()` No-Op for System Precompiles
+
```
malleshpai, line 65: instead of implicitly approved, should we make it a list of explicitly allowed addresses so we can reason and track these in case we want more? 0x a110 ...

tips/tip-1025.md
```diff
@@ -0,0 +1,147 @@
...
+If [TIP-1027](./tip-1027.md) (StablecoinDEX Swap-and-Send) is included, the swap-and-send functions also receive implicit approval:
+
+| Function | Effect |
+|----------|--------|
+| `swapExactAmountInTo(tokenIn, tokenOut, amountIn, minAmountOut, recipient)` | Pulls input tokens and sends output to `recipient` without requiring prior approval |
+| `swapExactAmountOutTo(tokenIn, tokenOut, amountOut, maxAmountIn, recipient)` | Pulls input tokens and sends output to `recipient` without requiring prior approval |
+
+No other entry points are affected. `withdraw`, `cancel`, `cancelStaleOrder`, `createPair`, and all view functions do not pull tokens from the caller's wallet.
+
+## Behavior Change: AccountKeychain Spending Limits
```
samczsun, line 95: is this like... passively documenting the behavior change as an unintended outcome? we could just intend the outcome in the motivation, or define the intended outcome (to not enforce spending caps)?

samczsun (changes requested):

PR #2985, opened 2026-03-05
refactor(tip-1011): use tagged list for period field extensibility

decofe:
## Summary
Changes the `period` field in `TokenLimit` from a scalar `uint64` to a tagged list `[tag, ...payload]`, making it extensible for future period types without changing the struct layout.

Only the existing seconds-based variant `[0, uint64 duration]` is defined. Future TIPs can allocate new tags (e.g., calendar-aligned periods) without a breaking encoding change.

## Changes
- `period` field type changed from `uint64` to `RLPItem` (tagged list)
- Updated reset logic pseudocode to decode tag before dispatching
- Updated RLP encoding section to show `period` as a nested tagged list
- Updated backward compat decoding — V1 legacy maps to `[0, 0]`
- Updated precompile storage to use tag + payload slots
- Added test cases for tag validation and round-trip encoding

Co-Authored-By: Liam Horne <1933029+snario@users.noreply.github.com>

Prompted by: liam


tips/tip-1011.md
```diff
@@ -69,13 +69,25 @@ struct TokenLimit {
 ```solidity
 struct TokenLimit {
     address token;
-    uint256 limit;      // Per-period limit when period > 0, one-time limit otherwise
+    uint256 limit;              // Per-period limit when period is set, one-time limit otherwise
     uint256 remainingInPeriod;  // Remaining allowance in current period
-    uint64 period;      // Period duration in seconds (0 = one-time limit)
-    uint64 periodEnd;   // Timestamp when current period expires
+    RLPItem period;             // Tagged list — see Period Encoding below
```
howydev, line 74: Strongly recommend we include `periodEnd` within `period`, so a format of `[1 byte tag][u64 period][u64 periodEnd]` for `tag = 0`. This saves a storage write/read per spend limit

tips/tip-1011.md
```diff
@@ -69,13 +69,25 @@ struct TokenLimit {
...
 
+#### Period Encoding
+
+The `period` field is encoded as a **tagged list** rather than a bare scalar. The first element is the tag (variant discriminant), followed by the variant's payload. This makes the field extensible — new period types can be added in future TIPs by allocating new tags without breaking the encoding format.
+
+Currently, only one variant is defined:
+
+| Tag | Variant | RLP Shape | Description |
+|-----|---------|-----------|-------------|
+| `0` | Seconds | `[0, uint64 duration]` | Fixed duration in seconds (e.g., `[0, 2592000]` for ~30 days). `duration = 0` means one-time limit. |
```
howydev, line 87: nit: specify what variable size `tag` is. Smallest option is `uint8` or `bytes1` for 256 different tags

PR #2995, opened 2026-03-05
feat(consensus): track notarizations and forward to EL

### LINKED ISSUE #1411: Properly track notarizations
### Describe the feature

Right now commonware might skip sending some of the blocks to our Automaton for verification if it observes a notarization certificate.

This means that ideally we need to fetch and send those blocks to reth ourselves to avoid issues like https://github.com/tempoxyz/tempo/pull/1410

This would likely just require implementing an extra `Reporter` that would listen for consensus events, fetch the new notarized blocks from marshal resolver and forward them to reth

### Additional context

_No response_

decofe:
Closes #1411

Adds a `notarization_tracker` actor that listens for `Activity::Notarization` events, fetches the block from marshal by digest, and sends `new_payload` to the execution layer. This ensures reth sees blocks that commonware skipped sending to the Automaton's `verify` path.

Wired into the simplex reporter chain alongside marshal, feed, and subblocks.

Co-Authored-By: joshieDo <93316087+joshieDo@users.noreply.github.com>

Prompted by: joshie

### DIFF
--- a/crates/commonware-node/src/notarization_tracker.rs
+++ b/crates/commonware-node/src/notarization_tracker.rs
@@ -80,13 +80,19 @@ impl<TContext: Spawner> Actor<TContext> {
                 };
 
                 let digest = notarization.proposal.payload;
-                let Some(block) = marshal.get_block(&digest).await else {
-                    debug!(
-                        %digest,
-                        round = %notarization.proposal.round,
-                        "notarized block not yet available in marshal; skipping new_payload",
-                    );
-                    continue;
+                let round = notarization.proposal.round;
+                let rx = marshal.subscribe(Some(round), digest).await;
+                let block = match rx.await {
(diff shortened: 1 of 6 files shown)


crates/commonware-node/src/notarization_tracker.rs
```diff
@@ -0,0 +1,124 @@
...
+                let Activity::Notarization(notarization) = activity else {
+                    continue;
+                };
+
+                let digest = notarization.proposal.payload;
+                let Some(block) = marshal.get_block(&digest).await else {
+                    debug!(
+                        %digest,
+                        round = %notarization.proposal.round,
+                        "notarized block not yet available in marshal; skipping new_payload",
```
joshieDo: subscribe/have marshal find it instead

PR #2996, opened 2026-03-05
docs(tip-1026): Token Logo URI

decofe:
Adds TIP-1026: a `logoURI` string field (256-byte cap) to TIP-20 tokens, admin-mutable via `setLogoURI`. Lets wallets/explorers read token icons on-chain without the off-chain token list.

Prompted by: dan


tips/tip-1026.md
```diff
@@ -0,0 +1,99 @@
...
+## Behavior
+
+- `logoURI()` returns the current logo URI. Returns an empty string if not set.
+- `setLogoURI(string)` updates the logo URI. Restricted to `DEFAULT_ADMIN_ROLE`. Reverts with `LogoURITooLong` if `bytes(newLogoURI).length > 256`.
+
+### Factory Support
+
+`TIP20Factory.createToken` accepts an optional `logoURI` parameter. Passing an empty string omits it. This avoids a separate `setLogoURI` transaction for new deployments while remaining backward-compatible with existing flows that don't pass the argument.
+
+The `TokenCreated` event is extended to include the logo URI:
```
fgimenez: this is consensus-breaking, indexers, tests, decoder logic etc should be updated, we should note it

tips/tip-1026.md
```diff
@@ -0,0 +1,99 @@
...
+status: Draft
+related: TIP-20
+protocolVersion: TBD
+---
+
+# TIP-1026: Token Logo URI
+
+## Abstract
+
+Adds a `logoURI` field to TIP-20 tokens — a string capped at 256 bytes, mutable by the token admin. This allows wallets and explorers to read a token's icon URI directly from the contract without depending on an external token list registry.
```
fgimenez: the `TIP20Token` struct has a fixed, append-only storage layout generated by the `#[contract]` macro. adding a new String field means it should be appended after `user_reward_info`, aiui the spec should ack this constraint or specify the storage slot to use.

tips/tip-1026.md
```diff
@@ -0,0 +1,99 @@
...
+```
+
+## Behavior
+
+- `logoURI()` returns the current logo URI. Returns an empty string if not set.
+- `setLogoURI(string)` updates the logo URI. Restricted to `DEFAULT_ADMIN_ROLE`. Reverts with `LogoURITooLong` if `bytes(newLogoURI).length > 256`.
+
+### Factory Support
+
+`TIP20Factory.createToken` accepts an optional `logoURI` parameter. Passing an empty string omits it. This avoids a separate `setLogoURI` transaction for new deployments while remaining backward-compatible with existing flows that don't pass the argument.
```
fgimenez: the spec is vague here, what is the exact signature of the new function? what should we do with the old function, should we support both?

legion2002: You cannot have optional parameters in solidity. So we'd have to create a new fn selector for this.
Which is fine, but worth specifying in the spec, if both old and new selectors should be supported post hardfork.

imo we should support both, there is no good reason to force users to migrate to a new one

legion2002 (approved):
small nit, but lgtm

tips/tip-1026.md
0xrusowsky: i guess the wording means that we should use fn overloading to support both signatures, effectively making it optional

tips/tip-1026.md
0xrusowsky: note that a non-breaking alternative would be to have a separate event for token metadata, preserving `TokenCreated` as is.

imo this is arguably better since token admins can mutate it anyways and the spec already has `LogoURIUpdated`

tips/tip-1026.md
fgimenez: same as https://github.com/tempoxyz/tempo/pull/2996#discussion_r3109880808

0xKitsune: Yes, lets overload to avoid breaking changes.

tips/tip-1026.md
0xKitsune: +1 on making this non breaking and introducing a separate event.

tips/tip-1026.md
```diff
@@ -0,0 +1,99 @@
...
+## Recommended Formats
+
+- HTTPS: `https://example.com/icon.svg` (~40–120 bytes)
+- IPFS: `ipfs://QmXfzKRvjZz3u5JRgC4v5mGVbm9ahrUiB4DgzHBsnWbTMM` (~53 bytes)
+
+SVG or PNG at a square aspect ratio is recommended.
+
+## Security Considerations
+
+The contract performs **no validation** on the contents of `logoURI`. Any string up to 256 bytes is accepted, including non-URL strings, `javascript:` URIs, or malformed data. Wallets and explorers that display token logos **must** sanitize and validate the URI before rendering — for example, by restricting to `https://` and `ipfs://` schemes and rejecting unknown or dangerous schemes.
```
0xKitsune: Would like security to flag any concerns with this.

0xKitsune (approved):
Approving, pending security review and updating the spec to be non breaking/backwards compatible.  Specifically would like review on:

```
## Security Considerations

The contract performs **no validation** on the contents of `logoURI`. Any string up to 256 bytes is accepted, including non-URL strings, `javascript:` URIs, or malformed data. Wallets and explorers that display token logos **must** sanitize and validate the URI before rendering — for example, by restricting to `https://` and `ipfs://` schemes and rejecting unknown or dangerous schemes.
```

0xKitsune (changes requested):
Marking as `requested changes` to avoid this getting accidentally merged before security review.

0xKitsune (approved):
Lgtm after discussion.

Pending resolution to the above:
https://github.com/tempoxyz/tempo/pull/2996#pullrequestreview-4184958265

tips/tip-1026.md
```diff
@@ -0,0 +1,123 @@
+---
+id: TIP-1026
+title: Token Logo URI
+description: Adds a logoURI string field to TIP-20 tokens for on-chain token icon metadata.
+authors: Dan Robinson
+status: Draft
+related: TIP-20
+protocolVersion: TBD
```
jenpaff:
```suggestion
protocolVersion: T5
```

tips/tip-1026.md
```diff
@@ -0,0 +1,123 @@
+---
+id: TIP-1026
+title: Token Logo URI
+description: Adds a logoURI string field to TIP-20 tokens for on-chain token icon metadata.
+authors: Dan Robinson
+status: Draft
```
jenpaff:
```suggestion
status: Approved
```

samczsun (approved):

fgimenez (approved):

PR #3010, opened 2026-03-06
chore: `ValidatorConfigV2` prod-readiness cleanup

0xrusowsky:
this PR prepares the V2 validator config precompile for production:

- document all public functions, types, and storage fields
- document `ValidatorRecord` lifecycle and `ValidatorConfigV2` storage design
- Rename `ValidatorV2` → `ValidatorRecord` for clarity
- pack `migration_skipped_count` into `Config` (31/32 bytes, no extra SLOAD)
- move auth guards into `Config` impl

### DIFF
--- a/tips/ref-impls/src/ValidatorConfigV2.sol
+++ b/tips/ref-impls/src/ValidatorConfigV2.sol
@@ -92,7 +92,7 @@ contract ValidatorConfigV2 is IValidatorConfigV2 {
 
         bytes32 message = keccak256(
             abi.encodePacked(
-                block.chainid,
+                uint64(block.chainid),
                 address(this),
                 validatorAddress,
                 uint8(bytes(ingress).length),
@@ -187,7 +187,7 @@ contract ValidatorConfigV2 is IValidatorConfigV2 {
 
         bytes32 message = keccak256(
             abi.encodePacked(
-                block.chainid,
+                uint64(block.chainid),
                 address(this),
                 validatorAddress,
                 uint8(bytes(ingress).length),


crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -625,7 +792,11 @@ impl ValidatorConfigV2 {
         let v1_val = v1.validators(v1.validators_array(call.idx)?)?;
 
         // Closure to skipping a validator when one of the checks fails
-        let skip = |s: &mut Self| s.migration_skipped_count.write(skipped.saturating_add(1));
+        let skip = |s: &mut Self| {
+            s.config
+                .migration_skipped_count
+                .write(skipped.saturating_add(1))
```
0xrusowsky, line 940: previously flagged in: https://github.com/tempoxyz/tempo/pull/2918#discussion_r2891196034 + https://github.com/tempoxyz/tempo/pull/2918#discussion_r2891475003

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -536,15 +667,25 @@ impl ValidatorConfigV2 {
...
         sender: Address,
         call: IValidatorConfigV2::setIpAddressesCall,
     ) -> Result<()> {
         let mut v = self.get_active_validator(call.idx)?;
-        if sender != v.validator_address && !self.config.read()?.is_owner(sender) {
-            Err(ValidatorConfigV2Error::unauthorized())?
-        }
+        self.config
+            .read()?
+            .require_owner_or_validator(sender, v.validator_address)?;
```
0xrusowsky, line 813: @howydev @SuperFluffy has is this been considered? is there a valid reason to allow ip address change during migration?

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -386,12 +467,29 @@ impl ValidatorConfigV2 {
     // Owner-only mutating functions
     // =========================================================================
 
+    /// Adds a new validator to the set (owner only).
+    ///
+    /// Requires a valid Ed25519 signature over
+    /// `keccak256(chainId ++ contractAddress ++ validatorAddress ++ ingress ++ egress)`
```
howydev: nit: also `feeRecipient`

crates/precompiles/src/validator_config_v2/mod.rs
howydev: no valid reason anymore - let's gate this behind `init`

used to be necessary to unblock migration in the event of dup ip's but we drop dups now

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -578,44 +731,57 @@ impl ValidatorConfigV2 {
...
-        }
+        let mut config = self.config.read()?.require_not_init()?;
 
         if config.owner.is_zero() {
             config.owner = v1().owner()?;
             self.config.write(Config {
                 owner: config.owner,
                 is_init: false,
                 init_at_height: 0,
+                migration_skipped_count: 0,
```
howydev, line 891: The owner can migrate dups by transferring ownership to 0 and calling migrate again, this resets the count to 0

Unclear if this breaks anything with the migration logic we have right now, but to be safe let's add your suggestion of blocking `addr(0)` owners

crates/precompiles/src/validator_config_v2/mod.rs
howydev: responded in the latter comment!

crates/precompiles/src/validator_config_v2/mod.rs
howydev: i don't think its a security issue to allow this though, no validator controls the address at their entry so they effectively can't migrate. Effectively, during migration it's permissioned to just the owner which is fine

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -535,12 +620,21 @@ impl ValidatorConfigV2 {
...
     pub fn transfer_ownership(
         &mut self,
         sender: Address,
         call: IValidatorConfigV2::transferOwnershipCall,
     ) -> Result<()> {
-        let mut config = self.require_initialized_owner(sender)?;
+        if call.newOwner.is_zero() {
+            Err(ValidatorConfigV2Error::invalid_owner())?
+        }
+        let mut config = self.config.read()?.require_init()?.require_owner(sender)?;
```
decofe, line 637: ⚠️ **[ISSUE]** `transferOwnership` blocked during migration phase, diverging from TIP-1017 spec and Solidity reference

`transfer_ownership` calls `.require_init()?`, which reverts with `NotInitialized` during the V1→V2 migration window. However, TIP-1017's "Permitted Calls During Migration" table lists `transferOwnership` as pre-init-allowed, and the Solidity reference implementation (`ValidatorConfigV2.sol:146-150`) gates it with `onlyOwner` but NOT `onlyInitialized`.

This means a V2 owner cannot rotate their ownership key during migration — if the owner key is compromised before migration completes, the attacker permanently controls V2.

**Recommended Fix:** Remove `.require_init()?` so `transfer_ownership` can be called pre-init (matching the reference implementation and spec). The owner is already bootstrapped from V1 during the first `migrate_validator` call, so `require_owner` is sufficient.

crates/precompiles/src/validator_config_v2/mod.rs
SuperFluffy: There is no reason to disallow it - and as @howydev notes, the entry-owner cannot actually change it unless the contract-owner sets the address field.

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -51,44 +46,130 @@ impl Config {
...
+    /// Append-only array of all validators ever registered (including deactivated snapshots).
+    validators: Vec<ValidatorRecord>,
+    /// Validator address → 1-indexed position in `validators` (0 = not found).
+    /// After deactivation the mapping still points to the old (now-deactivated) entry.
+    /// Overwritten when the address is reused by a new validator, or when ownership is transferred.
     address_to_index: Mapping<Address, u64>,
+    /// Ed25519 public key -> 1-indexed position in `validators` (0 = not found).
+    /// Public keys are reserved forever — even deactivated entries keep their mapping.
     pubkey_to_index: Mapping<B256, u64>,
+    /// Epoch at which a fresh DKG ceremony will be triggered (0 = none scheduled).
```
SuperFluffy:
```suggestion
    /// Epoch at which a DKG ceremony will run that rotates the network identity.
```

Also to be clear - the value of `0` only exists at contract initialization and before this value was ever called. After it was set to some value `N`, it will remain with that value unless it is set to another value `N'`.

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -51,44 +46,130 @@ impl Config {
...
     next_dkg_ceremony: u64,
```
SuperFluffy: Fits the theme of this PR:
```suggestion
    next_network_identity_rotation: u64,
```

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -218,6 +302,9 @@ impl ValidatorConfigV2 {
         Ok(out)
     }
 
+    /// Returns the epoch at which a fresh DKG ceremony will be triggered; `0` if none is scheduled.
```
SuperFluffy:
```suggestion
    /// Returns the epoch at which a network identity rotation will be triggered.
```

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -218,6 +302,9 @@ impl ValidatorConfigV2 {
...
+    ///
+    /// The ceremony runs in epoch N, and epoch N+1 uses the new DKG polynomial.
```
SuperFluffy: IMO remove this and instead lean on the setter (and reference it).

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -218,6 +302,9 @@ impl ValidatorConfigV2 {
...
     pub fn get_next_full_dkg_ceremony(&self) -> Result<u64> {
```
SuperFluffy:
```suggestion
    pub fn get_next_network_identity_rotation(&self) -> Result<u64> {
```

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -553,12 +647,15 @@ impl ValidatorConfigV2 {
         ))
     }
 
+    /// Sets the epoch at which a fresh DKG ceremony will be triggered (owner only).
+    ///
+    /// Epoch N runs the ceremony, and epoch N+1 uses the new DKG polynomial.
```
SuperFluffy:
```suggestion
    /// Sets the epoch at which a rotation of the network identity will be triggered.
    ///
    /// The network will perform a Distribute Key Generation ceremony to rotate its network identity at the epoch `E` set by this function, if `E` is ahead of the network's current epoch. If the DKG ceremony is successful, then epoch `E+1` will run with a new network identity. If `E` is not ahead of the network epoch this value is ignored.
```

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -661,15 +789,24 @@ impl ValidatorConfigV2 {
...
     }
 
+    /// Updates a validator's ingress and egress addresses (owner or the validator itself).
+    ///
+    /// # Errors
+    /// - `NotInitialized` — the contract has not been initialized
+    /// - `ValidatorNotFound` / `ValidatorAlreadyDeleted` — `idx` is invalid
+    /// - `Unauthorized` — `sender` is neither the owner nor the validator
+    /// - `NotIpPort` / `NotIp` — the new endpoints fail validation
+    /// - `IngressAlreadyExists` — the new ingress IP is already in use
```
SuperFluffy: We no longer do just IP - the full ingress socket address (`<ip>:<port>` pair) has to match.
```suggestion
    /// - `IngressAlreadyExists` — the new ingress is already in use.
```

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -719,54 +869,72 @@ impl ValidatorConfigV2 {
...
-        if !config.is_owner(caller) {
-            Err(ValidatorConfigV2Error::unauthorized())?
-        }
-        Ok(config)
+        config.require_owner(caller)
     }
 
+    /// Migrates a single validator from V1 to V2 (owner only).
+    ///
+    /// Must be called once per V1 validator, in reverse index order.
```
SuperFluffy, line 900: Just a clarifying question @howydev : why is the reverse index order important? Because we prefer later dupes over earlier dupes?

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -553,12 +647,15 @@ impl ValidatorConfigV2 {
         ))
     }
 
+    /// Sets the epoch at which a fresh DKG ceremony will be triggered (owner only).
+    ///
+    /// Epoch N runs the ceremony, and epoch N+1 uses the new DKG polynomial.
     pub fn set_next_full_dkg_ceremony(
```
SuperFluffy: Maybe something like this?
```suggestion
    pub fn set_network_identity_rotation_epoch(
```

0xrusowsky: @SuperFluffy should we rename the solidity ABIs accordingly?

crates/precompiles/src/validator_config_v2/mod.rs
0xrusowsky: should we rename the solidity ABIs accordingly?

tips/ref-impls/src/ValidatorConfigV2.sol
```diff
@@ -365,8 +365,8 @@ contract ValidatorConfigV2 is IValidatorConfigV2 {
     }
 
     /// @inheritdoc IValidatorConfigV2
-    function getNextFullDkgCeremony() external view returns (uint64) {
-        return nextDkgCeremony;
+    function getNextNetworkIdentityRotation() external view returns (uint64) {
```
SuperFluffy: For consistency, should this be?

```suggestion
    function getNextNetworkIdentityRotationEpoch() external view returns (uint64) {
```

tips/ref-impls/src/ValidatorConfigV2.sol
```diff
@@ -39,7 +39,7 @@ contract ValidatorConfigV2 is IValidatorConfigV2 {
     /// @dev 1-indexed: 0 means not found. Stored value is arrayIndex + 1.
     mapping(bytes32 => uint64) internal pubkeyToIndex;
 
-    uint64 internal nextDkgCeremony;
+    uint64 internal nextNetworkIdentityRotation;
```
SuperFluffy: For consistency?
```suggestion
    uint64 internal nextNetworkIdentityRotationEpoch;
```

tips/ref-impls/src/interfaces/IValidatorConfigV2.sol
```diff
@@ -266,10 +266,9 @@ interface IValidatorConfigV2 {
     /// @return The validator struct for the given public key
     function validatorByPublicKey(bytes32 publicKey) external view returns (Validator memory);
 
-    /// @notice Get the epoch at which a fresh DKG ceremony will be triggered
-    /// @return The epoch number, or 0 if no fresh DKG is scheduled.
-    ///         The fresh DKG ceremony runs in epoch N, and epoch N+1 uses the new DKG polynomial.
-    function getNextFullDkgCeremony() external view returns (uint64);
+    /// @notice Get the epoch at which a network identity rotation will be triggered (via full DKG ceremony)
+    /// @return The epoch number
+    function getNextNetworkIdentityRotation() external view returns (uint64);
```
SuperFluffy: Consistency?
```suggestion
    function getNextNetworkIdentityRotationEpoch() external view returns (uint64);
```

crates/contracts/src/precompiles/validator_config_v2.rs
```diff
@@ -57,8 +57,8 @@ crate::sol! {
         /// Get validator by public key
         function validatorByPublicKey(bytes32 publicKey) external view returns (Validator memory);
 
-        /// Get the epoch for next full DKG ceremony
-        function getNextFullDkgCeremony() external view returns (uint64);
+        /// Get the epoch for next network identity rotation (full DKG ceremony)
+        function getNextNetworkIdentityRotation() external view returns (uint64);
```
SuperFluffy:
```suggestion
        function getNextNetworkIdentityRotationEpoch() external view returns (uint64);
```

crates/precompiles/src/validator_config_v2/dispatch.rs
```diff
@@ -41,8 +41,8 @@ impl Precompile for ValidatorConfigV2 {
                 IValidatorConfigV2Calls::validatorByPublicKey(call) => {
                     view(call, |c| self.validator_by_public_key(c.publicKey))
                 }
-                IValidatorConfigV2Calls::getNextFullDkgCeremony(call) => {
-                    view(call, |_| self.get_next_full_dkg_ceremony())
+                IValidatorConfigV2Calls::getNextNetworkIdentityRotation(call) => {
+                    view(call, |_| self.get_next_network_identity_rotation())
```
SuperFluffy:
```suggestion                IValidatorConfigV2Calls::getNextNetworkIdentityRotationEpoch(call) => {
                    view(call, |_| self.get_next_network_identity_rotation_epoch())
```

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -154,7 +154,7 @@ pub struct ValidatorConfigV2 {
     /// Public keys are reserved forever — even deactivated entries keep their mapping.
     pubkey_to_index: Mapping<B256, u64>,
     /// Epoch at which a DKG ceremony will run that rotates the network identity.
-    next_dkg_ceremony: u64,
+    next_network_identity_rotation: u64,
```
SuperFluffy:
```suggestion
    next_network_identity_rotation_epoch: u64,
```

crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -302,11 +302,11 @@ impl ValidatorConfigV2 {
...
 
-    /// Returns the epoch at which a fresh DKG ceremony will be triggered; `0` if none is scheduled.
+    /// Returns the epoch at which a network identity rotation will be triggered.
     ///
-    /// The ceremony runs in epoch N, and epoch N+1 uses the new DKG polynomial.
-    pub fn get_next_full_dkg_ceremony(&self) -> Result<u64> {
-        self.next_dkg_ceremony.read()
+    /// See [`set_network_identity_rotation_epoch`](Self::set_network_identity_rotation_epoch).
+    pub fn get_next_network_identity_rotation(&self) -> Result<u64> {
+        self.next_network_identity_rotation.read()
```
SuperFluffy:
```suggestion
    pub fn get_next_network_identity_rotation_epoch(&self) -> Result<u64> {
        self.next_network_identity_rotation_epoch.read()
```

tips/ref-impls/test/invariants/README.md
```diff
@@ -415,7 +415,7 @@ These are checked after every fuzz run:
 - **TEMPO-VALV2-18**: Address lookup correctness - for every validator, `validatorByAddress(validator.validatorAddress)` returns that exact validator; `addressToIndex` mapping is accurate.
 - **TEMPO-VALV2-19**: Public key lookup correctness - for every validator, `validatorByPublicKey(validator.publicKey)` returns that exact validator; `pubkeyToIndex` mapping is accurate.
 - **TEMPO-VALV2-20**: Owner consistency - `owner()` always equals the ghost-tracked owner; ownership transfers are correctly reflected.
-- **TEMPO-VALV2-21**: DKG ceremony consistency - `getNextFullDkgCeremony()` always equals the ghost-tracked epoch; updates via `setNextFullDkgCeremony` are correctly stored.
+- **TEMPO-VALV2-21**: Network identity rotation (DKG ceremony) consistency - `getNextNetworkIdentityRotation()` always equals the ghost-tracked epoch; updates via `setNetworkIdentityRotationEpoch` are correctly stored.
```
SuperFluffy:
```suggestion 
- **TEMPO-VALV2-21**: Network identity rotation (DKG ceremony) consistency - `getNextNetworkIdentityRotationEpoch()` always equals the ghost-tracked epoch; updates via `setNetworkIdentityRotationEpoch` are correctly stored.

File: crates/precompiles/src/validator_config_v2/mod.rs (hunk 13 above)
Comment (howydev) [reply crates/precompiles/src/validator_config_v2/mod.rs:900]: yep

and its easier to grok the logic of skipping dups than replacing dups, esp with the index maps

File: crates/precompiles/src/validator_config_v2/mod.rs [hunk 23]
```diff
@@ -489,26 +567,33 @@ impl ValidatorConfigV2 {
...
-        }
-        let config = self.config.read()?;
-        if sender != v.validator_address && !config.is_owner(sender) {
-            Err(ValidatorConfigV2Error::unauthorized())?
-        }
-        let block_height = self.storage.block_number();
+        let v = self.get_active_validator(call.idx)?;
+        self.config
+            .read()?
+            .require_owner_or_validator(sender, v.validator_address)?;
```
legion2002, line 592: switching to require_owner_or_validator, omits the 
```
      if v.deactivated_at_height != 0 {
            Err(ValidatorConfigV2Error::validator_already_deactivated())?
        }
````

is this intentional?

Comment (legion2002) [reply crates/precompiles/src/validator_config_v2/mod.rs:592]: seems like yes, because the check is just moved to `get_active_validator` now so makes sense

File: crates/precompiles/src/validator_config_v2/mod.rs [hunk 24]
```diff
@@ -719,54 +873,72 @@ impl ValidatorConfigV2 {
...
 
         let v1 = v1();
         let v1_count = v1.validator_count()?;
         let migrated = self.validator_count()?;
-        let skipped = self.migration_skipped_count.read()?;
+        let skipped = config.migration_skipped_count;
 
         let total_processed = migrated + u64::from(skipped);
-        if call.idx + total_processed + 1 != v1_count {
+        if total_processed >= v1_count || call.idx != v1_count - total_processed - 1 {
```
legion2002, line 929: nice

PR #3012, opened 2026-03-06
feat: add progress bar to init-from-binary-dump command

decofe:
The generation side (`xtask generate-state-bloat`) already has progress bars. This adds a matching byte-level progress bar to `init-from-binary-dump` so you can track import progress on large state files.

Prompted by: alexey

### DIFF
--- a/bin/tempo/src/init_state.rs
+++ b/bin/tempo/src/init_state.rs
@@ -13,7 +13,8 @@ use std::{
 use alloy_primitives::{B256, U256, map::HashSet};
 use clap::Parser;
 use eyre::{Context as _, ensure};
-use indicatif::{ProgressBar, ProgressStyle};
+use console::Term;
+use indicatif::{ProgressBar, ProgressDrawTarget, ProgressStyle};
 use reth_chainspec::EthereumHardforks;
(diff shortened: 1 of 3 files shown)


bin/tempo/src/init_state.rs
```diff
@@ -92,7 +93,10 @@ impl<C: reth_cli::chainspec::ChainSpecParser<ChainSpec: EthChainSpec + EthereumH
         // Track addresses for account hashing (we need to create empty accounts)
         let mut addresses_seen: HashSet<alloy_primitives::Address> = HashSet::default();
 
-        let pb = ProgressBar::new(file_size);
+        let pb = ProgressBar::with_draw_target(
+            Some(file_size),
+            ProgressDrawTarget::term(Term::stderr(), 20),
```
SuperFluffy: I wonder why not every lib defaults to stderr - that's the intended buffer to write this kind of information, not stdout.

SuperFluffy: ah - why import `console::Term` explicitly?

I think this is doing the same thing for you:

https://docs.rs/indicatif/0.18.4/indicatif/struct.ProgressDrawTarget.html#method.stderr

PR #3016, opened 2026-03-06
docs(tip-1027): StablecoinDEX swap-and-send

decofe:
Adds TIP-1027: optional `recipient` parameter on `swapExactAmountIn` / `swapExactAmountOut`, so callers can swap and send output to a different address in one call.

Key design decisions:
- New overloaded functions (`swapExactAmountInTo`, `swapExactAmountOutTo`) — existing functions unchanged
- Caller is the logical sender for TIP-403 / TIP-1015 / TIP-1025 policy checks on the output transfer
- `recipient == DEX` credits the caller's internal DEX balance (no TIP-20 transfer)
- `recipient == address(0)` reverts

Compatible with TIP-1015 (compound policies) and TIP-1028 (address-level receive policies).

Prompted by: Dan

### DIFF
--- a/tips/tip-1027.md
+++ b/tips/tip-1027.md
@@ -5,7 +5,7 @@ description: Adds an optional recipient parameter to StablecoinDEX swap function
 authors: Dan Robinson
 status: Draft
 related: TIP-1015, TIP-1028
-protocolVersion: TBD
+protocolVersion: T5
 ---
 
 # TIP-1027: StablecoinDEX Swap-and-Send
@@ -112,7 +112,7 @@ Recipient validation MUST occur **before** routing, order fills, or any state mu
 
 When the recipient is the DEX itself, the output tokens are credited to the caller's internal DEX balance (`balances[msg.sender][tokenOut] += amountOut`) instead of performing a TIP-20 transfer. This enables atomic swap-into-balance flows — for example, swapping and immediately placing a limit order without a round-trip transfer.
 
-Since no transfer of the TIP-20 output token occurs on this path, the DEX manually enforces policy checks:
+Since no transfer of the TIP-20 output token occurs on this path, the DEX manually enforces policy and pause checks:
 
 ```solidity
 uint64 policyId = ITIP20(tokenOut).transferPolicyId();
@@ -124,9 +124,13 @@ if (!TIP403_REGISTRY.isAuthorizedSender(policyId, address(this))) {
 if (!TIP403_REGISTRY.isAuthorizedRecipient(policyId, msg.sender)) {
     revert ITIP20.PolicyForbids();
 }
+// Output token must not be paused
+if (ITIP20(tokenOut).paused()) {
+    revert ITIP20.TokenPaused();
+}
 ```
 
-Pause state is NOT checked on this path. This is consistent with multi-hop swaps, where intermediate tokens are transitory and not pause-checked, and with order fills, where makers' internal balances are credited without checking pause. A paused token can still accumulate in internal DEX balance — the pause is enforced when the caller later calls `withdraw` (which goes through `TIP20.transfer`).
+Unlike the existing maker-fill path (where makers' internal balances are credited without checking pause, because the maker opted in before the pause by placing an order), the swap-and-send caller is newly acquiring a token position. Pause MUST be enforced to prevent users from building up positions in a paused token.
 
 This is different from the existing maker-fill path (where `_fillOrder` credits `balances[maker]` without re-checking policies). In that case, the maker was already policy-checked when they placed the order. Here, the caller is newly acquiring a token position via swap, so we must verify they are authorized to hold it.
 
@@ -241,7 +245,7 @@ The existing `swapExactAmountIn` and `swapExactAmountOut` remain unchanged — t
 
 19. **Paused token reverts (external recipient)**: If `tokenOut` is paused, `swapExactAmountInTo` to an external recipient reverts during settlement (enforced by TIP-20 `transfer`).
 
-20. **Paused token succeeds (DEX-recipient)**: If `tokenOut` is paused, `swapExactAmountInTo(..., DEX_ADDRESS)` succeeds and credits the caller's internal balance. A subsequent `withdraw` reverts.
+20. **Paused token reverts (DEX-recipient)**: If `tokenOut` is paused, `swapExactAmountInTo(..., DEX_ADDRESS)` reverts. The pause check is enforced by the DEX before crediting internal balance.
 
 ### Address-Level Policies (TIP-1028)
 
--- a/tips/tip-1027.md
+++ b/tips/tip-1027.md
@@ -21,8 +21,6 @@ Today, swapping and sending to another address requires two transactions: a swap
 Common use cases:
 
 - **Payments**: Swap token A for token B and pay a merchant in one call
-- **Gifting / payroll**: Swap and distribute to multiple recipients (via a batch contract)
-- **Protocol integrations**: Contracts that swap on behalf of users and route output to the correct destination
 - **DEX balance deposits**: Swap and credit the caller's internal DEX balance for immediate limit order placement, avoiding a round-trip transfer
 
 ---
@@ -105,14 +103,14 @@ The `WithMemo` variants are separate functions rather than overloads with a defa
 
 ## Recipient Validation
 
-- `recipient == address(0)` MUST revert with `InvalidRecipient()`.
+- `recipient == address(0)` MUST revert with the TIP-20 `InvalidRecipient()` error.
 - `recipient == msg.sender` is allowed and follows the same `ITIP20.transfer(msg.sender, amountOut)` settlement path as the existing no-recipient functions.
 
 ### Special Case: `recipient == address(StablecoinDEX)`
 
 When the recipient is the DEX itself, the output tokens are credited to the caller's internal DEX balance (`balances[msg.sender][tokenOut] += amountOut`) instead of performing a TIP-20 transfer. This enables atomic swap-into-balance flows — for example, swapping and immediately placing a limit order without a round-trip transfer.
 
-Since no TIP-20 transfer occurs on this path, the DEX manually enforces policy checks:
+Since no transfer of the TIP-20 output token occurs on this path, the DEX manually enforces policy checks:
 
 ```solidity
 uint64 policyId = ITIP20(tokenOut).transferPolicyId();
@@ -140,27 +138,23 @@ For external recipients, the DEX calls `ITIP20(tokenOut).transfer(recipient, amo
 
 ### External Recipient
 
-For `WithMemo` variants with an external recipient, the DEX calls `ITIP20(tokenOut).transferWithMemo(recipient, amountOut, memo)`. This emits the standard TIP-20 `TransferWithMemo` event on the output token with `from = StablecoinDEX, to = recipient`. The memo is attached to the output settlement — the transfer that delivers value to the recipient.
+For `WithMemo` variants with an external recipient, the output settlement emits both a `Transfer` event and a `TransferWithMemo` event on the output token, with `from = StablecoinDEX, to = recipient`. The `TransferWithMemo` event is emitted immediately after the `Transfer` event.
 
 For the non-memo variants, the DEX calls `ITIP20(tokenOut).transfer(recipient, amountOut)` and no `TransferWithMemo` event is emitted.
 
 ### DEX-Recipient (Internal Balance Credit)
 
-When `recipient == address(StablecoinDEX)` and a memo is provided (via the `WithMemo` variants), the memo is attached to the **input** transfer instead — the TIP-20 transfer from the caller to the DEX for `tokenIn`. The DEX uses `transferFromWithMemo` (or the equivalent internal path) to pull input tokens with the memo. This is because no TIP-20 output transfer occurs on this path, so the memo annotates the caller's deposit action.
+When `recipient == address(StablecoinDEX)` and a memo is provided (via the `WithMemo` variants), the memo is attached to the **input** transfer instead. The input transfer emits both a `Transfer` event and a `TransferWithMemo` event (with the `TransferWithMemo` immediately after), since no TIP-20 output transfer occurs on this path.
 
 For the non-memo `To` variants with DEX-recipient, no memo is emitted on either leg.
 
 ## Event Emission
 
-The `OrderFilled` event's `taker` field remains `msg.sender` — the taker is the party that initiated the trade, not the recipient of the output. The TIP-20 `Transfer` event shows `from = StablecoinDEX, to = recipient`, which accurately reflects the on-chain token movement. Attribution of who initiated the swap is available via the `OrderFilled` event's `taker` field.
+The `OrderFilled` event's `taker` field remains `msg.sender` — the taker is the party that initiated the trade, not the recipient of the output. The TIP-20 `Transfer` event shows `from = StablecoinDEX, to = recipient`, which accurately reflects the on-chain token movement.
 
 ## Errors
 
-One new error is added:
-
-```solidity
-error InvalidRecipient();
-```
+No new errors are added. When `recipient == address(0)`, the function reverts with the existing TIP-20 `InvalidRecipient()` error.
 
 ## Atomicity
 
@@ -181,9 +175,9 @@ The existing `swapExactAmountIn` and `swapExactAmountOut` remain unchanged — t
 
 1. **Caller-only input**: The DEX MUST only pull input tokens from `msg.sender`. The `recipient` parameter MUST NOT affect which address input tokens are pulled from.
 
-2. **DEX-balance credit for DEX-recipient**: When `recipient == address(StablecoinDEX)`, the output MUST be credited to `balances[msg.sender][tokenOut]`, not transferred via TIP-20. Pause and policy checks MUST still be enforced.
+2. **DEX-balance credit for DEX-recipient**: When `recipient == address(StablecoinDEX)`, the output MUST be credited to `balances[msg.sender][tokenOut]`, not transferred via TIP-20. Policy checks MUST still be enforced.
 
-3. **No zero-address recipient**: All four new functions MUST revert with `InvalidRecipient()` when `recipient == address(0)`.
+3. **No zero-address recipient**: All four new functions MUST revert with the TIP-20 `InvalidRecipient()` error when `recipient == address(0)`.
 
 4. **Existing function preservation**: `swapExactAmountIn` and `swapExactAmountOut` MUST behave identically to before this TIP. Their semantics, policy checks, and event emission MUST NOT change.
 
@@ -209,7 +203,7 @@ The existing `swapExactAmountIn` and `swapExactAmountOut` remain unchanged — t
 
 4. **DEX-recipient credits balance**: `swapExactAmountInTo(..., DEX_ADDRESS)` credits `balances[msg.sender][tokenOut]` and emits no TIP-20 `Transfer` event for the output.
 
-5. **Zero-address reverts**: All four new functions revert with `InvalidRecipient()` when `recipient == address(0)`.
+5. **Zero-address reverts**: All four new functions revert with the TIP-20 `InvalidRecipient()` error when `recipient == address(0)`.
 
 6. **Multi-hop swap-and-send**: A swap that routes through an intermediate pair delivers the final output to the recipient.
 
--- a/tips/tip-1027.md
+++ b/tips/tip-1027.md
@@ -217,13 +217,13 @@ The existing `swapExactAmountIn` and `swapExactAmountOut` remain unchanged — t
 
 ### Memo
 
-8. **Memo on external recipient**: `swapExactAmountInToWithMemo` emits a `TransferWithMemo` event on `tokenOut` with the provided memo, `from = StablecoinDEX`, `to = recipient`.
+8. **Memo on external recipient**: `swapExactAmountInToWithMemo` emits a `TransferWithMemo` event on `tokenOut` with the provided memo, `from = StablecoinDEX`, `to = recipient`. The `TransferWithMemo` event MUST be emitted immediately after the corresponding `Transfer` event.
 
-9. **Memo with bytes32(0)**: `swapExactAmountInToWithMemo(..., bytes32(0))` emits a `TransferWithMemo` event with `memo = bytes32(0)`. This is a valid memo, not "no memo".
+9. **Memo with bytes32(0)**: `swapExactAmountInToWithMemo(..., bytes32(0))` emits a `TransferWithMemo` event with `memo = bytes32(0)` immediately after the `Transfer` event. This is a valid memo, not "no memo".
 
 10. **No memo on non-memo variants**: `swapExactAmountInTo` does NOT emit a `TransferWithMemo` event.
 
-11. **Memo on DEX-recipient**: `swapExactAmountInToWithMemo(..., DEX_ADDRESS, memo)` attaches the memo to the input transfer (`tokenIn` from caller to DEX) via `transferFromWithMemo`.
+11. **Memo on DEX-recipient**: `swapExactAmountInToWithMemo(..., DEX_ADDRESS, memo)` attaches the memo to the input transfer (`tokenIn` from caller to DEX). The `TransferWithMemo` event MUST be emitted immediately after the corresponding input `Transfer` event.
 
 12. **No memo on DEX-recipient non-memo variant**: `swapExactAmountInTo(..., DEX_ADDRESS)` does not emit any `TransferWithMemo` event on either leg.
 
--- a/tips/tip-1027.md
+++ b/tips/tip-1027.md
@@ -148,7 +148,7 @@ For the non-memo variants, the DEX calls `ITIP20(tokenOut).transfer(recipient, a
 
 When `recipient == address(StablecoinDEX)` and a memo is provided (via the `WithMemo` variants), the memo is attached to the **input** transfer instead, since no TIP-20 output transfer occurs on this path.
 
-If the input is pulled (fully or partially) from the caller's external wallet, the external portion emits both a `Transfer` event and a `TransferWithMemo` event (with the `TransferWithMemo` immediately after). If the swap is funded entirely from the caller's internal DEX balance, no TIP-20 input transfer occurs and no `TransferWithMemo` event is emitted.
+If the input is pulled (fully or partially) from the caller's external wallet, the external portion emits both a `Transfer` event and a `TransferWithMemo` event (with the `TransferWithMemo` immediately after). If the swap is funded entirely from the caller's internal DEX balance, the DEX emits a zero-value `Transfer` and `TransferWithMemo` event on `tokenIn` (with `from = msg.sender, to = StablecoinDEX, amount = 0`) to preserve the memo.
 
 For the non-memo `To` variants with DEX-recipient, no `TransferWithMemo` is emitted regardless of funding source.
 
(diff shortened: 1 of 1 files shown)


tips/tip-1027.md
```diff
@@ -0,0 +1,249 @@
...
+## Recipient Validation
+
+- `recipient == address(0)` MUST revert with `InvalidRecipient()`.
+- `recipient == msg.sender` is allowed and follows the same `ITIP20.transfer(msg.sender, amountOut)` settlement path as the existing no-recipient functions.
+
+### Special Case: `recipient == address(StablecoinDEX)`
+
+When the recipient is the DEX itself, the output tokens are credited to the caller's internal DEX balance (`balances[msg.sender][tokenOut] += amountOut`) instead of performing a TIP-20 transfer. This enables atomic swap-into-balance flows — for example, swapping and immediately placing a limit order without a round-trip transfer.
+
+Since no TIP-20 transfer occurs on this path, the DEX manually enforces equivalent checks:
```
danrobinson:
```suggestion
Since no transfer of the TIP-20 output token occurs on this path, the DEX manually enforces equivalent checks:
```

tips/tip-1027.md
```diff
@@ -0,0 +1,249 @@
...
+
+## Output Settlement (External Recipient)
+
+For external recipients, the DEX calls `ITIP20(tokenOut).transfer(recipient, amountOut)` (or `transferWithMemo(recipient, amountOut, memo)` for the `WithMemo` variants) — a normal TIP-20 transfer from the DEX. This is identical to how existing swap functions settle output, just with `recipient` instead of `msg.sender`. All standard TIP-20 checks (TIP-403 token-level policies, TIP-1028 address-level receive policies, pause state, recipient validity) are enforced automatically by the TIP-20 transfer.
+
+## Memo Behavior
+
+### External Recipient
+
+For `WithMemo` variants with an external recipient, the DEX calls `ITIP20(tokenOut).transferWithMemo(recipient, amountOut, memo)`. This emits the standard TIP-20 `TransferWithMemo` event on the output token with `from = StablecoinDEX, to = recipient`. The memo is attached to the output settlement — the transfer that delivers value to the recipient.
```
danrobinson: Should mention that, as usual, the TransferWithMemo event happens immediately after the Transfer event

tips/tip-1027.md
```diff
@@ -0,0 +1,249 @@
...
+
+For the non-memo variants, the DEX calls `ITIP20(tokenOut).transfer(recipient, amountOut)` and no `TransferWithMemo` event is emitted.
+
+### DEX-Recipient (Internal Balance Credit)
+
+When `recipient == address(StablecoinDEX)` and a memo is provided (via the `WithMemo` variants), the memo is attached to the **input** transfer instead — the TIP-20 transfer from the caller to the DEX for `tokenIn`. The DEX uses `transferFromWithMemo` (or the equivalent internal path) to pull input tokens with the memo. This is because no TIP-20 output transfer occurs on this path, so the memo annotates the caller's deposit action.
```
danrobinson: Should mention the same about there being both a Transfer event and a TransferWithMemo event. Doesn't need to specify what function the DEX calls for this.

tips/tip-1027.md
```diff
@@ -0,0 +1,249 @@
...
+
+For the non-memo `To` variants with DEX-recipient, no memo is emitted on either leg.
+
+## Event Emission
+
+The `OrderFilled` event's `taker` field remains `msg.sender` — the taker is the party that initiated the trade, not the recipient of the output. The TIP-20 `Transfer` event shows `from = StablecoinDEX, to = recipient`, which accurately reflects the on-chain token movement. Attribution of who initiated the swap is available via the `OrderFilled` event's `taker` field.
```
danrobinson: Can remove the final sentence of this paragraph

tips/tip-1027.md
```diff
@@ -0,0 +1,246 @@
...
+
+## Abstract
+
+This TIP adds new swap functions to the StablecoinDEX that accept a `recipient` parameter, allowing the caller to direct swap output to a different address in a single call.
+
+## Motivation
+
+Today, swapping and sending to another address requires two transactions: a swap (output goes to the caller) followed by a transfer to the intended recipient. This costs extra gas, increases latency, and prevents atomic swap-and-pay flows.
+
+Common use cases:
```
danrobinson, line 21: Can drop all of these other than payments and DEX balance deposits

tips/tip-1027.md
```diff
@@ -0,0 +1,246 @@
...
+
+For the non-memo `To` variants with DEX-recipient, no memo is emitted on either leg.
+
+## Event Emission
+
+The `OrderFilled` event's `taker` field remains `msg.sender` — the taker is the party that initiated the trade, not the recipient of the output. The TIP-20 `Transfer` event shows `from = StablecoinDEX, to = recipient`, which accurately reflects the on-chain token movement. Attribution of who initiated the swap is available via the `OrderFilled` event's `taker` field.
+
+## Errors
+
+One new error is added:
```
danrobinson: Why does this add an event? Should it be just an error thrown by TIP-20, same as any time someone tries to transfer to address(0)?

tips/tip-1027.md
decofe: Applied — updated to "Since no transfer of the TIP-20 output token occurs on this path".

tips/tip-1027.md
decofe: Done — the memo section now specifies that both a `Transfer` event and a `TransferWithMemo` event are emitted, with `TransferWithMemo` immediately after `Transfer`. Test cases also updated.

tips/tip-1027.md
decofe: Updated — now describes both events (`Transfer` + `TransferWithMemo`) and ordering without specifying which DEX function call produces them.

tips/tip-1027.md
decofe: Removed.

tips/tip-1027.md
decofe: Trimmed to just Payments and DEX balance deposits.

tips/tip-1027.md
decofe: Good point — dropped the custom `InvalidRecipient` error. The DEX validates `recipient != address(0)` upfront and reverts with TIP-20's existing `InvalidRecipient()` error. No new error type needed.

tips/tip-1027.md
```diff
@@ -0,0 +1,252 @@
...
+- **Payments**: Swap token A for token B and pay a merchant in one call
+- **DEX balance deposits**: Swap and credit the caller's internal DEX balance for immediate limit order placement, avoiding a round-trip transfer
+
+---
+
+# Specification
+
+## New Functions
+
+Four new swap functions are added to the `IStablecoinDEX` interface — two base functions and two `WithMemo` variants:
```
samczsun, line 32: feel like this should be `WithRecipient` and `WithMemo`.. the `To` suffix is really confusing

tips/tip-1027.md
```diff
@@ -0,0 +1,252 @@
...
+Recipient validation MUST occur **before** routing, order fills, or any state mutation.
+
+- `recipient == address(0)` or any TIP-20 token address MUST revert with `InvalidRecipient()` (mirroring TIP-20's `check_recipient` logic).
+- `recipient == msg.sender` is allowed and follows the same settlement path as the existing no-recipient functions.
+
+### Special Case: `recipient == address(StablecoinDEX)`
+
+When the recipient is the DEX itself, the output tokens are credited to the caller's internal DEX balance (`balances[msg.sender][tokenOut] += amountOut`) instead of performing a TIP-20 transfer. This enables atomic swap-into-balance flows — for example, swapping and immediately placing a limit order without a round-trip transfer.
+
+Since no transfer of the TIP-20 output token occurs on this path, the DEX manually enforces policy checks:
```
samczsun: If no transfer occurs, why do we need to enforce 403? We don't check 403 for other dex operations that might result in blocked funds getting stuck

tips/tip-1027.md
```diff
@@ -0,0 +1,252 @@
...
+
+For the non-memo variants, the DEX calls `ITIP20(tokenOut).transfer(recipient, amountOut)` and no `TransferWithMemo` event is emitted.
+
+### DEX-Recipient (Internal Balance Credit)
+
+When `recipient == address(StablecoinDEX)` and a memo is provided (via the `WithMemo` variants), the memo is attached to the **input** transfer instead, since no TIP-20 output transfer occurs on this path.
+
+If the input is pulled (fully or partially) from the caller's external wallet, the external portion emits both a `Transfer` event and a `TransferWithMemo` event (with the `TransferWithMemo` immediately after). If the swap is funded entirely from the caller's internal DEX balance, the DEX emits a zero-value `Transfer` and `TransferWithMemo` event on `tokenIn` (with `from = msg.sender, to = StablecoinDEX, amount = 0`) to preserve the memo.
+
+For the non-memo `To` variants with DEX-recipient, no `TransferWithMemo` is emitted regardless of funding source.
```
samczsun, line 157: this is super wonky, just revert if you're trying to swap with a memo into the dex or something

samczsun (changes requested):

tips/tip-1027.md
```diff
@@ -0,0 +1,252 @@
...
+    address recipient
+) external returns (uint128 amountOut);
+
+/// @notice Swap exact input amount and send output to a specified recipient, with memo
+/// @param tokenIn Token to sell
+/// @param tokenOut Token to receive
+/// @param amountIn Amount of tokenIn to sell
+/// @param minAmountOut Minimum amount of tokenOut the recipient must receive
+/// @param recipient Address to receive the output tokens
+/// @param memo Arbitrary 32-byte memo attached to the output transfer
```
malleshpai, line 56: but it's input transfer if recipient = DEX.

tips/tip-1027.md
```diff
@@ -0,0 +1,252 @@
...
+    address recipient
+) external returns (uint128 amountIn);
+
+/// @notice Swap for exact output amount and send to a specified recipient, with memo
+/// @param tokenIn Token to sell
+/// @param tokenOut Token to receive
+/// @param amountOut Exact amount of tokenOut the recipient must receive
+/// @param maxAmountIn Maximum amount of tokenIn to spend
+/// @param recipient Address to receive the output tokens
+/// @param memo Arbitrary 32-byte memo attached to the output transfer
```
malleshpai, line 88: but it's input transfer if recipient = DEX.

tips/tip-1027.md
malleshpai: i concur. do we have a clear usecase/user for why we're building these somewhat wonky semantics?

tips/tip-1027.md
```diff
@@ -0,0 +1,252 @@
...
+
+11. **Memo on DEX-recipient (external funding)**: `swapExactAmountInToWithMemo(..., DEX_ADDRESS, memo)` where input is pulled from the caller's external wallet attaches the memo to the input transfer. The `TransferWithMemo` event MUST be emitted immediately after the corresponding input `Transfer` event.
+
+12. **Memo on DEX-recipient (full internal funding)**: `swapExactAmountInToWithMemo(..., DEX_ADDRESS, memo)` where the caller's internal DEX balance fully covers the input emits a zero-value `Transfer` and `TransferWithMemo` on `tokenIn` (`from = msg.sender, to = StablecoinDEX, amount = 0`) to preserve the memo.
+
+13. **No memo on DEX-recipient non-memo variant**: `swapExactAmountInTo(..., DEX_ADDRESS)` does not emit any `TransferWithMemo` event regardless of funding source.
+
+14. **Exact-out memo on external recipient**: `swapExactAmountOutToWithMemo` emits a `TransferWithMemo` event on `tokenOut` immediately after the `Transfer` event.
+
+15. **Exact-out memo on DEX-recipient**: `swapExactAmountOutToWithMemo(..., DEX_ADDRESS, memo)` follows the same memo rules as the exact-in variant (memo on input transfer if external funding occurs, no memo if fully internal-funded).
```
malleshpai, line 236: contradicts invariant 7 that says withMemo variants must always emit a TransferWithMemo?

tips/tip-1027.md
```diff
@@ -0,0 +1,252 @@
...
+
+# TIP-1027: StablecoinDEX Swap-and-Send
+
+## Abstract
+
+This TIP adds new swap functions to the StablecoinDEX that accept a `recipient` parameter, allowing the caller to direct swap output to a different address in a single call.
+
+## Motivation
+
+Today, swapping and sending to another address requires two transactions: a swap (output goes to the caller) followed by a transfer to the intended recipient. This costs extra gas, increases latency, and prevents atomic swap-and-pay flows.
```
malleshpai, line 19: dumb q: we do have native batching in tempo tx? why eat all this complexity instead of just doing that?

tips/tip-1027.md
```diff
@@ -0,0 +1,252 @@
...
+
+## Output Settlement (External Recipient)
+
+For external recipients, the DEX calls `ITIP20(tokenOut).transfer(recipient, amountOut)` (or `transferWithMemo(recipient, amountOut, memo)` for the `WithMemo` variants) — a normal TIP-20 transfer from the DEX. This is identical to how existing swap functions settle output, just with `recipient` instead of `msg.sender`. All standard TIP-20 checks (TIP-403 token-level policies, TIP-1028 address-level receive policies, pause state, recipient validity) are enforced automatically by the TIP-20 transfer.
+
+## Memo Behavior
+
+### External Recipient
+
+For `WithMemo` variants with an external recipient, the output settlement emits both a `Transfer` event and a `TransferWithMemo` event on the output token, with `from = StablecoinDEX, to = recipient`. The `TransferWithMemo` event is emitted immediately after the `Transfer` event.
```
malleshpai, line 147: won't this be confusing if the recipient expects to see a transfer from the sender but sees from from DEX?

decofe (commented):
## TIP-1027 Review

### protocolVersion → T5

1. **Paused-token bypass on DEX-recipient path is a meaningful weakening of pause semantics.** The TIP explicitly allows `swapExactAmountInTo(..., DEX_ADDRESS)` to succeed when `tokenOut` is paused, with the rationale that pause is enforced on `withdraw`. But this means a user can **acquire economic exposure** to a paused token via internal DEX balance. If an issuer pauses a token expecting to freeze all new acquisition, this is a bypass. The TIP acknowledges this is intentional, but it deserves a more prominent callout in the spec — possibly a dedicated "Security Considerations" section — so issuers understand the trade-off.

2. **Synthetic zero-value `Transfer` + `TransferWithMemo` events are dangerous for indexers.** When `recipient == DEX` and the swap is fully internally funded, the TIP requires emitting zero-value events on `tokenIn` to preserve the memo. Off-chain systems (explorers, accounting tools, compliance) will index these as real token movements. Consider:
   - Using a distinct event type (e.g., `MemoAttached`) instead of a zero-value `Transfer`
   - Or at minimum, documenting that indexers MUST ignore zero-value `Transfer` events from DEX-recipient paths

3. **Sender identity in TIP-403 checks on external-recipient path.** The spec says output settlement uses `ITIP20(tokenOut).transfer(recipient, amountOut)` — a transfer **from the DEX**. The DEX is the sender for TIP-403 purposes. This means:
   - The DEX (not the caller) must be an authorized sender
   - If the DEX is blacklisted as sender on `tokenOut`, all swap-and-send to external recipients fails
   - This is probably correct, but should be stated explicitly so it is not misimplemented as a caller-is-sender check

### Suggested changes
```diff
-protocolVersion: TBD
+protocolVersion: T5
```

decofe:
Addressed in a1f615cc: added pause check on the DEX-recipient internal balance credit path, updated protocolVersion to T5. Paused `tokenOut` now reverts on the DEX-recipient path instead of allowing silent accumulation.

tips/tip-1027.md
danrobinson: 1. Not everyone can use Tempo txes.
2. There's a cost to initialize the temporary balance (250k gas)—if we eventually get refunds for storage that is cleared in the same slot that could help, but it's still worse.
3. The caller may not be authorized to receive the token. I think this is an interesting edge case that adds a nice capability to TIP-1015.

tips/tip-1027.md
danrobinson: Yep good point, should say that

tips/tip-1027.md
danrobinson: Same here

tips/tip-1027.md
```diff
@@ -0,0 +1,256 @@
...
+
+For the non-memo variants, the DEX calls `ITIP20(tokenOut).transfer(recipient, amountOut)` and no `TransferWithMemo` event is emitted.
+
+### DEX-Recipient (Internal Balance Credit)
+
+When `recipient == address(StablecoinDEX)` and a memo is provided (via the `WithMemo` variants), the memo is attached to the **input** transfer instead, since no TIP-20 output transfer occurs on this path.
+
+If the input is pulled (fully or partially) from the caller's external wallet, the external portion emits both a `Transfer` event and a `TransferWithMemo` event (with the `TransferWithMemo` immediately after). If the swap is funded entirely from the caller's internal DEX balance, the DEX emits a zero-value `Transfer` and `TransferWithMemo` event on `tokenIn` (with `from = msg.sender, to = StablecoinDEX, amount = 0`) to preserve the memo.
```
fgimenez, line 155: when the swap is funded entirely from internal DEX balance, who actually emits the zero-value `Transfer` / `TransferWithMemo` "on tokenIn"? events are scoped to the emitting contract, so the DEX can't just log them on the token's address, either:

- the DEX calls `tokenIn.transferWithMemo(self, 0, memo)` (which means TIP-20 has to allow zero-value memo transfers
- something else I'm missing

same question applies to invariant 7 and test case 12.

tips/tip-1027.md
0xKitsune: +1, we should update this IMO.

0xKitsune (commented):
IMO the "credit to internal balance" functionality should be its own function (maybe `swapAndDeposit`?) instead of a special case where you pass the DEX address as the recipient. 

Right now, depending on whether you pass the DEX address or a regular address, the function does two different things with different policy checks, no actual token transfer, the memo gets attached to a different leg, and `TIP-1028` doesn't apply. 

Splitting it into two functions means each does one thing clearly. `swapAndSend` always transfers to the recipient, and `swapAndDeposit` doesn't need a memo since nothing is moving externally. 

If `swapAndDeposit` is useful, I think we could consider it for another TIP.

tips/tip-1027.md
danrobinson: Ah interesting idea, but `WithRecipient` and `WithMemo` sounds like the one with the memo doesn't have a recipient, but it does.

An alternative would just be overloading the function with 1-2 arguments and not changing the name. How do we feel about overloading?

tips/tip-1027.md
danrobinson: You're right, let's just have this one revert; the special cases just kinda piled up a bit

tips/tip-1027.md
```diff
@@ -0,0 +1,256 @@
...
+
+19. **Paused token reverts (external recipient)**: If `tokenOut` is paused, `swapExactAmountInTo` to an external recipient reverts during settlement (enforced by TIP-20 `transfer`).
+
+20. **Paused token reverts (DEX-recipient)**: If `tokenOut` is paused, `swapExactAmountInTo(..., DEX_ADDRESS)` reverts. The pause check is enforced by the DEX before crediting internal balance.
+
+### Address-Level Policies (TIP-1028)
+
+21. **Recipient address-level receive policy**: If `recipient` has a whitelist receive policy that does not include the DEX address, the swap-and-send reverts (enforced by TIP-20 `transfer`).
+
+22. **Recipient token set**: If `recipient` has a token set that does not include `tokenOut`, the swap-and-send reverts.
```
0xalpharush, line 254: Is this purposeful or the use of the `ESCROW_ADDRESS` in https://github.com/tempoxyz/tempo/pull/3791 was added later and the tokens should be redirected there?

tips/tip-1027.md
```diff
@@ -0,0 +1,256 @@
...
+
+14. **Exact-out memo on external recipient**: `swapExactAmountOutToWithMemo` emits a `TransferWithMemo` event on `tokenOut` immediately after the `Transfer` event.
+
+15. **Exact-out memo on DEX-recipient**: `swapExactAmountOutToWithMemo(..., DEX_ADDRESS, memo)` follows the same memo rules as the exact-in variant (memo on input transfer if external funding occurs, no memo if fully internal-funded).
+
+### Policy (TIP-403 / TIP-1015)
+
+16. **Policy: blocked recipient**: If `recipient` is blocked by `tokenOut`'s recipient policy, the swap-and-send reverts.
+
+17. **DEX frozen as sender**: If `tokenOut`'s policy blacklists the DEX as a sender, both `swapExactAmountInTo` (external recipient) and `swapExactAmountInTo` (DEX-balance credit) revert.
```
fgimenez, line 242: let's make clear when `swapExactAmountInTo` is called with `DEAX_ADDRESS` as recipient like in other tests
```suggestion
+17. **DEX frozen as sender**: If `tokenOut`'s policy blacklists the DEX as a sender, both `swapExactAmountInTo(..., recipient)` (external recipient) and `swapExactAmountInTo(..., DEX_ADDRESS)` (DEX-balance credit) revert.
```

tips/tip-1027.md
```diff
@@ -0,0 +1,256 @@
...
+## New Functions
+
+Four new swap functions are added to the `IStablecoinDEX` interface — two base functions and two `WithMemo` variants:
+
+```solidity
+/// @notice Swap exact input amount and send output to a specified recipient
+/// @param tokenIn Token to sell
+/// @param tokenOut Token to receive
+/// @param amountIn Amount of tokenIn to sell
+/// @param minAmountOut Minimum amount of tokenOut the recipient must receive
```
fgimenez, line 39:
```suggestion
/// @param minAmountOut Minimum acceptable output amount; the swap reverts if the resulting tokenOut would be less
```

tips/tip-1027.md
```diff
@@ -0,0 +1,256 @@
...
+    uint128 amountIn,
+    uint128 minAmountOut,
+    address recipient
+) external returns (uint128 amountOut);
+
+/// @notice Swap exact input amount and send output to a specified recipient, with memo
+/// @param tokenIn Token to sell
+/// @param tokenOut Token to receive
+/// @param amountIn Amount of tokenIn to sell
+/// @param minAmountOut Minimum amount of tokenOut the recipient must receive
```
fgimenez, line 54:
```suggestion
/// @param minAmountOut Minimum acceptable output amount; the swap reverts if the resulting tokenOut would be less
```

tips/tip-1027.md
```diff
@@ -0,0 +1,256 @@
...
+
+## Implementation Outline
+
+All four new functions MUST execute the same routing, pricing, fill, and input-debit logic as the corresponding existing swap function. The only semantic differences are the output settlement target and memo handling:
+
+- If `recipient == address(StablecoinDEX)`: check policies, credit `balances[msg.sender][tokenOut]`. For `WithMemo` variants, if the input is pulled from the caller's external wallet (fully or partially), use `transferFromWithMemo` for the external portion so the memo is attached to the input leg. If the swap is funded entirely from internal DEX balance, emit a zero-value `Transfer` and `TransferWithMemo` on `tokenIn` to preserve the memo.
+- Otherwise: settle via `ITIP20(tokenOut).transfer(recipient, amountOut)` for base variants, or `ITIP20(tokenOut).transferWithMemo(recipient, amountOut, memo)` for `WithMemo` variants.
+
+The existing `swapExactAmountIn` and `swapExactAmountOut` remain unchanged — they are not refactored to delegate to the new functions, preserving identical behavior for existing callers.
```
fgimenez, line 178: maybe worth moving this to a separate `## Backward Compatibility` section

tips/tip-1027.md
```diff
@@ -0,0 +1,256 @@
...
+    address tokenIn,
+    address tokenOut,
+    uint128 amountOut,
+    uint128 maxAmountIn,
+    address recipient,
+    bytes32 memo
+) external returns (uint128 amountIn);
+```
+
+The existing `swapExactAmountIn` and `swapExactAmountOut` functions are unchanged.
```
fgimenez, line 100: duplicated in https://github.com/tempoxyz/tempo/pull/3016/changes#diff-94b11f7085b7b8623a2fd53a1b70b42339e32ea064d5a0e650ddd6c72b893de0R178 better keep the latter (maybe extracted to separate compatibility section) which placement makes more sense than here and explains no-delegation. 
```suggestion
```

tips/tip-1027.md
```diff
@@ -0,0 +1,256 @@
...
+
+9. **Memo placement — DEX-recipient**: For `WithMemo` variants where `recipient == address(StablecoinDEX)`, the memo MUST be attached to the input transfer (the `tokenIn` transfer from caller to DEX), since no TIP-20 output transfer occurs.
+
+## Test Cases
+
+### Core Swap-and-Send
+
+1. **Basic swap-and-send**: `swapExactAmountInTo` sends output to a third-party address. Verify recipient receives the tokens and caller's balance decreases by `amountIn`.
+
+2. **Swap-and-send exact out**: `swapExactAmountOutTo` sends exact output to recipient. Verify `amountIn <= maxAmountIn`.
```
fgimenez, line 208: those could be merged

tips/tip-1027.md
```diff
@@ -0,0 +1,256 @@
...
+
+5. **Invalid recipient reverts early**: All four new functions revert with `InvalidRecipient()` when `recipient == address(0)` or a TIP-20 token address, **before** any order fills or state mutation.
+
+6. **Multi-hop swap-and-send**: A swap that routes through an intermediate pair delivers the final output to the recipient.
+
+7. **Existing functions unchanged**: `swapExactAmountIn` and `swapExactAmountOut` continue to work identically (gas, events, policy checks) before and after this TIP.
+
+### Memo
+
+8. **Memo on external recipient**: `swapExactAmountInToWithMemo` emits a `TransferWithMemo` event on `tokenOut` with the provided memo, `from = StablecoinDEX, to = recipient`. The `TransferWithMemo` event MUST be emitted immediately after the corresponding `Transfer` event.
```
fgimenez, line 222: could be merged with 14

tips/tip-1027.md
```diff
@@ -0,0 +1,256 @@
...
+
+9. **Memo with bytes32(0)**: `swapExactAmountInToWithMemo(..., bytes32(0))` emits a `TransferWithMemo` event with `memo = bytes32(0)` immediately after the `Transfer` event. This is a valid memo, not "no memo".
+
+10. **No memo on non-memo variants**: `swapExactAmountInTo` and `swapExactAmountOutTo` do NOT emit a `TransferWithMemo` event.
+
+11. **Memo on DEX-recipient (external funding)**: `swapExactAmountInToWithMemo(..., DEX_ADDRESS, memo)` where input is pulled from the caller's external wallet attaches the memo to the input transfer. The `TransferWithMemo` event MUST be emitted immediately after the corresponding input `Transfer` event.
```
fgimenez, line 228: could be merged with 15

tips/tip-1027.md
```diff
@@ -0,0 +1,256 @@
...
+
+16. **Policy: blocked recipient**: If `recipient` is blocked by `tokenOut`'s recipient policy, the swap-and-send reverts.
+
+17. **DEX frozen as sender**: If `tokenOut`'s policy blacklists the DEX as a sender, both `swapExactAmountInTo` (external recipient) and `swapExactAmountInTo` (DEX-balance credit) revert.
+
+18. **DEX-recipient + blocked caller**: `swapExactAmountInTo(..., DEX_ADDRESS)` reverts if `msg.sender` is blocked by `tokenOut`'s recipient policy.
+
+19. **Paused token reverts (external recipient)**: If `tokenOut` is paused, `swapExactAmountInTo` to an external recipient reverts during settlement (enforced by TIP-20 `transfer`).
+
+20. **Paused token reverts (DEX-recipient)**: If `tokenOut` is paused, `swapExactAmountInTo(..., DEX_ADDRESS)` reverts. The pause check is enforced by the DEX before crediting internal balance.
```
fgimenez, line 248: same, could be merged

tips/tip-1027.md
```diff
@@ -0,0 +1,256 @@
...
+
+## Atomicity
+
+If output settlement fails for any reason (policy denial, paused token, invalid recipient, etc.), the entire swap MUST revert — including all prior order fills, orderbook mutations, and balance changes.
+
+## Implementation Outline
+
+All four new functions MUST execute the same routing, pricing, fill, and input-debit logic as the corresponding existing swap function. The only semantic differences are the output settlement target and memo handling:
+
+- If `recipient == address(StablecoinDEX)`: check policies, credit `balances[msg.sender][tokenOut]`. For `WithMemo` variants, if the input is pulled from the caller's external wallet (fully or partially), use `transferFromWithMemo` for the external portion so the memo is attached to the input leg. If the swap is funded entirely from internal DEX balance, emit a zero-value `Transfer` and `TransferWithMemo` on `tokenIn` to preserve the memo.
```
fgimenez, line 175:
```suggestion
- If `recipient == address(StablecoinDEX)`: follow the [Special Case](#special-case-recipient--addressstablecoindex) above (policy checks and internal balance credit) and the [DEX-Recipient (Internal Balance Credit)](#dex-recipient-internal-balance-credit) section under Memo Behavior for memo handling.
```

decofe:
Closing as stale. Please reopen or open a fresh PR if this is still needed.

PR #3024, opened 2026-03-07
feat(precompiles): add no_std support

mattsse:
## Summary

Extends no_std support to `tempo-precompiles`, building on the chainspec no_std work in #3019. The crate now compiles with `--no-default-features --target riscv32imac-unknown-none-elf`.

The main challenge was the `alloy` meta-crate which doesn't support no_std. This was solved by replacing it with individual sub-crates (`alloy-primitives`, `alloy-sol-types`, `alloy-consensus`) and updating all imports across the crate and its proc-macro companion.

Changes:
- Replace `alloy` meta-crate with `alloy-primitives`, `alloy-sol-types`, `alloy-consensus`
- Replace `std::` imports with `core::`/`alloc::` equivalents throughout
- Use `hashbrown` for HashMap/HashSet, `once_cell::sync::Lazy` for LazyLock
- Feature-gate `scoped-tls`, `std::net`, `std::time`, `commonware-cryptography`, `commonware-codec` behind `std`
- Provide no_std stubs for thread-local storage, IP validation, and signature verification
- Update proc-macro generated code to reference `alloy_primitives` instead of `alloy::primitives`
- Add `tempo-precompiles` to `check_no_std.sh`

### DIFF
--- a/.github/scripts/check_no_std.sh
+++ b/.github/scripts/check_no_std.sh
@@ -5,6 +5,7 @@ set -eo pipefail
 # List of crates to check for no_std compatibility.
 # These crates are expected to build without std on bare-metal targets.
 no_std_crates=(
+    tempo-chainspec
     tempo-contracts
     tempo-primitives
 )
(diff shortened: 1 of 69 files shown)


0xrusowsky:
@mattsse i am lacking ctx on the motivation for this change

can u explain why do we need `tempo-precompiles` to be able to run in a `no_std` env?

PR #3025, opened 2026-03-08
ci: run flaky e2e tests exclusively to reduce contention

mattsse:
Several e2e tests flake consistently under resource contention on CI (32-vCPU runner):

- `validator_can_fast_sync_after_full_dkg` fails in ~50% of runs, needing all 3 retries in some cases
- `subblocks_are_included` and `can_restart_after_joining_from_snapshot` flake occasionally

These tests were running concurrently with up to 3 other e2e tests (each reserving 8 of 32 slots) plus integration tests.

This adds a nextest override that reserves all test slots (`threads-required = "num-test-threads"`) for these tests so they run exclusively. Also sets `failure-output = "immediate"` for the CI profile so failures produce debuggable output instead of being suppressed by the existing `failure-output = "never"`.

ref https://github.com/tempoxyz/tempo/actions/runs/22798126108/job/66135699771?pr=3023

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -4,7 +4,8 @@ retries = 2
 test-threads = "num-cpus"
 
 # Heavy e2e tests that flake under resource contention — run exclusively.
-# threads-required = "num-test-threads" means "use all slots", so no other test runs concurrently.
+# threads-required = "num-test-threads" reserves all slots, so no other test runs concurrently.
+# Scheduled last to avoid blocking the rest of the suite.
 [[profile.ci.overrides]]
 filter = """
   package(tempo-e2e) & (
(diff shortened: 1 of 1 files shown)


.config/nextest.toml
```diff
@@ -3,6 +3,19 @@ fail-fast = false
...
+filter = """
+  package(tempo-e2e) & (
+    test(can_restart_after_joining_from_snapshot) |
+    test(validator_can_fast_sync_after_full_dkg) |
+    test(subblocks_are_included)
+  )
+"""
+failure-output = "immediate"
+threads-required = "num-test-threads"
+
```
0xrusowsky, line 20: i think it would also make sense to add `priority = "low"` so that they are scheduled last, and don't block other tests.

usually these we won'y be making changes that break these tests, so i'd rather have them run last

mattsse: > usually these we won'y be making changes that break these tests
unclear what this means

since these run exclusiely now, it doesnt matter when they run

0xrusowsky (approved):

PR #3029, opened 2026-03-09
refactor(payload): extract pool selection stats and metrics helpers

yongkangc:
Moves inline closures for pool-selection and subblock metrics recording
into private methods on `TempoPayloadBuilder`, and groups the three pool
tx counters into a `PoolSelectionStats` struct.

No behavior changes — same metrics, same return types, same side effects.


decofe:
Closing — this incremental cleanup doesn't move the needle enough on its own. Going straight to the full phase pipeline refactor instead.

PR #3035, opened 2026-03-09
refactor(payload): introduce PoolDecision enum to separate policy from execution

decofe:
Reduce duplicated code, simplify code logic and allow us to test `check_pool_tx_admission`. 

changes: 
- Extracts pool transaction admission checks into a `decide_pool_tx` function that returns `PoolDecision::{Execute, Skip, Stop}`. The pool loop now reads as domain logic:

## Testing
`cargo test -p tempo-payload-builder` — all 3 tests pass.

Co-Authored-By: YK <46377366+yongkangc@users.noreply.github.com>

Prompted by: yk

### DIFF
--- a/crates/payload/builder/src/lib.rs
+++ b/crates/payload/builder/src/lib.rs
@@ -62,40 +62,44 @@ use tempo_transaction_pool::{
 };
 use tracing::{Level, debug, debug_span, error, info, instrument, trace, warn};
 
-/// Pre-execution policy decision for a pool transaction.
-///
-/// Separates admission policy from EVM execution: `decide_pool_tx` returns one of these
-/// based on gas budget, block size, and cancellation state, so the main loop reads as
-/// pure domain logic.
-enum PoolDecision {
-    /// Transaction passes all admission checks — execute it.
-    Execute(PoolCandidate),
-    /// Transaction should be skipped for the given reason.
-    Skip(PoolSkipReason),
-    /// Building should stop (interrupted or cancelled).
-    Stop,
-}
-
-/// A pool transaction that passed admission checks and is ready for EVM execution.
-struct PoolCandidate {
-    /// Whether this is a TIP-20 payment transaction.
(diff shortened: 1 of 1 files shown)


decofe (commented):
Yes — logically equivalent, and slightly safer.

The three admission checks (gas-limit, non-payment-gas, osaka-block-size) are **independent**: none reads a value produced by another, so reordering them into a single function doesn't change behavior.

The only observable difference: the original gas-limit error path used plain subtraction (`non_shared_gas_limit - cumulative_gas_used`) which could theoretically underflow in debug builds. This refactor uses `saturating_sub`, which is strictly better.

The Osaka block-size check moved from *after* `let is_payment = ...` to *before* it, but `is_payment` was never used in the size check — it only gates `payment_transactions += 1` and `non_payment_gas_used` accounting, both of which happen post-execution. So moving it earlier is safe.

tl;dr: same semantics, `saturating_sub` is a minor improvement.

crates/payload/builder/src/lib.rs
```diff
@@ -467,75 +550,37 @@ where
...
-                        limit: MAX_RLP_BLOCK_SIZE,
-                    },
-                );
-                pool_transactions_skipped += 1;
-                self.metrics
-                    .pool_transactions_skipped_oversized_block
-                    .increment(1);
-                continue;
-            }
-
```
yongkangc, line 530: it's moved into `check_pool_tx_admission()`

crates/payload/builder/src/lib.rs
```diff
@@ -70,6 +114,46 @@ fn has_expired_transactions(subblock: &RecoveredSubBlock, timestamp: u64) -> boo
     })
 }
 
+/// Returns `Some(reason)` if the transaction should be skipped, `None` if it can execute.
+fn check_pool_tx_admission(
```
yongkangc, line 117: this now allows us to test the logic

PR #3042, opened 2026-03-09
feat(ext): introduce tempo cli extensions

Slokh:
Adds `crates/ext/`, the extension system for the Tempo CLI. Extensions are standalone binaries (e.g., `tempo-wallet`, `tempo-mpp`) that are discovered, installed, and updated automatically — users just run `tempo wallet` and the right binary is located or fetched on demand.

## How it works

When a user runs `tempo <name>`, the launcher searches for a `tempo-<name>` binary next to the running executable, in `~/.local/bin`, and on `PATH`. If not found, it fetches the signed release manifest from `https://cli.tempo.xyz/extensions/tempo-<name>/manifest.json`, downloads and verifies the binary, and dispatches the command. Installed extensions are auto-updated in the background at most once every 6 hours.

## Management commands

- `tempo add <ext> [version]` — install an extension. Specifying a version pins it (auto-update will notify but not install newer versions).
- `tempo update [ext]` — without arguments, updates tempo itself via `tempoup` then updates all installed extensions. With an extension name, updates that single extension and unpins it.
- `tempo remove <ext>` — uninstalls the binary and any associated agent skill files.
- `tempo list` — shows installed extensions with their version, pinned status, and description.

## Release manifest

Extensions are published as a JSON manifest at a well-known URL:

```
https://cli.tempo.xyz/extensions/tempo-{name}/manifest.json            # latest
https://cli.tempo.xyz/extensions/tempo-{name}/v{version}/manifest.json # pinned
```

Schema:

```json
{
  "version": "1.2.0",
  "description": "Manage your Tempo Wallet",
  "binaries": {
    "tempo-wallet-darwin-arm64": {
      "url": "https://cdn.example.com/tempo-wallet-darwin-arm64",
      "sha256": "b94d27b9...",
      "signature": "untrusted comment: ...\nRWT..."
    }
  },
  "skill": "https://cdn.example.com/SKILL.md",
  "skill_sha256": "e3b0c442...",
  "skill_signature": "untrusted comment: ...\nRWT..."
}
```

Binary keys follow `tempo-{name}-{os}-{arch}` (`darwin`/`linux`, `arm64`/`amd64`). The `description`, `skill`, `skill_sha256`, and `skill_signature` fields are optional. Signatures are minisign signature boxes (multi-line, base64-encoded).

## Security

Every binary and skill file is verified before installation:

1. **URL scheme enforcement** — download URLs must be `https://` or `file://`; `http://` and other schemes are rejected.
2. **SHA-256 checksum** — verified before signature check to reject corrupted downloads cheaply.
3. **Minisign signature** — verified against a public key compiled into the binary. The key can only be overridden via `TEMPO_EXT_PUBLIC_KEY` in debug/test builds (`#[cfg(debug_assertions)]`).
4. **Trusted comment anti-substitution** — after signature verification, the trusted comment is checked against the expected artifact identity (`file:tempo-<name>-<os>-<arch>` for binaries, `skill:tempo-<name>` for skills) to prevent cross-extension substitution attacks.
5. **Downgrade prevention** — `tempo update` uses semver comparison and only installs if the manifest version is strictly newer. Non-semver versions fall back to string equality.
6. **Atomic installs** — binaries are written to a unique temp file, permissions are set via the open file handle (no TOCTOU), and `persist()` atomically moves into place.

## Local Registry

Installed versions, update-check timestamps, pinned status, and descriptions are persisted to `$TEMPO_HOME/extensions.json`. The registry is not file-locked — concurrent writes are last-writer-wins, which is acceptable since the data is limited to timestamps and version strings (worst case: a redundant update check). The registry is also used by auto-update to track which extensions need checking and to skip pinned extensions. Auto-update is disabled when `TEMPO_HOME` is set to avoid interfering with managed or test environments.

## Agent skills

Extensions can optionally ship a `SKILL.md` that gets installed into all detected coding agent directories (Claude Code, Amp, Cursor, Copilot, Gemini CLI, Codex, Goose, Windsurf, Continue, Roo, Kiro, Augment, Trae) so the extension's capabilities are available to AI agents.

## Testing

```
cargo build -p tempo

# Show installed extensions
./target/debug/tempo list

# Add extension implicitly
./target/debug/tempo wallet --help

# Verify installation
./target/debug/tempo list

# Add extension explicitly
./target/debug/tempo add request

# Update extension
./target/debug/tempo update request

<can run any wallet / request commands>

# Remove extensions
./target/debug/tempo remove wallet
./target/debug/tempo remove request

# Verify removal
./target/debug/tempo list
```

### DIFF
--- a/bin/tempo/src/main.rs
+++ b/bin/tempo/src/main.rs
@@ -135,7 +135,10 @@ fn print_extensions_footer() {
         return;
     }
 
-    let extensions = tempo_ext::installed_extensions();
+    let extensions = match tempo_ext::installed_extensions() {
+        Ok(e) => e,
+        Err(_) => return,
+    };
     if extensions.is_empty() {
         return;
     }
--- a/crates/ext/src/installer/mod.rs
+++ b/crates/ext/src/installer/mod.rs
@@ -27,13 +27,14 @@ use tempfile::TempDir;
 
 const HTTP_TIMEOUT: Duration = Duration::from_secs(30);
 
+/// Builds an HTTP client with a 30-second timeout for manifest/binary fetches.
 pub(super) fn http_client() -> Result<reqwest::blocking::Client, InstallerError> {
     Ok(reqwest::blocking::Client::builder()
         .timeout(HTTP_TIMEOUT)
         .build()?)
 }
 
-/// Parse a `file://` URL into a local path using the `url` crate.
+/// Parses a `file://` URL into a local path using the `url` crate.
 ///
 /// Returns `None` if the URL isn't a valid `file://` URL or can't be
 /// converted to a platform path (e.g. `file://remote/share` on Windows).
@@ -44,44 +45,53 @@ pub(super) fn file_url_to_path(url_str: &str) -> Option<PathBuf> {
         .and_then(|u| u.to_file_path().ok())
 }
 
+/// Where to fetch and verify an extension release from.
 #[derive(Debug, Clone)]
 pub(crate) struct InstallSource {
+    /// URL (or `file://` path) of the signed release manifest JSON.
     pub(crate) manifest: Option<String>,
+    /// Base64-encoded minisign public key for signature verification.
     pub(crate) public_key: Option<String>,
 }
 
+/// Handles downloading, verifying, and placing extension binaries.
 #[derive(Debug, Clone)]
 pub(crate) struct Installer {
+    /// Directory where extension binaries are installed.
     pub(crate) bin_dir: PathBuf,
 }
 
+/// Returned after a successful install with the version and description
+/// from the release manifest.
 #[derive(Debug, Clone)]
 pub(crate) struct InstallResult {
+    /// Installed version string (e.g. `"1.0.0"`).
     pub(crate) version: String,
+    /// Short description from the release manifest.
     pub(crate) description: String,
 }
 
+/// Fully resolved install plan: all paths and verification material ready.
 #[derive(Debug)]
 struct ResolvedInstall {
+    /// Version from the release manifest.
     version: String,
+    /// Short description from the release manifest.
     description: String,
     /// Path to the downloaded binary. `None` in dry-run mode.
     src: Option<PathBuf>,
+    /// Final destination path in `bin_dir`.
     dst: PathBuf,
+    /// Optional URL for the extension's Claude Code skill file.
     skill_url: Option<String>,
+    /// Expected SHA-256 hex digest of the skill file.
     skill_sha256: Option<String>,
+    /// Base64-encoded minisign signature of the skill file.
     skill_signature: Option<String>,
+    /// Decoded public key used for signature verification.
     public_key: PublicKey,
+    /// Temp directory holding the download; kept alive until install completes.
     _download_dir: TempDir,
-}
-
-/// The fallback install directory: `TEMPO_HOME/bin` if set, else `~/.local/bin`.
-pub(crate) fn fallback_bin_dir() -> Option<PathBuf> {
-    if let Some(home) = env::var_os("TEMPO_HOME") {
-        Some(PathBuf::from(home).join("bin"))
-    } else {
-        default_local_bin().ok()
-    }
 }
 
 impl Installer {
@@ -97,7 +107,7 @@ impl Installer {
         Ok(Self { bin_dir })
     }
 
-    /// Install an extension and return the installed version and description.
+    /// Installs an extension and returns the installed version and description.
     pub(crate) fn install(
         &self,
         extension: &str,
@@ -108,7 +118,7 @@ impl Installer {
         self.install_inner(extension, source, None, dry_run, quiet)
     }
 
-    /// Check if a newer version is available without installing.
+    /// Checks if a newer version is available without installing.
     /// Returns `Some(latest_version)` if the manifest version is strictly
     /// newer, `None` if already up to date.
     pub(crate) fn check_latest_version(
@@ -131,7 +141,7 @@ impl Installer {
         }
     }
 
-    /// Install an extension only if the manifest version is newer than
+    /// Installs an extension only if the manifest version is newer than
     /// `installed_version`. Returns `Some(result)` if an update was
     /// performed, `None` if already at the latest version.
     pub(crate) fn install_if_changed(
@@ -157,6 +167,7 @@ impl Installer {
         Ok(Some(result))
     }
 
+    /// Shared implementation for `install` and `install_if_changed`.
     fn install_inner(
         &self,
         extension: &str,
@@ -190,6 +201,7 @@ impl Installer {
         Ok(result)
     }
 
+    /// Removes an extension's binary and skill files.
     pub(crate) fn remove(&self, extension: &str, dry_run: bool) -> Result<(), InstallerError> {
         let binary = format!("tempo-{extension}");
         self.remove_binary(&binary, dry_run)?;
@@ -197,6 +209,7 @@ impl Installer {
         Ok(())
     }
 
+    /// Fetches the manifest, downloads the binary, and verifies checksums/signatures.
     fn resolve_install(
         &self,
         extension: &str,
@@ -263,6 +276,7 @@ impl Installer {
         })
     }
 
+    /// Atomically places the downloaded binary at its destination path.
     fn copy_binary(
         &self,
         resolved: &ResolvedInstall,
@@ -307,6 +321,7 @@ impl Installer {
         Ok(())
     }
 
+    /// Deletes the named binary from `bin_dir`.
     fn remove_binary(&self, binary: &str, dry_run: bool) -> Result<(), InstallerError> {
         let path = self.bin_dir.join(executable_name(binary));
 
@@ -320,6 +335,7 @@ impl Installer {
         Ok(())
     }
 
+    /// Creates `bin_dir` if it doesn't exist and verifies it is writable.
     fn ensure_dirs(&self, dry_run: bool) -> Result<(), InstallerError> {
         if dry_run {
             println!("dry-run: ensure dir {}", self.bin_dir.display());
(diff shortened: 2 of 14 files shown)


crates/ext/src/launcher.rs
```diff
@@ -0,0 +1,1042 @@
...
+        Commands::Extension(ext_args) => launcher.handle_extension(ext_args),
+    }
+}
+
+/// Run `tempoup` to update the tempo binary itself.
+///
+/// Passes `TEMPO_BIN_DIR` so tempoup installs into the same directory as the
+/// running binary. If tempoup is not found on `PATH`, it is installed first
+/// via `https://tempo.xyz/install`.
+fn run_tempoup(bin_dir: &Path) -> Result<bool, LauncherError> {
```
Slokh, line 166: intentionally not removing `tempoup` as part of this change to keep scope isolated. can remove it in a follow-up since we need to update more references in docs and stuff

crates/ext/src/registry.rs
```diff
@@ -0,0 +1,270 @@
...
+            } else {
+                // XDG_STATE_HOME, defaulting to ~/.local/state
+                env::var_os("XDG_STATE_HOME")
+                    .map(PathBuf::from)
+                    .unwrap_or_else(|| home.join(".local/state"))
+            };
+            base.join("tempo").join("extensions.json")
+        })
+    }
+}
```
klkvr, line 172: nit: reth is using `dirs_next` for that https://github.com/paradigmxyz/reth/blob/2580304b417a67870de1a984ca66e8b07bffc2d7/crates/node/core/src/dirs.rs#L17-L22

crates/ext/src/registry.rs
```diff
@@ -0,0 +1,270 @@
...
+    pub(crate) fn load() -> Self {
+        let path = match state_path() {
+            Some(p) => p,
+            None => return Self::default(),
+        };
+        match fs::read_to_string(&path) {
+            Ok(content) => serde_json::from_str(&content).unwrap_or_default(),
+            Err(_) => Self::default(),
+        }
+    }
```
klkvr: i believe serde and fs errors might be worth propagating here? also could simplify with `let Some(_) else { return }` syntax

crates/ext/src/registry.rs
```diff
@@ -0,0 +1,270 @@
...
+    pub(crate) installed_version: String,
+    /// When true, auto-update will not install newer versions — only
+    /// log that an update is available. Set when the user installs a
+    /// specific version via `tempo add <ext> <version>`.
+    #[serde(default)]
+    pub(crate) pinned: bool,
+    /// Short description from the release manifest.
+    #[serde(default)]
+    pub(crate) description: String,
+}
```
klkvr, line 41: can we document all fields here?

crates/ext/src/installer/mod.rs
klkvr, line 1: let's document all of this

crates/ext/src/installer/mod.rs
```diff
@@ -0,0 +1,512 @@
...
+/// The fallback install directory: `TEMPO_HOME/bin` if set, else `~/.local/bin`.
+pub(crate) fn fallback_bin_dir() -> Option<PathBuf> {
+    if let Some(home) = env::var_os("TEMPO_HOME") {
+        Some(PathBuf::from(home).join("bin"))
+    } else {
+        default_local_bin().ok()
+    }
+}
+
+impl Installer {
```
klkvr, line 97: should be at the top of the file right after Installer struct definition

crates/ext/src/launcher.rs
```diff
@@ -0,0 +1,1057 @@
...
+    #[arg(long)]
+    dry_run: bool,
+}
+
+pub fn run<I, T>(args: I) -> Result<i32, LauncherError>
+where
+    I: IntoIterator<Item = T>,
+    T: Into<OsString> + Clone,
+{
+    let exe_dir = env::current_exe()
```
klkvr: can we move this to the top of the file? maybe even to `lib.rs`

crates/ext/src/registry.rs
Slokh: Yeah, updating to exit on error with hint to remove file

crates/ext/src/launcher.rs
```diff
@@ -0,0 +1,1057 @@
...
+    let status = match Command::new("tempoup")
+        .env("TEMPO_BIN_DIR", bin_dir)
+        .status()
+    {
+        Ok(s) => s,
+        Err(err) if err.kind() == std::io::ErrorKind::NotFound => {
+            println!("tempoup not found, installing...");
+            let install_status = Command::new("sh")
+                .arg("-c")
+                .arg("curl -fsSL https://tempo.xyz/install | bash")
```
decofe, line 176: 🚨 **[SECURITY — High]** Unsigned `tempoup` Bootstrap Executes Untrusted Script Content

When `tempoup` is not found, this falls back to `sh -c "curl -fsSL https://tempo.xyz/install | bash"`, executing whatever script bytes are returned with no signature or checksum validation. This bypasses the minisign-based trust model enforced everywhere else in `crates/ext`. An attacker who compromises the CDN/origin or performs TLS interception gains arbitrary code execution.

**Recommended Fix:** Replace `curl | bash` with a signed artifact flow consistent with the extension installer: download a binary/manifest, verify a minisign signature (with trusted comment binding), then execute only verified bytes.

crates/ext/src/installer/mod.rs
```diff
@@ -0,0 +1,512 @@
...
+        semver::Version::parse(
+            manifest_version
+                .strip_prefix('v')
+                .unwrap_or(manifest_version),
+        ),
+    ) {
+        manifest_v > installed_v
+    } else {
+        // Non-semver fallback: only skip if identical.
+        installed != manifest_version
```
decofe, line 457: 🚨 **[SECURITY — Medium]** Version Downgrade via Stale Manifest Replay for Non-Semver Versions

The non-semver fallback `installed != manifest_version` is symmetric — it permits downgrades as well as upgrades. A stale or replayed manifest with an older version string (e.g. `nightly-2025-01-01` vs installed `nightly-2025-03-09`) is treated as "newer" because the strings differ. Since old binaries remain validly signed, signature verification does not prevent rollback.

**Recommended Fix:** Use lexicographic `>` comparison if non-semver versions are chronologically sortable (like `nightly-YYYY-MM-DD`), or require explicit user confirmation for non-semver version changes.

bin/tempo/src/main.rs
```diff
@@ -145,12 +178,38 @@ fn main() -> eyre::Result<()> {
...
+    let mut cli = match TempoCli::command()
+        .about("Tempo")
+        .try_get_matches_from(std::env::args_os())
+        .and_then(|matches| TempoCli::from_arg_matches(&matches))
+    {
+        Ok(cli) => cli,
+        Err(err) => {
+            if err.kind() == clap::error::ErrorKind::InvalidSubcommand {
+                // Unknown subcommand — try the extension launcher.
+                let code = match tempo_ext::run(std::env::args_os()) {
```
decofe, line 193: 🚨 **[SECURITY — Medium]** Nested Built-In Typos Can Trigger External Binary Execution

This unconditionally routes every `InvalidSubcommand` error to `tempo_ext::run`, but Reth's CLI includes built-in commands with nested subcommand trees (`db`, `stage`, `p2p`). A typo under a valid nested command (e.g. `tempo db <typo>`) also produces `InvalidSubcommand`, causing the first token (`db`) to be interpreted as an extension name. If `tempo-db` exists in extension directories or `PATH`, it executes instead of showing a CLI error.

**Recommended Fix:** Guard the fallback by checking that the unknown token is not a known built-in top-level command name before delegating to `tempo_ext::run`.

crates/ext/src/registry.rs
```diff
@@ -0,0 +1,270 @@
...
+            let _ = fs::create_dir_all(parent);
+        }
+        let json = match serde_json::to_string_pretty(self) {
+            Ok(j) => j,
+            Err(err) => {
+                tracing::warn!("registry serialize failed: {err}");
+                return;
+            }
+        };
+        let tmp = path.with_extension("tmp");
```
decofe, line 89: ⚠️ **[ISSUE — Low]** Concurrent Registry Saves Use Hardcoded Temp File

`path.with_extension("tmp")` produces the same temp path for all concurrent CLI invocations. Parallel launches (e.g. scripts running `tempo wallet` concurrently) can race on this file, potentially corrupting the registry JSON. On parse failure, `Registry::load()` silently falls back to `Self::default()`, wiping state including version pins.

**Recommended Fix:** Use `tempfile::NamedTempFile` (or `tempfile::Builder`) to create unique temp files in the same directory, then `persist()` to the target path.

decofe (commented):
:eye: **Cyclops Review — PR #3042**

This PR adds a CLI extension manager (`tempo-ext`) with minisign signature verification, manifest URL restrictions, and trusted-comment binding. The extension trust model is well-designed overall. Four actionable findings were identified across the new code.

**Findings summary:**
| # | Tier | Title | File |
|---|------|-------|------|
| 1 | 🚨 SECURITY (High) | Unsigned `tempoup` bootstrap executes untrusted script content | `crates/ext/src/launcher.rs:169-172` |
| 2 | 🚨 SECURITY (Medium) | Version downgrade via stale manifest replay for non-semver versions | `crates/ext/src/installer/mod.rs:428-431` |
| 3 | 🚨 SECURITY (Medium) | Nested built-in typos can trigger external binary execution | `bin/tempo/src/main.rs:188-197` |
| 4 | ⚠️ ISSUE (Low) | Concurrent registry saves use hardcoded temp file | `crates/ext/src/registry.rs:67-72` |

<details>

- ⚡ **Update Trust Boundary**: `run_tempoup` is the only code path that executes remote content without cryptographic verification — all other install paths enforce minisign signatures.
- ⚡ **CLI Error Routing Boundary**: Re-check all parse error kinds intercepted in `main()` and ensure only intentional extension entrypoints are redirected. Consider maintaining an explicit built-in command name list for the guard.
- ⚡ **Extension `is_newer` Logic**: The non-semver fallback is also exercised by `check_latest_version`, so the downgrade gap affects both explicit `tempo update` and periodic auto-updates.
- ⚡ **Global CLI Flags**: `tempo_ext::run` receives raw `args_os()` and uses a strict clap parser — global flags like `--log.level` placed before the extension name will cause `UnknownArgument` errors.
</details>

crates/ext/src/launcher.rs
Slokh: invalid, this is our normal install steps

crates/ext/src/installer/mod.rs
Slokh: We aren't using non-semver for extensions

bin/tempo/src/main.rs
Slokh: We don't want to do this since we want to easily be able to launch new extensions later. The s3 bucket is one we own so will only have safe binaries.

crates/ext/src/registry.rs
Slokh: Doesn't really matter since registry just tracks current version

PR #3044, opened 2026-03-09
chore: bump CW to 2026.3.0

hamdiallam:


crates/commonware-node/src/dkg/manager/actor/state.rs
```diff
@@ -143,7 +143,7 @@ where
         self.events
             .append(
                 section,
-                Event::Ack {
+                &Event::Ack {
```
SuperFluffy, line 146: That's so silly - we can pass the argument by reference, but still need to clone the contents. Some view type would be nice. 🫠

PR #3045, opened 2026-03-09
feat(consensus): add --consensus.enable-subblocks flag

decofe:
Adds `--consensus.enable-subblocks` CLI flag (default: `false`) to control whether the node participates in subblock processing.

Co-Authored-By: Arsenii Kulikov <62447812+klkvr@users.noreply.github.com>

Prompted by: klkvr

### DIFF
--- a/crates/commonware-node/src/consensus/engine.rs
+++ b/crates/commonware-node/src/consensus/engine.rs
@@ -624,6 +624,8 @@ where
 
         if let Some(subblocks) = self.subblocks {
             tasks.push(self.context.spawn(|_| subblocks.run(subblocks_channel)));
+        } else {
+            drop(subblocks_channel);
         }
 
         try_join_all(tasks)


crates/commonware-node/src/lib.rs
```diff
@@ -112,6 +112,7 @@ pub async fn run_consensus_stack(
         time_to_build_subblock: config.time_to_build_subblock.into_duration(),
         subblock_broadcast_interval: config.subblock_broadcast_interval.into_duration(),
         fcu_heartbeat_interval: config.fcu_heartbeat_interval.into_duration(),
+        enable_subblocks: !config.no_subblocks,
```
SuperFluffy:
```suggestion
        with_subblocks: !config.no_subblocks,
```

crates/commonware-node/src/subblocks.rs
```diff
@@ -155,6 +160,34 @@ impl<TContext: Spawner + Metrics + Pacer> Actor<TContext> {
             impl Receiver<PublicKey = PublicKey>,
         ),
     ) {
+        if !self.enable_subblocks {
+            info!("subblocks are disabled");
```
SuperFluffy: event outside of scope

SuperFluffy: The actor should just not run if subblocks are disabled.

crates/commonware-node/src/subblocks.rs
```diff
@@ -155,6 +160,34 @@ impl<TContext: Spawner + Metrics + Pacer> Actor<TContext> {
...
+                    biased;
+
+                    Some(action) = self.actions_rx.next() => {
+                        // Only handle GetSubBlocks (return empty), ignore everything else.
+                        if let Message::GetSubBlocks { response, .. } = action {
+                            let _ = response.send(Vec::new());
+                        }
+                    },
+                    // Drain subblock transactions to prevent channel backpressure.
+                    result = self.subblock_transactions_rx.recv() => {
```
SuperFluffy: Likewise here - we shouldn't do work if the feature is disabled.

crates/commonware-node/src/subblocks.rs
```diff
@@ -98,7 +98,6 @@ pub(crate) struct Actor<TContext> {
     subblock_broadcast_interval: Duration,
     /// The epoch strategy used by tempo.
     epoch_strategy: FixedEpocher,
-
```
klkvr, line 101: revert

crates/commonware-node/src/lib.rs
```diff
@@ -80,7 +80,11 @@ pub async fn run_consensus_stack(
     );
     let marshal = network.register(MARSHAL_CHANNEL_IDENT, backfill_quota, message_backlog);
     let dkg = network.register(DKG_CHANNEL_IDENT, DKG_LIMIT, message_backlog);
-    let subblocks = network.register(SUBBLOCKS_CHANNEL_IDENT, SUBBLOCKS_LIMIT, message_backlog);
+    let subblocks = if config.no_subblocks {
+        None
+    } else {
+        Some(network.register(SUBBLOCKS_CHANNEL_IDENT, SUBBLOCKS_LIMIT, message_backlog))
+    };
```
klkvr: use `bool::then`

crates/commonware-node/src/epoch/manager/actor.rs
```diff
@@ -336,10 +336,16 @@ where
...
+                reporter: match self.config.subblocks.clone() {
+                    Some(subblocks) => Reporters::from((
+                        subblocks,
+                        Reporters::from((self.config.marshal.clone(), self.config.feed.clone())),
+                    )),
+                    None => Reporters::from((
+                        None::<crate::subblocks::Mailbox>,
+                        Reporters::from((self.config.marshal.clone(), self.config.feed.clone())),
+                    )),
+                },
```
klkvr: `Reporters` supports `From<(Option<R1>, R2)>`

crates/commonware-node/src/lib.rs
```diff
@@ -80,7 +80,8 @@ pub async fn run_consensus_stack(
     );
     let marshal = network.register(MARSHAL_CHANNEL_IDENT, backfill_quota, message_backlog);
     let dkg = network.register(DKG_CHANNEL_IDENT, DKG_LIMIT, message_backlog);
-    let subblocks = network.register(SUBBLOCKS_CHANNEL_IDENT, SUBBLOCKS_LIMIT, message_backlog);
+    let subblocks = (!config.no_subblocks)
```
SuperFluffy: I think this one is dangerous - IIRC my conversations with commonware, you *must* have this subchannel registered, even if you don't listen on it. Otherwise you will consider those actors that send you messages over unregistered subchannels as Byzantine. This should be double-checked!

crates/commonware-node/src/epoch/manager/actor.rs
```diff
@@ -336,7 +336,7 @@ where
                 blocker: self.config.blocker.clone(),
                 automaton: self.config.application.clone(),
                 relay: self.config.application.clone(),
-                reporter: Reporters::from((
+                reporter: Reporters::<_, crate::subblocks::Mailbox, _>::from((
```
SuperFluffy, line 339: Wish this API was neater.

crates/commonware-node/src/lib.rs
SuperFluffy: Indeed: peers that send on unregistered channels are considered Byzantine and will be blocked: https://github.com/commonwarexyz/monorepo/blob/314802f2c3e16fde253973598ece1db9000ead9a/p2p/src/authenticated/lookup/actors/peer/actor.rs#L218-L222

When fixing this, please add a comment why we still register the channel even if we don't have an actor for it.

crates/commonware-node/src/args.rs
```diff
@@ -120,6 +120,13 @@ pub struct Args {
     )]
     pub minimum_time_before_propose: PositiveDuration,
 
+    /// Disable subblock processing.
+    ///
+    /// When set, the node will not build or broadcast subblocks, and will
+    /// ignore any incoming subblocks from the network.
+    #[arg(long = "consensus.no-subblocks", default_value_t = false)]
+    pub no_subblocks: bool,
```
0xKitsune: Should subblocks be disabled by default?

crates/commonware-node/src/args.rs
```diff
@@ -120,6 +120,13 @@ pub struct Args {
     )]
     pub minimum_time_before_propose: PositiveDuration,
 
+    /// Disable subblock processing.
+    ///
+    /// When set, the node will not build or broadcast subblocks, and will
+    /// ignore any incoming subblocks from the network.
+    #[arg(long = "consensus.no-subblocks", default_value_t = false)]
```
klkvr:
```suggestion
    #[arg(long = "consensus.no-subblocks", default_value_t = true)]
```

crates/commonware-node/src/lib.rs
```diff
@@ -80,9 +80,7 @@ pub async fn run_consensus_stack(
     );
     let marshal = network.register(MARSHAL_CHANNEL_IDENT, backfill_quota, message_backlog);
     let dkg = network.register(DKG_CHANNEL_IDENT, DKG_LIMIT, message_backlog);
-    let subblocks = config
-        .enable_subblocks
-        .then(|| network.register(SUBBLOCKS_CHANNEL_IDENT, SUBBLOCKS_LIMIT, message_backlog));
+    let subblocks = network.register(SUBBLOCKS_CHANNEL_IDENT, SUBBLOCKS_LIMIT, message_backlog);
```
SuperFluffy, line 86: Comment gone again?

crates/commonware-node/src/consensus/engine.rs
```diff
@@ -621,12 +620,17 @@ where
...
-            subblocks,
-        ])
-        .await
-        .map(|_| ())
-        // TODO: look into adding error context so that we know which
-        // component failed.
-        .wrap_err("one of the consensus engine's actors failed")
+        ];
+
+        if let Some(subblocks) = self.subblocks {
```
decofe, line 625: 🚨 **[SECURITY]** Disabled Subblocks Path Allows Remote Memory Exhaustion

When `with_subblocks` is `false`, no task consumes `subblocks_channel`, but the receiver stays alive for the process lifetime. Upstream `commonware_p2p` allocates a bounded `mpsc::channel(backlog)` per registered channel and enqueues `(peer, IoBuf)` via `try_send`, only dropping when full. With defaults (`message_backlog=16_384`, max msg ~8 MiB), an authenticated peer can pin ~128 GiB in this unpolled queue, causing OOM.

**Recommended Fix:** Drop the `subblocks_channel` receiver immediately when subblocks are disabled, or spawn a lightweight discard task. Dropping is simplest — `try_send` returns `TrySendError::Closed` which the router already handles safely:

```rust
if let Some(subblocks) = self.subblocks {
    tasks.push(self.context.spawn(|_| subblocks.run(subblocks_channel)));
} else {
    drop(subblocks_channel);
}
```

PR #3048, opened 2026-03-09
chore(txpool): move keychain validation after amm checks

0xKitsune:
This PR moves `validate_against_keychain()` after amm checks to move expensive checks later in the validation path.

### DIFF
--- a/crates/alloy/src/network.rs
+++ b/crates/alloy/src/network.rs
@@ -167,6 +167,7 @@ impl TransactionBuilder<TempoNetwork> for TempoTransactionRequest {
             || self.key_id.is_some()
             || self.valid_before.is_some()
             || self.valid_after.is_some()
+            || self.fee_payer_signature.is_some()
         {
             TempoTxType::AA
         } else {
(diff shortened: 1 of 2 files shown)


crates/transaction-pool/src/validator.rs
```diff
@@ -651,18 +660,13 @@ where
...
-                    transaction,
-                    InvalidPoolTransactionError::other(err),
-                );
-            }
-            Err(err) => {
-                return TransactionValidationOutcome::Error(*transaction.hash(), Box::new(err));
-            }
+        // Validate keychain signature versions early so that permanently invalid
+        // errors before cheaper economic checks that would mask them.
+        if let Err(err) = self.validate_keychain_version(&transaction, spec) {
```
legion2002, line 665: It makes sense to move these checks separately, to validate versions early. Just calling out that there maybe a footgun here, where people assume that `validate_against_keychain` would do this automatically.

This is not a problem for now since this function is only used by `validate_one` , and has proper documentation.

crates/transaction-pool/src/validator.rs
```diff
@@ -834,6 +838,20 @@ where
             }
         }
 
+        // Validate transactions that involve keychain keys.
+        match self.validate_against_keychain(&transaction, &mut state_provider) {
```
decofe, line 842: :rotating_light: **[SECURITY]** Mempool DoS Amplification via Deferred Keychain Authentication

**Severity:** High

Moving `validate_against_keychain` here — after `ensure_aa_intrinsic_gas` (EC recoveries), fee-token validation (~4 SLOADs), balance checks, and AMM liquidity checks — means unauthenticated keychain transactions can spoof any `user_address` and force the node through all these expensive operations before being rejected.

`TempoSignature::recover_signer` for keychain signatures returns the embedded `user_address` without verifying on-chain authorization (that's solely this function's job). The eventual rejection maps to `is_bad_transaction() == false`, so the attacker's peer is **never penalized** and can repeat indefinitely.

**Recommended Fix:** Move `validate_against_keychain` back to immediately after `validate_keychain_version` (around line 670), before `ensure_aa_intrinsic_gas` and all state-dependent economic checks. Authentication must precede any operations that trust the claimed sender identity.

PR #3051, opened 2026-03-09
TIP-1028: Address-Level Receive Policies

malleshpai:
Extends TIP-403 with token sets and address-level receive policies.

### DIFF
--- a/tips/tip-1028.md
+++ b/tips/tip-1028.md
@@ -495,7 +495,7 @@ DEX swaps (`swapExactAmountIn`, `swapExactAmountOut`) transfer the output token 
 
 Order fills (`partial_fill_order`, `fill_order`) credit tokens to the maker's *DEX internal balance* via `increment_balance`, which does not trigger TIP-20 transfers. Address-level controls are NOT checked on internal balance credits. Controls are checked when the maker later calls `withdraw` to move tokens out of the DEX to their wallet (see below).
 
-> **DEX Internal Balance Bypass**: This is an intentional design choice, not a loophole. The DEX's internal accounting is an escrow mechanism — tokens are locked in the DEX contract, not delivered to the maker's wallet. Address-level controls govern what enters an address's wallet, and the `withdraw` path enforces this. Attempting to enforce controls on internal balance credits would break order matching (the maker is not the caller during fills) and would be redundant with the `withdraw` check.
+> **DEX Internal Balance Bypass**: This is an intentional design choice, not a loophole. The DEX's internal accounting is an escrow mechanism — tokens are locked in the DEX contract, not delivered to the maker's wallet. Address-level controls govern what enters an address's wallet, and the `withdraw` path enforces this. Attempting to enforce controls on internal balance credits would break order matching.
 
 ### Order Placement
 


tips/tip-1028.md
```diff
@@ -0,0 +1,734 @@
...
+
+Burning (`burn`, `burnBlocked`) does not deliver tokens to a new recipient, so address-level controls are not applicable. The existing behavior is unchanged.
+
+### Self-Transfers
+
+Self-transfers (`from == to`) are treated as normal transfers and are subject to the receive policy check. If an address sets a whitelist receive policy, it SHOULD include itself if self-transfers are needed.
+
+### Fee Transfers
+
+Fee transfers via `transferFeePreTx` and `transferFeePostTx` bypass address-level policy checks since they are system operations that directly manipulate balances without going through `ensureTransferAuthorized`. The fee collection path (`collect_fee_pre_tx`) checks token-level `ensureTransferAuthorized(fee_payer, FeeManager)` but does NOT check address-level controls on the FeeManager address — this is correct because FeeManager is a system precompile, not an address that opts into receive policies.
```
danrobinson, line 475: This contradicts line 595

tips/tip-1028.md
```diff
@@ -0,0 +1,734 @@
...
+        bool inSet = tokenSetMembers[tokenSet][token];
+        bool authorized = (tokenSetType == PolicyType.WHITELIST) ? inSet : !inSet;
+        if (!authorized) return false;
+    }
+    
+    return true;
+}
+```
+
+#### isAddressMintAuthorized
```
danrobinson: Why doesn't this just use `isAddressTransferAuthorized`?

tips/tip-1028.md
```diff
@@ -0,0 +1,734 @@
...
+
+```solidity
+mapping(address => uint256) public addressPolicies;
+```
+
+The `addressPolicies` mapping packs a presence flag, policy IDs, token set IDs, and their cached types into a single 256-bit storage slot:
+
+| Bits (inclusive) | Size | Field | Description |
+|------------------|------|-------|-------------|
+| 0 | 1 bit | `hasReceivePolicy` | 1 if address has set receive policies, 0 if not |
```
danrobinson: Perhaps let's make this bit 255 rather than bit 0? That makes the layout of these other fields cleaner

tips/tip-1028.md
```diff
@@ -0,0 +1,734 @@
...
+| 1–64 | 64 bits | `receivePolicyId` | TIP-403 policy checked when this address is the receiver |
+| 65–72 | 8 bits | `receivePolicyType` | Cached type of receive policy (0 = whitelist, 1 = blacklist) |
+| 73–136 | 64 bits | `tokenSetId` | TIP-403 token set checked on tokens being received |
+| 137–144 | 8 bits | `tokenSetType` | Cached type of token set (0 = whitelist, 1 = blacklist) |
+| 145–255 | 111 bits | Reserved | For future use (must be 0) |
+
+When `hasReceivePolicy` is 0 (the default for all uninitialized storage), no address-level controls are set and the address is always authorized at the address level. The remaining fields are ignored.
+
+When `hasReceivePolicy` is 1, the `receivePolicyId` and `tokenSetId` use standard TIP-403 semantics: ID 0 = always-reject, ID 1 = always-allow. This means:
```
danrobinson: I don't think this paragraph and bullets are necessary

tips/tip-1028.md
```diff
@@ -0,0 +1,734 @@
...
+
+- `transfer(address to, uint256 amount)`
+- `transferFrom(address from, address to, uint256 amount)`
+- `transferWithMemo(address to, uint256 amount, bytes32 memo)`
+- `transferFromWithMemo(address from, address to, uint256 amount, bytes32 memo)`
+- `systemTransferFrom(address from, address to, uint256 amount)`
+
+### Mint Behavior
+
+Minting checks both the token-level policy AND the recipient's full address-level controls (receive policy and token set). This is important because TIP-20 creation is permissionless — anyone can create a token and attempt to mint to any address.
```
danrobinson, line 462: I think could basically just say this works the same way as transfer; don't need all this rationale stuff for it

tips/tip-1028.md
```diff
@@ -0,0 +1,734 @@
...
+
+### Cancel Stale Order
+
+`cancelStaleOrder` checks token-level sender authorization to determine if an order is stale. It does not deliver tokens to a new recipient (cancelled escrow is credited to the maker's internal DEX balance via `increment_balance`). Address-level controls are not involved.
+
+## Integration with Fee Manager
+
+### Fee Collection
+
+`collect_fee_pre_tx` calls `ensureTransferAuthorized(fee_payer, FeeManager)` followed by `transfer_fee_pre_tx` which directly manipulates balances. The `ensureTransferAuthorized` check will now include the FeeManager's address-level controls. Since FeeManager is a system precompile that does not set address-level policies, this adds only the baseline SLOAD cost.
```
danrobinson: This is confusing. It shouldn't add an SLOAD anyway

tips/tip-1028.md
```diff
@@ -0,0 +1,734 @@
...
+
+- **`claimRewards()`**: Transfers accumulated rewards from the token contract to the caller. This calls `ensureTransferAuthorized(tokenAddress, msg.sender)`, which will now check the caller's address-level receive policy (is the token contract authorized to send to me?) and token set (is this token authorized?). This means a regulated entity that has a receive policy whitelist must include the token contract address to claim rewards, and must have the token in their token set. This is correct — claiming rewards is receiving tokens, and should respect the receiver's controls. If the token is already in their token set (which it must be for them to hold it), the token set check passes naturally. The receive policy check against the token contract address requires the address to whitelist it, or not set a receive policy.
```
danrobinson: > If the token is already in their token set (which it must be for them to hold it), the token set check passes naturally. The receive policy check against the token contract address requires the address to whitelist it, or not set a receive policy.

Very tautological sentences; would just drop them

tips/tip-1028.md
```diff
@@ -0,0 +1,734 @@
...
+- **Token set**: The output token must be in the sender's token set (if set). This means a regulated entity cannot accidentally receive unauthorized tokens via a swap — the token set check will reject it.
+
+Order fills (`partial_fill_order`, `fill_order`) credit tokens to the maker's *DEX internal balance* via `increment_balance`, which does not trigger TIP-20 transfers. Address-level controls are NOT checked on internal balance credits. Controls are checked when the maker later calls `withdraw` to move tokens out of the DEX to their wallet (see below).
+
+> **DEX Internal Balance Bypass**: This is an intentional design choice, not a loophole. The DEX's internal accounting is an escrow mechanism — tokens are locked in the DEX contract, not delivered to the maker's wallet. Address-level controls govern what enters an address's wallet, and the `withdraw` path enforces this. Attempting to enforce controls on internal balance credits would break order matching (the maker is not the caller during fills) and would be redundant with the `withdraw` check.
```
danrobinson: > and would be redundant with the `withdraw` check.

I don't think this is true. `withdraw` checks the token's 403 policy but obviously can't check your address-level policy against everyone you traded with.

I think can just say that enforcing address-level controls when limit orders are filled would break order matching and leave it at that.

tips/tip-1028.md
```diff
@@ -0,0 +1,734 @@
...
+1. `ensureTransferAuthorized(sender, DEX_ADDRESS)` for the escrow token (sender sends to DEX)
+2. `ensureTransferAuthorized(DEX_ADDRESS, sender)` for the non-escrow token (DEX will eventually send to sender when the order fills)
+
+Check (2) will now include address-level controls on the sender as receiver. This is correct — it validates upfront that the sender's receive policy and token set would allow receiving the output token from the DEX when the order is eventually filled. This prevents placing orders that would be unfillable due to the maker's own receive controls.
+
+### Withdraw
+
+`withdraw(token, amount)` transfers tokens from the DEX internal balance to the user's wallet via `TIP20.transfer(DEX_ADDRESS, user, amount)`. This goes through `ensureTransferAuthorized(DEX_ADDRESS, user)`, which will now check the user's address-level controls. A user with a token set whitelist can only withdraw tokens that are in their token set.
+
+### Cancel Stale Order
```
danrobinson: I think can cut this section out

tips/tip-1028.md
```diff
@@ -0,0 +1,734 @@
...
+### State Creation Costs
+
+Per TIP-1000, Tempo charges 250,000 gas for new state element creation (SSTORE zero→non-zero).
+
+| Operation | Gas Cost |
+|-----------|----------|
+| First call to `setAddressReceivePolicy` (creates slot) | ~250,000 gas |
+| Subsequent updates to address receive policy | ~5,000 gas |
+| Clearing address receive policy (non-zero→zero) | ~5,000 gas (with 4,800 gas refund per EIP-3529) |
+| Adding address to policy membership | ~250,000 gas (first time) |
```
danrobinson: Not just first time, right? Every time an address is added

tips/tip-1028.md
```diff
@@ -0,0 +1,734 @@
...
+`collect_fee_post_tx` refunds unused fees directly via balance manipulation (not through `ensureTransferAuthorized`), so address-level controls do not apply to refunds.
+
+## Gas Cost Analysis
+
+### Baseline Cost Discussion
+
+This TIP introduces the **first per-address persistent storage read** (`addressPolicies[to]`) on the TIP-20 transfer hot path. Every transfer incurs a ~2,100 gas cold SLOAD to check whether the receiver has set address-level controls, even if they have not. This is a new class of per-transfer overhead — existing features like access keys and TIP-403 membership either short-circuit via transient storage or are keyed by policy ID rather than by address.
+
+We investigated several approaches to avoid this cost:
+- **Packing the flag into existing per-address reads** (balances, policySet): Balances are full `uint256` with no spare bits. `policySet[policyId][to]` is only read when `transferPolicyId >= 2`, and many stablecoins use policy 1 (always-allow), making this unreliable.
```
danrobinson: > Balances are full `uint256` with no spare bits.

Not true; balances have a max of 128. The reason we can't do this is that it is set by an address for ALL tokens, not for a single token.

tips/tip-1028.md
```diff
@@ -0,0 +1,734 @@
...
+### Per-Transfer Overhead (Incremental)
+
+| Scenario | Additional Gas |
+|----------|----------------|
+| Receiver has no controls set | ~2,100 gas (1 SLOAD for addressPolicies[to]) |
+| Receiver has receive policy only | ~4,200 gas (+1 policySet SLOAD) |
+| Receiver has token set only | ~4,200 gas (+1 tokenSetMembers SLOAD) |
+| Receiver has receive policy + token set | ~6,300 gas (+2 SLOADs) |
+
+### Breakdown
```
danrobinson: I think this whole breakdown section is overkill; could just drop it

tips/tip-1028.md
malleshpai: ok, done.

tips/tip-1028.md
malleshpai: removed.

tips/tip-1028.md
malleshpai: 🤦 originally mint behavior was different. now it isn't, but this bit didn't update to reflect that it could be simpler 🤦

tips/tip-1028.md
malleshpai: fixed

tips/tip-1028.md
malleshpai: fixed. this section is an artifact of how we got here (to the extent that's an excuse)

PR #3053, opened 2026-03-09
fix(txpool): consider priority when evicting txs

0xKitsune:
This PR updates txpool eviction logic to consider `submission_id` as a tie breaker between transactions.

### DIFF
--- a/crates/transaction-pool/src/tt_2d_pool.rs
+++ b/crates/transaction-pool/src/tt_2d_pool.rs
@@ -5273,7 +5273,8 @@ mod tests {
     }
 
     #[test]
-    fn cross_pool_eviction_same_priority_evicts_newer() {
+    fn eviction_same_priority_evicts_newer() {
+        // Direction 1: newer expiring tx evicted over older 2D txs
         let mut pool = eviction_test_pool();
         let sender = Address::random();
 
@@ -5307,7 +5308,6 @@ mod tests {
             )
             .unwrap();
 
-        // Same priority — newer expiring nonce tx evicted, older 2D txs preserved
         let AddedTransaction::Pending(pending) = result else {
             panic!("expected pending")
         };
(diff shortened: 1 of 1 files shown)


crates/transaction-pool/src/tt_2d_pool.rs
```diff
@@ -1014,34 +1014,39 @@ impl AA2dPool {
...
         match (worst_2d, worst_expiring) {
-            (Some((id, priority_2d)), Some((hash, priority_exp))) => {
-                // Evict the one with lower priority
-                if priority_exp < priority_2d {
+            (Some((id, pri_2d, sid_2d)), Some((hash, pri_exp, sid_exp))) => {
+                // Same ordering as EvictionKey::Ord: lower priority first, newer first.
+                let evict_expiring = pri_exp
+                    .cmp(&pri_2d)
+                    .then_with(|| sid_2d.cmp(&sid_exp))
+                    .is_le();
```
decofe, line 1041: :rotating_light: **[SECURITY]** Over-eviction via Descendant Demotion

This new tie-break (`.is_le()` — newer first) guarantees an attacker's freshly submitted root tx is evicted first in `evict_lowest_priority`. That triggers `demote_descendants()`, which mass-demotes children from pending→queued. But the caller's `for _ in 0..count` loop uses a stale count and continues evicting honest pending txs beyond what's needed.

**Recommended Fix:** Change `evict_lowest_priority` to dynamically recheck `pending_count > limit` each iteration instead of using a pre-computed fixed loop bound.

---

:warning: **[ISSUE]** Success-on-Drop Mismatch

When equal-priority expiring txs self-evict here, `TempoTransactionPool::add_validated_transaction` still returns `Ok(hash)` without checking `pending.discarded`. Callers see success for an already-evicted tx. Mirror upstream reth's `DiscardedOnInsert` normalization.

decofe (commented):
:eye: **Cyclops Review** — PR #3053

This PR correctly aligns cross-pool eviction tie-breaking with `EvictionKey::Ord`. Two findings survived dedup and verification:

---

### :rotating_light: [SECURITY] Over-eviction of Honest Pending Transactions via Descendant Demotion (DoS)
**Severity:** High 
**File:** `crates/transaction-pool/src/tt_2d_pool.rs:987-995` (not in diff)

`evict_lowest_priority` computes a static `pending_excess` and runs a fixed `for _ in 0..count` loop. Each `evict_2d_pending_tx` call invokes `demote_descendants()`, which can demote an entire chain from pending→queued in one shot. The loop does not recheck pending count, so it continues evicting honest txs after the pool is already under limit. This PR's `submission_id` tie-break makes the attack more reliable — an attacker's newest root tx is guaranteed to be evicted first, triggering mass demotion while the stale loop purges honest transactions.

**Recommended Fix:** Replace the fixed loop with a dynamic `while self.pending_count() > limit` check, or subtract demoted descendants from the counter.

---

### :warning: [ISSUE] Equal-Priority Expiring Txs Return Success While Already Dropped
**Severity:** Low 
**File:** `crates/transaction-pool/src/tempo_pool.rs:485-504` (not in diff)

The new tie-break can self-evict a just-inserted expiring tx (newer loses). `TempoTransactionPool::add_validated_transaction` still returns `Ok(hash)` without checking `pending.discarded`, so RPC callers get success for an already-evicted hash. Upstream reth normalizes this to `Err(DiscardedOnInsert)`.

**Recommended Fix:** Check if `pending.discarded` contains the inserted hash and return `PoolErrorKind::DiscardedOnInsert`.

---

<details>

- :zap: **Eviction loop / `discard()`**: The static `pending_excess` + `demote_descendants` interaction nullifies pool size protection. Queued eviction branch may have a similar variant.
- :zap: **`submission_id` wraparound**: Uses `wrapping_add` — long-lived nodes could theoretically wrap, inverting ordering. Low practical risk.
- :zap: **RPC semantics**: Verify `eth_sendRawTransaction` behavior when AA2D pending pool is full with equal-priority expiring txs.
</details>

crates/transaction-pool/src/tt_2d_pool.rs
0xrusowsky: the success-on-drop finding seems to be valid and partially introduced by this PR. With the new `submission_id` tie-break, the newest tx loses, effectively self-evicting.

the fix should be as simple as:
```rust
if let Some(pending) = added.as_pending() {
    if pending.discarded.iter().any(|tx| *tx.hash() == hash) {
        return Err(PoolError::new(hash, PoolErrorKind::DiscardedOnInsert));
    }
    self.protocol_pool.inner().on_new_pending_transaction(pending);
}
```

decofe: **Over-eviction claim — not valid.**

`demote_descendants()` only flips `is_pending` to `false` — it doesn't remove txs. The `for _ in 0..count` loop calls `evict_one_pending()`, which filters with `.find(|key| key.is_pending())`. Demoted txs are skipped, so the loop does *less* work after demotion, not more. Worst case it evicts exactly `count` pending txs (the pre-computed excess). Demoted txs shift to queued, but queued overflow is handled separately in `discard()`.

**Success-on-Drop — valid.** With the new tie-break (newer first), a freshly inserted equal-priority tx self-evicts and lands in `pending.discarded`, but `add_validated_transaction` still returns `Ok(hash)`. Will fix by checking `discarded` post-insert.

0xrusowsky (commented):
all 3 tests have the expiring tx as the newest, so the `submission_id` tie-break is only tested in one direction.

we should add a test where a newer 2D tx is evicted over an older expiring tx at equal priority.

crates/transaction-pool/src/tt_2d_pool.rs
```diff
@@ -5252,6 +5257,208 @@ mod tests {
...
+            queued_limit: SubPoolLimit {
+                max_txs: 10,
+                max_size: usize::MAX,
+            },
+            ..Default::default()
+        })
+    }
+
+    #[test]
+    fn eviction_same_priority_evicts_newer() {
```
legion2002, line 5276: These tests can be made less verbose by combining into 1

PR #3059, opened 2026-03-10
fix(txpool): update keychain fee token resolution

0xKitsune:
This PR updates txpool validation to fix fee token resolution for keychain txs.

### DIFF
--- a/crates/transaction-pool/src/validator.rs
+++ b/crates/transaction-pool/src/validator.rs
@@ -3522,10 +3522,7 @@ mod tests {
             let transaction =
                 create_aa_with_keychain_signature(user_address, &access_key_signer, None);
 
-            let fee_token = transaction
-                .inner()
-                .fee_token()
-                .unwrap_or(DEFAULT_FEE_TOKEN);
+            let fee_token = transaction.inner().fee_token().unwrap_or(DEFAULT_FEE_TOKEN);
 
             let validator = setup_validator_with_spending_limit(
                 &transaction,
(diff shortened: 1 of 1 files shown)


crates/transaction-pool/src/validator.rs
```diff
@@ -259,35 +259,71 @@ where
...
 
-        Ok(Ok(()))
+        let Ok(key_id) = sig.key_id(&tx.signature_hash()) else {
+            return Ok(());
+        };
+
+        // Skip if same-tx key authorization (key will be authorized during execution)
+        if tx.tx().key_authorization.is_some() {
+            return Ok(());
+        }
```
klkvr, line 294: not immediately obvious why we short-circuit the spending limits check here? shouldn't it still be done even if key is just being authorized in this tx?

crates/transaction-pool/src/validator.rs
```diff
@@ -259,35 +259,71 @@ where
...
-
-            if fee_cost > remaining_limit {
-                return Ok(Err(TempoPoolTransactionError::SpendingLimitExceeded {
-                    fee_token,
-                    cost: fee_cost,
-                    remaining: remaining_limit,
-                }));
-            }
+    /// Checks the access key spending limit for the given fee token and payer.
+    fn check_keychain_spending_limit(
```
fgimenez, line 266: style nit, to align with the rest of methods, we should name it `ensure_...` instead `check_...`

fgimenez (approved):

crates/transaction-pool/src/validator.rs
```diff
@@ -259,35 +259,71 @@ where
...
-                .unwrap_or(U256::ZERO);
-
-            if fee_cost > remaining_limit {
-                return Ok(Err(TempoPoolTransactionError::SpendingLimitExceeded {
-                    fee_token,
-                    cost: fee_cost,
-                    remaining: remaining_limit,
-                }));
-            }
+    /// Checks the access key spending limit for the given fee token and payer.
```
mattsse, line 265: what does this mean/include exactly?

PR #3063, opened 2026-03-10
fix(rpc): AA roundtrip conversion

0xrusowsky:
ref: `AUD-38`

this PR ensures consistent conversion between `TempoTransaction` and `TempoTransactionRequest` so that roundtrip conversions are idempotent

### DIFF
--- a/crates/alloy/src/rpc/request.rs
+++ b/crates/alloy/src/rpc/request.rs
@@ -313,9 +313,9 @@ impl From<TempoTransaction> for TempoTransactionRequest {
             fee_token: tx.fee_token,
             inner: TransactionRequest {
                 from: None,
-                // AA calls are stored in the `calls` field below — leave `to`, `value`,
-                // and `input` unset so that `build_aa()` / `try_into_tx_env()` don't
-                // append a duplicate call from the envelope fields.
+                // AA transactions store their calls in `calls` below.
+                // `to`, `value`, `input` must stay unset to avoid the builder
+                // creating a duplicate call from the envelope fields.
                 to: None,
                 gas: Some(tx.gas_limit()),
                 gas_price: tx.gas_price(),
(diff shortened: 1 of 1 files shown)


crates/alloy/src/rpc/request.rs
```diff
@@ -313,13 +313,16 @@ impl From<TempoTransaction> for TempoTransactionRequest {
             fee_token: tx.fee_token,
             inner: TransactionRequest {
                 from: None,
-                to: Some(tx.kind()),
+                // AA calls are stored in the `calls` field below — leave `to`, `value`,
+                // and `input` unset so that `build_aa()` / `try_into_tx_env()` don't
+                // append a duplicate call from the envelope fields.
```
0xKitsune:
```suggestion
                // AA transactions store their calls in `calls` below.
                // `to`, `value`, `input` must stay unset to avoid the builder
                // creating a duplicate call from the envelope fields.
```

crates/alloy/src/rpc/request.rs
```diff
@@ -604,4 +607,58 @@ mod tests {
             "build_aa must preserve key_authorization from the request"
         );
     }
+
+    #[test]
+    fn test_aa_roundtrip_preserves_count() {
```
0xKitsune, line 612: Can we also test that rountrip via `try_into_tx_env` is correct?

0xrusowsky: addressed in [`f06741a` (this PR)](https://github.com/tempoxyz/tempo/pull/3063/commits/f06741a204caf8338d361e46669348df05390f16)

crates/alloy/src/rpc/request.rs
0xrusowsky: addressed in [`f06741a` (this PR)](https://github.com/tempoxyz/tempo/pull/3063/commits/f06741a204caf8338d361e46669348df05390f16)

PR #3065, opened 2026-03-10
fix(transaction-pool): validate inline key authorization state

legion2002:
Closes CHAIN-886

Inline key_authorization transactions no longer return early after signature checks in pool validation. They now validate key storage state (reject existing/revoked keys) and enforce fee-token spending limits from inline limits before entering the mempool, preventing deterministic execution failures from being gossiped.

### DIFF
--- a/crates/transaction-pool/src/tempo_pool.rs
+++ b/crates/transaction-pool/src/tempo_pool.rs
@@ -240,6 +240,12 @@ where
             // from state for affected (account, key_id, fee_token) combos and evict if
             // the pending tx's fee cost now exceeds the remaining limit.
             if !updates.spending_limit_spends.is_empty()
+                // NOTE: sponsored txs don't consume the sender's key limits.
+                && tx
+                    .transaction
+                    .inner()
+                    .as_aa()
+                    .is_none_or(|aa| aa.tx().fee_payer_signature.is_none())
                 && let Some(ref mut provider) = state_provider
                 && let Some(ref subject) = keychain_subject
                 && subject.matches_spending_limit_update(&updates.spending_limit_spends)
@@ -1040,15 +1046,24 @@ pub(crate) fn exceeds_spending_limit(
     subject: &crate::transaction::KeychainSubject,
     fee_token_cost: alloy_primitives::U256,
 ) -> bool {
-    let limit_key = AccountKeychain::spending_limit_key(subject.account, subject.key_id);
     // Spec doesn't affect raw storage reads (sload), so default is safe here.
     let spec = TempoHardfork::default();
-    match provider.with_read_only_storage_ctx(spec, || {
-        AccountKeychain::new().spending_limits[limit_key][subject.fee_token].read()
-    }) {
-        Ok(remaining) => fee_token_cost > remaining,
-        Err(_) => false,
-    }
+    let limit_key = AccountKeychain::spending_limit_key(subject.account, subject.key_id);
+
+    provider
+        .with_read_only_storage_ctx(spec, || -> TempoPrecompileResult<bool> {
+            let keychain = AccountKeychain::new();
+            if !keychain.keys[subject.account][subject.key_id]
+                .read()?
+                .enforce_limits
+            {
+                return Ok(false);
+            }
+
+            let remaining = keychain.spending_limits[limit_key][subject.fee_token].read()?;
+            Ok(fee_token_cost > remaining)
+        })
+        .unwrap_or_default()
 }
 
 /// Reads the transfer policy ID for a TIP-20 token, using a cache to avoid redundant lookups.
@@ -1111,7 +1126,10 @@ mod tests {
     use crate::transaction::KeychainSubject;
     use reth_provider::test_utils::{ExtendedAccount, MockEthProvider};
     use reth_storage_api::StateProviderFactory;
-    use tempo_precompiles::{ACCOUNT_KEYCHAIN_ADDRESS, account_keychain::AccountKeychain};
+    use tempo_precompiles::{
+        ACCOUNT_KEYCHAIN_ADDRESS,
+        account_keychain::{AccountKeychain, AuthorizedKey},
+    };
 
     fn provider_with_spending_limit(
         account: Address,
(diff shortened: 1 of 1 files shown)


crates/transaction-pool/src/validator.rs
```diff
@@ -229,6 +220,53 @@ where
...
+                    .unwrap_or(U256::ZERO);
+
+                if fee_cost > remaining_limit {
+                    return Ok(Err(TempoPoolTransactionError::SpendingLimitExceeded {
+                        fee_token,
+                        cost: fee_cost,
+                        remaining: remaining_limit,
+                    }));
+                }
+            }
```
klkvr: can we unify the limits check with the one below?

crates/transaction-pool/src/validator.rs
```diff
@@ -121,10 +139,34 @@ where
...
+        self.validate_against_keychain_with_fee_context(
+            transaction,
+            state_provider,
+            transaction.sender(),
+            transaction
+                .inner()
+                .fee_token()
+                .unwrap_or(tempo_precompiles::DEFAULT_FEE_TOKEN),
+        )
+    }
```
0xKitsune: Do we need this helper function or can we simply inline the logic in this function?

crates/transaction-pool/src/validator.rs
```diff
@@ -135,15 +135,57 @@ where
...
```
klkvr: is this only used in tests now? should this be removed?

crates/transaction-pool/src/validator.rs
```diff
@@ -211,10 +253,38 @@ where
         };
 
         // Ensure that if key auth is present, it is for the same key as the keychain signature.
+        if let Some(auth) = auth
```
0xKitsune: nit: There are two separate `if let Some(auth) = auth` blocks (one for key-id mismatch and one for state validation). It would be nice to unify these.

crates/transaction-pool/src/validator.rs
```diff
@@ -839,7 +905,12 @@ where
         }
 
         // Validate transactions that involve keychain keys.
-        match self.validate_against_keychain(&transaction, &mut state_provider) {
+        match self.validate_against_keychain_with_fee_context(
+            &transaction,
+            &mut state_provider,
+            fee_payer,
+            fee_token,
+        ) {
```
klkvr: let's keep this called `validate_against_keychain`?

crates/transaction-pool/src/validator.rs
```diff
@@ -211,10 +253,38 @@ where
...
+        }
+
+        // Compute storage slot using helper function
+        let storage_slot = AccountKeychain::new().keys[transaction.sender()][key_id].base_slot();
+
+        // Read storage slot from state provider
+        let slot_value = state_provider
+            .storage(ACCOUNT_KEYCHAIN_ADDRESS, storage_slot.into())?
+            .unwrap_or(U256::ZERO);
+
```
klkvr: should this use readonly provider? ref https://github.com/tempoxyz/tempo/pull/3062

crates/transaction-pool/src/validator.rs
```diff
@@ -211,10 +253,38 @@ where
...
+        // Decode AuthorizedKey using helper
+        let authorized_key = AuthorizedKey::decode_from_slot(slot_value);
```
0xrusowsky: can we use the `ReadOnlyStorageProvider` and not manually read/decode slots?

crates/transaction-pool/src/validator.rs
0xKitsune: +1 if we are able to use the read only provider here.

legion2002: Yup updated in https://github.com/tempoxyz/tempo/pull/3065/changes/9e59ffcb69db7ae55f9da4006125b2eb164804c6

crates/transaction-pool/src/validator.rs
```diff
@@ -20,9 +20,10 @@ use tempo_chainspec::{
     hardfork::{TempoHardfork, TempoHardforks},
 };
 use tempo_evm::TempoEvmConfig;
+#[cfg(test)]
+use tempo_precompiles::{ACCOUNT_KEYCHAIN_ADDRESS, account_keychain::AuthorizedKey};
```
klkvr, line 24: are those imports used?

decofe (commented):
:eye: **Cyclops Review**

This PR improves mempool spending-limit enforcement: admission-time checks for inline `KeyAuthorization`, resolved fee-token caching, wildcard token matching for spend updates, and `enforce_limits` gating. Two issues found.

<details>

- ⚡ **Pool/Execution Parity**: Any mempool gate should use execution semantics (`fee_payer == sender`), not envelope shape (`fee_payer_signature` presence). This mismatch is the root cause of the inline finding below and has been a recurring source of issues.
- ⚡ **TOCTOU Pattern in maintain.rs**: Every new admission-time invariant in `validator.rs` needs a corresponding invalidation path in `maintain.rs` + `tempo_pool.rs`. The `KeyAuthorized` gap is a direct instance of the SIGP-TMPO2-04 pattern.
</details>

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -240,6 +240,12 @@ where
             // from state for affected (account, key_id, fee_token) combos and evict if
             // the pending tx's fee cost now exceeds the remaining limit.
             if !updates.spending_limit_spends.is_empty()
+                // NOTE: sponsored txs don't consume the sender's key limits.
+                && tx
+                    .transaction
+                    .inner()
+                    .as_aa()
+                    .is_none_or(|aa| aa.tx().fee_payer_signature.is_none())
```
decofe, line 248: ⚠️ **[ISSUE]** Self-Sponsored Fee-Payer Signature Bypasses Spending-Limit Spend Eviction

`fee_payer_signature.is_none()` is an incorrect proxy for "sender pays fees." An AA tx can include a `fee_payer_signature` that recovers to the sender (`fee_payer == sender`), so sender key spending limits still apply at execution. These self-sponsored txs are skipped by this maintenance recheck and persist in-pool after their limit is exhausted, causing deterministic execution failures.

Impact is bounded by per-sender cap (16) and stale-pending eviction (~30 min), but the semantic mismatch should be fixed.

**Recommended Fix:** Replace the signature-presence gate with a resolved fee-payer check: skip spend-recheck only when resolved `fee_payer != sender`. Cache the resolved fee-payer identity during validation (similar to `resolved_fee_token`).

crates/transaction-pool/src/validator.rs
```diff
@@ -210,35 +231,60 @@ where
...
+        // fee-token spending limits to prevent deterministic execution failures from entering
+        // the pool.
         if let Some(auth) = auth {
             if auth.key_id != key_id {
                 return Ok(Err(TempoPoolTransactionError::Keychain(
                     "KeyAuthorization key_id does not match Keychain signature key_id",
                 )));
             }
 
+            if authorized_key.expiry > 0 {
```
decofe, line 250: ⚠️ **[ISSUE]** Missing KeyAuthorized Event Eviction in Pool Maintenance

This admission-time check correctly rejects inline `KeyAuthorization` when the key already exists. However, `maintain.rs` does not parse `KeyAuthorized` events. If multiple txs with the same inline `KeyAuthorization` are admitted concurrently (via 2D expiring nonces), they all pass admission. When the first executes and creates the key, the rest become deterministically invalid but are never evicted — a recurrence of the SIGP-TMPO2-04 TOCTOU pattern.

Impact bounded by 30s expiring nonce window and per-sender cap (16).

**Recommended Fix:** Parse `IAccountKeychain::KeyAuthorized` events in `TempoPoolUpdates::from_chain` and add a corresponding eviction branch in `evict_invalidated_transactions`.

PR #3069, opened 2026-03-10
docs(tip-1029): fee recipient from validator config v2

decofe:
Proposes to use `ValidatorConfigV2.Validator.feeRecipient` as the `suggestedFeeRecipient` when proposing blocks.

### DIFF
--- a/tips/tip-1029.md
+++ b/tips/tip-1029.md
@@ -85,7 +85,7 @@ like so, with the contract value taking priority if set:
 
 # Invariants
 
-The invariants listed here are only local to a given node and consistute an
+The invariants listed here are only local to a given node and constitute an
 implementation detail. They are not expected to be enforced by validators.
 
 - The fee recipient used in payload building MUST equal the on-chain
--- a/tips/tip-1029.md
+++ b/tips/tip-1029.md
@@ -1,5 +1,5 @@
 ---
-id: TIP-XXXX
+id: TIP-1029
 title: Fee Recipient from Validator Config V2
 description: Validators declare a fee recipient address in ValidatorConfigV2; the node reads it and passes it to the payload builder.
 authors: @superfluffy
@@ -8,26 +8,13 @@ related: TIP-1017
 protocolVersion: TBD
 ---
 
-# TIP-XXXX: Fee Recipient from Validator Config V2
+# TIP-1029: Fee Recipient from Validator Config V2
 
 ## Abstract
 
 Proposers use the `feeRecipient` field added to validator entries in [TIP-1017](./tip-1017.md).
 When a validator proposes a block, the node reads this on-chain field and uses it as the fee recipient in the payload builder.
 The command line flag `--consensus.fee-recipient` is removed.
-
-## Proposal
-
-1. Nodes, when building blocks to be proposed, use the
-  `ValidatorConfigV2.Validator.feeRecipient` field of their `Validator` entry
-   at the chain state of the notarized parent block.
-2. The state of the notarized (not finalized) block is used because:
-    + the state must be available to propose a block.
-    + because Tempo consensus uses a VRF to determine the next leader, a node
-      can be elected leader several blocks in a row. Waiting for the fee receipient
-      rotation to be actived by a finalized block would make it non-deterministic.
-3. Remove the `--consensus.fee-recipient` command line argument to provide a
-   single source of truth and make compliance easier.
 
 ## Motivation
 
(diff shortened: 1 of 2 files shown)


tips/tip-1029.md
```diff
@@ -0,0 +1,76 @@
...
+related: TIP-1017
+protocolVersion: TBD
+---
+
+# TIP-1029: Fee Recipient from Validator Config V2
+
+## Abstract
+
+Proposers use the `feeRecipient` field added to validator entries in [TIP-1017](./tip-1017.md).
+When a validator proposes a block, the node reads this on-chain field and uses it as the fee recipient in the payload builder.
```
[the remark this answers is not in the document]
SuperFluffy: This is correct; the relevant change lives in https://github.com/tempoxyz/tempo/pull/2910.

tips/tip-1029.md
```diff
@@ -0,0 +1,76 @@
...
+status: Draft
+related: TIP-1017
+protocolVersion: TBD
+---
+
+# TIP-1029: Fee Recipient from Validator Config V2
+
+## Abstract
+
+Proposers use the `feeRecipient` field added to validator entries in [TIP-1017](./tip-1017.md).
```
decofe: ⚠️ **[ISSUE]** TIP References Non-Existent `feeRecipient` Field in ValidatorConfigV2

This line (and the spec throughout) references `ValidatorConfigV2.Validator.feeRecipient` and `setFeeRecipient` as fields added in TIP-1017. However, these do not exist in TIP-1017, the `ValidatorV2` Rust struct, the generated ABI bindings, or the Solidity interface (`IValidatorConfigV2.sol`). Implementing this TIP as written is impossible without first extending ValidatorConfigV2.

**Recommended Fix:** Either update TIP-1017 and the ValidatorConfigV2 contract to add `feeRecipient` / `setFeeRecipient`, or explicitly state this TIP depends on a prerequisite contract change.

tips/tip-1029.md
```diff
@@ -0,0 +1,76 @@
...
+hardfork period and removed in a subsequent hardfork (TBD). The node will behave
+like so, with the contract value taking priority if set.
+
+1. Use `v.feeRecipient` if set to a value that is *not* the zero address `0x0`.
+2. If `v.feeRecipient` is set to the zero address `0x0`, use `--consensus.fee-recipient`.
+3. `--consensus.fee-recipient` will retain a default value of `0x0`.
+
+# Invariants
+
+- The fee recipient used in payload building MUST equal the on-chain `ValidatorConfigV2.Validator.feeRecipient` value read at the notarized parent block's state.
```
decofe: ⚠️ **[ISSUE]** Contradictory Invariants Cause Consensus Split During Migration

Invariant 1 says fee recipient MUST equal the on-chain value. Invariant 2 (line 75) says if the on-chain value is `0x0`, fall back to `--consensus.fee-recipient`. A proposer with on-chain `0x0` uses CLI (per rule 2), but verifiers checking rule 1 compare against `0x0` and reject the block — guaranteed liveness failure.

Additionally, using `0x0` as "unset" means treasury cannot intentionally burn fees to the zero address; an infra engineer controlling the CLI flag would redirect them.

**Recommended Fix:** Exempt the migration fallback from invariant 1 explicitly (e.g., "verifiers MUST NOT enforce strict equality when `feeRecipient` is `0x0` during migration"). Consider a dedicated sentinel other than `0x0` so intentional burn-to-zero remains possible.

tips/tip-1029.md
```diff
@@ -0,0 +1,76 @@
...
+requirements (for example, requiring a higher multisig threshold), while keeping
+every day ops more manageable (by requiring a lower multisig threshold).
+
+---
+
+# Specification
+
+1. Nodes, when building blocks to be proposed, use the
+   `ValidatorConfigV2.Validator.feeRecipient` field of their `Validator` entry
+   at the chain state of the notarized parent block.
```
decofe: 💡 **[SUGGESTION]** Clarify State Dependency for Back-to-Back Leader Slots

The spec requires reading `feeRecipient` at the "notarized parent block's state." When a node is elected leader for consecutive blocks (N and N+1), block N is not yet notarized when N+1 building begins. Should the node use the post-state of its own proposed-but-not-yet-notarized block, or wait for notarization (stalling pipelining)?

**Recommended Fix:** Add a clarification such as: "For back-to-back leader slots, the node MAY use the post-state of its own proposed parent block before notarization."

decofe (commented):
:eye: **Cyclops Review**

This draft TIP specifies reading the fee recipient from on-chain `ValidatorConfigV2` instead of a CLI flag. Two specification issues were identified: the referenced `feeRecipient` field does not exist in ValidatorConfigV2 (TIP-1017 or implementations), and the invariants section contains contradictory rules that would cause consensus failures during migration if implemented as written.

<details>

- ⚡ **No Execution-Layer Enforcement**: When implemented, the execution layer must validate `SubBlock.fee_recipient` against the on-chain `ValidatorConfigV2` value. Currently it trusts whatever is in the signed `SubBlockMetadata`.
- ⚡ **Subblock Relationship TBD**: The "Relationship with subblock building?" section must be resolved before implementation since subblock fee recipients follow the same concern.
- ⚡ **Txpool Revalidation**: Dynamic fee recipient changes may require the transaction pool to monitor `ValidatorConfigV2` state changes in addition to `FeeManager.ValidatorTokenSet` events.
</details>

tips/tip-1029.md
SuperFluffy: TIP 1017 is now merged and the reference be correct now.

tips/tip-1029.md
SuperFluffy: Clarified that the invariants are local to the node only. It is not expected that any validator enforces these variants on a proposer.

tips/tip-1029.md
SuperFluffy: We currently do not have back-to-back leader slots that skip notarizations. A notarization is a prerequisite to a proposal. But added this anyway since it doesn't hurt.

PR #3071, opened 2026-03-10
docs(tip-1030): allow same-tick flip orders

decofe:
Adds TIP-1030, which is the "allow same-tick flip orders" portion of TIP-1002 without the "prevent crossed orders" change.

Relaxes `placeFlip` validation so `flipTick == tick` is allowed, enabling use cases like instant token convertibility where an issuer places flip orders on both sides at the same tick.

Prompted by: Dan

### DIFF
--- a/tips/tip-1030.md
+++ b/tips/tip-1030.md
@@ -41,7 +41,7 @@ No new errors.
 # Implications
 
 - **Locked books**: Same-tick flip orders can result in `best_bid_tick == best_ask_tick`. This forecloses any future upgrade that would forbid bids and asks from resting at the same tick, since same-tick flip orders legitimately create that state.
-- **MEV**: Tighter flip orders (where `flipTick` is closer to or equal to `tick`) increase the likelihood of certain kinds of MEV, such as backrunning, since the price at which the new opposite-side order will appear is more predictable.
+- **MEV**: Tighter flip orders (where `flipTick` is closer to or equal to `tick`) increase the likelihood of certain kinds of MEV, such as backrunning, since the new opposite-side order appears at a better price for the backrunner.
 
 # Invariants
 
--- a/tips/tip-1030.md
+++ b/tips/tip-1030.md
@@ -5,17 +5,19 @@ description: Relaxes the placeFlip validation to allow flipTick to equal tick, e
 authors: Dan Robinson
 status: Draft
 related: TIP-1002
+protocolVersion: T5
+supersedes: TIP-1002
 ---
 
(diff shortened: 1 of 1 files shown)


tips/tip-1030.md
```diff
@@ -0,0 +1,44 @@
...
+# Specification
+
+## Modified behavior
+
+The `placeFlip` validation is relaxed to allow `flipTick == tick`:
+
+- **Current behavior**: For bids, `flipTick > tick` required; for asks, `flipTick < tick` required
+- **New behavior**: For bids, `flipTick >= tick` required; for asks, `flipTick <= tick` required
+
+This applies to both the precompile implementation and the Solidity reference implementation.
```
danrobinson: Can remove this

samczsun (approved):

tips/tip-1030.md
```diff
@@ -0,0 +1,42 @@
...
+authors: Dan Robinson
+status: Draft
+related: TIP-1002
+---
+
+# TIP-1030: Allow same-tick flip orders
+
+## Abstract
+
+Relaxes the `placeFlip` validation to allow `flipTick == tick`, enabling flip orders that flip to the same price. This is the "allow same-tick flip orders" portion of TIP-1002, without the "prevent crossed orders" change.
```
malleshpai: does this supersede tip-1002?

decofe (commented):
## TIP-1030 Review

### protocolVersion → T5

### Correctness

1. **Queue-position semantics need to be explicit.** When a flip order triggers at the same tick, the new order MUST get:
   - A **new orderId**
   - Appended to the **tail** of the opposite-side FIFO queue (i.e., fresh queue position, no priority retention)

   Without this, implementations could preserve the original order's priority incorrectly.

2. **Self-referential fill loop.** With `flipTick == tick`, a flip order on one side creates a new order at the same price on the opposite side. If that new order is immediately matchable (e.g., a crossed book), define whether it can be filled in the same `fill_order` traversal or only in a subsequent swap call. If it can self-fill, you get recursive flip churn within a single tx.

### Suggested changes

Add to frontmatter:
```yaml
protocolVersion: T5
```

Add to Specification:
```markdown
## Queue Position

A flip order placed at `flipTick` receives a new `orderId` and is appended to the tail
of the opposite-side order queue at that tick. It does not inherit the original order's
queue position or priority.
```

0xrusowsky (approved):

howydev (approved):

0xKitsune (approved):

PR #3074, opened 2026-03-10
chore: dkg e2e test for wiped consensus state

hamdiallam:
Ensure the node recovers a share in the subsequent epoch


crates/e2e/src/tests/dkg/share_loss.rs
```diff
@@ -45,46 +61,97 @@ fn validator_lost_share_but_gets_share_in_next_epoch() {
...
+        let target_idx = validators.len() - 1;
+        let uid = validators[target_idx].uid().to_string();
+
+        join_all(validators.iter_mut().map(|v| v.start(&context))).await;
+
+        'setup: loop {
+            context.sleep(Duration::from_secs(1)).await;
+            let metrics = context.encode();
+
+            // Should have dealings in epoch 1.
```
SuperFluffy: We can check for this explicitly - stop immediately if `acks_received > 0` (metric in the DKG actor) and delete state

crates/e2e/src/tests/dkg/share_loss.rs
```diff
@@ -45,46 +61,97 @@ fn validator_lost_share_but_gets_share_in_next_epoch() {
...
+
+        let (mut validators, _execution_runtime) =
+            setup_validators(&mut context, setup.clone()).await;
+
+        let target_idx = validators.len() - 1;
+        let uid = validators[target_idx].uid().to_string();
+
+        join_all(validators.iter_mut().map(|v| v.start(&context))).await;
+
+        'setup: loop {
```
SuperFluffy, line 116: Let's add a condition to the loop that this should have happened in epoch 0 - `assert!(epoch < 1)`

crates/e2e/src/tests/dkg/share_loss.rs
```diff
@@ -45,46 +61,97 @@ fn validator_lost_share_but_gets_share_in_next_epoch() {
...
+
+        validators[target_idx].start(&context).await;
+
+        let uid = validators[target_idx].metric_prefix();
+
+        // The node should finalize the current ceremony as an observer (hitting
+        // MissingPlayerDealing) and then recover a share in the next epoch.
+        let mut detected_missing_dealings = false;
+
+        'recover: loop {
```
SuperFluffy, line 146: Let's panic if the network goes to epoch 2 so this test does not run indefinitely.

crates/e2e/src/testing_node.rs
```diff
@@ -122,6 +122,15 @@ where
         &self.uid
     }
 
+    /// Get the metric prefix used by the most recently started instance.
+    ///
+    /// # Panics
+    /// Panics if the node has was never started.
+    pub fn metric_prefix(&self) -> String {
```
SuperFluffy, line 148: Is that part of the metrics? TIL

crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -1439,6 +1443,13 @@ impl Metrics {
             bad_dealings.clone(),
         );
 
+        let missing_player_dealings = Counter::default();
```
SuperFluffy: I think this metric is not very useful in practice, as it will only ever be `0` or `1` - either the state was missing on restart or it wasn't. During the entire lifetime of the program this will never change again.

You can infer what the node was doing by using `how_often_dealer` - if the node does not have a share, it will not increment this metric in the followup epoch.

So the flow of the test would be like this:

1. on start, `how_often_dealer = 1` in epoch 0
2. delete/move state
3. on restart, `how_often_dealer = 0` in epoch 0, ` = 0` in epoch 1`, ` = 1` in epoch 2

crates/e2e/src/tests/dkg/share_loss.rs
```diff
@@ -45,46 +61,97 @@ fn validator_lost_share_but_gets_share_in_next_epoch() {
...
+                && v > 0
+            {
+                break 'acquire_share;
+            }
+        }
+    });
+}
+
+#[test_traced]
+fn validator_loses_consensus_state_becomes_observer() {
```
SuperFluffy, line 96: IMO let's not stop here but see if it can recover.

hamdiallam: when it is a dealer again it has recovered right?

crates/commonware-node/src/dkg/manager/actor/mod.rs
hamdiallam: Nice, this is clean

SuperFluffy (approved):

PR #3075, opened 2026-03-10
chore: speed up tests in CI via parallelisation

emmajam:
chore: speed up tests in CI via parallelisation

Screenshot 2026-03-10 at 19 13 11


sds (approved):
Hell yeah

PR #3076, opened 2026-03-10
feat: add bootnodes as trusted peers

decofe:
## Summary
Bootnodes are now also added as trusted peers at node startup, so connections to them are always maintained and not subject to peer eviction.

## Changes
- In the `apply` block in `bin/tempo/src/main.rs`, added bootnodes from the chainspec to `network.trusted_peers`

## Testing
`cargo check -p tempo` passes.

Prompted by: zygis


SuperFluffy (approved):
Change lgtm. But the PR message is too noisy, the testing bit makes no sense.

shekhirin (approved):

PR #3081, opened 2026-03-10
ci(bench): wait for node before using faucets

Rjected:
If we don't wait for the node to produce a block before funding, it gets stuck with an older / stale block timestamp, which then will reject all of the faucet / funding transactions.


decofe:
cc @Rjected

🚀 Benchmark started! [View job](https://github.com/tempoxyz/tempo/actions/runs/23023424950/job/66865400667)

**Config:** preset: `tip20`, duration: `300s`, bloat: `102400 MiB`, tps: `10000`, baseline: `main`, feature: `dan/fix-bench-expiry`

PR #3083, opened 2026-03-10
fix(transaction-pool): cache resolved fee token for pool maintenance

0xrusowsky:
Caches the resolved fee token on `TempoPooledTransaction` so `keychain_subject()` returns the validated token instead of the raw `tx.fee_token`. Uses `Address::ZERO` as a wildcard in spending-limit eviction for block txs without an explicit fee_token.

### DIFF
--- a/crates/transaction-pool/src/maintain.rs
+++ b/crates/transaction-pool/src/maintain.rs
@@ -120,7 +120,7 @@ impl TempoPoolUpdates {
                     updates.spending_limit_changes.insert(
                         event.account,
                         event.publicKey,
-                        event.token,
+                        Some(event.token),
                     );
                 }
             }
@@ -180,13 +180,14 @@ impl TempoPoolUpdates {
                 continue;
             }
             // Resolving the fee token requires state (AMM routing), which we don't have here.
-            // `Address::ZERO` wildcards the token in `SpendingLimitUpdates::contains`, so every
-            // pending tx for this (account, key_id) is re-checked.
-            // Safe because tx-pool gates eviction on `exceeds_spending_limit()`, which reads state.
-            let fee_token = aa_tx.tx().fee_token.unwrap_or(Address::ZERO);
-            updates
-                .spending_limit_spends
-                .insert(keychain_sig.user_address, key_id, fee_token);
+            // `None` wildcards the token in `SpendingLimitUpdates::contains`, so every pending tx
+            // for this (account, key_id) is re-checked. Safe because the main pool still gates
+            // eviction on `exceeds_spending_limit()`, which can read state.
+            updates.spending_limit_spends.insert(
+                keychain_sig.user_address,
+                key_id,
+                aa_tx.tx().fee_token,
+            );
         }
 
         updates
@@ -1444,7 +1445,7 @@ mod tests {
             updates.spending_limit_spends.insert(
                 Address::random(),
                 Address::random(),
-                Address::random(),
+                Some(Address::random()),
             );
             assert!(updates.has_invalidation_events());
         }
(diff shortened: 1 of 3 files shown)


crates/transaction-pool/src/maintain.rs
```diff
@@ -179,13 +179,15 @@ impl TempoPoolUpdates {
...
-            let fee_token = aa_tx
-                .tx()
-                .fee_token
-                .unwrap_or(tempo_precompiles::DEFAULT_FEE_TOKEN);
-            updates
-                .spending_limit_spends
-                .insert(keychain_sig.user_address, key_id, fee_token);
+            // Resolving the fee token requires state (AMM routing), which we don't have here.
+            // `None` wildcards the token in `SpendingLimitUpdates::contains`, so every pending tx
+            // for this (account, key_id) is re-checked. Safe because the main pool still gates
```
decofe, line 184: :rotating_light: **[SECURITY]** Wildcard Spending-Limit Eviction Ignores `enforce_limits` and Sponsored Fee Payers

This comment claims eviction is "safe because the main pool still gates eviction on `exceeds_spending_limit()`, which can read state." However, `exceeds_spending_limit()` (`tempo_pool.rs:1038-1052`) only reads `spending_limits[account][key_id][fee_token]` from state — it does **not** check `enforce_limits` or `fee_payer`.

For unlimited keys (`enforce_limits == false`), no spending limit is written to state, so remaining defaults to `0`. `exceeds_spending_limit` returns `cost > 0` → `true`, falsely evicting valid pending txs. Similarly, sponsored txs (`fee_payer != sender`) are evicted based on the sender's irrelevant limit.

**Recommended Fix:** Update `exceeds_spending_limit` (or its call site at `tempo_pool.rs:246`) to:
1. Skip eviction when `fee_payer != subject.account`
2. Skip eviction when `enforce_limits == false` for the key
3. Skip eviction when the pending tx has an inline `KeyAuthorization`

decofe (commented):
:eye: **Cyclops Review**

**1 finding survived dedup and verification.**

---

### :rotating_light: [SECURITY] Wildcard Spending-Limit Eviction Ignores `enforce_limits` and Sponsored Fee Payers
**Severity:** High
**File:** `crates/transaction-pool/src/tempo_pool.rs:1038-1052` (not in diff — see inline comment on `maintain.rs` for context)

The new wildcard `None` token in `spending_limit_spends` causes every pending tx for a given `(account, key_id)` to be re-checked via `exceeds_spending_limit()`. That function only reads `spending_limits[account][key_id][fee_token]` from state and returns `fee_token_cost > remaining`. It does **not** check:

1. **Unlimited keys** (`enforce_limits == false`) — state defaults to `0`, so `cost > 0` is always `true`, falsely evicting valid txs.
2. **Sponsored txs** (`fee_payer != sender`) — the validator skips spending-limit checks for sponsored txs, but `exceeds_spending_limit` always checks the sender's limit.
3. **Inline `KeyAuthorization`** — the validator uses `auth.limits`, but `exceeds_spending_limit` reads on-chain state limits.

This diverges from the validator (`validator.rs:318-333`) and execution (`account_keychain/mod.rs:443-445`).

**Recommended Fix:** Update `exceeds_spending_limit` (or its call site at `tempo_pool.rs:246`) to skip eviction when `fee_payer != subject.account` or `enforce_limits == false`.

<details>

- :zap: **Pool/Execution parity**: The validator and execution layer both gate spending-limit enforcement behind `fee_payer == sender` and `enforce_limits`, but the pool eviction path lacks these gates. The wildcard broadening amplifies the blast radius.
- :zap: **Wildcard fanout scope**: `None` wildcarding means every included keychain tx triggers re-evaluation of all pending txs for that `(account, key_id)`. Confirm this is acceptable once the eviction guard is fixed.
</details>

PR #3085, opened 2026-03-10
fix: gate spending-limit eviction on enforce_limits

0xrusowsky:
addresses cyclops finding https://github.com/tempoxyz/tempo/pull/3083#discussion_r2914664048

### DIFF
--- a/crates/transaction-pool/src/tempo_pool.rs
+++ b/crates/transaction-pool/src/tempo_pool.rs
@@ -240,6 +240,12 @@ where
             // from state for affected (account, key_id, fee_token) combos and evict if
             // the pending tx's fee cost now exceeds the remaining limit.
             if !updates.spending_limit_spends.is_empty()
+                // NOTE: sponsored txs don't consume the sender's key limits.
+                && tx
+                    .transaction
+                    .inner()
+                    .as_aa()
+                    .is_none_or(|aa| aa.tx().fee_payer_signature.is_none())
                 && let Some(ref mut provider) = state_provider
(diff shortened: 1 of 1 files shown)


crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -1040,13 +1040,23 @@ pub(crate) fn exceeds_spending_limit(
     subject: &crate::transaction::KeychainSubject,
     fee_token_cost: alloy_primitives::U256,
 ) -> bool {
-    let limit_key = AccountKeychain::spending_limit_key(subject.account, subject.key_id);
     // Spec doesn't affect raw storage reads (sload), so default is safe here.
     let spec = TempoHardfork::default();
-    match provider.with_read_only_storage_ctx(spec, || {
-        AccountKeychain::new().spending_limits[limit_key][subject.fee_token].read()
+    let limit_key = AccountKeychain::spending_limit_key(subject.account, subject.key_id);
```
legion2002, line 1051: Is there a reason you removed the fee payer check? 
It was in the first commit. I think it is correct to have it

legion2002 (commented):
I think this PR addresses 2 of the 3 cases mentioned in the cyclops audit, but not this one
3. Skip eviction when the pending tx has an inline KeyAuthorization

Still thinking if we need this or not

0xKitsune:
> 3. Skip eviction when the pending tx has an inline KeyAuthorization

Cross posting here for vis, after review/discussion it looks like this is checked at validation.

0xKitsune (approved):

PR #3086, opened 2026-03-10
fix: check inline key auth

0xKitsune:


legion2002 (approved):
makes sense, because you're checking the limits from the access key about to be authorized, if it is present.

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -246,14 +246,39 @@ where
...
+                let fee_cost = tx.transaction.fee_token_cost();
+                let should_evict = if let Some(auth) = tx
+                    .transaction
+                    .inner()
+                    .as_aa()
+                    .and_then(|aa| aa.tx().key_authorization.as_ref())
+                {
+                    // Inline KeyAuth: check limits from the tx itself.
+                    // limits == None means unlimited spending → never evict.
+                    auth.limits.as_ref().is_some_and(|limits| {
```
legion2002, line 261: we probably shouldn't duplicate this logic

PR #3092, opened 2026-03-11
feat(tip): encoding consensus metadata in tempo blocks

hamdiallam:
Rendered: https://github.com/tempoxyz/tempo/blob/tips/tip-1031/tips/tip-1031.md

### DIFF
--- a/tips/tip-1031.md
+++ b/tips/tip-1031.md
@@ -3,7 +3,7 @@ id: TIP-1031
 title: Embed consensus context in the block Header
 description: Embed consensus context into the block header
 authors: Hamdi, Janis
-status: Draft
+status: Approved
 related: N/A
 protocolVersion: T4
 ---


tips/tip-1031.md
```diff
@@ -0,0 +1,144 @@
...
+
+This TIP prepends a fixed-size consensus context prefix to `extra_data`. The consensus context is encoded using the `commonware_codec::Write` as defined upstream:
+
+```
+[ varint(epoch) ][ varint(view) ][ pubkey_bytes(32) ][ varint(parent_view) ][ digest_bytes(32) ]
+```
+
+Depending on the epoch and view counts, this is roughly ~70B (For views/epoch in the order of billions it only grows to 80 bytes). Well within the limits of the max extra data size (10KiB).
+
+Because the context has a known fixed-width trailer (32-byte pubkey + varint + 32-byte digest) but variable-width varint prefix, and the DKG payload that follows is also variable-length, a **1-byte discriminant** is prepended to distinguish the DKG payload type:
```
hamdiallam: Want to double-highlight that we **dont** have to do this if we think it's unnecessary. I personally think encoding the dkg type is useful

tips/tip-1031.md
```diff
@@ -0,0 +1,144 @@
...
+| Block type | Current `extra_data` contents |
+|---|---|
+| Epoch boundary (last block of epoch) | `OnchainDkgOutcome` encoded via `commonware_codec` |
+| Regular block with DKG dealing | `SignedDealerLog` encoded via `commonware_codec` |
+| Regular block (no dealing) | Empty |
+
+This TIP prepends a fixed-size consensus context prefix to `extra_data`. The consensus context is encoded using the `commonware_codec::Write` as defined upstream:
+
+```
+[ varint(epoch) ][ varint(view) ][ pubkey_bytes(32) ][ varint(parent_view) ][ digest_bytes(32) ]
```
Zygimantass: is there a reason we keep adding data into extra_data? is there no way to extend the header? seems like we'll eventually run into conflicts / make the header harder to decode

tips/tip-1031.md
```diff
@@ -0,0 +1,144 @@
...
+certify():
+  await background result  →  pass  →  vote finalize
+                           →  fail  →  vote nullify
+```
+
+This migration to `Deferred` verification does not have to be coupled with this TIP. A validator that chooses to synchronize verification with their notarization will not affect consensus. The node simply does not contribute to any decrease in finalization latency.
+
+## Genesis Block
+
+The genesis block encodes a sentinel context (epoch 0, view 0, zero pubkey, zero parent) for consistency in the extra data wire format.
```
Zygimantass: how would this work for chains that already have built the genesis block?

tips/tip-1031.md
```diff
@@ -0,0 +1,144 @@
...
+status: Draft
+related: TIP-1017
+protocolVersion: TBD
+---
+
+# TIP-1031: Embed Consensus Context in Block Header
+
+## Abstract
+
+Tempo blocks currently do not carry any consensus `Context` (round, leader, parent view+digest) inside the block itself. Recovering any consensus information about a block requires the Notarization or Finalization certificates, completely separate from the execution layer. There is no verifiable linkage of consensus metadata in the execution layer.
```
Zygimantass: quacky idea, but should we embed the finalization of block N in block N+1? it might not work right – because the block N+1 isn't guaranteed to get the finalization cert?

Zygimantass: but if it is possible, it might be a nice way of guaranteeing that we have the finalization certs for all blocks?

SuperFluffy: Reduce this to what TIP is proposing and what it would unlock. Dump background stuff (current situation) in motivation.

tips/tip-1031.md
```diff
@@ -0,0 +1,144 @@
...
+
+Introducing consensus metadata into the blocks allows for stateless validation of this information prior to any notarization or finalization certificates. Commonware's new `CertifiableBlock` trait requires this in their new `Deferred` verification mechanism to support optimistic notarizations of blocks, allowing network latency to mask finalization latency with async execution verification.
```
SuperFluffy: Let's link the relevant commonware logic (using their recent release) for reference.

tips/tip-1031.md
hamdiallam: Ooh yea I like this idea a lot. Maybe at least for the view 0 of every epoch to encode the finalization of the boundary. since that IS required.

> but if it is possible, it might be a nice way of guaranteeing that we have the finalization certs for all blocks?

Need to think about this piece some more

hamdiallam: @SuperFluffy this re-surfaces extending the "notarization chain" beyond the locally-finalized tip. Doing this on every block would now prevents this since requiring a finalization would be required to progress. I think this should be okay. Is there any reason we would want the network to have an extended notarized chain that isn't finalized?

A validator is be able to query/validate any certs it doesn't have over the p2p channel. Just some first pass thoughts

tips/tip-1031.md
hamdiallam: Depends on the activation time for the fork. If not activated, the extra_data **must not** have this information.

I'll update this TIP to make it clear this requires a hard fork. Took that as an implied assumption

tips/tip-1031.md
SuperFluffy: I am not sure how this would work. In order to include the finalization of block N in block N+1, you'd need to have that cert available almost immediately upon entering the view that will propose N+1.

But that kind of defeats the purpose of using the whole simplex construction in the first place (namely, that you don't require waiting for finalization in order to have the chain advance).

tips/tip-1031.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Motivation
+
+Opt-into `Deferred` verification mechanism in the consensus engine. Commonware's consensus engine supports a deferred verification flow through two traits:
```
SuperFluffy: Mentioned above, let's provide some references to this, pointing both at the CW github and maybe docs.rs.

hamdiallam: Updated with links to the implementation

tips/tip-1031.md
hamdiallam: Resolving as unrelated.

Have a good thread with commonware on why this wouldn't work for finalizations

tips/tip-1031.md
hamdiallam: Updated to be a dedicated field now

tips/tip-1031.md
```diff
@@ -0,0 +1,103 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Default, RlpEncodable, RlpDecodable)]
+#[cfg_attr(feature = "reth-codec", derive(reth_codecs::Compact))]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct Context {
+  pub epoch: u64,
+  pub view: u64,
+  pub leader: B256,
+  pub parent_view: u64,
+  pub parent_digest: B256,
+}
```
klkvr, line 39: isn't `parent_digest` redundant to `parent_hash`? and `epoch` can technically be derived from block number (unless we've found a way to have dynamic epoch length?)

also could consider carrying `leader` as some kind of index into valset to save some bytes in each block header

tips/tip-1031.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Motivation
+
+The Tempo application currently implements [Automaton::verify](https://github.com/tempoxyz/tempo/blob/main/crates/commonware-node/src/consensus/application/actor.rs#L618), directly passed into simplex, which fully validates the payload with the execution engine *before* a notarize can be cast.
+
+The `Deferred` wrapper speeds up consensus by masking finalization latency behind the network. The notarize is cast [immediately after a cheap check](https://github.com/commonwarexyz/monorepo/blob/main/consensus/src/marshal/standard/deferred.rs#L487) on consensus metadata, spawning verification in the background which gates finalization. Thus forming a notarization certificate is not blocked on verification.
+
+In order to leverage `Deferred`, Tempo blocks must be [CertifiableBlock](https://github.com/commonwarexyz/monorepo/blob/2a588e4e341548333a4bd753c016e814ae0ecca0/consensus/src/lib.rs#L57) which exposes the consensus context that is checked.
```
klkvr: imo would be nice to have an explanation here on why do we need this information on EL block vs before commonware was fully managing consensus-layer context on its own. it's not entirely obvious to me tbh

tips/tip-1031.md
SuperFluffy: Good catch: `parent_digest` can be derived from the block already and is not necessary here.

`epoch` can be made dynamic and should stay here. It is constant right now through `FixedEpocher`, but there is no reason it should stay this way. For future compatibility, we need to maintain it.

Also good point on `leader` - we must always have the participant set for a given epoch (via the DKG outcome at the boundary). And so we can always derive the `leader` from it. It'd be more elegant if this were self-contained, but using 8 vs 32 bytes is a good reason to be economic.

hamdiallam: yep good call on `parent_digest`. Agree with @SuperFluffy , if the 24 bytes are worth the saving, we can use an index

SuperFluffy: We should even body use u16 - two bytes are more than enough

tips/tip-1031.md
SuperFluffy: This goes a bit too far for this TIP. But TL;DR, commonware's execution model changes. Before deferred verification: you got a certificate (notarization) for view V, and immediately exit it and enter V+1.

After deferred verification: you get a certificate for view V and "hop into" the view. To exit the view, you need to execute the block. But to do that, you need to know the leader that proposed it (we actually do when checking dealer logs!). But you only have the certificate for V, not V-1, and so you can't derive the leader.

That's why we need the info on the block.

(This became very non-TL;DR)

tips/tip-1031.md
SuperFluffy: What's nicer about having the full pubkey here is that reconciliation with the dkg outcome isn't necessary. The understanding of the block can stay entirely local to the block.

This might matter if we ever want do to something with this info in the EVM.

klkvr: > epoch can be made dynamic and should stay here. It is constant right now through FixedEpocher, but there is no reason it should stay this way. For future compatibility, we need to maintain it.

looking at `Epocher` API any implementation would define `height -> epoch` conversion? so tracking `epoch` for any epocher would still be redundant i think?

SuperFluffy: > looking at Epocher API any implementation would define height -> epoch conversion? so tracking epoch for any epocher would still be redundant i think?

Yeah, you are right. We should throw `epoch` out and construct it on the fly.

hamdiallam: +1, nice call. I'll update this tip soon with the feedback

tips/tip-1031.md
klkvr: > After deferred verification: you get a certificate for view V and "hop into" the view. To exit the view, you need to execute the block. But to do that, you need to know the leader

I see, and you can trust the leader value in the block header because the notarization condition enforces it?

I think an important high-level part that could be worth mentioning in the TIP is that after this change notarization no longer guarantees that a block is valid but only that consensus context in this block is valid wrt view.

Also seems worth being explicit about which exact validations should happen before "optimistically" notarizing the block

hamdiallam: Sounds good

>Also seems worth being explicit about which exact validations should happen before "optimistically" notarizing the block

Is the Block Verification section not clear?

hamdiallam: Also important to note post activation, this does not mean blocks will be optimistically notarized. This TIP simply includes the data so that we __can__ do this.

It's not consensus breaking to optimistically notarize or not. So I didn't want to couple usage of `Deferred` to this TIP. Explained in the block verification section

hamdiallam: Lifted that point to a bullet in `Motivation` to make clearer. As well as block validity guarantees when `Deferred` is used

tips/tip-1031.md
```diff
@@ -0,0 +1,103 @@
...
+
+During `Automaton::verify()` step:
+
+1. If not activated, the context **MUST** be `None`.
+2. If activated, the context **MUST** be set and match the information provided by the consensus engine.
+3. Continue with verification as-is. 
+
+Important to note the immediate switch to `Deferred` is __not strictly required__. For example a validator can choose an implementation that preserves synchronous verification. This validator simply does not contribute to the reduced latency in forming the notarization certificate.
+
+## Genesis Block
```
SuperFluffy, line 67: I think this needs to be updated to match the context above.

tips/tip-1031.md
```diff
@@ -0,0 +1,103 @@
...
+If activated at genesis, genesis block encodes a sentinel context for consistency in the setting of this field.
+
+```rust
+let sentinel_context = Context {
+  epoch: 0,
+  view: 0,
+  leader: B256::ZERO,
+  parent_view: 0,
+  parent_digest: B256::ZERO,
+};
```
malleshpai, line 77: this doesn't match the struct above

tips/tip-1031.md
```diff
@@ -0,0 +1,103 @@
...
+
+* The migration to `Deferred` implies that the notarization certified will no longer carry any guarantees on the validity of the block. Only ensuring the consensus view is valid.
+
+* The migration to `Deferred` is also not stricly coupled with this TIP. This TIP only asserts the inclusion and validity of the consensus context in the header as a prerequisite to using the `Deferred` implementation. Usage of `Deferred` is not a consensus breaking change, explained in [Block Verification](#block-verification).
+
+---
+
+# Specification
+
+This TIP will be activated in a future hardfork (TBD).
```
jenpaff: THis can be removed, the it already says on the top which protocolVersion this will be activated

malleshpai (approved):
it's nbd i would just simplify the tip about the whole deferred verification stuff. 

Right now, iiuc, the TIP only adds context and verifies that the context is correct. someone reading the motivation might think it is also doing some deferred verification (my mind went there when reading the TIP). if you have a cycle maybe just clarify the motivation + abstract. something crisper like: 

Abstract: This TIP puts consensus context in header. after activation, every block will have correct consensus context in hearder. 

Motivation: This is a necessary step for us to implement deferred verification (blah blah). That will be enabled separately/ later in a subsequent TIP.

tips/tip-1031.md
```diff
@@ -0,0 +1,100 @@
+---
+id: TIP-1031
+title: Embed consensus context in the block Header
+description: Embed consensus context into the block header
+authors: Hamdi, Janis
+status: Draft
+related: N/A
+protocolVersion: TBD
```
jenpaff:
```suggestion
protocolVersion: T4
```

tips/tip-1031.md
```diff
@@ -0,0 +1,100 @@
+---
+id: TIP-1031
+title: Embed consensus context in the block Header
+description: Embed consensus context into the block header
+authors: Hamdi, Janis
+status: Draft
```
jenpaff:
```suggestion
status: Approved
```

PR #3094, opened 2026-03-11
chore(precompiles): assert LIFO order in nested checkpoint usage

0xrusowsky:
ref: `ZC-103`

## Motivation

right now it is (theoretically) possible for nested guards to NOT be resolved in LIFO order - thus committing the wrong checkpoint.

## Solution

pass the checkpoint handle through `checkpoint_commit(checkpoint)`, enabling providers to verify LIFO ordering.

this PR hardens both providers, which now assert correct resolution order in debug/test builds.


crates/precompiles/src/storage/evm.rs
```diff
@@ -243,20 +248,53 @@ impl<'a> PrecompileStorageProvider for EvmPrecompileStorageProvider<'a> {
...
+/// LIFO checkpoint validation (debug builds only).
+///
+/// Since `EvmInternals` doesn't expose revm's journal depth, we mirror it by
+/// recording each checkpoint's `journal_i` on creation and asserting that
+/// commits/reverts always resolve the most recent checkpoint first.
+#[cfg(debug_assertions)]
+impl EvmPrecompileStorageProvider<'_> {
+    /// Records a newly created checkpoint for later LIFO validation.
+    fn track_checkpoint(&mut self, cp: &JournalCheckpoint) {
+        self.checkpoint_stack.push(cp.journal_i);
```
decofe: 🛡️ **[DEFENSE-IN-DEPTH]** Debug LIFO checker tracks only `journal_i`, not full checkpoint identity

Upstream revm's `JournalCheckpoint` is a `(log_i, journal_i)` tuple. `log()` operations mutate only `logs` without growing `journal`, so nested checkpoints created with only LOG operations between them can share the same `journal_i`. The debug assertion would then treat distinct checkpoints as equivalent and miss out-of-order resolution.

Not exploitable today (only one production checkpoint user exists), but weakens the intended "detect misuse early in debug" guarantee for future nested checkpoint users.

**Recommended Fix:** Track the full `JournalCheckpoint` (both `log_i` and `journal_i`) in debug mode:
```rust
#[cfg(debug_assertions)]
checkpoint_stack: Vec<(usize, usize)>,

fn track_checkpoint(&mut self, cp: &JournalCheckpoint) {
    self.checkpoint_stack.push((cp.journal_i, cp.log_i));
}

fn assert_lifo(&mut self, cp: &JournalCheckpoint, op: &str) {
    let top = self.checkpoint_stack.pop()
        .unwrap_or_else(|| panic!("checkpoint_{op}: no active checkpoint"));
    assert_eq!(
        (cp.journal_i, cp.log_i), top,
        "out-of-order checkpoint {op} (expected top of stack)"
    );
}
```

0xrusowsky: addressed in [`5788878` (this PR)](https://github.com/tempoxyz/tempo/pull/3094/commits/57888786298984e4543329273865d712d2a7f125)

PR #3096, opened 2026-03-11
feat: introduce consensus pubkey label for pushed telemetry metrics & logs

hamdiallam:
Metrics: VMAuth supports merging `extra_labels` query args. We can use that to inject the consensus id
Logs: extra_fields does not support merging. However we can wire into `OTEL_RESOURCE_ATTRIBUTES` such that the open telemetry sdk picks it up internally prior to initialization

### DIFF
--- a/bin/tempo/src/main.rs
+++ b/bin/tempo/src/main.rs
@@ -223,22 +223,22 @@ fn main() -> eyre::Result<()> {
             .try_to_config()
             .wrap_err("failed to parse telemetry config")?
     {
-        let consensus_id = node_cmd
+        let consensus_pubkey = node_cmd
             .ext
             .consensus
             .public_key()
             .wrap_err("failed parsing consensus key")?
             .map(|k| k.to_string());
 
-        if let Some(id) = &consensus_id {
+        if let Some(pubkey) = &consensus_pubkey {
             // VictoriaMetrics does not support merging `extra_fields` query args like `extra_labels` for
             // metrics. A workaround for now is to directly hook into the `OTEL_RESOURCE_ATTRIBUTES` env var
             // used at startup to capture contextual information.
             let current = std::env::var("OTEL_RESOURCE_ATTRIBUTES").unwrap_or_default();
             let new_attrs = if current.is_empty() {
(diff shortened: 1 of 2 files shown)


crates/node/src/telemetry.rs
```diff
@@ -78,3 +82,87 @@ pub fn install_prometheus_metrics(
 
     Ok(())
 }
+
+fn attach_labels(metrics: &str, labels: &HashMap<String, String>) -> String {
```
SuperFluffy: This is one heavy op - worried we are going to hammer the nodes with this.

hamdiallam: extra_labels work as long as vmauth merges

bin/tempo/src/main.rs
```diff
@@ -298,10 +236,13 @@ fn main() -> eyre::Result<()> {
 
                 // Start the unified metrics exporter if configured
                 if let Some(config) = telemetry_config {
+                    let consensus_id = args.consensus.public_key()?.map(|k| k.to_string());
```
SuperFluffy: please provide error context

crates/node/src/telemetry.rs
```diff
@@ -41,6 +43,13 @@ pub fn install_prometheus_metrics(
     let endpoint = config.endpoint.to_string();
     let auth_header = config.auth_header;
 
+    let query_params = config
+        .consensus_id
+        .map(|k| format!("?extra_label=consensus_id={k}"))
```
SuperFluffy: This works as long as we only put the public key in here, but will fail in the general case; let's use https://docs.rs/urlencoding/latest/urlencoding/

bin/tempo/src/main.rs
hamdiallam: ~~Will do `public_key()` unilaterally returns `Ok`~~. nvm i see the propagated err

crates/node/src/telemetry.rs
hamdiallam: Moved to a better way of adding query params to the `Url` resource directly before `to_string` which should handle this

PR #3097, opened 2026-03-11
fix(txpool): enforce exact calldata length for TIP-20 payment lane

decofe:
Closed in favor of #3099.

### DIFF
--- a/crates/primitives/src/transaction/envelope.rs
+++ b/crates/primitives/src/transaction/envelope.rs
@@ -14,28 +14,27 @@ use core::fmt;
 /// Same as TIP20_TOKEN_PREFIX
 pub const TIP20_PAYMENT_PREFIX: [u8; 12] = hex!("20C000000000000000000000");
 
-/// ERC-20 `transfer(address,uint256)` selector
-const TRANSFER_SELECTOR: [u8; 4] = hex!("a9059cbb");
-
-/// TIP-20 `transferWithMemo(address,uint256,bytes32)` selector
-const TRANSFER_WITH_MEMO_SELECTOR: [u8; 4] = hex!("95777d59");
-
-/// Expected calldata length for `transfer(address,uint256)`: 4 + 32 + 32 = 68
-const TRANSFER_CALLDATA_LEN: usize = 68;
-
-/// Expected calldata length for `transferWithMemo(address,uint256,bytes32)`: 4 + 32 + 32 + 32 = 100
-const TRANSFER_WITH_MEMO_CALLDATA_LEN: usize = 100;
+/// Valid TIP-20 payment selectors with their expected calldata lengths.
+const PAYMENT_CALLDATA: &[([u8; 4], usize)] = &[
+    (hex!("a9059cbb"), 68),  // transfer(address,uint256)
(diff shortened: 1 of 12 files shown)


decofe (commented):
:eye: **Cyclops Review**

Hardfork-aware payment classification via `PaymentRules` is well-structured. Consensus-critical paths correctly derive rules from block timestamp. One defense-in-depth finding below.

<details>

- ⚡ **Dead cached `is_payment` in `TempoPooledTransaction`**: The `is_payment` field is computed at pool insertion but never consumed by consensus paths. Both the block executor and payload builder recompute classification dynamically from the envelope + active hardfork. The field and all its plumbing (`new()` taking `payment_rules`, `maintain.rs` querying `reorg_payment_rules`, `from_pooled` hardcoding `LATEST`) serve no purpose and could be removed.
- ⚡ **`PaymentRules::LATEST` coupling**: `from_pooled()` hardcodes `PaymentRules::LATEST`. Safe today since consensus paths recompute from timestamp, but will silently misclassify near hardfork boundaries if future pool logic ever relies on the cached value.

</details>

crates/primitives/src/transaction/envelope.rs
```diff
@@ -156,30 +192,35 @@ impl TempoTxEnvelope {
...
-                call.to
-                    .to()
-                    .is_some_and(|to| to.starts_with(&TIP20_PAYMENT_PREFIX))
-            }),
+            Self::Legacy(tx) => is_tip20_payment(tx.tx().to.to(), &tx.tx().input),
+            Self::Eip2930(tx) => is_tip20_payment(tx.tx().to.to(), &tx.tx().input),
+            Self::Eip1559(tx) => is_tip20_payment(tx.tx().to.to(), &tx.tx().input),
+            Self::Eip7702(tx) => is_tip20_payment(Some(&tx.tx().to), &tx.tx().input),
+            Self::AA(tx) => {
+                !tx.tx().calls.is_empty()
```
decofe, line 217: 🛡️ **[DEFENSE-IN-DEPTH]** AA empty-calls guard applied unconditionally across all `PaymentRules` variants

The `!tx.tx().calls.is_empty()` check is applied regardless of `PaymentRules`. Before this PR, `is_payment()` returned `true` for AA txs with empty `calls` (vacuous `all()`). Now `is_payment(PaymentRules::Genesis)` returns `false` for the same input — a semantic change to historical classification.

This is **not exploitable**: empty-call AA txs are structurally rejected by `validate_calls()` (in both pool validation and `validate_env()`) before `validate_tx` ever runs, so the payment classification for empty calls is never reached in practice. However, it breaks the design intent of `PaymentRules::Genesis` preserving pre-T2 semantics.

**Recommended Fix:** Scope the guard to `T2` only:
```rust
Self::AA(tx) => match rules {
    PaymentRules::Genesis => tx
        .tx()
        .calls
        .iter()
        .all(|call| is_tip20_payment(call.to.to(), &call.input)),
    PaymentRules::T2 => {
        !tx.tx().calls.is_empty()
            && tx
                .tx()
                .calls
                .iter()
                .all(|call| is_tip20_payment(call.to.to(), &call.input))
    }
},
```

PR #3099, opened 2026-03-12
fix(txpool): strict payment calldata validation in pool and builder

### LINKED ISSUE #1379: Enforce exact calldata length for TIP-20 payment lane

decofe:
## Motivation

Payment lane classification currently only checks the TIP-20 address prefix. Transactions with unrecognized or oversized calldata can be classified as payments, bypassing the 30M general gas limit and accessing the full 450M+ payment lane budget.

## Solution

Adds `ITIP20Calls::is_payment()` for type-safe calldata matching using `SolCall::SELECTOR` and `SolType::ENCODED_SIZE`, and introduces `is_payment_v2()` on `TempoTxEnvelope` which requires both TIP-20 prefix and recognized calldata. The existing prefix-only check is preserved as `is_payment()` for the block executor (consensus). The tx pool and payload builder use `is_payment_v2()`, so no hardfork gate is needed.

Recognized selectors: `transfer`, `transferWithMemo`, `transferFrom`, `transferFromWithMemo`, `mint`, `mintWithMemo`, `burn`, `burnWithMemo`.

Closes #1379. Supersedes #2575, #2601, #3040, #3097, #3103.

Prompted by: dan

### DIFF
--- a/crates/contracts/src/precompiles/tip20.rs
+++ b/crates/contracts/src/precompiles/tip20.rs
@@ -1,6 +1,7 @@
 pub use IRolesAuth::{IRolesAuthErrors as RolesAuthError, IRolesAuthEvents as RolesAuthEvent};
 pub use ITIP20::{ITIP20Errors as TIP20Error, ITIP20Events as TIP20Event};
 use alloy_primitives::{Address, U256};
+use alloy_sol_types::{SolCall, SolType};
 
 /// USD currency string constant.
 pub const USD_CURRENCY: &str = "USD";
@@ -184,8 +185,7 @@ impl ITIP20::ITIP20Calls {
     ///
     /// [TIP-20 payment]: <https://docs.tempo.xyz/protocol/tip20/overview#get-predictable-payment-fees>
     pub fn is_payment(input: &[u8]) -> bool {
-        fn is_call<C: alloy_sol_types::SolCall>(input: &[u8]) -> bool {
-            use alloy_sol_types::SolType;
+        fn is_call<C: SolCall>(input: &[u8]) -> bool {
             input.first_chunk::<4>() == Some(&C::SELECTOR)
                 && input.len()
                     == 4 + <C::Parameters<'_> as SolType>::ENCODED_SIZE.unwrap_or_default()
@@ -195,6 +195,7 @@ impl ITIP20::ITIP20Calls {
             || is_call::<ITIP20::transferWithMemoCall>(input)
             || is_call::<ITIP20::transferFromCall>(input)
             || is_call::<ITIP20::transferFromWithMemoCall>(input)
+            || is_call::<ITIP20::approveCall>(input)
             || is_call::<ITIP20::mintCall>(input)
             || is_call::<ITIP20::mintWithMemoCall>(input)
             || is_call::<ITIP20::burnCall>(input)
@@ -321,23 +322,22 @@ impl TIP20Error {
 }
 
 #[cfg(test)]
-#[rustfmt::skip]
 mod test {
     use super::*;
-    use alloy_primitives::{Address, B256, U256};
-    use alloy_sol_types::SolCall;
     use alloc::vec::Vec;
+    use alloy_primitives::{Address, B256, U256};
 
+    #[rustfmt::skip]
     /// Returns valid ABI-encoded calldata for every recognized TIP-20 payment selector.
-    fn payment_calldatas() -> [Vec<u8>; 8] {
-        let (to, from, amount, memo) =
-            (Address::random(), Address::random(), U256::random(), B256::random());
+    fn payment_calldatas() -> [Vec<u8>; 9] {
+        let (to, from, amount, memo) = (Address::random(), Address::random(), U256::random(), B256::random());
 
         [
             ITIP20::transferCall { to, amount }.abi_encode(),
             ITIP20::transferWithMemoCall { to, amount, memo }.abi_encode(),
             ITIP20::transferFromCall { from, to, amount }.abi_encode(),
             ITIP20::transferFromWithMemoCall { from, to, amount, memo }.abi_encode(),
+            ITIP20::approveCall { spender: to, amount }.abi_encode(),
             ITIP20::mintCall { to, amount }.abi_encode(),
             ITIP20::mintWithMemoCall { to, amount, memo }.abi_encode(),
             ITIP20::burnCall { amount }.abi_encode(),
@@ -345,6 +345,7 @@ mod test {
         ]
     }
 
+    #[rustfmt::skip]
     /// Returns ABI-encoded calldata for TIP-20 selectors NOT recognized as payments.
     fn non_payment_calldatas() -> [Vec<u8>; 3] {
         let mut data = ITIP20::transferCall { to: Address::random(), amount: U256::random() }.abi_encode();
--- a/crates/primitives/src/transaction/envelope.rs
+++ b/crates/primitives/src/transaction/envelope.rs
@@ -677,13 +677,14 @@ mod tests {
 
     #[rustfmt::skip]
     /// Returns valid ABI-encoded calldata for every recognized TIP-20 payment selector.
-    fn payment_calldatas() -> Vec<Bytes> {
+    fn payment_calldatas() -> [Bytes; 9] {
         let (to, from, amount, memo) = (Address::random(), Address::random(), U256::random(), B256::random());
-        vec![
+        [
             ITIP20::transferCall { to, amount }.abi_encode().into(),
             ITIP20::transferWithMemoCall { to, amount, memo }.abi_encode().into(),
             ITIP20::transferFromCall { from, to, amount }.abi_encode().into(),
             ITIP20::transferFromWithMemoCall { from, to, amount, memo }.abi_encode().into(),
+            ITIP20::approveCall { spender: to, amount }.abi_encode().into(),
             ITIP20::mintCall { to, amount }.abi_encode().into(),
             ITIP20::mintWithMemoCall { to, amount, memo }.abi_encode().into(),
             ITIP20::burnCall { amount }.abi_encode().into(),


crates/primitives/src/transaction/envelope.rs
```diff
@@ -14,6 +14,36 @@ use core::fmt;
...
+const PAYMENT_CALLDATA: &[([u8; 4], usize)] = &[
+    (hex!("a9059cbb"), 68),  // transfer(address,uint256)
+    (hex!("95777d59"), 100), // transferWithMemo(address,uint256,bytes32)
+    (hex!("23b872dd"), 100), // transferFrom(address,address,uint256)
+    (hex!("929c2539"), 132), // transferFromWithMemo(address,address,uint256,bytes32)
+    (hex!("40c10f19"), 68),  // mint(address,uint256)
+    (hex!("e44f0b12"), 100), // mintWithMemo(address,uint256,bytes32)
+    (hex!("42966c68"), 36),  // burn(uint256)
+    (hex!("38f23b0b"), 68),  // burnWithMemo(uint256,bytes32)
+];
```
0xrusowsky: this is too manual imo, better to use `SELECTOR` from `ITIP20Calls`

0xrusowsky (changes requested):
needs better docs, and could be cleaned up

made an improvement proposal in https://github.com/tempoxyz/tempo/pull/3103

crates/primitives/src/transaction/envelope.rs
```diff
@@ -183,6 +213,31 @@ impl TempoTxEnvelope {
         }
     }
 
+    /// Strict payment classification: TIP-20 prefix AND recognized calldata.
+    ///
+    /// Like [`is_payment`](Self::is_payment), but additionally requires the calldata to have a
+    /// recognized payment selector (`transfer`, `transferWithMemo`, `transferFrom`,
+    /// `transferFromWithMemo`) with exact ABI-encoded length. This prevents transactions with
```
0xrusowsky: would defeat the purpose of the payment lane. invalid

crates/primitives/src/transaction/envelope.rs
0xrusowsky: irrelevant in terms of what this PR tries to achieve

0xrusowsky: addressed in [`ac28930` (this PR)](https://github.com/tempoxyz/tempo/pull/3099/commits/ac2893025dda061462604952d87bf18e7fb9c1f8)

crates/primitives/src/transaction/envelope.rs
```diff
@@ -154,32 +166,51 @@ impl TempoTxEnvelope {
             && self.nonce() == 0
     }
 
-    /// Classify a transaction as payment or non-payment.
+    /// [TIP-20 payment] classification: `to` address has the `0x20c0` prefix.
     ///
-    /// Currently uses classifier v1: transaction is a payment if the `to` address has the TIP20 prefix.
-    pub fn is_payment(&self) -> bool {
```
mattsse, line 160: whats the motivation for changign this function?

i assume we want to keep the exact same behaviour for this? then why even touch it?

0xrusowsky: for clarity and not duplicating code, it should be equivalent 1:1

crates/primitives/src/transaction/envelope.rs
```diff
@@ -9,11 +9,23 @@ use alloy_consensus::{
...
+/// Returns `true` if `to` has the TIP-20 payment prefix.
+fn is_tip20_call(to: Option<&Address>) -> bool {
+    to.is_some_and(|to| to.starts_with(&TIP20_PAYMENT_PREFIX))
+}
+
+/// Returns `true` if `to` has the TIP-20 payment prefix and `input` is recognized payment
+/// calldata (selector + exact ABI-encoded length).
+fn is_tip20_payment(to: Option<&Address>, input: &[u8]) -> bool {
+    is_tip20_call(to) && ITIP20::ITIP20Calls::is_payment(input)
+}
```
mattsse: these dont belong top of file

mattsse (changes requested):

crates/primitives/src/transaction/envelope.rs
0xrusowsky: addressed in [`8b226dd` (this PR)](https://github.com/tempoxyz/tempo/pull/3099/commits/8b226dd7e46662017c12e693d0637b370f6d1d11)

crates/primitives/src/transaction/envelope.rs
0xrusowsky: reverted changes in [`04f2cee` (this PR)](https://github.com/tempoxyz/tempo/pull/3099/commits/04f2cee814936fb5a38726a90ed61c39ee0cad3f) to keep the PR focused

crates/primitives/src/transaction/envelope.rs
```diff
@@ -183,6 +191,34 @@ impl TempoTxEnvelope {
         }
     }
 
+    /// Strict [TIP-20 payment] classification: `0x20c0` prefix AND recognized calldata.
+    ///
+    /// Like [`is_payment`](Self::is_payment), but additionally requires calldata to match a
+    /// recognized payment selector with exact ABI-encoded length.
+    ///
+    /// # NOTE
+    /// Builder-level classifier, used by the transaction pool and payload builder to prevent DoS of
```
decofe, line 200: ⚠️ **[ISSUE]** Strict payment classifier ignores gas-heavy sidecar fields, allowing payment-lane capacity abuse

`is_payment_v2()` checks only `to` and `input` but ignores `access_list` (EIP-2930/1559/AA), `authorization_list` (EIP-7702), `tempo_authorization_list`, and `key_authorization` (AA). A transaction with valid TIP-20 `transfer` calldata can attach a large authorization list and still be classified as a payment, bypassing the `general_gas_limit`.

The pool's `max_tx_input_bytes` cap (128 KiB) bounds the maximum list size, but even within that cap, a single EIP-7702 payment-classified transaction can consume ~26M gas through authorization list processing (each `nonce == 0` auth costs 250k gas) — nearly the entire 30M non-payment budget.

**Recommended Fix:** Require sidecar fields to be empty:
```rust
pub fn is_payment_v2(&self) -> bool {
    match self {
        Self::Legacy(tx) => is_tip20_payment(tx.tx().to.to(), &tx.tx().input),
        Self::Eip2930(tx) => tx.tx().access_list.is_empty()
            && is_tip20_payment(tx.tx().to.to(), &tx.tx().input),
        Self::Eip1559(tx) => tx.tx().access_list.is_empty()
            && is_tip20_payment(tx.tx().to.to(), &tx.tx().input),
        Self::Eip7702(tx) => tx.tx().access_list.is_empty()
            && tx.tx().authorization_list.is_empty()
            && is_tip20_payment(Some(&tx.tx().to), &tx.tx().input),
        Self::AA(tx) => {
            tx.tx().access_list.is_empty()
                && !tx.tx().calls.is_empty()
                && tx.tx().calls.iter().all(|call| is_tip20_payment(call.to.to(), &call.input))
        }
    }
}
```

crates/contracts/src/precompiles/tip20.rs
```diff
@@ -171,6 +171,37 @@ crate::sol! {
...
+    /// - `transferFrom` / `transferFromWithMemo`
+    /// - `mint` / `mintWithMemo`
+    /// - `burn` / `burnWithMemo`
+    ///
+    /// # NOTES
+    /// - Only validates calldata; the caller must check the TIP-20 address prefix on `to`.
+    /// - Only selector and exact ABI-encoded length match, no decoding (better performance).
+    ///
+    /// [TIP-20 payment]: <https://docs.tempo.xyz/protocol/tip20/overview#get-predictable-payment-fees>
+    pub fn is_payment(input: &[u8]) -> bool {
```
decofe, line 187: 💡 **[SUGGESTION]** Strict classifier and runtime TIP-20 decoder disagree on trailing calldata

`is_payment()` requires exact ABI-encoded length, but runtime TIP-20 dispatch uses `ITIP20Calls::abi_decode` which accepts trailing bytes for static parameter tuples. A `transfer(address,uint256)` call with valid args plus trailing bytes will execute successfully at runtime but be classified as non-payment by `is_payment_v2()`. This is a conservative mismatch (false-negative, not false-positive), so not exploitable, but creates policy/runtime coherence drift.

**Recommended Fix:** Either align the runtime decoder to reject trailing bytes, or document the intentional conservative behavior.

crates/contracts/src/precompiles/tip20.rs
```diff
@@ -171,6 +171,37 @@ crate::sol! {
...
+        fn is_call<C: alloy_sol_types::SolCall>(input: &[u8]) -> bool {
+            use alloy_sol_types::SolType;
+            input.first_chunk::<4>() == Some(&C::SELECTOR)
+                && input.len()
+                    == 4 + <C::Parameters<'_> as SolType>::ENCODED_SIZE.unwrap_or_default()
+        }
+
+        is_call::<ITIP20::transferCall>(input)
+            || is_call::<ITIP20::transferWithMemoCall>(input)
```
0xKitsune, line 195: Can we add tests for this function?

crates/primitives/src/transaction/envelope.rs
```diff
@@ -413,6 +449,17 @@ impl From<TempoTransaction> for TempoTypedTransaction {
...
+/// Returns `true` if `to` has the TIP-20 payment prefix.
+fn is_tip20_call(to: Option<&Address>) -> bool {
+    to.is_some_and(|to| to.starts_with(&TIP20_PAYMENT_PREFIX))
+}
+
+/// Returns `true` if `to` has the TIP-20 payment prefix and `input` is recognized payment
+/// calldata (selector + exact ABI-encoded length).
+fn is_tip20_payment(to: Option<&Address>, input: &[u8]) -> bool {
+    is_tip20_call(to) && ITIP20::ITIP20Calls::is_payment(input)
+}
```
0xKitsune, line 461: nit: is there a reason these are two separate functions rather than just having `is_tip20_payment()` and inlining the `is_tip20_call` logic?

0xKitsune (approved):

crates/contracts/src/precompiles/tip20.rs
0xrusowsky: added here: [`5d0917f` (this PR)](https://github.com/tempoxyz/tempo/pull/3099/commits/5d0917fcaf7fa539719f3a95afdc3711947ed020)

crates/primitives/src/transaction/envelope.rs
0xrusowsky: was initially used to refactor `is_payment`(v1), so that both v1 and v2 versions used the same helper.

i left it cause i'll use it in the followup cleanup PR

0xrusowsky: addressed in https://github.com/tempoxyz/tempo/pull/3123

PR #3107, opened 2026-03-12
chore(precompiles): cleanup

0xrusowsky:
this PR:
- drops `encode_to_slot` / `decode_from_slot` in favor of typed patterns leveraging `StorageCtx`
- remove dead dispatch code and unused helpers
- exposes precompile constants from `tempo-contracts` so alloy consumers don't need `tempo-precompiles` deps
- clean up stale TODOs and comments

> NOTE: no behavioral change

### DIFF
--- a/crates/precompiles/src/stablecoin_dex/mod.rs
+++ b/crates/precompiles/src/stablecoin_dex/mod.rs
@@ -767,11 +767,9 @@ impl StablecoinDEX {
 
     /// Fill an order and delete from storage. Returns the next best order and price level.
     ///
-    /// NOTE: Maker transfer policy authorization is intentionally NOT checked here.
-    /// Fills must never fail due to unexecutable orders from blacklisted makers, as that
-    /// would force takers to pay gas eating through stale liquidity. Instead, anyone can
-    /// call [`cancel_stale_order`](Self::cancel_stale_order) to remove orders from
-    /// blacklisted makers before they are filled.
+    /// NOTE: Maker transfer policy is not enforced here to not block swaps on the pair.
+    /// Note that TIP403 checks on order placement and withdraws are enforced.
+    /// [`cancel_stale_order`](Self::cancel_stale_order) can be used to remove orders.
     fn fill_order(
         &mut self,
         book_key: B256,
--- a/tips/ref-impls/src/StablecoinDEX.sol
+++ b/tips/ref-impls/src/StablecoinDEX.sol
@@ -551,10 +551,9 @@ contract StablecoinDEX is IStablecoinDEX {
     }
 
     /// @notice Fill an order and handle cleanup when fully filled
-    /// @dev Maker transfer policy authorization is intentionally NOT checked here.
-    /// Fills must never fail due to unexecutable orders from blacklisted makers, as that
-    /// would force takers to pay gas eating through stale liquidity. Instead, anyone can
-    /// call `cancelStaleOrder` to remove orders from blacklisted makers before they are filled.
+    /// @dev Maker transfer policy is not enforced here to not block swaps on the pair.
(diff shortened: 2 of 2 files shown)


crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -766,6 +766,12 @@ impl StablecoinDEX {
     }
 
     /// Fill an order and delete from storage. Returns the next best order and price level.
+    ///
+    /// NOTE: Maker transfer policy authorization is intentionally NOT checked here.
+    /// Fills must never fail due to unexecutable orders from blacklisted makers, as that
+    /// would force takers to pay gas eating through stale liquidity. Instead, anyone can
+    /// call [`cancel_stale_order`](Self::cancel_stale_order) to remove orders from
+    /// blacklisted makers before they are filled.
```
0xKitsune: nit: The reason that Maker transfer policy auth is not checked is to ensure that swaps can complete, rather than for gas efficiency. Consider updating the comment to something like the following:



```
    /// NOTE: Maker transfer policy is not enforced here to not block swaps on the pair. 
    /// Note that TIP403 checks on order placement and withdraws are enforced.
    /// [`cancel_stale_order`](Self::cancel_stale_order) can be used to remove orders. 
```

tips/ref-impls/src/StablecoinDEX.sol
```diff
@@ -551,6 +551,10 @@ contract StablecoinDEX is IStablecoinDEX {
     }
 
     /// @notice Fill an order and handle cleanup when fully filled
+    /// @dev Maker transfer policy authorization is intentionally NOT checked here.
+    /// Fills must never fail due to unexecutable orders from blacklisted makers, as that
+    /// would force takers to pay gas eating through stale liquidity. Instead, anyone can
+    /// call `cancelStaleOrder` to remove orders from blacklisted makers before they are filled.
```
0xKitsune: nit: same thing here, it is important to not fail on 403 policies during swaps to ensure swap execution rather than this being a gas optimization.

crates/precompiles/src/tip_fee_manager/dispatch.rs
```diff
@@ -44,7 +44,7 @@ impl Precompile for TipFeeManager {
                 view(call, |c| self.user_tokens(c))
             }
             TipFeeManagerCall::FeeManager(IFeeManagerCalls::validatorTokens(call)) => {
-                view(call, |c| self.validator_tokens(c))
+                view(call, |c| self.get_validator_token(c.validator))
```
0xKitsune, line 47: I think this should be fine, but lets make sure re-execution passes. Previously `self.validator_tokens` returned a Vec.

0xrusowsky: i dropped it cause it returned an `Address`, so it's identical:

```rs
    /// Reads the stored fee token preference for a validator, defaulting to [`DEFAULT_FEE_TOKEN`].
    pub fn validator_tokens(&self, call: IFeeManager::validatorTokensCall) -> Result<Address> {
        let token = self.validator_tokens[call.validator].read()?;

        if token.is_zero() {
            Ok(DEFAULT_FEE_TOKEN)
        } else {
            Ok(token)
        }
    }
```
```rs
    /// Returns the validator's preferred fee token, falling back to [`DEFAULT_FEE_TOKEN`].
    pub fn get_validator_token(&self, beneficiary: Address) -> Result<Address> {
        let token = self.validator_tokens[beneficiary].read()?;

        if token.is_zero() {
            Ok(DEFAULT_FEE_TOKEN)
        } else {
            Ok(token)
        }
    }
```

PR #3108, opened 2026-03-12
chore: bump reth to latest main

klkvr:


bin/tempo/Cargo.toml
```diff
@@ -74,13 +74,13 @@ name = "tempo"
 path = "src/main.rs"
 
 [features]
-default = ["asm-keccak", "jemalloc", "otlp"]
+efault = ["asm-keccak", "jemalloc", "otlp"]
```
joshieDo:
```suggestion
default = ["asm-keccak", "jemalloc", "otlp"]
```

klkvr: oops good catch lol

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -575,13 +551,53 @@ where
...
-        let validated = self
-            .protocol_pool
+
+        // Fully delegate to protocol pool for non-2D transactions
+        if !transactions.iter().any(|tx| tx.is_aa_2d()) {
+            return self
+                .protocol_pool
+                .add_transactions(origin, transactions)
+                .await;
+        }
```
joshieDo, line 561: why dont we only add the ones which are aa 2d ? if one is, all are?

Cargo.toml
```diff
@@ -118,68 +118,68 @@ tempo-contracts = { path = "crates/contracts", default-features = false }
...
-reth-rpc-builder = { git = "https://github.com/paradigmxyz/reth", branch = "tempo/v1.10.2" }
-reth-rpc-convert = { git = "https://github.com/paradigmxyz/reth", branch = "tempo/v1.10.2" }
-reth-rpc-eth-api = { git = "https://github.com/paradigmxyz/reth", branch = "tempo/v1.10.2" }
-reth-rpc-eth-types = { git = "https://github.com/paradigmxyz/reth", branch = "tempo/v1.10.2" }
-reth-rpc-server-types = { git = "https://github.com/paradigmxyz/reth", branch = "tempo/v1.10.2" }
-reth-storage-api = { git = "https://github.com/paradigmxyz/reth", branch = "tempo/v1.10.2" }
-reth-tracing = { git = "https://github.com/paradigmxyz/reth", branch = "tempo/v1.10.2" }
-reth-transaction-pool = { git = "https://github.com/paradigmxyz/reth", branch = "tempo/v1.10.2" }
-reth-trie-common = { git = "https://github.com/paradigmxyz/reth", branch = "tempo/v1.10.2" }
+reth-basic-payload-builder = { git = "https://github.com/paradigmxyz/reth", rev = "9060c50" }
```
decofe, line 121: 🚨 **[SECURITY]** New upstream auth-RPC `reth_newPayload` feeds validator-less payloads into Tempo's trusted-block path

The reth bump to `9060c50` silently adds `reth_newPayload` and `reth_forkchoiceUpdated` to Tempo's auth/JWT module. Upstream `create_auth_module()` now unconditionally merges `reth_*` endpoints, and the auth server is enabled by default.

When a block arrives via `reth_newPayload(BlockRlp)`, it routes through `TempoPayloadTypes::block_to_payload()` which always sets `validator_set: None`. In Tempo's EVM, `None` is a trusted/finalized shortcut that skips subblock validator membership, signature verification, duplicate-subblock detection, and per-validator shared-gas checks (`crates/evm/src/block.rs:216-290`). A JWT-authenticated caller can submit blocks with forged `SubBlockMetadata` and drive them to head via `reth_forkchoiceUpdated`.

**Recommended Fix:**
1. Disable the auth server by default for Tempo nodes using `NoopEngineApiBuilder`, **or**
2. Override auth module construction so `reth_newPayload`/`reth_forkchoiceUpdated` are not registered, **or**
3. Reject payloads arriving without a trusted `validator_set`.

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -575,13 +551,53 @@ where
...
             .validate_transactions_with_origin(origin, transactions)
-            .await;
+            .await
+            .into_iter()
+            .map(|outcome| self.add_validated_transaction(origin, outcome))
+            .collect()
+    }
 
-        self.add_validated_transactions(origin, validated)
+    async fn add_transactions_with_origins(
```
decofe, line 572: ⚠️ **[ISSUE]** Mixed AA-2D batch insertion bypasses upstream batch-discard reconciliation

When any AA-2D transaction is present, the batch validates then inserts each tx one-by-one via `add_validated_transaction()`. This per-item path never runs the upstream end-of-batch `discard_worst()` reconciliation, so an earlier `Ok(hash)` can become non-resident when a later higher-priority insertion evicts it. Batching is enabled by default (`max_batch_size=50000`), so this is the normal high-throughput RPC path.

**Recommended Fix:** Preserve the upstream atomic batch path for the non-2D subset, or collect all per-item results and run a final discarded-hash reconciliation pass before returning.

Cargo.toml
klkvr: this is fine, auth server should not be exposed and requires auth anyway

crates/transaction-pool/src/tempo_pool.rs
klkvr: not sure if it really matters and this is the case on current main as well so would prefer this to be a follow up if we choose to fix

shekhirin (approved):

PR #3116, opened 2026-03-12
test(rpc): add devnet test matrix

0xrusowsky:
ths PR refactors remote RPC tests: renames `Testnet` → `RpcEnv`, moves `testnet.rs` → `rpc.rs`, and adds a `test_matrices_devnet` test that mirrors the testnet matrix.

Both testnet and devnet tests require their respective env var
(`TEMPO_TESTNET_RPC_URL`, `TEMPO_DEVNET_RPC_URL`) — skipped when unset.

> NOTE: since T1C has already been activated on all networks, we can now drop `fn uses_legacy_keyauth_pool_validation`

### DIFF
--- a/crates/node/tests/it/tempo_transaction/rpc.rs
+++ b/crates/node/tests/it/tempo_transaction/rpc.rs
@@ -83,10 +83,6 @@ impl super::types::TestEnv for RpcEnv {
         self.hardfork
     }
 
-    fn uses_legacy_keyauth_pool_validation(&self) -> bool {
-        false
-    }
-
     async fn fund_account(&mut self, addr: Address) -> eyre::Result<U256> {
         let tx_hashes: Vec<B256> = self
             .provider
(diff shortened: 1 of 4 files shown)


crates/node/tests/it/tempo_transaction/rpc.rs
```diff
@@ -74,7 +84,7 @@ impl super::types::TestEnv for Testnet {
     }
 
     fn uses_legacy_keyauth_pool_validation(&self) -> bool {
-        true
+        false
```
fgimenez: could we add a note about why do we need this change?

0xrusowsky: i should actually simply deprecate the fn now that T1C is active everywhere

0xrusowsky: addressed in [`180eaef` (this PR)](https://github.com/tempoxyz/tempo/pull/3116/commits/180eaef994bf087ed0ce39f0873168159e60fc65)

PR #3135, opened 2026-03-15
feat(payload): add per-tx and state change tracing spans

decofe:
Adds two new `debug_span!` emissions to the payload builder for blockscope observability:

**Per-tx execution span** (`execute_tx`)
- Wraps each transaction execution in `block_fill` with `tx_hash` and `gas_used` attributes
- Enables blockscope to show per-tx breakdown within a block build

**Block-level state changes span** (`state_changes`)
- Emitted after `finish_block` with `accounts_changed`, `storage_slots_changed`, `accounts_with_storage_changes`
- Gated behind `tracing::enabled!` to avoid unnecessary iteration when debug tracing is off
- Unblocks RETH-594/595/596 (per-tx detail, LLM API, correlation graphs in blockscope)

Both follow the existing `debug_span!(target: "payload_builder", ...)` convention.

Prompted by: yk

### DIFF
--- a/crates/payload/builder/src/lib.rs
+++ b/crates/payload/builder/src/lib.rs
@@ -448,8 +448,16 @@ where
                 .then(|| format!("{:?}", pool_tx.transaction))
                 .unwrap_or_default();
 
+            let tx_hash = *pool_tx.hash();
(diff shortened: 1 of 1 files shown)


yongkangc (approved):

yongkangc (dismissed):
-

PR #3136, opened 2026-03-15
fix(tips): harden TempoStreamChannel close/tombstone and deposit checks

legion2002:
Deletes channel structs on close/withdraw and stores a tombstone bit so closed channel IDs cannot be reopened while reducing per-channel state retained after finalization.

Also hardens token handling by requiring TIP20 addresses, so that we can use transfer/transferFrom, without performing pre/post balance checks

Adds comprehensive Solidity tests for tombstone behavior, deleted-state semantics, reopen prevention, invalid token rejection, and short-credit open/topUp failures.

### DIFF
--- a/.github/workflows/specs.yml
+++ b/.github/workflows/specs.yml
@@ -184,7 +184,7 @@ jobs:
           # Update all tempo dependencies to point to the current commit
           # Using # as delimiter since URLs contain /
           sed -i -E \
-            "s#(tempo-[a-z-]+) = \{ git = \"https://github.com/tempoxyz/tempo\", rev = \"[^\"]+\" \}#\1 = { git = \"https://github.com/tempoxyz/tempo\", rev = \"$COMMIT_SHA\" }#g" \
+            "s#(tempo-[a-z-]+) = \{ git = \"https://github.com/tempoxyz/tempo\", rev = \"[^\"]+\"([^}]*)\}#\1 = { git = \"https://github.com/tempoxyz/tempo\", rev = \"$COMMIT_SHA\"\2}#g" \
             Cargo.toml
 
           echo "Updated Cargo.toml:"


tips/ref-impls/src/TempoStreamChannel.sol
```diff
@@ -380,26 +395,26 @@ contract TempoStreamChannel is EIP712 {
...
+        emit ChannelClosed(channelId, payer, payee, settledAmount, refund);
     }
 
     // --- View Functions ---
 
     /**
      * @notice Get channel state.
      */
-    function getChannel(bytes32 channelId) external view returns (Channel memory) {
+    function getChannel(bytes32 channelId) external view returns (Channel memory channelState) {
```
danrobinson: Why this change?

0xrusowsky: agreed with dan, this change seems useless?

tips/ref-impls/src/TempoStreamChannel.sol
```diff
@@ -467,4 +482,17 @@ contract TempoStreamChannel is EIP712 {
...
+
+    function _clearAndFinalize(Channel storage channel) internal {
+        channel.payer = address(0);
+        channel.payee = address(0);
+        channel.token = address(0);
+        channel.authorizedSigner = address(0);
+        channel.deposit = 0;
+        channel.settled = 0;
+        channel.closeRequestedAt = 0;
+        channel.finalized = true;
```
0xrusowsky: nit: isn't it clearer to delete the struct and then set `channel.finalized = true`?

tips/ref-impls/src/TempoStreamChannel.sol
```diff
@@ -158,9 +163,11 @@ contract TempoStreamChannel is EIP712 {
             finalized: false
         });
 
-        bool success = ITIP20(token).transferFrom(msg.sender, address(this), deposit);
-        if (!success) {
-            revert TransferFailed();
+        if (deposit > 0) {
```
horsefacts: not a safety issue but why allow zero deposit at all?

tips/ref-impls/src/TempoStreamChannel.sol
horsefacts, line 189: slightly weird behavior here: `topUp(0)` will make no transfers but cancel an open close request.

tips/ref-impls/src/TempoStreamChannel.sol
```diff
@@ -467,4 +482,17 @@ contract TempoStreamChannel is EIP712 {
         }
     }
 
+    // --- Internal Functions ---
+
+    function _clearAndFinalize(Channel storage channel) internal {
```
horsefacts: since we are clearing storage here it might be nice for quality of life to emit all the values in an event before tombstoning: 

```solidity
  event ChannelFinalized(
      bytes32 indexed channelId,
      address indexed payer,
      address indexed payee,
      address token,
      address authorizedSigner,
      uint128 deposit,
      uint128 settled,
      uint64 closeRequestedAt,
  );
```

legion2002: I don't think it is worth the gas, we already emit `ChannelClosed` events wherever we use it.

tips/ref-impls/src/TempoStreamChannel.sol
legion2002: yeah just blocking this

tips/ref-impls/src/TempoStreamChannel.sol
legion2002: yup resolved

tips/ref-impls/src/TempoStreamChannel.sol
legion2002: yeah think bot picked up some solidity convention. removed it.

PR #3140, opened 2026-03-16
chore(e2e): el peering after startup, avoid manual port reservation

hamdiallam:
The kernel can re-assign the free'd port before reth is able to bind it


crates/e2e/src/tests/restart.rs
```diff
@@ -202,69 +215,14 @@ async fn ensure_no_progress(context: &Context, tries: u32) {
 /// This is the simplest possible restart case: the network stops because we
 /// dropped below quorum. The node should be able to pick up after.
 #[test_traced]
-fn network_resumes_after_restart_with_el_p2p() {
```
hamdiallam, line 205: This never actually tested anything compared to the `without_el_p2p` test. Since the network halts, CL Engine will still continue with the restart.

Didn't make sense to adjust since we already have sync tests with el p2p

SuperFluffy: Yeah, that makes sense. Looking at the test, there is no mechanism to force just the EL sync.

crates/e2e/src/tests/v2_at_genesis/restart.rs
SuperFluffy: I refactored this one a bit compared to the pre-v2 test.

Should we remove the connection here as well?

It might make sense to keep it because the restart is sufficiently far in the future (triggering a pipeline sync). But there is nothing in the test that actually asserts that.

hamdiallam: Yea for this test it's needed when there is a gap of 3 epochs, or the test takes too long to run

PR #3141, opened 2026-03-16
feat(alloy): ergonomic access key support

onbjerg:
Add `AccessKeyAccount`, `TempoWallet`. `AccessKeyFiller`, and `TempoProviderBuilder` so that access key transactions can be sent via the standard `send_transaction` flow without manual hash computation or raw transaction submission.

For authorizations use `AccessKeyAccount::authorize_with`, otherwise omit it.


crates/alloy/src/fillers/access_key.rs
```diff
@@ -0,0 +1,98 @@
...
+///
+/// When an [`AccessKeyAccount`] is configured, this filler sets:
+/// - `from` to the root account address (the sender)
+/// - `key_id` for accurate gas estimation and AA type detection
+/// - `key_authorization` on the first transaction only (consumed after use)
+#[derive(Debug)]
+pub struct AccessKeyFiller {
+    account: AccessKeyAccount,
+    /// Key authorization consumed on first use.
+    key_authorization: Mutex<Option<SignedKeyAuthorization>>,
```
0xrusowsky: i think since this is consumed on first usage, it doesn't really need shared mutable state?

instead we could leverage alloy's filler pipeline with: `type Fillable = Option<SignedKeyAuthorization>`.

`fn fill_sync()` would only handle `from` and `key_id`, while `key_authorization` would flows through the `fn prepare()` + `fn fill()`, getting consumed exactly once per request.

crates/alloy/src/account.rs
```diff
@@ -0,0 +1,114 @@
...
+///
+/// Skip `authorize_with` — only the signer and root address are needed:
+///
+/// ```ignore
+/// let access_key = AccessKeyAccount::new(existing_signer, root.address());
+/// ```
+#[derive(Clone, Debug)]
+pub struct AccessKeyAccount {
+    /// The access key signer.
+    signer: PrivateKeySigner,
```
0xrusowsky, line 39: note: `PrivateKeySigner` is hardcoded, but the protocol also supports `P256` and `WebAuthn` access keys.

despite alloy doesn't have `SignerSync` impl for those types, should we have it as a todo?

crates/alloy/src/provider/tempo_builder.rs
```diff
@@ -0,0 +1,48 @@
...
+/// [`AccessKeyFiller`], a nonce filler, and the [`TempoWallet`] so that
+/// `send_transaction` works seamlessly for access key accounts.
+///
+/// # Example
+///
+/// ```ignore
+/// let provider = TempoProviderBuilder::with_access_key(access_key)
+///     .on_http(rpc_url);
+/// ```
+pub struct TempoProviderBuilder;
```
0xrusowsky: there's already `TempoProviderBuilderExt` in `ext.rs` with `with_random_2d_nonces()`, `with_expiring_nonces()`, etc.

imo `with_access_key()`should live there instead of creating a new standalone struct.

0xrusowsky (approved):
nice, way better with the `AccessKeyFiller`

PR #3143, opened 2026-03-16
docs(tip-1033): two-hop FeeAMM routing

decofe:
Adds TIP-1033: a minimal two-hop fallback for FeeAMM swaps when there's no direct liquidity between userToken and validatorToken.

Single path only: `userToken → userToken.quoteToken() → validatorToken`. Both hops at standard M rate (0.997), validator gets M² ≈ 0.994009.

Key design points:
- Reserve liquidity in both pools during pre-tx (extends T1C+ reservation)
- Store intermediate token in transient storage for post-tx
- Block `completeQuoteTokenUpdate` on the fee token during two-hop txs to prevent routing inconsistency

Design-only TIP, no implementation.

Prompted by: dan


decofe (commented):
## TIP-1033 Review

### protocolVersion → T5

### Security / Correctness Issues (High Severity)

1. **Rounding must be sequential, not combined — this is consensus-critical.** The spec shows `out2 = actualSpending × 9970 × 9970 / (10000 × 10000)` as an explanation, but also says "Integer division at each step." These produce different results. The spec MUST be unambiguous:
   ```
   out1 = floor(actualSpending * 9970 / 10000)
   out2 = floor(out1 * 9970 / 10000)
   ```
   vs.
   ```
   out2 = floor(actualSpending * 99400900 / 100000000)
   ```
   If different clients implement different formulae, they will disagree on block validity.

2. **Reservation must cover both pools atomically.** The spec says "Reserve liquidity in both pools (transient storage, per T1C+)." But it must also specify:
   - Whether public AMM operations (`mint`, `burn`, `rebalanceSwap`) during the tx must respect both reservations
   - Whether a second two-hop tx in the same block can double-reserve the intermediate pool
   - What happens if the second pool has insufficient liquidity for `compute_amount_out(out1)` where `out1` already includes the first hop's rounding

3. **`quoteToken()` returning zero / invalid / self-referential.** The edge case table says "`userToken.quoteToken() == userToken` cannot happen." This should be an explicit revert condition, not an assumption. If a future TIP-20 token violates this invariant, the fee path should fail cleanly rather than loop.

4. **What if `intermediateToken.quoteToken() != userToken.quoteToken()`?** The spec only checks one hop: `userToken → userToken.quoteToken() → validatorToken`. But it should clarify that no further hops are attempted and no chain-walking occurs.

### Suggested changes
```diff
-protocolVersion: TBD
+protocolVersion: T5
```

Make the rounding rule normative:
```markdown
### Rounding Rule (Normative)

Two-hop output MUST be computed as two sequential integer divisions:

    out1 = floor(actualSpending × 9970 / 10000)
    out2 = floor(out1 × 9970 / 10000)

Implementations MUST NOT combine the two steps into a single multiplication.
```

tips/tip-1033.md
```diff
@@ -0,0 +1,126 @@
...
+
+Total overhead is bounded: ~5 additional storage operations. No loops or unbounded computation.
+
+---
+
+# Rationale
+
+### Why only one path?
+
+A general router (LCA walk, BFS, etc.) adds complexity, gas cost, and attack surface for minimal benefit. The `quoteToken` chain provides a natural routing hint that covers the common case: `userToken → userToken.quoteToken() → validatorToken` is almost always available if the direct pool isn't.
```
danrobinson: I wouldn't say "almost always available"—I would say "is arguably the most likely two-step path to be available if the direct pool isn't".

tips/tip-1033.md
```diff
@@ -0,0 +1,126 @@
...
+
+Total overhead is bounded: ~5 additional storage operations. No loops or unbounded computation.
+
+---
+
+# Rationale
+
+### Why only one path?
+
+A general router (LCA walk, BFS, etc.) adds complexity, gas cost, and attack surface for minimal benefit. The `quoteToken` chain provides a natural routing hint: `userToken → userToken.quoteToken() → validatorToken` is arguably the most likely two-step path to be available if the direct pool isn't.
```
0xrusowsky, line 143: another reason worth mentioning is that `quoteToken` can be set (are not static).
therefore, we can expect TIP20 admins to configure it to the most liquid tokens, benefiting DEX and FeeAMM routing.

0xrusowsky (approved):

tips/tip-1033.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Abstract
+
+When a user's fee token differs from the validator's fee token, the FeeManager swaps the fee through the FeeAMM at a fixed rate of M = 0.9970 (30 bps). Today, if the direct pool (`userToken → validatorToken`) has insufficient liquidity, the transaction is rejected.
+
+TIP-1033 adds a single fallback path: `userToken → userToken.quoteToken() → validatorToken`. Both hops use the same M rate, so the validator receives `amount × M²` (≈ 0.994009, or ~60 bps total). No general routing or path search is introduced—only one specific two-hop path is checked.
+
+## Motivation
+
+As the number of TIP-20 tokens grows, maintaining deep FeeAMM liquidity for every possible `(userToken, validatorToken)` pair becomes impractical. Most tokens have a `quoteToken` that points to a liquid hub token. By routing through this intermediate token, we can support fee swaps between any pair of tokens that share a common `quoteToken` ancestor, without requiring direct pool liquidity.
```
malleshpai: wait why is this common quoteToken? shouldn't it just be as long as userToken -> quoteToken and quoteToken -> validatorToken have liquidity, even if userToken and validatorToken don't share a common quoteToken?

tips/tip-1033.md
```diff
@@ -0,0 +1,126 @@
...
+| `userToken.quoteToken() == validatorToken` | Two-hop degenerates to single-hop pair, which already failed. Revert. |
+| `userToken.quoteToken() == userToken` | Cannot happen—TIP-20 token graph does not allow self-quoting. |
+| Either two-hop pool has insufficient liquidity | Revert with `InsufficientLiquidity`. |
+| Transaction drains intermediate pool during execution | Prevented by liquidity reservation in both pools (T1C+). |
+| `completeQuoteTokenUpdate` called on fee token during tx | No effect on fee routing. The intermediate token was captured in transient storage at pre-tx time. |
+
+## Gas Overhead
+
+Two-hop routing adds:
+- 2 additional storage reads in pre-tx (second pool check + `quoteToken()`)
```
howydev, line 107: nit: 1 more transient storage write (reserve liq on the 2nd pool)

howydev (approved):
1 nit otherwise LGTM

tips/tip-1033.md
danrobinson: Yes, good call

tips/tip-1033.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Overview
+
+The change affects two protocol-level calls on the FeeManager: `collect_fee_pre_tx` and `collect_fee_post_tx`. A new transient storage field stores the intermediate token address when two-hop routing is used.
+
+## Pre-Transaction: `collect_fee_pre_tx`
+
+When `userToken != validatorToken`:
+
+1. **Try direct pool.** Check liquidity in pool `(userToken, validatorToken)`. If sufficient, proceed as today (single hop). Reserve liquidity per T1C+.
```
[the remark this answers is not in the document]
0xKitsune: @danrobinson this finding seems valid but low/informational. Looks like the if the pool has < 30M a user could submit a tx where `gas_limit > pool_liquidity < 30M` and force a 2 hop swap. The liquidity needed to not hit this trivially low and this should just be documented, no spec change needed IMO.

0xKitsune (approved):
Agreed with @malleshpai [here](https://github.com/tempoxyz/tempo/pull/3143#discussion_r3060175030). Pending this update, looks good to me.

tips/tip-1033.md
```diff
@@ -0,0 +1,126 @@
...
+id: TIP-1033
+title: Two-Hop FeeAMM Routing
+description: Adds a two-hop fallback path through the FeeAMM when there is insufficient liquidity in the direct userToken→validatorToken pool.
+authors: Daniel Robinson
+status: Draft
+related: TIP-1000, FeeAMM, FeeManager
+protocolVersion: T5
+---
+
+# TIP-1033: Two-Hop FeeAMM Routing
```
0xKitsune, line 11: The spec is missing an `Invariants` section. Are there any new invariants that should be tested with this change? One that comes to mind: The validator-credited amount should equal `floor(floor(actualSpending × M / SCALE) × M / SCALE)` exactly.

horsefacts (approved):

tips/tip-1033.md
0xrusowsky: addressed in [`0f0e1fc` (this PR)](https://github.com/tempoxyz/tempo/pull/3143/commits/0f0e1fc37d4b0b5a7e5e69984ea44efab6e21523)

tips/tip-1033.md
0xrusowsky: addressed in [`149865a` (this PR)](https://github.com/tempoxyz/tempo/pull/3143/commits/149865a6c9033ed5383c8b241eedb2cfea519efd)

tips/tip-1033.md
```diff
@@ -0,0 +1,149 @@
+---
+id: TIP-1033
+title: Two-Hop FeeAMM Routing
+description: Adds a two-hop fallback path through the FeeAMM when there is insufficient liquidity in the direct userToken→validatorToken pool.
+authors: Daniel Robinson
+status: Draft
```
jenpaff:
```suggestion
status: Approved
```

tips/tip-1033.md
danrobinson: I think can leave that implicit

tips/tip-1033.md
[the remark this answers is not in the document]
decofe: Added a note directly in the spec (step 1 of `collect_fee_pre_tx`) documenting this behavior: high gas limits can force two-hop routing, but pools need only modest liquidity to avoid it unless gas prices are very high. No spec change needed—this is informational.

tips/tip-1033.md
decofe: Good catch — updated to 2 transient storage writes (intermediate token + second pool reservation) and bumped the total to ~6 operations.

tips/tip-1033.md
```diff
@@ -0,0 +1,126 @@
...
+
+When a fee swap is needed and `intermediateToken` is set (non-zero) in transient storage:
+
+1. Read `intermediateToken` from transient storage.
+2. Execute two chained swaps:
+   ```
+   out1 = execute_fee_swap(userToken, intermediateToken, actualSpending)
+   out2 = execute_fee_swap(intermediateToken, validatorToken, out1)
+   ```
+3. Accumulate `out2` as the validator's collected fees.
```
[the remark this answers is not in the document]
decofe: This is not an actual issue. When `userToken = pathUSD`, its `quoteToken` is `Address::ZERO`. The fallback path would be `pathUSD → Address::ZERO → validatorToken`, but there is no FeeAMM pool with `Address::ZERO` as one of the tokens, so the two-hop liquidity check will fail and the tx reverts with `InsufficientLiquidity` — the sentinel is never written to transient storage. In other words, zero-root tokens simply don't get two-hop routing, which is the correct behavior since `pathUSD` is a hub token that should have direct pools to everything.

PR #3144, opened 2026-03-16
fix: never propose blocks with timestamp before the parent

decofe:
## Summary

If a malicious proposer builds a block with a timestamp in the future, honest proposers will continue setting timestamp to the current time, which results in their blocks being invalid.

This PR ensures parent timestamp is taken into account when choosing the timestamp for the next block.

## Changes

- Use `parent.timestamp_millis() + 1` when local time is behind the parent (`crates/commonware-node/src/consensus/application/actor.rs`)
- Reduce allowed future block time from 3s to 500ms (`crates/consensus/src/lib.rs`)
- Switch timestamp validation to millisecond precision
- Add `parent_ahead_of_local_time` counter metric

## Testing

Cherry-picked from [tempo-private-fork#26](https://github.com/tempoxyz/tempo-private-fork/pull/26) — 5 commits by @klkvr.

Prompted by: klkvr

### DIFF
--- a/crates/commonware-node/src/consensus/application/actor.rs
+++ b/crates/commonware-node/src/consensus/application/actor.rs
@@ -543,6 +543,9 @@ impl Inner<Init> {
         };
 
         // Use current timestamp but make sure that if parent's timestamp is in the future, we account for that.
+        //
+        // We don't expect this being hit in practice because we validate the
+        // timestamp is not in the future during EL validation.
         let current_timestamp = context.current().epoch_millis();
         let timestamp_millis = if current_timestamp <= parent.timestamp_millis() {
             self.metrics.parent_ahead_of_local_time.inc();
--- a/crates/consensus/src/lib.rs
+++ b/crates/consensus/src/lib.rs
@@ -23,6 +23,11 @@ use tempo_primitives::{
 };
 
 /// How far in the future the block timestamp can be.
+///
+/// We are setting this to 0 to not allow any drift of the block time in the future.
+/// We are considering this safe because with the way CL works currently block time would
+/// be consistent and thus an honest proposer should never produce a block that appears
+/// to be in the future even assuming 50-100ms clock drift.
 pub const ALLOWED_FUTURE_BLOCK_TIME_MILLIS: u64 = 0;
 
 /// Divisor for calculating shared gas limit (payment lane capacity).


crates/consensus/src/lib.rs
```diff
@@ -23,7 +23,7 @@ use tempo_primitives::{
 };
 
 /// How far in the future the block timestamp can be.
-pub const ALLOWED_FUTURE_BLOCK_TIME_SECONDS: u64 = 3;
+pub const ALLOWED_FUTURE_BLOCK_TIME_MILLIS: u64 = 500;
```
SuperFluffy: I feel like a skew of 500ms is overly generous. From what I can find a range from low millis to maybe 50ms is expected in distributed systems.

Taking this into account and noting that a proposer sets the timestamp *before* proposing a block, we have the following timestamps and latencies:

1. `T_0`: enter view and set timestamp for `PayloadBuilderAttributes`; assuming `time.enter_view ~= time.start_build` here
2. `dT1`: latency to build payload
3. `dT2`: latency to broadcast to other nodes

So some skew `S` needs to account for `dT1 + dT2` so that we don't accidentally reject a valid view just because of skew.

Now our current implementation already has a wait time before a block is sent out to smooth block times when we have low activity: https://github.com/tempoxyz/tempo/blob/9f771897e3ed43298b869134bfaffa7a054b51fe/crates/commonware-node/src/consensus/application/actor.rs#L596


```suggested
pub const ALLOWED_CLOCK_SKEW_MILLIS: u64 = 500;
```

crates/consensus/src/lib.rs
```diff
@@ -47,21 +47,19 @@ impl TempoConsensus {
...
+    ) -> Result<(), ConsensusError> {
         self.inner.validate_header(header)?;
 
-        let present_timestamp = std::time::SystemTime::now()
-            .duration_since(std::time::SystemTime::UNIX_EPOCH)
-            .expect("system time should never be before UNIX EPOCH")
-            .as_secs();
-
-        if header.timestamp() > present_timestamp + ALLOWED_FUTURE_BLOCK_TIME_SECONDS {
+        if header.timestamp_millis() > present_timestamp_millis + ALLOWED_FUTURE_BLOCK_TIME_MILLIS {
```
SuperFluffy, line 71: Note my comment on the constant - I actually think that together with the wait-time we introduced, we can require that `header.timestamp_millis() < present_timestamp_millis` always

crates/consensus/src/lib.rs
```diff
@@ -47,21 +47,26 @@ impl TempoConsensus {
...
-            .as_secs();
+        // Validate the timestamp milliseconds part
+        if header.timestamp_millis_part >= 1000 {
+            return Err(ConsensusError::Other(
+                "Timestamp milliseconds part must be less than 1000".to_string(),
+            ));
+        }
 
-        if header.timestamp() > present_timestamp + ALLOWED_FUTURE_BLOCK_TIME_SECONDS {
+        if header.timestamp_millis() > present_timestamp_millis + ALLOWED_FUTURE_BLOCK_TIME_MILLIS {
```
SuperFluffy, line 71: Bit confusing since the value is 0 - why not just remove the flag entirely?

crates/consensus/src/lib.rs
```diff
@@ -23,7 +23,7 @@ use tempo_primitives::{
 };
 
 /// How far in the future the block timestamp can be.
-pub const ALLOWED_FUTURE_BLOCK_TIME_SECONDS: u64 = 3;
+pub const ALLOWED_FUTURE_BLOCK_TIME_MILLIS: u64 = 0;
```
SuperFluffy, line 31: Let's explain this value (with an eye toward how CL is driving this right now).

What I think would be better is to make this configurable at the CL level and explain it there.

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -523,10 +542,19 @@ impl Inner<Init> {
             }
         };
 
+        // Use current timestamp but make sure that if parent's timestamp is in the future, we account for that.
```
SuperFluffy, line 545: Explain that this shouldn't be hit in practice and is a precaution?

PR #3146, opened 2026-03-16
refactor(cli): use hashed state tables in init-from-binary-dump

decofe:
Switches `init-from-binary-dump` from writing to `PlainAccountState`/`PlainStorageState` to writing directly to `HashedAccounts`/`HashedStorages`, matching the reth storage v2 canonical state representation.

This eliminates the two-pass approach (write plain state → copy to hashed state via `HashingWriter`) and instead writes hashed state in a single pass using `keccak256` on addresses and storage keys.

Co-Authored-By: Dan Cline <6798349+Rjected@users.noreply.github.com>

Prompted by: Dan

### DIFF
--- a/bin/tempo/src/init_state.rs
+++ b/bin/tempo/src/init_state.rs
@@ -4,13 +4,12 @@
 //! to the genesis state. The binary format is produced by `tempo-xtask generate-state-bloat`.
 
 use std::{
-    collections::HashMap,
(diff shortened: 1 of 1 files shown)


klkvr (commented):
should we make it conditional on storage version? for now most of the tempo nodes are on storage v1 which is still the default

PR #3147, opened 2026-03-16
fix(consensus): snapshot V1 validator count during migration

howydev:
Snapshots V1 validator count on the first `migrateValidator` call to prevent V1 mutations from breaking migration index ordering. Adds `EmptyV1ValidatorSet` error to reject migration when V1 has no validators.

Updates TIP-1017 spec, Solidity ref-impl, and Rust precompile.


crates/precompiles/src/validator_config_v2/mod.rs
```diff
@@ -37,7 +37,8 @@ struct Config {
     init_at_height: u64,
     /// Number of V1 validators skipped during migration (bad pubkey, duplicate, etc).
     /// Packed alongside `is_init` and `init_at_height` since all are migration lifecycle state.
-    migration_skipped_count: u16,
+    migration_skipped_count: u8,
+    v1_validator_count: u8,
```
0xrusowsky, line 51: nit: needs docs

PR #3150, opened 2026-03-16
fix(evm): charge ecrecover gas for TIP-20 permit

howydev:
Adds a 3,000 gas charge for ecrecover signature verification in the TIP-20 permit precompile, which was previously unmetered.

### DIFF
--- a/crates/precompiles/src/tip20/mod.rs
+++ b/crates/precompiles/src/tip20/mod.rs
@@ -11,7 +11,7 @@ pub use tempo_contracts::precompiles::{
 pub use slots as tip20_slots;
 
 use crate::{
-    PATH_USD_ADDRESS, TIP_FEE_MANAGER_ADDRESS,
+    ECRECOVER_GAS, PATH_USD_ADDRESS, TIP_FEE_MANAGER_ADDRESS,
     account_keychain::AccountKeychain,
     error::{Result, TempoPrecompileError},
     storage::{Handler, Mapping},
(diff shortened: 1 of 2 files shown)


crates/precompiles/src/tip20/mod.rs
0xrusowsky, line 580: personally i'd implement `fn recover_signer()` at the storage provider level, so that we don't need to deal with gas directly in the precompile impl.

if we keep it inlined though, i'd move the `ECRECOVER_GAS` deduction prior to reading nonces and computing hashes, to avoid useless work in case of OOG

would like to know what @rakita thinks

crates/precompiles/src/tip20/mod.rs
```diff
@@ -566,7 +566,10 @@ impl TIP20Token {
             .concat(),
         );
 
-        // 4. Validate ECDSA signature
+        // 4. Charge gas for ecrecover
+        self.storage.deduct_gas(ECRECOVER_GAS)?;
```
rakita, line 570: I would add additional gas for all the keccaks that we are doing.

Quick math is giving me 150gas:
First keccak: 32 (b256) + 20 (address) + 20 (address) + 8 (value) + 4 (nonce) + 4 (time) == 88 bytes == 57gas
Second keccak: 2 (prefix) + 32 (domain_separator) + 32 (hash) == 64 bytes == 48gas
And there is abi encoding that is hard to estimate as it is rust, so putting gas to 150 looks 
reasonable.

crates/precompiles/src/tip20/mod.rs
rakita: > if we keep it inlined though, i'd move the ECRECOVER_GAS deduction prior to reading nonces and computing hashes, to avoid useless work in case of OOG

I like this, I would put in between the first and second ( Or it could even go in first step, if there is no special handling for `TIP20Error::permit_expired()`)

howydev:
closing in favor of https://github.com/tempoxyz/tempo/pull/3155

PR #3151, opened 2026-03-16
docs(tip-1034): specify single-slot TIP-20 escrow channels

legion2002:
Specifies TIP-1034 as a descriptor-derived channel escrow precompile where only one packed mutable storage slot is kept on-chain per channel. Immutable channel identity (`payer`, `payee`, `token`, `authorizedSigner`, `salt`) moves to calldata and `ChannelOpened` logs, while `settled`, `deposit`, `expiresAt`, and `closeData` make up the single stored channel state word.

The reference interface, reference implementation, and tests now follow the same design, including `closeData` sentinels (`0` = active, `1` = finalized, `>=2` = close requested at timestamp) so terminal channels keep a tombstone and cannot be reopened with the same descriptor.

### DIFF
--- a/tips/tip-1034.md
+++ b/tips/tip-1034.md
@@ -2,7 +2,7 @@
 id: TIP-1034
 title: TIP-20 Channel Escrow Precompile
 description: Enshrines TIP-20 channel escrow as a Tempo precompile with payment-lane admission and native escrow transfer semantics.
-authors: Tanishk Goyal
+authors: Tanishk Goyal, Brendan Ryan
 status: Draft
 related: TIP-20, TIP-1000, TIP-1020, Tempo Session, Tempo Charge
 protocolVersion: TBD
--- a/tips/tip-1034.md
+++ b/tips/tip-1034.md
@@ -42,10 +42,10 @@ address constant TIP20_CHANNEL_ESCROW = 0x4D505000000000000000000000000000000000
 
 ## Implementation Details
 
-This TIP is normative. The current reference implementations are informative and live at:
+This TIP is normative. The current reference Solidity artifacts are informative and live at:
 
-1. [`tips/ref-impls/src/interfaces/ITIP20ChannelEscrow.sol`](ref-impls/src/interfaces/ITIP20ChannelEscrow.sol)
-2. [`tips/ref-impls/src/TIP20ChannelEscrow.sol`](ref-impls/src/TIP20ChannelEscrow.sol)
+1. [`tips/verify/src/interfaces/ITIP20ChannelEscrow.sol`](verify/src/interfaces/ITIP20ChannelEscrow.sol)
+2. [`tips/verify/src/TIP20ChannelEscrow.sol`](verify/src/TIP20ChannelEscrow.sol)
 
 Implementations SHOULD keep those reference artifacts aligned with the normative interface and execution rules defined below.
 
@@ -124,7 +124,7 @@ channelId = keccak256(
 
 ### Interface
 
-The canonical interface for this TIP is [`tips/ref-impls/src/interfaces/ITIP20ChannelEscrow.sol`](ref-impls/src/interfaces/ITIP20ChannelEscrow.sol).
+The canonical interface for this TIP is [`tips/verify/src/interfaces/ITIP20ChannelEscrow.sol`](verify/src/interfaces/ITIP20ChannelEscrow.sol).
 
 Implementations MUST expose an external interface that is semantically identical to that file,
 including the `ChannelDescriptor`, `ChannelState`, and `Channel` structs, the descriptor-based
--- a/tips/tip-1034.md
+++ b/tips/tip-1034.md
@@ -51,23 +51,53 @@ Implementations SHOULD keep those reference artifacts aligned with the normative
 
 Channels are unidirectional (`payer -> payee`) and token-specific.
 
-### Channel State Layout
+### Channel Identity And Packed State
 
 ```solidity
-struct Channel {
-    bool finalized;
-    uint64 closeRequestedAt;
+struct ChannelDescriptor {
     address payer;
     address payee;
-    uint64 expiresAt;
     address token;
+    bytes32 salt;
     address authorizedSigner;
-    uint128 deposit;
-    uint128 settled;
+}
+
+struct ChannelState {
+    bool finalized;
+    uint32 closeRequestedAt;
+    uint32 expiresAt;
+    uint96 deposit;
+    uint96 settled;
+}
+
+struct Channel {
+    ChannelDescriptor descriptor;
+    ChannelState state;
 }
 ```
 
-This layout uses 5 storage slots total.
+Implementations MUST store only `ChannelState` on-chain, keyed by `channelId` and packed into a
+single 32-byte storage slot. `ChannelDescriptor` is immutable channel identity and MUST NOT be
+stored on-chain; it MUST instead be supplied in calldata for post-open operations and emitted in
+`ChannelOpened` so indexers and counterparties can recover it.
+
+The packed slot layout is:
+
+```text
+bits   0..95   settled        uint96
+bits  96..191  deposit        uint96
+bits 192..223  expiresAt      uint32
+bits 224..255  closeData      uint32
+```
+
+`closeData` MUST be encoded as:
+
+1. `0` for an active channel with no close request.
+2. `1` for a finalized channel tombstone.
+3. `closeRequestedAt + 2` for an active channel with a close request.
+
+Implementations MUST decode `closeRequestedAt` as `0` when `closeData < 2`, otherwise as
+`closeData - 2`.
 
 `channelId` MUST use the following deterministic domain-separated construction:
 
@@ -89,14 +119,16 @@ channelId = keccak256(
 
 The canonical interface for this TIP is [`tips/ref-impls/src/interfaces/ITIP20ChannelEscrow.sol`](ref-impls/src/interfaces/ITIP20ChannelEscrow.sol).
 
-Implementations MUST expose an external interface that is semantically identical to that file, including the `Channel` state layout, events, errors, and function signatures.
+Implementations MUST expose an external interface that is semantically identical to that file,
+including the `ChannelDescriptor`, `ChannelState`, and `Channel` view structs, the descriptor-based
+post-open methods, the events, the errors, and the function signatures.
 
 ### Execution Semantics
 
 Voucher signatures use the following EIP-712 type:
 
 ```solidity
-Voucher(bytes32 channelId,uint128 cumulativeAmount)
+Voucher(bytes32 channelId,uint96 cumulativeAmount)
 ```
 
 Voucher signatures MUST be verified via the TIP-1020 Signature Verification Precompile at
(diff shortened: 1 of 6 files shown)


tips/tip-1034.md
```diff
@@ -0,0 +1,191 @@
+---
+id: TIP-1034
+title: Enshrined MPP Channels Precompile
```
jxom: I could see this TIP being non-exclusive to MPP? Like "Channeled TIP20 Escrow"?

Idea is that we could repurpose this TIP to be used for payment + capture use cases (where we can either do full capture with a max voucher; or partial/incremental capture with multiple vouchers).

brendanjryan: +1 - Maybe even "voucher TIP20 Escrow"?

legion2002: The TIP proposes exemptions, which allows the calls of this precompile to be included in the privileged payment lane.
repurposing it to a general escrow, might spam the blockspace.

Although atm we don't have any onchain protections for people to not use this wihout MPP. So it's more of a branding thing.

tips/tip-1034.md
```diff
@@ -0,0 +1,196 @@
...
+---
+
+# Specification
+
+## Precompile Address
+
+This TIP introduces a new system precompile at:
+
+```solidity
+address constant MPP_CHANNEL_PRECOMPILE = 0x4D50500000000000000000000000000000000000;
```
brendanjryan: why this address?

legion2002: encodes MPP (4D 50 50)

decofe (commented):
:eye: **Cyclops Review**

TIP-1034 is well-structured but contains **6 specification-level security flaws** that would produce exploitable vulnerabilities when implemented. All findings are against the normative TIP text (no executable code is changed in this PR).

**Critical (2):** Payer can rug channels by shrinking expiry via `topUp`; `systemTransferFrom` bypasses token approval enabling theft from contract wallets.
**High (2):** Unbounded `signature` field enables payment lane abuse; `close` invariant locks payee funds when voucher exceeds deposit.
**Medium (2):** Ambiguous `expiresAt` boundary creates consensus split risk; auth side effects bypass `general_gas_limit` via payment lane.

<details>

- ⚡ **Monotonic Expiry**: The word "extend" in semantic #2 implies monotonicity, but the MUST clause doesn't enforce it. This is the highest-priority fix.
- ⚡ **Precompile Composability**: `systemTransferFrom` fundamentally breaks EVM composability assumptions. Any future contract that interacts with arbitrary addresses becomes vulnerable.
- ⚡ **Classification Strictness**: Payment-lane classifiers must be extremely strict about dynamic types and side-effectful metadata. Both the `bytes signature` length and auth-list presence need explicit bounds.
- ⚡ **Partial Capture Consistency**: The invariant chain `captureAmount <= cumulativeAmount <= deposit` was inherited from the reference model where `captureAmount` didn't exist. It needs restructuring for the new partial-capture semantics.
- ⚡ **Exact Boundary Operators**: Every `expiresAt` comparison must use explicit `<`/`<=`/`>`/`>=` operators, not prose, since this is consensus-critical.
</details>

tips/tip-1034.md
brendanjryan: ahh nice

tips/tip-1034.md
```diff
@@ -0,0 +1,196 @@
+---
+id: TIP-1034
+title: Enshrined MPP Channels Precompile
+description: Enshrines MPP payment channels as a Tempo precompile with payment-lane admission and native escrow transfer semantics.
+authors: Tanishk Goyal
+status: Draft
+related: TIP-20, TIP-1000, mpp-specs (tempo session/auth-capture)
```
brendanjryan: can we link to these on paymentauth.org?

tips/tip-1034.md
```diff
@@ -0,0 +1,196 @@
...
+```
+
+`MPP_CHANNEL_PRECOMPILE` MUST refer to this address throughout this specification.
+
+## Implementation Details
+
+This precompile MUST follow the current reference implementation in:
+
+1. `tips/ref-impls/src/interfaces/ITempoStreamChannel.sol`
+2. `tips/ref-impls/src/TempoStreamChannel.sol`
```
brendanjryan: can we link to these?

tips/tip-1034.md
```diff
@@ -0,0 +1,196 @@
...
+```solidity
+function close(
+    bytes32 channelId,
+    uint128 cumulativeAmount,
+    uint128 captureAmount,
+    bytes calldata signature
+) external;
+```
+
+Compared to the reference interface, this adds `expiresAt` to channel lifecycle (`open`, `topUp`) and adds explicit final `captureAmount` to `close`.
```
brendanjryan: instead of comparing to the reference interface, since this is a formal TIP -- I think it may make sense to include all context and define interfaces/flows from first principles here

tips/tip-1034.md
```diff
@@ -0,0 +1,196 @@
...
+Execution semantics are the same as the reference implementation except:
+
+1. `open` MUST validate `expiresAt > block.timestamp` and persist `expiresAt`.
+2. `topUp` MAY extend expiry using `newExpiresAt`; when non-zero, it MUST be strictly in the future.
+3. `settle` and `close` MUST reject any capture-increasing action after `expiresAt`.
+4. `close` MUST enforce `previousSettled <= captureAmount <= cumulativeAmount <= deposit`.
+5. `close` MUST settle `captureAmount - previousSettled` to payee and refund `deposit - captureAmount` to payer.
+6. `withdraw` MUST be allowed after close grace OR after expiry.
+7. Signer MUST be `authorizedSigner` when set, otherwise `payer`.
+
```
legion2002, line 224: @brendanjryan was this behaviour allowing shrinking of expiry intentional? I think this concern looks pretty valid

tips/tip-1034.md
```diff
@@ -0,0 +1,196 @@
...
+## Native Escrow Movement
+
+In this precompile, escrow transfers MUST use system TIP-20 movement semantics equivalent to `systemTransferFrom`.
```
legion2002, line 233: @brendanjryan what's the expected behaviour here? 
my understanding is that servers should be checking this offchain , that the voucher is collateralized onchain.

But in the edge case, that such a voucher gets through. Should we block it completely, or allow partial capture? (my opinion is that the latter makes more sense)

tips/tip-1034.md
brendanjryan: I think this is a safer invariant -- much like the vouchers this should only be monotonically increasing

tips/tip-1034.md
```diff
@@ -0,0 +1,196 @@
...
+
+Required behavior:
+
+1. `open` escrows `deposit` from `payer` to channel escrow state without requiring a prior user `approve` transaction.
+2. `topUp` escrows `additionalDeposit` the same way.
+3. `settle`, `close`, and `withdraw` payout paths continue to transfer TIP-20 value using protocol-native token movement.
```
brendanjryan, line 239: cc @legion2002 for your thoughts here -- I think this may be safer now that we have enshrined permit as well?

tips/tip-1034.md
brendanjryan: this should be a server side concern, but the contract should not lock users funds out. I'm +1 to partial capture

tips/tip-1034.md
legion2002: we do the same for stablecoin dex as well. So it is the same security policy. MPP does feel more exposed, as it is a more general purpose escrow contract.
On the other hand, contracts should not be calling random functions on random addresses, that's a big security risk in itself.

So I think it is fine to leave it, as it does meaningfully improve UX. and enshrined permit does not work for passkey accounts.

tips/tip-1034.md
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/cc014e7e16bc985582f00d24b2adfa82c4c3cf99

tips/tip-1034.md
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/cc014e7e16bc985582f00d24b2adfa82c4c3cf99

tips/tip-1034.md
```diff
@@ -0,0 +1,196 @@
...
+
+1. `open` MUST validate `expiresAt > block.timestamp` and persist `expiresAt`.
+2. `topUp` MAY extend expiry using `newExpiresAt`; when non-zero, it MUST be strictly in the future.
+3. `settle` and `close` MUST reject any capture-increasing action after `expiresAt`.
+4. `close` MUST enforce `previousSettled <= captureAmount <= cumulativeAmount <= deposit`.
+5. `close` MUST settle `captureAmount - previousSettled` to payee and refund `deposit - captureAmount` to payer.
+6. `withdraw` MUST be allowed after close grace OR after expiry.
+7. Signer MUST be `authorizedSigner` when set, otherwise `payer`.
+
+## Native Escrow Movement
```
legion2002, line 231: Addressed in https://github.com/tempoxyz/tempo/commit/29cbe48fd152bc53db1162d16379fc8d3f9afa12

tips/tip-1034.md
```diff
@@ -0,0 +1,196 @@
...
+
+1. `to == MPP_CHANNEL_PRECOMPILE`.
+2. `input` selector is one of `{open, settle, topUp, close, requestClose, withdraw}`.
+3. Calldata length/encoding is valid for that selector.
+
+For AA transactions, payment classification MUST satisfy both:
+
+1. `calls.length > 0`.
+2. Every call satisfies either TIP-20 strict payment classification or `is_mpp_channel_payment`.
+
```
legion2002, line 287: Addressed in https://github.com/tempoxyz/tempo/commit/29cbe48fd152bc53db1162d16379fc8d3f9afa12

tips/tip-1034.md
```diff
@@ -0,0 +1,196 @@
...
+An AA transaction with an empty `calls` array MUST be classified as non-payment.
+
```
legion2002, line 289: Addressed in https://github.com/tempoxyz/tempo/commit/29cbe48fd152bc53db1162d16379fc8d3f9afa12

tips/tip-1034.md
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/55b74e1b823a46c9bc10a173bb5d2233ffef7390

tips/tip-1034.md
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/55b74e1b823a46c9bc10a173bb5d2233ffef7390

tips/tip-1034.md
legion2002: Has been resolved and changed to `TIP20ChannelEscrow`

tips/tip-1034.md
```diff
@@ -0,0 +1,204 @@
...
+The precompile is introduced to reduce execution overhead, remove the separate `approve` UX via native escrow movement, and make channel operations first-class payment-lane traffic under congestion.
+
+## Motivation
+
+TIP-20 channel escrow is currently specified as a Solidity contract reference implementation. Enshrining that behavior as a precompile is motivated by three protocol-level goals:
+
+1. **Efficiency**: Channel operations become native precompile execution instead of generic contract execution, reducing overhead and making gas behavior more predictable for high-frequency payment flows.
+2. **Payment-Lane Access**: Channel operations become payment-lane transactions, so they are not throttled by the non-payment `general_gas_limit` path during block contention.
+3. **Approve-less Escrow UX**: `open` and `topUp` can escrow TIP-20 funds through native system transfer (`systemTransferFrom` semantics), removing the extra `approve` transaction from the user flow.
```
brendanjryan: 👍

tips/tip-1034.md
```diff
@@ -0,0 +1,204 @@
...
+`TIP20_CHANNEL_ESCROW` MUST refer to this address throughout this specification.
+
+## Implementation Details
+
+This TIP is normative. The current reference implementations are informative and live at:
+
+1. [`tips/ref-impls/src/interfaces/ITIP20ChannelEscrow.sol`](ref-impls/src/interfaces/ITIP20ChannelEscrow.sol)
+2. [`tips/ref-impls/src/TIP20ChannelEscrow.sol`](ref-impls/src/TIP20ChannelEscrow.sol)
+
+Implementations SHOULD keep those reference artifacts aligned with the normative interface and execution rules defined below.
```
brendanjryan, line 50: 👍

tips/tip-1034.md
```diff
@@ -0,0 +1,250 @@
...
+### What This Means Operationally
+
+With this integration, channel lifecycle calls consume payment-lane capacity rather than non-payment capacity. Under high congestion, these transactions continue to compete in the same lane as other payment traffic instead of being excluded by non-payment limits.
+
+---
+
+# Invariants
+
+1. `settled <= deposit` MUST hold in all reachable states.
+2. `settled` is monotonic and can never decrease.
```
brendanjryan: can we just say `settled` is monotonically increasing here? vs. the negative?

tips/tip-1034.md
```diff
@@ -0,0 +1,268 @@
...
+
+The precompile is introduced to reduce execution overhead, remove the separate `approve` UX via native escrow movement, and make channel operations first-class payment-lane traffic under congestion.
+
+## Motivation
+
+TIP-20 channel escrow is currently specified as a Solidity contract reference implementation. Enshrining that behavior as a precompile is motivated by three protocol-level goals:
+
+1. **Efficiency**: Channel operations become native precompile execution instead of generic contract execution, reducing overhead and making gas behavior more predictable for high-frequency payment flows.
+2. **Payment-Lane Access**: Channel operations become payment-lane transactions, so they are not throttled by the non-payment `general_gas_limit` path during block contention.
+3. **Approve-less Escrow UX**: `open` and `topUp` can escrow TIP-20 funds through native system transfer (`systemTransferFrom` semantics), removing the extra `approve` transaction from the user flow.
```
klkvr: i believe for (2) and (3) we don't need enshrining? given implicit approval lists

tips/tip-1034.md
```diff
@@ -0,0 +1,268 @@
...
+2. `open` MUST persist only the packed `ChannelState` slot and MUST emit the full immutable descriptor in `ChannelOpened`.
+3. Post-open methods (`settle`, `topUp`, `close`, `requestClose`, `withdraw`, and descriptor-based views) MUST recompute `channelId` from the supplied descriptor and use that derived id for storage lookup.
+4. If `closeData >= 2`, a successful `topUp` MUST clear it back to `0` and emit `CloseRequestCancelled`.
+5. `requestClose` MUST set `closeData = uint32(block.timestamp)` on the first successful call and leave it unchanged on later successful calls.
+6. `close` MUST validate the voucher signature via TIP-1020 for any capture-increasing close.
+7. Signer MUST be `authorizedSigner` from the supplied descriptor when set, otherwise `payer` from the supplied descriptor.
+8. `close` MUST enforce `previousSettled <= captureAmount <= cumulativeAmount`.
+9. `close` MUST reject when `captureAmount > deposit`, even if `cumulativeAmount > deposit`.
+10. A `close` voucher with `cumulativeAmount > deposit` remains valid for signature verification; `captureAmount` is the escrow-bounded amount that may actually be paid out.
```
klkvr: why do we allow such vouchers? afaict `settle` rejects them

tips/tip-1034.md
legion2002: we can do 2 without enshrining. We can't do 3 in the way that [TIP 1035](https://tips.sh/1035) is written right now.

tips/tip-1034.md
legion2002: addressed in - https://github.com/tempoxyz/tempo/pull/3151/commits/0cc095847af8384a338ae042ba64697de1e79f45

tips/tip-1034.md
legion2002: settle pays out exactly what the cumulativeAmount says instantly.
Whereas close is supposed to facilitate the auth+capture usecase, where you initially auth a high amount (this is the cumulative amount), but you might capture an amount lower than this.

I think this is a common credit card flow, and we must be making sure on the server side that the capture amount never exceeds deposit.

Also even in the worst case when there is some error in the server, and server expects 100 USD, but deposit is only 60 USD. We want the server to be able to claim atleast the 60 USD. There is no point in reverting here.

But @brendanjryan can say more about this usecase.

tips/tip-1034.md
```diff
@@ -0,0 +1,325 @@
...
+This TIP enshrines TIP-20 channel escrow in Tempo as a native precompile. The implementation follows the existing reference channel model (`open`, `settle`, `topUp`, `requestClose`, `close`, `withdraw`) and keeps the same EIP-712 voucher flow, with explicit partial-capture updates defined in this spec.
+
+The precompile is introduced to reduce execution overhead, remove the separate `approve` UX via native escrow movement, and make channel operations first-class payment-lane traffic under congestion.
+
+## Motivation
+
+TIP-20 channel escrow is currently specified as a Solidity contract reference implementation. Enshrining that behavior as a precompile is motivated by three protocol-level goals:
+
+1. **Efficiency**: Channel operations become native precompile execution instead of generic contract execution, reducing overhead and making gas behavior more predictable for high-frequency payment flows.
+2. **Payment-Lane Access**: Channel operations become payment-lane transactions, so they are not throttled by the non-payment `general_gas_limit` path during block contention.
```
jenpaff:
```suggestion
2. **Canonical implementation, evolved at network-upgrade cadence**: The escrow lives at a single canonical address on every Tempo node (`0x4D50500…`, ASCII "MPP"), with no admin or proxy admin, and ships behavior via network upgrades alongside the rest of the protocol. Wallets, indexers, and partner integrations bind to one stable surface rather than to a moving set of redeployments.
```

tips/tip-1034.md
```diff
@@ -0,0 +1,325 @@
...
+3. **Approve-less Escrow UX**: `open` and `topUp` can escrow TIP-20 funds through native system transfer (`systemTransferFrom` semantics), removing the extra `approve` transaction from the user flow.
```
jenpaff:
```suggestion
3. **Approve-less Escrow UX**: `open` and `topUp` can escrow TIP-20 funds through native system transfer (`systemTransferFrom` semantics), removing the extra `approve` transaction from the user flow and works uniformly for passkey accounts.
```

brendanjryan (approved):

samczsun (approved):

PR #3155, opened 2026-03-17
fix(precompiles): built-in support for ecrecover + keccak

0xrusowsky:
this PR adds built-in support to the `trait PrecompileStorageProvider` for `fn keccak256()` and `fn ecrecover()` so that consumers don't need to worry about gas metering.

supersedes https://github.com/tempoxyz/tempo/pull/3150

### DIFF
--- a/crates/precompiles/src/storage/mod.rs
+++ b/crates/precompiles/src/storage/mod.rs
@@ -116,8 +116,8 @@ pub trait PrecompileStorageProvider {
     ///
     /// Implementations should use this over naked `keccak256` call to ensure gas is accounted for.
     fn keccak256(&mut self, data: &[u8]) -> Result<B256> {
-        let num_words = u64::try_from(data.len().div_ceil(32))
-            .map_err(|_| TempoPrecompileError::OutOfGas)?;
+        let num_words =
+            u64::try_from(data.len().div_ceil(32)).map_err(|_| TempoPrecompileError::OutOfGas)?;
         let price = KECCAK256WORD
             .checked_mul(num_words)
             .and_then(|w| w.checked_add(KECCAK256))


crates/precompiles/src/storage/mod.rs
```diff
@@ -104,6 +106,35 @@ pub trait PrecompileStorageProvider {
     ///
     /// Prefer [`CheckpointGuard`] (auto-reverts on drop).
     fn checkpoint_revert(&mut self, checkpoint: JournalCheckpoint);
+
+    /// Computes keccak256 and charges the appropriate gas.
+    ///
+    /// Implementations should use this over naked `keccak256` call to ensure gas is accounted for.
+    fn keccak256(&mut self, data: &[u8]) -> Result<B256> {
+        let price = KECCAK256 + KECCAK256WORD * data.len().div_ceil(32) as u64;
```
howydev: what happens if this overflows?

is it even a risk? attacker needs to figure out how to generate a large data size here without paying for calldata?

howydev: in val config v2 we might be hashing arbitrary string lengths (IP) but attacker has to pay for calldata there

crates/revm/src/handler.rs
```diff
@@ -36,6 +36,7 @@ use tempo_contracts::precompiles::{
     IAccountKeychain::SignatureType as PrecompileSignatureType, TIPFeeAMMError,
 };
 use tempo_precompiles::{
+    ECRECOVER_GAS,
```
howydev, line 39: huh, amp suggested that i do this and i assumed it was a dumb suggestion lol

crates/precompiles/src/storage/mod.rs
0xrusowsky: good call. even if users always pay for calldata it is better to make it fallible

crates/revm/src/handler.rs
0xrusowsky: haha no, cause then we have a single source of truth

crates/precompiles/src/storage/mod.rs
0xrusowsky: addressed in [`836efa8` (this PR)](https://github.com/tempoxyz/tempo/pull/3155/commits/836efa8cf3fac2350ce17de69af0707cc1ec8342)

klkvr (approved):

PR #3158, opened 2026-03-17
fix(ext): do not call `tempoup` second time

shekhirin:
Fixes failing tests https://github.com/tempoxyz/tempo/actions/runs/23193066677/job/67394024072?pr=3156

There's no need to call `tempoup` second time, because installing `tempoup` with curl already calls it https://github.com/tempoxyz/tempo/blob/2497751c8b8036fc39f5f079276e931b5df75b75/tempoup/install#L85-L89

Also includes the dependency bump from https://github.com/tempoxyz/tempo/pull/3156

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5762,9 +5762,9 @@ dependencies = [
 
 [[package]]
 name = "lz4_flex"
-version = "0.12.0"
+version = "0.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ab6473172471198271ff72e9379150e9dfd70d8e533e0752a27e515b48dd375e"
+checksum = "98c23545df7ecf1b16c303910a69b079e8e251d60f7dd2cc9b4177f2afaf1746"
 
 [[package]]
 name = "mach2"


crates/ext/src/launcher.rs
```diff
@@ -174,14 +174,14 @@ fn run_tempoup(bin_dir: &Path) -> Result<bool, LauncherError> {
             let install_status = Command::new("sh")
                 .arg("-c")
                 .arg("curl -fsSL https://tempo.xyz/install | bash")
```
SuperFluffy, line 176: Can't we just download it ourselves? Calling `curl` from a rust binary, to then launch a subshell seems ... too much.

crates/ext/src/launcher.rs
```diff
@@ -174,14 +174,14 @@ fn run_tempoup(bin_dir: &Path) -> Result<bool, LauncherError> {
...
+                .env("TEMPO_BIN_DIR", bin_dir)
                 .status()?;
             if !install_status.success() {
                 tracing::error!("failed to install tempoup");
```
SuperFluffy, line 180: Should probably return an error....

PR #3170, opened 2026-03-17
feat(bench): resolve mpp-contract-address from chain ID

decofe:
Resolves `--mpp-contract-address` automatically based on chain ID fetched from the first RPC URL when the flag is not explicitly provided. Mainnet (4217) uses `0x33b901018174ddabe4841042ab76ba85d4e24f25`, testnet (42431) uses `0xe1c4d3dce17bc111181ddf716f75bae49e61a336`. Unknown chain IDs error with a message to pass the flag explicitly.

Co-Authored-By: Alexey Shekhirin <5773434+shekhirin@users.noreply.github.com>

Prompted by: alexey

### DIFF
--- a/bin/tempo-bench/src/cmd/max_tps.rs
+++ b/bin/tempo-bench/src/cmd/max_tps.rs
@@ -16,7 +16,7 @@ use alloy::{
     consensus::BlockHeader,
     eips::Encodable2718,
     network::{EthereumWallet, ReceiptResponse, TransactionBuilder, TxSignerSync},
-    primitives::{Address, B256, BlockNumber, U256},
+    primitives::{address, Address, B256, BlockNumber, U256},
     providers::{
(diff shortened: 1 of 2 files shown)


bin/tempo-bench/src/cmd/max_tps.rs
```diff
@@ -151,8 +151,9 @@ pub struct MaxTpsArgs {
     #[arg(long, default_value_t = 0.0)]
     mpp_weight: f64,
 
-    /// Address of a deployed TempoStreamChannel contract. Required when `--mpp-weight` > 0.
-    #[arg(long, default_value = "0x33b901018174ddabe4841042ab76ba85d4e24f25")]
+    /// Address of a deployed TempoStreamChannel contract. Resolved automatically from chain ID
+    /// if not provided (mainnet: 0x33b9...4f25, testnet: 0xe1c4...a336).
```
shekhirin:
```suggestion
    /// if not provided.
```

PR #3171, opened 2026-03-17
fix(pool): run invalidation, pause events, and expiring nonce cleanup on reorg segment

fgimenez:
Closes SIGP-182

On reorg, the maintenance handler `continue`d before reaching `TempoPoolUpdates::from_chain()`, so invalidation events (key revocations, spending limit changes, validator/user token changes, blacklist additions), pause/unpause events, and expiring nonce cleanup in the new canonical segment were silently ignored.

Extracts updates from the reorg segment and runs eviction, pause event handling, and expiring nonce cleanup before the `continue`. Factors the duplicated logic into shared helpers (`evict_invalidated_from_pool`, `evict_invalidated_from_paused_pool`, `handle_pause_events`) used by both the reorg and commit paths.

Also addresses three issues found during review:

- **Synthesize inverse events on reorg**: `from_chain(&new)` missed reverted "positive" state changes from the old chain (e.g., reverted unpauses, blacklist removals, whitelist additions). Added `from_reorg(old, new)` that generates synthetic inverse events for these cases.

- **Align reorg eviction ordering**: the reorg path ran main pool eviction before pause handling, so a tx hit by both a pause and a blacklist was permanently dropped instead of parked. Reordered to match the commit path: pause first, then paused pool eviction, then main pool eviction.

- **Remove `spending_limit_spends` from paused pool eviction**: the paused pool lacks state access for `exceeds_spending_limit` checks and the sponsor exemption, so it was incorrectly evicting transactions. Limits are re-checked on unpause via full validation.

### DIFF
--- a/crates/transaction-pool/src/maintain.rs
+++ b/crates/transaction-pool/src/maintain.rs
@@ -892,15 +892,10 @@ fn evict_invalidated_from_pool<Client>(
 /// Removes paused transactions whose keychain keys were revoked or whose
 /// spending limits changed.
 fn evict_invalidated_from_paused_pool(state: &mut TempoPoolState, updates: &TempoPoolUpdates) {
-    if !updates.revoked_keys.is_empty()
-        || !updates.spending_limit_changes.is_empty()
-        || !updates.spending_limit_spends.is_empty()
-    {
-        state.paused_pool.evict_invalidated(
-            &updates.revoked_keys,
-            &updates.spending_limit_changes,
-            &updates.spending_limit_spends,
-        );
+    if !updates.revoked_keys.is_empty() || !updates.spending_limit_changes.is_empty() {
+        state
+            .paused_pool
+            .evict_invalidated(&updates.revoked_keys, &updates.spending_limit_changes);
     }
 }
 
--- a/crates/transaction-pool/src/maintain.rs
+++ b/crates/transaction-pool/src/maintain.rs
@@ -635,13 +635,15 @@ where
                         // including synthetic inverse events for reverted state changes
                         // in the old segment (e.g., reverted unpauses, blacklist removals).
                         let updates = TempoPoolUpdates::from_reorg(&old, &new);
-                        evict_invalidated_from_pool(&pool, &mut state, &metrics, &updates);
-                        evict_invalidated_from_paused_pool(&mut state, &updates);
 
-                        // Handle pause/unpause events from the reorg segment.
+                        // Order matches the commit path: pause first (so affected txs are
+                        // parked rather than permanently evicted), then evict from paused
+                        // pool, then evict from main pool.
                         handle_pause_events(
                             &pool, &mut state, &metrics, &updates,
                         );
+                        evict_invalidated_from_paused_pool(&mut state, &updates);
+                        evict_invalidated_from_pool(&pool, &mut state, &metrics, &updates);
 
                         // Remove included expiring nonce txs from the new chain.
                         pool.remove_included_expiring_nonce_txs(
--- a/crates/transaction-pool/src/maintain.rs
+++ b/crates/transaction-pool/src/maintain.rs
@@ -560,6 +560,14 @@ where
                             error!(target: "txpool", ?err, "AMM liquidity cache repopulate after reorg failed");
                         }
 
+                        // Extract invalidation events from the new canonical segment
+                        // and evict affected transactions. Without this, events like key
+                        // revocations, spending limit changes, validator/user token changes,
+                        // and blacklist additions in the reorg segment would be ignored.
+                        let updates = TempoPoolUpdates::from_chain(&new);
+                        evict_invalidated_from_pool(&pool, &mut state, &metrics, &updates);
+                        evict_invalidated_from_paused_pool(&mut state, &updates);
+
                         continue;
                     }
                     CanonStateNotification::Commit { new } => new,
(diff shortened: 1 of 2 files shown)


0xKitsune (commented):
​

crates/transaction-pool/src/maintain.rs
```diff
@@ -560,6 +560,14 @@ where
                             error!(target: "txpool", ?err, "AMM liquidity cache repopulate after reorg failed");
                         }
 
+                        // Extract invalidation events from the new canonical segment
+                        // and evict affected transactions. Without this, events like key
+                        // revocations, spending limit changes, validator/user token changes,
+                        // and blacklist additions in the reorg segment would be ignored.
+                        let updates = TempoPoolUpdates::from_chain(&new);
+                        evict_invalidated_from_pool(&pool, &mut state, &metrics, &updates);
+                        evict_invalidated_from_paused_pool(&mut state, &updates);
```
0xKitsune: IIUC, this only checks for invalid transactions in the paused pool. It doesn’t handle moving transactions into or out of the paused pool when the pause status changes.

crates/transaction-pool/src/maintain.rs
```diff
@@ -890,6 +868,63 @@ where
     count
 }
 
+/// Evicts invalidated transactions from the main pool in a single scan.
+///
+/// Checks revoked keys, spending limit changes, validator token changes,
+/// blacklist additions, and whitelist removals.
+fn evict_invalidated_from_pool<Client>(
```
0xKitsune, line 854: Does this need to be a separate function from `evict_invalidated_from_paused_pool`. There seem to be many different functions to handle various evictions from the pool. This is something we can cleanup in a future PR but noting that it looks like these might be able to one fn (unless there are cases where one is called without the other).

crates/transaction-pool/src/maintain.rs
fgimenez: exactly, now the reorg path does not handle pause state management (pause/unpause events), the changes are meant to address SIGP-182, ignored invalidation events. lmk if we should extend the logic to handle pause state management too, or do it on a separate PR

crates/transaction-pool/src/maintain.rs
fgimenez: The `evict_invalidated_from_paused_pool` and `evict_invalidated_from_pool` helpers are extracted from the commit loop to avoid duplication and reuse in the reorg code.

in the original location you can see that they are called separately, `evict_invalidated_from_paused_pool` in step 5 and `evict_invalidated_from_pool` in step 9, does this make sense, should we keep these helpers?

crates/transaction-pool/src/maintain.rs
fgimenez: similar to https://github.com/tempoxyz/tempo/pull/3171#discussion_r2954736966 but including expiring noce cleanup. this is pre-existing, but given that the changes in the PR touch this code let's fix it as part of the PR

fgimenez: done here https://github.com/tempoxyz/tempo/pull/3171/commits/35d3c9248d0b0f321b8de5d4453331bd81c6bebb similar to invalidation, extracted the logic existing in the commit loop to a separate helper and call it from the commit loop and from the reorg segment, ptal

fgimenez: addressed here https://github.com/tempoxyz/tempo/pull/3171/commits/35d3c9248d0b0f321b8de5d4453331bd81c6bebb

crates/transaction-pool/src/maintain.rs
```diff
@@ -560,6 +560,26 @@ where
                             error!(target: "txpool", ?err, "AMM liquidity cache repopulate after reorg failed");
                         }
 
+                        // Extract invalidation events from the new canonical segment
+                        // and evict affected transactions. Without this, events like key
+                        // revocations, spending limit changes, validator/user token changes,
+                        // and blacklist additions in the reorg segment would be ignored.
+                        let updates = TempoPoolUpdates::from_chain(&new);
```
fgimenez: valid, addressed here https://github.com/tempoxyz/tempo/pull/3171/commits/fcd741ca165239949870d51bb6bf93c4b13f480e

crates/transaction-pool/src/maintain.rs
```diff
@@ -560,6 +560,26 @@ where
...
+                        evict_invalidated_from_pool(&pool, &mut state, &metrics, &updates);
+                        evict_invalidated_from_paused_pool(&mut state, &updates);
+
+                        // Handle pause/unpause events from the reorg segment.
+                        handle_pause_events(
```
fgimenez, line 642: ordering concern is valid addressed here https://github.com/tempoxyz/tempo/pull/3171/commits/e24423fb50628725b46b72527f542369d970b538

fgimenez: orphaned AA 2D txs missing the pause scan is a pre-existing issue not related to this PR

crates/transaction-pool/src/maintain.rs
```diff
@@ -560,6 +560,26 @@ where
                             error!(target: "txpool", ?err, "AMM liquidity cache repopulate after reorg failed");
                         }
 
+                        // Extract invalidation events from the new canonical segment
+                        // and evict affected transactions. Without this, events like key
+                        // revocations, spending limit changes, validator/user token changes,
+                        // and blacklist additions in the reorg segment would be ignored.
+                        let updates = TempoPoolUpdates::from_chain(&new);
+                        evict_invalidated_from_pool(&pool, &mut state, &metrics, &updates);
+                        evict_invalidated_from_paused_pool(&mut state, &updates);
```
fgimenez: valid, addressed here https://github.com/tempoxyz/tempo/pull/3171/commits/2aa668a2966b9103d4c597a8ebe1529c173610bd

PR #3173, opened 2026-03-17
TIP-1028: Address-Level Receive Policies

jenpaff:
Extends TIP-403 with token sets and address-level receive policies.


tips/tip-1028.md
```diff
@@ -0,0 +1,642 @@
...
+3. **Address-level token filtering** (token filter) — the ability for individual addresses to control which tokens they can receive.
+
+Currently, TIP-403 policies are set at the token level — each TIP-20 token has a single `transferPolicyId` that governs all transfers. This proposal adds the ability for individual addresses (EOAs or contracts) to set their own inbound controls across two dimensions:
+
+1. **Receive policy** (counterparty filter): Control who can send to this address
+2. **Token set** (token filter): Control which tokens can be sent to this address
+
+This enables regulated entities (banks, exchanges, etc.) and orchestrator contracts to enforce inbound controls at the address level, independent of and in addition to token-level policies.
+
+Outbound send restrictions are explicitly out of scope. Controlling what an address can send is best handled by mechanisms like multisigs, offchain policy engines, or smart contract wallets — not protocol-level transfer hooks.
```
0xrusowsky: similar to what you did in the virtual address spec, imo it is probably worth adding a reminder/disclaimer that address-level receive policies only apply to TIP-20 tokens.

regulated entities which opt-in to the 403 policies can still be spammed with regular ERC-20 tokens deployed on Tempo, since those are not subject to these controls.

tips/tip-1028.md
```diff
@@ -0,0 +1,642 @@
...
+
+Token sets are a new primitive in the TIP-403 Registry for managing sets of token addresses.
+
+### Storage
+
+```solidity
+uint64 public tokenSetIdCounter = 2; // Skip 0 (always-reject) and 1 (always-allow)
+
+struct TokenSetData {
+    PolicyType setType;  // WHITELIST or BLACKLIST (not COMPOUND)
```
0xrusowsky: probably worth adding a one-liner explicitly explaining why `COMPOUND` doesn't apply to address-level policies

tips/tip-1028.md
```diff
@@ -0,0 +1,642 @@
...
+The `addressPolicies` mapping packs a presence flag, policy IDs, token set IDs, and their cached types into a single 256-bit storage slot:
+
+| Bits (inclusive) | Size | Field | Description |
+|------------------|------|-------|-------------|
+| 0–63 | 64 bits | `receivePolicyId` | TIP-403 policy checked when this address is the receiver |
+| 64–71 | 8 bits | `receivePolicyType` | Cached type of receive policy (0 = whitelist, 1 = blacklist) |
+| 72–135 | 64 bits | `tokenSetId` | TIP-403 token set checked on tokens being received |
+| 136–143 | 8 bits | `tokenSetType` | Cached type of token set (0 = whitelist, 1 = blacklist) |
+| 144–254 | 111 bits | Reserved | For future use (must be 0) |
+| 255 | 1 bit | `hasReceivePolicy` | 1 if address has set receive policies, 0 if not |
```
0xrusowsky: this table is inconsistent with the encoding/decoding impl in the next section (which imo makes more sense).

it should be:

```suggestion
| Bits (inclusive) | Size | Field | Description |
|------------------|------|-------|-------------|
| 0 | 1 bit | `hasReceivePolicy` | 1 if address has set receive policies, 0 if not |
| 1–64 | 64 bits | `receivePolicyId` | TIP-403 policy checked when this address is the receiver |
| 65–72 | 8 bits | `receivePolicyType` | Cached type of receive policy (0 = whitelist, 1 = blacklist) |
| 73–136 | 64 bits | `tokenSetId` | TIP-403 token set checked on tokens being received |
| 137–144 | 8 bits | `tokenSetType` | Cached type of token set (0 = whitelist, 1 = blacklist) |
| 145–255 | 111 bits | Reserved | For future use (must be 0) |
```

0xrusowsky (changes requested):

tips/tip-1028.md
```diff
@@ -0,0 +1,642 @@
...
+        uint64 tokenSetId, PolicyType tokenSetType
+    ) 
+{
+    return decodeAddressPolicies(addressPolicies[account]);
+}
+```
+
+#### isAddressTransferAuthorized
+
+This function checks the **receiver's** controls only — the sender has no address-level policy checked. Types are embedded in `addressPolicies`, so no extra SLOADs for `policyData` or `tokenSetData`.
```
0xrusowsky: should `isAddressTransferAuthorized` bypass the counterparty check when `from` is the DEX or FeeManager?

i think that would be a better that relying on users reading the "System Precompile Addresses and Whitelists" section

tips/tip-1028.md
```diff
@@ -0,0 +1,642 @@
...
+
+    return data.setType == PolicyType.WHITELIST
+        ? tokenSetMembers[tokenSetId][token]
+        : !tokenSetMembers[tokenSetId][token];
+}
+```
+
+---
+
+## Part 2: Address-Level Receive Policies
```
horsefacts: Capturing some external discussions: There are too many edge cases around receive policies bricking contracts and batch transfers that we need to consider. It might be possible to make policy enforcement nonreverting, but we need to go back to the drawing board on the design.

tips/tip-1028.md
```diff
@@ -0,0 +1,642 @@
...
+
+### Overview
+
+The TIP-403 Registry is extended with a new mapping that allows any address to set its own inbound controls:
+
+- **Receive policy**: Controls which addresses can send tokens to this address (references a TIP-403 policy)
+- **Token set**: Controls which tokens this address can receive (references a TIP-403 token set)
+
+These controls are checked on every TIP-20 transfer and mint in addition to the existing token-level policy check. Since TIP-20 is implemented as a precompile, a hardfork that adds this functionality will apply to all tokens automatically.
```
horsefacts: Will raise externally, but I'd like to better understand our view on changing TIP-20 behavior over time.  This approach seems tricky.

horsefacts (changes requested):

tips/tip-1028.md
0xKitsune: +1, there are some sharp edge cases and I think we should revisit the underlying design.

tips/tip-1028.md
```diff
@@ -0,0 +1,720 @@
...
+It MUST NOT reference a `COMPOUND` policy.
+
+`claimerWhitelistId` MUST reference:
+
+- a simple `WHITELIST` policy, or
+- built-in policy `0` or `1`
+
+It MUST NOT reference a `BLACKLIST` or `COMPOUND` policy.
+
+### 3.2 Recovery Modes
```
danrobinson: Can we just have the user specify a recovery contract, and that recovery contract is what can call the escrow address to claim the tokens? Then we don't need a bunch of manually specified recovery modes

decofe:
### Suggestion: Pre-warm `ESCROW_ADDRESS` balance slot at token creation

Section 6.1 notes that the first blocked inbound *ever* for a given token adds ~250k gas because `balances[ESCROW_ADDRESS]` goes from zero → nonzero.

This cost can be moved to token initialization — set `balances[ESCROW_ADDRESS] = 0` (or a sentinel `1`) during TIP-20 token deployment. That way no individual transfer ever eats the extra 250k for warming that slot. The cost is amortized once per token at creation time, which is a much more predictable place to pay it.

decofe:
### Suggestion: Replace `RecoveryMode` + `claimerWhitelistId` with a single `recoveryContract` address

The current design has 4 recovery modes, a `claimerWhitelistId` referencing TIP-403, and two claim entrypoints with branching authorization logic. This is a lot of protocol-level surface area for claim authorization.

**Proposed simplification:** the receiver sets a `recoveryContract` address. The escrow only checks:

- `msg.sender == recoveryContract` → allowed
- `msg.sender == receiver` when `recoveryContract == address(0)` → allowed (default fallback)
- otherwise → revert

All claim authorization logic lives in the recovery contract — originator self-claim, delegate whitelists, multisig, timelocks, whatever. The protocol has zero opinion on claim policy.

This drops `RecoveryMode`, `claimerWhitelistId`, `claimOriginatorBucketToOriginator`, and the entire §4.3 authorization matrix from the spec. The recovery contract is a userland concern that can be iterated without protocol changes.

If `recoveryContract == address(0)`, only the receiver can claim (safe default — no contract deployment needed to get started). The receiver can upgrade to a custom recovery contract later to enable delegation or originator self-claim.

decofe:
### Suggestion: Lock recovery contract into the bucket key at transfer time

The current TIP notes that "recovery-mode changes are retroactive" — changing claim authorization affects existing escrowed funds. This is a footgun: a receiver can rug escrowed funds by swapping their recovery contract after the fact.

**Fix:** make the recovery contract part of the bucket key, locked in at the time the transfer is blocked:

```
key = hash(token, receiver, originator, recoveryContract)
value = accumulated amount
```

- At transfer time, read the receiver's current `recoveryContract` and include it in the key
- Accumulation still works — repeated blocks from the same `(token, receiver, originator, recoveryContract)` tuple still land in the same bucket (2.9k after first)
- If the receiver changes their recovery contract, new blocked transfers go to new buckets, but old buckets remain governed by the original recovery contract
- At claim time, the escrow checks `msg.sender == recoveryContract` from the key (or `msg.sender == receiver` when `recoveryContract == address(0)`)

This gives originators certainty: at the time their transfer was blocked, they can see exactly which contract governs recovery.

decofe:
### Suggestion: Use a flat hash key instead of nested mappings for escrow storage

Since this is a precompile (implemented in Rust, not Solidity), the nested mapping notation is just spec readability sugar. The actual storage should use a single flat hash:

```
slot = keccak256(abi.encode(token, receiver, originator, recoveryContract))
value = uint256 amount
```

Compared to the nested `mapping(address => mapping(address => mapping(address => uint256)))`:

- **Cheaper**: one keccak256 over the concatenated components vs four sequential hashes
- **Simpler**: one storage key derivation in the Rust implementation
- **Equivalent**: both produce a single deterministic slot, neither supports onchain enumeration, both require knowing all key components for lookup

The TIP can describe it as a logical mapping for readability while noting the actual storage layout is a flat hash.

tips/tip-1028.md
0xKitsune: This is nice, this also allows for recovery contract upgrades in the future without having.

tips/tip-1028.md
```diff
@@ -0,0 +1,825 @@
...
+        );
+
+    function classifyAddressInbound(address token, address originator, address to)
+        external
+        view
+        returns (
+            bool authorized,
+            address recoveryContract,
+            BlockedReason blockedReason
+        );
```
0xrusowsky: i think this fn is overscoped, since the `recoveryContract` info adds an extra SLOAD and that info is already available via `addressReceivePolicy`.

if we think it's still useful we can keep it, but IMO the API only needs a minimal version that allows contracts to pre-verify whether a transfer will succeed (without the extra SLOAD for the recovery contract):

```solidity
verifyAddressInbound(address token, address originator, address to)
        external
        view
        returns (bool authorized, BlockedReason blockedReason);
```

tips/tip-1028.md
```diff
@@ -0,0 +1,825 @@
...
+}
+```
+
+Changing `recoveryContract` affects only future blocked buckets. Existing buckets remain governed by the `recoveryContract` recorded when they were blocked.
+
+### 3.5 Authorization Logic
```
0xrusowsky, line 266: if we adopt the proposed `verifyAddressInbound`, this section should be updated to use that fn. Additionally, it should explicitly say that impl should only read `addressRecoveryContract[to]` if the inbound is actually blocked

tips/tip-1028.md
```diff
@@ -0,0 +1,924 @@
...
+error InvalidReceivePolicyType();
+error InvalidRecoveryContract();
+```
+
+## 4. Escrow Precompile
+
+Blocked inbounds are recorded in a dedicated escrow precompile. The raw TIP-20 balance is held at `ESCROW_ADDRESS` inside each TIP-20 token; the precompile stores receipt metadata and claimability.
+
+```solidity
+ESCROW_ADDRESS = 0xFDC1000000000000000000000000000000000000
```
malleshpai: 0xE5c00000

tips/tip-1028.md
malleshpai: done

tips/tip-1028.md
malleshpai: done

tips/tip-1028.md
```diff
@@ -0,0 +1,792 @@
...
+
+```solidity
+uint64 public blockedReceiptNonce = 1;
+mapping(bytes32 => uint256) internal blockedReceiptAmount;
+```
+
+The persistent escrow key for a blocked inbound is:
+
+```text
+receiptKey = keccak256(
```
malleshpai, line 312: add domain separator / version number

tips/tip-1028.md
```diff
@@ -0,0 +1,802 @@
...
+protocolVersion: TBD
+---
+
+# TIP-1028: Address-Level Receive Policies
+
+## Abstract
+
+This TIP extends TIP-403 with token sets for TIP-20 token addresses, address-level receive controls, and non-reverting escrow for blocked TIP-20 inbounds. A blocked inbound credits `ESCROW_ADDRESS` and creates one fine-grained receipt bucket.
+
+The escrow precompile stores only one keyed amount for each open blocked receipt. The rest of the receipt identity — receipt version, requested recipient, block reason, transfer-vs-mint kind, memo, timestamp, and nonce — is authenticated by the receipt witness and emitted in the blocked event. Claims may unwind to the receiver or reroute elsewhere; claim-to-receiver is an unwind, while reroutes are new spends. This TIP applies only to TIP-20 precompile flows; ordinary ERC-20 contracts remain out of scope.
```
0xKitsune: nit: the abstract jumps into details like receipt version, transfer-vs-mint kind, etc. before the reader actually is introduced to this context. I think the abstract should be a bit higher level. 


```suggestion
TIP-1028 extends TIP-403 with address-level receive policies and token sets, allowing a receiver to control which originators and which TIP-20 tokens may credit it. When a receive policy blocks an inbound transfer or mint, the operation still succeeds at the protocol level, but the funds are credited to `ESCROW_ADDRESS` instead of the receiver and a corresponding escrow receipt is recorded.

The receiver or a designated recovery contract may later claim these escrowed funds. Claims back to the receiver act as an unwind of the original inbound, while claims to other addresses are treated as new transfers.

Token-level authorization remains unchanged and continues to revert on failure. This behavior applies only to TIP-20 precompile flows; ordinary contracts and other precompiles are unaffected.
```

tips/tip-1028.md
```diff
@@ -0,0 +1,802 @@
...
+
+## Motivation
+
+TIP-403 lets token issuers decide who may use a token, but some receivers also need inbound controls. A revert-based receiver policy creates a liveness problem: once a relationship exists, the receiver can later change policy and cause future transfers or mints to revert.
+
+This TIP keeps issuer-side semantics unchanged and changes only receiver-side failure handling. Token-level failures still revert; receiver-side failures are escrowed instead. The escrow representation must preserve more than an aggregate amount because TIP-1022 virtual addresses carry attribution in the literal `to` address, memo-bearing transfers carry memo data, recovery contracts may want time-based or memo-based rules, and offchain recovery flows need the original sender identity.
```
0xKitsune:
```suggestion
TIP-403 allows token issuers to control who may use a token, but it does not give receivers control over which inbounds they accept. Some applications require receivers to restrict incoming transfers or mints based on the originator or token.

A revert-based receiver policy introduces a liveness problem: once a sender relationship exists, a receiver can later change its policy and cause future transfers or mints to revert, breaking integrations and making delivery unreliable.

This TIP introduces receiver-side controls while preserving liveness. Instead of reverting when a receiver blocks an inbound, the transfer or mint succeeds at the protocol level and the funds are escrowed. This ensures that senders and integrations are not disrupted, while still giving receivers control over acceptance.

Blocked inbounds must remain recoverable and attributable, allowing the escrow model to preserve per-transfer context such as originator, requested recipient, and metadata, enabling flexible recovery and offchain handling.
```

tips/tip-1028.md
```diff
@@ -0,0 +1,802 @@
...
+
+If a token-level TIP-403 or TIP-1015 policy rejects the operation, the operation MUST revert exactly as it does today. If the receiver's address-level controls reject the inbound, the operation MUST be escrowed instead.
+
+If `to` is a TIP-1022 virtual address, TIP-1022 recipient resolution MUST occur before TIP-1028 receiver-side authorization. In that case:
+
+- TIP-1028 applies to the resolved master address, not the literal virtual address;
+- if virtual-address resolution fails, the operation MUST revert rather than escrow;
+- the blocked receipt's `receiver` MUST be the resolved master address;
+- the blocked receipt's `requestedRecipient` MUST be the literal `to` address so offchain systems can recover the TIP-1022 `userTag`; and
+- if the inbound is authorized, the success path MUST then follow TIP-1022 forwarding semantics.
```
0xKitsune: This is talking about "blocked receipt" before its introduced/defined. Virtual addresses should be reintroduced when talking about the claim path in `Claim Authorization` or `Release Semantics`. I would remove it from here and simplify. 


```suggestion
If `to` is a TIP-1022 virtual address, TIP-1022 recipient resolution MUST occur before TIP-1028 receiver-side authorization. If virtual-address resolution fails, the operation MUST revert.
```

tips/tip-1028.md
```diff
@@ -0,0 +1,802 @@
...
+
+TIP-1028 does **not** alter `approve`, `permit`, `burn`, fee refunds via `transfer_fee_post_tx`, non-TIP-20 tokens deployed as ordinary contracts, or future recipient-bearing system-credit paths that do not identify a concrete originator.
```
0xKitsune: > non-TIP-20 tokens deployed as ordinary contracts

I dont think its necessary to mention this again. We should mention that "TIP-1028 only applies to TIP-20 precompile flows; ordinary contracts and other precompiles are unaffected" once in the overview, rather than restate this.

tips/tip-1028.md
```diff
@@ -0,0 +1,802 @@
...
+
+DEX internal balances are not TIP-20 wallet balances and are not subject to address-level receive checks until withdrawn back onto the TIP-20 ledger.
```
0xKitsune: Similar to the comment above, I dont think we need to mention this as this is implied when stating that only TIP-20 precompiles are affected.

tips/tip-1028.md
```diff
@@ -0,0 +1,802 @@
...
+
+The reward subsystem is not receiver-side escrowable, but it is not exempt from recipient consent. `distributeReward` remains token-internal accounting; `setRewardRecipient` MUST reject a nonzero recipient whose address-level receive controls would block the holder as originator; and `claimRewards` MUST revalidate the actual payout recipient's address-level receive controls against the token contract as originator and MUST revert on failure rather than escrow.
+
+For reward accounting, `ESCROW_ADDRESS` is a reward-exempt always-opted-out synthetic sink/source: blocked transfers, blocked mints, and claim releases MUST preserve the same opted-in-supply effects as a movement into or out of an always-opted-out address, and implementations MUST NOT create, update, or consult per-user reward state for `ESCROW_ADDRESS`.
```
0xKitsune: This feels a bit over-specified, once the scope is limited to recipient-bearing TIP-20 ledger flows, internal accounting flows like rewards or DEX internal balances should already be implied out of scope. We should avoid enumerating every internal balance system unless there’s a non-obvious exception.

If we want to keep something to explictly callout rewards since they are part of TIP20s, I would suggest that we update to:


```suggestion
Reward flows are not subject to escrow. Reward payouts MUST still respect address-level receive controls and MUST revert if the recipient would reject the inbound.
```

tips/tip-1028.md
```diff
@@ -0,0 +1,802 @@
...
+
+High-level flow:
+
+1. If `to` is a TIP-1022 virtual address, resolve it first to `effectiveReceiver`; otherwise `effectiveReceiver = to`.
+2. Run the existing TIP-20 sender-side or issuer-side checks and the token's TIP-403 and TIP-1015 checks using `effectiveReceiver` wherever the current path is recipient-sensitive.
```
0xKitsune: nit:
```suggestion
2. Run existing TIP-20 sender/issuer checks, including TIP-403 and TIP-1015, using `effectiveReceiver` where applicable.
```

tips/tip-1028.md
```diff
@@ -0,0 +1,802 @@
...
+3. Run the receiver's token-set and receive-policy checks against `effectiveReceiver`.
+4. If the inbound is authorized, credit the receiver normally.
+5. If the inbound is blocked by the receiver, credit `ESCROW_ADDRESS`, create one blocked receipt, and emit a blocked-inbound event naming both `receiver` and `requestedRecipient`.
+6. Later, the receiver or the receipt's recovery contract may claim the funds.
+
```
0xKitsune, line 69: It would be nice to have a simple sequence daigram demonstrating the high level flow here.

```suggestion

Below is a simplified sequence diagram showing the TIP-20 inbound flow, including escrow of blocked inbounds and how funds are claimed.

sequenceDiagram
    actor Sender
    participant TIP20 as TIP-20
    participant Escrow
    actor Receiver

    Sender->>TIP20: transfer / mint to Receiver
    TIP20->>TIP20: resolve receiver and evaluate inbound rules
    TIP20->>Escrow: record blocked inbound receipt
    TIP20->>TIP20: credit ESCROW_ADDRESS
    TIP20-->>Sender: success

    Receiver->>Escrow: claim blocked receipt
    Escrow->>TIP20: release from ESCROW_ADDRESS
    TIP20-->>Receiver: credit claimed funds

```

tips/tip-1028.md
```diff
@@ -0,0 +1,802 @@
...
+## 2. Token Sets
+
+Token sets are a dedicated TIP-403 primitive for TIP-20 token addresses. They answer a different question from address policies:
+
+- address policy: "is this originator authorized?"
+- token set: "is this token authorized?"
+
+Token sets use a separate ID space from policy IDs. They are not aliases for ordinary TIP-403 policy lists and do not reuse the compound-policy surface, but they mirror the ordinary TIP-403 list ergonomics, including create-with-members and batched membership updates.
```
0xKitsune: It feels like token sets are introduced too early without context. The core mental model of this TIP is the inbound flow (credit vs escrow), it would be clearer to introduce escrow and the overall flow first, then define token sets and receive policies as part of the receiver-side authorization step.

tips/tip-1028.md
```diff
@@ -0,0 +1,802 @@
...
+    address admin;
+}
+
+mapping(uint64 => TokenSetData) internal _tokenSetData;
+mapping(uint64 => mapping(address => bool)) internal tokenSetMembers;
+```
+
+Built-in meanings: `0` = always reject; `1` = always allow.
+
+Token sets MUST satisfy: `setType` is `WHITELIST` or `BLACKLIST`; `COMPOUND` token sets are forbidden; token-set type is immutable after creation; and membership is mutable by the token-set admin.
```
0xKitsune, line 119: Similar to the comment above, storage and constraints are introduced without actually introducing what contract this relates to. Would should introduce the escrow precompile first to make it easier to build a mental model of the subsequent moving parts.

tips/tip-1028.md
```diff
@@ -0,0 +1,802 @@
...
+interface ITIP403TokenSets {
+    function createTokenSet(address admin, PolicyType setType)
+        external
+        returns (uint64 newTokenSetId);
+
+    function createTokenSetWithTokens(
+        address admin,
+        PolicyType setType,
+        address[] calldata tokens
+    ) external returns (uint64 newTokenSetId);
```
0xKitsune, line 133: It’s not clear what contract implements the `ITIP403TokenSets` interface. Im under the impression it should be the TIP403 precompile but the spec does not explicitly state this. Would be good to make this explicit to avoid confusion.

tips/tip-1028.md
```diff
@@ -0,0 +1,802 @@
...
+error TokenSetNotFound();
+error InvalidTokenSetType();
+error TokenSetBatchLengthMismatch();
+```
+
+## 3. Address-Level Receive Controls
+
+Any address MAY configure three fields: `receivePolicyId` for which originators may credit the address, `tokenSetId` for which TIP-20 token addresses may credit the address, and `recoveryContract` for an optional claimer authorized to recover blocked receipts one receipt at a time; `address(0)` means the receiver claims directly. If an address has no configured receive controls, address-level authorization defaults to allow.
+
+TIP-1022 virtual addresses are forwarding aliases, not canonical TIP-20 holders. `setAddressReceivePolicy()` MUST reject TIP-1022 virtual addresses and require configuration on the resolved master address instead.
```
0xKitsune, line 194: “Any address MAY configure…” is unclear. Where is this configuration stored and which contract exposes `setAddressReceivePolicy()`? It would help to explicitly state which component owns address-level receive controls. At the moment, the reader has to infer it which is confusing.

PR #3181, opened 2026-03-17
fix(dex): check recipient auth on payout token in cancel_stale_order

fgimenez:
Closes SIGP-163

`cancel_stale_order` now also checks recipient authorization on the payout token (T4+), so orders from makers blacklisted as recipients can be cleaned up by third parties. Pre-T4 behavior is unchanged.

### DIFF
--- a/crates/precompiles/src/stablecoin_dex/mod.rs
+++ b/crates/precompiles/src/stablecoin_dex/mod.rs
@@ -1202,38 +1202,19 @@ impl StablecoinDEX {
     /// T4+: also checks recipient authorization on the payout token (bid=base, ask=quote).
     fn is_maker_authorized(&self, order: &Order) -> Result<bool> {
         let book = self.books[order.book_key()].read()?;
-        let registry = TIP403Registry::new();
 
-        let escrow_token = if order.is_bid() {
-            book.quote
+        let (token_in, token_out) = if order.is_bid() {
+            (book.quote, book.base)
         } else {
-            book.base
+            (book.base, book.quote)
         };
-        let escrow_policy_id = TIP20Token::from_address(escrow_token)?.transfer_policy_id()?;
-        let sender_authorized =
-            match registry.is_authorized_as(escrow_policy_id, order.maker(), AuthRole::sender()) {
-                Ok(authorized) => authorized,
-                Err(e) if is_policy_lookup_error(&e) => false,
-                Err(e) => return Err(e),
-            };
 
-        if !sender_authorized {
+        if !is_authorized_for_token(token_in, order.maker(), AuthRole::sender())? {
             return Ok(false);
         }
 
         if self.storage.spec().is_t4() {
-            let payout_token = if order.is_bid() {
-                book.base
-            } else {
-                book.quote
-            };
-            let payout_policy_id = TIP20Token::from_address(payout_token)?.transfer_policy_id()?;
-            match registry.is_authorized_as(payout_policy_id, order.maker(), AuthRole::recipient())
-            {
-                Ok(authorized) => Ok(authorized),
-                Err(e) if is_policy_lookup_error(&e) => Ok(false),
-                Err(e) => Err(e),
-            }
+            is_authorized_for_token(token_out, order.maker(), AuthRole::recipient())
         } else {
             Ok(true)
         }
@@ -1548,6 +1529,18 @@ impl StablecoinDEX {
         }
 
         Ok(amount_out)
+    }
+}
+
+/// Checks whether `address` is authorized under the transfer policy of `token` for the given
+/// `role`. Returns `false` instead of erroring when the policy lookup fails.
+fn is_authorized_for_token(token: Address, address: Address, role: AuthRole) -> Result<bool> {
+    let policy_id = TIP20Token::from_address(token)?.transfer_policy_id()?;
+    let registry = TIP403Registry::new();
+    match registry.is_authorized_as(policy_id, address, role) {
+        Ok(authorized) => Ok(authorized),
+        Err(e) if is_policy_lookup_error(&e) => Ok(false),
+        Err(e) => Err(e),
     }
 }
 
--- a/crates/precompiles/src/stablecoin_dex/mod.rs
+++ b/crates/precompiles/src/stablecoin_dex/mod.rs
@@ -882,32 +882,6 @@ impl StablecoinDEX {
         Ok((amount_out, next_tick_info))
     }
 
-    /// T2+: cancel stale orders at the head of the queue until an authorized maker is found.
-    /// Returns the updated level and order, or `None` if no authorized orders remain.
-    fn skip_stale_orders(
-        &mut self,
-        book_key: B256,
-        bid: bool,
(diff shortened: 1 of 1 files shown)


0xrusowsky (approved):
lgtm, noting that it will have merge conflicts with https://github.com/tempoxyz/tempo/pull/3015 depending on which one is merged first

crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -1184,18 +1185,45 @@ impl StablecoinDEX {
...
+            true
+        };
 
-        let policy_id = TIP20Token::from_address(token)?.transfer_policy_id()?;
-        // Invalid policy ids throw under_overflow. Treat as unauthorized to clear the orders.
-        match TIP403Registry::new().is_authorized_as(policy_id, order.maker(), AuthRole::sender()) {
-            Ok(true) => Err(StablecoinDEXError::order_not_stale().into()),
-            Err(e) if e != TempoPrecompileError::under_overflow() => Err(e),
-            _ => self.cancel_active_order(order),
+        if sender_authorized && recipient_authorized {
```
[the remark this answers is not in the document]
fgimenez: valid but pre-existing, we can address separately

crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -1184,18 +1185,45 @@ impl StablecoinDEX {
...
             book.quote
         } else {
             book.base
         };
+        let escrow_policy_id = TIP20Token::from_address(escrow_token)?.transfer_policy_id()?;
+        let sender_authorized =
+            match registry.is_authorized_as(escrow_policy_id, order.maker(), AuthRole::sender()) {
+                Ok(authorized) => authorized,
+                // Invalid policy ids throw under_overflow. Treat as unauthorized.
+                Err(e) if e == TempoPrecompileError::under_overflow() => false,
```
legion2002: Noting that this error will change after we merge the fix for InvalidPolicyTypes. Not sure how to schedule this, but it is important that we don't forget to update this error once the fix is merged.

CC: @0xrusowsky @0xKitsune

crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -1184,18 +1185,45 @@ impl StablecoinDEX {
...
+            let payout_token = if order.is_bid() {
+                book.base
+            } else {
+                book.quote
+            };
+            let payout_policy_id = TIP20Token::from_address(payout_token)?.transfer_policy_id()?;
+            match registry.is_authorized_as(payout_policy_id, order.maker(), AuthRole::recipient())
+            {
+                Ok(authorized) => authorized,
+                Err(e) if e == TempoPrecompileError::under_overflow() => false,
```
legion2002: same for this one

legion2002 (approved):
lgtm, but I think there is some race condition with if we fix the InvalidPolicyType error first or later.

jenpaff (changes requested):
lets put this behind T3 for now , this will also need a new TIP

crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -885,8 +911,14 @@ impl StablecoinDEX {
...
-        let mut level = self.get_best_price_level(book_key, bid)?;
-        let mut order = self.orders[level.head].read()?;
+        let (mut level, mut order) = match self.skip_stale_orders(book_key, bid)? {
+            Some(result) => result,
+            None => {
+                let level = self.get_best_price_level(book_key, bid)?;
+                let order = self.orders[level.head].read()?;
+                (level, order)
+            }
+        };
```
0xKitsune: I dont think swap logic should be changed here. Lets make sure @danrobinson is in the loop on the proposed changeset.

0xKitsune (changes requested):

crates/precompiles/src/stablecoin_dex/mod.rs
howydev: good call, IIRC there was a slack thread and in the discussion we did not want to incur the additional storage reads while iterating thru orders

crates/precompiles/src/stablecoin_dex/mod.rs
fgimenez: should be fixed now, ptal

crates/precompiles/src/stablecoin_dex/mod.rs
fgimenez: fixed here https://github.com/tempoxyz/tempo/pull/3181/commits/3b65897892611ffa792594acdd2eab812bdb2c84 ptal

0xrusowsky (approved):

crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -1188,21 +1189,53 @@ impl StablecoinDEX {
...
+        if !sender_authorized {
+            return Ok(false);
+        }
+
+        if self.storage.spec().is_t4() {
+            let payout_token = if order.is_bid() {
+                book.base
+            } else {
+                book.quote
+            };
```
0xKitsune: Since the `token_in` / `token_out` logic is identical here as above, we can consolidate it into one place.

```rust

  let (token_in, token_out) = if order.is_bid() {
            (book.quote, book.base)
        } else {
            (book.base, book.quote)
        };
```

crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -1188,21 +1189,53 @@ impl StablecoinDEX {
...
         } else {
             book.base
         };
+        let escrow_policy_id = TIP20Token::from_address(escrow_token)?.transfer_policy_id()?;
+        let sender_authorized =
+            match registry.is_authorized_as(escrow_policy_id, order.maker(), AuthRole::sender()) {
+                Ok(authorized) => authorized,
+                Err(e) if is_policy_lookup_error(&e) => false,
+                Err(e) => return Err(e),
+            };
```
0xKitsune: Similar to the above comment, this logic is identical when checking the token in and token out. Can we create a helper function for this?

```rust
fn is_authorized(
    policy_id: PolicyId,
    address: Address,
    role: AuthRole,
) -> Result<bool, Error> {
    let registry = TIP403Registry::new();

    match registry.is_authorized_as(policy_id, address, role) {
        Ok(authorized) => Ok(authorized),
        Err(e) if is_policy_lookup_error(&e) => Ok(false),
        Err(e) => Err(e),
    }
}
```

0xKitsune (approved):

crates/precompiles/src/stablecoin_dex/mod.rs
fgimenez: good point done

crates/precompiles/src/stablecoin_dex/mod.rs
fgimenez: same, done

jenpaff (approved):

PR #3192, opened 2026-03-18
fix(dex): check recipient auth in cancel_stale_order

howydev:
On T2+, also check payout token recipient authorization so orders from deauthorized recipients can be cleaned up.

Closes ZELLIC-66

### DIFF
--- a/crates/precompiles/src/stablecoin_dex/mod.rs
+++ b/crates/precompiles/src/stablecoin_dex/mod.rs
@@ -1184,19 +1184,34 @@ impl StablecoinDEX {
         }
 
         let book = self.books[order.book_key()].read()?;
-        let token = if order.is_bid() {
-            book.quote
+        let (escrow_token, payout_token) = if order.is_bid() {
+            (book.quote, book.base)
         } else {
-            book.base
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -1184,19 +1184,34 @@ impl StablecoinDEX {
...
         }
+
+        // (T2+) Check recipient auth on payout token
+        if self.storage.spec().is_t2() {
+            let payout_policy = TIP20Token::from_address(payout_token)?.transfer_policy_id()?;
+            match registry.is_authorized_as(payout_policy, order.maker(), AuthRole::recipient()) {
+                Ok(true) => {}
+                Err(e) if e != TempoPrecompileError::under_overflow() => return Err(e),
+                _ => return self.cancel_active_order(order),
+            }
```
fgimenez, line 1211: we should change this after #3015 merges, we should use `is_policy_lookup_error`

fgimenez (approved):
lgtm, should be merged after #3015

PR #3196, opened 2026-03-18
fix(pool): resolve compound policy sub-policy IDs for eviction checks

fgimenez:
Closes SIGP-162

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7762,7 +7762,7 @@ checksum = "1e061d1b48cb8d38042de4ae0a7a6401009d6143dc80d2e2d6f31f0bdd6470c7"
 [[package]]
 name = "reth-basic-payload-builder"
 version = "1.11.3"
-source = "git+https://github.com/paradigmxyz/reth?rev=9060c50#9060c5059e0ca15813ab97a72e33d04cd2e7d998"
+source = "git+https://github.com/paradigmxyz/reth?rev=a0b0d88#a0b0d8854c87683390d94ba6fb3a535fc79a0db3"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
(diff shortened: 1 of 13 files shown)


crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -1091,26 +1099,49 @@ pub(crate) fn exceeds_spending_limit(
...
+        if let Ok(data) = registry.policy_records[policy_id].base.read()
+            && data.is_compound()
+            && let Ok(compound) = registry.policy_records[policy_id].compound.read()
+        {
+            for sub_id in [
+                compound.sender_policy_id,
+                compound.recipient_policy_id,
+                compound.mint_recipient_policy_id,
+            ] {
+                if sub_id != 0 && !ids.contains(&sub_id) {
```
howydev: nice

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -1091,26 +1099,49 @@ pub(crate) fn exceeds_spending_limit(
...
+                    ids.push(sub_id);
+                }
+            }
+        }
+
+        cache.insert(fee_token, ids.clone());
```
howydev, line 1136: sanity check - compound policies have immutable references to the simple policies, so we wouldn’t need to insert it into cache also right?

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -1091,26 +1099,49 @@ pub(crate) fn exceeds_spending_limit(
...
+
+        // Resolve compound policies to include sub-policy IDs
+        let registry = TIP403Registry::new();
+        if let Ok(data) = registry.policy_records[policy_id].base.read()
+            && data.is_compound()
+            && let Ok(compound) = registry.policy_records[policy_id].compound.read()
+        {
+            for sub_id in [
+                compound.sender_policy_id,
+                compound.recipient_policy_id,
```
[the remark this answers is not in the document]
fgimenez: addressed here https://github.com/tempoxyz/tempo/pull/3196/commits/203be8baa0429a005ceb168ecb54ae302711043b

crates/transaction-pool/src/tempo_pool.rs
fgimenez: yep good point, this cache aims to prevent multiple SLOADs when txs share the same fee token, added a comment about it, lmk if it makes sense

PR #3200, opened 2026-03-18
fix(fees): reject self-sponsored fee payer signatures

legion2002:
Addresses CHAIN-959.

Rejects AA transactions when a `fee_payer_signature` resolves back to the sender, so self-sponsored signatures cannot bypass fee-payer assumptions. This is enforced in txpool validation and EVM fee-payer resolution, and the fee-payer negative e2e scenario now includes a self-sponsored rejection case.

### DIFF
--- a/crates/node/tests/it/tempo_transaction/rpc.rs
+++ b/crates/node/tests/it/tempo_transaction/rpc.rs
@@ -226,7 +226,9 @@ impl super::types::TestEnv for RpcEnv {
             .await;
 
         match result {
-            Ok(_) => panic!("Transaction should be rejected, but was accepted"),
+            Ok(_) => Err(eyre::eyre!(
+                "Transaction should be rejected, but was accepted"
+            )),
             Err(RpcError::Transport(_)) => {
                 // Transport error that persisted through all retries — not a
                 // real rejection, so we must not count it as a test pass.


crates/node/tests/it/tempo_transaction/rpc.rs
```diff
@@ -209,11 +209,19 @@ async fn wait_for_receipt(
...
+                "eth_getTransactionReceipt".into(),
+                [tx_hash],
+            )
+            .await
+        {
+            Ok(Some(receipt)) => return Ok(receipt),
+            Ok(None) => {}
+            // Remote endpoints occasionally return transient Cloudflare 403 pages.
+            // Retry instead of failing the whole matrix immediately.
+            Err(err) if err.to_string().contains("HTTP error 403") => {}
```
fgimenez: better match the actual transport error status
```suggestion
            Err(alloy::transports::RpcError::Transport(
                alloy::transports::TransportErrorKind::HttpError(ref e),
            )) if e.status == 403 => {}
```

crates/revm/src/tx.rs
```diff
@@ -94,7 +94,11 @@ impl TempoTxEnv {
     /// Resolves fee payer from the signature.
     pub fn fee_payer(&self) -> Result<Address, TempoInvalidTransaction> {
         if let Some(fee_payer) = self.fee_payer {
-            fee_payer.ok_or(TempoInvalidTransaction::InvalidFeePayerSignature)
+            let fee_payer = fee_payer.ok_or(TempoInvalidTransaction::InvalidFeePayerSignature)?;
+            if fee_payer == self.caller() {
+                return Err(TempoInvalidTransaction::SelfSponsoredFeePayer);
```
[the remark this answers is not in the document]
legion2002: Fixed in https://github.com/tempoxyz/tempo/pull/3200/commits/3a74288ec197741fc808063a524cd5301f908a0c

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -240,12 +240,14 @@ where
...
-                // NOTE: sponsored txs don't consume the sender's key limits.
+                // Sponsored txs (fee payer != sender) don't consume sender spending limits.
+                // Use recovered fee payer instead of signature presence so self-sponsored
+                // fee payer signatures are treated as sender-paid.
                 && tx
                     .transaction
                     .inner()
-                    .as_aa()
-                    .is_none_or(|aa| aa.tx().fee_payer_signature.is_none())
+                    .fee_payer(tx.transaction.sender())
```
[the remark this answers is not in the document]
legion2002: Fixed in https://github.com/tempoxyz/tempo/pull/3200/changes/d0d5a7afa2d4997d64e5387762bf77628b17329d

crates/revm/src/tx.rs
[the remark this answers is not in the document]
legion2002: Fixed in https://github.com/tempoxyz/tempo/commit/3a74288ec197741fc808063a524cd5301f908a0c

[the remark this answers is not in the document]
legion2002: Fixed here https://github.com/tempoxyz/tempo/pull/3200/commits/6c10eba276c0dd67f0fe23abdd8a8bf3132f4910

crates/revm/src/tx.rs
```diff
@@ -100,6 +100,11 @@ impl TempoTxEnv {
         }
     }
 
+    /// Returns true if transaction carries a fee payer signature.
+    pub fn has_fee_payer_signature(&self) -> bool {
+        self.fee_payer.is_some()
+    }
```
0xKitsune, line 106: Remind me why we use this format to represent fee_payer?

```rust

/// Optional fee payer specified for the transaction.
    ///
    /// - Some(Some(address)) corresponds to a successfully recovered fee payer
    /// - Some(None) corresponds to a failed recovery and means that transaction is invalid
    /// - None corresponds to a transaction without a fee payer
```

Rather than `Some(address)` and `None`. Seems counter intuitive to represent `fee_payer` like this when we could use explicit enums or a simple Option (compared to representing some meaning via `Some(Some())`).

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -230,6 +230,12 @@ where
             if !updates.spending_limit_changes.is_empty()
                 && let Some(ref subject) = keychain_subject
                 && subject.matches_spending_limit_update(&updates.spending_limit_changes)
+                // Sponsored txs (fee payer != sender) don't consume sender spending limits.
+                && tx
+                    .transaction
+                    .inner()
+                    .fee_payer(tx.transaction.sender())
+                    .map_or(true, |fee_payer| fee_payer == tx.transaction.sender())
```
0xKitsune: Why do we need this (or to check for `fee_payer == tx.transaction.sender()`) anywhere in tx pool maintenace if we are rejecting these txs during validation?

legion2002: We technically don't need these checks, because the gating already happens outside. 
We just do it for completeness, so that the pool eviction logic is correct in isolation of the external validation.

Would aso be fine to remove it, but the check is cheap, so kept it as a defense in depth improvement.

crates/revm/src/tx.rs
legion2002: I am not sure actually. Seems like it was introduced in this PR - https://github.com/tempoxyz/tempo/pull/351

going through the reason.

legion2002: amp says
```
This Some(Some(_)) / Some(None) / None shape is an internal tri-state, not a wire format.

It encodes two independent facts in one field: whether a fee-payer signature was present, and whether recovery succeeded. That behavior is documented and consumed in TempoTxEnv.fee_payer and has_fee_payer_signature.
A plain Option<Address> would collapse “no signature” and “bad signature” into the same None, but those are semantically different for validation and policy checks.
In RPC conversion, we intentionally preserve “signature provided but not recoverable” as Some(None) so estimate/fill paths can still construct AA envs before final validation, as shown in try_into_tx_env and the test test_estimate_gas_invalid_fee_payer_signature_keeps_unresolved_fee_payer.
Final execution still rejects invalid signatures via ctx.tx.fee_payer()?, and there’s an explicit unit test for this in test_fee_payer_invalid_signature_rejected.

You’re right that nested Option is not very ergonomic; an explicit enum would be clearer while preserving the same semantics.
```

crates/transaction-pool/src/tempo_pool.rs
0xKitsune: I think we should remove this if we are checking during tx pool validation and at the protocol level. This is an unnecessary check IMO.

crates/revm/src/tx.rs
0xKitsune: We should remove if there is not a clear reason for this and prefer `Option<Address>` instead. We can do this in a followup.

legion2002: yup we can update in a followup

crates/transaction-pool/src/tempo_pool.rs
legion2002: removed here https://github.com/tempoxyz/tempo/pull/3200/changes/2ebcc270c83ca3f3fcbc173033139245cce706f1

PR #3202, opened 2026-03-18
fix(account-keychain): require tx.origin for admin ops on T2

legion2002:
Ref AUD-39

T2 now scopes AccountKeychain admin operations to the caller account by requiring both `transaction_key == Address::ZERO` and `msg_sender == tx_origin` before `authorizeKey`, `revokeKey`, and `updateSpendingLimit` can execute. This prevents contract-mediated confused-deputy flows from reusing transaction-global root-key status to self-administer keys.

Adds a T2 regression test covering all three admin operations and updates existing T2 replay-protection setup to seed `tx_origin` in tests.

### DIFF
--- a/tips/ref-impls/test/invariants/AccountKeychain.t.sol
+++ b/tips/ref-impls/test/invariants/AccountKeychain.t.sol
@@ -8,6 +8,7 @@ import { InvariantBaseTest } from "./InvariantBaseTest.t.sol";
 /// @notice Fuzz-based invariant tests for the AccountKeychain precompile
 /// @dev Tests invariants TEMPO-KEY1 through TEMPO-KEY19 for access key management
 ///      Note: TEMPO-KEY20/21 require integration tests (transient storage for transaction_key)
+/// forge-config: default.isolate = true
 contract AccountKeychainInvariantTest is InvariantBaseTest {
 
     /// @dev Starting offset for key ID address pool (distinct from zero address)


crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -385,6 +371,33 @@ impl AccountKeychain {
...
+    ///
+    /// `tx_origin` is seeded by the handler before tx execution. Some direct precompile
+    /// harnesses don't populate it, in which case we preserve legacy behavior.
+    fn ensure_admin_caller(&self, msg_sender: Address) -> Result<()> {
+        let transaction_key = self.transaction_key.t_read()?;
+        if transaction_key != Address::ZERO {
+            return Err(AccountKeychainError::unauthorized_caller().into());
+        }
+
+        if self.storage.spec().is_t2() {
```
legion2002, line 399: not valid

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -385,6 +374,43 @@ impl AccountKeychain {
...
+            let tx_origin = self.tx_origin.t_read()?;
+            if tx_origin != Address::ZERO
+                && tx_origin != msg_sender
+                && self.is_contract_account(msg_sender)?
```
0xrusowsky: let's drop `fn is_contract_account` and reuse existing patterns: `self.storage.has_bytecode(msg_sender)?`

NOTE: u'll have to move `StorageCtx`'s `fn has_bytecode()` out of the `test-utils` cfg

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -385,6 +374,43 @@ impl AccountKeychain {
...
+    ///
+    /// The T2 check prevents transaction-global root-key status from being reused by
+    /// intermediate contracts (confused-deputy self-administration).
+    ///
+    /// `tx_origin` is seeded by the handler before tx execution. Some test harnesses may
+    /// emulate `msg.sender`/`tx.origin` with EOA mismatches, so we scope the additional
+    /// check to contract callers where confused-deputy risk exists.
+    fn ensure_admin_caller(&self, msg_sender: Address) -> Result<()> {
+        let transaction_key = self.transaction_key.t_read()?;
+        if transaction_key != Address::ZERO {
```
0xrusowsky: nit: we could inline `if !self.transaction_key.t_read()?.is_zero() {`

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -385,6 +374,43 @@ impl AccountKeychain {
...
+                && self.is_contract_account(msg_sender)?
+            {
+                return Err(AccountKeychainError::unauthorized_caller().into());
+            }
+        }
+
+        Ok(())
+    }
+
+    /// Returns true if the account has deployed bytecode.
```
fgimenez: nit, maybe we could explain how the fn can return Err

legion2002: added in https://github.com/tempoxyz/tempo/pull/3202/changes/0353a294ad47c6a6021926ce1122910c972030b7

crates/precompiles/src/account_keychain/mod.rs
legion2002: added in https://github.com/tempoxyz/tempo/pull/3202/changes/0353a294ad47c6a6021926ce1122910c972030b7

crates/precompiles/src/account_keychain/mod.rs
legion2002: I've actually removed the `is_contract_account` altogethere in this commit - https://github.com/tempoxyz/tempo/pull/3202/changes/0353a294ad47c6a6021926ce1122910c972030b7

we strictly require msg.sender == tx.origin now. We were making this exemption before for cases like eth_estimateGas and eth_call where tx.origin may not be seeded properly.

But we fix this now, by seeding the tx.origin in the `inspect_run` and `run` both. Lmk if this makes sense.

0xrusowsky: i think that makes sense

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -385,6 +374,36 @@ impl AccountKeychain {
         self.tx_origin.t_write(origin)
     }
 
+    /// Ensures admin operations are authorized for this caller.
+    ///
+    /// Rules:
+    /// - transaction must be signed by the main key (`transaction_key == Address::ZERO`)
+    /// - T2+: if tx.origin is set, caller must match tx.origin
```
0xKitsune: >  T2+: if tx.origin is set, caller must match tx.origin

nit: is there a case where the tx origin wouldnt be set? Seems like this should be updated.

crates/revm/src/handler.rs
```diff
@@ -674,14 +696,6 @@ where
...
-        // This must be done for ALL transactions so precompiles can access it.
-        StorageCtx::enter_evm(journal, block, cfg, tx, || {
-            let mut keychain = AccountKeychain::new();
-            keychain.set_tx_origin(tx.caller())
-        })
-        .map_err(|e| EVMError::Custom(e.to_string()))?;
-
         // Validate fee token has TIP20 prefix before loading balance.
         // This prevents panics in get_token_balance for invalid fee tokens.
         // Note: Full fee token validation (currency check) happens in load_fee_fields,
```
0xKitsune, line 709: Why was the relevant logic in `seed_tx_origin` was moved out of `validate_against_state_and_deduct_caller` and into run()/inspect_run()?


The `FoundryHandler` in `tempo-foundry` overrides `inspect_run` and only calls `load_fee_fields`, so `tx_origin` stays zero, and the `tx_origin != Address::ZERO` guard in `ensure_admin_caller` silently passes.

The simplest remediation is keeping `set_tx_origin` where it was in `validate_against_state_and_deduct_caller` and additionally calling it from `run()/inspect_run()` for the eth_call/eth_estimateGas simulation path — it's idempotent, requires no API changes, and doesn't break downstream handler consumers.

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -745,6 +765,124 @@ mod tests {
         })
     }
 
+    #[test]
+    fn test_admin_operations_require_tx_origin_on_t2() -> eyre::Result<()> {
```
0xKitsune, line 771: Missing a happy path test where a contract account is both `msg.sender` and `tx.origin`. The two existing tests cover rejection but not that a legitimate contract-origin caller can still `authorizeKey`/`revokeKey`/`updateSpendingLimit` on `T2`.

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -745,6 +765,124 @@ mod tests {
         })
     }
 
+    #[test]
```
0xKitsune, line 770: Add a matching test on a pre-T2 spec where `msg.sender != tx.origin` and assert admin ops succeed.

crates/revm/src/handler.rs
legion2002: ah good point, fixing this.

legion2002: Fixed in https://github.com/tempoxyz/tempo/pull/3202/changes/130562a8b57bad53d22b0eeadefa44f0772f4758

crates/precompiles/src/account_keychain/mod.rs
legion2002: Added in https://github.com/tempoxyz/tempo/pull/3202/changes/130562a8b57bad53d22b0eeadefa44f0772f4758

crates/precompiles/src/account_keychain/mod.rs
legion2002: Added in https://github.com/tempoxyz/tempo/pull/3202/changes/130562a8b57bad53d22b0eeadefa44f0772f4758

crates/precompiles/src/account_keychain/mod.rs
legion2002: updated in https://github.com/tempoxyz/tempo/pull/3202/changes/130562a8b57bad53d22b0eeadefa44f0772f4758

crates/revm/src/handler.rs
```diff
@@ -674,8 +696,8 @@ where
         let cfg = &evm.inner.ctx.cfg;
         let journal = &mut evm.inner.ctx.journaled_state;
 
-        // Set tx.origin in the keychain's transient storage for spending limit checks.
-        // This must be done for ALL transactions so precompiles can access it.
+        // Keep tx.origin seeding here for downstream custom handlers that override
+        // `inspect_run` and call directly into validation/setup paths.
         StorageCtx::enter_evm(journal, block, cfg, tx, || {
             let mut keychain = AccountKeychain::new();
             keychain.set_tx_origin(tx.caller())
```
0xKitsune: Why do we duplicate this logic here rather than just calling `seed_tx_origin()` here instead of in `run()`? Is there a reason that we call this in run?

legion2002: Yeah was initially doing it according to your suggestion here.
https://github.com/tempoxyz/tempo/pull/3202#discussion_r2962648494

But it seems like this is not needed. The CI fails now because of a tempo-foundry bug, PR to fix it [here](https://github.com/tempoxyz/tempo-foundry/pull/344)

legion2002: addressed in this PR https://github.com/tempoxyz/tempo/pull/3202/changes/17cb132533e5660737ce9913169fe734743764d4

PR #3203, opened 2026-03-18
fix(precompiles): craft built-in policy types' data

0xrusowsky:
ref: `SIGP-185`

### DIFF
--- a/crates/precompiles/src/tip403_registry/mod.rs
+++ b/crates/precompiles/src/tip403_registry/mod.rs
@@ -12,12 +12,12 @@ pub use tempo_contracts::precompiles::{
     ITIP403Registry::{self, PolicyType},
     TIP403RegistryError, TIP403RegistryEvent,
 };
-use tempo_precompiles_macros::{contract, Storable};
+use tempo_precompiles_macros::{Storable, contract};
 
 use crate::{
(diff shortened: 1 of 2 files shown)


crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -884,6 +896,30 @@ mod tests {
         })
     }
 
+    #[test]
+    fn test_policy_data_returns_correct_type_for_builtin_policies() -> eyre::Result<()> {
```
fgimenez: would be good to have an equivalent pre-t2 `policy_data` test that documents the behaviour `test_policy_data_returns_correct_type_for_builtin_policies_pre_t2`, and rename this one to `test_policy_data_returns_correct_type_for_builtin_policies_t2`, wdyt?

0xrusowsky: good call, will do!

0xrusowsky: addressed in [`1d43341` (this PR)](https://github.com/tempoxyz/tempo/pull/3203/commits/1d43341bff50d34dfe8ec061acefafcc6f24d72c)

PR #3204, opened 2026-03-18
fix(precompile): check token paused in internal balance swaps

0xrusowsky:
ref: `QED-5`


crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -197,7 +197,14 @@ impl StablecoinDEX {
...
         amount: u128,
     ) -> Result<()> {
-        TIP20Token::from_address(token)?.ensure_transfer_authorized(user, self.address)?;
+        // Ensure that the token can be transferred
+        let tip20 = TIP20Token::from_address(token)?;
+        tip20.ensure_transfer_authorized(user, self.address)?;
+
+        // Ensure that the token is not paused (spec: T2+)
+        if self.storage.spec().is_t2() {
+            tip20.check_not_paused()?;
```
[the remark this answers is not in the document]
fgimenez: looks valid

fgimenez (commented):
i believe we should address https://github.com/tempoxyz/tempo/pull/3204#discussion_r2956372098 other than that looks good

crates/precompiles/src/stablecoin_dex/mod.rs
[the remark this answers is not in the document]
0xrusowsky: addressed in [`471d5f3` (this PR)](https://github.com/tempoxyz/tempo/pull/3204/commits/471d5f3d1a46d2cca08b6aabff9276b0a0d8ebdb)

i moved the validation to `fn validate_and_build_route()` because all swaps (direct and multi-hop) use it.

we don't need to worry about external interactions with paused tip20 tokens because those are caught by the built in TIP20 checks in `transfer`/`transfer_from`

fgimenez (approved):

crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -1383,6 +1389,13 @@ impl StablecoinDEX {
 
             let (base, quote) = {
                 let token_in_tip20 = TIP20Token::from_address(token_in)?;
+
+                // Ensure that the token is not paused (spec: T2+)
+                // Necessary because TIP20 transfer checks don't cover internal DEX balance updates
+                if self.storage.spec().is_t2() {
```
0xKitsune: Looks like this needs to be bumped to T3.

crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -1383,6 +1389,13 @@ impl StablecoinDEX {
...
+                    token_in_tip20.check_not_paused()?;
```
0xKitsune, line 1396: Only `token_in` is checked, the final `token_out` is not currently checked for paused status.

This should be fine for actual swaps since they should revert but quote functions will return valid quotes for routes ending in a paused token.

Not this is necessary to check, given actual swaps wont complete but flagging for visibility.

crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -1383,6 +1389,13 @@ impl StablecoinDEX {
 
             let (base, quote) = {
                 let token_in_tip20 = TIP20Token::from_address(token_in)?;
+
+                // Ensure that the token is not paused (spec: T2+)
```
0xKitsune: nit: comment should mention T3 or a subsequent fork.

crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -4949,4 +4963,130 @@ mod tests {
...
+                }
+
+                Ok::<_, eyre::Report>(())
+            })?;
+        }
+        Ok(())
+    }
+
+    #[test]
+    fn test_swap_paused_intermediate_token_allowed_pre_t2_blocked_on_t2() -> eyre::Result<()> {
```
0xKitsune: All tests should be updated to reference T3.

crates/precompiles/src/stablecoin_dex/mod.rs
0xrusowsky: already told to auditors that this is done for perf reasons (happy path), but we could add the redundant check if u want to fail early

crates/precompiles/src/stablecoin_dex/mod.rs
0xrusowsky: addressed in [`2c484be` (this PR)](https://github.com/tempoxyz/tempo/pull/3204/commits/2c484be389a3048bdaa9eeb60dd89f098e59667b)

crates/precompiles/src/stablecoin_dex/mod.rs
0xrusowsky: addressed in [`2c484be` (this PR)](https://github.com/tempoxyz/tempo/pull/3204/commits/2c484be389a3048bdaa9eeb60dd89f098e59667b)

crates/precompiles/src/stablecoin_dex/mod.rs
0xrusowsky: addressed in [`2c484be` (this PR)](https://github.com/tempoxyz/tempo/pull/3204/commits/2c484be389a3048bdaa9eeb60dd89f098e59667b)

crates/precompiles/src/stablecoin_dex/mod.rs
0xrusowsky: @0xKitsune is this g2g? or do u want the extra check to fail fast?

PR #3207, opened 2026-03-18
fix(precompiles): properly handle account creation with T2 gas and nonce logic

rakita:
## Summary

Rename `set_code` to `create_account` and add T2-specific gas deduction for account creation operations. On T2, accounts being created should deduct GACC gas and have their nonce incremented to properly reflect the create operation semantics.

## Changes

- Rename `set_code` to `create_account` across all storage provider implementations
- Add GACC (account creation) gas deduction when creating accounts on T2
- Bump nonce on account creation for T2 to ensure proper state tracking
- Update all callsites and test cases to use the new method name
- Update documentation strings to reflect the new semantics

## Files Modified

- `crates/precompiles/src/storage/mod.rs` - Trait definition
- `crates/precompiles/src/storage/evm.rs` - Main EVM implementation with T2 logic
- `crates/precompiles/src/storage/hashmap.rs` - HashMap provider implementation
- `crates/precompiles/src/storage/thread_local.rs` - Thread-local wrapper
- `crates/precompiles-macros/src/layout.rs` - Macro codegen
- `crates/revm/src/common.rs` - Read-only context implementation


jenpaff:
fixed as part of 1016

PR #3211, opened 2026-03-19
test(invariants): improve ValidatorConfigV2 coverage

grandizzy:
Improves the V2 validator invariant test coverage based on gap analysis between the Foundry invariant tests and the Rust fuzzer (tempo-fuzzer).

### Ed25519 signature fix
`_signAdd` and `_signRotate` were missing **length-prefixed** ingress/egress strings in the message hash. The Rust precompile hashes `chainId || contractAddr || validatorAddr || len(ingress) || ingress || len(egress) || egress || feeRecipient`, but the test was omitting the `len()` bytes. This caused **every** `addValidator` and `rotateValidator` call to revert with `InvalidSignature`, meaning all success-path invariant assertions were dead code.

### Ghost state fix for deactivated migration
`_ghostAddressInUse` was set to `true` unconditionally for all migrated validators. Deactivated V1 validators should not mark their address as in-use — the contract correctly allows reusing deactivated addresses. Changed to `_ghostAddressInUse[...] = v1Vals[idx].active`.

### Caller selection bias
Added `_selectOwnerOrRandom` (75% owner, 25% random) and `_selectDualAuthCaller` (50% owner, 25% validator, 25% random) helpers. Previously, handlers like `_deactivateValidator`, `_setIpAddresses`, `_rotateValidator`, `_transferOwnership`, etc. selected callers from `_selectPotentialValidator` — a 500-address pool that never included the owner or validator addresses, so every call reverted with `Unauthorized`.

### Invariant fixes
- **VALV2-18** (`_invariantAddressLookupCorrectness`): Was finding the "last occurrence" by array index. After rotation the active entry stays at the original (lower) index while the deactivated snapshot is appended — so the last entry is the wrong one. Fixed to find the **active** entry.
- **VALV2-24** (`_invariantMigrationIdentity`): Was asserting migrated slot pubkey matches V1 pubkey. After rotation the slot pubkey changes. Fixed to verify the original V1 pubkey still exists in V2 (as a deactivated snapshot) via `validatorByPublicKey`.

### Migration in setUp
Completes all 15 V1→V2 migrations and calls `initializeIfMigrated()` in `setUp()`, so all 1000 fuzzer calls exercise post-init operations instead of burning ~300 calls on migration.

### `afterInvariant` edge cases
Deactivates all validators post-campaign to exercise:
- `_addValidator` dupIP mode with no active validators
- `_deactivateValidator` on already-deactivated validators

### IPv6 address generation
`_generateIngress` / `_generateEgress` now produce 4 address types (~20% each): IPv4-mapped `::ffff:`, native IPv6 loopback `::1`, documentation range `2001:db8::`, and plain IPv4.

### Block advancement
New `handler_advanceBlock(delta)` rolls `block.number` forward by 1–100, making height-based invariants (VALV2-10) non-trivial.

### Inactive V1 validators in migration
setUp deactivates V1 validators at indices 3, 7, 11 so VALV2-25 (migration activity preservation) tests the deactivated path.

### `setFeeRecipient` handler
New `handler_setFeeRecipient` with dual-auth testing and init gate assertion.

### Fix spec `setFeeRecipient`
Spec allowed setting fees on non initialized validators 

## Coverage Results
With 5 runs × 1000 depth: **62/62 instrumented branches executed (100%)**, including all success paths for `addValidator`, `deactivateValidator`, `rotateValidator`, `setIpAddresses`, `setFeeRecipient`, `transferOwnership`, `transferValidatorOwnership`, and `setNextDkgCeremony`.

### DIFF
--- a/tips/ref-impls/test/invariants/ValidatorConfigV2.t.sol
+++ b/tips/ref-impls/test/invariants/ValidatorConfigV2.t.sol
@@ -1338,8 +1338,7 @@ contract ValidatorConfigV2InvariantTest is InvariantBaseTest {
             uint256 activeIdx = 0;
             for (uint256 j = 0; j < vals.length; j++) {
                 if (vals[j].validatorAddress == addr && vals[j].deactivatedAtHeight == 0) {
-                    hasActive = true;
-                    activeIdx = j;
+                    (hasActive, activeIdx) = (true, j);
                     break;
                 }
             }
@@ -1459,15 +1458,15 @@ contract ValidatorConfigV2InvariantTest is InvariantBaseTest {
 
             // If rotated, the original pubkey must exist in a deactivated snapshot.
             // Verify via pubkey lookup — the contract keeps all pubkeys forever.
-            IValidatorConfigV2.Validator memory looked =
+            IValidatorConfigV2.Validator memory snapshotVal =
                 validatorConfigV2.validatorByPublicKey(v1Vals[v1Idx].publicKey);
             assertEq(
-                looked.publicKey,
+                snapshotVal.publicKey,
                 v1Vals[v1Idx].publicKey,
                 "TEMPO-VALV2-24: Migrated pubkey must still exist in V2 (possibly as snapshot)"
             );
             assertTrue(
-                looked.deactivatedAtHeight != 0,
+                snapshotVal.deactivatedAtHeight != 0,
                 "TEMPO-VALV2-24: Rotated-out migrated pubkey must be in a deactivated snapshot"
             );
         }
@@ -1480,19 +1479,23 @@ contract ValidatorConfigV2InvariantTest is InvariantBaseTest {
     function afterInvariant() public {
         // Deactivate all active validators, track first deactivated index
         uint64 firstDeactivatedIdx = type(uint64).max;
-        for (uint256 i = 0; i < _ghostTotalCount; i++) {
-            // forge-lint: disable-next-line(unsafe-typecast)
-            uint64 idx = uint64(i);
-            if (_ghostDeactivatedAtHeight[idx] == 0) {
+        for (uint64 i = 0; i < _ghostTotalCount; i++) {
+            if (_ghostDeactivatedAtHeight[i] == 0) {
                 vm.startPrank(_ghostOwner);
-                try validatorConfigV2.deactivateValidator(idx) {
+                try validatorConfigV2.deactivateValidator(i) {
                     vm.stopPrank();
-                    _ghostDeactivatedAtHeight[idx] = uint64(block.number);
-                    delete _ghostAddressInUse[_ghostAddress[idx]];
-                    delete _ghostActiveIngressIpHashes[keccak256(bytes(_ghostIngress[idx]))];
-                    if (firstDeactivatedIdx == type(uint64).max) firstDeactivatedIdx = idx;
-                } catch {
+                    _ghostDeactivatedAtHeight[i] = uint64(block.number);
+                    delete _ghostAddressInUse[_ghostAddress[i]];
+                    delete _ghostActiveIngressIpHashes[keccak256(bytes(_ghostIngress[i]))];
+                    if (firstDeactivatedIdx == type(uint64).max) firstDeactivatedIdx = i;
+                } catch (bytes memory reason) {
                     vm.stopPrank();
+                    fail(
+                        string.concat(
+                            "TEMPO-VALV2-TEARDOWN: deactivateValidator reverted for active validator: ",
+                            vm.toString(reason)
+                        )
+                    );
                 }
             }
         }


tips/ref-impls/test/invariants/ValidatorConfigV2.t.sol
```diff
@@ -1195,49 +1312,76 @@ contract ValidatorConfigV2InvariantTest is InvariantBaseTest {
...
+            }
+            if (alreadyChecked) continue;
+
+            // Find the active entry for this address (if any)
+            bool hasActive = false;
+            uint256 activeIdx = 0;
+            for (uint256 j = 0; j < vals.length; j++) {
+                if (vals[j].validatorAddress == addr && vals[j].deactivatedAtHeight == 0) {
+                    hasActive = true;
+                    activeIdx = j;
```
howydev: ultra nit: we can do `(hasActive, activeIdx) = (true, j)`

tips/ref-impls/test/invariants/ValidatorConfigV2.t.sol
```diff
@@ -1309,11 +1453,64 @@ contract ValidatorConfigV2InvariantTest is InvariantBaseTest {
             // forge-lint: disable-next-line(unsafe-typecast)
             uint64 v2Idx = uint64(i);
             uint64 v1Idx = _ghostV2ToV1Index[v2Idx];
+
+            // If the slot hasn't been rotated, pubkey should still match directly
+            if (_ghostPubKey[v2Idx] == v1Vals[v1Idx].publicKey) continue;
+
+            // If rotated, the original pubkey must exist in a deactivated snapshot.
+            // Verify via pubkey lookup — the contract keeps all pubkeys forever.
+            IValidatorConfigV2.Validator memory looked =
```
howydev: nit: is there a better var name than `looked`?

tips/ref-impls/test/invariants/ValidatorConfigV2.t.sol
```diff
@@ -1309,11 +1453,64 @@ contract ValidatorConfigV2InvariantTest is InvariantBaseTest {
...
+    }
+
+    /// @notice Runs after invariant campaign to exercise edge cases unreachable during fuzzing.
+    /// @dev Covers:
+    ///   - _addValidator dupIP mode with no active validators
+    ///   - ValidatorAlreadyDeactivated revert on already-deactivated validator
+    function afterInvariant() public {
+        // Deactivate all active validators, track first deactivated index
+        uint64 firstDeactivatedIdx = type(uint64).max;
+        for (uint256 i = 0; i < _ghostTotalCount; i++) {
```
howydev: nit:
```suggestion
        for (uint64 i = 0; i < _ghostTotalCount; i++) {
```

howydev: then we can remove L1485

tips/ref-impls/test/invariants/ValidatorConfigV2.t.sol
```diff
@@ -1309,11 +1453,64 @@ contract ValidatorConfigV2InvariantTest is InvariantBaseTest {
...
+            if (_ghostDeactivatedAtHeight[idx] == 0) {
+                vm.startPrank(_ghostOwner);
+                try validatorConfigV2.deactivateValidator(idx) {
+                    vm.stopPrank();
+                    _ghostDeactivatedAtHeight[idx] = uint64(block.number);
+                    delete _ghostAddressInUse[_ghostAddress[idx]];
+                    delete _ghostActiveIngressIpHashes[keccak256(bytes(_ghostIngress[idx]))];
+                    if (firstDeactivatedIdx == type(uint64).max) firstDeactivatedIdx = idx;
+                } catch {
+                    vm.stopPrank();
```
howydev, line 1492: should we fail here? if `_ghostDeactivatedAtHeight` exists, we should be able to deactivate the validator right

howydev (approved):
Very nice, thanks for hardening the invariant tests!

Have some non-blocking nit comments

tips/ref-impls/test/invariants/ValidatorConfigV2.t.sol
grandizzy: changed in https://github.com/tempoxyz/tempo/pull/3211/changes/724587de8e86122ae48e35a668d6910af818e99f

tips/ref-impls/test/invariants/ValidatorConfigV2.t.sol
grandizzy: renamed to `snapshotVal` in https://github.com/tempoxyz/tempo/pull/3211/changes/724587de8e86122ae48e35a668d6910af818e99f

tips/ref-impls/test/invariants/ValidatorConfigV2.t.sol
grandizzy: changed in https://github.com/tempoxyz/tempo/pull/3211/changes/724587de8e86122ae48e35a668d6910af818e99f

tips/ref-impls/test/invariants/ValidatorConfigV2.t.sol
grandizzy: added fail in https://github.com/tempoxyz/tempo/pull/3211/changes/724587de8e86122ae48e35a668d6910af818e99f

PR #3193, opened 2026-03-18
docs(tip-1035): payment transaction classification

0xrusowsky:
Extracted from #3187.

Adds TIP-1035 specifying consensus rules for classifying transactions as payment lane eligible, including the calldata selector allowlist enforced at the T2 hardfork.

PR #3219, opened 2026-03-19
docs(tip-1035): Implicit Approval List — system_transfer_from only

decofe:
Defines an in-protocol Implicit Approval List of precompile addresses that may call `system_transfer_from` to pull TIP-20 tokens without requiring prior `approve()` calls.

**Scope**: The only behavioral change is that listed precompiles can call `system_transfer_from`, a precompile-only internal function (not part of the ABI) that skips allowance checks. All other TIP-20 semantics — `approve`, `permit`, `allowance`, `transferFrom` — are completely unchanged.

**Initial list**: TipFeeManager (`0xfeEC...`) + StablecoinDEX (`0xDEc0...`)

**Discoverability**: `AddressRegistry.isImplicitlyApproved(addr)` returns whether an address is on the list for the active hardfork.


tips/tip-1035.md
```diff
@@ -0,0 +1,121 @@
...
+
+1. Checks that `msg.sender` (the direct caller) is on the Implicit Approval List. If not, the call reverts.
+2. Enforces TIP-403 transfer policies via `ensure_transfer_authorized`.
+3. Enforces AccountKeychain spending limits via `check_and_update_spending_limit`.
+4. Debits `from` and credits `to`. Reverts with `InsufficientBalance` if `from` has insufficient balance.
+5. Emits a standard TIP-20 `Transfer(from, to, amount)` event.
+
+### Non-precompile contracts
+
+Non-precompile contracts on the Implicit Approval List (such as MPP Channels) are already deployed and call the standard `transferFrom`. For these contracts, TIP-20's `transferFrom` itself checks whether `msg.sender` is on the Implicit Approval List. If so, the allowance check and allowance decrement are skipped. All other checks (balance, transfer policies, spending limits, events) still apply.
```
danrobinson: It actually adds a spending limit check and update, right? It's not just that it doesn't skip it

tips/tip-1035.md
```diff
@@ -0,0 +1,121 @@
...
+2. **No DELEGATECALL**: The contract MUST NOT contain any `DELEGATECALL` opcodes. If a listed contract were to delegatecall into external code, that code would run with the listed contract's address as `address(this)`. A subsequent call to `transferFrom` on a TIP-20 token would have `msg.sender` equal to the listed contract's address, inheriting the implicit approval privilege. This would allow the external code to call `transferFrom` with an arbitrary `from` address and bypass the allowance check.
+3. **No proxy pattern**: The contract MUST NOT be an upgradeable proxy. An upgradeable proxy delegatecalls to its implementation, so a malicious upgrade would inherit the implicit approval privilege (same mechanism as above).
+4. **Immutable code**: For non-precompile contracts, the deployed bytecode must be immutable. There must be no mechanism (e.g., `SELFDESTRUCT` + redeployment, metamorphic contracts) by which the code at the listed address could change.
+
+For precompiles, criteria 2–4 are inherently satisfied: precompile code is part of the node implementation and is not executed via EVM opcodes.
```
danrobinson: Should point out that precompiles can be upgraded through hard forks of the protocol; security of such upgrades is the responsibility of the protocol.

tips/tip-1035.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Eligibility Criteria
+
+A contract MAY be added to the Implicit Approval List only if it satisfies all of the following:
+
+1. **Caller-only pulls**: Every reachable code path that pulls tokens via `transferFrom` or `system_transfer_from` SHOULD use `from == msg.sender` for the current call, or transfer with a non-replayable signed authorization from the transferor. This is a security guideline — other patterns could theoretically be secure, but should be vetted with significantly more scrutiny before a contract is added to the list.
+2. **No DELEGATECALL**: The contract MUST NOT contain any `DELEGATECALL` opcodes. If a listed contract were to delegatecall into external code, that code would run with the listed contract's address as `address(this)`. A subsequent call to `transferFrom` on a TIP-20 token would have `msg.sender` equal to the listed contract's address, inheriting the implicit approval privilege. This would allow the external code to call `transferFrom` with an arbitrary `from` address and bypass the allowance check.
```
danrobinson: I think probably this is a SHOULD as well. If it DELEGATECALLS, then the contract it delegates to MUST be vetted using the same rules.

tips/tip-1035.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Integrator Guidance
+
+Implicitly-approved precompiles (TipFeeManager, StablecoinDEX) are part of the protocol and integrators may treat these as stable unless superseded by a future TIP.
+
+Implicitly-approved deployed contracts (such as TempoStreamChannel) carry no such guarantee. The Implicit Approval List may be amended at any hardfork, and a deployed contract may be removed from the list — for example, if it is replaced by a new deployment or promoted to a precompile. Integrators MUST NOT assume that a deployed contract will remain implicitly approved.
+
+Non-upgradeable integrations that interact with an implicitly-approved deployed contract SHOULD continue to perform their existing `approve()`-before-use flow, rather than rely on a one-time approval, because `approve()` does not grant allowance while the spender is listed. This ensures the integration continues to work if the contract is removed in a future hardfork. Upgradeable integrations can omit the `approve()` call and adapt if the list changes.
```
0xrusowsky: @danrobinson i feel like this is bad UX for integrators, and we should help them a little bit more.

imo we should leverage the `TIP20Registry` precompile introduced by TIP-1022 (virtual addresses). Such registry is the perfect candidate for a helper fn:
```solidity
function isImplicitlyApproved(address addr) public pure returns (bool);
```

this makes integrator's life way easier:
- is hardfork aware, so is the canonical truth
- from a protocol POC it doesn't require any state + would also help us keep the logic clean, so that we implicitly approve based on what this hardfork-aware fn tells us

decofe (commented):
## TIP-1035 Numbering Conflict

**Blocker**: This PR and [PR #3193](https://github.com/tempoxyz/tempo/pull/3193) both use **TIP-1035**. One is "Payment Transaction Classification" and the other is "Implicit Approval List." This will cause real ambiguity in hardfork manifests, client feature flags, test names, and future cross-references.

**Action**: Renumber one of them before merge.

decofe (commented):
## TIP-1035 (Implicit Approval List) Review

### protocolVersion → T5

1. **`approve`/`permit` semantic changes are the highest-risk part of this TIP.** The TIP redefines core ERC-20 semantics for listed spenders:
   - `approve` succeeds but does nothing
   - `permit` validates signature, increments nonce, but does nothing to allowance
   - `allowance()` returns stale data

   Wallets and integrators universally assume `approve` has the effect of setting an allowance. Returning success without actually doing anything will silently break approval-tracking UIs, and `allowance()` returning stale data is a footgun for any contract that checks allowance before calling `transferFrom`.

   **Suggestion:** Consider making `approve` on a listed spender revert with a descriptive error (e.g., `ImplicitlyApproved()`) rather than silently succeeding. This is safer for integrators — they learn immediately that approval is unnecessary, rather than silently succeeding and then being confused when allowance reads don't match.

2. **`system_transfer_from` does not restrict `from` parameter.** The TIP acknowledges this: "a listed precompile could pass any address, not just `msg.sender`." This means the security model entirely depends on the correctness of each listed precompile's code. If a precompile bug allows arbitrary `from`, it's an instant drain of any token. The "Eligibility Guidelines" section is advisory, not enforced. Consider:
   - Adding `from == msg.sender` as a hard constraint in `system_transfer_from` itself
   - Or explicitly documenting why this is intentionally left unconstrained (e.g., FeeManager needs `from = tx.origin` not `msg.sender`)

3. **List permanence ("MUST NOT be removed") is a strong commitment.** If a listed precompile is later found to have a bug in its token-pull logic, the protocol cannot revoke its implicit approval privilege without violating this invariant. Consider softening to "SHOULD NOT be removed" or adding an explicit escape hatch via governance/hardfork.

4. **Pre-existing approval + spending limit interaction.** The TIP says pre-existing allowances become "inert" but spending limits already deducted at `approve` time are not refunded. This creates an asymmetry: users who approved before the fork lose spending limit capacity with no recourse. While noted as "rare," it should be documented as a known migration artifact.

### Suggested changes

Add to frontmatter:
```yaml
protocolVersion: T5
```

tips/tip-1035.md
```diff
@@ -0,0 +1,130 @@
...
+
+# Specification
+
+## Implicit Approval List
+
+The protocol maintains an explicit set of addresses called the Implicit Approval List. For any spender address on this list:
+
+1. The precompile may transfer TIP-20 tokens from `msg.sender` without a prior `approve()` call (see [Implementation](#implementation)).
+2. `approve(spender, amount)` where `spender` is on the list: the call succeeds but does not affect allowance state, does not emit an `Approval` event, and does not deduct AccountKeychain spending limits. This applies regardless of `amount`.
+3. `permit()` where `spender` is on the list: the call MUST still enforce all normal validity checks (including deadline and signature validation). If validation fails, the call MUST revert and MUST NOT change the nonce or any other state. If validation succeeds, the owner's nonce increments, but allowance state is unchanged and no `Approval` event is emitted.
```
samczsun: What are the tradeoffs to performing the full permit validation? I'd like to see the rationale explained in a security considerations section

tips/tip-1035.md
```diff
@@ -0,0 +1,130 @@
...
+4. Token pulls through listed precompiles still enforce balance checks, TIP-403 transfer policies, AccountKeychain spending limits, and `Transfer` event emission. The only check bypassed is the allowance check.
+
+## Implementation
+
+Implicitly-approved precompiles MUST use `system_transfer_from()` to pull tokens. The TipFeeManager already uses `system_transfer_from` today, so this TIP does not change its behavior. `system_transfer_from` is a privileged function on the TIP-20 precompile that performs a transfer without checking the allowance. It:
```
samczsun: Why bother describing `system_transfer_from` here?

tips/tip-1035.md
```diff
@@ -0,0 +1,130 @@
...
+
+1. Checks that `msg.sender` (the direct caller) is on the Implicit Approval List. If not, the call reverts.
+2. Enforces TIP-403 transfer policies via `ensure_transfer_authorized`.
+3. Enforces AccountKeychain spending limits via `check_and_update_spending_limit`.
+4. Debits `from` and credits `to`. Reverts with `InsufficientBalance` if `from` has insufficient balance.
+5. Emits a standard TIP-20 `Transfer(from, to, amount)` event.
+
+### Security model
+
+`system_transfer_from` does not restrict the `from` parameter — a listed precompile could pass any address, not just `msg.sender`. The caller-only-pulls property is enforced by the listed precompile's own code, not by the implicit approval system. The security of this TIP depends on strict admission criteria for the list. The implicit approval system itself provides no protection against a listed precompile that passes a `from` address other than `msg.sender`. Since precompile code is part of the node implementation, the security of each listed precompile is the responsibility of the protocol.
```
samczsun: This paragraph says nothing about the actual security model and also doesn't explain why `system_transfer_from` doesn't just assert `from == msg.sender`?

tips/tip-1035.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Initial List
+
+| Address | Contract | Rationale |
+|---------|----------|-----------|
+| `0xfeEC000000000000000000000000000000000000` | TipFeeManager (FeeAMM) | Already uses `system_transfer_from`; pulls from `msg.sender` via `rebalanceSwap` and `mint`, and from `tx.origin` to collect transaction fees |
```
samczsun: I thought this TIP only supported `msg.sender` pulls?

tips/tip-1035.md
```diff
@@ -0,0 +1,130 @@
...
+| `0xDEc0000000000000000000000000000000000000` | StablecoinDEX | Only pulls from `msg.sender` via `place`, `placeFlip`, `swapExactAmountIn`, `swapExactAmountOut` |
+
+## Pre-existing Approvals
+
+Users who called `approve()` on a listed address before the fork activation may have stale allowance state in storage. While a spender remains on the Implicit Approval List, any pre-existing allowance for that spender is ignored and has no effect on token-pull behavior. Stale slots are inert while the spender remains listed.
+
+`allowance(owner, spender)` continues to return the raw stored value and may therefore be stale for listed spenders. While a spender is listed, effective token-pull authorization is determined by list membership and the listed contract's behavior, not by the allowance mapping.
```
samczsun: Why not also return `type(uint).max` for listed addresses?

tips/tip-1035.md
```diff
@@ -0,0 +1,130 @@
...
+
+The following are security guidelines for adding a precompile to the Implicit Approval List. Precompiles that do not satisfy these guidelines may still be added, but the amendment TIP MUST include an explicit safety argument explaining why the alternative pattern is secure.
+
+1. **Caller-only pulls**: Every reachable code path that pulls tokens via `system_transfer_from` SHOULD use `from == msg.sender` for the current call, or transfer with a non-replayable signed authorization from the transferor. Other patterns could theoretically be secure, but should be vetted with significantly more scrutiny.
+
+Since precompile code is part of the node implementation and is not executed via EVM opcodes, concerns like DELEGATECALL, proxy patterns, and code mutability do not apply. Precompiles can be upgraded through hardforks of the protocol; the security of such upgrades is the responsibility of the protocol.
+
+## Integrator Guidance
+
+Precompiles on the Implicit Approval List MUST NOT be removed. Contracts MAY depend on implicitly-approved precompiles never requiring approvals.
```
samczsun: Why?

tips/tip-1035.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Future Amendments
+
+Additional precompiles may be added to the Implicit Approval List via future TIPs that reference and amend this one. Each addition should include a safety argument demonstrating that the precompile satisfies the eligibility guidelines above.
+
+---
+
+# Alternatives Considered
+
+## Default-to-Unlimited Allowance
```
samczsun: Maybe we should just default to unlimited allowances with no option to opt out. This is basically what we want anyways, and there's no use case to opting out

tips/tip-1035.md
0xrusowsky: my old cmnt is still unaddressed after the refactor, so i'll post it again here, as it was targeted towards integrator guidance:

"""
@danrobinson i feel like this is bad UX for integrators, and we should help them more.

imo we should leverage the `AddressRegistry` precompile introduced by TIP-1022 (virtual addresses). Such registry is the perfect candidate for a helper fn:
```solidity
function isImplicitlyApproved(address addr) public pure returns (bool);
```

this makes integrator's life way easier:
- is hardfork aware, so is the canonical truth
- from a protocol POC it doesn't require any state + would also help us keep the logic clean, so that we implicitly approve based on what this hardfork-aware fn tells us
"""

tips/tip-1035.md
legion2002: may break indexers if we implicitly change the allowance for listed addresses, without a corresponding event.

tips/tip-1035.md
```diff
@@ -0,0 +1,130 @@
...
+
+An alternative design would keep standard ERC-20 allowance semantics but default the allowance for listed spenders to `type(uint256).max` (unlimited). `approve()` would still write to storage and `transferFrom` would still check the allowance — users would simply never need to explicitly approve because the default is already unlimited. Users could call `approve(spender, 0)` to revoke if they wanted to opt out.
+
+### Sentinel encoding
+
+To distinguish "never set" (default unlimited) from "explicitly set to 0" (revoked) without an extra SLOAD on every read, a sentinel value would be used. The raw allowance slot encodes as:
```
legion2002: why is it necessary to distinguish "never set" from "explicitly set to 0" here?
I would assume we'd basically force all users to always have infinite allowance for these addresses with no option to opt out?

tips/tip-1035.md
```diff
@@ -0,0 +1,130 @@
...
+
+**Advantages:**
+
+- `approve()`, `allowance()`, and `permit()` behave as standard ERC-20. Wallets and integrators do not need to special-case listed spenders.
+- Users can revoke by calling `approve(spender, 0)`, providing an explicit opt-out mechanism.
+
+**Disadvantages:**
+
+- **Gas overhead on every transaction.** Every token pull by a listed precompile adds an allowance SLOAD (~2,100 gas cold). The current design (`system_transfer_from`) skips this entirely. For the FeeManager, this cost is paid on every transaction in the network.
+- **Potential SSTORE on every transaction.** If a user sets a finite (non-max) allowance, every fee collection or swap decrements it, adding a warm SSTORE (~2,900 gas). Users who set finite allowances would also need to periodically re-approve, reintroducing the UX friction this TIP aims to eliminate.
```
legion2002: can't we block a user from setting a finite allowance? Or ignore the finite allowance if it has already been set.

tips/tip-1035.md
legion2002: do we need to add permit support here? 
Do we have any existing precompiles that would need permit support for system transfers?

legion2002 (commented):
It would make more sense to frame this TIP as a new precompile transfer type, instead of an update to existing approval semantics

The proposed solution has very little to do with ERC20 approvals, because it doesn't emit the events, or touch the approval balance slots.

We can just call this a new transfer type (`system_transfer_from`), only accessible to trusted precompiles.

0xrusowsky:
> It would make more sense to frame this TIP as a new precompile transfer type, instead of an update to existing approval semantics
> 
> The proposed solution has very little to do with ERC20 approvals, because it doesn't emit the events, or touch the approval balance slots.
> 
> We can just call this a new transfer type (`system_transfer_from`), only accessible to trusted precompiles.

note that if we went this route, then u couldn't be able to use it on MPP's escrow contract (until it becomes a precompile)

tips/tip-1035.md
danrobinson: The reason to allow this one-way ratchet rule would be so that non-upgradeable contracts don't have to write in unnecessary approval steps or checks like this. But if you guys would prefer, I'm OK saying instead that non-upgradeable contracts MUST NOT depend on implicit approval lasting forever, and instead should call approve() to be safe.

I think it's a good idea to add an isImplicitlyApproved function to the AddressRegistry, but IMO the best pattern in this case would just be to call approve(), since it is a no-op if the contract is implicitly approved.

@decofe can you update the TIP to add isImplicitlyApproved to the AddressRegistry, and to consistently change the rule to basically say that contracts that are implicitly approved may have that status change in the future, and that non-upgradeable contracts MUST NOT depend on an implicitly approved contract remaining implicitly approved forever? Contracts or integrations MAY call approve() for implicitly approved contracts in order to continue to work properly even if that implicit approval changes.

tips/tip-1035.md
danrobinson: This section is under "alternative proposals" and is discussing the variant where we give users the ability to opt out. I don't like it as much

tips/tip-1035.md
danrobinson: You just mean that we should update the main proposal to have allowance return uint256 max instead of 0 (or whatever allowance the address had already given to the implicitly approved contract)? I think that makes sense

samczsun (approved):

tips/tip-1035.md
```diff
@@ -0,0 +1,105 @@
...
+```
+
+This function returns `true` if and only if `addr` is on the Implicit Approval List for the active hardfork. Before TIP-1035 activates, it returns `false` for all addresses.
+
+## Initial List
+
+| Address | Contract | Rationale |
+|---------|----------|-----------|
+| `0xfeEC000000000000000000000000000000000000` | TipFeeManager (FeeAMM) | Already relies on `system_transfer_from` for fee collection and liquidity operations |
+| `0xDEc0000000000000000000000000000000000000` | StablecoinDEX | Removes redundant approvals for DEX order placement and swap flows |
```
legion2002, line 75: If we add it to T5, we should also add enshrined MPP here.

tips/tip-1035.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Security Model
+
+`system_transfer_from` does not restrict the `from` parameter — a listed precompile could pass any address, not just `msg.sender`. The caller-only-pulls property is enforced by the listed precompile's own code, not by the implicit approval system. The security of this TIP therefore depends on strict admission criteria for the list and on review of each listed precompile's token-pull logic.
```
legion2002, line 79: maybe we can refactor our precompile setup, so that we can pass the msg.sender into `system_transfer_from` so that we can guarantee that the "caller-only pulls" invariant is never violated

legion2002 (approved):

samczsun (approved):

PR #3220, opened 2026-03-19
fix: check 403 policy updates against fee manager

0xKitsune:
Previously pool eviction only checked sender-side policies. When the fee manager was blacklisted or removed from a whitelist as a recipient, affected transactions remained in the pool despite being invalid at execution. This updates the eviction logic to account for this case.

### DIFF
--- a/crates/transaction-pool/src/tempo_pool.rs
+++ b/crates/transaction-pool/src/tempo_pool.rs
@@ -34,8 +34,11 @@ use tempo_chainspec::{
     hardfork::{TempoHardfork, TempoHardforks},
 };
 use tempo_precompiles::{
-    TIP_FEE_MANAGER_ADDRESS, account_keychain::AccountKeychain,
-    error::Result as TempoPrecompileResult, nonce::NonceManager, storage::Handler,
+    TIP_FEE_MANAGER_ADDRESS,
+    account_keychain::AccountKeychain,
+    error::Result as TempoPrecompileResult,
+    nonce::NonceManager,
+    storage::Handler,
     tip20::TIP20Token,
     tip403_registry::{REJECT_ALL_POLICY_ID, TIP403Registry},
 };


crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -1138,6 +1181,41 @@ fn get_sender_policy_ids(
...
+fn get_recipient_policy_ids(
+    provider: &mut impl StateProvider,
+    fee_token: Address,
+) -> Option<Vec<u64>> {
+    let spec = TempoHardfork::default();
+    provider.with_read_only_storage_ctx(spec, || {
+        let policy_id = TIP20Token::from_address(fee_token)
+            .and_then(|t| t.transfer_policy_id())
+            .ok()
+            .filter(|&id| id != 0)?;
```
mattsse: why 0?

dont we have constants for policies now?

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -1138,6 +1181,41 @@ fn get_sender_policy_ids(
...
+/// Returns the set of policy IDs that can affect recipient authorization for a token.
+///
+/// For simple (non-compound) policies, the transfer policy applies symmetrically to both
+/// sender and recipient, so the set contains just the policy ID. For compound policies
+/// (TIP-1015) it contains both the compound root and the recipient sub-policy, since
+/// fee transfer authorization checks the fee manager via `AuthRole::Recipient`.
+///
+/// Unlike `get_sender_policy_ids` this is uncached — it's only called on the rare path
+/// where the fee manager itself is blacklisted or un-whitelisted.
+fn get_recipient_policy_ids(
```
mattsse, line 1237: i think this fn is out of order, but so is get_sender_policy_ids -.-

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -1138,6 +1181,41 @@ fn get_sender_policy_ids(
...
+    provider: &mut impl StateProvider,
+    fee_token: Address,
+) -> Option<Vec<u64>> {
+    let spec = TempoHardfork::default();
```
mattsse: why is default here okay?

couldnt this end up being wrong if a hf activates?

crates/transaction-pool/src/tempo_pool.rs
0xKitsune: I pushed an update to remove `0` and use the const for the blacklist policy Id.


Ref: https://github.com/tempoxyz/tempo/pull/3220/commits/59f76124df8c449711c6ddb7ac4421a979c11e86

crates/transaction-pool/src/tempo_pool.rs
0xKitsune: Updated to move these in https://github.com/tempoxyz/tempo/pull/3220/changes/74e43926cf4cc36780e3a86e135b38db5e800aee

crates/transaction-pool/src/tempo_pool.rs
0xKitsune: At the moment, the TIP20/TIP403 reads don't check `spec()`, so default() works. But yeah if a future hardfork adds spec gated logic to these then using `default()` would silently return wrong results. I updated to derive the spec from the timestamp.

PR #3227, opened 2026-03-19
docs: add TIP-1037 FeeAMM TIP-403 policy exemptions

decofe:
Adds TIP-1037, which exempts the FeeAMM address from TIP-403 checks during fee collection while preserving full enforcement on public AMM operations (mint, burn, rebalanceSwap, distributeFees). Also adds a new authorization gate on `mint` requiring the caller to be authorized on the `userToken`.

Prompted by: dan

### DIFF
--- a/tips/tip-1037.md
+++ b/tips/tip-1037.md
@@ -11,7 +11,7 @@ related: TIP-403, TIP-1015, TIP-20
 
 ## Abstract
 
-This TIP modifies how TIP-403 transfer policies interact with the FeeAMM/FeeManager precompile. The FeeManager address is exempted from TIP-403 authorization checks during the protocol fee collection path (`collectFeePreTx`, `transferFeePreTx`, `executeFeeSwap`, `transferFeePostTx`), while full TIP-403 enforcement is preserved on all public AMM operations (`mint`, `burn`, `rebalanceSwap`, `distributeFees`). Additionally, `mint` is extended with a new authorization check: the caller must be authorized as both sender and recipient on the `userToken`, even when only depositing `validatorToken`.
+This TIP modifies how TIP-403 transfer policies interact with the FeeAMM/FeeManager precompile. The FeeManager address is exempted from TIP-403 authorization checks during the protocol fee collection path (`collectFeePreTx`, `transferFeePreTx`, `executeFeeSwap`, `transferFeePostTx`), while full TIP-403 enforcement is preserved on all public AMM operations (`mint`, `burn`, `rebalanceSwap`, `distributeFees`). Additionally, `mint` is extended with a new authorization check that treats the operation as if the caller were transferring both tokens to the pool: the caller must be authorized as a sender on the `userToken` and the FeeManager must be an authorized recipient, even when only `validatorToken` is deposited.
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


tips/tip-1037.md
```diff
@@ -0,0 +1,195 @@
...
+
+# TIP-1037: FeeAMM TIP-403 Policy Exemptions
+
+## Abstract
+
+This TIP modifies how TIP-403 transfer policies interact with the FeeAMM/FeeManager precompile. The FeeManager address is exempted from TIP-403 authorization checks during the protocol fee collection path (`collectFeePreTx`, `transferFeePreTx`, `executeFeeSwap`, `transferFeePostTx`), while full TIP-403 enforcement is preserved on all public AMM operations (`mint`, `burn`, `rebalanceSwap`, `distributeFees`). Additionally, `mint` is extended with a new authorization check: the caller must be authorized as both sender and recipient on the `userToken`, even when only depositing `validatorToken`.
+
+## Motivation
+
+Currently, TIP-403 policy checks apply uniformly to all participants in FeeManager transfers, including the FeeManager address itself. This creates unnecessary friction for token issuers:
```
danrobinson: Rewrite this section, this isn't the motivation. The explicit whitelisting of the FeeAMM IS still necessary after this change, because otherwise nobody can mint.

The main motivation is that 403 checks during the transaction flow complicate block building and impose costs on every transaction without meaningful benefit (since the check on whether the FeeManager is allowed can be done at the time of the mint).

tips/tip-1037.md
```diff
@@ -0,0 +1,195 @@
...
+
+At the same time, the FeeManager address is shared with the public AMM surface (`mint`, `burn`, `rebalanceSwap`). A blanket exemption would allow the AMM to become an alternate transfer/swap path around TIP-403 policies. Token issuers who intentionally exclude trading contracts from their whitelist would find their tokens tradeable through the fee AMM.
+
+This TIP threads the needle: fee collection works without requiring the FeeManager to be whitelisted, while issuers retain full control over whether their token participates in AMM liquidity and trading.
+
+## Assumptions
+
+- The FeeManager address (`0xfeec000000000000000000000000000000000000`) is a protocol-level singleton that cannot be controlled by any external account.
+- `transferFeePreTx` and `transferFeePostTx` are only callable by the protocol (`msg.sender == address(0)`), so exempting the FeeManager from TIP-403 on these paths does not create a user-exploitable bypass.
```
danrobinson: I would remove this paragraph, it's sort of not really accurate since anyone can cause those functions to be called by using a feeToken in a transaction.

tips/tip-1037.md
```diff
@@ -0,0 +1,195 @@
...
+            || !TIP403_REGISTRY.isAuthorizedRecipient(userTokenPolicyId, msg.sender)
+    ) {
+        revert ITIP20.PolicyForbids();
+    }
+
+    // ... rest of mint logic (unchanged)
+}
+```
+
+Note: There is an inherent TOCTOU (time-of-check-time-of-use) gap — a policy may change between when the LP mints and when the pool is used for fee swaps or rebalancing. This is a best-effort check and is consistent with how TIP-403 operates elsewhere (e.g., a token holder may be blacklisted after receiving tokens).
```
danrobinson, line 132: Can say something like "Additionally, since rebalanceSwap checks the relevant policies, a token can prevent the liquidity on the FeeAMM from being refreshed, which means it will eventually be exhausting. This should give token admins an adequate tool for disabling the FeeAMM for their token even if someone manages to create liquidity on the pool before the admin sets that policy."

PR #3228, opened 2026-03-19
feat: ERC-1271 support for TempoStreamChannel

Fbartoli:
## Summary

- Replace `ECDSA.recoverCalldata` with `SignatureCheckerLib.isValidSignatureNowCalldata` in `settle()` and `close()` to support smart contract signers (ERC-1271)
- Extract shared `_verifyVoucher()` helper to reduce code duplication
- Add `MockERC1271Signer` test mock and 5 new tests

## Motivation

Smart contracts (ERC4626 vaults, multisigs, AA wallets) cannot currently be `authorizedSigner` or act as payers that sign vouchers on payment channels. This blocks use cases like:

- **Lending vaults** that open channels on behalf of AI agents (vault is payer, admin key signs vouchers via ERC-1271)
- **Multisig wallets** that authorize streaming payments
- **Account abstraction wallets** that use contract-based signatures

## Change

One function swap — `SignatureCheckerLib` from solady (already a dependency):

```diff
- address signer = ECDSA.recoverCalldata(digest, signature);
- if (signer != expectedSigner) revert InvalidSignature();

+ if (!SignatureCheckerLib.isValidSignatureNowCalldata(expectedSigner, digest, signature))
+     revert InvalidSignature();
```

`SignatureCheckerLib.isValidSignatureNowCalldata`:
1. Tries `ecrecover` first (EOA — identical to current behavior)
2. Falls back to `IERC1271.isValidSignature()` (contract signers)

**Fully backward compatible.** Every existing EOA signature works exactly as before.

## Tests

5 new tests added to `TempoStreamChannel.t.sol`:

| Test | What it proves |
|------|---------------|
| `test_erc1271_contractPayer_eoaSigner` | Contract pays, EOA signs (vault + admin key) |
| `test_erc1271_contractAsSigner` | Contract is authorizedSigner, delegates to EOA via isValidSignature |
| `test_erc1271_rejectsWrongKey` | Wrong EOA key rejected by contract's isValidSignature |
| `test_erc1271_closeWithContractSigner` | Close with partial settle + final voucher via ERC-1271 |
| `test_erc1271_existingEOATestsStillPass` | Backward compatibility — existing EOA flow unchanged |

## Context

Built during the Tempo hackathon while building an [Agent Credit Protocol](https://github.com/Fbartoli/tempo-hack) — an ERC4626 vault that lends USDC.e to AI agents for MPP service consumption. The vault needed to open payment channels as payer, which required ERC-1271 support on the escrow.

### DIFF
--- a/tips/ref-impls/src/TempoStreamChannel.sol
+++ b/tips/ref-impls/src/TempoStreamChannel.sol
@@ -5,6 +5,7 @@ import { TempoUtilities } from "./TempoUtilities.sol";
 import { ITIP20 } from "./interfaces/ITIP20.sol";
 import { ITempoStreamChannel } from "./interfaces/ITempoStreamChannel.sol";
 import { ECDSA } from "solady/utils/ECDSA.sol";
+import { SignatureCheckerLib } from "solady/utils/SignatureCheckerLib.sol";
 import { EIP712 } from "solady/utils/EIP712.sol";
 
 /**
@@ -129,16 +130,7 @@ contract TempoStreamChannel is ITempoStreamChannel, EIP712 {
             revert AmountNotIncreasing();
         }
 
-        bytes32 structHash = keccak256(abi.encode(VOUCHER_TYPEHASH, channelId, cumulativeAmount));
-        bytes32 digest = _hashTypedData(structHash);
-        address signer = ECDSA.recoverCalldata(digest, signature);
-
-        address expectedSigner =
-            channel.authorizedSigner != address(0) ? channel.authorizedSigner : channel.payer;
-
-        if (signer != expectedSigner) {
-            revert InvalidSignature();
-        }
+        _verifyVoucher(channel, channelId, cumulativeAmount, signature);
 
         uint128 delta = cumulativeAmount - channel.settled;
         channel.settled = cumulativeAmount;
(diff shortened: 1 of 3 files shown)


tips/ref-impls/src/TempoStreamChannel.sol
```diff
@@ -129,16 +130,7 @@ contract TempoStreamChannel is ITempoStreamChannel, EIP712 {
...
-        bytes32 digest = _hashTypedData(structHash);
-        address signer = ECDSA.recoverCalldata(digest, signature);
-
-        address expectedSigner =
-            channel.authorizedSigner != address(0) ? channel.authorizedSigner : channel.payer;
-
-        if (signer != expectedSigner) {
-            revert InvalidSignature();
-        }
+        _verifyVoucher(channel, channelId, cumulativeAmount, signature);
```
legion2002, line 133: It's dangerous to have a 1271 stateful voucher signer, because they could  revoke their signature before you can call settle.
Effectively blocking the payment.

I think a better way to do this would be using protocol enforced stateless signature schemes. We have a WIP progress TIP here - https://github.com/tempoxyz/tempo/pull/2633

Lmk if that would solve for your usecase.

Fbartoli: @legion2002 Thanks for your feedback, I will close and review my idea. I wanted an erc4626 vault (or any smart contracts) to be able to open sessions but I am not sure this is the best idea to do so now.

PR #3250, opened 2026-03-20
chore: failed-closed account keychain

0xrusowsky:
this PR hardens `fn ensure_admin_caller` to ensure that the `tx.origin` has been set

### DIFF
--- a/crates/precompiles/src/account_keychain/mod.rs
+++ b/crates/precompiles/src/account_keychain/mod.rs
@@ -388,9 +388,8 @@ impl AccountKeychain {
     /// The T2 check prevents transaction-global root-key status from being reused by
     /// intermediate contracts (confused-deputy self-administration).
     ///
-    /// `tx_origin` is seeded by the handler before validation/execution on in-repo flows.
-    /// The zero-origin fallback here exists for out-of-tree integrations that bypass
-    /// handler setup.
+    /// `tx_origin` is seeded by the handler before validation/execution.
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -388,17 +388,16 @@ impl AccountKeychain {
...
+    /// If origin is not seeded (zero), admin ops are rejected.
     fn ensure_admin_caller(&self, msg_sender: Address) -> Result<()> {
         if !self.transaction_key.t_read()?.is_zero() {
             return Err(AccountKeychainError::unauthorized_caller().into());
         }
 
         if self.storage.spec().is_t2() {
             let tx_origin = self.tx_origin.t_read()?;
-            if tx_origin != Address::ZERO && tx_origin != msg_sender {
+            if tx_origin.is_zero() || tx_origin != msg_sender {
```
legion2002, line 400: Does this break system txs?

0xrusowsky: i think they (will) never touch access keys, so it should be fine

wdyt @0xkitsune?

fgimenez: > i think they (will) never touch access keys, so it should be fine

agree, currently system txs never call `authorize_key`, `revoke_key` or `update_spending_liimt`, they nly process subblock metadata. if later we add a new system tx that tries to do admin keychain ops these changes would make it fail until origin is explicitly set

PR #3253, opened 2026-03-20
fix: update fee payer fallback

0xKitsune:
Restores sender-paid fallback in `compat.rs` when fee-payer signature recovery fails to fix `eth_estimateGas` functionality.

### DIFF
--- a/crates/transaction-pool/src/tempo_pool.rs
+++ b/crates/transaction-pool/src/tempo_pool.rs
@@ -257,14 +257,24 @@ where
             }
 
             // Check 2: Spending limit updates
-            // Only evict if the transaction's fee token matches the token whose limit changed.
+            // Only evict sender-paid keychain txs whose fee token matches the changed limit.
+            // Sponsored txs (fee payer != sender) are not subject to the sender's spending
+            // limit, so evicting them here would be a false positive.
             if !updates.spending_limit_changes.is_empty()
                 && let Some(ref subject) = keychain_subject
                 && subject.matches_spending_limit_update(&updates.spending_limit_changes)
             {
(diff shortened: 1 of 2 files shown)


crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -257,30 +257,49 @@ where
...
             {
-                to_remove.push(*tx.hash());
-                spending_limit_count += 1;
-                continue;
+                let sender = *tx.transaction.sender_ref();
+                let sender_paid = tx
+                    .transaction
+                    .inner()
+                    .fee_payer(sender)
+                    .map_or(true, |fee_payer| fee_payer == sender);
```
fgimenez, line 272: these new lines are duplicated in both 2 and 2b checks, maye be could add a `is_sender_paid` helper to refactor

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -257,30 +257,49 @@ where
...
+            // Sponsored txs (fee payer != sender) are not subject to the sender's spending
+            // limit, so evicting them here would be a false positive.
             if !updates.spending_limit_changes.is_empty()
                 && let Some(ref subject) = keychain_subject
                 && subject.matches_spending_limit_update(&updates.spending_limit_changes)
             {
-                to_remove.push(*tx.hash());
-                spending_limit_count += 1;
-                continue;
+                let sender = *tx.transaction.sender_ref();
```
fgimenez, line 267: in other checks we have `tx.transaction.sender()` instead of this deref of a borrow, given that `Address` is `Copy` both work, but why using this instead? should we align?

PR #3254, opened 2026-03-20
feat: tip-1031 implementation. consensus context in the block header

hamdiallam:
* When activated, consensus context must be set.

* life timestamp/timestamp_millis_part outside of `TempoPayloadAttributes` to avoid re-constructing the timestamp

* Since Reth's `Compact` does not support trailing options past the last  `inner: Header` field so we manually implement the Compact for TempoHeader.


crates/primitives/src/header.rs
```diff
@@ -49,6 +76,68 @@ impl AsRef<Self> for TempoHeader {
     }
 }
 
+impl Encodable for TempoHeader {
```
hamdiallam: Requires an update to `Compat`.

Manually encoding to unblock for now

crates/primitives/src/header.rs
```diff
@@ -28,6 +48,13 @@ pub struct TempoHeader {
...
+    /// Consensus metadata for the block. `None` for pre-fork blocks.
+    #[cfg_attr(
+        feature = "serde",
+        serde(default, skip_serializing_if = "Option::is_none")
+    )]
+    pub consensus_context: Option<TempoConsensusContext>,
+
     /// Inner Ethereum [`Header`].
     #[cfg_attr(feature = "serde", serde(flatten))]
     pub inner: Header,
```
hamdiallam: I think we can probably can workaround these encoding/decoding nonsense by a wrapper over the `inner` type that flattens the ethereum header along with anything trailing that we might add

crates/primitives/src/header.rs
```diff
@@ -140,3 +213,71 @@ impl Sealable for TempoHeader {
...
+        let header = presto_block_1();
+        let hash = header.hash_slow();
+
+        // Presto block 1 on-chain hash. If this changes, RLP encoding has broken.
+        let expected = "0x76e86f9739fbe17669b01b24e976ac214742c4b1bbc6ae0c083a87e43a5e9b0f";
+        assert_eq!(format!("{hash:#x}"), expected);
+    }
+
+    #[test]
+    fn presto_block_1_rlp_roundtrip() {
```
joshieDo: can we add a compact roundtrip too, from a serialized blob

joshieDo: https://github.com/tempoxyz/tempo/pull/3467 ended up adding here

hamdiallam: Added a test here as well just to be sure

crates/primitives/src/header.rs
```diff
@@ -2,16 +2,32 @@ use alloy_consensus::{BlockHeader, Header, Sealable};
...
+    pub proposer: B256,
+    pub parent_view: u64,
+}
+
+/// Tempo block header.
+///
+/// RLP-encoded as `[general_gas_limit, shared_gas_limit, timestamp_millis_part, inner,
+/// consensus_context?]`. The `consensus_context` is trailing and omitted for pre-fork blocks.
+///
+/// RLP is implemented manually because `consensus_context` must be trailing (optional) while
```
hamdiallam: Need to update this. RLP is no longer implemented manually, compact is

crates/primitives/src/header.rs
```diff
@@ -140,91 +163,3 @@ impl Sealable for TempoHeader {
         keccak256(alloy_rlp::encode(self))
     }
 }
-
-#[cfg(all(test, feature = "reth-codec"))]
-mod tests {
```
hamdiallam, line 153: moved into reth_compt directory with the added tests

joshieDo (approved):

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -569,13 +574,29 @@ impl Inner<Init> {
             current_timestamp
         };
 
+        let consensus_context = if self
+            .execution_node
+            .chain_spec()
+            .is_t4_active_at_timestamp(timestamp_millis / 1000)
```
SuperFluffy: Should this be `block.timestamp()`?

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -569,13 +574,29 @@ impl Inner<Init> {
...
+        {
+            Some(tempo_primitives::TempoConsensusContext {
+                epoch: round.epoch().get(),
+                view: round.view().get(),
+                proposer: alloy_primitives::B256::from_slice(leader.as_ref()),
```
SuperFluffy: A bit further down we have `public_key_to_b256`. in `crate::utils`

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -686,6 +707,32 @@ impl Inner<Init> {
             }
         }
 
+        if self
```
SuperFluffy: Let's merge this with `verify_header_extra_data` into a single `verify_header`.

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -686,6 +707,32 @@ impl Inner<Init> {
...
+            .execution_node
+            .chain_spec()
+            .is_t4_active_at_timestamp(block.timestamp())
```
SuperFluffy: Here you use `block.timestamp()`, where above you used `block.timestamp_millis() / 1000`.

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -686,6 +707,32 @@ impl Inner<Init> {
...
+        {
+            if let Some(ctx) = &block.header().consensus_context {
+                let expected_context = TempoConsensusContext {
+                    epoch: round.epoch().get(),
+                    view: round.view().get(),
+                    proposer: alloy_primitives::B256::from_slice(proposer.as_ref()),
```
SuperFluffy: `crate::utils::public_key_to_b256`

crates/commonware-node/src/consensus/application/ingress.rs
```diff
@@ -22,11 +22,11 @@ impl Mailbox {
 }
 
 /// Messages forwarded from consensus to application.
-// TODO: add trace spans into all of these messages.
+/// TODO: add trace spans into all of these messages.
```
SuperFluffy: Undo - this is not intended to be a doc comment.

crates/commonware-node/src/consensus/block.rs
```diff
@@ -133,6 +137,29 @@ impl commonware_consensus::Block for Block {
     }
 }
 
+impl commonware_consensus::CertifiableBlock for Block {
+    type Context = Context<Digest, PublicKey>;
+
+    fn context(&self) -> Self::Context {
+        let context = self.consensus_context.unwrap_or_default();
+        let leader = PublicKey::read(&mut context.proposer.as_ref())
+            .expect("valid ed25519 public key in consensus context");
```
SuperFluffy: Are we absolutely sure that an attack can't circumvent this? Unless we actually put a `PublicKey` into `context.proposer`, we cannot be sure that this is the case.

crates/primitives/src/header.rs
```diff
@@ -2,16 +2,29 @@ use alloy_consensus::{BlockHeader, Header, Sealable};
...
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, Default, RlpEncodable, RlpDecodable)]
 #[cfg_attr(feature = "reth-codec", derive(reth_codecs::Compact))]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
 #[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
 #[cfg_attr(test, reth_codecs::add_arbitrary_tests(compact, rlp))]
+pub struct TempoConsensusContext {
+    pub epoch: u64,
+    pub view: u64,
+    pub proposer: B256,
```
SuperFluffy: Flagging that this `B256` does not provide any type level guarantees on this actually being a well formed ed25519 public key.

Can we ensure well-formedness by providing a refinement type `PublicKey(B256)`? This is critical since you assume that all proposers stored in consensus context can be mapped to a ed25519 pubkey.

crates/commonware-node/src/consensus/application/actor.rs
hamdiallam: Because at this point we don't have the block available. We're preparing the block. And the timestamp taken and supplied to the payload attributes are in milliseconds

crates/commonware-node/src/consensus/block.rs
hamdiallam: Hm good callout. I think this is worth the effort to refine the type but still safe since these are generated from the proposer keys supplied by commonware and not from any node config. So an invalid block would be rejected since it would not match the serialized b256 created from the proposer

crates/commonware-node/src/consensus/application/actor.rs
SuperFluffy: Good point re the availability of `timestamp_millis` vs not - IMO, don't worry about the consensus context here. Always construct it and pass it to the payload builder.

Let the payload builder worry about when to provide context or not (so that it's all unified).

hamdiallam: Nice yea that is clean

hamdiallam: Lifting the seconds/millis part seems like a cleaner approach to consolidate the the `is_t4_active` checks into the application actor.

Lmk what you think of this. Happy to revert and move it into the assembler too, don't feel strongly at all

crates/commonware-node/src/consensus/block.rs
SuperFluffy: Did some double checking and I think this is a problem: our plan is to eventually move to commonware's `Deferred` wrapper, which calls `CertifiableBlock::context` before doing any other checks here: https://github.com/commonwarexyz/monorepo/blob/3dc4e1aa0ca31aecd362da05535db99d71c419f8/consensus/src/marshal/standard/deferred.rs#L487

This means that this is an attack vector that can be exploited by a single malicious proposer to crash the entire network.

SuperFluffy (changes requested):
Blocked on panic in `<Block as CertifiableBlock>::context`, as that is an easily exploitable attack vector; see comment and
https://github.com/commonwarexyz/monorepo/blob/3dc4e1aa0ca31aecd362da05535db99d71c419f8/consensus/src/marshal/standard/deferred.rs#L487

crates/primitives/src/header.rs
```diff
@@ -2,16 +2,62 @@ use alloy_consensus::{BlockHeader, Header, Sealable};
...
+/// adds Ed25519 key validation on top.
+#[derive(RlpDecodable)]
+struct TempoConsensusContextRlpDecodable {
+    epoch: u64,
+    view: u64,
+    proposer: B256,
+    parent_view: u64,
+}
+
+impl alloy_rlp::Decodable for TempoConsensusContext {
```
hamdiallam: cc @SuperFluffy 

Should we move this validation to `block.rs` within `commonware-node` which already depends on `commonware-cryptography`. And we wouldn't need  the `TempoConsensusContextRlpDecodable` mirror type

Specifically here in `Read::read_cfg` implementation.

https://github.com/tempoxyz/tempo/blob/main/crates/commonware-node/src/consensus/block.rs#L72

Bit awkward to introduce this here and supporting no_std.

hamdiallam: Leaning towards this direction and keeping `tempo-primitives` opaque

.config/zepter.yaml
```diff
@@ -17,6 +17,9 @@ workflows:
         "--left-side-feature-missing=ignore",
         # Ignore the case that `A` is outside of the workspace. Otherwise it will report errors in external dependencies that we have no influence on.
         "--left-side-outside-workspace=ignore",
+        # PublicKey serializes via B256, not ed25519-consensus's serde impl.
+        # Propagating would also break no_std (ed25519-consensus/serde pulls serde/std).
+        "--ignore-missing-propagate=tempo-primitives/serde:ed25519-consensus/serde",
```
hamdiallam, line 22: Needed since ed25519-consensus doesn't have `default-features = false`.

Will try upstream a fix so we can remove this ignore. Since we compile no_std we're safe

crates/commonware-node/src/utils.rs
```diff
@@ -16,6 +16,13 @@ pub(crate) fn public_key_to_b256(key: &PublicKey) -> B256 {
         .expect("ed25519 pub keys always map to B256")
 }
 
+pub(crate) fn public_key_to_tempo_primitive(
```
SuperFluffy, line 19: Let's add a roundtrip test here to ensure we can go between the two (probably don't go crazy; just use a static seed).

Should we look into fuzzzing? Probably overkill.

crates/commonware-node/src/consensus/block.rs
```diff
@@ -133,6 +140,31 @@ impl commonware_consensus::Block for Block {
...
+        match self.consensus_context {
+            Some(ctx) => Context {
+                leader: ctx.proposer.get().into(),
+                round: Round::new(Epoch::new(ctx.epoch), View::new(ctx.view)),
+                parent: (View::new(ctx.parent_view), self.parent_digest()),
+            },
+            None => {
+                // Pre-T4: Unused deterministic sentinel.
+                // Post-T4: Panic, all blocks where this is called must have a context.
+                //          Historical Pre-T4 blocks are never replayed in consensus
```
SuperFluffy, line 156: Panicking here would not work because then a malicious actor could just nuke the network.

We just need to rely on the network rejecting such blocks because the context never matches.

SuperFluffy (approved):

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -889,17 +909,39 @@ async fn verify_block<TContext: Pacer>(
...
 ) -> eyre::Result<()> {
     let epoch_info = epoch_strategy
         .containing(block.height())
         .expect("epoch strategy is for all heights");
+
+    if chainspec.is_t4_active_at_timestamp(block.timestamp()) {
+        let ctx = block
+            .header()
+            .consensus_context
+            .ok_or_eyre("missing consensus context")?;
```
SuperFluffy, line 929: mb "after t4 hardfork activation"?

crates/primitives/src/header.rs
```diff
@@ -31,6 +49,13 @@ pub struct TempoHeader {
     /// Inner Ethereum [`Header`].
     #[cfg_attr(feature = "serde", serde(flatten))]
     pub inner: Header,
+
+    /// Consensus metadata for the block. `None` for pre-fork blocks.
```
SuperFluffy, line 53: Add a comment that this must remain the last field so that we can grow `TempoConsensusContext` downward?

crates/commonware-node/src/consensus/block.rs
hamdiallam: Ah yea that's right, the only malicious payload would be omitting the context entirely

crates/commonware-node/src/utils.rs
hamdiallam: Yeah I was thinking about fuzzing but serialization into bytes is deterministic so comparing the bufs for a fixed seed should be enough

crates/primitives/src/header.rs
hamdiallam: So it doesn't have to be the last field. Other trailing fields can be added onto the header. Only constraint is this this must be populated.

For reference see all the trailing fields in the inner Header. Many options, post activation they can never be `None` again.

PR #3258, opened 2026-03-21
Fix provider output and correctly fetch chain ID

evshank:
Originally attempted to print the entire provider object instead of the actual chain ID, which is incorrect. Fixed by calling `get_chain_id()` to make a proper RPC request and retrieve the real network `chain ID`. Also added the `Provider` trait import to enable access to this method.


klkvr (approved):
ty

onbjerg (approved):

PR #3268, opened 2026-03-23
fix(tx-pool): track AA authorization authorities in validator

legion2002:
Closes QED-71

Ensures AA transactions contribute recovered authorities from `tempo_authorization_list` to pool validation metadata, matching EIP-7702 authority tracking behavior.

Adds a regression test that validates an AA transaction with a signed authorization entry and asserts the recovered authority is surfaced in `TransactionValidationOutcome::Valid.authorities`.

### DIFF
--- a/crates/transaction-pool/src/validator.rs
+++ b/crates/transaction-pool/src/validator.rs
@@ -665,7 +665,7 @@ mod tests {
         tip20::{TIP20Token, slots as tip20_slots},
     };
     use tempo_primitives::{
-        Block, TempoHeader, TempoPrimitives, TempoTxEnvelope,
+        Block, TempoHeader, TempoPrimitives, TempoTxEnvelope, TempoTxType,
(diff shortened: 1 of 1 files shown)


crates/transaction-pool/src/validator.rs
```diff
@@ -4642,6 +4661,60 @@ mod tests {
...
 
+    #[tokio::test]
+    async fn test_aa_authorization_list_authorities_tracked() {
+        use alloy_eips::eip7702::Authorization;
+        use alloy_signer::SignerSync;
+        use alloy_signer_local::PrivateKeySigner;
+        use tempo_primitives::transaction::{
+            TempoSignedAuthorization,
+            tt_signature::{PrimitiveSignature, TempoSignature},
+        };
```
fgimenez: nit, can we move this to the the top of the enclosing module?

PR #3271, opened 2026-03-23
fix(consensus): DKG ignores blocks from prior epochs

SuperFluffy:
The DKG loop now drops blocks from historic epochs.

Fixes the actor being overly strict and exiting on restarts when a boundary block was replayed even though the DKG actor already finalized it, causing the node to crash.

### DIFF
--- a/crates/commonware-node/src/dkg/manager/actor/mod.rs
+++ b/crates/commonware-node/src/dkg/manager/actor/mod.rs
@@ -852,11 +852,33 @@ where
             .containing(block.height())
             .expect("epoch strategy is covering all block heights");
 
-        ensure!(
-            epoch_info.epoch() == round.epoch(),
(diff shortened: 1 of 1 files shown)


crates/e2e/src/tests/sync.rs
```diff
@@ -312,12 +312,7 @@ fn can_restart_after_joining_from_snapshot() {
                     }
 
                     if metric.contains(&receiver.uid) {
-                        // -1 to account for stopping on boundaries.
-                        assert!(
-                            epoch >= last_epoch_before_stop.saturating_sub(1),
-                            "when starting from snapshot, older epochs must never \
-                            had consensus engines running"
-                        );
+                        assert!(epoch > 0, "validator should never boot into genesis epoch");
```
SuperFluffy, line 315: exists for all other tests, forgot to update it here.

PR #3275, opened 2026-03-23
refactor(revm): use standard InvalidTransaction for intrinsic gas errors

joohhnnn:
## Summary

Replace Tempo's custom `InsufficientGasForIntrinsicCost` error with standard `InvalidTransaction::CallGasCostMoreThanGasLimit` for intrinsic gas validation.

## Motivation

reth's `eth_estimateGas` uses a binary search that recognizes standard error types to decide whether to adjust gas limits up or down. The custom Tempo error mapped to `EthApiError::EvmCustom` (a generic string), which reth could not classify — causing it to abort the search instead of continuing.

With the standard error, `is_gas_too_low()` returns `true` and the binary search correctly raises the lower bound.

### Bug context

When calling `eth_estimateGas` for a native transfer with `maxFeePerGas` on presto, the RPC returns a raw internal error string instead of a structured gas estimation error:

```json
// Request
{"method":"eth_estimateGas","params":[{"from":"0xd5ad...","to":"0xd5ad...","value":"0x0","maxFeePerGas":"0x9502f9001"}]}

// Response — broken
{"error":{"message":"gas_limit 0 < intrinsic_gas 21000"}}
```

This happens because `InsufficientGasForIntrinsicCost` is a Tempo-specific error that reth doesn't know how to handle. It gets serialized as a string and returned as-is. Tools like Foundry's `cast send` can't parse this and just fail.

The underlying cause (fee token resolution for native transfers) is tracked separately.
This PR only fixes the error type so that reth can produce a proper RPC error response.

## Changes

- `handler.rs`: 5 call sites switched from `InsufficientGasForIntrinsicCost` to `InvalidTransaction::CallGasCostMoreThanGasLimit`
- `error.rs`: removed the now-unused `InsufficientGasForIntrinsicCost` variant
- `evm.rs`, `handler.rs`, `validator.rs`: updated test assertions and comments


crates/revm/src/handler.rs
```diff
@@ -1650,9 +1651,9 @@ where
 
     // Validate floor gas (Prague+)
     if !evm.ctx.cfg.is_eip7623_disabled() && gas_limit < batch_gas.floor_gas {
-        return Err(TempoInvalidTransaction::InsufficientGasForIntrinsicCost {
+        return Err(InvalidTransaction::CallGasCostMoreThanGasLimit {
             gas_limit,
-            intrinsic_gas: batch_gas.floor_gas,
+            initial_gas: batch_gas.floor_gas,
         }
```
klkvr, line 1656: this should actually use `GasFloorMoreThanGasLimit` i believe https://github.com/decofe/revm/blob/6ef10066478605c2c5029fbbdea6ec197b20a269/crates/handler/src/validation.rs#L247-L254

crates/revm/src/handler.rs
```diff
@@ -1382,20 +1382,22 @@ where
...
             }
 
             // Validate floor gas (Prague+)
             if !evm.ctx.cfg.is_eip7623_disabled() && gas_limit < init_gas.floor_gas {
-                return Err(TempoInvalidTransaction::InsufficientGasForIntrinsicCost {
+                return Err(InvalidTransaction::CallGasCostMoreThanGasLimit {
                     gas_limit,
-                    intrinsic_gas: init_gas.floor_gas,
+                    initial_gas: init_gas.floor_gas,
                 }
```
klkvr: same here

crates/revm/src/handler.rs
```diff
@@ -1382,20 +1382,22 @@ where
                 init_gas.floor_gas = 0u64;
             }
 
-            // Validate gas limit is sufficient for initial gas
+            // Validate gas limit is sufficient for initial gas.
+            // Use standard InvalidTransaction so reth's gas estimation binary search
+            // recognizes this as a "gas too low" error (is_gas_too_low() returns true).
```
klkvr: no need to change this comment

crates/revm/src/handler.rs
joohhnnn: Yeah nice catch

PR #3277, opened 2026-03-23
feat(evm): TIP-1020 signature verification precompile

howydev:
Precompile at `0x7369676E...` (`hex('sign')`) that verifies Tempo signatures (secp256k1, P256, WebAuthn) and returns the recovered signer address. Reuses `PrimitiveSignature::from_bytes()` and `recover_signer()` from tempo-primitives.


crates/precompiles/src/signature_verifier/mod.rs
```diff
@@ -0,0 +1,206 @@
...
+impl SignatureVerifier {
+    pub fn verify(
+        &mut self,
+        call: ISignatureVerifier::verifyCall,
+    ) -> Result<Address> {
+        // Parse and validate signature (handles size checks + type disambiguation).
+        let sig = PrimitiveSignature::from_bytes(&call.signature)
+            .map_err(|_| SignatureVerifierError::invalid_signature_format())?;
+
+        // Charge verification gas before crypto (SV5).
```
howydev:
```suggestion
        // Charge verification gas before performing verification (SV5).
```

crates/precompiles/src/signature_verifier/mod.rs
```diff
@@ -0,0 +1,194 @@
...
+pub struct SignatureVerifier {}
+
+impl SignatureVerifier {
+    pub fn initialize(&mut self) -> Result<()> {
+        self.__initialize()
+    }
+
+    pub fn verify(&mut self, call: ISignatureVerifier::verifyCall) -> Result<Address> {
+        // Parse and validate signature (handles size checks + type disambiguation).
+        let sig = PrimitiveSignature::from_bytes(&call.signature)
```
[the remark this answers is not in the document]
0xrusowsky: should we also add:
```rs
if let PrimitiveSignature::Secp256k1(ref s) = sig {
    let v = s.v().to_u64();
    if v != 27 && v != 28 {
        return Err(SignatureVerifierError::invalid_signature_format().into());
    }
}
```

to be consistent with TIP-1004 like in TIP20's permit impl?

crates/precompiles/src/signature_verifier/mod.rs
```diff
@@ -0,0 +1,194 @@
...
+            SignatureType::WebAuthn => WEBAUTHN_VERIFY_GAS,
+        };
+        self.storage.deduct_gas(verify_gas)?;
+
+        // Verify and recover signer (SV1, SV2, SV4).
+        let signer = sig
+            .recover_signer(&call.hash)
+            .map_err(|_| SignatureVerifierError::signature_verification_failed())?;
+
+        Ok(signer)
```
0xrusowsky: style nit:

```suggestion
        sig
            .recover_signer(&call.hash)
            .map_err(|_| SignatureVerifierError::signature_verification_failed())
```

crates/contracts/src/precompiles/signature_verifier.rs
```diff
@@ -0,0 +1,26 @@
...
+    #[derive(Debug, PartialEq, Eq)]
+    #[sol(abi)]
+    interface ISignatureVerifier {
+        /// @notice Verifies a Tempo signature (secp256k1, P256, WebAuthn).
+        /// @param hash The message hash that was signed
+        /// @param signature The encoded signature (see Tempo Transaction spec for formats)
+        /// @return signer Address of the signer if valid, reverts otherwise
+        function verify(bytes32 hash, bytes calldata signature) external view returns (address signer);
+
+        error InvalidSignatureFormat();
```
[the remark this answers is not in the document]
0xrusowsky: we should update the spec accordingly for richer errors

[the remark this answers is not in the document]
0xrusowsky: addressed in https://github.com/tempoxyz/tempo/pull/3306

crates/precompiles/src/signature_verifier/mod.rs
0xrusowsky: addressed in [`750a1fe` (this PR)](https://github.com/tempoxyz/tempo/pull/3277/commits/750a1fe9a6f2412c3c9da88cf673cf5607ae5506)

crates/precompiles/src/signature_verifier/dispatch.rs
```diff
@@ -0,0 +1,57 @@
...
+
+        // Pre-T3: behave like an empty contract (call succeeds, no execution)
+        if !self.storage.spec().is_t3() {
+            return Ok(PrecompileOutput::new(
+                self.storage.gas_used(),
+                Default::default(),
+            ));
+        }
+
+        dispatch_call(
```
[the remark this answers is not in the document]
0xrusowsky: addressed in [`750a1fe` (this PR)](https://github.com/tempoxyz/tempo/pull/3277/commits/750a1fe9a6f2412c3c9da88cf673cf5607ae5506)

crates/precompiles/src/signature_verifier/mod.rs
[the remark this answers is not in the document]
0xrusowsky: we want to be ecrecover compatible

crates/contracts/src/precompiles/signature_verifier.rs
```diff
@@ -0,0 +1,34 @@
...
+        /// @param signature The encoded signature (see Tempo Transaction spec for formats)
+        /// @return Address of the signer if valid, reverts otherwise
+        function recover(bytes32 hash, bytes calldata signature) external view returns (address signer);
+
+        /// @notice Verifies a signer against a Tempo signature (secp256k1, P256, WebAuthn).
+        /// @param signer The input address verified against the recovered signer
+        /// @param hash The message hash that was signed
+        /// @param signature The encoded signature (see Tempo Transaction spec for formats)
+        /// @return True if the input address signed, false otherwise. Reverts on invalid signatures.
+        function verify(address signer, bytes32 hash, bytes calldata signature) external view returns (bool);
```
legion2002, line 19: We discussed haveing a separate `verifyWithState` function in the precompile for access keys, because those sigs can be manipulated after signing. And blocking stateful key types from being verified using standard `verify`

crates/precompiles/src/signature_verifier/mod.rs
```diff
@@ -0,0 +1,197 @@
...
+
+        // Charge verification gas before crypto (SV5).
+        let verify_gas = match sig.signature_type() {
+            SignatureType::Secp256k1 => SECP256K1_VERIFY_GAS,
+            SignatureType::P256 => P256_VERIFY_GAS,
+            SignatureType::WebAuthn => WEBAUTHN_VERIFY_GAS,
+        };
+        self.storage.deduct_gas(verify_gas)?;
+
+        // Verify and recover signer (SV1, SV2, SV4).
```
legion2002: nit: not clear what SV1/2/4 is here

crates/precompiles/src/signature_verifier/dispatch.rs
```diff
@@ -0,0 +1,151 @@
...
+
+        // Pre-T3: behave like an empty contract (call succeeds, no execution)
+        if !self.storage.spec().is_t3() {
+            return Ok(PrecompileOutput::new(
+                self.storage.gas_used(),
+                Default::default(),
+            ));
+        }
+
+        if calldata.len() > MAX_CALLDATA_LEN {
```
legion2002, line 22: This check if redundant I think, because we already verify webauthn sig size, when converting using from_bytes here - https://github.com/tempoxyz/tempo/blob/d4fe0adcc78e8c3de0eb7f593a0cb125f825e12e/crates/primitives/src/transaction/tt_signature.rs#L156-L160

legion2002 (approved):
lgtm, pending the removal of the redundant sig size check

crates/precompiles/src/signature_verifier/dispatch.rs
0xrusowsky: while that's true, it is added here to avoid abi decoding on huge calldatas

legion2002: ah makes sense then

crates/contracts/src/precompiles/signature_verifier.rs
0xrusowsky: as discussed internally, we won't do this in T3 as it's simply an extension

samczsun (approved):

crates/evm/src/block.rs
```diff
@@ -390,6 +390,23 @@ where
...
+                .unwrap_or_default();
+            if info.is_empty_code_hash() {
+                let code = Bytecode::new_legacy([0xef].into());
+                info.code_hash = code.hash_slow();
+                info.code = Some(code);
+                let mut account: Account = info.into();
+                account.mark_touch();
+                db.commit(EvmState::from_iter([(SIGNATURE_VERIFIER_ADDRESS, account)]));
+            }
+        }
```
0xKitsune: Can we add tests to assert the expected code is deployed pre/post hardfork boundary?

crates/evm/src/block.rs
```diff
@@ -390,6 +390,23 @@ where
             }
         }
 
+        // Deploy 0xEF marker bytecode to SignatureVerifier when T3 activates.
+        if self.inner.spec.is_t3_active_at_timestamp(timestamp) {
+            let db = self.evm_mut().ctx_mut().journaled_state.db_mut();
+            let mut info = db
```
0xKitsune: Since this is the same code as the `ValidatorConfigV2` deployment above and we will also have the same logic for the `TIP20Registry`, consider creating a `deploy_precompile_at_boundary()` fn that takes the precompile address and the hardfork at which it should activate.

crates/precompiles/src/signature_verifier/dispatch.rs
```diff
@@ -0,0 +1,151 @@
...
+use revm::precompile::{PrecompileError, PrecompileOutput, PrecompileResult};
+use tempo_contracts::precompiles::{
+    ISignatureVerifier::ISignatureVerifierCalls as ISVCalls, SignatureVerifierError,
+};
+use tempo_primitives::MAX_WEBAUTHN_SIGNATURE_LENGTH;
+
+/// Maximum valid calldata size: `verify(address,bytes32,bytes)` with a WebAuthn signature is the
+/// worst case (4 + 32 + 32 + 32 + 32 + [1 + MAX_WEBAUTHN_SIGNATURE_LENGTH] = 2181).
+const MAX_CALLDATA_LEN: usize =
+    (4 + 32 * 4 + MAX_WEBAUTHN_SIGNATURE_LENGTH + 1).next_multiple_of(32);
```
0xKitsune: The `MAX_CALLDATA_LEN` formula rounds the entire sum `(selector + args + sig)` to a 32-byte boundary, but ABI encoding pads the dynamic bytes field independently so the 4-byte selector eats into the padding and the limit ends up 4 bytes too small (2208 vs 2212). This means a `verify()` call with a max length WebAuthn signature (2049 bytes) will be wrongly rejected with `InvalidFormat`. Fix by only rounding the dynamic portion: `4 + 32 * 4 + (MAX_WEBAUTHN_SIGNATURE_LENGTH + 1).next_multiple_of(32).`

```rust
#[test]
fn test_max_webauthn_verify_passes_size_guard() -> eyre::Result<()> {
    let mut storage = HashMapStorageProvider::new_with_spec(1, TempoHardfork::T3);
        let mut sig = vec![0x02u8];
        sig.extend_from_slice(&[0u8; MAX_WEBAUTHN_SIGNATURE_LENGTH]);

        let calldata = ISignatureVerifier::verifyCall {
            signer: Address::ZERO,
            hash: B256::ZERO,
            signature: sig.into(),
        }
        .abi_encode();

        let result = SignatureVerifier::new().call(&calldata, Address::ZERO)?;
        // Should NOT be rejected by the size guard, should fail later at signature validation
        assert!(
            SignatureVerifierError::abi_decode(&result.bytes)
                .map(|e| e != SignatureVerifierError::invalid_format())
                .unwrap_or(true),
            "max-size WebAuthn calldata was wrongly rejected by size guard"
        );
        Ok(())
    })
}
```

This asserts a max-size WebAuthn call should not return `InvalidFormat`, which currently fails because the size guard rejects it before it reaches signature validation.

0xrusowsky: good catch!

addressed in [`174b25f` (this PR)](https://github.com/tempoxyz/tempo/pull/3277/commits/174b25fc6812773871d9b068204c88113c4344fe)

crates/evm/src/block.rs
0xrusowsky: addressed in [`174b25f` (this PR)](https://github.com/tempoxyz/tempo/pull/3277/commits/174b25fc6812773871d9b068204c88113c4344fe)

crates/evm/src/block.rs
0xrusowsky: addressed in [`174b25f` (this PR)](https://github.com/tempoxyz/tempo/pull/3277/commits/174b25fc6812773871d9b068204c88113c4344fe)

crates/precompiles/src/signature_verifier/dispatch.rs
```diff
@@ -0,0 +1,178 @@
...
+            .deduct_gas(input_cost(calldata.len()))
+            .map_err(|_| PrecompileError::OutOfGas)?;
+
+        // Pre-T3: behave like an empty contract (call succeeds, no execution)
+        if !self.storage.spec().is_t3() {
+            return Ok(PrecompileOutput::new(
+                self.storage.gas_used(),
+                Default::default(),
+            ));
+        }
```
klkvr: this should be removed and precompile should instead be conditionally inserted in `extend_tempo_precompiles`

klkvr (changes requested):

crates/evm/src/block.rs
```diff
@@ -152,6 +152,27 @@ where
         }
     }
 
+    /// Deploys `0xEF` marker bytecode to a precompile address if it doesn't already have code.
+    fn deploy_precompile_at_boundary(
```
0xrusowsky, line 158: let's use the helper version implemented in https://github.com/tempoxyz/tempo/pull/3397

crates/precompiles/src/signature_verifier/dispatch.rs
0xrusowsky: addressed in https://github.com/tempoxyz/tempo/commit/49a01b35ef89fbeac614f7eeaf84ef95a40a191e

klkvr (approved):

crates/precompiles/src/signature_verifier/mod.rs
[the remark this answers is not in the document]
howydev: Late, but I like preventing this/doing this check here. Currently sc devs rely on libraries (solady, OZ) to do this check

The risk here is invalidating valid sigs, but the worst case here is that the user has to sign again? Doesn't seem too bad

PR #3280, opened 2026-03-24
fix: bounded-memory state bloat generation and ingestion

decofe:
Closes RETH-216

State bloat generation and ingestion OOM at >=10GB because both tools hold all data in memory. Generation materializes ~33GB of vectors for 100GB target; ingestion collects all entries in a `HashMap` before flushing, consuming ~1.6× file size in heap.

Fix: stream both phases. Generator writes 256k-entry chunks. Loader batches `insert_storage_for_hashing` every 40M entries (~2.8 GB heap cap) with periodic MDBX commits, bounding heap usage regardless of input size.


bin/tempo/src/init_state.rs
```diff
@@ -77,24 +79,28 @@ impl<C: reth_cli::chainspec::ChainSpecParser<ChainSpec: EthChainSpec + EthereumH
...
-            match reader.read_exact(&mut header_buf) {
-                Ok(()) => {}
-                Err(e) if e.kind() == std::io::ErrorKind::UnexpectedEof => break,
+            match reader.read(&mut header_buf[..1]) {
+                Ok(0) => break, // clean EOF between blocks
+                Ok(_) => {}
                 Err(e) => return Err(e).wrap_err("failed to read block header"),
             }
+            reader
+                .read_exact(&mut header_buf[1..])
```
yongkangc: previously used `read_exact` on all 40 header bytes and treated UnexpectedEof as a clean EOF, masking truncated files. 

now we first checks for true EOF with a 1-byte read, then uses read_exact for the remaining 39 bytes so truncation surfaces as an error.

bin/tempo/src/init_state.rs
```diff
@@ -77,24 +79,28 @@ impl<C: reth_cli::chainspec::ChainSpecParser<ChainSpec: EthChainSpec + EthereumH
...
-            HashMap::new();
+        let mut total_blocks = 0u64;
+        let mut total_commits = 0u64;
 
-        // Track addresses for account hashing (we need to create empty accounts)
+        // Track addresses for account hashing (small — only token addresses)
         let mut addresses_seen: HashSet<alloy_primitives::Address> = HashSet::default();
 
+        // Batch buffer for hashed storage inserts (multi-address)
+        let mut batch: HashMap<alloy_primitives::Address, Vec<StorageEntry>> = HashMap::default();
```
shekhirin: can use `AddressMap`

bin/tempo/src/init_state.rs
```diff
@@ -77,24 +79,28 @@ impl<C: reth_cli::chainspec::ChainSpecParser<ChainSpec: EthChainSpec + EthereumH
...
-
-        // Collect storage entries per address for hashing
-        let mut storage_for_hashing: HashMap<alloy_primitives::Address, Vec<StorageEntry>> =
-            HashMap::new();
+        let mut total_blocks = 0u64;
+        let mut total_commits = 0u64;
 
-        // Track addresses for account hashing (we need to create empty accounts)
+        // Track addresses for account hashing (small — only token addresses)
         let mut addresses_seen: HashSet<alloy_primitives::Address> = HashSet::default();
```
shekhirin: `AddressSet`

bin/tempo/src/init_state.rs
```diff
@@ -77,24 +79,28 @@ impl<C: reth_cli::chainspec::ChainSpecParser<ChainSpec: EthChainSpec + EthereumH
...
-                Ok(()) => {}
-                Err(e) if e.kind() == std::io::ErrorKind::UnexpectedEof => break,
+            match reader.read(&mut header_buf[..1]) {
+                Ok(0) => break, // clean EOF between blocks
+                Ok(_) => {}
                 Err(e) => return Err(e).wrap_err("failed to read block header"),
             }
+            reader
+                .read_exact(&mut header_buf[1..])
+                .wrap_err("truncated block header (got partial header at end of file)")?;
```
shekhirin: do we need this? I think we always expect the input to be properly formed

xtask/src/generate_state_bloat.rs
```diff
@@ -100,137 +111,105 @@ impl GenerateStateBloat {
...
-            writer.write_all(&total_supply_slot_bytes)?;
-            writer.write_all(&total_supply_bytes)?;
+        let mut chunk_buf = Vec::with_capacity(chunk_size.min(total_accounts) * 64);
+
+        for chunk_idx in 0..num_chunks {
+            let chunk_start = chunk_idx * chunk_size;
+            let chunk_end = (chunk_start + chunk_size).min(total_accounts);
+            let chunk_len = chunk_end - chunk_start;
+
+            // Derive addresses and compute slot bytes for this chunk only
```
shekhirin, line 171: this makes sense

xtask/src/generate_state_bloat.rs
```diff
@@ -100,137 +111,105 @@ impl GenerateStateBloat {
...
-
-            // Write total_supply entry
-            writer.write_all(&total_supply_slot_bytes)?;
-            writer.write_all(&total_supply_bytes)?;
+        let mut chunk_buf = Vec::with_capacity(chunk_size.min(total_accounts) * 64);
+
+        for chunk_idx in 0..num_chunks {
+            let chunk_start = chunk_idx * chunk_size;
+            let chunk_end = (chunk_start + chunk_size).min(total_accounts);
+            let chunk_len = chunk_end - chunk_start;
```
shekhirin: can we just do `for chunk in (0..total_accounts).chunks(chunk_size) { ... }` here?

shekhirin (commented):
the approach makes sense! Some comments

bin/tempo/src/init_state.rs
```diff
@@ -165,10 +171,26 @@ impl<C: reth_cli::chainspec::ChainSpecParser<ChainSpec: EthChainSpec + EthereumH
...
                 total_entries += 1;
 
+                // Flush batch when it reaches the threshold
+                if batch_total >= HASHING_BATCH_SIZE {
+                    // Drop cursor before commit
+                    drop(storage_cursor);
+
+                    provider_rw.insert_storage_for_hashing(batch.drain())?;
+                    batch_total = 0;
+                    provider_rw.commit()?;
```
shekhirin: why do we need this? As far as I know, MDBX handles big transactions ok and do not keep the whole set of changes in memory, but slowly spills them into the disk

bin/tempo/src/init_state.rs
yongkangc: thanks! addressed this

bin/tempo/src/init_state.rs
yongkangc: i was afraid there might be some case where we accidentally have the wrong input in the future, yeah but i guess not needed - removing it to reduce diff

xtask/src/generate_state_bloat.rs
yongkangc: yup thanks for that, simpler to read

bin/tempo/src/init_state.rs
yongkangc: i see, i was intiially afraid that the oom could be due to big txns held in memory. thanks for the note, simplified

bin/tempo/src/init_state.rs
```diff
@@ -165,10 +167,21 @@ impl<C: reth_cli::chainspec::ChainSpecParser<ChainSpec: EthChainSpec + EthereumH
...
 
+                // Flush batch when it reaches the threshold to bound heap usage
+                if batch_total >= HASHING_BATCH_SIZE {
+                    drop(storage_cursor);
+
+                    provider_rw.insert_storage_for_hashing(batch.drain())?;
+                    batch_total = 0;
+
+                    let tx = provider_rw.tx_ref();
+                    storage_cursor = tx.cursor_dup_write::<tables::PlainStorageState>()?;
```
shekhirin: cursor reinitialisation can be removed

bin/tempo/src/init_state.rs
```diff
@@ -165,10 +167,21 @@ impl<C: reth_cli::chainspec::ChainSpecParser<ChainSpec: EthChainSpec + EthereumH
                 storage_cursor.upsert(address, &entry)?;
 
                 // Collect for hashed storage
-                storage_entries.push(entry);
-
+                batch.entry(address).or_default().push(entry);
```
shekhirin: can we still do `batch.entry(address)` once above to not go through hashmap every time?

decofe:
cc @shekhirin

⚠️ Benchmark cancelled. [View logs](https://github.com/tempoxyz/tempo/actions/runs/23568517675/job/68625670705)

decofe:
cc @shekhirin

❌ Benchmark failed. [View logs](https://github.com/tempoxyz/tempo/actions/runs/23569811847/job/68629677441)

decofe:
cc @shekhirin

🚀 Benchmark started! [View job](https://github.com/tempoxyz/tempo/actions/runs/23570731536/job/68632474244)

**Config:** preset: `tip20`, duration: `300s`, bloat: `100000 MiB`, tps: `10000`, baseline: `yk/fix-state-bloat-oom`, feature: `yk/fix-state-bloat-oom`

PR #3286, opened 2026-03-24
feat(precompiles): TIP20 virtual address support

0xrusowsky:
implements the spec defined in
- https://github.com/tempoxyz/tempo/pull/2852

### DIFF
--- a/crates/primitives/src/reth_compat/transaction/envelope.rs
+++ b/crates/primitives/src/reth_compat/transaction/envelope.rs
@@ -71,6 +71,9 @@ mod codec {
                 }
                 TempoTxType::AA => {
                     let (tx, buf) = TempoTransaction::from_compact(buf, buf.len());
+                    // The provided `signature` is unused for AA transactions. The real
+                    // `TempoSignature` was appended to the buffer in `to_tx_compact` and
+                    // is decoded here instead.
                     let (sig_bytes, buf) = Bytes::from_compact(buf, buf.len());
                     let aa_sig = TempoSignature::from_bytes(&sig_bytes)
                         .map_err(|e| panic!("Failed to decode AA signature: {e}"))
@@ -106,7 +109,13 @@ mod codec {
                 Self::Eip1559(tx) => tx.signature(),
                 Self::Eip7702(tx) => tx.signature(),
                 Self::AA(_tx) => {
-                    // TODO: Will this work?
+                    // The `Envelope` trait requires `&Signature` (ECDSA), but AA transactions
+                    // use `TempoSignature` which is a different type. We return a dummy zero
+                    // signature here because `CompactEnvelope::to_compact` calls this to
+                    // serialize a signature into the buffer. The actual `TempoSignature` is
+                    // encoded separately in `ToTxCompact::to_tx_compact` and decoded back in
+                    // `FromTxCompact::from_tx_compact`, where the dummy signature passed in
+                    // is ignored for the AA variant.
                     &TEMPO_SYSTEM_TX_SIGNATURE
                 }
             }


crates/precompiles/src/tip20/mod.rs
```diff
@@ -1,11 +1,12 @@
...
 //!
 //! Provides ERC-20-like balances, allowances, and transfers with Tempo extensions:
-//! role-based access control, pausability, supply caps, transfer policies ([TIP-403]),
-//! opt-in staking rewards,EIP-2612 permits (post-T2) and quote-token graphs.
+//! role-based access control, pausability, supply caps, transfer policies ([TIP-403]), opt-in
+//! staking rewards, EIP-2612 permits (T2+), quote-token graphs, and virtual addresses ([TIP-1022]).
 //!
 //! [TIP-20]: <https://docs.tempo.xyz/protocol/tip20>
 //! [TIP-403]: <https://docs.tempo.xyz/protocol/tip403>
+//! [TIP-1022]: <https://docs.tempo.xyz/protocol/tip1022>
```
0xrusowsky, line 9: note: these links don't exist yet, i'll open a PR to docs too

tips/tip-1022.md
```diff
@@ -177,7 +177,7 @@ This layout packs all metadata for a `masterId` into a single storage slot, enab
 ### Interface
 
 ```solidity
-interface IVirtualAddressRegistry {
+interface ITIP20Registry {
```
0xKitsune: very minor nit: I think the previous naming convention was a bit more explicit.

crates/precompiles/src/address_registry/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+/// 6-byte user tag occupying the trailing bytes of a virtual address.
+pub type UserTag = FixedBytes<6>;
+
+/// 10-byte magic value identifying virtual addresses at bytes `[4:14]`.
+pub const VIRTUAL_MAGIC: [u8; 10] = [0xFD; 10];
+
+/// Returns `true` if `addr` matches the [TIP-1022] virtual-address format
+/// (bytes `[4:14]` == [`VIRTUAL_MAGIC`]).
+///
+/// [TIP-1022]: <https://docs.tempo.xyz/protocol/tip1022>
```
0xKitsune, line 35: nit: I dont think we need to add `/// [TIP-1022]: <https://docs.tempo.xyz/protocol/tip1022>`
 to every function.

crates/precompiles/src/tip20_registry/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+/// A valid master address:
+/// - MUST NOT be `address(0)`
+/// - MUST NOT itself match the virtual-address format
+/// - MUST NOT be a [TIP-20] token address
+///
+/// [TIP-20]: <https://docs.tempo.xyz/protocol/tip20>
+/// [TIP-1022]: <https://docs.tempo.xyz/protocol/tip1022>
+pub fn is_master_address(addr: Address) -> bool {
+    !addr.is_zero() && !is_virtual_address(addr) && !is_tip20_prefix(addr)
+}
```
0xKitsune: This will return `true` for all precompiles other than TIP20s. IIUC looks like this isnt exposed externally so should be fine though.

Also this function name is a bit misleading. Also `is_master_address()` could make users think it checks whether an address is the active master address, rather than whether the address is simply valid for use as a master address. Consider updating to `is_valid_master_address()` or similar.

tips/tip-1022.md
0xrusowsky: there were discussions to use the master ID concept for other sorts of address identification, so i think it makes sense to use this a general-purpose TIP20 registry, which is limited to virtual addresses for the time being.

i.e. @danrobinson has a TIP for some addresses to be implicitly pre-approved on TIP20, and i think this registry precompile is the perfect place for that (ref: https://github.com/tempoxyz/tempo/pull/3219#discussion_r2980587993 )

crates/precompiles/src/address_registry/dispatch.rs
```diff
@@ -0,0 +1,134 @@
...
+            .deduct_gas(input_cost(calldata.len()))
+            .map_err(|_| PrecompileError::OutOfGas)?;
+
+        // Pre-T3: behave like an empty contract (call succeeds, no execution)
+        if !self.storage.spec().is_t3() {
+            return Ok(PrecompileOutput::new(
+                self.storage.gas_used(),
+                Default::default(),
+            ));
+        }
```
0xKitsune: nit: consider adding a test for pre-t3 calls to the registry. This should be covered in re-execution tests but consider adding as a unit test here as well.

crates/precompiles/src/tip20/mod.rs
```diff
@@ -1,11 +1,12 @@
 //! [TIP-20] token standard — Tempo's native fungible token implementation.
 //!
 //! Provides ERC-20-like balances, allowances, and transfers with Tempo extensions:
-//! role-based access control, pausability, supply caps, transfer policies ([TIP-403]),
-//! opt-in staking rewards,EIP-2612 permits (post-T2) and quote-token graphs.
+//! role-based access control, pausability, supply caps, transfer policies ([TIP-403]), opt-in
+//! staking rewards, EIP-2612 permits (T2+), quote-token graphs, and virtual addresses ([TIP-1022]).
 //!
 //! [TIP-20]: <https://docs.tempo.xyz/protocol/tip20>
 //! [TIP-403]: <https://docs.tempo.xyz/protocol/tip403>
```
0xKitsune, line 8: Can we also add TIP20 integration tests for virtual address transfers (transfer/mint/transferFrom to virtual)?

crates/precompiles/src/tip20/mod.rs
```diff
@@ -390,19 +466,27 @@ impl TIP20Token {
...
+    /// Mints `amount` tokens to the resolved target `to` address:
+    /// - Enforces mint-recipient compliance via [`TIP403Registry`] and validates against supply cap
+    /// - Resolves `to` via the [`TIP20Registry`]. If `to` is a virtual address, credits the
+    ///   resolved master and emits a two-hop `Transfer` and `Mint(virtual, amount)` events
     ///
     /// # Errors
     /// - `PolicyForbids` — TIP-403 policy rejects the mint recipient
     /// - `Unauthorized` — caller does not hold the `ISSUER_ROLE` role
     /// - `SupplyCapExceeded` — minting would push total supply above the cap
     pub fn mint(&mut self, msg_sender: Address, call: ITIP20::mintCall) -> Result<()> {
```
0xKitsune, line 404: Missing unit tests covering mints to registered virtual addresses. We should assert that the master addr is credited and all other storage values are correct.

crates/precompiles/src/tip20/mod.rs
```diff
@@ -412,7 +496,8 @@ impl TIP20Token {
         msg_sender: Address,
         call: ITIP20::mintWithMemoCall,
     ) -> Result<()> {
-        self._mint(msg_sender, call.to, call.amount)?;
+        let to = Recipient::resolve(call.to)?;
```
0xKitsune, line 425: Same thing here looks like we are missing tests to cover virtual address behavior (event emission, etc.).

crates/precompiles/src/tip20/mod.rs
```diff
@@ -67,13 +69,87 @@ pub fn validate_usd_currency(token: Address) -> Result<()> {
...
+/// For non-virtual recipients both fields are equal. For [TIP-1022] virtual addresses, `event` is
+/// used in `Transfer` events while `target` is the resolved master whose balance is updated.
+///
+/// [TIP-1022]: <https://docs.tempo.xyz/protocol/tip1022>
+pub struct Recipient {
+    /// Address used in `Transfer` events — the literal recipient supplied by the caller.
+    pub event: Address,
+    /// Address whose balance is actually credited.
+    pub target: Address,
+}
```
0xKitsune: IMO the `Recipient` struct in its current form makes the TIP20 logic harder to reason about. 

This introduces the concept of an `event` address into the recipient logic and also has the `Recipient` struct construct `Transfer` events, so now when reading through the transfer/mint paths you have to keep cached that `event` is just `call.to`, that `target` is the resolved master, and that the struct is also responsible for building log entries. It's a lot of indirection to hold in your head.

A simpler approach would be to just have a `resolve_to` fn that returns the resolved address, then at the callsite check if the address is virtual and emit the hop event inline. No extra struct or functions and is easier to follow when reading the code.

0xrusowsky: the dual-address complexity isn't introduced by `Recipient` — it's inherent to virtual addresses. IMHO abstracting away the virtual logic is the right call as it allows us to reason about the transfer path cleanly.

with that being said, i think that u make a good point about `event`/`target` naming  being more complex that it needs to... what if we reshape the struct to match the mental model?

```rs
pub struct Recipient {
    /// The effective (resolved) address where the balance is credited.
    pub target: Address,
    /// The virtual address, if registered.
    pub virtual: Option<Address>,
}
```

the trade-off is 1 extra byte (the `Option` discriminant) for better comprehension, and the transfer paths stay virtual-address-agnostic, with just an `if let Some(event) = to.virtual_transfer()` at each callsite.

crates/precompiles/src/tip20/mod.rs
0xrusowsky: addressed in [`5f45a2d` (this PR)](https://github.com/tempoxyz/tempo/pull/3286/commits/5f45a2d698a0edfc1cac77bc20080fc3c0256589)

crates/precompiles/src/tip20/mod.rs
0xrusowsky: addressed in [`5f45a2d` (this PR)](https://github.com/tempoxyz/tempo/pull/3286/commits/5f45a2d698a0edfc1cac77bc20080fc3c0256589)

crates/precompiles/src/tip20/mod.rs
0xrusowsky: addressed in [`5f45a2d` (this PR)](https://github.com/tempoxyz/tempo/pull/3286/commits/5f45a2d698a0edfc1cac77bc20080fc3c0256589)

crates/precompiles/src/address_registry/dispatch.rs
0xrusowsky: addressed in [`5f45a2d` (this PR)](https://github.com/tempoxyz/tempo/pull/3286/commits/5f45a2d698a0edfc1cac77bc20080fc3c0256589)

crates/precompiles/src/tip20/mod.rs
0xrusowsky: addressed in [`0baa76c` (this PR)](https://github.com/tempoxyz/tempo/pull/3286/commits/0baa76c12bb64ae117077e39fab0f0325f77bc2f)

crates/precompiles/src/address_registry/mod.rs
0xrusowsky: addressed in [`9b38c4b` (this PR)](https://github.com/tempoxyz/tempo/pull/3286/commits/9b38c4b58c252903fb3eeb4f5af224d486295de6)

crates/precompiles/src/tip20_registry/mod.rs
0xrusowsky: addressed in [`c0b4cbb` (this PR)](https://github.com/tempoxyz/tempo/pull/3286/commits/c0b4cbb98d6db0786e82b906a4d81b2ac340deb3)

crates/evm/src/block.rs
```diff
@@ -390,6 +390,23 @@ where
...
+                .unwrap_or_default();
+            if info.is_empty_code_hash() {
+                let code = Bytecode::new_legacy([0xef].into());
+                info.code_hash = code.hash_slow();
+                info.code = Some(code);
+                let mut account: Account = info.into();
+                account.mark_touch();
+                db.commit(EvmState::from_iter([(TIP20_REGISTRY_ADDRESS, account)]));
+            }
+        }
```
0xKitsune: We should have a unit/inttegration test to assert the the registry is deployed correctly at T3 activation.

crates/node/tests/it/tip20.rs
```diff
@@ -931,3 +935,206 @@ async fn test_tip20_pause_blocks_fee_collection() -> eyre::Result<()> {
...
+    Ok((setup, http_url, admin, token, virtual_addr))
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_tip20_virtual_mint() -> eyre::Result<()> {
+    reth_tracing::init_test_tracing();
+    let (_setup, _http_url, admin, token, virtual_addr) = setup_virtual_test().await?;
+
+    let gas = 2_000_000u64;
+    let gas_price = TEMPO_T1_BASE_FEE as u128;
```
0xKitsune: very minor nit: Do we need to specify gas and gas price? Very minor, but would reduce the amount of code with all of these new tests.

crates/node/tests/it/tip20.rs
```diff
@@ -931,3 +935,206 @@ async fn test_tip20_pause_blocks_fee_collection() -> eyre::Result<()> {
...
+    let token_addr = *token.address();
+    let transfers: Vec<_> = mint_receipt
+        .logs()
+        .iter()
+        .filter(|log| log.address() == token_addr)
+        .filter_map(|log| ITIP20::Transfer::decode_log(&log.inner).ok())
+        .collect();
+    assert_eq!(transfers.len(), 2);
+    assert_eq!(transfers[0].from, Address::ZERO);
+    assert_eq!(transfers[0].to, virtual_addr);
```
0xKitsune: Can we also assert the amount here (and in other tests)?

Consider simply comparing the whole event with:

```rust

    assert_eq!(
        transfers[0].data,
        ITIP20::Transfer {
            from: Address::ZERO,
            to: virtual_addr,
            amount: mint_amount,
        }
    );
```

crates/precompiles/src/tip20_registry/mod.rs
```diff
@@ -0,0 +1,438 @@
...
+pub fn is_virtual_address(addr: Address) -> bool {
+    addr.as_slice()[4..14] == VIRTUAL_MAGIC
+}
+
+/// Returns `true` if `addr` is eligible to be a virtual-address master per TIP-1022.
+///
+/// A valid master address:
+/// - MUST NOT be `address(0)`
+/// - MUST NOT itself match the virtual-address format
+/// - MUST NOT be a TIP-20 token address
```
0xKitsune: nit: these coments are unnecessary, the same information can be inferred from reading the one line of code.

crates/precompiles/src/tip20/mod.rs
```diff
@@ -66,13 +68,88 @@ pub fn validate_usd_currency(token: Address) -> Result<()> {
...
+            return Err(TIP20Error::invalid_recipient().into());
+        }
+        Ok(())
+    }
+
+    /// Builds the primary `Transfer(from, to, amount)` event.
+    ///
+    /// For virtual recipients `to` is the virtual address (first hop); for regular
+    /// recipients this is the only `Transfer` event needed.
+    pub fn transfer(&self, from: Address, amount: U256) -> TIP20Event {
```
0xKitsune: nit: when reading this in the code, it looks like `Recipient::transfer()` which at first glance (without inspecting the function's code) seems like this should be transferring something. Consider updating to `build_transfer_event` or something for clarity.

Same thing for `virtual_transfer` -> `build_virtual_transfer_event`.

crates/precompiles/src/address_registry/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+#[contract(addr = TIP20_REGISTRY_ADDRESS)]
+pub struct TIP20Registry {
+    /// Maps `masterId → RegistryData` (master address + metadata).
+    data: Mapping<MasterId, RegistryData>,
+}
+
+/// Storage record for a registered master. Packed into a single 32-byte slot.
+#[derive(Debug, Clone, Default, Storable)]
+struct RegistryData {
```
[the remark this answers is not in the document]
0xrusowsky: modified spec instead

crates/precompiles/src/tip20/mod.rs
0xrusowsky: addressed in [`7348aa2` (this PR)](https://github.com/tempoxyz/tempo/pull/3286/commits/7348aa2d1150fb18719b4b85b5f105bcae1049f1)

crates/precompiles/src/tip20_registry/mod.rs
0xrusowsky: addressed in [`7348aa2` (this PR)](https://github.com/tempoxyz/tempo/pull/3286/commits/7348aa2d1150fb18719b4b85b5f105bcae1049f1)

crates/node/tests/it/tip20.rs
0xrusowsky: addressed in [`7348aa2` (this PR)](https://github.com/tempoxyz/tempo/pull/3286/commits/7348aa2d1150fb18719b4b85b5f105bcae1049f1)

crates/node/tests/it/tip20.rs
0xrusowsky: addressed in [`7348aa2` (this PR)](https://github.com/tempoxyz/tempo/pull/3286/commits/7348aa2d1150fb18719b4b85b5f105bcae1049f1)

crates/evm/src/block.rs
0xrusowsky: addressed in [`d679be5` (this PR)](https://github.com/tempoxyz/tempo/pull/3286/commits/d679be568a84f914fc78f64cfa7b7953e3a6e34f) + [`a6b8c72` (this PR)](https://github.com/tempoxyz/tempo/pull/3286/commits/a6b8c72dbf3038139f82bac75a95bc1013af229b)

crates/evm/src/block.rs
```diff
@@ -390,6 +390,23 @@ where
             }
         }
 
+        // Deploy 0xEF marker bytecode to AddressRegistry when T3 activates.
+        if self.inner.spec.is_t3_active_at_timestamp(timestamp) {
```
0xrusowsky, line 410: use the helper introduced in https://github.com/tempoxyz/tempo/pull/3397

crates/precompiles/src/address_registry/dispatch.rs
```diff
@@ -0,0 +1,154 @@
...
+use tempo_contracts::precompiles::IAddressRegistry::IAddressRegistryCalls;
+
+impl Precompile for AddressRegistry {
+    fn call(&mut self, calldata: &[u8], msg_sender: Address) -> PrecompileResult {
+        self.storage
+            .deduct_gas(input_cost(calldata.len()))
+            .map_err(|_| PrecompileError::OutOfGas)?;
+
+        // Pre-T3: behave like an empty contract (call succeeds, no execution)
+        if !self.storage.spec().is_t3() {
```
0xrusowsky: this should be removed and precompile should instead be conditionally inserted in `extend_tempo_precompiles`

crates/evm/src/block.rs
0xrusowsky: addressed in https://github.com/tempoxyz/tempo/commit/42d9b7e8d49e86a6e860fb1b0967634f75cccec2

crates/precompiles/src/address_registry/dispatch.rs
0xrusowsky: addressed in https://github.com/tempoxyz/tempo/commit/42d9b7e8d49e86a6e860fb1b0967634f75cccec2

.github/workflows/specs.yml
```diff
@@ -158,7 +158,7 @@ jobs:
         uses: actions/checkout@v6
         with:
           repository: tempoxyz/tempo-foundry
-          ref: tempo
+          ref: rusowsky/init-tip20-registry
```
klkvr:
```suggestion
          ref: tempo
```

crates/precompiles/src/tip20/mod.rs
```diff
@@ -422,17 +508,24 @@ impl TIP20Token {
...
+        }
+        Ok(())
     }
 
-    /// Internal helper to mint new tokens and update balances
-    fn _mint(&mut self, msg_sender: Address, to: Address, amount: U256) -> Result<()> {
+    /// Internal helper to mint new tokens and update balances.
+    fn _mint(&mut self, msg_sender: Address, to: &Recipient, amount: U256) -> Result<()> {
         self.check_role(msg_sender, *ISSUER_ROLE)?;
         let total_supply = self.total_supply()?;
```
fgimenez, line 447: on solidity we add in this PR a call to `_validateRecipient` in `_mint`, should we do the same here?

crates/precompiles/src/address_registry/mod.rs
```diff
@@ -0,0 +1,427 @@
...
+
+    /// Resolves a transfer recipient using virtual address semantics.
+    ///
+    /// Non-virtual addresses are returned unchanged.
+    /// Virtual addresses are resolved to their registered master.
+    ///
+    /// # Errors
+    /// - `VirtualAddressUnregistered` — `to` is a virtual address whose `masterId` is not registered
+    pub fn resolve_recipient(&self, to: Address) -> Result<Address> {
+        // Explicit check because it isn't exclusibly a view function.
```
fgimenez:
```suggestion
        // Explicit check because it isn't exclusively a view function.
```

crates/precompiles/src/tip20/mod.rs
```diff
@@ -66,13 +68,88 @@ pub fn validate_usd_currency(token: Address) -> Result<()> {
     Ok(())
 }
 
+/// Resolved transfer recipient for [TIP-1022] virtual address support.
+///
+/// `to` is always the effective (resolved) address where the balance is credited. For virtual
+/// recipients, `virtual_addr` carries the original virtual address for event emission.
+///
+/// [TIP-1022]: <https://docs.tempo.xyz/protocol/tip1022>
+pub struct Recipient {
```
fgimenez: `Recipient` is a helper type used in `Tip20Token`, also used in rewards.rs, but if we keep it in this module, could we move its definition after the main type? also maybe `pub(crate)` instead of `pub`

0xKitsune: Addressed in https://github.com/tempoxyz/tempo/pull/3286/changes/6c0e84554272cde3a303f70f384082ceb7a60d31, https://github.com/tempoxyz/tempo/pull/3286/changes/88326d6bd7f9a6e8e668fb17c8eacda12340192d

crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -457,11 +474,17 @@ impl TIP403Registry {
...
     /// - `IncompatiblePolicyType` — the policy is not a blacklist
     /// - `PolicyNotFound` — the policy ID does not exist (T2+)
+    /// - `VirtualAddressNotAllowed` — virtual addresses are forbidden (T3+)
     pub fn modify_policy_blacklist(
         &mut self,
         msg_sender: Address,
         call: ITIP403Registry::modifyPolicyBlacklistCall,
     ) -> Result<()> {
+        // TIP-1022: virtual addresses are forwarding aliases, not valid policy members (spec: T3+)
+        if self.storage.spec().is_t3() && is_virtual_address(call.account) {
```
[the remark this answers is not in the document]
0xKitsune: Virtual addresses require bytes `[4:14]` to exactly match `VIRTUAL_MAGIC` (`0xFDFDFDFDFDFDFDFD FDFD`) which is 80 bits of fixed entropy. The probability of anyone holding a private key for an address matching this pattern is 1 in 2⁸⁰ (~1.2 × 10²⁴), effectively zero. 

Post-T3, TIP-20 transfers resolve virtual addresses to their registered master, so tokens can't land at the literal address through normal TIP-20 paths. 

If someone did deposit tokens at a virtual-format address, the effect is equivalent to burning  the tokens are irrecoverable regardless of `burn_blocked`, and the spec explicitly calls this out as an accepted limitation for non-TIP-20 transfers.

crates/precompiles/src/tip20/rewards.rs
```diff
@@ -126,12 +127,19 @@ impl TIP20Token {
...
+    /// - `InvalidRecipient` — TIP-1022 virtual addresses are rejected
     pub fn set_reward_recipient(
         &mut self,
         msg_sender: Address,
         call: ITIP20::setRewardRecipientCall,
     ) -> Result<()> {
         self.check_not_paused()?;
+
+        // TIP-1022: reject virtual addresses as reward recipients
+        if self.storage.spec().is_t3() && tip20_registry::is_virtual_address(call.recipient) {
```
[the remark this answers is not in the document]
0xKitsune: Same probability analysis as the [previous cyclops comment](https://github.com/tempoxyz/tempo/pull/3286#discussion_r2981514591) and requires someone to have deliberately set a reward recipient to an address matching the 80-bit `VIRTUAL_MAGIC` pattern pre-T3. 

Even if it happened, the impact is limited: only the delegated reward share is affected. The holder can change their reward recipient to stop further reward accrual to the phantom address.

crates/precompiles/src/tip20/mod.rs
0xrusowsky: that was an error when generating the solidity ref impl (rust is what matters), which i addressed in [`6478133` (this PR)](https://github.com/tempoxyz/tempo/pull/3286/commits/647813323d816a61c9208b16ee040feb697e2d7b).

in the pre-virtual addresses TIP20 code we don't validate mint recipients, so i preserved that behavior in the udpated impl. However, i do wonder whether we should also call:
```rs
        self.check_not_paused()?;
        to.validate()?;
```
in mints @danrobinson @malleshpai

0xrusowsky: as discussed internally, we will do this in a follow-up PR

PR #3287, opened 2026-03-24
chore: bump revm/reth deps and adapt to EIP-8037 changes

rakita:
## Summary
- Bump reth to `2a94eed`, bump revm/alloy-evm/revm-inspectors for EIP-8037 support
- Fix `last_frame_result` signature and simplify `OutOfGas` error conversion in precompile dispatch
- Default `storage.v2` to false
- Includes additional fixes and improvements merged into the branch: fee payer signature validation, precompile policy data crafting, ValidatorConfigV2 test coverage, chainspec no_std support, bench profiling, CI improvements, and more

## Test plan
- [ ] CI passes (unit tests, integration tests, RPC matrix tests)
- [ ] Verify EVM execution with updated revm deps
- [ ] Confirm precompile dispatch error handling works correctly


crates/precompiles/src/lib.rs
```diff
@@ -89,7 +92,7 @@ pub trait Precompile {
     ///
     /// Business-logic errors are returned as reverted [`PrecompileOutput`]s with ABI-encoded
     /// error data, while fatal failures (e.g. out-of-gas) are returned as [`PrecompileError`].
```
fgimenez:
```suggestion
   /// Business-logic errors are returned as reverted [`PrecompileOutputExt`]s with ABI-encoded
   /// error data, while fatal failures (e.g. out-of-gas) are returned as [`PrecompileErrorExt`].
```

crates/precompiles/src/error.rs
```diff
@@ -160,18 +162,22 @@ impl TempoPrecompileError {
...
             .abi_encode()
             .into(),
             Self::Fatal(msg) => {
-                return Err(PrecompileError::Fatal(msg));
+                return Err(PrecompileError::Fatal(msg).into());
             }
         };
-        Ok(PrecompileOutput::new_reverted(gas, bytes))
+        Ok(PrecompileOutputExt {
+            gas: GasTracker::new(gas, 0, 0),
```
fgimenez: according to upstream API `GasTracker::new(limit, remaining, reservoir)` but we are passing here `gas` which is gas used?

crates/precompiles/src/error.rs
```diff
@@ -160,18 +161,22 @@ impl TempoPrecompileError {
             Self::ValidatorConfigV2Error(e) => e.abi_encode().into(),
             Self::AccountKeychainError(e) => e.abi_encode().into(),
             Self::OutOfGas => {
-                return Err(PrecompileError::OutOfGas);
+                return Err(PrecompileError::OutOfGas.into());
```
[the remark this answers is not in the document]
fgimenez: addressed in https://github.com/tempoxyz/tempo/pull/3287/commits/a8103928cd0ee56348c6764767aba70ae348b28a

crates/precompiles/src/error.rs
```diff
@@ -160,18 +161,22 @@ impl TempoPrecompileError {
...
             .abi_encode()
             .into(),
             Self::Fatal(msg) => {
-                return Err(PrecompileError::Fatal(msg));
+                return Err(PrecompileError::Fatal(msg).into());
             }
         };
-        Ok(PrecompileOutput::new_reverted(gas, bytes))
+        Ok(PrecompileOutputExt {
+            gas: GasTracker::new(gas, 0, 0),
```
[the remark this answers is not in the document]
fgimenez: same as https://github.com/tempoxyz/tempo/pull/3287#discussion_r2988688124, addressed in https://github.com/tempoxyz/tempo/pull/3287/commits/a8103928cd0ee56348c6764767aba70ae348b28a

bin/tempo/src/main.rs
mattsse, line 1: dont understand why we need this in this pr

crates/payload/types/src/lib.rs
```diff
@@ -115,8 +115,12 @@ impl ExecutionPayload for TempoExecutionData {
         self.block.gas_used()
     }
 
+    fn gas_limit(&self) -> u64 {
+        self.block.gas_limit()
+    }
+
     fn slot_number(&self) -> Option<u64> {
-        self.block.slot_number()
+        None
```
mattsse: why?

crates/precompiles/src/lib.rs
```diff
@@ -310,11 +313,6 @@ fn mutate_void<T: SolCall>(
 #[inline]
 fn fill_precompile_output(mut output: PrecompileOutput, storage: &StorageCtx) -> PrecompileOutput {
     output.gas_used = storage.gas_used();
-
-    // add refund only if it is not reverted
-    if !output.reverted {
-        output.gas_refunded = storage.gas_refunded();
-    }
```
mattsse, line 317: why remove this check?

commit_msg.txt
mattsse, line 1: remove?

Cargo.toml
```diff
@@ -350,15 +350,3 @@ vergen-git2 = "9.1.0"
 # reth-trie-sparse-parallel = { path = "../reth/crates/trie/sparse-parallel" }
 
 # [patch.crates-io]
-# # Commonware at HEAD after PR #3594 was merged
-# commonware-broadcast = { git = "https://github.com/commonwarexyz/monorepo", rev = "f4e49d2d8ce81b1ff594adb436e7f7560b23703c" }
```
mattsse, line 356: why remove?

mattsse (commented):
not super familiar with the revm related changes especially the error mappings/variants

crates/precompiles/src/lib.rs
fgimenez: yep, revm 37 removed `gas_refunded` from `PrecompileOutput`. #2684 handles it by encoding the refund in the `reservoir` field on T4+ and having the `PrecompileProvider` wrapper apply it via `record_refund`, does it make sense? lmk if we should move the logic here

crates/payload/types/src/lib.rs
fgimenez: this was not intended, reverted

bin/tempo/src/main.rs
fgimenez: yep not needed, during the rebase the reth-discv5 dep was removed, reverted

commit_msg.txt
fgimenez: yep leftover, done

Cargo.toml
fgimenez: reverted

crates/evm/src/evm.rs
```diff
@@ -190,7 +190,7 @@ where
                 );
             };
 
-            *gas = ResultGas::default().with_limit(tx.inner.gas_limit);
+            *gas = ResultGas::default();
```
klkvr, line 193: why do we no longer set the limit here?

crates/evm/src/lib.rs
```diff
@@ -162,6 +162,7 @@ impl ConfigureEvm for TempoEvmConfig {
                 suggested_fee_recipient: attributes.suggested_fee_recipient,
                 prev_randao: attributes.prev_randao,
                 gas_limit: attributes.gas_limit,
+                slot_number: None,
```
klkvr: should this use inner `slot_number`? we can skip setting it at the engine level but ideally we propagate everything here

crates/payload/builder/src/lib.rs
```diff
@@ -376,6 +376,7 @@ where
                         parent_beacon_block_root: attributes.parent_beacon_block_root,
                         withdrawals: attributes.withdrawals.clone().map(Into::into),
                         extra_data: attributes.extra_data().clone(),
+                        slot_number: None,
```
klkvr: same here, let's not hardcode `None` everywhere and propagate from attributes. if needed, it will be `None` there

crates/payload/types/src/attrs.rs
```diff
@@ -183,7 +183,7 @@ impl PayloadAttributes for TempoPayloadAttributes {
     }
 
     fn slot_number(&self) -> Option<u64> {
-        self.inner.slot_number()
+        None
```
klkvr: why was this reverted?

crates/precompiles/src/account_keychain/dispatch.rs
```diff
@@ -27,9 +27,9 @@ const T3_DROPPED: &[[u8; 4]] = &[IAccountKeychain::getRemainingLimitCall::SELECT
 
 impl Precompile for AccountKeychain {
     fn call(&mut self, calldata: &[u8], msg_sender: Address) -> PrecompileResult {
-        self.storage
-            .deduct_gas(input_cost(calldata.len()))
-            .map_err(|_| PrecompileError::OutOfGas)?;
+        if self.storage.deduct_gas(input_cost(calldata.len())).is_err() {
+            return Ok(PrecompileOutput::halt(PrecompileHalt::OutOfGas, 0));
+        }
```
klkvr: that seems wrong? why would we return zero reservoir on halt?

crates/precompiles/src/address_registry/dispatch.rs
```diff
@@ -6,14 +6,14 @@ use crate::{
...
 use tempo_contracts::precompiles::IAddressRegistry::IAddressRegistryCalls;
 
 impl Precompile for AddressRegistry {
     fn call(&mut self, calldata: &[u8], msg_sender: Address) -> PrecompileResult {
-        self.storage
-            .deduct_gas(input_cost(calldata.len()))
-            .map_err(|_| PrecompileError::OutOfGas)?;
+        if self.storage.deduct_gas(input_cost(calldata.len())).is_err() {
+            return Ok(PrecompileOutput::halt(PrecompileHalt::OutOfGas, 0));
+        }
```
klkvr: we should extract this into a separate fn imo

crates/precompiles/src/signature_verifier/dispatch.rs
```diff
@@ -15,14 +15,15 @@ const MAX_CALLDATA_LEN: usize =
...
+        if self.storage.deduct_gas(input_cost(calldata.len())).is_err() {
+            return Ok(PrecompileOutput::halt(PrecompileHalt::OutOfGas, 0));
+        }
 
         if calldata.len() > MAX_CALLDATA_LEN {
-            return Ok(PrecompileOutput::new_reverted(
+            return Ok(PrecompileOutput::revert(
                 self.storage.gas_used(),
                 SignatureVerifierError::invalid_format().abi_encode().into(),
+                0,
```
klkvr: same here, i'd prefer if we return the reservoir unchanged for now

crates/precompiles/src/lib.rs
klkvr: @rakita why did revm remove the refund from `PrecompileOutput`? i missed this change

crates/precompiles/src/lib.rs
```diff
@@ -392,11 +390,11 @@ pub(crate) fn dispatch_call<T>(
...
             return Ok(fill_precompile_output(
-                PrecompileOutput::new_reverted(0, Bytes::new()),
+                PrecompileOutput::revert(0, Bytes::new(), 0),
                 &storage,
             ));
         } else {
-            return Err(PrecompileError::Other(
+            return Err(PrecompileError::Fatal(
                 "Invalid input: missing function selector".into(),
             ));
```
klkvr: `Fatal` here is incorrect, `Other` was getting mapped to a halt. `Fatal` woulc cause entire block execution to crash as soon as any tx triggers this error

crates/revm/src/handler.rs
```diff
@@ -1988,7 +1994,7 @@ fn check_gas_limit(
     tx: &TempoTxEnv,
     adjusted_gas: &InitialAndFloorGas,
 ) -> Option<FrameResult> {
-    if spec.is_t0() && tx.gas_limit() < adjusted_gas.initial_gas {
+    if spec.is_t0() && tx.gas_limit() < adjusted_gas.initial_total_gas {
```
klkvr, line 1998: hmm @rakita so for state gas charges both `initial_total_gas` and `initial_state_gas` would have to be implemented? why not just have `initial_regular_gas`

crates/revm/src/handler.rs
```diff
@@ -1690,7 +1695,8 @@ where
         evm: &mut TempoEvm<DB, I>,
     ) -> Result<ValidationContext, EVMError<DB::Error, TempoInvalidTransaction>> {
         self.validate(evm)?;
-        self.pre_execution(evm)?;
+        let mut init_and_floor_gas = InitialAndFloorGas::default();
+        self.pre_execution(evm, &mut init_and_floor_gas)?;
```
klkvr: instead of creating empty `InitialAndFloorGas` this should use returned value from `validate` call above

klkvr (changes requested):

crates/evm/src/evm.rs
fgimenez: yep, `ResultGas` no longer has a `limit` field in revm 37, `with_limit` was removed. ~the old `used()` needed limit for its calculation (`limit - spent + refund`), but the new one is just `max(spent - refund, floor)`~ even in revm 36, `used()` was `max(spent - refund, floor)` and didn't use `limit`, it was only used by `remaining()` which nothing calls on system tx results. `ResultGas::default()` (all zeros) correctly gives `tx_gas_used() = 0` for system txs, is this correct @rakita

crates/evm/src/lib.rs
fgimenez: good call, now propagating `attributes.slot_number` in both `next_evm_env` and the payload builder

crates/payload/builder/src/lib.rs
fgimenez: done

crates/payload/types/src/attrs.rs
fgimenez: yep, this was not intended, fixed

crates/precompiles/src/account_keychain/dispatch.rs
fgimenez: good point, `reservoir: 0` on halt loses the reservoir gas. #2684 doesn't address this specific case either (the `deduct_gas` OOG halt at dispatch entry is unchanged there). the `Precompile::call` trait doesn't have access to the reservoir, so aiui it'd need to be threaded through or handled at the `tempo_precompile!` macro level where `$input.reservoir` is available, wdyt? any thoughts @rakita?

crates/precompiles/src/address_registry/dispatch.rs
fgimenez: done, extracted `charge_input_cost` helper in `lib.rs` and replaced all 11 dispatch files

crates/precompiles/src/lib.rs
fgimenez: good catch, fixed here https://github.com/tempoxyz/tempo/pull/3287/commits/50fd07550c4d21aca04469910f4b198d2d898008

crates/revm/src/handler.rs
fgimenez: indeed done here https://github.com/tempoxyz/tempo/pull/3287/commits/261225adc9078da2ccd8769000d9707ed5c6999a

klkvr (approved):

klkvr:
@copilot resolve the merge conflicts in this pull request

crates/precompiles/src/account_keychain/dispatch.rs
rakita: It looks like it needs fixing, the reservoir needs to be returned

PR #3297, opened 2026-03-24
feat: bump reth to payload attrs changes

klkvr:
bumps reth to commit including https://github.com/paradigmxyz/reth/pull/23202 refactor

this now allows replacing the `payload_builder_handle.send_new_payload` with proper FCU

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12180,7 +12180,6 @@ dependencies = [
  "rand 0.9.2",
  "reqwest 0.13.2",
  "reth-e2e-test-utils",
- "reth-engine-local",
  "reth-errors",
  "reth-ethereum",
  "reth-evm",
--- a/crates/node/Cargo.toml
+++ b/crates/node/Cargo.toml
@@ -39,7 +39,6 @@ reth-rpc-builder.workspace = true
 reth-node-api.workspace = true
 reth-rpc-eth-types.workspace = true
 reth-node-ethereum.workspace = true
-reth-engine-local.workspace = true
 
 alloy-serde.workspace = true
 alloy-eips.workspace = true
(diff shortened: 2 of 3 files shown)


crates/payload/types/src/attrs.rs
```diff
@@ -125,37 +134,15 @@ impl From<EthPayloadBuilderAttributes> for TempoPayloadBuilderAttributes {
...
-        self.inner.parent()
+impl PayloadAttributes for TempoPayloadAttributes {
+    fn payload_id(&self, parent_hash: &B256) -> PayloadId {
+        // XXX: derives the payload ID from the parent so that
+        // overlong payload builds will eventually succeed on the
+        // next iteration: if all other nodes take equally as long,
+        // the consensus engine will kill the proposal task. Then eventually
+        // consensus will circle back to an earlier node, which then
+        // has the chance of picking up the old payload.
+        payload_id_from_block_hash(parent_hash)
```
klkvr, line 147: this is now default behavior for all attributes which i think should be fine

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -905,14 +895,6 @@ async fn verify_header_extra_data(
     Ok(())
 }
 
-/// Constructs a [`PayloadId`] from the first 8 bytes of `block_hash`.
-fn payload_id_from_block_hash(block_hash: &B256) -> PayloadId {
```
SuperFluffy, line 909: I'm not sure we really every used this.

But is ID bit handled by the EL now, not in case block building takes too long and we need to pick up from an earlier build?

klkvr: no it's still done like this, it's just that payload attributes type is now responsible for generating its own `PayloadId` so i've moved this deeper https://github.com/tempoxyz/tempo/pull/3297#discussion_r2983476589

PR #3300, opened 2026-03-24
chore: remaining validator consensus commands

hamdiallam:
v2 only operations
* set-validator-ip-address
* set-validator-fee-recipient
* transfer-validator-ownership
* deactivate-validator

v1/v2 auto-support
* validator
* validators-info

Confirm prompt on any onchain action with a summary of the function + calldata

Tested against localnet. Both self serve and contract owner

### DIFF
--- a/bin/tempo/src/tempo_cmd.rs
+++ b/bin/tempo/src/tempo_cmd.rs
@@ -140,7 +140,7 @@ pub(crate) enum ConsensusSubcommand {
     SetValidatorFeeRecipient(SetValidatorFeeRecipient),
     /// Transfer validator ownership
     TransferValidatorOwnership(TransferValidatorOwnership),
-    /// Look up a validator by address or index.
+    /// Look up a validator by etheruem address, e25519 public key, or index.
     Validator(ValidatorInfo),
     /// Query validator info from the previous epoch's DKG outcome and current contract state.
     ValidatorsInfo(ValidatorsInfo),
@@ -472,7 +472,7 @@ impl AddValidator {
 
 #[derive(Debug, clap::Args)]
 pub(crate) struct TransferValidatorOwnership {
-    /// Validator ethereum address, ed25519 pubkey, or index
+    /// Validator ethereum address, ed25519 public key, or index
     #[arg(long)]
     id: ValidatorId,
 
@@ -670,7 +670,7 @@ impl CreateRotateValidatorSignatureArgs {
 
 #[derive(Debug, clap::Args)]
 pub(crate) struct SetValidatorIpAddress {
-    /// Validator ethereum address, ed25519 pubkey, or index
+    /// Validator ethereum address, ed25519 public key, or index
     #[arg(long)]
     id: ValidatorId,
     /// The inbound address for the validator.
@@ -721,7 +721,7 @@ impl SetValidatorIpAddress {
 
 #[derive(Debug, clap::Args)]
 pub(crate) struct DeactivateValidator {
-    /// Validator ethereum address, ed25519 pubkey, or index
+    /// Validator ethereum address, ed25519 public key, or index
     #[arg(long)]
     id: ValidatorId,
 


bin/tempo/src/tempo_cmd.rs
```diff
@@ -116,12 +116,18 @@ impl ExtendedCommand for TempoSubcommand {
...
+    /// Transfer validator ownership
+    TransferValidatorOwnership(TransferValidatorOwnership),
     /// Create an ed25519 signature for `addValidator`.
     CreateAddValidatorSignature(CreateAddValidatorSignatureArgs),
     /// Rotate a validator to a new identity.
     RotateValidator(RotateValidator),
     /// Create an ed25519 signature for `rotateValidator`.
     CreateRotateValidatorSignature(CreateRotateValidatorSignatureArgs),
+    /// Set the validator Ip Address
+    SetValidatorIPAddress(SetValidatorIPAddress),
```
SuperFluffy: Use `SetValidatorIpAddress` - Rust convention.

bin/tempo/src/tempo_cmd.rs
```diff
@@ -490,6 +572,160 @@ impl CreateRotateValidatorSignatureArgs {
...
+pub(crate) struct SetValidatorIPAddress {
+    /// The validator's address
+    #[arg(long, value_name = "ETHEREUM_ADDRESS")]
+    validator_address: Address,
+    /// The inbound address for the validator.
+    #[arg(long, value_name = "IP:PORT")]
+    ingress: SocketAddr,
+    /// The outbound address for the validator.
+    #[arg(long, value_name = "IP")]
+    egress: IpAddr,
```
SuperFluffy: We should allow making these `Option`s - if you don't specify one, then the other will set based on the latest state.

bin/tempo/src/tempo_cmd.rs
```diff
@@ -490,6 +572,160 @@ impl CreateRotateValidatorSignatureArgs {
     }
 }
 
+#[derive(Debug, clap::Args)]
+pub(crate) struct SetValidatorIPAddress {
```
SuperFluffy:
```suggestion
pub(crate) struct SetValidatorIpAddresses {
```

bin/tempo/src/tempo_cmd.rs
```diff
@@ -490,6 +572,160 @@ impl CreateRotateValidatorSignatureArgs {
...
+            .wrap_err("failed constructing async runtime")?
+            .block_on(self.run_async())
+    }
+
+    async fn run_async(self) -> eyre::Result<()> {
+        let private_key_bytes =
+            std::fs::read(&self.submit.private_key).wrap_err("failed reading private key")?;
+        let private_key =
+            B256::try_from(private_key_bytes.as_slice()).wrap_err("invalid private key")?;
+        let signer = PrivateKeySigner::from_bytes(&private_key)?;
```
SuperFluffy: Factor this into a separate method?

bin/tempo/src/tempo_cmd.rs
```diff
@@ -490,6 +572,160 @@ impl CreateRotateValidatorSignatureArgs {
...
+    submit: ValidatorTransactionArgs,
+}
+
+impl SetValidatorFeeRecipient {
+    fn run(self) -> eyre::Result<()> {
+        tokio::runtime::Builder::new_current_thread()
+            .enable_all()
+            .build()
+            .wrap_err("failed constructing async runtime")?
+            .block_on(self.run_async())
```
SuperFluffy: Feels like you could also factor this `run` method into a self-standing `fn run(impl Future<Output = eyre::Result<()>>)`

bin/tempo/src/tempo_cmd.rs
```diff
@@ -134,9 +140,12 @@ impl ConsensusSubcommand {
     fn run(self) -> eyre::Result<()> {
         match self {
             Self::AddValidator(args) => args.run(),
+            Self::TransferValidatorOwnership(args) => args.run(),
```
SuperFluffy: Sort alphabetically? :-)

bin/tempo/src/tempo_cmd.rs
hamdiallam: #[tokio::main(flavor = "current_thread")]
    async fn run(self) -> eyre::Result<()> {
    }
    
There's a tikio macro for this which is pretty clean

bin/tempo/src/tempo_cmd.rs
```diff
@@ -20,15 +20,17 @@ use commonware_cryptography::{
 };
 use commonware_math::algebra::Random as _;
 use commonware_utils::NZU64;
-use eyre::{OptionExt as _, Report, WrapErr as _, eyre};
+use eyre::{Ok, OptionExt as _, Report, WrapErr as _, eyre};
```
SuperFluffy: Don't like this `Ok` method shadowing the normal `Ok` in prelude. Is it necessary?

bin/tempo/src/tempo_cmd.rs
```diff
@@ -116,34 +118,68 @@ impl ExtendedCommand for TempoSubcommand {
...
+    SetValidatorFeeRecipient(SetValidatorFeeRecipient),
+    /// Transfer validator ownership
+    TransferValidatorOwnership(TransferValidatorOwnership),
     /// Query validator info from the previous epoch's DKG outcome and current contract state.
     ValidatorsInfo(ValidatorsInfo),
 }
 
 impl ConsensusSubcommand {
-    fn run(self) -> eyre::Result<()> {
+    #[tokio::main(flavor = "current_thread")]
```
SuperFluffy: Weird seeing this here, but cool. Works for me.

@mattsse Is that fine (just factors out manual construction of tokio) or is there another blessed way of running these commands from reth cli context?

bin/tempo/src/tempo_cmd.rs
```diff
@@ -264,6 +346,68 @@ impl AddValidator {
...
+        let validator = read_validator_from_contract(&provider, self.validator_address).await?;
+
+        let calldata = IValidatorConfigV2::transferValidatorOwnershipCall {
+            idx: validator.index,
+            newAddress: new_validator_address,
+        };
+
+        self.submit.confirm(&serde_json::json!({
+            "to": VALIDATOR_CONFIG_V2_ADDRESS,
+            "function": "transferValidatorOwnership",
```
SuperFluffy: Let's change this to `"signature"` and use the sol cal `SIGNATURE` associated const.

bin/tempo/src/tempo_cmd.rs
```diff
@@ -264,6 +346,68 @@ impl AddValidator {
...
+        let new_validator_address = new_signer.address();
+
+        let validator = read_validator_from_contract(&provider, self.validator_address).await?;
+
+        let calldata = IValidatorConfigV2::transferValidatorOwnershipCall {
+            idx: validator.index,
+            newAddress: new_validator_address,
+        };
+
+        self.submit.confirm(&serde_json::json!({
```
SuperFluffy: Can we type this out to avoid repetition? We should be able to format this by passing a `fn f<T: SolCall>(t: T)`. Would clean this up nicely and avoid mistakes.

bin/tempo/src/tempo_cmd.rs
mattsse: this creates a new runtime, don't think we need this?

mattsse: can use 

https://github.com/paradigmxyz/reth/blob/7d31bb176cb90f328fe32763215dccce13da1948/crates/cli/runner/src/lib.rs#L191-L194

I guess, but unsure whwere this is called from

SuperFluffy: Double checked: https://github.com/tempoxyz/tempo/blob/f2ecd7be2c592258315ebd47e4424b1a8fa47698/bin/tempo/src/tempo_cmd.rs#L96-L102

This is how we do it elsewhere

bin/tempo/src/tempo_cmd.rs
hamdiallam: ahh no, auto-import slipped this in

bin/tempo/src/tempo_cmd.rs
hamdiallam: Ah nice, yea parameterizing this is clean

bin/tempo/src/tempo_cmd.rs
```diff
@@ -116,34 +121,87 @@ impl ExtendedCommand for TempoSubcommand {
...
     CreateRotateValidatorSignature(CreateRotateValidatorSignatureArgs),
     /// Generates an ed25519 signing key pair to be used in consensus.
     GeneratePrivateKey(GeneratePrivateKey),
-    /// Calculates the public key from an ed25519 signing key.
-    CalculatePublicKey(CalculatePublicKey),
+    /// Rotate a validator to a new identity.
+    RotateValidator(RotateValidator),
+    /// Set the validator Ip Address
+    SetValidatorIpAddress(SetValidatorIpAddress),
+    /// SEt the validator fee recipient
```
SuperFluffy: typo

bin/tempo/src/tempo_cmd.rs
```diff
@@ -187,66 +245,119 @@ pub(crate) struct SignatureArgs {
...
+}
+
+impl ValidatorTransactionArgs {
+    fn confirm<T: SolCall + Serialize>(&self, call: &T) -> eyre::Result<()> {
+        eprintln!(
+            "{}",
+            &serde_json::json!({
+                "to": VALIDATOR_CONFIG_V2_ADDRESS,
+                "signature": T::SIGNATURE,
+                "call": serde_json::to_value(call)?,
```
SuperFluffy: I think you don't need this - just:
```suggestion
                "call": call,
```

bin/tempo/src/tempo_cmd.rs
```diff
@@ -187,66 +245,119 @@ pub(crate) struct SignatureArgs {
...
+        eprint!("\nSubmit this transaction? [y/N] ");
+        std::io::stderr().flush()?;
+
+        let mut input = String::new();
+        std::io::stdin().read_line(&mut input)?;
+
+        if !matches!(input.trim(), "y" | "Y" | "yes" | "YES") {
+            Err(eyre!("transaction cancelled by user"))
+        } else {
+            Ok(())
```
SuperFluffy, line 383: neat

bin/tempo/src/tempo_cmd.rs
```diff
@@ -187,66 +245,119 @@ pub(crate) struct SignatureArgs {
...
+    yes: bool,
+}
+
+impl ValidatorTransactionArgs {
+    fn confirm<T: SolCall + Serialize>(&self, call: &T) -> eyre::Result<()> {
+        eprintln!(
+            "{}",
+            &serde_json::json!({
+                "to": VALIDATOR_CONFIG_V2_ADDRESS,
+                "signature": T::SIGNATURE,
```
SuperFluffy, line 365: cool

bin/tempo/src/tempo_cmd.rs
```diff
@@ -490,6 +592,102 @@ impl CreateRotateValidatorSignatureArgs {
     }
 }
 
+#[derive(Debug, clap::Args)]
+pub(crate) struct SetValidatorIpAddress {
+    /// The validator's address
+    #[arg(long, value_name = "ETHEREUM_ADDRESS")]
+    validator_address: Address,
```
SuperFluffy: It'd be pretty easy to make this just `validator` and do this:

```suggestion
enum ValidatorId {
    Address(Address),
    Index(u16),
    PublicKey(PublicKey),
}

impl FromStr for ValidatorId {
    type Err = eyre::Report;
    fn from_str(s: &str) -> Result<Self, Self::Err> {
        if let Ok(idx) = s.parse::<u16>() {
            return Ok(Self::Index(idx)); // maybe also account for hex int?
        }
        if let Ok(address) = s.parse::<Address>()> {
            return Ok(Self::Address(address));
        }
        if let Ok(bytes) = const_hex::decode(s)
        && let Ok(public_key) = PublicKey::decode(&bytes) // cw DecodeExt
        {
            return Ok(Self::PublicKey(public_key));
        }
        Err(eyre!("validator identifier must be integer, address, or ed25519 public key"))
    }
}
```

bin/tempo/src/tempo_cmd.rs
```diff
@@ -540,12 +739,96 @@ impl CalculatePublicKey {
...
+
+    /// Look up by validator address.
+    #[arg(long, conflicts_with = "index")]
+    address: Option<Address>,
+
+    /// Look up by validator index.
+    #[arg(long, conflicts_with = "address")]
+    index: Option<u64>,
+
+    /// Look up a validator on the v1 config contract
```
SuperFluffy: You mean `v2`?

bin/tempo/src/tempo_cmd.rs
```diff
@@ -540,12 +739,96 @@ impl CalculatePublicKey {
...
+    #[arg(long, default_value = "false")]
```
SuperFluffy: default values for bool are always false

bin/tempo/src/tempo_cmd.rs
```diff
@@ -540,12 +739,96 @@ impl CalculatePublicKey {
...
+    #[arg(long, default_value = "https://rpc.presto.tempo.xyz")]
+    rpc_url: String,
+
+    /// Look up by validator address.
+    #[arg(long, conflicts_with = "index")]
+    address: Option<Address>,
+
+    /// Look up by validator index.
+    #[arg(long, conflicts_with = "address")]
+    index: Option<u64>,
```
SuperFluffy: Look at the FromStr example I gave above - we can make this much neater! :-)

bin/tempo/src/tempo_cmd.rs
```diff
@@ -540,12 +739,96 @@ impl CalculatePublicKey {
...
+
+impl ValidatorInfo {
+    async fn run(self) -> eyre::Result<()> {
+        let provider = ProviderBuilder::new_with_network::<TempoNetwork>()
+            .connect(&self.rpc_url)
+            .await
+            .wrap_err("failed to connect to RPC")?;
+
+        // Once migrated, this block will be removed
+        if !self.v2 {
```
SuperFluffy: IMO just hide this from users? You can use `v2.getInitializedHeight` to figure out if v2 was active at the last epoch boundary (whether we use v1 or v2 is decided on the boundary's parent; also check `isInitialized` if `getInitializedHeight == 0`).

Then you can just read v1 or v2 depending on that result.

bin/tempo/src/tempo_cmd.rs
```diff
@@ -596,21 +879,17 @@ pub(crate) struct ValidatorsInfo {
...
 
-    /// Whethr to include historic validators (deactivated and not in the current committee).
-    #[arg(long)]
+    /// Whether to include historic validators (deactivated and not in the current committee).
+    #[arg(long, conflicts_with = "v2")]
     with_historic: bool,
+
+    /// Use the v2 config contract (active validators only).
+    #[arg(long)]
+    v2: bool,
```
SuperFluffy: likewise here - we can hide this from users so that they need not worry about that.

bin/tempo/src/tempo_cmd.rs
```diff
@@ -116,34 +121,87 @@ impl ExtendedCommand for TempoSubcommand {
...
+    TransferValidatorOwnership(TransferValidatorOwnership),
+    /// Query a single validator by address or index.
+    ValidatorInfo(ValidatorInfo),
     /// Query validator info from the previous epoch's DKG outcome and current contract state.
     ValidatorsInfo(ValidatorsInfo),
 }
 
 impl ConsensusSubcommand {
-    fn run(self) -> eyre::Result<()> {
+    async fn run(self) -> eyre::Result<()> {
```
SuperFluffy, line 150: Wrwapping the command for deactivating validators is also useful.

bin/tempo/src/tempo_cmd.rs
```diff
@@ -116,34 +121,87 @@ impl ExtendedCommand for TempoSubcommand {
...
+    /// Rotate a validator to a new identity.
+    RotateValidator(RotateValidator),
+    /// Set the validator Ip Address
+    SetValidatorIpAddress(SetValidatorIpAddress),
+    /// SEt the validator fee recipient
+    SetValidatorFeeRecipient(SetValidatorFeeRecipient),
+    /// Transfer validator ownership
+    TransferValidatorOwnership(TransferValidatorOwnership),
+    /// Query a single validator by address or index.
+    ValidatorInfo(ValidatorInfo),
```
SuperFluffy: What if we renamed this just `Validator` -> `tempo consensus validator <idx>`

bin/tempo/src/tempo_cmd.rs
SuperFluffy: IMO make these positional arguments so that you can do:
```shell
tempo consensus validator <identifier>
```

bin/tempo/src/tempo_cmd.rs
hamdiallam: Ahh yea this is much cleaner, nice

bin/tempo/src/tempo_cmd.rs
hamdiallam: Ah nice this is neat

bin/tempo/src/tempo_cmd.rs
hamdiallam: ah nice catch, I missed this function

bin/tempo/src/tempo_cmd.rs
```diff
@@ -116,32 +122,144 @@ impl ExtendedCommand for TempoSubcommand {
...
+
+impl FromStr for ValidatorId {
+    type Err = eyre::Report;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if let Ok(idx) = s.parse::<u64>() {
+            Ok(Self::Index(idx))
+        } else if let Ok(address) = s.parse::<Address>() {
+            Ok(Self::Address(address))
+        } else if let Ok(pubkey) = s.parse::<B256>() {
```
SuperFluffy, line 183: Ah, indeed. I forgot

bin/tempo/src/tempo_cmd.rs
```diff
@@ -116,32 +122,144 @@ impl ExtendedCommand for TempoSubcommand {
...
+    }
+
+    let height = IValidatorConfigV2::getInitializedAtHeightCall::abi_decode_returns(&resp)
+        .wrap_err("failed to decode v2 initialized height response")?;
+
+    let block_number = provider
+        .get_block_number()
+        .await
+        .wrap_err("failed to fetch block number")?;
+    let initialized = height >= block_number;
```
SuperFluffy: Note: this is not sufficient if the contract is activated at genesis - for that you need to also check `isInit`.

bin/tempo/src/tempo_cmd.rs
```diff
@@ -490,6 +652,142 @@ impl CreateRotateValidatorSignatureArgs {
     }
 }
 
+#[derive(Debug, clap::Args)]
+pub(crate) struct SetValidatorIpAddress {
+    /// The validator's address
+    #[arg(long, value_name = "ETHEREUM_ADDRESS")]
+    validator_address: Address,
```
SuperFluffy: Use the `ValidatorId` you defined above?

bin/tempo/src/tempo_cmd.rs
```diff
@@ -490,6 +652,142 @@ impl CreateRotateValidatorSignatureArgs {
...
+
+        Ok(())
+    }
+}
+
+#[derive(Debug, clap::Args)]
+pub(crate) struct DeactivateValidator {
+    /// The validator's address
+    #[arg(long, value_name = "ETHEREUM_ADDRESS")]
+    validator_address: Address,
```
SuperFluffy: Use `ValidatorId` from above?

bin/tempo/src/tempo_cmd.rs
```diff
@@ -490,6 +652,142 @@ impl CreateRotateValidatorSignatureArgs {
...
+
+        Ok(())
+    }
+}
+
+#[derive(Debug, clap::Args)]
+pub(crate) struct SetValidatorFeeRecipient {
+    /// The validator's address
+    #[arg(long, value_name = "ETHEREUM_ADDRESS")]
+    validator_address: Address,
```
SuperFluffy: `ValidatorId`?

bin/tempo/src/tempo_cmd.rs
hamdiallam: Ah good call, updated

SuperFluffy: Wrong order? It's initialized if `initHeight <= currentHeight`

hamdiallam: Nope totally right 🫠

PR #3303, opened 2026-03-25
docs(tip-1037): extended memo support

decofe:
Adds TIP-1037: Extended Memo Support. Introduces `WithLongMemo` variants of transfer, transferFrom, mint, and burn that accept dynamic `bytes memo` (up to 256 bytes). Events emit both `bytes32 indexed memoHash` for topic filtering and raw `bytes memo` for readability. Cap set to 256 bytes to discourage ordinals-style inscription abuse.

Prompted by: dan

### DIFF
--- a/tips/tip-1037.md
+++ b/tips/tip-1037.md
@@ -1,0 +1,165 @@
+---
+id: TIP-1037
+title: Extended Memo Support
+description: Adds variable-length memo support to TIP-20 transfers, mints, and burns via new functions that accept dynamic bytes alongside a bytes32 memo hash for efficient log indexing.
+authors: Daniel Robinson (@danrobinson)
+status: Draft
+related: TIP-20
+protocolVersion: TBD
+---
+
+# TIP-1037: Extended Memo Support
+
+## Abstract
+
+TIP-1037 extends the TIP-20 memo system beyond the current 32-byte limit. New `WithLongMemo` variants of `transfer`, `transferFrom`, `mint`, and `burn` accept a dynamic `bytes memo` of up to 1024 bytes. Events emit both a `bytes32 indexed memoHash` (the keccak256 of the memo) for efficient topic-based filtering and the raw `bytes memo` in log data for full readability. The existing 32-byte `WithMemo` functions remain unchanged.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


tips/tip-1037.md
```diff
@@ -0,0 +1,76 @@
...
+```
+
+These overload the existing `bytes32` variants — Solidity resolves them by parameter type, and they produce distinct function selectors.
+
+## Behavior
+
+- Reverts with `MemoTooLong()` if `memo.length > 256`. Empty memos are permitted.
+- Transfer, allowance, policy, and pause logic is identical to the existing `bytes32` variants.
+- Events set `memoHash = keccak256(memo)`, computed by the contract (never caller-supplied).
+- For a max-length 256-byte memo, the additional gas cost beyond a standard transfer is approximately 3,500 gas.
```
malleshpai, line 57: are we sure about this number?  should we add a line with something like should be determined based on benchmarking?

tips/tip-1037.md
```diff
@@ -0,0 +1,76 @@
+---
+id: TIP-1037
+title: Extended Memo Support
+description: Variable-length memo support for TIP-20 transfers, mints, and burns.
+authors: Daniel Robinson (@danrobinson)
+status: Draft
+related: TIP-20, TIP-1035
```
malleshpai, line 7: add TIP-1022

malleshpai: not sure where this should go in the TIP but events generated upon transfers to virtual addresses with bytes memos should be properly updated

tips/tip-1037.md
```diff
@@ -0,0 +1,76 @@
...
+---
+
+# Invariants
+
+1. Any call with `memo.length > 256` MUST revert with `MemoTooLong()`.
+2. `memoHash` in every `TransferWithBytesMemo` event MUST equal `keccak256(memo)`.
+3. Transfer behavior (balances, allowances, policies, pause) MUST be identical to the corresponding `bytes32` variant.
+4. Existing `bytes32` memo functions and `TransferWithMemo(address,address,uint256,bytes32)` event MUST remain unchanged.
+
+## Critical Test Cases
```
malleshpai, line 69: should add one to test admission into payments lane?

PR #3309, opened 2026-03-25
refactor: request proposals through FCU requests

klkvr:
Changes CL logic to request proposals through FCU requests with payload attributes, like it's done by other chains.

This would allow to inherit reth optimizations we're working on that are sharing caches and other perf infrastructure from engine with payload builder directly

### DIFF
--- a/crates/commonware-node/src/consensus/application/actor.rs
+++ b/crates/commonware-node/src/consensus/application/actor.rs
@@ -28,11 +28,7 @@ use prometheus_client::metrics::counter::Counter;
 
 use commonware_utils::{SystemTimeExt, channel::oneshot};
 use eyre::{OptionExt as _, WrapErr as _, bail, ensure, eyre};
-use futures::{
-    StreamExt as _, TryFutureExt as _,
-    channel::mpsc,
-    future::{ready, try_join},
-};
+use futures::{StreamExt as _, TryFutureExt as _, channel::mpsc, future::try_join};
 use rand_08::{CryptoRng, Rng};
 use reth_ethereum::chainspec::EthChainSpec as _;
 use reth_node_builder::{Block as _, BuiltPayload, ConsensusEngineHandle};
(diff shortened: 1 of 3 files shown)


crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -569,15 +560,15 @@ impl Inner<Init> {
 
         let interrupt_handle = attrs.interrupt_handle().clone();
 
-        let payload_id = self
-            .execution_node
-            .payload_builder_handle
-            .send_new_payload(parent_hash, attrs)
-            .pace(&context, Duration::from_millis(20))
```
shekhirin, line 576: do we not need this anymore for deterministic runtime?

shekhirin: I guess we don't because we already do it in `canonicalize`?

klkvr: yep

crates/commonware-node/src/executor/ingress.rs
```diff
@@ -75,7 +79,8 @@ pub(super) enum Command {
 pub(super) struct CanonicalizeHead {
     pub(super) height: Height,
     pub(super) digest: Digest,
-    pub(super) ack: oneshot::Sender<()>,
```
SuperFluffy, line 78: What does it mean if it returns `None`?

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -569,15 +560,15 @@ impl Inner<Init> {
...
-            .await
-            .map_err(|_| eyre!("channel was closed before a response was returned"))
-            .and_then(|ret| ret.wrap_err("execution layer rejected request"))
-            .wrap_err("failed requesting new payload from the execution layer")?;
+        let payload_id = ready(self.state.executor.canonicalize_head(
```
SuperFluffy: Please provide a new command for this so that becomes clearer (for example):
```suggest
        let payload_id = ready(self.state.executor.build_proposal(
```

crates/commonware-node/src/executor/ingress.rs
```diff
@@ -19,12 +21,14 @@ impl Mailbox {
         &self,
         height: Height,
         digest: Digest,
-    ) -> eyre::Result<oneshot::Receiver<()>> {
+        attributes: Option<TempoPayloadAttributes>,
```
SuperFluffy: Mentioned in the application actor: I would like not overloading this here because it's confusing (you need intimate knowledge of how FCU works).

Let's create a second method && message type for this.

crates/commonware-node/src/executor/ingress.rs
klkvr: means there was no attributes or FCU failed

PR #3310, opened 2026-03-25
feat(consensus): proposer takes fee recipient from validator config v2

decofe:
When the V2 contract is active (T2 hardfork + initialized), the proposer
reads its fee recipient from the contract by looking up its own validator entry
by public key. If the on-chain fee recipient is `Address::ZERO`, falls back to
the CLI `--consensus.fee-recipient` (which itself now defaults to
`Address::ZERO` and is no longer required).

Subscribes to `canonical_state_stream()` (a `broadcast::Receiver`) *before*
the first read to avoid missing updates between a failed read and subscribing.
On failure, waits for a canonical event whose committed chain segment contains
the parent block at the expected height with a matching hash, then retries. The
outer `select!` in `handle_propose` races this against `response.closed()` so
consensus can time out the proposal.

Includes an e2e test (`v2_at_genesis::fee_recipient`) that sets a non-zero fee
recipient in the V2 contract at genesis and verifies produced blocks use it as
their beneficiary.

Prompted by: janis


crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -619,6 +621,65 @@ impl Inner<Init> {
...
+            Err(error) => {
+                warn!(
+                    %error,
+                    "failed reading proposer fee recipient from validator \
+                     config v2; waiting for canonical state events",
+                );
+            }
+        }
+
+        loop {
```
Zygimantass, line 655: do we need a sleep in this loop? I assume not, because we wait for canonical_events stream?

SuperFluffy: No, and yeah - that's the reason.

PR #3323, opened 2026-03-25
feat(account-keychain): implement TIP-1011 T3 semantics

legion2002:
## Overview

Implements TIP-1011 end-to-end behind the T3 gate: transaction/primitives encoding, precompile ABI and dispatch, `AccountKeychain` state transitions, runtime validation in the REVM handler, and txpool-side validation.

Pre-T3 behavior remains backward-compatible by keeping legacy `authorizeKey` / `getRemainingLimit` selector behavior and rejecting TIP-1011-only fields (`period`, `allowed_calls`) before activation.

## Review Options

Reviewers can use either path:
- Review the stacked PRs for a cleaner, layered review
- Review this PR directly if you want the full end-to-end diff in one place

We are keeping this PR open so Cyclops can continue to run on the full integrated diff.

We will also keep the final diff of this PR equivalent to the final diff of the stacked PRs.

## Stacked PRs

- [#3384](https://github.com/tempoxyz/tempo/pull/3384) `feat(primitives): add TIP-1011 wire format and ABI`
- [#3385](https://github.com/tempoxyz/tempo/pull/3385) `feat(precompiles): implement TIP-1011 restrictions`
- [#3386](https://github.com/tempoxyz/tempo/pull/3386) `feat(node): enforce TIP-1011 restrictions`

Stack structure:
- `#3384` is the base PR on `main`
- `#3385` stacks on `#3384`
- `#3386` stacks on `#3385`

## What Lands Here

This PR includes the whole TIP-1011 change together:
- periodic spending limits
- target / selector / recipient call scoping
- new `AccountKeychain` ABI methods and readers
- precompile storage and mutation logic
- handler-side enforcement and intrinsic gas changes
- txpool validation updates

If you want to understand how the ideas interact across layers, this PR is the easiest place to see the full picture.


crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -21,17 +21,95 @@ pub struct TokenLimit {
 
     /// Maximum spending amount for this token (enforced over the key's lifetime)
     pub limit: U256,
+
+    /// Period duration in seconds.
+    ///
+    /// `0` means one-time limit. `>0` means the limit resets periodically.
+    pub period: u64,
```
legion2002, line 29: Fixed in commit 69aefaabf. `TokenLimit` now encodes canonically as `[token, limit]` when `period == 0`, and decode rejects explicit 3-field `period=0` payloads. This removes the pre-T3 wire-shape split and preserves signature-hash compatibility for legacy 2-field payloads.

crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -60,6 +138,12 @@ pub struct KeyAuthorization {
     /// - `Some([])` = no spending allowed (enforce_limits=true but no tokens allowed)
     /// - `Some([TokenLimit{...}])` = specific limits enforced
     pub limits: Option<Vec<TokenLimit>>,
+
+    /// Optional call scopes for this key.
+    /// - `None` (RLP 0x80) = unrestricted calls
+    /// - `Some([])` = scoped mode with no allowed calls (deny-all)
+    /// - `Some([CallScope{...}])` = explicit target/selector scope list
+    pub allowed_calls: Option<Vec<CallScope>>,
```
legion2002, line 193: Fixed in commit 69aefaabf. `KeyAuthorization` now has custom decode logic that rejects an explicit sixth-field `allowed_calls = None` encoding (`0x80`) on the wire; unrestricted mode must be represented by omission of the trailing field.

crates/revm/src/handler.rs
```diff
@@ -126,6 +126,35 @@ fn tempo_signature_verification_gas(signature: &TempoSignature) -> u64 {
     }
 }
 
+#[inline]
+fn key_authorization_scope_slots(
```
legion2002: Fixed in commit 69aefaabf. Intrinsic gas accounting now charges periodic limits as 4 slots per limit on T3, and `key_authorization_scope_slots` now accounts for `Set` write/insert costs (targets/selectors/recipients) instead of 1 slot per item.

crates/transaction-pool/src/validator.rs
```diff
@@ -199,6 +303,27 @@ where
...
+                    .limits
+                    .as_ref()
+                    .is_some_and(|limits| limits.iter().any(|limit| limit.period != 0))
+                {
+                    return Ok(Err(TempoPoolTransactionError::Keychain(
+                        "periodic token limits are not active before T3",
+                    )));
+                }
+
+                if auth.allowed_calls.is_some() {
```
legion2002: Fixed in commit 69aefaabf. Txpool now validates call scopes in `validate_against_keychain` for T3+ both for inline `key_authorization` scope constraints and for already-authorized keys via `AccountKeychain::validate_call_scope_for_transaction`.

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -331,16 +498,181 @@ impl AccountKeychain {
...
+        if !self.storage.spec().is_t3() {
+            return Ok(Vec::new());
+        }
+
+        let account_key = Self::account_key(call.account, call.keyId);
+        let mode = self.key_scopes[account_key].mode.read()?;
+        if mode == 0 {
+            return Ok(Vec::new());
+        }
+        if mode == 2 {
```
legion2002, line 636: Already fixed in commit 97eabb305b (`feat(account-keychain): implement TIP-1011 T3 semantics`). `get_allowed_calls()` returns the explicit deny-all sentinel (`target = address(0)`) when mode is deny-all, preserving round-trip disambiguation from unrestricted mode.

crates/transaction-pool/src/validator.rs
```diff
@@ -154,6 +159,200 @@ where
...
+
+        recipients.contains(&Address::from_slice(&recipient_word[12..]))
+    }
+
+    fn validate_inline_t3_call_scopes(
+        tx: &TempoTransaction,
+        auth: &tempo_primitives::transaction::SignedKeyAuthorization,
+    ) -> Result<(), TempoPoolTransactionError> {
+        for call in &tx.calls {
+            if !Self::call_scope_allows_call(auth.allowed_calls.as_deref(), &call.to, &call.input) {
```
legion2002, line 353: Fixed in commit `1bdf359a2`. Txpool now rejects access-key `CREATE` calls in the inline T3 path by short-circuiting both `call_scope_allows_call()` and `validate_inline_t3_call_scopes()`, so pool admission matches `AccountKeychain::validate_call_scope_for_transaction()`.

crates/revm/src/handler.rs
```diff
@@ -1099,8 +1173,22 @@ where
...
                 |mut keychain: AccountKeychain| {
-                    // Skip keychain validation when authorizing this key in the same tx
-                    if !is_authorizing_this_key {
+                    if is_authorizing_this_key {
+                        if spec.is_t1()
+                            && tempo_tx_env
+                                .key_authorization
+                                .as_ref()
+                                .is_some_and(|key_auth| {
+                                    key_auth.key_type != keychain_sig.signature.signature_type()
```
legion2002, line 1200: Fixed in commit `1bdf359a2`. The txpool inline-auth branch now mirrors the handlers T1+ parity check and rejects transactions where `key_authorization.key_type` does not match the actual keychain signature type.

crates/revm/src/handler.rs
```diff
@@ -163,7 +163,24 @@ fn calculate_key_authorization_gas(
         let sload_cost =
             gas_params.warm_storage_read_cost() + gas_params.cold_storage_additional_cost();
 
-        sig_gas + sload_cost + sstore_cost * (1 + num_limits) + BUFFER
+        let limit_slots = if spec.is_t3() {
+            // T3 periodic limits write 4 set-slots per token:
+            // spending_limits[token] + period_state.{max,period,period_end}
+            num_limits.saturating_mul(4)
```
legion2002: Fixed in commit `1bdf359a2`. T3 intrinsic gas now charges `3` periodic-limit storage writes per token instead of `4`, matching the packed `SpendingLimitPeriodState` layout, and the handler test plus TIP note were updated to reflect the actual write count.

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -331,16 +503,181 @@ impl AccountKeychain {
...
+        let key = self.load_active_key(msg_sender, call.keyId)?;
+        let current_timestamp = self.storage.timestamp().saturating_to::<u64>();
+        if current_timestamp >= key.expiry {
+            return Err(AccountKeychainError::key_expired().into());
+        }
+
+        let account_key = Self::account_key(msg_sender, call.keyId);
+
+        // Any explicit target update implies restricted mode.
+        self.key_scopes[account_key].mode.write(1)?;
```
legion2002: Fixed in commit `1bdf359a2`. `set_allowed_calls()` now canonicalizes an empty post-update target set to explicit deny-all (`mode = 2`), and `get_allowed_calls()` returns the deny-all sentinel for that state, so delete-equivalent updates no longer leave a hidden zero-target restricted mode.

crates/contracts/src/precompiles/account_keychain.rs
```diff
@@ -44,18 +61,31 @@ crate::sol! {
...
+            uint256 remainingLimit
+        );
+
         /// Authorize a new key for the caller's account
         /// @param keyId The key identifier (address derived from public key)
         /// @param signatureType 0: secp256k1, 1: P256, 2: WebAuthn
         /// @param expiry Block timestamp when the key expires (u64::MAX for never expires)
         /// @param enforceLimits Whether to enforce spending limits for this key
         /// @param limits Initial spending limits for tokens (only used if enforceLimits is true)
+        /// @param enforceAllowedCalls Whether call scope restrictions are enabled for this key
```
legion2002: Fixed in commit `1bdf359a2`. I added a legacy `IAccountKeychainLegacy` ABI alongside the T3 interface so pre-T3 callers still have the 5-argument `authorizeKey` surface and legacy `getRemainingLimit` shape available from the exported bindings.

crates/alloy/src/provider/ext.rs
```diff
@@ -40,14 +40,17 @@ pub trait TempoProviderExt: Provider<TempoNetwork> {
...
     {
-        self.account_keychain()
+        let ret = self
+            .account_keychain()
             .getRemainingLimit(account, key_id, token)
             .call()
-            .await
+            .await?;
+
+        Ok((ret.remaining, ret.periodEnd))
```
legion2002: Fixed in commit `1bdf359a2`. `TempoProviderExt::get_keychain_remaining_limit()` now attempts the T3 tuple decode first and falls back to the legacy single-`uint256` decode, returning `periodEnd = 0` on pre-T3 nodes.

crates/alloy/src/provider/ext.rs
```diff
@@ -40,14 +54,26 @@ pub trait TempoProviderExt: Provider<TempoNetwork> {
...
-    ) -> ContractResult<U256>
+    ) -> ContractResult<(U256, u64)>
     where
         Self: Sized,
     {
-        self.account_keychain()
+        let raw = self
+            .account_keychain()
             .getRemainingLimit(account, key_id, token)
+            .clear_decoder()
```
gakonst: ?

crates/alloy/src/network.rs
```diff
@@ -292,8 +292,56 @@ impl RecommendedFillers for TempoNetwork {
     }
 }
 
+/// Tempo wallet wrapper around [`EthereumWallet`].
+///
+/// This local wrapper avoids coherence conflicts if upstream alloy adds broader
+/// `NetworkWallet` blanket impls for `EthereumWallet`.
```
gakonst: why?

crates/contracts/src/precompiles/account_keychain.rs
```diff
@@ -44,18 +61,31 @@ crate::sol! {
...
+        /// @param allowedCalls Initial call scopes for this key
         function authorizeKey(
             address keyId,
             SignatureType signatureType,
             uint64 expiry,
             bool enforceLimits,
-            TokenLimit[] calldata limits
+            TokenLimit[] calldata limits,
+            bool enforceAllowedCalls,
+            CallScope[] calldata allowedCalls
```
gakonst: interface starting to get big

crates/contracts/src/precompiles/account_keychain.rs
```diff
@@ -104,6 +146,46 @@ crate::sol! {
...
+}
+
+crate::sol! {
+    /// Legacy Account Keychain interface exposed for pre-T3 networks.
+    ///
+    /// Current nodes before T3 still accept the original 5-argument `authorizeKey`
+    /// selector and return a single `uint256` from `getRemainingLimit`.
+    #[derive(Debug, PartialEq, Eq)]
+    #[sol(abi)]
+    interface IAccountKeychainLegacy {
```
gakonst: curious why we do new account keychain vs extend the current one / polymorphism for new args?

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -8,26 +8,103 @@
...
-    storage::{Handler, Mapping, packing::insert_into_word},
+    storage::{Handler, Mapping, Set, packing::insert_into_word},
+    tip20_factory::TIP20Factory,
 };
-use alloy::primitives::{Address, B256, U256};
+use alloy::primitives::{Address, B256, TxKind, U256};
 use tempo_precompiles_macros::{Storable, contract};
 
+/// Maximum number of call scopes per account key.
+pub const MAX_CALL_SCOPES: u8 = 8;
```
legion2002, line 42: These numbers are eyeballed atm to a tight enough amount. The cap is more for defense against weird decoding attacks. But we can discuss if these should be tuned.

crates/precompiles/src/account_keychain/dispatch.rs
```diff
@@ -1,22 +1,104 @@
...
+                        amount: limit.amount,
+                        period: 0,
+                    })
+                    .collect(),
+                enforceAllowedCalls: false,
+                allowedCalls: vec![],
+            };
+
+            return mutate_void(call, msg_sender, |sender, c| self.authorize_key(sender, c));
+        }
```
gakonst: this is the kind of thing that i'd hope we can have much more cleanly implemented in the code vs large if/else blocks - not sure how to solve

crates/precompiles/src/account_keychain/dispatch.rs
```diff
@@ -65,4 +178,117 @@ mod tests {
...
     }
+
+    #[test]
+    fn test_legacy_authorize_key_selector_supported_pre_t3() -> eyre::Result<()> {
+        let mut storage = HashMapStorageProvider::new_with_spec(1, TempoHardfork::T1C);
+        let account = Address::random();
+        let key_id = Address::random();
+        let token = Address::random();
+
+        StorageCtx::enter(&mut storage, || {
```
gakonst, line 168: really dislike this pattern obviously

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -8,26 +8,103 @@
...
+/// Maximum number of selector rules per call scope.
+pub const MAX_SELECTOR_RULES_PER_SCOPE: u8 = 16;
+/// Maximum number of recipients per selector rule.
+pub const MAX_RECIPIENTS_PER_SELECTOR: u8 = 16;
```
gakonst, line 46: how is this chosen?

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -8,26 +8,103 @@
...
+
+/// Allowed TIP-20 selectors for recipient-constrained rules.
+const TIP20_TRANSFER_SELECTOR: [u8; 4] = [0xa9, 0x05, 0x9c, 0xbb];
+const TIP20_APPROVE_SELECTOR: [u8; 4] = [0x09, 0x5e, 0xa7, 0xb3];
+const TIP20_TRANSFER_WITH_MEMO_SELECTOR: [u8; 4] = [0x95, 0x77, 0x7d, 0x59];
+const TIP20_APPROVE_WITH_MEMO_SELECTOR: [u8; 4] = [0xc2, 0xda, 0x9a, 0xed];
```
gakonst: do we not have a macro fro this?

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -8,26 +8,103 @@
...
+    pub token: Address,
+    pub limit: U256,
+    pub period: u64,
+}
+
+/// Internal selector rule configuration used by handler-side key authorization.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct SelectorRuleConfig {
+    pub selector: [u8; 4],
+    pub recipients: Option<Vec<Address>>,
```
gakonst: this is confusing to me - the selector rule should be "if the selector is a transfer selector then the recipients and amounts can only be such and such" how is a selector in the general case associated with a recipient, doesn't make sense

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -8,26 +8,103 @@
...
+    pub selectors: Set<u32>,
+    pub selector_scopes: Mapping<u32, SelectorScope>,
+}
+
+/// Key-level call scope.
+///
+/// mode:
+/// - 0 => unrestricted
+/// - 1 => scoped
+/// - 2 => deny-all
```
gakonst, line 91: these should be enums and we should have a Storable for enums @0xrusowsky

crates/precompiles/src/account_keychain/mod.rs
legion2002: Yeah I left a comment about this [here](https://github.com/tempoxyz/tempo/pull/3323#discussion_r2998095546)

they're eyeballed to a reasonable amount atm. Since we charge fairly for the gas, there is no concrete attack vector why we need this cap. This is just here for sanity to protect against weird decoding or other DOS vectors.

crates/contracts/src/precompiles/account_keychain.rs
legion2002: yeah, this is the most minimal version I could think of, to have all the features we want. The bool is necessary, so we can differentiate between "no allowed calls" and "all calls allowed"

crates/precompiles/src/account_keychain/mod.rs
legion2002: Agree that this is confusing, I was trying to think of a cleaner interface, but couldn't come up with anything good.

So we want to express 2 things - 
1. any random target, can scope by random selectors (but without recipients)
2. if the target is TIP 20, and the selectors are transfer/approve, then the recipients scoping is allowed.

We enforce the second one in the handler, but to also make this clear it in the interface, we would need to maintain 2 separate interfaces for general selectors, and selectors where `recipients` is allowed.

crates/contracts/src/precompiles/account_keychain.rs
legion2002: We had to do this legacy one because `getRemainingLimit` changed return type, and the `TokenLimit` struct added a field.  so we couldn't just extend it.
But I think we can just use a different fn selector for the new getRemainingLimit, and just extend the old interface. Updated here - https://github.com/tempoxyz/tempo/pull/3323/commits/575727b5e0b590688796d5ed4a488d75cc77ad54

crates/precompiles/src/account_keychain/mod.rs
legion2002: updated in https://github.com/tempoxyz/tempo/pull/3323/commits/575727b5e0b590688796d5ed4a488d75cc77ad54

crates/precompiles/src/account_keychain/dispatch.rs
legion2002: sidestepped this in the PR, by keeping the old function selector in the ABI. https://github.com/tempoxyz/tempo/pull/3323/commits/575727b5e0b590688796d5ed4a488d75cc77ad54

crates/contracts/src/precompiles/account_keychain.rs
0xrusowsky: only cosmetic, but we could introduce a `KeyRestrictions` type to keep interface minimal (although it doesn't help with calldata size)

```solidity
struct KeyRestrictions {
    uint64 expiry,
    bool enforceLimits;
    TokenLimit[] limits;
    bool enforceAllowedCalls;
    CallScope[] allowedCalls;
}

function authorizeKey(
    address keyId,
    SignatureType signatureType,
    KeyRestrictions calldata config
) external;
```

crates/precompiles/src/account_keychain/dispatch.rs
```diff
@@ -1,22 +1,104 @@
...
+            TokenLimit[] calldata limits
+        ) external;
+
+        function getRemainingLimit(
+            address account,
+            address keyId,
+            address token
+        ) external view returns (uint256);
+    }
+}
```
0xrusowsky: this is redundant; it has already been defined in `crates/contracts/src/precompiles/account_keychain.rs`

import `IAccountKeychainLegacy` instead

crates/precompiles/src/account_keychain/mod.rs
0xrusowsky: these are already generated by the `sol!` macro that defines `TIP20`

you can import the corresponding Call struct for each method and use its `::SELECTOR`

crates/precompiles/src/account_keychain/mod.rs
0xrusowsky: i think that the type system needs some work.
rust's type system (enums) is reacher than solidity's, so you can express smth like:

```rs
pub enum SelectorRuleConfig {
    /// Selector allowed, no recipient constraints
    Unrestricted([u8; 4]),
    /// TIP-20 selector constrained to recipient allowlist
    Tip20 { selector: [u8; 4], recipients: Vec<Address> },
}
```

then we can convert from solidity's flat `SelectorRule` struct via smth like `fn from_abi(rule: &SelectorRule, target: Address)` which has all the validation logic from `abi_call_scope_to_config()` + `validate_selector_rules()`

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -48,6 +134,17 @@ pub struct AuthorizedKey {
     pub is_revoked: bool,
 }
 
+/// Per-token periodic spending limit state (T3+).
+#[derive(Debug, Clone, Default, PartialEq, Eq, Storable)]
+pub struct SpendingLimitPeriodState {
+    /// Maximum amount allowed per period.
+    pub max: U256,
```
0xrusowsky: perf: given that TIP20 supply is capped at `u128` (cause 6 decimals), we should turn this into `u128` so that the struct packs in 1 slot.

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -8,26 +8,112 @@
...
+/// Selector-level scope for one selector under one target.
+///
+/// mode:
+/// - 0 => unset/disabled
+/// - 1 => selector allowed for any recipient
+/// - 2 => selector allowed only for recipients in the set
+#[derive(Debug, Clone, Storable, Default)]
+pub struct SelectorScope {
+    pub mode: u8,
+    pub recipients: Set<Address>,
```
0xrusowsky, line 70: flagging that `Set` is an unaudited primitive (despite being a impl of OpenZeppelin's EnumerableSet) so we must remember to tell auditors to look into it

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -134,6 +236,22 @@ impl AccountKeychain {
         keccak256(data)
     }
 
+    /// Key for account-scoped permission/storage rows.
+    #[inline]
+    pub fn account_key(account: Address, key_id: Address) -> B256 {
+        Self::spending_limit_key(account, key_id)
+    }
```
0xrusowsky: why did we introduce this? let's drop it and always use `spending_limit_key` (or rename it)

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -374,6 +725,371 @@ impl AccountKeychain {
...
+                SpendingLimitPeriodState {
+                    max: limit.limit,
+                    period: limit.period,
+                    period_end,
+                },
+            )?;
+        }
+
+        let account_key = Self::account_key(account, key_id);
+        self.replace_allowed_calls(account_key, allowed_calls)
```
0xrusowsky: this is the already computed `limit_key`, so we shouldn't compute it again (hashing is expensive)

```suggestion
        self.replace_allowed_calls(limit_key, allowed_calls)
```

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -195,14 +314,57 @@ impl AccountKeychain {
...
-        if call.enforceLimits {
+        let limit_configs = call
+            .limits
+            .iter()
+            .map(|limit| TokenLimitConfig {
+                token: limit.token,
+                limit: limit.amount,
+                period: limit.period,
+            })
+            .collect::<Vec<_>>();
```
0xrusowsky: this should go inside the `if is_t3 {` branch.

additionally we are doing an unnecessary vec alloc, to iterate over it later. Instead we should pass an iter and change the `fn apply_key_authorization_t3()` sig to:
```rs
    pub fn apply_key_authorization_t3(
        &mut self,
        account: Address,
        key_id: Address,
        limits: impl IntoIterator<Item = TokenLimitConfig>,
        allowed_calls: Option<&[CallScopeConfig]>,
    ) -> Result<()> {
```

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -469,6 +1185,61 @@ impl AccountKeychain {
...
+            period_state.period,
+            current_timestamp,
+        );
+
+        Ok((period_state.max, next_end))
+    }
+
+    /// Computes the period end for the current rollover window, saturating on
+    /// all intermediate operations to avoid overflow in extreme timestamps.
+    fn compute_next_period_end(period_end: u64, period: u64, current_timestamp: u64) -> u64 {
```
0xrusowsky: nit: this fn belongs into `SpendingLimitPeriodState` impl block

```rs
fn compute_next_period_end(&self, current_timestamp: u64) -> u64 {
    let elapsed = current_timestamp.saturating_sub(self.period_end);
    let periods_elapsed = (elapsed / self.period).saturating_add(1);
    let advance = self.period.saturating_mul(periods_elapsed);
    self.period_end.saturating_add(advance)
}
```

crates/precompiles/src/account_keychain/mod.rs
0xrusowsky: will prio `Storable` support for enums, but in the meantime you can already use enums in the impls with `From<u8>`/`Into<u8>` conversions at read/write boundaries (note that the storage field stays `u8`)

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -134,6 +236,22 @@ impl AccountKeychain {
...
+
+    #[inline]
+    fn selector_key(selector: [u8; 4]) -> u32 {
+        u32::from_be_bytes(selector)
+    }
+
+    #[inline]
+    fn selector_from_u32(selector: u32) -> [u8; 4] {
+        selector.to_be_bytes()
+    }
```
0xrusowsky, line 241: these are not needed if we use `FixedBytes<4>` (which is also `Storable`), and we can simply use `FixedBytes<4>::from(selector)`

crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -21,18 +21,142 @@ pub struct TokenLimit {
...
+/// `selector_rules` uses tri-state semantics:
+/// - `None` => allow any selector for this target
+/// - `Some([])` => deny all selectors for this target
+/// - `Some([..])` => allow exactly the listed selector rules
+#[derive(Clone, Debug, PartialEq, Eq, Hash, alloy_rlp::RlpEncodable, alloy_rlp::RlpDecodable)]
+#[rlp(trailing)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
+#[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
+pub struct CallScope {
```
0xrusowsky, line 126: `TokenLimit`, `CallScope`, and `SelectorRule` are identical to the types defined in `precompiles/`.

in `handler.rs` we have to do multiple verbose conversions from these identical types.

should we make `tempo-primitives` (just with `std` feature) a dep of `tempo-precompiles`? if we decide to do it, let's ask @mattsse first in case he is opinionated of adding new crate interdeps

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -8,26 +8,112 @@
...
+    )
+}
+
+/// Internal token limit configuration used by handler-side key authorization.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TokenLimitConfig {
+    pub token: Address,
+    pub limit: U256,
+    pub period: u64,
+}
```
klkvr: why do we need this struct? it seems identical to `TokenLimit`

crates/precompiles/src/account_keychain/mod.rs
klkvr: +1 here, it's `spending_limit_key` semantically in both cases

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -134,6 +236,22 @@ impl AccountKeychain {
...
+    /// Key for account-scoped permission/storage rows.
+    #[inline]
+    pub fn account_key(account: Address, key_id: Address) -> B256 {
+        Self::spending_limit_key(account, key_id)
+    }
+
+    #[inline]
+    fn selector_key(selector: [u8; 4]) -> u32 {
+        u32::from_be_bytes(selector)
+    }
```
klkvr: can't we put `[u8; 4]` as is into the storage?

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -195,14 +314,57 @@ impl AccountKeychain {
...
+            self.apply_key_authorization_t3(
+                msg_sender,
+                call.keyId,
+                if call.enforceLimits {
+                    &limit_configs
+                } else {
+                    &[]
+                },
+                allowed_call_configs.as_deref(),
+            )?;
```
klkvr: not immediately clear why this is needed. ideally we unify as much codepaths as possible with T3/T2

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -374,6 +725,371 @@ impl AccountKeychain {
...
+        }
+
+        let limit_key = Self::spending_limit_key(account, key_id);
+        let now = self.storage.timestamp().saturating_to::<u64>();
+
+        let mut seen_tokens = HashSet::new();
+        for limit in limits {
+            if !seen_tokens.insert(limit.token) {
+                return Err(AccountKeychainError::invalid_call_scope().into());
+            }
```
klkvr: looks like we didn't have this check pre-T3?

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -8,26 +8,112 @@
...
+/// mode:
+/// - 0 => unrestricted
+/// - 1 => scoped
+/// - 2 => deny-all
+#[derive(Debug, Clone, Storable, Default)]
+pub struct KeyScope {
+    pub mode: u8,
+    pub targets: Set<Address>,
+    pub target_scopes: Mapping<Address, TargetScope>,
+}
```
klkvr, line 97: should we have `EnumerableMap` implementation? the current datastructure here seems suboptimal because we need to do 2x hashing due to 2 separate maps (`Set` and a `Mapping`)

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -8,26 +8,112 @@
...
+/// mode:
+/// - 0 => unset/disabled
+/// - 1 => all selectors allowed
+/// - 2 => only selectors in the set are allowed
+#[derive(Debug, Clone, Storable, Default)]
+pub struct TargetScope {
+    pub mode: u8,
+    pub selectors: Set<u32>,
+    pub selector_scopes: Mapping<u32, SelectorScope>,
+}
```
klkvr, line 84: same here

crates/contracts/src/precompiles/account_keychain.rs
```diff
@@ -21,10 +25,31 @@ crate::sol! {
...
             uint256 amount;
+            uint64 period;
+        }
+
+        /// Selector-level recipient rule.
+        struct SelectorRule {
+            bytes4 selector;
+            bool allowAllRecipients;
+            address[] recipients;
+        }
```
klkvr, line 47: isn't this bool redundant? we can infer this from empty recipients

this should also allow us to remove `SelectorRuleConfig`

crates/contracts/src/precompiles/account_keychain.rs
```diff
@@ -21,10 +25,31 @@ crate::sol! {
...
+
+        /// Per-target call scope.
+        struct CallScope {
+            address target;
+            bool allowAllSelectors;
+            SelectorRule[] selectorRules;
         }
```
klkvr, line 56: same here i believe? empty `selectorRules` could be inferred as `allowAllSelectors`

crates/revm/src/handler.rs
```diff
@@ -163,7 +163,24 @@ fn calculate_key_authorization_gas(
...
             gas_params.warm_storage_read_cost() + gas_params.cold_storage_additional_cost();
 
-        sig_gas + sload_cost + sstore_cost * (1 + num_limits) + BUFFER
+        let limit_slots = if spec.is_t3() {
+            // T3 periodic limits write 3 storage slots per token:
+            // spending_limits[token] + period_state.max + packed {period,period_end}
+            num_limits.saturating_mul(3)
+        } else {
+            num_limits
+        };
```
klkvr, line 172: should we make this fair and don't count zero writes?

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -111,6 +208,11 @@ pub struct AccountKeychain {
     // spendingLimits[(account, keyId)][token] -> amount
     // Using a hash of account and keyId as the key to avoid triple nesting
     spending_limits: Mapping<B256, Mapping<Address, U256>>,
+    // spendingLimitPeriodState[(account, keyId)][token] -> { max, period, period_end }
+    spending_limit_period_state: Mapping<B256, Mapping<Address, SpendingLimitPeriodState>>,
```
klkvr: doesn't storage layout allow us to re-use the `spending_limits` mapping with something like this?
```rust
struct SpendingLimitState {
    pub remaining: U256
    pub max: U256,
    pub period: u64,
    pub period_end: u64,
}
```

i believe `remaining` would be in the first slot as it is right now

crates/contracts/src/precompiles/account_keychain.rs
```diff
@@ -72,22 +119,44 @@ crate::sol! {
             uint256 newLimit
         ) external;
 
+        /// Set or replace allowed calls for a key+target pair.
+        function setAllowedCalls(
+            address keyId,
+            address target,
+            bool allowAllSelectors,
+            SelectorRule[] calldata selectorRules
+        ) external;
```
klkvr, line 133: this should take `key_id` and `CallScope`

crates/precompiles/src/account_keychain/mod.rs
0xrusowsky: we can use as storage but not as mapping key.
we should use `FixedBytes<4>` instead, which can do both and already impl `From` trait

suggested here: https://github.com/tempoxyz/tempo/pull/3323#discussion_r3000438313

crates/precompiles/src/account_keychain/mod.rs
0xrusowsky: I missed this one, good call.

yes, it would be backwards compatible since collisions are virtually impossible statistically speaking

crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -113,13 +319,74 @@ impl KeyAuthorization {
...
+
     /// Calculates a heuristic for the in-memory size of the key authorization
     pub fn size(&self) -> usize {
         size_of::<Self>()
             + self
                 .limits
                 .as_ref()
                 .map_or(0, |limits| limits.capacity() * size_of::<TokenLimit>())
+            + self.allowed_calls.as_ref().map_or(0, |scopes| {
+                scopes.capacity() * size_of::<CallScope>()
```
fgimenez, line 365: nit: looks like `size()` double-counts? `scopes.capacity() * size_of::<CallScope>()` already covers the per-element fixed size, so the `size_of::<CallScope>()` (and `size_of::<SelectorRule>()`) inside the `.map()` iterators add it again. aiui harmless since it's a heuristic, just slightly over-reports memory

crates/contracts/src/precompiles/account_keychain.rs
legion2002: We need this bool to differentiate between "no recipients allowed" and "all selectors allowed"
there are no optional params in solidity, so we have to represent it with a bool.

legion2002: and although it doesn't make sense for someone to auth a key with no recipients. We allow them to do this, in case they want to quickly pause a key.

ig one other way to do it could be with some token `address(0)` value for recipients representing "no recipients allowed"

But it seems like a footgun, where someone might send an empty array expecting the key to be blocked, but it actually gives it unlimited access.

crates/contracts/src/precompiles/account_keychain.rs
legion2002: answered [here](https://github.com/tempoxyz/tempo/pull/3323/changes#r3002484530)

crates/contracts/src/precompiles/account_keychain.rs
legion2002: yeah good idea, looks cleaner.

crates/precompiles/src/account_keychain/mod.rs
0xrusowsky: PR here: https://github.com/tempoxyz/tempo/pull/3354

crates/contracts/src/precompiles/account_keychain.rs
klkvr: hmm but to remove a given `SelectorRule` you could just call `setAllowedCalls` with the `selectorRules` array not containing the target one

the reason i want to change it is so that we can unify `SelectorRuleConfig` and `SelectorRule`

legion2002: hmm that makes sense, I'll update this.

crates/contracts/src/precompiles/account_keychain.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/0955404e6ad89436c280cba4a2f36d7405bd5e3f

crates/precompiles/src/account_keychain/mod.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/e0ac1c7d0ce2f99a70b71eefddc6d30cca29716a

crates/precompiles/src/account_keychain/mod.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/e0ac1c7d0ce2f99a70b71eefddc6d30cca29716a

crates/contracts/src/precompiles/account_keychain.rs
legion2002: addressed here - https://github.com/tempoxyz/tempo/pull/3323/changes/0955404e6ad89436c280cba4a2f36d7405bd5e3f

crates/contracts/src/precompiles/account_keychain.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/pull/3323/changes/0955404e6ad89436c280cba4a2f36d7405bd5e3f

crates/precompiles/src/account_keychain/mod.rs
legion2002: addressed here - https://github.com/tempoxyz/tempo/pull/3323/changes/e0ac1c7d0ce2f99a70b71eefddc6d30cca29716a

crates/primitives/src/transaction/key_authorization.rs
legion2002: addressed in - https://github.com/tempoxyz/tempo/pull/3323/changes/ca3c699569d478a1272292e4d037ffa429bdbaa9

crates/precompiles/src/account_keychain/mod.rs
legion2002: Yup good one - updated here https://github.com/tempoxyz/tempo/pull/3323/changes/01e231617c827e961c2833b65125cd61c9e04e02

crates/precompiles/src/account_keychain/mod.rs
legion2002: nice okay, updated here - https://github.com/tempoxyz/tempo/pull/3323/commits/ab709a12d7cd61980ea32846922c175da93581b8

crates/precompiles/src/account_keychain/mod.rs
legion2002: Updated here - https://github.com/tempoxyz/tempo/pull/3323/commits/ab709a12d7cd61980ea32846922c175da93581b8

crates/precompiles/src/account_keychain/mod.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/ce826a2c0b11501d7f4579ad96f26923fbb3005f

crates/precompiles/src/account_keychain/mod.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/ce826a2c0b11501d7f4579ad96f26923fbb3005f

crates/precompiles/src/account_keychain/mod.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/ce826a2c0b11501d7f4579ad96f26923fbb3005f

crates/primitives/src/transaction/key_authorization.rs
mattsse: tempo-primitives is already part of the dep graph of the precompiles crate via chainspec, so it makes sense to me to move them to the primitives crate

legion2002: addressed here - https://github.com/tempoxyz/tempo/pull/3323/changes/d41004b3b0d3c28d28f6dd37a1f88f9e78a8c286

crates/contracts/src/precompiles/account_keychain.rs
legion2002: addressed here - https://github.com/tempoxyz/tempo/pull/3323/changes/d41004b3b0d3c28d28f6dd37a1f88f9e78a8c286

crates/precompiles/src/account_keychain/mod.rs
legion2002: yeah this was accidental. We don't need this check because we add the limits if there are duplicates.
Fixed here - https://github.com/tempoxyz/tempo/pull/3323/commits/c7c89b9d9b2c071682a874bbadc7fe64ce113ca4

crates/precompiles/src/account_keychain/mod.rs
legion2002: Yup updated here - https://github.com/tempoxyz/tempo/pull/3323/commits/c7c89b9d9b2c071682a874bbadc7fe64ce113ca4

crates/transaction-pool/src/validator.rs
```diff
@@ -154,6 +160,189 @@ where
         Ok(())
     }
 
+    fn validate_t3_key_authorization_shape(
```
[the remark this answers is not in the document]
0xrusowsky: @legion2002 looks valid, we either do what cyclops propose or change `TokenLimit` so that `limit: u128` rather than `U256`

crates/precompiles/src/account_keychain/mod.rs
legion2002: working on adding an enumerable map impl in #3370

crates/revm/src/handler.rs
```diff
@@ -1008,18 +1043,59 @@ where
...
+                                        .map(|rules| {
+                                            rules
+                                                .iter()
+                                                .map(|rule| {
+                                                    tempo_precompiles::account_keychain::SelectorRule {
+                                                        selector: rule.selector.into(),
+                                                        recipients: rule
+                                                            .recipients
+                                                            .clone()
+                                                            .unwrap_or_default(),
```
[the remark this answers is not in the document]
legion2002: addressed in - https://github.com/tempoxyz/tempo/commit/b32b60f7a915d34cc30cdd2aabfabec26d805069

crates/transaction-pool/src/validator.rs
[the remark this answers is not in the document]
legion2002: addressed in - https://github.com/tempoxyz/tempo/commit/b32b60f7a915d34cc30cdd2aabfabec26d805069

crates/revm/src/handler.rs
```diff
@@ -1008,18 +1043,59 @@ where
                             .map(|limit| TokenLimit {
                                 token: limit.token,
```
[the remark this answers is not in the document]
legion2002: addressed in - https://github.com/tempoxyz/tempo/commit/0848c18c13fdb3d603d52f8d37826f0783aa312a and https://github.com/tempoxyz/tempo/commit/b32b60f7a915d34cc30cdd2aabfabec26d805069

crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -21,18 +21,142 @@ pub struct TokenLimit {
 
     /// Maximum spending amount for this token (enforced over the key's lifetime)
     pub limit: U256,
+
+    /// Period duration in seconds.
+    ///
+    /// `0` means one-time limit. `>0` means the limit resets periodically.
+    pub period: u64,
```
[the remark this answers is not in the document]
legion2002: addressed in - https://github.com/tempoxyz/tempo/commit/38c9fbc741f9c05beee95af9bc714ccee1d6c64c

crates/alloy/src/provider/ext.rs
```diff
@@ -40,14 +41,20 @@ pub trait TempoProviderExt: Provider<TempoNetwork> {
...
         Self: Sized,
     {
-        self.account_keychain()
-            .getRemainingLimit(account, key_id, token)
+        let getRemainingLimitReturn {
+            remaining,
+            periodEnd,
+        } = self
+            .account_keychain()
+            .getRemainingLimitWithPeriod(account, key_id, token)
```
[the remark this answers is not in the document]
legion2002: We addressed this by removing the premature `tempo-alloy` helper change from the PR and keeping `TempoProviderExt::get_keychain_remaining_limit()` on the legacy `getRemainingLimit` selector for now.

Reasoning:
- T3 is not active yet, so the helper must remain compatible with live T1C/T2 nodes.
- Switching the helper to `getRemainingLimitWithPeriod` in this PR would make the SDK depend on a future-only selector before activation.

Commit:
- https://github.com/tempoxyz/tempo/commit/1b8c0283ee4dc5e04b12d09b114e3dcdb4a8ba9c

crates/revm/src/handler.rs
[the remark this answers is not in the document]
legion2002: We chose to reject duplicate token entries on T3+ instead of keeping last-write-wins.

Reasoning: pre-T3 duplicates were effectively tolerable because each token only had a single simple limit slot, but T3 adds period-bearing limit semantics, so duplicate entries for the same token become ambiguous to merge safely (especially if one entry is one-time and another is periodic). Rejecting them is simpler, safer, and restores alignment with the TIP.

Commits:
- https://github.com/tempoxyz/tempo/commit/0848c18c13fdb3d603d52f8d37826f0783aa312a
- https://github.com/tempoxyz/tempo/commit/b32b60f7a915d34cc30cdd2aabfabec26d805069

PR #3325, opened 2026-03-26
perf: use ETL in init-state

shekhirin:
Closes RETH-216

State bloat generation and ingestion OOM at >=10GB because both tools hold all data in memory. Generation materializes ~33GB of vectors for 100GB target; ingestion collects all entries in a HashMap before flushing, consuming ~1.6× file size in heap.

Fixes by using ETL

### DIFF
--- a/bin/tempo/src/init_state.rs
+++ b/bin/tempo/src/init_state.rs
@@ -8,6 +8,7 @@ use std::{
     io::{BufReader, Read},
     path::PathBuf,
     sync::mpsc,
+    thread,
     time::{Duration, Instant},
 };
 
@@ -43,11 +44,11 @@ const VERSION: u16 = 1;
 /// ETL collector file size (200 MiB per temp file before spilling a new one).
 const ETL_FILE_SIZE: usize = 200 * 1024 * 1024;
 
-/// Maximum number of storage entries to hash per rayon worker job.
+/// Maximum number of storage entries to hash per worker batch.
 const WORKER_CHUNK_SIZE: usize = 100;
 
-/// Maximum number of channels that can exist in memory before draining.
-const MAXIMUM_CHANNELS: usize = 10_000;
+/// Bounded channel depth for the hashing worker thread.
+const HASH_WORKER_QUEUE_DEPTH: usize = 256;
 
 /// Initialize state from a binary dump file.
 #[derive(Debug, Parser)]
@@ -101,12 +102,36 @@ impl<C: reth_cli::chainspec::ChainSpecParser<ChainSpec: EthChainSpec + EthereumH
         // ETL collectors: accumulate entries sorted, spill to disk when full
         let mut plain_collector: Collector<Vec<u8>, CompactU256> =
             Collector::new(ETL_FILE_SIZE, None);
-        let mut hashed_collector: Collector<Vec<u8>, CompactU256> =
-            Collector::new(ETL_FILE_SIZE, None);
-
         let mut hash_chunk: Vec<(alloy_primitives::Address, B256, CompactU256)> =
             Vec::with_capacity(WORKER_CHUNK_SIZE);
-        let mut pending_channels: Vec<mpsc::Receiver<(Vec<u8>, CompactU256)>> = Vec::new();
+
+        // Single worker thread for keccak hashing: owns the hashed ETL collector, receives
+        // batches over a bounded channel, and returns the collector when the sender drops.
+        let (hash_tx, hash_rx) = mpsc::sync_channel::<
+            Vec<(alloy_primitives::Address, B256, CompactU256)>,
+        >(HASH_WORKER_QUEUE_DEPTH);
+        let hash_worker =
+            thread::spawn(move || -> eyre::Result<Collector<Vec<u8>, CompactU256>> {
+                let mut hashed_collector: Collector<Vec<u8>, CompactU256> =
+                    Collector::new(ETL_FILE_SIZE, None);
+                while let Ok(chunk) = hash_rx.recv() {
+                    let mut last_addr = alloy_primitives::Address::ZERO;
+                    let mut hashed_addr = B256::ZERO;
+                    for (address, slot, value) in chunk {
+                        if address != last_addr {
+                            last_addr = address;
+                            hashed_addr = keccak256(address);
+                        }
+                        let mut hashed_key = Vec::with_capacity(64);
+                        hashed_key.extend_from_slice(hashed_addr.as_slice());
+                        hashed_key.extend_from_slice(keccak256(slot).as_slice());
+                        hashed_collector
+                            .insert(hashed_key, value)
+                            .wrap_err("hashed ETL insert failed")?;
+                    }
+                }
+                Ok(hashed_collector)
+            });
 
         // Process blocks from binary file
         loop {
@@ -149,7 +174,9 @@ impl<C: reth_cli::chainspec::ChainSpecParser<ChainSpec: EthChainSpec + EthereumH
             );
 
             // Ensure account exists in plain state (only on first encounter).
-            // Read the existing account from genesis, or insert a default empty one.
+            // The binary dump is chunked: generate-state-bloat writes one block per token
+            // per chunk, so the same token address appears in multiple blocks. This entry
+            // is Vacant on the first chunk and Occupied on subsequent ones.
             // Preserving the genesis account is critical: TIP20 tokens have bytecode (0xEF)
             // set during genesis, and overwriting with Account::default() would clear the
             // code hash, making the token appear uninitialized.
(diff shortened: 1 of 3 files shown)


bin/tempo/src/init_state.rs
```diff
@@ -157,78 +143,142 @@ impl<C: reth_cli::chainspec::ChainSpecParser<ChainSpec: EthChainSpec + EthereumH
...
+                hashed_key.extend_from_slice(hashed_address.as_slice());
+                hashed_key.extend_from_slice(keccak256(slot).as_slice());
+                hashed_collector
+                    .insert(hashed_key, compact_value)
+                    .wrap_err("ETL insert failed")?;
+
+                total_entries += 1;
+
+                let now = Instant::now();
+                if now.duration_since(last_log) >= Duration::from_secs(5) || i + 1 == pair_count {
```
yongkangc: we might wanna clarify / simplify what this is doing

bin/tempo/src/init_state.rs
```diff
@@ -157,78 +143,142 @@ impl<C: reth_cli::chainspec::ChainSpecParser<ChainSpec: EthChainSpec + EthereumH
...
+
+            let (addr_slot, value) = item.wrap_err("ETL iteration failed")?;
+            plain_cursor.append_dup(
+                alloy_primitives::Address::from_slice(&addr_slot[..20]),
+                StorageEntry {
+                    key: B256::from_slice(&addr_slot[20..]),
+                    value: CompactU256::decompress_owned(value)?.into(),
+                },
+            )?;
+        }
```
yongkangc: maybe some commemnts here might be good.

bin/tempo/src/init_state.rs
```diff
@@ -157,78 +143,142 @@ impl<C: reth_cli::chainspec::ChainSpecParser<ChainSpec: EthChainSpec + EthereumH
...
+
+            let (addr_key, value) = item.wrap_err("ETL iteration failed")?;
+            hashed_cursor.append_dup(
+                B256::from_slice(&addr_key[..32]),
+                StorageEntry {
+                    key: B256::from_slice(&addr_key[32..]),
+                    value: CompactU256::decompress_owned(value)?.into(),
+                },
+            )?;
+        }
```
yongkangc: seems like a repeated patter, maybe we could dry this?

decofe:
cc @decofe

🚀 Benchmark started! [View job](https://github.com/tempoxyz/tempo/actions/runs/23580849539/job/68663221126)

**Config:** preset: `tip20`, duration: `300s`, bloat: `100000 MiB`, tps: `10000`, baseline: `alexey/init-state-etl`, feature: `alexey/init-state-etl`

decofe:
cc @decofe

🚀 Benchmark started! [View job](https://github.com/tempoxyz/tempo/actions/runs/23580854467/job/68663235770)

**Config:** preset: `tip20`, duration: `300s`, bloat: `100000 MiB`, tps: `10000`, baseline: `alexey/init-state-etl`, feature: `alexey/init-state-etl`

.github/workflows/bench.yml
```diff
@@ -381,7 +381,7 @@ jobs:
     needs: tempo-bench-ack
     name: tempo-bench
     runs-on: [self-hosted, Linux, X64, bare-metal]
-    timeout-minutes: 300
+    timeout-minutes: 1200
```
yongkangc:
```suggestion
    timeout-minutes: 300
```

xtask/src/generate_state_bloat.rs
```diff
@@ -100,138 +112,108 @@ impl GenerateStateBloat {
...
-        // - Remaining addresses use fast keccak-based derivation (not signable, just for bloat)
-        let actual_signable = signable_count.min(total_accounts);
-        let fast_count = total_accounts - actual_signable;
-
-        println!(
-            "\nDeriving {total_accounts} user addresses ({actual_signable} signable, {fast_count} fast)..."
-        );
-
-        // Parse mnemonic and derive parent key once (this is the slow PBKDF2 step)
+        // Step 1: Derive parent key (slow PBKDF2 step)
```
yongkangc:
```suggestion
        // Step 1: Derive parent key
```

bin/tempo/src/init_state.rs
```diff
@@ -188,20 +237,194 @@ impl<C: reth_cli::chainspec::ChainSpecParser<ChainSpec: EthChainSpec + EthereumH
...
+                        "{:.2}%",
+                        (index as f64 / total_plain as f64) * 100.0
+                    ),
+                    "Inserting plain storage"
+                );
+            }
+
+            let (addr_slot, value) = item.wrap_err("ETL iteration failed")?;
+            if let Some((ref prev_key, ref prev_val)) = pending_plain
+                && *prev_key != addr_slot
```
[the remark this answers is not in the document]
yongkangc: think this is okay 

The "last value seen" dedup here is not insertion-order-stable: Collector sorts by key only within each flush (par_sort_unstable_by(key)), but the merge heap orders equal keys by (key, value) lexicographically. So for duplicate slots the surviving value depends on byte-ordering of the compressed value, not on which was inserted later.

In practice this is safe today because (1) the generator never emits duplicate (address, slot) pairs and (2) clear wipes genesis before load, so no collision path exists.

bin/tempo/src/init_state.rs
```diff
@@ -123,35 +145,33 @@ impl<C: reth_cli::chainspec::ChainSpecParser<ChainSpec: EthChainSpec + EthereumH
...
+                let account = match account_cursor.seek_exact(address)? {
+                    Some((_, account)) => account,
+                    None => {
+                        let account = Account::default();
+                        account_cursor.upsert(address, &account)?;
+                        account
+                    }
+                };
+                e.insert(account);
+            }
```
shekhirin, line 195: I don't think it's ever the case that `accounts_seen.entry` will return `Occupied`? Because accounts are unique in the `state_bloat.bin`.

bin/tempo/src/init_state.rs
```diff
@@ -165,20 +185,50 @@ impl<C: reth_cli::chainspec::ChainSpecParser<ChainSpec: EthChainSpec + EthereumH
...
+                            if address != last_addr {
+                                last_addr = address;
+                                hashed_addr = keccak256(address);
+                            }
+                            let mut hashed_key = Vec::with_capacity(64);
+                            hashed_key.extend_from_slice(hashed_addr.as_slice());
+                            hashed_key.extend_from_slice(keccak256(slot).as_slice());
+                            let _ = tx.send((hashed_key, value));
+                        }
+                    });
```
shekhirin: can we instead spawn a single rayon thread and send the batches to it? without keeping that array of `pending_channels`

bin/tempo/src/init_state.rs
```diff
@@ -188,20 +238,194 @@ impl<C: reth_cli::chainspec::ChainSpecParser<ChainSpec: EthChainSpec + EthereumH
...
                     last_log = now;
                 }
             }
 
-            total_tokens += 1;
+            total_blocks += 1;
         }
 
+        // Spawn leftover chunk
+        if !hash_chunk.is_empty() {
```
shekhirin, line 244: suggestion with a single rayon thread that reads a batch from channel would also eliminate this, because the thread will drop once the last sender is dropped

bin/tempo/src/init_state.rs
yongkangc: that is much simpler, thanks for the note makes sense

bin/tempo/src/init_state.rs
yongkangc: think `Occupied` might still happen since the generator streams in chunks and so each chunk could emit a separate block for the same token address ([here](https://github.com/tempoxyz/tempo/blob/ecc7febb090d8ce2ae0b50b849a881bcd746a763/xtask/src/generate_state_bloat.rs#L190-L194)).

bin/tempo/src/init_state.rs
yongkangc: removed to a single std thread in ecc7feb 

think we still need `if !hash_chunk.is_empty()` to flushes the partial local batch before dropping the sender so we dont lose any entries in local `hash_chunk` `Vec` that wasnt sent

shekhirin: makes sense, yeah!

bin/tempo/src/init_state.rs
```diff
@@ -247,3 +443,28 @@ impl<C: reth_cli::chainspec::ChainSpecParser<ChainSpec: EthChainSpec + EthereumH
...
+/// Log collection progress every 5 seconds and on the final entry.
+fn log_collection_progress(
+    address: &alloy_primitives::Address,
+    index: u64,
+    total: u64,
+    start: Instant,
+    last_log: &mut Instant,
+) {
+    let now = Instant::now();
+    if now.duration_since(*last_log) >= Duration::from_secs(5) || index + 1 == total {
```
shekhirin:
```suggestion
    if last_log.elapsed() >= Duration::from_secs(5) || index + 1 == total {
```

bin/tempo/src/init_state.rs
```diff
@@ -165,43 +212,201 @@ impl<C: reth_cli::chainspec::ChainSpecParser<ChainSpec: EthChainSpec + EthereumH
...
+        if let Some((key, val)) = pending_hashed {
+            hashed_cursor.append_dup(
+                B256::from_slice(&key[..32]),
+                StorageEntry {
+                    key: B256::from_slice(&key[32..]),
+                    value: CompactU256::decompress_owned(val)?.into(),
+                },
+            )?;
+        }
+        drop(hashed_cursor);
```
shekhirin: this can be moved to a single function that we call with different tables as a generic, i think?

yongkangc: ah yes thanks for tha note, thats cleaner extracted it out to `load_etl_to_cursor` https://github.com/tempoxyz/tempo/pull/3325/changes#diff-9305376fc9c02e6e3c8fc28d661858b748454f8ae8009ce5cfab0b54b40beffcR404-R425

PR #3329, opened 2026-03-26
fix(init-state): compute and persist trie nodes after bulk load

yongkangc:
Closes the block-0 stall at 100GB+ scale.

init-from-binary-dump writes leaf data (HashedStorageState, PlainStorageState) but not intermediate trie nodes. The sparse trie cache on block 1 sees stale genesis trie nodes, mismatches, and falls back to a full sequential rebuild that takes 68+ seconds, stalling the node.

Adds a trie computation step after hashed data is written using reth's `StateRoot::root_with_progress()` loop to compute the full merkle trie and persist intermediate nodes to AccountsTrie/StoragesTrie.

Stacked on #3325.

### DIFF
--- a/bin/tempo/src/init_state.rs
+++ b/bin/tempo/src/init_state.rs
@@ -307,51 +307,29 @@ impl<C: reth_cli::chainspec::ChainSpecParser<ChainSpec: EthChainSpec + EthereumH
             );
         }
 
-        // Load sorted entries from the ETL collector into the database table.
+        // Load sorted entries from each ETL collector into its database table.
         // Strategy: iterate the sorted collector, deduplicate consecutive entries with
-        // the same composite key (address ++ slot), and bulk-insert via append_dup.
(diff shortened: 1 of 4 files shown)


bin/tempo/src/init_state.rs
```diff
@@ -430,35 +416,94 @@ impl<C: reth_cli::chainspec::ChainSpecParser<ChainSpec: EthChainSpec + EthereumH
...
+                        .seek_exact(*addr)
+                        .ok()
+                        .flatten()
+                        .map(|(_, acc)| acc)
+                        .unwrap_or_default();
+                    (*addr, Some(account))
+                })
+                .collect();
+            provider_rw.insert_account_for_hashing(accounts)?;
+        }
```
shekhirin: hmm, why are we doing this instead of collecting into `accounts_seen` first?

PR #3360, opened 2026-03-29
fix(transaction-pool): reject queued txs discarded on insert

joohhnnn:
Returns `DiscardedOnInsert` when an AA2d queued transaction is immediately evicted by `discard()` during insertion instead of reporting it as parked.

keeps AA2d pool insert semantics aligned with the standard pool and avoids reporting success for transactions that are no longer in the pool.

### DIFF
--- a/crates/transaction-pool/src/tt_2d_pool.rs
+++ b/crates/transaction-pool/src/tt_2d_pool.rs
@@ -305,16 +305,7 @@ impl AA2dPool {
         }
 
         // Call discard for queued transactions too
-        let discarded = self.discard();
-        if discarded
-            .iter()
-            .any(|discarded_tx| *discarded_tx.hash() == *transaction.hash())
-        {
-            return Err(PoolError::new(
-                *transaction.hash(),
-                PoolErrorKind::DiscardedOnInsert,
-            ));
-        }
+        let _ = self.discard();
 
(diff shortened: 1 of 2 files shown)


mattsse (changes requested):
this has unrelated changes

please reduce this to just the fix

crates/transaction-pool/src/tt_2d_pool.rs
mattsse, line 303: this can technically also happen here

mattsse (commented):
I believe this is correct and we should return an error here because it's possible that this will discard the tx itself if its queued.

but this now introduces some side-effects because this will then bypass:

https://github.com/tempoxyz/tempo/blob/e331e2f488a368ea9ca76c422e12fedbfb6eb3b4/crates/transaction-pool/src/tempo_pool.rs#L475-L495

so idk we should make this change like this

crates/transaction-pool/src/tt_2d_pool.rs
joohhnnn: moved the discard check from `tt_2d_pool.rs` to `tempo_pool.rs`. For parked txs, uses `!aa_2d_pool.contains(&hash)` within the write lock to detect if discard() evicted it. 
Both pending and queued paths are now handled at the same layer

PR #3366, opened 2026-03-30
docs(tip-1040): epoch-scoped temporary storage precompile

decofe:
Adds TIP-1040: a precompile for epoch-scoped temporary key-value storage. Values are available for 2^16 to 2^17 blocks then automatically expire. Nodes can prune anything older than the previous epoch.

Key design points:
- Epochs are 2^16 blocks. Current + previous epoch storage are live; older epochs are forgotten.
- Storage slot = `keccak256(sender || key)` scoped per `uint16(epoch)` — sender-isolated, no cross-contract collisions.
- Store costs a flat 40k gas (no state gas). Load follows EIP-2929 cold/warm model (2,100 / 100 gas).
- Load checks current epoch first, falls back to previous epoch, returns zero if both miss.

Prompted by: dankrad

### DIFF
--- a/tips/tip-1040.md
+++ b/tips/tip-1040.md
@@ -121,9 +121,10 @@ Loads the value for the derived slot, checking the current epoch first and falli
 1. Compute `hash = keccak256(msg.sender || key)`.
 2. Compute `current_epoch_index = uint32(epoch(block.number))`.
 3. Compute `current_key = current_epoch_index ++ hash[4:32]`. Look up `current_key` in the precompile's storage. If a non-zero value exists, return it.
-4. Compute `previous_epoch_index = uint32(epoch(block.number) - 1)`.
-5. Compute `previous_key = previous_epoch_index ++ hash[4:32]`. Look up `previous_key` in the precompile's storage. If a non-zero value exists, return it.
-6. Return `bytes32(0)`.
+4. If `current_epoch_index == 0`, return `bytes32(0)` (no previous epoch exists).
+5. Compute `previous_epoch_index = uint32(epoch(block.number) - 1)`.
+6. Compute `previous_key = previous_epoch_index ++ hash[4:32]`. Look up `previous_key` in the precompile's storage. If a non-zero value exists, return it.
+7. Return `bytes32(0)`.
 
 **Gas**: Follows the EIP-2929 access-list model. Each `storage_key` is tracked independently for hot/cold purposes:
 - First access to a given `storage_key` within the transaction: `TEMPORARY_LOAD_GAS_COLD` (2,100 gas).
--- a/tips/tip-1040.md
+++ b/tips/tip-1040.md
@@ -28,12 +28,12 @@ Epoch-scoped temporary storage solves this by:
 ## Assumptions
 
 - The block number is always available to the precompile at execution time.
-- Nodes maintain at least two complete epoch storage trees (current and previous) at all times. A node that has pruned the previous epoch's tree is non-conforming.
-- The precompile address is reserved and not used by any existing contract.
+- Nodes maintain at least two epoch accounts (current and previous) at all times. A node that has pruned the previous epoch's account is non-conforming.
+- The precompile address and the address range `[PRECOMPILE_ADDRESS + 1, ...]` are reserved and not used by any existing contract or EOA.
 - Epoch boundaries are deterministic: `epoch = block_number >> 16`.
-- Storage keys are collision-resistant due to the `keccak256(sender, key)` derivation; two different senders cannot write to the same slot.
+- Storage keys are collision-resistant due to the full `keccak256(sender, key)` derivation (256 bits); two different senders cannot write to the same slot.
 
-If the assumption that nodes keep two epochs is violated (e.g., a node prunes the previous epoch early), `temporaryLoad` may incorrectly return zero for data that should still be accessible.
+If the assumption that nodes keep two epoch accounts is violated (e.g., a node prunes the previous epoch account early), `temporaryLoad` may incorrectly return zero for data that should still be accessible.
 
 ---
 
@@ -62,18 +62,21 @@ The current epoch is `epoch(block.number)`. The previous epoch is `epoch(block.n
 
 ## Storage Layout
 
-All temporary storage lives in the precompile's own EVM storage trie (i.e., the account at `PRECOMPILE_ADDRESS`). Each slot is a 32-byte key composed of two parts:
+Each epoch's temporary storage lives in a **separate account**. Epoch `n` is stored in the account at address `PRECOMPILE_ADDRESS + n + 1`. The `+ 1` offset reserves `PRECOMPILE_ADDRESS` itself for the precompile dispatch logic.
 
 ```
-storage_key = uint32(epoch) ++ keccak256(sender || key)[4:32]
+epoch_account(n) = PRECOMPILE_ADDRESS + n + 1
 ```
 
-| Bytes | Field | Description |
-|-------|-------|-------------|
-| `[0:4)` | `uint32(epoch)` | Big-endian epoch index (2^32 epochs ≈ 2^48 blocks before wrap-around) |
-| `[4:32)` | `keccak256(sender \|\| key)[4:32]` | Last 28 bytes of the hash of `msg.sender` concatenated with the caller-provided key |
+Within each epoch account, storage keys are derived from the caller and key:
 
-This packs the epoch tag and the sender-derived slot into a single 32-byte EVM storage key. A 4-byte epoch index supports 2^32 epochs (2^48 blocks) before wrap-around, making the scheme effectively eternal. The 28-byte hash truncation provides 224 bits of collision resistance, which is sufficient given the sender-scoping.
+```
+storage_key = keccak256(sender || key)
+```
+
+This is the full 32-byte hash — no truncation is needed since the epoch is encoded in the account address rather than the storage key, giving 256 bits of collision resistance.
+
+**Pruning advantage**: Nodes can prune an entire expired epoch by dropping the account at `PRECOMPILE_ADDRESS + n + 1` and its storage trie in one operation, rather than scanning individual slots within a single large trie.
 
 ## Precompile Interface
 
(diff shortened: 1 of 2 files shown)


tips/tip-1040.md
```diff
@@ -0,0 +1,157 @@
...
+
+1. **Eliminating cleanup cost** — stale data becomes inaccessible automatically, no explicit deletion needed.
+2. **Reducing long-term state growth** — nodes can prune expired storage, keeping the state trie bounded.
+3. **Cheaper writes** — a flat 40,000 gas per store with no state gas overhead (no refund accounting, no cold/warm store distinction on writes).
+
+Alternatives considered:
+
+- **EIP-1153 transient storage (`TSTORE`/`TLOAD`)**: Too short-lived — data is discarded at the end of a single transaction. Does not cover multi-block use cases.
+- **Application-level expiry with `SSTORE`**: Requires explicit cleanup transactions, is more expensive, and leaves dead state in the trie permanently.
+- **Off-chain data with on-chain commitments**: Adds complexity and trust assumptions around data availability.
```
dankrad:
```suggestion
```

tips/tip-1040.md
```diff
@@ -0,0 +1,152 @@
...
+
+## Motivation
+
+Certain use cases need on-chain data availability for a bounded time window rather than permanent storage — for example time-limited approvals, ephemeral session state, or temporary commitments. Today these patterns require permanent `SSTORE` writes and separate cleanup transactions, wasting gas and bloating state.
+
+Epoch-scoped temporary storage solves this by:
+
+1. **Eliminating cleanup cost** — stale data becomes inaccessible automatically, no explicit deletion needed.
+2. **Reducing long-term state growth** — nodes can prune expired storage, keeping the state trie bounded.
+3. **Cheaper writes** — a flat 40,000 gas per store with no state gas overhead (no refund accounting, no cold/warm store distinction on writes).
```
dankrad:
```suggestion
3. **Cheaper writes** — a flat 40,000 gas per store with no additional state gas (no refund accounting, no cold/warm store distinction on writes).
```

tips/tip-1040.md
```diff
@@ -0,0 +1,253 @@
...
+
+Within each epoch account, storage keys are derived from the caller and key:
+
+```
+storage_key = keccak256(sender || key)
+```
+
+This is the full 32-byte hash — no truncation is needed since the epoch is encoded in the account address rather than the storage key, giving 256 bits of collision resistance.
+
+**Pruning advantage**: Nodes can prune an entire expired epoch by dropping the account at `PRECOMPILE_ADDRESS + n + 1` and its storage trie in one operation, rather than scanning individual slots within a single large trie.
```
dankrad:
```suggestion
**Pruning advantage**: Nodes can prune an entire expired epoch by dropping the account at `PRECOMPILE_ADDRESS + n + 1` and its storage trie in one operation, rather than scanning individual slots within a single large trie. Note that this dropping of storage is done out of consensus -- there is no actual deletion happening and the account trie does not change, the node can simply forget the storage for the account because it is guaranteed to not be needed again any time in the future.
```

tips/tip-1040.md
```diff
@@ -0,0 +1,253 @@
...
+
+## Epoch Transition
+
+When `block.number` crosses an epoch boundary (i.e., `epoch(block.number) != epoch(block.number - 1)`):
+
+1. The account at `PRECOMPILE_ADDRESS + epoch(block.number) - 1` (two epochs ago) becomes unreachable by the precompile and may be pruned by the node.
+2. The account at `PRECOMPILE_ADDRESS + epoch(block.number) + 1` (current epoch) begins accepting writes.
+3. The account at `PRECOMPILE_ADDRESS + epoch(block.number)` (previous epoch) remains accessible for read fallback.
+
+Nodes MUST retain the accounts for the current and previous epoch. Nodes MAY delete any older epoch account and its entire storage trie.
```
dankrad:
```suggestion
Nodes MUST retain the accounts for the current and previous epoch. Nodes MAY delete any older epoch account and its entire storage trie, however they MUST retain all information that is still needed for future operation, such as the Merkle root of the storage trie in order to be able to update the accounts trie in the future.
```

tips/tip-1040.md
```diff
@@ -0,0 +1,253 @@
...
+1. **Eliminating cleanup cost** — stale data becomes inaccessible automatically, no explicit deletion needed.
+2. **Reducing long-term state growth** — nodes can prune expired storage, keeping the state trie bounded.
+3. **Cheaper writes** — 40,000 gas for new slots; overwrites within the same epoch follow standard SSTORE hot/cold pricing, much cheaper than a fresh write.
+
+
+## Assumptions
+
+- The block number is always available to the precompile at execution time.
+- Nodes maintain at least two epoch accounts (current and previous) at all times. A node that has pruned the previous epoch's account is non-conforming.
+- The precompile address and the address range `[PRECOMPILE_ADDRESS + 1, ...]` are reserved and not used by any existing contract or EOA.
```
Rjected: end of the range we expect being, explicitly, `PRECOMPILE_ADDRESS + u64::MAX + 1`?

tips/tip-1040.md
```diff
@@ -0,0 +1,253 @@
...
+
+Within each epoch account, storage keys are derived from the caller and key:
+
+```
+storage_key = keccak256(sender || key)
+```
+
+This is the full 32-byte hash — no truncation is needed since the epoch is encoded in the account address rather than the storage key, giving 256 bits of collision resistance.
+
+**Pruning advantage**: Nodes can prune an entire expired epoch by dropping the account at `PRECOMPILE_ADDRESS + n + 1` and its storage trie in one operation, rather than scanning individual slots within a single large trie. Note that this dropping of storage is done out of consensus -- there is no actual deletion happening and the account trie does not change, the node can simply forget the storage for the account because it is guaranteed to not be needed again any time in the future.
```
Rjected: > rather than scanning individual slots within a single large trie

> there is no actual deletion happening

IMO this can be worded better, because we will be deleting the storage of expired epochs and their trie nodes from the node. And while we are not reading the entire trie before deleting it, we do have to _delete_ all the storage and trie pages. So I am not sure what the value of this section is / what this is an advantage _over_, other than a hypothetical implementation. it might be be easier to just mention that the account trie does not change, despite nodes deleting the underlying data.

lightclient: > Nodes can prune an entire expired epoch by dropping the account

this can also be worded better, because you would still need to keep the account around for things like `EXTCODE*` and `EXTBALANCE`

tips/tip-1040.md
```diff
@@ -0,0 +1,253 @@
...
+Epoch-scoped temporary storage solves this by:
+
+1. **Eliminating cleanup cost** — stale data becomes inaccessible automatically, no explicit deletion needed.
+2. **Reducing long-term state growth** — nodes can prune expired storage, keeping the state trie bounded.
+3. **Cheaper writes** — 40,000 gas for new slots; overwrites within the same epoch follow standard SSTORE hot/cold pricing, much cheaper than a fresh write.
+
+
+## Assumptions
+
+- The block number is always available to the precompile at execution time.
```
dankrad:
```suggestion
```

Deleting because this is obvious (has an EVM opcode)

tips/tip-1040.md
dankrad:
```suggestion
- The precompile address and the address range `[PRECOMPILE_ADDRESS + 1, ..., PRECOMPILE_ADDRESS + u64::MAX]` are reserved and not used by any existing contract or EOA.
```

tips/tip-1040.md
```diff
@@ -0,0 +1,253 @@
...
+- Nodes maintain at least two epoch accounts (current and previous) at all times. A node that has pruned the previous epoch's account is non-conforming.
+- The precompile address and the address range `[PRECOMPILE_ADDRESS + 1, ...]` are reserved and not used by any existing contract or EOA.
+- Epoch boundaries are deterministic: `epoch = block_number / EPOCH_LENGTH`.
+- Storage keys are collision-resistant due to the full `keccak256(sender, key)` derivation (256 bits); two different senders cannot write to the same slot.
+
+If the assumption that nodes keep two epoch accounts is violated (e.g., a node prunes the previous epoch account early), `temporaryLoad` may incorrectly return zero for data that should still be accessible.
+
```
dankrad:
```suggestion
```

tips/tip-1040.md
```diff
@@ -0,0 +1,253 @@
...
+status: Draft
+related: N/A
+protocolVersion: TBD
+---
+
+# TIP-1040: Epoch-Scoped Temporary Storage Precompile
+
+## Abstract
+
+This TIP introduces a precompile that provides temporary key-value storage scoped to epochs. An epoch is 172,800 blocks, which is approximately 24 hours at 500ms slot times. Data written in a given epoch is readable during that epoch and the next, then automatically becomes inaccessible. This gives stored values a lifetime of between 172,800 and 345,600 blocks depending on when within the epoch the write occurs. Nodes may prune storage from epochs older than the previous one.
```
dankrad:
```suggestion
This TIP introduces a precompile that provides temporary key-value storage scoped to epochs. An epoch is 172,800 blocks, which is approximately 24 hours at 500ms slot times (the actual time can vary due to block time variation and skipped slots; consumers need to take this into account and can only rely on the minimum time since there is no strict upper bound). Data written in a given epoch is readable during that epoch and the next, then automatically becomes inaccessible. This gives stored values a lifetime of between 172,800 and 345,600 blocks depending on when within the epoch the write occurs. Nodes may prune storage from epochs older than the previous one.
```

tips/tip-1040.md
dankrad:
```suggestion
**Pruning advantage**: Nodes can prune an entire expired epoch by dropping the account at `PRECOMPILE_ADDRESS + n + 1` and its storage trie in one operation, rather than scanning individual slots within a single large trie. Note that this dropping of storage is done out of consensus -- there is no actual deletion happening inside consensus logic and the account trie does not change, the node can simply forget the storage for the account because it is guaranteed to not be needed again any time in the future.
```

dankrad: Right, there is database deletion, but from the point of view of consensus logic, nothing actually gets deleted. Adding "inside consensus logic" to clarify this.

tips/tip-1040.md
```diff
@@ -0,0 +1,253 @@
+---
+id: TIP-1040
+title: Epoch-Scoped Temporary Storage Precompile
+description: A precompile providing cheap temporary key-value storage that automatically expires after one to two epochs.
+authors: Dankrad
```
dankrad:
```suggestion
authors: Dankrad Feist @dankrad
```

PR #3384, opened 2026-03-30
feat: implement TIP-1011 enhanced access key permissions

legion2002:
Implements TIP-1011 end to end for T3 access keys. Adds periodic TIP-20 spending limits, per-target and per-selector call scopes, and recipient-constrained token calls across the spec, authorization payload, AccountKeychain precompile, runtime enforcement, and txpool admission.

The key-authorization payload now carries exact RLP and JSON semantics for `limits` and `allowed_calls`, including the distinction between unrestricted keys and deny-all scoped keys. The precompile persists the new restriction state, validates uniqueness and constrained-selector invariants, exposes the new allowed-call and spending-limit interfaces, hides stale scopes for revoked or expired keys, and keeps the T3 access-key creation restrictions intact.

Handler and txpool validation now reject malformed TIP-1011 authorizations before execution and enforce selector and recipient scoping consistently with onchain storage. The diff also extends coverage for RLP roundtrips, periodic rollover, allowed-call normalization, recipient-constrained TIP-20 selectors, and access-key execution paths.


crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -21,18 +21,142 @@ pub struct TokenLimit {
...
     /// Maximum spending amount for this token (enforced over the key's lifetime)
     pub limit: U256,
+
+    /// Period duration in seconds.
+    ///
+    /// `0` means one-time limit. `>0` means the limit resets periodically.
+    pub period: u64,
+}
+
+impl Decodable for TokenLimit {
```
klkvr: could this not be done with an `rlp(trailing)`?

if we wanted to avoid an extra `Option` we could still rely on derive macro here via a helper type

klkvr: also i think it's worth looking into unifying this with struct defined in `tempo-contracts`

crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -60,6 +184,74 @@ pub struct KeyAuthorization {
...
     pub limits: Option<Vec<TokenLimit>>,
+
+    /// Optional call scopes for this key.
+    /// - `None` (RLP 0x80) = unrestricted calls
+    /// - `Some([])` = scoped mode with no allowed calls (deny-all)
+    /// - `Some([CallScope{...}])` = explicit target/selector scope list
+    pub allowed_calls: Option<Vec<CallScope>>,
+}
+
+impl Decodable for KeyAuthorization {
```
klkvr: did derive not work properly for this?

crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -113,13 +317,68 @@ impl KeyAuthorization {
         Ok(())
     }
 
+    /// Returns the number of storage rows written for scoped-call state.
+    ///
+    /// This mirrors the writes performed by `authorize_key -> replace_allowed_calls ->
+    /// upsert_target_scope` in the account keychain precompile.
+    pub fn call_scope_storage_slots(&self) -> u64 {
```
klkvr: doesn't feel like this belongs here

crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -21,18 +21,142 @@ pub struct TokenLimit {
...
+    /// Optional selector rules for this target.
+    pub selector_rules: Option<Vec<SelectorRule>>,
+}
+
+/// Selector-level rule within a [`CallScope`].
+///
+/// `recipients` semantics:
+/// - `None` => no recipient constraint
+/// - `Some([..])` => first ABI address argument must be in this list
+#[derive(Clone, Debug, PartialEq, Eq, Hash, alloy_rlp::RlpEncodable, alloy_rlp::RlpDecodable)]
```
0xKitsune, line 69: `SelectorRule` uses derived `RlpDecodable` and `recipients: Some([])` decodes successfully, however the spec says it `MUST` be rejected. This should have a custom Decodable impl. Alternatively we need to make sure this case is rejected downstrea.

crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -21,18 +21,142 @@ pub struct TokenLimit {
...
+#[rlp(trailing)]
```
0xKitsune: `CallScope` and `SelectorRule` use #[rlp(trailing)] which omits `None` fields instead of encoding `0x80` (which the spec specifies).

crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -113,13 +318,68 @@ impl KeyAuthorization {
...
+    }
+
     /// Calculates a heuristic for the in-memory size of the key authorization
     pub fn size(&self) -> usize {
         size_of::<Self>()
             + self
                 .limits
                 .as_ref()
                 .map_or(0, |limits| limits.capacity() * size_of::<TokenLimit>())
+            + self.allowed_calls.as_ref().map_or(0, |scopes| {
```
fgimenez, line 208: nit, the rest of the codebase delegates heap accounting to child `size()` methods (like `TempoTransaction::size()` -> `call.size()`, `key_auth.size()`). in this we are inlining 5 levels of nested iterators instead. maybe we could add `heap_size()` helper on `CallScope` and `SelectorRule` so this stays flat, wdyt?

crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -60,6 +185,74 @@ pub struct KeyAuthorization {
     /// - `Some([])` = no spending allowed (enforce_limits=true but no tokens allowed)
     /// - `Some([TokenLimit{...}])` = specific limits enforced
     pub limits: Option<Vec<TokenLimit>>,
+
+    /// Optional call scopes for this key.
+    /// - `None` (RLP 0x80) = unrestricted calls
```
fgimenez: this would match the spec and what the code actually does
```suggestion
    /// - `None` (omitted) = unrestricted calls
```

crates/contracts/src/precompiles/account_keychain.rs
```diff
@@ -72,22 +123,48 @@ crate::sol! {
...
         /// @param publicKey The public key
         /// @return Key information
         function getKey(address account, address keyId) external view returns (KeyInfo memory);
 
-        /// Get remaining spending limit
+        /// Get remaining spending limit using the legacy pre-T3 return shape.
         /// @param account The account address
         /// @param publicKey The public key
         /// @param token The token address
-        /// @return Remaining spending amount
```
fgimenez, line 85: should we keep this comment?

crates/primitives/src/transaction/key_authorization.rs
legion2002: yeah we have checks that ensure this is rejected downstream.

crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -21,18 +21,142 @@ pub struct TokenLimit {
...
+}
+
+/// Per-target call scope for an access key.
+///
+/// `selector_rules` uses tri-state semantics:
+/// - `None` => allow any selector for this target
+/// - `Some([])` => deny all selectors for this target
+/// - `Some([..])` => allow exactly the listed selector rules
+#[derive(Clone, Debug, PartialEq, Eq, Hash, alloy_rlp::RlpEncodable, alloy_rlp::RlpDecodable)]
+#[rlp(trailing)]
```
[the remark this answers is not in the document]
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/59bc969aa

crates/primitives/src/transaction/key_authorization.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/59bc969aa

crates/primitives/src/transaction/key_authorization.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/697bba0acc86d1ab639fe34edcd275fb08ba9e1c

crates/primitives/src/transaction/key_authorization.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/59bc969aa

crates/primitives/src/transaction/key_authorization.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/59bc969aa

crates/primitives/src/transaction/key_authorization.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/59bc969aa

crates/contracts/src/precompiles/account_keychain.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/59bc969aa

crates/primitives/src/transaction/key_authorization.rs
legion2002: I can move this to the handler.

legion2002: addressed in https://github.com/tempoxyz/tempo/pull/3384/changes/9a628b05037d3baedf8913664b87d93c44229407

and added in 3386 here
https://github.com/tempoxyz/tempo/pull/3386/changes/f5c06cf10f6fac4f09beb12976d80d03c6b1fefd

crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -21,17 +21,121 @@ pub struct TokenLimit {
...
+            period: (value.period != 0).then_some(value.period),
+        }
+    }
+}
+
+impl Decodable for TokenLimit {
+    fn decode(buf: &mut &[u8]) -> alloy_rlp::Result<Self> {
+        Ok(TokenLimitWire::decode(buf)?.into())
+    }
+}
```
klkvr: let's move those impls and helper structs to a `mod rlp` at the end of the file

legion2002: addressed in https://github.com/tempoxyz/tempo/commit/3283ab1be32d7478e2021a5040b19e180810bc34

tips/tip-1011.md
```diff
@@ -1,313 +1,620 @@
...
+1. `selectorRules = []` sets `selector_rules = None` semantics (any selector allowed on `target`).
+2. `removeAllowedCalls(keyId, target)` disables that target scope.
+3. Implementations MUST enforce at most one scope per target for each `(account, key)`.
+4. Selector rules MUST be unique by `selector` within a target scope.
+5. If any rule has `recipients = Some([..])`, `target` MUST be a TIP-20 token address.
+6. If any rule has `recipients = Some([..])`, its `selector` MUST be one of:
+   1. `0xa9059cbb` (`transfer(address,uint256)`)
+   2. `0x095ea7b3` (`approve(address,uint256)`)
+   3. `0x95777d59` (`transferWithMemo(address,uint256,bytes32)`)
+8. If any rule has `recipients = Some([..])`, each recipient in that list MUST be non-zero.
```
0xKitsune: minor nit: this skips 7

tips/tip-1011.md
```diff
@@ -1,313 +1,620 @@
...
+16. `setAllowedCalls` create-or-replace semantics are enforced.
+17. `removeAllowedCalls(keyId, target)` removes that target scope and leaves the key in deny-all mode if no target scopes remain.
+19. Address-only scopes allow selectorless/fallback-style calls to the scoped target.
+20. Access-key transactions with CREATE as first call are rejected.
+21. Access-key transactions with any CREATE in a batch are rejected.
+22. For constrained TIP-20 selectors (`transfer`, `approve`, `transferWithMemo`), calls succeed iff calldata argument `0` is in the configured recipient set.
+23. Single-recipient and multi-recipient selector rules both enforce the same membership rule.
+24. Reject the transaction before execution when a selector rule with a recipient allowlist is matched and calldata is shorter than `4 + 32` bytes.
+25. Reject the transaction before execution when a selector rule with a recipient allowlist is matched and ABI argument `0` is not canonically encoded as an address.
+27. Reject selector rules with recipient allowlists for selectors outside the fixed constrained-selector set.
```
0xKitsune: minor nit: same thing here, this skips 26

tips/tip-1011.md
```diff
@@ -1,313 +1,620 @@
...
+9. Existing keys continue to function after the fork.
+10. Batch validation rejects the transaction before execution when any call is invalid.
+11. Shared key IDs across accounts cannot overwrite each other’s scopes.
+12. Reject `allowed_calls` entries with `S > MAX_CALL_SCOPES`.
+13. Reject any target scope with selector-rule count `> MAX_SELECTOR_RULES_PER_SCOPE`.
+14. Reject calls that do not provide at least 4 selector bytes when explicit selector matching is required.
+15. `setAllowedCalls(..., scope.selectorRules = [])` allows any selector on that target.
+16. `setAllowedCalls` create-or-replace semantics are enforced.
+17. `removeAllowedCalls(keyId, target)` removes that target scope and leaves the key in deny-all mode if no target scopes remain.
+19. Address-only scopes allow selectorless/fallback-style calls to the scoped target.
```
0xKitsune: minor nit: this skips 18

0xKitsune (commented):
Overall lgtm but left a few additional comments regarding missing tests.

0xKitsune (commented):
Flagging a few more test cases we should add to cover the new types introduced.

0xKitsune (commented):
Wire format test gaps

crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -266,14 +351,265 @@ mod tests {
...
+
+        let mut reencoded = Vec::new();
+        decoded.encode(&mut reencoded);
+
+        assert_eq!(reencoded, encoded);
+    }
+
+    #[test]
+    fn test_call_scope_decode_accepts_omitted_none_selector_rules() {
+        let target = Address::repeat_byte(0x11);
```
0xKitsune, line 563: Lets add a roundtrip test for `CallScope` with `selector_rules = Some([])` (deny-all). Only `None` and `Some([rule])` are covered.

crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -21,17 +21,77 @@ pub struct TokenLimit {
...
+#[rlp(trailing)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
+#[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
+pub struct SelectorRule {
+    /// 4-byte function selector.
+    pub selector: [u8; 4],
+    /// Optional recipient allowlist.
+    pub recipients: Option<Vec<Address>>,
+}
```
0xKitsune, line 83: Can we add a roundtrip test for `SelectorRule` with a non-empty recipients list. Only `None` and empty are tested.

crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -60,6 +120,12 @@ pub struct KeyAuthorization {
     /// - `Some([])` = no spending allowed (enforce_limits=true but no tokens allowed)
     /// - `Some([TokenLimit{...}])` = specific limits enforced
     pub limits: Option<Vec<TokenLimit>>,
+
+    /// Optional call scopes for this key.
+    /// - `None` (canonically omitted, explicit 0x80 accepted) = unrestricted calls
+    /// - `Some([])` = scoped mode with no allowed calls (deny-all)
```
0xKitsune: Can we add a decode test for `allowed_calls = Some([])` (deny-all). The existing test covers `0x80` -> `None` but not `0xc0 → Some([])`.

0xKitsune (approved):

crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -328,3 +664,55 @@ mod tests {
         assert_eq!(err.got, 999);
     }
 }
+
+mod rlp {
```
klkvr: sorry it should still be above `mod tests` but below any definitions/impl blocks

crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -21,17 +21,77 @@ pub struct TokenLimit {
...
+/// `selector_rules` uses tri-state semantics:
+/// - `None` => allow any selector for this target
+/// - `Some([])` => deny all selectors for this target
+/// - `Some([..])` => allow exactly the listed selector rules
+#[derive(Clone, Debug, PartialEq, Eq, Hash, alloy_rlp::RlpEncodable, alloy_rlp::RlpDecodable)]
+#[rlp(trailing)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
+#[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
+pub struct CallScope {
```
fgimenez, line 42: could we add here `#[cfg_attr(test, reth_codecs::add_arbitrary_tests(rlp))]`?

crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -21,17 +21,77 @@ pub struct TokenLimit {
...
+///
+/// `recipients` semantics:
+/// - `None` => no recipient constraint
+/// - `Some([..])` => first ABI address argument must be in this list
+#[derive(Clone, Debug, PartialEq, Eq, Hash, alloy_rlp::RlpEncodable, alloy_rlp::RlpDecodable)]
+#[rlp(trailing)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
+#[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
+pub struct SelectorRule {
```
fgimenez, line 74: same, missing `#[cfg_attr(test, reth_codecs::add_arbitrary_tests(rlp))]`

fgimenez (approved):

crates/primitives/src/transaction/key_authorization.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/pull/3384/commits/92ecda564f7e02a059a7876411bab3b8a657b5ac

crates/primitives/src/transaction/key_authorization.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/pull/3384/commits/92ecda564f7e02a059a7876411bab3b8a657b5ac

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -273,19 +419,158 @@ impl AccountKeychain {
...
+        self.remove_target_scope(key_hash, call.target)?;
+
+        Ok(())
+    }
+
+    /// Returns whether an account key is call-scoped together with its configured call scopes.
+    ///
+    /// `isScoped = false` means unrestricted. `isScoped = true` with an empty `scopes` vec means
+    /// the key is scoped but currently allows no targets.
+    pub fn get_allowed_calls(&self, call: getAllowedCallsCall) -> Result<getAllowedCallsReturn> {
```
0xrusowsky, line 523: > this fn is flawed, as it doesn't account for revoked/expired keys.
>
> despite it is intended to be a public getter, i think this is dangerous and prone to misuse. IMO we should add revoked + expiry checks.
>
> if we decide we don't need to (which i think would be an error), at the very least we need clear doc comments here and in the solidity interfaces.

it was marked as resolved in https://github.com/tempoxyz/tempo/pull/3385#discussion_r3021834321 but IIUIC it wasn't resolved

legion2002: hmm that's weird, I even responded to it, checking.

legion2002: sorry we had fixed this for get_remaining_limits, which also had the same issue.
I have addressed this for get_approved_calls here - https://github.com/tempoxyz/tempo/pull/3384/changes/16cde1ec2f9b6a7cbd66fa9a11bbee1791558576

crates/primitives/src/transaction/key_authorization.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/pull/3384/commits/5e87eae2422ac4e8ade5d92c06b69ea51c58a381

crates/primitives/src/transaction/key_authorization.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/pull/3384/commits/5e87eae2422ac4e8ade5d92c06b69ea51c58a381

crates/primitives/src/transaction/key_authorization.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/pull/3384/commits/5e87eae2422ac4e8ade5d92c06b69ea51c58a381

crates/primitives/src/transaction/key_authorization.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/pull/3384/commits/5e87eae2422ac4e8ade5d92c06b69ea51c58a381

crates/revm/src/handler.rs
```diff
@@ -126,6 +129,97 @@ fn tempo_signature_verification_gas(signature: &TempoSignature) -> u64 {
...
+                }
+            }
+
+            // Storage write accounting:
+            // - account mode write: 1
+            // - each target insertion + target mode write: 3 + 1
+            // - each selector insertion + selector mode write: 3 + 1
+            // - recipient-constrained selectors also write recipient set length: +1 per selector
+            // - recipient set values+positions: +2 per recipient
+            1 + scopes.len() as u64 * 4 + selectors * 4 + constrained_selectors + recipients * 2
```
0xrusowsky: i think this should be `3` rather than `4`, cause there is no target mode nor selector mode now
```suggestion
            // Storage write accounting:
            // - account mode write: 1
            // - each target insertion: 3
            // - each selector insertion: 3
            // - recipient-constrained selectors also write recipient set length: +1 per selector
            // - recipient set values+positions: +2 per recipient
            1 + scopes.len() as u64 * 3 + selectors * 3 + constrained_selectors + recipients * 2
```

crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -21,17 +21,80 @@ pub struct TokenLimit {
...
+#[rlp(trailing)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
+#[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
+#[cfg_attr(test, reth_codecs::add_arbitrary_tests(rlp))]
+pub struct SelectorRule {
+    /// 4-byte function selector.
+    pub selector: [u8; 4],
+    /// Optional recipient allowlist.
+    pub recipients: Option<Vec<Address>>,
```
klkvr: can we remove this `Option`?

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -319,6 +621,323 @@ impl AccountKeychain {
...
+        self.replace_allowed_calls(limit_key, allowed_calls)
+    }
+
+    /// Validates a top-level call against scoped permissions for this key.
+    ///
+    /// Validation walks the scope tree from coarse to fine:
+    /// - `is_scoped = false` => unrestricted key
+    /// - target missing from `targets` => target denied
+    /// - target present with `selectors = []` => allow any selector on that target
+    /// - selector missing from `selectors` => selector denied
```
0xrusowsky, line 675: addressed in [`6abaffe` (this PR)](https://github.com/tempoxyz/tempo/pull/3384/commits/6abaffe1283c925ea89aed17c62f458633f8310e)

crates/revm/src/handler.rs
```diff
@@ -1008,18 +1137,26 @@ where
...
                     keyId: access_key_addr,
                     signatureType: signature_type,
-                    expiry,
-                    enforceLimits: enforce_limits,
-                    limits: precompile_limits,
+                    config: KeyRestrictions {
+                        expiry,
+                        enforceLimits: enforce_limits,
+                        limits: precompile_limits,
+                        enforceAllowedCalls: enforce_allowed_calls,
```
klkvr: i'm wondering if a better name for this would be `allowAnyCalls`? right now it's easy to forget whether `enforceAllowedCalls=false` means that no calls are allowed or that all calls are allowed

crates/revm/src/handler.rs
```diff
@@ -1099,8 +1237,22 @@ where
...
+                                .is_some_and(|key_auth| {
+                                    key_auth.key_type != keychain_sig.signature.signature_type()
+                                })
+                        {
+                            return Err(TempoInvalidTransaction::KeychainValidationFailed {
+                                reason: "key authorization key_type does not match the keychain signature type"
+                                    .to_string(),
+                            }
+                            .into());
+                        }
```
klkvr, line 1220: this check seems new? what's the motivation for it?

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -273,19 +419,175 @@ impl AccountKeychain {
...
+
+        let current_timestamp = self.storage.timestamp().saturating_to::<u64>();
+        self.load_active_key(msg_sender, call.keyId, current_timestamp)?;
+
+        let key_hash = Self::spending_limit_key(msg_sender, call.keyId);
+        let scopes = call.scopes;
+
+        if scopes.is_empty() {
+            return Ok(());
+        }
```
klkvr, line 480: this is still slightly confusing to me. imo this should either allow any calls from the key or flp the `is_scoped` flag

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -8,23 +8,50 @@
```
fgimenez, line 8: this file almost doubled its size to ~4K LOC, could we extract the new scope tree operations and periodic limit logic to separate module(s)? not needed in this PR ofc, maybe followup

crates/precompiles/src/account_keychain/dispatch.rs
```diff
@@ -27,10 +74,56 @@ impl Precompile for AccountKeychain {
                         self.update_spending_limit(sender, c)
                     })
                 }
+                IAccountKeychainCalls::setAllowedCalls(call) => {
+                    if !self.storage.spec().is_t3() {
+                        return unknown_selector(
+                            setAllowedCallsCall::SELECTOR,
+                            self.storage.gas_used(),
+                        );
+                    }
```
0xKitsune, line 83: nit: we have this pattern in many of the dispatch files. Might make sense to make a helper function, not important now but just noting for a future PR.

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -319,6 +621,323 @@ impl AccountKeychain {
...
+        let target = scope.target;
+
+        // The public API uses the absence of a target to block it, so persisting address(0) as a
+        // real target is always confusing and serves no useful purpose.
+        if target.is_zero() {
+            return Err(AccountKeychainError::invalid_call_scope().into());
+        }
+
+        if !scope.selectorRules.is_empty() {
+            self.validate_selector_rules(target, &scope.selectorRules)?;
```
klkvr, line 853: this should be unified with https://github.com/tempoxyz/tempo/blob/7499dd126aa7211d8d29cb1b7578ca7a4395914f/crates/transaction-pool/src/validator.rs#L229-L272

fgimenez (approved):
lgtm!

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -273,19 +419,175 @@ impl AccountKeychain {
...
+        )?;
+
+        Ok(getRemainingLimitReturn {
+            remaining,
+            periodEnd: period_end,
+        })
+    }
+
+    /// Root-only create-or-replace updates for one or more target call scopes.
+    pub fn set_allowed_calls(
```
0xKitsune, line 461: This seems valid, is there a reason we dont emit an event when scope is changed?

crates/precompiles/src/account_keychain/mod.rs
legion2002: if we just flip the `is_scoped` flag, then we'll have kind of invalid state. Because someone would add 3 scopes, then call this fn with empty scope.
We'd flip the flag to allow all scopes.

Then someone adds some scopes back in. We'll flip the flag back to allow the new scopes. But the old scopes that were stale in the mapping would also be reactivated.

crates/precompiles/src/account_keychain/mod.rs
legion2002: yeah can cleanup in a followup!

crates/precompiles/src/account_keychain/mod.rs
0xKitsune: Isnt this what the remaining logic is doing? If scopes is empty we are returning early and allowing any calls from the key. If scopes is not empty, we upsert the target scope and flips the `is_scoped` flag. Lmk if I am missing something though.

crates/precompiles/src/account_keychain/mod.rs
legion2002: addressed in - https://github.com/tempoxyz/tempo/commit/bd16e5dfb2df1e8e0e9f71e07657ffc50423f13a

crates/revm/src/handler.rs
```diff
@@ -126,6 +129,97 @@ fn tempo_signature_verification_gas(signature: &TempoSignature) -> u64 {
...
+                    for rule in rules {
+                        if let Some(rule_recipients) = rule.recipients.as_ref() {
+                            constrained_selectors += 1;
+                            recipients += rule_recipients.len() as u64;
+                        }
+                    }
+                }
+            }
+
+            // Storage write accounting:
```
legion2002, line 155: addressed in - https://github.com/tempoxyz/tempo/commit/f13e3f3f0147cf1f244b5a927dde7afed3c7fcc3

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -273,19 +419,175 @@ impl AccountKeychain {
...
+                        .selector_scopes[selector]
+                        .recipients
+                        .read()?
+                        .into();
+
+                    rules.push(SelectorRule {
+                        selector,
+                        recipients,
+                    });
+                }
```
0xKitsune, line 576: nit: consider using collect, feel free to ignore though if preferred.

```rust
let rules = selectors
    .into_iter()
    .map(|selector| {
        let recipients: Vec<Address> = self.key_scopes[key_hash].target_scopes[target]
            .selector_scopes[selector]
            .recipients
            .read()?
            .into();
        Ok(SelectorRule { selector, recipients })
    })
    .collect::<Result<Vec<_>>>()?;
```

0xKitsune (approved):

crates/precompiles/src/account_keychain/mod.rs
legion2002: Yes, but we don't allow the user to explicitly flip the `is_scoped` flag in any of the flows. For key authorizations, this is the init of the key, so we can be sure that there is no stale set.

For `remove_allowed_calls` we don't flip the flag if all targets are removed. Because this would be a big footgun, you remove all allowed calls from a key, and suddenly the key is given unrestricted access.

legion2002: But I agree we shouldn't skip silently here, I'll add a revert for this.

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -442,20 +1124,62 @@ impl AccountKeychain {
...
+        }
+
+        let mut limit_state = self.spending_limits[limit_key][token].read()?;
+        let mut remaining = limit_state.remaining;
+        let is_periodic = limit_state.period != 0;
+
+        if is_periodic && current_timestamp >= limit_state.period_end {
+            let next_end = limit_state.compute_next_period_end(current_timestamp);
+
+            remaining = U256::from(limit_state.max);
```
legion2002, line 1147: We've discussed this, and decided that this is valid, but no good way to fix it untill we restrict access keys to only have transient approvals. We will propose this as a TIP for T4/T5.

But it is okay to leave this in the PR for now, since for most cases this is just equivalent to transferring out the amount every period.

crates/precompiles/src/account_keychain/mod.rs
legion2002: yes, we will fix this in a followup pool PR

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -276,7 +276,13 @@ where
                 && let Some(ref subject) = keychain_subject
                 && subject.matches_spending_limit_update(&updates.spending_limit_spends)
                 && let Some(ref mut provider) = state_provider
-                && exceeds_spending_limit(provider, subject, tx.transaction.fee_token_cost())
+                && exceeds_spending_limit(
```
legion2002, line 279: we will address this in a follow up tx pool PR

crates/precompiles/src/account_keychain/mod.rs
legion2002: addressing in #3422

crates/transaction-pool/src/validator.rs
```diff
@@ -154,6 +160,211 @@ where
...
+        let Some(scopes) = auth.allowed_calls.as_ref() else {
+            return Ok(Ok(()));
+        };
+
+        if scopes.len() > MAX_CALL_SCOPES as usize {
+            return Ok(Err(TempoPoolTransactionError::Keychain(
+                "too many call scopes in key authorization",
+            )));
+        }
+
```
legion2002, line 201: addressing in #3422

crates/transaction-pool/src/tempo_pool.rs
legion2002: addressing in #3422

crates/primitives/src/transaction/key_authorization.rs
legion2002: Stacked PR here - https://github.com/tempoxyz/tempo/pull/3421/

crates/revm/src/handler.rs
legion2002: Amped summary 

The problem is a signature-type consistency bug.

`key_id` by itself is not enough to identify the signature scheme: for P256 and WebAuthn, the same public-key coordinates derive the same address, so the same `key_id` can be paired with different actual signature types. One of the TIP-1011 invariants is that the authorized/stored key type must match the signature type actually used by the keychain transaction.

We were already enforcing that invariant in the normal keychain-validation path, but there were two gaps:
1. Same-tx `KeyAuthorization + use-the-same-key` bypassed the normal keychain validation, so it could skip the key-type parity check entirely.
2. Txpool validation for an already-authorized key loaded the stored key but did not compare `authorized_key.signature_type` against the actual keychain signature type, so a mismatched tx could enter the pool and deterministically fail later during execution.

The solution was to enforce the same invariant explicitly in both places:
1. In the handler, for same-tx auth+use, reject if `key_auth.key_type != keychain_sig.signature.signature_type()`.
2. In txpool validation, reject if the stored key's `signature_type` does not match `sig.signature.signature_type()` for the existing-key path.
3. Gate both checks at T3 so we preserve historical pre-T3 reexecution behavior and only enforce the stricter TIP-1011 invariant starting at the TIP-1011 hardfork boundary.

Why this belongs in this PR:
This PR is where TIP-1011 access-key semantics are being made concrete and enforced end-to-end. Without these checks, the implementation would violate a TIP invariant after T3: a transaction could present a key authorization / stored key type that does not match the actual signature type used on the wire. So this change is not adding a new rule arbitrarily; it is closing a gap so the implementation actually satisfies the TIP invariant consistently across execution and pool admission.

crates/transaction-pool/src/validator.rs
```diff
@@ -332,6 +577,31 @@ where
             }
         }
 
+        if spec.is_t3() {
```
legion2002, line 572: addressing in #3422

crates/primitives/src/transaction/key_authorization.rs
legion2002: merged.

crates/precompiles/src/account_keychain/mod.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/pull/3384/commits/405602b391a7c129548c11b140be5d5f89e5c696

crates/revm/src/handler.rs
```diff
@@ -1130,7 +1282,25 @@ where
                     // The TIP20 precompile will read this during execution to enforce spending limits
                     keychain
                         .set_transaction_key(access_key_addr)
-                        .map_err(|e| EVMError::Custom(e.to_string()))
+                        .map_err(|e| EVMError::Custom(e.to_string()))?;
```
legion2002, line 1251: addressed in - https://github.com/tempoxyz/tempo/pull/3384/commits/5162f3cfedaaa6e00eed1cecbef3cde933ace54d

crates/revm/src/handler.rs
legion2002: It was incorrectly using is_t1  though, I have updated to is_t3

crates/precompiles/src/account_keychain/mod.rs
legion2002: This seeems like it might require a bigger refactor, we can do it in a follow up PR

crates/revm/src/handler.rs
legion2002: addressed here - https://github.com/tempoxyz/tempo/pull/3384/commits/79c51afde05b8d7c6e12e36d0b5af17210c0840e

crates/precompiles/src/account_keychain/dispatch.rs
legion2002: tried doing this, doesn't really save many LOC, because you still have to abstract out the hardfork version.

klkvr (approved):

PR #3388, opened 2026-03-31
docs(tip-1041): add batch tip-403 policy updates

malleshpai:
Adds TIP-1041 for batch updates to TIP-403 whitelist and blacklist policies.

The TIP keeps the scope narrow by adding only two batch registry methods, reusing existing per-account semantics and events, and noting the limited gas savings relative to Tempo transaction batching.

### DIFF
--- a/tips/tip-1041.md
+++ b/tips/tip-1041.md
@@ -1,0 +1,95 @@
+---
+id: TIP-1041
+title: Batch TIP-403 Policy Updates
+description: Adds batch whitelist and blacklist update functions to the TIP-403 policy registry.
+authors: Mallesh Pai
+status: Draft
+related: TIP-403
+protocolVersion: TBD
+---
+
+# TIP-1041: Batch TIP-403 Policy Updates
+
+## Abstract
+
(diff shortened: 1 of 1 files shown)


decofe (commented):
## TIP-1041 Review

### protocolVersion → T5

### Correctness

This TIP is straightforward and well-specified. The equivalence to sequential single-account calls is the right model. A few items:

1. **Gas limit / array size.** No maximum array size is specified. Very large batches could exceed block gas limits. Consider noting that callers are responsible for batch sizing, or add a maximum (e.g., 1000) to prevent accidental OOG on large compliance updates.

2. **Zero address in batch.** If `accounts` contains `address(0)`, does the underlying `modifyPolicyWhitelist` / `modifyPolicyBlacklist` accept it? If not, the batch reverts atomically — which is probably correct but should be explicitly tested.

3. **Atomicity is well-stated** ("if the call reverts, none of the account updates take effect"). Good.

### Suggested changes
```diff
-protocolVersion: TBD
+protocolVersion: T5
```

PR #3410, opened 2026-04-01
refactor(account-keychain): use enumerable maps for scopes

legion2002:
## What This PR Does

This PR now contains only the enumerable-map refactor that remained after the TIP-1011 stack merged.

- adds `EnumerableMap<K, V>` to the precompile storage types, backed by a `Set<K>` for authoritative key membership and a `Mapping<K, V>` for values
- exports `EnumerableMap` from `crates/precompiles/src/storage/types/mod.rs`
- changes `AccountKeychain` scope storage so `KeyScope.targets` becomes `EnumerableMap<Address, TargetScope>` and `TargetScope.selectors` becomes `EnumerableMap<FixedBytes<4>, SelectorScope>`
- updates the `AccountKeychain` read and mutation paths to use the enumerable-map API instead of parallel `Set + Mapping` structures when enumerating, inserting, removing, and clearing target and selector scopes
- preserves the existing TIP-1011 behavior: `is_scoped` still distinguishes unrestricted vs scoped deny-all, and empty child sets still mean "no further restriction"

This does not change the external TIP-1011 interface or validation model. It only changes how scoped target/selector state is stored and maintained internally.

## Verification

- `cargo fmt --all --check`
- `cargo test -p tempo-precompiles --lib`

### DIFF
--- a/crates/precompiles/src/storage/types/enumerable_map.rs
+++ b/crates/precompiles/src/storage/types/enumerable_map.rs
@@ -145,8 +145,11 @@ where
         V: Storable,
         V::Handler: Handler<V>,
     {
-        self.values.at_mut(key).delete()?;
-        self.keys.remove(key)
+        if self.keys.remove(key)? {
+            self.values.at_mut(key).delete()?;
+            Ok(true)
+        }
+        Ok(false)
     }
 
     /// Removes all key-value pairs from the map.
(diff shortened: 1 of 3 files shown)


crates/precompiles/src/storage/types/enumerable_map.rs
```diff
@@ -0,0 +1,408 @@
...
+    /// Returns `true` if the key was removed from the map.
+    #[inline]
+    pub fn remove(&mut self, key: &K) -> Result<bool>
+    where
+        K::Handler: Handler<K>,
+        V: Storable,
+        V::Handler: Handler<V>,
+    {
+        self.values.at_mut(key).delete()?;
+        self.keys.remove(key)
```
0xrusowsky: since `remove(key)` returns `bool`, we could do:

```suggestion
        if self.keys.remove(key)? {
            self.values.at_mut(key).delete()?;
            Ok(true)
        }
        Ok(false)
```

PR #3412, opened 2026-04-01
feat(sidecar): make simple arb event-aware

ming1523:
Updates `tempo-sidecar simple-arb` to avoid sweeping every pool on every loop.

This change keeps the existing fallback full rescan, but shifts the steady-state path to an event-aware dirty-pair workflow:

- track pool-affecting events
- enqueue dirty pairs
- rebalance only changed pools between full rescans
- expose basic dirty-pair/full-sweep metrics

The result is a cheaper and more production-like rebalancer without changing protocol behavior.


ming1523:
Reviving this lane. The PR is still awaiting the requested review from @Zygimantass, and there are no visible failing checks on the current head. Keeping it active instead of letting the stale marker speak for the current state.

PR #3413, opened 2026-04-01
feat(sidecar): add payment lane benchmark command

ming1523:
Adds a new `payment-lane-bench` subcommand to `tempo-sidecar` for quickly exercising Tempo's payment-lane behavior under synthetic load.

The command supports:

- `payments-only`
- `non-payment-only`
- `mixed-load`

It submits load, watches mined blocks, and reports summary metrics including submitted/observed tx counts and p50/p95 landing latencies for payment vs non-payment transactions.


ming1523:
Reviving this lane. The PR is still awaiting the requested review from @Zygimantass, and there are no visible failing checks on the current head. Keeping it active instead of letting the stale marker speak for the current state.

PR #3425, opened 2026-04-01
docs: add TIP-1042 FeeAMM TIP-403 policy exemptions

decofe:
Reopen of #3227 with new TIP number.

Adds TIP-1042, which exempts the FeeManager address from TIP-403 authorization checks during the protocol fee-collection path while preserving TIP-403 enforcement on public AMM operations (`mint`, `burn`, `rebalanceSwap`, `distributeFees`).

This version also incorporates the #3976 review fixes for SIGP-260 / TMPO6-02 by pinning LP authorization to the LP lifecycle:

- `mint(..., to)` requires the caller to be sender-authorized on `userToken`, FeeManager to be recipient-authorized on `userToken`, and `to` to be recipient-authorized on both underlying tokens.
- `burn(...)` requires `msg.sender` to be sender-authorized on both underlying tokens, in addition to the existing underlying transfer checks for FeeManager and `to`.

Also updates the behavioral summary, invariants, test cases, TOCTOU wording, and rationale around withdraw-only mode using TIP-1015 compound policies.

Prompted by: @danrobinson


decofe (commented):
## TIP-1042 Review

### protocolVersion → T5

1. **Exemption scope must be call-site-scoped, not address-wide.** The spec correctly describes exempting FeeManager in the fee-collection path only and preserving checks on public AMM ops. But the implementation must be careful: if the exemption is implemented as "skip TIP-403 when `msg.sender == FeeManager` or `to == FeeManager`," it could accidentally exempt public-facing functions. The exemption should be gated on the specific internal code path (e.g., a flag or call context), not just address equality.

2. **`mint` authorization check — single-sided mint clarification.** The `mint` function accepts `userToken` and `validatorToken` but only transfers `validatorToken`. The new check treats mint as if the caller were transferring `userToken` to the pool. This is a good design, but:
   - If there are multiple `mint` overloads or paths (single-sided, dual-sided), specify which ones get the new check.
   - The check uses `isAuthorizedSender(userTokenPolicyId, msg.sender)` — this checks whether the *caller* can send `userToken`. But the caller isn't actually sending `userToken`. Make sure the semantic intent ("caller is opting into userToken ecosystem") is clear.

3. **Stranded fees edge case is well-documented** — validators setting their fee token to a non-whitelisted token will have uncollectable fees. Wallet/validator tooling should surface this. Good that it's called out.

4. **Interaction with TIP-1033 (two-hop routing).** If two-hop routing goes through an intermediate token, does the fee-collection exemption apply to both hops? The spec should confirm that the FeeManager TIP-403 exemption covers the intermediate transfer as well, since both hops occur within the fee-collection path.

### Suggested changes

Add to frontmatter:
```yaml
protocolVersion: T5
```

tips/tip-1042.md
```diff
@@ -0,0 +1,192 @@
...
+
+No change. Calls `systemTransferFrom(msg.sender, address(this), amountIn)` and `transfer(to, amountOut)`, which enforce TIP-403 on all participants.
+
+### `mint` — Enhanced Authorization
+
+In addition to the existing TIP-403 enforcement on the `validatorToken` transfer, `mint` adds an authorization check on the `userToken` for the caller.
+
+**Rationale**: The caller is providing liquidity to a pool that facilitates swaps for `userToken`. By minting, they become an economic participant in that token's ecosystem — pool reserves shift during fee swaps and rebalancing, and on `burn` the LP receives pro-rata shares of both tokens. The caller should be authorized to interact with the `userToken` even though no `userToken` moves during `mint`.
+
+**Change**: Before processing the mint, check that the caller is authorized as a sender on the `userToken` and that the FeeManager is an authorized recipient — as if the caller were transferring `userToken` into the pool:
```
legion2002: Nice, we can use `ensure_transfer_authorized` here directly

tips/tip-1042.md
```diff
@@ -0,0 +1,192 @@
...
+The changes are divided into two categories:
+
+1. **Fee collection path**: Remove TIP-403 checks on the FeeManager address during protocol-initiated fee operations.
+2. **Public AMM operations**: Preserve existing TIP-403 enforcement, plus add a new authorization gate on `mint`.
+
+## Fee Collection Path — Exempt FeeManager
+
+The following operations skip TIP-403 authorization checks for the FeeManager address. The counterparty (user, validator) is still checked.
+
+### `collectFeePreTx`
```
legion2002, line 44: Can we add a similar section explicitly specifying changes for `collectFeePostTx`
i am assuming there will be no change here, because we don't do any checks atm.

legion2002 (dismissed):
makes sense

tips/tip-1042.md
```diff
@@ -0,0 +1,192 @@
...
+        revert ITIP20.PolicyForbids();
+    }
+
+    // ... rest of mint logic (unchanged)
+}
+```
+
+Note: There is an inherent TOCTOU (time-of-check-time-of-use) gap — a policy may change between when the LP mints and when the pool is used for fee swaps or rebalancing. This is a best-effort check and is consistent with how TIP-403 operates elsewhere (e.g., a token holder may be blacklisted after receiving tokens).
+
+Additionally, since `rebalanceSwap` checks the relevant policies, a token admin can prevent the liquidity on the FeeAMM from being refreshed, which means it will eventually be exhausted. This should give token admins an adequate tool for disabling the FeeAMM for their token even if someone manages to create liquidity on the pool before the admin sets that policy.
```
0xKitsune: The spec enforces TIP-403 on FeeManager as sender for `burn`, `distributeFees`, and `rebalanceSwap`. If an issuer revokes `FeeManager` from the whitelist after LPs have minted or fees have been collected, those outflows all revert. 

The spec acknowledges this but another option could be exempting the `FeeManager` from TIP-403 on the outflow paths (burn, distributeFees, rebalanceSwap) in addition to fee collection. The mint gate could stay as an issuer's opt-in for AMM participation (or we could consider adding a mapping for fee token enabled/disabled).  

We should consider an approach where once funds are in the pool, they arent able to be frozen due to TIP-403 checks against the `FeeManager`. Existing checks against the user should remain unchanged as the spec suggests.

tips/tip-1042.md
```diff
@@ -0,0 +1,192 @@
...
+The FeeManager address exemption does **not** remove issuer control. Instead, it shifts the control mechanism from "whitelist the FeeManager address" to "whitelist determines AMM participation":
+
+| Issuer action | Fee payment | AMM pools | `distributeFees` |
+|---------------|-------------|-----------|-------------------|
+| Whitelist FeeManager | ✅ | ✅ Pools can be created, rebalanced | ✅ Fees distributed |
+| Don't whitelist FeeManager | ✅ Same-token fees work. Cross-token fees require pool liquidity (see below). | ❌ `mint` blocked (FeeManager not authorized as recipient on either token). No pools → no cross-token fee swaps. | ❌ `transfer` from FeeManager blocked |
+
+### Edge Case: Stranded Fees
+
+If a token does not whitelist the FeeManager address, same-token fees (where `userToken == validatorToken`) will be collected successfully — the fee collection path is exempt. However, `distributeFees` calls `transfer()` which checks the FeeManager as sender. If the FeeManager is not whitelisted, the validator cannot claim these fees and they remain stranded in the FeeManager.
```
0xKitsune, line 194: Its also possible to hit the case where LP funds are frozen if the token issuer initially whitelists (or doesnt blacklist) the fee token, users LP into the pool and then the token issuer removes the `FeeManager` from the whitelist. Currently there is no recovery mechanism for this case, once `collectedFees[v][x]` accrues for a non-whitelisted token, the value is stuck.

As [mentioned above](https://github.com/tempoxyz/tempo/pull/3425#discussion_r3118723593), this would be fixed by exempting `burn` from `TIP-403` checks.

tips/tip-1042.md
```diff
@@ -0,0 +1,192 @@
...
+
+### `rebalanceSwap`
+
+No change. Calls `systemTransferFrom(msg.sender, address(this), amountIn)` and `transfer(to, amountOut)`, which enforce TIP-403 on all participants.
+
+### `mint` — Enhanced Authorization
+
+In addition to the existing TIP-403 enforcement on the `validatorToken` transfer, `mint` adds an authorization check on the `userToken` for the caller.
+
+**Rationale**: The caller is providing liquidity to a pool that facilitates swaps for `userToken`. By minting, they become an economic participant in that token's ecosystem — pool reserves shift during fee swaps and rebalancing, and on `burn` the LP receives pro-rata shares of both tokens. The caller should be authorized to interact with the `userToken` even though no `userToken` moves during `mint`.
```
0xKitsune: The new sender check on `msg.sender` is bypassable.

Mint only checks the `msg.sender`, rather than the `to` address. Burn only checks `to`, and not the LP. Either gate can be passed by routing through a whitelisted helper, neither side of the lifecycle actually pins authorization to the LP itself.

danrobinson: I think makes sense to check `to` as a recipient on `mint` and as a sender on `burn` (on the user token and the validator token), roughly consistent with what we do on the DEX.

tips/tip-1042.md
danrobinson: Yeah, I think could make sense to pave the path where the fee token can turn off the fee manager and put it into "withdraw only" mode.

In that case, we should skip the check on whether FeeManager is an authorized sender on `burn` and `distributeFees` but not the checks on `rebalanceSwap`.

danrobinson: This would mean that in the non-pool case, there's no way to prevent fees from being paid in the token to validators who set it as their fee token (other than by blacklisting validators). But I think this is OK? The FeeManager is acting as a passthrough. The token can blacklist the validator

danrobinson: Eh, on thinking about it more, I think it's OK if blacklisting it as the sender causes funds to be stuck inside; that's ordinary behavior for a contract. If a token wants to put it in withdraw-only mode, it can set a compound policy

tips/tip-1042.md
```diff
@@ -0,0 +1,192 @@
...
+
+No change. This function already does not perform TIP-403 checks.
+
+## Public AMM Operations — Full TIP-403 Enforcement
+
+The following operations continue to enforce TIP-403 on all participants, including the FeeManager address, via the underlying `transfer()` and `systemTransferFrom()` calls on TIP-20 tokens.
+
+### `distributeFees`
+
+No change. Calls `IERC20(token).transfer(validator, amount)`, which enforces TIP-403 on both sender (FeeManager) and recipient (validator).
```
fgimenez: nit
```suggestion
No change. Calls `ITIP20(token).transfer(validator, amount)`, which enforces TIP-403 on both sender (FeeManager) and recipient (validator).
```

tips/tip-1042.md
```diff
@@ -0,0 +1,192 @@
...
+
+### `burn`
+
+No change. Calls `IERC20(userToken).transfer(to, amountUserToken)` and `IERC20(validatorToken).transfer(to, amountValidatorToken)`, which enforce TIP-403 on both sender (FeeManager) and recipient (`to`).
```
fgimenez:
```suggestion
No change. Calls `ITIP20(userToken).transfer(to, amountUserToken)` and `ITIP20(validatorToken).transfer(to, amountValidatorToken)`, which enforce TIP-403 on both sender (FeeManager) and recipient (`to`).
```

tips/tip-1042.md
```diff
@@ -0,0 +1,192 @@
...
+    // Caller must be authorized as sender, FeeManager as recipient.
+    uint64 userTokenPolicyId = ITIP20(userToken).transferPolicyId();
+    if (
+        !TIP403_REGISTRY.isAuthorizedSender(userTokenPolicyId, msg.sender)
+            || !TIP403_REGISTRY.isAuthorizedRecipient(userTokenPolicyId, address(this))
+    ) {
+        revert ITIP20.PolicyForbids();
+    }
+
+    // ... rest of mint logic (unchanged)
```
fgimenez: should we add here an explicit `systemTransferFrom(msg.sender, address(this), amountValidatorToken)` line? this is the call that enforces the validatorToken policy as per the Behavioral Summary "FeeManager checked as recipient (both tokens) / Caller checked as sender (both tokens)"

fgimenez: it is technically part of "rest of mint logic (unchanged)" but maybe worth mentioned as we do for `rebalanceSwap` https://github.com/tempoxyz/tempo/pull/3425/changes#diff-96299f1d34b0d26567f7873b97e841bc32f07a9fb1e3761f6c522f932b025803R91

tips/tip-1042.md
```diff
@@ -0,0 +1,192 @@
...
+    if (amountValidatorToken == 0) revert InvalidAmount();
+    _requireUSDTIP20(userToken);
+    _requireUSDTIP20(validatorToken);
+
+    // New: treat mint as if caller were transferring userToken to the pool.
+    // Caller must be authorized as sender, FeeManager as recipient.
+    uint64 userTokenPolicyId = ITIP20(userToken).transferPolicyId();
+    if (
+        !TIP403_REGISTRY.isAuthorizedSender(userTokenPolicyId, msg.sender)
+            || !TIP403_REGISTRY.isAuthorizedRecipient(userTokenPolicyId, address(this))
```
fgimenez, line 154: `to` is not checked as an authorized LP-share recipient, should we add here `|| !TIP403_REGISTRY.isAuthorizedRecipient(userTokenPolicyId, to) || !TIP403_REGISTRY.isAuthorizedRecipient(ITIP20(validatorToken).transferPolicyId, to)`

tips/tip-1042.md
```diff
@@ -0,0 +1,192 @@
...
+    ) {
+        revert ITIP20.PolicyForbids();
+    }
+
+    // ... rest of mint logic (unchanged)
+}
+```
+
+Note: There is an inherent TOCTOU (time-of-check-time-of-use) gap — a policy may change between when the LP mints and when the pool is used for fee swaps or rebalancing. This is a best-effort check and is consistent with how TIP-403 operates elsewhere (e.g., a token holder may be blacklisted after receiving tokens).
```
fgimenez: per invariant 1, fee swaps don't check the FeeManager
```suggestion
Note: There is an inherent TOCTOU (time-of-check-time-of-use) gap — a policy may change between when the LP mints and when the pool is used for subsequent `rebalanceSwap`, `burn`, or `distributeFees`. This is a best-effort check and is consistent with how TIP-403 operates elsewhere (e.g., a token holder may be blacklisted after receiving tokens).
```

tips/tip-1042.md
fgimenez: following up from https://github.com/tempoxyz/tempo/pull/3425/changes#r3150523058

fgimenez (commented):
minor issues, in general lgtm, most relevant https://github.com/tempoxyz/tempo/pull/3425/changes#r3150523058 imo

decofe:
Closing in favor of the fresh review PR: #6604

PR #3438, opened 2026-04-02
docs(tip-1043): add Tempo block access lists draft

mediocregopher:
Adds TIP-1043: Tempo Block Access Lists. Tempo blocks carry a residual access list (`RAL`) whose header hash is computed over the RLP-encoded `RAL`, while `IAL` is local-only validator I/O preparation and is not used to reconstruct a full upstream BAL.

The initial scope defines TIP-20 `transfer` and `transferWithMemo` handling: builders produce a valid `RAL`, validators compare against the expected residual, and transfer-path `storage_reads` are excluded from the block-carried payload.

### DIFF
--- a/tips/tip-1043.md
+++ b/tips/tip-1043.md
@@ -1,0 +1,162 @@
+---
+id: TIP-1043
+title: Tempo Block Access Lists
+description: Tempo-specific decomposition of block access lists into implicit and residual components.
+authors: Brian (@mediocregopher)
+status: Draft
+related: EIP-7928
+---
+
+# TIP-1043: Tempo Block Access Lists
+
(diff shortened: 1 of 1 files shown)


mediocregopher:
Given recent improvements to TIP-20 transfer precompile implementation there are no longer any read-only slots on the happy path, and therefore excluding storage_reads for these transactions doesn't really net us anything significant. See [this bench](https://github.com/tempoxyz/tempo/pull/4717#issuecomment-4594043313), the changes to block size are barely significant.

Given this we're going to close this TIP for now. If BAL changes end up being necessary for Tempo they will be of a different sort and we can open a new TIP for them.

PR #3443, opened 2026-04-02
fix: use consistent rlp encoding

klkvr:
Changes `KeyAuthorization` and `TokenLimit` to use consistent rlp encoding which ensures that each value can only be represented with a single RLP payload

This would be breaking for tooking that was encoding `expiry: None` as 0x80 instead of omitting it

Fixes SIGP 204

### DIFF
--- a/crates/primitives/src/transaction/key_authorization.rs
+++ b/crates/primitives/src/transaction/key_authorization.rs
@@ -118,7 +118,7 @@ pub struct KeyAuthorization {
     /// Unix timestamp when key expires.
     /// - `None` (RLP 0x80) = key never expires (stored as u64::MAX in precompile)
     /// - `Some(timestamp)` = key expires at this timestamp
-    /// 
+    ///
     /// Note: Some(0) will get decoded as None after RLP roundtrip.
     #[cfg_attr(feature = "serde", serde(with = "alloy_serde::quantity::opt"))]
     pub expiry: Option<u64>,
@@ -293,7 +293,11 @@ mod rlp {
 
     impl From<&TokenLimit> for TokenLimitWire {
         fn from(value: &TokenLimit) -> Self {
-            let TokenLimit { token, limit, period } = value;
+            let TokenLimit {
+                token,
+                limit,
+                period,
+            } = value;
             Self {
                 token: *token,
                 limit: *limit,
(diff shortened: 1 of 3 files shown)


crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -119,6 +118,8 @@ pub struct KeyAuthorization {
     /// Unix timestamp when key expires.
     /// - `None` (RLP 0x80) = key never expires (stored as u64::MAX in precompile)
     /// - `Some(timestamp)` = key expires at this timestamp
+    ///
+    /// Note: Some(0) will get decoded as None after RLP roundtrip.
```
legion2002, line 122: This could be problematic, None expiry means a key authorization with infinite expiry.
It could be a footgun, although it is nonsensical for someone to send Some(0).

Also maybe this fails reexec, because before this Some(0) would revert a key auth, because `expiry < current timestamp`

legion2002: ig it doesn't fail reexec, because we already reencode txs inside the node, so those txs would still treat this as None

legion2002 (approved):

crates/primitives/src/transaction/key_authorization.rs
klkvr: it was never possible to decode a `Some(0)` before so it would be ok

mattsse (approved):

crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -325,6 +331,72 @@ mod rlp {
...
+                key_id,
+                expiry,
+                limits,
+                allowed_calls,
+            } = value;
+            Self {
+                chain_id: *chain_id,
+                key_type: *key_type,
+                key_id: *key_id,
+                expiry: expiry.and_then(NonZeroU64::new),
```
0xKitsune, line 361: `Some(0)` silently collapses to `None` here via `expiry.and_then(NonZeroU64::new)`, which looks like will result in RLP roundtrips to produce a different value:

```rust
#[test]
fn test_expiry_some_zero_roundtrip() {
    let auth = make_auth(Some(0), None);

    let mut encoded = Vec::new();
    auth.encode(&mut encoded);

    let decoded =
        <KeyAuthorization as Decodable>::decode(&mut encoded.as_slice()).expect("decode");
    assert_eq!(auth, decoded); // fails: expiry is now None
}
```

When decoding, the missing field is read back as `None`. Downstream in `handler.rs:1081`, `None` is interpreted as "never expires" via `unwrap_or(u64::MAX)`. So a key that was meant to expire immediately becomes one that never expires after a roundtrip.

0xKitsune (approved):

PR #3468, opened 2026-04-07
docs(tip-1044): waive storage init cost for DEX flip order placement

decofe:
Adds TIP-1044: waive the SSTORE initialization gas cost when a flip order creates a new order during fill, since the filled order's storage slots are cleared in the same call — making the operation net-zero on state growth.

The discount is applied as a precompile-internal gas refund, bypassing the EIP-3529 cap, scoped only to `StablecoinDEX.fill_order` when processing successful flips.

Related: TIP-1002

Prompted by: Dan


tips/tip-1044.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Motivation
+
+Flip orders are designed as perpetual two-sided liquidity: when one side fills, a new order is automatically placed on the opposite side. Because the old order is deleted and the new order is created in the same `fill_order` call, the net state growth is zero — the same number of storage slots are cleared as are initialized.
+
+However, the EVM's SSTORE gas schedule prices each slot independently. The new order's slots transition from zero → non-zero (costing 20,000 gas each), while the old order's slots transition from non-zero → zero (earning a refund capped at 20% of total transaction gas by EIP-3529). The refund cap means swappers never recover the full initialization cost, even though no net storage is consumed.
+
+This makes filling flip orders substantially more expensive than filling regular orders of the same size, discouraging their use and increasing costs for swappers who interact with them. Since the StablecoinDEX is a precompile with full control over its own gas accounting, it can apply a targeted discount without modifying the general EVM gas schedule.
```
danrobinson: It doesn't really discourage their use, can cut those words

decofe (commented):
## TIP-1044 Review

### protocolVersion → T5

1. **Exact discounted write set must be enumerated.** The TIP says "the `Order` struct's storage footprint" but doesn't list the exact slots. Different `Storable` derive expansions could produce different slot counts. The spec should either:
   - Enumerate the exact number of slots (e.g., "3 slots per order")
   - Or tie the discount to `Order::SLOT_COUNT` and require re-evaluation if the struct changes

2. **Tick-level storage exclusion is correct** — the TIP rightly excludes tick metadata from the discount since ticks may differ. Good.

3. **Failed flip gas accounting.** The spec says "If `place_flip()` returns a business logic error, no discount is applied since no new order was created." But what about the delete of the old order? If the old order was already deleted before `place_flip()` fails, does the clearing refund apply normally? The spec should clarify the ordering:
   - Is it: fill → delete old → place new (discount)? Or: fill → place new (discount) → delete old?
   - If delete happens before place and place fails, the delete should get its normal refund.

4. **Quote/simulation must not use the discount.** If TIP-1024's shared fill engine runs in simulation mode, it must not apply this gas discount. Otherwise quote gas estimates would be wrong.

5. **Same-tick flip (TIP-1030) interaction.** Same-tick flips make churn cheaper. Combined with gas-neutral storage, a flip order at the same tick can be filled and re-placed repeatedly with minimal gas. This is probably fine (it's real economic activity), but worth noting that the gas floor for flip-order churn is now very low.

### Suggested changes

Add to frontmatter:
```yaml
protocolVersion: T5
```

Add to Specification:
```markdown
## Slot Count

The current `Order` struct uses N storage slots (as determined by the `Storable` derive).
The discount exempts exactly N slots from state creation cost and N slots from clearing
refund. If the `Order` struct layout changes in a future TIP, the discount must be
re-evaluated.
```

tips/tip-1044.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Motivation
+
+Flip orders are designed as perpetual two-sided liquidity: when one side fills, a new order is automatically placed on the opposite side. Because the old order is deleted and the new order is created in the same `fill_order` call, the net state growth is zero — the same number of storage slots are cleared as are initialized.
+
+However, the SSTORE gas schedule prices each slot independently. Per TIP-1000, Tempo charges 250,000 gas for writing a new state element (zero → non-zero), while clearing a slot (non-zero → zero) earns only a 15,000 gas refund capped at 20% of total transaction gas by EIP-3529. The refund is negligible compared to the initialization cost, so swappers pay the full state creation price even though no net storage is consumed.
```
0xKitsune, line 21: The 15,000-gas clearing refund is correct. REFUND_SSTORE_CLEARS in upstream revm is 15,000, and tempo_gas_params() does not override this constant.

tips/tip-1044.md
```diff
@@ -0,0 +1,85 @@
...
+
+This makes filling flip orders substantially more expensive than filling regular orders of the same size, increasing costs for swappers who interact with them. Since the StablecoinDEX is a precompile with full control over its own gas accounting, it can apply a targeted discount without modifying the general EVM gas schedule.
+
+## Assumptions
+
+- The `Order` struct occupies 6 storage slots (determined by the `Storable` derive macro). This does not change between the filled order and the newly placed flip order.
+- The new order is always placed (`place_flip()`) and the old order is always deleted (`orders[old_id].delete()`) within the same `fill_order` call. This is guaranteed by the `fill_order` control flow.
```
0xKitsune, line 28: Minor but seems with putting some language in to mention that the refund is only applied when the new order succeeds.

tips/tip-1044.md
```diff
@@ -0,0 +1,85 @@
...
+2. **Charge `WARM_SSTORE_RESET` on the old order's 6 clears**: When `orders[old_id].delete()` clears the old order's 6 slots (non-zero → zero), each slot is charged at the standard `WARM_SSTORE_RESET` (2,900 gas) — the normal cost for modifying an existing slot — but the 15,000-gas clearing refund is **not granted**. Total: 6 × 2,900 = 17,400 gas.
+3. **Net effect**: The flip's 12 storage operations cost 137,400 gas total, compared to 1,500,000 gas under standard accounting. The state growth penalty is discounted since the net state change is zero.
+
+The discount applies to the `fill_order` call as a whole — the create (`place_flip`) and delete (`orders[old_id].delete()`) are treated as a single logical operation regardless of their execution order within the call.
+
+This is not a post-hoc refund — the state creation cost is never charged and the clearing refund is never granted. This means the gas is never counted against the transaction's gas limit, avoiding situations where a transaction would fail due to insufficient gas even though the net state change is zero.
```
0xKitsune, line 50: @0xrusowsky @klkvr we should think through how to implement this gracefully.

tips/tip-1044.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Hardfork gating
+
+This change is gated behind a hardfork flag. The discount is only applied when the active spec satisfies the gating check.
+
+---
+
+# Invariants
+
+- **Net state neutrality**: A flip order fill always deletes exactly 6 `Order` storage slots and creates 6 new ones. The number of slots discounted from state creation cost must equal the number of slots where the clearing refund is withheld.
```
0xKitsune, line 81: This seems correct, we will need to account for this case. @danrobinson can you update the spec to mention this case?

0xKitsune (commented):
The spec says state creation cost is "never charged, even transiently" but also says the discount doesn't apply if `place_flip()` fails. These contradict each other,  at the time of the writes inside `place_flip()`, we don't know yet if it'll succeed or fail. If we never charge the cost up front, there's no discount to reverse on failure. 

We will need to think through how to meter and refund this @0xrusowsky. It's different from existing refunds that get applied post-tx, and I'm not sure we have the plumbing for this right now.

tips/tip-1044.md
0xrusowsky: also worth flagging that if it fails it is not an issue cause only the delete side happens and there is no state creation

tips/tip-1044.md
0xrusowsky: imo we need to integrate this into the `PrecompileStorageProvider` so that we can conditionally charge `STORAGE_MOVE` rather than the 250k.

the impl/design will depend on the security guarantees that we want to offload to the dev writing the precompile

0xrusowsky:
> The spec says state creation cost is "never charged, even transiently" but also says the discount doesn't apply if `place_flip()` fails. These contradict each other, at the time of the writes inside `place_flip()`, we don't know yet if it'll succeed or fail. If we never charge the cost up front, there's no discount to reverse on failure.
> 
> We will need to think through how to meter and refund this @0xrusowsky. It's different from existing refunds that get applied post-tx, and I'm not sure we have the plumbing for this right now.

i think the "discount doesn't apply if `place_flip()` fails" is technically true, but doesn't make sense as worded. Personally i'd rewrite for clarity, since the writes only happen in `place_flip()`.

tips/tip-1044.md
klkvr: yeah this definitely sounds feasible - either by adding API for a free `SSTORE` and handling the gas accounting in the DEX precompile, or by encapsulating the "move" usecase in the storage provider

tips/tip-1044.md
```diff
@@ -0,0 +1,85 @@
+---
+id: TIP-1044
+title: Discount 1,500,000-gas storage initialization cost for DEX flip order placement
+description: Discounts the 1,500,000-gas state creation cost (6 slots × 250,000 gas) when a flip order creates a new order, since the filled order's 6 storage slots are cleared in the same call.
+authors: Dan Robinson (@danrobinson)
+status: Draft
```
jenpaff:
```suggestion
status: Rejected
```

tips/tip-1044.md
```diff
@@ -0,0 +1,85 @@
...
+related: TIP-1000, TIP-1002, TIP-1016
+protocolVersion: T5
```
jenpaff:
```suggestion
protocolVersion: N/A
```

tips/tip-1044.md
```diff
@@ -0,0 +1,83 @@
+---
+id: TIP-1044
+title: Discount 1,500,000-gas storage initialization cost for DEX flip order placement
+description: Discounts the 1,500,000-gas state creation cost (6 slots × 250,000 gas) when a flip order creates a new order, since the filled order's 6 storage slots are cleared in the same call.
+authors: Dan Robinson (@danrobinson)
+protocolVersion: N/A
```
jenpaff, line 8:
```suggestion
status: Rejected
authors: Dan Robinson (@danrobinson)
protocolVersion: N/A
```

jenpaff:
superseded by https://github.com/tempoxyz/tempo/pull/3755

PR #3483, opened 2026-04-07
fix(account-keychain): clamp T3 refund spending limits

legion2002:
Refs ZELLIC-189

Clamp `refund_spending_limit` on T3 so defensive refund paths cannot restore more spending room than the key's configured max.

Adds a regression test covering an over-refund on T3 while leaving legacy pre-T3 behavior unchanged, since the legacy layout does not persist a max bound.

### DIFF
--- a/crates/precompiles/src/account_keychain/mod.rs
+++ b/crates/precompiles/src/account_keychain/mod.rs
@@ -1155,6 +1155,8 @@ impl AccountKeychain {
     ///
     /// Restores the spending limit by the refunded amount.
     /// Should be called after a fee refund to avoid permanently reducing the spending limit.
+    /// On T3, this should never restore more than the configured max in the current fee flow,
+    /// but we still clamp as defense in depth in case a future caller violates that invariant.
     pub fn refund_spending_limit(
         &mut self,
         account: Address,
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -3117,6 +3127,102 @@ mod tests {
...
+            keychain.set_tx_origin(eoa)?;
+
+            keychain.authorize_transfer(eoa, token, U256::from(10))?;
+            keychain.refund_spending_limit(eoa, token, U256::from(50))?;
+
+            let after_refund = keychain.get_remaining_limit(getRemainingLimitCall {
+                account: eoa,
+                keyId: access_key,
+                token,
+            })?;
```
0xKitsune, line 3185: The existing clamp test only asserts `remaining == max`, which would also pass if the code unconditionally returned max. Consider adding a test case that spends 60 from a limit of 100, refunds 30, and asserts `remaining == 70` rather than clamping to max.

0xKitsune (approved):

crates/precompiles/src/account_keychain/mod.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/pull/3483/commits/f4ab7f9c52191552196416d364bb0a5ee4eaa13a

PR #3490, opened 2026-04-08
docs(tip-1024): note gas cost changes in API compatibility section

decofe:
Adds a bullet to the API compatibility section noting that gas costs will change for quote() and swap(), per review feedback from @malleshpai.

Prompted by: Dan


decofe:
Superseded by the full spec rewrite in #3509, which includes this gas cost note and all other review comment resolutions.

PR #3496, opened 2026-04-08
docs(tip-1045): payment lane classification

0xrusowsky:
Adds a new TIP-1045 specifying consensus rules for classifying transactions as payment lane eligible, including the calldata selector allow-list enforced at the T5 hardfork.

Besides the allow-list, not only accepts `TIP20` calls, but also calls to the new `TIP20ChannelEscrow` precompile proposed in TIP-1034


jenpaff:
Closing since this is fixed in the builder for now

tips/tip-1045.md
```diff
@@ -0,0 +1,83 @@
...
+A call qualifies as a payment call if, for some entry in the payment call allow-list table, all of the following hold:
+
+1. It is not a contract-creation call. Calls with `to = null`, `CREATE`, or `CREATE2` never match.
+2. Its target satisfies the entry's target-match predicate.
+3. The first 4 bytes of its calldata equal the entry's selector.
+4. Its total calldata length equals the entry's listed ABI-encoded length (selector + parameters, with no trailing bytes).
+
+The initial payment call allow-list is:
+
+| Target | Selector | Function | Calldata length (bytes) |
```
legion2002: We can also define a simple formula for the calldata length
```
 calldata length bytes = 4 + num_params * 32 bytes
```

tips/tip-1045.md
```diff
@@ -0,0 +1,83 @@
...
+| TIP-20 | `mint(address,uint256)` | Token mint | 68 bytes |
+| TIP-20 | `mintWithMemo(address,uint256,bytes32)` | Mint with memo | 100 bytes |
+| TIP-20 | `burn(uint256)` | Token burn | 36 bytes |
+| TIP-20 | `burnWithMemo(uint256,bytes32)` | Burn with memo | 68 bytes |
+| TIP20ChannelEscrow | `open(address,address,uint96,bytes32,address,uint32)` | Open channel | 196 bytes |
+| TIP20ChannelEscrow | `topUp((address,address,address,bytes32,address),uint96,uint32)` | Top up channel | 228 bytes |
+| TIP20ChannelEscrow | `settle((address,address,address,bytes32,address),uint96,bytes)` | Settle channel | up to 2048 bytes |
+| TIP20ChannelEscrow | `close((address,address,address,bytes32,address),uint96,uint96,bytes)` | Close channel | up to 2048 bytes |
+
+The `Target` column may specify either an exact target precompile address or an address class. For TIP-20 precompiles, the match is against the `0x20C000000000000000000000` prefix.
```
legion2002, line 68: is it just against the prefix, or do we also want to do the code check to ensure the TIP 20 is actually deploye?

tips/tip-1045.md
```diff
@@ -0,0 +1,83 @@
...
+
+Any transaction carrying a non-empty `access_list` MUST be classified as general, regardless of its type, targets, or calldata. Access lists can carry arbitrary addresses and storage keys, enabling arbitrary data injection at payment lane gas prices.
+
+## Transaction Types
+
+For legacy, EIP-2930, EIP-1559, and EIP-7702 transactions, the single top-level call must match the payment call allow-list and `access_list` must be empty.
+
+For account-abstraction (type `0x76`) transactions, `calls` must be non-empty, every call in the batch must individually match the payment call allow-list, and `access_list` must be empty.
+
+No other transaction fields participate in consensus payment classification. In particular, `authorization_list`, `tempo_authorization_list`, `key_authorization`, signature fields, fee sponsorship fields, and validity-window fields do not affect payment versus general classification.
```
legion2002: Think we should disallow both authorization lists, key authorization have a legitimate use case because they are used for user onboarding so we can keep it in.

tips/tip-1045.md
0xrusowsky: current impl only checks the prefix. if they call a non-deployed TIP20 it will revert with `Uninitialized()`.

i don't think the extra SLOAD is worth it, since it doesn't look like real spam concern?

tips/tip-1045.md
0xrusowsky: addressed in [`7f4459b` (this PR)](https://github.com/tempoxyz/tempo/pull/3496/commits/7f4459b71151eacceb63eaa50a893059c50dcfea)

tips/tip-1045.md
0xrusowsky: addressed in [`60a03e5` (this PR)](https://github.com/tempoxyz/tempo/pull/3496/commits/60a03e5de41e96b3d1f8eee19399083f85793346)

tips/tip-1045.md
```diff
@@ -0,0 +1,95 @@
+---
+id: TIP-1045
+title: Payment Transaction Classification
+description: Defines consensus rules for classifying transactions as payment lane eligible using an allow-list of call targets/selectors and empty access lists.
+authors: @0xrusowsky, Arsenii Kulikov @klkvr, Dan Robinson @danrobinson, Tanishk Goyal @legion2002
+status: Draft
+related: TIP-1010, TIP-1000, Payment Lane Specification
```
legion2002: would add TIP 1034 to related, since we have MPP fn selectors here.

legion2002 (approved):
lgtm

tips/tip-1045.md
0xrusowsky:
```suggestion
related: TIP-1010, TIP-1000, Payment Lane Specification, TIP-20, TIP-1034
```

tips/tip-1045.md
```diff
@@ -0,0 +1,95 @@
...
+
+1. It is not a contract-creation call. Calls with `to = null`, `CREATE`, or `CREATE2` never match.
+2. Its target satisfies the entry's target-match predicate.
+3. The first 4 bytes of its calldata equal the entry's selector.
+4. Its total calldata length equals the entry's listed ABI-encoded length (selector + parameters, with no trailing bytes).
+
+Calldata length is derived from the entry's signature rather than listed per row:
+
+- For entries whose parameters are all static ABI types, the expected calldata length is exactly `4 + num_params * 32` bytes (tuples of static types are inlined, so each leaf static type contributes 32 bytes).
+- For entries containing any dynamic ABI type (`bytes`, `string`, dynamic arrays), the encoded length depends on the payload. The constraint becomes `calldata_length <= 2048`, and the encoding must still be well-formed with no trailing bytes beyond the ABI-encoded payload.
```
klkvr: enforcing the well-formed encoding would be kind of tricky right? i'm wondering if for TIP we should just define the `<= 2048` part and leave out the specific encoding check for the builder-side filtering

klkvr (approved):

tips/tip-1045.md
0xrusowsky: with the sol! macro is quite easy, we already do it:

https://github.com/tempoxyz/tempo/blob/869dcf2123e606bf9e05966ca6cf32a189b86436/crates/contracts/src/precompiles/tip20.rs#L189-L193

0xrusowsky: rewritten for clarity [`48a9fdf` (this PR)](https://github.com/tempoxyz/tempo/pull/3496/commits/48a9fdf586b5764df9baa1539ac4ac7b87ba5cd6)

tips/tip-1045.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Access Lists
+
+Any transaction carrying a non-empty `access_list` MUST be classified as general, regardless of its type, targets, or calldata. Access lists can carry arbitrary addresses and storage keys, enabling arbitrary data injection at payment lane gas prices.
+
+## Authorization Lists
+
+Any transaction carrying a non-empty `authorization_list` (EIP-7702) or non-empty `tempo_authorization_list` MUST be classified as general. Like access lists, both fields are unbounded in entry count and each entry carries attacker-chosen signed payloads, enabling arbitrary data injection at payment lane gas prices.
+
+`key_authorization` is exempt from this restriction and MAY be present in payment transactions. It is a single, size-bounded structure tied to onboarding the transaction's sender, and removing it from the payment lane would break first-transaction UX without offering a comparable data-injection surface.
```
danrobinson, line 80: Can't a key_authorization have up to 256 token limits? So something like 16 KB? Seems really big

klkvr: I believe we're planning to limit them at the builder level

0xrusowsky: it was to not have to validate them and reuse the webauthn tx limit size constant.

we could also not enforce an explicit limit on the calldata length for dyn types, and simply fine-tune it at the builder level

danrobinson: I think some kind of limit on total transaction size for the payment lane makes sense, in consensus. Doesn't necessarily need to micromanage every field on it

legion2002: Yeah we can have a sufficiently big limit. We should just avoid keeping it v tight initially, otherwise we'll have to update it every time we add a new feature

0xrusowsky: addressed in [`a895cb3` (this PR)](https://github.com/tempoxyz/tempo/pull/3496/commits/a895cb3b621fefd4d50f728f0ed47adcd0a9a9b1)

danrobinson (approved):

tips/tip-1045.md
```diff
@@ -0,0 +1,100 @@
...
+status: Draft
+related: TIP-1010, TIP-1000, Payment Lane Specification, TIP-20, TIP-1034
+protocolVersion: T5
+---
+
+# TIP-1045: Payment Transaction Classification
+
+## Abstract
+
+This TIP formalizes the consensus rules for classifying a transaction as payment lane eligible under TIP-1010. Starting with the T5 hardfork, consensus requires every call in the transaction to match a payment call allow-list entry (target + selector + exact ABI-encoded calldata length) and requires `access_list` to be empty.
```
samczsun: There are other requirements too beyond `access_list`, might be better to say "and other restrictions" or something?

tips/tip-1045.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Motivation
+
+TIP-1010 defines the payment lane gas budget but leaves the classification criteria underspecified — only requiring that the transaction targets a TIP-20 address. As payment lane eligibility expands beyond TIP-20 calls, consensus needs a generic but stable predicate that can be maintained indefinitely for historical validation.
+
+This TIP therefore enshrines only the minimum long-term consensus rule: a transaction is payment lane eligible if every call it contains matches an allow-listed payment call shape and the transaction carries no access list. Builders MAY apply stricter local policy when selecting transactions for the payment lane, but such heuristics are intentionally left out of scope.
```
samczsun: same

tips/tip-1045.md
```diff
@@ -0,0 +1,100 @@
...
+
+---
+
+# Specification
+
+A transaction qualifies for the payment lane when **every** call it contains is an allow-listed payment call and the transaction carries an empty `access_list`.
```
samczsun: This sentence says that (1) every call is an allowed call [good, well defined] and (2) empty `access_list` but fails to mention the authorization lists or key authorizations. We should enumerate all these restrictions up-front and then elaborate in each section rather than burying the lede

tips/tip-1045.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Payment Call Allow-List
+
+A call qualifies as a payment call if, for some entry in the payment call allow-list table, all of the following hold:
+
+1. It is not a contract-creation call. Calls with `to = null`, `CREATE`, or `CREATE2` never match.
+2. Its target satisfies the entry's target-match predicate.
+3. The first 4 bytes of its calldata equal the entry's selector.
+4. Its calldata satisfies the size and encoding constraints derived from the entry's signature, as defined below.
```
samczsun, line 43: Might be worth clarifying whether it also needs to be canonical encoding, given that it affects whether `len(calldata) ∝ gas` (specifically, calldata could be less than 2048 bytes but point into itself in order to get more memory usage out of the same number of bytes)

tips/tip-1045.md
```diff
@@ -0,0 +1,100 @@
...
+| TIP20ChannelEscrow | `settle((address,address,address,bytes32,address),uint96,bytes)` | Settle channel |
+| TIP20ChannelEscrow | `close((address,address,address,bytes32,address),uint96,uint96,bytes)` | Close channel |
+
+The `Target` column may specify either an exact target precompile address or an address class. For TIP-20 precompiles, the match is against the `0x20C000000000000000000000` prefix.
+
+Calls with empty calldata, unrecognized selectors, target/selector combinations not present in the allow-list, malformed ABI encoding, or calldata that violates the matched entry's calldata length constraint, do not match any allow-list entry.
+
+## Access Lists
+
+Any transaction carrying a non-empty `access_list` MUST be classified as general, regardless of its type, targets, or calldata. Access lists can carry arbitrary addresses and storage keys, enabling arbitrary data injection at payment lane gas prices.
```
samczsun: Do access lists actually inject data...?

tips/tip-1045.md
```diff
@@ -0,0 +1,100 @@
...
+
+`key_authorization` is exempt from this restriction and MAY be present in payment transactions. It is a single, size-bounded structure tied to onboarding the transaction's sender, and removing it from the payment lane would break first-transaction UX without offering a comparable data-injection surface.
+
+To bound the data-injection surface that `key_authorization` itself can carry, payment transactions MUST satisfy `len(rlp(key_authorization)) <= 1024` bytes. Any payment-eligible transaction whose `key_authorization` exceeds this bound MUST be classified as general. The 1 KB ceiling comfortably fits realistic provisioning payloads with limits and scopes.
+
+No consensus-level cap is placed on the overall transaction size: payment transactions are free to batch many TIP-20 calls in a single tx, and any tx-size shaping is left to the builder.
+
+## Transaction Types
```
samczsun, line 86: This feels like it should be at the top of this section

tips/tip-1045.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Pre-T5 (Backward Compatibility)
+
+Before T5, only the legacy TIP-20 address prefix check is enforced at the consensus level.
+
+# Invariants
+
+1. **Classification completeness**: Every transaction MUST be classified as exactly one of payment or general — never both, never neither.
+2. **Allow-list strictness**: A transaction containing any call that does not match an allow-listed `(target, selector, calldata constraint)` entry MUST be classified as general and subject to `general_gas_limit`.
+3. **No trailing bytes**: Calldata longer than the expected ABI-encoded length for a matched allow-list entry MUST be classified as general.
```
samczsun: This is only true for calls with static parameters

tips/tip-1045.md
```diff
@@ -0,0 +1,100 @@
...
+
+# Specification
+
+A transaction qualifies for the payment lane when **every** call it contains is an allow-listed payment call and the transaction carries an empty `access_list`.
+
+## Payment Call Allow-List
+
+A call qualifies as a payment call if, for some entry in the payment call allow-list table, all of the following hold:
+
+1. It is not a contract-creation call. Calls with `to = null`, `CREATE`, or `CREATE2` never match.
```
samczsun: What does `CREATE` or `CREATE2` mean in a transaction context?

tips/tip-1045.md
```diff
@@ -0,0 +1,100 @@
...
+
+Before T5, only the legacy TIP-20 address prefix check is enforced at the consensus level.
+
+# Invariants
+
+1. **Classification completeness**: Every transaction MUST be classified as exactly one of payment or general — never both, never neither.
+2. **Allow-list strictness**: A transaction containing any call that does not match an allow-listed `(target, selector, calldata constraint)` entry MUST be classified as general and subject to `general_gas_limit`.
+3. **No trailing bytes**: Calldata longer than the expected ABI-encoded length for a matched allow-list entry MUST be classified as general.
+4. **No contract creation**: Any transaction containing `CREATE`, `CREATE2`, or `to = null` MUST be classified as general.
+5. **AA atomicity**: An AA transaction is a payment only if `calls` is non-empty and **all** of its calls independently match the payment call allow-list. A single non-qualifying call MUST cause the entire transaction to be classified as general.
```
samczsun, line 104: Isn't this invariant 2?

samczsun (approved):

tips/tip-1045.md
```diff
@@ -0,0 +1,100 @@
+---
+id: TIP-1045
+title: Payment Transaction Classification
+description: Defines consensus rules for classifying transactions as payment lane eligible using an allow-list of call targets/selectors and empty access lists.
+authors: @0xrusowsky, Arsenii Kulikov @klkvr, Dan Robinson @danrobinson, Tanishk Goyal @legion2002
+status: Draft
```
jenpaff:
```suggestion
status: Approved
```

tips/tip-1045.md
0xrusowsky: addressed in [`acaec75` (this PR)](https://github.com/tempoxyz/tempo/pull/3496/commits/acaec75ea53031dde49a45ff366dd656e2fcf084)

tips/tip-1045.md
0xrusowsky: addressed in [`acaec75` (this PR)](https://github.com/tempoxyz/tempo/pull/3496/commits/acaec75ea53031dde49a45ff366dd656e2fcf084)

tips/tip-1045.md
0xrusowsky: addressed in [`acaec75` (this PR)](https://github.com/tempoxyz/tempo/pull/3496/commits/acaec75ea53031dde49a45ff366dd656e2fcf084)

tips/tip-1045.md
0xrusowsky: addressed in [`acaec75` (this PR)](https://github.com/tempoxyz/tempo/pull/3496/commits/acaec75ea53031dde49a45ff366dd656e2fcf084)

tips/tip-1045.md
0xrusowsky: addressed in [`acaec75` (this PR)](https://github.com/tempoxyz/tempo/pull/3496/commits/acaec75ea53031dde49a45ff366dd656e2fcf084)

tips/tip-1045.md
0xrusowsky: addressed in [`acaec75` (this PR)](https://github.com/tempoxyz/tempo/pull/3496/commits/acaec75ea53031dde49a45ff366dd656e2fcf084)

tips/tip-1045.md
0xrusowsky: addressed in [`acaec75` (this PR)](https://github.com/tempoxyz/tempo/pull/3496/commits/acaec75ea53031dde49a45ff366dd656e2fcf084)

tips/tip-1045.md
0xrusowsky: addressed in [`acaec75` (this PR)](https://github.com/tempoxyz/tempo/pull/3496/commits/acaec75ea53031dde49a45ff366dd656e2fcf084)

tips/tip-1045.md
0xrusowsky: addressed in [`acaec75` (this PR)](https://github.com/tempoxyz/tempo/pull/3496/commits/acaec75ea53031dde49a45ff366dd656e2fcf084)

PR #3501, opened 2026-04-08
fix(primitives): use nonzero aa validity bounds

legion2002:
Follow-up alternative to #3482.

Makes `TempoTransaction.valid_before` and `TempoTransaction.valid_after` use `Option<NonZeroU64>` at the primitives layer so `Some(0)` cannot be represented in core AA transaction types. The PR keeps request-side APIs on plain `u64`, rejects `0` explicitly at RPC/JSON boundaries, and converts back to raw `u64` only in downstream execution and pool components that store or compare timestamps.

Verified with `cargo check -p tempo-primitives -p tempo-alloy -p tempo-revm -p tempo-transaction-pool -p tempo-payload-builder`, `cargo test -p tempo-primitives --lib transaction::tempo_transaction -- --nocapture`, `cargo test -p tempo-alloy validity_window -- --nocapture`, `cargo test -p tempo-revm test_from_recovered_tx_expiring_nonce_hash -- --nocapture`, `cargo test -p tempo-transaction-pool test_expiring_nonce_intrinsic_gas_uses_lower_cost -- --nocapture`, and `cargo fmt --all --check`.


crates/alloy/src/rpc/request.rs
```diff
@@ -258,6 +259,28 @@ impl TempoTransactionRequest {
...
+        let valid_after = match self.valid_after {
+            Some(0) => {
+                return Err(ValueError::new(
+                    self,
+                    "'valid_after' cannot be 0; omit the field when no lower bound is intended.",
+                ));
+            }
+            Some(value) => Some(NonZeroU64::new(value).unwrap()),
+            None => None,
+        };
```
klkvr: can we not use `NonZeroU64` on the request type as well?

crates/primitives/src/reth_compat/transaction/tempo_transaction.rs
```diff
@@ -5,3 +5,21 @@ impl reth_primitives_traits::InMemorySize for TempoTransaction {
...
+        self.encode(buf);
+        self.length()
+    }
+
+    fn from_compact(mut buf: &[u8], _len: usize) -> (Self, &[u8]) {
+        let item = alloy_rlp::Decodable::decode(&mut buf)
+            .expect("Failed to decode TempoTransaction from compact");
+        (item, buf)
+    }
+}
```
klkvr: hmm this looks like a breaking change? we can't just change storage encoding schema like that

does `Compact` not work well with `NonZeroU64`? we can fix this in reth upstream

crates/primitives/src/transaction/tempo_transaction.rs
```diff
@@ -903,6 +884,34 @@ impl<'a> arbitrary::Arbitrary<'a> for TempoTransaction {
     }
 }
 
+#[cfg(feature = "serde")]
+mod serde_nonzero_quantity_opt {
```
klkvr: can we dedup this with the `KeyAuthorization` helper mod?

crates/primitives/src/transaction/tempo_transaction.rs
```diff
@@ -529,27 +535,8 @@ impl TempoTransaction {
...
-                buf.advance(1);
-                None
-            } else {
-                Some(Decodable::decode(buf)?)
-            }
-        } else {
-            return Err(alloy_rlp::Error::InputTooShort);
-        };
+        let valid_before = decode_optional_nonzero_u64(buf)?;
+        let valid_after = decode_optional_nonzero_u64(buf)?;
```
klkvr: isn't this same thing as `u64::decode(buf).map(NonZeroU64::new)?`?

crates/primitives/src/reth_compat/transaction/tempo_transaction.rs
legion2002: Yes it's because Compact doesn't have a NonZeroU64 impl. If we can make this change upstream, that would be best.
Opened a PR here - https://github.com/paradigmxyz/reth-core/pull/9

crates/primitives/src/transaction/tempo_transaction.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/pull/3501/changes/d1cad18863131b3756800a0e015370831238042a

crates/primitives/src/transaction/tempo_transaction.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/pull/3501/changes/d1cad18863131b3756800a0e015370831238042a and https://github.com/tempoxyz/tempo/pull/3501/commits/f9b5ea37b39ce5924b1908f6a16d802de0429bf0

crates/alloy/src/rpc/request.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/pull/3501/commits/b22eb1caf974e71226043694f19c4ab95b93a4c4

crates/primitives/src/reth_compat/transaction/tempo_transaction.rs
legion2002: updated here - https://github.com/tempoxyz/tempo/pull/3501/commits/6345f226addeb0052aed61e14b859a8856c41120

klkvr (approved):

0xrusowsky (approved):

PR #3500, opened 2026-04-08
fix(primitives): use nonzero key expiry

legion2002:
Alternative to #3482.

Makes `KeyAuthorization.expiry` an `Option<NonZeroU64>` so `Some(0)` cannot be represented in primitives and therefore cannot silently normalize to `None` during encoding or decoding. The PR keeps the existing wire shape for absent expiries and adds serde coverage to reject `0x0` explicitly.

Verified with `cargo test -p tempo-primitives --lib transaction::key_authorization -- --nocapture`, `cargo test -p tempo-transaction-pool test_key_authorization_valid_expiry_accepted -- --nocapture`, and `cargo fmt --all --check`.

PR #3527, opened 2026-04-09
test(rpc): gate remote fill compatibility cases

legion2002:
Remote devnet/testnet RPCs can lag main for a bit, so this gates selector-scoped key auth fill/estimate cases on known hardfork support instead of assuming the current branch's RPC behavior is deployed everywhere.

It also keeps the `eth_fillTransaction` revert-decoding regression local-only, so we still validate the branch's formatting behavior without failing against older remote environments.

This should remove the false negatives that were blocking PR #3500 and similar rebases.

### DIFF
--- a/crates/node/tests/it/tempo_transaction/rpc.rs
+++ b/crates/node/tests/it/tempo_transaction/rpc.rs
@@ -63,6 +63,7 @@ pub(super) struct RpcEnv {
     provider: alloy::providers::RootProvider,
     chain_id: u64,
     hardfork: TempoHardfork,
+    supports_scoped_key_auth_rpc: bool,
 }
 
 impl RpcEnv {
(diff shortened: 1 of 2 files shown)


crates/node/tests/it/tempo_transaction/rpc.rs
```diff
@@ -95,12 +96,17 @@ impl RpcEnv {
             .get_block_by_number(Default::default())
             .await?
             .ok_or_else(|| eyre::eyre!("latest block missing"))?;
-        let hardfork = chain_spec.tempo_hardfork_at(latest_block.header.timestamp());
+        let latest_timestamp = latest_block.header.timestamp();
+        let hardfork = chain_spec.tempo_hardfork_at(latest_timestamp);
+        let supports_scoped_key_auth_rpc =
+            TempoHardfork::from_chain_and_timestamp(chain_id, latest_timestamp)
+                .is_some_and(|fork| fork.is_t3());
```
0xrusowsky, line 103: i don't think this is a good solution... `supports_scoped_key_auth_rpc` shouldn't be needed at all.

instead, we should simply drop `chainspec` and build `hardfork` from `TempoHardfork::from_chain_and_timestamp`

0xrusowsky (changes requested):

crates/node/tests/it/tempo_transaction/rpc.rs
0xrusowsky: also according to https://github.com/tempoxyz/tempo/pull/3527#discussion_r3058521212, this method only works for mainnet and testnet, so for devnet scoped_key_auth_rpc is always false

PR #3532, opened 2026-04-09
fix: re-validate transactions on policy change

klkvr:
fixes SIGP-193

On token policy changes, transactions using this token as a fee token might get invalidated.

This might affect a lot of transactions and checking every single one would require potentially expensive storage lookups, thus we don't want to perform this check in maintenance loop directly to ensure that we don't leave pool in an inconsistent state for too long

Instead, we are removing all of the transactions for this fee token from the transaction pool immediately and re-add them forcing full re-validation to run and filter out transactions that became invalid after the policy change

### DIFF
--- a/crates/transaction-pool/src/maintain.rs
+++ b/crates/transaction-pool/src/maintain.rs
@@ -68,6 +68,10 @@ pub struct TempoPoolUpdates {
     pub whitelist_removals: Vec<(u64, Address)>,
     /// Fee token pause state changes: (token, is_paused).
     pub pause_events: Vec<(Address, bool)>,
+    /// Tokens whose transfer policy was changed via `changeTransferPolicyId()`.
+    /// Pending transactions using these tokens as fee tokens need to be re-validated
+    /// because the new policy may forbid the fee payer or fee manager.
+    pub transfer_policy_updates: HashSet<Address>,
     /// Fee token balance changes keyed by token.
     ///
     /// We only track the debited `from` account from TIP20 `Transfer` logs because credits to the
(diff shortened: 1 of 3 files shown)


crates/transaction-pool/src/maintain.rs
```diff
@@ -582,6 +589,59 @@ where
...
+                // 5b. Handle transfer policy updates
+                // When a token's transfer policy changes, pending transactions using that
+                // token may become invalid under the new policy. We remove them and re-add
+                // so they go through full validation against the updated policy.
+                if !updates.transfer_policy_updates.is_empty() {
+                    let all_txs = pool.all_transactions();
+                    let hashes: Vec<TxHash> = all_txs
+                        .pending
+                        .iter()
+                        .chain(all_txs.queued.iter())
```
mattsse, line 601: I believe there should be a helper for this chained iterator now

mattsse (approved):

crates/transaction-pool/src/maintain.rs
klkvr: hmm yeah there is but it yields `Recovered<ConsensusTx>` instead of full pool transactions

crates/transaction-pool/src/maintain.rs
```diff
@@ -160,6 +165,8 @@ impl TempoPoolUpdates {
             else if is_tip20_prefix(log.address) {
                 if let Ok(event) = ITIP20::PauseStateUpdate::decode_log(log) {
                     updates.pause_events.push((log.address, event.isPaused));
+                } else if ITIP20::TransferPolicyUpdate::decode_log(log).is_ok() {
```
fgimenez, line 168: should we check `receipt.success == true` here?

klkvr: logs are not emitted for events that were reverted

PR #3537, opened 2026-04-09
fix(revm): move scoped key validation to pre-execution

legion2002:
## Problem

T3 scoped access keys currently validate every batch call against call scopes in the transaction-validation path. That makes call-scope matching part of transaction validity rather than part of execution.

This matters because validation work is not paid work. Even if we meter the storage reads performed during validation, nodes still do that work before the transaction has entered execution, so they are not guaranteed to get paid for it. A malicious sender can therefore force validators or mempool nodes to spend CPU on transactions that are eventually discarded.

This is also the first check in the handler that adds unbounded validation complexity. Most existing validity checks are constant-time or otherwise tightly bounded by transaction structure. Scoped call validation, by contrast, scales with the number of calls in the batch, and the configured limits are now high enough that this becomes a real DoS concern.

The existing implementation also extends the dynamic `initial_gas` pattern: it runs the scope checks first, measures the cost afterwards, and then folds that cost back into `initial_gas`. That pattern has already caused multiple bugs in this part of the handler and makes the control flow harder to reason about.

## Solution

This change moves full scoped-call matching out of `validate_against_state_and_deduct_caller` and into a metered pre-execution phase at the start of AA batch execution.

The new behavior is:

- transaction validation still performs only the cheap keychain checks needed to determine whether the transaction is structurally acceptable
- full per-call scope matching happens after validation succeeds, but before the first user call executes
- if any call fails scope validation, the batch fails atomically before any user call runs
- that failure is now an execution failure with normal gas accounting, rather than an invalid-transaction rejection

This is the best fix because it solves the actual problem from first principles:

- metering validation alone does not help, because unpaid validation work is still unpaid
- adding more worst-case upfront charging and refund logic would make the already fragile `initial_gas` flow even more complex
- moving the checks into pre-execution keeps validity cheap and bounded, preserves atomic user protection, and ensures scoped-call failures are paid execution failures

TIP-1011 is updated to reflect the new semantics.

## Tests

- `cargo test -p tempo-revm scope_validation_moves_to_execution`
- `cargo test -p tempo-revm deny_all_scopes_blocks_same_tx_call`


crates/revm/src/handler.rs
```diff
@@ -396,6 +488,23 @@ where
 
         let mut final_result = None;
 
+        if let Some(mut frame_result) =
+            self.prevalidate_keychain_call_scopes(evm, &calls, &mut remaining_gas)?
+        {
+            let total_gas_spent = (gas_limit - remaining_gas) + frame_result.gas().spent();
+
```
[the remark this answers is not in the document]
legion2002: we block CREATE for access keys, and we keep the check a part of tx invalidation in this commit - https://github.com/tempoxyz/tempo/pull/3537/changes/dbeae69ce1f31158690068a334a1bcbada9d0c2f

So we won't face this issue, because txs which have access keys and CREATE will just be considered invalid, and should not have a nonce bump.

crates/revm/src/handler.rs
```diff
@@ -311,6 +311,102 @@ impl<DB, I> TempoEvmHandler<DB, I>
...
+            .first()
+            .expect("AA transactions must contain at least one call")
+            .to;
+
+        let ctx = evm.ctx_mut();
+        let internals = EvmInternals::new(&mut ctx.journaled_state, &ctx.block, &ctx.cfg, &ctx.tx);
+        let mut provider =
+            EvmPrecompileStorageProvider::new_with_gas_limit(internals, &ctx.cfg, *remaining_gas);
+
+        let validation = StorageCtx::enter(&mut provider, || {
```
klkvr: can we use some of the helper APIs on `StorageCtx` here? like `StorageCtx::enter_ctx`

crates/revm/src/handler.rs
```diff
@@ -396,6 +492,26 @@ where
 
         let mut final_result = None;
 
+        if let Some(mut frame_result) =
+            self.prevalidate_keychain_call_scopes(evm, &calls, &mut remaining_gas)?
+        {
+            // This return happens before any call frame is constructed. `make_create_frame` has
+            // not run yet, so the protocol nonce is still untouched and there is nothing to bump
+            // or preserve here.
```
klkvr: let's clarify that this is only reachable for keychain transactions for which CREATE is not allowed which means that nonce was already bumped

crates/revm/src/handler.rs
```diff
@@ -396,6 +492,26 @@ where
...
+            let total_gas_spent = (gas_limit - remaining_gas) + frame_result.gas().spent();
+
+            let mut corrected_gas = Gas::new(gas_limit);
+            if frame_result.instruction_result().is_revert() {
+                corrected_gas.set_spent(total_gas_spent);
+            } else {
+                corrected_gas.spend_all();
+            }
+            corrected_gas.set_refund(0);
+            *frame_result.gas_mut() = corrected_gas;
```
klkvr: i'm wondering if we could unify this somehow with the logic below when individual calls fail?

crates/revm/src/handler.rs
```diff
@@ -453,19 +571,18 @@ where
                     }
                 }
 
-                // Include gas from all previous successful calls + failed call
-                let gas_spent_by_failed_call = frame_result.gas().spent();
-                let total_gas_spent = (gas_limit - remaining_gas) + gas_spent_by_failed_call;
```
klkvr, line 458: we shouldn't need to change any of the logic here i think?

crates/revm/src/handler.rs
```diff
@@ -396,6 +492,26 @@ where
 
         let mut final_result = None;
```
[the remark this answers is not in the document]
legion2002: > [ISSUE] Scope-denied same-tx auth+use still leaves the access key authorized (Low — T3-only)

I think we want to keep the authorized key in this case, because the user has already paid for it. And call scope related reverts during execution should not have any impact on key authorization. 

> [ISSUE] Queued access-key txs survive scope tightening and are mined as paid reverts (Medium — T3-only)

we can address this separately in ths tx pool PR we have for this #3440

crates/revm/src/handler.rs
legion2002: we can't use `enter_ctx` here because we want to provide a specific limit, and `enter_ctx` always build the provider with max gas

I can create new helpers with this functionality if you want

crates/revm/src/handler.rs
legion2002: yeah that was just a variable name change. addressed in  - https://github.com/tempoxyz/tempo/pull/3537/commits/3b382237e9ba709634eab9d75df5f5564e11c25f

crates/revm/src/handler.rs
legion2002: addressed in -https://github.com/tempoxyz/tempo/pull/3537/commits/3b382237e9ba709634eab9d75df5f5564e11c25f

crates/revm/src/handler.rs
legion2002: addressed in - https://github.com/tempoxyz/tempo/pull/3537/commits/3b382237e9ba709634eab9d75df5f5564e11c25f

crates/revm/src/handler.rs
```diff
@@ -311,6 +332,102 @@ impl<DB, I> TempoEvmHandler<DB, I>
...
+                    FrameResult::Create(revm::interpreter::CreateOutcome::new(
+                        revm::interpreter::InterpreterResult::new(
+                            revm::interpreter::InstructionResult::Revert,
+                            Default::default(),
+                            gas,
+                        ),
+                        None,
+                    ))
+                };
+
```
klkvr: let's not use fully-qualified paths here so that it's more readable

also i think we should propagate the revert reason bytes when applicable? so that when revert is returned on rpc it can be parsed to understand what went wrong

crates/revm/src/handler.rs
```diff
@@ -419,12 +545,14 @@ where
...
-            let mut frame_result = frame_result?;
+            let frame_result = frame_result?;
 
             // Check if call succeeded
             let instruction_result = frame_result.instruction_result();
             if !instruction_result.is_ok() {
-                // Revert checkpoint - rolls back ALL state changes from ALL calls
+                let mut frame_result = frame_result;
+
+                // Revert checkpoint - rolls back ALL state changes from all executed calls.
```
klkvr: this change seems unnecessary as well?

crates/revm/src/handler.rs
legion2002: addressed in - https://github.com/tempoxyz/tempo/pull/3537/changes/4cfa46e4f390c20136839ca883b1a22c416dd643

crates/revm/src/handler.rs
legion2002: addressed in - https://github.com/tempoxyz/tempo/pull/3537/changes/4cfa46e4f390c20136839ca883b1a22c416dd643

crates/revm/src/handler.rs
```diff
@@ -163,6 +163,27 @@ fn call_scope_storage_slots(auth: &tempo_primitives::transaction::KeyAuthorizati
...
+    let mut corrected_gas = Gas::new(gas_limit);
+    if frame_result.instruction_result().is_revert() {
+        corrected_gas.set_spent(total_gas_spent);
+    } else {
+        corrected_gas.spend_all();
+    }
+    corrected_gas.set_refund(0); // No refunds when batch fails and all state is reverted.
+    *frame_result.gas_mut() = corrected_gas;
+    frame_result
+}
```
klkvr: should this take `&mut frame_result`?

crates/revm/src/handler.rs
```diff
@@ -311,6 +332,97 @@ impl<DB, I> TempoEvmHandler<DB, I>
...
+            };
+
+            let access_key_addr = if let Some(override_key_id) = tempo_tx_env.override_key_id {
+                override_key_id
+            } else {
+                keychain_sig
+                    .key_id(&tempo_tx_env.signature_hash)
+                    .map_err(|_| TempoInvalidTransaction::AccessKeyRecoveryFailed)?
+            };
+
```
klkvr, line 377: nitpick: this should ideally panic, returning `InvalidTransaction` when we're already in execution is very unexpected

crates/revm/src/handler.rs
klkvr: that would be nice but obv not a blocker

crates/revm/src/handler.rs
```diff
@@ -311,6 +332,97 @@ impl<DB, I> TempoEvmHandler<DB, I>
...
+            Ok(()) => {
+                *remaining_gas = remaining_gas.saturating_sub(gas_used);
+                Ok(None)
+            }
+            Err(TempoPrecompileError::OutOfGas) => Ok(Some(oog_frame_result(kind, *remaining_gas))),
+            Err(TempoPrecompileError::Fatal(err)) => Err(EVMError::Custom(err)),
+            Err(err) => {
+                let revert_output = err
+                    .into_precompile_result(gas_used)
+                    .expect("non-fatal prevalidation errors must encode as a revert");
```
klkvr: this panic actually seems a bit dangerous to me, ideally we should refactor this so that any error is converted to a `PrecompileResult` and then OOG/Fatal errors are handled separately

crates/revm/src/handler.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/pull/3537/commits/e542c279175184a2db39a9dc8f547d9b5e099750

crates/revm/src/handler.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/pull/3537/commits/e542c279175184a2db39a9dc8f547d9b5e099750

crates/revm/src/handler.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/pull/3537/commits/e542c279175184a2db39a9dc8f547d9b5e099750

we convert it into a Custom EVM error

crates/revm/src/handler.rs
legion2002: added in - https://github.com/tempoxyz/tempo/pull/3537/commits/7bab788cc38a2e6bb6d53327e2889400e16bbd9a

crates/revm/src/handler.rs
```diff
@@ -163,6 +164,22 @@ fn call_scope_storage_slots(auth: &tempo_primitives::transaction::KeyAuthorizati
     }
 }
 
+fn correct_failed_batch_gas(frame_result: &mut FrameResult, gas_limit: u64, remaining_gas: u64) {
```
0xrusowsky: nit: can we add docs and/or rename for clarity? it is unclear at usage site what this fn does

crates/revm/src/handler.rs
```diff
@@ -3115,21 +3206,131 @@ mod tests {
...
-                Some(TempoInvalidTransaction::EthInvalidTransaction(
-                    InvalidTransaction::CallGasCostMoreThanGasLimit { .. }
-                ))
+                result.instruction_result(),
+                revm::interpreter::InstructionResult::OutOfGas
             ),
-            "expected CallGasCostMoreThanGasLimit, got: {err:?}"
+            "expected scope validation to fail during execution with OOG, got: {:?}",
+            result.instruction_result()
         );
```
0xrusowsky, line 3234: let's also assert that `correct_failed_batch_gas` consumes the tx gas budget?

0xrusowsky (approved):
i think it makes sense

crates/revm/src/handler.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/pull/3537/commits/84fce48369bba3e7fee756e1cf98a4f2355a52d2

crates/revm/src/handler.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/pull/3537/commits/84fce48369bba3e7fee756e1cf98a4f2355a52d2

crates/revm/src/handler.rs
```diff
@@ -311,6 +338,103 @@ impl<DB, I> TempoEvmHandler<DB, I>
...
+
+        match validation {
+            Ok(()) => {
+                *remaining_gas = remaining_gas.saturating_sub(gas_used);
+                Ok(None)
+            }
+            Err(err) => match err.into_precompile_result(gas_used) {
+                Ok(revert_output) => {
+                    let mut gas = Gas::new(gas_used);
+                    gas.set_spent(gas_used);
```
klkvr, line 407: i guess passing `gas_used` to `Gas::new` is not entirely correct here but it's fine because we're not really using the resulting `Gas` value after

klkvr (approved):

PR #3538, opened 2026-04-10
feat(consensus): enrich validator command with DKG role info

decofe:
The `tempo consensus validator` command now includes `is_dkg_dealer`, `is_dkg_player`, and `in_committee` fields — same role info as `validators-info` but for a single node lookup. Adds `--chain` flag for epoch length resolution.

Prompted by: Janis

### DIFF
--- a/bin/tempo/src/tempo_cmd.rs
+++ b/bin/tempo/src/tempo_cmd.rs
@@ -865,7 +865,7 @@ pub(crate) struct ValidatorInfo {
     #[arg()]
     id: ValidatorId,
 
-    /// Chain to query (presto, testnet, moderato, or path to chainspec file)
+    /// Chain to query (mainnet, testnet, moderato, or path to chainspec file)
     #[arg(long, short, default_value = "mainnet", value_parser = tempo_chainspec::spec::chain_value_parser)]
     chain: Arc<TempoChainSpec>,
 
@@ -1087,7 +1087,7 @@ struct ValidatorEntry {
 
 #[derive(Debug, clap::Args)]
 pub(crate) struct ValidatorsInfo {
-    /// Chain to query (presto, testnet, moderato, or path to chainspec file)
+    /// Chain to query (mainnet, testnet, moderato, or path to chainspec file)
     #[arg(long, short, default_value = "mainnet", value_parser = tempo_chainspec::spec::chain_value_parser)]
     chain: Arc<TempoChainSpec>,
 
--- a/bin/tempo/src/tempo_cmd.rs
+++ b/bin/tempo/src/tempo_cmd.rs
@@ -881,10 +881,8 @@ pub(crate) struct ValidatorInfo {
 /// Output for the single-validator lookup enriched with DKG role and epoch context.
 #[derive(Debug, Serialize)]
 struct SingleValidatorOutput {
-    #[serde(skip_serializing_if = "Option::is_none")]
-    current_epoch: Option<u64>,
-    #[serde(skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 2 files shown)


bin/tempo/src/tempo_cmd.rs
```diff
@@ -864,13 +864,35 @@ pub(crate) struct ValidatorInfo {
     /// Validator ethereum address, ed25519 public key, or index
     #[arg()]
     id: ValidatorId,
+
+    /// Chain to query (presto, testnet, moderato, or path to chainspec file)
```
SuperFluffy: `presto` is not a valid entry. Only `mainnet`, `testnet`, `moderato`, `dev`, and `<PATH>` are.

Check if you can take this value from the `SUPPORTED_CHAINS` const in the `tempo-chainspec` crate.

SuperFluffy (approved):

bin/tempo/src/tempo_cmd.rs
```diff
@@ -864,13 +864,37 @@ pub(crate) struct ValidatorInfo {
     /// Validator ethereum address, ed25519 public key, or index
     #[arg()]
     id: ValidatorId,
+
+    /// Chain to query (mainnet, testnet, moderato, or path to chainspec file)
+    #[arg(long, short, default_value = "mainnet", value_parser = tempo_chainspec::spec::chain_value_parser)]
+    chain: Arc<TempoChainSpec>,
```
hamdiallam: Can we infer this using the chain id from the rpc?

bin/tempo/src/tempo_cmd.rs
```diff
@@ -864,13 +864,37 @@ pub(crate) struct ValidatorInfo {
...
+
     /// RPC URL to query.
     #[arg(long, default_value = "https://rpc.presto.tempo.xyz")]
     rpc_url: String,
+
+    /// Skip crosschecking the validator with the last DKG round.
+    #[arg(long)]
+    no_dkg_information: bool,
```
hamdiallam, line 880: Why would anyone want to skip? Since we're doing rpc here makes more sense to always output this info?

hamdiallam (approved):
LGTM, just a couple comments

bin/tempo/src/tempo_cmd.rs
SuperFluffy: On the one hand probably no - on the other I want to leave the ability to just wrap the smart contract call without any extra steps.

crates/chainspec/src/spec.rs
```diff
@@ -561,4 +572,19 @@ mod tests {
...
+    #[test]
+    fn chainspec_from_chain_id_roundtrips_supported_chains() {
+        use reth_chainspec::EthChainSpec;
+
+        for &name in super::SUPPORTED_CHAINS {
+            let spec = super::chain_value_parser(name)
+                .unwrap_or_else(|e| panic!("chain_value_parser({name}) failed: {e}"));
+
+            let resolved = super::chainspec_from_chain_id(spec.chain().id())
+                .unwrap_or_else(|| panic!("chainspec_from_chain_id failed for {name}"));
```
SuperFluffy: These are tests, so no need to over optimize. Use in both instances:
```
expect(format!("failed to parse chain `{name}`"))
```

bin/tempo/src/tempo_cmd.rs
SuperFluffy: Done: inferring the chainspec in both instances from the chain ID unless `--chain` is specified. Both in `--validator` and `--validator-info` (naming is a bit confusing now).

PR #3531, opened 2026-04-09
chore(precompiles): add hardfork-gating to `dispatch_call`

0xrusowsky:
ref: `SIGP-227`

this PR adds built-in hardfork-gating capabilities to `fn dispatch_call` so that they don't need to be manually implemented. Instead, we can simply pass added/removed selectors at each hardfork.

example usage:

```rs
dispatch_call(
    calldata,
    // NEW INPUT: define a slice `SelectorSchedule` with the added/dropped selectors
    &[SelectorHardforkDiff::new(TempoHardfork::T3).add(ADDED).drop(DROPPED)],
    IAccountKeychainCalls::abi_decode,
    |call| match call { .. }
)
```

PR #3540, opened 2026-04-10
chore(precompiles): migrate manual hardfork gates to SelectorHardforkRule arrays

decofe:
Ref: #3531

Migrates the remaining inline `is_t1()`/`is_t2()` selector checks into static
`SelectorHardforkRule` arrays so `dispatch_call` rejects gated selectors before
ABI decode. Fixes the invariant where selector-only calldata for a gated function
would hit decode first and return an empty revert instead of `UnknownFunctionSelector`.

- **TIP20**: `permit`, `nonces`, `DOMAIN_SEPARATOR` → T2 added
- **TIP403Registry**: `isAuthorizedSender`, `isAuthorizedRecipient`, `isAuthorizedMintRecipient`, `compoundPolicyData`, `createCompoundPolicy` → T2 added
- **ValidatorConfig**: `changeValidatorStatusByIndex` → T1 added

Prompted by: rusowsky

### DIFF
--- a/crates/precompiles/src/account_keychain/dispatch.rs
+++ b/crates/precompiles/src/account_keychain/dispatch.rs
@@ -37,10 +37,9 @@ impl Precompile for AccountKeychain {
             calldata,
             &[(
                 TempoHardfork::T3,
-                SelectorHardforkDiff {
-                    added: T3_ADDED_SELECTORS,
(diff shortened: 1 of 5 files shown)


0xrusowsky:
confirmed that re-execution tests passed

0xrusowsky (approved):

PR #3543, opened 2026-04-10
fix(account-keychain): propagate refund system errors

legion2002:
Propagate system-level storage failures from refund_spending_limit while still treating revoked, missing, and expired keys as no-op refund cases.

Adds a regression test that injects an sload failure to ensure refund_spending_limit surfaces fatal storage errors instead of silently swallowing them.

### DIFF
--- a/crates/precompiles/src/account_keychain/mod.rs
+++ b/crates/precompiles/src/account_keychain/mod.rs
@@ -1300,14 +1300,11 @@ mod tests {
     use super::*;
     use crate::{
         error::TempoPrecompileError,
-        storage::{PrecompileStorageProvider, StorageCtx, hashmap::HashMapStorageProvider},
+        storage::{StorageCtx, hashmap::HashMapStorageProvider},
         test_util::TIP20Setup,
     };
-    use alloy::primitives::{Address, B256, LogData, TxKind, U256};
-    use revm::{
-        context::journaled_state::JournalCheckpoint,
-        state::{AccountInfo, Bytecode},
-    };
+    use alloy::primitives::{Address, B256, TxKind, U256};
+    use revm::state::Bytecode;
     use tempo_chainspec::hardfork::TempoHardfork;
     use tempo_contracts::precompiles::{DEFAULT_FEE_TOKEN, IAccountKeychain::SignatureType};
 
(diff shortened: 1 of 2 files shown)


crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -1332,6 +1336,117 @@ mod tests {
         }
     }
 
+    struct FailingSloadStorageProvider {
+        inner: HashMapStorageProvider,
+        fail_on_sload: Option<(Address, U256)>,
+    }
```
0xrusowsky: this feels overkill.
can we extend `HashMapStorageProvider` with `fail_on_sload` instead?

legion2002: addressed in https://github.com/tempoxyz/tempo/pull/3543/commits/db575861a731f3c67de655745c71d596e2d38a6a

crates/precompiles/src/storage/hashmap.rs
```diff
@@ -15,6 +15,8 @@ pub struct HashMapStorageProvider {
     internals: HashMap<(Address, U256), U256>,
     transient: HashMap<(Address, U256), U256>,
     accounts: HashMap<Address, AccountInfo>,
+    #[cfg(any(test, feature = "test-utils"))]
```
0xrusowsky: nit: since HashMapStorageProvider is test-only, this is technically not needed, and it is enough with the cfg in the impl block

crates/precompiles/src/storage/hashmap.rs
```diff
@@ -48,6 +50,8 @@ impl HashMapStorageProvider {
             internals: HashMap::new(),
             transient: HashMap::new(),
             accounts: HashMap::new(),
+            #[cfg(any(test, feature = "test-utils"))]
```
0xrusowsky: same here

0xrusowsky (approved):

crates/precompiles/src/storage/hashmap.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/pull/3543/changes/65915686384114e43ac61c29777df68a77ae0312

crates/precompiles/src/storage/hashmap.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/pull/3543/changes/65915686384114e43ac61c29777df68a77ae0312

PR #3545, opened 2026-04-11
chore: remove val config v1 support from consensus cli

hamdiallam:
* --with-historic is no longer an applicable flag for the validators-info cmd
* simply some code within validators-info

### DIFF
--- a/bin/tempo/src/tempo_cmd.rs
+++ b/bin/tempo/src/tempo_cmd.rs
@@ -30,9 +30,8 @@ use tempo_alloy::TempoNetwork;
 use tempo_chainspec::spec::{TempoChainSpec, TempoChainSpecParser};
 use tempo_commonware_node_config::SigningKey;
 use tempo_contracts::precompiles::{
-    IValidatorConfig::{self},
     IValidatorConfigV2::{self, Validator},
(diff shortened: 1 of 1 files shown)


bin/tempo/src/tempo_cmd.rs
```diff
@@ -1211,117 +1088,50 @@ impl ValidatorsInfo {
...
+                    .into(),
+            );
+
+        let resp = provider
+            .call(tx.into())
+            .await
+            .wrap_err("failed to call getActiveValidators")?;
+
+        let validators = IValidatorConfigV2::getActiveValidatorsCall::abi_decode_returns(&resp)
+            .wrap_err("failed to decode getActiveValidators response")?;
```
SuperFluffy, line 1105: This is not sufficient as it only returns those validators that are marked active in the contract. It ignores those validators that are members of the committee as per the DKG outcome in the boundary block.

I am just now noticing that the V2 version of this is borked.

SuperFluffy (changes requested):
Validator info is incomplete.

SuperFluffy (changes requested):
Validator info is incomplete.

SuperFluffy (approved):
Approving - fix should happen in a followup PR to not mix these two concerns up.

PR #3557, opened 2026-04-13
chore: remove val config v1 from tempo-e2e

hamdiallam:
* Default to v2 at genesis, divorcing from t2_time
* remove migration_from_v1_to_v2
* v2_at_genesis is the default state
* remove setting t2_time in setup


hamdiallam:
Pushed the wrong branch

PR #3558, opened 2026-04-13
chore: remove val config v1 from tempo-e2e

hamdiallam:
* Default to v2 at genesis, divorcing from t2_time
* remove migration_from_v1_to_v2
* v2_at_genesis is the default state
* remove setting t2_time in setup


crates/e2e/src/tests/dkg/common.rs
```diff
@@ -11,21 +11,6 @@ use tempo_dkg_onchain_artifacts::OnchainDkgOutcome;
 
 use crate::{CONSENSUS_NODE_PREFIX, TestingNode};
 
-/// Returns the target epoch to wait for depending on `event_height`.
```
SuperFluffy, line 14: Why move this around? Let's not do refactors in large chore-removal steps.

crates/e2e/src/tests/dkg/dynamic.rs
SuperFluffy, line 1: why all these changes?

SuperFluffy (changes requested):
Let's keep this patch confined to just removing V1 logic, not extra refactoring and cleanup.

crates/e2e/src/tests/dkg/dynamic.rs
hamdiallam: The test here differed from the test within v2_at_genesis. I believe we want to prefer the v2_at_genesis test which is updated one. I'll double check to make sure the tests are effectively the same

crates/e2e/src/tests/dkg/common.rs
hamdiallam: Oh i'm not sure why this was moved. Let me fix

hamdiallam: Ah think in v2_at_genesis the common.rs had it that way which is why it's showing this change. But I undid it here

crates/e2e/src/tests/dkg/dynamic.rs
hamdiallam: Confirming the files differed between the two but that one we want now is v2_at_genesis

SuperFluffy (approved):

PR #3560, opened 2026-04-13
chore(dex): fail early on order placement for paused tokens

0xrusowsky:
ref: `ZELLIC-190`

### DIFF
--- a/crates/precompiles/src/stablecoin_dex/mod.rs
+++ b/crates/precompiles/src/stablecoin_dex/mod.rs
@@ -191,7 +191,7 @@ impl StablecoinDEX {
     }
 
     /// Attempts to satisfy `amount` from the user's internal DEX balance.
-    /// Applies transfer authorization and, when requested, verifies the token's pause state.
+    /// Applies transfer authorization and optionally (spec: T3+) verifies the token's pause state.
     ///
     /// Returns `None` when the full amount was debited, or `Some(remaining)` when the caller must
     /// source the remaining amount via `transfer_from` without mutating user internal balance.
@@ -221,7 +221,7 @@ impl StablecoinDEX {
     }
 
     /// Decrement user's internal balance and/or transfer from external wallet.
-    /// Applies transfer authorization and, when requested, verifies the token's pause state.
+    /// Applies transfer authorization and optionally (spec: T3+) verifies the token's pause state.
     fn debit_or_transfer_from_with_pause_policy(
         &mut self,
         user: Address,
@@ -5197,15 +5197,10 @@ mod tests {
     }
 
     #[test]
-    fn test_place_orders_with_internal_balance() -> eyre::Result<()> {
-        enum EscrowToken {
-            Base,
-            Quote,
-        }
-
+    fn test_place_orders_on_paused_token_respects_internal_balance_path() -> eyre::Result<()> {
         fn assert_paused_token_order_path<F>(
-            escrow_token_for_order: EscrowToken,
             internal_balance_amount: u128,
+            is_bid: bool,
             mut place_order: F,
         ) -> eyre::Result<()>
         where
@@ -5223,10 +5218,7 @@ mod tests {
                     let (base_token, quote_token) =
                         setup_test_tokens(admin, alice, exchange.address, 500_000_000u128)?;
                     exchange.create_pair(base_token)?;
-                    let escrow_token = match escrow_token_for_order {
-                        EscrowToken::Base => base_token,
-                        EscrowToken::Quote => quote_token,
-                    };
+                    let escrow_token = if is_bid { quote_token } else { base_token };
                     exchange.set_balance(alice, escrow_token, internal_balance_amount)?;
 
                     let mut escrow_tip20 = TIP20Token::from_address(escrow_token)?;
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -492,13 +533,13 @@ impl StablecoinDEX {
             (token, amount, quote_token)
         };
 
-        // Check policy on non-escrow token (escrow token is checked in decrement_balance_or_transfer_from)
+        // Check policy on non-escrow token (escrow token is checked in `try_debit_internal_balance`)
         // Direction: DEX → sender (order placer receives non-escrow token when filled)
         TIP20Token::from_address(non_escrow_token)?
             .ensure_transfer_authorized(self.address, sender)?;
```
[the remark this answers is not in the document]
0xrusowsky: the reason we check the escrow token's paused state is cause it is the only asset actually moved during order placement: via `transfer_from` (checked on the tip20 side) or by decrementing the user's internal balance (now checked with  `try_debit_internal_balance`).

the non-escrow token is not transferred, thus, its pause state is not directly relevant to the placement step itself. Given that paused input tokens are rejected during `validate_and_build_route`, and paused output tokens fail on the final tip20 `transfer`, i don't see an issue.

crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -190,27 +190,63 @@ impl StablecoinDEX {
...
         user: Address,
         token: Address,
         amount: u128,
-    ) -> Result<()> {
-        // Ensure that the token can be transferred
-        TIP20Token::from_address(token)?.ensure_transfer_authorized(user, self.address)?;
+        enforce_pause_on_internal_balance: bool,
+    ) -> Result<Option<u128>> {
+        let tip20 = TIP20Token::from_address(token)?;
+        tip20.ensure_transfer_authorized(user, self.address)?;
```
legion2002, line 206: noting that this seems to have the same storage ops after refactor, so reexec should be good.

crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -190,27 +190,63 @@ impl StablecoinDEX {
...
 
         let user_balance = self.balance_of(user, token)?;
         if user_balance >= amount {
-            self.sub_balance(user, token, amount)
+            if enforce_pause_on_internal_balance && self.storage.spec().is_t3() {
```
legion2002, line 210: noting that we don't need this pause enforcement in the other branch, because TIP 20 transferFrom would already check this.

legion2002 (approved):
lgtm

crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -190,27 +190,63 @@ impl StablecoinDEX {
...
+    fn debit_or_transfer_from_with_pause_policy(
+        &mut self,
+        user: Address,
+        token: Address,
+        amount: u128,
+        enforce_pause_on_internal_balance: bool,
+    ) -> Result<()> {
+        if let Some(remaining) =
+            self.try_debit_internal_balance(user, token, amount, enforce_pause_on_internal_balance)?
+        {
```
0xKitsune, line 234: Why are we returning an option rather then simply returning the remaining amount? If the amount is 0, we can return this directly. This would remove some of the `if let Some()` logic.

crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -190,27 +190,63 @@ impl StablecoinDEX {
...
+    /// Applies transfer authorization and optionally (spec: T3+) verifies the token's pause state.
+    fn debit_or_transfer_from_with_pause_policy(
+        &mut self,
+        user: Address,
+        token: Address,
+        amount: u128,
+        enforce_pause_on_internal_balance: bool,
+    ) -> Result<()> {
+        if let Some(remaining) =
+            self.try_debit_internal_balance(user, token, amount, enforce_pause_on_internal_balance)?
```
0xKitsune, line 233: Can you remind me why we need to add an `enforce_pause_on_internal_balance` bool? This seems to be the cause of most of the complexity in this PR. Would it be possible to simply to the following where the tip20 pause status is only checked when decrementing the balance and `user_balance >= amount`?

```rust

    /// Decrement user's internal balance or transfer from external wallet
    fn decrement_balance_or_transfer_from(
        &mut self,
        user: Address,
        token: Address,
        amount: u128,
    ) -> Result<()> {
        
        // --snip--
        if user_balance >= amount {
            let remaining = self.sub_balance(user, token, amount);
            if self.storage.spec().is_t3() {
                tip20.check_not_paused()?;
            }

            remaining
        } else {
           // --snip--
        }
```

0xrusowsky: that's not possible if we want to avoid redundant SLOADs cause then swaps would check paused state twice (here and in `validate_and_build_route`)

crates/precompiles/src/stablecoin_dex/mod.rs
0xrusowsky: i thought it was semantically clean, but checking `.is_zero()` should be equivalent, good call

0xrusowsky:
closing in favor of https://github.com/tempoxyz/tempo/pull/3586

PR #3561, opened 2026-04-13
chore: remove val config v1 from consensus

hamdiallam:
* v2 is initialized at genesis & mainnet is migrated
* v1 pathways in the peer and dkg manager will never be hit again. remove them
* subscribe_finalized is no longer needed in the v2 implementation

### DIFF
--- a/crates/commonware-node/src/dkg/manager/actor/mod.rs
+++ b/crates/commonware-node/src/dkg/manager/actor/mod.rs
@@ -1,6 +1,6 @@
-use std::{collections::BTreeMap, num::NonZeroU32, task::Poll, time::Duration};
+use std::{collections::BTreeMap, num::NonZeroU32, task::Poll};
 
-use alloy_consensus::{BlockHeader as _, Sealable as _};
+use alloy_consensus::BlockHeader as _;
 use alloy_primitives::B256;
 use bytes::{Buf, BufMut};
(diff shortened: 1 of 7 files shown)


crates/commonware-node/src/dkg/manager/actor/mod.rs
```diff
@@ -805,22 +798,13 @@ where
...
-        .wrap_err("failed reading contract to determine syncers")?;
-
         Ok(Some(state::State {
             epoch: onchain_outcome.epoch,
             seed: Summary::random(&mut self.context),
             output: onchain_outcome.output.clone(),
             share,
             players: onchain_outcome.next_players,
-            syncers,
+            syncers: ordered::Set::default(),
```
SuperFluffy, line 813: Should make a followup to hide this as an implementation detail (or do some migration logic?).

SuperFluffy (approved):
nice cleanup

crates/commonware-node/src/dkg/manager/actor/mod.rs
hamdiallam: Yeah agreed, I think we can migrate the consensus state to remove this entirely

hamdiallam:
Squash + Rebasing this since the conflicts are more complicated to merge

PR #3573, opened 2026-04-14
docs(tip-1022): address SIGP-228 and SIGP-229 audit findings

decofe:
ref: `SIGP-228`, `SIGP-229`

Clarifies that the zero-balance invariant (invariant 4) applies from T3 activation onwards — pre-T3 transfers to virtual-format addresses credit the literal address since resolution isn't active yet, and those balances are permanently stranded.

Documents the externally-triggerable `VirtualAddressUnregistered` revert as a new recipient-dependent failure mode, with integrator guidance for batch transfer contracts.

Prompted by: rusowsky


0xrusowsky (approved):

malleshpai (approved):
lgtm

PR #3574, opened 2026-04-14
fix(precompiles): `Set<T>` overflow protection

0xrusowsky:
ref: `SIGP-231`


crates/precompiles/src/storage/types/set.rs
```diff
@@ -229,6 +229,18 @@ where
     }
 }
 
+/// Converts a 0-based index to a 1-based position for storage.
+///
+/// Returns an error if the result would overflow `u32`, which would corrupt the
+/// sentinel value (`0` means "not present") used by `contains()` and `remove()`.
+#[inline]
+fn checked_position(index: usize) -> Result<u32> {
```
legion2002, line 237: the fn name could be more clear

legion2002 (approved):

PR #3579, opened 2026-04-14
fix(revm): charge T3 key authorization scope gas

legion2002:
Refs SIGP-234

Charge the remaining T3 key authorization scope housekeeping gas in `calculate_key_authorization_gas`, including unrestricted and deny-all scope setup plus fresh set insert/delete costs.

This aligns intrinsic gas with the storage work performed by `replace_allowed_calls` for new keys and closes the remaining gap left after the earlier row-counting change.


legion2002:
Closed in favour of #3595

PR #3581, opened 2026-04-14
fix(account-keychain): align call restriction polarity

legion2002:
Refs SIGP-237

Renames `allowAnyCalls` to `enforceCallRestrictions` and flips its meaning so the keychain restriction booleans now use the same positive polarity.

Updates the precompile ABI, handler translation, SDK encoding, TIP text, and reference-interface/tests together so the breaking interface change stays consistent everywhere.


0xrusowsky (commented):
this one will be painful cause it will require a new alloy release

let's hold on merging until we have `tempo-foundry` fully upstreamed and we have migrated to use `foundry` in our CI (should happen today/tomorrow)

legion2002:
Yeah maybe we don't do this change, seems painful for all tooling, and it isn't a big deal.

legion2002:
Closing this PR, because it's not a big deal, and it will cause breaking changes to a lot of tooling at this point.

PR #3586, opened 2026-04-14
fix(dex): check paused state on internal balance debit paths

decofe:
Adds `check_not_paused()` in `decrement_balance_or_transfer_from` and `place_flip`'s internal-balance-only path when `user_balance >= amount` on T4+, since TIP-20 `transferFrom` won't run in that case. A `check_pause` bool avoids redundant SLOADs on the swap path (already covered by `validate_and_build_route`).

Prompted by: daniel

### DIFF
--- a/crates/precompiles/src/stablecoin_dex/mod.rs
+++ b/crates/precompiles/src/stablecoin_dex/mod.rs
@@ -5511,6 +5511,133 @@ mod tests {
     }
 
     #[test]
+    fn test_place_orders_on_paused_token_respects_internal_balance_path() -> eyre::Result<()> {
+        fn assert_paused_token_order_path<F>(
+            internal_balance_amount: u128,
+            is_bid: bool,
+            mut place_order: F,
+        ) -> eyre::Result<()>
+        where
+            F: FnMut(&mut StablecoinDEX, Address, Address, u128) -> Result<u128>,
+        {
+            for spec in [TempoHardfork::T2, TempoHardfork::T3] {
+                let mut storage = HashMapStorageProvider::new_with_spec(1, spec);
+                StorageCtx::enter(&mut storage, || {
+                    let mut exchange = StablecoinDEX::new();
+                    exchange.initialize()?;
+
+                    let (alice, admin) = (Address::random(), Address::random());
+                    let amount = MIN_ORDER_AMOUNT;
+
+                    let (base_token, quote_token) =
+                        setup_test_tokens(admin, alice, exchange.address, 500_000_000u128)?;
+                    exchange.create_pair(base_token)?;
+                    let escrow_token = if is_bid { quote_token } else { base_token };
+                    exchange.set_balance(alice, escrow_token, internal_balance_amount)?;
+
+                    let mut escrow_tip20 = TIP20Token::from_address(escrow_token)?;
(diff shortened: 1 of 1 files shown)


0xrusowsky (approved):

0xalpharush:
place / place_flip can still create resting orders when the paused token is only the non-escrow side of the pair. Is that intended because execution/fill fails via `validate_and_build_route`, or should placement reject any paused base/quote token on T4?

fgimenez:
> place / place_flip can still create resting orders when the paused token is only the non-escrow side of the pair. Is that intended because execution/fill fails via `validate_and_build_route`, or should placement reject any paused base/quote token on T4?

@0xalpharush good call, should reject either side on T4. `validate_and_build_route` only catches swap fills, and dead resting orders pollute the book until cancel/unpause

added `check_not_paused()` on the non-escrow token in `place`/`place_flip` (T4+) and a `test_place_orders_on_paused_non_escrow_token_blocked_on_t4` covering both order types x ask/bid x internal-only and transferFrom paths, ptal

fgimenez (approved):

crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -1408,9 +1441,9 @@ impl StablecoinDEX {
             let (base, quote) = {
                 let token_in_tip20 = TIP20Token::from_address(token_in)?;
 
-                // Ensure that the token is not paused (spec: T3+)
+                // Ensure that the token is not paused (spec: T4+)
                 // Necessary because TIP20 transfer checks don't cover internal DEX balance updates
-                if self.storage.spec().is_t3() {
+                if self.storage.spec().is_t4() {
```
0xKitsune: This is already active on T3, if we update this then this will break consensus for nodes with T3 active.

crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -5151,8 +5184,227 @@ mod tests {
...
+        ) -> eyre::Result<()>
+        where
+            F: FnMut(&mut StablecoinDEX, Address, Address, u128) -> Result<u128>,
+        {
+            for spec in [TempoHardfork::T3, TempoHardfork::T4] {
+                let mut storage = HashMapStorageProvider::new_with_spec(1, spec);
+                StorageCtx::enter(&mut storage, || {
+                    let mut exchange = StablecoinDEX::new();
+                    exchange.initialize()?;
+
```
0xKitsune: Do we need both `assert_paused_non_escrow_token_order` and `assert_paused_token_order_path` or can we consolidate.

crates/precompiles/src/stablecoin_dex/mod.rs
fgimenez: good catch, fixed here https://github.com/tempoxyz/tempo/pull/3586/commits/e5c1fbe046756b6f56cfc95471b5abb69254f188

crates/precompiles/src/stablecoin_dex/mod.rs
fgimenez: yep, consolidated here https://github.com/tempoxyz/tempo/pull/3586/commits/62a2acf84ad4c0ecaa496343bea85c566d9bc286

PR #3588, opened 2026-04-14
nit: small tip-1031 followups

hamdiallam:


crates/commonware-node/src/consensus/block.rs
```diff
@@ -152,8 +152,7 @@ impl commonware_consensus::CertifiableBlock for Block {
             },
             None => {
                 // Pre-T4: Unused deterministic sentinel.
-                // Post-T4: Panic, all blocks where this is called must have a context.
-                //          Historical Pre-T4 blocks are never replayed in consensus
+                // Post-T4: This dummy context will fail verification with consensus values
```
SuperFluffy, line 155: Let's unit test that.

SuperFluffy (approved):

crates/commonware-node/src/consensus/block.rs
hamdiallam: Would require a bit of a refactor in our e2e setup. Would need 1 validator to have t4=0 while the others are activated so that the node proposals would fail

i'll look into this

PR #3591, opened 2026-04-14
perf: make sure prewarming prewarms correct expiring nonce slots

klkvr:
RIght now if there are many expiring nonce transactions in the block prewarming won't work well because the slots accessed by expiring nonce code depend on number of expiring nonce transactions that were executed before the current one, and this number is different in prewarming vs execution.

This PR changes logic to make transactions carry `expiring_nonce_idx` which is a number of expiring transactions before that one in the block. This value is only used in prewarming (enforced by unsetting it in block executor), and changes the expiring nonce contract state to match what it'd acutally look like during this transaction execution


crates/evm/src/engine.rs
```diff
@@ -39,9 +39,18 @@ impl ConfigureEngineEvm<TempoExecutionData> for TempoEvmConfig {
...
     ) -> Result<impl ExecutableTxIterator<Self>, Self::Error> {
         let block = payload.block.clone();
-        let transactions: Vec<_> = (0..payload.block.body().transactions.len())
-            .map(move |i| (block.clone(), i))
-            .collect();
+        let mut transactions = Vec::with_capacity(payload.block.body().transactions.len());
+        let mut expiring_nonce_idx = 0;
+
+        for idx in 0..payload.block.body().transactions.len() {
+            let tx = &payload.block.body().transactions[idx];
```
shekhirin:
```suggestion
        for (idx, tx) in 0..payload.block.body().transactions.iter().enumerate() {
```

shekhirin (approved):

0xrusowsky (approved):

PR #3595, opened 2026-04-15
fix(revm): round key auth scope gas surcharge

legion2002:
`SIGP 234`

T3 key authorizations were still undercharging intrinsic gas for the call-scope helper path. The existing intrinsic calculation already charged the dynamic storage rows for the persisted scope tree, but it did not account for the surrounding bookkeeping in `replace_allowed_calls`: clearing the empty tree for a fresh key, maintaining target and selector sets, and writing recipient sets.

This patch keeps the fix intentionally small. Instead of trying to exactly mirror every storage helper operation, it adds a rounded surcharge on top of the existing scope-slot charge:
- 5,000 gas base for the always-run scope bookkeeping
- 7,000 gas per target
- 7,000 gas per selector
- 5,000 gas per recipient

That makes unrestricted and small scoped authorizations materially less underpriced, while also scaling with larger scope trees so the gap does not stay flat as scopes grow. The code comment in `call_scope_extra_gas` explains what each part of the surcharge is covering, and it includes a TODO to fold this and the other intrinsic surcharges into a shared intrinsic-gas accounting model in the future.

The regression test in `crates/revm/src/handler.rs` now checks both the unrestricted T3 case and a scoped authorization case against the rounded surcharge formula.

Tested with:
- `cargo test -p tempo-revm test_key_authorization_gas_with_limits --lib`
- `cargo fmt --all --check`


0xrusowsky (approved):
makes sense as a quick temp solution that should be refined down the line

PR #3597, opened 2026-04-15
ci: add release attestation verification workflow

decofe:
Runs `gh release verify` and `gh release verify-asset` on every published release to confirm the immutable release attestation and all assets are valid. Fires on `release: [published]` so it runs exactly once when a draft is published.

Prompted by: georgen

### DIFF
--- a/.github/workflows/verify-release.yml
+++ b/.github/workflows/verify-release.yml
@@ -1,0 +1,31 @@
+name: Verify Release
+
+on:
+  release:
+    types: [published]
+
+permissions: {}
+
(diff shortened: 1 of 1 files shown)


.github/workflows/verify-release.yml
```diff
@@ -0,0 +1,31 @@
+name: Verify Release
```
jenpaff, line 1: i think it should be step of existing workflow https://github.com/tempoxyz/tempo/blob/main/.github/workflows/release.yml

PR #3606, opened 2026-04-15
chore(chainspec): isolate reth deps

0xrusowsky:
this PR prepares `tempo-chainspec` for sanitization (previous to publishing) to strip the code that deps on `reth`

### DIFF
--- a/crates/chainspec/src/spec.rs
+++ b/crates/chainspec/src/spec.rs
@@ -1,16 +1,12 @@
+pub use crate::constants::gas::*;
+
 use crate::{
     bootnodes::{moderato_nodes, presto_nodes},
     hardfork::{TempoHardfork, TempoHardforks},
 };
 use alloc::{boxed::Box, sync::Arc, vec::Vec};
 use alloy_eips::eip7840::BlobParams;
-use alloy_evm::{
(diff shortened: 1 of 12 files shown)


crates/chainspec/src/spec.rs
```diff
@@ -26,49 +22,10 @@ use reth_network_peers::NodeRecord;
 use std::sync::LazyLock;
 use tempo_primitives::TempoHeader;
 
-/// T0 base fee: 10 billion attodollars (1×10^10)
-///
-/// Attodollars are the atomic gas accounting units at 10^-18 USD precision.
-/// Basefee is denominated in attodollars.
-pub const TEMPO_T0_BASE_FEE: u64 = 10_000_000_000;
```
mattsse, line 33: why move this?

mattsse (approved):

crates/chainspec/src/spec.rs
0xrusowsky: cause `spec.rs` needs to be gated under `reth` cfg, and `hardfork.rs` (which is not) uses the consts too.

by moving them to `constants.rs` (which is not gated), we can use them everywhere.

PR #3611, opened 2026-04-15
chore: tempo consensus cli wallet options

hamdiallam:
--ledger
--trezor
--aws
--gcp

When https://github.com/foundry-rs/foundry/issues/13800 is addressed, we can simply rely on foundry rather than re-doing it here

### DIFF
--- a/bin/tempo/src/tempo_cmd.rs
+++ b/bin/tempo/src/tempo_cmd.rs
@@ -293,11 +293,14 @@ async fn read_validator_from_contract(
 /// Shared validator identity arguments used across add/rotate/sign commands.
 #[derive(Debug, clap::Args)]
 pub(crate) struct ValidatorIdentityArgs {
-    /// The validator's address
+    /// The validator's Ethereum address
     #[arg(long, value_name = "ETHEREUM_ADDRESS")]
     validator_address: Address,
-    /// The identity key of the validator (0x-prefixed hex).
-    #[arg(long = "consensus.public-key", value_name = "IDENTITY_KEY")]
+    /// The validator's signing key address (0x-prefixed hex).
+    #[arg(
+        long = "consensus.public-key",
+        value_name = "IDENTITY_PUBLIC_KEY_ADDRESS"
+    )]
     public_key: B256,
     /// The inbound address for the validator.
     #[arg(long, value_name = "IP:PORT")]
@@ -327,7 +330,7 @@ pub(crate) struct ValidatorSignatureArgs {
     #[arg(long, value_name = "SIGNATURE")]
     signature: Option<Bytes>,
 
-    /// Path to the ed25519 signing key file. The signature is computed
+    /// Path to the ed25519 signing private key file. The signature is computed
     /// automatically so a separate `create-*-signature` step is not needed.
     #[arg(long = "consensus.signing-key", value_name = "FILE")]
     signing_key: Option<PathBuf>,
@@ -354,7 +357,7 @@ impl ValidatorSignatureArgs {
 #[derive(Debug, clap::Args)]
 #[group(required = true, multiple = false)]
 pub(crate) struct WalletArgs {
-    /// Path to the file holding the Ethereum private key.
+    /// Path to the file holding the validator's Ethereum private key.
     #[arg(long, value_name = "FILE", help_heading = "Wallet options - raw")]
     wallet_key: Option<PathBuf>,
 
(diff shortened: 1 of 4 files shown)


bin/tempo/src/tempo_cmd.rs
```diff
@@ -40,6 +45,16 @@ use tempo_validator_config::ValidatorConfig;
...
 
+fn get_env(key: &str) -> eyre::Result<String> {
+    match std::env::var(key) {
+        Ok(val) => Ok(val),
+        Err(std::env::VarError::NotPresent) => bail!("environment variable {key:?} not found"),
+        Err(std::env::VarError::NotUnicode(s)) => {
+            bail!("environment variable {key:?} was not valid unicode: {s:?}")
+        }
+    }
+}
```
SuperFluffy, line 51: Don't manually type these out. When `eyre::Report` gets returned you get that for free. So to get the benefit of telling the user which env failed, just:

```suggestion
fn get_env(key: &str) -> eyre::Result<String> {
    std::env::var(key).wrap_err_with(|| format!("failed reading environment variable `{key}`))
}
```

bin/tempo/src/tempo_cmd.rs
```diff
@@ -341,12 +356,36 @@ impl ValidatorSignatureArgs {
     }
 }
 
+#[derive(Debug, clap::Args)]
+#[group(required = true, multiple = false)]
+pub(crate) struct WalletArgs {
+    /// Path to the file holding the Ethereum private key.
+    #[arg(long, value_name = "FILE", help_heading = "Wallet options - raw")]
+    private_key: Option<PathBuf>,
```
SuperFluffy: Maybe so that we don't have to explain that `--private-key` (or `--consensus.private-key`) is totally different from `--consensus.signing-key`?
```suggestion
    wallet_key: Option<PathBuf>,
```

bin/tempo/src/tempo_cmd.rs
```diff
@@ -385,24 +424,72 @@ impl ValidatorTransactionArgs {
...
+            .await
+            .wrap_err("failed to create GCP KMS client")?;
+
+            let signer = GcpSigner::new(client, specifier, None)
+                .await
+                .wrap_err("failed to create GCP KMS signer")?;
 
-        Ok(PrivateKeySigner::from_bytes(&private_key)?)
+            Ok(EthereumWallet::new(signer))
+        } else if let Some(ref path) = self.wallet.private_key {
```
SuperFluffy: agents...
```suggestion
        } else if let Some(path) = &self.wallet.private_key {
```

bin/tempo/src/tempo_cmd.rs
```diff
@@ -385,24 +424,72 @@ impl ValidatorTransactionArgs {
...
+            let private_key_bytes =
+                std::fs::read(path).wrap_err("failed reading validator ethereum private key")?;
+            let private_key = B256::try_from(private_key_bytes.as_slice())
+                .wrap_err("invalid validator ethereum private key")?;
+            let signer = PrivateKeySigner::from_bytes(&private_key)?;
```
SuperFluffy: There is `PrivateKeySigner::from_slice` so you don't need to do this manually:
```suggestion
            let signer = PrivateKeySigner::from_slice(&private_key_bytes).wrap_err("failed decoding file contents to private key")?;
```

bin/tempo/src/tempo_cmd.rs
```diff
@@ -385,24 +424,72 @@ impl ValidatorTransactionArgs {
...
+
+            let signer = GcpSigner::new(client, specifier, None)
+                .await
+                .wrap_err("failed to create GCP KMS signer")?;
 
-        Ok(PrivateKeySigner::from_bytes(&private_key)?)
+            Ok(EthereumWallet::new(signer))
+        } else if let Some(ref path) = self.wallet.private_key {
+            let private_key_bytes =
+                std::fs::read(path).wrap_err("failed reading validator ethereum private key")?;
```
SuperFluffy: Report which file we failed to read (here and when constructing the key - nice to know what failed).

bin/tempo/src/tempo_cmd.rs
```diff
@@ -385,24 +424,72 @@ impl ValidatorTransactionArgs {
...
+            let private_key = B256::try_from(private_key_bytes.as_slice())
```
SuperFluffy: After reading these, let's stuff them into a `Zeroizizing` immediately: https://docs.rs/zeroize/latest/zeroize/struct.Zeroizing.html

SuperFluffy (approved):
Neat, just a bunch of smaller suggestions.

bin/tempo/src/tempo_cmd.rs
hamdiallam: gah amp

PR #3615, opened 2026-04-16
fix(consensus): accept hex-encoded private keys in CLI commands

kuyziss:


bin/tempo/src/tempo_cmd.rs
```diff
@@ -422,12 +422,23 @@ impl ValidatorTransactionArgs {
...
                 format!("failed reading private key from `{}`", path.display())
             })?);
 
-            let signer = PrivateKeySigner::from_slice(private_key_bytes.as_slice())
-                .wrap_err_with(|| format!("invalid private key in `{}`", path.display()))?;
+            // Support both raw 32-byte keys and hex-encoded strings (with optional 0x prefix)
+            let private_key = if raw.len() == 32 {
+                B256::from_slice(&raw)
+            } else {
+                let hex_str = std::str::from_utf8(&raw)
```
SuperFluffy: Use `const_hex` for this to handle both `0x` and not-`0x` prefixed

kuyziss: @SuperFluffy good now?

SuperFluffy (approved):

PR #3622, opened 2026-04-16
refactor(precompiles): use GasTracker in EvmPrecompileStorageProvider

rakita:
## Summary
- Replace individual `gas_remaining`, `gas_refunded`, `gas_limit`, and `reservoir` fields with revm's `GasTracker` struct in `EvmPrecompileStorageProvider`
- Delegate all gas accounting to `GasTracker` methods (`record_regular_cost`, `record_refund`, `limit`, `remaining`, `refunded`, `reservoir`)
- Remove the standalone `deduct_gas` helper function that is no longer needed

## Test plan
- [x] All 788 existing tests pass (`cargo test -p tempo-precompiles`)
- [x] `cargo check` passes with no warnings

### DIFF
--- a/crates/precompiles/src/lib.rs
+++ b/crates/precompiles/src/lib.rs
@@ -953,10 +953,10 @@ mod tests {
         let output = AlloyEvmPrecompile::call(&precompile, input).expect("transfer should succeed");
         assert!(output.is_success(), "transfer should be successful");
 
-        // T4+: gas refund must be encoded in the reservoir field
+        // T4+: gas refund must be encoded in the gas_refunded field
         assert!(
-            output.reservoir > 0,
-            "T4+ successful precompile with SSTORE refund must encode refund in reservoir, got 0"
+            output.gas_refunded != 0,
+            "T4+ successful precompile with SSTORE refund must encode refund in gas_refunded, got 0"
         );
     }
 
--- a/crates/precompiles/src/lib.rs
+++ b/crates/precompiles/src/lib.rs
@@ -436,6 +436,7 @@ pub(crate) fn dispatch_call<T>(
 
     match result {
         Ok(call) => f(call).map(|mut res| {
+            // TODO: fix this, each precompile handler should either return output with proper gas values or don't return any gas values at all.
             res.gas_used = storage.gas_used();
             fill_state_gas(&mut res, &storage);
             res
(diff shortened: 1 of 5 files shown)


crates/precompiles/src/storage/evm.rs
```diff
@@ -127,14 +122,16 @@ impl<'a> PrecompileStorageProvider for EvmPrecompileStorageProvider<'a> {
...
-        deduct_gas(
-            &mut self.gas_remaining,
-            self.gas_params.warm_storage_read_cost(),
-        )?;
+        if !self
+            .gas_tracker
+            .record_regular_cost(self.gas_params.warm_storage_read_cost())
+        {
+            return Err(TempoPrecompileError::OutOfGas);
+        }
```
klkvr: should this just return a `Result` like `deduct_state_gas` does?

klkvr: ah actually should this be just `self.deduct_gas`?

klkvr: ok that's not feasible because the account is borrowed, then we should stick to `deduct_gas` helper imo

crates/precompiles/src/lib.rs
```diff
@@ -331,12 +331,20 @@ pub(crate) fn charge_input_cost(
...
+            // T4+: propagate SSTORE refunds via reservoir so the TempoPrecompileProvider
+            // wrapper can apply them with record_refund. Pre-T4 blocks were executed
+            // without refund propagation, so we cannot change their gas accounting.
             output.gas_refunded = storage.gas_refunded();
+        } else {
+            // On revert or halt: state changes are undone, so ALL state gas returns
+            // to the parent's reservoir.
+            output.reservoir = storage.state_gas_used() + storage.reservoir();
+            output.state_gas_used = 0;
         }
```
klkvr, line 348: so looking at this logic there's no really reason for returning `reservoir` field?

basically the logic is 
```
if output.is_success() {
     reservoir = previous_reservoir.saturating_sub(state_gas_used);
} else {
     reservoir = previous_reservoir
}
```

am I missing something?

klkvr (approved):

crates/precompiles/src/lib.rs
rakita: `in else {` we need `previous_reservoir + state_part_of_regular_gas`

crates/precompiles/src/storage/evm.rs
rakita: Yeah it is borrow checked problem

PR #3628, opened 2026-04-16
fix(txpool): avoid duplicate invalidation evictions

### LINKED ISSUE #3627: txpool invalidation can double-count a transaction when blacklist and whitelist updates both match
### Describe the bug

Affected crate: tempo-transaction-pool (`crates/transaction-pool/src/tempo_pool.rs`).

`evict_invalidated_transactions()` can push the same tx hash twice when one pending AA transaction matches both Check 4 (blacklist addition) and Check 5 (whitelist removal) in the same maintenance scan.

On current HEAD, Checks 1-3 `continue` after pushing to `to_remove`, but Checks 4-5 do not. The same hash can therefore be returned twice, while `crates/transaction-pool/src/maintain.rs` increments `metrics.transactions_invalidated` with `evicted.len()`. This can inflate the Prometheus metric and txpool debug counters, and `state.untrack(hash)` can run twice for the same tx.

### Steps to reproduce

1. Add a pending AA transaction with an explicit fee token whose transfer policy is compound.
2. In the same maintenance scan, include:
 - a blacklist addition for the fee payer on the sender-side policy
 - a whitelist removal for `TIP_FEE_MANAGER_ADDRESS` on the recipient-side policy
3. Run `evict_invalidated_transactions()`.
4. 4. The same tx hash can be pushed once in Check 4 and again in Check 5 because neither branch exits the outer loop after the first invalidation.

Confirmed by code inspection on current HEAD. I also verified a local fix with a targeted regression test.

### Logs

```text

```

### Platform(s)

_No response_

### Container Type

Not running in a container

Current HEAD on main: `d5b27eaa`

_No response_

### Code of Conduct

RekCuy63:
Closes #3627

Tracks each invalidated transaction at most once when blacklist and whitelist updates both match in the same maintenance scan.

Adds a regression test covering compound sender/recipient policy updates so eviction metrics and follow-up bookkeeping stay aligned with the actual number of removed transactions.

### DIFF
--- a/crates/transaction-pool/src/tempo_pool.rs
+++ b/crates/transaction-pool/src/tempo_pool.rs
@@ -349,6 +349,7 @@ where
                 if sender_evicted || recipient_evicted {
                     to_remove.push(*tx.hash());
                     blacklisted_count += 1;
+                    continue;
                 }
             }
 
@@ -394,6 +395,7 @@ where
                 if sender_evicted || recipient_evicted {
                     to_remove.push(*tx.hash());
                     unwhitelisted_count += 1;
+                    continue;
                 }
             }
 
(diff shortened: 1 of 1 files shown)


mattsse (commented):
dont think this is impactful/matters

PR #3637, opened 2026-04-16
feat(primitives): add `TempoAddressExt` with built-in helper fns

0xrusowsky:
this PR encapsulates all the helper fns related to the `Address` into a single `trait TempoAddressExt`, so that consumers of `tempo-primitives` can easily use all those helpers.

review note:
- the trait is defined in `crates/primitives/src/address.rs`
- the other changes simply make use of the new ergonomic APIs

### DIFF
--- a/crates/precompiles/src/tip20/mod.rs
+++ b/crates/precompiles/src/tip20/mod.rs
@@ -37,6 +37,7 @@ use alloy::{
 use std::sync::LazyLock;
 use tempo_precompiles_macros::contract;
 use tempo_primitives::TempoAddressExt;
+pub use tempo_primitives::is_tip20_prefix;
 use tracing::trace;
 
 /// u128::MAX as U256
--- a/crates/precompiles/src/tip20/mod.rs
+++ b/crates/precompiles/src/tip20/mod.rs
@@ -31,30 +31,18 @@ use crate::{
     tip403_registry::{AuthRole, ITIP403Registry, TIP403Registry},
 };
 use alloy::{
-    hex,
     primitives::{Address, B256, U256, keccak256, uint},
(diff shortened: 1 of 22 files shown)


crates/precompiles/src/tip20/mod.rs
```diff
@@ -31,31 +31,19 @@ use crate::{
...
-/// The full address is: TIP20_TOKEN_PREFIX (12 bytes) || derived_bytes (8 bytes)
-const TIP20_TOKEN_PREFIX: [u8; 12] = hex!("20C000000000000000000000");
-
-/// Returns true if the address has the TIP20 prefix.
-///
-/// NOTE: This only checks the prefix, not whether the token was actually created.
-/// Use `TIP20Factory::is_tip20()` for full validation.
-pub fn is_tip20_prefix(token: Address) -> bool {
-    token.as_slice().starts_with(&TIP20_TOKEN_PREFIX)
-}
```
mattsse, line 57: removing this here is technically breaking

crates/primitives/src/address.rs
mattsse, line 1: I'd prefer if we keep the fns as is and then reuse the standalone fn in the trait impl

mattsse (commented):
style suggestion

crates/primitives/src/address.rs
0xrusowsky: restored here, but defined in `tempo-primitives` due to deps [`5665c46` (this PR)](https://github.com/tempoxyz/tempo/pull/3637/commits/5665c465c55de50c9d3405f1c443883b9152791e)

mattsse (approved):
lgtm

PR #3645, opened 2026-04-17
tip: admin access keys

jxom:
Adds TIP-1049: Admin Access Keys.

Extends access keys with an `admin` flag that grants key-management and TIP-1020 signature verification capabilities. Admin keys can authorize/revoke other keys and participate in onchain signature verification.


legion2002 (commented):
:eye: **Cyclops Review** — `fb93bd48327c`

TIP-1049 introduces admin access keys — a single-flag extension to `AuthorizedKey` that grants key-management and TIP-1020 signature verification capabilities. The spec contains **3 high-severity security defects** and **5 medium/low issues** that must be resolved before advancing past Draft.

**Critical issues:**
1. V1 keychain signatures (`0x03`) are admitted in the TIP-1020 path, re-enabling the CAN-1/CAN-2 cross-account replay attack fixed at T1C.
2. The shared `ensure_admin_caller` gate protects 5 mutators (not just authorize/revoke), so relaxing it gives admin keys unintended write access to sibling key spending limits and call scopes.
3. The spec simultaneously requires root-signed `KeyAuthorization` and claims rootless device onboarding — a self-contradictory invariant.

**Additional issues:** The `admin?` RLP trailing field is not backward-compatible under the current ordered-optional layout; hardfork gating is absent; `recover()`/`verify()` return semantics are ambiguous; signature-type binding is omitted from the admin gate; and gas costs for the new stateful TIP-1020 path are unspecified.

<details>

- ⚡ **Signing-domain overlap**: V2 effective hash `keccak256(0x04 || hash || user_address)` is shared between tx auth and proposed TIP-1020 admin verification with no purpose separator. Consider a distinct domain byte (e.g., `0x04 || 0x20 || hash || user_address`).
- ⚡ **Pool validator**: `AccessKeyCannotAuthorizeOtherKeys` rejection becomes state-dependent for admin keys; pool needs a state read for `is_admin`.
- ⚡ **ABI selector & event-topic migration**: Adding fields to `KeyRestrictions` and `KeyAuthorized` changes selectors/topics; needs explicit migration schedule (cf. T3 `legacyAuthorizeKey` pattern).
- ⚡ **Inline authorize ordering**: `transaction_key` is seeded after inline `KeyAuthorization` is processed; admin-key authorization needs an explicit handler-side lookup.

</details>

tips/tip-1049.md
```diff
@@ -0,0 +1,203 @@
...
+Byte 11:    is_admin (bool)        ← NEW
+```
+
+This adds one byte to the existing packed slot. No new storage slots are introduced for the admin flag.
+
+### Root key revocation slot
+
+Root-key revocation reuses the existing `keys[account][keyId]` mapping by treating `keyId == account` (the EOA address itself) as the canonical slot for the root key's `AuthorizedKey`. When the root key is revoked, this slot is written with `is_revoked = true` and all other fields cleared, identical to how an access key is marked revoked. No additional storage layout is introduced.
+
+The slot at `keys[account][account]` is reserved for root-key state under this TIP and MUST NOT be used as a regular access-key slot. Authorization of an access key with `keyId == account` MUST be rejected with `InvalidKeyId`.
```
0xKitsune: Pre TIP-1049, nothing rejects `keyId == account` at authorization, so any legacy entry at this slot would be silently reinterpreted as a root-revocation marker at activation, and a later access key `revokeKey` against it would brick the account via `RootKeyRevoked`. If this is expected behavior, the spec should define behavior for pre fork entries.

tips/tip-1049.md
```diff
@@ -0,0 +1,203 @@
...
+
+Revocation of an admin key does NOT automatically revoke keys that it authorized. Each key is independently tracked; cascading revocation is out of scope for this TIP. Accounts that need cascading revocation should revoke downstream keys explicitly.
+
+## Transaction Encoding
+
+The `KeyAuthorization` RLP encoding is bumped to V2 to carry the `admin` field unambiguously. The version is indicated by a new leading byte:
+
+```
+V1 (existing): [chain_id, key_type, key_id, expiry?, limits?, allowed_calls?]
+V2 (this TIP):  [0x02, chain_id, key_type, key_id, expiry?, limits?, allowed_calls?, admin]
```
0xrusowsky: should we consider moving the tag outside the list, mirroring [EIP-2718](https://eips.ethereum.org/EIPS/eip-2718)?

```
V1: rlp([...])                  // unchanged (legacy, like pre-2718 txs)
V2: 0x02 || rlp([..., admin])   // typed envelope
```

dispatch becomes a one-byte peek (≥ 0xc0 → V1, 0x02 → V2), disjoint by RLP construction — same structural argument EIP-2718 relies on for legacy vs typed transactions.

the benefit is that it disambiguates V2 regardless of the chain_id, whereas i believe single-byte chain ids may cause issues with the current proposal

tips/tip-1049.md
```diff
@@ -0,0 +1,203 @@
...
+/// `keyId` and then checks `isAdminKey(account, keyId)`. Returns true on
+/// success, false on signature failure or non-admin key.
+function verifySignature(
+    address account,
+    bytes32 digest,
+    bytes calldata signature
+) external view returns (bool);
+```
+
+`verifySignature` is the canonical entry point for ERC-1271-style contracts that want to accept admin access keys. It is a thin composition of TIP-1020 verification with an admin-status check; TIP-1020 itself is not modified.
```
0xrusowsky: personally i'm not a huge fan of this suggestion from a devex perspective, as we are telling devs to verify sigs via 2 different precompiles. Instead, i'd propose to extend `SignatureVerifier` with stateful checks that read `AccountKeychain`'s storage:
```solidity
ISignatureVerifier {
   // EXISTING(stateless): pure crypto
   function verify(address signer, bytes32 digest, bytes calldata sig)
       external pure returns (bool);

   // NEW(stateful): true iff signer is a non-revoked, non-expired authorized key of `account`
   function verifyAccount(address account, bytes32 digest, bytes calldata sig)
      external view returns (bool);

   // NEW(stateful): like `verifyAccount` + is_admin
   function verifyAdminAccount(address account, bytes32 digest, bytes calldata sig)
      external view returns (bool);
}
```
imo this is a more comprehensive story

tips/tip-1049.md
```diff
@@ -0,0 +1,203 @@
...
+
+2. **Onchain account ↔ key association.** Contracts that want to gate actions on "this key belongs to this account" have no canonical way to resolve a public key back to the account that authorized it, or to check whether the key has admin capabilities.
+
+Adding an `admin` flag plus a small set of read helpers in the `AccountKeychain` precompile solves both problems without entangling TIP-1020. TIP-1020 remains a pure signature-verification primitive; contracts that want admin-key semantics call the keychain explicitly.
+
+## Assumptions
+
+- The root EOA key is implicitly admin until it is explicitly revoked via `revokeKey`. The `admin` flag only applies to access keys registered in the AccountKeychain precompile.
+- Admin keys MAY be authorized with spending limits, call scopes, and expiry just like non-admin keys. However, callers should be aware that an admin key can effectively bypass its own spending-limit and call-scope restrictions by authorizing a new, unrestricted key. Expiry remains a meaningful constraint because the admin key cannot extend its own expiry.
```
legion2002: This feels like a big footgun. An admin key with a spending limit doesn't make natural sense. Is there a use case for this?

tips/tip-1049.md
```diff
@@ -0,0 +1,203 @@
...
+- An admin key cannot grant the `admin` flag to keys it authorizes unless the admin key itself was authorized with that capability. This prevents privilege escalation: only the root key (or another admin key) can create admin keys.
```
legion2002: Do we want admin keys to be exactly the same as root keys? We can probably remove this restriction.

tips/tip-1049.md
legion2002: agree with this. We should probably make the names more descriptive so people know these are stateful.
`verifyAccountStateful`?

tips/tip-1049.md
0xrusowsky: +1, feels counter-intuitive that it is limited, but can enable new keys to effectively bypass the limits

tips/tip-1049.md
```diff
@@ -25,10 +25,9 @@ Adding an `admin` flag plus a small set of read helpers in the `AccountKeychain`
 
 ## Assumptions
 
-- The root EOA key is implicitly admin until it is explicitly revoked via `revokeKey`. The `admin` flag only applies to access keys registered in the AccountKeychain precompile.
+- The root EOA key is implicitly admin. The `admin` flag only applies to access keys registered in the AccountKeychain precompile.
 - Admin keys MAY be authorized with spending limits, call scopes, and expiry just like non-admin keys. However, callers should be aware that an admin key can effectively bypass its own spending-limit and call-scope restrictions by authorizing a new, unrestricted key. Expiry remains a meaningful constraint because the admin key cannot extend its own expiry.
 - An admin key cannot grant the `admin` flag to keys it authorizes unless the admin key itself was authorized with that capability. This prevents privilege escalation: only the root key (or another admin key) can create admin keys.
```
samczsun: This sentence is a little confusing - don't admin keys implicitly have this capability? There's no distinction between "is admin" and "can create new admin" right now

jxom: ah yep

tips/tip-1049.md
```diff
@@ -0,0 +1,180 @@
...
+
+2. **Onchain account ↔ key association.** Contracts that want to gate actions on "this key belongs to this account" have no canonical way to resolve a public key back to the account that authorized it, or to check whether the key has admin capabilities.
+
+Adding an `admin` flag plus a small set of read helpers in the `AccountKeychain` precompile solves both problems without entangling TIP-1020. TIP-1020 remains a pure signature-verification primitive; contracts that want admin-key semantics call the keychain explicitly.
+
+## Assumptions
+
+- The root EOA key is implicitly admin. The `admin` flag only applies to access keys registered in the AccountKeychain precompile.
+- Admin keys MAY be authorized with spending limits, call scopes, and expiry just like non-admin keys. However, callers should be aware that an admin key can effectively bypass its own spending-limit and call-scope restrictions by authorizing a new, unrestricted key. Expiry remains a meaningful constraint because the admin key cannot extend its own expiry.
```
samczsun: Can an admin key not create a new key with no expiry and circumvent that too?

tips/tip-1049.md
```diff
@@ -0,0 +1,180 @@
...
+    uint64 expiry;
+    bool enforceLimits;
+    bool isRevoked;
+    bool isAdmin;             // NEW
+}
+```
+
+### `KeyAuthorized` Event
+
+Add `isAdmin` to the `KeyAuthorized` event:
```
samczsun: This is not backwards compatible

tips/tip-1049.md
```diff
@@ -0,0 +1,180 @@
...
+Byte 11:    is_admin (bool)        ← NEW
+```
+
+This adds one byte to the existing packed slot. No new storage slots are introduced for the admin flag.
+
+## Interface Changes
+
+### `KeyRestrictions`
+
+Add an `admin` field to the `KeyRestrictions` struct:
```
samczsun: This is not backwards compatible

tips/tip-1049.md
```diff
@@ -0,0 +1,180 @@
...
+Three read-only helpers are added to the AccountKeychain precompile so external contracts can reason about key ↔ account associations and verify admin-key signatures:
+
+```solidity
+/// Returns true if `keyId` is currently an authorized, non-revoked,
+/// non-expired key on `account` with `is_admin = true`.
+function isAdminKey(address account, address keyId) external view returns (bool);
+
+/// Returns the admin status of `keyId` for `account`, plus its full KeyInfo.
+/// Reverts if the key is unknown.
+function adminInfo(address account, address keyId)
```
samczsun: `admin` is already in `KeyInfo`, why also return it as a bool?

tips/tip-1049.md
```diff
@@ -0,0 +1,180 @@
...
+
+`verifySignature` is the canonical entry point for ERC-1271-style contracts that want to accept admin access keys. It is a thin composition of TIP-1020 verification with an admin-status check; TIP-1020 itself is not modified.
+
+## Behavioral Changes
+
+### Key Authorization by Access Keys
+
+Currently, the handler rejects any transaction where an access key attempts to authorize a different key (`AccessKeyCannotAuthorizeOtherKeys`). With this TIP:
+
+- If the signing access key has `is_admin = true`, the transaction MAY include a `KeyAuthorization` for a different `key_id`. The `KeyAuthorization` must still be signed by the root account key.
```
samczsun: I thought the point was that if the signing access key was an admin, then the `KeyAuthorization` didn't need to be signed by the root account anymore?

tips/tip-1049.md
samczsun: +1, this feels like poorly separated interfaces

tips/tip-1049.md
```diff
@@ -0,0 +1,180 @@
...
+- If the signing access key has `is_admin = false`, the existing restriction remains: the key can only authorize itself (same-tx auth+use).
+
+### Key Revocation by Access Keys
+
+Currently, only the root key can call `revokeKey`. With this TIP:
+
+- Admin access keys MAY call `revokeKey` to revoke other access keys on the same account.
+- Non-admin access keys MUST NOT call `revokeKey`.
```
samczsun: Semantically this is backwards, rather than enforcing that non-admin keys 'must not' call the admin function, it's more correct to say that the admin function must revert if called by a non-admin key

tips/tip-1049.md
```diff
@@ -0,0 +1,180 @@
...
+Currently, the handler rejects any transaction where an access key attempts to authorize a different key (`AccessKeyCannotAuthorizeOtherKeys`). With this TIP:
+
+- If the signing access key has `is_admin = true`, the transaction MAY include a `KeyAuthorization` for a different `key_id`. The `KeyAuthorization` must still be signed by the root account key.
+- If the signing access key has `is_admin = false`, the existing restriction remains: the key can only authorize itself (same-tx auth+use).
+
+### Key Revocation by Access Keys
+
+Currently, only the root key can call `revokeKey`. With this TIP:
+
+- Admin access keys MAY call `revokeKey` to revoke other access keys on the same account.
```
samczsun: And itself?

tips/tip-1049.md
samczsun: oops, should've done this before, but alas. agree with moving it out

tips/tip-1049.md
```diff
@@ -0,0 +1,186 @@
...
+
+2. **Onchain account ↔ key association.** Contracts that want to gate actions on "this key belongs to this account" have no canonical way to resolve a public key back to the account that authorized it, or to check whether the key has admin capabilities.
+
+Adding an `admin` flag plus a small set of read helpers in the `AccountKeychain` precompile solves both problems without entangling TIP-1020. TIP-1020 remains a pure signature-verification primitive; contracts that want admin-key semantics call the keychain explicitly.
+
+## Assumptions
+
+- The root EOA key is implicitly admin. The `admin` flag only applies to access keys registered in the AccountKeychain precompile.
+- Admin keys MUST NOT carry spending limits, call scopes, or expiry. Authorizations that combine `admin = true` with non-default `enforceLimits`, `allowedCalls`, or `expiry` MUST be rejected.
+- TIP-1020 (`SignatureVerifier`) is extended with a stateful `verifyStateful` method that returns true only for signatures produced by the account's root key or an admin access key. Non-admin access keys MUST NOT satisfy `verifyStateful`.
```
samczsun: This is probably not the right name - "stateful" means nothing to users. I'd suggest something like `verifyAdmin`/`verifyAccountAdmin`/`verifyAccessKeyAdmin`

tips/tip-1049.md
```diff
@@ -0,0 +1,176 @@
...
+authors: Jake Moxey (@jxom), Georgios Konstantopoulos (@gakonst), Tanishk Goyal (@tanishkgoyal)
+status: Draft
+related: TIP-1011, TIP-1020
+---
+
+# TIP-1049: Admin Access Keys
+
+## Abstract
+
+This TIP adds an `admin` boolean flag to the `AuthorizedKey` storage struct and the `KeyRestrictions` configuration. When `admin` is `true`, the access key may authorize and revoke other access keys on behalf of the account. The TIP also adds helper views to the `AccountKeychain` precompile so external contracts can resolve a key id back to its root account and inspect its admin status, without modifying TIP-1020 itself.
```
samczsun: This TIP no longer touches `KeyRestrictions`. This TIP also modifies TIP-1020 now

tips/tip-1049.md
```diff
@@ -0,0 +1,176 @@
...
+
+## Motivation
+
+Access keys today are strictly subordinate: they can sign transactions within their spending limits and call scopes, but they cannot authorize new keys or participate in onchain key-management flows. This creates two practical problems:
+
+1. **Cross-device key management.** A user with a passkey on their phone cannot authorize a second passkey on their laptop without a transaction from the root EOA key. In practice, multi-device setups require the root key to be available for every new device onboarding, which is inconvenient and fragile.
+
+2. **Onchain account ↔ key association.** Contracts that want to gate actions on "this key belongs to this account" have no canonical way to resolve a public key back to the account that authorized it, or to check whether the key has admin capabilities.
```
samczsun: Technically you could do this with `getKey` already

tips/tip-1049.md
```diff
@@ -0,0 +1,176 @@
...
+
+Adding an `admin` flag plus a small set of read helpers in the `AccountKeychain` precompile solves both problems without entangling TIP-1020. TIP-1020 remains a pure signature-verification primitive; contracts that want admin-key semantics call the keychain explicitly.
```
samczsun: This is false now

tips/tip-1049.md
```diff
@@ -0,0 +1,176 @@
...
+Byte 9:     enforce_limits (bool)
+Byte 10:    is_revoked (bool)
+Byte 11:    is_admin (bool)        ← NEW
+```
+
+This adds one byte to the existing packed slot. No new storage slots are introduced for the admin flag.
+
+## Interface Changes
+
+### `authorizeAdminKey`
```
samczsun, line 54: What happens if I call `authorizeAdminKey(rootAccount, ...)`?

tips/tip-1049.md
```diff
@@ -0,0 +1,176 @@
...
+/// Authorize a new admin key for the caller's account.
+/// @param keyId The key identifier (address derived from public key)
+/// @param signatureType 0: secp256k1, 1: P256, 2: WebAuthn
+function authorizeAdminKey(
+    address keyId,
+    SignatureType signatureType
+) external;
+```
+
+No `KeyRestrictions` argument is accepted: admin keys carry no limits, scopes, or expiry (Invariant 7). The accompanying `KeyAuthorization` (transported on the transaction in the same way `authorizeKey` consumes one) MUST be a V2 envelope with the `admin` bit set to `1`, and MUST be signed by **either the root account or admin access key** of the account. All other existing `authorizeKey` validation (T2 `tx_origin == msg_sender`, signature type, etc.) applies identically.
```
samczsun: What accompanying `KeyAuthorization`?

tips/tip-1049.md
```diff
@@ -0,0 +1,176 @@
...
+
+Today, the AccountKeychain precompile gates the following mutators on the root key via `ensure_admin_caller`:
+
+- `authorizeKey`
+- `revokeKey`
+- `updateSpendingLimit`
+- `setAllowedCalls`
+- `removeAllowedCalls`
+
+With this TIP, `ensure_admin_caller` is widened to also accept admin access keys. Admin access keys MAY call any of the five mutators above (including on itself).
```
samczsun: What does it mean for an admin key to call `updateSpendingLimit` / `setAllowedCalls` / `removeAllowedCalls` on itself?

tips/tip-1049.md
```diff
@@ -0,0 +1,176 @@
...
+
+Revocation of an admin key does NOT automatically revoke keys that it authorized. Each key is independently tracked; cascading revocation is out of scope for this TIP. Accounts that need cascading revocation should revoke downstream keys explicitly.
+
+## Transaction Encoding
+
+The `KeyAuthorization` encoding is bumped to V2 to carry the `admin` field. V2 uses a typed envelope mirroring [EIP-2718](https://eips.ethereum.org/EIPS/eip-2718): a single leading version byte followed by the RLP list.
+
+```
+V1 (legacy):    rlp([chain_id, key_type, key_id, expiry?, limits?, allowed_calls?])
+V2 (this TIP):  0x02 || rlp([chain_id, key_type, key_id, expiry?, limits?, allowed_calls?, admin])
```
samczsun: Maybe this should just be the "admin key envelope type" and the encoding is `0x02 || rlp([chain_id, key_type, key_id])`. If `admin = false`, use legacy. If `admin = true`, the optional fields must be omitted anyways. This also avoids weird "required field after optional field" rlp logic

tips/tip-1049.md
jxom: was only meant to state for revoke. updating.

tips/tip-1049.md
jxom: stale content. updating.

tips/tip-1049.md
jxom: yep, defs

tips/tip-1049.md
```diff
@@ -0,0 +1,177 @@
...
+
+## Motivation
+
+Access keys today are strictly subordinate: they can sign transactions within their spending limits and call scopes, but they cannot authorize new keys or participate in onchain key-management flows. This creates two practical problems:
+
+1. **Cross-device key management.** A user with a passkey on their phone cannot authorize a second passkey on their laptop without a transaction from the root EOA key. In practice, multi-device setups require the root key to be available for every new device onboarding, which is inconvenient and fragile.
+
+2. **Onchain admin signature verification.** Contracts that want to gate actions on "an admin key of account X signed this" (e.g. ERC-1271-style flows) have no canonical primitive: today they would have to combine raw signature recovery with their own key-status lookups. A single stateful verification method that composes recovery with admin-status checks removes that footgun.
```
samczsun: Now this isn't even a practical problem. There is no such situation where a contract would need to `gate actions on "an admin key of account X signed this"` because there is no such thing as an admin key today. This could be a logical motivation for why we want the stateful verification method _after_ this TIP is implemented, but this is being presented as one of two practical problems resulting from `Access keys today are strictly subordinate`. This is a nitpick but this is the sort of internal inconsistency that arises as a result of just having Amp patch over review comments instead of just doing it properly

tips/tip-1049.md
```diff
@@ -0,0 +1,177 @@
...
+/// Authorize a new admin key for the caller's account.
+/// @param keyId The key identifier (address derived from public key)
+/// @param signatureType 0: secp256k1, 1: P256, 2: WebAuthn
+function authorizeAdminKey(
+    address keyId,
+    SignatureType signatureType
+) external;
+```
+
+No `KeyRestrictions` argument is accepted: admin keys carry no limits, scopes, or expiry (Invariant 7). Like `authorizeKey`, the call is bound to a `KeyAuthorization` carried on the same transaction (TIP-1011) — for `authorizeAdminKey` this MUST be the V2 admin envelope (`0x02 || rlp([chain_id, key_type, key_id])`, see [Transaction Encoding](#transaction-encoding)), and MUST be signed by **either the root account or an admin access key** of the account. All other existing `authorizeKey` validation (T2 `tx_origin == msg_sender`, signature type, etc.) applies identically.
```
samczsun: Here is another example of Amp not understanding the _actual problem_ being pointed out in my previous review comment. I asked `What accompanying KeyAuthorization` because there's no reason to mention `KeyAuthorization` in this section, or at the very least there is no point in making it the headlining topic of the paragraph. I didn't ask it so the AI could re-define what the envelope format looks like and make it even more of the primary point of focus in this section.

We are defining a new function on the precompile which _accounts can call_ in order to authorize a new admin key. This section should describe how this function works. If we want to talk about the fact that it's possible to authorize-and-use a signed `KeyAuthorization` in a Tempo transaction, we should do that in a different section.

Also, TIP-1011 is for `Enhanced Access Key Permissions`. It doesn't define the `KeyAuthorization` type, so why is it being included in brackets? Did Amp hallucinate this too?

tips/tip-1049.md
```diff
@@ -0,0 +1,177 @@
...
+Byte 11:    is_admin (bool)        ← NEW
+```
+
+This adds one byte to the existing packed slot. No new storage slots are introduced for the admin flag.
+
+## Interface Changes
+
+### `authorizeAdminKey`
+
+`KeyRestrictions` is **unchanged** to preserve the ABI selector of the existing `authorizeKey` function. Admin key authorization is exposed via a new precompile method that mirrors the shape of `authorizeKey` but omits the `KeyRestrictions` config:
```
samczsun: Yet another case of Amp simply patching the file over and over with no understanding of what is actually happening. Why are we emphasizing that `KeyRestrictions` is unchanged in a section that does not touch `KeyRestrictions` at all? Is it because a previous version of the TIP changed the struct, and then we reverted that change, but the AI felt it very important to maintain this language in this part of the file, perhaps? If we are maintaining backwards compatibility, it should go in a backwards compatibility section,  along with all the other backwards compatibility notes in this doc

tips/tip-1049.md
```diff
@@ -0,0 +1,177 @@
...
+
+Today, the AccountKeychain precompile gates the following mutators on the root key via `ensure_admin_caller`:
+
+- `authorizeKey`
+- `revokeKey`
+- `updateSpendingLimit`
+- `setAllowedCalls`
+- `removeAllowedCalls`
+
+With this TIP, `ensure_admin_caller` is widened to also accept admin access keys. Admin access keys MAY call `authorizeKey`, `revokeKey`, `updateSpendingLimit`, `setAllowedCalls`, and `removeAllowedCalls`.
```
samczsun: See, the problem with the AI just stochastically editing files is that it will decide that it's a great idea to over-constrain the language and then suddenly your specification is enumerating a list of calls that admin keys are optionally allowed to make for no apparent reason. And then it forgets about `authorizeAdminKey` as a valid call because it wasn't in the bullet list and it didn't think hard about the fact that this TIP is introducing a new function which is also going to be gated on `ensure_admin_caller`

tips/tip-1049.md
```diff
@@ -0,0 +1,177 @@
...
+- First byte `0x02` → Admin envelope, authorizes an admin key.
+
+    - The admin envelope carries no `expiry`, `limits`, or `allowed_calls` because admin keys carry no restrictions (Invariant 7).
+- Future versions reserve typed-envelope bytes `0x03..0x7f`.
+
+# Invariants
+
+1. **Admin-only keychain mutation.** `authorizeKey` (outside the same-tx auth+use flow), `revokeKey`, `updateSpendingLimit`, `setAllowedCalls`, and `removeAllowedCalls` MUST revert with `AccountKeychainError::UnauthorizedCaller` when invoked by a non-admin access key.
+
+2. **Root-equivalence for admin keys.** Admin access keys MUST be accepted by `ensure_admin_caller` for all five keychain mutators (`authorizeKey`, `revokeKey`, `updateSpendingLimit`, `setAllowedCalls`, `removeAllowedCalls`). A `KeyAuthorization` MUST be accepted when signed by the root account OR admin access key of the account; all other existing checks (signature type, etc.) apply. The T2 `tx_origin == msg_sender` guard MUST be enforced for admin key callers.
```
samczsun: Now that I'm looking at it, these invariants also forgot about `authorizeAdminKey`.

tips/tip-1049.md
```diff
@@ -0,0 +1,177 @@
...
+
+3. **TIP-1020 stateless methods unchanged.** TIP-1020's `recover` and `verify` MUST behave identically to pre-TIP-1049 semantics. The new stateful `verifyAdmin` is additive; it is the only path that combines signature recovery with account/admin-status checks.
+
+4. **Backward compatibility.** Existing access keys (authorized before this TIP activates) default to `is_admin = false`. Their behavior is unchanged.
+
+5. **No cascading revocation.** Revoking an admin key MUST NOT automatically revoke keys that were authorized while that admin key was active. Each key's revocation status is independent.
```
samczsun: This is new normative behavior and should not be presented for the first time as an invariant

tips/tip-1049.md
```diff
@@ -0,0 +1,177 @@
...
+
+6. **Root key is not revocable.** `revokeKey` MUST reject any attempt to revoke the root EOA key. The root key remains the ultimate authority for the account under this TIP.
+
+7. **Admin keys carry no restrictions.** Admin authorizations MUST go through `authorizeAdminKey`, which structurally cannot accept `KeyRestrictions`. `authorizeKey` MUST reject any V2 `KeyAuthorization` whose `admin` bit is `1`. Admin keys are always unconditional and non-expiring.
```
samczsun: This still references "V2 `KeyAuthorization`" which is a remnant from previous iterations that Amp forgot to clean up. It's also needlessly verbose

tips/tip-1049.md
```diff
@@ -0,0 +1,177 @@
...
+/// was produced by the root key or an admin key of `account`. Returns false
+/// for non-admin access keys, unknown keys, and invalid signatures.
+function verifyAdmin(
+    address account,
+    bytes32 digest,
+    bytes calldata signature
+) external view returns (bool);
+```
+
+`verifyAdmin` MUST internally recompute the account-bound digest `keccak256(0x04 || digest || account)` before recovery, mirroring the V2 keychain signature binding. This prevents cross-account replay when the same admin key is registered on multiple accounts. Callers sign over the raw `digest`; the binding is applied inside the precompile and cannot be opted out of.
```
samczsun: Why do we bind the digest to the account when it's the application's responsibility to ensure that the digest is bound to the correct parameters? This also makes it impossible for most users to sign using an admin key, because nothing supports this non-standard signing scheme

tips/tip-1049.md
```diff
@@ -0,0 +1,177 @@
...
+- Future versions reserve typed-envelope bytes `0x03..0x7f`.
+
+# Invariants
+
+1. **Admin-only keychain mutation.** `authorizeKey` (outside the same-tx auth+use flow), `revokeKey`, `updateSpendingLimit`, `setAllowedCalls`, and `removeAllowedCalls` MUST revert with `AccountKeychainError::UnauthorizedCaller` when invoked by a non-admin access key.
+
+2. **Root-equivalence for admin keys.** Admin access keys MUST be accepted by `ensure_admin_caller` for all five keychain mutators (`authorizeKey`, `revokeKey`, `updateSpendingLimit`, `setAllowedCalls`, `removeAllowedCalls`). A `KeyAuthorization` MUST be accepted when signed by the root account OR admin access key of the account; all other existing checks (signature type, etc.) apply. The T2 `tx_origin == msg_sender` guard MUST be enforced for admin key callers.
+
+3. **TIP-1020 stateless methods unchanged.** TIP-1020's `recover` and `verify` MUST behave identically to pre-TIP-1049 semantics. The new stateful `verifyAdmin` is additive; it is the only path that combines signature recovery with account/admin-status checks.
```
samczsun: `The new stateful ... with account/admin-status checks` is not a testable invariant, nor is it relevant to `TIP-1020 stateless methods unchanged`

tips/tip-1049.md
samczsun: This still references "V2 `KeyAuthorization`" which is a remnant from previous iterations that Amp forgot to clean up. It's also needlessly verbose when it could just say `For all admin keys, expiry MUST equal zero, enforceLimits MUST equal false, and isScoped MUST equal false`

tips/tip-1049.md
jxom: I think it's moreso that in order to achieve the "access key verification" behavior, they would need to navigate via these flows.

jxom: So I think the correct language here is `...want to gate actions on "an access key of account X signed this"`

tips/tip-1049.md
jxom: I actually did this line without AI lmao, so this is my bad.

tips/tip-1049.md
```diff
@@ -0,0 +1,156 @@
...
+An admin access key MAY authorize new keys (admin or non-admin) by signing the `KeyAuthorization` itself, enabling delegation chains. This is intentional: it allows multi-device setups where any trusted device can onboard new devices without the root key being available at all.
+
+Revocation of an admin key does NOT automatically revoke keys that it authorized. Each key is independently tracked; cascading revocation is out of scope for this TIP. Accounts that need cascading revocation should revoke downstream keys explicitly.
+
+## Transaction Encoding
+
+A new **versioned envelope** for admin keys is introduced for `authorizeAdminKey`. It is a typed envelope mirroring [EIP-2718](https://eips.ethereum.org/EIPS/eip-2718): a single leading version byte followed by the RLP list.
+
+```
+0x02 || rlp([chain_id, key_type, key_id])
```
legion2002: This new wire format feels a little weird, think this isn't consistent with proper RLP encoding?

Should we consider adding the admin flag as an optional field to the existing key authorization?

jxom: Yeah, that was what was originally proposed but amended as per https://github.com/tempoxyz/tempo/pull/3645#discussion_r3155136228

agree an extra optional field is probably ok for now

tips/tip-1049.md
```diff
@@ -0,0 +1,156 @@
...
+) external view returns (bool);
+```
+
+`verifyAdmin` recovers the signer from `digest` directly using the same recovery rules as the existing stateless `verify(signer, digest, sig)`, then checks that the recovered key is an admin key of `account` (the root key, or an admin access key with `is_admin = true`).
+
+The existing stateless `verify(signer, digest, sig)` is unchanged.
+
+### `AccountKeychain.isAdminKey`
+
+A single read-only helper is added to the AccountKeychain precompile:
```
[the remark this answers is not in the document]
legion2002: think this is valid,, it could be a footgun. Ideally we discourage people from adding the same access key to multiple accounts, and if they do, they should make sure the digest has an account field.

Not sure if we can do much about this, other than highlighting in the docs, and add a big warning about it in the natspec

I don't think it makes sense to mandate the account prefix to all digests, it would defeat the purpose of the TIP.

tips/tip-1049.md
legion2002: @samczsun @0xrusowsky think it is safer to keep the original proposed version. 
We will have to build custom RLP decoding if we keep the tag outside the rlp list, because the RLP schemes that we derive from macros will expect the traditional RLP bytes as the leading fields.

Leading to issues like this - 
https://github.com/tempoxyz/tempo/pull/3645#discussion_r3240537160

tips/tip-1049.md
```diff
@@ -0,0 +1,156 @@
...
+
+- Guarded by `ensure_admin_caller`.
+- MUST reject `keyId == account` with `AccountKeychainError::InvalidKeyId`.
+- On success, sets `keys[account][keyId].is_admin = true` and emits `AdminKeyAuthorized`.
+
+### `AdminKeyAuthorized` Event
+
+The existing `KeyAuthorized` event is **unchanged** to preserve backward compatibility for indexers and decoders. A new event is added and emitted **alongside** `KeyAuthorized` whenever `admin = true`:
+
+```solidity
```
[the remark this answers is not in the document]
legion2002: @jxom think this is a valid problem. The TIP should explicitly specify that authorizeAdminKey should only be possible for keys with no spending limits etc.

Or if those key have some limits/scopes they have to be cleared. Maybe just ignoring the scopes would work. but I am worried about having such state inconsistencies onchain, which could lead to footguns. So would prefer to explicitly clear the storage for such keys.

tips/tip-1049.md
```diff
@@ -0,0 +1,156 @@
...
+A new **versioned envelope** for admin keys is introduced for `authorizeAdminKey`. It is a typed envelope mirroring [EIP-2718](https://eips.ethereum.org/EIPS/eip-2718): a single leading version byte followed by the RLP list.
+
+```
+0x02 || rlp([chain_id, key_type, key_id])
+```
+
+- First byte `≥ 0xc0` → Existing envelpe, authorizes a "non-admin" key.
+- First byte `0x02` → Admin envelope, authorizes an admin key.
+
+    - The admin envelope carries no `expiry`, `limits`, or `allowed_calls` because admin keys carry no restrictions.
```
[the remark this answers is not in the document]
legion2002: being discussed here - 
https://github.com/tempoxyz/tempo/pull/3645#discussion_r3173275751

think this is a valid issue.

tips/tip-1049.md
```diff
@@ -0,0 +1,156 @@
...
+
+### Admin Key Privilege Propagation
+
+An admin access key MAY authorize new keys (admin or non-admin) by signing the `KeyAuthorization` itself, enabling delegation chains. This is intentional: it allows multi-device setups where any trusted device can onboard new devices without the root key being available at all.
+
+Revocation of an admin key does NOT automatically revoke keys that it authorized. Each key is independently tracked; cascading revocation is out of scope for this TIP. Accounts that need cascading revocation should revoke downstream keys explicitly.
+
+## Transaction Encoding
+
+A new **versioned envelope** for admin keys is introduced for `authorizeAdminKey`. It is a typed envelope mirroring [EIP-2718](https://eips.ethereum.org/EIPS/eip-2718): a single leading version byte followed by the RLP list.
```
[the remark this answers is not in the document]
legion2002: @jxom This one is also an important bug to fix. In the current form, if the same admin key is authorized in multiple root accounts.

Then a key auth from this admin key can be replayed across all those root accounts. So once we go back to the original wiring format, along with is_admin, we should also add an explicit mandatory account field.

tips/tip-1049.md
legion2002: Also this is a valid issue - https://github.com/tempoxyz/tempo/pull/3645#discussion_r3240537151

tips/tip-1049.md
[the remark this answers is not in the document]
legion2002: Maybe the cleanest way to do this is to reject existing keys from becoming admin keys. 

only new keys can be marked as admin. This feels like a safer design.

legion2002 (commented):
A couple of important issues that need to be fixed in the wire format.

tips/tip-1049.md
0xrusowsky: i think the original ambiguity concern still applies... if we put a numeric version as the first item in the same RLP list as `chain_id`, it can be confused with a legacy auth whose first field is `chain_id == 2`.

i didn't think about the extra impl complications though, given that this would be a nested field and not a top-level EIP-2718 tx.

maybe a good compromise is to RLP-encode the version as a string rather than as a numeric value? this way there can't be confusion

```text
V1: rlp([chain_id, ...])
V2: rlp(["v2", chain_id, ...])
```

will defer final soundness assessment to @samczsun though

tips/tip-1049.md
```diff
@@ -0,0 +1,156 @@
...
+### Admin-Gated Keychain Mutators
+
+`ensure_admin_caller` is widened to additionally accept admin access keys. Every mutator gated by `ensure_admin_caller` (including the new `authorizeAdminKey`) MAY now be called by either the root key or an admin access key, and MUST revert with `AccountKeychainError::UnauthorizedCaller` for any other caller.
+
+When the target `keyId` resolves to an admin key (including the caller targeting itself):
+
+- `updateSpendingLimit`, `setAllowedCalls`, `removeAllowedCalls` MUST reject with `AccountKeychainError::InvalidKeyId`. Admin keys have no `KeyRestrictions`.
+- `revokeKey` MAY be called and revokes the admin key (including self-revocation).
+
+`revokeKey` MUST NOT be used to revoke the root key; any attempt to revoke the root EOA key MUST be rejected.
```
0xalpharush, line 129:
```suggestion
`revokeKey` CANNOT be used to revoke the root key; any attempt to revoke the root EOA key MUST be rejected.
```

tips/tip-1049.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Motivation
+
+Access keys today are strictly subordinate: they can sign transactions within their spending limits and call scopes, but they cannot authorize new keys or participate in onchain key-management flows. This creates two practical problems:
+
+1. **Cross-device key management.** A user with a passkey on their phone cannot authorize a second passkey on their laptop without a transaction from the root EOA key. In practice, multi-device setups require the root key to be available for every new device onboarding, which is inconvenient and fragile.
+
+2. **Onchain access key signature verification.** Contracts that want to gate actions on "an access key of account X signed this" (e.g. ERC-1271-style flows) have no canonical primitive: today they would have to combine raw signature recovery with their own key-status lookups. A single stateful verification method that composes recovery with key-status checks removes that footgun.
```
samczsun: This doesn't get addressed by the TIP though, only "an /admin/ access key of account X signed this" is enabled

tips/tip-1049.md
samczsun: Oh, I sorta assumed we had a primitive for this sort of optional-version-prefix decoding, given that it shows up elsewhere. I think anything is fine as long as it allows us to concretely disambiguate between different messages

tips/tip-1049.md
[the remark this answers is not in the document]
samczsun: Ok yeah let's bind the target account somehow

samczsun (approved):

tips/tip-1049.md
```diff
@@ -0,0 +1,161 @@
+---
+id: TIP-1049
+title: Admin Access Keys
+description: Extend access keys with an admin flag that grants key-management capabilities
+authors: Jake Moxey (@jxom), Georgios Konstantopoulos (@gakonst), Tanishk Goyal (@tanishkgoyal)
+status: Draft
```
jenpaff:
```suggestion
status: Approved
related: TIP-1011, TIP-1020
protocolVersion: T6
```

tips/tip-1049.md
```diff
@@ -0,0 +1,161 @@
...
+
+### `SignatureVerifier.verifyAdmin`
+
+TIP-1020's `SignatureVerifier` precompile is extended with a single stateful method:
+
+```solidity
+/// @notice Returns true if `signature` over `digest` was produced by the
+/// root key or an admin access key of `account`. Returns false for non-admin
+/// access keys, unknown keys, and invalid signatures.
+function verifyAdmin(
```
horsefacts, line 92: I'm not totally sold on this function. Why should admin access keys be uniquely privileged here? If we're going to introduce stateful verification, why can't I check whether a non-admin access key signature is currently valid? And what [changed our minds](https://github.com/tempoxyz/tempo/blob/main/tips/tip-1020.md#keychain-signature-rejection) about all this?

jxom: I guess you can't really scope non-admin access keys – what if a key I delegate to an app maliciously modifies some state of value?

0xrusowsky (approved):

tips/tip-1049.md
```diff
@@ -0,0 +1,163 @@
...
+An admin access key MAY authorize new keys (admin or non-admin) by signing the `KeyAuthorization` itself, enabling delegation chains. This is intentional: it allows multi-device setups where any trusted device can onboard new devices without the root key being available at all.
+
+Revocation of an admin key does NOT automatically revoke keys that it authorized. Each key is independently tracked; cascading revocation is out of scope for this TIP. Accounts that need cascading revocation should revoke downstream keys explicitly.
+
+## Transaction Encoding
+
+`KeyAuthorization` remains an RLP list. TIP-1049 adds two trailing optional fields:
+
+```
+rlp([chain_id, key_type, key_id, expiry?, limits?, allowed_calls?, is_admin?, account?])
```
mattsse, line 146: this should also include the TIP-1053 witness I believe

```suggestion
rlp([chain_id, key_type, key_id, expiry?, limits?, allowed_calls?, witness?, is_admin?, account?])
```

tips/tip-1049.md
```diff
@@ -0,0 +1,163 @@
...
+```
+
+- If `is_admin` is omitted or `false`, `account` MUST be omitted and the authorization creates a non-admin key.
+- If `is_admin` is `true`, `account` MUST be present and MUST equal the target account being modified.
+- If `is_admin` is `true`, `expiry`, `limits`, and `allowed_calls` MUST be omitted or encoded as empty optional values because admin keys carry no restrictions.
```
mattsse, line 151: and these updated accordingly for witness?

PR #3647, opened 2026-04-17
fix: charge CREATE cost on set_code

klkvr:


crates/precompiles/src/storage/evm.rs
```diff
@@ -108,9 +108,18 @@ impl<'a> PrecompileStorageProvider for EvmPrecompileStorageProvider<'a> {
         // Track state gas for code deposit
         self.deduct_state_gas(self.gas_params.code_deposit_state_gas(code_len))?;
 
-        self.internals
-            .load_account_mut(address)?
-            .set_code_and_hash_slow(code);
+        let was_empty = {
+            let mut account = self.internals.load_account_mut(address)?;
+            let was_empty = account.data.account().info.is_empty();
+            account.set_code_and_hash_slow(code);
```
fgimenez, line 114: is it ok to charge gas after state mutation?

fgimenez (approved):

crates/precompiles/src/storage/evm.rs
klkvr: this is how `sload` works already

PR #3648, opened 2026-04-17
docs(tip-1050): add P384 and SHA-384 signature verifier extension

Zygimantass:
Adds a draft TIP extending TIP-1020 with P384 signer recovery and raw ES384/SHA-384 helper methods.

The spec covers calldata encoding, gas pricing, compatibility expectations, and invariants for certificate and attestation verification flows.

### DIFF
--- a/tips/tip-1050.md
+++ b/tips/tip-1050.md
@@ -1,0 +1,237 @@
+---
+id: TIP-1050
+title: P384 and SHA-384 Extension for the Signature Verification Precompile
+description: Extends TIP-1020 with P384 signature verification and SHA-384 helper methods for onchain attestation and certificate validation flows.
+authors: Zygimantas Magelinskas (@zhygis)
+status: Draft
+related: TIP-1020, Tempo Transaction Spec
+protocolVersion: TBD
+---
+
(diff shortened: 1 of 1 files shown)


tips/tip-1050.md
```diff
@@ -0,0 +1,237 @@
...
+1. `0x05` is the P384 type byte. `0x03` and `0x04` remain reserved for keychain signatures, so `0x05` is the next free discriminant.
+2. `r`, `s`, `x`, and `y` are 48-byte big-endian field elements.
+3. `prehash` is a single byte.
+4. If `prehash == 0`, the `hash` argument passed to `recover()` / `verify()` is treated as the final SHA-256 digest to verify.
+5. If `prehash != 0`, the precompile MUST first compute `sha256(hash)` and verify that derived 32-byte digest.
+
+Successful P384 recovery derives the signer address as:
+
+```text
+address = last_20_bytes(keccak256(x || y))
```
0xalpharush, line 121: Why do we need to synthesize an address if this just used for TEE verification?

We could just do:
```
verifyES384(bytes48 digest, bytes signature, bytes publicKey) returns (bool)
sha384(bytes data) returns (bytes48)
```

And limit callers to zone's instead of overloading TIP-1020

PR #3650, opened 2026-04-17
fix(evm): exclude state gas from payment lane and incentive gas accounting

fgimenez:


crates/evm/src/block.rs
```diff
@@ -496,8 +503,18 @@ where
...
+
+        // T4+: use block_regular_gas_used (excludes state gas, ignores refunds per EIP-7778)
+        // for payment lane and incentive gas accounting, matching block gas limit semantics.
+        // Pre-T4: use tx_gas_used (state gas is 0 anyway).
+        let timestamp = self.evm().block().timestamp.to::<u64>();
+        let gas_used = if self.inner.spec.is_t4_active_at_timestamp(timestamp) {
+            self.inner.block_regular_gas_used - prev_regular
+        } else {
+            gas_output.tx_gas_used()
+        };
```
klkvr: we could use `inner` result's gas value here instead of the `prev_regular` hack

klkvr (approved):

PR #3655, opened 2026-04-17
chore(alloy): add `is_active_harfork` to provider

0xrusowsky:
this PR adds a new `is_active_hardfork()` method to the `TempoProviderExt` trait.

additionally, it also adds `tempo-chainspec` to the alloy release pipeline so that `TempoHardfork` is available to the sdk users.

### DIFF
--- a/crates/alloy/src/provider/ext.rs
+++ b/crates/alloy/src/provider/ext.rs
@@ -11,6 +11,8 @@ use tempo_contracts::precompiles::{
 };
 use tempo_primitives::transaction::CallScope;
 
+use tempo_chainspec::hardfork::TempoHardfork;
+
 use crate::{
     TempoFillers, TempoNetwork,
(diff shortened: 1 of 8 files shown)


crates/alloy/src/provider/ext.rs
```diff
@@ -94,6 +95,29 @@ pub trait TempoProviderExt: Provider<TempoNetwork> {
...
+    ) -> Result<bool, alloy_transport::TransportError>
+    where
+        Self: Sized,
+    {
+        #[derive(Debug, serde::Deserialize)]
+        struct Response {
+            active: String,
+        }
+
+        let resp: Response = self.raw_request("tempo_forkSchedule".into(), ()).await?;
```
mattsse, line 114: we could make this a separate fn as well

mattsse (approved):

PR #3656, opened 2026-04-17
fix(evm): set reservoir gas on frame result when gas limit check fails

rakita:
## Summary
- Compute `reservoir` before calling `check_gas_limit` so the OOG frame result returned on gas-limit failure reports the correct reservoir value
- Thread `reservoir` into `check_gas_limit` and apply it via `output.gas_mut().set_reservoir(reservoir)` when producing the OOG frame result
- Refactor context access to pull `tx`/`cfg` from `evm.ctx_ref().all()` once, reused across the gas-limit check and reservoir computation (both in execute and inspect paths)

## Test plan
- [ ] `cargo check -p revm`
- [ ] Existing TIP-1016 reservoir/gas-limit tests pass

### DIFF
--- a/crates/revm/src/handler.rs
+++ b/crates/revm/src/handler.rs
@@ -750,18 +750,18 @@ where
         evm: &mut Self::Evm,
         init_and_floor_gas: &InitialAndFloorGas,
     ) -> Result<FrameResult, Self::Error> {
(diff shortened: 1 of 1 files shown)


klkvr:
this OOG is not reachable post-T4 so it's fine to keep as is

PR #3664, opened 2026-04-20
feat: default bootnodes endpoint

decofe:
Default to `https://peers.tempo.xyz`. No new flags — opt out with `--tempo.bootnodes-endpoint=none` (or `TEMPO_BOOTNODES_ENDPOINT=none`).

Prompted by: kuyziss

### DIFF
--- a/bin/tempo/src/main.rs
+++ b/bin/tempo/src/main.rs
@@ -117,8 +117,7 @@ struct TempoArgs {
     ///
     /// Bootnodes for the current chain are added as peer hints to the discovery service.
     ///
-    /// Defaults to the chain-specific endpoint (e.g. `https://peers.tempo.xyz` on mainnet,
-    /// `https://testnet-peers.tempo.xyz` on moderato). Set to "none" to disable.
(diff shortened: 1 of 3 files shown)


crates/chainspec/src/spec.rs
```diff
@@ -165,6 +167,8 @@ pub struct TempoChainSpec {
     pub info: TempoGenesisInfo,
     /// Default RPC URL for following this chain.
     pub default_follow_url: Option<&'static str>,
+    /// Default bootnodes endpoint URL for this chain.
+    pub default_bootnodes_endpoint: Option<&'static str>,
```
mattsse: why do we even need this here?

bin/tempo/src/main.rs
mattsse, line 126: I think we can add `default_value = "http..."`

which should have the same effect as layering another config setting, especially since we never use `chainspec.default_bootnodes_endpoint` anywhere

mattsse (changes requested):

mattsse (approved):

PR #3667, opened 2026-04-20
Spike on rework CL -> EL pipeline

SuperFluffy:


crates/commonware-node/src/executor/actor.rs
```diff
@@ -174,405 +194,670 @@ where
...
+            certified_digest: digest,
+            certified_height: height,
+        }
+    }
+
+    fn to_fcu_state(&self) -> ForkchoiceState {
+        ForkchoiceState {
+            head_block_hash: self.certified_digest.0,
+            safe_block_hash: self.finalized_digest.0,
+            finalized_block_hash: self.certified_digest.0,
```
joshieDo, line 838:
```suggestion
            finalized_block_hash: self.finalized_digest.0,
```

crates/commonware-node/src/executor/actor.rs
```diff
@@ -174,405 +194,670 @@ where
...
-            neither valid nor syncing: `{payload_status}`"
-        );
-
-        acknowledgment.acknowledge();
+    /// 1. backfilling has highest priority: if the marshal actor's finalization
+    ///    view is ahead of the execution layer, the node likely suffered a
+    ///    persistence loss after shutdown/restart.
+    /// 2. `canonicalize-and-build` have highest priority so that proposers can
+    ///    return as proposal as soon as follow.
+    /// 3. certified blocks follow to ensure the node stays at the tip of the
```
hamdiallam, line 500: I believe this should take priority over canonicalize-and-build since a proposal requires being at tip?

hamdiallam: I think there might be a good `Sync` abstraction that all these can feed into such that the build is not triggered until we know the head hash has reached the target.

And the notarizations can hook into the same sync with no post-sync action

decofe:
Closing as stale. Please reopen or open a fresh PR if this is still needed.

PR #3675, opened 2026-04-20
fix(cli): validators-info lists all players, renamed to info with alias

hamdiallam:
* Read state from the same block height
* Rename validators-info to validator-committee

### DIFF
--- a/bin/tempo/src/tempo_cmd.rs
+++ b/bin/tempo/src/tempo_cmd.rs
@@ -153,7 +153,8 @@ pub(crate) enum ConsensusSubcommand {
     /// Look up a validator by etheruem address, e25519 public key, or index.
     Validator(ValidatorInfo),
     /// Query current committee information from the previous epoch's DKG outcome and current contract state.
-    ValidatorCommittee(ValidatorCommittee),
+    #[command(alias = "validators-info")]
+    Info(Info),
 }
 
 impl ConsensusSubcommand {
@@ -170,7 +171,7 @@ impl ConsensusSubcommand {
             Self::GeneratePrivateKey(args) => args.run(),
             Self::CalculatePublicKey(args) => args.run(),
             Self::Validator(args) => args.run().await,
-            Self::ValidatorCommittee(args) => args.run().await,
+            Self::Info(args) => args.run().await,
         }
     }
 }
@@ -1081,7 +1082,7 @@ struct ValidatorEntry {
 }
 
 #[derive(Debug, clap::Args)]
-pub(crate) struct ValidatorCommittee {
+pub(crate) struct Info {
     /// RPC URL to query. Defaults to <https://rpc.presto.tempo.xyz>
     #[arg(long, default_value = "https://rpc.presto.tempo.xyz")]
     rpc_url: String,
@@ -1092,7 +1093,7 @@ pub(crate) struct ValidatorCommittee {
     chain: Option<Arc<TempoChainSpec>>,
 }
 
-impl ValidatorCommittee {
+impl Info {
     async fn run(self) -> eyre::Result<()> {
         use alloy_consensus::BlockHeader;
         use alloy_provider::ProviderBuilder;
--- a/bin/tempo/src/tempo_cmd.rs
+++ b/bin/tempo/src/tempo_cmd.rs
@@ -152,8 +152,8 @@ pub(crate) enum ConsensusSubcommand {
     TransferValidatorOwnership(TransferValidatorOwnership),
     /// Look up a validator by etheruem address, e25519 public key, or index.
     Validator(ValidatorInfo),
-    /// Query validator info from the previous epoch's DKG outcome and current contract state.
-    ValidatorsInfo(ValidatorsInfo),
+    /// Query current committee information from the previous epoch's DKG outcome and current contract state.
+    ValidatorCommittee(ValidatorCommittee),
 }
 
 impl ConsensusSubcommand {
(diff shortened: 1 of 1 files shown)


bin/tempo/src/tempo_cmd.rs
```diff
@@ -1201,27 +1187,59 @@ impl ValidatorsInfo {
...
+
+        let active_validator_by_pubkey = ordered::Map::from_iter_dedup(
+            active_validators.iter().map(|v| (v.publicKey, v.clone())),
+        );
+
+        let players = dkg_outcome.players();
+        let next_players = dkg_outcome.next_players();
+
+        let mut validator_entries: Vec<ValidatorEntry> = Vec::new();
+        for public_key in ordered::Set::from_iter_dedup(players.iter().chain(next_players)) {
```
SuperFluffy: Isn't this missing the validators that have been added to the contract but not yet picked up by DKG yet?

Before this patch: we walked the result of `getActiveValidators`

After this patch: we walk current dealers + players and match against the result of `getActiveValidators`.

What we need:

1. collect the result of `getActiveValidators` into a growable map (so not the commonware `ordered::Map` but a btreemap keyed by public key)
2. walk dealers + players; if one is missing, look it up by pubkey, and push it into the map.
3. emit the resulting map.

(suggesting a btreemap over a hashmap so that ordering is stable)

SuperFluffy (changes requested):
Before we were missing validators that are about to fall out of committee. Now we are missing entries that have been added to the contract but were not yet picked up as players.

bin/tempo/src/tempo_cmd.rs
hamdiallam: Should we instead print by ascending index? Since that is stable for a validator. Easy to key in on

SuperFluffy: Problem is that there will be multiple entries with the same index (old and new validator after rotation).

hamdiallam: The old entry gets a new index at the end of the list?

bin/tempo/src/tempo_cmd.rs
```diff
@@ -1201,27 +1188,74 @@ impl ValidatorsInfo {
...
+            .wrap_err("failed to call getActiveValidators")?;
+
+        let active_validators = IValidatorConfigV2::getActiveValidatorsCall::abi_decode_returns(
+            &active_validators_result,
+        )
+        .wrap_err("failed to decode getActiveValidators response")?;
+
+        let active_validators_by_public_key = std::collections::BTreeMap::from_iter(
+            active_validators.iter().map(|v| (v.publicKey, v.clone())),
+        );
```
SuperFluffy: More idiomatic:
```suggestion
        let active_validators = active_validators
            .into_iter()
            .map(|v| (v.publicKey, v))
            .collect::<BTreeMap<_, _>>();
```

bin/tempo/src/tempo_cmd.rs
```diff
@@ -1201,27 +1188,74 @@ impl ValidatorsInfo {
...
+        let mut validators_by_public_key = active_validators_by_public_key.clone();
+
+        // Add validators that are in the dkg outcome but no longer active onchain
+        for public_key in dkg_players {
+            let key = B256::from_slice(public_key.as_ref());
+            if let std::collections::btree_map::Entry::Vacant(e) =
+                validators_by_public_key.entry(key)
+            {
+                let id = ValidatorId::PublicKey(key);
+                let validator = read_validator_from_contract(&provider, id).await?;
```
SuperFluffy: `wrap_err_with` and state which validators you were not able to look up.

bin/tempo/src/tempo_cmd.rs
```diff
@@ -1201,27 +1188,74 @@ impl ValidatorsInfo {
...
+                active,
+                onchain_address: validator.validatorAddress,
+                public_key: key.encode_hex(),
+                inbound_address: validator.ingress.clone(),
+                outbound_address: validator.egress.clone(),
+                fee_recipient: Some(validator.feeRecipient),
+                index: Some(validator.index),
+                is_dkg_dealer: Some(players.position(&public_key).is_some()),
+                is_dkg_player: Some(next_players.position(&public_key).is_some()),
+                in_committee: Some(players.position(&public_key).is_some()),
```
SuperFluffy: We should actually list the activation and deactivation height. The `let active = ...` dance is not necessary in that case.

bin/tempo/src/tempo_cmd.rs
```diff
@@ -1201,27 +1188,74 @@ impl ValidatorsInfo {
...
+                Ok(key) => key,
+                Err(e) => {
+                    let index = validator.index;
+                    eprintln!("invalid ed25519 public key found on validator index {index}: {e}",);
+                    continue;
+                }
+            };
+
+            validator_entries.push(ValidatorEntry {
+                active,
```
SuperFluffy: With the new format this is silly IMO - let's use the actual values we store for each entry (activation and deactivation height).

SuperFluffy (approved):
Smol change requests.

bin/tempo/src/tempo_cmd.rs
hamdiallam: Ahh nice figured there was a better way to do this

bin/tempo/src/tempo_cmd.rs
hamdiallam: Yea makes sense, Perhaps we can use this struct for both commands so that we have the height/epoch context too

struct SingleValidatorOutput {
    current_epoch: u64,
    current_height: u64,
    #[serde(flatten)]
    validator: ValidatorEntry,
}
```

SuperFluffy: Yeah, we do! we can probably streamline the construction for both cases.

PR #3679, opened 2026-04-21
fix: export OTLP traces/spans via telemetry-url

joshieDo:
Only logs were being exported because `cli.traces.otlp` was never set.
Adds the traces OTLP endpoint (`opentelemetry/v1/traces`) alongside the
existing logs endpoint so span names like `handle_propose` are exported.


SuperFluffy (approved):
IMO worth trying.

PR #3680, opened 2026-04-21
fix(revm): charge scope set length resets

legion2002:
Addresses SIGP-240.

Counts repeated `Set` length-slot writes in T3 key authorization scope gas as warm resets instead of fresh `SSTORE_SET` rows. This matches how `SetHandler::insert()` and `VecHandler::push()` actually update the shared length slot after the first insertion.

Adds a regression case covering repeated target and selector insertions in `calculate_key_authorization_gas()`.

### DIFF
--- a/crates/revm/src/handler.rs
+++ b/crates/revm/src/handler.rs
@@ -339,12 +339,15 @@ fn calculate_key_authorization_gas(
         let mut num_sstores = 1 + limit_slots;
         let mut num_reset_sstores = 0;
 
-        // T3+: include scoped-call storage rows in intrinsic gas.
-        if spec.is_t3() {
+        // T4+ splits repeated set length-slot writes out so T3 reexecution keeps the existing
+        // all-rows-as-SSTORE_SET accounting from current main.
+        if spec.is_t4() {
             let scope_slots = call_scope_storage_slots(&key_auth.authorization);
             let scope_reset_rows = call_scope_length_reset_rows(&key_auth.authorization);
             num_sstores += scope_slots.saturating_sub(scope_reset_rows);
             num_reset_sstores += scope_reset_rows;
+        } else if spec.is_t3() {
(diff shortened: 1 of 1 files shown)


crates/revm/src/handler.rs
```diff
@@ -172,8 +172,24 @@ fn call_scope_storage_slots(auth: &tempo_primitives::transaction::KeyAuthorizati
...
+fn call_scope_dirty_length_rows(auth: &tempo_primitives::transaction::KeyAuthorization) -> u64 {
+    let Some(scopes) = auth.allowed_calls.as_ref() else {
+        return 0;
+    };
+
+    scopes.len().saturating_sub(1) as u64
+        + scopes
+            .iter()
+            .map(|scope| scope.selector_rules.len().saturating_sub(1) as u64)
+            .sum::<u64>()
```
klkvr: is it really worth counting each 100 of gas here? imo it's ok to charge only for cold writes

0xrusowsky (approved):

crates/revm/src/handler.rs
legion2002: we don't add this helper fn just to count the extra 100 gas, we count it so that we can subtract it from the 250k sstore slots below. (which is basically the huge overcharge that we are trying to solve in this PR)

Since we have this number anyways, might as well charge the extra 100.

klkvr: if we'd only charge for cold SSTOREs it's not necessary to count warm ones separately i believe? we could just count number of call scope targets

legion2002: yeah, LOC are similar for both approaches, but yeah this approach is cleaner.
Updated it here - https://github.com/tempoxyz/tempo/pull/3680/commits/fe0db202b627f196414206512fa0f608a0eb6d3e

klkvr (approved):

Issue #1485, opened 2025-12-12
Change key authorization signing to EIP-712


legion2002:
## Summary
--
Key authorization signatures in the protocol are currently RLP-encoded, which makes them difficult to sign with common wallets like MetaMask. We should migrate key authorization signing to the EIP-712 typed data standard so that wallets can natively sign the payload.
 
## Motivation
- Wallet compatibility: MetaMask and other wallets do not support signing arbitrary RLP payloads, but they already support EIP-712 typed data.
- Developer ergonomics: Using the same EIP-712 flow as other Tempo transactions simplifies integrations and reduces bespoke signing paths.
- User clarity: Typed data provides clear signing prompts, reducing risk of incorrect or opaque approvals.
 
## Proposal
- Define an EIP-712 domain and typed data schema for key authorizations that captures the current RLP fields (including spend limits and permissions).
- Update the keychain/key authorization flow to produce and verify EIP-712 signatures instead of RLP-encoded payloads.
- Ensure backwards compatibility plan is documented (e.g., if RLP signatures should be deprecated or temporarily supported).
- Update developer documentation and examples to reflect the new signing format.
 
## Acceptance Criteria
- Key authorization signatures are generated and verified using EIP-712 typed data.
- Wallets that support EIP-712 (e.g., MetaMask) can sign key authorization payloads without custom flows.
- Documentation includes the new typed data schema and guidance for integrators.
- Tests cover EIP-712 signing and verification for key authorizations.

PR #3682, opened 2026-04-21
docs(tip-1051): EIP-712 signing support for KeyAuthorization and Tempo Transaction

legion2002:
Related #1485

Drafts TIP-1051 for EIP-712 signing support on Tempo transactions and key authorizations. The spec keeps canonical wire encodings unchanged, adds an explicit signature envelope to tag the signing scheme, and defines the typed-data, activation, replay, and verification rules needed to implement the feature.


tips/tip-1051.md
```diff
@@ -0,0 +1,420 @@
...
+Tempo-native signing path without changing the canonical wire encoding of
+`TempoTransaction` or `KeyAuthorization`.
+
+### Envelope Format
+
+When a signature uses the EIP-712 signature type, the encoded signature bytes
+are:
+
+```text
+0x7f || inner_signature_bytes
```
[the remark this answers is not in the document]
legion2002: addressed in https://github.com/tempoxyz/tempo/pull/3682/changes/3eb3f10a8bf3de30343657d67b7ef923e95d7444

tips/tip-1051.md
```diff
@@ -0,0 +1,420 @@
...
+
+## `KeyAuthorization` EIP-712 Signing
+
+When a `KeyAuthorization` signature is wrapped with `0x7f`, the signed digest is
+an EIP-712 digest under the `Tempo Key Authorization` domain.
+
+The typed data MUST commit to the same semantic fields as the current unsigned
+`KeyAuthorization`, including:
+
+1. `chain_id`
```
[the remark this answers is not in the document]
legion2002: addressed in https://github.com/tempoxyz/tempo/pull/3682/changes/59ff94ee29957bfa293084ce1db950a7fc8c3bc8

tips/tip-1051.md
```diff
@@ -0,0 +1,445 @@
...
+
+| Field | Value |
+|-------|-------|
+| name | `"Tempo Transaction"` |
+| version | `"1"` |
+| chainId | `tx.chain_id` |
+
+The chain ID is present in both the domain and the message payload. This is
+intentional. The domain binds the EIP-712 replay domain, while the message
+payload preserves the existing transaction and authorization fields.
```
klkvr, line 141: this doesn't explain why do we need the redundancy. the domain would be populated from the transaction's field anyway

tips/tip-1051.md
```diff
@@ -0,0 +1,445 @@
...
+    bytes4 selector;
+    address[] recipients;
+}
+
+struct CallScope712 {
+    address target;
+    SelectorRule712[] selectorRules;
+}
+
+struct KeyAuthorization712 {
```
klkvr: 712 suffixes look strange and iirc they are gonna be part of the payload so let's get rid of them?

tips/tip-1051.md
```diff
@@ -0,0 +1,445 @@
...
+
+1. `chainId = authorization.chain_id`
+2. `keyType = authorization.key_type as uint8`
+3. `keyId = authorization.key_id`
+4. If `authorization.expiry == None`:
+   - `hasExpiry = false`
+   - `expiry = 0`
+5. If `authorization.expiry == Some(x)`:
+   - `hasExpiry = true`
+   - `expiry = x`
```
klkvr: why do we need the `hasExpiry` bool if it can be inferred from `expiry != 0`?

tips/tip-1051.md
```diff
@@ -0,0 +1,445 @@
...
+1. `aaAuthorizationListHash = keccak256(rlp(tx.tempo_authorization_list))`
+2. If `tx.key_authorization == None`:
+   - `hasKeyAuthorization = false`
+   - `keyAuthorizationHash = 0x0000000000000000000000000000000000000000000000000000000000000000`
+3. If `tx.key_authorization == Some(x)`:
+   - `hasKeyAuthorization = true`
+   - `keyAuthorizationHash = keccak256(rlp(x))`
+
+This commits to the exact attached signed sidecars, including their signature
+bytes and their own signing-scheme envelopes if present.
```
klkvr: we define 712 schema for key authorization signature anyway so there's always a chance that it's gonna require versioning, why not extend this to transactions as well?

tips/tip-1051.md
```diff
@@ -0,0 +1,445 @@
...
+   - `validBefore = 0`
+10. If `tx.valid_before == Some(x)`:
+   - `hasValidBefore = true`
+   - `validBefore = x`
+11. If `tx.valid_after == None`:
+   - `hasValidAfter = false`
+   - `validAfter = 0`
+12. If `tx.valid_after == Some(x)`:
+   - `hasValidAfter = true`
+   - `validAfter = x`
```
klkvr: those bools are redundant because `Some(0)` is an unreachable state

tips/tip-1051.md
```diff
@@ -0,0 +1,445 @@
...
+data defined in this TIP.
+
+## Signature Envelope Encoding
+
+When a signature uses the EIP-712 signature type, the encoded signature bytes
+are:
+
+```text
+0x7f || inner_signature_bytes
+```
```
klkvr, line 64: this feels somewhat hacky, we already have signatures tagged internally and we could just extend this scheme instead of adding an outer tag

dankrad:
Did we look into the dangers of phishing with this change? One of the downsides is, EIP-712 transactions will not be recognized as normal transactions by wallets, so it's easier to hide phishing payloads in them. This seems to have become a significant problem with permit2 (which also uses EIP-712)

tips/tip-1051.md
legion2002: isn't this the same as extending the signature scheme with a new tag which takes in other internal schemes recursively?

can you recommend how we should do this

tips/tip-1051.md
legion2002: yeah we can just skip the chainId in the domain, 712 allows that. 
I just thought keeping it in here is useful for consistency, so that in the future we don't accidentally miss chainId for items that might not have them internally

tips/tip-1051.md
legion2002: yeah I think this is the RLP optional encoding thing again.

 iirc None and Some(0) give the same output during RLP roundtrips, so for all these optional fields we can infer None from 0 fields? (Just double checking before updating the TIP)

tips/tip-1051.md
```diff
@@ -0,0 +1,445 @@
...
+
+The current unwrapped signature encoding remains the canonical Tempo-native
+signing path. Implementations MUST NOT wrap a non-EIP-712 signature with `0x7f`.
+
+Implementations MUST reject all of the following:
+
+1. An empty signature byte string.
+2. A byte string beginning with `0x7f` whose total length is 65 bytes or
+   shorter.
+3. A nested envelope, where `inner_signature_bytes` itself begins with `0x7f`.
```
legion2002: addressed in 3085db9f91

tips/tip-1051.md
```diff
@@ -0,0 +1,445 @@
...
+   `TempoSignature` decoding rules.
+
+The current unwrapped signature encoding remains the canonical Tempo-native
+signing path. Implementations MUST NOT wrap a non-EIP-712 signature with `0x7f`.
+
+Implementations MUST reject all of the following:
+
+1. An empty signature byte string.
+2. A byte string beginning with `0x7f` whose total length is 65 bytes or
+   shorter.
```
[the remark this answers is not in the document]
legion2002: addressed in 3085db9f91

tips/tip-1051.md
```diff
@@ -0,0 +1,445 @@
...
+continues to be encoded as a single RLP byte string.
+
+## Activation and Compatibility
+
+This TIP activates at protocol version `T5`.
+
+At `T5`, the protocol adds support for the `0x7f || inner_signature_bytes`
+EIP-712 signature type defined in this TIP. The protocol continues to support
+existing unwrapped signatures. Implementations MUST verify signatures against
+exactly one digest determined by the signature-discrimination rules above.
```
legion2002: addressed in 3085db9f91

tips/tip-1051.md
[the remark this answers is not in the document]
legion2002: addressed in 9f6f33109

tips/tip-1051.md
klkvr: yep `Some(0)` encodes as `None` in RLP and we've migrated to `NonZeroU64` to fix that ambiguity so effectively `Some(0)` is unreachable now

tips/tip-1051.md
legion2002: addressed in 73e06ed30

tips/tip-1051.md
klkvr: the idea i had is that we could use last bit of the first byte of the signature encoding as a marker of whether it's using a EIP-712 payload. i.e right now if `signature[0] == 100000000` it's treated as P256 and we could treat `10000001` as P256 with EIP-712 digest

but i didn't realize that for secp256k1 sigs the only way to encode them is a legacy format with 65 bytes which means that for them we don't really have any byte tag we could pack this flag into, so i guess there's no way to fix this without adding an external tag

another option is to add a way to encode secp256k1 with a byte tag, like all other signature types

legion2002: yeah I think I prefer the external tag here, it provides us more flexibility in the future.
doing te bit approach is 1 byte more efficient, but feels like multiplying our supported sig types by 2.

tips/tip-1051.md
legion2002: addressed in 1e1dbcdd5

tips/tip-1051.md
legion2002: addressed in 1e1dbcdd5

PR #3687, opened 2026-04-21
chore: bump rand as per RUSTSEC advisory

0xrusowsky:
ref: `SIGP-242`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -178,7 +178,7 @@ dependencies = [
  "either",
  "k256",
  "once_cell",
- "rand 0.8.5",
+ "rand 0.8.6",
  "secp256k1 0.30.0",
  "serde",
  "serde_json",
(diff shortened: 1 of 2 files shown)


legion2002 (commented):
is the tempo-std bump accidental?

Cargo.lock
```diff
@@ -178,7 +178,7 @@ dependencies = [
  "either",
  "k256",
  "once_cell",
- "rand 0.8.5",
+ "rand 0.8.6",
```
legion2002, line 181: SIGP issue explicitly recommends switching to 0.9.3
Is this bump to 0.8.6 intentional?

klkvr: we can't remove the 0.8 dep entirely because a bunch of our dependencies depend on it

klkvr (approved):

klkvr:
tempo-std bump needs to be reverted i guess?

Cargo.lock
0xrusowsky: SIGP recommends switching 0.9.3 to 0.9.4

0.8.5 is a transitive dep, which also had a RUSTSEC advisory, so bumping it to 0.8.6 allowed us the remove it

0xrusowsky:
yup, reverted

PR #3689, opened 2026-04-21
fix(revm): gate key auth scope surcharge at T4

legion2002:
Follow-up to #3595.

Keeps the existing T3 scoped-call storage row pricing in place, but moves the rounded helper surcharge introduced in PR 3595 to T4.

The regression test now covers both sides of the split: T3 charges only the storage rows, while T4 adds the rounded helper overhead. TIP-1011 is updated to describe the surcharge as a T4 behavior.

### DIFF
--- a/crates/revm/src/handler.rs
+++ b/crates/revm/src/handler.rs
@@ -172,7 +172,7 @@ fn call_scope_storage_slots(auth: &tempo_primitives::transaction::KeyAuthorizati
     }
 }
 
-/// Charges the unpriced scope-helper bookkeeping for T3 key authorizations.
+/// Charges the unpriced scope-helper bookkeeping for T4 key authorizations.
 ///
 /// The dynamic SSTORE rows are already counted by `call_scope_storage_slots()`. What remains is the
 /// helper work around them: clearing the empty scope tree for fresh keys, target/set maintenance,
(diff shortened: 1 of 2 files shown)


crates/revm/src/handler.rs
```diff
@@ -314,11 +314,14 @@ fn calculate_key_authorization_gas(
...
-                .saturating_add(
-                    sstore_cost.saturating_mul(call_scope_storage_slots(&key_auth.authorization)),
-                )
-                .saturating_add(call_scope_extra_gas(&key_auth.authorization));
+            total = total.saturating_add(
+                sstore_cost.saturating_mul(call_scope_storage_slots(&key_auth.authorization)),
+            );
+        }
+
+        // T4+: add the rounded scope-helper surcharge introduced in PR 3595.
```
0xrusowsky: nit: can we remove the "introduced in PR 3595"?

legion2002: addressed https://github.com/tempoxyz/tempo/pull/3689/changes/fccb7e2873974dc83c3c6e4aea8ffd98d1471b25

0xrusowsky (approved):
makes sense, let's wait for @0xKitsune to also have a look before merging

shekhirin (approved):

0xKitsune (approved):

tips/tip-1011.md
```diff
@@ -435,7 +435,7 @@ Implementations may also charge a small rounded helper overhead for scoped-key a
 
 This overhead exists because fresh scope persistence includes additional bookkeeping such as clearing the empty scope tree, maintaining per-layer set metadata, and materializing recipient sets.
 
-Tempo's T3 implementation rounds this overhead upward using the same scope cardinalities:
```
jenpaff, line 438: I think we should add this to a T4 Meta TIP instead

PR #3696, opened 2026-04-22
fix(transaction-pool): avoid panic when add_transactions returns no row

### LINKED ISSUE #2347: Transaction Pool unwrap() in Production Code
## Affected Component
Transaction Pool (`crates/transaction-pool/src/tempo_pool.rs`)

## Description
The Tempo transaction pool contains `unwrap()` calls in production code at lines 352 and 361. While these unwraps are **protected by Reth's API contract** (which guarantees one result per input transaction), they could theoretically panic if Reth's implementation changes behavior.

## Root Cause
The code calls `add_transactions()` with single-item inputs and assumes it will always return at least one result:

```rust
// Line 449
.add_transactions(origin, std::iter::once(TransactionValidationOutcome::Valid { ... }))
    .pop()
    .unwrap()

// Line 468
.add_transactions(origin, Some(invalid))
    .pop()
    .unwrap()
```

The `.pop()` returns `Option<T>` and `.unwrap()` will panic if the Vec is empty.

### Steps to reproduce

Empty Result from Valid Transaction (Theoretical Only)

1. The current code path only executes for transactions with `nonce_key = 0` (standard protocol nonces)
2. Would require Reth's `add_transactions()` to return empty Vec
3. If Reth returns empty Vec for any reason, the `.pop().unwrap()` will panic
4. This would crash the transaction pool service

### Logs

```text

```

### Platform(s)

Linux (x86)

### Container Type

Not running in a container

v1.0.2 - the Presto release

_No response_

### Code of Conduct

crazywriter1:
Closes #2347

`add_validated_transaction` called `add_transactions(...).pop().unwrap()` for the
protocol pool path. Reth currently returns one `PoolResult` per input, but
trusting that with `unwrap` can still crash the node if the returned `Vec` is
ever empty.

- Capture the transaction hash before moving the validation outcome.
- Replace `unwrap` with `map_or`: empty `pop` returns `PoolError::other` with a
  clear message; non-empty passes through the inner `PoolResult` unchanged.

Same intent as the stale #3435; that PR did not land on `main`.


mattsse (commented):
these are fine because this is guaranteed to return the same legnth as the input iter

crazywriter1:
I opened this PR based on the issue, thanks for the clarification. Please don’t forget to close the issue as well. @mattsse

PR #3697, opened 2026-04-22
fix(consensus): bump commonware and adapt application actor

SuperFluffy:
Updates all commonware dependencies to the most recent head at https://github.com/commonwarexyz/monorepo/commit/2a7dd423f0a241276a5a38db8cc3d05f11de0c03.

This notably contains the patches https://github.com/commonwarexyz/monorepo/pull/3543 and
https://github.com/commonwarexyz/monorepo/pull/3641 to improve durability and persistence of proposed and verified blocks in the marshal actor.

In certain scenarios (especially around epoch transitions, but also in single-node tests), there were races between simplex engines asking for new proposals (overwriting the last-proposed-block mutex), and broadcasts of proposal parents.

This patch removes this logic, instead relying on the marshal actor to provide the blocks.


crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -762,7 +755,23 @@ impl Inner<Init> {
         .await
         .wrap_err("failed verifying block against execution layer")?;
 
-        Ok((block, is_good))
+        let block_height = block.height();
+        let block_digest = block.digest();
+
+        // Persist the block in the marshal actor and execution layer.
+        if !self.marshal.verified(round, block).await {
```
hamdiallam: Why do verify with the marshal if `is_good=false`?

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -490,18 +466,24 @@ impl Inner<Init> {
         // next epoch.
         if parent_epoch_info.last() == parent.height() && parent_epoch_info.epoch() == round.epoch()
         {
+            if !self.marshal.verified(round, parent.clone()).await {
```
hamdiallam, line 469: Why do we need to verify here? Isn't `marshal.proposed` enough?

For the new block we do build down below we don't call `verified` on it, just `proposed`

hamdiallam (commented):
Looks good, just a couple comments

crates/commonware-node/src/consensus/application/actor.rs
SuperFluffy: This was incorrect, good catch! ce898f37f

crates/commonware-node/src/consensus/application/actor.rs
SuperFluffy: `marshal.verified` and `marshal.proposed` are the same thing if you look at what marshal is doing under the hood. The only difference is that `marshal.proposed` keeps the last proposed block in memory for faster lookup in case it is being asked to rebroadcast.

I can only guess as to why this is done, but I assume it's to optimize for the new epoch instead of the old.

BTW, lifted all of this from commonware's impl.

joshieDo (approved):

PR #3677, opened 2026-04-21
feat(xtask): abi check

0xrusowsky:
this PR adds an ABI compatibility checker (cargo xtask check-abi) that compares rust `sol!` bindings used in the precompile impls against `tempo-std`'s Solidity interfaces, catching drift between the two surfaces. CI integration via `just check-abi` is wired up in `specs.yml`.

additionally, the PR removes the stale error types and precompiles (created during the development process, but which were never used).

finally, it also bumps `tempo-std` to its latest (alignment) commit.

```sh
>  just check-abi

  ✓  INonce
  ✓  IAccountKeychain
  ✓  ITIP20
  ✓  ITIP20Factory
  ✓  ITIP20RolesAuth
  ✓  ITIP403Registry
  ✓  IFeeAMM
  ✓  IFeeManager
  ✓  IStablecoinDEX
  ✓  IValidatorConfig
  ✓  IValidatorConfigV2

Summary: 11/11 interfaces are ABI-compatible.
```

PR #3699, opened 2026-04-22
refactor(spec): drop precompile Solidity reference impls

0xrusowsky:
build on top of:
- https://github.com/tempoxyz/tempo/pull/3677

this PR drops the Solidity reference precompile impls under `tips/ref-impls/src` and updates the spec tests to target rust precompiles directly via `tempo-std`.

since the PR removes the parallel Solidity-only path, the Foundry profiles, CI workflow, and README now reflect a single Tempo-native test setup. Additionally, i also deleted the (now useless) `./tempo-forge` and `./tempo-cast` wrapper scripts.

> NOTE: since profiles have been simplified and now we always use `default`, there needs to be a parallel PR updating the invariant test workflows


grandizzy (approved):
lgtm! further cleanup can be done in foundry.toml and remove ci profile altogether and have single place where we define hardfork / optimizer, we can follow-up

0xrusowsky:
> lgtm! further cleanup can be done in foundry.toml and remove ci profile altogether and have single place where we define hardfork / optimizer, we can follow-up

sounds good, i was also thinking we should do a followup to rename `ref-impls/` as that doesn't apply anymore. i left it there for the time being to minimize the diff

PR #3700, opened 2026-04-22
fix(fast_sync_after_full_dkg): race between epoch_manager hints and gap repair

hamdiallam:
When the epoch manager hints, the gap repair might finish before the next hint comes in. Thus the late validator never ends up skipping a round

Marshal's `MAX_REPAIR=20` on the consensus engine makes this race very likely ~40%. Since the epoch length matches, it can gap repair this in one go. By increasing the epoch length, it requires two round trips and the second hint should thus process causing a skip of epoch 0.


SuperFluffy (approved):
Awesome finding, thank you!

So the DKG manager never actually skipped any epochs because it was ripping through the 20 blocks too quickly.

PR #3704, opened 2026-04-23
feat(tip-1034): implement channel escrow precompile 

legion2002:
Depends on #3151

Adds the native TIP-20 channel escrow precompile from TIP-1034, including packed channel state, approve-less funding via system transfers, and voucher settlement, close, and withdraw flows.

The feature is now gated behind T5 end-to-end: the precompile is only registered and deployed once T5 is active. Payment-lane classification now lives in a stacked follow-up PR on top of this branch.

Note: The payment lane classification rules for MPP will be done as a separate follow up PR.


crates/precompiles/src/lib.rs
```diff
@@ -120,6 +121,8 @@ pub fn extend_tempo_precompiles(precompiles: &mut PrecompilesMap, cfg: &CfgEnv<T
             Some(TIP20Token::create_precompile(*address, &cfg))
         } else if *address == TIP20_FACTORY_ADDRESS {
             Some(TIP20Factory::create_precompile(&cfg))
+        } else if *address == TIP20_CHANNEL_ESCROW_ADDRESS && cfg.spec.is_t5() {
+            Some(TIP20ChannelEscrow::create_precompile(&cfg))
```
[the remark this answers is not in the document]
legion2002: will be included in a separate stacked PR.

crates/precompiles/src/tip20_channel_escrow/mod.rs
```diff
@@ -0,0 +1,909 @@
...
+        }
+
+        // `close_data` reserves 0 and 1 as sentinels, so tests and local fixtures that run
+        // with synthetic block timestamps of 0 or 1 can encode inconsistent channel state.
+        // Mainnet/testnet timestamps are guaranteed to be > 1, so this only matters outside
+        // real network execution.
+        let close_requested_at = self.now_u32();
+        let batch = self.storage.checkpoint();
+        state.close_data = close_requested_at;
+        self.channel_states[channel_id].write(state)?;
```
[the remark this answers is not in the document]
legion2002: block timestamp cannot be 1 for all cases we care about

crates/precompiles/src/tip20_channel_escrow/mod.rs
```diff
@@ -0,0 +1,909 @@
...
+            call.payee,
+            call.token,
+            call.salt,
+            call.authorizedSigner,
+        )?;
+        if self.channel_states[channel_id].read()?.exists() {
+            return Err(TIP20ChannelEscrowError::channel_already_exists().into());
+        }
+
+        let batch = self.storage.checkpoint();
```
klkvr: why do we need the checkpoint here?

crates/precompiles/src/tip20_channel_escrow/mod.rs
```diff
@@ -0,0 +1,909 @@
...
+            .ok_or_else(TIP20ChannelEscrowError::deposit_overflow)?;
+
+        if call.newExpiresAt != 0 {
+            if call.newExpiresAt as u64 <= self.now() || call.newExpiresAt <= state.expires_at {
+                return Err(TIP20ChannelEscrowError::invalid_expiry().into());
+            }
+        }
+
+        let had_close_request = state.close_requested_at().is_some();
+        let batch = self.storage.checkpoint();
```
klkvr: same here, seems unnecessary given that this would either succeed or return err and revert all state changes?

crates/precompiles/src/tip20_channel_escrow/mod.rs
klkvr, line 1: this has basically no docs right now, let's document all of the consts, structs, methods, etc

crates/contracts/src/precompiles/tip20_channel_escrow.rs
```diff
@@ -0,0 +1,245 @@
+pub use ITIP20ChannelEscrow::{
+    ITIP20ChannelEscrowErrors as TIP20ChannelEscrowError,
+    ITIP20ChannelEscrowEvents as TIP20ChannelEscrowEvent,
+};
+use alloy_primitives::{Address, address};
+
+pub const TIP20_CHANNEL_ESCROW_ADDRESS: Address =
+    address!("0x4D50500000000000000000000000000000000000");
+
+crate::sol! {
```
klkvr, line 12: also worth documenting all of this, it's not immediately clear what every field/method means if you're not familiar with the spec

crates/precompiles/src/tip20_channel_escrow/mod.rs
```diff
@@ -0,0 +1,909 @@
...
+            return Ok(());
+        }
+
+        // `close_data` reserves 0 and 1 as sentinels, so tests and local fixtures that run
+        // with synthetic block timestamps of 0 or 1 can encode inconsistent channel state.
+        // Mainnet/testnet timestamps are guaranteed to be > 1, so this only matters outside
+        // real network execution.
+        let close_requested_at = self.now_u32();
+        let batch = self.storage.checkpoint();
+        state.close_data = close_requested_at;
```
legion2002: invalid. block timestamp cannot be 1 for any cases we care about.

crates/contracts/src/precompiles/tip20_channel_escrow.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/pull/3704/commits/4d041622c3a9db3e87759fe8233389ab86cfdcb5

crates/precompiles/src/tip20_channel_escrow/mod.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/4d041622c3a9db3e87759fe8233389ab86cfdcb5

crates/precompiles/src/tip20_channel_escrow/mod.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/4d041622c3a9db3e87759fe8233389ab86cfdcb5

crates/precompiles/src/tip20_channel_escrow/mod.rs
legion2002: removed in https://github.com/tempoxyz/tempo/commit/4d041622c3a9db3e87759fe8233389ab86cfdcb5

crates/revm/src/tx.rs
```diff
@@ -264,7 +289,19 @@ impl IntoTxEnv<Self> for TempoTxEnv {
...
+            EthereumTxEnvelope::Eip7702(tx) => channel_open_context_hash(tx.tx(), sender),
+        };
+
+        Self {
+            inner: TxEnv::from_recovered_tx(tx, sender),
+            channel_open_context_hash: Some(channel_open_context_hash),
+            ..Default::default()
+        }
     }
 }
```
klkvr: i think the impl for `EthereumTxEnvelope` can just be removed

crates/revm/src/tx.rs
```diff
@@ -376,12 +414,25 @@ impl FromRecoveredTx<TempoTxEnvelope> for TempoTxEnv {
...
+            },
+            TempoTxEnvelope::Eip1559(tx) => Self {
+                inner: TxEnv::from_recovered_tx(tx.tx(), sender),
+                channel_open_context_hash: Some(channel_open_context_hash(tx.tx(), sender)),
+                ..Default::default()
+            },
+            TempoTxEnvelope::Eip7702(tx) => Self {
+                inner: TxEnv::from_recovered_tx(tx.tx(), sender),
+                channel_open_context_hash: Some(channel_open_context_hash(tx.tx(), sender)),
+                ..Default::default()
```
klkvr: can we have `fn channel_open_context_hash` on `TempoTxEnvelope`? also ideally we unify naming with `expiring_nonce_hash`, maybe something like `fn unique_tx_identifier` etc

crates/revm/src/tx.rs
```diff
@@ -157,6 +172,16 @@ impl From<TxEnv> for TempoTxEnv {
...
 }
 
+fn channel_open_context_hash<T>(tx: &T, sender: Address) -> B256
+where
+    T: SignableTransaction<Signature>,
+{
+    let mut buf = Vec::with_capacity(tx.payload_len_for_signature() + sender.as_slice().len());
+    tx.encode_for_signing(&mut buf);
+    buf.extend_from_slice(sender.as_slice());
+    keccak256(buf)
```
klkvr: this duplicates `expiring_nonce_hash` logic rn, we should unify

crates/revm/src/tx.rs
```diff
@@ -84,6 +86,11 @@ pub struct TempoTxEnv {
     /// Whether the transaction is a system transaction.
     pub is_system_tx: bool,
 
+    /// Replay-protected context hash used to derive channel escrow IDs for `open`.
+    ///
+    /// Synthetic transaction environments used by tests and simulations may leave this unset.
+    pub channel_open_context_hash: Option<B256>,
+
```
klkvr: we should remove `expiring_nonce_hash` from `TempoBatchCallEnv` now that we have this

crates/alloy/src/rpc/reth_compat.rs
```diff
@@ -198,6 +200,32 @@ impl TryIntoTxEnv<TempoTxEnv, TempoHardfork, TempoBlockEnv> for TempoTransaction
...
+
+fn channel_open_context_hash_from_sim_tx(tx: &TempoTxEnvelope, sender: Address) -> B256 {
+    match tx {
+        TempoTxEnvelope::Legacy(tx) => channel_open_context_hash_from_signable(tx.tx(), sender),
+        TempoTxEnvelope::Eip2930(tx) => channel_open_context_hash_from_signable(tx.tx(), sender),
+        TempoTxEnvelope::Eip1559(tx) => channel_open_context_hash_from_signable(tx.tx(), sender),
+        TempoTxEnvelope::Eip7702(tx) => channel_open_context_hash_from_signable(tx.tx(), sender),
+        TempoTxEnvelope::AA(tx) => tx.expiring_nonce_hash(sender),
+    }
+}
```
klkvr: why do we need this? the channel context hash we'd get here would still likely be different from the actual one due to different gas values

crates/precompiles/src/tip20_channel_escrow/mod.rs
```diff
@@ -0,0 +1,1490 @@
...
+    /// Opens a channel and pulls the initial deposit from the payer into escrow.
+    ///
+    /// Payees cannot be zero or TIP-20-prefix addresses. TIP-20-prefix payees would be token
+    /// contracts rather than ordinary recipients, which can make later payee payouts fail.
+    pub fn open(
+        &mut self,
+        msg_sender: Address,
+        call: ITIP20ChannelEscrow::openCall,
+    ) -> Result<B256> {
+        if call.payee == Address::ZERO || is_tip20_prefix(call.payee) {
```
legion2002, line 113: This pattern is a documented limitation of the TIP 1022 virtual address design. And we don't need to solve for it explicitly in the MPP contract.

crates/revm/src/tx.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/0f6282e6c

crates/revm/src/tx.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/0f6282e6c

crates/revm/src/tx.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/0f6282e6c

crates/revm/src/tx.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/0f6282e6c

crates/alloy/src/rpc/reth_compat.rs
legion2002: yeah we can hardcode a value here.

legion2002: addressed in https://github.com/tempoxyz/tempo/commit/78f985228

crates/contracts/src/precompiles/tip20_channel_escrow.rs
```diff
@@ -0,0 +1,410 @@
...
+        Self::ChannelNotFound(ITIP20ChannelEscrow::ChannelNotFound {})
+    }
+
+    pub const fn not_payer() -> Self {
+        Self::NotPayer(ITIP20ChannelEscrow::NotPayer {})
+    }
+
+    pub const fn not_payee() -> Self {
+        Self::NotPayee(ITIP20ChannelEscrow::NotPayee {})
+    }
```
0xrusowsky: i think this one is unused? if so, we could then drop that unused error type

crates/contracts/src/precompiles/tip20_channel_escrow.rs
```diff
@@ -0,0 +1,410 @@
...
+        Self::CloseNotReady(ITIP20ChannelEscrow::CloseNotReady {})
+    }
+
+    pub const fn deposit_overflow() -> Self {
+        Self::DepositOverflow(ITIP20ChannelEscrow::DepositOverflow {})
+    }
+
+    pub const fn transfer_failed() -> Self {
+        Self::TransferFailed(ITIP20ChannelEscrow::TransferFailed {})
+    }
```
0xrusowsky: same here, in rust errors are propagated so transfers fail with the corresponding TIP20 error

crates/contracts/src/precompiles/tip20_channel_escrow.rs
```diff
@@ -0,0 +1,410 @@
...
+
+/// TIP-1045 Maximum calldata length (in bytes) for payment-eligible calls with dynamic params.
+pub const MAX_PAYMENT_CALLDATA_LEN: usize = 2048;
+
+impl ITIP20ChannelEscrow::ITIP20ChannelEscrowCalls {
+    /// Returns `true` if `input` matches one of the recognized [TIP-20 channel escrow payment]
+    /// selectors: `open`, `topUp`, `settle`, `close`, `requestClose`, `withdraw`.
+    ///
+    /// # NOTES
+    /// - Only validates calldata; caller must check the that `to == TIP20_CHANNEL_ESCROW_ADDRESS`.
```
0xrusowsky:
```suggestion
    /// - Only validates calldata; caller must check that `to == TIP20_CHANNEL_ESCROW_ADDRESS`.
```

crates/contracts/src/precompiles/tip20_channel_escrow.rs
legion2002: Fixed in https://github.com/tempoxyz/tempo/commit/3ab1ba4e9.

crates/contracts/src/precompiles/tip20_channel_escrow.rs
legion2002: Fixed in https://github.com/tempoxyz/tempo/commit/3ab1ba4e9.

crates/contracts/src/precompiles/tip20_channel_escrow.rs
legion2002: Fixed in https://github.com/tempoxyz/tempo/commit/3ab1ba4e9.

crates/evm/src/block.rs
```diff
@@ -156,6 +161,7 @@ where
     ) -> Self {
         Self {
             incentive_gas_used: 0,
+            hardfork: chain_spec.tempo_hardfork_at(evm.block().timestamp.to::<u64>()),
```
klkvr: this is already available as `evm.cfg.spec`

crates/precompiles/src/tip20_channel_escrow/mod.rs
```diff
@@ -0,0 +1,1507 @@
...
+    }
+
+    /// Converts packed native storage to the public Solidity ABI shape.
+    fn to_sol(self) -> ITIP20ChannelEscrow::ChannelState {
+        ITIP20ChannelEscrow::ChannelState {
+            settled: self.settled,
+            deposit: self.deposit,
+            closeRequestedAt: self.close_requested_at,
+        }
+    }
```
klkvr, line 72: i'm wondering if it could makes sense to make `Storable` derive work directly on sol types?

crates/precompiles/src/tip20_channel_escrow/mod.rs
```diff
@@ -0,0 +1,1507 @@
...
+    fn enclosing_channel_open_context_hash(&self) -> Result<B256> {
+        let hash = self.channel_open_context_hash.t_read()?;
+        if hash.is_zero() {
+            return Err(TIP20ChannelEscrowError::expiring_nonce_hash_not_set().into());
+        }
+        Ok(hash)
+    }
+
+    /// Computes the channel id including chain and precompile domain separation.
+    #[allow(clippy::too_many_arguments)]
```
klkvr:
```suggestion
    #[expect(clippy::too_many_arguments)]
```

crates/precompiles/src/tip20_channel_escrow/mod.rs
```diff
@@ -0,0 +1,1507 @@
...
+            .keccak256(&(*VOUCHER_TYPEHASH, channel_id, cumulative_amount).abi_encode())?;
+        let domain_separator = self.domain_separator_inner()?;
+
+        let mut digest_input = [0u8; 66];
+        digest_input[0] = 0x19;
+        digest_input[1] = 0x01;
+        digest_input[2..34].copy_from_slice(domain_separator.as_slice());
+        digest_input[34..66].copy_from_slice(struct_hash.as_slice());
+        self.storage.keccak256(&digest_input)
+    }
```
klkvr, line 627: we could probably delegate a bunch of the logic here to alloy, all structs have eip712 helpers like `fn eip712_signing_hash`

klkvr: tho i guess maybe it's harder given dynamic pricing

crates/revm/src/tx.rs
```diff
@@ -371,32 +374,37 @@ impl FromRecoveredTx<AASigned> for TempoTxEnv {
...
 impl FromRecoveredTx<TempoTxEnvelope> for TempoTxEnv {
     fn from_recovered_tx(tx: &TempoTxEnvelope, sender: Address) -> Self {
+        let unique_tx_identifier = Some(tx.unique_tx_identifier(sender));
+
         match tx {
             tx @ TempoTxEnvelope::Legacy(inner) => Self {
                 inner: TxEnv::from_recovered_tx(inner.tx(), sender),
                 fee_token: None,
                 is_system_tx: tx.is_system_tx(),
+                unique_tx_identifier,
```
klkvr, line 384: isn't the `Option` redundant here? it's always `Some` it seems?

klkvr (commented):
only nits

crates/precompiles/src/tip20_channel_escrow/mod.rs
```diff
@@ -0,0 +1,1507 @@
...
+            return Err(TIP20ChannelEscrowError::amount_exceeds_deposit().into());
+        }
+
+        if capture > previous_settled {
+            self.validate_voucher(
+                &call.descriptor,
+                channel_id,
+                call.cumulativeAmount,
+                &call.signature,
+            )?;
```
0xKitsune, line 347: `close` skips voucher validation when `captureAmount <= previous_settled`, which is mostly fine but `ensure_payee_or_operator` lets any nonzero operator call `close`. 

This allows a third party operator to immediately close a channel and refund without a voucher. Any unsettled but signed voucher the payee holds is effectively unredeemable.

What is the reason for allowing `close` without a voucher?

crates/primitives/src/transaction/envelope.rs
```diff
@@ -475,14 +494,21 @@ impl From<TempoTransaction> for TempoTypedTransaction {
...
-fn is_tip20_payment(to: Option<&Address>, input: &[u8]) -> bool {
-    is_tip20_call(to) && ITIP20::ITIP20Calls::is_payment(input)
+fn is_tip1045_call(to: Option<&Address>, input: &[u8]) -> bool {
+    match to {
+        // TIP20 call + payment calldata constraints
+        Some(to) if to.is_tip20() => ITIP20::ITIP20Calls::is_payment(input),
+        // TIP20ChannelEscrow call + payment calldata constraints
+        Some(to) if *to == TIP20_CHANNEL_ESCROW_ADDRESS => {
+            ITIP20ChannelEscrowCalls::is_payment(input)
+        }
```
0xKitsune, line 509: `TempoPooledTransaction::new` now always caches `is_payment_v2()`, but execution doesnt switch to `v2` until T5.

Since escrow calls are now included in `is_payment_v2`, pre-T5 payload building can classify escrow txs as payments while execution classifies them as non payments.

A malicious user could spam txs targeting the escrow contract, causing builders to exceed the real pre-T5 gas budget, fail payload finalization without clearing these txs from the pool.

If proposers all have these txs in the pool, all validators will not be able to propose a block until these txs were cleared.

We should make sure payment lane logic in the pool/builder is fork aware.

crates/precompiles/src/tip20_channel_escrow/mod.rs
legion2002: The operator has the same trust assumptions as the payee. So it should be considered a trusted entity.
We allow close without a voucher, to enable the auth + capture use case like credit cards, where you can get permission for a larger amount of funds, but only charge some smaller value depending on actual usage.

crates/precompiles/src/tip20_channel_escrow/mod.rs
legion2002: Yeah it doesn't work well if you want to meter keccak pricing, ends up increasing LOC

crates/precompiles/src/tip20_channel_escrow/mod.rs
legion2002: yeah makes sense. We can do this in a follow up PR. CC: @0xrusowsky

crates/revm/src/tx.rs
legion2002: In onchain code paths, yes we always use Some. But there are simulation codepaths which might leave this as None.

crates/precompiles/src/tip20_channel_escrow/mod.rs
0xrusowsky: i believe this should be possible, but it would need some refactor work, cause the `tempo-precompiles` create deps on `tempo-contracts`.

we'd have to isolate the storage providers/primitives into its own crate first

crates/revm/src/tx.rs
klkvr: should simulation codepaths use a dummy value?

crates/evm/src/block.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/pull/3704/changes/632b73498ace6015e50a385fb9d1a4e30f920afa

crates/primitives/src/transaction/envelope.rs
legion2002: Yeah this is valid, addressed in - https://github.com/tempoxyz/tempo/pull/3704/commits/a8f6e613a535090a9bc313fca752c6cc96dd0746

crates/precompiles/src/address_registry/mod.rs
```diff
@@ -29,7 +29,11 @@ pub use tempo_primitives::{MasterId, TempoAddressExt, UserTag};
 /// Precompiles on this list are authorized to call
 /// [`crate::tip20::TIP20Token::system_transfer_from`], pulling TIP-20 tokens from a user without a
 /// prior `approve()`. The list is gated on `TempoHardfork::T5`; before activation it is empty.
-pub const IMPLICIT_APPROVAL_LIST: &[Address] = &[TIP_FEE_MANAGER_ADDRESS, STABLECOIN_DEX_ADDRESS];
+pub const IMPLICIT_APPROVAL_LIST: &[Address] = &[
+    TIP_FEE_MANAGER_ADDRESS,
+    STABLECOIN_DEX_ADDRESS,
+    TIP20_CHANNEL_ESCROW_ADDRESS,
```
jenpaff, line 35: please also update TIP-1035

crates/contracts/src/precompiles/tip20_channel_escrow.rs
```diff
@@ -0,0 +1,398 @@
...
+    pub fn is_payment(input: &[u8]) -> bool {
+        fn is_call<C: SolCall>(input: &[u8]) -> bool {
+            if input.first_chunk::<4>() != Some(&C::SELECTOR) {
+                return false;
+            }
+
+            if let Some(canonical_size) = <C::Parameters<'_> as SolType>::ENCODED_SIZE {
+                input.len() == 4 + canonical_size
+            } else {
+                input.len() <= MAX_PAYMENT_CALLDATA_LEN && C::abi_decode_validate(input).is_ok()
```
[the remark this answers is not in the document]
0xrusowsky: this is done on purpose for perf reasons. the 1045 spec explicitly allows up to 2KB on dynamic types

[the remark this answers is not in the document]
0xrusowsky: TIP-1045 doesn't require full abi decoding validation for perf reasons.
it only checks valid fn selector and length constraints.

crates/revm/src/handler.rs
```diff
@@ -396,22 +397,30 @@ impl<DB, I> TempoEvmHandler<DB, I> {
...
             &ctx.cfg,
             &ctx.tx,
             || {
                 let mut keychain = AccountKeychain::new();
-                keychain.set_tx_origin(ctx.tx.caller())
+                keychain.set_tx_origin(ctx.tx.caller())?;
+
+                if let Some(channel_open_context_hash) = channel_open_context_hash {
+                    let mut channel_escrow = TIP20ChannelEscrow::new();
+                    channel_escrow.set_channel_open_context_hash(channel_open_context_hash)?;
```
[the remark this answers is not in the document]
legion2002: we won't replace this precompile address with another random precompile address before the T5 hardfork activates. So this should not be a problem.

crates/precompiles/src/address_registry/mod.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/pull/3704/commits/a4a5a96ceba04d4eedd213fadacdeb84094c983b

klkvr (approved):

0xrusowsky (approved):

0xKitsune (approved):

PR #3707, opened 2026-04-23
refactor(consensus): use typed consensus errors

figtracer:
Replaces string-based consensus validation errors in `tempo-consensus` with a typed `TempoConsensusError` enum and updates the crate tests to assert on downcast variants instead of message text.

This keeps the existing validation behavior while making downstream error handling and tests more precise.

### DIFF
--- a/crates/consensus/src/lib.rs
+++ b/crates/consensus/src/lib.rs
@@ -3,8 +3,11 @@
 #![cfg_attr(not(test), warn(unused_crate_dependencies))]
 #![cfg_attr(docsrs, feature(doc_cfg))]
 
+mod error;
+
 use alloy_consensus::{BlockHeader, Transaction, transaction::TxHashRef};
 use alloy_evm::block::BlockExecutionResult;
+pub use error::TempoConsensusError;
 use reth_chainspec::EthChainSpec;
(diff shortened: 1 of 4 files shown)


SuperFluffy:
~~Appreciate the effort, but typing these errors does not make them actionable. This is intended as a binary only crate. We will revisit this should we decide to break out components.~~

crates/consensus/src/lib.rs
```diff
@@ -22,6 +23,43 @@ use tempo_primitives::{
     Block, BlockBody, TempoHeader, TempoPrimitives, TempoReceipt, TempoTxEnvelope,
 };
 
+/// Tempo-specific consensus errors.
+#[derive(Debug, thiserror::Error)]
+pub enum TempoConsensusError {
```
mattsse: lets move this to error.rs

dont like this top of file here

mattsse (changes requested):
this is better, in case we need to upcast this and relax certain checks such as:

https://github.com/paradigmxyz/reth/blob/62d99888d2eb70fdbf4836714d0ee4c1804d0e05/crates/consensus/consensus/src/lib.rs#L113-L123

SuperFluffy:
My bad, this actually makes sense.

mattsse (approved):
lgtm

fyi @joshieDo

PR #3710, opened 2026-04-23
feat(cli): --dry-run to show consensus requests without sending

hamdiallam:
* Wallet Arg is not required, runtime fail if not provided and --dry-run is not set
* --dry-run simply emits tx details

### DIFF
--- a/bin/tempo/src/tempo_cmd.rs
+++ b/bin/tempo/src/tempo_cmd.rs
@@ -330,10 +330,6 @@ pub(crate) struct WalletArgs {
     /// GCP_KEY_VERSION env vars
     #[arg(long, help_heading = "Wallet options - remote")]
     gcp: bool,
-
-    /// Output the unsigned transaction and exit without signing or sending.
-    #[arg(long, help_heading = "Wallet options")]
-    unsigned: bool,
 }
 
 impl WalletArgs {
@@ -391,7 +387,7 @@ impl WalletArgs {
 
             Ok(EthereumWallet::new(signer))
         } else {
-            bail!("a signing wallet option must be set")
+            bail!("a wallet option must be set")
         }
     }
 }
@@ -409,6 +405,10 @@ pub(crate) struct ValidatorTransactionArgs {
     /// Skip the interactive confirmation prompt.
     #[arg(long, short = 'y')]
     yes: bool,
+
+    /// Skip signing and submitting the transaction. Only printing out the details
+    #[arg(long)]
+    dry_run: bool,
 }
 
 impl ValidatorTransactionArgs {
@@ -417,7 +417,7 @@ impl ValidatorTransactionArgs {
             .to(VALIDATOR_CONFIG_V2_ADDRESS)
             .input(call.abi_encode().into());
 
-        if self.wallet.unsigned {
+        if self.dry_run {
             println!("{}", &serde_json::json!(tx));
             return Ok(());
         }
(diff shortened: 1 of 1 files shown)


bin/tempo/src/tempo_cmd.rs
```diff
@@ -330,65 +330,27 @@ pub(crate) struct WalletArgs {
...
-    /// The RPC URL to submit the transaction to.
-    #[arg(long, default_value = "https://rpc.presto.tempo.xyz")]
-    rpc_url: String,
 
-    /// Skip the interactive confirmation prompt.
-    #[arg(long, short = 'y')]
-    yes: bool,
+    /// Output the unsigned transaction and exit without signing or sending.
+    #[arg(long, help_heading = "Wallet options")]
+    unsigned: bool,
```
SuperFluffy: I suggest so simplify this. Instead of providing yet another flag, make the wallet args themselves optional.

No wallet args -> (raw) transaction is emitted to STDOUT.

hamdiallam: Not opposed, but if unsigned is the default with nothing set, do you think it could mistaken some folks to think they must sign transaction details?

hamdiallam: updated to `--dry-run`

bin/tempo/src/tempo_cmd.rs
```diff
@@ -432,17 +389,76 @@ impl ValidatorTransactionArgs {
...
+            .to(VALIDATOR_CONFIG_V2_ADDRESS)
+            .input(call.abi_encode().into());
+
+        if self.dry_run {
+            println!("{}", &serde_json::json!(tx));
+            return Ok(());
+        }
+
+        if !self.yes {
+            eprintln!("{}", &serde_json::json!(tx));
```
SuperFluffy: IMO always print out what will be submitted (and write to STDERR to make our lives easier). We can always add `--quiet` later - or the user can just `2> /dev/null`:

```suggestion
        eprintln!("{}", &serde_json::json!(tx));
        if self.dry_run {
            return Ok(());
        }
        if !self.yes {
```

bin/tempo/src/tempo_cmd.rs
```diff
@@ -432,17 +389,76 @@ impl ValidatorTransactionArgs {
...
+
+            let mut input = String::new();
+            std::io::stdin().read_line(&mut input)?;
+
+            if !matches!(input.trim(), "y" | "Y" | "yes" | "YES") {
+                bail!("transaction cancelled by user")
+            }
+        }
+
+        let wallet = self.wallet.build().await?;
```
SuperFluffy:
```suggestion
        let wallet = self.wallet.build().await.wrap_err("failed to open wallet to send transaction")?;
```

bin/tempo/src/tempo_cmd.rs
```diff
@@ -432,17 +389,76 @@ impl ValidatorTransactionArgs {
             bail!("a wallet option must be set")
```
SuperFluffy:
```suggestion
            bail!("no wallet provided")
```

bin/tempo/src/tempo_cmd.rs
```diff
@@ -432,17 +389,76 @@ impl ValidatorTransactionArgs {
...
+
+    /// The RPC URL to submit the transaction to.
+    #[arg(long, default_value = "https://rpc.presto.tempo.xyz")]
+    rpc_url: String,
+
+    /// Skip the interactive confirmation prompt.
+    #[arg(long, short = 'y')]
+    yes: bool,
+
+    /// Skip signing and submitting the transaction. Only printing out the details
```
SuperFluffy:
```suggestion
    /// Prints transaction and exits. Does not send sign or send the transaction.
```

SuperFluffy (approved):

PR #3711, opened 2026-04-23
docs(tip-1049): System-Contract Transfer Policy Exemption

dankrad:
## Summary

Introduces **TIP-1049: System-Contract Transfer Policy Exemption**.

This TIP defines a new privileged TIP-20 function, `systemForceTransfer(from, to, amount)`, that moves tokens while **skipping the TIP-403 policy check** and enforcing every other TIP-20 check (pause, balance, zero-recipient, AccountKeychain spending limits). It is gated behind a protocol-maintained **Transfer Policy Exemption List** of system contracts authorized to call it, with list entries being either literal addresses or *authority predicates* (contracts that vouch for a dynamic family of system contracts).

## Why

A system contract that custodies user tokens on behalf of a refund path needs a liveness guarantee: the refund must succeed regardless of TIP-403 policy drift between the time the custody was established and the time the refund is processed. The canonical example is Zones deposit/withdrawal bounce-backs, where today a `ZonePortal`'s refund `transfer` can revert if the recipient is subsequently blocked by a TIP-403 policy — leaving funds stranded and, in the current zones reference implementation, risking a stall in `ZoneInbox.advanceTempo` via a re-enqueued bounce with `fallbackRecipient = address(0)`.

Neither of the two existing bypasses is suitable:
- `systemTransferFrom` (TIP-1025/[TIP-1035](https://github.com/tempoxyz/tempo/blob/main/tips/tip-1038.md)) bypasses **allowance**, not policy — it still calls `ensure_transfer_authorized`.
- Hard-coding each system contract into TIP-20 or granting per-token roles doesn't scale to per-zone `ZonePortal` instances.

TIP-1049 introduces a dedicated, narrowly-scoped function and a dedicated list with its own admission criteria.

## What's in the TIP

- New TIP-20 function `systemForceTransfer(from, to, amount)` with precompile semantics spelled out step-by-step (pause → zero-recipient → caller gating → skip TIP-403 → spending limit → balance debit/credit → `Transfer` + new `SystemForceTransfer` events).
- A new `SystemForceTransfer(caller, from, to, amount)` event emitted alongside the standard `Transfer`, so compliance tooling can audit every policy-exempt refund separately.
- The **Transfer Policy Exemption List** structure, with literal-address and authority-predicate entry kinds, plus an `ITransferPolicyExemptAuthority` interface for the predicate form.
- Initial list on Tempo mainnet: a single authority entry for the canonical Tempo-side `ZoneFactory`, which exposes `isAuthorizedSystemContract(c) := isZonePortal(c)`. This covers every per-zone `ZonePortal` automatically and requires no per-zone amendment.
- Eligibility Guidelines: custody-only refunds, destination pre-validated at custody time (the existing deposit-time TIP-403 check on `bouncebackRecipient` anchors this), bounded amount, no EOA on the hot path, upgrade via hardfork, observable origin events.
- **No `mint` variant** — bypassing policy on mint would let listed contracts create supply at arbitrary addresses, which is materially wider than moving already-custodied tokens and is out of scope.
- Zones integration guidance: `ZonePortal.processWithdrawal` on the `gasLimit == 0` path should switch to `systemForceTransfer(address(this), withdrawal.to, amount)`. The existing `try/catch` stays for defense in depth but becomes exceptional rather than load-bearing, which in turn defuses the latent recursive-bounce issue with `fallbackRecipient = 0` ([zones#167](https://github.com/tempoxyz/zones/issues/167)).
- Backward compatibility: no existing transfer that TIP-403 authorized before this TIP is denied after it; `transfer`, `transferFrom`, `transferWithMemo`, `systemTransferFrom`, `allowance`, `permit`, and `mint` are unchanged.
- Hardfork-gated activation, list permanence, and explicit invariants and test cases.

## Relationship to existing TIPs

- **TIP-20**: extended with one new function and one new event.
- **TIP-403**: unchanged. Policy authors retain full control of `isAuthorizedSender`/`isAuthorizedRecipient`/`isAuthorizedMintRecipient`.
- **TIP-1025 / TIP-1035** (Implicit Approval List): independent axis. Allowance vs. policy bypasses are intentionally kept separate so the two don't get conflated in the security model.

## Test plan

- [ ] `systemForceTransfer` reverts with `NotPolicyExempt` for unlisted callers.
- [ ] Succeeds when TIP-403 would otherwise deny the transfer on `from` or `to`.
- [ ] Still reverts on insufficient balance, on `to == address(0)`, and when the token is paused.
- [ ] Emits exactly one `Transfer` and one `SystemForceTransfer` per successful call.
- [ ] Does not read or mutate allowance state.
- [ ] Enforces AccountKeychain spending limits on `from`.
- [ ] Authority-predicate integration: calls from a registered `ZonePortal` succeed; an unregistered look-alike reverts.
- [ ] End-to-end: a Zones deposit whose `bouncebackRecipient` is subsequently TIP-403-blocked still bounces back successfully on Tempo.
- [ ] No state changes on revert (balances, allowances, spending-limit counters, `totalSupply`, events).
- [ ] Hardfork gating: pre-fork calls to the `systemForceTransfer` selector revert with "function not found" and post-fork calls follow the new semantics.

### DIFF
--- a/tips/tip-1049.md
+++ b/tips/tip-1049.md
@@ -1,0 +1,230 @@
+---
+id: TIP-1049
+title: System-Contract Transfer Policy Exemption
+description: Defines a privileged TIP-20 transfer path that bypasses TIP-403 policy checks for a protocol-maintained list of system contracts that need guaranteed liveness for refunds and recoveries.
+authors: Dankrad Feist
+status: Draft
+related: TIP-20, TIP-403, TIP-1025, TIP-1035
+---
+
+# TIP-1049: System-Contract Transfer Policy Exemption
+
+## Abstract
+
+This TIP introduces a privileged TIP-20 transfer function, `systemForceTransfer`, that moves tokens while skipping TIP-403 transfer-policy authorization checks. It may only be invoked by an explicit, enumerated set of system contracts — the **Transfer Policy Exemption List**. Entries on the list may be literal addresses or *authority predicates* (contracts whose `isAuthorizedSystemContract(address)` view returns true), which lets the list grow to cover families of contracts that are deployed dynamically, such as per-zone [`ZonePortal`](../../zones/specs/spec.md#izoneportal) instances.
+
+This is a narrow, auditable exemption. It does not affect balance checks, spending-limit checks, event emission, or the allowance rules introduced by [TIP-1035](./tip-1035.md). It specifically addresses scenarios where a system contract must *return* tokens that it already custodies — most notably deposit and withdrawal bounce-backs for Zones — and where failing the transfer would either strand user funds or stall protocol progress.
+
+## Motivation
+
+### The problem: policy drift on refund paths
+
+Some system contracts custody user tokens and owe the user a refund under well-defined conditions. Two concrete examples in the current protocol:
+
+1. **Zone deposit bounce-backs.** A user deposits a TIP-20 token on Tempo and the sequencer relays the deposit to a zone. If the zone-side mint reverts (for example, because a TIP-403 policy active on the zone at processing time forbids minting to the recipient), the deposit must bounce back to a `bouncebackRecipient` on Tempo. The refund transfer is executed by the `ZonePortal` contract and is `ITIP20.transfer(bouncebackRecipient, amount)`.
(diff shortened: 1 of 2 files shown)


dankrad:
Superseded by #3723 — renumbered to **TIP-1052** to resolve a collision with #3645 (`tip: admin access keys`), which had an earlier claim on TIP-1049. The branch rename via GitHub's `branches/rename` API auto-closed this PR; #3723 continues from the same commits (renamed to `tip/1052`, file moved to `tips/tip-1052.md`, three in-body references updated).

PR #3715, opened 2026-04-24
perf: fix O(n²) scan in PendingStalenessTracker

rachit77:
`check_and_update` now performs a single-pass comparison between the current pending snapshot and the previous one. Entries already present in the previous snapshot are identified for stale eviction, while new entries form the next snapshot.

This removes the earlier `Vec::contains`-based filtering and ensures the update runs in linear time. In the benchmark (10k transactions with 50% stale), execution time improved from ~27.6ms to ~0.8ms.


mattsse (approved):
thanks, we can even improve this further on unify this in one scan

fgimenez (approved):
lgtm

PR #3718, opened 2026-04-24
feat: add T5 hardfork

fgimenez:
Adds T5 hardfork variant to the Tempo chain. No activation timestamps set for mainnet/moderato yet.

Changes follow the documented `Adding a New Hardfork` guide in `hardfork.rs`:
- `TempoHardfork::T5` variant in `tempo_hardfork!` macro
- `t5_time` field in `TempoGenesisInfo`
- `fork_time()` match arm
- `dev.json` and `test-genesis.json` with `t5Time: i64::MAX` (disabled)
- CLI arg in `genesis_args.rs`

Linear: https://linear.app/tempoxyz/issue/CHAIN-1139/t5-hardfork-meta-tip

### DIFF
--- a/xtask/src/genesis_args.rs
+++ b/xtask/src/genesis_args.rs
@@ -179,7 +179,7 @@ pub(crate) struct GenesisArgs {
     t4_time: u64,
 
     /// T5 hardfork activation time.
-    #[arg(long, default_value = "9223372036854775807")]
+    #[arg(long, default_value = "0")]
     t5_time: u64,
 }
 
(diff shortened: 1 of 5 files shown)


legion2002 (approved):

xtask/src/genesis_args.rs
```diff
@@ -177,6 +177,10 @@ pub(crate) struct GenesisArgs {
     /// T4 hardfork activation time.
     #[arg(long, default_value = "0")]
     t4_time: u64,
+
+    /// T5 hardfork activation time.
+    #[arg(long, default_value = "9223372036854775807")]
+    t5_time: u64,
```
mattsse, line 183: looks like this is causing a test failure

fgimenez: should be fixed now

0xKitsune (approved):

PR #3721, opened 2026-04-24
fix(consensus): treat future timestamps as transient

mattsse:
Marks `ConsensusError::TimestampIsInFuture` as transient in Tempo consensus so future-dated blocks are retried after the local clock catches up. This can happen during brief clock drift between nodes.

Validated with `cargo test -p tempo-consensus`.

### DIFF
--- a/crates/consensus/src/lib.rs
+++ b/crates/consensus/src/lib.rs
@@ -207,6 +207,11 @@ impl Consensus<Block> for TempoConsensus {
 
         self.inner.validate_block_pre_execution(block)
     }
+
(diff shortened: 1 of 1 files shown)


joshieDo (approved):

SuperFluffy:
> Marks ConsensusError::TimestampIsInFuture as transient in Tempo consensus so future-dated blocks are retried after the local clock catches up.

when would that retry happen?

mattsse:
@SuperFluffy clock drifts, added smol note

joshieDo:
>when would that retry happen?

as is, only in the next view/block - it would backfill. But at the very least, it shouldn't stall the node

PR #3724, opened 2026-04-24
feat(dex): TIP-1030 allow same-tick flip orders for T5+

fgimenez:
ref https://linear.app/tempoxyz/issue/PRODUCT-401/tip-1030-allow-same-tick-flip-orders

Relaxes `placeFlip` validation to allow `flipTick == tick`, gated behind T5

Linear: https://linear.app/tempoxyz/issue/CHAIN-1146/tip-1030-allow-same-tick-flip-orders


crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -675,8 +681,14 @@ impl StablecoinDEX {
...
-        let order = Order::new_flip(order_id, sender, book_key, amount, tick, is_bid, flip_tick)
-            .map_err(|_| StablecoinDEXError::invalid_flip_tick())?;
+        let order = if self.storage.spec().is_t5() {
+            Order::new_flip_with_same_tick(
+                order_id, sender, book_key, amount, tick, is_bid, flip_tick,
+            )
+        } else {
+            Order::new_flip(order_id, sender, book_key, amount, tick, is_bid, flip_tick)
+        }
+        .map_err(|_| StablecoinDEXError::invalid_flip_tick())?;
```
0xrusowsky: imo it would be simpler to modify `new_flip` to take a `TempoHardfork` param and apply the logic internally

crates/precompiles/src/stablecoin_dex/order.rs
```diff
@@ -124,12 +124,59 @@ impl Order {
...
+    fn new_flip_inner(
+        order_id: u128,
+        maker: Address,
+        book_key: B256,
+        amount: u128,
+        tick: i16,
+        is_bid: bool,
+        flip_tick: i16,
+        allow_same_tick: bool,
+    ) -> Result<Self, OrderError> {
```
0xrusowsky: then this would become the canonical impl, where u only need to replace `allow_same_tick: bool` for `spec: TempoHardfork`

0xrusowsky (commented):
current impl works, but i think we can reduce the LOC and make the change more succinct

crates/precompiles/src/stablecoin_dex/mod.rs
fgimenez: good point done here https://github.com/tempoxyz/tempo/pull/3724/commits/dec0a0c62ef383b799a52523e3c051639f9a8198 ptal

crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -633,8 +634,13 @@ impl StablecoinDEX {
...
-        if (is_bid && flip_tick <= tick) || (!is_bid && flip_tick >= tick) {
+        // Validate flip_tick relationship to tick based on order side.
+        // TIP-1030 (T5): allow flip_tick == tick for same-tick flip orders.
+        if self.storage.spec().is_t5() {
+            if (is_bid && flip_tick < tick) || (!is_bid && flip_tick > tick) {
+                return Err(StablecoinDEXError::invalid_flip_tick().into());
+            }
+        } else if (is_bid && flip_tick <= tick) || (!is_bid && flip_tick >= tick) {
             return Err(StablecoinDEXError::invalid_flip_tick().into());
         }
```
0xrusowsky: nit: i think this is equivalent and imo more comprehensive
```rs
if (flip_tick == tick && !self.storage.spec().is_t5()) 
    || (is_bid && flip_tick < tick)
    || (!is_bid && flip_tick > tick)
{
    return Err(StablecoinDEXError::invalid_flip_tick().into());
}
```

0xrusowsky (approved):

crates/precompiles/src/stablecoin_dex/mod.rs
0xKitsune: +1, lets update to this.

crates/precompiles/src/stablecoin_dex/order.rs
```diff
@@ -123,13 +129,27 @@ impl Order {
...
+
         if is_bid {
-            if flip_tick <= tick {
+            if bid_invalid {
                 return Err(OrderError::InvalidBidFlipTick { tick, flip_tick });
             }
-        } else if flip_tick >= tick {
+        } else if ask_invalid {
             return Err(OrderError::InvalidAskFlipTick { tick, flip_tick });
         }
```
0xKitsune, line 148: Consider simplifying/reducing to something like the following:

```rust

let t5_active = hardfork.is_t5();
let invalid = if is_bid {
    flip_tick < tick || (!t5_active && flip_tick == tick)
} else {
    flip_tick > tick || (!t5_active && flip_tick == tick)
};

if invalid {
    return Err(OrderError::invalid_flip(is_bid, tick, flip_tick));
}
```

crates/precompiles/src/stablecoin_dex/order.rs
```diff
@@ -123,13 +129,27 @@ impl Order {
...
+        let bid_invalid = if allow_same_tick {
+            flip_tick < tick
+        } else {
+            flip_tick <= tick
+        };
+        let ask_invalid = if allow_same_tick {
+            flip_tick > tick
+        } else {
+            flip_tick >= tick
+        };
```
0xKitsune: The Solidity invariant test for `TEMPO-DEX17` need to be updated. `_pickFlipTick` still uses strict inequality (t > tick / t < tick), so it never produces `flipTick == tick` and the assertion still claims `flipTick > tick / flipTick < tick`. Without this we do not have invariant coverage of the new T5 same tick path.

crates/precompiles/src/stablecoin_dex/order.rs
```diff
@@ -123,13 +129,27 @@ impl Order {
...
+
         if is_bid {
-            if flip_tick <= tick {
+            if bid_invalid {
                 return Err(OrderError::InvalidBidFlipTick { tick, flip_tick });
```
0xKitsune: nit: `OrderError::InvalidBidFlipTick` / `InvalidAskFlipTick` doc comments and `#[error("…must be greater than tick…")]` strings do not accurately describe T5+ behavior.

crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -633,8 +634,13 @@ impl StablecoinDEX {
...
 
-        // Validate flip_tick relationship to tick based on order side
-        if (is_bid && flip_tick <= tick) || (!is_bid && flip_tick >= tick) {
+        // Validate flip_tick relationship to tick based on order side.
+        // TIP-1030 (T5): allow flip_tick == tick for same-tick flip orders.
+        if self.storage.spec().is_t5() {
+            if (is_bid && flip_tick < tick) || (!is_bid && flip_tick > tick) {
+                return Err(StablecoinDEXError::invalid_flip_tick().into());
+            }
+        } else if (is_bid && flip_tick <= tick) || (!is_bid && flip_tick >= tick) {
```
0xKitsune: nit: we are duplicating this check both here and in `Order::new_flip()`. This pre-existed this change and we might need to keep the check here for backwards compatibility, but it looks like we will always fail early here and never actually use the `Order::new_flip()` errors in practice. There are no storage costs to this and its an inexpensive check but flagging in case we want to do something about it.

crates/precompiles/src/stablecoin_dex/mod.rs
fgimenez: sure done here https://github.com/tempoxyz/tempo/pull/3724/commits/140cd48341e3c8f4b3bcd16673e4254fbdf692f5

crates/precompiles/src/stablecoin_dex/order.rs
fgimenez: nice done here https://github.com/tempoxyz/tempo/pull/3724/commits/38d0d81c38cbc2d6cc9c9256e08cfd223e38b822

crates/precompiles/src/stablecoin_dex/order.rs
fgimenez: good catch done here https://github.com/tempoxyz/tempo/pull/3724/commits/1476fae1e18bd454bcdae61ec21c4b2a66998afc

crates/precompiles/src/stablecoin_dex/order.rs
fgimenez: done here https://github.com/tempoxyz/tempo/pull/3724/commits/8ef48b4b327d90d323e0f83544a6668b0489030b

crates/precompiles/src/stablecoin_dex/mod.rs
fgimenez: indeed, agree with the backwards compatibility point, added a note about it here https://github.com/tempoxyz/tempo/pull/3724/commits/45ac765451f7bff5a5c3e2df57eab2e80140af48 lmk if we should change anything else

crates/precompiles/src/stablecoin_dex/error.rs
```diff
@@ -14,9 +15,10 @@ pub enum OrderError {
     },
 
     /// Flip tick constraint violated for an ask flip order.
-    /// For asks: flip_tick must be < tick
+    /// Pre-T5: `flip_tick` must be `< tick`.
+    /// T5+ (TIP-1030): `flip_tick` must be `<= tick`.
     #[error(
-        "invalid flip tick for ask order: flip_tick ({flip_tick}) must be less than tick ({tick})"
+        "invalid flip tick for ask order: flip_tick ({flip_tick}) must be <= tick ({tick}) (< tick before T5)"
```
0xKitsune: nit: consider removing the `(>tick before T5)`. We can open an issue to update these after t5 activation

crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -2152,6 +2169,271 @@ mod tests {
...
+
+            let stored = exchange.orders[order_id].read()?;
+            assert_eq!(stored.tick(), tick);
+            assert_eq!(stored.flip_tick(), tick);
+            assert!(stored.is_bid());
+            assert!(stored.is_flip());
+
+            Ok(())
+        })
+    }
```
0xKitsune, line 2217: nit: consider consolidating these tests

crates/precompiles/src/stablecoin_dex/order.rs
```diff
@@ -112,9 +113,14 @@ impl Order {
...
     ///
+    /// The `hardfork` parameter controls flip-tick validation:
+    /// - Pre-T5: for bid flips `flip_tick > tick`; for ask flips `flip_tick < tick`.
+    /// - T5+ (TIP-1030): for bid flips `flip_tick >= tick`; for ask flips `flip_tick <= tick`.
+    ///
     /// # Errors
-    /// - `InvalidBidFlipTick` — `is_bid` is true and `flip_tick <= tick`
-    /// - `InvalidAskFlipTick` — `is_bid` is false and `flip_tick >= tick`
+    /// - `InvalidBidFlipTick` - `is_bid` is true and `flip_tick` is on the wrong side of `tick`
+    /// - `InvalidAskFlipTick` - `is_bid` is false and `flip_tick` is on the wrong side of `tick`
```
0xKitsune: nit: consider updating for T5:

```suggestion
    /// - `InvalidBidFlipTick` - `is_bid` is true and `flip_tick < tick`
    /// - `InvalidAskFlipTick` - `is_bid` is false and `flip_tick > tick`
```

0xKitsune (approved):

crates/precompiles/src/stablecoin_dex/error.rs
fgimenez: addressed here https://github.com/tempoxyz/tempo/pull/3724/commits/66f22b2865a810bb8968a7d3a47505dea03ad955

crates/precompiles/src/stablecoin_dex/mod.rs
fgimenez: addressed here https://github.com/tempoxyz/tempo/pull/3724/commits/1722fb3c01dc8fc986197f781bcf0fd414bee912

crates/precompiles/src/stablecoin_dex/order.rs
fgimenez: done here https://github.com/tempoxyz/tempo/pull/3724/commits/6c0b956757deddc26ecba146b0bb0950e347b5f0 (diff was stale and could not commit suggestion)

PR #3733, opened 2026-04-27
test(e2e): add backfill_on_start_after_crash restart test

joshieDo:
Adds a test that exercises the `backfill_on_start` code path in the executor, which was previously untested.

## What

- `TestingNode::unwind(n)` — simulates a crash by unwinding reth's EL by `n` blocks using `Pipeline::unwind` with `DefaultStages` + noop downloaders. Returns `(height_before, height_after)`.
- `backfill_on_start_after_crash` test — single-validator network, advance to height 10+, stop, unwind EL by 2, restart, assert EL recovers and node continues producing.

## Why

The `backfill_on_start` stream (`executor/actor.rs:191-204`) fires when CL finalized height > EL persisted height on restart. This happens after unclean shutdowns where reth loses a few unpersisted blocks. No existing test covers this path — all restart tests use clean shutdowns where EL and CL heights are in sync.


crates/e2e/src/tests/restart.rs
```diff
@@ -554,3 +555,47 @@ impl AssertNodeRecoversAfterFinalizingBlock {
         });
     }
 }
+
+/// After a clean stop, unwind reth's DB by 2 blocks, then restart. The CL's
```
SuperFluffy, line 559: Add a comment that this also simulates regular expected EL behavior, where the last few blocks are not guaranteed to be persisted even if finalized.

SuperFluffy (approved):

PR #3734, opened 2026-04-27
feat(precompiles): add U96 storage primitive

legion2002:
Adds a reusable `U96` type to the precompile storage system, including Alloy `Uint<96, 2>` conversions, storage/packing trait impls, and focused roundtrip tests.

This pulls the low-level packed-width support out of the larger TIP-1034 branch so follow-up escrow work can build on a smaller dependency PR.

Linear: https://linear.app/tempoxyz/issue/CHAIN-1147/tip-1034-enshrined-tip-20-escrow-channel

### DIFF
--- a/crates/precompiles/src/storage/types/mod.rs
+++ b/crates/precompiles/src/storage/types/mod.rs
@@ -17,7 +17,6 @@ pub use set::{Set, SetHandler};
 
 pub mod bytes_like;
 mod primitives;
-pub use alloy::primitives::aliases::U96;
 
 use crate::{
     error::Result,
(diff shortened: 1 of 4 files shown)


crates/precompiles/src/storage/types/mod.rs
```diff
@@ -17,6 +17,7 @@ pub use set::{Set, SetHandler};
 
 pub mod bytes_like;
 mod primitives;
+pub use alloy::primitives::aliases::U96;
```
0xrusowsky: nit: do we need this `pub`?

0xrusowsky (approved):

crates/precompiles/src/storage/types/mod.rs
legion2002: removed

PR #3736, opened 2026-04-27
fix(consensus): trim whitespace when parsing hex signing keys

kuyziss:
SigningKey/SigningShare::try_from_hex now trim input before hex decoding so files with a trailing newline (the common case) no longer fail with "odd number of digits". Complements #3615 which only covered --wallet-key and --new-private-key.

### DIFF
--- a/crates/commonware-node-config/src/lib.rs
+++ b/crates/commonware-node-config/src/lib.rs
@@ -27,11 +27,11 @@ impl SigningKey {
 
     pub fn read_from_file<P: AsRef<Path>>(path: P) -> Result<Self, SigningKeyError> {
         let hex = std::fs::read_to_string(path).map_err(SigningKeyErrorKind::Read)?;
-        Self::try_from_hex(&hex)
+        Self::try_from_hex(hex.trim())
     }
 
     pub fn try_from_hex(hex: &str) -> Result<Self, SigningKeyError> {
-        let bytes = const_hex::decode(hex.trim()).map_err(SigningKeyErrorKind::Hex)?;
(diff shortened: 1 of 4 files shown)


crates/commonware-node-config/src/lib.rs
```diff
@@ -31,7 +31,7 @@ impl SigningKey {
     }
 
     pub fn try_from_hex(hex: &str) -> Result<Self, SigningKeyError> {
-        let bytes = const_hex::decode(hex).map_err(SigningKeyErrorKind::Hex)?;
+        let bytes = const_hex::decode(hex.trim()).map_err(SigningKeyErrorKind::Hex)?;
```
SuperFluffy: If you do this, move this to `read_from_file` since  `try_from_hex` is not `try_from_hex_and_maybe_whitespace` :-)

crates/commonware-node-config/src/lib.rs
```diff
@@ -96,7 +96,7 @@ impl SigningShare {
     }
 
     pub fn try_from_hex(hex: &str) -> Result<Self, SigningShareError> {
-        let bytes = const_hex::decode(hex).map_err(SigningShareErrorKind::Hex)?;
+        let bytes = const_hex::decode(hex.trim()).map_err(SigningShareErrorKind::Hex)?;
```
SuperFluffy: Likewise, move this to the `read_from_file` logic.

SuperFluffy (approved):
IMO user error, but fine to use this for better UX.

PR #3738, opened 2026-04-27
feat: add xtask to get enode without starting a full node

laibe:
enode is needed for validator configuration, currently generating it requires starting up the node. This adds small xtask to generate it given private key and address.

### DIFF
--- a/xtask/src/enode_from_secret.rs
+++ b/xtask/src/enode_from_secret.rs
@@ -1,0 +1,51 @@
+use std::net::SocketAddr;
+
+use eyre::{Context, ensure};
+use reth_network_peers::pk2id;
+
+/// Derive the enode URL from a 32-byte secp256k1 secret key.
+///
+/// The enode identifier is the uncompressed secp256k1 public key with the
+/// leading 0x04 tag stripped (64 bytes = 128 hex characters).
+#[derive(Debug, clap::Parser)]
+pub(crate) struct EnodeFromSecret {
+    /// 64-character hex string of the 32-byte secp256k1 secret key.
+    ///
+    /// May optionally start with "0x".
+    #[arg(long)]
+    secret_key: String,
+
+    /// Optional socket address (ip:port) to produce the full enode URL.
+    ///
+    /// If omitted only the public key is printed.
+    #[arg(long)]
+    address: Option<SocketAddr>,
+}
+
+impl EnodeFromSecret {
+    pub(crate) fn run(self) -> eyre::Result<()> {
+        let mut hex = self.secret_key.trim();
+        hex = hex.strip_prefix("0x").unwrap_or(hex);
+        hex = hex.strip_prefix("0X").unwrap_or(hex);
+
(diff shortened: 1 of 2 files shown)


xtask/src/enode_from_secret.rs
```diff
@@ -0,0 +1,51 @@
...
+    /// If omitted only the public key is printed.
+    #[arg(long)]
+    address: Option<SocketAddr>,
+}
+
+impl EnodeFromSecret {
+    pub(crate) fn run(self) -> eyre::Result<()> {
+        let mut hex = self.secret_key.trim();
+        hex = hex.strip_prefix("0x").unwrap_or(hex);
+        hex = hex.strip_prefix("0X").unwrap_or(hex);
```
SuperFluffy, line 29: Unnecessary. `const_hex` (and `alloy_primitive::hex`, which is a reexport) does that afor you.

xtask/src/enode_from_secret.rs
```diff
@@ -0,0 +1,51 @@
...
+///
+/// The enode identifier is the uncompressed secp256k1 public key with the
+/// leading 0x04 tag stripped (64 bytes = 128 hex characters).
+#[derive(Debug, clap::Parser)]
+pub(crate) struct EnodeFromSecret {
+    /// 64-character hex string of the 32-byte secp256k1 secret key.
+    ///
+    /// May optionally start with "0x".
+    #[arg(long)]
+    secret_key: String,
```
SuperFluffy, line 16: Passing this in as a plain text command line arg makes me very uncomfortable.

Is this for testing purposes only or is there a chance this might be used for production nodes? If so, let's read the secret key via a file, or alternatively?

xtask/src/enode_from_secret.rs
```diff
@@ -0,0 +1,51 @@
...
+        hex = hex.strip_prefix("0x").unwrap_or(hex);
+        hex = hex.strip_prefix("0X").unwrap_or(hex);
+
+        ensure!(
+            hex.len() == 64,
+            "secret key must be exactly 64 hex characters (32 bytes), got {}",
+            hex.len()
+        );
+
+        let bytes = const_hex::decode(hex).wrap_err("invalid hex string")?;
```
SuperFluffy, line 37: Let const_hex handle that:

```suggestion
        let raw_secret_key = <[u8; 32]>::from_hex(&s).wrap_err("invalid input")?;
```

xtask/src/enode_from_secret.rs
```diff
@@ -0,0 +1,51 @@
...
+
+        let sk = secp256k1::SecretKey::from_slice(&bytes)
```
SuperFluffy, line 39: `from_slice` is deprecated; use https://docs.rs/secp256k1/latest/secp256k1/struct.SecretKey.html#method.from_byte_array:

```suggestion
        let sk = secp256k1::SecretKey::from_byte_array(bytes);
```

PR #3739, opened 2026-04-27
feat: TIP-1033 precompile + tx-pool impls

0xrusowsky:
this PR implements the TIP-1033 spec defined in #3143 for the `FeeManager` precompile and also integrates it with the transaction pool


crates/transaction-pool/src/amm.rs
```diff
@@ -64,56 +66,98 @@ impl AmmLiquidityCache {
...
+                let direct = inner
+                    .pool_cache
+                    .get(&(user_token, validator_token))
+                    .copied();
+                if matches!(direct, Some(r) if r >= amount_out) {
+                    return Ok(true);
+                }
+
+                if current_fork.is_t5() {
+                    match inner.quote_token_cache.get(&user_token).copied() {
```
0xrusowsky: addressed in [`1575824` (this PR)](https://github.com/tempoxyz/tempo/pull/3739/commits/157582463bc96745551c8a100933a57a64e9e0ef)

crates/transaction-pool/src/amm.rs
```diff
@@ -258,16 +302,25 @@ impl AmmLiquidityCache {
                 .collect();
         }
 
+        // Refresh the cached active hardfork from the latest seen header.
+        self.inner.write().current_fork = client.chain_spec().tempo_hardfork_at(latest_timestamp);
```
0xrusowsky: this is not a problem

crates/transaction-pool/src/amm.rs
```diff
@@ -64,56 +66,98 @@ impl AmmLiquidityCache {
...
-                    inner
-                        .slot_to_pool
-                        .insert(slot, (user_token, validator_token));
-
-                    // If the pool has enough liquidity, short circuit and return true
-                    if reserve >= amount_out {
+            .with_read_only_storage_ctx(current_fork, || -> TempoResult<bool> {
+                let manager = TipFeeManager::new();
+                for validator_token in deferred {
+                    if let Some(plan) = manager.plan_fee_swap(user_token, validator_token, fee)? {
```
0xrusowsky: addressed in [`554b31a` (this PR)](https://github.com/tempoxyz/tempo/pull/3739/commits/554b31ac69f60e78c39472c87f7b353785f9042b)

crates/precompiles/src/tip_fee_manager/amm.rs
```diff
@@ -488,6 +492,78 @@ impl TipFeeManager {
...
+        ));
+        if amount_out <= U256::from(direct.reserve_validator_token) {
+            data.route = Some(FeeRoute::Direct);
+            return Ok(data);
+        }
+
+        // T5+: two-hop fallback through `userToken.quoteToken()`.
+        if !self.storage.spec().is_t5() {
+            return Ok(data);
+        }
```
0xKitsune, line 529: Do we need to set the `data.route` here? Also applies to other if blocks that return data early before adding the second hop.

crates/precompiles/src/tip_fee_manager/amm.rs
```diff
@@ -86,6 +86,29 @@ impl PoolKey {
...
+/// Output of [`TipFeeManager::plan_fee_route`]. The chosen `route` plus state-data for every pool
+/// read by the planner, so callers can do pool introspection without redundant SLOADs.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct FeePlan {
+    /// Chosen route, or `None` if no path has sufficient liquidity.
+    pub route: Option<FeeRoute>,
+    /// Pools read during planning, paired with their observed validator-token reserve.
+    /// At most 3 entries: direct + (optionally) two legs. Order is irrelevant.
+    pub pools: Vec<((Address, Address), u128)>,
+}
```
0xKitsune: Do we need two different structs here?

We could simply have:


```rust
/// AMM path [`TipFeeManager`] will take to swap `user_token` into `validator_token` for fee collection.
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub enum FeeRoute {
    Direct,
    OneHop((Address, Address, u128)),
    TwoHop([(Address, Address, u128); 2]),
}
```

crates/precompiles/src/tip_fee_manager/amm.rs
0xrusowsky: no, cause pre-T5 there was no possible route.
when we don't set the route it returns `None`

crates/precompiles/src/tip_fee_manager/amm.rs
0xrusowsky: imo it's better with 2 routes, cause we always want to return the inspected state.

otherwise u need to account for the following scenarios:

- Same token: nothing inspected
- Direct: inspected direct liquidity
- 2hop: inspected direct + 2hop liquidity
- No route: inspected direct liquidity
- No route: inspected direct + 2hop liquidity

that's why this is better suited a enum (route) + struct which also contains the pool state (up to 3 pools inspected)

0xrusowsky: i guess we could drop the struct and make the return type more "opaque" by returning a tuple 
```rs
Result<(Option<FeeRoute>, Vec<(Address, Address, u128)>)>
```

i wouldn't mine doing that if u think it's cleaner

0xKitsune: > `Result<(Option<FeeRoute>, Vec<(Address, Address, u128)>)>`

This sgtm

0xrusowsky: addressed in [`1296acc` (this PR)](https://github.com/tempoxyz/tempo/pull/3739/commits/1296accce78641fe4edd7b7638c8c983cbd0d14f)

0xKitsune (approved):
Proposed a few small changes in https://github.com/tempoxyz/tempo/pull/3809.

Also can we make sure to update invariant tests to reflect cases added in the spec?

crates/transaction-pool/src/amm.rs
```diff
@@ -65,55 +68,86 @@ impl AmmLiquidityCache {
...
 
-                if let Some(validator_reserve) = inner.cache.get(&(user_token, *validator_token)) {
-                    if *validator_reserve >= amount_out {
-                        return Ok(true);
+                let direct = inner
+                    .pool_cache
+                    .get(&(user_token, validator_token))
+                    .copied();
+                if matches!(direct, Some(r) if r >= amount_out) {
+                    return Ok(true);
```
klkvr: nitpick: let's use `if let Some(r) = direct && r >= amount_out`

klkvr: we can also just add a `else` branch that puts the token into `missing_in_cache` immediately

crates/transaction-pool/src/amm.rs
```diff
@@ -65,55 +68,87 @@ impl AmmLiquidityCache {
...
+                    if !pools.is_empty() {
+                        let mut inner = self.inner.write();
+                        for &(pair, reserve) in &pools {
+                            let id = manager.pool_id(pair.0, pair.1);
+                            let slot = manager.pools[id].base_slot();
+                            inner.pool_cache.insert(pair, U256::from(reserve));
+                            inner.slot_to_pool.insert(slot, pair);
+                        }
+                        if let Some(FeeRoute::TwoHop(hop)) = route {
+                            inner.quote_token_cache.insert(user_token, hop);
```
klkvr: this seems a bit problematic because `plan_fee_route` might return `None` but still load the quote token which we won't cache meaning that any tx with the same user token would still do a cold storage read

crates/transaction-pool/src/amm.rs
```diff
@@ -148,30 +183,40 @@ impl AmmLiquidityCache {
...
             }
         }
+
+        // Process TIP-20 quote token updates: invalidate stale entries.
+        inner.quote_token_cache.retain(|token, _| {
+            execution_outcome
+                .account_state(token)
+                .and_then(|acc| acc.storage.get(&tip20::slots::QUOTE_TOKEN))
+                .is_none()
+        });
```
klkvr, line 214: maybe we should also re-validate all txs in the pool on quote token change, like we do for policy changes https://github.com/tempoxyz/tempo/blob/869dcf2123e606bf9e05966ca6cf32a189b86436/crates/transaction-pool/src/maintain.rs#L572-L622

klkvr (approved):
the precompile changes look good tome, some notes around pool code

klkvr:
> Payload builder scoring uses single-hop haircut for two-hop fee swaps, breaking payload_fees <= collected_delta invariant at T5+

regarding this imo it's worth considering extending `TempoTxResult` with the concrete fee paid to validator (propagated via `TempoEvm`) so that we don't have to duplicate the fee computation logic here for every AMM change we do
https://github.com/tempoxyz/tempo/blob/869dcf2123e606bf9e05966ca6cf32a189b86436/crates/payload/builder/src/lib.rs#L531

crates/transaction-pool/src/amm.rs
0xrusowsky: addressed in [`29023cb` (this PR)](https://github.com/tempoxyz/tempo/pull/3739/commits/29023cbca1722e852cf2ad5a09ccd6d0042aefbc)

crates/transaction-pool/src/amm.rs
0xrusowsky: addressed in [`29023cb` (this PR)](https://github.com/tempoxyz/tempo/pull/3739/commits/29023cbca1722e852cf2ad5a09ccd6d0042aefbc)

crates/transaction-pool/src/amm.rs
0xrusowsky: addressed in [`29023cb` (this PR)](https://github.com/tempoxyz/tempo/pull/3739/commits/29023cbca1722e852cf2ad5a09ccd6d0042aefbc)

klkvr (approved):

PR #3741, opened 2026-04-27
fix: only spend regular gas on subblocks fee payment halt

klkvr:
Changes the logic for subblocks fee payment failure to never produce regular gas spending exceeding the cap


fgimenez (commented):
lgtm, proposed regression test here https://github.com/tempoxyz/tempo/pull/3777

fgimenez (approved):

PR #3742, opened 2026-04-27
Deploy ERC2470 and NanoUniversalDeployer at T5 boundary

ScreamingHawk:
Adds [ERC2470 SingletonDeployer](https://eips.ethereum.org/EIPS/eip-2470) and [Nano Universal Deployer](https://gist.github.com/Agusx1211/de05dabf918d448d315aa018e2572031) as predeploys via two mechanisms:

- Genesis alloc (xtask/src/genesis_args.rs) — new networks get them at block 0
- T5 hardfork boundary (crates/evm/src/block.rs) — live mainnet gets them injected at the T5 activation block via deploy_contract_at_boundary, which is idempotent (skips if code already present)

Includes ABI JSON files and Rust bindings for both contracts in crates/contracts and integration tests verifying T5 boundary deployment and that both deployers' addresses matches the deterministic formula. 

---

This PR is required to support [Trails](https://trails.build/) cross chain interoperability on Tempo. Trails is built on the [Sequence v3 wallet](https://github.com/0xsequence/wallet-contracts-v3) contracts, which must be deployed at identical addresses on every supported chain. This is achieved by routing all deployments through a shared factory address with CREATE2, so the same bytecode and salt always resolve to the same address regardless of chain.

Both deployers are part of the [Sequence live-contracts deployment flow](https://github.com/0xsequence/live-contracts). Without them present on their canonical addresses, the Sequence deployment pipeline cannot be executed on Tempo and Trails integration is blocked.


mikeyb:
Would love to see Trails & Tempo united

avious00:
image

PR #3744, opened 2026-04-28
fix(consensus): drain CL->EL backfill before sending newer FCUs

joshieDo:
On restart with a CL/EL gap, `select_biased!` in the executor run loop
would fall through to the mailbox branch when `marshal.get_block()` was
momentarily pending between backfill blocks. A `Finalize` from live
consensus would get processed, its FCU would trigger a download from
EL peers, and the gap (>32 blocks) would cause reth to start pipeline
sync — breaking all subsequent backfill.

Moves the backfill stream drain into a dedicated `while let` loop that
completes fully before the main `select_biased!` loop starts.

Adds an e2e test that reproduces the bug with 4 validators, EL peers
connected, a 50-block unwind, and asserts `pipeline_runs == 0`.

CLOSES CHAIN-1115

### DIFF
--- a/crates/commonware-node/src/executor/actor.rs
+++ b/crates/commonware-node/src/executor/actor.rs
@@ -132,7 +132,7 @@ pub(crate) struct Actor<TContext> {
     pending_backfill: OptionFuture<BoxFuture<'static, (u64, Option<Block>)>>,
 
     /// Blocks received from the marshal actor that are awaiting execution and
-    /// acknowedgement. FuturesOrdered because it is nicer to use as a stream
+    /// acknowledgement. FuturesOrdered because it is nicer to use as a stream
     /// in a select-loop.
     pending_finalizations: FuturesOrdered<Ready<(Span, Block, Exact)>>,
 
--- a/crates/commonware-node/src/executor/actor.rs
+++ b/crates/commonware-node/src/executor/actor.rs
@@ -279,6 +279,10 @@ where
 
                 msg = self.mailbox.next() => {
                     let Some(msg) = msg else { break; };
+
+                    // CanonicalizeAndBuild will fail when the EL is in a syncing state during backfill.
+                    // CanonicalizeHead is safe to ignore for parents which will be forward finalized and
+                    // verification requests will also fail when the EL is syncing during backfill.
                     if let Command::Finalize(update) = msg.command {
                         pending_finalizations.push_back((msg.cause, update));
                     }
--- a/crates/commonware-node/src/executor/actor.rs
+++ b/crates/commonware-node/src/executor/actor.rs
@@ -195,40 +195,36 @@ where
                     let marshal = marshal.clone();
                     async move { (height, marshal.get_block(Height::new(height)).await) }
                 })
-                .fuse()
             )
         };
 
+        while let Some((height, block)) = backfill_on_start.next().await {
+            match block {
+                Some(block) => {
+                    let (ack, _wait) = Exact::handle();
+                    let span = info_span!("backfill_on_start", height);
+                    let _ = self.forward_finalized(span, block, ack).await;
+                }
+                None => {
+                    warn_span!("backfill_on_start", height).in_scope(|| {
+                        warn!(
+                            "marshal actor did not have block even though \
+                        it must have finalized it previously",
+                        )
+                    });
+                }
+            }
+        }
+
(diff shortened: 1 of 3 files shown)


crates/commonware-node/src/executor/actor.rs
```diff
@@ -195,40 +195,36 @@ where
                     let marshal = marshal.clone();
                     async move { (height, marshal.get_block(Height::new(height)).await) }
                 })
-                .fuse()
             )
         };
 
+        while let Some((height, block)) = backfill_on_start.next().await {
```
SuperFluffy: Don't do it this way; depending on how long it takes to backfill, you can end up in a situation where the mailbox ends up with hundreds of items.

Instead, keep draining the mailbox and buffer the finalized blocks / keep the latest tip.

SuperFluffy (changes requested):
Actors should always actively drain their mailboxes.

hamdiallam:
@joshieDo Taking over this PR. Looks like the added test was not passing to begin with

crates/commonware-node/src/executor/actor.rs
hamdiallam: Hm, if we don't `ack` the block can the pending message ever grow > 1?

crates/commonware-node/src/executor/actor.rs
```diff
@@ -139,16 +158,17 @@ where
...
-            .provider
-            .block_hash(last_execution_finalized_height)
-            .map_or_else(
-                |e| Err(Report::new(e)),
-                |hash| hash.ok_or_eyre("execution layer does not have the block hash"),
-            )
-            .wrap_err("failed to read the last finalized block hash")?;
-        let last_execution_finalized_height = Height::new(last_execution_finalized_height);
+
+        let canonical_state = execution_node.provider.canonical_in_memory_state();
```
SuperFluffy, line 162: Adjusted this slightly to get correct details from the execution layer (i.e. finalized -> finalized, canonical -> head, instead of using best/head for both as we did before).

crates/commonware-node/src/executor/actor.rs
```diff
@@ -158,15 +178,21 @@ where
...
+                finalized_height: Height::new(finalized_num_hash.number),
             },
             fcu_heartbeat_interval,
             fcu_heartbeat_timer,
+
+            finalized_heights_to_backfill,
+            pending_backfill: OptionFuture::none(),
+            pending_finalizations: FuturesOrdered::new(),
+
+            latest_observed_finalized_tip: None,
```
SuperFluffy, line 195: Working with these pending queues now so that we don't need 2 loops.

crates/commonware-node/src/executor/actor.rs
```diff
@@ -184,51 +210,81 @@ where
...
+            }
+
+            let finalized_tip_has_moved =
+                self.latest_observed_finalized_tip
+                    .is_some_and(|(height, digest)| {
+                        self.last_canonicalized
+                            != self.last_canonicalized.update_finalized(height, digest)
+                    });
+
+            select! {
```
SuperFluffy, line 230: Using a tokio select macro here instead, which allows us to conditionally enable/disable arms we don't want to run (like draining the finalized blocks and sending an FCU for the tip).

crates/e2e/src/tests/restart.rs
```diff
@@ -602,3 +603,69 @@ fn backfill_on_start_after_crash() {
...
+
+            if current >= el_before + 5 {
+                recovered = true;
+                break;
+            }
+        }
+
+        assert!(recovered, "unwound validator did not recover");
+
+        let new_pipeline_runs = get_pipeline_runs(metrics_recorder) - pipeline_runs_before;
```
SuperFluffy, line 668: Good find, @hamdiallam

SuperFluffy (approved):

hamdiallam (approved):
LGTM! Not a strong pref was nice to have the backfill logic tightly scoped to just startup vs leak a bit more into the actor

crates/commonware-node/src/executor/actor.rs
SuperFluffy: I just saw this - but yeah, this can grow in the general case (right now, the marshal actor is set to only send 1 finalized block, but it can in theory send more).

PR #3746, opened 2026-04-28
feat: allow omitting empty subblocks metadata system tx

klkvr:
Right now all blocks include an empty subblocks metadata tx which is somewhat redundant. This PR changes building and validation logic to allow omitting subblocks metadata transaction if there are no subblocks included

### DIFF
--- a/crates/consensus/src/lib.rs
+++ b/crates/consensus/src/lib.rs
@@ -254,7 +254,7 @@ mod tests {
     use std::time::{SystemTime, UNIX_EPOCH};
     use tempo_chainspec::{
         hardfork::TempoHardfork,
-        spec::{MODERATO, TempoChainSpec},
+        spec::{DEV, MODERATO, TempoChainSpec},
     };
 
     fn current_timestamp_millis() -> u64 {
@@ -827,7 +827,7 @@ mod tests {
     }
 
     #[test]
-    fn test_validate_block_pre_execution_no_system_tx() {
+    fn test_validate_block_pre_execution_pre_t4_missing_system_tx() {
         let consensus = TempoConsensus::new(MODERATO.clone());
         let chain_id = MODERATO.chain().id();
 
@@ -850,6 +850,23 @@ mod tests {
                 )),
             "Expected MissingEndOfBlockSystemTxs, got: {err:?}"
         );
+    }
+
+    #[test]
+    fn test_validate_block_pre_execution_t4_allows_missing_system_tx() {
+        let consensus = TempoConsensus::new(DEV.clone());
+        let chain_id = DEV.chain().id();
+
+        let user_tx = create_tx(chain_id);
+
+        let header = TestHeaderBuilder::default()
+            .gas_limit(30_000_000)
+            .timestamp(0)
+            .build();
+        let block = create_valid_block(header, vec![user_tx]);
+        let sealed = SealedBlock::seal_slow(block);
+
+        assert!(consensus.validate_block_pre_execution(&sealed).is_ok());
     }
 
     #[test]
--- a/crates/consensus/src/lib.rs
+++ b/crates/consensus/src/lib.rs
@@ -174,9 +174,19 @@ impl Consensus<Block> for TempoConsensus {
             .into());
         }
 
+        let expected_system_tx_count = if self
+            .inner
+            .chain_spec()
+            .is_t4_active_at_timestamp(block.header().timestamp())
+        {
+            0
+        } else {
+            SYSTEM_TX_COUNT
+        };
+
         // Get the last END_OF_BLOCK_SYSTEM_TX_COUNT transactions and validate they are end-of-block system txs
         let end_of_block_system_txs = transactions
-            .get(transactions.len().saturating_sub(SYSTEM_TX_COUNT)..)
+            .get(transactions.len().saturating_sub(expected_system_tx_count)..)
             .map(|slice| {
                 slice
                     .iter()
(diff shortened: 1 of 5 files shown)


crates/payload/builder/src/lib.rs
```diff
@@ -127,9 +123,14 @@ impl<Provider: ChainSpecProvider<ChainSpec = TempoChainSpec>> TempoPayloadBuilde
...
     fn build_seal_block_txs(
         &self,
-        block_env: &BlockEnv,
+        evm: &TempoEvm<impl Database>,
         subblocks: &[RecoveredSubBlock],
     ) -> Vec<Recovered<TempoTxEnvelope>> {
+        if subblocks.is_empty() && evm.cfg.spec.is_t4() {
+            // Post-T4, omit the subblocks metadata transaction if there are no subblocks
+            return vec![];
+        }
```
0xKitsune, line 132: Currently, if spec is t4 and subblocks is not empty, the builder will still build the subblocks metadata tx. Is there a reason we cant just do something like:

```rust
if evm.cfg.spec.is_t4(){
    return vec![];
}
```

crates/evm/src/block.rs
```diff
@@ -564,6 +564,18 @@ where
...
+        let seen_subblock_signatures = match self.section {
+            BlockSection::System {
+                seen_subblocks_signatures,
+            } => seen_subblocks_signatures,
+            _ => false,
+        };
+
+        // Post T4, if subblocks metadata transaction was not seen, imply empty metadata.
+        if !seen_subblock_signatures && self.evm().cfg.spec.is_t4() {
+            self.validate_shared_gas(&[])?;
```
0xKitsune, line 577: Why do we need to call this post T4?

crates/payload/builder/src/lib.rs
klkvr: we must include the system transaction if subblocks are not empty, otherwise the validation of the block would fail. we make sure that subblocks are empty by other means tho by not building any, so this shouldn't really happen in practice

crates/evm/src/block.rs
klkvr: to fail block validation if there were any subblocks included or incentive gas used

0xKitsune (approved):

crates/consensus/src/lib.rs
```diff
@@ -174,9 +174,19 @@ impl Consensus<Block> for TempoConsensus {
...
 
+        let expected_system_tx_count = if self
+            .inner
+            .chain_spec()
+            .is_t4_active_at_timestamp(block.header().timestamp())
+        {
+            0
+        } else {
+            SYSTEM_TX_COUNT
+        };
```
0xKitsune, line 185: Can we add tests for both branches here? T4 block with 0 end-of-block system txs accepted, and pre‑T4 block missing the metadata tx still rejected.

crates/evm/src/block.rs
```diff
@@ -214,6 +214,10 @@ where
                 ));
             }
 
+            if self.evm().cfg.spec.is_t4() {
+                return Err(BlockValidationError::msg("subblocks are disabled in T4+"));
+            }
```
0xKitsune, line 219: Can we add a test  that a metadata tx in a T4 block is rejected with this error.

crates/evm/src/block.rs
```diff
@@ -561,6 +565,18 @@ where
...
+            BlockSection::System {
+                seen_subblocks_signatures,
+            } => seen_subblocks_signatures,
+            _ => false,
+        };
+
+        // Post T4, if subblocks metadata transaction was not seen, imply empty metadata.
+        if !seen_subblock_signatures && self.evm().cfg.spec.is_t4() {
+            self.validate_shared_gas(&[])?;
+        }
```
0xKitsune, line 578: Can we add tests for:
- T4 activated and no metadata seen passes when incentive_gas_used == 0
- fails when > 0.

crates/evm/src/lib.rs
```diff
@@ -216,7 +216,7 @@ impl ConfigureEvm for TempoEvmConfig {
             .rev()
             .filter(|tx| (*tx).to() == Some(Address::ZERO))
             .find_map(|tx| Vec::<SubBlockMetadata>::decode(&mut tx.input().as_ref()).ok())
-            .ok_or(TempoEvmError::NoSubblockMetadataFound)?
+            .unwrap_or_default()
```
0xKitsune, line 219: The deleted `test_context_for_block_no_subblock_metadata` should be replaced with a test asserting empty subblock_fee_recipients here, ideally gated on T4.

0xKitsune (approved):
Changset lgtm, add a few comments where we should add tests. Can we also add an e2e test asserting blocks are produced without the subblocks system tx?

crates/payload/builder/src/lib.rs
```diff
@@ -127,9 +123,14 @@ impl<Provider: ChainSpecProvider<ChainSpec = TempoChainSpec>> TempoPayloadBuilde
     /// - Subblocks signatures - validates subblock signatures
     fn build_seal_block_txs(
         &self,
-        block_env: &BlockEnv,
+        evm: &TempoEvm<impl Database>,
         subblocks: &[RecoveredSubBlock],
     ) -> Vec<Recovered<TempoTxEnvelope>> {
+        if subblocks.is_empty() && evm.cfg.spec.is_t4() {
```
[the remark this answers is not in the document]
klkvr: this is unreachable and subblocks service is disabled by default for a while now

PR #3749, opened 2026-04-28
chore: migrate to marshal standard inline wrapper

hamdiallam:
Instead of implementing the Automaton trait ourselves, marshal provides wrappers, `Inline` & `Deferred` which drive implements the trait and drives the application.

`Inline` preserves our existing behavior so we start with that


crates/commonware-node/src/consensus/application/mod.rs
```diff
@@ -1,75 +1,668 @@
...
+                subblocks
+                    .as_ref()
+                    .and_then(|s| s.get_subblocks(parent_hash).ok())
+                    .unwrap_or_default()
+            },
+        );
+
+        let interrupt_handle = attrs.interrupt_handle().clone();
+
+        // NOTE(hamdi): Revisit compared to the application actor which handled cancellation
```
hamdiallam: Rebased this PR, adapted real quick just to run the test suite. will revisit what the updated application actor was doing

crates/commonware-node/src/consensus/application/mod.rs
```diff
@@ -1,75 +1,758 @@
...
+            &dealer == proposer,
+            "proposer `{proposer}` is not the dealer `{dealer}` of the dealing \
+            in the block",
+        );
+    }
+
+    Ok(())
+}
+
+struct CancellablePayloadReceiver<E: Spawner> {
```
hamdiallam, line 674: @SuperFluffy @klkvr to review the updated cancel semantics

PR #3750, opened 2026-04-28
refactor:  simplify types with `FeeRoute` enum

0xKitsune:
This PR consolidates the `SwapInfo`, `FeeSwapPlan`, and `FeeSwapAttempt` logic into a simplified two-variant `FeeRoute` enum. It also removes the txpool’s `quote_token_cache`.

On the hot path, we can determine whether a direct pool swap is sufficient using `pool_cache`. The slow path remains fast enough that re-running it for T5+ two-hop admissions should be negligible.

Additionally, `plan_fee_swap` has been refactored into a read-only `get_fee_route` function that returns `Result<Option<FeeRoute>>`.

Behavior is consistent with the implementation in `rus/tip-1033`. The trade-off is that the cold path now re-reads the direct pool slot to warm `pool_cache`, introducing one additional in-memory read (for the cold path only).

### DIFF
--- a/crates/precompiles/src/tip_fee_manager/amm.rs
+++ b/crates/precompiles/src/tip_fee_manager/amm.rs
@@ -86,128 +86,17 @@ impl PoolKey {
     }
 }
 
-/// Result of a per-pool liquidity check used to build a [`FeeSwapPlan`].
-#[derive(Debug, Clone, Copy, PartialEq)]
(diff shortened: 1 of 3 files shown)


0xrusowsky:
closing in favor of https://github.com/tempoxyz/tempo/pull/3758

thanks for pushing for a simpler design!

PR #3751, opened 2026-04-28
fix: mitigate race condition in payload cancellation

klkvr:
https://github.com/tempoxyz/tempo/issues/3451 added logic to resolve/cancel payload job upon cancellation. however, the current implementation is a subject to a race condition in cases when proposal is interrupted after dispatching FCU but before the FCU completes (and starts the payload build). In that case `resolve_kind_fut` would return `None` because the payload build did not yet begin, and the payload job would still get spawned and will run until deadline.

This PR fixes it by maintaining shared state between `handle_propose` and `propose` that is used to properly await the FCU response if cancellation happens while it's still in-flight

### DIFF
--- a/crates/commonware-node/src/consensus/application/actor.rs
+++ b/crates/commonware-node/src/consensus/application/actor.rs
@@ -32,14 +32,16 @@ use commonware_runtime::{
 };
 use prometheus_client::metrics::counter::Counter;
 
-use commonware_utils::{SystemTimeExt, channel::oneshot};
+use commonware_utils::SystemTimeExt;
 use eyre::{OptionExt as _, WrapErr as _, bail, ensure, eyre};
-use futures::{StreamExt as _, TryFutureExt as _, channel::mpsc, future::try_join};
+use futures::{
+    StreamExt as _, TryFutureExt as _,
+    channel::{mpsc, oneshot},
+    future::try_join,
+};
 use rand_08::{CryptoRng, Rng};
 use reth_ethereum::chainspec::EthChainSpec as _;
-use reth_node_builder::{
-    Block as _, BuiltPayload, ConsensusEngineHandle, PayloadAttributes, PayloadKind,
-};
+use reth_node_builder::{Block as _, BuiltPayload, ConsensusEngineHandle, PayloadKind};
 use tempo_chainspec::{TempoChainSpec, hardfork::TempoHardforks as _};
 use tempo_dkg_onchain_artifacts::OnchainDkgOutcome;
 use tempo_node::{TempoExecutionData, TempoFullNode, TempoPayloadTypes};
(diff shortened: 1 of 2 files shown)


crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -598,18 +615,29 @@ impl Inner<Init> {
...
+            .as_mut()
+            .expect("just set")
             .await
+            .wrap_err("executor dropped response")?
             .wrap_err("failed requesting a new payload build")?;
 
+        // Once we hold `payload_id`, the cancel branch no longer needs the rx;
+        // it can cancel directly by id.
+        *payload_id_rx = None;
+        *payload_id_slot = Some(payload_id);
```
SuperFluffy: Why not just always push the ID into the channel? We don't need the extra `payload_id_slot`.

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -414,13 +426,18 @@ impl Inner<Init> {
...
+    )]
     async fn propose<TContext: Pacer>(
         self,
         context: TContext,
         parent_view: View,
         parent_digest: Digest,
         round: Round,
-        payload_id_tx: oneshot::Sender<PayloadId>,
+        payload_id_rx: &mut Option<oneshot::Receiver<eyre::Result<PayloadId>>>,
+        payload_id_slot: &mut Option<PayloadId>,
```
SuperFluffy: I see why you did that, but IMO remove this slot to make this all a bit easier.

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -598,18 +615,29 @@ impl Inner<Init> {
...
-            .executor
-            .canonicalize_and_build(parent.height(), parent.digest(), attrs)
+        // Share the dispatch receiver with the cancel branch so that, if cancellation
+        // hits between dispatch send and receiving `payload_id`, the cancel branch can
+        // still drain the rx, learn `payload_id`, and cancel the now-registered job.
+        *payload_id_rx = Some(self.state.executor.canonicalize_and_build(
+            parent.height(),
+            parent.digest(),
+            attrs,
+        )?);
```
SuperFluffy, line 614: context

crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -311,17 +313,26 @@ impl Inner<Init> {
             leader,
         } = request;
 
-        let (payload_id_tx, mut payload_id_rx) = oneshot::channel();
+        let mut payload_id_rx: Option<oneshot::Receiver<eyre::Result<PayloadId>>> = None;
+        let mut payload_id: Option<PayloadId> = None;
```
SuperFluffy: Just retain the channel IMO.

SuperFluffy (approved):
Looks fine to me, but IMO just keep the channel. Once it is resolved inside `fn propose`, you can just do
```rust
let (tx, rx) = oneshot::channel();
*payload_id_rx = rx;
tx.send(payload);
```

PR #3766, opened 2026-04-30
fix(p2p-proxy): cap header request expansion

mattsse:
Caps `GetBlockHeaders` number expansion to `HEADER_BATCH_SIZE` before building the request vector, preventing peers from forcing unbounded memory growth with oversized limits.

Validated with `cargo check -p tempo --bin tempo`.


fgimenez (approved):
makes sense

PR #3770, opened 2026-04-30
Strip blockTimestamp from transaction RPC responses

emmajam:
## Summary
- Clear Alloy/Reth's optional transaction block timestamp in Tempo's RPC transaction converter
- Keep receipt log block timestamps unchanged
- Add a regression assertion for eth_getTransactionByHash responses

## Why
run-viem-tests started failing because viem snapshots for getTransaction received a volatile blockTimestamp field on transaction responses.

## Test plan
- cargo fmt --check
- cargo check -p tempo-node
- cargo test -p tempo-node --test it eth_transactions::test_get_transaction_by_sender_and_nonce -- --exact


mattsse (commented):
I think we can keep this as is

PR #3774, opened 2026-04-30
feat: default to enabling discv5 on the same port as discv4

klkvr:


bin/tempo/src/main.rs
```diff
@@ -562,13 +562,6 @@ fn main() -> eyre::Result<()> {
         } = builder
             .node(TempoNode::new(&args.node_args, validator_key))
             .apply(|mut builder: WithLaunchContext<_>| {
-                // Enable discv5 peer discovery
-                builder
-                    .config_mut()
-                    .network
-                    .discovery
-                    .enable_discv5_discovery = true;
-
```
klkvr, line 571: this is now true by default

mattsse (approved):
very nice

emmajam (approved):
lets go!

PR #3779, opened 2026-04-30
feat(tip-1034): derive channel ids from open tx hash

legion2002:
Makes TIP-20 channel escrow stateless on close by including the opening transaction hash in channel ID derivation and deleting terminal state instead of keeping a tombstone.

This threads the top-level tx hash through revm and the precompile storage context, updates the TIP-1034 spec and Solidity reference artifacts, and adds regression coverage for channel reopen and tx-hash scoping behavior.

Verification:
- cargo test -p tempo-precompiles
- cargo test -p tempo-revm


legion2002:
Closed in favour of #3837

PR #3784, opened 2026-04-30
refactor: offload some of the keychain checks to stateless validation

klkvr:
`validate_against_state_and_deduct_caller` performs keychain validation which is qutie hard to keep track of partially because of mixing stateless and stateful checks

this PR moves keychain-related checks that do not require access to state into `validate_env`, making `validate_against_state_and_deduct_caller` slightly simpler and allowing to reject transactions slightly earlier during validation

### DIFF
--- a/crates/revm/src/handler.rs
+++ b/crates/revm/src/handler.rs
@@ -1113,67 +1113,6 @@ where
         if let Some(tempo_tx_env) = tx.tempo_tx_env.as_ref()
             && let Some(key_auth) = &tempo_tx_env.key_authorization
         {
-            // Check if this TX is using a Keychain signature (access key)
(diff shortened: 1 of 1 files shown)


crates/revm/src/handler.rs
```diff
@@ -1233,16 +1172,6 @@ where
...
 
-                // Validate expiry is not in the past
-                let current_timestamp = block.timestamp().saturating_to::<u64>();
-                if expiry <= current_timestamp {
-                    return Err(TempoInvalidTransaction::AccessKeyExpiryInPast {
-                        expiry,
-                        current_timestamp,
-                    }
-                    .into());
-                }
```
klkvr, line 1244: this check is removed entirely as it's performed within `authorize_key` anyway

legion2002 (approved):
lgtm

PR #3786, opened 2026-04-30
fix(precompiles): bind system_transfer_from to msg.sender

legion2002:
Addresses the review feedback on PR #3219 by removing the caller-supplied source address from `system_transfer_from` and binding it to the external `msg.sender` captured at the precompile boundary.

This threads `msg.sender` through `StorageCtx`, updates the TIP-20 helper and AMM call sites, and adds regression coverage for the scoped sender context.

Validation: `cargo nextest run -p tempo-precompiles`

Linear: https://linear.app/tempoxyz/issue/CHAIN-1144/tip-1035-implicit-approvals-list

### DIFF
--- a/crates/precompiles/src/storage/thread_local.rs
+++ b/crates/precompiles/src/storage/thread_local.rs
@@ -21,6 +21,7 @@ use crate::{
 };
 
 scoped_thread_local!(static STORAGE: RefCell<&mut dyn PrecompileStorageProvider>);
+scoped_thread_local!(static MSG_SENDER: Address);
 
 /// Thread-local storage accessor that implements `PrecompileStorageProvider` without the trait bound.
 ///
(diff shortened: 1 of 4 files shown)


crates/precompiles/src/storage/thread_local.rs
```diff
@@ -21,6 +21,7 @@ use crate::{
 };
 
 scoped_thread_local!(static STORAGE: RefCell<&mut dyn PrecompileStorageProvider>);
+scoped_thread_local!(static MSG_SENDER: Address);
```
klkvr, line 24: hmm not sure i understand the idea here. all precompiles propagate `msg_sender` manually now so adding this introduces a bit of an inconsistency to how it's handled

also i don't think we need to keep it as a thread local variable, it could just be a part of `PrecompileStorageProvider` context i believe

legion2002: Yeah we want to remove manually propogating msg_sender to system transfer from, so we can have better guarantees about TIP 1035. 

We want `system_transfer_from`  to receive a trusted msg_sender, instead of trusting whatever the precompile passes to it.

PR #3790, opened 2026-05-01
refactor: charge fee before key authorization

klkvr:
Right now we invoke `collectFeePreTx` only after processing a key authorization which is a potential DOS vector because `authorizeKey` might perform a pretty much unbounded amount of i/o work to write all of the storage slots for its key which we don't know if user can pay for.

This PR changes order in `validate_against_state_and_deduct_caller` so that fee payment is done after validating keychain signature but before processing the key authorization.

To work around the fact that `collectFeePreTx` might need the current key to be authorized already during same tx auth+use case, we do the following:
1. Pre-validate that spending limit would cover the max fee spending. Don't authorize the key or set the transient key_id yet
2. Charge the fee — this step would decrement spending limit for already-authorized keys and just charge the fee directly from the caller for same tx auth+use
3. Authorize the key
4. If it's same tx auth+use, invoke `keychain.authorize_transfer`


crates/revm/src/handler.rs
```diff
@@ -1109,7 +1109,151 @@ where
...
+                    user_address: *user_address,
+                    caller: tx.caller,
+                }
+                .into());
+            }
+
+            if let Some(key_auth) = tempo_tx_env.key_authorization.as_ref() {
+                // If this is a same tx auth+use, validate that spending limit is enough to cover the fee.
+                //
+                // `collectFeePreTx` would not validate the spending limit because they key is not authorized yet and we are not setting the transient key_id.
```
```suggestion
                // `collectFeePreTx` would not validate the spending limit because the key is not authorized yet and we are not setting the transient key_id.
```

crates/revm/src/handler.rs
```diff
@@ -1244,152 +1385,35 @@ where
...
+                    journal,
+                    block,
+                    cfg,
+                    tx,
+                    |mut keychain: AccountKeychain| {
+                        keychain
+                            .set_transaction_key(key_auth.key_id)
+                            .map_err(|e| EVMError::Custom(e.to_string()))?;
+
+                        if evm.collected_fee.is_zero() {
```
legion2002, line 1401: this shouldn't really happen, because `collected_fee` should never be 0. but no harm in early returning.

legion2002 (approved):
good catch!

crates/revm/src/handler.rs
klkvr: it might be zero during gas estimation

PR #3791, opened 2026-05-01
docs(tips): TIP-1028, Address-Level Receive Policies

0xKitsune:
Supersedes #3173, rewrites the spec for readability without changing the protocol design.

This PR introduces TIP-1028, which extends TIP-403 with address-level receive policies that let a receiver control which TIP-20 tokens they accept and who can send them. When a transfer or mint is blocked by a receive policy, the funds are sent to escrow instead and can later be claimed by the receiver or a designated recovery contract.


tips/tip-1028.md
```diff
@@ -0,0 +1,517 @@
...
+Token filters must satisfy the following:
+
+- `filterType` is `WHITELIST` or `BLACKLIST`.
+- `COMPOUND` token filters are forbidden.
+- Filter type is immutable after creation.
+- Membership is mutable by the filter admin.
+
+`isTokenAllowed(tokenFilterId, token)` returns `false` for filter `0`, `true` for filter `1`, and otherwise reads the stored membership bit and applies it according to the filter's `filterType`.
+
+## Sender Policies
```
0xrusowsky: this should go before tokens sets or inside receiver policy evaluation

tips/tip-1028.md
```diff
@@ -0,0 +1,517 @@
...
+
+### Escrow Address
+
+```solidity
+address constant ESCROW_ADDRESS = 0xE5C0000000000000000000000000000000000000;
+```
+
+The blocked balance for each TIP-20 token sits in that token's `balances[ESCROW_ADDRESS]` slot.
+
+### Restrictions on `ESCROW_ADDRESS`
```
0xrusowsky: should also mention that escrow address doesn't accrue tip20 rewards

tips/tip-1028.md
```diff
@@ -0,0 +1,517 @@
...
+The following restrictions apply:
+
+- A TIP-20 transfer or mint with `to == ESCROW_ADDRESS` MUST revert with `EscrowAddressReserved()`. This applies to `transfer`, `transferFrom`, `transferWithMemo`, `transferFromWithMemo`, `systemTransferFrom`, `mint`, and `mintWithMemo`.
+- A reroute claim with `to == ESCROW_ADDRESS` MUST revert.
+- `setReceivePolicy(...)` MUST reject `account == ESCROW_ADDRESS` and any `recoveryContract == ESCROW_ADDRESS`.
+- Any TIP-20 logic that protects DEX or FeeManager balances as system balances MUST extend the same protection to `ESCROW_ADDRESS`.
+
+The first zero-to-nonzero write to `balances[ESCROW_ADDRESS]` for a token can add roughly 250,000 gas to the first blocked transfer or mint. TIP-20 implementations SHOULD move that cost to deployment by initializing the slot when the token is created. One acceptable pattern is an implementation-private escrow reserve created at deployment that is not claimable by users or recovery contracts and is preserved by release and burn logic.
+
+### Escrow Model
```
0xrusowsky: should further explain that requires the full info to be able to apply arbitrary rules thanks to the witness provided at claim time.

imo requires further ctx for the reader so that they can understand why this is needed. i would introduce some of the stuff in the following sections about storing blocked transfers/mints so that it is easier to understand later

tips/tip-1028.md
```diff
@@ -0,0 +1,517 @@
...
+
+TIP-1028 introduces three main changes:
+
+- The TIP-403 precompile is extended to add per-address receive policies that define which tokens and senders are allowed to send to a given address. Receive policies are configured via newly added `setReceivePolicy(...)` and enforced via `validateReceivePolicy(...)` functions.
+- A new escrow precompile is introduced that holds blocked funds at `ESCROW_ADDRESS` and records a receipt for each blocked transfer or mint, which can be claimed later by the receiver or a recovery contract.
+- TIP-20 transfer and mint flows are updated to check receive policies before crediting a receiver. If allowed, the transfer proceeds as normal. If blocked, the funds are sent to the escrow precompile and a receipt is recorded.
+
+The sequence diagram below shows the high level flow for a transfer that is blocked by a receive policy and later claimed from escrow.
+
+```mermaid
```
0xrusowsky, line 52: i think it's important to state that the TIP20 and the token-level 403 policies are still perform and that those checks revert on failure.

0xrusowsky (commented):
i think this progressive narrative describing components incrementally works way better, nice

tips/tip-1028.md
0xKitsune: Do we want to explicitly reject this? Would allowing it break any assumptions?

It feels more like a user error that can be reversed, but otherwise expected behavior? This would be the same as a user setting their rewards address as an address they do not have access to. Lmk if I am overlooking something here though.

tips/tip-1028.md
0xKitsune: Addressed in https://github.com/tempoxyz/tempo/pull/3791/commits/4d8de8e7b5d10bb1c8d504730ec3d54e65fb42b4

tips/tip-1028.md
0xKitsune: Moved in https://github.com/tempoxyz/tempo/pull/3791/commits/13b7f07e66a8617498f2f71a91b88ab6bb40e2cb

tips/tip-1028.md
0xKitsune: Updated the spec to address. Commit for ref:
 https://github.com/tempoxyz/tempo/pull/3791/commits/fbba6b9d4e37c2e958e1ad2ff6aaf39540275ff4

tips/tip-1028.md
malleshpai: no strong opinion. if it's easy enough i would say escrow can't receive rewards, just so that we're minimizing chances of unclaimable rewards floating around?

tips/tip-1028.md
```diff
@@ -0,0 +1,543 @@
...
+
+The receiver or a designated recovery contract can later claim these funds. Claims back to the receiver unwind the original inbound while claims to other addresses are treated as new transfers.
+
+TIP-403 authorization checks are unchanged and continue to revert on failure. TIP-1028 policies only apply to TIP-20 precompile flows. Ordinary contracts and other precompiles are unaffected.
+
+<br>
+
+## Motivation
+
+TIP-403 allows token issuers to control who may use a token, but it does not give receivers control over which transfers or mints they accept. Some applications require receivers to restrict incoming funds based on the sender or the token.
```
malleshpai:
```suggestion
TIP-403 allows token issuers to control who may use a token, but it does not give receivers control over which transfers or mints they accept. However, receivers may wish to restrict incoming funds based on the sender or the token. For example:

- a regulated entity might wish to receive incoming transfers only from addresses held by individuals they have KYC'ed, 
- orchestrators  and exchanges may wish to receive only specific tokens at their deposit addresses to prevent unrecoverable funds.

```

tips/tip-1028.md
```diff
@@ -0,0 +1,543 @@
...
+
+If a receiver rejects transfers, sending them tokens will fail. This can cause swaps, payouts, and other operations to revert, even though the sender and token are valid.
+
+TIP-1028 lets receivers filter incoming transfers without causing those transfers to fail. If a receiver blocks a transfer or mint, the call still succeeds and the funds are sent to escrow instead.
+
+The receiver (or a recovery contract) can claim those funds later. Each escrow entry keeps enough information to identify the original transfer, so it can be handled correctly offchain.
```
malleshpai:
```suggestion
The receiver (or a recovery contract designated by the receiver) can claim those funds later. Each escrow entry keeps enough information to identify the original transfer, so it can be handled correctly offchain.
```

tips/tip-1028.md
```diff
@@ -0,0 +1,543 @@
...
+
+If both checks pass, the transfer or mint proceeds as normal. If either check fails, the call still succeeds but the funds are sent to `ESCROW_ADDRESS` instead.
+
+Token level checks controlled by the issuer (`TIP-403` / `TIP-1015`) are unchanged and continue to revert on failure.
+
+Note that TIP-1028 only applies to the TIP-20 operations listed above. It does not apply to `approve`, `permit`, or `burn` and does not affect fee deposits or refunds via `transfer_fee_pre_tx` or `transfer_fee_post_tx`. Additionally, TIP-1028 does not interact with TIP-20 escrowed rewards or internal balances.
+
+### TIP-1022 Interaction
+
+If `to` is a TIP-1022 virtual address, it is resolved to its master address before any checks run. All receive policy checks use the master address. If resolution fails, the operation reverts.
```
malleshpai:
```suggestion
If `to` is a TIP-1022 virtual address, it is resolved to its master address before any checks run. All receive policy checks use the master address. If resolution fails, the operation reverts as before.
```

tips/tip-1028.md
```diff
@@ -0,0 +1,543 @@
...
+A receive policy defines which transfers and mints an address accepts. It controls two things:
+
+- Which TIP-20 tokens are allowed.
+- Which senders are allowed.
+
+Receive policies are stored per address in the TIP-403 registry. Each address can configure:
+
+- A TIP-403 policy (`receivePolicyId`) that defines which senders are allowed.
+- A token filter (`tokenFilterId`) that defines which TIP-20 tokens are allowed.
+- An optional `recoveryContract` that can claim blocked funds. If not set, the receiver claims directly.
```
malleshpai:
```suggestion
- An optional `recoveryContract` that can claim blocked funds. If not set, only the receiver can claim directly.
```

tips/tip-1028.md
```diff
@@ -0,0 +1,543 @@
...
+
+`recoveryContract` is stored separately because a 160-bit address does not fit in the packed config slot.
+
+An address that wants to functionally disable filtering SHOULD set `receivePolicyId = 1` and `tokenFilterId = 1`. The slot remains allocated.
+
+Constraints on `setReceivePolicy(...)`:
+
+- `receivePolicyId` MUST reference a simple `WHITELIST` or `BLACKLIST` TIP-403 policy, or built-in policy `0` or `1`. `COMPOUND` policies are not valid.
+- `tokenFilterId` MUST reference an existing token filter, or built-in `0` (reject all) or `1` (allow all).
+- `recoveryContract` MAY be `address(0)`. If nonzero, it MUST NOT equal `ESCROW_ADDRESS` and MUST NOT be a TIP-1022 virtual address.
```
malleshpai: Dumb question--- if we're protecting this set, should we also disallow it being other clearly wrong addresses (DEX, FEE AMM etc)?

tips/tip-1028.md
```diff
@@ -0,0 +1,543 @@
...
+- The caller MUST NOT be a TIP-1022 virtual address. Virtual addresses are forwarding aliases and configure receive policies on their resolved master address.
```
malleshpai:
```suggestion
- The caller MUST NOT be a TIP-1022 virtual address. Virtual addresses are forwarding aliases and the user should instead configure receive policies on their resolved master address.
```

tips/tip-1028.md
```diff
@@ -0,0 +1,543 @@
...
+
+The destination `to` decides whether the claim is an **unwind** back to the receiver or a **reroute** to a new address. The two cases differ only in what they enforce on the destination side.
+
+When `to == receiver`, the claim is an unwind of a previously authorized inbound. It bypasses the receiver's receive policy, token-level recipient authorization for the receiver, and AccountKeychain spending-limit metering. This is intentional. A blocked mint, for example, has already passed the token's original mint-recipient check on the inbound path. Releasing escrow back to the same receiver is not a new transfer and MUST NOT be rechecked against transfer-recipient authorization.
```
malleshpai:
```suggestion
When `to == receiver`, the claim is an unwind of a previously authorized inbound. It bypasses the receiver's receive policy, token-level recipient authorization for the receiver, and AccountKeychain spending-limit metering. This is intentional. A blocked mint, for example, has already passed the token's original mint-recipient check on the inbound path. Releasing escrow back to the same receiver is not a new transfer and MUST NOT be rechecked against the recipient's  receive policy.
```

tips/tip-1028.md
```diff
@@ -0,0 +1,543 @@
...
+
+When `to != receiver`, the claim is a reroute and is treated as a new spend by the receiver. The reroute MUST reject `to == ESCROW_ADDRESS` and TIP-1022 virtual addresses, enforce token-level transfer-recipient authorization for `to`, and enforce `to`'s receive policy against the consumed receipt's `originator`. If either destination check fails, the call MUST revert with `ClaimDestinationUnauthorized()`. If `recoveryContract == address(0)` and the reroute is initiated through an access key, the claim MUST meter the claimed amount against the receiver's AccountKeychain spending limit as an ordinary TIP-20 spend by the receiver. If the receiver uses a custom `recoveryContract`, any equivalent delegation, timelock, multisig, or key-policy enforcement is the responsibility of that contract.
```
malleshpai:
```suggestion
When `to != receiver`, the claim is a reroute and is treated as a new spend by the claimant. The reroute MUST reject `to == ESCROW_ADDRESS` and TIP-1022 virtual addresses, enforce token-level transfer-recipient authorization for `to`, and enforce `to`'s receive policy against the consumed receipt's `originator`. If either destination check fails, the call MUST revert with `ClaimDestinationUnauthorized()`. If `recoveryContract == address(0)` and the reroute is initiated through an access key, the claim MUST meter the claimed amount against the receiver's AccountKeychain spending limit as an ordinary TIP-20 spend by the receiver. If the receiver uses a custom `recoveryContract`, any equivalent delegation, timelock, multisig, or key-policy enforcement is the responsibility of that contract.
```

tips/tip-1028.md
```diff
@@ -0,0 +1,543 @@
...
+
+`storeBlocked(...)` MUST be callable only by TIP-20 precompiles or protocol-internal system code. User callers MUST NOT be able to fabricate receipts. Without this restriction, an attacker could mint synthetic receipts by replaying or fabricating witnesses without any backing escrow balance.
+
+### Claiming
+
+A claim consumes one full receipt and releases the funds to a single destination. `claimBlocked(...)` takes the receipt witness and a destination `to`, recomputes `receiptKey`, requires `blockedReceiptAmount[receiptKey] > 0`, deletes the slot, and releases the stored amount. Partial claims are not allowed. Claims MUST consume whole receipts.
```
malleshpai: should we clarify that consuming a receipit frees that slot? i see no reason to keep it around, and we should be storage friendly if we can be.

tips/tip-1028.md
malleshpai: also why are we disallowing virtual addresses here?

0xKitsune: Not strictly necessary, we could resolve virtual addresses on every receive policy check. The rationale is to push resolution to `setReceivePolicy(...)` time so `validateReceivePolicy(...)` can assume the receiver is already a master address and skip TIP-1022 resolution on every inbound.

tips/tip-1028.md
```diff
@@ -0,0 +1,547 @@
...
+
+<br>
+
+## Abstract
+
+TIP-1028 extends TIP-403 with address-level receive policies, letting a receiver control which TIP-20 tokens they accept and who can send them.
+
+When a receive policy blocks a TIP-20 transfer or mint, the operation still succeeds, but the funds go to `ESCROW_ADDRESS` instead of the receiver and an escrow receipt is recorded.
+
+The receiver or a designated recovery contract can later claim these funds. Claims back to the receiver unwind the original inbound while claims to other addresses are treated as new transfers.
```
samczsun: Just from the abstract, it's not intuitive what "unwinding the original inbound" means

tips/tip-1028.md
```diff
@@ -0,0 +1,547 @@
...
+TIP-403 authorization checks are unchanged and continue to revert on failure. TIP-1028 policies only apply to TIP-20 precompile flows. Ordinary contracts and other precompiles are unaffected.
+
+<br>
+
+## Motivation
+
+TIP-403 allows token issuers to control who may use a token, but it does not give receivers control over which transfers or mints they accept. However, receivers may wish to restrict incoming funds based on the sender or the token. For example:
+
+- a regulated entity might wish to receive incoming transfers only from addresses held by individuals they have KYC'ed, 
+- orchestrators  and exchanges may wish to receive only specific tokens at their deposit addresses to prevent unrecoverable funds.
```
samczsun:
```suggestion
- orchestrators and exchanges may wish to receive only specific tokens at their deposit addresses to prevent unrecoverable funds.
```

tips/tip-1028.md
```diff
@@ -0,0 +1,547 @@
...
+    TIP20-->>Receiver: transfer funds
+```
+
+<br>
+
+## TIP-20 Operations
+
+TIP-1028 applies to the following TIP-20 operations: `transfer`, `transferFrom`, `transferWithMemo`, `transferFromWithMemo`, `systemTransferFrom`, `mint`, `mintWithMemo`.
+
+For each of these operations, TIP-20 checks the receiver's policy configuration before crediting them. It verifies whether the token is allowed by the receiver's token filter and whether the sender is allowed by the receiver's policy. For `transfer`, `transferFrom`, `transferWithMemo`, `transferFromWithMemo`, and `systemTransferFrom`, the policy checks the `from` parameter as the sender. For `mint` and `mintWithMemo`, the policy checks `msg.sender` as the sender.
```
samczsun: For mints, should the sender be `msg.sender` or `address(0x00)`? The former gives you control over the individual triggering the mint, but the latter allows you to have the flexibility to allow mints while rejecting transfers. I'm not sure if this distinction actually matters, but it would bring the params in parity with the `Transfer` event, which might be abstractly desirable

tips/tip-1028.md
```diff
@@ -0,0 +1,547 @@
...
+| Bits | Size | Field |
+|---|---:|---|
+| `0` | 1 | `hasReceivePolicy` |
+| `1..64` | 64 | `receivePolicyId` |
+| `65..72` | 8 | `receivePolicyType` |
+| `73..136` | 64 | `tokenFilterId` |
+| `137..144` | 8 | `tokenFilterType` |
+| `145..255` | 111 | reserved, MUST be zero |
+
+When `hasReceivePolicy == 0`, the address has no receive policy and all transfers and mints are allowed. The cached type fields are valid because policy type and token filter type are immutable after creation.
```
samczsun, line 144: The naming here is a little ambiguous - does `hasReceivePolicy` refer to the tuple of `(receivePolicy, tokenFilter)` or just the first component of that tuple? In other words, if `hasReceivePolicy == 0`, can `tokenFilterId != 0`? I assume yes, in which case this should probably be named `hasReceiveConfig`

tips/tip-1028.md
```diff
@@ -0,0 +1,547 @@
...
+
+## Sender Policies
+
+A receive policy points at an existing TIP-403 policy through `receivePolicyId`. The sender side of receive checks reuses TIP-403 policy evaluation directly. A receiver can reuse an existing simple TIP-403 `WHITELIST` or `BLACKLIST` policy, create a new one through the existing TIP-403 interface, or use built-in policy `0` (reject all) or `1` (allow all).
+
+`COMPOUND` policies are not valid for `receivePolicyId`. A receive check only asks one question: may this sender send to this receiver. A `COMPOUND` policy splits authorization across sender, transfer-recipient, and mint-recipient roles. Allowing `COMPOUND` would conflate those roles.
+
+<br>
+
+## Token Filters
```
samczsun, line 170: Hot take - should token filters just be regular address lists? We're mirroring a data structure which, at the end of the day, does the same thing: stores a list of addresses. Whether the addresses are "senders" or "tokens" shouldn't really matter here, and potentially keeps the implementation clean

tips/tip-1028.md
```diff
@@ -0,0 +1,547 @@
...
+`validateReceivePolicy(token, sender, receiver)` returns whether a transfer or mint to `receiver` is allowed and, if not, why. If the receiver has not configured a receive policy (`hasReceivePolicy == 0`), it returns `(true, NONE)`. Otherwise it checks `token` against the receiver's token filter and `sender` against the receiver's receive policy, and returns one of `(true, NONE)` when both pass, `(false, TOKEN_FILTER)` when only the token filter rejects, `(false, RECEIVE_POLICY)` when only the receive policy rejects, or `(false, TOKEN_FILTER_AND_RECEIVE_POLICY)` when both reject.
+
+The `BlockedReason` values used in the second return slot are:
+
+```solidity
+enum BlockedReason {
+    NONE,
+    TOKEN_FILTER,
+    RECEIVE_POLICY,
+    TOKEN_FILTER_AND_RECEIVE_POLICY
```
samczsun: Should we just omit this? It means that policy evaluation can't short-circuit, and also creates an expectation that if we ever add more reasons in the future, we have to enumerate all permutations. Instead, maybe we define a canonical ordering for what is checked, and the blocked reason is defined as the first check which fails, or `NONE` if all succeed.

tips/tip-1028.md
```diff
@@ -0,0 +1,547 @@
...
+
+### Escrow Address
+
+```solidity
+address constant ESCROW_ADDRESS = 0xE5C0000000000000000000000000000000000000;
+```
+
+The blocked balance for each TIP-20 token sits in that token's `balances[ESCROW_ADDRESS]` slot.
+
+### Restrictions on `ESCROW_ADDRESS`
```
samczsun: What happens if the token admin sets a 403 deny policy on the escrow contract?

tips/tip-1028.md
```diff
@@ -0,0 +1,547 @@
...
+
+The following restrictions apply:
+
+- A TIP-20 transfer or mint with `to == ESCROW_ADDRESS` MUST revert with `EscrowAddressReserved()`. This applies to `transfer`, `transferFrom`, `transferWithMemo`, `transferFromWithMemo`, `systemTransferFrom`, `mint`, and `mintWithMemo`.
+- A reroute claim with `to == ESCROW_ADDRESS` MUST revert.
+- `setReceivePolicy(...)` MUST reject `account == ESCROW_ADDRESS` and any `recoveryContract == ESCROW_ADDRESS`.
+- Any TIP-20 logic that protects DEX or FeeManager balances as system balances MUST extend the same protection to `ESCROW_ADDRESS`.
+
+The first zero-to-nonzero write to `balances[ESCROW_ADDRESS]` for a token can add roughly 250,000 gas to the first blocked transfer or mint. TIP-20 implementations SHOULD move that cost to deployment by initializing the slot when the token is created. One acceptable pattern is an implementation-private escrow reserve created at deployment that is not claimable by users or recovery contracts and is preserved by release and burn logic.
```
samczsun: This feels really hacky, imagine telling a stablecoin issuer that you have to permanently burn 1 unit of your token because your users might hit unexpected gas charges. Can we just define the TIP20 initializer to charge an extra cold sstore and then never charge cold writes on escrow transfers?

tips/tip-1028.md
```diff
@@ -0,0 +1,547 @@
...
+
+### Escrow Model
+
+When a transfer or mint is blocked, the funds are credited to `ESCROW_ADDRESS` instead of the receiver. The escrow precompile records one receipt per blocked transfer or mint. Each receipt captures the full context of the blocked operation, including the original sender, the requested recipient, whether it was a transfer or mint, the memo, and the reason for blocking. This gives a claimer (or a recovery contract acting on the receiver's behalf) enough information to apply arbitrary rules when deciding whether and how to claim.
+
+The receipt is identified by a `receiptKey` derived from these fields. The escrow precompile only stores the keyed amount per receipt. The other fields are emitted in the blocked event when the receipt is created, and the claimer supplies them again as a witness at claim time. This keeps onchain state minimal while letting recovery logic stay flexible.
```
samczsun: The downside is that it's impossible for a smart contract to semi-automate sweeping blocked funds. There will always need to be some off-chain component to maintain state and submit transactions

tips/tip-1028.md
```diff
@@ -0,0 +1,547 @@
...
+uint64 public blockedReceiptNonce = 1;
+mapping(bytes32 => uint256) internal blockedReceiptAmount;
+```
+
+The persistent escrow key for a blocked transfer or mint is:
+
+```text
+receiptKey = keccak256(
+    abi.encode(
+        receiptVersion,
```
samczsun: Maybe we should put the version outside, that way it's possible to change the inner struct without too much hassle

tips/tip-1028.md
```diff
@@ -0,0 +1,547 @@
...
+        blockedNonce
+    )
+)
+```
+
+where:
+
+- `receiptVersion`: one-byte bucketing tag. MUST be `1` for receipts created under this TIP. Future receipt-key formats MUST use a different value.
+- `token`: the TIP-20 token whose balance ledger holds the blocked amount at `ESCROW_ADDRESS`.
+- `receiver`: the canonical TIP-20 holder that owns the receipt. For TIP-1022 inbounds this is the resolved master address.
```
samczsun: I feel like we should just store the `requestedRecipient` (and call it `to` or `recipient`) - virtual address resolution is a deterministic algorithm and it doesn't hurt off-chain apps to make an extra call to perform that resolution (they already have to track chain state to load the receipt), and keeps the receipt simple

tips/tip-1028.md
```diff
@@ -0,0 +1,547 @@
...
+- `originator`: `from` for transfers, `msg.sender` for mints.
+- `requestedRecipient`: the literal `to` supplied at the TIP-20 entrypoint. For non-virtual inbounds, `requestedRecipient == receiver`.
+- `recoveryContract`: the receiver's recovery contract at the time the receipt was created, or `address(0)`.
+- `blockedReason`: `TOKEN_FILTER`, `RECEIVE_POLICY`, or `TOKEN_FILTER_AND_RECEIVE_POLICY`.
+- `kind`: `TRANSFER` or `MINT`.
+- `memo`: original memo for memo-bearing paths, `bytes32(0)` otherwise.
+- `blockedAt`: block timestamp captured at receipt creation.
+- `blockedNonce`: monotonically increasing global disambiguator assigned at receipt creation.
```
samczsun, line 275: Why do we need a nonce?

tips/tip-1028.md
```diff
@@ -0,0 +1,547 @@
...
+
+When `to != receiver`, the claim is a reroute and is treated as a new spend by the claimant. The reroute MUST reject `to == ESCROW_ADDRESS` and TIP-1022 virtual addresses, enforce token-level transfer-recipient authorization for `to`, and enforce `to`'s receive policy against the consumed receipt's `originator`. If either destination check fails, the call MUST revert with `ClaimDestinationUnauthorized()`. If `recoveryContract == address(0)` and the reroute is initiated through an access key, the claim MUST meter the claimed amount against the receiver's AccountKeychain spending limit as an ordinary TIP-20 spend by the receiver. If the receiver uses a custom `recoveryContract`, any equivalent delegation, timelock, multisig, or key-policy enforcement is the responsibility of that contract.
```
samczsun: No claiming to other virtual addresses?

tips/tip-1028.md
samczsun: Now it's a little clearer, maybe instead of unwind, `Claims back to the receiver resume the original transfer`?

0xKitsune: Sgtm, will update to this.

tips/tip-1028.md
0xKitsune: I don’t see a strong justification for explicitly preventing `recoveryContract` from being set to `ESCROW_ADDRESS`, so we could remove this restriction.

If we start enumerating “invalid” addresses (DEX, fee AMM, precompiles, predeploys, etc.), we may be taking on ongoing maintenance where every new system address could cause consideration to update this list.

Unless there’s a concrete case we are protecting against, it seems cleaner to drop this provision. If you agree, I can update accordingly.

tips/tip-1028.md
0xKitsune: This makes sense, addressed in https://github.com/tempoxyz/tempo/pull/3791/commits/45e8f40976468d9980856d3af87d7753b3f92c3a

tips/tip-1028.md
0xKitsune: Updated in https://github.com/tempoxyz/tempo/pull/3791/commits/cd537c84c8691457a2ba382127a9aa1ab4005c16

tips/tip-1028.md
0xKitsune: Yeah we can update this. Wdyt about simply updating  to `senderPolicyId`? 

```
ReceivePolicy(account) = (
    senderPolicyId,      // TIP-403 policy ref indicating which senders are allowed
    tokenFilterId,       // token filter, indicating which TIP-20 tokens are allowed
    recoveryContract   
)
```

tips/tip-1028.md
0xKitsune: @malleshpai any thoughts here?

tips/tip-1028.md
0xKitsune: Sgtm, addressed in https://github.com/tempoxyz/tempo/pull/3791/commits/f5a7f90a78b99e0086b39df2fbda221b31a4d5f2

tips/tip-1028.md
0xKitsune: Sgtm, updated in https://github.com/tempoxyz/tempo/pull/3791/commits/d4432aad9ff73e43ffa97363bc4331737adfc82a

tips/tip-1028.md
malleshpai: yes i think maybe we can remove it right now.

tips/tip-1028.md
malleshpai: I defer to @0xKitsune on this. I'm ok with removing it and just having it be a specific instance of a 403 list (where all entries are tip-20 tokens)

tips/tip-1028.md
malleshpai: yeah this was a hard tradeoff we needed to make. the context is that every new slot is 250k gas, and we don't want a transfer to be much more than that in the worst case. storing the blocked context onchain needed 2-3 slots.

tips/tip-1028.md
malleshpai: suppose the same block has multiple blocked transfers with the same other entries (e.g. same sender, no memo, transfer, etc etc). the nonce ensures uniqueness

tips/tip-1028.md
malleshpai: The reason for blocked reason to be in there is something like: sometimes i get sent the wrong token (USDT instead of USDC) and sometimes I get stuff from blocked addresses. In the latter case I don't want this claimable by sender, but in the former case i'm ok with it. 

This can be captured with what sam suggests so i defer to you @0xKitsune on what's better for implementation

tips/tip-1028.md
malleshpai: transaction reverts i believe

tips/tip-1028.md
malleshpai: @0xKitsune i defer to you on this

tips/tip-1028.md
malleshpai: making it address(0) would make all mints indistinguishable at the policy layer. it would therefore make impossible things like this: i want no association with address A. in response address A creates token B, and mints B into my account. 

my suggestion would be to keep this one as is

tips/tip-1028.md
malleshpai: yeah i agree I think we should just allow reroute to a virtual address.

tips/tip-1028.md
0xKitsune: Sounds good removed with https://github.com/tempoxyz/tempo/pull/3791/commits/98cc0611f5bf49f0debb9070cffc25f5e6f29f59

tips/tip-1028.md
0xKitsune: Updated in https://github.com/tempoxyz/tempo/pull/3791/commits/0c592550a534b6cd681e5c951b896bd268039645

tips/tip-1028.md
0xKitsune: In the current iteration of the spec, if a token’s TIP-403 policy blacklists `ESCROW_ADDRESS` as a sender, claim release still succeeds. The spec mentions the release path “bypasses the token-level TIP-403 sender check for ESCROW_ADDRESS.”

If the token issuer blacklists `ESCROW_ADDRESS` as a recipient, the spec does not explicitly say “bypass recipient check for escrow,” but the blocked path credits `ESCROW_ADDRESS` only after original token-level TIP-403/TIP-1015 checks have already passed and the receive policy should have failed.

tips/tip-1028.md
0xKitsune: Sgtm, updated in https://github.com/tempoxyz/tempo/pull/3791/commits/1b99d3c3add60e821ef2e143e58211cf0e2cc0d2

tips/tip-1028.md
0xKitsune: Updated in https://github.com/tempoxyz/tempo/pull/3791/commits/7abe64662920125558023ddd45eae4e488fdcecb

tips/tip-1028.md
0xKitsune: Updated to simplify to just use `TOKEN_FILTER` and `RECEIVE_POLICY`.

 Ref: https://github.com/tempoxyz/tempo/pull/3791/commits/88c85b2805c9befc4ac3fa890fdc54c210b2f242

tips/tip-1028.md
```diff
@@ -0,0 +1,569 @@
...
+
+The sequence diagram below shows the high level flow for a transfer that is blocked by a receive policy and later claimed from escrow.
+
+```mermaid
+sequenceDiagram
+    participant Receiver
+    participant Sender
+    participant TIP20
+    participant TIP403
+    participant Escrow
```
0xrusowsky: nit: personally i think the diagram is more comprehensive with sender and receiver being on opposite positions.

i'd either do:
`sender, tip20, tip403, escrow, receiver`

or:
`receiver, escrow, tip403, tip20, sender`

tips/tip-1028.md
0xrusowsky: i think sam's proposal makes sense. We can get rid of `TokenFilterData` and use `PolicyData` instead, with an extra validation to ensure the addresses in the set are TIP20 tokens.

malleshpai: in that case do we even need to check addresses are tip20 tokens? that's just user error, it's basically your (user's problem) to make sure your tokenlist is indeed a list of tokens.

tips/tip-1028.md
0xKitsune: The receive policy check already used `msg.sender` for mints, I also updated the blocked receipt originator to use `msg.sender` for mints so escrow claims preserve the same `mint` operator identity instead of collapsing mints to `address(0)`.


Ref: https://github.com/tempoxyz/tempo/pull/3791/commits/f361c33b41f2f799e5ca787ea10bb80e441344cf

tips/tip-1028.md
0xKitsune: Addressed in https://github.com/tempoxyz/tempo/pull/3791/commits/f9270a69e7259519b2ba66742b4ae3d1719f4198

tips/tip-1028.md
0xKitsune: Updated to simplify to use existing 403 policies rather than new token lists.



Ref: https://github.com/tempoxyz/tempo/pull/3791/commits/3fe294a9ef12b7b51bf9cd310545c3275eda1e64

tips/tip-1028.md
```diff
@@ -0,0 +1,495 @@
...
+
+<br>
+
+## Abstract
+
+TIP-1028 extends TIP-403 with address-level receive policies, letting a receiver control which TIP-20 tokens they accept and who can send them.
+
+When a receive policy blocks a TIP-20 transfer or mint, the operation still succeeds, but the funds go to `ESCROW_ADDRESS` instead of the receiver and an escrow receipt is recorded.
+
+The receiver or a designated recovery contract can later claim these funds. Claims back to the receiver resume the original transfer while claims to other addresses are treated as new transfers.
```
0xalpharush: Is the distinction b/w "original" vs "new" important e.g. distinct events? IIUC the important piece is it's optional and a pull-based model.

tips/tip-1028.md
```diff
@@ -0,0 +1,495 @@
...
+
+```text
+ReceivePolicy(account) = (
+    senderPolicyId,    // TIP-403 policy ref indicating which senders are allowed
+    tokenFilterId,     // TIP-403 policy ref indicating which TIP-20 tokens are allowed
+    recoveryContract
+)
+```
+
+`recoveryContract` is optional. If not set, only the receiver can claim blocked funds directly.
```
0xalpharush: Could clarify the behavior upon setting `recoveryContract` i.e. whether both can claim or its mutally exclusive

0xalpharush: Later on it's clarified
> `addressRecoveryContract[account]` stores the recovery contract for the address. If it is `address(0)`, the receiver claims blocked funds directly.

tips/tip-1028.md
0xalpharush: Ah
> Releasing escrow back to the same receiver is not a new transfer and MUST NOT be rechecked against the recipient's receive policy.

> When `to != receiver`, the claim is a reroute and is treated as a new spend by the claimant. The reroute MUST reject `to == ESCROW_ADDRESS`.

tips/tip-1028.md
```diff
@@ -0,0 +1,495 @@
...
+
+TIP-1028 introduces three main changes:
+
+- The TIP-403 precompile is extended to add per-address receive policies that define which tokens and senders are allowed to send to a given address. Receive policies are configured via newly added `setReceivePolicy(...)` and enforced via `validateReceivePolicy(...)` functions.
+- A new escrow precompile is introduced that holds blocked funds at `ESCROW_ADDRESS` and records a receipt for each blocked transfer or mint, which can be claimed later by the receiver or a recovery contract.
+- TIP-20 transfer and mint flows are updated to check receive policies before crediting a receiver. The existing TIP-20 checks (pause, balance, allowance) and the token-level TIP-403 / TIP-1015 checks still run first and continue to revert on failure. Only a receive policy failure diverts the funds to the escrow precompile, where a receipt is recorded.
+
+The sequence diagram below shows the high level flow for a transfer that is blocked by a receive policy and later claimed from escrow.
+
+```mermaid
```
0xrusowsky, line 52: i think we should either have a plain text section or enhance the mermaid diagram so that users can easily see the emitted events and how tokens move.

```
// REGULAR TRANSFER - HAPPY PATH (NON-BLOCKED)
1. Transfer(from, to, amount)

// VIRTUAL TRANSFER - HAPPY PATH
1. Transfer(from, virtual, amount)
2. Transfer(virtual, master, amount)

// REGULAR TRANSFER - BLOCKED
1. Transfer(from, ESCROW, amount)
2. TransferBlocked(...)

// VIRTUAL TRANSFER - BLOCKED
1. Transfer(from, ESCROW, amount)
2. TransferBlocked(...)
```

and the second leg when claiming would also be useful i guess

tips/tip-1028.md
```diff
@@ -0,0 +1,495 @@
...
+        bytes32 memo
+    ) external returns (uint64 blockedNonce, uint64 blockedAt);
+}
+```
+
+<br>
+
+## Invariants
+
+- For every TIP-20 token, `balances[ESCROW_ADDRESS]` equals exactly the sum of `blockedReceiptAmount[receiptKey]` over all open receipts for that token.
```
0xrusowsky:
```suggestion
- For every TIP-20 token, `balances[ESCROW_ADDRESS]` equals exactly the sum of `blockedReceiptAmount[receiptKey]` over all open receipts for that token.
- Unlike token-level (TIP-403 and TIP-1015) policies, address-level inbound policies CANNOT make transactions revert.
- Only the allowed claimer —address owner (when `recoveryContract = address(0)`) or `recoveryContract`— can claim the funds from the ESCROW precompile
```

tips/tip-1028.md
0xrusowsky: proposal: https://github.com/tempoxyz/tempo/pull/3815

malleshpai (approved):

tips/tip-1028.md
0xKitsune: Updated 3815 and merged

0xrusowsky (approved):

tips/tip-1028.md
```diff
@@ -0,0 +1,506 @@
...
+The receiver, the originator, or a recovery authority designated by the receiver can claim those funds later. Each escrow entry keeps enough information to identify the original transfer, so it can be handled correctly.
+
+<br>
+
+# Specification
+
+TIP-1028 introduces three main changes:
+
+- The TIP-403 precompile is extended to add per-address receive policies that define which tokens and senders are allowed to send to a given address. Receive policies are configured via newly added `setReceivePolicy(...)` and enforced via `validateReceivePolicy(...)` functions.
+- A new escrow precompile is introduced that holds blocked funds at `ESCROW_ADDRESS` and records a receipt for each blocked transfer or mint, which can be claimed later by the receiver, the originator, or a recovery authority.
```
jenpaff: would add a 1-liner why this design was chosen either here or in the section below

tips/tip-1028.md
```diff
@@ -0,0 +1,506 @@
...
+    TIP20-->>Receiver: transfer funds
+```
+
+<br>
+
+## TIP-20 Operations
+
+TIP-1028 applies to the following TIP-20 operations: `transfer`, `transferFrom`, `transferWithMemo`, `transferFromWithMemo`, `systemTransferFrom`, `mint`, `mintWithMemo`.
+
+For each of these operations, TIP-20 checks the receiver's policy configuration before crediting them. It verifies whether the token is allowed by the receiver's token filter and whether the sender is allowed by the receiver's policy. For `transfer`, `transferFrom`, `transferWithMemo`, `transferFromWithMemo`, and `systemTransferFrom`, the policy checks the `from` parameter as the sender. For `mint` and `mintWithMemo`, the policy checks `msg.sender` as the sender.
```
jenpaff: How much overhead is the additional check? Do we have before/after benches?

tips/tip-1028.md
```diff
@@ -0,0 +1,506 @@
...
+        bytes32 memo
+    ) external returns (uint64 blockedNonce, uint64 blockedAt);
+}
+```
+
+<br>
+
+## Invariants
+
+- For every TIP-20 token, `balances[ESCROW_ADDRESS]` equals exactly the sum of `blockedReceiptAmount[receiptKey]` over all open receipts for that token.
```
jenpaff: is that the only invariant?

malleshpai: Yes. @0xKitsune can comment further but he argued (and I agree) that we've conflated invariants with test cases (and what a lot of our tips contain are really test cases).

tips/tip-1028.md
malleshpai: Done here: [3006633](https://github.com/tempoxyz/tempo/pull/3791/commits/30066331199ff345bc8ea22478dfa1ab74cc7cac)

tips/tip-1028.md
```diff
@@ -500,6 +500,14 @@ interface IBlockedInboundEscrow {
...
+
+- reverting the transaction that effects a send which is disallowed by the receiver's policy, and
+- bouncing back a transfer when disallowed by the receiver's policy.
+
+While these would be conceptually simpler, they raise several issues. For example, the standard ERC-20 (and by extension TIP-20) model is that when the sender has sufficient balance and the parties are allowed by the token issuer, a transfer succeeds. As a result, several smart contracts may break in unexpected ways from reverts when failing receive policy. Similarly a single receiver setting a receive policy may invalidate a batch transfer, e.g., a payroll run. Bouncebacks suffer from similar issues.  Having the transfer succeed but as a new receipt in the blocked transfer smart contract rather than the address intended by the sender is thus both less disruptive, and also unlocks additional functionality.  for example, as described above, a regulated party can individually screen, off-chain, every blocked transfer and then proceed to claim only the ones that succeed this screening.
 
 ## Invariants
```
jenpaff, line 538: Can we add a compatibility section to highlight if any existing tooling/partners are affected?

0xrusowsky: what is expected to go there? the TIP won't have any breaking changes, it is add-only

tips/tip-1028.md
```diff
@@ -0,0 +1,514 @@
...
+
+## Abstract
+
+TIP-1028 extends TIP-403 with address-level receive policies, letting a receiver control which TIP-20 tokens they accept and who can send them.
+
+When a receive policy blocks a TIP-20 transfer or mint, the operation still succeeds, but the funds go to `ESCROW_ADDRESS` instead of the receiver and an escrow receipt is recorded.
+
+The receiver also designates who can later claim transfers blocked by the receiver: this could be the receiver theselves, the originator, or a third party recovery authority. Claims to the receiver by the receiver or its selected recovery authority resume the original inbound; originator-authorized claims and claims to other addresses are reroutes.
+
+TIP-403 authorization checks are unchanged and continue to revert on failure. TIP-1028 policies only apply to TIP-20 precompile flows. Ordinary contracts and other precompiles are unaffected.
```
0xalpharush: Does "bypasses the token-level TIP-403 sender check for `ESCROW_ADDRESS" not count as a change?

tips/tip-1028.md
```diff
@@ -0,0 +1,516 @@
...
+
+# TIP-1028: Address-Level Receive Policies
+
+<br>
+
+## Abstract
+
+TIP-1028 extends TIP-403 with address-level receive policies, letting a receiver control which TIP-20 tokens they accept and who can send them.
+
+When a user's receive policy blocks an inbound TIP-20 transfer, delivery is redirected to `TIP1028Guard`, a proof is recorded, and the amount becomes claimable by the recovery authority designated by the user.
```
horsefacts: Not a huge fan of putting TIP names in components. How about `ReceivePolicyGuard`?

tips/tip-1028.md
```diff
@@ -0,0 +1,516 @@
...
+- A reroute claim with `to == TIP1028_GUARD_ADDRESS` MUST revert.
+- `setReceivePolicy(...)` MUST reject `account == TIP1028_GUARD_ADDRESS`.
+- The TIP-20 `burnBlocked` function MUST revert when `from == TIP1028_GUARD_ADDRESS`. Funds leave `TIP1028Guard` only by claiming a proof. Burning the `TIP1028_GUARD_ADDRESS` balance directly would leave proofs pointing at money that no longer exists.
+
+Without intervention, the first blocked transfer or mint for a token would pay an unexpected cold-sstore surcharge (~22,100 gas) for the zero-to-nonzero write to `balances[TIP1028_GUARD_ADDRESS]`, charged to whichever user happens to trigger the first block. To avoid this, the TIP-20 initializer MUST charge the deployer a one-time fee equivalent to a cold sstore on the token's `balances[TIP1028_GUARD_ADDRESS]` slot, and every subsequent write to `balances[TIP1028_GUARD_ADDRESS]` for that token MUST be priced as a warm sstore. No tokens are minted to `TIP1028Guard` at deployment and no implementation-private reserve is required; the cold cost is paid exactly once, at deploy time, by the issuer.
```
horsefacts: is this gas value correct?

tips/tip-1028.md
0xKitsune: I like this, @malleshpai @0xrusowsky any objections?

0xrusowsky: yup, good for me

malleshpai: ditto

tips/tip-1028.md
```diff
@@ -0,0 +1,483 @@
...
```
malleshpai:
```suggestion
Without intervention, the first blocked transfer or mint for a token would pay an unexpected cold-sstore surcharge (~250,000 gas) for the zero-to-nonzero write to `balances[TIP1028_GUARD_ADDRESS]`, charged to whichever user happens to trigger the first block. To avoid this, the TIP-20 initializer MUST charge the deployer a one-time fee equivalent to a cold sstore on the token's `balances[TIP1028_GUARD_ADDRESS]` slot, and every subsequent write to `balances[TIP1028_GUARD_ADDRESS]` for that token MUST be priced as a warm sstore. No tokens are minted to `TIP1028Guard` at deployment and no implementation-private reserve is required; the cold cost is paid exactly once, at deploy time, by the issuer.
```

tips/tip-1028.md
malleshpai: nope, corrected.

tips/tip-1028.md
```diff
@@ -0,0 +1,483 @@
...
+
+### Claiming
+
+A claim consumes one full proof and releases the funds to a single destination. `claim(...)` takes an encoded proof and a destination `to`. For `proof.version == 1`, it decodes the witness as `ClaimProofV1`, derives the canonical receiver as `tip1022.resolve(proof.recipient)`, recomputes `proofKey`, requires `blockedProofAmount[proofKey] > 0`, zeroes the slot to free its storage, and releases the stored amount from `proof.token`. Once consumed, a proof is permanently retired: its `proofKey` slot is empty and any subsequent claim against the same witness MUST revert with `InvalidProof()`. Partial claims are not allowed. Claims MUST consume whole proofs.
+
+The release path inside the TIP-20 token debits `balances[TIP1028Guard]`, credits the destination, emits `Transfer(TIP1028Guard, to, amount)` followed by `ProofClaimed(...)`, bypasses the token-level TIP-403 sender check for `TIP1028Guard`, and treats `TIP1028Guard` as a reward-exempt always-opted-out source. Releasing a previously blocked mint does not emit a fresh `Mint` event because the claim is a transfer out of blocked transfers precompile, not a new mint.
```
0xalpharush: Should there be something like: "TIP-403 whitelist/blacklist membership mutation MUST reject TIP1028_GUARD_ADDRESS"?

Or it will just be ignored even if it's added/removed?

tips/tip-1028.md
```diff
@@ -0,0 +1,483 @@
...
+- A reroute claim with `to == TIP1028_GUARD_ADDRESS` MUST revert.
+- `setReceivePolicy(...)` MUST reject `account == TIP1028_GUARD_ADDRESS`.
+- The TIP-20 `burnBlocked` function MUST revert when `from == TIP1028_GUARD_ADDRESS`. Funds leave `TIP1028Guard` only by claiming a proof. Burning the `TIP1028_GUARD_ADDRESS` balance directly would leave proofs pointing at money that no longer exists.
+
+Without intervention, the first blocked transfer or mint for a token would pay an unexpected cold-sstore surcharge (~250,000 gas) for the zero-to-nonzero write to `balances[TIP1028_GUARD_ADDRESS]`, charged to whichever user happens to trigger the first block. To avoid this, the TIP-20 initializer MUST charge the deployer a one-time fee equivalent to a cold sstore on the token's `balances[TIP1028_GUARD_ADDRESS]` slot, and every subsequent write to `balances[TIP1028_GUARD_ADDRESS]` for that token MUST be priced as a warm sstore. No tokens are minted to `TIP1028Guard` at deployment and no implementation-private reserve is required; the cold cost is paid exactly once, at deploy time, by the issuer.
```
0xalpharush: For already deployed TIP-20's, a user will have to pay the cost or it will just be omitted and only charged for newly deployed TIP-20's?

tips/tip-1028.md
```diff
@@ -0,0 +1,483 @@
...
...re reroutes. This includes every originator-authorized claim, even if `to == receiver`. A reroute MUST reject `to == TIP1028_GUARD_ADDRESS`. If `to` is a TIP-1022 virtual address, it is resolved to its master address before destination checks and final credit; if resolution fails, the call MUST revert with `InvalidClaimAddress()`. The reroute enforces token-level transfer-recipient authorization and receive policy checks against the resolved destination, using the consumed proof's `originator` for the receive policy check. If either destination check fails, the call MUST revert with `InvalidClaimAddress()`. The `ProofClaimed` event preserves the literal `to` supplied by the caller for attribution. If a direct receiver-authorized reroute is initiated through an access key, the claim MUST meter the claimed amount against the receiver's AccountKeychain spending limit as an ordinary TIP-20 spend by the receiver. If an originator-authorized reroute is initiated through an access key, it MUST meter against the originator. If the receiver uses a third-party recovery authority, any equivalent delegation, timelock, multisig, or key-policy enforcement is the responsibility of that authority.
```
0xalpharush: > If a direct receiver-authorized reroute is initiated through an access key, the claim MUST meter the claimed amount against the receiver's AccountKeychain spending limit as an ordinary TIP-20 spend by the receiver. If an originator-authorized reroute is initiated through an access key, it MUST meter against the originator. 

Not sure if it's worth adding an invariant: "A keychain should not be able to send more than its limits using TIP-1028 reroutes". It's already the case but a bit subtle

horsefacts (approved):

tips/tip-1028.md
```diff
@@ -0,0 +1,483 @@
+---
+id: TIP-1028
+title: Address-Level Receive Policies
+description: Extends TIP-403 with address-level receive policies, token filters, and proof-based recovery for blocked TIP-20 transfers and mints.
+authors: Mallesh Pai, 0xrusowsky, 0xKitsune
+status: Draft
+related: TIP-403, TIP-1015, TIP-20, TIP-1016, TIP-1022
+protocolVersion: TBD
```
jenpaff:
```suggestion
protocolVersion: T6
```

tips/tip-1028.md
```diff
@@ -0,0 +1,483 @@
+---
+id: TIP-1028
+title: Address-Level Receive Policies
+description: Extends TIP-403 with address-level receive policies, token filters, and proof-based recovery for blocked TIP-20 transfers and mints.
+authors: Mallesh Pai, 0xrusowsky, 0xKitsune
+status: Draft
```
jenpaff:
```suggestion
status: Approved
```

tips/tip-1028.md
```diff
@@ -0,0 +1,483 @@
...
+        InboundKind kind;
+        bytes32 memo;
+    }
+
+    function balanceOf(bytes calldata proof) external view returns (uint256 amount);
+
+    function claim(address to, bytes calldata proof) external;
+
+    function storeBlocked(
+        address token,
```
samczsun, line 515: maybe this should just be taken from msg.sender

tips/tip-1028.md
```diff
@@ -0,0 +1,483 @@
...
+The aggregate blocked balance for each TIP-20 token sits in that token's `balances[BLOCKED_TRANSFER_ADDRESS]` slot.
+
+### Restrictions on `TIP1028_GUARD_ADDRESS`
+
+The following restrictions apply:
+
+- A TIP-20 transfer or mint with `to == TIP1028_GUARD_ADDRESS` MUST revert with `AddressReserved()`. This applies to `transfer`, `transferFrom`, `transferWithMemo`, `transferFromWithMemo`, `systemTransferFrom`, `mint`, and `mintWithMemo`.
+- A reroute claim with `to == TIP1028_GUARD_ADDRESS` MUST revert.
+- `setReceivePolicy(...)` MUST reject `account == TIP1028_GUARD_ADDRESS`.
+- The TIP-20 `burnBlocked` function MUST revert when `from == TIP1028_GUARD_ADDRESS`. Funds leave `TIP1028Guard` only by claiming a proof. Burning the `TIP1028_GUARD_ADDRESS` balance directly would leave proofs pointing at money that no longer exists.
```
samczsun: I think we need a `burnBlockedProof` or something, or a malicious user can reject transfers from themselves, self-transfer, and now they have a proof against a claim that the issuer can't touch, even if they block the user's address (because the originator is never checked against 403). If the originator is blocked, then the burn blocked role should be able to burn any valid proof that has them as an originator. Similarly, trying to claim a proof where the originator is blocked should reject.

tips/tip-1028.md
```diff
@@ -0,0 +1,483 @@
...
+    Receiver->>TIP403: setReceivePolicy(...)
+    Sender->>TIP20: transfer(receiver, amount)
+    TIP20->>TIP403: validateReceivePolicy(token, sender, receiver)
+    TIP403-->>TIP20: blocked
+    TIP20->>BlockedTransfers: storeBlocked(...)
+    TIP20-->>Sender: success (funds recorded as blocked)
+
+    Receiver->>BlockedTransfers: claim(...)
+    BlockedTransfers->>TIP20: release from TIP1028Guard
+    TIP20-->>Receiver: transfer funds
```
0xalpharush, line 73: Related to https://github.com/tempoxyz/tempo/pull/3791#pullrequestreview-4322802080. I think this is missing 403 checks for receiver

tips/tip-1028.md
malleshpai: done

tips/tip-1028.md
malleshpai: based on a directive from @0xKitsune we're viewing an invariant as something that should always be true in any reachable state. what you're describing is what i think we'd call a test case, something the implementation should be checked for but not strictly an invariant?

tips/tip-1028.md
0xrusowsky: this is just an internal function.

inside precompiles there is no notion of calls (thus `msg.sender` remains as the caller of the precompile entry point).

we internally ensure that the token addressed passed is the caller precompile's address

tips/tip-1028.md
malleshpai: added a line

tips/tip-1028.md
```diff
@@ -0,0 +1,540 @@
...
+    Receiver->>TIP403: setReceivePolicy(...)
+    Sender->>TIP20: transfer(receiver, amount)
+    TIP20->>TIP403: check token's TIP-403 policy(sender, receiver)
+    TIP403-->>TIP20: allowed
+    TIP20->>TIP403: validateReceivePolicy(token, sender, receiver)
+    TIP403-->>TIP20: blocked
+    TIP20->>ReceivePolicyGuard: storeBlocked(...)
+    TIP20-->>Sender: success (funds recorded as blocked)
+
+    Receiver->>ReceivePolicyGuard: claim(...)
```
malleshpai:
```suggestion
    RecoveryAuthority->>ReceivePolicyGuard: claim(...)
```

jenpaff (approved):

tips/tip-1028.md
0xrusowsky: imo best to simply ignore (this prevents us from needing custom logic for whitelist policies)

tips/tip-1028.md
```diff
@@ -0,0 +1,541 @@
...
+`addressReceiveConfig[account]` is a packed `uint256` with the following layout:
+
+| Bits | Size | Field |
+|---|---:|---|
+| `0` | 1 | `hasReceivePolicy` |
+| `1..64` | 64 | `senderPolicyId` |
+| `65..72` | 8 | `senderPolicyType` |
+| `73..136` | 64 | `tokenFilterId` |
+| `137..144` | 8 | `tokenFilterType` |
+| `145..255` | 111 | reserved, MUST be zero |
```
0xrusowsky: needs update with `recoveryMode` enum, which allows us to save 1 SLOAD
```suggestion
| `145` | 1 | `recoveryMode` |
| `146..255` | 110 | reserved, MUST be zero |
```

tips/tip-1028.md
```diff
@@ -0,0 +1,541 @@
...
+
+```text
+proofKey = keccak256(abi.encode(proof))
+```
+
+where:
+
+- `version`: one-byte discriminator inside the proof. MUST be `1` for proofs created under this TIP. Future proof-key formats MUST use a different value and MAY define a different version-specific proof body.
+- `token`: the TIP-20 token whose balance ledger holds the blocked amount at `RECEIVE_POLICY_GUARD_ADDRESS`.
+- `recoveryAuthority`: the encoded recovery authority captured when the proof was created: `address(0)` for originator recovery (default), or a nonzero address authorized to claim the proof.
```
0xrusowsky: updated this in the impl to be more user friendly. recovery authority now always reflects the address with recovery rights. 

this means that when we store `address(0)` the proof will already use `recoveryAuthority = originator`


```suggestion
- `recoveryAuthority`: the recovery authority captured when the proof was created. This address is the only caller authorized to claim the proof.
```

tips/tip-1028.md
```diff
@@ -0,0 +1,541 @@
...
+```
+
+#### Claim authority
+
+The proof only tells `ReceivePolicyGuard` which proof to consume. It does not grant a right to claim. Blocked events are public, so anyone can build a valid witness, but only the authorized claimer may call `claim(...)`.
+
+Each proof is governed by the `recoveryAuthority` captured for that proof at block time:
+
+- if `recoveryAuthority == address(0)`, only `proof.originator` may call `claim(...)`;
+- otherwise, only `recoveryAuthority` may call `claim(...)`.
```
0xrusowsky: proofs don't show `address(0)` anymore, now they already pass `originator`

```suggestion
Each proof is governed by the `recoveryAuthority` captured for that proof at block time. That is the only authorized caller for claims.
```

tips/tip-1028.md
```diff
@@ -0,0 +1,541 @@
...
+Changing `addressRecoveryAuthority[receiver]` affects future proofs only. Existing proofs remain governed by the recovery authority captured in their key. If a receiver configures originator recovery and the originator is a system or protocol address that cannot call `ReceivePolicyGuard`, the proof may be unclaimable; this is a receiver configuration risk. 
+
+#### Policy subject
+
+Claim authority and TIP-403 policy status are separate. The authorized claimer is determined by `recoveryAuthority`. The policy subject is the address whose TIP-403 sender status is checked for reroutes and proof burns:
+
+- for originator recovery, the policy subject is `proof.originator`;
+- for any nonzero `recoveryAuthority`, the policy subject is `receiver`.
+
+A nonzero recovery authority acts as the receiver's delegate for this purpose. Its own TIP-403 policy status does not control whether the proof can be rerouted or burned.
```
0xrusowsky: same
```suggestion
- otherwise, the policy subject is `receiver`.

A non-originator recovery authority acts as the receiver's delegate for this purpose. Its own TIP-403 policy status does not control whether the proof can be rerouted or burned.
```

tips/tip-1028.md
```diff
@@ -0,0 +1,541 @@
...
+
+#### Resume claims
+
+A claim resumes the original inbound when `recoveryAuthority != address(0)` and `to == receiver`. A resume:
```
0xrusowsky:
```suggestion
A claim resumes the original inbound when `recoveryAuthority != originator` and `to == receiver`. A resume:
```

tips/tip-1028.md
```diff
@@ -0,0 +1,541 @@
...
+- Which TIP-20 tokens are allowed.
+- Which senders are allowed.
+
+Receive policies are stored per address in the TIP-403 registry. Conceptually, each account has:
+
+```text
+ReceivePolicy(account) = (
+    senderPolicyId,    // TIP-403 policy ref indicating which senders are allowed
+    tokenFilterId,     // TIP-403 policy ref indicating which TIP-20 tokens are allowed
+    recoveryAuthority   // address(0) or a claiming address
```
0xrusowsky:
```suggestion
    recoveryAuthority   // transfer originator or a claiming address
```

tips/tip-1028.md
```diff
@@ -0,0 +1,541 @@
...
+| `0` | 1 | `hasReceivePolicy` |
+| `1..64` | 64 | `senderPolicyId` |
+| `65..72` | 8 | `senderPolicyType` |
+| `73..136` | 64 | `tokenFilterId` |
+| `137..144` | 8 | `tokenFilterType` |
+| `145..255` | 111 | reserved, MUST be zero |
+
+When `hasReceivePolicy == 0`, the address has no receive policy and all transfers and mints are allowed. The cached type fields are valid because policy type and token filter type are immutable after creation.
+
+`addressRecoveryAuthority[account]` stores the encoded recovery authority for the address. `address(0)` selects originator recovery (the default for uninitialized storage). Any nonzero address selects that address as the recovery authority.
```
0xrusowsky:
```suggestion
for gas efficiency purposes, the packed struct holds a `recoveryMode` enum (`uint8`), which indicates the recovery authority:
`RecoveryMode::Originator`, `RecoveryMode::Receiver`, or `RecoveryMode:ThirdParty`. When the recovery authority is a third-party, the protocol also stores its address in a dedicated slot `addressRecoveryAuthority[account]`.
```

tips/tip-1028.md
```diff
@@ -0,0 +1,541 @@
...
+
+Token filters use the existing TIP-403 `PolicyData` and policy membership set. A receive policy references one by `tokenFilterId`, and the policy's members are interpreted as TIP-20 token addresses.
+
+The registry does not need to validate that policy members are TIP-20 contracts. Non-token addresses in a token filter are inert configuration mistakes.
+
+<br>
+
+## Receive Policy Evaluation
+
+`validateReceivePolicy(token, sender, receiver)` returns whether a transfer or mint to `receiver` is allowed and, if not, why. If the receiver has not configured a receive policy (`hasReceivePolicy == 0`), it returns `(true, NONE)`.
```
samczsun, line 180: Maybe this should just return the enum, and instead of BlockedReason it should be PolicyResult and the first value should be ALLOWED so then you don't have this invalid state of (false, PolicyResult.ALLOWED)

tips/tip-1028.md
```diff
@@ -0,0 +1,542 @@
...
+```
+
+#### Claim authority
+
+The receipt only tells `ReceivePolicyGuard` which receipt to consume. It does not grant a right to claim. Blocked events are public, so anyone can build a valid witness, but only the authorized claimer may call `claim(...)`.
+
+Each receipt is governed by the `recoveryAuthority` captured for that receipt at block time:
+
+- if `recoveryAuthority == address(0)`, only `receipt.originator` may call `claim(...)`;
+- otherwise, only `recoveryAuthority` may call `claim(...)`.
```
0xrusowsky:
```suggestion
Each receipt is governed by the `recoveryAuthority` captured for that receipt at block time. This is the only address that has claim capabilities over the blocked funds.
```

tips/tip-1028.md
```diff
@@ -0,0 +1,540 @@
...
+
+| Bits | Size | Field |
+|---|---:|---|
+| `0` | 1 | `hasReceivePolicy` |
+| `1..64` | 64 | `senderPolicyId` |
+| `65..72` | 8 | `senderPolicyType` |
+| `73..136` | 64 | `tokenFilterId` |
+| `137..144` | 8 | `tokenFilterType` |
+| `145` | 1 | `recoveryMode` |
+| `146..255` | 110 | reserved, MUST be zero |
```
0xrusowsky:
```suggestion
| `145-152` | 8 | `recoveryMode` |
| `153..255` | 102 | reserved, MUST be zero |
```

PR #3793, opened 2026-05-01
fix: validate call scopes before writing

klkvr:
Changes key authorization logic to pre-validate all call scopes before writing them to avoid DOS vectors where we are tricked into inserting a huge batch of call scopes where last one fails validation and invalidates entire batch

Additionally changes selector scopes validation to not enforce that TIP20 is initialized but just validate the address

Gated behind T4


mattsse (approved):
lgtm

0xKitsune (approved):

PR #3800, opened 2026-05-04
feat(tip-1028): implement address-level receive policies

0xKitsune:
This PR Implements TIP-1028 receive policies, token filters, and receipt based escrow for blocked TIP-20 transfers and mints. 

Changes are gated behind the T6 hardfork. Unit and e2e coverage is also added for receipt lifecycle, claim authorization, recovery address flows, and escrow balance accounting.

Ref #3791


crates/contracts/src/precompiles/mod.rs
```diff
@@ -4,6 +4,7 @@ pub mod common_errors;
 pub mod nonce;
 pub mod signature_verifier;
 pub mod stablecoin_dex;
+pub mod tip1028_escrow;
```
0xrusowsky: nit: can we simply name `escrow`?

crates/contracts/src/precompiles/mod.rs
```diff
@@ -43,3 +45,4 @@ pub const ADDRESS_REGISTRY_ADDRESS: Address =
     address!("0xFDC0000000000000000000000000000000000000");
 pub const SIGNATURE_VERIFIER_ADDRESS: Address =
     address!("0x5165300000000000000000000000000000000000");
+pub const TIP1028_ESCROW_ADDRESS: Address = address!("0xE5C0000000000000000000000000000000000000");
```
0xrusowsky: same nit, i'd simply do `ESCROW_ADDRESS`

crates/contracts/src/precompiles/escrow.rs
```diff
@@ -0,0 +1,69 @@
...
+};
+
+crate::sol! {
+    #[derive(Debug, PartialEq, Eq)]
+    #[sol(abi)]
+    interface ITIP1028Escrow {
+        enum BlockedReason {
+            NONE,
+            TOKEN_FILTER,
+            RECEIVE_POLICY
```
0xrusowsky: technically misses `TOKEN_FILTER_AND_RECEIVE_POLICY` as per mallesh's original spec

crates/contracts/src/precompiles/tip20.rs
```diff
@@ -170,6 +178,9 @@ crate::sol! {
         error InvalidTransferPolicyId();
         error PermitExpired();
         error InvalidSignature();
+        error EscrowAddressReserved();
+        error ClaimDestinationUnauthorized();
+        error InsufficientEscrowBalance();
```
0xrusowsky: these errors belong to the escrow precompile, not to TIP20

crates/contracts/src/precompiles/tip403_registry.rs
```diff
@@ -13,6 +13,12 @@ crate::sol! {
             COMPOUND
         }
 
+        enum BlockedReason {
+            NONE,
+            TOKEN_FILTER,
+            RECEIVE_POLICY
+        }
```
0xrusowsky, line 20: i think we should probably only define `BlockedReason` here, and have tip20 and escrow import the type rather than redefining it

crates/precompiles/src/tip20/mod.rs
```diff
@@ -413,32 +435,54 @@ impl TIP20Token {
...
         Ok(())
     }
 
     /// Internal helper to mint new tokens and update balances.
-    fn _mint(&mut self, msg_sender: Address, to: &Recipient, amount: U256) -> Result<()> {
+    fn _mint(
+        &mut self,
+        msg_sender: Address,
+        recipient: &Recipient,
+        to: Address,
```
0xrusowsky: this info is already available in `Recipient`

0xrusowsky (commented):
overall i think this is a great start.

i think we should lean more into the `Recipient` struct to simplify and offload some of the cross-precompile logic out of the TIP20 transfer/mint paths, similar to what we did with virtual addresses.

i'll open a draft with a proposal to build on top of this foundation

crates/precompiles/src/tip20/mod.rs
```diff
@@ -449,16 +493,38 @@ impl TIP20Token {
...
             .ok_or(TempoPrecompileError::under_overflow())?;
-        self.set_balance(to.target, new_to_balance)?;
+        self.set_balance(credit_to.target, new_to_balance)?;
+
+        self.emit_event(credit_to.build_transfer_event(Address::ZERO, amount))?;
+
+        if let Some(blocked) = blocked {
+            return self
+                .store_blocked_inbound(
+                    Address::ZERO,
```
0xrusowsky: based on https://github.com/tempoxyz/tempo/pull/3791#discussion_r3178322931 i think this should be `msg_sender`

crates/contracts/src/precompiles/tip1028_escrow.rs
```diff
@@ -0,0 +1,69 @@
...
+            uint64 blockedAt;
+            uint64 blockedNonce;
+            BlockedReason blockedReason;
+            InboundKind kind;
+            bytes32 memo;
+        }
+
+        function blockedReceiptBalance(address token, address recoveryContract, uint8 receiptVersion, bytes calldata receipt) external view returns (uint256 amount);
+        function claimBlocked(address token, address recoveryContract, uint8 receiptVersion, bytes calldata receipt, address to) external;
+        function storeBlocked(address token, address originator, address receiver, address recipient, address recoveryContract, uint256 amount, BlockedReason blockedReason, InboundKind kind, bytes32 memo) external returns (uint64 blockedNonce, uint64 blockedAt);
```
0xrusowsky: this fn should be internal to the precompile

crates/precompiles/src/tip1028_escrow/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+        {
+            return Err(TIP1028EscrowError::invalid_receipt_claim().into());
+        }
+
+        let receiver = AddressRegistry::new()
+            .resolve_recipient(call.recipient)
+            .map_err(|_| TIP1028EscrowError::claim_destination_unauthorized())?;
+        if receiver != call.receiver {
+            return Err(TIP1028EscrowError::invalid_receipt_claim().into());
+        }
```
0xrusowsky: this performs redundant work.

since this fn should be internal to the precompile (not exposed in the dispatcher) + the caller (TIP20) already has resolved the virtual address, we don't need to re-validate this here.

crates/contracts/src/precompiles/mod.rs
0xKitsune: Sgtm, addressed in https://github.com/tempoxyz/tempo/pull/3800/commits/56a621a6ae1e67e168ebac76ae8feca67deaaa1e

crates/contracts/src/precompiles/mod.rs
0xKitsune: Pushed a fix in https://github.com/tempoxyz/tempo/pull/3800/commits/1ed1a2e2655bd53de8f0278eaa4e682dc869f7e1

crates/contracts/src/precompiles/escrow.rs
0xKitsune: This is no longer in the current spec.

crates/contracts/src/precompiles/tip20.rs
0xKitsune: Moved errors in https://github.com/tempoxyz/tempo/pull/3800/commits/8f8fec79b07d652d11892eccd795b72cc586f181

crates/precompiles/src/tip1028_escrow/mod.rs
0xKitsune: Addressed in https://github.com/tempoxyz/tempo/pull/3800/commits/474c96c7c77b449bac5be9cb9d9e4c4b062ab527

crates/contracts/src/precompiles/escrow.rs
```diff
@@ -0,0 +1,63 @@
...
+            address recipient;
+            uint64 blockedAt;
+            uint64 blockedNonce;
+            uint8 blockedReason;
+            InboundKind kind;
+            bytes32 memo;
+        }
+
+        function blockedReceiptBalance(address token, address recoveryContract, uint8 receiptVersion, bytes calldata receipt) external view returns (uint256 amount);
+        function claimBlocked(address token, address recoveryContract, uint8 receiptVersion, bytes calldata receipt, address to) external;
```
0xrusowsky: nit: should we rename to simply `claim`?

crates/precompiles/src/tip1028_escrow/mod.rs
```diff
@@ -0,0 +1,1406 @@
...
+            return Err(TIP1028EscrowError::unauthorized_claimer().into());
+        }
+
+        let key = self.receipt_key(call.receiptVersion, call.token, recovery_address, &receipt)?;
+        let amount = self.blocked_receipt_amount[key].read()?;
+        if amount.is_zero() {
+            return Err(TIP1028EscrowError::invalid_receipt_claim().into());
+        }
+
+        let guard = self.storage.checkpoint();
```
0xrusowsky: why do we need to checkpoint here? afaict any error after this point propagates to the precompile boundary, so revm would revert the whole frame (as usual).

this is only necessary when there's a caller that swallows the error and continues, which doesn't seem to be the case

crates/precompiles/src/tip20/mod.rs
```diff
@@ -414,8 +427,20 @@ impl TIP20Token {
         call: ITIP20::mintWithMemoCall,
     ) -> Result<()> {
         let to = Recipient::resolve(call.to)?;
-        self._mint(msg_sender, &to, call.amount)?;
+        self.check_role(msg_sender, *ISSUER_ROLE)?;
+        self.validate_mint(&to)?;
```
0xrusowsky: this is technically breaking if `validate_mint()` fails:
```
BEFORE inside fn _mint():
1. check_role(..)?;
2. total_supply()?; // SLOAD
3. validate_mint(..)?;

NOW
1. check_role(..)?;
2. validate_mint(..)?;
3. total_supply()?;
```

that's in my previous proposal to enhance `Receiver` i was performing the inbound validation inside `_mint()`

crates/precompiles/src/tip20/mod.rs
```diff
@@ -1009,6 +1068,118 @@ impl TIP20Token {
...
+        }
+        let registry = TIP403Registry::new();
+        let Some(reason) = registry.validate_receive_policy(self.address, originator, to.target)?
+        else {
+            return Ok(false);
+        };
+        let recovery = registry.receive_policy_recovery(to.target)?;
+        let recipient = to.virtual_addr.unwrap_or(to.target);
+
+        let guard = self.storage.checkpoint();
```
0xrusowsky: why do we checkpoint here? same question as before, afaik we always propagate errors?

crates/precompiles/src/tip20/mod.rs
```diff
@@ -1009,6 +1068,118 @@ impl TIP20Token {
         self.emit_event(to.build_transfer_event(from, amount))
     }
 
+    /// Validates the TIP-1028 receive-policy check for the destination address. If the receive
+    /// policy prohibits the action, the funds are escrowed.
+    fn validate_or_escrow_funds(
```
0xrusowsky: nits:
- can we document what when the return type means "whether the funds were escrowed or not", and explicitly explain that the caller should not move the funds if that's the case?
- can we maybe rename the fn to indicate what we are validating? personally i like `validate_inbound_or_escrow` but if we want a shorter version we could also use `validate_inbound` or `validate_receive_policies`

0xrusowsky: the refactor with this function is an elegant way to simplify, i really like it.
we only need to fix the re-execution issue on mints, but this is great.

crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -48,6 +48,16 @@ pub struct TIP403Registry {
     /// value is `true` when the address is allowed; for blacklists it is `true` when the
     /// address is restricted.
     policy_set: Mapping<u64, Mapping<Address, bool>>,
+    /// Account receive policy configuration.
+    address_receive_config: Mapping<Address, ReceivePolicyConfig>,
```
0xrusowsky: nit: can we keep consistent naming with pre-existing vars and use "policy"?

```suggestion
    /// Account-level receive policy configuration.
    receive_policy_config: Mapping<Address, ReceivePolicyConfig>,
}
```

crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -230,6 +239,53 @@ impl TIP403Registry {
         })
     }
 
+    /// Returns `account`'s receive-policy configuration.
+    pub fn receive_policy(
+        &self,
+        call: ITIP403Registry::receivePolicyCall,
```
0xrusowsky: let's flatten the call and take `account: Address` directly

crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -276,6 +332,39 @@ impl TIP403Registry {
         Ok(new_policy_id)
     }
 
+    /// Sets the caller's TIP-1028 receive policy.
+    pub fn set_receive_policy(
+        &mut self,
+        msg_sender: Address,
+        call: ITIP403Registry::setReceivePolicyCall,
+    ) -> Result<()> {
+        if msg_sender == ESCROW_ADDRESS {
```
0xrusowsky: shouldn't this also verify the recovery address?

it shouldn't be a TIP20 nor the ESCROW precompile

tips/verify/lib/tempo-std
0xrusowsky, line 1: this requires a PR on its own.

crates/precompiles/src/tip20/mod.rs
0xrusowsky: proposed fix [`06ea050` (#3845)](https://github.com/tempoxyz/tempo/pull/3845/commits/06ea0509011822215e473e7884682cf5c069c99a)

crates/precompiles/src/tip20/mod.rs
```diff
@@ -1009,6 +1068,118 @@ impl TIP20Token {
...
+                return Err(TIP20Error::policy_forbids().into());
+            }
+            if registry
+                .validate_receive_policy(self.address, originator, destination.target)?
+                .is_some()
+            {
+                return Err(TIP20Error::policy_forbids().into());
+            }
+            if meter_spending_limit {
+                self.check_and_update_spending_limit(receiver, amount)?;
```
0xrusowsky: should be forbidden too, they should use the sentinel `address(0)`. repeated comment in `setReceivePolicy`

crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -276,6 +332,39 @@ impl TIP403Registry {
         Ok(new_policy_id)
     }
 
+    /// Sets the caller's TIP-1028 receive policy.
+    pub fn set_receive_policy(
```
0xrusowsky, line 398: valid. already flagged here https://github.com/tempoxyz/tempo/pull/3800#discussion_r3199464331

0xrusowsky: fix proposal: [`3c27ebd` (#3845)](https://github.com/tempoxyz/tempo/pull/3845/commits/3c27ebd26155c3f738dbe58db0280bf7cdcd27d7)

crates/precompiles/src/tip403_registry/mod.rs
0xrusowsky: fix proposal: [`3c27ebd` (#3845)](https://github.com/tempoxyz/tempo/pull/3845/commits/3c27ebd26155c3f738dbe58db0280bf7cdcd27d7)

crates/precompiles/src/tip20/mod.rs
```diff
@@ -1009,6 +1068,118 @@ impl TIP20Token {
...
+            kind,
+            memo,
+        )?;
+        guard.commit();
+        Ok(true)
+    }
+
+    /// Releases escrowed funds to `to`. Self-recovery skips policy checks. Redirects
+    /// revalidate the transfer and receive policies and meter the spending limit when set.
+    pub(crate) fn release_from_escrow(
```
0xrusowsky: proposed fix: [`67cfee6` (#3845)](https://github.com/tempoxyz/tempo/pull/3845/commits/67cfee62382fa398389a070dc701b7002420232b)

0xrusowsky:
> 🚨 [SECURITY] burn_blocked permanently destroys escrowed funds — ESCROW_ADDRESS missing from protected-address allow-list

proposed fix [`3067470` (#3845)](https://github.com/tempoxyz/tempo/pull/3845/commits/3067470cfacdf136ff85e840b5c1aee7483d4d96)

crates/precompiles/src/tip20/mod.rs
```diff
@@ -1009,6 +1068,118 @@ impl TIP20Token {
...
+        meter_spending_limit: bool,
+    ) -> Result<()> {
+        if to == ESCROW_ADDRESS {
+            return Err(TIP1028EscrowError::escrow_address_reserved().into());
+        }
+
+        let destination = Recipient::resolve(to)?;
+        destination.validate()?;
+
+        if destination.target != receiver {
```
0xrusowsky: i previously discussed this with @malleshpai and i believe current impl is correct, and this would be overreaching with validations. however, it would be nice if we got an explicit ACK that we can disregard this concern

tips/verify/lib/tempo-std
0xKitsune: Addressed in https://github.com/tempoxyz/tempo/pull/3800/changes/c155db018c3b6224b61048a5cac31d40e3b649ed

crates/contracts/src/precompiles/escrow.rs
0xKitsune: Addressed in https://github.com/tempoxyz/tempo/pull/3800/changes/e5c3ed5050e5e70556c1dfe24c321df520f7d8c9

crates/precompiles/src/tip20/mod.rs
0xKitsune: Removed checkpoint in https://github.com/tempoxyz/tempo/pull/3800/changes/b38a2656aa8da8ddec826bc19a40807bf1d8a53a

crates/precompiles/src/tip1028_escrow/mod.rs
0xKitsune: Addressed in https://github.com/tempoxyz/tempo/pull/3800/changes/1cfe8f832e63d25e5153a1675e7e5216b33ebbcb

crates/precompiles/src/tip20/mod.rs
```diff
@@ -1014,9 +1079,21 @@ impl TIP20Token {
...
     }
 
-    /// Ensures that the recipient is authorized to receive mints.
-    /// Additionally (+T3) checks pause state, validates the effective recipient.
-    fn validate_mint(&self, to: &Recipient) -> Result<()> {
+    /// Resolves the effective recipient, checks that `msg_sender` has the issuer role, preserves
+    /// the pre-T6 `total_supply` read position, validates that the token is not paused, and the
+    /// recipient can receive mints. On T6+, the `total_supply` read is deferred to `_mint`.
+    ///
+    /// Returns the resolved recipient and, pre-T6, the total supply.
```
0xrusowsky:
```suggestion
    /// Resolves the effective recipient and verifies that ⁠`msg_sender` has the issuer role.
    /// To preserve re-execution of old transactions (pre-T6), also reads ⁠total_supply.
    /// Additionally (+T3) checks pause state and validates the effective recipient.
    ///
    /// Returns the resolved [⁠`Recipient`] and, pre-T6, the ⁠total_supply.
```

crates/precompiles/src/tip1028_escrow/mod.rs
```diff
@@ -0,0 +1,1500 @@
...
+            blockedAt: blocked_at,
+            blockedNonce: blocked_nonce,
+            blockedReason: blocked_reason as u8,
+            kind,
+            memo,
+        };
+        let key = self.receipt_key(BLOCKED_RECEIPT_VERSION, token, recovery_address, &receipt)?;
+        self.blocked_receipt_amount[key].write(amount)?;
+
+        if kind == ITIP1028Escrow::InboundKind::TRANSFER {
```
0xrusowsky: imo makes sense to emit `TransferBlocked` here too

crates/precompiles/src/receive_policy_guard/mod.rs
```diff
@@ -0,0 +1,1500 @@
...
+    Originator(Address),
+    Contract(Address),
+}
+
+impl RecoveryAuthority {
+    fn from_address(address: Address, receiver: Address, originator: Address) -> Self {
+        if address == RECOVERY_RECEIVER {
+            Self::Receiver(receiver)
+        } else if address == RECOVERY_SENDER {
+            Self::Originator(originator)
```
0xrusowsky: i think this would be solved if we stored the msg sender as the originator, rather than address(0).

this was an open discussion point, and may be a good reason to use msg_sender as the originator for mints

crates/precompiles/src/tip20/mod.rs
```diff
@@ -1119,6 +1196,119 @@ impl TIP20Token {
...
+            if !registry.is_authorized_as(policy_id, destination.target, AuthRole::recipient())? {
+                return Err(TIP20Error::policy_forbids().into());
+            }
+            if registry
+                .validate_receive_policy(self.address, originator, destination.target)?
+                .is_some()
+            {
+                return Err(TIP20Error::policy_forbids().into());
+            }
+            self.check_and_update_spending_limit(receiver, amount)?;
```
0xrusowsky: i think impl is aligned with spec, so it is good as is in regards to policy checks.

crates/precompiles/src/tip1028_guard/mod.rs
```diff
@@ -0,0 +1,1500 @@
...
+        if amount.is_zero() {
+            return Err(TIP1028EscrowError::invalid_receipt_claim().into());
+        }
+
+        let guard = self.storage.checkpoint();
+        self.blocked_receipt_amount[key].write(U256::ZERO)?;
+
+        let reroute = match recovery_authority {
+            RecoveryAuthority::Originator(_) => true,
+            RecoveryAuthority::Receiver(_) | RecoveryAuthority::Contract(_) => call.to != receiver,
```
0xrusowsky: i think that's aligned with the spec

crates/precompiles/src/tip1028_escrow/mod.rs
0xKitsune: Sgtm to add `TransferBlocked`

0xrusowsky: addressed in [`8c30632` (this PR)](https://github.com/tempoxyz/tempo/pull/3800/commits/8c306325c62f6125db3d4762c25b976551c51820)

crates/precompiles/src/tip20/mod.rs
0xrusowsky: agreed internally that what matters is the origin of funds:
```
sender -> escrow -> receiver
```

it doesn't matter who is the claimer, as the validation was already performed vs the original sender

crates/precompiles/src/receive_policy_guard/mod.rs
0xrusowsky: we will use `msg_sender` as the originator for mints rather than address zero to fix this issue and align with the spec

0xrusowsky: addressed in [`5d26573` (this PR)](https://github.com/tempoxyz/tempo/pull/3800/commits/5d26573af3a766aa9d45fb158a72911ec2143f12)

crates/precompiles/src/tip1028_escrow/mod.rs
```diff
@@ -0,0 +1,1501 @@
...
+
+    /// Records a blocked inbound transfer or mint and emits `TransferBlocked` for
+    /// transfers. Caller moves the funds into escrow in the same checkpoint.
+    #[allow(clippy::too_many_arguments)]
+    pub(crate) fn store_blocked(
+        &mut self,
+        token: Address,
+        originator: Address,
+        receiver: Address,
+        recipient: Address,
```
klkvr: `receiver` and `recipient` are very confusing, should this just take `Recipient`?

crates/precompiles/src/tip20/mod.rs
```diff
@@ -1119,6 +1191,119 @@ impl TIP20Token {
...
+        }
+        if to.target == ESCROW_ADDRESS {
+            return Err(TIP1028EscrowError::escrow_address_reserved().into());
+        }
+        let registry = TIP403Registry::new();
+        let Some(reason) = registry.validate_receive_policy(self.address, originator, to.target)?
+        else {
+            return Ok(false);
+        };
+        let recovery = registry.receive_policy_recovery(to.target)?;
```
klkvr: afaict `recovery` could be resolved inside of the `store_blocked`

klkvr: also worth considering just moving this entire logic into the registry, similarly to how this is handled for keychain hooks https://github.com/tempoxyz/tempo/blob/45ad010035ae6a73416392c2b9eeb3f6df45f6ad/crates/precompiles/src/tip20/mod.rs#L1165-L1171

crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -48,6 +48,16 @@ pub struct TIP403Registry {
...
+    /// Account receive policy configuration.
+    receive_policy_config: Mapping<Address, ReceivePolicyConfig>,
+}
+
+#[derive(Debug, Clone, Default, Storable)]
+struct ReceivePolicyConfig {
+    has_receive_policy: bool,
+    sender_policy_id: u64,
+    token_filter_id: u64,
+    recovery_address: Address,
```
klkvr: needs a small doc

crates/precompiles/src/tip1028_guard/mod.rs
```diff
@@ -0,0 +1,1501 @@
...
+        ITIP1028Escrow::ClaimReceiptV1::abi_decode(receipt)
+            .map_err(|_| TIP1028EscrowError::invalid_receipt_claim().into())
+    }
+
+    /// Content hash over every receipt field. Any mutation yields a different empty slot.
+    fn receipt_key(
+        &self,
+        receipt_version: u8,
+        token: Address,
+        recovery_address: Address,
```
klkvr: could `token` and `recovery_address` be part of receipt payload?

crates/precompiles/src/tip1028_guard/mod.rs
```diff
@@ -0,0 +1,1501 @@
...
+                token,
+                receipt.originator,
+                receipt.recipient,
+                recovery_address,
+                U256::from(receipt.blockedReason),
+                receipt.kind,
+                receipt.memo,
+                U256::from(receipt.blockedAt),
+                U256::from(receipt.blockedNonce),
+            )
```
klkvr: this feels somewhat bloated — nonce already enforces uniqueness, could we only keep fields that actually matter?

crates/precompiles/src/tip20/mod.rs
```diff
@@ -1119,6 +1191,119 @@ impl TIP20Token {
...
+            if !registry.is_authorized_as(policy_id, destination.target, AuthRole::recipient())? {
+                return Err(TIP20Error::policy_forbids().into());
+            }
+            if registry
+                .validate_receive_policy(self.address, originator, destination.target)?
+                .is_some()
+            {
+                return Err(TIP20Error::policy_forbids().into());
+            }
+            self.check_and_update_spending_limit(receiver, amount)?;
```
klkvr: this seems wrong and different from the spec

`
If a direct receiver-authorized reroute is initiated through an access key, the claim MUST meter the claimed amount against the receiver's AccountKeychain spending limit as an ordinary TIP-20 spend by the receiver. If an originator-authorized reroute is initiated through an access key, it MUST meter against the originator. If the receiver uses a third-party recovery authority, any equivalent delegation, timelock, multisig, or key-policy enforcement is the responsibility of that authority.
`

so basically my understanding is that here we should update spending limits for the resolved recovery authority?

otherwise if a user has blocked transfers eligible for refund an access key would be able to initiate a tx from that account that'd claim funds to an external account bypassing the spending limits

crates/precompiles/src/tip1028_escrow/mod.rs
0xKitsune: Yeah can be simplified, updated in https://github.com/tempoxyz/tempo/pull/3800/commits/c3e74a943d701395271a9bfc6f638c3ef294caf0

crates/precompiles/src/tip403_registry/mod.rs
0xKitsune: Added in https://github.com/tempoxyz/tempo/pull/3800/commits/88571b8dca62940885a488cc1cf5e6ce53c6bc08

crates/precompiles/src/tip20/mod.rs
0xKitsune: Good catch, updated in https://github.com/tempoxyz/tempo/pull/3800/commits/77c6bf7c048549cb6a129071ba1be0ba0e5db2cf

crates/precompiles/src/tip20/mod.rs
0xKitsune: Moved `validate_or_escrow()` into the 403 registry and updated to fetch the `recovery` address with the same `SLOAD` when fetching the receive policy config.


Ref: https://github.com/tempoxyz/tempo/pull/3800/commits/2304deee7cd5338936cb5031a7c43a23240a6024

crates/precompiles/src/tip1028_guard/mod.rs
0xrusowsky: afaik this is a deliberate decision. we do not only care about uniqueness here, we are also committing to the canonical receipt contents.

this allows maximally flexible recovery contracts, which can use all the available metadata and come up with arbitrary redemption flows/rules.

crates/precompiles/src/tip1028_guard/mod.rs
0xrusowsky: idk if @malleshpai had a good reason to keep them out. i guess they kinda act like a domain separator?

i agree that it seems compelling to have them inside the payload

crates/precompiles/src/tip1028_guard/mod.rs
0xrusowsky: however i think we could use a ~equivalent form:
```rs
keccak256(
    U256::from(receipt_version), token, recovery_address, receipt.abi_encode()
).abi_encode()
```

it produces a different key value, but give us a more succinct impl

malleshpai: yes the idea was to have these additional fields available for possible smart contract recovery (i.e. the smart contract determines who is authorized based on these fields, an example is receiver can claim for first 24 hrs, sender can claim after that) / "off-label" use. no opinion on the implementation, but i think we'd like the fields in there.

crates/precompiles/src/tip1028_guard/mod.rs
malleshpai: hmmm yes you're right i think. we should gavel on this, i can update the TIP.

crates/precompiles/src/tip20/mod.rs
```diff
@@ -488,9 +497,19 @@ impl TIP20Token {
...
 
+        if self.validate_inbound_or_escrow(
+            msg_sender,
+            &to,
+            call.amount,
+            InboundKind::MINT,
+            B256::ZERO,
+        )? {
+            return Ok(());
+        }
```
klkvr: looks like it would've been cleaner to do this inside `_mint`? the only context that's missing there is memo

crates/precompiles/src/tip20/mod.rs
```diff
@@ -830,8 +892,20 @@ impl TIP20Token {
...
+
+        if self.validate_inbound_or_escrow(
+            call.from,
+            &to,
+            call.amount,
+            InboundKind::TRANSFER,
+            call.memo,
+        )? {
+            return Ok(true);
+        }
```
klkvr: same here, don't think there is a reason to pull `validate_transfer` and allowance change from `_transfer_from`?

crates/precompiles/src/tip1028_escrow/mod.rs
klkvr: looks like the checkpoint got commited back 😅

crates/precompiles/src/tip1028_escrow/mod.rs
```diff
@@ -0,0 +1,1504 @@
...
+        ))?;
+
+        Ok((blocked_nonce, blocked_at))
+    }
+
+    /// Releases escrowed receipt funds to the authorized recipient.
+    pub fn claim(&mut self, msg_sender: Address, call: ITIP1028Escrow::claimCall) -> Result<()> {
+        if !call.token.is_tip20() {
+            return Err(TIP1028EscrowError::invalid_token().into());
+        }
```
klkvr: nit: this check is redunant because no receipt can exist for a non-tip20 token

crates/precompiles/src/tip1028_escrow/mod.rs
```diff
@@ -0,0 +1,1504 @@
...
+
+        if call.to == ESCROW_ADDRESS {
+            return Err(TIP1028EscrowError::invalid_claim_address().into());
+        }
+
+        let receipt = Self::decode_v1(call.receiptVersion, &call.receipt)?;
+        let receiver = AddressRegistry::new()
+            .resolve_recipient(receipt.recipient)
```
klkvr: it seems that we now have a naming convention where `recipient` is maybe a virtual address and `receiver` is the actual receiver. should this fn be renamed to `resolve_receiver` then?

crates/precompiles/src/receive_policy_guard/mod.rs
```diff
@@ -0,0 +1,1504 @@
...
+        self.blocked_receipt_amount[key].write(U256::ZERO)?;
+
+        let reroute = match recovery_authority {
+            RecoveryAuthority::Originator(_) => true,
+            RecoveryAuthority::Receiver(_) | RecoveryAuthority::Contract(_) => call.to != receiver,
+        };
+        let recovery_addr = match recovery_authority {
+            RecoveryAuthority::Receiver(addr) | RecoveryAuthority::Originator(addr) => Some(addr),
+            RecoveryAuthority::Contract(_) => None,
+        };
```
klkvr: i'm not sure if it makes sense to me that for `Contract` we don't check spending limits. i don't think it would hurt?

also i'm not sure we need `Contract`/`Receiver` distinction at all? could we not simplify this so that there are just 2 options
1. return to sender
2. let X handle this where X might be either receiver itself or their trusted party or a contract

crates/precompiles/src/tip1028_guard/mod.rs
klkvr: if we do this could also consider making the version field part of receipt as well

that way we'd be able to just `abi_encode` receipt here and have a single argument in `claim` fn. if we were to ever add a v2 receipt we'd be able to classify it based on the first word of the ABI payload

crates/precompiles/src/tip20/mod.rs
```diff
@@ -1119,6 +1208,110 @@ impl TIP20Token {
...
+            destination.target,
+            to_balance
+                .checked_add(amount)
+                .ok_or(TempoPrecompileError::under_overflow())?,
+        )?;
+
+        self.emit_event(destination.build_transfer_event(ESCROW_ADDRESS, amount))?;
+        if let Some(hop) = destination.build_virtual_transfer_event(amount) {
+            self.emit_event(hop)?;
+        }
```
klkvr: could invoke `_transfer` here?

crates/precompiles/src/receive_policy_guard/mod.rs
0xrusowsky: thoughts @malleshpai ?

crates/precompiles/src/tip1028_escrow/mod.rs
0xrusowsky: addressed in [`4b8c915` (this PR)](https://github.com/tempoxyz/tempo/pull/3800/commits/4b8c91587895812da74c8c61f24aa8f6e1d69fb5)

crates/precompiles/src/tip20/mod.rs
0xrusowsky: addressed in [`0bfa03a` (this PR)](https://github.com/tempoxyz/tempo/pull/3800/commits/0bfa03ad7bfa0e0d31cdafc33adb1bf545c26722)

crates/precompiles/src/tip20/mod.rs
0xrusowsky: addressed in [`0bfa03a` (this PR)](https://github.com/tempoxyz/tempo/pull/3800/commits/0bfa03ad7bfa0e0d31cdafc33adb1bf545c26722)

crates/precompiles/src/tip20/mod.rs
0xrusowsky: addressed in [`09be233` (this PR)](https://github.com/tempoxyz/tempo/pull/3800/commits/09be2337974e6fb21a4d3a0de75f2db36c14a9ae)

crates/precompiles/src/tip1028_guard/mod.rs
```diff
@@ -0,0 +1,1225 @@
...
+        to: &Recipient,
+        recovery_address: Address,
+        amount: U256,
+        blocked_reason: BlockedReason,
+        kind: InboundKind,
+        memo: B256,
+    ) -> Result<(u64, u64)> {
+        debug_assert!(token.is_tip20(), "TIP1028Guard only accepts TIP20 tokens");
+
+        if matches!(
```
0xalpharush: Would prefer to explicitly match here as any additional variant needs to be handled, which this hides.

crates/precompiles/src/tip1028_guard/mod.rs
```diff
@@ -0,0 +1,1225 @@
...
+        let key = self.proof_key(&proof)?;
+        self.balances[key].write(amount)?;
+
+        self.emit_event(proof.blocked_event(receiver, amount))?;
+        Ok((blocked_nonce, blocked_at))
+    }
+
+    /// Given a valid proof, releases blocked funds to the authorized recipient.
+    pub fn claim(&mut self, msg_sender: Address, to: Address, proof: Bytes) -> Result<()> {
+        if to == TIP1028_GUARD_ADDRESS {
```
0xalpharush: Any reason to check here and in `release_blocked_funds`?

crates/precompiles/src/tip20/mod.rs
```diff
@@ -128,6 +130,13 @@ pub static ISSUER_ROLE: LazyLock<B256> = LazyLock::new(|| keccak256(b"ISSUER_ROL
 /// Role hash that authorizes burning tokens from blocked accounts.
 pub static BURN_BLOCKED_ROLE: LazyLock<B256> = LazyLock::new(|| keccak256(b"BURN_BLOCKED_ROLE"));
 
+#[rustfmt::skip]
+/// System custody addresses added to burn-blocked protection at each hardfork.
```
0xalpharush, line 135: This needs to be rebased bc it is missing T5. Would make reviewing a bit easier and allow e2e interaction with channels

0xrusowsky: done [`46c0adc` (this PR)](https://github.com/tempoxyz/tempo/pull/3800/commits/46c0adcaee9ae7aa31ac0813e57fb13da4a7f69a)

crates/precompiles/src/receive_policy_guard/mod.rs
0xrusowsky: @klkvr distinction between contract and receiver is now justified due to gas savings.

https://github.com/tempoxyz/tempo/pull/4057 introduced a `RecoveryMode` enum that allows us to pack the originator/receiver info within the same slot, rather than having to use a sentinel address (which goes into its own slot).

crates/precompiles/src/tip1028_guard/mod.rs
0xrusowsky: addressed in [`bc1da54` (this PR)](https://github.com/tempoxyz/tempo/pull/3800/commits/bc1da547b4a641d7acc4da5d83bcdf6add64eadf)

crates/precompiles/src/tip1028_guard/mod.rs
0xrusowsky: addressed in [`bc1da54` (this PR)](https://github.com/tempoxyz/tempo/pull/3800/commits/bc1da547b4a641d7acc4da5d83bcdf6add64eadf)

0xrusowsky: maybe we could make it a `debug_assert!` since the caller is always TIP20, and in prod this should be a dead branch.

wdyt? @0xalpharush

crates/precompiles/src/tip20/mod.rs
```diff
@@ -1106,6 +1182,88 @@ impl TIP20Token {
...
+        recovery_addr: Option<Address>,
+    ) -> Result<()> {
+        debug_assert!(
+            to != TIP1028_GUARD_ADDRESS,
+            "checked in TIP1028Guard::claim"
+        );
+
+        self.check_not_paused()?;
+        let destination = Recipient::resolve(to)?;
+        destination.validate()?;
```
fgimenez, line 1234: looks like if there's a faillure here `AddrRegistryError` or `TIP20Error::InvalidRecipient` would be propagated instead of `TIP1028GuardError::InvalidClaimAddress` as the spec mandates https://github.com/tempoxyz/tempo/pull/3791/changes#diff-ecfb80eee8e6f2ece98beed1eb21fe1296f94256d674c58d72693df5fc0379ceR356

crates/precompiles/src/tip20/mod.rs
```diff
@@ -499,9 +506,13 @@ impl TIP20Token {
...
     ) -> Result<()> {
-        let to = Recipient::resolve(call.to)?;
-        self._mint(msg_sender, &to, call.amount)?;
+        let Some((to, total_supply)) =
+            self.validate_mint(msg_sender, call.to, call.amount, call.memo)?
+        else {
+            return Ok(());
+        };
 
+        self._mint(&to, call.amount, total_supply)?;
```
fgimenez: Blocked mint never emits the spec-mandated `Mint(TIP1028Guard, amount)` event, `validate_mint` returns `None` and both `mint()` / `mint_with_memo()` skip the `TIP20Event::Mint` emission, spec ref here https://github.com/tempoxyz/tempo/pull/3791/changes#diff-ecfb80eee8e6f2ece98beed1eb21fe1296f94256d674c58d72693df5fc0379ceR283

crates/precompiles/src/receive_policy_guard/mod.rs
```diff
@@ -0,0 +1,1247 @@
...
+
+    fn assert_unauthorized(result: Result<()>) {
+        assert_eq!(
+            result.unwrap_err(),
+            TIP1028GuardError::unauthorized_claimer().into()
+        );
+    }
+
+    #[test]
+    fn test_claim_transfer() -> eyre::Result<()> {
```
fgimenez: we could merge here `test_claim_self_recovery`, `test_claim_via_recovery_contract`,  `test_claim_binds_recovery_contract`, `test_claim_mint` and add event assertions to the fixture

crates/precompiles/src/tip20/mod.rs
0xrusowsky: imo it's better to propagate and leave as is

crates/precompiles/src/tip20/mod.rs
0xrusowsky: nice catch, ty!

addressed in [`95bd6ab` (this PR)](https://github.com/tempoxyz/tempo/pull/3800/commits/95bd6aba0ce8978698b6ad38602deab5b4caa1bb)

crates/precompiles/src/receive_policy_guard/mod.rs
0xrusowsky: addressed in [`95bd6ab` (this PR)](https://github.com/tempoxyz/tempo/pull/3800/commits/95bd6aba0ce8978698b6ad38602deab5b4caa1bb)

crates/precompiles/src/receive_policy_guard/mod.rs
0xrusowsky: as per this cmnt, i dropped the sentinel `address(1)` here for better UX [`9fe4987` (this PR)](https://github.com/tempoxyz/tempo/pull/3800/commits/9fe498755708fe778a4f5fe4df8ae1900675964b)

crates/precompiles/src/tip1028_escrow/mod.rs
0xrusowsky: addressed in [`9b183cf` (this PR)](https://github.com/tempoxyz/tempo/pull/3800/commits/9b183cfe29a79bff0cc87bae2ba7d4b66bf7ca4f)

crates/precompiles/src/tip20/mod.rs
0xalpharush, line 1: https://github.com/tempoxyz/tempo/blob/4ad8fca1f25c588dbce4a3ee2447bc722e6ffc9b/tips/tip-1028.md?plain=1#L223

This seems to be missing?

0xrusowsky: proposal to address it here: [`96aefd5` (this PR)](https://github.com/tempoxyz/tempo/pull/3800/changes/96aefd553986e3dd783a5157f8234141150aa934)

0xalpharush: mm now this is no longer true. it will always be off-by-one https://github.com/tempoxyz/tempo/blob/4ad8fca1f25c588dbce4a3ee2447bc722e6ffc9b/tips/tip-1028.md?plain=1#L541

0xalpharush: tbh I think we should omit this and dust it manually. that also addresses https://github.com/tempoxyz/tempo/pull/3791#discussion_r3262878536. but ig that we can't transfer to `RECEIVE_POLICY_GUARD_ADDRESS`...

0xalpharush: an alternative would be new tips mint 1 token to guard address 🤔

crates/precompiles/src/address_registry/mod.rs
```diff
@@ -151,7 +151,7 @@ impl AddressRegistry {
     ///
     /// # Errors
     /// - `VirtualAddressUnregistered` — `to` is a virtual address whose `masterId` is not registered
-    pub fn resolve_recipient(&self, to: Address) -> Result<Address> {
+    pub fn resolve_receiver(&self, to: Address) -> Result<Address> {
```
0xKitsune: nit: why are we changing this in this PR?

crates/node/tests/it/gas/snapshots/it__gas__tip20_transfers__tip20_transfer_gas_snapshots.snap
```diff
@@ -2,88 +2,88 @@
...
-"virtual_address::policy:needs_whitelist::amount:some": 62766
-"virtual_address::policy:needs_whitelist::amount:some::rewards_sender_delegates_a": 62766
-"virtual_address::policy:needs_whitelist::amount:some::rewards_sender_delegates_a_with_delta": 328566
-"virtual_address::policy:needs_whitelist::amount:some::rewards_sender_delegates_b": 62766
-"virtual_address::policy:needs_whitelist::amount:some::rewards_sender_delegates_b_with_delta": 328566
-"virtual_address::policy:needs_whitelist::amount:some::rewards_sender_opted_out": 317766
-"virtual_address::policy:needs_whitelist::amount:some::rewards_sender_opted_out_with_delta": 326866
+"policy:all_whitelisted::amount:full::rewards_recipient_opted_out": 54846
+"policy:all_whitelisted::amount:full::rewards_recipient_opted_out_with_delta": 555558
+"policy:all_whitelisted::amount:full::rewards_sender_delegates_a_recipient_opted_out": 54858
```
klkvr, line 7: why are those snapshots different now?

crates/precompiles/src/tip20/mod.rs
```diff
@@ -516,14 +527,18 @@ impl TIP20Token {
...
+    /// If `None`, reads from storage.
+    pub(crate) fn _mint(
+        &mut self,
+        to: &Recipient,
+        amount: U256,
+        total_supply: Option<U256>,
+    ) -> Result<()> {
         let new_supply = total_supply
+            .map(Ok)
+            .unwrap_or_else(|| self.total_supply())?
```
klkvr: similarly to the other PR imo it's fine to just read `total_supply` here

crates/precompiles/src/receive_policy_guard/mod.rs
```diff
@@ -0,0 +1,1123 @@
...
+
+/// Recovery authority for blocked inbound funds.
+#[derive(Debug, Clone, Copy, Default, Storable, PartialEq)]
+#[repr(u8)]
+pub(crate) enum RecoveryMode {
+    #[default]
+    Originator,
+    Receiver,
+    ThirdParty,
+}
```
klkvr: this type should be moved below `ReceivePolicyGuard` impl block

crates/precompiles/src/receive_policy_guard/mod.rs
```diff
@@ -0,0 +1,1123 @@
...
+    }
+
+    /// Returns the account charged for access-key spending limits, if any.
+    pub(crate) fn spending_account(self, recovery_authority: Address) -> Option<Address> {
+        match self {
+            Self::Originator | Self::Receiver => Some(recovery_authority),
+            Self::ThirdParty => None,
+        }
+    }
+}
```
klkvr: it is still not clear to me why for claims handled by a third party spending limit checks do not apply

crates/precompiles/src/receive_policy_guard/mod.rs
```diff
@@ -0,0 +1,1123 @@
...
+            to,
+            amount,
+            recovery_mode,
+            receipt.recoveryAuthority,
+        )?;
+
+        self.emit_event(receipt.claimed_event(receiver, msg_sender, to, amount))
+    }
+
+    pub fn burn_blocked_receipt(&mut self, msg_sender: Address, receipt: Bytes) -> Result<()> {
```
klkvr, line 135: can we document this?

crates/precompiles/src/tip20/mod.rs
```diff
@@ -582,44 +598,53 @@ impl TIP20Token {
...
 
         // Prevent burning from system custody addresses to protect accounting invariants.
-        let hardfork = self.storage.spec();
-        if PROTECTED
-            .iter()
-            .any(|(fork, addrs)| hardfork >= *fork && addrs.contains(&call.from))
+        if !policed_by_guard
+            && PROTECTED
+                .iter()
+                .any(|(fork, addrs)| hardfork >= *fork && addrs.contains(&owner))
```
klkvr: can we document that this is required so that the `PROTECTED` check does not fire on burns invoked through the guard

crates/precompiles/src/receive_policy_guard/mod.rs
```diff
@@ -0,0 +1,1123 @@
...
+
+impl ReceivePolicyGuard {
+    /// One-time storage initialization.
+    pub fn initialize(&mut self) -> Result<()> {
+        self.__initialize()
+    }
+
+    /// Returns the unclaimed amount for a receipt, or zero if unknown or already claimed.
+    pub fn balance_of(&self, receipt: Bytes) -> Result<U256> {
+        let receipt = ClaimReceiptV1::try_from(receipt)?;
```
0xKitsune, line 45: nit: should we pass in the decoded receipt instead of decoding in this function?

crates/precompiles/src/tip20/mod.rs
```diff
@@ -964,6 +997,12 @@ impl TIP20Token {
         self.supply_cap.write(U128_MAX)?;
         self.transfer_policy_id.write(1)?;
 
+        // HACK(T6+): deployer must incur TIP-1028 cold storage (0 -> 1) cost.
+        // We virtually assign 1 unit without increasing supply to warm the address.
+        if self.storage.spec().is_t6() {
+            self.balances[RECEIVE_POLICY_GUARD_ADDRESS].write(U256::ONE)?;
+        }
+
```
klkvr: flagging that this is breaking an invariant and might be problematic for e.g https://github.com/tempoxyz/tempo/pull/4161

crates/node/tests/it/gas/snapshots/it__gas__tip20_transfers__tip20_transfer_gas_snapshots.snap
klkvr: i guess it makes sense because we need to check if receive policy is configured for the receiver on every transfer

crates/precompiles/src/receive_policy_guard/mod.rs
0xKitsune: Yes good catch, moved in https://github.com/tempoxyz/tempo/pull/3800/commits/d876f077242e7b9c3e6d211582f9270ba25c0efb

0xKitsune: Yes good catch, moved in https://github.com/tempoxyz/tempo/pull/3800/commits/d876f077242e7b9c3e6d211582f9270ba25c0efb

crates/precompiles/src/tip20/mod.rs
0xKitsune: Updated to remove this.

crates/precompiles/src/receive_policy_guard/mod.rs
0xKitsune: Originator recovery is metered because claiming to any address other than the original receiver is effectively the originator sending those funds somewhere else.

Receiver recovery is not metered when funds are claimed directly to the receiver, because that is just completing the original inbound transfer. However, receiver recovery is metered when the receiver reroutes the claim to another address, because that becomes equivalent to the receiver sending those funds out.

Third-party recovery currently skips receiver/originator limits because the third party, not the receiver or originator, chooses the claim destination.

If the third party is an account using its own access key, it might make sense to meter the third party’s own spending limit.

crates/precompiles/src/tip20/mod.rs
0xKitsune: Agreed, removed in [cfac277](https://github.com/tempoxyz/tempo/pull/3800/commits/cfac277b479637fb3a57c43c7ac73382a59be5d9)

crates/precompiles/src/tip20/mod.rs
0xKitsune: Updated to remove the special case write to the balances mapping to avoid breaking invariants or having unnecessary mints during TIP20 initialization. We can explore optimizations for this in a subsequent update.

crates/precompiles/src/tip20/mod.rs
```diff
@@ -582,44 +585,52 @@ impl TIP20Token {
...
                 .ok_or(TIP20Error::insufficient_balance(
                     total_supply,
-                    call.amount,
+                    amount,
                     self.address,
                 ))?;
         self.set_total_supply(new_supply)?;
 
-        self.emit_event(TIP20Event::burn_blocked(call.from, call.amount))
+        self.emit_event(TIP20Event::burn_blocked(owner, amount))
```
klkvr, line 636: is it intentionally that we're emitting `BurnBlocked` with `from` different from the transfer event corresponding to the burn?

crates/precompiles/src/receive_policy_guard/mod.rs
klkvr: > If the third party is an account using its own access key, it might make sense to meter the third party’s own spending limit.

yes this is the scenario i was thinking about. if a third party has control over some of the blocked funds it doesn't seem reasonable that any access key it authorized has full control over those funds and spending limits do not apply?

klkvr (approved):

crates/precompiles/src/tip20/mod.rs
0xrusowsky: i think so, cause once funds are blocked they belong to the recovery authority.

so burning a proof requires the recovery authority to be blocked.

cc: @malleshpai to verify my claims

crates/precompiles/src/tip20/mod.rs
```diff
@@ -1106,6 +1169,90 @@ impl TIP20Token {
...
+    /// authorized and the caller should proceed with the normal transfer or mint.
+    pub(crate) fn validate_inbound_or_block(
+        &mut self,
+        originator: Address,
+        to: &Recipient,
+        amount: U256,
+        kind: InboundKind,
+        memo: B256,
+    ) -> Result<bool> {
+        if !self.storage.spec().is_t6() {
```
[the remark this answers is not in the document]
klkvr: we can't change behavior pre-T6

crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -272,6 +394,55 @@ impl TIP403Registry {
...
+            return Err(TIP403RegistryError::invalid_recovery_authority().into());
+        }
+        if msg_sender.is_virtual() {
+            return Err(TIP403RegistryError::virtual_address_not_allowed().into());
+        }
+
+        let recovery_address = call.recoveryAuthority;
+        let (recovery_mode, recovery_write) = RecoveryMode::encode(recovery_address, msg_sender);
+
+        if recovery_address == RECEIVE_POLICY_GUARD_ADDRESS
```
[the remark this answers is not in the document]
klkvr: imo even existing checks are somewhat excessive but there's not much we can do to prevent user from setting authority that doesn't make sense

crates/precompiles/src/tip20/mod.rs
```diff
@@ -996,17 +1015,57 @@ impl TIP20Token {
...
-    /// Ensures that the recipient is authorized to receive mints.
-    /// Additionally (+T3) checks pause state, validates the effective recipient.
-    fn validate_mint(&self, to: &Recipient) -> Result<()> {
+    /// Resolves `to`, checks the issuer role, and ensures TIP-403 mint-recipient authorization.
+    /// Additionally (+T3) checks pause state and validates the effective recipient; also
+    /// (+T6) applies TIP-1028 address-level receive policies.
+    ///
+    /// Returns `Some(to)` when the caller should proceed with the regular mint.
+    /// Returns `None` when funds were minted and blocked, and the caller should return immediately.
+    fn validate_mint(
```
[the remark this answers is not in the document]
klkvr: that one seems right, pre-T6 we will now only read total supply after paused and recipient checks while on main it would be read before

crates/precompiles/src/tip20/mod.rs
```diff
@@ -996,17 +1015,57 @@ impl TIP20Token {
...
-        self.ensure_transfer_authorized(from, to.target)
+        self.ensure_transfer_authorized(from, to.target)?;
+
+        if let Some(spender) = spender {
+            self.consume_allowance(from, spender, amount)?;
+        } else {
+            self.check_and_update_spending_limit(from, amount)?;
+        }
+
+        if self.validate_inbound_or_block(from, &to, amount, InboundKind::TRANSFER, memo)? {
```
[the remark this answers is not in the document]
klkvr: you can always send funds to an intermediate address to bypass the receive policy and escrow contract is just another case of this, not sure if worth fixing explicitly

crates/precompiles/src/receive_policy_guard/mod.rs
```diff
@@ -0,0 +1,1127 @@
...
+        } else if authority == msg_sender {
+            (Self::Receiver, Address::ZERO)
+        } else {
+            (Self::ThirdParty, authority)
+        }
+    }
+
+    /// Resolves the recovery mode for a receipt and resolved receiver.
+    pub(crate) fn from(receipt: &ClaimReceiptV1, receiver: Address) -> Self {
+        if receipt.recoveryAuthority == receipt.originator {
```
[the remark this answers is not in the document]
klkvr: this does seem slightly strange but logic-wise i don't think there's an attack vector here given that originator is less trusted than the recovery authority and thus this edge case just causes some extra checks to be performed

notable edge case here is that if originator == recovery thrid party and their transfer to receiver end up blocked, they won't be able to claim directly to receiver address because it would be triggered as a reroute

crates/precompiles/src/tip20/mod.rs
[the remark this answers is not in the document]
klkvr: fixed in 0702bf743

crates/precompiles/src/tip20/mod.rs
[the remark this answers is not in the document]
0xrusowsky: imo not worth it, as u can whitelist if u want to be super restrictive of who can send u funds.

crates/precompiles/src/receive_policy_guard/mod.rs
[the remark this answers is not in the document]
0xrusowsky: do u think we should revert back to recovery authority == address(0) for originator? or encode recoveryMode too in the proof to avoid any sentinels?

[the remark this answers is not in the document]
0xrusowsky: fixed in [`5ef6e57` (this PR)](https://github.com/tempoxyz/tempo/pull/3800/commits/5ef6e57e8210d0dfe96c91e8d06c01a7a4b63bfc)

0xrusowsky (approved):

PR #3804, opened 2026-05-04
ci(release): add SLSA provenance, SBOM attestation, binary checksum, reproducible build

grandizzy:
Adds SLSA build provenance, SBOM generation + attestation, a bare-binary checksum, `--locked` builds, and the foundation for byte-deterministic reproducible builds — letting downstream consumers verify each release artifact's origin, contents, and integrity, and letting independent rebuilders confirm a published binary corresponds bit-for-bit to its source.

### Supply-chain attestations
- add anchore/sbom-action to generate SPDX-JSON per archive
- add actions/attest in SBOM mode binding the SBOM to archive + binary digests
- add actions/attest in provenance mode (SLSA v1) for archive + binary
- grant id-token/attestations write perms on the build-release job
- record both archive and bare-binary sha256 (`.tar.gz.sha256` + `.sha256`)
- pass `--locked` to `cargo build` to lock the dep graph
- README: replace the prior multi-step recipe with two independent paths under the Security section — Path A (`sha256sum -c` + `gpg --verify`, offline / no GitHub auth) or Path B (`gh attestation verify` against the SLSA provenance). Either path proves the archive came from the tagged commit, signed by tempoxyz. SBOM verification is dropped from the operator recipe.
- `tempoup`: verifies the archive checksum, GPG signature, and SLSA build provenance (via `gh attestation verify`) on every install. Hard-fails by default if `gpg` or `gh` are missing, or if `gh` is not authenticated; pass `--unsafe-skip-verify` to downgrade those (tool-availability) failures to warnings. Cryptographic failures (bad signature, bad attestation) still abort regardless of the flag. Error messages distinguish "gh not found" from "gh not authenticated" so the remediation is actionable. SBOM attestation verification dropped (operators don't consume it).

### Reproducible build (x86_64-unknown-linux-gnu)
- `Dockerfile.reproducible`: hermetic build environment pinned to a specific `rust:1.93.0-bookworm` digest, with `/etc/apt/sources.list` repointed at `snapshot.debian.org` so apt-installed deps (libjemalloc, mold, libclang) can't drift between an in-CI build and a later rebuild. RUSTFLAGS strip every known source of non-determinism (`SOURCE_DATE_EPOCH`, `--remap-path-prefix`, `--build-id=none`, `symbol-mangling-version=v0`, `metadata=`).
- `Cargo.toml`: new `[profile.reproducible]` inheriting `release` with `panic = "abort"`, `strip = "none"`, `debug = "none"`, `codegen-units = 1`, `incremental = false`. Diverges from `maxperf` only in the bits required for byte-determinism — `panic = "abort"` is the deliberate choice for this profile because it eliminates unwinding tables and the associated codegen paths, the most reliable source of byte drift between toolchain versions. The user-shipped binary still uses `maxperf` (with `panic = "unwind"`) so production backtraces are unchanged.
- `scripts/reproducible-build.sh`: single shared docker-build wrapper. Both the in-CI workflow and any external rebuilder run the literal same command so the comparison can never silently diverge through someone editing one site and forgetting the other.
- `.github/workflows/reproducible-build.yml`: standalone workflow with two triggers — `push` to `main` (continuous reproducibility canary on every merged PR) and `workflow_dispatch` with a `ref` input (manual builds for ad-hoc verification by independent rebuilders). Unsigned hash, 7-day artifact retention. 3-attempt retry on the docker step to absorb the dominant transient failure mode (snapshot.debian.org throttling). Concurrency is set so back-to-back merges to `main` cancel earlier in-progress canary runs.

Locally verified that the recipe produces byte-identical hashes (`92b754ce…`) on Mac/OrbStack/Rosetta and on Ubuntu 24.04 native.

**TODO (after tested with an independent rebuilder against this manual workflow):**
- add a `workflow_call` trigger to `reproducible-build.yml` plus a GPG-sign step gated on a `sign` input
- add a `build-reproducible` job in `release.yml` that calls it (`uses: ./.github/workflows/reproducible-build.yml`, `sign: true`)
- so each tagged release ships a signed `.reproducible.sha256` sidecar attached automatically via the existing `gh release create … artifacts/**/*` glob
- extend the README's Security-section recipe with a step 7 that `gpg --verify`s the sidecar and confirms an independent rebuild from source matches it


.github/workflows/release.yml
```diff
@@ -165,6 +178,36 @@ jobs:
...
+          format: spdx-json
+          output-file: ${{ steps.prepare.outputs.archive_name }}.spdx.json
+          # The release workflow uploads SBOMs explicitly via actions/upload-artifact
+          # below, so disable the action's own artifact / release upload to keep
+          # the asset list deterministic.
+          upload-artifact: false
+          upload-release-assets: false
+
+      - name: Attest SBOM
+        uses: actions/attest-sbom@c604332985a26aa8cf1bdc465b92731239ec6b9e # v4.1.0
```
grandizzy: As of version 4.0.0 this action is being deprecated in favor of [actions/attest](https://github.com/action/attest). actions/attest-sbom will continue to function as a wrapper on top of actions/attest for some period of time, but applications should make plans to migrate.

All of the existing action inputs are compatible with the actions/attest interface.

.github/workflows/release.yml
```diff
@@ -165,6 +178,36 @@ jobs:
           echo -n "$GPG_SIGNING_KEY" | base64 --decode | gpg --batch --import
           printf '%s' "$GPG_PASSPHRASE" | gpg --passphrase-fd 0 --pinentry-mode loopback --batch -ab "${{ steps.prepare.outputs.archive_path }}"
 
+      - name: Generate SBOM (SPDX-JSON) for archive
+        uses: anchore/sbom-action@f8bdd1d8ac5e901a77a92f111440fdb1b593736b # v0.20.6
```
grandizzy: new version https://github.com/anchore/sbom-action/releases/tag/v0.24.0

.github/workflows/release.yml
```diff
@@ -165,6 +178,36 @@ jobs:
...
+          # Attest both the archive and the bare binary so consumers can verify
+          # provenance whether they hold the .tar.gz or the unpacked ELF
+          # (containers, distro packages, the independent rebuilder).
+          subject-path: |
+            ${{ steps.prepare.outputs.archive_path }}
+            ${{ steps.prepare.outputs.binary_name }}
+          sbom-path: ${{ steps.prepare.outputs.archive_name }}.spdx.json
+
+      - name: Attest build provenance (SLSA v1)
+        uses: actions/attest-build-provenance@00014ed6ed5efc5b1ab7f7f34a39eb55d41aa4f8 # v3.1.0
```
grandizzy: As of version 4, actions/attest-build-provenance is simply a wrapper on top of [actions/attest](https://github.com/actions/attest).

Existing applications may continue to use the attest-build-provenance action, but new implementations should use actions/attest instead.

README.md
```diff
@@ -144,6 +144,46 @@ Our contributor guidelines can be found in [`CONTRIBUTING.md`](https://github.co
 
 See [`SECURITY.md`](https://github.com/tempoxyz/tempo?tab=security-ov-file). Note: Tempo is still undergoing audit and does not have an active bug bounty. Submissions will not be eligible for a bounty until audits have concluded.
 
+### Verifying release binaries
```
jenpaff, line 147: could we also do this via `tempoup` ?

grandizzy: added in https://github.com/tempoxyz/tempo/pull/3804/changes/611f64a54b926ef94e4e6c7a69bddb328f084fef for SLSA provenance and SBOM. warns if `gh` not installed

0xalpharush:
> push to main (continuous reproducibility canary on every merged PR) 

Do we want to do this check aside from releases?

> The user-shipped binary still uses maxperf (with panic = "unwind") so production backtraces are unchanged.

Isn't the point of this people could run the binaries in production after verifying their integrity?

tempoup/tempoup
```diff
@@ -402,6 +402,34 @@ main() {
...
+        if gh attestation verify "$ARCHIVE_PATH" --repo "$REPO" \
+                --predicate-type https://spdx.dev/Document >/dev/null 2>&1; then
+            info "SBOM attestation verified ✓"
+        else
+            warn "SBOM attestation not found or failed verification."
+            warn "(Older releases may not carry attestations.)"
+        fi
+    else
+        warn "gh CLI not found or not authenticated. Skipping SLSA + SBOM attestation verification."
+        warn "Install gh and run 'gh auth login' to enable full attestation verification."
```
0xalpharush: Wdyt of hard fail and requiring a flag `--unsafe-skip-verify`

README.md
```diff
@@ -144,6 +144,48 @@ Our contributor guidelines can be found in [`CONTRIBUTING.md`](https://github.co
...
+  -p "$ARCHIVE" -p "$ARCHIVE.sha256" -p "$ARCHIVE.asc" \
+  -p "$ARCHIVE.spdx.json" -p "$BIN.sha256"
+
+# 2. Checksum the archive (and, after extraction, the bare binary).
+sha256sum -c "$ARCHIVE.sha256"
+tar xzf "$ARCHIVE" && sha256sum -c "$BIN.sha256"
+
+# 3. GPG signature. See https://docs.tempo.xyz/guide/node/installation#verifying-releases
+#    for the public key, fingerprint, and `gpg --recv-keys` command.
+gpg --verify "$ARCHIVE.asc" "$ARCHIVE"
```
0xalpharush, line 166: Can we not go all-in on sigstore and remove gpg altogether?

grandizzy:
> > push to main (continuous reproducibility canary on every merged PR)
> 
> Do we want to do this check aside from releases?
> 

yeah, what I want to achieve here is to catch early any cache poisoning / source that introduce non determinism or cargo source replacement. This is complemented by an external (non GHA) task that will build and compare hashes

> > The user-shipped binary still uses maxperf (with panic = "unwind") so production backtraces are unchanged.
> 
> Isn't the point of this people could run the binaries in production after verifying their integrity?

for x86_64-linux (and arm-linux) this can be done by switching build-release from the bare runner cargo build to Docker (needs a refactor of the existing release path, which is why I didn't want to add it into this PR, happy to follow up). Mac targets stay non-reproducible per my checks, so we won’t be able to cover all binaries.

tempoup/tempoup
grandizzy: good point, @jenpaff wdyt about the change in installer, would this be OK for partners (they'll need gh installed and authed in order to `tempoup`, otherwise they'll have to pass the `--unsafe-skip-verify` flag)

README.md
grandizzy: I'd keep GPG for now as the offline (no gh required) verification path, wdyt?

tempoup/tempoup
jenpaff: Fine with it- double checking: does one need to be authed for the checksum? And it's only needed for certifying checksum, correct?

grandizzy: indeed, here's the issue https://github.com/cli/cli/issues/11803. will add it and remind to update docs

grandizzy: added in https://github.com/tempoxyz/tempo/pull/3804/changes/e3564303b6a94676e8c67c4c8e837038641e0079

0xalpharush:
> for x86_64-linux (and arm-linux) this can be done by switching build-release from the bare runner cargo build to Docker (needs a refactor of the existing release path, which is why I didn't want to add it into this PR, happy to follow up). Mac targets stay non-reproducible per my checks, so we won’t be able to cover all binaries.

Can follow-up with this. Idt we need to support MacOS for reproducible builds

README.md
```diff
@@ -144,6 +144,48 @@ Our contributor guidelines can be found in [`CONTRIBUTING.md`](https://github.co
...
+
+# 4. GitHub release attestation (tag + asset digests, signed by GitHub).
+gh release verify "$TAG" --repo tempoxyz/tempo
+
+# 5. SLSA build provenance (proves the workflow + commit that built it).
+#    Both the archive and the bare binary are listed as subjects of the
+#    same attestation, so either path verifies.
+gh attestation verify "$ARCHIVE" --repo tempoxyz/tempo \
+  --predicate-type https://slsa.dev/provenance/v1
+gh attestation verify "$BIN"     --repo tempoxyz/tempo \
```
0xalpharush: I think the instructions could be reduced down to this step? I don't anticipate the SBOM being relevant to operators

grandizzy: updated in https://github.com/tempoxyz/tempo/pull/3804/changes/f8740778a540059a9f1aedbd9f9740452dc624d8

0xalpharush (approved):
Leaving a few items as follow-up
https://github.com/tempoxyz/tempo/pull/3804#issuecomment-4388999597

PR #3807, opened 2026-05-04
refactor(consensus): rename `leader` to `proposer` and clarify sentinel-context comment

fgimenez:
ref https://linear.app/tempoxyz/issue/SIGP-253

- Renames the `Context` field in TIP-1031 from `leader` to `proposer` so the spec matches the codebase, where `proposer` is the established term.
- Clarifies the `CertifiableBlock::context()` `None`-branch comment to state that it returns a deterministic sentinel (the prior comment incorrectly suggested a post-T4 panic).

### DIFF
--- a/crates/commonware-node/src/consensus/block.rs
+++ b/crates/commonware-node/src/consensus/block.rs
@@ -151,8 +151,13 @@ impl commonware_consensus::CertifiableBlock for Block {
                 parent: (View::new(ctx.parent_view), self.parent_digest()),
             },
             None => {
-                // Pre-T4: Unused deterministic sentinel.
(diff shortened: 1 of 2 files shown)


SuperFluffy:
Let's rather adjust the spec to mention `proposer`. The term is clearer and more pervasive in our codebase.

fgimenez:
> Let's rather adjust the spec to mention `proposer`. The term is clearer and more pervasive in our codebase.

@SuperFluffy makes sense, updated ptal

SuperFluffy (approved):
Looks good, please update PR title and body prior to merge.

PR #3808, opened 2026-05-04
docs(tip-0001): add tempo transaction spec

legion2002:
Adds TIP-0001 by copying the existing Tempo Transactions spec into the repo as a retroactive TIP.

Tempo Transactions shipped at genesis and are already live on mainnet, so this PR makes the existing spec discoverable alongside the rest of the TIP set. It also points readers to the related follow-on transaction TIPs: TIP-1007, TIP-1009, TIP-1011, and TIP-1020.


tips/tip-0001.md
```diff
@@ -0,0 +1,1299 @@
...
+- **Cost**: Effectively doubles the free account-access work for sponsored transactions
+- **Mitigation**: Cost is still bounded to a single additional account read
+
+#### Comparison to Ethereum
+
+The introduction of EIP-7702 delegated accounts already created complex cross-transaction dependencies in the mempool, which prevents static pool checks from being fully useful. A single transaction can invalidate multiple others by spending balances of multiple accounts.
+
+**Assessment:** While this transaction type introduces additional pre-execution validation costs, all costs are bounded to reasonable limits. The mempool complexity issues around cross-transaction dependencies already exist in Ethereum due to EIP-7702 and accounts with code, so the incremental cost from this transaction type is acceptable given these existing constraints.
+
+## T2 -> T3 Migration
```
jenpaff, line 1266: unsure if we still need this section since we're at T4 now

jenpaff (approved):

tips/tip-0001.md
legion2002: yup, can update in the docs too

PR #3810, opened 2026-05-04
refactor: offload inbound checks to `Recipient`

0xrusowsky:


crates/precompiles/src/tip20/mod.rs
```diff
@@ -409,21 +394,14 @@ impl TIP20Token {
...
-            self.emit_event(hop)?;
+        let mut to = Recipient::resolve(call.to)?;
+        self._mint(msg_sender, &mut to, call.amount, B256::ZERO)?;
+
+        self.emit_event(to.build_mint_event(call.amount))?;
+        if let Some(ev) = to.build_blocked_event(self.address) {
+            self.emit_event(ev)?;
+        } else if let Some(ev) = to.build_virtual_transfer_event(call.amount) {
+            self.emit_event(ev)?;
         }
```
0xrusowsky: i think we can probably move this logic into `Recipient` too since it is aware of both virtual addresses and blocked transfers (and this pattern repeats in all transfer/mint vraiants), but wanted to keep the changes focused for this first pass

PR #3811, opened 2026-05-04
fix(precompiles): fixed-size array packing

0xalpharush:
## Summary

Two paired correctness bugs in `[T; N]` storage codegen for element types
where `T::BYTES <= 16` but `32 % T::BYTES != 0` (in practice: `[U96; N]`,
`[I96; N]`, and odd-sized `[FixedBytes<N>; M]`).

## The bugs

1. **Bulk and indexed paths use mismatched packing rules.**
   `storable_primitives::is_packable` required `32 % byte_count == 0`,
   excluding 12-byte types from the packed bulk path. `ArrayHandler::compute_handler`
   gates only on `T::BYTES <= 16` and uses the packed path. A bulk write
   followed by an indexed read on the same `[U96; N]` returns the wrong
   element; an indexed write mutates the wrong slot.

2. **Packed slot-count formula undercounts the trailing partial slot.**
   `(N * byte_count).div_ceil(32)` treats storage as contiguous and undercounts when
   `byte_count` does not divide 32. `[U96; 5]` reported 2 slots instead
   of 3, and the bulk-store loop dropped the final element.

Both bugs are currently dormant (no in-tree consumer of `[U96; N]` /
`[I96; N]`), but a future precompile schema using these types would hit
silent storage corruption.

## Changes

- `test`: regression tests for `[U96; 2]` (bulk-vs-indexed agreement)
  and `[U96; 5]` (slot-count formula).
- `fix`: relax `is_packable` to `byte_count < 32`, matching the trait-level
  `Layout::is_packable()` already used by indexed access.
- `fix`: route both array-codegen call sites through
  `packing::calc_packed_slot_count` (the same helper used by `Vec`/`Set`).

### DIFF
--- a/crates/precompiles/src/storage/types/array.rs
+++ b/crates/precompiles/src/storage/types/array.rs
@@ -212,6 +212,7 @@ mod tests {
         storage::{Layout, LayoutCtx, PrecompileStorageProvider, StorageCtx},
         test_util::setup_storage,
     };
+    use alloy::primitives::aliases::U96;
     use proptest::prelude::*;
 
     // Strategy for generating random U256 slot values that won't overflow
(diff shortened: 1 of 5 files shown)


crates/precompiles/src/storage/types/array.rs
0xrusowsky, line 1: fine for me to keep these unit tests, but could we also update the integration test `fn test_arrays_layout()` in `crates/precompiles/tests/storage_tests/solidity/primitives.rs`?

those are verified vs solc's storage layout

0xrusowsky (commented):
nice catch, ty!

crates/precompiles/src/storage/types/array.rs
0xalpharush: https://github.com/tempoxyz/tempo/pull/3811/changes/137bad3c61a397cd8959a08d481ff3b88f615b5c

Can follow up in another PR for additional test coverage of other types

0xrusowsky (approved):

crates/precompiles/tests/storage_tests/solidity/testdata/arrays.layout.json
```diff
@@ -1,98 +1,136 @@
 {
```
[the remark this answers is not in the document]
0xrusowsky: the PR updates the file already

PR #3812, opened 2026-05-04
chore: cleanup dkg state migration code

hamdiallam:
* Remove migration logic
* Remove syncing_players metric which is always zero
* Remove `syncers` from State and hide wire format within Read/Write impls

### DIFF
--- a/crates/commonware-node/src/dkg/manager/actor/state.rs
+++ b/crates/commonware-node/src/dkg/manager/actor/state.rs
@@ -522,8 +522,7 @@ impl Builder {
             let initial_state = match initial_state {
                 None => {
                     return Err(eyre!(
-                        "states metadata was empty, legacy journal was empty, \
-                         and initializer was not set"
+                        "states metadata was empty and initializer was not set"
                     ));
                 }
                 Some(initial_state) => initial_state
--- a/crates/commonware-node/src/dkg/manager/actor/state.rs
+++ b/crates/commonware-node/src/dkg/manager/actor/state.rs
@@ -1,6 +1,5 @@
 use std::{
     collections::{BTreeMap, HashMap},
-    net::SocketAddr,
     num::{NonZeroU16, NonZeroU32, NonZeroUsize},
 };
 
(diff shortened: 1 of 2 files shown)


crates/commonware-node/src/dkg/manager/actor/state.rs
```diff
@@ -747,7 +675,7 @@ impl EncodeSize for State {
             + self.output.encode_size()
             + self.share.encode_size()
             + self.players.encode_size()
-            + self.syncers.encode_size()
+            + self.legacy_syncers().encode_size()
```
SuperFluffy, line 680: Add comment that this must remain for compatibility reasons.

crates/commonware-node/src/dkg/manager/actor/state.rs
```diff
@@ -759,7 +687,7 @@ impl Write for State {
         self.output.write(buf);
         self.share.write(buf);
         self.players.write(buf);
-        self.syncers.write(buf);
+        self.legacy_syncers().write(buf);
```
SuperFluffy, line 694: Add comment that this must remain for compatibility reasons.

crates/commonware-node/src/dkg/manager/actor/state.rs
```diff
@@ -771,97 +699,28 @@ impl Read for State {
...
-    epoch: Epoch,
-    seed: Summary,
-    output: Output<MinSig, PublicKey>,
-    share: Option<Share>,
-    dealers: ordered::Map<PublicKey, SocketAddr>,
-    players: ordered::Map<PublicKey, SocketAddr>,
-    syncers: ordered::Map<PublicKey, SocketAddr>,
-    is_full_dkg: bool,
-}
+        // Legacy syncers (Read but ignored. Should be empty)
```
SuperFluffy: Add comment that this must remain for compatibility reasons.

This comment is a bit too weak.

SuperFluffy (approved):

PR #3813, opened 2026-05-04
fix(p2p-proxy): limit response to 1MB; early-exit on missing hash

0xalpharush:
similar to https://github.com/tempoxyz/tempo/pull/3766 but for `GetBlockBodies`

### DIFF
--- a/bin/tempo/src/p2p_proxy.rs
+++ b/bin/tempo/src/p2p_proxy.rs
@@ -6,6 +6,7 @@ use alloy::{
     providers::{Provider, ProviderBuilder},
     rpc::client::BatchRequest,
 };
+use alloy_rlp::Encodable;
 use clap::Parser;
 use eyre::{Context, Result};
 use futures::StreamExt;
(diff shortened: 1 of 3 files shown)


bin/tempo/src/p2p_proxy.rs
```diff
@@ -671,19 +673,30 @@ async fn resolve_bodies(
...
+        };
 
-        if let Some(body) = fetch_body_by_hash(provider, cache, hash).await {
-            bodies.push(body);
+        if next_total_bytes > BODY_RESPONSE_SIZE_LIMIT {
+            break;
         }
+
+        total_bytes = next_total_bytes;
+        bodies.push(body);
```
mattsse: we need to swap the order here, otherwise we can't serve 1 body larger than the soft limit and a body can be up to ~8MB atm I believe (in theory)

mattsse (changes requested):
smol nit

we can go over the softlimit for 1 body,

bin/tempo/src/p2p_proxy.rs
0xalpharush: https://github.com/tempoxyz/tempo/pull/3813/commits/67ef2bf12f1f07c6c825ef14414bdf6c68f6d21d

mattsse (approved):

PR #3824, opened 2026-05-05
fix(consensus): reject system txs outside required tail

Nepali123:
Enforces a strict consensus invariant for Tempo system transactions: blocks must contain exactly the expected number of system transactions at the end of the block, and no system transaction may appear before the required tail window.

This prevents acceptance of blocks with extra or misplaced system transactions that can bypass intended block-body constraints and increase validation load. Adds consensus tests for misplaced system transactions and for rejecting any system transactions once T4 is active.


SuperFluffy:
This is already covered by https://github.com/tempoxyz/tempo/blob/addc044bf5ca597f73a54b873f660454fe082fa6/crates/evm/src/block.rs#L428-L433:

1. in `validate_block_pre_execution` we ensure that we have the expected number of system transactions in the tail.
2. in `validate_tx` we ensure that regular transactions never follow system transactions (effectively this PR).

PR #3828, opened 2026-05-05
feat: TIP-1035

klkvr:
Closes CHAIN-1144

Adds `isImplicitlyApproved` method to `AddressRegistry` and changes stablecoind DEX to use `system_transfer_from` post-T5.

### DIFF
--- a/crates/precompiles/src/address_registry/mod.rs
+++ b/crates/precompiles/src/address_registry/mod.rs
@@ -16,9 +16,30 @@ use alloy::{
     primitives::{Address, FixedBytes, keccak256},
     sol_types::SolValue,
 };
-pub use tempo_contracts::precompiles::{AddrRegistryError, AddrRegistryEvent, IAddressRegistry};
+use tempo_chainspec::hardfork::TempoHardfork;
+pub use tempo_contracts::precompiles::{
+    AddrRegistryError, AddrRegistryEvent, IAddressRegistry, STABLECOIN_DEX_ADDRESS,
+    TIP_FEE_MANAGER_ADDRESS,
+};
 use tempo_precompiles_macros::{Storable, contract};
 pub use tempo_primitives::{MasterId, TempoAddressExt, UserTag};
+
+/// TIP-1035 Implicit Approval List.
+///
+/// Precompiles on this list are authorized to call
+/// [`crate::tip20::TIP20Token::system_transfer_from`], pulling TIP-20 tokens from a user without a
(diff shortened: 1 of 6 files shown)


crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -179,15 +179,23 @@ impl StablecoinDEX {
...
-    fn transfer_from(&mut self, token: Address, from: Address, amount: u128) -> Result<()> {
-        TIP20Token::from_address(token)?.transfer_from(
-            self.address,
-            ITIP20::transferFromCall {
-                from,
-                to: self.address,
-                amount: U256::from(amount),
-            },
-        )?;
+    fn transfer_from(&mut self, token: Address, sender: Address, amount: u128) -> Result<()> {
```
0xrusowsky, line 182: should we create this fn as a new TIP20 helper (i.e `transfer_from_internal(&mut self, caller: Address, from: Address, amount: U256)` which routes based on  `AddressRegistry::is_implicitly_approved()`?

it would ensure that impl and registry are always aligned + that `to` is always the caller precompile

crates/precompiles/src/address_registry/mod.rs
```diff
@@ -16,10 +16,31 @@ use alloy::{
...
+};
 use tempo_precompiles_macros::{Storable, contract};
 pub use tempo_primitives::{MasterId, TempoAddressExt, UserTag};
 
+/// TIP-1035 Implicit Approval List.
+///
+/// Precompiles on this list are authorized to call
+/// [`crate::tip20::TIP20Token::system_transfer_from`], pulling TIP-20 tokens from a user without a
+/// prior `approve()`. The list is gated on `TempoHardfork::T5`; before activation it is empty.
+pub const IMPLICIT_APPROVAL_LIST: &[Address] = &[TIP_FEE_MANAGER_ADDRESS, STABLECOIN_DEX_ADDRESS];
```
legion2002, line 32: Can we also add `TIP_20_ESCROW_PRECOMPILE` to this, ig it also needs to be updated in the TIP itself.
CC: @danrobinson

klkvr: we don't yet have this const defined so i'd suggest we do it in scope of TIP-1034

crates/precompiles/src/stablecoin_dex/mod.rs
klkvr: added here be93800

0xrusowsky (approved):

legion2002 (approved):

PR #3833, opened 2026-05-06
ci: use foundry no-default build for specs

emmajam:
Unpins the specs workflow from the temporary no-OP Foundry SHA and checks out Foundry master again. Forge is now built with `--no-default-features`, matching the supported OP-disabled lane from foundry-rs/foundry#14612.

The patch script now resolves Foundry metadata with the same no-default feature set so OP-only dependencies do not enter the patched lockfile before the build.

Validation:
- `bash -n scripts/foundry-patch.sh`
- `shellcheck scripts/foundry-patch.sh`
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/specs.yml")'`
- fresh-worktree patch against Foundry master, then `cargo +1.94 tree -p forge --no-default-features` with no `op-*` / `alloy-op-*` matches

### DIFF
--- a/.github/workflows/specs.yml
+++ b/.github/workflows/specs.yml
@@ -16,8 +16,6 @@ on:
 env:
   CARGO_TERM_COLOR: always
   RUSTC_WRAPPER: "sccache"
-  # Pinned to mablr/alloy_bump_without_op (alloy 2.0.4 + OP support stripped)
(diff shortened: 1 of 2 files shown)


emmajam:
Matching Workflows PR: tempoxyz/workflows#434

zerosnacks (approved):
lgtm

0xrusowsky (approved):

PR #3836, opened 2026-05-06
feat: enforce tranfer from msg.sender in system_transfer_from

klkvr:


crates/precompiles/src/tip20/mod.rs
```diff
@@ -780,6 +780,10 @@ impl TIP20Token {
             return Err(TIP20Error::unauthorized().into());
         }
 
+        if self.storage.msg_sender() != Some(from) {
```
0xalpharush, line 783: Should this be an assert?

PR #3838, opened 2026-05-06
feat(account-keychain): implement TIP-1053 nonces

legion2002:
Implements TIP-1053 key authorization nonces behind T5 using the simplified nonce-presence semantics from the spec PR. Present nonce fields are included in the signed key authorization and consumed per account on successful authorization; absent nonce fields preserve legacy byte-equivalent encodings. Adds T5-gated AccountKeychain selectors to authorize with a nonce, burn a nonce, and query nonce consumption.


crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -263,6 +314,10 @@ impl AccountKeychain {
             None
         };
 
+        if let Some(nonce) = nonce {
+            self.consume_key_authorization_nonce(msg_sender, nonce)?;
```
legion2002: addressed in https://github.com/tempoxyz/tempo/pull/3838/changes/e3d525868f9d7ae14e68cc41d6b5aa598078c4c0

crates/revm/src/handler.rs
```diff
@@ -347,15 +348,24 @@ fn calculate_key_authorization_gas(
...
 
+        if has_t5_nonce {
+            num_sstores += 1;
+        }
+
         let sstore_cost = gas_params.get(GasId::sstore_set_without_load_cost());
         let mut total_gas = sig_gas + sload_cost + sstore_cost * num_sstores + BUFFER;
 
+        if has_t5_nonce {
+            total_gas += sload_cost;
```
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/4b3a0acf472e38e3180969b08b90df2b96b09670

tips/verify/lib/tempo-std
0xrusowsky, line 1: this requires a separate PR to tempo-std (changes should be reverted)

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -198,13 +201,49 @@ impl AccountKeychain {
...
+        call: authorizeKeyWithNonceCall,
+    ) -> Result<()> {
+        self.authorize_key_inner(
+            msg_sender,
+            call.keyId,
+            call.signatureType,
+            call.config,
+            Some(call.nonce),
+        )
+    }
```
0xrusowsky: nit (don't need to act on it, feel free to ignore):
personally i'd prefer a single `fn authourize_key(&mut self, msg_sender, key_id, sig_ty, config, nonce)` and have the dispatcher flatten the `SolCall` with/without nonce

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -263,6 +302,15 @@ impl AccountKeychain {
             None
         };
 
+        // Nonce-bearing authorizations must commit atomically: if any later key,
+        // restriction, or event write fails, the TIP-1053 nonce must remain reusable.
+        // No-nonce authorizations keep the legacy mutation/error ordering unchanged.
+        let checkpoint = nonce.map(|_| self.storage.checkpoint());
```
0xrusowsky: why do we need to checkpoint here? afaik any error after this point propagates to the precompile boundary, so revm should revert the whole frame anyway.

this seems unnecessary unless there's a caller that swallows and error and continues (which doesn't seem to be the case)

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -977,6 +1053,32 @@ impl AccountKeychain {
...
+        }
+
+        let transaction_key = self.transaction_key.t_read()?;
+        if !transaction_key.is_zero() {
+            let current_timestamp = self.storage.timestamp().saturating_to::<u64>();
+            self.load_active_key(msg_sender, transaction_key, current_timestamp)?;
+        }
+
+        Ok(())
+    }
```
0xrusowsky: this is almost identical to `fn ensure_admin_caller()` i think it is worth unifying into a single fn:

```rs
   fn ensure_account_caller(&self, msg_sender: Address, allow_active_key: bool) -> Result<()> {
       let transaction_key = self.transaction_key.t_read()?;

       if !transaction_key.is_zero() {
           if allow_active_key {
               self.load_active_key(msg_sender, transaction_key, current_timestamp)?;
           } else {
               return Err(AccountKeychainError::unauthorized_caller().into());
           }
       }

       if self.storage.spec().is_t2() {
           let tx_origin = self.tx_origin.t_read()?;
           if tx_origin.is_zero() || tx_origin != msg_sender {
               return Err(AccountKeychainError::unauthorized_caller().into());
           }
       }

       Ok(())
   }
```

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -1379,6 +1481,222 @@ mod tests {
...
+
+    fn unrestricted_restrictions() -> KeyRestrictions {
+        KeyRestrictions {
+            expiry: u64::MAX,
+            enforceLimits: false,
+            limits: vec![],
+            allowAnyCalls: true,
+            allowedCalls: vec![],
+        }
```
0xrusowsky: this is equivalent to `tempo_alloy::provider::keychain::KeyRestrictions::default().into()`

also note that the alloy `KeyRestrictions` struct has some useful builders that you could leverage in tests, and simply `.into()`

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -1379,6 +1481,222 @@ mod tests {
...
+        match error {
+            TempoPrecompileError::AccountKeychainError(e) => {
+                assert!(
+                    matches!(e, AccountKeychainError::KeyAuthorizationNonceAlreadyUsed(_)),
+                    "Expected KeyAuthorizationNonceAlreadyUsed error, got: {e:?}"
+                );
+            }
+            _ => panic!("Expected AccountKeychainError, got: {error:?}"),
+        }
+    }
```
0xrusowsky: replace for:
```rs
assert_eq!(error, AccountKeychainError::key_authorization_nonce_already_used().into()); 
```

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -290,11 +338,31 @@ impl AccountKeychain {
...
+
+    /// Burns a TIP-1053 nonce without authorizing a key.
+    pub fn burn_key_authorization_nonce(
+        &mut self,
+        msg_sender: Address,
+        call: burnKeyAuthorizationNonceCall,
+    ) -> Result<()> {
+        if !self.storage.spec().is_t5() {
+            return Err(AccountKeychainError::invalid_key_authorization_nonce().into());
+        }
```
0xrusowsky: this is not necessary, cause we already hardfork-gate the fn in the dispatcher, and it has no internal callers.

also rather than an error it would be more of a no-op

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -198,13 +201,49 @@ impl AccountKeychain {
...
+        config: KeyRestrictions,
+        nonce: Option<B256>,
+    ) -> Result<()> {
+        let config = &config;
         self.ensure_admin_caller(msg_sender)?;
         let is_t3 = self.storage.spec().is_t3();
 
+        if nonce.is_some() && !self.storage.spec().is_t5() {
+            return Err(AccountKeychainError::invalid_key_authorization_nonce().into());
+        }
```
0xrusowsky: technically unreachable as it's already hardfork-gated by the dispatcher

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -263,6 +302,15 @@ impl AccountKeychain {
...
         };
 
+        // Nonce-bearing authorizations must commit atomically: if any later key,
+        // restriction, or event write fails, the TIP-1053 nonce must remain reusable.
+        // No-nonce authorizations keep the legacy mutation/error ordering unchanged.
+        let checkpoint = nonce.map(|_| self.storage.checkpoint());
+
+        if let Some(nonce) = nonce {
+            self.consume_key_authorization_nonce(msg_sender, nonce)?;
+        }
```
klkvr: i think it's worth clarifying what trust assumptions are we operating under here

we now do an extra SSTORE for key authorizations with nonce which should only help in cases when you've signed 2 different authorizations with the same nonce for some reason, which seems like a weird thing to happen?

it's also not entirely clear to me how exactly nonces help seurity-wise given that end user has no idea about what nonces are and even if we display it likely won't keep track of them

crates/precompiles/src/account_keychain/mod.rs
klkvr: yeah no need for checkpoints here

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -977,6 +1053,32 @@ impl AccountKeychain {
         Ok(())
     }
 
+    fn ensure_nonce_burn_caller(&self, msg_sender: Address) -> Result<()> {
+        let tx_origin = self.tx_origin.t_read()?;
+        if tx_origin.is_zero() || tx_origin != msg_sender {
+            return Err(AccountKeychainError::unauthorized_caller().into());
+        }
```
klkvr: why is this check necessary?

crates/revm/src/handler.rs
```diff
@@ -347,15 +352,24 @@ fn calculate_key_authorization_gas(
             num_limits
         };
 
+        let has_t5_nonce = spec.is_t5() && key_auth.has_nonce();
```
```suggestion
        let has_t5_nonce = key_auth.has_nonce();
```

tx would be rejected if it's not T5 anyway

crates/transaction-pool/src/paused.rs
```diff
@@ -215,29 +220,38 @@ impl PausedFeeTokenPool {
         for meta in self.by_token.values_mut() {
             let before = meta.entries.len();
             meta.entries.retain(|entry| {
-                let Some(subject) = entry.tx.transaction.keychain_subject() else {
+                if let Some(subject) = entry.tx.transaction.keychain_subject() {
```
klkvr: why was this refactored? can we keep `let Some(subject)` pattern?

crates/precompiles/src/account_keychain/mod.rs
legion2002: yeah I added it initially just for peace of mind, because the whole KeyAuthorization OOG case was a bit messy, because of the partial write bug we solved in the T1B hardfork.

But think it should be safe to remove.

tips/verify/lib/tempo-std
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/a954aad50d7b5422204e93ad5d3417f3f7b121f30

crates/precompiles/src/account_keychain/mod.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/a954aad50d7b5422204e93ad5d3417f3f7b121f30

crates/precompiles/src/account_keychain/mod.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/a954aad50d7b5422204e93ad5d3417f3f7b121f30

crates/precompiles/src/account_keychain/mod.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/a954aad50d7b5422204e93ad5d3417f3f7b121f30

crates/precompiles/src/account_keychain/mod.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/a954aad50d7b5422204e93ad5d3417f3f7b121f30

crates/precompiles/src/account_keychain/mod.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/a954aad50d7b5422204e93ad5d3417f3f7b121f30

crates/precompiles/src/account_keychain/mod.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/a954aad50d7b5422204e93ad5d3417f3f7b121f30

crates/revm/src/handler.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/a954aad50d7b5422204e93ad5d3417f3f7b121f30

crates/transaction-pool/src/paused.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/a954aad50d7b5422204e93ad5d3417f3f7b121f30

crates/precompiles/src/account_keychain/mod.rs
legion2002: It's a security guardrail stops a random contract from calling into the account keychain whenever it is called by the access key.
We require root addresses to directly call the keychain precompile if they want to do any admin operations.

crates/precompiles/src/account_keychain/mod.rs
legion2002: The main reason to add the nonce is so that root keys have the ability to revoke signed key authorizations that are not onchain yet.

Doing this is complicated in viem right now. If a user decides to cancel an access key after signing it, they have to sign the same access key again, and revoke it onchain to prevent the DApp from using it again.

crates/precompiles/src/account_keychain/mod.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/cf505300fb207217932c9aaadbbdf9cd75f35ed3

crates/precompiles/src/account_keychain/mod.rs
klkvr: > they have to sign the same access key again, and revoke it onchain to prevent the DApp from using it again

this doesn't sound that bad (still one tx), and we could also allow revoking non-authorized keys

but my question was about why is it necessary to consume the nonce when key is authorized? you can't isn't nonce only useful before the key authorization lands on chain?

0xrusowsky (approved):
changes in #3914 make sense to me.

pending gas snapshots update to clear CI

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -281,20 +295,41 @@ impl AccountKeychain {
...
         )?;
 
+        if let Some(witness) = witness {
+            self.emit_event(AccountKeychainEvent::KeyAuthorizationWitness(
+                IAccountKeychain::KeyAuthorizationWitness {
+                    account: msg_sender,
+                    witness,
+                },
+            ))?;
+        }
```
klkvr, line 310: do we actually need the event?

klkvr (approved):

crates/precompiles/src/account_keychain/mod.rs
legion2002: Think good to have it, enables more use cases of the witness field

PR #3840, opened 2026-05-06
fix(storage): cleanup tail in dyn types

0xrusowsky:
## Motivation

Overwriting a dynamic storable (`Vec<T>`, `String`, `Bytes`) with a shorter value left stale tail slots populated, so subsequent reads observed garbage past the new length.

## Solution

Ensure that "shrinking" writes on dyn types clear their stale tails.

Additionally, introduces a new sentinel `LayoutCtx::INIT` for hot paths that know the destination is virgin (`Vec::push`), letting them skip the extra SLOAD + cleanup.

### DIFF
--- a/crates/precompiles-macros/src/storable.rs
+++ b/crates/precompiles-macros/src/storable.rs
@@ -547,8 +547,10 @@ fn gen_store_impl(fields: &[(&Ident, &Type)], packing: &Ident) -> TokenStream {
                         storage.store(base_slot + ::alloy::primitives::U256::from(offset), pending_val)?;
                         pending_offset = None;
                     }
-                    // Dynamic fields inherit ctx so LayoutCtx::INIT propagates; Static ones ignore it.
+                    // Dynamic fields need INIT to skip stale-tail cleanup on virgin storage.
+                    // Static fields ignore INIT, so we always use FULL for them.
                     let ctx = if <#ty as crate::storage::StorableType>::IS_DYNAMIC { ctx } else { crate::storage::LayoutCtx::FULL };
+                    debug_assert!(ctx.is_full(), "Struct types can only use full-slot LayoutCtx (FULL or INIT)");
                     <#ty as crate::storage::Storable>::store(&self.#name, storage, #slot_addr, ctx)?;
                 }
 
@@ -581,8 +583,10 @@ fn gen_store_impl(fields: &[(&Ident, &Type)], packing: &Ident) -> TokenStream {
                         pending_offset = None;
                     }
                 } else {
-                    // Dynamic fields inherit ctx so LayoutCtx::INIT propagates; Static ones ignore it.
+                    // Dynamic fields need INIT to skip stale-tail cleanup on virgin storage.
+                    // Static fields ignore INIT, so we always use FULL for them.
                     let ctx = if <#ty as crate::storage::StorableType>::IS_DYNAMIC { ctx } else { crate::storage::LayoutCtx::FULL };
+                    debug_assert!(ctx.is_full(), "Struct types can only use full-slot LayoutCtx (FULL or INIT)");
                     <#ty as crate::storage::Storable>::store(&self.#name, storage, #slot_addr, ctx)?;
                 }
             }}
--- a/crates/precompiles-macros/src/storable.rs
+++ b/crates/precompiles-macros/src/storable.rs
@@ -127,7 +127,7 @@ fn derive_struct_impl(input: &DeriveInput, data_struct: &DataStruct) -> syn::Res
                 ctx: crate::storage::LayoutCtx
             ) -> crate::error::Result<Self> {
                 use crate::storage::Storable;
-                debug_assert_eq!(ctx, crate::storage::LayoutCtx::FULL, "Struct types can only be loaded with LayoutCtx::FULL");
+                debug_assert!(ctx.is_full(), "Struct types can only be loaded with a full-slot LayoutCtx (FULL or INIT)");
 
                 #load_impl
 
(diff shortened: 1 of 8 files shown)


crates/precompiles/src/storage/types/vec.rs
```diff
@@ -303,9 +297,18 @@ where
...
+        if T::BYTES <= 16 {
+            let mut elem_slot = Self::compute_handler(self.data_slot(), self.address, length);
+            elem_slot.write(value)?;
+        } else {
+            // Handlers always use `FULL` ctx:
+            // `T::Handler::write(v)` → `self.as_slot().write(v)` → `Slot::<T>::new(s, a).write(v)`.
+            // Since the slot we push to is guaranteed empty, we build the `Slot<T>` directly to
+            // thread `INIT` into `T::store` and skip its tail-cleanup SLOAD for dynamic types.
+            let elem_slot = self.data_slot() + U256::from(length * T::SLOTS);
+            Slot::<T>::new_with_ctx(elem_slot, LayoutCtx::INIT, self.address).write(value)?;
```
0xrusowsky: NOTE: i believe this workaround is safe because of the reasons explained in the cmnt

in order to use `Self::compute_handler` with `LayoutCtx::INIT` we'd have to change all non-primitive handlers to use and hold ctx (which currently ignore)

klkvr: can we just call `T::store` directly here?

crates/precompiles/src/storage/types/vec.rs
```diff
@@ -394,19 +397,45 @@ pub(crate) fn calc_data_slot(len_slot: U256) -> U256 {
...
+    if from >= to {
+        return Ok(());
+    }
+    if T::BYTES <= 16 {
+        // Vec elements can't be split across slots.
+        let from_slots = calc_packed_slot_count(from, T::BYTES);
+        let to_slots = calc_packed_slot_count(to, T::BYTES);
+        for slot_idx in from_slots..to_slots {
+            storage.store(data_start + U256::from(slot_idx), U256::ZERO)?;
+        }
```
klkvr, line 428: looks like if `from_slots == to_slots` this won't clear anything?

crates/precompiles-macros/src/storable.rs
```diff
@@ -579,7 +581,9 @@ fn gen_store_impl(fields: &[(&Ident, &Type)], packing: &Ident) -> TokenStream {
                         pending_offset = None;
                     }
                 } else {
-                    <#ty as crate::storage::Storable>::store(&self.#name, storage, #slot_addr, crate::storage::LayoutCtx::FULL)?;
+                    // Dynamic fields inherit ctx so LayoutCtx::INIT propagates; Static ones ignore it.
+                    let ctx = if <#ty as crate::storage::StorableType>::IS_DYNAMIC { ctx } else { crate::storage::LayoutCtx::FULL };
+                    <#ty as crate::storage::Storable>::store(&self.#name, storage, #slot_addr, ctx)?;
```
klkvr: is the assumption here is that there's a dynamic field within a struct than we have either `FULL` or `INIT` ctx? and thus it's ok to propagate it directly? i'd document that explicitly and add an assertion

0xrusowsky: addressed in [`273b2c3` (this PR)](https://github.com/tempoxyz/tempo/pull/3840/commits/273b2c3345d81673370bee0a0d30508f7f948ad3)

crates/precompiles/src/storage/types/vec.rs
0xrusowsky: yes, and i think that's correct (i should probably document it better).

my reasoning is as follows:
- delete always uses `clear_elements(.., from=0, to=length)`, which means it always deletes smth
- tail cleanup doesn't delete the same element cause it is overriden by the later SSTORE, so it would be redundant

klkvr (approved):

crates/precompiles/src/storage/types/vec.rs
0xrusowsky: addressed in [`a201aeb` (this PR)](https://github.com/tempoxyz/tempo/pull/3840/commits/a201aebe0cc7c4e7959d1b32ce3ed73be570aba2)

crates/precompiles/src/storage/types/bytes_like.rs
```diff
@@ -201,22 +201,47 @@ where
...
+    storage: &mut S,
+    base_slot: U256,
+    ctx: LayoutCtx,
+) -> Result<()> {
+    let new_len = bytes.len();
+    let new_is_long = new_len > 31;
+    let mut data_slot: Option<U256> = None;
+
+    // (T5+) Cleanup stale tail if necessary.
+    if !ctx.skip_tail_cleanup() && StorageCtx.spec().is_t5() {
```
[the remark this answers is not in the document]
0xrusowsky: expected behavior

crates/precompiles/src/storage/types/vec.rs
```diff
@@ -61,27 +61,35 @@ where
     }
 
     fn store<S: StorageOps>(&self, storage: &mut S, len_slot: U256, ctx: LayoutCtx) -> Result<()> {
-        debug_assert_eq!(ctx, LayoutCtx::FULL, "Dynamic arrays cannot be packed");
+        debug_assert!(ctx.is_full(), "Dynamic arrays cannot be packed");
+        let data_start = calc_data_slot(len_slot);
+
+        // (T5+) Cleanup stale tail, if necessary.
+        if !ctx.skip_tail_cleanup() && StorageCtx.spec().is_t5() {
+            let (prev_len, new_len) = (load_checked_len(storage, len_slot)?, self.len());
```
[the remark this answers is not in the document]
0xrusowsky: expected behavior

PR #3845, opened 2026-05-07
fix(tip-1028): impl bugs and nits

0xrusowsky:
this PR fixes:
- re-execution in `mint` paths
- validate recovery address when setting receive policies
- block claims when token is paused
- add escrow to the list of protected addresses

additionally it performs some formatting and documentation nits

### DIFF
--- a/crates/precompiles/src/tip20/mod.rs
+++ b/crates/precompiles/src/tip20/mod.rs
@@ -400,9 +400,9 @@ impl TIP20Token {
     /// - `PolicyForbids` — TIP-403 policy rejects the mint recipient
     /// - `SupplyCapExceeded` — minting would push total supply above the cap
     pub fn mint(&mut self, msg_sender: Address, call: ITIP20::mintCall) -> Result<()> {
-        let (to, total_supply) = self.preflight_mint(msg_sender, call.to)?;
+        let (to, total_supply) = self.validate_mint(msg_sender, call.to)?;
 
-        if self.validate_inbound_or_escrow(
+        if self.validate_or_escrow_funds(
             Address::ZERO,
             &to,
             call.amount,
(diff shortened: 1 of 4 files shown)


crates/precompiles/src/tip20/mod.rs
```diff
@@ -998,6 +1004,29 @@ impl TIP20Token {
...
+
+        let total_supply = if self.storage.spec().is_t6() {
+            None
+        } else {
+            Some(self.total_supply()?)
+        };
+
+        self.validate_mint(&to)?;
+        Ok((to, total_supply))
+    }
```
0xKitsune: Can we roll this into `validate_mint`?

crates/precompiles/src/tip20/mod.rs
```diff
@@ -1068,9 +1097,11 @@ impl TIP20Token {
...
     }
 
-    /// Validates the TIP-1028 receive-policy check for the destination address. If the receive
-    /// policy prohibits the action, the funds are escrowed.
-    fn validate_or_escrow_funds(
+    /// Validates inbound TIP-1028 receive policy for `to`.
+    ///
+    /// Returns `true` when receive-policies block the inbound and this function escrows the funds.
+    /// Callers must ONLY move funds when `false` is returned.
+    fn validate_inbound_or_escrow(
```
0xKitsune: nit: any reason we need this changed? IMO `validate_or_escrow()` is short and descriptive.

crates/precompiles/src/tip20/mod.rs
```diff
@@ -1127,6 +1158,8 @@ impl TIP20Token {
         amount: U256,
         meter_spending_limit: bool,
     ) -> Result<()> {
+        self.check_not_paused()?;
+
```
0xKitsune, line 1152: Nice, good catch

0xKitsune (approved):

PR #3846, opened 2026-05-07
docs: document zero-init behaviour of packed struct stores

decofe:
ref: `SIGP-254`

addresses the informational audit finding from PR #2976

### DIFF
--- a/crates/precompiles-macros/src/storable.rs
+++ b/crates/precompiles-macros/src/storable.rs
@@ -490,11 +490,9 @@ fn gen_load_impl(fields: &[(&Ident, &Type)], packing: &Ident) -> TokenStream {
 ///
 /// # Zero-init behaviour (T4+)
 ///
-/// From T4 onward each packed slot group starts from `U256::ZERO` instead of a pre-`SLOAD`,
-/// so any byte not written by a declared packed field is zeroed on every store. If a struct's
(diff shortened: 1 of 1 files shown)


0xrusowsky (approved):

crates/precompiles-macros/src/storable.rs
```diff
@@ -487,6 +487,14 @@ fn gen_load_impl(fields: &[(&Ident, &Type)], packing: &Ident) -> TokenStream {
...
 /// For consecutive packable fields sharing a slot, accumulates changes in memory
 /// and writes once, avoiding redundant SLOAD + SSTORE pairs.
+///
+/// # Zero-init behaviour (T4+)
+///
+/// From T4 onward each packed slot group starts from `U256::ZERO` instead of a pre-`SLOAD`,
+/// so any byte not written by a declared packed field is zeroed on every store. If a struct's
+/// footprint later shrinks (e.g. a trailing field is removed), those formerly-occupied bytes
+/// will be cleared on the next write — benign, since they are unreachable via the layout
+/// system, but raw `SLOAD` consumers should not rely on stale padding surviving stores.
```
0xrusowsky:
```suggestion
/// Each packed slot group starts from `U256::ZERO` instead of a previous SLOAD, so any byte not
/// written by a declared packed field is zeroed on every store. If a struct later on removes a
/// trailing field, those formerly-occupied bytes will be cleared on the next write.
```

PR #3854, opened 2026-05-07
feat: set shared gas to 0 post-T4

klkvr:
Sets `shared_gas_limit` to 0 post-T4


0xKitsune (approved):
Nice, lgtm. A few spots in `subblocks.rs` would break at the T4 boundary if any validators are sending empty subblocks but with subblocks hardcoded to false it should be a non issue.

PR #3856, opened 2026-05-07
test(tip-1033): invariants

0xrusowsky:


grandizzy:
@0xrusowsky think should be good for now, I added 3 precompile-level unit tests that cover TIP-1033 paths the forge invariants can't reach (Foundry zeros tx fees so the real precompile path is unreachable, and vm.load can't observe another contract's transient storage)

PR #3859, opened 2026-05-07
docs (TIP -1054): Fees in non-USD TIP-20 tokens

malleshpai:
TIP-1054 lets users pay transaction fees on Tempo with non-USD TIP-20 tokens.

For a non-USD token to be used for fees the issuer must designate a price oracle for their token which will be used by the feeAMM. This version supports only direct FeeAMM pools from the user's non-USD token to the validator's USD fee token.

Gas accounting does not change. Transactions still use the existing signed gas fields, Tempo still prices gas in attodollars, and validators still receive fees in USD.


0xrusowsky (commented):
nit: vars are sometimes named `usd6` and sometimes `priceX18`.

personally i'd use scientific notation-like naming (`usdE6` and `priceE18`), so the decimal scale is explicit. imo `X` can be confused with uniV3 binary fixed-point notation, where values like `X96`/`X128` are scaled by `2**N`

0xrusowsky:
directionally looks good, and i think such design could work.

my main concerns with this design thought, are around oracle trust and data freshness... as specified, FX direct pools all pricing/oracle risk on LPs, without giving them any protocol-level guarantees.

a relatively simple mitigation would be to store quote update time/block and have FeeAMM stop quoting once the price has not been updated for more than `N` blocks. `N` could either be defined by the protocol or be configurable per non-USD token (by the issuer). This would at least provide clearer liveness/freshness guarantees.

a more complex (and opinionated) solution would be to add some kind of rolling average / volatility or max-change check, and stop quoting or reject updates when price changes exceed a threshold.

0xrusowsky:
the new proposed changes partially mitigate my previous concerns; the ceiling + freshness policy is a clean way to bound oracle risk while keeping the mechanism simple.

i think we are now in a way better spot than before, however, i would like to have a small dedicated subsection spelling out the trust assumptions LPs are making (what they're protected against, what they're not, etc). Feels like useful hygiene given the oracle/admin surface here.

malleshpai:
> nit: vars are sometimes named `usd6` and sometimes `priceX18`.
> 
> personally i'd use scientific notation-like naming (`usdE6` and `priceE18`), so the decimal scale is explicit. imo `X` can be confused with uniV3 binary fixed-point notation, where values like `X96`/`X128` are scaled by `2**N`

@0xrusowsky updated this.

tips/tip-1054.md
```diff
@@ -0,0 +1,408 @@
...
+    FeeManager->>User: refund unused FX token
+```
+
+<br>
+
+# Specification
+
+## TIP-20 Changes
+
+The fee oracle lives in the TIP-20 precompile. This TIP adds storage and entry points to the token rather than defining a separate oracle contract.
```
danrobinson: I was thinking we could have the fee oracle stored in its own precompile rather than on each token, so we don't need to do separate oracle pushes to every token. My expectation would be that most tokens for a given currency would pick some standard shared oracle.

samczsun: People might want to charge rent for this, like Chainlink perhaps

tips/tip-1054.md
```diff
@@ -0,0 +1,408 @@
...
+
+### Fee Oracle
+
+A non-USD TIP-20 token opts into letting users pay fees in that token by setting a `feeQuoteUpdater` and a fee quote policy. Only the token's authorized admin MAY set or clear them. Successful changes MUST emit `FeeQuoteUpdaterSet` or `FeeQuotePolicySet`, and changing either value MUST clear the stored quotes. `feeQuoteUpdater = address(0)` means the token is not configured for FX fees.
```
samczsun: Should this just be a role instead of a single address?

tips/tip-1054.md
```diff
@@ -0,0 +1,408 @@
...
+
+USD tokens MUST reject nonzero fee quote updaters and nonzero fee quote policies.
+
+### Fee Quotes
+
+A fee quote has two fields:
+- `quoteId` is an updater-supplied monotonically increasing identifier for the quote.
```
samczsun: Does it need to be? If so, we need to introduce some guardrails so you can't just brick a token by passing `quoteId = MaxInt`

tips/tip-1054.md
```diff
@@ -0,0 +1,408 @@
...
+The updater MAY call `updateLatestFeeQuote(quoteId, pathUsdPerTokenE18)` to push a new quote. The call MUST revert unless all of these conditions hold:
+- The caller is `feeQuoteUpdater`.
+- The token has a valid fee quote policy.
+- `quoteId` is greater than both `usableQuote.quoteId` and `latestQuote.quoteId`.
+- `pathUsdPerTokenE18` is nonzero.
+
+The first accepted quote is trusted after these checks. After a `latestQuote` exists, each later quote MUST satisfy the token's quote-change limit:
+
+```text
+abs(pathUsdPerTokenE18 - latestQuote.pathUsdPerTokenE18) * BPS_SCALE
```
samczsun: Shouldn't this be comparing with `usableQuote`? Otherwise I can just push a bunch of updates all in one block and blow past the max change bps

tips/tip-1054.md
```diff
@@ -0,0 +1,408 @@
...
+
+A non-USD TIP-20 token opts into letting users pay fees in that token by setting a `feeQuoteUpdater` and a fee quote policy. Only the token's authorized admin MAY set or clear them. Successful changes MUST emit `FeeQuoteUpdaterSet` or `FeeQuotePolicySet`, and changing either value MUST clear the stored quotes. `feeQuoteUpdater = address(0)` means the token is not configured for FX fees.
+
+USD tokens MUST reject nonzero fee quote updaters and nonzero fee quote policies.
+
+### Fee Quotes
+
+A fee quote has two fields:
+- `quoteId` is an updater-supplied monotonically increasing identifier for the quote.
+- `pathUsdPerTokenE18` is the value of one whole token in `pathUSD`, scaled by `QUOTE_SCALE`. For example, if one token is worth `2.50 pathUSD`, the updater pushes `pathUsdPerTokenE18 = 2.5e18`.
```
samczsun: maybe this should just be `usdPerTokenE18`

tips/tip-1054.md
```diff
@@ -0,0 +1,408 @@
...
+
+The fee quote policy has two fields:
+- `maxFeeQuoteAgeBlocks` bounds how old the usable quote may be.
+- `maxQuoteChangeBps` bounds how far a newly pushed quote may move from the latest accepted quote.
```
horsefacts: nitpick: naming is inconsistent here, drop the "fee," it's cleaner.

```suggestion
- `maxQuoteAgeBlocks` bounds how old the usable quote may be.
- `maxQuoteChangeBps` bounds how far a newly pushed quote may move from the latest accepted quote.
```

tips/tip-1054.md
malleshpai: @danrobinson Resolved by introducing the shared FeeOracle precompile and having tokens store only a feeOracleFeedId, allowing multiple tokens to reuse the same feed.

@samczsun the feeOraclefeed gets to specify what token(s) are authorized to use the feed, but the tip also clarifies that the payment is offchain (line 77)

tips/tip-1054.md
malleshpai: accepted, see below

tips/tip-1054.md
malleshpai: Fixed by comparing new quotes against usableQuote and explicitly preventing same-block update walking past maxQuoteChangeBps; see lines 116-125.

tips/tip-1054.md
```diff
@@ -0,0 +1,343 @@
...
+
+Before `updateCurrentFeeQuote` writes the new quote, the token MUST apply `rollIfEligible`. The call then writes the pushed quote as `currentQuote`, sets `currentQuoteBlock = block.number`, and emits `FeeQuoteUpdated`. If more than one update is pushed for the same token in block `N`, only the first call can roll a quote from an earlier block. Same-block updates only overwrite `currentQuote` and cannot affect fees in block `N`.
+
+Before any FeeAMM operation uses a token quote, FeeAMM MUST call `rollFeeQuote()` for that token and then read `previousQuote`. `rollFeeQuote()` MAY be restricted to FeeAMM. If it is exposed more broadly, it MUST perform only `rollIfEligible`.
+
+A previous quote is valid only if `previousQuoteBlock != 0`, `previousQuote.roundId != 0`, and `previousQuote.pathUsdPerTokenX18 != 0`. If the previous quote is invalid after `rollFeeQuote()`, the FX token is not fee-enabled for the current block and any operation that requires a valid quote MUST revert.
+
```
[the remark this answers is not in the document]
malleshpai: fixed across mulitple commits but basically we added all the things recommended.

tips/tip-1054.md
```diff
@@ -0,0 +1,343 @@
...
+### Transaction Fee Settlement
+
+For a transaction that resolves to a USD user fee token, existing fee-collection semantics are unchanged.
+
+For a transaction that resolves to an FX user fee token, pre-tx collection MUST happen before user code runs:
+
+1. FeeManager asks FeeAMM to call `rollFeeQuote()` for `userToken` and to read and return `userToken.previousFeeQuote()`.
+2. Fee collection reverts if the resolved FX token is not executable in the current block.
+3. FeeManager computes `maxUserTokenFee = fxTokenAmountForUsd6(maxFeeUsd6, quote.pathUsdPerTokenX18)`.
+4. FeeManager transfers `maxUserTokenFee` from the fee payer through the existing precharge path.
```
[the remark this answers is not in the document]
malleshpai: fixed.

tips/tip-1054.md
```diff
@@ -0,0 +1,343 @@
...
+### Mint and Burn
+
+For the first `mint(userToken, validatorToken, amountValidatorToken, to)` on an FX direct pool, FeeAMM MUST call `rollFeeQuote()` and MUST revert unless `userToken` has a valid previous quote for the current block. The caller deposits only `validatorToken`, `MIN_LIQUIDITY` remains permanently locked, and the bootstrap formula is unchanged. The quote is used only to confirm that the pool's FX token is currently fee-enabled; it is not used in the bootstrap formula.
+
+For later mints, FeeAMM MUST call `rollFeeQuote()` and use the previous quote for `userToken` in the current block. Let `U = reserve_user_token`, `V = reserve_validator_token`, `S = totalSupply`, `priceX18` be the previous quote, and `amountValidatorToken` be the caller's deposit. The minted liquidity MUST be:
+
+```text
+discountedUserReserveUsd6
+    = floor(usd6ValueFloor(U, priceX18) * FEE_AMM_N / FEE_AMM_SCALE)
+
```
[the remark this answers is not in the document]
malleshpai: Fixed in commit ff05baa02

tips/tip-1054.md
malleshpai: addressed in line 77: the feed updater may be a contract or multisig that implements its own roles/aggregation/access rules

tips/tip-1054.md
malleshpai: done.

tips/tip-1054.md
```diff
@@ -0,0 +1,477 @@
...
+### Feeds
+
+A feed is keyed by `feedId` and has a feed admin, a feed updater, a quote policy, token authorizations, and two quote tuples:
+
+```text
+(usableQuote, usableQuoteBlock)
+(latestQuote, latestQuoteBlock)
+```
+
+`latestQuote` is the newest accepted quote. `usableQuote` is the quote FeeAMM may use. Each quote block stores the Tempo block in which that quote was accepted.
```
dankrad: I think this should be modified: we should have a `latestQuote` and a `previousQuote`.

When a user calls to get a quote, if `latestQuoteBlock < currentBlock`, return latest, and otherwise return previous
Posting a new quote moves `latestQuote` to `previousQuote` and updates `latestQuote`

tips/tip-1054.md
```diff
@@ -0,0 +1,477 @@
...
+
+<br>
+
+## PriceOracle Precompile
+
+The PriceOracle precompile is deployed at `0xFEED000000000000000000000000000000000000` and stores shared token-price feeds. This TIP specifies only how fee settlement uses those feeds. Payments, charging, and off-protocol oracle operation are outside this TIP. A feed updater MAY be a contract or multisig that implements its own roles, aggregation, billing, or access rules.
+
+### Feeds
+
+A feed is keyed by `feedId` and has a feed admin, a feed updater, a quote policy, token authorizations, and two quote tuples:
```
dankrad, line 80: Should we sequentially assign feedIds instead or assign using a hash that includes the sender?
Otherwise, it can be frontrun

tips/tip-1054.md
```diff
@@ -0,0 +1,477 @@
...
+
+## Transaction Fee Token Selection
+
+`setValidatorToken(token)` is unchanged and MUST remain USD-only.
+
+### Selecting a Fee Token
+
+Tempo resolves a transaction's user fee token in the same order as before this TIP with the change that the token MAY be either a USD token or a fee-quoting FX token.
+
+`setUserToken(token)` MUST continue to accept every USD token accepted today. It MUST also accept a valid non-USD TIP-20 token when that token has a nonzero `priceOracleFeedId` and the selected feed authorizes the token. For an FX token, `setUserToken(token)` checks only that a feed is configured and authorized. It MUST NOT require the token to be unpaused or to have a valid usable quote, because those conditions are block-dependent and are checked when a transaction is charged.
```
dankrad: Do we need the condition that it has a feed implemented? We can't prevent users from setting unusable fee tokens (e.g. those having zero liquidity in feeAMMs) so arbitrarly saying "non-USD tokens must fulfill this additional criterion" does not add any value but feels less currency-neutral

dankrad:
I'm not convinced that we should have `PROTOCOL_MAX_QUOTE_CHANGE_BPS`. This trades trust in the oracle feed with exposure in sudden price movements. And while on most days price movements in FX may be tiny, there are big movements even in major currencies (vividly remember June 2024 2016 when GBP crashed 10% in minutes after the result of the Brexit vote became clear)

tips/tip-1054.md
malleshpai: Thanks, that's much cleaner... updated to use latestQuote / previousQuote. usableQuote now returns latestQuote if it was accepted before the current block; otherwise it returns previousQuote. Same-block updates remain ineligible for same-block fee use.

tips/tip-1054.md
malleshpai: Updated feed IDs to use the same creator+salt pattern as TIP-20.

tips/tip-1054.md
malleshpai: Agreed. setUserToken now accepts any valid non-USD TIP-20 token

malleshpai:
> I'm not convinced that we should have `PROTOCOL_MAX_QUOTE_CHANGE_BPS`. This trades trust in the oracle feed with exposure in sudden price movements. And while on most days price movements in FX may be tiny, there are big movements even in major currencies (vividly remember June 2024 2016 when GBP crashed 10% in minutes after the result of the Brexit vote became clear)

my take is that oracle mistakes/ corruption are much more likely than a jump move (i looked back in the past year and rarely saw USD move more than 50bps in a day against other G20s, will get you the exact numbers), so i'd rather have guards against the more likely event. brexit specifics are more nuanced--- the results came in overnight...

0xrusowsky (approved):

dankrad:
> i looked back in the past year and rarely saw USD move more than 50bps in a day

In a day? According to claude the USDGBP daily volatility is 0.6% so that's an average day!
Anyway, a reasonable per block BPS should cover it because a block is much shorter than a day

PR #3865, opened 2026-05-08
perf(builder): boilerplate of prewarming for best transactions

shekhirin:
Adds an opt-in `--builder.enable-prewarming` flag that enables prewarming thread that consumes `best_txs` at its own speed, and buffers transactions for payload building to consume. Prewarming accounts for invalidations for buffered transactions.

This PR doesn't do any prewarming yet, next step would be to start executing transactions on top of latest state and populating shared cache.

The flag is disabled by default, so existing payload building behavior is unchanged.


crates/payload/builder/src/prewarming.rs
```diff
@@ -0,0 +1,192 @@
...
+                // TODO: prewarm state by executing the transaction `tx` on top of latest state
+            } else {
+                // No more best transactions for now. We do not break the loop,
+                // because there may be more transactions later.
+                if ctx.events_tx.send(BestTransactionsEvent::Empty).is_err() {
+                    break;
+                }
+            }
+
+            while let Ok(command) = ctx.commands_rx.try_recv() {
```
klkvr, line 65: this would effectively hotloop until entire iterator is drained right? i guess maybe it's fine because we're just draining Arcs but we could probably change this so that the builder thread directly requests new transactions via something like `BestTransactionsCommand::Advance(oneshot::Receiver)`

in practice most of the time prewarming should be ahead so the buffer of transactions we've drained for prewarming workers should be enough for execution to advance, so it shouldn't be necessary to drain the entire iter

klkvr (approved):

PR #3870, opened 2026-05-08
feat(consensus): prunable blocks

SuperFluffy:
Introduces a hybrid storage model, which writes finalized blocks to a prunable archive that sits on top of the execution layer. The prunable archive acts as a cache that gets pruned as the execution layer's finalized watermark rises. The consensus layer cannot write through to the execution layer (that's the domain of the new-payload + forkchoice-updated APIs), and so it maintains a layer on top of it.

To ensure that nodes can be rolled back to previous versions, this patch does not yet delete the (legacy) immutable archive, but write 2 copies of blocks - one to the new hybrid store, one to the legacy archive. Nodes can opt out of this behavior by setting `--consensus.no-legacy-archive`. Starting a node without a legacy archive and then restarting it with one is not supported and users are advised against doing this.


crates/commonware-node/src/consensus/engine.rs
```diff
@@ -174,6 +186,27 @@ where
...
+        let last_finalized_height =
+            marshal_stored_height.max(Height::new(reth_finalized_height));
+        if last_finalized_height > marshal_stored_height {
+            info!(
+                marshal_stored = %marshal_stored_height,
+                reth_finalized = reth_finalized_height,
+                "advancing marshal sync floor to reth's finalized block"
+            );
+            marshal_mailbox.set_floor(last_finalized_height).await;
+        }
```
joshieDo: dupped, helper?

SuperFluffy: Nice: 6381de52bd09866e409b75397eb43a4fa3914466

crates/commonware-node/src/storage/hybrid/mod.rs
```diff
@@ -0,0 +1,497 @@
...
+    ///   happens; the cache temporarily grows past `retention_blocks`.
+    /// - Eviction tracks reth's progress monotonically. Once reth
+    ///   finalizes height `H`, the cache may drop everything below
+    ///   `H - retention_blocks + 1` on the next put.
+    ///
+    /// `prune(min)` keeps `min` and above, and the archive rounds `min`
+    /// *down* to a section boundary, so the actual retained window can
+    /// exceed `retention_blocks` by up to `items_per_section − 1` items.
+    /// See the module docs ("Section-rounding") for the full story.
+    async fn evict_below_reth_finalized_floor(&mut self) -> Result<(), archive::Error> {
```
hamdiallam: should we remove `reth` from the name since we've abstracted away the provider?

hamdiallam: `evict_below_finalized_floor`?

crates/commonware-node/src/storage/hybrid/mod.rs
```diff
@@ -0,0 +1,497 @@
...
+            );
+        }
+        Ok(())
+    }
+
+    async fn sync(&mut self) -> Result<(), Self::Error> {
+        if let Some(legacy) = self.legacy.as_mut() {
+            archive::Archive::sync(legacy).await?;
+        }
+        archive::Archive::sync(&mut self.prunable).await?;
```
hamdiallam, line 365: why this notation over self.prunable::sync? Applies to everywhere in this file

Guessing this is amp?

crates/commonware-node/src/alias.rs
```diff
@@ -1,28 +1,202 @@
...
+    {
+        let finalizations_by_height = storage::init_finalizations_archive(
+            &context,
+            &config.partition_prefix,
+            page_cache.clone(),
+        )
+        .await
+        .wrap_err("failed to initialize finalizations by height archive")?;
+
+        let finalized_blocks = storage::init_hybrid_finalized_blocks(
```
hamdiallam: do we need `hybrid` in the name here, it's fully abstracted away internally. I think we should keep `init_finalized_blocks`

crates/commonware-node/src/storage/mod.rs
```diff
@@ -0,0 +1,227 @@
...
+        retention_blocks > 0,
+        "finalized blocks retention must be greater than zero",
+    );
+
+    let mut prunable =
+        init_prunable_finalized_blocks_archive(context, partition_prefix, page_cache.clone())
+            .await
+            .wrap_err("failed to initialize prunable finalized blocks archive")?;
+
+    let legacy = if dual_write_to_legacy {
```
hamdiallam: Why have `dual_write_to_legacy` as something configurable? Instead always dual writing then in an upcoming release, delete the legacy archive completely.

Feels awkward since this means someone can dual write -> stop dual write -> dual write again? The legacy archive will then miss blocks in that intermediary phase. Perhaps it's fine since it's written in prunable?

Feels better for it to be an explicit transitionary phase rather than parameterized

hamdiallam (approved):
Approach LGTM though!

crates/commonware-node/src/storage/hybrid/mod.rs
SuperFluffy: Not amp actually - I want down the rabbit hole of removing this, but in some instances it did not know which `sync` to delegate to because there were multiple matching definitions in the namespace. Figured that I actually like how amp made it explicit which definition is called even in the cases where it was clear.

crates/commonware-node/src/storage/hybrid/mod.rs
SuperFluffy: I have mostly replaced `reth` by `execution` or `EL` in docs: 328a173b6

SuperFluffy: I have mostly replaced `reth` by `execution` or `EL` in docs: 328a173b6.

crates/commonware-node/src/alias.rs
SuperFluffy: Good call; done in 05620bde

crates/commonware-node/src/storage/mod.rs
SuperFluffy: That's fair; removed in c10b3c6a

PR #3875, opened 2026-05-08
feat(rpc): add expiring nonce status RPC

Slokh:
## Summary

Adds `tempo_getExpiringNonceStatus`, a Tempo RPC for classifying a signed expiring-nonce transaction as `pending`, `included`, `expired`, or `unavailable` using Tempo's canonical transaction parsing and finalized nonce-manager state.

Also adds `tempo_alloy::provider::TempoProviderExt::get_expiring_nonce_status` so Alloy clients can call the RPC without hand-rolling the raw JSON-RPC request.

## Motivation

Expiring-nonce transactions are intentionally replaceable only after their validity window has passed, but `eth_getTransactionReceipt(txHash) == null` is not enough information for a client to decide that replacement is safe. A missing receipt can mean the transaction has not executed yet, that the client is looking at the wrong hash variant, or that historical replay state is no longer available.

This RPC gives integrators a node-level answer based on Tempo semantics instead of requiring every client to derive replay keys and read nonce-manager storage by hand.

## RPC Interface

Request:

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "tempo_getExpiringNonceStatus",
  "params": [
    {
      "signedTransaction": "0x76..."
    }
  ]
}
```

Response:

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "status": "pending"
  }
}
```

Status values:

| Status | Meaning |
| --- | --- |
| `pending` | No finalized canonical block at or after `validBefore` exists yet. Replacement is not known to be safe. |
| `included` | Finalized canonical pre-expiry replay state proves the expiring nonce was seen. Replacement is not safe. |
| `expired` | Finalized canonical pre-expiry replay state proves the expiring nonce was not seen once expiry has been crossed. This is the only replace-safe status. |
| `unavailable` | The node cannot access the historical state needed to classify the transaction. Replacement is not known to be safe. |

## Implementation

- Decodes and validates the signed transaction using Tempo transaction semantics.
- Rejects malformed, non-Tempo, non-expiring-nonce, and missing-`validBefore` inputs as invalid params.
- Recovers the signer and computes the expiring-nonce replay hash internally.
- Uses a binary search over retained finalized headers to find the first finalized block with `timestamp >= validBefore`.
- Reads `NonceManager.expiringNonceSeen[expiringNonceHash]` at the latest retained finalized block before `validBefore`, rather than latest/post-expiry state, so bounded replay-state cleanup cannot make an included transaction look expired.
- Classifies zero storage as `expired`, exact `validBefore` storage as `included`, and unexpected nonzero storage as `unavailable`.

## Testing

Tests cover:

- `pending`, `included`, and `expired` happy paths.
- The `timestamp == validBefore` expiry boundary.
- Invalid input handling for malformed, legacy, non-expiring-nonce, and missing-`validBefore` transactions.
- Storage classification rules for zero, exact-`validBefore`, and unexpected nonzero storage.
- Pre-expiry proof-block selection when historical state is and is not retained.
- Alloy provider helper success and transport-error propagation.

Commands run:

- `cargo fmt`
- `cargo check -p tempo-node -p tempo-alloy`
- `cargo test -p tempo-alloy get_expiring_nonce_status`
- `cargo test -p tempo-node rpc::expiring_nonce_status::tests --lib`
- `cargo test -p tempo-node --test it expiring_nonce_status -- --nocapture`
- `cargo clippy -p tempo-node -p tempo-alloy --tests -- -D warnings`
- `git diff --check`


crates/node/src/rpc/expiring_nonce_status.rs
```diff
@@ -0,0 +1,266 @@
...
+    N: FullNodeTypes<Types = TempoNode>,
+    N::Provider: BlockIdReader
+        + BlockNumReader
+        + ChainSpecProvider<ChainSpec = TempoChainSpec>
+        + HeaderProvider
+        + Send
+        + Sync
+        + 'static,
+{
+    async fn get_expiring_nonce_status(
```
klkvr, line 85: could we just check that `receipt(tx_hash).is_none() && tip.timestamp > expiry` to know if the status is `expired`?

if we want to only report inclusion in finalized blocks here we could use `finalized` as `tip`, though i'm not entirely sure that our RPC nodes handle `finalized` tag properly

also seems that this is not specific to expiring nonce transactions and we could support this for any expiring tx?

Slokh: > could we just check that receipt(tx_hash).is_none() && tip.timestamp > expiry to know if the status is expired?

I think technically this wouldn’t be 100% correct because the replay key for expiring nonces isn’t the tx hash. For example, if fee payer data changes, the tx hash changes but the expiring nonce hash stays the same, so `receipt(tx_hash) == null` could miss that the replay key was already consumed.

> if we want to only report inclusion in finalized blocks here we could use finalized as tip, though i'm not entirely sure that our RPC nodes handle finalized tag properly

Yeah, we want to use finalized. Isn't that what I'm doing with `provider.finalized_block_num_hash`. Opted to not use the tag, and if we're not confident in it, then should avoid it I think.

> also seems that this is not specific to expiring nonce transactions and we could support this for any expiring tx?

That's fair, didn't even think about that. I'd be for moving this to be something even more generic like `tempo_getTransactionReplayStatus` that will cover all cases.

klkvr: > if fee payer data changes, the tx hash changes but the expiring nonce hash stays the same

yeah this malleability is a known issue and it's also not specific to expiring nonce transactions. any transaction that specifies a fee payer signature is malleable and tracking it by hash is potentially wrong. should we have a more general RPC endpoint for this given that any tempo transaction is potentially affected? or have a more low-level fix to get rid of this footgun?

PR #3877, opened 2026-05-08
fix(rpc): improve fee token errors

brendanjryan:
## Summary

Adds typed node errors for invalid Tempo fee tokens and maps them to transaction-rejected JSON-RPC responses. Mirrors the client-side validation shape from wevm/viem#4590.


crates/revm/src/common.rs
```diff
@@ -208,6 +208,28 @@ pub trait TempoStateAccess<M = ()> {
...
+        self.with_read_only_storage_ctx(spec, || {
+            // SAFETY: caller must ensure prefix is already checked
+            let token = TIP20Token::from_address_unchecked(fee_token);
+            let len = token.currency.len()?;
+            if len > 31 {
+                return Ok(format!("<{len} bytes>"));
+            }
+            token.currency.read()
+        })
+    }
```
0xrusowsky: this is inefficient, instead let's update the existing `fn is_tip20_usd()` to also return the currency

0xrusowsky: addressed in [`db937fa` (this PR)](https://github.com/tempoxyz/tempo/pull/3877/commits/db937fa3dc0398325d5ab2765bb55877be520ddc)

0xrusowsky (approved):
let's wait for @klkvr's review too as this touches handler code, which is always sensible

klkvr (approved):

PR #3881, opened 2026-05-10
fix(tempoup): require gpg verification by default

brendanjryan:
## Summary

Streamlines binary install process on new hosts which may not have all deps via flags + fallbacks

Requires GPG signature verification by default for signed releases while keeping checksum verification mandatory. 

Users can explicitly opt out with `--unsafe-skip-verify`, including through the bootstrap installer via `curl -fsSL https://tempo.xyz/install | bash -s -- --unsafe-skip-verify`.

Keeps GitHub CLI / SLSA attestation verification as best-effort: it runs when `gh` is installed and authenticated, but missing or unauthenticated `gh` no longer blocks installation.

### DIFF
--- a/tempoup/install
+++ b/tempoup/install
@@ -87,7 +87,7 @@ ENVEOF
     info "Running tempoup to install tempo..."
     echo ""
 
-    TEMPO_BIN_DIR="$BIN_DIR" "$BIN_DIR/tempoup"
+    TEMPO_BIN_DIR="$BIN_DIR" "$BIN_DIR/tempoup" "$@"
 
     echo ""
     local user_shell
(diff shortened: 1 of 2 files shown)


0xalpharush:
@grandizzy Maybe we should RIIR and take a dep. on sigstore so we can not require external deps

grandizzy:
> @grandizzy Maybe we should RIIR and take a dep. on sigstore so we can not require external deps

👍 foundry going that direction already https://github.com/foundry-rs/foundryup

grandizzy (approved):
lgtm!

brendanjryan:
ty!

PR #3886, opened 2026-05-11
ci: add Foundry resolver smoke test

emmajam:
## Summary
- add a lightweight Foundry resolver smoke job for dependency-sensitive PR changes
- include the smoke job in the existing specs success aggregate
- make foundry-patch retry cargo metadata failures with targeted cargo update -p fixes for lockfile constraint conflicts

## Verification
- bash -n scripts/foundry-patch.sh
- ruby -e 'require "yaml"; YAML.load_file(".github/workflows/specs.yml"); puts "specs.yml parsed"'\n- git diff --check\n- local smoke run against current foundry-rs/foundry@master passed

### DIFF
--- a/.github/workflows/specs.yml
+++ b/.github/workflows/specs.yml
@@ -6,10 +6,15 @@ on:
   push:
     branches: [main]
     paths:
+      - "Cargo.lock"
+      - "Cargo.toml"
+      - "crates/**/Cargo.toml"
       - "tips/verify/**"
       - "crates/contracts/**"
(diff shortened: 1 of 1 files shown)


scripts/foundry-patch.sh
0xrusowsky, line 1: i don't think we need to add changes to the script to add a smoke test?

emmajam: agree

fgimenez (approved):

PR #3903, opened 2026-05-11
fix: reject withdrawals

klkvr:
Rejects all blocks with non-empty withdrawals list.

We enforce it in executor to make sure that it never is tricked into actually commiting withdrawals even if block was not pre-validated beforehand

### DIFF
--- a/crates/evm/src/block.rs
+++ b/crates/evm/src/block.rs
@@ -454,7 +454,7 @@ where
             .as_ref()
             .is_some_and(|withdrawals| !withdrawals.is_empty())
         {
-            return Err(BlockExecutionError::msg("withdrawals are not permitted"));
+            return Err(BlockValidationError::msg("withdrawals are not permitted").into());
         }
 
         self.inner.apply_pre_execution_changes()?;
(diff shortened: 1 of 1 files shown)


crates/evm/src/block.rs
```diff
@@ -447,6 +447,16 @@ where
...
 
     fn apply_pre_execution_changes(&mut self) -> Result<(), alloy_evm::block::BlockExecutionError> {
+        if self
+            .inner
+            .ctx
+            .withdrawals
+            .as_ref()
+            .is_some_and(|withdrawals| !withdrawals.is_empty())
+        {
+            return Err(BlockExecutionError::msg("withdrawals must be empty"));
```
SuperFluffy: Maybe "withdrawals are not permitted"? "withdrawals must be empty" reads a bit roundabout.

Also, should we call these "eth withdrawals" or something like that so that it's not confused with normal transfers?

SuperFluffy (approved):

crates/evm/src/block.rs
klkvr:
```suggestion
            return Err(BlockExecutionError::msg("withdrawals are not permitted"));
```

PR #3904, opened 2026-05-11
test(tip20): demonstrate burn_blocked from ESCROW succeeds pre-T6

0xalpharush:
Stacked on top of #3901.

Extend the protected-address test with a T5 case that flips the transfer policy to REJECT_ALL and burns from `ESCROW_ADDRESS`, asserting the balance is reduced. Makes the T5↔T6 divergence in `PROTECTED` visible instead of relying on `assert_ne` against an unrelated error.

Prompted by: alpharush


0xrusowsky (approved):
ty

PR #3906, opened 2026-05-11
chore(consensus): report if reth did not return a payload ID

hamdiallam:
Qe support threading back errors in the response channel so we should populate it if we have no payload id on a build request rather than rely on dropping the channel

Context: When syncing, reth does not return a payload id

### DIFF
--- a/crates/commonware-node/src/executor/actor.rs
+++ b/crates/commonware-node/src/executor/actor.rs
@@ -517,11 +517,31 @@ where
                 let _ = response.send(Ok(()));
             }
             JustCanonicalizeOrAlsoBuild::AlsoBuild { response, .. } => {
-                if let Some(payload_id) = fcu_response.payload_id {
-                    let _ = response.send(Ok(payload_id));
+                let result = if fcu_response.is_syncing() {
+                    let canonical_block_number = self
+                        .execution_node
+                        .provider()
+                        .canonical_in_memory_state()
+                        .get_canonical_block_number();
+
+                    // Allow a syncing if only behind by 1 block
+                    let min_allowed = height.previous().unwrap_or_default().get();
+                    if canonical_block_number < min_allowed {
+                        Some(Err(eyre::eyre!(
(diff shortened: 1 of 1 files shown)


crates/commonware-node/src/executor/actor.rs
```diff
@@ -473,6 +473,22 @@ where
             return;
         }
 
+        let canonical_block_number = self
```
hamdiallam: @SuperFluffy decided against triggering the FCU then cancelling after. Felt that was more complex than just erroring early

crates/commonware-node/src/executor/actor.rs
```diff
@@ -473,6 +473,22 @@ where
...
+
+        // Allow syncing the parent block. Anything deeper will likely timeout
+        if canonical_block_number < height.previous().unwrap_or_default().get()
+            && let JustCanonicalizeOrAlsoBuild::AlsoBuild { response, .. } = maybe_build
+        {
+            let _ = response.send(Err(eyre::eyre!(
+                "build request on height {height} too far ahead of canonical height {canonical_block_number}"
+            )));
+            return;
+        }
```
SuperFluffy: This leads to not sending the FCU at all. What if we don't send the payload attributes? That's what triggers the payload build after all.

Send (and await?) the FCU but without the attributes; then, return an error on the payload ID response channel.

hamdiallam: Yeah makes sense. If we clear the attributes we can just error when `payload_id: None`

hamdiallam:
Looking in to an e2e test

hamdiallam:
@SuperFluffy since we're explicitly returning an error now, a local test with a stubbed mailbox isn't super useful.

Setting up something e2e is also racy

SuperFluffy:
Looks like we already inadvertently covered this use-case: if reth cannot propose a block because the parent is not known / not connected to the (local) canonical chain, it won't propose a block.

SuperFluffy (approved):

PR #3917, opened 2026-05-12
fix: refresh stale Tempo sources in foundry patch

legion2002:
The Foundry precompile CI patches Foundry to use the checked-out Tempo crates, then fails if any Tempo git sources remain in Foundry's lockfile. Cargo can still keep stale locked Tempo git packages after the path patches are applied, so the guard trips even though the local patches are present.

This updates the patch script to rewrite Foundry's patch entries portably and then run a targeted `cargo update -p ...` only for Tempo packages that still resolve from `git+https://github.com/tempoxyz/tempo`. That keeps the fix scoped to the lockfile entries that violate the CI invariant instead of doing a broad dependency update or weakening the guard.


legion2002:
closed in favour of #3919

PR #3929, opened 2026-05-12
fix(consensus): fall back to finalized tip on backfill gaps

joshieDo:
Handles missing blocks during startup backfill by abandoning the assumed contiguous replay range and sending FCU for the latest CL-finalized block from marshal.

Validated with `cargo +nightly fmt` and `cargo check -p tempo-commonware-node`.

### DIFF
--- a/crates/commonware-node/src/executor/actor.rs
+++ b/crates/commonware-node/src/executor/actor.rs
@@ -279,8 +279,13 @@ where
                             warn_span!("backfill_on_start", %height)
                             .in_scope(|| warn!(
                                 "marshal actor did not have block even though \
-                                it must have finalized it previously",
+                                it must have finalized it previously; \
+                                falling back to EL sync",
                             ));
+                            self.finalized_heights_to_backfill = RangeInclusive::new(1, 0);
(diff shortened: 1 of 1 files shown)


crates/commonware-node/src/executor/actor.rs
```diff
@@ -279,8 +279,13 @@ where
                             warn_span!("backfill_on_start", %height)
                             .in_scope(|| warn!(
                                 "marshal actor did not have block even though \
```
hamdiallam, line 281: Do we observe this? Marshal storage should be durable

crates/commonware-node/src/executor/actor.rs
```diff
@@ -279,8 +279,13 @@ where
...
-                                it must have finalized it previously",
+                                it must have finalized it previously; \
+                                falling back to EL sync",
                             ));
+                            self.finalized_heights_to_backfill = RangeInclusive::new(1, 0);
+                            if let Some(block) = self.marshal.get_block(self.last_consensus_finalized_height).await {
```
hamdiallam, line 286: We are requesting the block from marshal anyways. So the original backfill loop will still end up FCU'ing the latest block that marshal has. Not sure what this is doing differently

PR #3936, opened 2026-05-12
ci: full-crate Codecov coverage with PR delta comments

decofe:
Adds Codecov integration for full-crate coverage reporting with PR delta comments.

## Changes

- **Always instrument coverage** on PRs and main pushes — no longer gated on precompiles-only changes
- **Generate lcov for all crates** (not just `precompiles`/`contracts`) and upload to Codecov from both `cargo test` partitions and `forge test`
- **`codecov.yml` config** — PR comment shows coverage diff, patch coverage, per-flag breakdown; project/patch status checks are informational (won't block merge)
- **Existing precompiles HTML report** is unchanged — still triggers on precompiles changes for the detailed drill-down

## Setup required

Add `CODECOV_TOKEN` as a repository secret (get it from [codecov.io](https://app.codecov.io) after enabling the repo).

## How it works

Each CI run uploads lcov data with flags (`cargo-test-1`, `cargo-test-2`, `forge-test`). Codecov merges them, compares against the base branch baseline, and posts a PR comment showing:
- Overall project coverage change
- Patch coverage (lines added in this PR)
- Per-file breakdown of changed coverage

Flags use `carryforward: true` so partial runs (e.g. forge skipped) don't zero out prior data.

### DIFF
--- a/.github/workflows/specs.yml
+++ b/.github/workflows/specs.yml
@@ -182,22 +182,40 @@ jobs:
           workspaces: foundry
 
       - name: Install llvm-tools
+        if: github.event_name != 'merge_group'
         run: rustup component add llvm-tools-preview
 
       - name: Patch foundry to use checked-out tempo crates
         run: tempo/scripts/foundry-patch.sh "$GITHUB_WORKSPACE/tempo" "$GITHUB_WORKSPACE/foundry"
 
+      # Build and run WITH coverage (pull_request and main push)
       - name: Build foundry forge with coverage instrumentation
+        if: github.event_name != 'merge_group'
         working-directory: foundry
         run: RUSTFLAGS="-C instrument-coverage" cargo build -p forge --profile release --no-default-features
 
       - name: Run Forge tests with Rust precompiles (with coverage)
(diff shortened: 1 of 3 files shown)


.github/workflows/specs.yml
```diff
@@ -182,45 +182,48 @@ jobs:
...
+          chmod +x foundry/target/release/forge
+          "$LLVM_BIN/llvm-cov" export \
+            --format=lcov \
+            --instr-profile=coverage/forge-test.profdata \
+            --object=foundry/target/release/forge \
+            --ignore-filename-regex='/rustc/' \
+            --ignore-filename-regex='\.cargo/' \
+            --ignore-filename-regex='\.rustup/' \
+            --ignore-filename-regex='foundry/' \
+            > coverage/forge-lcov.info 2>/dev/null || true
```
grandizzy, line 234: should we fail the CI if export fails?

grandizzy (commented):
specs and test are also executed on `merge_group`, no benefits to generate coverage for that step so maybe let's gate these?

PR #3955, opened 2026-05-13
feat(p2p-proxy): expose prometheus metrics

kuyziss:
## Summary
- add `--metrics <addr>` to `tempo p2p-proxy` using the same socket parser as reth metrics args
- start a Prometheus scrape endpoint when the flag is set
- expose p2p proxy request and peer gauges/counters

## Validation
- cargo check -p tempo --bin tempo
- cargo build -p tempo --bin tempo
- cargo fmt --check
- ./target/debug/tempo p2p-proxy --help shows `--metrics <METRICS>`


kuyziss:
Closing this since p2p-proxy does not currently expose metrics and we do not want to add binary-side metrics support as part of this change.

PR #3965, opened 2026-05-13
chore: tempo.nu fixes & --epoch-length support

hamdiallam:
1. use a local reth.rpc file to avoid collisions
2. metrics ports should listen on 0.0.0.0. Obs stack will not work with the 127.0.0.X for the validators
3. Update follower to use the next available validator port mapping to work with the obs stack
4. Supply --epoch-length as a parameter. Default is too large. The localnet will just stay in epoch 0
5. Support --certify for the follow stack

### DIFF
--- a/tempo.nu
+++ b/tempo.nu
@@ -1517,6 +1517,7 @@ def "main follower" [
     } | math max) + 1)
 
     let consensus_port = ($follower_index * 100) + 8000
+    let consensus_ip = $"127.0.0.($follower_index + 1)"
     let http_port = 8545 + $follower_index
     let reth_metrics_port = 9001 + $follower_index
     let el_p2p_port = $consensus_port + 1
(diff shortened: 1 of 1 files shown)


shekhirin (approved):

tempo.nu
```diff
@@ -1521,25 +1506,40 @@ def "main follower" [
...
+    let consensus_metrics_port = $consensus_port + 2
+    let authrpc_port = $consensus_port + 3
+    let discv5_port = $consensus_port + 4
     let log_dir = $"($LOGS_DIR)/follower"
     mkdir $log_dir
 
     let args = (build-base-args $genesis_path $follower_dir $log_dir "0.0.0.0" $http_port $reth_metrics_port)
         | append [
             "--follow" $"ws://127.0.0.1:8545"
+            "--consensus.listen-address" $"127.0.0.1:($consensus_port)"
```
hamdiallam: Need to update this up address to follower the validator convention as well

PR #3967, opened 2026-05-14
fix(rpc): map Tempo-specific tx errors to -32003 instead of -32603

decofe:
## Problem

Tempo-specific transaction validation errors (e.g. `KeyAuthorizationChainIdMismatch: expected 42431, got 4217`) are returned as `-32603 Internal error` to RPC callers. These are client-side validation errors, not internal node failures.

### Root cause

reth's generic `From<EVMError<_, TxError>>` routes all custom `TxError` variants (where `as_invalid_tx_err()` returns `None`) into `EvmCustom(msg)` → `internal_rpc_err` → `-32603`.

## Fix

Add a `TempoTransactionRejected` variant to `TempoEthApiError` that intercepts Tempo-specific tx errors in the `From<EVMError>` impl and maps them to `-32003` (transaction rejected) with the original error message preserved.

**Before:** `{"code": -32603, "message": "Internal error"}`
**After:** `{"code": -32003, "message": "KeyAuthorization chain_id mismatch: expected 42431, got 4217"}`

### DIFF
--- a/crates/node/src/rpc/error.rs
+++ b/crates/node/src/rpc/error.rs
@@ -11,16 +11,36 @@ use reth_rpc_eth_types::{
 };
 use tempo_evm::TempoHaltReason;
 
+/// JSON-RPC error code for invalid input (EIP-1474 / geth convention).
+///
+/// Used for Tempo-specific transaction validation errors where the submitted
+/// input is fundamentally invalid (wrong chain_id, bad signature, etc.).
+const INVALID_INPUT: i32 = -32000;
+
 #[derive(Debug, thiserror::Error)]
 pub enum TempoEthApiError {
     #[error(transparent)]
     EthApiError(EthApiError),
+
+    /// Tempo-specific invalid transaction error.
+    ///
+    /// Mapped to `-32000` (invalid input) with structured `data` instead of the generic
+    /// `-32603` (internal error) that reth's catch-all produces for custom `InvalidTxError`
+    /// variants.
+    #[error("{message}")]
+    TempoInvalidInput {
+        message: String,
+        data: serde_json::Value,
+    },
 }
 
 impl From<TempoEthApiError> for jsonrpsee::types::error::ErrorObject<'static> {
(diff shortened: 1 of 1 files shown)


crates/node/src/rpc/error.rs
```diff
@@ -77,3 +109,72 @@ impl FromRevert for TempoEthApiError {
         }
     }
 }
+
+/// Build structured JSON `data` for a Tempo-specific transaction error.
+///
+/// Parses well-known error messages to extract structured fields (e.g. expected/got
+/// chain IDs). Falls back to just the error identifier for unrecognised messages.
+fn tempo_tx_error_data(message: &str) -> serde_json::Value {
```
mattsse, line 117: why do we need this?

crates/node/src/rpc/error.rs
```diff
@@ -77,3 +109,72 @@ impl FromRevert for TempoEthApiError {
...
+
+    #[test]
+    fn test_error_object_code() {
+        let err: jsonrpsee::types::error::ErrorObject<'static> =
+            TempoEthApiError::TempoInvalidInput {
+                message: "KeyAuthorization chain_id mismatch: expected 42431, got 4217"
+                    .to_string(),
+                data: tempo_tx_error_data(
+                    "KeyAuthorization chain_id mismatch: expected 42431, got 4217",
+                ),
```
mattsse, line 174: do we want this in both message and data?

crates/node/src/rpc/error.rs
deodad: I don't have strong opinions on how exactly these errors are modeled (code, message, data) other than they should not be internal errors when they are known validation errors so it's easier for downstream consumers to figure out

I don't have the context to make a judgement call—@mattsse would you take this over and do as you see fit?

crates/node/src/rpc/error.rs
deodad: in favor of simplifying to just message since that should work broadly

PR #3991, opened 2026-05-15
fix(tip-1033): surface tx `validator_fee` info to builder

0xrusowsky:
this PR fixes payload fee scoring for the feeAMM routes introduced by TIP-1033.

it uses the validator-credited amount reported during post-transaction fee collection so block value reflects actual proposer revenue after one-hop or two-hop AMM haircuts, instead of estimating it in the payload builder.

### DIFF
--- a/crates/evm/src/block.rs
+++ b/crates/evm/src/block.rs
@@ -99,6 +99,11 @@ pub struct TempoTxResult {
     tx: Option<TempoTxEnvelope>,
     /// Block gas consumed by this transaction. The block `gas_used` field will be incremented by this value.
     block_gas_used: u64,
+    /// Validator-credited fee (in the validator's fee token) reported by `collectFeePostTx`.
+    ///
+    /// Used by the payload builder to score blocks by actual proposer revenue. The value is the
+    /// post-feeAMM amount, regardless of route shape — absorbs any number of pool haircuts.
+    validator_fee: U256,
 }
 
 impl TempoTxResult {
@@ -110,6 +115,11 @@ impl TempoTxResult {
     /// Returns the state gas consumed by this transaction.
     pub fn state_gas_used(&self) -> u64 {
         self.inner.result.result.gas().state_gas_spent()
+    }
+
+    /// Returns the validator-credited fee amount (post-feeAMM haircut) for this transaction.
+    pub fn validator_fee(&self) -> U256 {
+        self.validator_fee
     }
 }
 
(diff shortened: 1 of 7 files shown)


crates/payload/builder/src/lib.rs
```diff
@@ -541,27 +535,12 @@ where
...
-                                .expect(
-                                    "execution succeeded, so compute_amount_out should not fail",
-                                );
-                        }
-                    } else {
-                        warn!("no resolved fee token for a pool transaction")
+                    // Score payload value by the validator-credited fee amount that the
+                    // FeeManager precompile actually wrote during this transaction.
+                    if pool_tx.transaction.resolved_fee_token().is_none() {
+                        warn!("no resolved fee token for a pool transaction");
```
klkvr: should be fine to remove this completely now

klkvr (approved):

crates/payload/builder/src/lib.rs
0xrusowsky: addressed in [`8f1662c` (this PR)](https://github.com/tempoxyz/tempo/pull/3991/commits/8f1662c564a5512ac7792bf0d441070a5e64b831)

crates/node/tests/it/block_building.rs
```diff
@@ -600,14 +615,35 @@ async fn test_payload_fees_account_for_amm_haircut() -> eyre::Result<()> {
 
     let fee_beneficiary = Address::ZERO;
 
-    // Create a custom fee token for the user
-    let user_fee_token =
-        setup_token_manual(&mut setup.node, &user_provider, &user_signer, chain_id).await?;
+    // Create a two-hop fee route: user_fee_token -> hop_fee_token -> PATH_USD.
+    // Track the next available nonce so CodeQL doesn't flag sequential test nonces as
+    // hard-coded cryptographic values.
+    let mut nonce = 0u64;
```
0xrusowsky:
```suggestion
    let mut nonce = 0u64;
```

crates/evm/src/block.rs
```diff
@@ -99,6 +99,11 @@ pub struct TempoTxResult {
     tx: Option<TempoTxEnvelope>,
     /// Block gas consumed by this transaction. The block `gas_used` field will be incremented by this value.
     block_gas_used: u64,
+    /// Validator-credited fee (in the validator's fee token) reported by `collectFeePostTx`.
+    ///
+    /// Used by the payload builder to score blocks by actual proposer revenue. The value is the
+    /// post-feeAMM amount, regardless of route shape — absorbs any number of pool haircuts.
```
0xKitsune, line 105: nit: what does pool haircuts mean here?

0xKitsune (approved):

crates/evm/src/block.rs
0xrusowsky: fee taken by each pool swap. should i rephrase?

PR #4004, opened 2026-05-15
feat: encode the network identity into the chainspec

hamdiallam:
By default, the network identity is parsed from `extra_data` from the genesis block. If there is a rotation, consensus args can be used for non-tempo networks to set the identity 
```
--consensus.network-identity
--consensus.network-identity-from-epoch
```

Encode the identities for testnet & mainnet in the binaries

### DIFF
--- a/crates/chainspec/src/network_identity.rs
+++ b/crates/chainspec/src/network_identity.rs
@@ -3,6 +3,7 @@
 use alloy_primitives::hex;
 use commonware_codec::ReadExt as _;
 use commonware_cryptography::bls12381::primitives::variant::{MinSig, Variant};
+use eyre::Context;
 use tempo_dkg_onchain_artifacts::OnchainDkgOutcome;
 
 const MODERATO_NETWORK_IDENTITY_EPOCH: u64 = 51;
@@ -50,11 +51,12 @@ impl NetworkIdentity {
         }
     }
 
-    pub(crate) fn from_extra_data(extra_data: &[u8]) -> Option<Self> {
+    pub(crate) fn from_extra_data(extra_data: &[u8]) -> eyre::Result<Self> {
         let mut extra_data = extra_data;
-        let outcome = OnchainDkgOutcome::read(&mut extra_data).ok()?;
+        let outcome =
+            OnchainDkgOutcome::read(&mut extra_data).wrap_err("unable to parse dkg outcome")?;
 
-        Some(Self {
+        Ok(Self {
             from_epoch: outcome.epoch.get(),
             identity: *outcome.network_identity(),
         })
--- a/crates/chainspec/src/network_identity.rs
+++ b/crates/chainspec/src/network_identity.rs
@@ -1,8 +1,23 @@
 //! Compiled network identities.
 
+use alloy_primitives::hex;
 use commonware_codec::ReadExt as _;
 use commonware_cryptography::bls12381::primitives::variant::{MinSig, Variant};
 use tempo_dkg_onchain_artifacts::OnchainDkgOutcome;
+
+const MODERATO_NETWORK_IDENTITY_EPOCH: u64 = 51;
+const MODERATO_NETWORK_IDENTITY: [u8; 96] = hex!(
+    "0x84591ad702a9ee67c0c64add2ff166c19a4666a1dc636cc530a810052957d34c"
+    "185bb1d2c7f5569983485a5af49baed70166ba17ae782bc8c75701099c704747"
+    "98ccc181d03b0c12054f1d01c7817b27b425bae4bfcf936218c0d097cccf3242"
+);
+
+const PRESTO_NETWORK_IDENTITY_EPOCH: u64 = 0;
+const PRESTO_NETWORK_IDENTITY: [u8; 96] = hex!(
+    "0xa217bb85001d4dcf8e5c50136f77af88cb2cab1857279b91c6240f41cca95c4f"
+    "43f6dcab3e0dfb87dafb3ecbeb6251e90a5df2e6c47432482821cd8b84665ee4"
+    "642589d2d9628a92b03e2bbfb00e006d038cd98def76d2a41b7c228c05f5a193"
+);
 
 /// A compiled weak-subjectivity anchor for the consensus network identity.
 ///
@@ -18,6 +33,26 @@ pub struct NetworkIdentity {
 }
 
 impl NetworkIdentity {
+    pub(crate) fn moderato() -> Self {
+        let identity = <MinSig as Variant>::Public::read(&mut MODERATO_NETWORK_IDENTITY.as_ref())
+            .expect("invalid network identity");
+
+        Self {
+            identity,
+            from_epoch: MODERATO_NETWORK_IDENTITY_EPOCH,
+        }
+    }
+
+    pub(crate) fn presto() -> Self {
+        let identity = <MinSig as Variant>::Public::read(&mut PRESTO_NETWORK_IDENTITY.as_ref())
+            .expect("invalid network identity");
+
+        Self {
+            identity,
+            from_epoch: PRESTO_NETWORK_IDENTITY_EPOCH,
+        }
+    }
+
     pub(crate) fn from_genesis_extra_data(extra_data: &[u8]) -> Option<Self> {
         let outcome = OnchainDkgOutcome::read(&mut extra_data.as_ref()).ok()?;
         Some(Self {
--- a/crates/chainspec/src/network_identity.rs
+++ b/crates/chainspec/src/network_identity.rs
@@ -19,7 +19,7 @@ const PRESTO_NETWORK_IDENTITY: [u8; 96] = hex!(
     "642589d2d9628a92b03e2bbfb00e006d038cd98def76d2a41b7c228c05f5a193"
 );
 
-/// This holds the key that known to be active. The genesis-derived
+/// This holds the key that is known to be active. The genesis-derived
 /// value must be updated with a newer identity after a full DKG rotation.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct NetworkIdentity {
--- a/crates/chainspec/src/network_identity.rs
+++ b/crates/chainspec/src/network_identity.rs
@@ -19,11 +19,8 @@ const PRESTO_NETWORK_IDENTITY: [u8; 96] = hex!(
     "642589d2d9628a92b03e2bbfb00e006d038cd98def76d2a41b7c228c05f5a193"
 );
 
-/// A compiled weak-subjectivity anchor for the consensus network identity.
-///
-/// This records the BLS threshold public key that a release knows is active for
-/// a chain. The genesis-derived value must be updated with a newer identity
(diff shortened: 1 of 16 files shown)


crates/commonware-node/src/args.rs
```diff
@@ -24,6 +24,20 @@ pub struct Args {
     #[arg(long = "consensus.signing-share")]
     pub signing_share: Option<PathBuf>,
 
+    /// Consensus network identity. Otherwise derived from genesis
+    #[arg(
+        long = "consensus.network-identity",
+        requires = "network_identity_from_epoch"
+    )]
+    pub network_identity: Option<String>,
```
hamdiallam: todo, proper clap arg type for this

crates/chainspec/src/network_identity.rs
```diff
@@ -0,0 +1,62 @@
...
+    /// First epoch for which `identity` is expected to verify finalizations.
+    pub from_epoch: u64,
+    /// BLS threshold public key.
+    pub identity: <MinSig as Variant>::Public,
+}
+
+impl NetworkIdentity {
+    pub(crate) fn moderato() -> Self {
+        let identity = <MinSig as Variant>::Public::read(&mut MODERATO_NETWORK_IDENTITY.as_ref())
+            .expect("invalid network identity");
```
SuperFluffy: Technically you expect a valid network identity :-)

```suggestion
            .expect("valid network identity");
```

crates/chainspec/src/network_identity.rs
```diff
@@ -0,0 +1,62 @@
...
+
+        Self {
+            identity,
+            from_epoch: MODERATO_NETWORK_IDENTITY_EPOCH,
+        }
+    }
+
+    pub(crate) fn presto() -> Self {
+        let identity = <MinSig as Variant>::Public::read(&mut PRESTO_NETWORK_IDENTITY.as_ref())
+            .expect("invalid network identity");
```
SuperFluffy: Likewise:

```suggestion
            .expect("valid network identity");
```

crates/chainspec/src/network_identity.rs
```diff
@@ -0,0 +1,62 @@
...
+/// value must be updated with a newer identity after a full DKG rotation.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct NetworkIdentity {
+    /// First epoch for which `identity` is expected to verify finalizations.
+    pub from_epoch: u64,
+    /// BLS threshold public key.
+    pub identity: <MinSig as Variant>::Public,
+}
+
+impl NetworkIdentity {
```
SuperFluffy, line 33: I guess let's have tests for these to ensure this doesn't suddently fail at runtime?

crates/chainspec/src/network_identity.rs
```diff
@@ -0,0 +1,62 @@
...
+        let identity = <MinSig as Variant>::Public::read(&mut PRESTO_NETWORK_IDENTITY.as_ref())
+            .expect("invalid network identity");
+
+        Self {
+            identity,
+            from_epoch: PRESTO_NETWORK_IDENTITY_EPOCH,
+        }
+    }
+
+    pub(crate) fn from_extra_data(extra_data: &[u8]) -> Option<Self> {
```
SuperFluffy: Why Option and not Result?

crates/chainspec/src/spec.rs
```diff
@@ -171,6 +177,8 @@ pub struct TempoChainSpec {
     /// [`ChainSpec`].
     pub inner: ChainSpec<TempoHeader>,
     pub info: TempoGenesisInfo,
+    /// Consensus network identity derived from genesis
+    pub network_identity: Option<NetworkIdentity>,
```
SuperFluffy, line 181: There should always be a network identity, right? It's either in the chainspec, or provided by some other means.

For test environments we should just provide a dummy value.

Can probably do that in a follow-up because this requires changing xtask.

crates/commonware-node/src/args.rs
```diff
@@ -24,6 +29,21 @@ pub struct Args {
...
+    )]
+    pub(crate) network_identity: Option<BlsPublicKeyParser>,
+
+    /// Epoch where --consensus.network-identity rotation occurred. If the network identity
+    /// was explicitly set, there must have been a network rotation after genesis.
+    #[arg(
+        long = "consensus.network-identity-from-epoch",
+        requires = "network_identity"
+    )]
+    pub(crate) network_identity_from_epoch: Option<u64>,
```
SuperFluffy: This acts as an override basically?

The PR text explains that this is for networks other than mainnet or testnet. But they could just update their chainspec as well.

I guess this would be interesting for mainnet/testnet as well if we have not yet gotten around to updating the chainspec and need a mechanism to override?

crates/commonware-node/src/args.rs
```diff
@@ -252,6 +272,21 @@ pub struct Args {
     pub storage_dir: Option<PathBuf>,
 }
 
+#[derive(Debug, Clone, Copy)]
+pub(crate) struct BlsPublicKeyParser(BlsPublicKey);
```
SuperFluffy: Let's call this `NetworkIdentity`.

crates/commonware-node/src/alias.rs
```diff
@@ -1,5 +1,9 @@
 //! A collection of aliases for frequently used (primarily commonware) types.
 
+use commonware_cryptography::bls12381::primitives::variant::{MinSig, Variant};
+
+pub(crate) type BlsPublicKey = <MinSig as Variant>::Public;
```
SuperFluffy: IMO instead of this alias introduce `NetworkIdentity` and make it parseable.

I know this clashes with the `NetworkIdentity` definition in chainspec - but the constant term of polynomial is literally known as the "network identity", so this would be nice here.

crates/commonware-node/src/feed/state.rs
SuperFluffy, line 43: Wondering if we should get rid of this functionality.

SuperFluffy (commented):
Agree with the general direction, but I think it is confusing that at times we call the `bls12381 public key` the network identity, and in other instasnces the tuple `(bls12381 public key, epoch)`.

Think we can streamline that.

crates/commonware-node/src/args.rs
hamdiallam: yea an override not ever used for our mainnet/testnet since we'll compile it in the binary (besides what you noted).


Seemed simpler to keep the chainspec oriented as a constant genesis spec. I initially updated the xtask to be able to avoid these flags but that it seemed odd to have a mutable chainspec? Open to switching back though

crates/commonware-node/src/feed/state.rs
hamdiallam: yea agreed not particularly useful. Probably good xtask to have as a general utility but not as an api

crates/chainspec/src/network_identity.rs
hamdiallam: the chainspec crates needs a bit of a refresh since errors can't be bubbled up as-is. Expects a valid chainspec to be generated

crates/chainspec/src/spec.rs
hamdiallam: yea exactly was getting hairy with the xtasks changes needed

hamdiallam: In the follower PR, when starting the engine we `.expect` on the network identity

crates/commonware-node/src/alias.rs
hamdiallam: Ah didn't see the not making it an alias, let me update

crates/chainspec/src/network_identity.rs
hamdiallam: Updated to be a Result and left a comment in `TempoChainSpec::from_genesis` on a todo for the Dev networks so that we can remove the Option<NetworkIdentity>

crates/chainspec/src/spec.rs
hamdiallam: left a todo in `from_genesis` regarding the followup

crates/commonware-node/src/args.rs
hamdiallam: Also note if this a user manually setting this, using the `xtask` requires also setting the appropriate fields for everything else as well

crates/commonware-node/src/network_identity.rs
```diff
@@ -0,0 +1,28 @@
...
+pub(crate) struct NetworkIdentity(pub <MinSig as Variant>::Public);
+
+impl NetworkIdentity {
+    pub(crate) fn public_key(&self) -> <MinSig as Variant>::Public {
+        self.0
+    }
+}
+
+impl FromStr for NetworkIdentity {
+    type Err = Box<dyn std::error::Error + Send + Sync + 'static>;
```
SuperFluffy: Let's type this error out with `thiserror`. Simple enough.

crates/chainspec/src/network_identity.rs
```diff
@@ -0,0 +1,92 @@
...
+
+const MODERATO_NETWORK_IDENTITY_EPOCH: u64 = 51;
+const MODERATO_NETWORK_IDENTITY: [u8; 96] = hex!(
+    "0x84591ad702a9ee67c0c64add2ff166c19a4666a1dc636cc530a810052957d34c"
+    "185bb1d2c7f5569983485a5af49baed70166ba17ae782bc8c75701099c704747"
+    "98ccc181d03b0c12054f1d01c7817b27b425bae4bfcf936218c0d097cccf3242"
+);
+
+const PRESTO_NETWORK_IDENTITY_EPOCH: u64 = 0;
+const PRESTO_NETWORK_IDENTITY: [u8; 96] = hex!(
```
SuperFluffy: Have a preference for using using `MAINNET*` and `TESTNET*`. IMO clearer than remembering names.

crates/chainspec/src/network_identity.rs
```diff
@@ -0,0 +1,92 @@
...
+);
+
+/// This holds the key that is known to be active. The genesis-derived
+/// value must be updated with a newer identity after a full DKG rotation.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct NetworkIdentity {
+    /// First epoch for which `identity` is expected to verify finalizations.
+    pub from_epoch: u64,
+    /// BLS threshold public key.
+    pub identity: <MinSig as Variant>::Public,
```
SuperFluffy, line 30: We should probably move these to some crate (so that tempo-chainspec does not import all of commonware-node), while avoiding typing this out everywhere.

SuperFluffy (approved):

crates/chainspec/src/network_identity.rs
hamdiallam: agreed, i'll switch them over

crates/chainspec/src/network_identity.rs
hamdiallam: This file generally? We only pull in the crypto dependencies, not all of commonware-node

We would still transitively depend on this stuff even if in an isolated crate right? Since we read/validate the key itself

PR #4006, opened 2026-05-15
feat(alloy): add RelayTransport for sponsor/fee-payer support

decofe:
## Summary

Add `RelayTransport` to `tempo-alloy` — a transport decorator that routes `eth_sendRawTransaction` to a sponsor service URL while forwarding all other RPC methods to the default transport.

This is the Alloy (Rust) equivalent of viem's [`withRelay`](https://viem.sh/docs/tempo/transports/withRelay) / `withFeePayer` transport decorator.

## Usage

```rust
use alloy_transport_http::Http;
use tempo_alloy::transport::RelayTransport;

let transport = RelayTransport::new(
    Http::new("https://rpc.tempo.xyz".parse()?),
    Http::new("https://sponsor.tempo.xyz/tp_abc123".parse()?),
);
// Use with ProviderBuilder — all sendRawTransaction calls
// go through the sponsor service automatically.
```

## How it works

- Implements `tower::Service<RequestPacket>` to intercept RPC dispatch
- Routes `eth_sendRawTransaction` → relay transport (sponsor URL)
- Routes everything else → default transport (RPC URL)
- The sponsor service (e.g. Tempo fee-payer Worker) decodes the user-signed tx, adds `fee_payer_signature`, and broadcasts

## Changes

- New `transport` module with `RelayTransport<D, R>` struct
- Added workspace deps: `alloy-json-rpc`, `tower`
- 2 unit tests verifying routing behavior

PR #4010, opened 2026-05-17
feat(alloy): add RelayTransport for sponsor/fee-payer support

0xrusowsky:
Adds Tempo Alloy sponsorship support for AA transactions through the fee-payer sponsor service. Supersedes https://github.com/tempoxyz/tempo/pull/4006

dual backend mode:
- default: RPC used for all requests that don't require sponsorship
- relay: sponsorship RPC

dual sponsorship mode: 
- `SignOnly`: expects the sponsorship service to only co-sign the txs and auth its sponsorship. The answer is validated and then broadcasted via the default rpc.
- `SignAndRelay`: expects the sponsorship service to sign and relay the tx.

example usage:

   ```rust
   use alloy::{
       network::EthereumWallet,
       providers::{ProviderBuilder, fillers::RecommendedFillers},
   };
   use tempo_alloy::{
       TempoNetwork,
       fillers::Random2DNonceFiller,
       provider::ext::TempoProviderBuilderExt,
       transport::SponsorshipMode,
   };

   let provider = ProviderBuilder::<_, _, TempoNetwork>::default()
       .filler(Random2DNonceFiller)
       .filler(<TempoNetwork as RecommendedFillers>::recommended_fillers())
       .wallet(EthereumWallet::from(signer))
       .sponsor(sponsor_url) // default: sponsor signs and relays
       .connect(rpc_url)
       .await?;
```
```rs
   // Or sign through the sponsor, then broadcast through `rpc_url`.
   let provider = ProviderBuilder::<_, _, TempoNetwork>::default()
       .filler(Random2DNonceFiller)
       .filler(<TempoNetwork as RecommendedFillers>::recommended_fillers())
       .wallet(EthereumWallet::from(signer))
       .sponsor_with_mode(sponsor_url, SponsorshipMode::SignOnly)
       .connect(rpc_url)
       .await?;
   ```

**NOTE:** also fixes Tempo AA transaction encoding for sponsor compatibility. The sponsor service treats `feePayerSignature === null` as the explicit sponsorship request, and `ox/tempo` encodes that marker as a single `0x00` byte. `tempo-primitives` now distinguishes absent fee-payer signatures, the sponsorship marker, and real fee-payer signatures in RLP so Rust emits the marker expected by the sponsor instead of a zero-valued signature tuple.

### CODE REVIEW

File: crates/alloy/Cargo.toml [hunk 1]
```diff
@@ -24,26 +24,31 @@ alloy-eips.workspace = true
 alloy-network.workspace = true
 alloy-primitives = { workspace = true, features = ["rand"] }
 alloy-provider.workspace = true
+alloy-rpc-client.workspace = true
```
figtracer: should we also enable `alloy-rpc-client/reqwest`?

crates/alloy/src/transport.rs
```diff
@@ -0,0 +1,786 @@
...
+    match mode {
+        SponsorshipMode::SignAndRelay => relay.call(RequestPacket::Single(request)).await,
+        SponsorshipMode::SignOnly => {
+            let sign_request = tx_request_with_metadata(SIGN_RAW_TX, raw_tx, &request)?;
+            let signed_tx: String = match relay.call(RequestPacket::Single(sign_request)).await? {
+                ResponsePacket::Single(response) => match response.payload {
+                    ResponsePayload::Success(payload) => serde_json::from_str(payload.get())
+                        .map_err(TransportErrorKind::non_retryable)?,
+                    ResponsePayload::Failure(err) => {
+                        return Err(TransportErrorKind::custom_str(&err.message));
```
figtracer: can we propagate error codes here instead of the custom str?

crates/alloy/src/transport.rs
```diff
@@ -0,0 +1,786 @@
...
+                    }
+                },
+                ResponsePacket::Batch(_) => {
+                    return Err(TransportErrorKind::custom_str(
+                        "sponsor returned a batch response to eth_signRawTransaction",
+                    ));
+                }
+            };
+
+            validate_signed_tempo_aa(&signed_tx)?;
```
figtracer: not sure if this is sufficient, should we decode and compare payloads as well?

crates/alloy/src/fillers/sponsor.rs
```diff
@@ -0,0 +1,50 @@
...
+
+    fn status(&self, tx: &N::TransactionRequest) -> FillerControlFlow {
+        if tx
+            .fee_payer_signature
+            .as_ref()
+            .is_some_and(|sig| *sig == FEE_PAYER_SIGNATURE_MARKER)
+        {
+            FillerControlFlow::Finished
+        } else {
+            FillerControlFlow::Ready
```
figtracer, line 24: if a user passes a sig this is `Ready` but `fill()` doesn't do anything; can't we get stuck in an infinite loop here? maybe i'm missing how this is supposed to work

crates/alloy/Cargo.toml
0xrusowsky: addressed in [`abca3ed` (this PR)](https://github.com/tempoxyz/tempo/pull/4010/commits/abca3ed9db06b081d13f8d8c602c8e15da00ba89)

crates/alloy/src/transport.rs
0xrusowsky: addressed in [`abca3ed` (this PR)](https://github.com/tempoxyz/tempo/pull/4010/commits/abca3ed9db06b081d13f8d8c602c8e15da00ba89)

crates/alloy/src/transport.rs
0xrusowsky: good call, hardened the validation to check all fields now.
addressed in [`abca3ed` (this PR)](https://github.com/tempoxyz/tempo/pull/4010/commits/abca3ed9db06b081d13f8d8c602c8e15da00ba89)

crates/alloy/src/fillers/sponsor.rs
0xrusowsky: i think u are right, i went too straightfwd with the impl.
addressed in [`abca3ed` (this PR)](https://github.com/tempoxyz/tempo/pull/4010/commits/abca3ed9db06b081d13f8d8c602c8e15da00ba89)

crates/alloy/src/transport.rs
```diff
@@ -0,0 +1,832 @@
...
+
+        assert_eq!(relay.methods(), vec![SIGN_RAW_TX]);
+        assert_eq!(relay.params(0), serde_json::json!([unsigned_raw_tx]));
+        assert_eq!(default.methods(), vec![SEND_RAW_TX]);
+        assert_eq!(default.params(0), serde_json::json!([signed_raw_tx]));
+    }
+
+    #[tokio::test]
+    async fn sign_only_rejects_sponsor_response_without_fee_payer_signature() {
+        let unsigned_raw_tx = signed_tempo_aa_raw_tx_with_nonce(false, 1);
```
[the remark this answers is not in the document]
decofe: this code is used in tests

crates/alloy/src/transport.rs
```diff
@@ -0,0 +1,832 @@
...
+            .unwrap_err();
+
+        assert!(err.to_string().contains("fee-payer signature placeholder"));
+        assert_eq!(relay.methods(), vec![SIGN_RAW_TX]);
+        assert!(default.methods().is_empty());
+    }
+
+    #[tokio::test]
+    async fn sign_only_rejects_sponsor_response_with_different_payload() {
+        let unsigned_raw_tx = signed_tempo_aa_raw_tx_with_nonce(false, 1);
```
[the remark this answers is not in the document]
decofe: this code is used in tests

crates/alloy/src/transport.rs
```diff
@@ -0,0 +1,832 @@
...
+            .await
+            .is_err()
+        );
+        assert_eq!(relay.methods(), vec![SEND_RAW_TX]);
+        assert!(default.methods().is_empty());
+    }
+
+    #[tokio::test]
+    async fn preserves_request_id_when_forwarding_to_relay() {
+        let raw_tx = signed_tempo_aa_raw_tx_with_nonce(false, 1);
```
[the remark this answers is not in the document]
decofe: this code is used in tests

crates/alloy/src/transport.rs
```diff
@@ -0,0 +1,832 @@
...
+        relay.push_success(&alloy_primitives::B256::ZERO);
+        let mut transport = RelayTransport::new(default, relay.clone());
+        let req = RequestPacket::Single(send_req_with_id(&raw_tx, Id::String("abc".into())));
+        assert!(tower::Service::call(&mut transport, req).await.is_ok());
+        assert_eq!(relay.ids(), vec![Id::String("abc".into())]);
+    }
+
+    #[tokio::test]
+    async fn preserves_request_headers_when_forwarding_to_relay() {
+        let raw_tx = signed_tempo_aa_raw_tx_with_nonce(false, 1);
```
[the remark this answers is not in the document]
decofe: this code is used in tests

crates/alloy/src/transport.rs
```diff
@@ -0,0 +1,832 @@
...
+                    }
+                },
+                ResponsePacket::Batch(_) => {
+                    return Err(TransportErrorKind::custom_str(
+                        "sponsor returned a batch response to eth_signRawTransaction",
+                    ));
+                }
+            };
+
+            let signed_tx_envelope = validate_signed_tempo_aa(&signed_tx)?;
```
figtracer: i think it's okay now, the node should reject a malformed sig anyway

crates/primitives/src/transaction/tempo_transaction.rs
mattsse, line 1: why is any of this necessary?

looks totally unrelated to this feature

crates/alloy/src/transport.rs
```diff
@@ -0,0 +1,846 @@
...
+fn tx_request_with_metadata(
+    method: &'static str,
+    tx: &str,
+    original: &SerializedRequest,
+) -> Result<SerializedRequest, TransportError> {
+    let mut request: SerializedRequest = Request::new(method, original.id().clone(), Some([tx]))
+        .try_into()
+        .map_err(TransportErrorKind::non_retryable)?;
+    if let Some(headers) = original.headers() {
+        request.headers_mut().extend(headers.clone());
```
stevencartavia, line 340: I think in `handle_sponsored_send_raw_transaction()`'s `match mode`, this clones every header from the user's original RPC request onto the sponsor request, so any per-request header meant for the default RPC gets sent to the sponsor.

crates/primitives/src/transaction/tempo_transaction.rs
0xrusowsky: done in a separate PR https://github.com/tempoxyz/tempo/pull/4033 using a helper that doesn't change RLP encoding

crates/alloy/src/transport.rs
0xrusowsky: i did this on purpose, but upon further reflection i understand that ur concern makes sense, as we don't want to leak headers to sponsor if its a different entity.

added user control over header forwarding, as well as auth capabilities for both RPCs in [`5f88745` (this PR)](https://github.com/tempoxyz/tempo/pull/4010/commits/5f8874532d2b57117ee64ecf3800ee84abed1c83)

crates/alloy/src/transport.rs
```diff
@@ -0,0 +1,913 @@
...
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Result<(), Self::Error>> {
+        futures::ready!(self.default.poll_ready(cx))?;
+        futures::ready!(self.relay.poll_ready(cx))?;
+        std::task::Poll::Ready(Ok(()))
+    }
+
+    fn call(&mut self, request: RequestPacket) -> Self::Future {
+        match request {
+            RequestPacket::Single(req) if req.method() == SEND_RAW_TX => {
```
figtracer: should we route `eth_sendRawTransactionSync` through the sponsor path too?

.changelog/jolly-rats-dry.md
```diff
@@ -0,0 +1,6 @@
+---
+tempo-alloy: minor
+---
+
+Added sponsored transaction support via a new `RelayTransport` that routes `eth_sendRawTransaction` requests through a sponsor service.
+Introduced `SponsorFiller`, `SponsoredProviderBuilder`, and `.sponsor()` / `.sponsor_with_mode()` provider builder extensions supporting both sign-and-relay and sign-only modes.
```
figtracer:
```suggestion
Introduced `SponsorFiller`, `SponsoredProviderBuilder`, and `.sponsor()` / `.sponsor_with_config()` provider builder extensions supporting both sign-and-relay and sign-only modes.
```

crates/alloy/src/transport.rs
0xrusowsky: addressed in [`6e982c2` (this PR)](https://github.com/tempoxyz/tempo/pull/4010/commits/6e982c2e22e5c46f852c358e8fb24366a3567163)

crates/alloy/src/transport.rs
```diff
@@ -0,0 +1,940 @@
+//! Relay transport for routing sponsored transactions through a fee payer service.
+//!
+//! `RelayTransport` wraps two transports:
+//! - a default Tempo RPC transport for ordinary requests and sign-only broadcasts.
+//! - a sponsor transport for signing or sign-and-relay raw transaction submissions.
+//!
+//! When a single `eth_sendRawTransaction` or `eth_sendRawTransactionSync` request is submitted, the
+//! raw unsigned Tempo AA transaction is locally preflighted. In `SponsorshipMode::SignAndRelay` it is
+//! forwarded unchanged to the sponsor service, which signs, broadcasts, and returns the transaction
```
stevencartavia:
```suggestion
//! re-encoded for the fee-payer service and forwarded to the sponsor, which signs, broadcasts, and returns the transaction
```

crates/alloy/src/provider/ext.rs
```diff
@@ -199,13 +335,29 @@ pub trait TempoProviderBuilderExt {
...
+        self.sponsor_with_config(sponsor_rpc, SponsorConfig::default())
+    }
+
+    fn sponsor_with_config(
+        self,
+        sponsor_rpc: impl Into<String>,
+        sponsor_config: SponsorConfig,
+    ) -> SponsoredProviderBuilder<L, JoinFill<F, SponsorFiller>, TempoNetwork> {
+        SponsoredProviderBuilder {
+            inner: self.filler(SponsorFiller),
```
mattsse: `self.filler(SponsorFiller)` appends the sponsor filler after the current filler stack. With the documented `.wallet(...).sponsor(...)` order, `F` already includes `WalletFiller`; Alloy runs fillers left-to-right, so wallet signing turns the builder into an envelope before `SponsorFiller::fill` can mutate `fee_payer_signature`. That leaves the raw tx without the sponsorship marker and makes `RelayTransport` reject it before calling the sponsor. Please make the sponsor marker run before wallet signing, or otherwise make this documented order safe.

mattsse (changes requested):
The documented `.wallet(...).sponsor(...)` path appears to append `SponsorFiller` after the wallet filler, so the wallet signs first and converts the builder into an envelope before the sponsor filler can set the fee-payer marker.

Permalinks:
- `.sponsor_with_config` appends `SponsorFiller` to the existing filler stack: https://github.com/tempoxyz/tempo/blob/6e982c2e22e5c46f852c358e8fb24366a3567163/crates/alloy/src/provider/ext.rs#L349-L356
- `SponsorFiller::fill` only mutates a builder, so it no-ops once the tx is already an envelope: https://github.com/tempoxyz/tempo/blob/6e982c2e22e5c46f852c358e8fb24366a3567163/crates/alloy/src/fillers/sponsor.rs#L39-L44
- The README advertises the affected `.wallet(...).sponsor(...)` order: https://github.com/tempoxyz/tempo/blob/6e982c2e22e5c46f852c358e8fb24366a3567163/crates/alloy/README.md#L127-L132

In that flow, the raw tx reaches `RelayTransport` without `FEE_PAYER_SIGNATURE_MARKER`, so it gets rejected by the unsigned-AA validation instead of being routed to the sponsor.

Suggested regression test: add a mocked-provider test using the README builder order (`Random2DNonceFiller` + recommended fillers + `.wallet(...)` + `.sponsor(...)`) and an AA request, then assert that the filled/signed envelope decodes with `fee_payer_signature == Some(FEE_PAYER_SIGNATURE_MARKER)` before transport submission. The current ordering should fail that assertion because wallet signing happens before the sponsor filler runs.

crates/alloy/src/provider/ext.rs
0xrusowsky: addressed in [`8c0d947` (this PR)](https://github.com/tempoxyz/tempo/pull/4010/commits/8c0d9470c2da90a54b2610d63092159d3fe572bd)

mattsse (approved):
lgtm

PR #4013, opened 2026-05-18
docs(tip-1059): add discounted payment settlement

malleshpai:
Adds the TIP-1059 spec for discounted pure TIP-20 payment transfers and burns. The proposal keeps normal base fee authorization while defining a narrower post-execution settlement discount for eligible low-gas payment-lane transactions.

### DIFF
--- a/tips/tip-1059.md
+++ b/tips/tip-1059.md
@@ -18,7 +18,7 @@ The goal is to make ordinary payments cheap while keeping the normal price for t
 
 ## Motivation
 
-Tempo's current fixed base fee is deliberately high enough to make worst-case spam and state growth expensive. That protects the chain, but it also means the most common payment path pays the same gas price as more expensive operations.
+Tempo's current fixed base fee is deliberately high enough to make worst-case spam and state growth expensive. That protects the chain, but it also means that transfers that do not create new state can be relatively expensive even in periods of low load. 
 
 Simple TIP-20 transfers to existing accounts and burns do not have the same long-term cost profile as state creation. They are the payment and redemption paths we most want to be cheap, predictable, and easy to explain. This TIP introduces a narrow discount for those paths without lowering the price of everything else.
 
@@ -59,10 +59,8 @@ A transaction is syntactically eligible for the discount only if all of the foll
    - `transferFromWithMemo(address from, address to, uint256 amount, bytes32 memo)`
    - `burn(uint256 amount)`
    - `burnWithMemo(uint256 amount, bytes32 memo)`
-3. The transaction does not create a contract.
-4. The transaction does not include an EIP-7702 authorization list.
 
-For account-abstraction transactions, every call in the batch must satisfy the TIP-20 operation requirement. A mixed batch is not eligible for the discount.
+For batch Tempo transactions, every call in the batch must satisfy the TIP-20 operation requirement. A mixed batch is not eligible for the discount.
 
 `transferFrom` is included because it is also a payment movement. The discount does not make allowance creation cheaper: approvals are not eligible.
 
@@ -137,9 +135,9 @@ This is the simplest change, but it makes storage creation and arbitrary contrac
 
 This is broader than the proposal in this TIP. The current payment-lane classifier includes more than simple payment and redemption operations. Discounting the whole lane risks making approvals, mints, channel operations, or future payment-lane operations cheaper even when they do not have the same cost profile as an ordinary payment.
 
-## Use a capped EIP-1559 fee for the payment lane
+## Use a capped EIP-1559 style fee controller
 
-A dynamic fee could make payments cheap most of the time while rising under load. It is also more complex: the protocol would need a second fee market, wallet behavior would need to change, and the interaction between payment-lane fees and normal fees would need careful design.
+A dynamic fee could make payments cheap most of the time while rising under load. It is also more complex: the protocol would need a fee market, wallet behavior would need to change, and a lot of permanent state could be added cheaply to the chain during periods of low prices. 
 
 ## Separate storage pricing from execution pricing
 
@@ -156,5 +154,5 @@ This is the clean long-term model. State creation could stay expensive while exe
 - The discount never applies to contract creation.
 - The discount never applies to a normal-fee-authorized transaction that uses more than 250,000 gas.
 - The discount never applies to a discount-only transaction with `gas_limit > 250,000`.
-- The discount never applies to a mixed account-abstraction batch that includes anything other than eligible TIP-20 transfers or burns.
+- The discount never applies to a mixed Tempo Transaction batch that includes anything other than eligible TIP-20 transfers or burns.
 - A transaction that authorizes the normal base fee remains valid whether or not discounted budget is available.
--- a/tips/tip-1059.md
+++ b/tips/tip-1059.md
@@ -14,7 +14,7 @@ protocolVersion: T6
 
 This TIP makes the first 50 million gas of simple TIP-20 transfers in each block cheaper than other transactions. Eligible transfers are charged 2 billion attodollars per gas instead of the normal 20 billion attodollars per gas.
 
-The goal is to make ordinary payments feel cheap while keeping the normal price for transactions that may create state, run arbitrary contracts, or consume the general lane.
+The goal is to make ordinary payments cheap while keeping the normal price for transactions that may create state, run arbitrary contracts, or consume the general lane.
 
 ## Motivation
 
@@ -26,10 +26,9 @@ Users and wallets do not need to rely on the discount for reliability. A transfe
 
 ## Assumptions
 
-- A TIP-20 transfer with `gas_limit <= 250,000` is not expected to create new account state. Transfers that create new account state are expected to require more gas.
+- A TIP-20 transfer with `gas_limit <= 250,000` cannot create new account state given current gas prices. 
 - The discount is a payment UX improvement, not a replacement for the normal base fee as the chain's main spam-resistance mechanism.
 - Wallets and partners may continue to set `max_fee_per_gas` to the normal base fee for reliable inclusion.
-- The discount is applied by protocol rules, not by trusting a block builder to voluntarily charge less.
 
 ---
 
@@ -104,7 +103,7 @@ Fee estimation for eligible pure TIP-20 transfers MAY show the discounted gas pr
 
 ## Lower the base fee for all transactions
 
-This is the simplest change, but it makes storage creation and arbitrary contract execution cheaper too. That weakens the main reason Tempo has a fixed base fee today: protecting the chain from cheap worst-case spam and permanent state growth.
+This is the simplest change, but it makes storage creation and arbitrary contract execution cheaper too. That weakens the protections the chain has from cheap worst-case spam and permanent state growth.
 
 ## Lower the payment-lane base fee for all payment-lane transactions
 
--- a/tips/tip-1059.md
+++ b/tips/tip-1059.md
@@ -12,7 +12,7 @@ protocolVersion: T6
 
 ## Abstract
 
-This TIP makes the first 50 million gas of simple TIP-20 transfers in each block cheaper than other transactions. Eligible transfers are charged 2 billion attodollars per gas instead of the normal 20 billion attodollars per gas.
+This TIP makes the first 50 million gas of simple TIP-20 payment operations in each block cheaper than other transactions. Eligible operations are charged 2 billion attodollars per gas instead of the normal 20 billion attodollars per gas.
 
 The goal is to make ordinary payments cheap while keeping the normal price for transactions that may create state, run arbitrary contracts, or consume the general lane.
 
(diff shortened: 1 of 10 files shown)


tips/tip-1059.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Block building
+
+Block builders SHOULD track remaining discounted gas while building a block. The payload fee total used for builder comparison MUST use the same fee amount that will be charged during settlement.
+
+## RPC and estimation
+
+Fee estimation for ordinary transactions is unchanged.
+
+Fee estimation for eligible pure TIP-20 payment operations MAY show the discounted gas price, but RPCs and wallets SHOULD make clear that transactions still authorize the normal base fee for reliable inclusion.
```
jenpaff: to clarify: wallet users will still see the normal base fee without the discount? I think this would cause a lot of confusion. should we consider adding a new rpc endpoint or updating estimate to share fees with discount?

tips/tip-1059.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Use a capped EIP-1559 style fee controller
+
+A dynamic fee could make payments cheap most of the time while rising under load. It is also more complex: the protocol would need a fee market, wallet behavior would need to change, and a lot of permanent state could be added cheaply to the chain during periods of low prices.
+
+## Separate storage pricing from execution pricing
+
+This is the clean long-term model. State creation could stay expensive while execution gets cheaper. It is also a larger protocol change than needed for the immediate goal of making ordinary payments cheaper.
+
+---
```
jenpaff: Lets add a compatibility section

tips/tip-1059.md
```diff
@@ -0,0 +1,141 @@
...
+A transaction is syntactically eligible for the discount only if all of the following are true:
+
+1. It satisfies the current payment-lane classification rules.
+2. Every user call in the transaction is one of:
+   - `transfer(address to, uint256 amount)`
+   - `transferWithMemo(address to, uint256 amount, bytes32 memo)`
+   - `transferFrom(address from, address to, uint256 amount)`
+   - `transferFromWithMemo(address from, address to, uint256 amount, bytes32 memo)`
+   - `burn(uint256 amount)`
+   - `burnWithMemo(uint256 amount, bytes32 memo)`
```
0xalpharush, line 54: 2 seems stricter than 1 following T5
https://github.com/tempoxyz/tempo/blob/7a5c8fca032f8ca9b0ce08f33bd2d1ddfc545e3c/tips/tip-1045.md?plain=1#L52-L68

tips/tip-1059.md
```diff
@@ -0,0 +1,141 @@
...
+
+For batch Tempo transactions, every call in the batch must satisfy the TIP-20 operation requirement. A mixed batch is not eligible for the discount.
+
+`transferFrom` is included because it is also a payment movement. The discount does not make allowance creation cheaper: approvals are not eligible.
+
+`burn` is included for redemption and off-ramp flows. It does not create recipient state.
```
0xrusowsky, line 63: there are some extra payment lane eligibility nuances that could we worth applying here.

imo it would be beneficial to add payment lane classification as pre-requisite for discounts, effectively making TIP-1045 a superset of TIP-1059

tips/tip-1059.md
```diff
@@ -0,0 +1,141 @@
...
+protocolVersion: T6
+---
+
+# TIP-1059: Discounted Pure Payment Transfers
+
+## Abstract
+
+This TIP makes the first 50 million gas of simple TIP-20 payment operations in each block cheaper than other transactions. Eligible operations are charged 2 billion attodollars per gas instead of the normal 20 billion attodollars per gas.
+
+The goal is to make ordinary payments cheap while keeping the normal price for transactions that may create state, run arbitrary contracts, or consume the general lane.
```
0xKitsune: nit: I would lead with the actual lever here, simple TIP-20 payment/redemption operations that do not exceed a specified gas limit can settle at a lower gas price, while wallets still authorize the normal base fee for reliable inclusion.


```suggestion
This TIP introduces a bounded per-block fee discount for simple TIP-20 payment and redemption operations that are below a specified gas limit. Transactions still authorize the normal base fee for reliable inclusion. After execution, if an eligible transfer, transferFrom, or burn stays within the per-transaction gas cap and discounted capacity remains in the block, the protocol charges the lower gas price and refunds the difference.

The discount is intentionally narrow. Approvals, mints, arbitrary contract execution, mixed batches, state-creating operations, and operations outside the discounted capacity continue to pay the normal base fee.
```

tips/tip-1059.md
0xalpharush: I guess this may be purposeful as "Lower the payment-lane base fee for all payment-lane transactions" is an alternative

tips/tip-1059.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Motivation
+
+Tempo's current fixed base fee is deliberately high enough to make worst-case spam and state growth expensive. That protects the chain, but it also means that transfers that do not create new state can be relatively expensive even in periods of low load.
+
+Simple TIP-20 transfers to existing accounts and burns do not have the same long-term cost profile as state creation. They are the payment and redemption paths we most want to be cheap, predictable, and easy to explain. This TIP introduces a narrow discount for those paths without lowering the price of everything else.
```
0xKitsune: minor nit: I would remove "easy to explain". Other drivers make sense here.

tips/tip-1059.md
```diff
@@ -0,0 +1,141 @@
...
+
+Users and wallets do not need to rely on the discount for reliability. A payment operation still authorizes the normal gas price. If it is actually cheap to execute and lands inside the per-block discounted gas cap, the chain charges the discounted price and refunds the difference.
```
0xKitsune: very minor nit: I think this section could be smoothed out a bit. Reads a little choppy in places. Consider something like:


```suggestion
Tempo keeps the fixed base fee high enough to price worst-case spam and state growth. That is the right default for arbitrary execution and state creating transactions, but it overprices simple payment flows whose execution does not create new state.

This TIP carves out a narrow discount for those low-cost payment and redemption paths: TIP-20 transfers to existing accounts, delegated transfers, and burns. The discount makes common payments cheaper without lowering the base fee for approvals, mints, contract execution, mixed batches, or state-creating operations.

Wallets still authorize the normal base fee. The chain applies the discount only after execution proves that the transaction is eligible, stays under the gas cap, and fits within the block’s remaining discount budget.
```

tips/tip-1059.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Assumptions
+
+- A TIP-20 payment operation that actually uses no more than 250,000 gas is treated as a cheap payment operation for this TIP.
```
0xKitsune: nit: Is this an assumption necessary to understand the TIP or could this just be a specification detail?

tips/tip-1059.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Parameters
+
+At the T6 hardfork, introduce the following parameters:
+
+| Parameter | Value | Meaning |
+| --- | ---: | --- |
+| Normal base fee | 20,000,000,000 | Existing gas price for ordinary transactions |
+| Discounted operation gas price | 2,000,000,000 | Gas price charged to eligible discounted operations |
+| Per-block discounted gas cap | 50,000,000 gas per block | Maximum gas that can receive the discount in one block |
```
0xKitsune: Why do we cap this? What if all txs in the block are state TIP20 operations <250k gas?

tips/tip-1059.md
0xKitsune: Yeah can this TIP effectively be simplified to "any payment lane tx (that also is in the set of functions eligible for the discount) below a gas threshold receive a discounted base fee" where payment lane constraints must hold?

tips/tip-1059.md
```diff
@@ -0,0 +1,141 @@
...
+
+For batch Tempo transactions, every call in the batch must satisfy the TIP-20 operation requirement. A mixed batch is not eligible for the discount.
+
+`transferFrom` is included because it is also a payment movement. The discount does not make allowance creation cheaper: approvals are not eligible.
+
+`burn` is included for redemption and off-ramp flows. It does not create recipient state.
+
+## Applying the discount
+
+Each block has a `discounted_operation_gas_left` counter initialized to 50,000,000 gas.
```
0xKitsune: Why do we have a cap here? Similar to my comment above, what if there are payment lane txs where every tx uses <250k gas? Can you clarify what the reasoning is to cap the discount? Does this then make it possible for users to spam, effectively using the entire discount, forcing other organizations to pay the undiscounted amount?

0xalpharush: I do think reconciling this TIP with profit-maxxing validator is worthwhile as IIUC they could just do a bunch of self-transfers to consume the cheap space and make more in fees (assuming cost of transfers > fees received from victim txs at a higher gas price)

tips/tip-1059.md
jenpaff: +1 - would add the rationale why the cap is needed

jenpaff (approved):

tips/tip-1059.md
```diff
@@ -0,0 +1,138 @@
...
+protocolVersion: T6
+---
+
+# TIP-1059: Discounted Pure Payment Transfers
+
+## Abstract
+
+This TIP lets simple TIP-20 payment and redemption operations settle at 12 billion attodollars per gas instead of the normal 20 billion attodollars per gas.
+
+The discount is intentionally narrow. It applies only to payment-lane transactions that call a small set of TIP-20 transfer or burn functions and actually use no more than 250,000 gas. Approvals, mints, arbitrary contract execution, mixed batches, and higher-gas transactions continue to pay the normal base fee.
```
0xKitsune: nit: remove "actually" here, reads a bit nicer

tips/tip-1059.md
```diff
@@ -0,0 +1,138 @@
...
+
+1. It satisfies the current payment-lane classification rules.
+2. Every user call in the transaction is one of:
+   - `transfer(address to, uint256 amount)`
+   - `transferWithMemo(address to, uint256 amount, bytes32 memo)`
+   - `transferFrom(address from, address to, uint256 amount)`
+   - `transferFromWithMemo(address from, address to, uint256 amount, bytes32 memo)`
+   - `burn(uint256 amount)`
+   - `burnWithMemo(uint256 amount, bytes32 memo)`
+3. The transaction uses no more than 250,000 gas.
```
0xKitsune: nit: It might be worth mentioning that 250k is the cost of new state writes, effectively making this discount apply to txs that do not create new state.

tips/tip-1059.md
```diff
@@ -0,0 +1,138 @@
...
+
+---
+
+# Invariants
+
+- The normal base fee remains unchanged for transactions that do not receive the discount.
+- The discount never applies to contract creation.
+- The discount never applies to a transaction that uses more than 250,000 gas.
+- The discount never applies to a mixed Tempo Transaction batch that includes anything other than eligible TIP-20 payment operations.
+- The discount never applies to a transaction that is not eligible for the payment lane.
```
0xKitsune: nit: some of these read more like eligibility constraints or correctness conditions than protocol invariants.

0xKitsune (approved):

tips/tip-1059.md
malleshpai:
```suggestion
The discount is intentionally narrow. It applies only to payment-lane transactions that call a small set of TIP-20 transfer or burn functions and use no more than 250,000 gas. Approvals, mints, arbitrary contract execution, mixed batches, and higher-gas transactions continue to pay the normal base fee.
```

tips/tip-1059.md
malleshpai:
```suggestion
3. The transaction uses no more than 250,000 gas. Since 250,000 gas is the cost of a new state write, this enforces the discount to apply only to transactions that do not create new state. 
```

tips/tip-1059.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Parameters
+
+At the T6 hardfork, introduce the following parameters:
+
+| Parameter | Value | Meaning |
+| --- | ---: | --- |
+| Normal base fee | 20,000,000,000 | Existing gas price for ordinary transactions |
+| Discounted payment gas price | 12,000,000,000 | Gas price charged to eligible payment operations |
+| Maximum discounted payment gas used | 250,000 gas | Per-transaction execution cap for the discount |
```
0xalpharush: Is this something that should be parameterized to track any changes that affect TIP-1000?
https://github.com/tempoxyz/tempo/blob/ac8b8237aec30412261419480bbaa677023262e3/crates/revm/src/gas_params.rs#L26

tips/tip-1059.md
```diff
@@ -0,0 +1,128 @@
...
+
+A transaction that receives the discount is charged the discounted payment gas price. A transaction that does not receive the discount is charged the normal base fee. If a transaction receives the discount, the unused authorized amount is refunded through the existing gas refund path.
+
+Wallets SHOULD continue to use the normal base fee by default. This gives users reliable inclusion while still letting them benefit from the discount when the transaction is eligible.
+
+## Block building
+
+The payload fee total used for builder comparison MUST use the same fee amount that will be charged during settlement.
+
+## RPC and estimation
```
jenpaff, line 83: We should check if any existing rpc's would break @klkvr @0xKitsune

tips/tip-1059.md
```diff
@@ -0,0 +1,128 @@
...
+2. Every user call in the transaction is one of:
+   - `transfer(address to, uint256 amount)`
+   - `transferWithMemo(address to, uint256 amount, bytes32 memo)`
+   - `transferFrom(address from, address to, uint256 amount)`
+   - `transferFromWithMemo(address from, address to, uint256 amount, bytes32 memo)`
+   - `burn(uint256 amount)`
+   - `burnWithMemo(uint256 amount, bytes32 memo)`
+3. The transaction uses no more than 250,000 gas. Since 250,000 gas is the cost of a new state write, this enforces the discount to apply only to transactions that do not create new state. 
+
+For batch Tempo transactions, every call in the batch must satisfy the TIP-20 operation requirement. A mixed batch is not eligible for the discount.
```
fgimenez: maybe we could be more explicit about batch txs, similar to TIP-1045 wording
```suggestion
For account-abstraction (type `0x76`) transactions, every call in `calls` must individually satisfy the eligible-function requirement. A transaction with any non-eligible call is not eligible for the discount.
```

fgimenez (approved):

tips/tip-1059.md
0xKitsune: This shouldnt break anything at the RPC level if we keep the discount as post execution settlement logic.

tips/tip-1059.md
0xKitsune: Good catch, I think it makes sense to parameterize this given the goal is to provide a discount for payment lane txs that do not create new state. @malleshpai any objections?

0xKitsune: Updated in https://github.com/tempoxyz/tempo/pull/4013/changes/6db44ea606158a889f78db689a1ab1ebe0fa7a29

0xalpharush (approved):

tips/tip-1059.md
```diff
@@ -0,0 +1,128 @@
+---
+id: TIP-1059
+title: Discounted Pure Payment Transfers
+description: Lets simple TIP-20 payment and redemption operations settle at a lower gas price.
+authors: Mallesh Pai
+status: Draft
```
jenpaff:
```suggestion
status: Approved
```

PR #4014, opened 2026-05-18
Allow tempoup to fall back to gh verification

decofe:
## Summary
- allow authenticated `gh` to satisfy tempoup release verification when `gpg` is unavailable, without requiring `--unsafe-skip-verify`
- consider both verifiers when both are installed: GPG remains enforced, and SLSA attestation verification is attempted too
- consolidate SLSA attestation handling into `verify_slsa_attestation <archive> <required|optional>` with short helper comments

## Behavior
- `gpg` present: checksum + GPG signature verification are required; authenticated `gh` also attempts SLSA verification as an optional additional check
- `gpg` missing, authenticated `gh` present: checksum + SLSA attestation verification are required
- neither `gpg` nor authenticated `gh` present: install fails unless `--unsafe-skip-verify` is used

## Testing
- `bash -n tempoup/tempoup`
- simulated PATH with `gh` present and `gpg` absent: verified preflight proceeds to required `gh attestation verify` and fails closed for an old release without attestations
- normal PATH with both `gpg` and authenticated `gh`: verified GPG succeeds and SLSA attestation is also attempted


0xalpharush:
@grandizzy i thought this could be simplified some i.e. we prioritize gh verification, fallback to gpg (do not do both), and then require passing flag to skip if neither are available. also think we should just call it "attestation" or "integrity" check versus SLSA

grandizzy:
> @grandizzy i thought this could be simplified some i.e. we prioritize gh verification, fallback to gpg (do not do both), and then require passing flag to skip if neither are available. also think we should just call it "attestation" or "integrity" check versus SLSA

@0xalpharush pls check https://github.com/tempoxyz/tempo/pull/4014/changes/5230f851323fa5b93cf3ff98b9c67ff83d851a5b

0xalpharush (approved):

PR #4015, opened 2026-05-18
TIP-1059: contract storage gas tokens

decofe:
## Summary
- add TIP-1059 for contract-level storage gas token accounting
- define mint/consume rules for deleted and newly-created contract storage slots
- specify the precompile interface for querying balances and setting default consume/preserve behavior
- explain how the mechanism can underlie TIP-1058 owner-level reusable storage

## Testing
- not run; docs-only TIP

### DIFF
--- a/tips/tip-1059.md
+++ b/tips/tip-1059.md
@@ -1,0 +1,164 @@
+---
+id: TIP-1059
+title: Contract Storage Gas Tokens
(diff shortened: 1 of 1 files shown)


decofe:
Closing in favor of a clean, unstacked TIP branch.  is already used by PR #4013 for a different TIP, so this proposal is moving to .

PR #4016, opened 2026-05-18
TIP-1060: contract storage credits

decofe:
## Summary
  - add TIP-1060 for per-account storage credits
  - define Refund, Preserve, and Direct storage creation modes
  - specify storage credit minting, consumption, settlement, and gas accounting rules
  - add the Storage Credits precompile for reading balances and setting account mode


tips/tip-1060.md
```diff
@@ -0,0 +1,164 @@
+---
+id: TIP-1060
+title: Contract Storage Gas Tokens
+description: Generalizes reusable storage to all smart contracts by giving each contract a protocol-managed counter for deleted storage slots that can offset later storage creation costs.
+authors: Dan Robinson
```
dankrad:
```suggestion
authors: Dankrad Feist
```

tips/tip-1060.md
```diff
@@ -0,0 +1,164 @@
...
+TIP-1058 applies reusable storage accounting to selected precompile-owned state with owner attribution. This TIP generalizes the underlying mechanism to all smart contracts at contract granularity. Contracts that need user-level attribution can implement their own owner accounting on top of the contract-level token balance; contracts that do not need attribution can use the default behavior and receive lower costs for storage churn automatically.
+
+## Assumptions
+
+- The protocol can identify storage transitions for each contract account, including `zero -> nonzero` creation and `nonzero -> zero` deletion.
+- The TIP-1000 state creation component can be omitted independently from the rest of the normal storage write cost.
+- Contract storage gas tokens are scoped to a single contract address and are not transferable across contracts.
+- A gas token represents previously deleted contract storage capacity, not a claim on gas refunds or execution gas.
+
+If the implementation cannot distinguish storage creation from storage update, or cannot debit the correct contract balance atomically with the storage write, this TIP MUST NOT activate.
```
dankrad:
```suggestion
```

tips/tip-1060.md
```diff
@@ -0,0 +1,159 @@
...
+    /// @notice Sets whether the calling contract consumes gas tokens by default for storage creation.
+    /// @param consumeTokens True to consume one gas token when available before paying the TIP-1000 creation cost; false to always pay gas and preserve tokens.
+    /// @dev This changes only the mode for msg.sender. EOAs MUST NOT be able to set the mode for a contract, and one contract MUST NOT be able to set the mode for another contract.
+    function setUseGasTokensByDefault(bool consumeTokens) external;
+}
+```
+
+`setUseGasTokensByDefault` applies to `msg.sender`. If `msg.sender` is an EOA, the call MUST revert. A contract may call this precompile during construction; in that case the mode applies to the address under construction even though runtime code has not yet been installed.
+
+The precompile MUST NOT expose any method that transfers gas tokens, mints gas tokens, burns gas tokens except through storage creation, or changes another contract's mode.
```
dankrad:
```suggestion
```

tips/tip-1060.md
```diff
@@ -0,0 +1,157 @@
...
+
+Contracts control this behavior through a new precompile that can query the contract's gas token balance and select whether storage creation should consume gas tokens by default. The default mode for all smart contracts is to consume gas tokens when available, allowing existing contracts to benefit from cheaper storage churn without modification.
+
+## Motivation
+
+TIP-1000 makes net-new storage creation expensive to protect Tempo from state growth attacks. That cost is appropriate when a contract increases its long-term state footprint, but it overcharges contracts that repeatedly delete and recreate storage slots while staying within a previously paid state footprint.
+
+TIP-1058 applies reusable storage accounting to selected precompile-owned state with owner attribution. This TIP generalizes the underlying mechanism to all smart contracts at contract granularity. Contracts that need user-level attribution can implement their own owner accounting on top of the contract-level token balance; contracts that do not need attribution can use the default behavior and receive lower costs for storage churn automatically.
```
0xKitsune: IMO I think it is cleaner to decouple this TIP from 1058 and just describe 1060 as a mechnaism to credit/apply contract storage gas to transactions.

tips/tip-1060.md
```diff
@@ -0,0 +1,157 @@
...
+contract_storage_gas_token_state: mapping(address contract => uint256 packed_state)
+```
+
+The packed state word uses the following layout:
+
+```text
+bits   0..63: gas_token_balance (uint64)
+bit       64: preserve_tokens flag
+bits 65..255: reserved for future hardfork-gated extensions
+```
```
0xKitsune: +1 on @danrobinson's comment, could this be a new precompile that tracks gas credits for all precompiles?

tips/tip-1060.md
```diff
@@ -0,0 +1,157 @@
...
+
+If `preserve_tokens == 0`, the contract is in consume-token mode. If `preserve_tokens == 1`, the contract is in preserve-token mode. The default packed state for every contract is zero, which means a zero balance and `ConsumeTokens` mode. Implementations MAY omit explicit storage for contracts with the default packed state.
```
0xKitsune: I may be missing something, but do we need this? Since precompiles already need custom logic for how these credits are used, could they just increment/decrement credits directly and implement whatever consumption rules they want internally?

dankrad: My idea on how this is used inside the precompile
```solidity
     
   mapping(address => uint256) public storageCredits;

   mapping(address => uint256) public userStorage;

    ...
    // User creates new storage
    if(storageCredits[msg.sender] > 0) {
       set_preserve_tokens(0);
       storageCredits[msg.sender] -= 1;
   } else {
      set_preserve_tokens(1);
   }
   userStorage[msg.sender] = value;

   // Restore default token preservation tracking
   set_preserve_tokens(1);
    ...
```

So the 1060 precompile gives contract-global tracking, which we can then, precompile per precompile (and maybe also for user contracts tbd) use to create per-account storage credits. 1060 says nothing about how that is implemented. 

A first step of implementing it could be:

- Enable 1060 only for the DEX and channels precompiles
- Create a new TIP that has the implementation of user storage credit tracking for these two precompiles

tips/tip-1060.md
dankrad: Can you clarify this, do you mean the per-account tracking of credits?
Why isn't it enough if 1060 tracks at the contract level, and then individual precompiles do the per-user tracking in their own storage?
The argument for this is that per-user tracking doesn't really generalize: Only the contract actually understands which storage slot should be attributable to which user, it isn't something that can generally be inferred. So the tracking logic will have to be custom anyway.

tips/tip-1060.md
dankrad: It is decoupled from 1058: There is no dependency of 1060 on 1058.

tips/tip-1060.md
```diff
@@ -0,0 +1,288 @@
...
+
+The mechanism applies uniformly to every account on Tempo, including ordinary smart contracts, precompile-owned state, and EOAs. EOAs do not normally hold storage, so in practice they will neither accumulate nor consume tokens, but the rule is account-keyed for uniformity.
+
+## Motivation
+
+TIP-1000 makes net-new storage creation expensive to protect Tempo from state growth attacks. That cost is appropriate when an account increases its long-term state footprint, but it overcharges accounts that repeatedly delete and recreate storage slots while staying within a previously paid state footprint. Examples include AMM orderbooks that constantly add and cancel orders, channel-style precompiles that open and close escrow slots, and ordinary contracts implementing transient data structures.
+
+TIP-1058 applies reusable storage accounting to selected precompile-owned state with owner attribution. This TIP provides the underlying account-level storage token mechanism that TIP-1058 and similar precompile-specific designs can build on, and makes the same primitive available to ordinary smart contracts through a single protocol-managed precompile.
+
```
dankrad:
```suggestion
```

tips/tip-1060.md
```diff
@@ -0,0 +1,288 @@
...
+`DirectTokens` mode charges less gas upfront when a token is available, but it creates a simulation-vs-inclusion failure mode: a transaction simulated against a state where `gas_token_balance[A] > 0` may be included against a state where another transaction has already drained the balance, in which case the included transaction can run out of gas and revert even though the user's intent and signed gas limit were unchanged.
+
+`RefundTokens` mode avoids that by always charging the full TIP-1000 creation cost upfront. When tokens are available at the end-of-transaction settlement, the user receives an equivalent refund per redeemed creation and pays the same net amount as `DirectTokens` would (subject to the EVM refund cap), but the worst-case upfront gas is bounded by the always-charged amount.
+
+End-of-transaction settlement also lets a single transaction self-finance its own creations: a transaction that creates one or more storage slots and later deletes any of its own storage slots in `RefundTokens` mode can use the just-minted tokens to refund the earlier creations, regardless of whether the account had a positive token balance when the transaction started.
+
```
dankrad:
```suggestion
```

tips/tip-1060.md
```diff
@@ -0,0 +1,288 @@
...
+`DirectTokens` mode remains available for accounts whose callers explicitly accept the simulation-vs-inclusion risk, for example because they reserve their own per-user storage credits inside the contract and only select `DirectTokens` mode when their accounting guarantees that a token is reserved for the specific operation about to execute.
```
dankrad:
```suggestion
`DirectTokens` mode is available for accounts whose callers explicitly accept the simulation-vs-inclusion risk, for example because they reserve their own per-user storage credits inside the contract and only select `DirectTokens` mode when their accounting guarantees that a token is reserved for the specific operation about to execute.
```

tips/tip-1060.md
```diff
@@ -0,0 +1,288 @@
...
+
+- `PreserveTokens` when it wants the user to fully pay the TIP-1000 creation cost.
+- `RefundTokens` when the user may be entitled to reuse deleted storage capacity but the contract cannot pre-confirm at the worst-case gas-limit boundary that a token will be available.
+- `DirectTokens` when the contract's per-user accounting guarantees that a token is available for this specific user at execution time.
+
+Alternatively, an account can leave the mode at the default `RefundTokens` and receive account-level reuse without attributing savings to specific users. In that mode, cheaper storage creation benefits the account as a whole and may not correspond to the user who originally caused the deletion.
+
+TIP-1064 defines the StablecoinDEX-specific owner accounting layer on top of this mechanism.
+
```
dankrad:
```suggestion
```

tips/tip-1060.md
```diff
@@ -0,0 +1,288 @@
...
+## Abuse and Scope
+
+Gas tokens are not transferable and cannot be created without deleting an account's own storage. This preserves TIP-1000's protection against net state growth: an account that increases its live storage footprint beyond its deleted-slot balance still pays the full storage creation cost for each net-new slot.
+
+The mechanism intentionally uses account-level accounting rather than protocol-level user attribution. This keeps the core rule simple and lets contracts that need stricter attribution layer their own per-user accounting on top of the account-level token balance and mode. Contracts that expose arbitrary storage writes to users while in `DirectTokens` mode remain responsible for ensuring their own economics are not exploitable.
+
+`setMode` is unauthenticated: any account can set its own mode at any time, but only its own. A contract that wants `setMode` to be invoked only under specific conditions MUST gate its own calls to the precompile accordingly. Calls from one account that attempt to alter another account's mode are not possible under this interface.
+
```
dankrad:
```suggestion
```

tips/tip-1060.md
```diff
@@ -0,0 +1,288 @@
...
+## Compatibility
+
+Existing contracts are unaffected unless they choose to interact with the Storage Gas Tokens precompile. Accounts that do not interact with the precompile keep the default `RefundTokens` mode and a zero balance until they delete one of their own existing storage slots.
```
dankrad:
```suggestion
Existing contracts don't incur any negative effects unless they choose to interact with the Storage Gas Tokens precompile. Accounts that do not interact with the precompile keep the default `RefundTokens` mode and a zero balance until they delete one of their own existing storage slots.
```

tips/tip-1060.md
```diff
@@ -0,0 +1,267 @@
...
+status: Draft
+related: TIP-1000, TIP-1058, TIP-1064
+protocolVersion: T5
+---
+
+# TIP-1060: Storage Gas Tokens
+
+## Abstract
+
+This TIP introduces *storage gas tokens*: a per-account counter that increments whenever the account deletes one of its own storage slots and that can later offset the TIP-1000 storage creation cost (250,000 gas per zero-to-nonzero SSTORE) when the same account creates a new slot. Each account chooses, via a new precompile, one of three storage creation modes: `RefundTokens` (pay the full TIP-1000 cost upfront; at end-of-transaction, each refund-eligible creation is matched against the account's then-current token balance to grant an equivalent refund), `PreserveTokens` (always pay the full cost, never consume tokens), or `DirectTokens` (when a token is available, consume it synchronously so the TIP-1000 creation cost is not charged at all).
```
danrobinson: Should we say "contract" rather than "account" throughout? I think talking about accounts makes it ambiguous whether it's about the EOAs that create the transactions.

tips/tip-1060.md
```diff
@@ -0,0 +1,267 @@
...
+
+Because `RefundTokens` settles at the end of the transaction, a transaction that creates storage and later deletes its own storage in the same transaction can use the just-minted token to refund the earlier creation. This is the main reason `RefundTokens` is the default: it is at least as cheap as `DirectTokens` in practice and avoids the simulation-vs-inclusion failure mode that synchronous consumption would expose.
```
danrobinson: I think "later deletes that storage (or other storage in the same transaction)"

tips/tip-1060.md
danrobinson: Hmm, I guess that wouldn't quite be accurate, and doesn't make it clear that it applies to precompiles.

tips/tip-1060.md
```diff
@@ -0,0 +1,267 @@
...
+
+# Specification
+
+## Terminology
+
+- **Account**: Any address that can own EVM storage, including ordinary smart contracts, precompiles, and externally owned accounts.
+- **Storage gas token balance**: A protocol-maintained `uint64` counter associated with one account, holding the number of unspent storage gas tokens for that account.
+- **Storage creation**: A storage slot transition from zero to nonzero in storage owned by some account.
+- **Storage deletion**: A storage slot transition from nonzero to zero in storage owned by some account.
+- **TIP-1000 storage creation cost**: The 250,000-gas state-creation component charged by TIP-1000 on every zero-to-nonzero SSTORE, in addition to the EIP-2929 access cost (2,100 gas cold, 100 gas warm) and any other base SSTORE gas. References elsewhere in this TIP to "the TIP-1000 storage creation cost", "the TIP-1000 creation cost", or "the TIP-1000 storage creation component" all denote this same 250,000-gas component. Storage gas tokens offset only this component; the EIP-2929 and base SSTORE gas remain payable in full.
```
danrobinson:
```suggestion
- **TIP-1000 storage creation cost**: The 250,000-gas state-creation component charged by TIP-1000 on every zero-to-nonzero SSTORE, separate from the EIP-2929 access cost (2,100 gas cold, 100 gas warm) and any other base SSTORE gas. References elsewhere in this TIP to "the TIP-1000 storage creation cost", "the TIP-1000 creation cost", or "the TIP-1000 storage creation component" all denote this same 250,000-gas component. Storage gas tokens offset only this component; the EIP-2929 and base SSTORE gas remain payable in full.
```

tips/tip-1060.md
```diff
@@ -0,0 +1,267 @@
...
+
+The synchronous parts (storage write, `DirectTokens`-mode token debit, `RefundTokens`-mode pending counter increment, and any mode changes) MUST be atomic with the storage write that caused them. If the storage write reverts, all of the above MUST revert according to normal EVM revert semantics.
+
+### End-of-Transaction Settlement
+
+After all EVM execution for the transaction has completed (after the top-level call frame returns) and before the EVM refund cap is applied to the transaction's total refund, the protocol MUST settle `RefundTokens`-mode creations for every account `A` with `pending_refund_eligible_creations[A] > 0`:
+
+1. Let `settled := min(pending_refund_eligible_creations[A], gas_token_balance[A])`.
+2. Decrement `gas_token_balance[A]` by `settled`.
+3. Add `settled * 250_000` gas to the transaction's gas refund counter (250,000 gas per redeemed creation, matching the TIP-1000 storage creation cost).
```
0xalpharush: Should we not apply credits to transactions that use less than 1,250,000 gas or when they'd be partially applied? I think this would save more $$$ in aggregate than smaller tx's getting partial refunds while consuming the entire credit

In practice this would look like:

-   300,000 gas used   -> cap 60,000   -> settle 0 credits
-   1,250,000 gas used -> cap 250,000  -> settle 1 credit
-   2,500,000 gas used -> cap 500,000  -> settle 2 credits

tips/tip-1060.md
```diff
@@ -0,0 +1,267 @@
...
+
+In addition to this persistent state, the protocol MUST maintain transaction-local accounting:
+
+```text
+pending_refund_eligible_creations: mapping(address account => uint64)
+```
+
+`pending_refund_eligible_creations` is not persisted across transactions. It is initialized to zero at the start of every transaction, journaled with EVM scopes (so it reverts with the call frame that incremented it, just like gas token balance and mode changes), and consulted only by the end-of-transaction settlement procedure defined below.
+
+## Gas Token Minting
```
0xalpharush: Wondering about the performance and implementation complexity of hooking every SSTORE, checking if it's a delete, and, if so, SLOAD+SSTORE'ing the updated storage gas token balance at the precompile address (seems awkward to write to another address' storage within the same call frame)

Would be curious your thoughts @rakita or @klkvr

tips/tip-1060.md
0xalpharush: This also makes it “at least as cheap as DirectTokens” true. Rn, the cap would make that untrue sometimes

tips/tip-1060.md
```diff
@@ -0,0 +1,267 @@
...
+- **Preserve-token mode**: A storage creation mode in which storage creation always pays the TIP-1000 creation cost and leaves the gas token balance unchanged.
+- **Direct-token mode**: A storage creation mode in which, if a gas token is available, one token is consumed and the TIP-1000 storage creation cost is not charged at all.
+- **Storage Gas Tokens precompile**: The protocol precompile, defined below, through which accounts read their balance and set their storage creation mode.
+
+## State
+
+The protocol MUST maintain the following state:
+
+```text
+storage_gas_token_state: mapping(address account => uint256 packed_state)
```
danrobinson: Where does this state live? I think should live in the precompile?

tips/tip-1060.md
```diff
@@ -0,0 +1,267 @@
...
+
+The mechanism applies uniformly to every account on Tempo, including ordinary smart contracts, precompile-owned state, and EOAs. EOAs do not normally hold storage, so in practice they will neither accumulate nor consume tokens, but the rule is account-keyed for uniformity.
+
+## Motivation
+
+TIP-1000 makes net-new storage creation expensive to protect Tempo from state growth attacks. That cost is appropriate when an account increases its long-term state footprint, but it overcharges accounts that repeatedly delete and recreate storage slots while staying within a previously paid state footprint. Examples include AMM orderbooks that constantly add and cancel orders, channel-style precompiles that open and close escrow slots, and ordinary contracts implementing transient data structures.
```
malleshpai: maybe mention permit's in a txn that use up the entire allowance within the tx's execution?

tips/tip-1060.md
```diff
@@ -0,0 +1,267 @@
...
+- **Storage deletion**: A storage slot transition from nonzero to zero in storage owned by some account.
+- **TIP-1000 storage creation cost**: The 250,000-gas state-creation component charged by TIP-1000 on every zero-to-nonzero SSTORE, in addition to the EIP-2929 access cost (2,100 gas cold, 100 gas warm) and any other base SSTORE gas. References elsewhere in this TIP to "the TIP-1000 storage creation cost", "the TIP-1000 creation cost", or "the TIP-1000 storage creation component" all denote this same 250,000-gas component. Storage gas tokens offset only this component; the EIP-2929 and base SSTORE gas remain payable in full.
+- **Refund-token mode**: The default storage creation mode. Each storage creation pays the TIP-1000 creation cost upfront and increments a transaction-local *pending refund-eligible creation* counter for the storage-owning account. At end-of-transaction, the protocol matches each pending refund-eligible creation against the account's gas token balance (as it stands at end-of-transaction, after all in-transaction deletions and mode changes) and credits a refund of the TIP-1000 creation cost per token consumed.
```
malleshpai: nit: maybe define the three modes together in this list.

tips/tip-1060.md
```diff
@@ -0,0 +1,267 @@
+---
+id: TIP-1060
```
malleshpai, line 2: One question i haven't thought through but just adding here: does this add any forward compatibility issues with 1016, or back us into a corner depending on the final form of EIP 8037?

tips/tip-1060.md
```diff
@@ -0,0 +1,267 @@
...
+
+`pending_refund_eligible_creations[A]` is then discarded with the transaction. The order in which accounts are settled MUST NOT affect any account's outcome, because settlement is per-account and gas tokens cannot move between accounts.
+
+Settlement uses `gas_token_balance[A]` and `pending_refund_eligible_creations[A]` as they stand at end-of-transaction, after all journaling has resolved. Tokens minted by storage deletions earlier in the same transaction are therefore available to fund refunds for earlier `RefundTokens`-mode creations in the same transaction. A transaction that creates storage and later (within the same transaction) deletes one of its own storage slots can receive a refund for the earlier creation against the just-minted token, even if the account's pre-transaction balance was zero.
+
+Refunds granted by settlement are then subject to the standard EVM refund cap together with any other refund accumulated in the transaction. A transaction whose total refund (storage gas token settlement plus any other refunds) exceeds the protocol-wide cap will have its effective refund clamped per existing rules.
```
malleshpai: the 20% evm refund cap would mean that a txn that costs 300k could refund at most 60k--- so for example a transaction that does a permit and then spend it down within the same txn is barely cheaper. 

that barely fixes our expensive gas cost problem for a lot of complexity...

tips/tip-1060.md
```diff
@@ -0,0 +1,267 @@
...
+    function setMode(Mode newMode) external;
+}
+```
+
+### Behavior
+
+- `balance()` and `mode()` are equivalent to `balanceOf(msg.sender)` and `modeOf(msg.sender)`. All four read functions MUST NOT modify state.
+- `balanceOf(account)` returns the `gas_token_balance` field of `storage_gas_token_state[account]`.
+- `modeOf(account)` returns the `storage_creation_mode` field of `storage_gas_token_state[account]`.
+- `setMode(newMode)` MUST revert with `InvalidMode` if `newMode` is the reserved value `3` (or any encoding outside `{0, 1, 2}`). Otherwise it MUST set `storage_creation_mode[msg.sender] := newMode`, emit `ModeUpdated(msg.sender, newMode)`, and consume only the ordinary protocol gas for the underlying state write per the *State* section. `setMode` MUST NOT incur a TIP-1000 creation charge and MUST NOT mint or consume gas tokens.
```
malleshpai: doesn't this effectively open up the same DOS vector that 1000 was supposed to get around? i guess you still have to initialize the address it self to setMode.

tips/tip-1060.md
```diff
@@ -0,0 +1,267 @@
...
+
+The storage gas token accounting state MUST be ordinary protocol state. Creating or updating the storage gas token counter, including the first creation of that counter's backing storage slot, MUST NOT itself mint or consume gas tokens and MUST NOT incur the 250,000 gas TIP-1000-mandated storage creation cost.
+
+In addition to this persistent state, the protocol MUST maintain transaction-local accounting:
+
+```text
+pending_refund_eligible_creations: mapping(address account => uint64)
+```
+
+`pending_refund_eligible_creations` is not persisted across transactions. It is initialized to zero at the start of every transaction, journaled with EVM scopes (so it reverts with the call frame that incremented it, just like gas token balance and mode changes), and consulted only by the end-of-transaction settlement procedure defined below.
```
klkvr: this is effectively transient storage right?

tips/tip-1060.md
```diff
@@ -0,0 +1,267 @@
...
+
+## Gas Token Minting
+
+When account `A` changes one of `A`'s storage slots from nonzero to zero:
+
+1. Apply the normal storage deletion gas behavior.
+2. Increment the `gas_token_balance` field for `A` by one.
```
klkvr: i think it probably worth clarifying edge cases around the 0->x->0 refunds that 8037 introduces? i think we wouldn't want to increment the gas token balance in that case

tips/tip-1060.md
danrobinson: Good point—I think we could implement this as transient storage in the precompile.

tips/tip-1060.md
danrobinson: My impression is that it's a lot harder to add new kinds of state to a contract, so I'm not sure we have another option for it.

tips/tip-1060.md
danrobinson: Wouldn't a 0->x->0 sequence with this TIP active result in the gas token balance getting incremented and then decremented, so the user will get a refund even without 8037?

Agree that we have to figure out if 8037 will have weird interactions with it

tips/tip-1060.md
malleshpai: minor: is the intent to refund 250k full or only the 230k state gas from tip 1000?

tips/tip-1060.md
```diff
@@ -0,0 +1,267 @@
...
+
+Deleting a slot mints exactly one gas token regardless of the deleted value, the slot key, the transaction sender, or the call path that caused the deletion. The token is credited to the account whose storage was modified. Under `DELEGATECALL` and `CALLCODE`, this is the calling account (the account whose storage the SSTORE actually wrote), per normal EVM semantics.
```
danrobinson:
```suggestion
Deleting a slot mints exactly one gas token regardless of the deleted value, the slot key, the transaction sender, or the call path that caused the deletion. The token is credited to the account (generally, the contract or precompile) whose storage was modified. Under `DELEGATECALL` and `CALLCODE`, this is the calling account (the account whose storage the SSTORE actually wrote), per normal EVM semantics.
```

tips/tip-1060.md
```diff
@@ -0,0 +1,267 @@
...
+If `storage_creation_mode == 3`, clients MUST treat the state as invalid before any future hardfork assigns semantics to that value.
+
+Reserved bits MUST be written as zero by this TIP. Future TIPs MAY assign reserved bits, but clients MUST NOT infer semantics from nonzero reserved bits unless those semantics have been activated by a hardfork.
+
+The storage gas token accounting state MUST be ordinary protocol state. Creating or updating the storage gas token counter, including the first creation of that counter's backing storage slot, MUST NOT itself mint or consume gas tokens and MUST NOT incur the 250,000 gas TIP-1000-mandated storage creation cost.
+
+In addition to this persistent state, the protocol MUST maintain transaction-local accounting:
+
+```text
+pending_refund_eligible_creations: mapping(address account => uint64)
```
malleshpai: maybe an implementation detail but at the end of the txn we also need a list of what addresses were touched, can't recover that from just the mapping.

tips/tip-1060.md
rakita: @0xrusowsky and me talked about this, you still need to do revert/halt on call frame level, so this mapping needs to live in Journal and it needs to create a journal entry on insertion.

tips/tip-1060.md
dankrad: Discussed today with @rakita, most likely manageable but will prototype once we have 1060 implementatino

tips/tip-1060.md
dankrad:
```suggestion
Because `RefundTokens` settles at the end of the transaction, a transaction that creates storage and later later deletes that storage (or other storage in the same contract) in the same transaction can use the just-minted token to refund the earlier creation. This is the main reason `RefundTokens` is the default: it is at least as cheap as `DirectTokens` in practice and avoids the simulation-vs-inclusion failure mode that synchronous consumption would expose.
```

tips/tip-1060.md
dankrad:
```suggestion
TIP-1000 makes net-new storage creation expensive to protect Tempo from state growth attacks. That cost is appropriate when an account increases its long-term state footprint, but it overcharges accounts that repeatedly delete and recreate storage slots while staying within a previously paid state footprint. Examples include AMM orderbooks that constantly add and cancel orders, channel-style precompiles that open and close escrow slots, using permit in a transaction that uses up the entire allowance, and ordinary contracts implementing transient data structures.
```

tips/tip-1060.md
dankrad: This state lives in the precompile at `STORAGE_GAS_TOKENS_ADDRESS`

tips/tip-1060.md
dankrad: Yeah I think contract creation costs will be too high for this to become a DOS vector

tips/tip-1060.md
rakita: Since the counter is persistent and needs to be loaded, we can just use storage to update it directly.

Using transient storage would probably be worse here.

tips/tip-1060.md
rakita: It is not perfect, but it is not that bad. We can isolate the cases of  x -> 0 and 0 -> x, and only when this happens do additional work of updating the loading and writing to an additional slot.

This becomes a lot easier because it is in precompile and not hooked inside evm instruction.

We should add additional gas cost, just to be in line of other work. For ref SSTORE update is 2800gas for work and 100/2200 (for warm/cold access)

0 -> x: so if the counter has some value, this is discounted by a lot. In any case, we should price in additional SSTORE, from the formula above.

x -> 0: It is priced the same as a write ( 2800gas +100/2200 (warn/cold access)). So additional gas is going to double it. Imo it is fine to double it and remove regular gas refund.

tips/tip-1060.md
```diff
@@ -0,0 +1,278 @@
+---
+id: TIP-1060
```
malleshpai, line 2: todo: if we run with this tip as is, we need to make sure that in worse case this is safe with the combo of (500M gas limit, ~20k  gas for SStore when there's lots of storage tokens available )

tips/tip-1060.md
```diff
@@ -0,0 +1,278 @@
...
+
+The default mode is intentionally refund-based rather than direct consumption: a caller must always provide enough gas for the full storage creation cost, so transactions do not unexpectedly fail when the gas token balance changes between simulation and inclusion.
+
+## Compatibility
+
+Existing contracts don't incur any negative effects unless they choose to interact with the Storage Gas Tokens precompile, and keep the default `RefundTokens` mode and a zero balance until they delete one of their own existing storage slots. Two protocol-wide changes do, however, alter storage gas economics for every account: the pre-existing storage-clearing gas refund (nonzero-to-zero) is removed and replaced by storage gas token minting, and the standard EVM refund cap is removed. A contract that previously received the immediate clearing refund now instead earns a storage gas token on deletion and redeems it (in the default `RefundTokens` mode) for a 230,000-gas refund at end-of-transaction settlement when it later creates a slot; a deletion that is not matched by a later creation in the same account yields a carried-forward token rather than an immediate refund.
```
malleshpai: the first sentence and the last sentence are slightly contradictory. previously if you cleared a slot you got some refund, the fact that you don't get it and get a gas token instead is somewhat a negative effect, no?

tips/tip-1060.md
```diff
@@ -0,0 +1,278 @@
...
+
+- `balance()` and `mode()` are equivalent to `balanceOf(msg.sender)` and `modeOf(msg.sender)`. All four read functions MUST NOT modify state.
+- `balanceOf(account)` returns the `gas_token_balance` field of `storage_gas_token_state[account]`.
+- `modeOf(account)` returns the `storage_creation_mode` field of `storage_gas_token_state[account]`.
+- `setMode(newMode)` MUST revert with `InvalidMode` if `newMode` is the reserved value `3` (or any encoding outside `{0, 1, 2}`). Otherwise it MUST set `storage_creation_mode[msg.sender] := newMode`, emit `ModeUpdated(msg.sender, newMode)`, and consume only the ordinary protocol gas for the underlying state write per the *State* section. `setMode` MUST NOT incur a TIP-1000 creation charge and MUST NOT mint or consume gas tokens.
+- Under `DELEGATECALL` or `CALLCODE`, `msg.sender` is propagated per normal EVM semantics, so the account whose mode is written is the same account whose storage gas tokens would be earned and consumed by subsequent SSTOREs in the same scope.
+- A mode write performed inside an execution scope that later reverts MUST itself revert, restoring the prior `storage_creation_mode` value for the same account.
+
```
malleshpai, line 334: do we need to specify explicitly how precompiles would call this? or a separate abi?

tips/tip-1060.md
danrobinson: @rakita This is just the storage for the current transaction to track how much needs to be refunded at the end of the tx; I don't think it's persisted?

tips/tip-1060.md
danrobinson: > This becomes a lot easier because it is in precompile and not hooked inside evm instruction.

Well, it is hooked inside SSTORE, isn't it?

tips/tip-1060.md
rakita: There are three modes, hence my confusion.

`DirectTokens` directly changes the storage of gas token contract.

`RefundTokens` does it at the end of transaction, and we can use transient storage to keep a track of this.

tips/tip-1060.md
rakita: It is, didn't know this is effecting every contract.

But even if it is inside SSTORE it is okay, as frame level handling is going to be done by using storage and transient storage.

Imo we should add additional gas for warm/cold loading of additional state and this will be fine.

tips/tip-1060.md
0xrusowsky: nice, question.

when doing the impl for the DEX i realized that while this API works, it adds perf overhead (and thus also gas costs) due to the multiple SLOAD/SSTOREs that it requires.

in practice, @rakita and i decided to add an internal method that allows us to bypass the storage layer for precompiles.

idk if worth mentioning it here.

tips/tip-1060.md
0xrusowsky: agreed, we should add some numbers in the threat model section

tips/tip-1060.md
0xrusowsky: the 230k state gas, althuogh subject to fine-tuning (could be even lowered)

tips/tip-1060.md
dankrad: We are not going to cap refunds under this TIP

tips/tip-1060.md
dankrad: 20K is still more expensive than a write, so we should test this but really it shouldn't be a problem

tips/tip-1060.md
dankrad: I think this comment is a misunderstanding on how credits work. First the refunds aren't capped, and second each credit is worth exactly one storage slot. A single credit lets you save 230,000 gas for creating new storage.

tips/tip-1060.md
dankrad: Done — moved `Pending refund-eligible creation` below the mode definitions so `Refund`, `Preserve`, and `Direct` are now listed together (e461b9248).

tips/tip-1060.md
dankrad: Good catch — the first sentence overclaimed. Reworded the Compatibility paragraph to drop "no negative effects" and flag the actual downside: a deletion that isn't matched by a later creation loses the small immediate clearing refund and instead holds a carried-forward credit that only pays off (230k) on a future storage creation. For delete-then-create flows it stays neutral-to-favorable, and removing the refund cap is favorable in all cases. (074ee7a4e)

dankrad: True, flagging the changes in refunds is a good addition.

tips/tip-1060.md
dankrad: Implementation detail -- the mapping is not visible from EVM so the notation was just for convenience.

tips/tip-1060.md
dankrad: Since TIP-1060 is rolled out before 1016, we will clarify edge cases at the time of 1016/EIP-8037 rollout. Makes no sense to do this ahead as details can still change

tips/tip-1060.md
```diff
@@ -0,0 +1,277 @@
...
+This exemption is not a discount for arbitrary user storage. It waives only the extra 250,000-gas storage-creation component for the protocol metadata slot that tracks an account's storage-credit state. For contracts, that metadata slot is coupled to an account whose creation already pays the normal account or contract creation overhead; charging another 250,000 gas for the bookkeeping slot would double-count the first per-account footprint. Since this path cannot create arbitrary contract-owned storage slots and mass creation of eligible accounts remains bounded by ordinary account and contract creation gas, the exemption is not a state-growth DoS vector.
```
samczsun: this is incorrect - when combined with expiring nonce transactions, a user can sign an expiring nonce transaction from a fresh account and a centralized fee payer which calls `setMode` to write one slot to storage for 48253 gas, or around 9.6M/TB

tips/tip-1060.md
jenpaff: bump- add this to the threat model section

legion2002:
cyclops audit

decofe:
cc @legion2002

Cyclops audit event published. [View workflow run](https://github.com/tempoxyz/tempo/actions/runs/27221499904)

tips/tip-1060.md
dankrad: Fixed now by only allowing contracts to call `setMode`

tips/tip-1060.md
```diff
@@ -0,0 +1,293 @@
...
+
+Deleting a slot mints exactly one storage credit regardless of the deleted value, the slot key, the transaction sender, or the call path that caused the deletion. The credit is credited to the account (generally, the contract or precompile) whose storage was modified. Under `DELEGATECALL` and `CALLCODE`, this is the calling account (the account whose storage the SSTORE actually wrote), per normal EVM semantics.
```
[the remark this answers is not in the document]
legion2002: we are already doing this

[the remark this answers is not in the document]
dankrad: This is intended behaviour #WONTFIX

tips/tip-1060.md
```diff
@@ -0,0 +1,293 @@
...
+
+```text
+pending_refund_eligible_creations: mapping(address account => uint64)
+```
+
+`pending_refund_eligible_creations` is not persisted across transactions. It is initialized to zero at the start of every transaction, journaled with EVM scopes (so it reverts with the call frame that incremented it, just like storage credit balance and mode changes), and consulted only by the end-of-transaction settlement procedure defined below.
+
+## Storage Credit Minting
+
+This TIP removes the pre-existing storage-clearing gas refund. Before this TIP, setting a nonzero storage slot to zero credited a gas refund to the transaction at end-of-transaction; this was the only storage-related gas refund on Tempo. That refund is removed and replaced by the storage credit minted on deletion: an account that clears a slot no longer receives an immediate refund but instead earns a credit it can apply to a future (or same-transaction) storage creation via the modes below. After this TIP, the end-of-transaction settlement defined below is the sole mechanism by which storage operations affect the transaction's gas refund counter.
```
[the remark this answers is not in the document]
dankrad: This finding does not apply as TIP-1016 is not activated on Tempo. We will apply appropriate fixes when EIP-8037/TIP-1016 are ready for Tempo activation.

tips/tip-1060.md
```diff
@@ -0,0 +1,293 @@
...
+
+## Storage Credit Consumption
+
+When account `A` changes one of `A`'s storage slots from zero to nonzero, the protocol selects exactly one of the following branches based on `storage_creation_mode[A]`:
+
+1. `storage_creation_mode[A] == Refund`:
+   - Charge the full TIP-1000-mandated storage creation cost.
+   - Increment `pending_refund_eligible_creations[A]` by one (transaction-local, journaled with the current EVM scope).
+   - Do NOT decrement `storage_credit_balance[A]` and do NOT credit a gas refund at this point. Credit consumption and the resulting refund happen during end-of-transaction settlement.
+2. `storage_creation_mode[A] == Preserve`:
```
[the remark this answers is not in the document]
dankrad: This finding is not a problem as the transaction would pay the fair amount at the 0->X transition. That the deletion occurs in the same transaction does not change the analysis.

tips/tip-1060.md
```diff
@@ -0,0 +1,293 @@
...
+- `setMode(newMode)` MUST revert with `InvalidMode` if `newMode` is the reserved value `3` (or any encoding outside `{0, 1, 2}`). Otherwise, for a permitted caller, it MUST set `storage_creation_mode[msg.sender] := newMode`, emit `ModeUpdated(msg.sender, newMode)`, and consume only the ordinary protocol gas for the underlying state write per the *State* section. The backing write MUST NOT incur a TIP-1000 storage creation charge and MUST NOT mint storage credits, consume storage credits, or increment `pending_refund_eligible_creations`, including when it creates the first nonzero backing slot for `msg.sender`.
+- Under `DELEGATECALL` or `CALLCODE`, `msg.sender` is propagated per normal EVM semantics, so the account whose mode is written is the same account whose storage credits would be earned and consumed by subsequent SSTOREs in the same scope.
```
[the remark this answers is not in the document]
dankrad: Fixed by requiring direct calls only. The precompile now reverts with `OnlyDirectCall` whenever it is reached via `DELEGATECALL`/`CALLCODE` (detected when the executing-context address isn't `STORAGE_CREDITS_ADDRESS`, matching the existing Tempo precompile guards), so `msg.sender` is always the actual calling account and a mode change is always keyed to the storage target. Removed the incorrect "msg.sender is propagated under DELEGATECALL" guidance and added invariant 18. Library code running as a DELEGATECALL callee can still make a normal direct CALL to the precompile, since that keys to the executing account. (18126e9fd)

tips/tip-1060.md
```diff
@@ -0,0 +1,295 @@
...
+
+The `storage_creation_mode` value `3` is reserved; clients MUST treat it as invalid until a future hardfork assigns it semantics, and `setMode` MUST reject it (see the precompile behavior).
+
+## Storage Credit Minting
+
+This TIP removes the pre-existing storage-clearing gas refund. Before this TIP, setting a nonzero storage slot to zero credited a gas refund to the transaction at end-of-transaction; this was the only storage-related gas refund on Tempo. That refund is removed and replaced by the storage credit minted on deletion: an account that clears a slot no longer receives an immediate refund but instead earns a credit it can apply to a future (or same-transaction) storage creation via the modes below. After this TIP, the end-of-transaction settlement defined below is the sole mechanism by which storage operations affect the transaction's gas refund counter.
+
+When account `A` changes one of `A`'s storage slots from nonzero to zero:
```
samczsun: Assume a slot is nonzero. Set mode to direct and clear the slot. Then, in a loop, set it to a new nonzero value that is different from the original value, then clear it again. Each loop will mint a new storage credit due to the clear, but the subsequent write to nonzero will only be charged 100 gas. This can then be used to offset zero-to-nonzero writes on new slots, resulting in a net cost of ~17k gas/slot, or ~4M/TB

tips/tip-1060.md
```diff
@@ -0,0 +1,324 @@
...
+
+1. Apply the normal storage deletion gas behavior, except that the pre-existing nonzero-to-zero storage-clearing gas refund is not credited (it is removed by this TIP, as described above).
+2. Increment the `storage_credit_balance` field for `A` by one.
```
0xKitsune: +1 on @samczsun's [comment](https://github.com/tempoxyz/tempo/pull/4016#discussion_r3390120669), can we specify how dirty slots are handled? (both x->0->x and 0->x->0).

tips/tip-1060.md
```diff
@@ -0,0 +1,324 @@
...
+
+Reserved bits MUST be written as zero by this TIP. Future TIPs MAY assign reserved bits, but clients MUST NOT infer semantics from nonzero reserved bits unless those semantics have been activated by a hardfork.
+
+Each account's `storage_credit_state` word is the only persistent slot this TIP introduces, and it is first written when the account mints its first storage credit, i.e., when it deletes one of its own nonzero storage slots. Writes whose storage owner is `STORAGE_CREDITS_ADDRESS`, including the first nonzero write that creates an account's `storage_credit_state` slot, MUST be journaled as ordinary protocol state but MUST NOT incur the 250,000-gas TIP-1000-mandated storage creation cost. Implementations MUST only charge ordinary protocol gas for the backing state write, such as the non-creating update/reset cost when a value changes. These writes also MUST NOT be treated as TIP-1060 storage-credit events: they MUST NOT themselves mint or consume storage credits and MUST NOT increment `pending_refund_eligible_creations`.
```
0xKitsune: Consider specifying that protocol/system writes dont participate in credit accounting. Atm only the storage credits precompile's own slot is exempt while credits apply to precompile owned state, however it seems like fee calls pre/post tx should also be exempt. This might be implied since these already do not consume gas but consider specifying.

tips/tip-1060.md
```diff
@@ -0,0 +1,324 @@
...
+- `setMode(newMode)` MUST revert with `InvalidMode` if `newMode` is the reserved value `3` (or any encoding outside `{0, 1, 2}`). Otherwise it MUST set the transient mode for `msg.sender` to `newMode`, set the transient Direct budget for `msg.sender` to `type(uint64).max` if `newMode == Direct` and zero otherwise, and emit `ModeUpdated(msg.sender, newMode)`. The writes target transient storage only: they create no persistent state, cost only ordinary transient-storage writes (EIP-1153 `TSTORE`-equivalent), MUST NOT incur a TIP-1000 storage creation charge, and MUST NOT mint storage credits, consume storage credits, or increment `pending_refund_eligible_creations`.
+- `setBudget(creditBudget)` MUST set the transient mode for `msg.sender` to `Direct`, set the transient Direct budget for `msg.sender` to `creditBudget`, and emit `ModeUpdated(msg.sender, Direct)`. It switches to `Direct` mode and caps the number of credits that may be spent directly by subsequent storage creations in the current transaction. If `creditBudget == 0`, the next storage creation follows the zero-budget branch, switches the account to `Preserve`, and pays the full TIP-1000 storage creation cost.
```
0xKitsune: Consider reverting on a zero budget or setting `Preserve` directly so the reported mode matches behavior. Right now `setBudget(0)` sets mode to `Direct` with a zero budget, so `modeOf` reports `Direct` even though the next creation immediately falls through to `Preserve`.

tips/tip-1060.md
```diff
@@ -0,0 +1,324 @@
...
+    ///         direct-spend budget.
+    /// @dev Sets the calling account's mode to `Direct` and initializes its
+    ///      remaining direct-spend budget to `creditBudget`.
+    /// @param creditBudget Maximum number of credits to spend directly after
+    ///                     this call in the current transaction.
+    function setBudget(uint64 creditBudget) external;
+}
+```
+
+### Behavior
```
0xKitsune, line 322: nit: seems like this section could fold into another section above when mentioning modes and budgets?

tips/tip-1060.md
```diff
@@ -0,0 +1,324 @@
...
+- A contract that itself runs as the callee of a `DELEGATECALL` (i.e. library code executing in another account's context) may still call the precompile, because that is a direct `CALL` to the precompile from within that context: `msg.sender` is then the account whose code is executing and whose storage the mode governs. Only `DELEGATECALL`/`CALLCODE` *into the precompile itself* is rejected.
+- A mode or budget write performed inside an execution scope that later reverts MUST itself revert, restoring the prior transient `storage_creation_mode` and `storage_credit_budget` values for the same account.
+
+## Scope
+
+This TIP applies to every account on Tempo. Storage credit minting, consumption, and mode selection follow the same rules whether the account is an ordinary smart contract, a precompile, or an EOA; any account may call `setMode` for itself. The effective SSTORE cost for accounts that never delete and re-create storage slots is unchanged, because their `storage_credit_balance` stays at zero and `Refund` mode degenerates to charging the full TIP-1000 cost with no refund.
+
+## Default Behavior
```
0xKitsune: nit: ame thing here, seems like we can combine the `Behavior` and `Default Behavior` sections into a single "Modes" section.

tips/tip-1060.md
```diff
@@ -0,0 +1,324 @@
...
+
+Traces and debug tooling SHOULD expose credit minting, synchronous `Direct` credit consumption, `Refund` pending counter increments, Direct budget reads and writes, end-of-transaction settlement (per-account credits consumed and refund credited), and storage creation mode reads or writes so developers can explain gas differences. Smart contracts that perform storage churn may observe lower effective costs after this TIP, including for transactions that net to zero new storage. SSTORE semantics for accounts that do not churn storage remain unchanged.
+
+---
+
+# Invariants
```
0xKitsune, line 355: Similar to previous TIPs, there are some of these that describe general constraints/behaviors rather than invariants. Can we do a pass on these to iron out which cases should be properly evaluated with invariant tests and which are general behavioral constraints?

0xKitsune (dismissed):
Following internal discussions, lgtm pending resolution of remaining comments.

tips/tip-1060.md
0xrusowsky: update, since we moved to transient storage, we will simply use the same ABI as any other contract

samczsun (dismissed):

0xrusowsky (dismissed):

0xalpharush (dismissed):
already approved https://github.com/tempoxyz/tempo/pull/4016/changes/e01e9b88fe42e64e8ff46e8acd5aece4d44b14ee

horsefacts (dismissed):

tips/tip-1060.md
malleshpai: there is still no threat model section, which is fine in itself but my question remains, especially now that we've cut it to 5k --- is that "sufficient" gas to be safe? cc @jenpaff

tips/tip-1060.md
```diff
@@ -0,0 +1,362 @@
...
+
+Settlement uses `storage_credit_balance[A]` and `pending_refund_eligible_creations[A]` as they stand at end-of-transaction, after all journaling has resolved. Credits minted by storage deletions earlier in the same transaction are therefore available to fund refunds for earlier `Refund`-mode creations in the same transaction. A transaction that creates storage and later (within the same transaction) deletes one of its own storage slots can receive a refund for the earlier creation against the just-minted credit, even if the account's pre-transaction balance was zero.
+
+The standard EVM refund cap (the limit of one fifth, i.e. 20%, of the transaction's gas used per EIP-3529) is removed, so no cap applies to the transaction's gas refund counter. As refund for clearing the storage is removed only refund left are for actions done in same transaction, this allow us to remove cap fully. The full `settled * STORAGE_CREDIT_VALUE` gas is credited regardless of the transaction's total gas used, and the preserved baseline SSTORE refunds (such as the restore-to-original refunds) are likewise credited in full.
```
danrobinson:
```suggestion
The standard EVM refund cap (the limit of one fifth, i.e. 20%, of the transaction's gas used per EIP-3529) is removed, so no cap applies to the transaction's gas refund counter. This is permissible because TIP-1060 removes the refund for storage cleared in a transaction (which was potentially subject to abuse), while adding a new type of refund for a user who paid full price for storage, when the contract was in REFUND mode and had an available gas credit that was consumed. This latter type of refund is not subject to abuse, because it only refunds gas that was paid for in the same transaction.```

File: tips/tip-1060.md [hunk 44]
```diff
@@ -0,0 +1,362 @@
...
+
+# Specification
+
+## Terminology
+
+- **Account**: Any address that can own EVM storage, including ordinary smart contracts, precompiles, and externally owned accounts.
+- **Storage credit balance**: A protocol-maintained `uint64` counter associated with one account, holding the number of unspent storage credits for that account.
+- **Storage creation**: A storage slot transition from zero to nonzero in storage owned by some account.
+- **Storage deletion**: A storage slot transition from nonzero to zero in storage owned by some account.
+- **TIP-1000 storage creation cost**: The state-creation component charged by TIP-1000 (`SSTORE_CREATE_COST = 250,000` gas), separate from the EIP-2929 access cost (2,100 gas cold, 100 gas warm) and any other base SSTORE gas. This TIP splits that component into two portions with different charging rules:
```
danrobinson, line 43: I think TIP-1000 defined the storage creation gas as 250,100 gas https://github.com/tempoxyz/tempo/blob/f71854c91b2d6c6af425e3817089a626dd0bb894/tips/tip-1000.md#new-state-element-creation

tips/tip-1060.md
```diff
@@ -0,0 +1,362 @@
...
+    ///         storage the mode and credits govern.
+    error DelegateCallNotAllowed();
+
+    /// @notice Emitted when an account's storage creation mode changes.
+    /// @param account The account whose mode changed: the direct caller
+    ///                (`msg.sender`) of `setMode`. Because the precompile
+    ///                rejects `DELEGATECALL`/`CALLCODE`, this is always the
+    ///                account that owns the storage the mode governs.
+    /// @param newMode The new storage creation mode.
+    event ModeUpdated(address indexed account, Mode newMode);
```
danrobinson: Do we really need a ModeUpdated event? Seems really expensive for something that might be called all the time

tips/tip-1060.md
0xrusowsky: iiuic TIP-1000 distinguishes the storage creation component (250k) from the access cost (warm: 100, cold: 2100), which is what we explain here.

imo it is fine as is

tips/tip-1060.md
0xrusowsky: dropped here [`33344f2` (this PR)](https://github.com/tempoxyz/tempo/pull/4016/commits/33344f2e14bf3c290657a46a455e895745f1cbee)

tips/tip-1060.md
0xrusowsky: addressed in https://github.com/tempoxyz/tempo/pull/6289

0xrusowsky (approved):

legion2002 (approved):

PR #4022, opened 2026-05-18
feat(tx): support EIP-712 Tempo signatures

figtracer:
## Summary
- add a browser-signable EIP-712 sender signing mode for Tempo transactions
- encode EIP-712 secp256k1 sender signatures as primitive signature type 0x05
- recover EIP-712 signatures against the typed-data digest and account for them as secp256k1 gas

The EIP-712 message commits to the existing Tempo transaction signing hash via `TempoTransaction(bytes32 signingHash)`, preserving the current transaction-field encoding while allowing browser wallets to use `eth_signTypedData_v4`.


0xalpharush:
think this will require a hardfork

0xKarl98:
Found that  new `Eip712Secp256k1` primitive is parsed through the shared `PrimitiveSignature` path, so other flows can accept it and still verify against a plain `sig_hash`. Only `AASigned::recover_signer()` switches to `eip712_signature_hash()`, which makes this signature type context-dependent.

Think it's better to keep this variant sender-only, or make the parse/recovery context explicit, so other consumers don't silently accept it with the wrong digest.

PR #4033, opened 2026-05-18
feat(primitives): use signature marker for sponsor requests

0xrusowsky:
this PR adds an explicit fee-payer service encoding path for sponsored Tempo transactions.

introduces a `0x00` service marker for the fee-payer signature field so clients can submit transactions that need sponsorship, while keeping canonical transaction RLP encoding unchanged.

this approach avoids changing tx hash behavior for transactions with an actual zero-valued fee-payer signature and keeps the sponsorship-specific encoding isolated to `encode_for_fee_payer_service`.

### DIFF
--- a/crates/primitives/src/transaction/tempo_transaction.rs
+++ b/crates/primitives/src/transaction/tempo_transaction.rs
@@ -20,9 +20,6 @@ pub const TEMPO_TX_TYPE_ID: u8 = 0x76;
 /// Magic byte for the fee payer signature
 pub const FEE_PAYER_SIGNATURE_MAGIC_BYTE: u8 = 0x78;
 
-/// In-memory sentinel for transactions that need a sponsor signature.
-pub const FEE_PAYER_SIGNATURE_MARKER: Signature = Signature::new(U256::ZERO, U256::ZERO, false);
-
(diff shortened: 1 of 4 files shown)


crates/primitives/src/transaction/tempo_transaction.rs
```diff
@@ -513,6 +519,29 @@ impl TempoTransaction {
...
+    ///
+    /// Fee-payer services accept an unsigned sponsorship request whose fee-payer signature field is
+    /// the single byte `0x00`. This is not the canonical transaction RLP encoding for a zero-valued
+    /// secp256k1 signature; it is a service-level marker that tells the sponsor where to insert the
+    /// real fee-payer signature. Keeping this path separate avoids changing canonical transaction
+    /// encoding and therefore avoids changing transaction hashes for transactions that contain an
+    /// actual all-zero fee-payer signature (which are invalid, but still have a canonical encoding).
+    ///
+    /// The fee token is omitted from this request encoding so the user does not commit to a specific
+    /// sponsorship token before the fee payer signs.
```
0xrusowsky:
```suggestion
```

klkvr (approved):

PR #4040, opened 2026-05-18
Fix replay bench cleanup and chart regression

decofe:
Fixes two hourly replay bench failure modes observed on May 18, 2026:\n\n- Guard replay cleanup so an unset tail process does not trip `set -u` after the benchmark exits.\n- Make chart regression skip non-finite or degenerate input and tolerate `np.polyfit` linear algebra failures.\n\nVerification:\n- `bash -n .github/scripts/bench-tempo-replay.sh`\n- `uv run --with numpy --with matplotlib python - <<'PY' ...` to exercise degenerate and valid regression inputs.

### DIFF
--- a/.github/scripts/bench-replay-charts.py
+++ b/.github/scripts/bench-replay-charts.py
@@ -156,8 +156,15 @@ def _add_regression(ax, x, y, color, label):
     """Add a linear regression line to the axes."""
     if len(x) < 2:
         return
-    xa, ya = np.array(x), np.array(y)
-    m, b = np.polyfit(xa, ya, 1)
+    xa, ya = np.array(x, dtype=float), np.array(y, dtype=float)
+    finite = np.isfinite(xa) & np.isfinite(ya)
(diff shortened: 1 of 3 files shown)


decofe:
Added the actual CLI compatibility fix from the bench log:  now probes  and only passes  when the checked-out txgen bench binary supports it. Older binaries will log that the flag was skipped instead of failing with .\n\nNote: I could not run a local Nushell parse check in the sandbox because  is not installed here.

decofe:
Correction on previous comment: added the actual CLI compatibility fix from the bench log. contrib/bench/txgen/helpers.nu now probes bench send --help and only passes --metrics-align when the checked-out txgen bench binary supports it. Older binaries will log that the flag was skipped instead of failing with unexpected argument --metrics-align.

Note: I could not run a local Nushell parse check in the sandbox because nu is not installed here.

decofe:
Closing per request in Slack. The proposed fix path was not the right one.

PR #4044, opened 2026-05-19
Fix ambiguous Tempo cargo update specs

decofe:
## Summary
- qualify stale Tempo git package specs with their Cargo source ID before running cargo update
- avoids ambiguity when Cargo.lock contains both crates.io and git entries for the same Tempo crate/version

## Validation
- bash -n scripts/foundry-patch.sh
- minimal Cargo.lock awk fixture emits git+https://github.com/tempoxyz/tempo?...#name@version specs


grandizzy (approved):

0xalpharush:
seems like it is working now

PR #4045, opened 2026-05-19
perf(tip-1028): pack `RecoveryMode` enum

0xrusowsky:
this PR effectively turns recovery authority from an `Address` with sentinel values —which cannot be packed, thus always requiring an extra SLOAD—, into an `enum RecoveryMode` + an optional address.

NOTE: our impl is equivalent to the following enum, but due to solidity's poor type system, it is not supported (yet)
```rs 
enum RecoveryAuthority {
   Originator,
   Receiver,
   ThirdParty(Address)
}
```


0xrusowsky:
closing in favor of https://github.com/tempoxyz/tempo/pull/4057

PR #4057, opened 2026-05-19
test(tip-1028): gas snapshots

0xalpharush:
Added gas snapshots, cherry-picked https://github.com/tempoxyz/tempo/pull/4045/changes/dd0cd03e69abd8237c3b8b75a2c7aa4031dd336f from https://github.com/tempoxyz/tempo/pull/4045, updated gas snapshots


0xalpharush:
If we go this route, we may want to clear the receiver address upon changing from third-party to receiver/originator. Otherwise, it will be left stale rn

crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -379,12 +393,19 @@ impl TIP403Registry {
...
-        })?;
+            recovery_mode: recovery_address.into(),
+        };
+        let recovery_mode = config.recovery_mode;
+        self.receive_policies[msg_sender].config.write(config)?;
+        if matches!(recovery_mode, RecoveryMode::ThirdParty) {
+            self.receive_policies[msg_sender]
+                .recovery_address
+                .write(recovery_address)?;
+        }
```
0xrusowsky: imo let's always write `Address::ZERO` in the other modes, to prevent stale values.

the savings come from saving an SLOAD in the hot-path. Changing config won't be nearly as common as receiving transfers.

0xrusowsky: addressed in [`25fe575` (this PR)](https://github.com/tempoxyz/tempo/pull/4057/commits/25fe57507e62240f44422351b98e16eac549e1f5)

0xrusowsky (approved):

crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -247,14 +256,14 @@ impl TIP403Registry {
...
-        let config = self.receive_policy_config[account].read()?;
+        let config = self.receive_policies[account].config.read()?;
         Ok(ITIP403Registry::receivePolicyReturn {
             hasReceivePolicy: config.has_receive_policy,
             senderPolicyId: config.sender_policy_id,
             senderPolicyType: self.receive_policy_type(config.sender_policy_id)?,
             tokenFilterId: config.token_filter_id,
             tokenFilterType: self.receive_policy_type(config.token_filter_id)?,
-            recoveryAuthority: config.recovery_address,
+            recoveryAuthority: self.receive_policy_recovery(account, config.recovery_mode)?,
```
0xalpharush, line 266: Is only having the sentinel address here going to limit any integrations?

crates/precompiles/src/tip403_registry/mod.rs
0xalpharush: LGTM

crates/precompiles/src/tip403_registry/mod.rs
0xrusowsky: i think it's fine as is? i guess we could replace `address(1)` for `account`.

is that what u are proposing?

0xalpharush: fine to leave as is. i was asking if the resolved `recoveryAuthority` is needed

PR #4062, opened 2026-05-19
feat: follower fast-sync via the network identity

hamdiallam:
Reworks `process_event`:

If the cert is in the future
1. Still hint the boundary for the current epoch
2. Register the network identity (if from_epoch is valid)

Instead of shortcutting early, we try to validate the cert. IF valid:
1. Jump directly to this finalization
2. Set the floor since we dont need to gap repair anything in the CL. We should get a marshal update immediately
3. On `process_update` ensure we update `current_epoch` accordingly

## Notes
We weren't validating certs in the follower. The marshal doesn't validate when calling `report` as that is an internal trusted api it uses when getting certs over p2p. Since we are using it directly, we need to validate the cert ourselves

### DIFF
--- a/crates/commonware-node/src/follow/driver.rs
+++ b/crates/commonware-node/src/follow/driver.rs
@@ -28,7 +28,7 @@ use commonware_runtime::{Clock, ContextCell, Spawner, spawn_cell};
 use commonware_utils::{Acknowledgement, vec::NonEmptyVec};
 use rand_08::{CryptoRng, Rng};
 
-use eyre::{OptionExt as _, Report, WrapErr as _, bail};
+use eyre::{Context, OptionExt as _, Report, bail, ensure};
 use reth_node_core::primitives::SealedBlock;
 use reth_provider::HeaderProvider as _;
 use tempo_chainspec::NetworkIdentity;
@@ -95,19 +95,6 @@ pub(super) fn try_init<TContext>(
     )
     .wrap_err("the genesis header did not contain a DKG outcome")?;
 
-    let network_identity = &config.network_identity;
-    if onchain_outcome.epoch.get() >= network_identity.from_epoch
-        && network_identity.identity != *onchain_outcome.network_identity()
-    {
-        warn!(
-            compiled_from_epoch = network_identity.from_epoch,
-            onchain_epoch = %onchain_outcome.epoch,
-            compiled_network_identity = %network_identity.identity,
-            onchain_network_identity = %onchain_outcome.network_identity(),
-            "Network identity differs from the on-chain DKG outcome!!! update the binary with the latest network identity"
-        );
-    }
-
     config.scheme_provider.register(
         onchain_outcome.epoch,
         Scheme::certificate_verifier(
@@ -116,12 +103,18 @@ pub(super) fn try_init<TContext>(
         ),
     );
 
+    let network_scheme = Arc::new(Scheme::certificate_verifier(
+        crate::config::NAMESPACE,
+        config.network_identity.identity,
+    ));
+
     let actor = Driver {
         context: ContextCell::new(context),
         config,
         mailbox: rx,
         current_epoch: epoch_info.epoch(),
         last_boundary,
+        network_scheme,
     };
     Ok((actor, mailbox))
 }
@@ -203,6 +196,7 @@ pub(super) struct Driver<TContext> {
 
     last_boundary: Height,
     current_epoch: Epoch,
+    network_scheme: Arc<Scheme<PublicKey, MinSig>>,
 }
 
 impl<C: Clock + Rng + CryptoRng> Driver<C>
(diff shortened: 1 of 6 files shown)


crates/commonware-node/src/follow/driver.rs
```diff
@@ -14,21 +14,24 @@ use commonware_consensus::{
...
     ed25519::{self, PublicKey},
 };
 use commonware_math::algebra::Random as _;
+use commonware_parallel::Sequential;
 use commonware_runtime::{Clock, ContextCell, Spawner, spawn_cell};
 use commonware_utils::{Acknowledgement, vec::NonEmptyVec};
 use rand_08::{CryptoRng, Rng};
 
-use eyre::{OptionExt as _, Report, WrapErr as _};
+use eyre::{Context, OptionExt as _, Report, bail, ensure};
```
SuperFluffy: Plz `WrapErr` over `Context`. `Context` is an alias they use to make migration from anyhow easier.

crates/commonware-node/src/lib.rs
```diff
@@ -167,12 +167,22 @@ pub async fn run_follow_stack(
...
+
+    let epoch_length = chain_spec
         .info
         .epoch_length()
         .ok_or_eyre("chainspec did not contain epochLength")?;
 
+    let chain_spec_network_identity = chain_spec
+        .network_identity
+        .clone()
+        .expect("chainspec has no genesis network identity");
```
SuperFluffy: Expect messages should contain what you expect to happen, not the opposite.

Also the term "genesis" is confusing here because this need not be the genesis chainspec AFAIU.

crates/commonware-node/src/lib.rs
```diff
@@ -167,12 +167,22 @@ pub async fn run_follow_stack(
...
+
+    let network_identity = config
+        .network_identity()
+        .unwrap_or(chain_spec_network_identity);
```
SuperFluffy, line 185: Let's communicate which network identity was used (the override or chainspec).

crates/commonware-node/src/follow/driver.rs
```diff
@@ -267,6 +278,19 @@ where
...
+            let network_identity = &self.config.network_identity;
+            if onchain_outcome.epoch.get() >= network_identity.from_epoch
+                && network_identity.identity != *onchain_outcome.network_identity()
+            {
+                warn!(
+                    compiled_from_epoch = network_identity.from_epoch,
+                    onchain_epoch = %onchain_outcome.epoch,
+                    compiled_network_identity = %network_identity.identity,
+                    onchain_network_identity = %onchain_outcome.network_identity(),
+                    "Network identity differs from the on-chain DKG outcome!!! update the binary with the latest network identity"
```
SuperFluffy: This one is super confusing if we haven't yet updated the chainspec to contain the latest network identity.

Either way, if my understanding of `heal_gap` is right, this is only hit on startup? So during normal operation the follower should transition alongside the the upstream node?

SuperFluffy (commented):
Trying to understand what happens with blocks in older epochs - if we have a finalization cert for a block in epoch E but we still want to feed blocks to the marshal actor in epoch E-3 - will that cause issues because it does not have a scheme verifier for E-3 or will it skip verification because it assumes all blocks in epochs <E implicitly finalized?

crates/commonware-node/src/follow/driver.rs
hamdiallam: Ah yea let me remove this. It's not particularly helpful. Had it here on the first iteration

And yes you're right in normal operation the follower will transition alongside so we only need this warning log in `process_update` at the time of transition. Not during startup

crates/commonware-node/src/lib.rs
hamdiallam: yea a bit confusing. I'll default identity specifically in the genesis block

crates/commonware-node/src/follow/driver.rs
hamdiallam: Ah gotcha, wasn't aware of the difference

hamdiallam:
> Trying to understand what happens with blocks in older epochs - if we have a finalization cert for a block in epoch E but we still want to feed blocks to the marshal actor in epoch E-3 - will that cause issues because it does not have a scheme verifier for E-3 or will it skip verification because it assumes all blocks in epochs <E implicitly finalized?

So there's two paths.
  - If the network identity is valid for epoch E, then it will skip verification and set the floor. No blocks <E will be sent and pipeline sync should fill in the gaps. So yes assumes validity <E

 - The network identity is invalid for epoch E. The boundary for epoch E-3 must be processed in order to correctly transition epochs. The follower hints the boundary and sets the floor for E-3 to try progress into E-2. The upstream needs to supply this boundary block for the follower to get "unstuck"

crates/commonware-node/src/lib.rs
hamdiallam: Added a log line

crates/commonware-node/src/lib.rs
```diff
@@ -177,12 +178,14 @@ pub async fn run_follow_stack(
     let chain_spec_network_identity = chain_spec
         .network_identity
         .clone()
-        .expect("chainspec has no genesis network identity");
+        .ok_or_eyre("chainspec has no dkg outcome in genesis header")?;
```
SuperFluffy, line 181: Coming back to this - is "genesis" really the right term for this?

SuperFluffy (approved):

crates/commonware-node/src/lib.rs
hamdiallam: any alternative? Technically this is the genesis identity for the network

PR #4072, opened 2026-05-19
TIP-1055: versioned DEX order storage

danrobinson:
## Summary
- add TIP-1055 for versioned Stablecoin DEX order storage behind the existing `orders` mapping
- introduce an `Orders` abstraction that decodes legacy version 0 orders and writes new version 1 orders
- move DEX reads, deletes, and hot field mutations (`remaining`, `prev`, `next`) through version-aware methods

## Notes
- The version discriminator is byte offset 31 of slot 0. Legacy orders store only the `uint128 order_id` at byte offset 0, so that byte is guaranteed to decode as zero.
- `getOrder(uint128)` stays ABI-compatible; versioned layouts synthesize `order_id` from the mapping key.
- Mixed-version linked lists are supported by updating neighbor pointers based on the neighbor order's own version.

## Validation
- `cargo test -p tempo-precompiles stablecoin_dex:: --lib`
- `cargo test -p tempo-precompiles --features test-utils --test storage test_stablecoin_dex_layout`

### DIFF
--- a/crates/precompiles/src/stablecoin_dex/mod.rs
+++ b/crates/precompiles/src/stablecoin_dex/mod.rs
@@ -4,7 +4,7 @@ pub mod error;
 pub mod order;
 pub mod orderbook;
 
-pub use order::Order;
+pub use order::{Order, Orders};
 pub use orderbook::{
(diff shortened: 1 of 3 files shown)


danrobinson:
Closing in favor of a replacement PR on the properly named tip/1056 branch with the TIP renumbered to TIP-1056.

PR #4077, opened 2026-05-20
chore: deny cast_lossless clippy lint

0xalpharush:
Replace strictly lossless upcasts with explicit From conversions. This preserves behavior while making widening casts intentional.

Denying cast_lossless makes the remaining as casts easier to audit because they are no longer mixed with behavior-preserving upcasts.

Remaining casts are potentially truncating, sign-changing, enum conversions, or otherwise outside this lint.

### DIFF
--- a/.github/scripts/bench-replay-summary.py
+++ b/.github/scripts/bench-replay-summary.py
@@ -500,14 +500,14 @@ def generate_markdown(
 
 
 def main():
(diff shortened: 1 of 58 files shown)


0xalpharush:
failing forge build seems unrelated to PR

mattsse (approved):

PR #4079, opened 2026-05-20
docs(tip): TIP-1063 Protocol Features

emmajam:
This TIP defines feature flags for Tempo protocol changes, decoupling individual feature activation from hardfork bundles.


tips/tip-1063.md
```diff
@@ -0,0 +1,346 @@
...
+
+## Abstract
+
+This TIP defines feature flags for Tempo protocol changes, decoupling individual feature activation from hardfork bundles.
+
+Each feature can be tracked separately, signaled by validators, and activated onchain. A feature becomes active only after the Tempo multisig schedules it and enough active validators have signaled support.
+
+## Motivation
+
+Tempo currently ships protocol changes through hardfork bundles (e.g. T3, T4, T5). This creates coordination overhead and often bundles unrelated feature updates under the same hardfork.
```
jenpaff:
```suggestion
Tempo currently ships protocol changes through hardfork bundles (e.g. T3, T4, T5). This forces a waterfall process: every protocol change in flight has to converge on the same activation boundary, regardless of whether the changes are related, ready at the same time. Features that finish early wait.
```

tips/tip-1063.md
```diff
@@ -0,0 +1,346 @@
...
+
+Feature flags separate code distribution from behavior activation. Validators can upgrade over time, and each feature can activate only after enough of the active validator set has signaled support. This does not remove release planning, audits, partner coordination, or hardforks; it reduces how often unrelated protocol changes must activate at the same boundary.
```
jenpaff:
```suggestion
Feature flags separate code distribution from behavior activation. Validators can upgrade over time, and each feature can activate only after enough of the active validator set has signaled support. This means shipping releases on a regular cadence, independently from setting an activation time, and the ability to disable features in case of an incident instead of shipping an emergency release.
```

emmajam: Agree with the motivation shift. I’d just avoid saying “disable features in an incident” because v1 only really supports cancelling/delaying before activation; post-activation rollback is still a new protocol change. 

I’ll reword around decoupling release cadence from activation and avoiding emergency releases to move/cancel activation.

tips/tip-1063.md
```diff
@@ -0,0 +1,346 @@
...
+
+After a feature is scheduled but before it is active, the activation signer can cancel or reschedule the feature.
+
+After activation, local disabling is not allowed. An operator must not be able to turn off an active feature through local configuration.
+
+Post-activation disabling through local configuration or the feature registry is not supported in v1. An active feature that needs to be changed, disabled, or replaced must be handled by a new protocol feature, hardfork, or emergency upgrade that supersedes the old behavior. This is especially important for features that change state, deploy contracts, perform migrations, or alter assumptions that are unsafe to unwind locally.
+
+## 11. Compatibility
+
+Feature code can ship before activation. Running a Tempo binary that supports an inactive feature must not change protocol behavior.
```
jenpaff: i think this section needs some re-work, what we want here is: Does this tip affect any existing tooling or require tooling changes.

tips/tip-1063.md
```diff
@@ -0,0 +1,346 @@
...
+
+A bitmap commit whose submitted words are identical to the proposer's stored bitmap does not change support counts and does not emit support-change events.
+
+The registry must emit events when validator bitmap words are committed and when feature support changes. The feature-support event identifies the affected feature by index and ID; current support counts are read from registry state.
+
+Validator set changes must be reflected in feature support counts. A validator entering the active set counts only after it is eligible under the validator set rules and has a committed bitmap. A validator leaving the active set no longer contributes to quorum.
+
+## 5. Quorum
+
+For v1, quorum is validator-count based and uses one global threshold for all features. The threshold is Tempo's consensus threshold, expressed as `n/m`, where `m` is the active validator set size and `n` is the required number of active validators with committed support for the feature.
```
jenpaff: i think v1 mentioned here sounds strange. the tip specifies the process being shipped, curious what other versions we have in mind?

tips/tip-1063.md
0xKitsune: Yeah agreed, will update.

tips/tip-1063.md
```diff
@@ -0,0 +1,336 @@
...
+
+Compatibility requirements for each individual feature must be documented by that feature's TIP. At minimum, each feature TIP states:
+
+- what behavior changes after activation
+- which releases support the feature
+- what old nodes do after activation
+- whether RPC, simulation, indexing, or partner behavior changes
+- whether the feature is reversible, supersedable, or state-changing
+
+## 12. Observability
```
jenpaff: would remove this section if we're not more explicit

tips/tip-1063.md
```diff
@@ -0,0 +1,305 @@
...
+
+Feature flags separate code distribution from behavior activation. Validators can upgrade over time, and each feature can activate only after enough of the active validator set has signaled support. This means releases can continue on a regular cadence, while activation can be scheduled, delayed, or cancelled independently instead of requiring an emergency release to move an activation.
+
+# Specification
+
+## 1. Scope
+
+This TIP defines the feature flag mechanism for Tempo protocol features.
+
+A feature flag controls activation for a named protocol feature. A protocol feature is a unit of protocol behavior that can be shipped inactive and later activated by chain history.
```
jenpaff: >and later activated by chain history.

what does this mean?

tips/tip-1063.md
```diff
@@ -0,0 +1,305 @@
...
+
+A protocol feature is appropriate when the change affects consensus validation, state transition behavior, transaction validity, system contract behavior, precompile behavior, block building, RPC simulation, or other protocol-facing logic.
+
+This TIP makes feature flags the normal replacement for shipping protocol changes through hardfork bundles. Hardforks can still be used when a change cannot be safely represented as a feature flag, including introduction of this mechanism, changes to feature signaling itself, or emergency changes that the feature mechanism cannot express.
```
jenpaff:
```suggestion
This TIP makes feature flags the default for shipping protocol changes through hardfork bundles. Hardforks will still be used in interim.
```

tips/tip-1063.md
```diff
@@ -0,0 +1,285 @@
...
+
+mapping(uint64 minimumVersionKey => MinimumVersionCheckpoint) minimumVersionCheckpoints;
+mapping(address validator => mapping(uint64 minimumVersionKey => bool)) validatorSupportsMinimumVersion;
+
+function minimum_required_version(uint64 minimumVersionKey) external onlyAuthorized;
+```
+
+Each `MinimumVersionCheckpoint` includes a `minimumVersionKey` and a `supportCount` indicating how many validators have reported compatibility with the specified version. `minimumVersionKey` is an encoded semver value where a given `x.y.z` version is encoded as `(x << 32) | (y << 16) | z`.
+
+Validators read the latest registered MRV and report support through a system transaction. The system transaction succeeds when the MRV checkpoint exists, the validator's encoded version is greater than or equal to `minimumVersionKey`, and the validator has not already reported support for that checkpoint. The registry then records `validatorSupportsMinimumVersion[validator][minimumVersionKey] = true` and increments `minimumVersionCheckpoints[minimumVersionKey].supportCount`.
```
hamdiallam: Should there be a delay (n epochs) period prior to reporting this system transaction in case of rollbacks? Or can another system tx cause the supportCount to decrement in the event of a rollback

Or a we ensuring the multisig is responsible for ensures socially that were no issues potentially causing validators to rollback after upgrades. This sounds fine as long as versions are given enough time to bake anyways

tips/tip-1063.md
```diff
@@ -0,0 +1,287 @@
...
+
+The registry specifies a designated signer authorized to register approved features onchain and schedule activation epochs.
+
+Each registry entry includes:
+
+- `feature_id`: a globally unique, lowercase, dot-separated feature identifier
+- `description`: a short description of the behavior; including the TIP reference
+- `minimum_supported_version`: the lowest Tempo version expected to validate, build, simulate, and serve the feature correctly after activation
+- `status`: the status of the feature. Available statuses are `pending` and `active`
+- `activation_epoch`: the epoch in which the feature becomes active, if scheduled; set to 0 means the feature is not scheduled
```
SuperFluffy: Does that mean there are features that will never be considered active at genesis? Important if we were to ever perform a re-genesis and reset back to height 0.

tips/tip-1063.md
```diff
@@ -0,0 +1,287 @@
...
+
+The registry should expose enough read methods for nodes, indexers, and operators to inspect feature state, validator reported versions, current support counts, and the global activation quorum threshold. Callers should be able to determine whether a feature is known, scheduled, or active, including for a specific block or epoch.
+
+Registry state changes must be observable through events for registration, activation scheduling, schedule updates, support count changes, activation, and cancellation.
+
+## Scheduling and Activation
+
+A registered feature can be scheduled for activation once the active validator set has reached quorum for the feature's `minimum_supported_version`. The authorised signer/multisig specified by the registry contract then schedules activation by submitting a registry transaction with the feature ID and a future activation epoch. The registry accepts the scheduling transaction only if the feature is registered, the status is `pending` and quorum over active validators is satisfied. The requested activation epoch must be strictly greater than the current epoch.
```
SuperFluffy: What's the story for when the active validator set changes during this activation process?

It is important that validators that dropped out are ignored here.

tips/tip-1063.md
```diff
@@ -0,0 +1,287 @@
...
+
+Scheduling writes the activation epoch to the feature metadata, an event is emitted to notify network participants and the scheduled epoch becomes the upgrade deadline for any nodes running a version older than the minimum supported version. Network quorum on version support is validated when the epoch is scheduled and is not revalidated at activation time.
```
SuperFluffy: Assume the event is emitted - what then?

Should there be a mechanism to remove nodes from the active peer set if they don't support the feature?

tips/tip-1063.md
```diff
@@ -0,0 +1,287 @@
...
+```
+
+Registered feature identity is immutable. After registration, the registry must not allow edits to `feature_id` or `minimum_supported_version`.
+
+Feature lifecycle state is mutable. Mutable state is limited to `status` and `activation_epoch`.
+
+The registry also tracks the Tempo version each validator reports. This is used to ensure that enough active validators are running a Tempo version that supports a given feature before the feature can be scheduled for activation. A version report counts only if it is deterministic from chain history. For a given feature, a validator supports the feature when its latest reported Tempo version is greater than or equal to the feature's `minimum_supported_version`.
+
+Releases are monotonic for feature support. A newer release supports every feature supported by earlier releases. Selectively opting out of features or non-monotonic support would require per-feature signaling and is not part of this design for simplicity.
+
```
0xalpharush, line 118: Discussing how fixes/modifications to not-yet-activated features that have already been given and ID and in a release would be good. That is, it's possible to have a semantic difference introduced (someone running the `minimum_supported_version` and `minimum_supported_version + 1` would fork) , and I think it's best to just require nearly every modification to change the ID.

tips/tip-1063.md
```diff
@@ -0,0 +1,287 @@
...
+## Feature Registry
+
+Tempo stores protocol feature data in an onchain feature registry. For a given feature, the registry records the feature ID, minimum supported version, status, and activation epoch. Registry state is the source of truth for feature activation on Tempo.
+
+The registry specifies a designated signer authorized to register approved features onchain and schedule activation epochs.
+
+Each registry entry includes:
+
+- `feature_id`: a globally unique, lowercase, dot-separated feature identifier
+- `description`: a short description of the behavior; including the TIP reference
```
horsefacts: What is the motivation behind storing these onchain vs something like an integer ID with offchain metadata? Do we expect anyone to consume this onchain?

tips/tip-1063.md
```diff
@@ -0,0 +1,287 @@
...
+- `status`: the status of the feature. Available statuses are `pending` and `active`
+- `activation_epoch`: the epoch in which the feature becomes active, if scheduled; set to 0 means the feature is not scheduled
+
+```solidity
+contract FeatureRegistry {
+  // --snip--
+
+  struct Feature {
+    string description;
+    string minimum_supported_version;
```
horsefacts: Should we store this in the same encoded format as `minimumVersionKey` below?

tips/tip-1063.md
horsefacts: Same question: what happens if I downgrade my validator but support was already recorded? Seems like we would need to check support every block at the top of the block.

tips/tip-1063.md
```diff
@@ -0,0 +1,287 @@
...
+
+Activation is bound to epoch boundaries rather than wall-clock timestamps for two reasons. First, epoch transitions in Tempo's consensus are not reorged, so once the chain crosses into the activation epoch the feature switch is irreversible and every node observes the same activation boundary. Second, anchoring activation to epoch boundaries lets the same feature flag mechanism gate consensus-layer features as well, since the CL can read the registry at epoch boundaries and apply activation in lockstep with the EL.
+
+## Feature Registry
+
+Tempo stores protocol feature data in an onchain feature registry. For a given feature, the registry records the feature ID, minimum supported version, status, and activation epoch. Registry state is the source of truth for feature activation on Tempo.
+
+The registry specifies a designated signer authorized to register approved features onchain and schedule activation epochs.
```
horsefacts: This is a bit underspecified. Is it hardcoded? Can it change? What is the path forward as Tempo decentralizes?

tips/tip-1063.md
```diff
@@ -0,0 +1,287 @@
...
+
+  mapping(string feature_id => Feature) features;
+
+  // --snip--
+}
+```
+
+Registered feature identity is immutable. After registration, the registry must not allow edits to `feature_id` or `minimum_supported_version`.
+
+Feature lifecycle state is mutable. Mutable state is limited to `status` and `activation_epoch`.
```
horsefacts: For very important components like this where we are not particularly gas sensitive I think it is good defense in depth to separate immutable from mutable state.

horsefacts:
We should better specify invariants, test cases, and security considerations in this TIP.

tips/tip-1063.md
horsefacts: > Network quorum on version support is validated when the epoch is scheduled and is not revalidated at activation time.

Why wouldn't we revalidate this at activation time?

tips/tip-1063.md
```diff
@@ -0,0 +1,287 @@
...
+
+# Specification
+
+## Overview
+
+A feature flag controls whether a specific Tempo protocol change is active. Feature gated code ships inactive in a release. Each feature is recorded in an onchain feature registry with a minimum supported Tempo version. A validator supports the feature when its reported version is greater than or equal to that `minimum_supported_version`.
+
+A feature can be scheduled for activation once enough active validators have reported a Tempo version that supports it. A designated multisig schedules the feature by setting a future activation epoch in an onchain feature registry. When the feature is scheduled, Tempo validates that enough active validators support the feature, and the activation epoch becomes the upgrade deadline. When the chain crosses into the scheduled epoch, the feature becomes active. Validators that have not upgraded must stop before processing blocks in epochs that require the active feature.
```
horsefacts: The definition of "must stop" seems underspecified. What actually happens to nodes that have not upgraded at activation? Dropped from the peer set? Dropped from the validator set? What are the failure modes here?

tips/tip-1063.md
```diff
@@ -0,0 +1,287 @@
...
+
+```solidity
+
+/// @title IFeatureRegistry
+/// @notice Registry for Tempo protocol feature activation.
+interface IFeatureRegistry {
+    enum FeatureStatus {
+        Pending,
+        Active,
+        Cancelled
```
horsefacts: I think this state is inconsistent with L#114

> Cancellation clears `activation_epoch` and leaves the feature `pending`.

tips/tip-1063.md
```diff
@@ -0,0 +1,287 @@
...
+    /// @notice Reports that an active validator is running at least the checkpoint version.
+    function reportMinimumVersionSupport(address validator, uint64 minimumVersionKey, uint64 validatorVersionKey) external;
+
+    /// @notice Schedules activation for a pending feature.
+    function scheduleActivation(string calldata featureId, uint64 activationEpoch) external;
+
+    /// @notice Replaces the activation epoch for a scheduled pending feature.
+    function rescheduleActivation(string calldata featureId, uint64 activationEpoch) external;
+
+    /// @notice Cancels a scheduled activation before the feature become active.
```
horsefacts:
```suggestion
    /// @notice Cancels a scheduled activation before the feature becomes active.
```

tips/tip-1063.md
```diff
@@ -0,0 +1,287 @@
...
+Tempo stores protocol feature data in an onchain feature registry. For a given feature, the registry records the feature ID, minimum supported version, status, and activation epoch. Registry state is the source of truth for feature activation on Tempo.
+
+The registry specifies a designated signer authorized to register approved features onchain and schedule activation epochs.
+
+Each registry entry includes:
+
+- `feature_id`: a globally unique, lowercase, dot-separated feature identifier
+- `description`: a short description of the behavior; including the TIP reference
+- `minimum_supported_version`: the lowest Tempo version expected to validate, build, simulate, and serve the feature correctly after activation
+- `status`: the status of the feature. Available statuses are `pending` and `active`
```
horsefacts: Can we simplify this to derived state based on activation epoch? Something like:

- `unscheduled` if `activation_epoch === 0`
- `scheduled` if `activation_epoch != 0 && now < activation_epoch`
- `active` if `activation_epoch != 0 && now >= activation_epoch`

"Cancelled" is a transition from scheduled back to unscheduled.

tips/tip-1063.md
horsefacts: (Something like a numeric ID seems cleaner here if we need to represent it as an enum and pass it around the code too)

horsefacts:
Maybe it's not necessary to capture in detail as part of this TIP spec, but since this has so much impact on validators, node operators, and engineering, we should play out the practical consequences for all of them. Specifically: 

- What changes about the release and upgrade process and timelines for validators and operators? What do we need to communicate when? What do different kinds of operators need to know about upgrading and rolling back? How will timelines change?
- What do feature flagged components look like in the Rust codebase? How do we prevent spaghetti complexity from many interacting flags? What can go wrong?
- How do we test across feature flagged components and guarantee that we don't introduce untested combinations? We will have dependencies between which features are on/off and the order in which they activate.

tips/tip-1063.md
```diff
@@ -0,0 +1,287 @@
...
+
+A feature can be scheduled for activation once enough active validators have reported a Tempo version that supports it. A designated multisig schedules the feature by setting a future activation epoch in an onchain feature registry. When the feature is scheduled, Tempo validates that enough active validators support the feature, and the activation epoch becomes the upgrade deadline. When the chain crosses into the scheduled epoch, the feature becomes active. Validators that have not upgraded must stop before processing blocks in epochs that require the active feature.
+
+Activation is bound to epoch boundaries rather than wall-clock timestamps for two reasons. First, epoch transitions in Tempo's consensus are not reorged, so once the chain crosses into the activation epoch the feature switch is irreversible and every node observes the same activation boundary. Second, anchoring activation to epoch boundaries lets the same feature flag mechanism gate consensus-layer features as well, since the CL can read the registry at epoch boundaries and apply activation in lockstep with the EL.
```
dankrad: I don't understand the "not reorged" argument. The same would be true for block number?

tips/tip-1063.md
```diff
@@ -0,0 +1,287 @@
...
+
+Tempo stores protocol feature data in an onchain feature registry. For a given feature, the registry records the feature ID, minimum supported version, status, and activation epoch. Registry state is the source of truth for feature activation on Tempo.
+
+The registry specifies a designated signer authorized to register approved features onchain and schedule activation epochs.
+
+Each registry entry includes:
+
+- `feature_id`: a globally unique, lowercase, dot-separated feature identifier
+- `description`: a short description of the behavior; including the TIP reference
+- `minimum_supported_version`: the lowest Tempo version expected to validate, build, simulate, and serve the feature correctly after activation
```
dankrad: Should there also be a field for `required_features` so that we can have a feature that requires other features to be activated first?

tips/tip-1063.md
```diff
@@ -0,0 +1,287 @@
...
+event FeatureScheduled(string featureId, uint64 activationEpoch);
+
+/// @notice Emitted when a pending feature's activation epoch is replaced.
+event FeatureRescheduled(string featureId, uint64 activationEpoch);
+
+/// @notice Emitted when a scheduled feature is cancelled.
+event FeatureScheduleCancelled(string featureId);
+
+/// @notice Emitted when a scheduled feature becomes active.
+event FeatureActivated(string featureId, uint64 activationEpoch);
```
dankrad: As part of which transaction is this event emitted? Is there a system transaction when the feature is activated?

dankrad:
My concern with this mechanism is that I think many TIPs actually depend on each other. It's much easier to write a TIP knowing exactly what comes before, than if you make no assumptions about that.

One example is all the things we do about state. The different TIPs about temporary state (e.g. 1040), changes in gas accounting (1016) and credits (1060) all interact with eachother. It would be 10x more effort to write them in a way so that they can be applied in any possible order.

tips/tip-1063.md
jenpaff: Agreed, “must stop” needs to be made explicit. I’ll update this section and add a clarifying security/failure mode section

The intended behavior is the same as a scheduled hardfork: Once a new feature is activated, nodes that do not support that active feature cannot safely validate blocks at/after the activation epoch. They will reject valid blocks, produce invalid blocks or fork from the network.

tips/tip-1063.md
jenpaff: Will elaborate in this section & update the interface below. The signer should be set in the registry when it is introduced, not hardcoded in node binaries with the ability to rotate the signer.

tips/tip-1063.md
jenpaff: makes sense to me, so we could have 

```
uint64 featureId;
uint64 minimumSupportedVersionKey;
uint64 activationEpoch;
```

on-chain and host an offchain metadata file with description, tip url etc

tips/tip-1063.md
jenpaff: Hm good catch, so maybe instead we should encode to `activation_epoch = uint64.max (2^64 - 1)`

tips/tip-1063.md
```diff
@@ -0,0 +1,287 @@
...
+
+The registry also tracks the Tempo version each validator reports. This is used to ensure that enough active validators are running a Tempo version that supports a given feature before the feature can be scheduled for activation. A version report counts only if it is deterministic from chain history. For a given feature, a validator supports the feature when its latest reported Tempo version is greater than or equal to the feature's `minimum_supported_version`.
+
+Releases are monotonic for feature support. A newer release supports every feature supported by earlier releases. Selectively opting out of features or non-monotonic support would require per-feature signaling and is not part of this design for simplicity.
+
+A feature has quorum when enough validators in the active validator set support it. The threshold is Tempo's consensus threshold, expressed as `n/m`, where `m` is the active validator set size and `n` is the required number of active validators that support the feature. Quorum does not activate the feature by itself. It only makes the feature eligible to be scheduled for activation.
```
jenpaff: The tip should specify `n` , it should not be consensus threshold but a specific `n` e.g. 80%

jenpaff: >Quorum does not activate the feature by itself. It only makes the feature eligible to be scheduled for activation.

quorum shouldn't gate scheduling, but activation

tips/tip-1063.md
jenpaff: Yes we should change this: Network quorum should be validated at activation time only

tips/tip-1063.md
jenpaff: We'll add a dependency bitmap to highlight dependencies of a feature. If features B depends on feature A, A must be activated before B (+1 epoch)

tips/tip-1063.md
jenpaff: When a feature flag changes throughout development (before its been activated) , the `minimum_supported_version` will be incremented, the ID will not change. 

Nodes that do not run the  `minimum_supported_version` will fall out of sync

tips/tip-1063.md
jenpaff: Validators that do not run the `minimum_supported_version` will fall out of sync and would not produce valid blocks

tips/tip-1063.md
```diff
@@ -0,0 +1,203 @@
...
+protocolVersion: TBD
+---
+
+# TIP-1063: Protocol Feature Flags
+
+## Abstract
+
+This TIP defines protocol feature flags for Tempo protocol changes, decoupling code distribution from activation.
+
+Each feature ships inactive in a Tempo release and is assigned a monotonically increasing feature ID. The chain stores a single active feature tip: the edge of the active feature-flag set and the highest active feature ID. A chain with feature tip `N` treats every feature ID in `1..=N` as active.
```
jenpaff: >The chain stores a single active feature tip: the edge of the active feature-flag set and the highest active feature ID. A chain with feature tip `N` treats every feature ID in `1..=N` as active.

>The registry owner schedules a higher feature tip for a future activation epoch. Feature activation occurs in the first block of that consensus epoch, and the scheduled tip becomes active if the current active validator set has quorum for it.

I think this section is confusing, lets riff on how to make this easy to understand

tips/tip-1063.md
```diff
@@ -0,0 +1,203 @@
...
+
+The registry owner schedules a higher feature tip for a future activation epoch. Feature activation occurs in the first block of that consensus epoch, and the scheduled tip becomes active if the current active validator set has quorum for it.
+
+## Motivation
+
+Tempo currently ships protocol changes through hardfork bundles (e.g. T3, T4, T5). This requires multiple protocol changes to converge on the same activation epoch even if they are production ready on different timelines.
+
+Feature flags separate code distribution from feature activation. Validators upgrade on a regular release cadence, report the highest feature ID they support, and the chain can advance the active feature tip once the network has quorum for the scheduled target. Activation can be scheduled, delayed, or cancelled without a subsequent release.
```
jenpaff: >report the highest feature ID they support

ah i see we changed this so no minimum version needed?

tips/tip-1063.md
```diff
@@ -0,0 +1,203 @@
...
+
+# Specification
+
+## Overview
+
+A protocol feature controls whether a specific Tempo protocol change is active. Feature-gated code ships inactive in a release. Each feature is assigned a globally unique `uint64` feature ID in strictly increasing order. Feature ID `0` is reserved to represent the absence of active protocol features.
+
+Feature IDs define a contiguous activation namespace. If a feature ID is skipped, withdrawn, or intentionally left unused, it must be recorded as a reserved no-op feature slot before any higher feature tip activates. Reserved no-op slots have no protocol effect, but they are treated as active when `features_tip` advances past them.
```
jenpaff: >Feature IDs define a contiguous activation namespace

what does this mean?

tips/tip-1063.md
```diff
@@ -0,0 +1,203 @@
...
+
+The active feature-flag set is represented by a single onchain cursor, `features_tip`, also called the active feature tip. If `features_tip == N`, every feature ID from `1` through `N` is active and every feature ID greater than `N` is inactive. The chain does not store per-feature activation structures or dependency metadata.
+
+Feature ordering is the dependency model. A feature that depends on another feature must receive a higher feature ID than the feature it depends on. Source-controlled metadata records human-readable feature names, descriptions, TIP references, and implementation notes keyed by `feature_id`; the chain stores only the state required for activation.
```
jenpaff: >Feature ordering is the dependency model. A feature that depends on another feature must receive a higher feature ID than the feature it depends on. 

In practise this means: If feature 1, 2, 3 were recorded and we don't want to ship feature 2, we will skip 2 and the feature will get a new id whenever its ready?

tips/tip-1063.md
```diff
@@ -0,0 +1,203 @@
...
+
+## Feature Registry
+
+Tempo stores protocol feature activation state in an onchain feature registry. The registry has an owner authorized to schedule or cancel feature tip activation.
+
+The registry stores:
+
+- `features_tip`: the highest active protocol feature ID
+- `scheduled_features_tip`: the next target feature tip, or `0` when no tip is scheduled
+- `scheduled_activation_epoch`: the earliest epoch in which the scheduled tip may activate, or `0` when no tip is scheduled
```
jenpaff: >- `scheduled_activation_epoch`: the earliest epoch in which the scheduled tip may activate, or `0` when no tip is scheduled

similar comment we've seen previously, needs to be max in case we regenesis

tips/tip-1063.md
```diff
@@ -0,0 +1,203 @@
+---
+id: TIP-1063
+title: Protocol Feature Flags
+description: Defines protocol feature flags and the active feature tip used to activate them.
+authors:
+status: Draft
+related: TIP-0000
```
jenpaff: TIP-1070

tips/tip-1063.md
```diff
@@ -0,0 +1,203 @@
...
+
+Validator feature-tip reports use the validator identity and authorization model defined by TIP-1070. The registry accepts `setSupportedFeaturesTip(validator, featuresTip)` only when the caller is authenticated as that validator under TIP-1070; otherwise it rejects the report as unauthorized. Reports are not owner-mediated and cannot be submitted on behalf of another validator by an arbitrary caller.
+
+```solidity
+mapping(address validator => uint64 supported_features_tip) validator_supported_features_tip;
+
+function setSupportedFeaturesTip(address validator, uint64 featuresTip) external;
+```
+
+Support counts must track the active validator set. Exited validators stop contributing to quorum, and newly active validators contribute according to their latest reported supported feature tip. Validator-set transitions must update support counts deterministically.
```
jenpaff: do we still have support counts?

tips/tip-1063.md
```diff
@@ -0,0 +1,203 @@
...
+
+For a target tip `N`, a validator contributes to quorum when `validator_supported_features_tip[validator] >= N`. A validator running a newer Tempo release can therefore support older scheduled tips without additional per-feature signaling.
+
+## Cancellation and Supersession
```
jenpaff, line 168: This section should also explain what happens if a feature in a feature set is skipped/not activated in the release and what happens after

tips/tip-1063.md
```diff
@@ -0,0 +1,203 @@
...
+
+Before activation, the registry owner can cancel a scheduled feature tip. Cancellation clears `scheduled_features_tip` and `scheduled_activation_epoch`. A new target tip can be scheduled later.
+
+After activation, a feature cannot be cancelled or locally disabled, and the active feature tip cannot be lowered. Active features are part of chain history, and removing or toggling off their code can break syncing, state assumptions, or protocol behavior where the feature was active.
+
+If an active feature needs to be fixed, disabled, or replaced, the change ships as a new protocol feature with a higher feature ID. The active feature tip only moves forward; historical behavior remains available for blocks where the earlier feature was active.
+
+## Interfaces
```
jenpaff, line 256: Invariants & Threat model

tips/tip-1063.md
```diff
@@ -0,0 +1,213 @@
...
+
+Feature flags separate code distribution from feature activation. Validators upgrade on a regular release cadence, report the highest feature ID they support, and the chain can advance the active feature tip once the network has quorum for the scheduled target. Activation can be scheduled, delayed, or cancelled without a subsequent release.
+
+# Specification
+
+## Overview
+
+A protocol feature controls whether a specific Tempo protocol change is active. Feature-gated code ships inactive in a release. Each feature is assigned a globally unique `uint64` feature ID in strictly increasing order. Feature ID `0` is reserved to represent the absence of active protocol features.
+
+Feature IDs form an ordered activation list. A skipped, withdrawn, or intentionally unused feature ID only matters if the chain activates a higher feature tip that passes over it. Before that higher feature tip activates, the skipped ID must be represented in source-controlled metadata as a reserved no-op feature slot. Reserved no-op slots have no protocol effect, but they are treated as active when `features_tip` advances past them.
```
0xalpharush: I still think the index-based approach is broken given the quorom (`validator_supported_features_tip[validator] >= N`) would count disagreeing validators who advertise support for `N`, e.g. one with a no-op feature slot and one with a breaking change toward the quorom. That is, quorom would be reached prematurely bc it's based on the height and not the behavior of activating `1..N`.

This is why I recommended a checksum of the list being the quorom quard (`validator_supported_features_tip[validator][n] == hash` where hash attests to a named identifier in the list from `1..N`).

tips/tip-1063.md
```diff
@@ -0,0 +1,213 @@
...
+/// @notice Registry for Tempo protocol feature tip scheduling.
+interface IFeatureRegistry {
+    /// @notice Returns the registry owner authorized to schedule feature tip activation.
+    function owner() external view returns (address);
+
+    /// @notice Returns the fixed global activation quorum threshold: 4/5, or 80%.
+    function activationQuorum() external view returns (uint256 numerator, uint256 denominator);
+
+    /// @notice Returns the highest active protocol feature ID.
+    function featuresTip() external view returns (uint64);
```
0xalpharush:
```suggestion
    function highestActiveTip() external view returns (uint64);
```

tips/tip-1063.md
```diff
@@ -0,0 +1,213 @@
...
+
+    /// @notice Returns the scheduled feature tip and activation epoch.
+    function scheduledFeaturesTip() external view returns (uint64 featuresTip, uint64 activationEpoch);
```
0xalpharush:
```suggestion
    function getActivationEpoch(uint64 featuresTip) external view returns uint64 activationEpoch;
```

0xalpharush: If multiple activate it'd make sense to return an array from `scheduledFeaturesTip`

tips/tip-1063.md
0xalpharush: Maybe `highestSupportedTip`, too?

tips/tip-1063.md
```diff
@@ -0,0 +1,213 @@
...
+
+# Specification
+
+## Overview
+
+A protocol feature controls whether a specific Tempo protocol change is active. Feature-gated code ships inactive in a release. Each feature is assigned a globally unique `uint64` feature ID in strictly increasing order. Feature ID `0` is reserved to represent the absence of active protocol features.
+
+Feature IDs form an ordered activation list. A skipped, withdrawn, or intentionally unused feature ID only matters if the chain activates a higher feature tip that passes over it. Before that higher feature tip activates, the skipped ID must be represented in source-controlled metadata as a reserved no-op feature slot. Reserved no-op slots have no protocol effect, but they are treated as active when `features_tip` advances past them.
+
+The active feature-flag set is represented by a single onchain cursor, `features_tip`, also called the active feature tip. If `features_tip == N`, every feature ID from `1` through `N` is active and every feature ID greater than `N` is inactive. The chain does not store per-feature activation structures or dependency metadata.
```
0xalpharush: When do we anticipate doing mass activations ` X..N` versus 1 per epoch?

tips/tip-1063.md
```diff
@@ -0,0 +1,213 @@
...
+
+# TIP-1063: Protocol Feature Flags
+
+## Abstract
+
+This TIP defines protocol feature flags for Tempo protocol changes, decoupling code distribution from activation.
+
+Each feature ships inactive in a Tempo release and is assigned a monotonically increasing feature ID. The chain stores `features_tip`, the highest active feature ID. A chain with feature tip `N` treats every feature ID in `1..=N` as active and every higher feature ID as inactive.
+
+The registry owner schedules a higher feature tip for a future activation epoch. During block processing for that epoch, the scheduled tip becomes active if the current active validator set has quorum for it.
```
0xalpharush: > During block processing for that epoch,

Would like to understand from consensus team trade-offs of intra-epoch versus activating at epoch transition i.e. `activationEpoch -1` activates up to `N` for `activationEpoch`

tips/tip-1063.md
```diff
@@ -0,0 +1,213 @@
...
+    function scheduledFeaturesTip() external view returns (uint64 featuresTip, uint64 activationEpoch);
+
+    /// @notice Records the highest protocol feature tip reported by a validator public key. System caller only.
+    function setSupportedFeaturesTip(bytes32 publicKey, uint64 featuresTip) external;
+
+    /// @notice Schedules a higher feature tip for activation at the target epoch.
+    function scheduleFeaturesTip(uint64 featuresTip, uint64 activationEpoch) external;
+
+    /// @notice Activates the scheduled feature tip during block processing. System caller only.
+    function activateScheduledFeaturesTip(uint64 currentEpoch) external;
```
0xalpharush:
```suggestion
    function activateScheduledFeaturesTip() external;
```
In other [places](https://github.com/tempoxyz/tempo/pull/5213) we assume we can get epoch # from block number, so maybe we don't need param?

tips/tip-1063.md
0xalpharush: From @SuperFluffy:
> boundary / end of epoch is guaranteed to have a finalization certificate.

> also, with the boundary block we can check readiness of exactly those validators that will be members in the next epoch.

Sounds like we should target end of epoch.

klkvr:
cyclops audit

decofe:
cc @klkvr

Cyclops audit event published. [View workflow run](https://github.com/tempoxyz/tempo/actions/runs/28531627241)

tips/tip-1063.md
```diff
@@ -0,0 +1,335 @@
...
+Registry state changes MUST be observable through events for scheduling, cancellation, validator readiness reports, and activation.
+
+## Scheduling and Activation
+
+The registry owner schedules activation by submitting a registry transaction with a target feature head and a future activation epoch.
+
+The registry MUST accept the scheduling transaction only if:
+
+- the target feature head is not `bytes32(0)`;
+- the target feature head is not already active;
```
klkvr: fixed in cb964c0

tips/tip-1063.md
```diff
@@ -0,0 +1,335 @@
...
+- no other feature head is scheduled; and
+- the requested activation epoch is strictly greater than the current epoch.
+
+Scheduling MUST write `scheduled_feature_head` and `scheduled_activation_epoch`, then emit an event to notify network participants. The scheduled activation epoch is the consensus epoch whose first block executes with the activated feature head if quorum was reached. Operators should upgrade before that epoch if they are running a release that does not support the scheduled feature head.
+
+At T9 and later, the block executor invokes the activation system call during end-of-block processing for the last block of each epoch, after readiness reports and other block transactions have executed and before the TIP-1070 current committee is updated for the next epoch. The call changes registry state only when a schedule targets the next epoch. `activateScheduledFeatureHead()` does not take an epoch argument and returns the activated feature head, or `bytes32(0)` if no feature head activated.
```
klkvr, line 136: fixed in ab99535

tips/tip-1063.md
```diff
@@ -0,0 +1,335 @@
...
+
+For quorum checks, the validator set is the TIP-1070 current committee exposed by the `CurrentCommittee` precompile at the end of the reporting epoch. The feature registry MUST NOT keep a separate validator roster for feature activation and MUST NOT derive quorum membership from `ValidatorConfigV2.getActiveValidators()`.
```
klkvr, line 138: i guess this is arguable. the reason why we make `E-1` committee rule the feature activation for next epoch is to make sure that every committee member gets a chance to report readiness. otherwise, if someone rotates their keys before epoch `E` their votes would just get lost

the outlined scneario would only be problematic if somehow it happens that at epoch boundary current committee has quorum over a feature but in the next epoch a bunch of new validators added and/or old validators are removed and the newly added validators no longer support the just activated feature.

this seems failry unlikely to me. and even if we did fix this and made `E`'s committee decide whether feature is getting activated, same kind of distruptive rotation could happen at `E+1` thus stalling the chain identically

klkvr:
cyclops audit

decofe:
cc @klkvr

Cyclops audit event published. [View workflow run](https://github.com/tempoxyz/tempo/actions/runs/28551554897)

tips/tip-1063.md
```diff
@@ -0,0 +1,341 @@
...
```
klkvr, line 138: same as https://github.com/tempoxyz/tempo/pull/4079#discussion_r3509353159

emmajam (commented):
LGTM

tips/tip-1063.md
```diff
@@ -0,0 +1,341 @@
...
+
+Scheduling MUST increment `schedule_nonce`, write `scheduled_feature_head` and `scheduled_activation_epoch`, then emit an event to notify network participants. The schedule nonce identifies the current schedule instance, so readiness from canceled, failed, or superseded schedules cannot count for a later schedule, even when the same feature head and activation epoch are scheduled again. The scheduled activation epoch is the consensus epoch whose first block executes with the activated feature head if quorum was reached. Operators should upgrade before that epoch if they are running a release that does not support the scheduled feature head.
+
+At T9 and later, the block executor invokes the activation system call during end-of-block processing for the last block of each epoch, after readiness reports and other block transactions have executed and before the TIP-1070 current committee is updated for the next epoch. The call changes registry state only when a schedule targets the next epoch. `activateScheduledFeatureHead()` does not take an epoch argument and returns the activated feature head, or `bytes32(0)` if no feature head activated.
```
0xKitsune, line 136: nit: This might update to T8.

tips/tip-1063.md
```diff
@@ -0,0 +1,341 @@
...
+
+For quorum checks, the validator set is the TIP-1070 current committee exposed by the `CurrentCommittee` precompile at the end of the reporting epoch. The feature registry MUST NOT keep a separate validator roster for feature activation and MUST NOT derive quorum membership from `ValidatorConfigV2.getActiveValidators()`.
+
+`scheduledFeatureSupport()` MUST evaluate readiness for the current schedule by iterating the current committee public keys and counting each public key whose `validator_confirmed_schedule_nonce[public_key] == schedule_nonce`. A scheduled feature head has quorum when at least 80% of current committee public keys have reported readiness for that schedule instance. The required validator count is `ceil(4 * current_committee_size / 5)`.
```
0xKitsune, line 140: > least 80% of current committee public keys

Do we want to make this match the consensus threshold exactly? Or was 80% chosen for a specific reason?

tips/tip-1063.md
```diff
@@ -0,0 +1,341 @@
...
+
+If no feature head is scheduled, no activation epoch is set, or the scheduled activation epoch is not the next epoch, `activateScheduledFeatureHead()` MUST return `bytes32(0)` without changing registry state.
+
+When `current_epoch + 1 == scheduled_activation_epoch`, `activateScheduledFeatureHead()` evaluates quorum and clears the scheduled fields exactly once. If quorum is satisfied, the chain MUST set `active_feature_head` to `scheduled_feature_head`, emit an activation event, and return the activated feature head. Otherwise, the chain MUST leave `active_feature_head` unchanged and return `bytes32(0)`. The registry owner can schedule the same or another feature head for a later epoch after the missing quorum condition is resolved.
```
0xKitsune, line 144: > The registry owner can schedule the same or another feature head for a later epoch after the missing quorum condition is resolved.

Another alternative is to keep the same feature head scheduled and automatically retry activation at the next epoch after a missed quorum. To avoid indefinite surprise activation, this could be bounded by a or max retry count or similar. Once the bound is reached, the schedule clears and emits a failure event.

tips/tip-1063.md
```diff
@@ -0,0 +1,341 @@
...
+
+Threat: Activation could be scheduled for the wrong target or before the ecosystem is ready.
+
+Mitigation: The registry owner can cancel or reschedule activation before the activation block, but cannot lower the active feature head after activation. Release notes and activation communications must distinguish validator quorum from ecosystem readiness because RPC nodes, indexers, monitoring, and partner systems do not contribute to validator quorum.
+
+### Post-Activation
+
+Threat: An operator could roll back or disable behavior after it becomes active chain history.
+
+Mitigation: Nodes must not validate, build, simulate, or serve post-activation protocol behavior for a chain whose active feature head is not present in the node's supported feature cache.
```
0xKitsune, line 216: nit: not sure if this should be considered in a thread model, this would just be like running a previous stf version and the node will fall out of sync.

tips/tip-1063.md
```diff
@@ -0,0 +1,341 @@
...
+
+Threat: A validator could report readiness for code it is not running.
+
+Mitigation: A node whose binary does not support the active feature head must fail closed rather than validate, build, simulate, or serve old behavior. After activation, a validator that misreported readiness will fail to follow the chain rather than safely continue with old behavior.
+
+### Activation Quorum
+
+Threat: Activation could use the wrong validator set, identity, or quorum denominator.
+
+Mitigation: Readiness reports use the current block proposer public key, and quorum checks use the TIP-1070 current committee public keys at the end of the reporting epoch. Readiness is bound to the current schedule nonce, which is bumped on each successful schedule, so stale reports from canceled or failed schedules do not count. The registry activation path gates activation on quorum before the scheduled activation epoch begins. If quorum is missing, the scheduled feature head MUST NOT activate and the active feature head MUST remain unchanged.
```
0xKitsune, line 204: nit: is this actually valid/possible if we are using 1070 to check the consensus committee? Consider removing this.

tips/tip-1063.md
```diff
@@ -0,0 +1,341 @@
...
+
+The tradeoff is that features do not activate in arbitrary order. If an earlier feature is deferred but a later feature should still activate, the earlier feature name must become a reserved no-op entry.
+
+### Numeric Feature Tips
+
+An earlier design stored a `uint64 featuresTip` high-water mark. This is compact, but it is ambiguous across releases if the same numeric tip refers to different implementation history.
+
+This TIP uses a hash-chain commitment instead. The feature head commits to the ordered feature-name prefix and lets a node decide locally whether a scheduled feature head is present in its supported chain.
+
+## Invariants
```
0xKitsune, line 234: Similar to previous TIPs, I think we can do a pass on the invariants and remove anything that describes generic system behavior/constraints but not quite invariants that must hold.

tips/tip-1063.md
0xalpharush: I believe @SuperFluffy recommended so we have some buffer for nodes going offline. Given we compute with current committee right before its rotated to the next (and some validators counted toward quorom drop off potentially), I think the buffer is important.

tips/tip-1063.md
```diff
@@ -0,0 +1,341 @@
...
+## Assumptions
+
+- TIP-1070 provides the validator identity and current-committee model used for readiness reporting and quorum checks.
+- Releases that support a feature head agree on the source-controlled ordered feature-name list that produced that head.
+- Feature readiness reporting is deterministic from the binary and chain state. Operators do not locally choose which subset of supported protocol features to report.
+- Active feature heads are forward-only. Fixing, disabling, or replacing active behavior requires appending a new feature name to the ordered list, or a hardfork if the feature registry mechanism cannot express the emergency change.
+
+# Specification
+
+## Feature Chain
```
legion2002, line 40: Do we need this hash based feature chain? Since features are linear now, why can't we use a monotonically increasing nonce here?

tips/tip-1063.md
```diff
@@ -0,0 +1,341 @@
...
+
+## Abstract
+
+This TIP defines protocol feature flags for Tempo protocol changes, decoupling code distribution from activation.
+
+Today, a Tempo hardfork ties protocol code and activation into the same upgrade bundle. With protocol feature flags, Tempo can ship releases when features are ready, let operators upgrade, observe validator readiness, and activate the new features per network.
+
+Protocol features are assigned canonical names in one source-controlled ordered list. Each binary constructs a hash chain over that list. A feature head is the `bytes32` digest at one tip of that chain, and the zero feature head means no protocol features are active.
+
+Activation is controlled by an onchain feature registry. The registry owner schedules a target feature head and future activation epoch. Validators report readiness for the scheduled feature head through the first transaction in blocks they propose. At the end of the epoch before the scheduled activation epoch, the registry checks readiness against the TIP-1070 current committee.
```
legion2002, line 21: are we comfortable with adding a registry owner here as a central point for any feature activation on the chain?

why can't we just check for validator quorum at the end of each epoch?

tips/tip-1063.md
```diff
@@ -0,0 +1,341 @@
...
+
+Tempo stores protocol feature activation state in an onchain feature registry. The registry stores the active feature head, one scheduled feature head, a schedule nonce, and the latest schedule readiness recorded for each proposer public key.
+
+The registry owner is authorized to schedule or cancel feature activation.
+
+The registry MUST store:
+
+- `active_feature_head`: the active feature head
+- `scheduled_feature_head`: the next target feature head, or `bytes32(0)` when none is scheduled
+- `scheduled_activation_epoch`: the epoch in which the scheduled feature head is attempted, or `0` when none is scheduled
```
legion2002, line 96: If the owner has to provide the `scheduled_activation_epoch` anyways, how is this design different from a standard timestamp based hardfork activation? 

The main diff is that you don't need a new binary, you can just change the timestamp value onchain?

tips/tip-1063.md
```diff
@@ -0,0 +1,341 @@
...
+
+This TIP uses a single ordered activation chain instead. An unordered feature set would require specs, implementations, audits, tests, monitoring, and operator guidance to account for many possible feature combinations and activation orders. It would also require validators to report exact supported feature sets rather than readiness for one scheduled feature head.
+
+The tradeoff is that features do not activate in arbitrary order. If an earlier feature is deferred but a later feature should still activate, the earlier feature name must become a reserved no-op entry.
+
+### Numeric Feature Tips
+
+An earlier design stored a `uint64 featuresTip` high-water mark. This is compact, but it is ambiguous across releases if the same numeric tip refers to different implementation history.
+
+This TIP uses a hash-chain commitment instead. The feature head commits to the ordered feature-name prefix and lets a node decide locally whether a scheduled feature head is present in its supported chain.
```
legion2002, line 232: but does the hash-chain commitment really solve the ambiguity problem? nodes could still have different implementations under the hood fo rhte same named prefix.

PR #4081, opened 2026-05-20
bench: add manual CodSpeed workflow

mediocregopher:
Adds a workflow_dispatch-only CodSpeed workflow that defaults to `tempo-evm` / `tip20_execution`, while allowing package, bench target, filter, and measurement-mode inputs.

Switches Criterion benchmarks to CodSpeed's compatibility crate and updates the TIP20 execution benchmark away from `iter_custom` so CodSpeed can measure it.

### DIFF
--- a/.github/workflows/codspeed-microbench.yml
+++ b/.github/workflows/codspeed-microbench.yml
@@ -49,9 +49,12 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - uses: rui314/setup-mold@725a8794d15fc7563f59595bd9556495c0564878 # v1
       - uses: mozilla-actions/sccache-action@7d986dd989559c6ecdb630a3fd2557667be217ad # v0.0.9
-      - uses: taiki-e/install-action@d0f23220b09a75c6db730f13bb37c4f8144b4382 # v2.75.9
-        with:
-          tool: cargo-codspeed@4.6.0
+      - name: Install cargo-codspeed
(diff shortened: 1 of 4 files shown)


shekhirin (approved):

.github/workflows/codspeed-microbench.yml
```diff
@@ -0,0 +1,112 @@
...
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: rui314/setup-mold@725a8794d15fc7563f59595bd9556495c0564878 # v1
+      - uses: mozilla-actions/sccache-action@7d986dd989559c6ecdb630a3fd2557667be217ad # v0.0.9
+      - uses: taiki-e/install-action@d0f23220b09a75c6db730f13bb37c4f8144b4382 # v2.75.9
+        with:
+          tool: cargo-codspeed@4.6.0
```
shekhirin: should probably pin to a rev?

mediocregopher: Done in fdb49af95: replaced the install-action tool version with `cargo install --git https://github.com/CodSpeedHQ/codspeed-rust --rev 0c6ccae11dc48654fd83c8dd77d231e386ebd9bd --locked`.

PR #4087, opened 2026-05-20
chore: default tips verification to T5

decofe:
## Summary
- set tips/verify Foundry default hardfork to tempo:T5
- update the tips verification README to mention the T5 default
- remove redundant inline tempo:T5 forge-config pins now covered by the default
- update StablecoinDEX tests for T5 DEX behavior:
  - TIP-1056: filled flip orders are rewritten in place, reuse the original order ID, and emit OrderFlipped instead of creating a new order and emitting OrderPlaced. Because no new flipped order is allocated, nextOrderId() stays at 2 instead of advancing to 3 in the updated test.
  - TIP-1030: same-tick flips are allowed, so bid validation is flipTick >= tick and ask validation is flipTick <= tick; only the wrong side still reverts
- temporarily disable Forge build-time lint because the GitHub `nightly` binary still panics in the calls-loop analyzer; source-built Foundry after PR 14845 passes locally with lint enabled

## Validation
- git diff --check
- local: built foundry-rs/foundry at 054857e2acd14a437dce38af5f2d39a1793944cc, which includes merged PR 14845, then ran tips/verify `forge build --sizes` successfully with lint enabled
- CI before this note: GitHub `nightly` still panicked in Forge Build, so lint remains disabled until a usable nightly includes PR 14845
- prior CI on this PR with lint disabled: Forge Build, Forge Fmt, Forge Test (Rust Precompiles), and specs success passed

### DIFF
--- a/tips/verify/test/invariants/FeeAMM.t.sol
+++ b/tips/verify/test/invariants/FeeAMM.t.sol
@@ -9,8 +9,6 @@ import { ITIP403Registry } from "tempo-std/interfaces/ITIP403Registry.sol";
 
 /// @title FeeAMM Invariant Test
 /// @notice Invariant tests for the FeeAMM/FeeManager implementation
-/// forge-config: default.hardfork = "tempo:T5"
-/// forge-config: fuzz500.hardfork = "tempo:T5"
 contract FeeAMMInvariantTest is InvariantBaseTest {
 
     /// @dev Constants from Rust tip_fee_manager/amm.rs
--- a/tips/verify/test/invariants/StablecoinDEX.t.sol
+++ b/tips/verify/test/invariants/StablecoinDEX.t.sol
@@ -11,8 +11,6 @@ import { ITIP20 } from "tempo-std/interfaces/ITIP20.sol";
 /// @dev Tests invariants TEMPO-DEX1 through TEMPO-DEX19 as documented in README.md.
 /// Pinned to T5 so TEMPO-DEX17 covers TIP-1030's same-tick flip path
 /// (`flipTick == tick`).
-/// forge-config: default.hardfork = "tempo:T5"
-/// forge-config: fuzz500.hardfork = "tempo:T5"
 contract StablecoinDEXInvariantTest is InvariantBaseTest {
 
     /// @dev Mapping of actor address to their placed order IDs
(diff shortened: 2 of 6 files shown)


tips/verify/test/invariants/FeeAMM.t.sol
```diff
@@ -9,8 +9,6 @@ import { ITIP403Registry } from "tempo-std/interfaces/ITIP403Registry.sol";
 
 /// @title FeeAMM Invariant Test
 /// @notice Invariant tests for the FeeAMM/FeeManager implementation
-/// forge-config: default.hardfork = "tempo:T5"
-/// forge-config: fuzz500.hardfork = "tempo:T5"
```
0xrusowsky, line 13: isn't it desirable to leave these, in case someone runs with older hardforks for some reason?

tips/verify/test/invariants/StablecoinDEX.t.sol
```diff
@@ -11,8 +11,6 @@ import { ITIP20 } from "tempo-std/interfaces/ITIP20.sol";
 /// @dev Tests invariants TEMPO-DEX1 through TEMPO-DEX19 as documented in README.md.
 /// Pinned to T5 so TEMPO-DEX17 covers TIP-1030's same-tick flip path
 /// (`flipTick == tick`).
-/// forge-config: default.hardfork = "tempo:T5"
-/// forge-config: fuzz500.hardfork = "tempo:T5"
```
0xrusowsky, line 15: same here, it kinda doesn't hurt?

tips/verify/test/invariants/TIP1026.t.sol
```diff
@@ -13,8 +13,6 @@ import { ITIP20RolesAuthErr } from "tempo-std/interfaces/ITIP20RolesAuth.sol";
 ///      TEMPO-1026-1: `bytes(logoURI()).length <= 256` must always hold.
 ///      TEMPO-1026-2: The legacy 6-argument `createToken` selector and the
 ///                    `TokenCreated` event signature are unchanged by this TIP.
-/// forge-config: default.hardfork = "tempo:T5"
-/// forge-config: fuzz500.hardfork = "tempo:T5"
```
0xrusowsky, line 17: and here

0xrusowsky (approved):

tips/verify/test/invariants/FeeAMM.t.sol
grandizzy: we could leave it but for example if they run with T3 the Dex tests will fail (as we just updated them for T5) so better to hard bump it + cleanup IMO

PR #4091, opened 2026-05-20
perf(transaction-pool): index expiring nonce eviction

mattsse:
Adds an ordered expiring nonce eviction index so pending eviction no longer scans every expiring nonce transaction to find the lowest-priority candidate. Expiring nonce txs keep a separate ordered index because regular 2D eviction order is keyed by `AA2dTransactionId` and carries pending/queued state.

Benchmarked with a temporary, uncommitted unit harness using `cargo test --release -p tempo-transaction-pool bench_expiring_nonce_discard_comparison --lib -- --nocapture`; values are median ns/eviction across 3 samples.

| Scenario | Pool size | `main` | This PR | Speedup |
|---|---:|---:|---:|---:|
| direct `evict_one_pending` on expiring nonce txs | 1,000 | 1,246 ns | 120 ns | 10.4x |
| direct `evict_one_pending` on expiring nonce txs | 10,000 | 19,234 ns | 190 ns | 101.2x |
| direct `evict_one_pending` on expiring nonce txs | 50,000 | 88,764 ns | 367 ns | 241.9x |
| over-limit `add_transaction` + `discard` | 1,000 | 3,353 ns | 1,007 ns | 3.3x |
| over-limit `add_transaction` + `discard` | 10,000 | 18,473 ns | 906 ns | 20.4x |
| over-limit `add_transaction` + `discard` | 50,000 | 95,515 ns | 1,791 ns | 53.3x |

### DIFF
--- a/crates/transaction-pool/src/tt_2d_pool.rs
+++ b/crates/transaction-pool/src/tt_2d_pool.rs
@@ -1043,24 +1043,25 @@ impl AA2dPool {
 
         let worst_expiring = self
             .expiring_nonce_eviction_order
-            .first()
-            .map(|key| (key.priority.clone(), key.submission_id));
+            .iter()
+            .next()
+            .map(|key| (key.expiring_hash, key.priority.clone(), key.submission_id));
 
         match (worst_2d, worst_expiring) {
-            (Some((id, pri_2d, sid_2d)), Some((pri_exp, sid_exp))) => {
+            (Some((id, pri_2d, sid_2d)), Some((hash, pri_exp, sid_exp))) => {
                 // Same ordering as EvictionKey::Ord: lower priority first, newer first.
                 let evict_expiring = pri_exp
                     .cmp(&pri_2d)
                     .then_with(|| sid_2d.cmp(&sid_exp))
                     .is_le();
                 if evict_expiring {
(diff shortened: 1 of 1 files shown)


crates/transaction-pool/src/tt_2d_pool.rs
```diff
@@ -1064,6 +1075,22 @@ impl AA2dPool {
...
+        let Some(pending_tx) = self.expiring_nonce_txs.remove(&eviction_key.expiring_hash) else {
+            debug_assert!(
+                false,
+                "expiring_nonce_eviction_order contains stale key {:?}",
+                eviction_key.expiring_hash
+            );
+            return None;
+        };
+        Some(self.finish_remove_expiring_nonce_tx(pending_tx))
+    }
```
klkvr: i guess we could just call `remove_expiring_nonce_tx` here with the hash from `eviction_key.expiring_hash` and not add an extra `finish_remove_expiring_nonce_tx`?

i'm fine with extra redundant `expiring_nonce_eviction_order` delete here tbh if we can unify the logic

klkvr (approved):
lgtm, just a nit

fgimenez (approved):
lgtm

PR #4089, opened 2026-05-20
ci(bench): internal perf dashboard link in bench-e2e output

decofe:
## Summary
- add an internal perf dashboard URL derived from the benchmark id to bench summary metadata
- include the internal perf dashboard in PR/job observability output beside Grafana logs/traces
- add Slack action buttons for the internal perf dashboard and Grafana metrics dashboard

## Validation
- node --check .github/scripts/bench-slack-notify.js
- git diff --check

Note: Nushell is not installed in this sandbox, so I could not run a local Nu parse check.

PR #4093, opened 2026-05-20
ci(bench): use txgen benchmark_id for bench uploads

decofe:
## Summary
- use the top-level txgen `benchmark_id` from `report.json` as the ClickHouse `run_id`
- keep generating a UUID for older reports that do not include `benchmark_id`

## Test
- dry-ran `contrib/bench/upload-clickhouse-txgen.sh` with a sample txgen report and fake ClickHouse curl; captured SQL used the report UUID for inserted block rows

Relates to tempoxyz/txgen#109


decofe:
Closing in favor of applying this change directly to #4089.

PR #4125, opened 2026-05-21
test: stabilize TIP20 transfer gas snapshots

decofe:
## Summary
- replace random virtual-address user tags and reward delegate addresses in the TIP20 transfer gas matrix with deterministic per-case values
- update the TIP20 transfer gas insta snapshot for those deterministic inputs

This addresses the snapshot instability seen while validating #4121, where the same virtual-address gas case could differ by a small amount across CI runs.

## Tests
- `INSTA_UPDATE=always cargo test -p tempo-node --test it gas::tip20_transfers::test_tip20_transfer_gas_snapshots -- --exact --nocapture`
- `cargo test -p tempo-node --test it gas::tip20_transfers::test_tip20_transfer_gas_snapshots -- --exact --nocapture`
- `cargo fmt --check`


crates/node/tests/it/gas/tip20_transfers.rs
```diff
@@ -533,6 +537,19 @@ fn tip20_transfer_gas_cases() -> Vec<TransferScenario> {
...
+}
+
+fn deterministic_user_tag(case_index: usize, domain: u64) -> FixedBytes<6> {
+    let word = deterministic_word(case_index, domain);
+    FixedBytes::from_slice(&word.as_slice()[26..32])
+}
+
+fn deterministic_address(case_index: usize, domain: u64) -> Address {
+    Address::from_word(deterministic_word(case_index, domain))
+}
```
0xrusowsky: let's get rid of `deterministic_word` and inline if necessary

0xrusowsky (changes requested):

0xrusowsky (approved):

PR #4130, opened 2026-05-21
Use Arc for Tempo AA tx env

decofe:
## Summary
- change TempoTxEnv::tempo_tx_env from Option<Box<TempoBatchCallEnv>> to Option<Arc<TempoBatchCallEnv>>
- update AA tx env construction sites to use Arc::new
- use Arc::make_mut at the few expiring_nonce_idx mutation points

## Tests
- cargo test -p tempo-revm tx:: --lib
- cargo test -p tempo-evm --lib
- cargo fmt --check
- git diff --check

### DIFF
--- a/crates/evm/src/block.rs
+++ b/crates/evm/src/block.rs
@@ -512,7 +512,7 @@ where
         let (mut tx_env, recovered) = tx.into_parts();
         // Remove any prewarming-specific context that was added to the tx env.
         if let Some(tempo_tx_env) = tx_env.tempo_tx_env.as_mut() {
-            tempo_tx_env.expiring_nonce_idx = None;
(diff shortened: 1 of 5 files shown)


crates/evm/src/block.rs
```diff
@@ -512,7 +512,7 @@ where
         let (mut tx_env, recovered) = tx.into_parts();
         // Remove any prewarming-specific context that was added to the tx env.
         if let Some(tempo_tx_env) = tx_env.tempo_tx_env.as_mut() {
-            tempo_tx_env.expiring_nonce_idx = None;
+            std::sync::Arc::make_mut(tempo_tx_env).expiring_nonce_idx = None;
```
mattsse, line 515: hmm @klkvr looks like this is a bit problematic
we could however change this to an atomic i suppose

but I dont really understand why we have  this

PR #4139, opened 2026-05-21
feat(tip-1059): discount pure payment settlement

0xKitsune:
Implements TIP-1059 settlement discounts for pure TIP-20 transfer and burn payments.

Ref #4013

### DIFF
--- a/crates/chainspec/src/constants.rs
+++ b/crates/chainspec/src/constants.rs
@@ -28,6 +28,14 @@ pub mod gas {
     /// - Economic: 1,000 microdollars = 0.001 USD = 0.1 cents
     pub const TEMPO_T1_BASE_FEE: u64 = 20_000_000_000;
 
+    /// TIP-1059 discounted gas price for pure payment transfers: 12 billion attodollars.
+    pub const TEMPO_T6_DISCOUNTED_PAYMENT_GAS_PRICE: u64 = 12_000_000_000;
+
+    /// TIP-1059 maximum gas used by a transaction that can receive the pure payment discount.
+    ///
+    /// This matches the TIP-1000 new-state-write gas cost at T6.
+    pub const TEMPO_T6_MAX_DISCOUNTED_PAYMENT_GAS_USED: u64 = 250_000;
+
     /// [TIP-1010] general (non-payment) gas limit: 30 million gas per block.
     /// Cap for non-payment transactions.
     ///
--- a/crates/chainspec/src/constants.rs
+++ b/crates/chainspec/src/constants.rs
@@ -31,11 +31,6 @@ pub mod gas {
     /// TIP-1059 discounted gas price for pure payment transfers: 12 billion attodollars.
     pub const TEMPO_T6_DISCOUNTED_PAYMENT_GAS_PRICE: u64 = 12_000_000_000;
 
-    /// TIP-1059 maximum gas used by a transaction that can receive the pure payment discount.
-    ///
-    /// This matches the TIP-1000 new-state-write gas cost at T6.
-    pub const TEMPO_T6_MAX_DISCOUNTED_PAYMENT_GAS_USED: u64 = 250_000;
-
     /// [TIP-1010] general (non-payment) gas limit: 30 million gas per block.
     /// Cap for non-payment transactions.
     ///
--- a/crates/chainspec/src/constants.rs
+++ b/crates/chainspec/src/constants.rs
@@ -50,6 +50,29 @@ pub mod gas {
         original_effective_gas_price.saturating_sub(TEMPO_T6_PAYMENT_GAS_PRICE_DISCOUNT)
     }
 
+    /// [TIP-1010] general (non-payment) gas limit: 30 million gas per block.
+    /// Cap for non-payment transactions.
+    ///
+    /// [TIP-1010]: <https://docs.tempo.xyz/protocol/tips/tip-1010>
+    pub const TEMPO_T1_GENERAL_GAS_LIMIT: u64 = 30_000_000;
+
+    /// TIP-1010 per-transaction gas limit cap: 30 million gas.
+    /// Allows maximum-sized contract deployments under [TIP-1000] state creation costs.
+    ///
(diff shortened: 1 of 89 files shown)


crates/contracts/src/precompiles/tip20.rs
```diff
@@ -207,6 +207,25 @@ impl ITIP20::ITIP20Calls {
...
+
+    /// Returns `true` if `input` matches one of the TIP-1059 discounted pure-payment call selectors.
+    ///
+    /// Approvals and mints remain payment-lane eligible via [`Self::is_payment`] but do not
+    /// receive the settlement discount.
+    pub fn is_discounted_payment_call(input: &[u8]) -> bool {
+        fn is_call<C: SolCall>(input: &[u8]) -> bool {
+            input.first_chunk::<4>() == Some(&C::SELECTOR)
+                && input.len()
+                    == 4 + <C::Parameters<'_> as SolType>::ENCODED_SIZE.unwrap_or_default()
```
0xalpharush: Applies to `is_payment` as well, but think we should avoid unwrap_or_default as it makes adding dynamic types incorrect.

```
  let Some(encoded_size) = <C::Parameters<'_> as SolType>::ENCODED_SIZE else {
      return false;
  };
```

0xalpharush: I think this actually makes it a compile-time error. Not really sure if it's better perf-wise
```
macro_rules! is_call {
    ($input:expr, $call:ty) => {{
        const CALLDATA_LEN: usize =
            match <<$call as SolCall>::Parameters<'_> as SolType>::ENCODED_SIZE {
                Some(len) => 4 + len,
                None => panic!("must have statically-sized ABI parameters"),
            };

        $input.first_chunk::<4>() == Some(&<$call as SolCall>::SELECTOR)
            && $input.len() == CALLDATA_LEN
    }};
}
```

0xKitsune: > we should avoid unwrap_or_default as it makes adding dynamic types incorrect.

Updated to apply your suggestion. Maybe we can consider adding the macro as a follow-on to keep the diff/scope of this PR small.

klkvr (approved):

mattsse (approved):
Approved.

crates/revm/src/handler.rs
```diff
@@ -1452,13 +1453,17 @@ where
...
         let basefee = u128::from(context.block().basefee());
-        let effective_gas_price = tx.effective_gas_price(basefee);
+        let mut effective_gas_price = tx.effective_gas_price(basefee);
         let gas = exec_result.gas();
+        let gas_used = gas.used().saturating_sub(gas.reservoir());
+        if context.cfg.spec.is_t6() && tx.is_discounted_payment() && gas_used <= SSTORE_SET_COST {
+            // For pure TIP-20 payment transactions, patch the transaction-derived effective gas
+            // price with TIP-1059's post-execution settlement discount before calculating actual spending:
+            // https://github.com/tempoxyz/tempo/blob/main/tips/tip-1059.md#applying-the-discount
+            effective_gas_price = u128::from(TEMPO_T6_DISCOUNTED_PAYMENT_GAS_PRICE);
```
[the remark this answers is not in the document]
legion2002: addressed in https://github.com/tempoxyz/tempo/pull/4139/changes/4c91d37278572aa533af55613cd18af65f4f3986

crates/chainspec/src/constants.rs
```diff
@@ -28,6 +28,26 @@ pub mod gas {
...
     pub const TEMPO_T1_BASE_FEE: u64 = 20_000_000_000;
 
+    /// TIP-1059 discounted gas price for pure payment transfers: 12 billion attodollars.
+    pub const TEMPO_T6_DISCOUNTED_PAYMENT_GAS_PRICE: u64 = 12_000_000_000;
+
+    /// TIP-1059 pure-payment discount from the T6 base fee.
+    ///
+    /// T6 inherits the T1 base fee of 20 billion attodollars/gas and discounts eligible pure
+    /// payments to 12 billion attodollars/gas, so the discount is 8 billion attodollars/gas.
+    pub const TEMPO_T6_PAYMENT_GAS_PRICE_DISCOUNT: u128 = 8_000_000_000;
```
mattsse, line 38: this works, we could enforce that TEMPO_T1_BASE_FEE is always TEMPO_T6_DISCOUNTED_PAYMENT_GAS_PRICE + TEMPO_T6_PAYMENT_GAS_PRICE_DISCOUNT

maybe even add a samll dummy test just for this condition

crates/chainspec/src/constants.rs
```diff
@@ -28,6 +28,26 @@ pub mod gas {
...
+
+    /// Returns the TIP-1059 discounted effective gas price.
+    ///
+    /// TIP-1059 subtracts the base-fee discount from the transaction-derived effective gas price.
+    /// Any effective priority-fee component remains payable, so high-priority bids cannot be
+    /// refunded away at settlement.
+    pub fn tempo_t6_discounted_payment_effective_gas_price(
```
mattsse, line 47: we could clarify here that this works because the discount is < TEMPO_T1_BASE_FEE
and this is equivalent to computing the effective gas price with a base fee of TEMPO_T6_DISCOUNTED_PAYMENT_GAS_PRICE

perhaps could also add a sanity test using:
https://github.com/alloy-rs/alloy/blob/3ac1a04a19b7fa2029c0d6ebb42452a80cbaf433/crates/eips/src/eip1559/helpers.rs#L14-L14

mattsse (approved):

crates/chainspec/src/constants.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/pull/4139/commits/fa8b059009ff6400e7ac445ac945c6d0b9f2a573

crates/chainspec/src/constants.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/pull/4139/commits/fa8b059009ff6400e7ac445ac945c6d0b9f2a573

PR #4140, opened 2026-05-21
feat(tip-1028): impl latest spec changes

0xrusowsky:
this PR implements the latest changes in the spec, mainly around:
- `burnBlockedProof`
- `releaseBlockedFunds`

it also gets rid of the impl/internal `enum recoveryAuthority` as it became overlapping/redundant now that we have `RecoveryMode` which is stored.

### DIFF
--- a/crates/contracts/src/precompiles/receive_policy_guard.rs
+++ b/crates/contracts/src/precompiles/receive_policy_guard.rs
@@ -72,6 +72,19 @@ crate::sol! {
         /// @param amount Amount of funds released.
         event ReceiptClaimed(address indexed token, address indexed receiver, uint8 receiptVersion, uint64 indexed blockedNonce, uint64 blockedAt, address originator, address recipient, address recoveryAuthority, address caller, address to, uint256 amount);
 
+        /// @notice Emitted when blocked funds are burned with a valid receipt.
+        /// @param token TIP-20 token burned by the guard.
+        /// @param receiver Resolved account where funds would settle; for virtual recipients its their master.
+        /// @param receiptVersion Claim receipt layout version.
+        /// @param blockedNonce Guard nonce from the burned receipt.
+        /// @param blockedAt Block timestamp from the burned receipt.
+        /// @param originator Original sender/originator from the burned receipt.
+        /// @param recipient Addressed recipient from the burned receipt; may be virtual.
+        /// @param recoveryAuthority Claim authority from the burned receipt.
+        /// @param caller Account that submitted the burn.
+        /// @param amount Amount of funds burned.
+        event ReceiptBurned(address indexed token, address indexed receiver, uint8 receiptVersion, uint64 indexed blockedNonce, uint64 blockedAt, address originator, address recipient, address recoveryAuthority, address caller, uint256 amount);
+
         error InvalidReceipt();
         error InvalidClaimAddress();
         error UnauthorizedClaimer();
@@ -148,6 +161,26 @@ impl IReceivePolicyGuard::ClaimReceiptV1 {
             blockedReason: self.blockedReason,
             recoveryAuthority: self.recoveryAuthority,
             memo: self.memo,
+        })
+    }
+
+    pub fn burned_event(
+        &self,
+        receiver: Address,
+        caller: Address,
+        amount: U256,
+    ) -> ReceivePolicyGuardEvent {
+        ReceivePolicyGuardEvent::ReceiptBurned(IReceivePolicyGuard::ReceiptBurned {
+            token: self.token,
+            receiver,
+            receiptVersion: self.version,
+            blockedNonce: self.blockedNonce,
+            blockedAt: self.blockedAt,
+            originator: self.originator,
+            recipient: self.recipient,
+            recoveryAuthority: self.recoveryAuthority,
+            caller,
+            amount,
         })
     }
 }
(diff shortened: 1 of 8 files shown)


crates/contracts/src/precompiles/receive_policy_guard.rs
```diff
@@ -91,6 +92,13 @@ impl IReceivePolicyGuard::ClaimProofV1 {
         kind: IReceivePolicyGuard::InboundKind,
         memo: B256,
     ) -> Self {
+        // Ensure proof doesn't use the `address(0)` sentinel
+        let recovery_auth = if recovery_address.is_zero() {
+            originator
+        } else {
+            recovery_address
+        };
+
```
0xKitsune: IIUC looks like this diverges from the spec. `address(0)` should stay encoded in the proof for originator recovery, but this rewrites it to the concrete `originator`.

This changes what gets stored/emitted, and later code has to guess the recovery type by comparing addresses instead of reading it directly from `recoveryAuthority`.

Can we preserve `address(0)` in the proof and branch on the encoded `recoveryAuthority` directly?

crates/precompiles/src/receive_policy_guard/mod.rs
```diff
@@ -54,6 +55,41 @@ impl RecoveryMode {
...
+            Self::Receiver | Self::ThirdParty => receiver,
+        }
+    }
+
+    /// Returns whether claiming to `to` redirects funds away from the receiver.
+    pub(crate) fn is_reroute(self, to: Address, receiver: Address) -> bool {
+        match self {
+            Self::Receiver => to != receiver,
+            Self::Originator | Self::ThirdParty => true,
+        }
```
0xKitsune, line 85: IIUC this makes third-party recovery claims to the receiver take the reroute path.

The spec says a claim resumes when `recoveryAuthority != originator` and `to == receiver`, so a nonzero recovery authority claiming back to the receiver should be a resume rather not a reroute.

We should align if the spec needs updating or if this should be adjusted.

crates/precompiles/src/tip20/mod.rs
```diff
@@ -1242,20 +1251,19 @@ impl TIP20Token {
         self.check_not_paused()?;
         let destination = Recipient::resolve(to)?;
         destination.validate()?;
+        self.ensure_transfer_authorized(
+            recovery_mode.policy_subject(originator, receiver),
+            destination.target,
+        )?;
```
0xKitsune: I think this applies the reroute transfer policy check to resume claims too.

For a resume, the spec only requires the receiver to be currently authorized as the destination. This checks both the policy subject as sender and the receiver as recipient, which can reject a valid resume if the receiver is allowed to receive but the policy subject is not allowed to send.

crates/precompiles/src/tip20/mod.rs
```diff
@@ -1242,20 +1251,19 @@ impl TIP20Token {
...
-            let policy_id = self.transfer_policy_id()?;
-            if !registry.is_authorized_as(policy_id, destination.target, AuthRole::recipient())? {
-                return Err(TIP20Error::policy_forbids().into());
-            }
-            if registry
+        if recovery_mode.is_reroute(to, receiver) {
+            if TIP403Registry::new()
                 .validate_receive_policy(self.address, originator, destination.target)?
                 .is_some()
             {
```
0xKitsune, line 1260: For reroutes, I think this should validate the destination receive policy using the policy subject as the sender, not always `originator`.

The spec defines the policy subject as `originator` only for originator recovery, otherwise it is the receiver.

crates/precompiles/src/receive_policy_guard/mod.rs
```diff
@@ -177,35 +217,14 @@ impl ReceivePolicyGuard {
...
-            Self::Receiver(claimer) | Self::Originator(claimer) | Self::Contract(claimer) => {
-                claimer
-            }
-        };
-        if msg_sender != authorized_claimer {
-            return Err(ReceivePolicyGuardError::unauthorized_claimer().into());
-        }
-        Ok(())
-    }
+    Ok((proof, receiver, recovery_mode))
```
0xKitsune: Looks like `burnBlockedProof` is missing the `ProofBurned` event from the spec.

After a successful burn, the spec says the guard should emit `ProofBurned` with the proof fields and amount. This clears the proof balance but does not emit anything from the guard.

crates/contracts/src/precompiles/receive_policy_guard.rs
0xrusowsky: i did this on purpose, cause i think it's a UX improvement.
left suggestions in the spec to adapt this new approach, so we only needs to align with @malleshpai on the way fwd.

crates/precompiles/src/receive_policy_guard/mod.rs
0xrusowsky: forgot to do it, my bad

0xrusowsky: addressed in [`6dbecb2` (this PR)](https://github.com/tempoxyz/tempo/pull/4140/commits/6dbecb22309eaa8893144480f6b1d5632203839e)

crates/precompiles/src/receive_policy_guard/mod.rs
0xrusowsky: yes, i flipped it... addressed here: [`c8bec64` (this PR)](https://github.com/tempoxyz/tempo/pull/4140/commits/c8bec6463dd3b3411219c84a2f17f0adc6b8cd52)

PR #4146, opened 2026-05-21
chore(consensus_context): cleanup post-t4 activation

hamdiallam:


crates/commonware-node/src/consensus/application/actor.rs
```diff
@@ -1006,24 +995,20 @@ async fn verify_header(
...
+        proposer: crate::utils::public_key_to_tempo_primitive(proposer),
+    };
 
-        if ctx != expected_ctx {
-            bail!("mismatching block consensus context");
-        }
-    } else if block.header().consensus_context.is_some() {
-        bail!("block consensus context set prior to activation");
+    if ctx != expected_ctx {
+        bail!("mismatching block consensus context");
```
SuperFluffy: While at it, let's a) use `eyre::ensure!` instead of `eyre::bail!`, and also b) report what the expected and actual consensus contexts were (can be useful for debugging).

crates/commonware-node/src/consensus/block.rs
```diff
@@ -153,11 +153,10 @@ impl commonware_consensus::CertifiableBlock for Block {
             None => {
                 // Returns a deterministic sentinel `Context`.
                 //
-                // Pre-T4: Unused; consensus does not consult this context.
-                // Post-T4: All blocks must carry a `consensus_context`, so reaching
-                // this branch indicates a malformed block. The sentinel intentionally
-                // does not match any real consensus values, so it will fail
-                // verification rather than panic.
+                // All consensus-produced blocks must carry a `consensus_context`, so
```
SuperFluffy, line 157: Let's in addition emit a warning here.

crates/e2e/src/tests/consensus_context.rs
```diff
@@ -53,7 +47,6 @@ fn blocks_have_consensus_context() {
             context.sleep(Duration::from_secs(1)).await;
         }
 
-        // Genesis block should not have a consensus context.
```
SuperFluffy, line 56: Why remove? Still seems valid?

crates/e2e/src/tests/consensus_context.rs
```diff
@@ -16,11 +14,7 @@ use crate::{CONSENSUS_NODE_PREFIX, Setup, setup_validators};
 fn blocks_have_consensus_context() {
     let _ = tempo_eyre::install();
 
-    let setup = Setup::new()
-        .how_many_signers(4)
-        .epoch_length(100)
-        .t4_time(0)
```
SuperFluffy, line 22: Remove `.t4_time` in the setup builder and hard code it to always be active?

SuperFluffy (approved):
Bunch of flyby comments to potentially improve this lib.

crates/e2e/src/tests/consensus_context.rs
hamdiallam: yea it was, codex too eager

PR #4151, opened 2026-05-22
fix(tip-1028): address recovery review comments

decofe:
## Summary
- preserve the originator recovery sentinel in claim receipts and resolve the authorized claimer from the encoded recovery authority
- treat claims back to the receiver as resumes, with recipient-only transfer policy validation
- validate reroute receive policies using the recovery policy subject and emit ProofBurned for burns

## Tests
- cargo test -p tempo-precompiles receive_policy_guard --lib
- cargo test -p tempo-precompiles tip1028_tests --lib


0xrusowsky:
already addressed

PR #4157, opened 2026-05-22
perf(txpool): add best transaction size hints

mattsse:
Adds `size_hint` implementations for Tempo's best-transaction iterators so callers can get bounded upper hints when updates are disabled.

The AA 2D iterator accounts for expiring nonce transactions separately, and the merge/state-aware wrappers forward conservative bounds from their inner iterators.

### DIFF
--- a/crates/transaction-pool/src/best.rs
+++ b/crates/transaction-pool/src/best.rs
@@ -86,6 +86,19 @@ impl Iterator for MergeBestTransactions {
     fn next(&mut self) -> Option<Self::Item> {
         self.next_best().map(|(tx, _)| tx)
     }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let cached = usize::from(self.next_protocol_pool.is_some())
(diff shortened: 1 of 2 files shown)


crates/transaction-pool/src/best.rs
```diff
@@ -86,6 +86,19 @@ impl Iterator for MergeBestTransactions {
...
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let cached = usize::from(self.next_protocol_pool.is_some())
+            .saturating_add(usize::from(self.next_aa_2d_pool.is_some()));
+        let (_, protocol_upper) = self.protocol_pool.size_hint();
+        let (_, aa_2d_upper) = self.aa_2d_pool.size_hint();
+        let upper = protocol_upper
+            .zip(aa_2d_upper)
+            .and_then(|(protocol, aa_2d)| protocol.checked_add(aa_2d))
+            .and_then(|upper| upper.checked_add(cached));
```
mattsse, line 98: actually, this is a bit of a mess

PR #4166, opened 2026-05-22
feat(tip-1028): impls spec-defined performant layout to minimize SLOADs

0xrusowsky:


crates/precompiles/src/tip403_registry/mod.rs
```diff
@@ -773,9 +811,10 @@ impl TIP403Registry {
     }
 
     /// Ensures `policy_id` is a built-in or an existing simple policy.
-    fn validate_receive_policy_id(&self, policy_id: u64) -> Result<()> {
+    /// Returns the policy type so that the caller can use it.
+    fn validate_receive_policy_id(&self, policy_id: u64) -> Result<u8> {
         if self.builtin_authorization(policy_id).is_some() {
-            return Ok(());
+            return Ok(policy_id as u8); // safe downcast as it's either 0 or 1.
```
legion2002, line 817: This feels a bit flaky, could be a footgun if we add some new built in policy type in the future at some bigger address.
Will add some cheap explicit checks here

0xrusowsky: this is safe cause we validate in the line above. that's why the cmnt says it's guaranteed to be 0 or 1

legion2002: ah, built in authorization explicitly checks this. yeah all good then.

legion2002 (approved):
lgtm

klkvr (approved):

PR #4176, opened 2026-05-23
perf(pool): batch expired tx membership checks

mattsse:
Adds a Tempo pool helper for expired local tracking entries that checks the AA 2D pool first under one read lock, then falls back to the protocol pool for misses.

This avoids read-locking the AA pool once per expired hash and skips protocol lookups for the common AA 2D case.

### DIFF
--- a/crates/transaction-pool/src/tempo_pool.rs
+++ b/crates/transaction-pool/src/tempo_pool.rs
@@ -70,6 +70,34 @@ impl<Client> TempoTransactionPool<Client> {
             aa_2d_pool: Arc::new(RwLock::new(aa_2d_pool)),
         }
     }
+
+    /// Retains hashes that are still present in either sub-pool.
+    ///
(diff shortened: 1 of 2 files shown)


crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -70,6 +70,34 @@ impl<Client> TempoTransactionPool<Client> {
...
+            }
+        }
+
+        checked
+            .into_iter()
+            .filter_map(|(hash, in_aa_2d_pool)| {
+                (in_aa_2d_pool || self.protocol_pool.contains(&hash)).then_some(hash)
+            })
+            .collect()
+    }
```
klkvr, line 84: can we use an actual retain here instead of allocating a new vec for results?

mattsse:
closing in favor https://github.com/paradigmxyz/reth/pull/24482

PR #4179, opened 2026-05-23
perf(transaction-pool): dispatch log decoding by topic

mattsse:
Dispatches `TempoPoolUpdates` log parsing on `topic0` with small local event enums before decoding generated event structs. This avoids the allocating invalid-signature error path for unrelated logs while preserving the same update extraction behavior.

The previous implementation would always end up allocating an error on mismatch
https://github.com/alloy-rs/core/pull/1109


DaniPopes:
Why don't we use the generated Events enum?

mattsse:
@DaniPopes because we dont need all variants and can avoid decoding those

DaniPopes (approved):

PR #4181, opened 2026-05-23
refactor(pool): reuse effective fee token helper

mattsse:
Uses `effective_fee_token()` when building keychain subjects so fee-token fallback and resolved-token handling stay in one place.

### DIFF
--- a/crates/transaction-pool/src/transaction.rs
+++ b/crates/transaction-pool/src/transaction.rs
@@ -46,11 +46,6 @@ pub struct TempoPooledTransaction {
     /// `Some(expiry)` for keychain transactions where expiry < u64::MAX (finite expiry).
     /// `None` for non-keychain transactions or keys that never expire.
     key_expiry: OnceLock<Option<u64>>,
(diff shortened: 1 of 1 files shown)


crates/transaction-pool/src/transaction.rs
```diff
@@ -151,12 +157,10 @@ impl TempoPooledTransaction {
...
-        let key_id = keychain_sig.key_id(&aa_tx.signature_hash()).ok()?;
-        let fee_token = self
-            .resolved_fee_token
-            .get()
-            .copied()
-            .unwrap_or_else(|| self.inner().fee_token().unwrap_or(DEFAULT_FEE_TOKEN));
+        let key_id = (*self
+            .keychain_key_id
+            .get_or_init(|| keychain_sig.key_id(&aa_tx.signature_hash()).ok()))?;
+        let fee_token = self.effective_fee_token();
```
mattsse, line 155: driveby simplification

DaniPopes (approved):

PR #4192, opened 2026-05-25
chore(precompiles): add `CheckedMath` trait to avoid err mapping

0xrusowsky:
this PR introduces a new `CheckedMath` trait, which provides built-in versions of the common `checked_<add/sub/mul/div>(rhs).ok_or_else(TempoPrecompileError::under_overflow())?;` pattern.

call sites are now much more compact and we can directly propagate the errors with `try_<add/sub/mul/div>?;`

additionally, it changes the return type of `fn base_to_quote` from `Option<u128>` to `Result<u128>` so that we can propagate directly without passing `TempoPrecompileError::under_overflow()`

### DIFF
--- a/crates/precompiles/src/stablecoin_dex/mod.rs
+++ b/crates/precompiles/src/stablecoin_dex/mod.rs
@@ -19,8 +19,8 @@ use tempo_contracts::precompiles::PATH_USD_ADDRESS;
 pub use tempo_contracts::precompiles::{IStablecoinDEX, StablecoinDEXError, StablecoinDEXEvents};
 
 use crate::{
-    STABLECOIN_DEX_ADDRESS,
-    error::{Result, TempoPrecompileError},
+    CheckedMath, STABLECOIN_DEX_ADDRESS,
+    error::Result,
     stablecoin_dex::orderbook::{MAX_PRICE, MIN_PRICE, compute_book_key},
     storage::{Handler, Mapping},
     tip20::{ITIP20, TIP20Token, validate_usd_currency},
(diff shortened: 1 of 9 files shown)


crates/precompiles/src/stablecoin_dex/mod.rs
```diff
@@ -499,8 +485,7 @@ impl StablecoinDEX {
         // Calculate escrow amount and token based on order side
         let (escrow_token, escrow_amount, non_escrow_token) = if is_bid {
             // For bids, escrow quote tokens based on price
-            let quote_amount = base_to_quote(amount, tick, RoundingDirection::Up)
-                .ok_or(StablecoinDEXError::insufficient_balance())?;
+            let quote_amount = base_to_quote(amount, tick, RoundingDirection::Up)?;
```
mattsse: This needs extra care because this changes the revert surface. Before this PR, bid escrow conversion overflow was explicitly mapped to `StablecoinDEXError::insufficient_balance()` in `place`, `place_flip`, and `flip_in_place`; with `base_to_quote(...)?` it now propagates `TempoPrecompileError::under_overflow()` / `Panic(UnderOverflow)`. That is observable to callers, and it also changes flip-fill behavior because `fill_order` swallows business errors but propagates system errors on T1A+, so a flip that previously failed to re-place cleanly can now revert the whole swap. Please preserve the explicit insufficient-balance mapping at this escrow boundary, or add tests if the revert change is intentional.

mattsse (changes requested):
This needs a closer pass before merging. The refactor is mostly mechanical, but the DEX escrow conversions are part of the precompile revert surface, and changing these mappings can break expected reverts and flip-order error handling.

PR #4258, opened 2026-05-26
fix: ensure virtual channel payees use operators

0xrusowsky:
ref: `SIGP-283`

### DIFF
--- a/crates/precompiles/src/tip20_channel_reserve/mod.rs
+++ b/crates/precompiles/src/tip20_channel_reserve/mod.rs
@@ -105,7 +105,7 @@ impl TIP20ChannelReserve {
 
     /// Opens a channel and pulls the initial deposit from the payer into reserve.
     ///
-    /// Payees cannot be zero or TIP-20-prefix addresses. Virtual payees require an operator.
+    /// Payees cannot be zero or TIP-20 addresses. Virtual payees require a non-virtual operator.
     /// This prevents channels whose payee cannot receive direct payouts or submit vouchers itself.
     pub fn open(
         &mut self,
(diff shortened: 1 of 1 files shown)


crates/precompiles/src/tip20_channel_reserve/mod.rs
```diff
@@ -104,14 +105,17 @@ impl TIP20ChannelReserve {
...
         &mut self,
         msg_sender: Address,
         call: ITIP20ChannelReserve::openCall,
     ) -> Result<B256> {
-        if call.payee == Address::ZERO || is_tip20_prefix(call.payee) {
+        if call.payee == Address::ZERO
+            || is_tip20_prefix(call.payee)
+            || (call.payee.is_virtual() && call.operator.is_zero())
+        {
             return Err(TIP20ChannelReserveError::invalid_payee().into());
```
mattsse, line 119: makes sense to reuse the same error here

crates/precompiles/src/tip20_channel_reserve/mod.rs
```diff
@@ -104,14 +105,17 @@ impl TIP20ChannelReserve {
...
+    /// This prevents channels whose payee cannot receive direct payouts or submit vouchers itself.
     pub fn open(
         &mut self,
         msg_sender: Address,
         call: ITIP20ChannelReserve::openCall,
     ) -> Result<B256> {
-        if call.payee == Address::ZERO || is_tip20_prefix(call.payee) {
+        if call.payee == Address::ZERO
+            || is_tip20_prefix(call.payee)
+            || (call.payee.is_virtual() && call.operator.is_zero())
```
mattsse: should virtual operators also be rejected here, like

```
    || (call.payee.is_virtual()
        && (call.operator.is_zero() || call.operator.is_virtual()))
```

mattsse (approved):

crates/precompiles/src/tip20_channel_reserve/mod.rs
0xrusowsky: addressed here [`f799bdf` (this PR)](https://github.com/tempoxyz/tempo/pull/4258/commits/f799bdf1fa7f05213d9ee7535e2534fb4b25b138)

PR #4265, opened 2026-05-26
feat(account-keychain): implement admin access keys

legion2002:
Implements TIP-1049 admin access keys on top of the rebased TIP branch. Adds admin key authorization, root/admin keychain permissions, account-bound admin key authorizations, and updates the TIP text for the ABI witness parameter.

Verification:
- cargo check -p tempo-precompiles -p tempo-primitives -p tempo-revm -p tempo-transaction-pool
- cargo test -p tempo-primitives key_authorization
- cargo test -p tempo-precompiles account_keychain::tests::test_t6
- cargo test -p tempo-revm handler::tests::keychain


tips/tip-1049.md
```diff
@@ -140,17 +143,18 @@ Revocation of an admin key does NOT automatically revoke keys that it authorized
...
+- If `admin_account` is omitted, the authorization creates a non-admin key.
+- If `admin_account` is present, the authorization creates an admin key and `admin_account` MUST equal the target account being modified.
+- If `admin_account` is present, `expiry`, `limits`, and `allowed_calls` MUST be omitted or encoded as empty optional values because admin keys carry no restrictions.
+- If `admin_account` is present, `witness` MAY be present or omitted for the transaction-encoded authorization. ABI calls to `authorizeAdminKey` always provide a `bytes32 witness`, including `bytes32(0)`.
 
-The target `account` and `is_admin` fields are part of the signed RLP payload. This binds admin key authorizations to a target account and prevents replaying an unrestricted non-admin authorization as an admin authorization.
+The target `admin_account` and `witness` fields are part of the signed RLP payload. This binds admin key authorizations to a target account and prevents replaying an unrestricted non-admin authorization as an admin authorization.
```
0xrusowsky: this makes sense to simplify

mattsse (commented):
### Review findings

**[P1] Admin-signed non-admin `KeyAuthorization`s are not account-bound**

[`admin_account` is only present when the authorization creates an admin key](https://github.com/tempoxyz/tempo/blob/6c1ca05cbbce70f9c358e96b6c1d311f8f82cf42/crates/primitives/src/transaction/key_authorization.rs#L221-L225), while the T6 signer validation accepts any signer that is currently admin for `tx.caller` ([validation path](https://github.com/tempoxyz/tempo/blob/6c1ca05cbbce70f9c358e96b6c1d311f8f82cf42/crates/revm/src/handler.rs#L1266-L1284)) and only enforces account binding when `admin_account` is present ([admin-only binding check](https://github.com/tempoxyz/tempo/blob/6c1ca05cbbce70f9c358e96b6c1d311f8f82cf42/crates/revm/src/handler.rs#L1727-L1748)). This leaves admin-signed non-admin authorizations replayable across accounts where the same admin key is authorized. If an admin key is admin for accounts A and B, a signature intended to add a non-admin child key to A can also add that same child key to B. The TIP calls out account-bound digests for `verifyAdmin`, but the transaction authorization path needs the same account binding.

**[P1] Admin delegation still runs same-tx auth+use handling**

The fee precheck is documented as same-tx auth+use, but it is not gated on `access_key_addr == key_auth.key_id` ([pre-fee limit check](https://github.com/tempoxyz/tempo/blob/6c1ca05cbbce70f9c358e96b6c1d311f8f82cf42/crates/revm/src/handler.rs#L1196-L1211)). Later, every keychain-signed transaction with a `KeyAuthorization` overwrites the transient transaction key with the newly authorized `key_auth.key_id` ([transient key overwrite](https://github.com/tempoxyz/tempo/blob/6c1ca05cbbce70f9c358e96b6c1d311f8f82cf42/crates/revm/src/handler.rs#L1504-L1527)). For admin delegation, the signer is the existing admin key and `key_auth.key_id` is the child key, so fee/spending checks and `getTransactionKey()` become attributed to a key that did not sign the transaction. These paths should only run for true same-tx auth+use.

**[P2] TIP-required `SignatureVerifier.verifyAdmin` is not implemented**

The TIP specifies `verifyAdmin(account, digest, signature)` as part of the feature ([TIP text](https://github.com/tempoxyz/tempo/blob/6c1ca05cbbce70f9c358e96b6c1d311f8f82cf42/tips/tip-1049.md#L87-L106)), but `ISignatureVerifier` still exposes only `recover` and `verify` ([interface](https://github.com/tempoxyz/tempo/blob/6c1ca05cbbce70f9c358e96b6c1d311f8f82cf42/crates/contracts/src/precompiles/signature_verifier.rs#L12-L22)), and dispatch still only handles those two calls ([dispatch](https://github.com/tempoxyz/tempo/blob/6c1ca05cbbce70f9c358e96b6c1d311f8f82cf42/crates/precompiles/src/signature_verifier/dispatch.rs#L28-L33)). If `verifyAdmin` is intended to ship in this TIP, the ABI, selector gating, stateful implementation, and coverage tests are missing.

**[P2] Admin signer state reads are not reflected in gas or pool invalidation**

T6 admin-signed authorizations add an `is_admin_key_for(tx.caller, auth_signer)` state read ([state read](https://github.com/tempoxyz/tempo/blob/6c1ca05cbbce70f9c358e96b6c1d311f8f82cf42/crates/revm/src/handler.rs#L1274-L1282)), but key-authorization intrinsic gas still prices only the existing key check/write and optional limits/witness/scope costs ([gas model](https://github.com/tempoxyz/tempo/blob/6c1ca05cbbce70f9c358e96b6c1d311f8f82cf42/crates/revm/src/handler.rs#L322-L381)). The transaction pool also only caches the returned `key_expiry` for expiry eviction ([expiry cache](https://github.com/tempoxyz/tempo/blob/6c1ca05cbbce70f9c358e96b6c1d311f8f82cf42/crates/transaction-pool/src/validator.rs#L421-L440)), while revoked-key invalidation is keyed off the outer keychain signature subject ([subject extraction](https://github.com/tempoxyz/tempo/blob/6c1ca05cbbce70f9c358e96b6c1d311f8f82cf42/crates/transaction-pool/src/transaction.rs#L152-L160)). For root-signed transactions carrying an admin-signed `KeyAuthorization`, the admin signer can expire or be revoked after pool admission without matching the existing keychain-subject eviction path, leaving stale transactions in the pool until inclusion-time validation.

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -217,6 +220,9 @@ impl AccountKeychain {
         if key_id == Address::ZERO {
             return Err(AccountKeychainError::zero_public_key().into());
         }
+        if self.storage.spec().is_t6() && key_id == msg_sender {
+            return Err(AccountKeychainError::invalid_key_id().into());
```
0xrusowsky, line 269: nit: requires fn docs update to document the new T6+ err

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -320,6 +327,74 @@ impl AccountKeychain {
...
+        if let Some(witness) = witness {
+            self.ensure_key_authorization_witness_not_burned(msg_sender, witness)?;
+        }
+
+        let signature_type = match signature_type {
+            SignatureType::Secp256k1 => 0,
+            SignatureType::P256 => 1,
+            SignatureType::WebAuthn => 2,
+            _ => return Err(AccountKeychainError::invalid_signature_type().into()),
+        };
```
0xrusowsky: nit: now that we do this in multiple places, it may be worth implementing the `TryFrom` trait

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -380,6 +458,9 @@ impl AccountKeychain {
 
         let current_timestamp = self.storage.timestamp().saturating_to::<u64>();
         let mut key = self.load_active_key(msg_sender, call.keyId, current_timestamp)?;
+        if self.storage.spec().is_t6() && key.is_admin {
```
0xrusowsky: i think technically we don't need the `is_t6()` check here, as pre-T6 `is_admin` is always false

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -494,7 +575,10 @@ impl AccountKeychain {
         self.ensure_admin_caller(msg_sender)?;
 
         let current_timestamp = self.storage.timestamp().saturating_to::<u64>();
-        self.load_active_key(msg_sender, call.keyId, current_timestamp)?;
+        let key = self.load_active_key(msg_sender, call.keyId, current_timestamp)?;
+        if self.storage.spec().is_t6() && key.is_admin {
```
0xrusowsky: same cmnt regarding T6 guard

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -617,6 +704,11 @@ impl AccountKeychain {
         self.key_authorization_witnesses[call.account][call.witness].read()
     }
 
+    /// Returns true for the root key or for an active admin access key.
+    pub fn is_admin_key(&self, call: isAdminKeyCall) -> Result<bool> {
+        self.is_admin_key_for(call.account, call.keyId)
+    }
```
0xrusowsky: nit: let's make `is_admin_key` the internal fn, and flatten the call in the dispatcher

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -521,7 +605,10 @@ impl AccountKeychain {
         self.ensure_admin_caller(msg_sender)?;
 
         let current_timestamp = self.storage.timestamp().saturating_to::<u64>();
-        self.load_active_key(msg_sender, call.keyId, current_timestamp)?;
+        let key = self.load_active_key(msg_sender, call.keyId, current_timestamp)?;
+        if self.storage.spec().is_t6() && key.is_admin {
```
0xrusowsky: same cmnt regarding T6 guard

0xrusowsky (commented):
the ai-audit findings seem legit, we should address them.

additionally, leaving some style/clarity cmnts

crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -216,6 +217,12 @@ pub struct KeyAuthorization {
     /// `None` means no witness. `Some(witness)` means the witness field is present, including when
     /// `witness == B256::ZERO`.
     pub witness: Option<B256>,
+
+    /// Account this admin authorization targets.
+    ///
+    /// `None` means this authorization creates a non-admin access key. `Some(account)` means this
+    /// authorization creates an admin key for `account`.
+    pub admin_account: Option<Address>,
```
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/565513832863362c5a58be82903d37150ad83501

legion2002:
Addressed `[P1] Admin-signed non-admin KeyAuthorization`s are not account-bound` in https://github.com/tempoxyz/tempo/commit/565513832863362c5a58be82903d37150ad83501

crates/revm/src/handler.rs
```diff
@@ -1142,6 +1142,54 @@ where
...
             if let Some(key_auth) = tempo_tx_env.key_authorization.as_ref() {
+                let access_key_addr = if let Some(override_key_id) = tempo_tx_env.override_key_id {
+                    override_key_id
+                } else {
+                    keychain_sig
+                        .key_id(&tempo_tx_env.signature_hash)
+                        .map_err(|_| TempoInvalidTransaction::AccessKeyRecoveryFailed)?
+                };
+
+                if cfg.spec.is_t6() && access_key_addr != key_auth.key_id {
```
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/3209629809aa9b05fb6d8eb40514868bece7da35

legion2002:
Addressed `[P1] Admin delegation still runs same-tx auth+use handling` in https://github.com/tempoxyz/tempo/commit/3209629809aa9b05fb6d8eb40514868bece7da35

legion2002:
Addressed `[P1] Limit key-type equality check to same-tx auth+use` in https://github.com/tempoxyz/tempo/commit/3209629809aa9b05fb6d8eb40514868bece7da35

crates/revm/src/handler.rs
```diff
@@ -1215,6 +1263,36 @@ where
...
+                .map_err(|_| TempoInvalidTransaction::KeyAuthorizationSignatureRecoveryFailed)?;
+            if auth_signer != tx.caller {
+                let is_admin = StorageCtx::enter_precompile(
+                    journal,
+                    block,
+                    cfg,
+                    tx,
+                    |keychain: AccountKeychain| {
+                        keychain
+                            .is_admin_key_for(tx.caller, auth_signer)
```
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/ad60b7f0ea0c4658ccd450250a4474e15b51e66a

legion2002:
Addressed `Admin KeyAuthorization signers bypass the stored signature-type check` in https://github.com/tempoxyz/tempo/commit/ad60b7f0ea0c4658ccd450250a4474e15b51e66a

crates/revm/src/handler.rs
```diff
@@ -1215,6 +1263,36 @@ where
...
 
+        if cfg.spec.is_t6()
+            && let Some(tempo_tx_env) = tx.tempo_tx_env.as_ref()
+            && let Some(key_auth) = tempo_tx_env.key_authorization.as_ref()
+        {
+            let auth_signer = key_auth
+                .recover_signer()
+                .map_err(|_| TempoInvalidTransaction::KeyAuthorizationSignatureRecoveryFailed)?;
+            if auth_signer != tx.caller {
+                let is_admin = StorageCtx::enter_precompile(
```
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/c26384fb9ab348277ace43966ca1a91ee53a72fe

legion2002:
Addressed `T6 admin-key authorization validation performs underpriced state access` in https://github.com/tempoxyz/tempo/commit/c26384fb9ab348277ace43966ca1a91ee53a72fe

crates/precompiles/src/account_keychain/mod.rs
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/f1d73eb17994243f48a81c5acb53acb58be56ea4

legion2002:
Addressed `Restriction mutators do not reject the implicit root/admin key` in https://github.com/tempoxyz/tempo/commit/f1d73eb17994243f48a81c5acb53acb58be56ea4

legion2002:
Addressed `Admin-key revocation does not evict pending transactions that depend on the revoked sidecar signer` in https://github.com/tempoxyz/tempo/commit/e55f560b2e08594c355062dc639a4bc2b2ba461b

crates/precompiles/src/account_keychain/mod.rs
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/7057f9bc4a22c4bf6e266920864879b396a95542

crates/precompiles/src/account_keychain/mod.rs
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/7057f9bc4a22c4bf6e266920864879b396a95542

crates/precompiles/src/account_keychain/mod.rs
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/7057f9bc4a22c4bf6e266920864879b396a95542

crates/precompiles/src/account_keychain/mod.rs
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/7057f9bc4a22c4bf6e266920864879b396a95542

crates/precompiles/src/account_keychain/mod.rs
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/7057f9bc4a22c4bf6e266920864879b396a95542

crates/precompiles/src/account_keychain/mod.rs
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/7057f9bc4a22c4bf6e266920864879b396a95542

crates/revm/src/handler.rs
```diff
@@ -1215,6 +1272,48 @@ where
...
         }
 
+        if cfg.spec.is_t6()
+            && let Some(tempo_tx_env) = tx.tempo_tx_env.as_ref()
+            && let Some(key_auth) = tempo_tx_env.key_authorization.as_ref()
+        {
+            let auth_signer = key_auth
+                .recover_signer()
+                .map_err(|_| TempoInvalidTransaction::KeyAuthorizationSignatureRecoveryFailed)?;
+            if auth_signer != tx.caller {
```
legion2002, line 1267: addressed in https://github.com/tempoxyz/tempo/commit/59b6aa615fb82cc0f09688e60e2f066b6c6dcc07

crates/revm/src/handler.rs
```diff
@@ -1215,6 +1272,48 @@ where
...
+                    tx,
+                    |keychain: AccountKeychain| {
+                        keychain
+                            .validate_keychain_authorization(
+                                tx.caller,
+                                auth_signer,
+                                block.timestamp().to::<u64>(),
+                                Some(key_auth.signature.signature_type().into()),
+                            )
+                            .map_err(|e| EVMError::Custom(e.to_string()))
```
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/f37db297773543cc859b2515e26388ac1829ea1f

crates/transaction-pool/src/transaction.rs
```diff
@@ -161,6 +161,26 @@ impl TempoPooledTransaction {
...
     }
 
+    /// Extracts the keychain subject for the signer of an inline `KeyAuthorization`.
+    ///
+    /// Used for revocation matching: if the access key that signed an inline authorization is
+    /// revoked while the transaction is still in the pool, the transaction must be revalidated.
+    pub fn key_authorization_signer_subject(&self) -> Option<KeychainSubject> {
+        let aa_tx = self.inner().as_aa()?;
+        let key_authorization = aa_tx.tx().key_authorization.as_ref()?;
+        let key_id = key_authorization.recover_signer().ok()?;
```
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/f37db297773543cc859b2515e26388ac1829ea1f

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -208,11 +208,18 @@ where
...
+                    .flatten();
 
                 // Check 1: Revoked keychain keys
                 if !updates.revoked_keys.is_empty()
-                    && let Some(ref subject) = keychain_subject
-                    && subject.matches_revoked(&updates.revoked_keys)
+                    && (keychain_subject
+                        .as_ref()
+                        .is_some_and(|subject| subject.matches_revoked(&updates.revoked_keys))
+                        || key_authorization_subject
```
legion2002, line 228: Addressed in https://github.com/tempoxyz/tempo/commit/f37db297773543cc859b2515e26388ac1829ea1f

mattsse:
Rolled back the `tips/verify/lib/tempo-std` submodule pointer from this PR. The matching AccountKeychain ABI update now lives in the separate tempo-std PR: https://github.com/tempoxyz/tempo-std/pull/113

crates/transaction-pool/src/transaction.rs
```diff
@@ -1259,3 +1315,10 @@ impl KeychainSubject {
         spending_limit_updates.contains(self.account, self.key_id, self.fee_token)
     }
 }
+
+impl KeyAuthorizationTargetSubject {
```
mattsse, line 1356: these two impls blocks are out of order, they're not below their struct defs

crates/primitives/src/transaction/key_authorization.rs
```diff
@@ -455,6 +491,90 @@ mod rlp {
     use super::*;
     use alloy_rlp::{Decodable, Encodable};
 
+    #[derive(
+        Clone, Debug, PartialEq, Eq, Hash, alloy_rlp::RlpEncodable, alloy_rlp::RlpDecodable,
+    )]
+    #[rlp(trailing(canonical))]
```
mattsse, line 543: I believe these new fields are fine with

`[rlp(trailing(canonical))]`

crates/revm/src/handler.rs
mattsse, line 1: side-note, this file's type order is a bit unhinged :D (private fns before the `pub struct TempoEvmHandler<DB, I> {` are driving me insane, but we keep this for now because not worth having two locations now

crates/revm/src/handler.rs
```diff
@@ -143,6 +143,25 @@ fn tempo_signature_verification_gas(signature: &TempoSignature) -> u64 {
     }
 }
 
+fn validate_t6_key_authorization_account_binding(
```
mattsse: needs docs

crates/revm/src/handler.rs
```diff
@@ -1215,6 +1291,53 @@ where
             }
         }
 
+        if cfg.spec.is_t6()
+            && let Some(tempo_tx_env) = tx.tempo_tx_env.as_ref()
+            && let Some(key_auth) = tempo_tx_env.key_authorization.as_ref()
+        {
+            let auth_signer = key_auth
+                .recover_signer()
+                .map_err(|_| TempoInvalidTransaction::KeyAuthorizationSignatureRecoveryFailed)?;
```
mattsse, line 1265: should we really do recovery here?

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -320,6 +356,69 @@ impl AccountKeychain {
...
+                    witness,
+                },
+            ))?;
+        }
+
+        self.emit_event(AccountKeychainEvent::key_authorized(
+            msg_sender,
+            key_id,
+            signature_type,
+            u64::MAX,
```
mattsse: can we add a note here that this never expires?

crates/revm/src/handler.rs
mattsse, line 1: I dont really feel comfortable reviewing these changes, not super familiar with this really

crates/revm/src/handler.rs
```diff
@@ -1215,6 +1291,53 @@ where
             }
         }
 
+        if cfg.spec.is_t6()
```
mattsse, line 1259: I'd be great if we could document why this hardfork specific logic exist/what this does

mattsse (commented):
a few nits

need to defer the handler review to someone familiar with this

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -65,6 +66,42 @@ pub struct AuthorizedKey {
...
+
+#[derive(Debug, Clone, Copy)]
+struct StoredSignatureType(u8);
+
+impl StoredSignatureType {
+    #[inline]
+    const fn as_u8(self) -> u8 {
+        self.0
+    }
+}
```
0xrusowsky: while this works, `Storable` macro now supports enums, so u could turn it into
```rs
#[repr(u8)]
#[derive(Debug, Clone, Copy, PartialEq, Storable)]
enum StoredSignatureType {
   Secp256k,
   P256,
   WebAuthn,
}
```

crates/revm/src/handler.rs
```diff
@@ -1142,11 +1167,62 @@ where
             }
 
             if let Some(key_auth) = tempo_tx_env.key_authorization.as_ref() {
+                let access_key_addr = if let Some(override_key_id) = tempo_tx_env.override_key_id {
+                    override_key_id
+                } else {
+                    keychain_sig
+                        .key_id(&tempo_tx_env.signature_hash)
+                        .map_err(|_| TempoInvalidTransaction::AccessKeyRecoveryFailed)?
+                };
```
0xrusowsky: should this assignment happen inside the T6 guarded block for backwards compat?

crates/revm/src/handler.rs
```diff
@@ -1142,11 +1167,62 @@ where
...
+                if cfg.spec.is_t6() && !same_tx_auth_use {
+                    let stored_key_expiry = StorageCtx::enter_precompile(
+                        journal,
+                        block,
+                        cfg,
+                        tx,
+                        |mut keychain: AccountKeychain| {
+                            let sig_type = spec
+                                .is_t1()
+                                .then_some(keychain_sig.signature.signature_type().into());
```
0xrusowsky: we are inside a T6 block, so this is guaranteed to be T1+

crates/revm/src/handler.rs
```diff
@@ -1142,11 +1167,62 @@ where
             }
 
             if let Some(key_auth) = tempo_tx_env.key_authorization.as_ref() {
```
0xrusowsky: a high-level block explaining what the block does wouldn't hurt

crates/revm/src/handler.rs
0xrusowsky: if we need to, probably safer to delay as much as possible and do it after `validate_t6_key_authorization_account_binding(key_auth, tx.caller)?;`

0xrusowsky: also depending on what is more expensive (keychain precompile validation or signer recovery) we could even consider moving recovery to the very end, only doing it when keychain validation fails

crates/revm/src/handler.rs
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/e743bf375571c154d58224d2ff54d6cbf961ae0e

crates/revm/src/handler.rs
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/e743bf375571c154d58224d2ff54d6cbf961ae0e

crates/revm/src/handler.rs
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/e743bf375571c154d58224d2ff54d6cbf961ae0e

crates/revm/src/handler.rs
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/1b29d6bfa2db138eb5f33254b6985de3b460f14f

crates/revm/src/handler.rs
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/e743bf375571c154d58224d2ff54d6cbf961ae0e

crates/revm/src/handler.rs
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/e743bf375571c154d58224d2ff54d6cbf961ae0e

crates/precompiles/src/account_keychain/mod.rs
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/8d26f0ae9b3fb9daca317f41ffc42d57e7c8201c

legion2002: we can probably also do this in a bunch of other places in the code now, so we don't have to do `as u8` conversions everywhere. But can do it in a follow up PR

crates/transaction-pool/src/transaction.rs
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/362217650c44fee2d74d0e510e8f826d8feae4d9

crates/precompiles/src/account_keychain/mod.rs
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/362217650c44fee2d74d0e510e8f826d8feae4d9

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -313,13 +347,77 @@ impl AccountKeychain {
...
+            signature_type as u8,
             config.expiry,
         ))?;
 
         Ok(())
     }
 
+    /// Registers a new unrestricted admin access key. Only newly authorized key IDs can become
+    /// admin keys; existing or previously revoked keys must not be upgraded in place.
+    pub fn authorize_admin_key(
```
klkvr, line 378: can we unify this with `authorize_key`?

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -217,6 +253,9 @@ impl AccountKeychain {
         if key_id == Address::ZERO {
             return Err(AccountKeychainError::zero_public_key().into());
         }
+        if self.storage.spec().is_t6() && key_id == msg_sender {
+            return Err(AccountKeychainError::invalid_key_id().into());
+        }
```
klkvr: i guess this seems reasonable but not entirely sure why we'd only add this check now

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -339,6 +437,9 @@ impl AccountKeychain {
     /// - `KeyNotFound` — no key registered with this ID
     pub fn revoke_key(&mut self, msg_sender: Address, call: revokeKeyCall) -> Result<()> {
         self.ensure_admin_caller(msg_sender)?;
+        if self.storage.spec().is_t6() && call.keyId == msg_sender {
+            return Err(AccountKeychainError::invalid_key_id().into());
+        }
```
klkvr: not sure if this makes sense? seems fine to allow revoking any key as long as it's authorized

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -378,8 +479,15 @@ impl AccountKeychain {
     ) -> Result<()> {
         self.ensure_admin_caller(msg_sender)?;
 
+        if self.storage.spec().is_t6() && call.keyId == msg_sender {
+            return Err(AccountKeychainError::invalid_key_id().into());
+        }
```
klkvr: same here

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -1012,6 +1129,17 @@ impl AccountKeychain {
...
+    /// Internal predicate for root/admin status.
+    pub fn is_admin_key(&self, account: Address, key_id: Address) -> Result<bool> {
+        if key_id == account {
+            return Ok(true);
+        }
+
+        let current_timestamp = self.storage.timestamp().saturating_to::<u64>();
+        let key = self.keys[account][key_id].read()?;
+        Ok(key.expiry != 0 && !key.is_revoked && current_timestamp < key.expiry && key.is_admin)
+    }
```
klkvr, line 1115: ideally we unify those checks with what's happening `load_active_key`

crates/revm/src/handler.rs
```diff
@@ -1219,6 +1310,56 @@ where
...
         }
 
+        // T6 defers `KeyAuthorization` signer checks from `validate_env` to this state-aware phase:
+        // root-signed authorizations remain valid, while admin-signed authorizations must name the
+        // caller account and prove that the sidecar signer is an active admin key for that account.
+        if cfg.spec.is_t6()
+            && let Some(tempo_tx_env) = tx.tempo_tx_env.as_ref()
+            && let Some(key_auth) = tempo_tx_env.key_authorization.as_ref()
+        {
+            validate_t6_key_authorization_account_binding(key_auth, tx.caller)?;
```
klkvr: this check is stateless so it's not obvious why we'd defer it?

crates/revm/src/handler.rs
```diff
@@ -363,6 +387,12 @@ fn calculate_key_authorization_gas(
         let sstore_cost = gas_params.get(GasId::sstore_set_without_load_cost());
         let mut total_gas = sig_gas + sload_cost + sstore_cost * num_sstores + BUFFER;
 
+        // T6 account-bound authorizations may be signed by an existing admin key instead of the
+        // root key. Charge one worst-case cold read for validating that admin signer row.
+        if spec.is_t6() && (key_auth.is_admin || key_auth.account.is_some()) {
+            total_gas += sload_cost;
+        }
```
klkvr: should this just check `account.is_some()`?

basically if `is_admin == true` and `account.is_none()` the key auth is either signed by the access key (and the tx is invalid) or by the root key and then there's no extra SLOAD to charge for

crates/revm/src/handler.rs
```diff
@@ -1146,11 +1176,72 @@ where
...
+
+                    let same_tx_auth_use = access_key_addr == key_auth.key_id;
+
+                    if !same_tx_auth_use {
+                        let stored_key_expiry = StorageCtx::enter_precompile(
+                            journal,
+                            block,
+                            cfg,
+                            tx,
+                            |mut keychain: AccountKeychain| {
```
klkvr: i find it a bit confusing that we now call `validate_keychain_authorization` in 3 different places instead of 1

feels like this could be done in a more pipelined manner? i.e
1. derive `same_tx_auth_use`
2. validate + load the key
3. check that key is an admin key, etc

crates/revm/src/handler.rs
```diff
@@ -1219,6 +1310,56 @@ where
...
+                .recover_signer()
+                .map_err(|_| TempoInvalidTransaction::KeyAuthorizationSignatureRecoveryFailed)?;
+
+            if auth_signer != tx.caller {
+                if key_auth.account.is_none() {
+                    return Err(TempoInvalidTransaction::KeychainValidationFailed {
+                        reason: "admin-signed key authorization account mismatch".to_string(),
+                    }
+                    .into());
+                }
```
klkvr: this also belongs to stateless checks imo

crates/revm/src/handler.rs
```diff
@@ -1219,6 +1310,56 @@ where
...
+
+                let admin_key = StorageCtx::enter_precompile(
+                    journal,
+                    block,
+                    cfg,
+                    tx,
+                    |keychain: AccountKeychain| {
+                        keychain
+                            .validate_keychain_authorization(
```
klkvr: looks like this `validate_keychain_authorization` will get executed even if we already did it above? for cases when tx is signed by admin key and carries a key auth that authorizes a different key

crates/revm/src/handler.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/pull/4265/commits/d224c8890fc323b0fa832ca1a4e8c39601d961d3

crates/revm/src/handler.rs
```diff
@@ -363,6 +387,12 @@ fn calculate_key_authorization_gas(
         let sstore_cost = gas_params.get(GasId::sstore_set_without_load_cost());
         let mut total_gas = sig_gas + sload_cost + sstore_cost * num_sstores + BUFFER;
 
+        // T6 account-bound authorizations may be signed by an existing admin key instead of the
+        // root key. Charge one worst-case cold read for validating that admin signer row.
+        if spec.is_t6() && (key_auth.is_admin || key_auth.account.is_some()) {
```
legion2002: Addressed in 6174d3cfc.

crates/precompiles/src/account_keychain/mod.rs
legion2002: imo better to have it completely separate, as it is a simple critical function with not a lot of related code, so feels cleaner to have it in a separate fn dispatch.

crates/precompiles/src/account_keychain/mod.rs
legion2002: TIP 1049 specifies this. ig it is an artifact of the fact that initially we also wanted this TIP to revoke the root key.
So we wanted to make sure that it could not authorize itself in different ways.

We don't have that root revocation in the TIP anymore, but think it's good to include this for the future anyways?

crates/precompiles/src/account_keychain/mod.rs
legion2002: yeah we can remove the revocation check here.

crates/revm/src/handler.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/6d6ea7879b390e45a2e76a922e4aeda64c2eddb4

crates/revm/src/handler.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/6d6ea7879b390e45a2e76a922e4aeda64c2eddb4

crates/precompiles/src/account_keychain/mod.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/76c7dd81f93beeda0e9910b66533b6c098fd923a

crates/precompiles/src/account_keychain/mod.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/a373e664d1ae71e38d758159057130121f4924d1

crates/revm/src/handler.rs
legion2002: addressed in https://github.com/tempoxyz/tempo/commit/f42486e6b69b2b1c1e966c3c139f4473caa16ef0

crates/precompiles/src/account_keychain/mod.rs
klkvr: given that it can be bypassed before the fork it seems that we if we were to add this kind of check it belongs to tx validation? i.e ensure that `keychain_sig.key_id != tx.caller`

not sure if it makes sense to me to add it right now, so i'd vote for removing to scope the PR down

crates/precompiles/src/account_keychain/mod.rs
klkvr: well pretty much everything this fn does is already being done in `authorize_key` besides the `AdminKeyAuthorized` event

i guess my biggest worry is that if we add a new validation for authorized keys that happens to also be relevant for admin keys this fn might be missed

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -1012,6 +1118,26 @@ impl AccountKeychain {
...
+        let current_timestamp = self.storage.timestamp().saturating_to::<u64>();
+        let key = match self.load_active_key(account, key_id, current_timestamp) {
+            Ok(key) => key,
+            Err(crate::error::TempoPrecompileError::AccountKeychainError(
+                AccountKeychainError::KeyAlreadyRevoked(_)
+                | AccountKeychainError::KeyNotFound(_)
+                | AccountKeychainError::KeyExpired(_),
+            )) => return Ok(false),
+            Err(err) => return Err(err),
+        };
```
klkvr, line 1112: let's use `err.is_system_error()` here

crates/revm/src/handler.rs
```diff
@@ -1145,40 +1165,37 @@ where
...
+                    access_key_addr == key_auth.key_id
                 } else {
-                    // Get the access key address (recovered during pool validation and cached)
-                    keychain_sig
-                        .key_id(&tempo_tx_env.signature_hash)
-                        .map_err(|_| TempoInvalidTransaction::AccessKeyRecoveryFailed)?
-                };
+                    true
+                }
+            });
```
```suggestion
            let same_tx_auth_use = key_auth.is_some_and(|key_auth| {
                access_key_addr == key_auth.key_id
            });
```

i think this should be enough? pre-T6 `validate_env` properly ensures that all auths are for the same key

crates/revm/src/handler.rs
```diff
@@ -1145,40 +1165,37 @@ where
...
 
-                // If this transaction is using an already-authorized key, validate the signature against stored key and set the transient key_id.
-                let stored_key_expiry = StorageCtx::enter_precompile(
+            let should_validate_existing_tx_key =
+                key_auth.is_none() || (cfg.spec.is_t6() && !same_tx_auth_use);
```
klkvr: that's basically just `!same_tx_auth_use`

crates/revm/src/handler.rs
```diff
@@ -1205,18 +1221,111 @@ where
                                 reason: format!("{e:?}"),
                             })?;
 
+                        // T6 adds admin delegation: a keychain signer may authorize a different
+                        // child key only if the acting transaction key is itself an active admin key.
+                        if key_auth.is_some() && !same_tx_auth_use && !key.is_admin {
```
```suggestion
                        if key_auth.is_some() && !key.is_admin {
```

crates/revm/src/handler.rs
```diff
@@ -1205,18 +1221,111 @@ where
...
                             .map_err(|e| EVMError::Custom(e.to_string()))?;
 
-                        Ok::<_, EVMError<_, TempoInvalidTransaction>>(key.expiry)
+                        Ok::<_, EVMError<_, TempoInvalidTransaction>>((
+                            key.expiry,
+                            LoadedTxAccessKey {
+                                key_id: access_key_addr,
+                                is_admin: key.is_admin,
+                                signature_type: key.signature_type as u8,
+                            },
```
klkvr: why not just put entire `AuthorizedKey` into `LoadedTxAccessKey` so that we don't have to pass this tuple with expiry around

crates/revm/src/handler.rs
```diff
@@ -1205,18 +1221,111 @@ where
...
+                        .iter()
+                        .rev()
+                        .find(|limit| limit.token == fee_token)
+                        .map(|limit| limit.limit)
+                        .unwrap_or_default();
+
+                    if gas_balance_spending > remaining {
+                        return Err(
+                            FeePaymentError::Other("SpendingLimitExceeded".to_string()).into()
+                        );
```
klkvr: would be easier to reason about it if it was the else branch of the if above that explicitly matches on `same_tx_auth_use`. also imo we should just `.unwrap()` the key_auth here

crates/revm/src/handler.rs
```diff
@@ -1205,18 +1221,111 @@ where
...
+        }
+
+        // T6 stateless signer/account checks run in `validate_env`. This state-aware phase only
+        // proves that a non-root sidecar signer is an active admin key for the caller account.
+        if cfg.spec.is_t6()
+            && let Some(tempo_tx_env) = tx.tempo_tx_env.as_ref()
+            && let Some(key_auth) = tempo_tx_env.key_authorization.as_ref()
+        {
+            let auth_signer = key_authorization_signer
+                .expect("T6 key authorization signer is set during validate_env");
```
klkvr: should we just have `OnceLock` cache for the signer on key auth?

crates/revm/src/handler.rs
klkvr: that's also a kind of weird requirement, curious what's the motivation? if i signed a key auth with my root key it seems fine to let any other key apply it? whether it's admin or not

crates/revm/src/handler.rs
```diff
@@ -1421,30 +1539,48 @@ where
...
+                        } else {
+                            keychain_sig
+                                .key_id(&tempo_tx_env.signature_hash)
+                                .map_err(|_| TempoInvalidTransaction::AccessKeyRecoveryFailed)?
+                        };
+
+                    access_key_addr == key_auth.key_id
+                } else {
+                    true
+                };
```
klkvr: let's just reuse the bool from above?

crates/revm/src/handler.rs
legion2002: yeah I think we can do this. Initially we just wanted to keep things locked down, so it was out of an abundance of caution where we only wanted root keys to be using key auths ( or same tx auth usecase )

crates/revm/src/handler.rs
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/46a95d47604afc9174e9282a7aedcb5e0442bb3a.

crates/revm/src/handler.rs
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/46a95d47604afc9174e9282a7aedcb5e0442bb3a.

crates/revm/src/handler.rs
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/46a95d47604afc9174e9282a7aedcb5e0442bb3a.

crates/revm/src/handler.rs
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/46a95d47604afc9174e9282a7aedcb5e0442bb3a and cleaned up in https://github.com/tempoxyz/tempo/commit/775fd023371f9764fde3e386cb79ae4d80680f28.

crates/revm/src/handler.rs
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/775fd023371f9764fde3e386cb79ae4d80680f28.

crates/revm/src/handler.rs
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/775fd023371f9764fde3e386cb79ae4d80680f28.

crates/revm/src/handler.rs
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/775fd023371f9764fde3e386cb79ae4d80680f28.

crates/revm/src/handler.rs
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/775fd023371f9764fde3e386cb79ae4d80680f28.

crates/revm/src/handler.rs
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/775fd023371f9764fde3e386cb79ae4d80680f28.

crates/revm/src/handler.rs
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/775fd023371f9764fde3e386cb79ae4d80680f28.

crates/precompiles/src/account_keychain/mod.rs
legion2002: makes sense, Addressed in https://github.com/tempoxyz/tempo/commit/74ce9f8d76c3de0b0d337681832da9870bae305b.

crates/precompiles/src/account_keychain/mod.rs
legion2002: yeah think we can descope this

legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/b4a04395fde527c4c2e7c488c4e3b3ff5440ed32.

crates/precompiles/src/account_keychain/mod.rs
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/066119722ed771a426e0e7802fff42b3495397c4.

crates/revm/src/handler.rs
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/11f243d4bfae94d80f3ce812d85d80e62a3bc9eb.

crates/revm/src/handler.rs
```diff
@@ -143,6 +143,28 @@ fn tempo_signature_verification_gas(signature: &TempoSignature) -> u64 {
...
+    key: AuthorizedKey,
+}
+
+fn recover_key_authorization_signer(
+    cache: &OnceLock<Address>,
+    key_auth: &SignedKeyAuthorization,
+) -> Result<Address, TempoInvalidTransaction> {
+    if let Some(signer) = cache.get() {
+        return Ok(*signer);
+    }
```
klkvr: i meant cache directly on the `SignedKeyAuthorization`, just like we do for all other types. that way we shouldn't need the `key_authorization_signer` field at all because it'd be cheap enough to invoke recovery once more

crates/revm/src/handler.rs
```diff
@@ -1205,18 +1244,68 @@ where
...
+                    {
+                        loaded_key.key.is_admin
+                    }
+                    Some(_) | None => {
+                        return Err(TempoInvalidTransaction::KeychainValidationFailed {
+                            reason:
+                                "admin-signed key authorization must be signed by transaction key"
+                                    .to_string(),
+                        }
+                        .into());
```
klkvr, line 1282: now that we enforce this i guess this check could be moved to the branch above where we just loaded the key?

also afaict this still allows key authorizations signed by the root key to be submitted by admin keys

klkvr (approved):

crates/revm/src/handler.rs
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/feab627121fcdff5267ce7a41cde486ce777847a.

crates/precompiles/src/account_keychain/mod.rs
```diff
@@ -318,13 +366,44 @@ impl AccountKeychain {
...
+            },
+            witness,
+            true,
+        )?;
+        self.emit_event(AccountKeychainEvent::AdminKeyAuthorized(
+            IAccountKeychain::AdminKeyAuthorized {
+                account: msg_sender,
+                publicKey: key_id,
+            },
+        ))
```
0xrusowsky: nit: now the `sol!` macro ships with builder methods, so we can do:
```suggestion
        self.emit_event(
            AccountKeychainEvent::admin_key_authorized(msg_sender, key_id)
        )
```

(needs proper fmt)

0xrusowsky (approved):
handler keyauth-related logic got so complex that its hard to have confidence we don't miss any edgecases

with that being said, everything makes sense and anything wrong stands out

crates/revm/src/handler.rs
legion2002: Yeah we block this root key behaviour in this commit https://github.com/tempoxyz/tempo/pull/4265/commits/e7835d5efe127672475569b19c09a1fe5eccc34d 

for consistency

PR #4270, opened 2026-05-26
Fix TIP-1045 YAML front matter

decofe:
## Summary
- Quote the TIP-1045 authors front-matter field so YAML parsers accept the leading @ handle.

## Validation
- uv run --with pyyaml python - <<'PY' ...

Prompted by: @max-digi


tips/tip-1045.md
```diff
@@ -2,7 +2,7 @@
 id: TIP-1045
 title: Payment Transaction Classification
 description: Defines consensus rules for classifying transactions as payment lane eligible using an allow-list of call targets/selectors and bounded auxiliary payloads.
-authors: @0xrusowsky, Arsenii Kulikov @klkvr, Dan Robinson @danrobinson, Tanishk Goyal @legion2002
+authors: "@0xrusowsky, Arsenii Kulikov @klkvr, Dan Robinson @danrobinson, Tanishk Goyal @legion2002"
```
jenpaff, line 5: if you just add Marc in the beginning should be fixed

PR #4272, opened 2026-05-26
fix(txpool): use resolved fee token in maintenance

mattsse:
Fixes SIGP-261

Uses the cached resolved fee token for txpool pause grouping and TIP-403 maintenance evictions, so implicitly resolved AA fee-token transactions are rechecked against the token they were validated with.

Also documents when resolved fee tokens are cached and when maintenance code should use the effective token helper.

### DIFF
--- a/crates/transaction-pool/src/maintain.rs
+++ b/crates/transaction-pool/src/maintain.rs
@@ -640,14 +640,18 @@ where
                 if !pause_tokens.is_empty() {
                     let all_txs = pool.all_transactions();
 
-                    // Group transactions by fee token for efficient batch processing.
+                    // Group transactions by effective fee token for efficient batch processing.
                     // This single pass over all transactions handles all pause events.
-                    let mut by_token: AddressMap<Vec<TxHash>> = AddressMap::default();
-                    for tx in all_txs.pending.iter().chain(all_txs.queued.iter()) {
-                        if let Some(fee_token) = tx.transaction.inner().fee_token() {
-                            by_token.entry(fee_token).or_default().push(*tx.hash());
-                        }
-                    }
+                    let mut by_token = all_txs.into_iter().fold(
+                        AddressMap::<Vec<TxHash>>::default(),
+                        |mut by_token, tx| {
+                            by_token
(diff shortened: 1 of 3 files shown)


crates/transaction-pool/src/maintain.rs
```diff
@@ -640,14 +640,20 @@ where
...
+                    let mut by_token = all_txs.into_iter().fold(
+                        AddressMap::<Vec<TxHash>>::default(),
+                        |mut by_token, tx| {
+                            if tx.transaction.inner().as_aa().is_some() {
+                                by_token
+                                    .entry(tx.transaction.effective_fee_token())
+                                    .or_default()
+                                    .push(*tx.hash());
+                            }
+                            by_token
```
klkvr: rn this works only for AA transactions but i'm assuming regular transactions should be covered too here?

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -322,13 +322,13 @@ where
...
 
-            // Check 4: Blacklisted fee payers
-            // Only check AA transactions with a fee token (non-AA transactions don't have
-            // a fee payer that can be blacklisted via TIP403)
+            // Check 4: Blacklisted fee payers.
+            // Only AA transactions have a fee payer that can be blacklisted via TIP403.
             if !updates.blacklist_additions.is_empty()
                 && let Some(ref mut provider) = state_provider
-                && let Some(fee_token) = tx.transaction.inner().fee_token()
+                && tx.transaction.inner().is_aa()
```
klkvr:
```suggestion
```

crates/transaction-pool/src/tempo_pool.rs
```diff
@@ -373,8 +373,9 @@ where
             // will fail validation at execution time.
             if !updates.whitelist_removals.is_empty()
                 && let Some(ref mut provider) = state_provider
-                && let Some(fee_token) = tx.transaction.inner().fee_token()
+                && tx.transaction.inner().is_aa()
```
klkvr:
```suggestion
```

klkvr (approved):

PR #4282, opened 2026-05-26
perf(payload-builder): dedup prewarm storage touches with HashSet

i-m-aditya:
Replace the O(n) linear scan with a HashSet for O(1) membership checks. Same unique touch list, same order — lower CPU on the prewarm worker path when fee and call paths overlap.


mattsse (changes requested):
can we keep the arg names as just `touches` then we can remove lik 80% of the diff here

mattsse:
derek bench mode=e2e

shekhirin:
TIP-20 specialized prewarming has been removed

PR #4289, opened 2026-05-26
fix: account for extra data when estimating block size

klkvr:
fixes SIGP-267

Right now when estimating block size in the builder we don't account for a potentially large extra data from DKG service. This PR fixes it so that estimate is more correct


mattsse (approved):

0xalpharush:
This is also VERIA-94

PR #4302, opened 2026-05-26
feat(signature-verifier): add verifyAdmin

legion2002:
Stacked on top of #4265 
Adds the T6 `verifyAdmin(account, hash, signature)` selector to `SignatureVerifier`.

The method recovers Tempo signatures with the existing rules and returns whether the recovered signer is the account root or an active admin access key.

### DIFF
--- a/crates/precompiles/src/signature_verifier/mod.rs
+++ b/crates/precompiles/src/signature_verifier/mod.rs
@@ -1,6 +1,6 @@
 pub mod dispatch;
 
-use crate::{SIGNATURE_VERIFIER_ADDRESS, error::Result};
+use crate::{SIGNATURE_VERIFIER_ADDRESS, account_keychain::AccountKeychain, error::Result};
 use alloy::primitives::{Address, B256, Bytes};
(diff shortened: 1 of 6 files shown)


crates/precompiles/src/signature_verifier/mod.rs
```diff
@@ -40,6 +44,15 @@ impl SignatureVerifier {
...
     }
+
+    pub fn verify_admin(&mut self, account: Address, hash: B256, signature: Bytes) -> Result<bool> {
+        let signer = match self.recover(hash, signature) {
+            Ok(signer) => signer,
+            Err(TempoPrecompileError::SignatureVerifierError(_)) => return Ok(false),
+            Err(err) => return Err(err),
+        };
+        AccountKeychain::new().is_admin_key(account, signer)
+    }
```
0xrusowsky: isn't it counterintuitive that it doesn't propagate signature verifier errors?

legion2002: yup fixed

0xrusowsky (approved):

PR #4317, opened 2026-05-27
fix(tip-1034): skip noop top-ups

0xrusowsky:
ref: `SIGP-273`


legion2002 (approved):
makes sense , we also have a comment on top of the fn that clarifies this behavior.

PR #4319, opened 2026-05-27
fix(txpool): evict implicit fee txs by fee payer

mattsse:
Fixes SIGP-268.

Implicit-fee txpool invalidation now compares `UserTokenSet` events against the recovered fee payer, matching EVM fee-token resolution. This evicts sponsored AA transactions when the sponsor preference changes and keeps them when only the sender preference changes.


0xrusowsky (approved):
makes sense

PR #4327, opened 2026-05-27
chore(tip-1034): use remaining slot space for u64 timestamps

0xrusowsky:
ref: `SIGP-281`


0xrusowsky:
we want to keep the reserve bytes

PR #4342, opened 2026-05-27
fix(txpool): skip stale aa2d demotion

mattsse:
fixes SIGP-274

Avoid demoting AA2D descendants when a removed transaction sits below the current independent nonce. This keeps expired stale ancestors from turning the current executable transaction into queued state while preserving independent cleanup for real nonce gaps.


mattsse:
closing because we can deal with this issue

PR #4346, opened 2026-05-27
perf(tip-1034): cache domain separator

0xrusowsky:
ref: `SIGP-284`

this PR addresses the misc informational issues from the T5 audit:
- caches the domain separator for TIP-1034.
- makes a DKG error description more accurate


legion2002 (approved):
makes sense

PR #4350, opened 2026-05-27
docs(dex): clarify multi-hop policy 403 behavior

0xrusowsky:
ref: `SIGP-278`


legion2002 (approved):
it makes sense to have this policy, where only input and output 403 policies are checked.
But should still get a confirmation from @danrobinson that this is intentional.

jenpaff (approved):

PR #4364, opened 2026-05-27
refactor(transaction-pool): avoid expiring nonce hash copy

mattsse:
Avoids copying the expiring transaction hash when removing it from the pool hash index by passing the existing hash reference directly.

This keeps the expiring nonce removal path slightly leaner without changing behavior.


shekhirin (approved):
performance

PR #4365, opened 2026-05-27
refactor(transaction-pool): simplify expiring nonce eviction order

mattsse:
Removes the expiring nonce hash tie-breaker from `ExpiringNonceEvictionKey` ordering. Submission IDs are unique for live entries, so the hash is only retained as removal payload.

This keeps the eviction key ordering aligned with the data that can actually affect ordering.

### DIFF
--- a/crates/transaction-pool/src/tt_2d_pool.rs
+++ b/crates/transaction-pool/src/tt_2d_pool.rs
@@ -1464,8 +1464,10 @@ impl AA2dInternalTransaction {
 /// Order:
 /// 1. Priority ascending (lowest priority evicted first)
 /// 2. Submission ID descending (newer transactions evicted first among equal priority)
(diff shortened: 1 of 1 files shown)


crates/transaction-pool/src/tt_2d_pool.rs
```diff
@@ -1482,12 +1484,19 @@ impl ExpiringNonceEvictionKey {
     }
 }
 
+impl PartialEq for ExpiringNonceEvictionKey {
+    fn eq(&self, other: &Self) -> bool {
+        self.submission_id == other.submission_id && self.priority == other.priority
```
shekhirin: do we need to compare priority here then?

mattsse: right we can remove

same thing we do on `EvictionKey` already

shekhirin (approved):

PR #4375, opened 2026-05-27
fix: reject keychain-wrapped channel reserve vouchers

legion2002:
## Summary

Fixes CHAIN-1195 by making TIP-1045 payment classification mirror TIP20ChannelReserve voucher signature acceptance.

The classifier previously accepted any `TempoSignature` encoding for channel-reserve `settle` and `close` calldata. Execution routes voucher recovery through `SignatureVerifier`, which only accepts `PrimitiveSignature` encodings. That meant Keychain-wrapped voucher signatures could occupy payment-lane capacity but later fail during precompile execution.

This change validates channel-reserve voucher signatures with `PrimitiveSignature::from_bytes`, matching execution, and adds a regression test for Keychain-wrapped signatures on both `settle` and `close`.

Linear: https://linear.app/tempoxyz/issue/CHAIN-1195/cyclops-t5-payment-classifier-admits-keychain-wrapped-vouchers

## Validation

- `cargo fmt --check`
- `cargo test -p tempo-primitives test_payment_v2_rejects_keychain_wrapped_channel_reserve_signature`


crates/primitives/src/transaction/envelope.rs
```diff
@@ -507,13 +505,19 @@ fn is_tip1045_call(to: Option<&Address>, input: &[u8]) -> bool {
...
+                is_tip20_channel_reserve_voucher_signature,
             )
         }
         _ => false,
     }
 }
 
+/// Returns `true` if the voucher signature can be decoded by TIP20ChannelReserve execution.
+#[inline]
+fn is_tip20_channel_reserve_voucher_signature(signature: &[u8]) -> bool {
```
0xrusowsky: can't we inline it directly since it is only called in a single place?

legion2002: yup, updated

0xrusowsky (approved):

PR #4383, opened 2026-05-27
chore(tip-1034): resolve virtual payee on open/top-up checks

0xrusowsky:
ref: `CHAIN-1192`

### DIFF
--- a/crates/precompiles/src/tip20_channel_reserve/mod.rs
+++ b/crates/precompiles/src/tip20_channel_reserve/mod.rs
@@ -683,7 +683,6 @@ mod tests {
     use tempo_contracts::precompiles::{
         ITIP20ChannelReserve::ITIP20ChannelReserveCalls, TIP20Error,
     };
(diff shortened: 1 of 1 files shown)


legion2002 (approved):

crates/precompiles/src/tip20_channel_reserve/mod.rs
```diff
@@ -153,7 +145,7 @@ impl TIP20ChannelReserve {
             return Err(TIP20ChannelReserveError::channel_already_exists().into());
         }
 
-        token.ensure_authorized_as(call.payee, AuthRole::Recipient)?;
+        token.ensure_authorized_as(Recipient::resolve(call.payee)?.target, AuthRole::Recipient)?;
```
legion2002, line 148: noting that we don't have to check both virtual and master addresses here, because virtual addresses are not allowed in TIP 403 policies.

If that changes in the future, we'd also need to update this.

PR #4391, opened 2026-05-27
perf(revm): do not compute unique tx identifier twice

shekhirin:
We computed it twice for AA transactions.


crates/revm/src/tx.rs
```diff
@@ -414,30 +414,30 @@ impl FromRecoveredTx<AASigned> for TempoTxEnv {
 
 impl FromRecoveredTx<TempoTxEnvelope> for TempoTxEnv {
     fn from_recovered_tx(tx: &TempoTxEnvelope, sender: Address) -> Self {
-        let unique_tx_identifier = Some(tx.unique_tx_identifier(sender));
-
+        // [`TempoTxEnv::from_recovered_tx`] calculates `unique_tx_identifier` on its own,
+        // so we compute it on demand only for non-AA transactions
```
mattsse: this is a classic slopmm cmment that has no value without this pr as context

mattsse (approved):

PR #4425, opened 2026-05-27
fix(pool): exclude mined txs from expiry eviction

mattsse:
Ensures expiry eviction explicitly filters out transactions included in the committed chain before checking pool membership.

This keeps mined transactions from being reported as expired if pool pruning and local expiry tracking disagree.


Rjected (approved):
makes sense

PR #4415, opened 2026-05-27
ci(bench): update e2e regenesis and node arg compatibility

decofe:
## Summary
- build a dedicated latest `origin/main` tempo binary when bench-e2e regenesis is requested
- use that main binary for snapshot regenesis instead of baseline or feature binaries
- keep the local reth arg list in the original single-list shape
- filter default local node args per benchmark binary using `tempo node --help`, so older/newer refs skip unsupported flags automatically

## Verification
- `nu --ide-check 0 bench-e2e.nu`
- YAML parse for `.github/workflows/bench.yml` and `.github/workflows/bench-e2e.yml`
- `git diff --check`

Prompted by: @shekhirin

PR #4431, opened 2026-05-27
ci(bench): tolerate missing invalid pool tx metric

decofe:
## Summary
- Treat all sampled counter metrics in `tempo.nu` summary generation as optional.
- Warn and emit `n/a`/zeroed summary fields when older benchmark binaries do not emit a metric instead of failing the job.

## Testing
- `nu --commands 'source tempo.nu'`

Prompted by: @shekhirin

### DIFF
--- a/tempo.nu
+++ b/tempo.nu
@@ -976,9 +976,9 @@ def generate-summary [
             | flatten
     }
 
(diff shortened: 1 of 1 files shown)


decofe:
Closing in favor of adding this change to #4415.

PR #4482, opened 2026-05-27
feat: include finalization in tempo snapshots 

hamdiallam:
Wrap Reth's `snapshot-manifest` and `download` commands

1. Encode the finalization cert for the manifest's block in the manifest.json file
2. Dump the cert on disk to be consumed on startup if the finalizations archive is empty

## Followups
When starting up the node, if the finalizations archive is empty, we can read the bootstrap finalization and set the floor from there. The snapshots ensure that the block is persisted in the EL


bin/tempo/src/main.rs
```diff
@@ -324,12 +316,15 @@ fn main() -> eyre::Result<()> {
     tempo_node::init_version_metadata();
     defaults::init_defaults();
 
-    let mut cli = match TempoCli::command()
+    // `tempo snapshot-manifest` and `tempo download` wrap the Reth variants, so they
+    // cannot be added without colliding. Mutate those subcommands with the wrapped ones.
+    let matches = match TempoCli::command()
```
SuperFluffy, line 330: Unfortunate that these hacks are needed. :-(

bin/tempo/src/snapshot_manifest.rs
```diff
@@ -0,0 +1,134 @@
...
+            .wrap_err_with(|| format!("failed to create output dir: {output_dir:?}"))?;
+
+        eprintln!("packaging execution layer");
+
+        let start = Instant::now();
+        self.inner
+            .execute()
+            .wrap_err("reth snapshot-manifest (EL packaging) failed")?;
+
+        eprintln!("execution layer snapshot finished in {:?}", start.elapsed());
```
SuperFluffy: Let's use `display_duration` from `tempo-telemetry-utils`

bin/tempo/src/main.rs
SuperFluffy: @klkvr

crates/commonware-node/src/storage/mod.rs
```diff
@@ -1,14 +1,14 @@
 //! This module defines consensus archive formats.
 //!
-//! Finalized blocks are stored in a [`Hybrid`] store which
+//! Finalized blocks are stored in a `Hybrid` store which
```
SuperFluffy: Bring back `[]` so it creates links? Did it complain?

crates/commonware-node/src/storage/mod.rs
```diff
@@ -1,14 +1,14 @@
...
 //! merges a prunable archive (holding the most recently finalized blocks) with
 //! a lookup into the execution layer (used for blocks below the prunable
 //! retention window).
 //!
 //! Older deployments stored finalized blocks in an immutable archive. To
 //! preserve the ability to roll back to one of those releases, the
-//! [`legacy`] module is opened on every restart and every newly
-//! finalized block is dual-written to it from [`Hybrid`]. The
+//! `legacy` module is opened on every restart and every newly
+//! finalized block is dual-written to it from `Hybrid`. The
```
SuperFluffy: likewise here - `[]` for links?

bin/tempo/src/snapshot_manifest.rs
```diff
@@ -0,0 +1,202 @@
...
+            "execution layer snapshot finished in {}",
+            display_duration(start.elapsed())
+        );
+
+        if self.skip_consensus {
+            eprintln!("--skip-consensus set. skipping consensus layer");
+            return Ok(());
+        }
+
+        let manifest_path = output_dir.join("manifest.json");
```
SuperFluffy, line 122: IMO toss this into a separate `read_manifest` to make error reporting neater:

```rust
# usage
let manifest_path = output_dir.join("manifest.json");
read_manifest(&manifest_path).wrap_err_with(|| format!("failed reading manifest from `{}`", manifest_pathdisplay()))?;

fn read_manifest(p: impl AsRef<Path>) -> eyre::Result<Manifest> {}
```

bin/tempo/src/snapshot_manifest.rs
```diff
@@ -0,0 +1,202 @@
...
+                "mismatch in --chain id {} and manifest chain id {}",
+                spec.chain_id(),
+                manifest.chain_id
+            ),
+        };
+
+        let block = execution_provider(chainspec, source_datadir)?
+            .block_by_number(manifest.block)
+            .wrap_err("failed to read block")?
+            .ok_or_eyre("missing block")?;
```
SuperFluffy: IMO too brittle. This happens to work for our certificate-enabled RPC nodes right now, but not for the general case (snapshots from verifiers, or some future followers that also take into account).

Ideally we fix this upstream and ensure the snapshot taken always points to the latest finalized block.

But even then we don't have a guarantee that the block is actually available because of reth's durability guarantees.

SuperFluffy (commented):
I think the general direction of this is fine, but IMO we should do something like this to reconcile the newest available finalization certificate in the CL with that highest available block in the EL (pseudo code):

```
certs := cl.certs_newest_to_latest()

cert := for cert in certs.next {
    if cl.contains(cert.digst) {
        cert
    }
}
```

crates/commonware-node/src/storage/mod.rs
hamdiallam: yea the docs linter was complaining about the types not being public, i'll try re-add them and see if it's still upset lol

bin/tempo/src/snapshot_manifest.rs
hamdiallam: Moved towards just encoded whatever latest finalization we have

bin/tempo/src/snapshot_manifest.rs
```diff
@@ -0,0 +1,173 @@
...
+        );
+
+        if skip_consensus {
+            eprintln!("--skip-consensus set. skipping consensus layer");
+            return Ok(());
+        }
+
+        let manifest_path = output_dir.join("manifest.json");
+        let manifest = read_manifest(&manifest_path)
+            .wrap_err_with(|| format!("failed reading manifest: {manifest_path:?}"))?;
```
SuperFluffy: -> `...{}", manifest_path.display()).`

bin/tempo/src/snapshot_manifest.rs
```diff
@@ -0,0 +1,173 @@
...
+
+        let consensus_dir = consensus_datadir.unwrap_or_else(|| source_datadir.join("consensus"));
+        eprintln!("embedding latest consensus finalization. consensus dir: {consensus_dir:?}");
```
SuperFluffy: use `Path::display()` instead of debug-fmt

bin/tempo/src/snapshot_manifest.rs
```diff
@@ -0,0 +1,173 @@
...
+            .insert(
+                TEMPO_CONSENSUS_MANIFEST_KEY.to_string(),
+                serde_json::to_value(consensus_manifest)
+                    .wrap_err("failed to serialize Tempo consensus manifest extension")?,
+            );
+
+        let manifest_json = serde_json::to_string_pretty(&manifest_json)
+            .wrap_err("failed to serialize manifest")?;
+        fs::write(&manifest_path, manifest_json)
+            .wrap_err_with(|| format!("failed to write {manifest_path:?}"))?;
```
SuperFluffy: `Path::display`

bin/tempo/src/snapshot_download.rs
```diff
@@ -0,0 +1,199 @@
...
+    #[arg(long, default_value_t = true)]
+    skip_consensus: bool,
+
+    /// Consensus storage directory. If not set, this will be derived from --datadir.
+    #[arg(long = "consensus.datadir", value_name = "PATH")]
+    consensus_datadir: Option<PathBuf>,
+}
+
+pub(crate) fn run(matches: &ArgMatches) -> eyre::Result<()> {
+    let args = Args::from_arg_matches(matches).map_err(|e| eyre!("{e}"))?;
```
SuperFluffy: `Args::from_arg_matches(matches).wrap_err` doesn't work? Is the error not a `std::error::Error`?

SuperFluffy (approved):
Approving because this is directionally ok - but I think we should still look at what's available in both the EL and CL and reconcile based off of that (or at least warn?).

crates/commonware-node/src/lib.rs
```diff
@@ -15,7 +15,7 @@ pub mod follow;
 pub mod metrics;
 pub(crate) mod network_identity;
 pub(crate) mod peer_manager;
-pub(crate) mod storage;
```
SuperFluffy, line 18: This change likely causes the doc lint to now fail.

hamdiallam: Yep nice catch

hamdiallam:
> Approving because this is directionally ok - but I think we should still look at what's available in both the EL and CL and reconcile based off of that (or at least warn?).

Agreed, lets followup with improvements here to unblock the marshal update?

bin/tempo/src/snapshot_manifest.rs
```diff
@@ -0,0 +1,244 @@
...
+) -> eyre::Result<(u64, TempoFinalization)> {
+    let runtime_config =
+        commonware_runtime::tokio::Config::default().with_storage_directory(consensus_dir);
+
+    let runner = commonware_runtime::tokio::Runner::new(runtime_config);
+    runner.start(|context| async move {
+        find_last_finalized_marker(&context, &execution_provider, max_depth)
+            .await?
+            .ok_or_eyre("no finalization marker found")
+    })
```
SuperFluffy, line 222: IMO report somewhere from/to which height the script has searched. Worth outputting that information.

bin/tempo/src/snapshot_manifest.rs
```diff
@@ -0,0 +1,244 @@
...
+    /// Skip encoding consensus state
+    #[arg(long, default_value_t = true)]
+    skip_consensus: bool,
+
+    /// Chain spec override for local/unknown chains.
+    #[arg(long, short, value_parser = chain_value_parser)]
+    chain: Option<Arc<TempoChainSpec>>,
+
+    /// Consensus storage directory. If not set, this will be derived from --datadir.
+    #[arg(long = "consensus.datadir", value_name = "PATH")]
```
SuperFluffy, line 68: double check that this matches the name for node bringup

bin/tempo/src/snapshot_manifest.rs
```diff
@@ -0,0 +1,244 @@
...
+#[derive(Debug, Parser)]
+#[command(
+    name = "snapshot-manifest",
+    about = "Generate snapshot archives and a manifest for the EL plus consensus floor certificate."
+)]
+pub(crate) struct Args {
+    #[command(flatten)]
+    inner: SnapshotManifestCommand,
+
+    /// Skip encoding consensus state
```
SuperFluffy: Note here that setting gives you vanilla-reth behavior.

bin/tempo/src/snapshot_manifest.rs
```diff
@@ -0,0 +1,244 @@
...
+        .get_one::<PathBuf>("source_datadir")
+        .cloned()
+        .expect("--source-dir must be set");
+
+    let output_dir = matches
+        .get_one::<PathBuf>("output_dir")
+        .cloned()
+        .expect("--output-dir must be set");
+
+    args.execute(&source_datadir, &output_dir)
```
SuperFluffy, line 89: We can't get these "for free" by using `Args`? :-/

bin/tempo/src/snapshot_manifest.rs
hamdiallam: Yep, matches!

bin/tempo/src/snapshot_manifest.rs
hamdiallam: sadly the command's fields are private

SuperFluffy (approved):

PR #4540, opened 2026-05-28
perf(precompiles): remove handler cache

decofe:
## Summary
- remove keyed `HandlerCache` reuse from mapping, array, and vec handlers
- add a non-keyed append-only `HandlerArena` so existing reference-returning `Index` APIs remain stable
- recompute child handlers on every access for comparison against the cached variants

## Testing
- `cargo fmt --check`
- `cargo test -p tempo-precompiles --features test-utils storage::types`


0xrusowsky:
closing in favor of https://github.com/tempoxyz/tempo/pull/4455

PR #4554, opened 2026-05-28
perf(transaction-pool): cache AA pool counts

figtracer:
Caches AA 2D pool pending/queued counts so count reads no longer scan every AA transaction. `discard()` calls this count on insertion, and it showed up in the P2P receive/import profile under `TransactionsManager::import_transactions`.

The behavior should be unchanged: the cached counters are updated on insert, replacement, removal, promotion, demotion, and expiring-nonce changes. Test-only invariants still recompute the counts by scanning the pool and assert they match the cached values.

Profile evidence from a live two-node P2P run using profiling builds, 5,000 deterministic AA txs, 1,000 accounts, concurrency 200:

- Baseline `070951c3`: `AA2dPool::discard` was 347 inclusive samples, 2.149% of the profile.
- This PR `49ff66d1`: `AA2dPool::discard` was 1 inclusive sample, 0.006% of the profile.

Profiles: baseline https://share.firefox.dev/4dNQZ81, this PR https://share.firefox.dev/49Ljhz9

The claim here is intentionally narrow: this removes a profile-visible O(N) AA pool count scan from P2P AA transaction import. 

Checked with `cargo +nightly fmt --all --check`, `cargo test -p tempo-transaction-pool`, and `cargo +nightly clippy -p tempo-transaction-pool --all-features --tests -- -D warnings`.


mattsse (approved):
looks like we have all changes covered here and this is better than scanning txs

PR #4561, opened 2026-05-28
fix: gate auth new account charge

klkvr:


0xrusowsky (approved):
makes sense, pending re-execution tests

legion2002 (approved):

PR #4277, opened 2026-05-26
feat(builder): add elastic payload build budget

joshieDo:
Replaces fixed proposal transaction timing with an elastic budget derived from `consensus.target-block-time - consensus.network-budget`. The network budget default is now `50ms`; `consensus.wait-for-proposal` stays `1200ms` for unhealthy views.

Mechanisms:
- Late `handle_propose` entry is subtracted before payload building, so delayed views do not get a fresh full budget.
- Consensus proposals pass the remaining return window through payload attributes; non-consensus builds keep `None` and behave as before.
- The builder stops pool tx execution when `leader_idle + 2 * predicted_replayable_work + 2 * estimated_marshal_persist >= budget`.
- Pool-empty idle counts once; replayable work and marshal persistence count twice because validators repeat them.
- Replayable work starts from `--builder.build-time-multiplier` (`1.35`), learns upward up to `1.70`, and decays down slowly.
- Marshal persistence is learned as ns per encoded block byte from `marshal.proposed`; tiny blocks under `128KiB` are ignored.
- Proposal return wait subtracts prebuild delay, actual build elapsed, payload validation work, and estimated validator marshal persistence.

Validated with:
- `cargo +nightly fmt --all --check`
- `cargo check -p tempo-payload-builder -p tempo-commonware-node -p tempo-e2e -p tempo-node -p tempo --tests`
- `cargo test -p tempo-payload-builder budget`
- `cargo test -p tempo-payload-types budget --lib`

closes RETH-867

PR #4570, opened 2026-05-28
docs: clarify elastic proposal budget model

joshieDo:
Clarifies the elastic proposal budget docs added in #4277, especially what value flows into `payload_build_budget`, how the proposal return budget is derived, and how replayable work, marshal persistence, and the build multiplier fit together.

This is docs-only and does not change budget behavior.

Tested with `cargo fmt --check` and `cargo check -p tempo-payload-types -p tempo-payload-builder -p tempo-commonware-node -p tempo-node --all-targets`.

### DIFF
--- a/crates/payload/types/src/budget.rs
+++ b/crates/payload/types/src/budget.rs
@@ -56,10 +56,6 @@ pub fn observe_marshal_persist(block_size_bytes: usize, elapsed: Duration) {
 }
 
 /// Point-in-time marshal persistence cost per encoded block byte.
-///
-/// The estimator is deliberately small and copyable so consensus can capture a
-/// stable rate at the start of a budget decision and pass it through without
-/// re-reading the global atomic mid-decision.
 #[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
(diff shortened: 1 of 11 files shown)


crates/payload/types/src/budget.rs
```diff
@@ -11,12 +11,22 @@ const MIN_SAMPLE_BYTES: usize = 128 * 1024;
...
+/// decision.
 pub fn marshal_persist_estimate() -> MarshalPersistEstimator {
     MarshalPersistEstimator::from_ns_per_byte(MARSHAL_PERSIST_NS_PER_BYTE.load(Ordering::Relaxed))
 }
 
 /// Records time spent persisting an encoded block through consensus marshal.
+///
+/// The observation is stored as nanoseconds per encoded block byte. Large
+/// blocks teach future build and return budgets how much size-dependent
+/// persistence time to reserve for both proposers and validators.
```
mattsse, line 27: how does this work, I dont see how this affects future values because this fn is stateless

mattsse (approved):

PR #4579, opened 2026-05-28
perf(txgen): use secp256k1 signing backend

mattsse:
Switches the txgen-style TIP20 execution benchmark to build mnemonic accounts as Alloy secp256k1-backed signers and enables the signer feature on the tempo-evm bench dev-dependency.

Generated signatures stay compatible while high-volume txgen signing uses the faster backend.


mattsse:
Closing; this change belongs in tempoxyz/txgen instead.

PR #4026, opened 2026-05-18
feat(bench): add optional BAL to Tempo execution payloads

decofe:
## Summary
- add `--builder.enable-bal` to opt into block access list generation
- wire Tempo payload building through Reth's BAL builder path
- preserve RLP-encoded BAL on `TempoExecutionData` and `block_to_payload`

## Verification
- `cargo fmt --check`
- `git diff --check`
- `cargo check -p tempo-payload-types -p tempo-payload-builder -p tempo-node -p tempo-evm -p tempo-commonware-node`

PR #4601, opened 2026-05-28
fix(consensus): tolerate experimental BAL hash mismatch

decofe:
Stacked on #4026.

Propagates the experimental BAL builder flag into `TempoConsensus` and uses it to ignore only `ConsensusError::BlockAccessListHashMismatch`. The default `TempoConsensus::new` path remains strict.

Validation:
- `cargo fmt`
- `cargo check -p tempo-consensus -p tempo-node`

Prompted by: @klkvr


decofe:
Closing; the fix was moved directly onto #4026.

PR #4609, opened 2026-05-28
test: add tip20 user state invariants

0xrusowsky:
this PR adds new TIP20 tests to verify the `UserState` invariants defined in TIP-1065

### DIFF
--- a/tips/verify/test/invariants/TIP20.t.sol
+++ b/tips/verify/test/invariants/TIP20.t.sol
@@ -1496,9 +1496,7 @@ contract TIP20InvariantTest is InvariantBaseTest {
         (address rewardRecipient,,) = token.userRewardInfo(account);
 
         assertEq(
-            decodedBalance,
-            exposedBalance,
-            "TIP-1065: low 128 bits must equal balanceOf(account)"
+            decodedBalance, exposedBalance, "TIP-1065: low 128 bits must equal balanceOf(account)"
         );
         assertLe(
             exposedBalance,
@@ -1508,9 +1506,8 @@ contract TIP20InvariantTest is InvariantBaseTest {
         assertEq(reservedBits, 0, "TIP-1065: reserved user-state bits must remain zero");
 
         if (rewardFlag != RewardFlag.Uninitialized) {
-            RewardFlag expectedFlag = rewardRecipient == address(0)
-                ? RewardFlag.OptedOut
-                : RewardFlag.OptedIn;
+            RewardFlag expectedFlag =
+                rewardRecipient == address(0) ? RewardFlag.OptedOut : RewardFlag.OptedIn;
             assertEq(
                 uint256(rewardFlag),
                 uint256(expectedFlag),
@@ -1528,9 +1525,8 @@ contract TIP20InvariantTest is InvariantBaseTest {
         view
     {
         (, RewardFlag rewardFlag,) = _userState(address(token), account);
-        RewardFlag expectedFlag = rewardRecipient == address(0)
-            ? RewardFlag.OptedOut
-            : RewardFlag.OptedIn;
+        RewardFlag expectedFlag =
+            rewardRecipient == address(0) ? RewardFlag.OptedOut : RewardFlag.OptedIn;
         assertEq(
             uint256(rewardFlag),
             uint256(expectedFlag),
--- a/tips/verify/test/invariants/TIP20.t.sol
+++ b/tips/verify/test/invariants/TIP20.t.sol
@@ -38,7 +38,7 @@ contract TIP20InvariantTest is InvariantBaseTest {
     uint256 internal constant USER_STATE_BALANCE_MASK = type(uint128).max;
     uint256 internal constant USER_STATE_REWARD_FLAG_MASK = uint256(3) << 128;
     uint256 internal constant USER_STATE_RESERVED_MASK = type(uint256).max << 130;
-    uint256 internal constant USER_STATE_BALANCES_SLOT = 10;
+    uint256 internal constant USER_STATE_BALANCES_SLOT = 9;
 
     enum RewardFlag {
         Uninitialized,
--- a/tips/verify/test/invariants/TIP20.t.sol
+++ b/tips/verify/test/invariants/TIP20.t.sol
@@ -1428,19 +1428,18 @@ contract TIP20InvariantTest is InvariantBaseTest {
             assertEq(totalSupply, expectedSupply, "TEMPO-TIP18: Supply conservation violated");
 
             // TEMPO-TIP20: Balance sum equals supply
-            address[] storage holders = _tokenHolders[tokenAddr];
-            uint256 balanceSum = 0;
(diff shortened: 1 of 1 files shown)


legion2002 (approved):

tips/verify/test/invariants/TIP20.t.sol
```diff
@@ -35,6 +35,17 @@ contract TIP20InvariantTest is InvariantBaseTest {
 
     /// @dev Constants
     uint256 internal constant ACC_PRECISION = 1e18;
+    uint256 internal constant USER_STATE_BALANCE_MASK = type(uint128).max;
+    uint256 internal constant USER_STATE_REWARD_FLAG_MASK = uint256(3) << 128;
+    uint256 internal constant USER_STATE_RESERVED_MASK = type(uint256).max << 130;
+    uint256 internal constant USER_STATE_BALANCES_SLOT = 10;
```
grandizzy: **Must-fix — this is the CI failure.** `balances` is at slot **9**, not 10. Slot 10 is `allowances`, so `vm.load(token, keccak256(abi.encode(account, 10)))` hashes into an empty outer mapping and returns `0`.

The `#[contract]` macro packs `next_quote_token` (Address, 20B) and `transfer_policy_id` (u64, 8B) into the same slot (slot 7, 20+8=28 ≤ 32). Verified locally by printing `tip20_slots::*`:

```
TOTAL_SUPPLY = 8
BALANCES     = 9
ALLOWANCES   = 10
```

```suggestion
    uint256 internal constant USER_STATE_BALANCES_SLOT = 9;
```

Recommended additional hardening: add a one-line sanity assertion in `setUp` so any future TIP20 field-order/packing change surfaces immediately instead of via a fuzz failure, e.g.:

```solidity
require(
    uint256(vm.load(address(_tokens[0]), keccak256(abi.encode(admin, USER_STATE_BALANCES_SLOT))))
        & type(uint128).max == _tokens[0].balanceOf(admin),
    "USER_STATE_BALANCES_SLOT out of sync with TIP20Token layout"
);
```

tips/verify/test/invariants/TIP20.t.sol
```diff
@@ -35,6 +35,17 @@ contract TIP20InvariantTest is InvariantBaseTest {
 
     /// @dev Constants
     uint256 internal constant ACC_PRECISION = 1e18;
+    uint256 internal constant USER_STATE_BALANCE_MASK = type(uint128).max;
+    uint256 internal constant USER_STATE_REWARD_FLAG_MASK = uint256(3) << 128;
+    uint256 internal constant USER_STATE_RESERVED_MASK = type(uint256).max << 130;
```
grandizzy, line 40: **Must-fix.** This encodes the wrong layout. In `crates/precompiles/src/tip20/user_state.rs` the `RewardFlag` derives `Storable`, so it occupies **a full byte** (bits 128..135), not 2 bits — see the explicit comment in that file.

With the current masks:
- `_userState`'s `assertLe(rewardFlagValue, type(RewardFlag).max)` only catches `value == 3`. Bad full-byte values like `0x04`, `0x80`, etc. get silently masked to `0/1/2` and are only caught later by the reserved-bits check.
- `_assertTip1065RewardFlagMatchesRecipient` ignores `reservedBits`, so flag-byte corruption above bit 129 wouldn't fail at that inline call site.

```suggestion
    uint256 internal constant USER_STATE_REWARD_FLAG_SHIFT = 128;
    uint256 internal constant USER_STATE_REWARD_FLAG_MASK = uint256(0xff) << USER_STATE_REWARD_FLAG_SHIFT;
    uint256 internal constant USER_STATE_RESERVED_MASK = type(uint256).max << 136;
```

(And update the `>> 128` in `_userState` to use `USER_STATE_REWARD_FLAG_SHIFT` for clarity.)

tips/verify/test/invariants/TIP20.t.sol
```diff
@@ -1423,6 +1435,13 @@ contract TIP20InvariantTest is InvariantBaseTest {
             }
             assertEq(balanceSum, totalSupply, "TEMPO-TIP20: Balance sum does not equal totalSupply");
 
+            // TIP-1065: T6 packed user-state balance slots must expose only low 128-bit balances
+            // through balanceOf, keep reserved bits zero, and keep initialized rewardFlag values
+            // consistent with rewardRecipient, which remains the source of truth.
+            for (uint256 j = 0; j < holders.length; j++) {
+                _assertTip1065UserState(token, holders[j]);
+            }
```
grandizzy, line 1441: **Perf — merge the two holder loops.** The existing `balanceSum` loop and the new `_assertTip1065UserState` loop both iterate `holders` and both call `token.balanceOf(holder)`. At fuzz500 depth × multiple tokens × ~35–45 holders this duplicates a lot of precompile calls.

Merge them and pass the already-read balance into the helper; also skip the `userRewardInfo` call when the flag is `Uninitialized`:

```solidity
for (uint256 j = 0; j < holders.length; j++) {
    address holder = holders[j];
    uint256 exposedBalance = token.balanceOf(holder);
    balanceSum += exposedBalance;
    _assertTip1065UserState(token, holder, exposedBalance);
}
assertEq(balanceSum, totalSupply, "TEMPO-TIP20: Balance sum does not equal totalSupply");
```

```solidity
function _assertTip1065UserState(ITIP20 token, address account, uint256 exposedBalance) internal view {
    (uint256 decodedBalance, RewardFlag rewardFlag, uint256 reservedBits) =
        _userState(address(token), account);

    assertEq(decodedBalance, exposedBalance, "TIP-1065: low 128 bits must equal balanceOf(account)");
    assertEq(reservedBits, 0, "TIP-1065: reserved user-state bits must remain zero");

    if (rewardFlag != RewardFlag.Uninitialized) {
        (address rewardRecipient,,) = token.userRewardInfo(account);
        RewardFlag expectedFlag =
            rewardRecipient == address(0) ? RewardFlag.OptedOut : RewardFlag.OptedIn;
        assertEq(uint256(rewardFlag), uint256(expectedFlag),
            "TIP-1065: initialized rewardFlag stale vs rewardRecipient");
    }
}
```

grandizzy (commented):
Test-only PR adding TIP-1065 invariants for the T6 packed `UserState`. Spec/implementation cross-check found two correctness bugs that cause the CI failure (`[FAIL: TIP-1065: low 128 bits must equal balanceOf(account): 0 != 1000100000000]`) and a perf cleanup. Verified slot layout locally against the `#[contract]`-generated `tip20_slots` constants.

tips/verify/test/invariants/TIP20.t.sol
0xrusowsky: while that's true, the spec says 2bits.
we used `Storable` to reuse the macro, but if we ever need to use the reserved bits, we will need to roll out a manual impl.

in practice, it should be the same, as `1u8 == 1u2`

grandizzy (approved):

PR #4623, opened 2026-05-28
feat(consensus): register new peers as secondary nodes

SuperFluffy:
Registers peers from the next DKG player set as secondary nodes while keeping the current DKG players primary. This lets newly added peers be tracked before they become primary participants, and preserves active validators that are outside the DKG sets as secondaries.

Adds focused `PeersBuilder::resolve_at_hash` coverage for unchanged, dropped, added, and non-DKG active validators. Narrows validator config reads to `trait ExecutionNode` so tests and production share the same EVM-backed read path without requiring a full node.

Tests: `cargo test -p tempo-commonware-node`


crates/commonware-node/src/validators.rs
```diff
@@ -76,36 +153,26 @@ pub(crate) fn read_active_and_known_peers_at_block_hash(
...
     debug!(height = header.number(), "header found");
 
     let db = State::builder()
         .with_database(StateProviderDatabase::new(
-            node.provider
-                .state_by_block_hash(block_hash)
-                .wrap_err_with(|| {
-                    format!("failed to get state from node provider for hash `{block_hash}`")
-                })?,
+            node.validator_config_state_by_block_hash(block_hash)?,
```
SuperFluffy: bring back error context with `wrap_err`

crates/commonware-node/src/validators.rs
```diff
@@ -76,36 +153,26 @@ pub(crate) fn read_active_and_known_peers_at_block_hash(
...
         ))
         .build();
 
-    let mut evm = node
-        .evm_config
-        .evm_for_block(db, &header)
-        .wrap_err("failed instantiating evm for block")?;
+    let mut evm = node.validator_config_evm_for_block(db, &header)?;
```
SuperFluffy: bring back error context with `wrap_err`

crates/commonware-node/src/validators.rs
```diff
@@ -12,23 +12,100 @@ use commonware_utils::{TryFromIterator, ordered};
...
 };
+use tempo_primitives::TempoHeader;
 
 use tracing::{Level, debug, instrument, warn};
 
 use crate::utils::public_key_to_b256;
 
+/// Minimal execution-node interface needed to read validator config state.
+///
+/// This keeps validator config consumers from depending on a full launched
```
SuperFluffy: Start this that in production we use `TempoFullNode` and that the purpose of this trait is to run tests with a mock.

crates/commonware-node/src/peer_manager/actor.rs
```diff
@@ -292,10 +292,174 @@ where
...
+                                        %error,
+                                        "this is a problem: a known peer could not \
+                                        be found in the smart contract or was \
+                                        malformed"
+                                    )
+                                })
+                                .map(|decoded| {
+                                    (decoded.public_key().clone(), decoded.to_p2p_address())
+                                })
+                                .ok()
```
SuperFluffy: Make this panic again - with this patch we started to emit warnings and then continue working. But actually this problem repeatedly showed that we have some bad execution layer misconfig.

crates/commonware-node/src/peer_manager/actor.rs
```diff
@@ -292,10 +292,174 @@ where
...
+                            .inspect_err(|error| {
+                                warn!(
+                                    %peer,
+                                    %error,
+                                    "this is a problem: a known peer could not \
+                                    be found in the smart contract or was \
+                                    malformed"
+                                )
+                            })
+                            .map(|decoded| decoded.to_p2p_address())
```
SuperFluffy: Likewise here - don't just emit an error and panic as it shows that an invariant in the execution layer was violated.

crates/commonware-node/src/validators.rs
```diff
@@ -12,23 +12,97 @@ use commonware_utils::{TryFromIterator, ordered};
...
+    ) -> eyre::Result<StateProviderBox>;
+
+    fn validator_config_evm_for_block(
+        &self,
+        db: State<StateProviderDatabase<StateProviderBox>>,
+        header: &TempoHeader,
+    ) -> eyre::Result<TempoEvm<State<StateProviderDatabase<StateProviderBox>>>>;
+}
+
+impl ValidatorConfigExecution for TempoFullNode {
```
hamdiallam: remove `validator_` from the function names. Just a passthrough

hamdiallam (approved):

PR #4698, opened 2026-05-29
chore(consensus): bump commonware to v2026.5.0

hamdiallam:
1. Bump Commonware
2. Marshal start argument supplied from the finalized floor
3. Remove --strict-startup. This can not be set to false as consensus data is required in this cw revision


crates/commonware-node/src/consensus/engine.rs
```diff
@@ -136,18 +137,20 @@ where
...
+        let mailbox_size =
+            NonZeroUsize::new(self.mailbox_size).ok_or_eyre("mailbox size must be non-zero")?;
 
         let alias::marshal::Initialized {
             actor: marshal,
             mailbox: marshal_mailbox,
             last_finalized_height,
         } = alias::marshal::init(
-            context.clone(),
+            context.child("marshal"),
```
hamdiallam: note: remove

crates/consensus/src/consensus/engine.rs
```diff
@@ -474,8 +478,11 @@ where
         let peer_manager = self.peer_manager.start();
 
         let broadcast = self.broadcast.start(broadcast_channel);
-        let resolver =
-            marshal::resolver::p2p::init(&self.context, self.resolver_config, marshal_channel);
+        let resolver = marshal::resolver::p2p::init(
```
hamdiallam, line 483: breaking for any metrics but appropriate

crates/consensus/src/consensus/engine.rs
```diff
@@ -512,7 +519,12 @@ where
         ];
 
         if let Some(subblocks) = self.subblocks {
-            tasks.push(self.context.spawn(|_| subblocks.run(subblocks_channel)));
+            tasks.push(
+                self.context
+                    .as_present()
```
hamdiallam, line 526: same here

crates/consensus/src/consensus/application/actor.rs
```diff
@@ -240,65 +235,8 @@ impl Inner<Init> {
...
 
-    #[instrument(
-        skip_all,
-        fields(
-            epoch = %genesis.epoch,
-        ),
-        ret(Display),
-        err(level = Level::ERROR)
-    )]
-    async fn handle_genesis<TContext: commonware_runtime::Clock>(
```
SuperFluffy, line 261: I removed the entire `handle_genesis` logic and moved it into epoch manager.

crates/consensus/src/consensus/engine.rs
```diff
@@ -213,7 +217,7 @@ where
 
         let subblocks = self.with_subblocks.then(|| {
             subblocks::Actor::new(subblocks::Config {
-                context: context.clone(),
+                context: context.child("subblocks"),
```
hamdiallam, line 222: This is a breaking change re metrics. However this is not used right now and should be scoped since the actor is owning this

crates/consensus/src/consensus/application/ingress.rs
```diff
@@ -164,11 +135,11 @@ impl Relay for Mailbox {
     type PublicKey = PublicKey;
     type Plan = commonware_consensus::simplex::Plan<PublicKey>;
 
-    async fn broadcast(&mut self, digest: Self::Digest, plan: Self::Plan) {
-        // TODO: panicking here is really not necessary. Just log at the ERROR or WARN levels instead?
```
SuperFluffy, line 168: nice

crates/e2e/src/lib.rs
```diff
@@ -50,6 +50,12 @@ mod tests;
 pub const CONSENSUS_NODE_PREFIX: &str = "consensus";
 pub const EXECUTION_NODE_PREFIX: &str = "execution";
 
+pub fn metric_name(metric: &str) -> &str {
```
SuperFluffy: Why this change? Seems orthogonal to the bump at first glance?

crates/primitives/src/ed25519.rs
```diff
@@ -11,106 +11,101 @@ impl core::fmt::Display for InvalidPublicKey {
...
     }
 }
 
 impl TryFrom<B256> for PublicKey {
     type Error = InvalidPublicKey;
 
     fn try_from(value: B256) -> Result<Self, Self::Error> {
-        let inner = VerificationKeyBytes::from(<[u8; 32]>::from(value))
-            .try_into()
+        let key = commonware_cryptography::ed25519::PublicKey::read(&mut value.as_slice())
```
SuperFluffy: CW is now providing native From/TryFrom. Do you want to leave that as a follow-up?

deny.toml
```diff
@@ -15,6 +15,11 @@ ignore = [
   # https://rustsec.org/advisories/RUSTSEC-2026-0173 proc-macro-error2 is unmaintained
   # and is pulled in transitively by upstream crates with no safe upgrade available.
   "RUSTSEC-2026-0173",
+  # https://rustsec.org/advisories/RUSTSEC-2025-0055 tracing-subscriber 0.2
```
SuperFluffy, line 18: we should raise this to commonware

crates/primitives/src/ed25519.rs
SuperFluffy, line 1: IMO fine, but we should have a close followup to use commonware's new impls to have a more convenient way to convert from/to `[u8; 32]`.

crates/e2e/src/testing_node.rs
```diff
@@ -283,7 +323,12 @@ where
         let engine = self
             .consensus_config
             .clone()
-            .try_init(context.with_label(&format!("{}_{}", self.uid, self.n_starts)))
+            .try_init(
+                context
+                    .child("consensus")
+                    .with_attribute("uid", &self.uid)
```
SuperFluffy: guess we now need to extra these?

crates/e2e/src/testing_node.rs
```diff
@@ -79,6 +79,10 @@ where
     /// The chain address of the node. Used for executing validator-config smart
     /// contract calls.
     pub chain_address: Address,
+    /// Raw encoded finalization to write to the bootstrap certificate path
+    /// before the next start. Used when tests attach an already-finalized EL
+    /// database to fresh consensus storage.
+    bootstrap_finalization: Option<Vec<u8>>,
```
SuperFluffy: Can we type this?

crates/consensus/src/metrics.rs
```diff
@@ -27,10 +27,11 @@ pub fn install(context: Context, listen_addr: SocketAddr) -> Handle<eyre::Result
             .wrap_err("failed to bind provided address")?;
 
         // Create a router for the metrics server
+        let context = Arc::new(context);
```
SuperFluffy, line 30: what made this necessary? `Context` no longer being `Clone`?

crates/consensus/src/alias.rs
```diff
@@ -144,53 +156,156 @@ pub(crate) mod marshal {
...
-        }
-
         Ok(Initialized {
             actor,
             mailbox,
             last_finalized_height,
         })
     }
+
+    async fn start<TContext>(
```
SuperFluffy: document what this is doing IMO

crates/consensus/src/alias.rs
```diff
@@ -2,46 +2,67 @@
 //! (primarily commonware) types.
 
 pub(crate) mod marshal {
```
SuperFluffy, line 4: this module is becoming so big, we should probably move it to a separte file

crates/consensus/src/bootstrap.rs
```diff
@@ -0,0 +1,43 @@
...
+pub(crate) const FINALIZATION_PATH: &str = "bootstrap/finalization.cert";
+
+pub(crate) fn read_finalization(
+    consensus_dir: &Path,
+) -> eyre::Result<Option<Finalization<Scheme<PublicKey, MinSig>, Digest>>> {
+    let path = consensus_dir.join(FINALIZATION_PATH);
+    if !std::fs::exists(path.clone()).wrap_err("could not determine file existence")? {
+        return Ok(None);
+    }
+
```
SuperFluffy: Don't check for file existance - try reading it and check the error:

```rust
match std::fs::read(path) {
    Err(error), if error.kind() == ErrorKind::NotFound => return Ok(None),
    ...
}
```

crates/consensus/src/alias.rs
```diff
@@ -144,53 +156,156 @@ pub(crate) mod marshal {
...
+            finalized_block_num == 0,
+            "Genesis start with finalized execution state; finalized height: `{finalized_block_num}` \
+            Finalization certificates or bootstrap finalization must be available.",
+        );
+
+        let genesis_hash = execution_node.chain_spec().genesis_hash();
+        let genesis_block = execution_node
+            .provider
+            .block_by_hash(genesis_hash)?
+            .expect("genesis block must exist");
```
SuperFluffy: IMO just return an error since we are exiting anyway.

crates/consensus/src/alias.rs
```diff
@@ -144,53 +156,156 @@ pub(crate) mod marshal {
...
+        finalized_blocks: &FinalizedBlocks<TContext>,
+        execution_node: &TempoFullNode,
+        config: &Config,
+        epoch: Epoch,
+    ) -> eyre::Result<()>
+    where
+        TContext:
+            Clock + Metrics + Spawner + Storage + BufferPooler + Rng + CryptoRng + Send + 'static,
+    {
+        if config.scheme_provider.scoped(epoch).is_some() {
```
SuperFluffy: When would this ever be the case at startup?

crates/consensus/src/alias.rs
```diff
@@ -144,53 +156,156 @@ pub(crate) mod marshal {
...
+            return Ok(());
+        }
+
+        let height = epoch.previous().map_or_else(Height::zero, |previous| {
+            config
+                .epoch_strategy
+                .last(previous)
+                .expect("epoch strategy is for all epochs")
```
SuperFluffy: technically "all heights" in this case

crates/consensus/src/alias.rs
```diff
@@ -144,53 +156,156 @@ pub(crate) mod marshal {
...
+
+        let extra_data = if let Some(block) = finalized_blocks
+            .get(Identifier::Index(height.get()))
+            .await?
+        {
+            block.extra_data().clone()
+        } else {
+            execution_node
+                .provider
+                .header_by_number(height.get())
```
SuperFluffy: This needs to ensure that the header is finalized.

crates/consensus/src/alias.rs
```diff
@@ -144,53 +156,156 @@ pub(crate) mod marshal {
...
+                .wrap_err("failed reading bootstrap finalization")?
+        };
+
+        let Some(finalization) = finalization else {
+            let genesis = genesis_start(execution_node)?;
+            return Ok((last_finalized_height, Start::Genesis(genesis)));
+        };
+
+        let epoch = finalization.epoch();
+        register_outcome(finalized_blocks, execution_node, config, epoch).await?;
```
SuperFluffy: is this correct? This function still takes the key read from `execution_node` as the source of truth. But in order to verify the cert, we need to pass in an "external" key (like the one we bake into the binary).

crates/consensus/src/consensus/engine.rs
SuperFluffy: Explain? Is this something we need to cover?

crates/consensus/src/consensus/engine.rs
SuperFluffy: It's fine IMO - nobody is looking at subblocks right now.

crates/consensus/src/dkg/manager/actor/mod.rs
```diff
@@ -277,6 +286,7 @@ where
                 )
             })?;
 
+        let ancestry_ctx = Arc::new(self.context.child("ancestry_stream"));
```
SuperFluffy, line 315: No need to assign assign a variable here.

`Arc::new` at the call site below.

crates/consensus/src/dkg/manager/actor/mod.rs
```diff
@@ -431,14 +441,20 @@ where
                                 )
                                 .await
                             {
-                                let stream = match self.config.marshal.ancestry((None, hole)).await {
+                                let stream = match self
+                                    .config
+                                    .marshal
+                                    .ancestry(
+                                        ancestry_ctx.clone(),
```
SuperFluffy, line 467: Just do `Arc::new(context.child("..."))` here so we don't have that extra variable up there.

crates/consensus/src/dkg/manager/actor/mod.rs
```diff
@@ -726,17 +754,15 @@ where
                 logs.record(k.clone(), v.clone());
             }
 
+            let ctx_mut = self.context.as_present_mut();
```
SuperFluffy, line 805: Can we no longer pass `&mut self.context` in directly?

deny.toml
hamdiallam: Agreed

crates/consensus/src/alias.rs
hamdiallam: Ah nope, nice catch

crates/consensus/src/alias.rs
hamdiallam: I think we still want to read from the execution node so that participants are set as well in the scheme.

Since the snapshot ensures the finalization cert must be <= reth's finalized tip, we just need to ensure the digest in the EL matches prior to trusting it as the source of truth. This part is missing

hamdiallam: This is actually a good callout. `verify_block` in the application actor will error if the participants are unknown. When registering the scheme, I wasn't setting the participants either.


Subtle edge case for any validator verifying blocks in the latest epoch after restart

hamdiallam: Pushed some changes. I dont think it's worth exiting early if it mismatches with the embedded network identity. The reason being on redkg, the node will fail on restarts until the operator intervenes. We still want the nodes to fall back to boundary-to-boundary progression with the node emitting warnings.

Bit worried if that if a simple restart exits the node, the cascading effects of this across validators could be bad.

But I added checks ensure the EL/CL is in a proper state with respect to the parsed finalization:
1. The finalization cert digest is present in execution storage (cl archives OK)
2. The finalization epoch matches up with the retrieved header height
3. The finalization can be verified with the outcome of the previous boundary.



Note that the network identity integration for the validators is still a todo atm. Going to address this in a followup to this PR

crates/e2e/src/lib.rs
hamdiallam: ah in this bump, `ctx.encode()` now also includes the labels in addition to the name which broke how we matched against the metric. This util parses out the name

crates/consensus/src/consensus/engine.rs
hamdiallam: I'll double check if any of our dashboards look at resolver metrics. Also renaming this to `p2p` since internally within `p2p::init`, the function does a `ctx.child('resolver')` call. It also now instantiates a child context for the handler as well


The rationale is that this init is taking ownership of the context so it should have it's own scope and not conflict with anything that may create more children at this level

crates/primitives/src/ed25519.rs
hamdiallam: Small enough to incorporate here. And we bumped to the new revision to get these utilities. Updated!

crates/e2e/src/testing_node.rs
hamdiallam: Yep since the raw metrics include the labels the tests now just `.contains(id)` on the metric. Removed the `metric_prefix()` that existed on the TestingNode.


The way we interact with the metrics in the e2e tests needs a refresh. Going to clean this up in a followup.

crates/consensus/src/consensus/engine.rs
```diff
@@ -82,6 +82,7 @@ pub struct Builder<TBlocker, TPeerManager> {
     pub with_subblocks: bool,
 
     pub feed_state: crate::feed::FeedStateHandle,
+    pub consensus_dir: PathBuf,
```
SuperFluffy: `consensus_dir` in the context of commonware is confusing: we already set the consensus dir when constructing the cw tokio runtime. Having an extra setting `consensus_dir` here is strange.

IMO, let's either provide an explicit `math_to_manifest` or use `<context as Storage>::open` to open the manifest for reading.

crates/consensus/src/consensus/engine.rs
```diff
@@ -476,8 +480,11 @@ where
         let peer_manager = self.peer_manager.start();
 
         let broadcast = self.broadcast.start(broadcast_channel);
-        let resolver =
-            marshal::resolver::p2p::init(&self.context, self.resolver_config, marshal_channel);
+        let resolver = marshal::resolver::p2p::init(
+            self.context.child("p2p"),
```
SuperFluffy, line 484: This is appropriate - but it is a breaking change for telemetry. Do we have to do this right now or can we move it to a separate PR? If yes, let's do that.

crates/consensus/src/dkg/manager/actor/mod.rs
```diff
@@ -556,11 +578,17 @@ where
...
-            }
+            // TODO(hamdi): We're relying on the fact that when in catch up we `hint_finalized` the
+            // boundary height in the epoch manager thus the finalization for this height should
+            // eventually be resolved.
+            //
+            // This deserves some clean up to consolidate the hint and set_floor. As well as factoring
+            // in the network idenetity in scenarios where jumping from boundary to boundary is no longer
+            // possible (no cert/block among peers).
+            match self.config.marshal.get_finalization(boundary_height).await {
+                Some(finalization) => self.config.marshal.set_floor(finalization),
```
SuperFluffy: Uff, good callout. This definitely needs a cleanup (and the whole `set_floor` logic should live outside the DKG actor; the DKG actor should just be able to "deal with it").

crates/consensus/src/dkg/manager/actor/mod.rs
```diff
@@ -981,40 +998,32 @@ where
...
 
-                    if let Err(error) = round_channel
-                        .send(
-                            Recipients::One(from.clone()),
-                            Message::Ack(ack).encode(),
-                            true,
-                        )
-                        .await
-                    {
-                        // FIXME(janis): the GATs in the Sender (and LimitedSender)
```
SuperFluffy, line 972: They fixed this? Nice.

crates/consensus/src/dkg/manager/actor/mod.rs
```diff
@@ -1322,9 +1327,6 @@ where
         Some(share)
     });
 
-    info!(%latest_boundary, "setting sync floor");
-    marshal.set_floor(Height::new(latest_boundary)).await;
```
SuperFluffy: Just curious - why?

crates/consensus/src/dkg/manager/actor/mod.rs
```diff
@@ -1354,101 +1356,81 @@ struct Metrics {
...
         );
 
-        let rounds_skipped = Counter::default();
-        context.register(
+        let rounds_skipped = context.counter(
             "rounds_skipped",
             "how many DKG rounds were skipped because the node fell too far behind and tried to catch up",
-            rounds_skipped.clone(),
         );
+        let ancestor_fetch_duration = Timed::new(context.histogram(
```
SuperFluffy, line 1468: Very confusing to have this in `chore` PR. Why is this necessary? For the ancestor stream? We should explain that here.

crates/consensus/src/follow/resolver.rs
```diff
@@ -81,18 +84,26 @@ pub(crate) struct Config {
...
     pub(super) upstream: super::upstream::Mailbox,
-    pub(super) mailbox_size: usize,
+    pub(super) mailbox_size: NonZeroUsize,
+}
+
+type FetchPool = AbortablePool<(Key<Digest>, Result<Bytes, bool>)>;
+type DeliveryPool = Pool<(Key<Digest>, bool)>;
+
+struct PendingRequest {
+    subscribers: BTreeSet<Annotation>,
```
SuperFluffy, line 88: I guess this is mainly because of commonware's marshal?

We don't really have a list of subscribers.

crates/consensus/src/follow/engine.rs
```diff
@@ -50,6 +50,9 @@ pub struct Config<TUpstream> {
     /// Partition prefix for storage.
     pub partition_prefix: String,
 
+    /// Consensus storage directory used for bootstrap material.
+    pub consensus_dir: PathBuf,
```
SuperFluffy: mentioned for the consensus engine - I think having this here is confusing. We should try and access this via `context`

crates/consensus/src/follow/engine.rs
SuperFluffy: Make this thing `NUsize!`

crates/consensus/src/follow/engine.rs
```diff
@@ -107,18 +109,19 @@ impl<TUpstream> Config<TUpstream> {
...
             last_finalized_height,
         } = alias::marshal::init(
-            context.clone(),
+            &context,
             page_cache_ref,
             self.execution_node.clone(),
             alias::marshal::Config {
                 partition_prefix: self.partition_prefix.clone(),
-                mailbox_size: self.mailbox_size,
+                mailbox_size: NZUsize!(self.mailbox_size),
```
SuperFluffy: Make `Config::mailbox_size: NonZeroUsize` instead of all these calls here.

crates/consensus/src/follow/stubs.rs
```diff
@@ -32,7 +34,7 @@ pub(super) fn null_broadcast<E: Clock + Spawner + Metrics + BufferPooler>(
 
     let config = buffered::Config {
         public_key,
-        mailbox_size,
+        mailbox_size: NonZeroUsize::new(mailbox_size).expect("mailbox size must be non-zero"),
```
SuperFluffy: likewise ... just make this thing `NonZeroUsize` in config.

crates/consensus/src/consensus/engine.rs
hamdiallam: Has to be done since p2p::init does not take a context reference but owns the passed context

crates/consensus/src/dkg/manager/actor/mod.rs
hamdiallam: Yea the Rng traits aren't forwarded anymore from the Cell wrapper

crates/consensus/src/dkg/manager/actor/mod.rs
hamdiallam: Ah the reason was so that the reference was actually shared. Instead of having to create two distinct child contexts with the same label

crates/consensus/src/feed/actor.rs
```diff
@@ -227,18 +232,34 @@ impl<TContext: Spawner> Actor<TContext> {
...
         // Update state and broadcast events
+        let height = certified.block.inner.number;
+        let subscribers = self.state.events_tx().receiver_count();
         match activity {
             Activity::Notarization(_) => {
-                let _ = self.state.events_tx().send(Event::Notarized {
-                    block: certified.clone(),
-                    seen: now_millis(),
-                });
+                if latest_notarized_height
```
hamdiallam: Only send increasing updates

crates/consensus/src/feed/actor.rs
```diff
@@ -248,14 +269,15 @@ impl<TContext: Spawner> Actor<TContext> {
...
             Activity::Finalization(_) => {
-                debug!(
-                    subscribers = self.state.events_tx().receiver_count(),
-                    "sending finalized event",
-                );
-                let _ = self.state.events_tx().send(Event::Finalized {
-                    block: certified.clone(),
-                    seen: now_millis(),
-                });
+                if latest_finalized_height
```
hamdiallam: Only send increasing updates

crates/consensus/src/storage/mod.rs
```diff
@@ -188,64 +186,3 @@ where
 
     archive
 }
-
```
hamdiallam, line 191: dead code that was leftover

PR #5139, opened 2026-06-01
fix(precompiles): reject state changes in static call context (#4104)

### LINKED ISSUE #4104: sstore/tstore/emit_event: is_static flag never checked in EvmPrecompileStorageProvider
## Summary

`EvmPrecompileStorageProvider` in `tempo/crates/precompiles/src/storage/evm.rs` stores `is_static` from the call context but never checks it in any of the three state-mutating methods: `sstore`, `tstore`, and `emit_event`.

If a Tempo precompile is invoked via `STATICCALL`, all three operations silently succeed and mutate EVM state. No error is returned, no gas is refunded — the mutation just happens.

## Affected methods

- `sstore` — writes persistent storage
- `tstore` — writes transient storage  
- `emit_event` — emits a log

All three are state-modifying operations. None of them guard against `self.is_static`.

## Spec

This is a direct violation of [EIP-214](https://eips.ethereum.org/EIPS/eip-214) (STATICCALL), which specifies:

> Any attempts to make state-changing operations inside an execution instance with `STATIC` set to `true` will instead throw an exception. These operations include `CREATE`, `CREATE2`, `LOG0`–`LOG4`, `SSTORE`, and `SELFDESTRUCT`.

`TSTORE` is also covered under EIP-1153, which applies the same static-context restriction.

The Yellow Paper formalises the `STATIC` execution environment flag introduced by EIP-214; once set, any state modification must result in an exceptional halt.

## Impact

A contract can invoke a Tempo precompile via `STATICCALL` and cause persistent or transient storage writes and log emissions that the caller's execution context assumes cannot happen. This breaks the core invariant that `STATICCALL` provides to callers: that no state has changed after the call returns.

Zhekinmaksim:
Closes #4104.

## Summary

`EvmPrecompileStorageProvider` stored `is_static` from the call context but never checked it in the state-mutating methods. A precompile invoked via `STATICCALL` could mutate persistent storage, transient storage, account code, and logs, which violates the invariant provided by `STATICCALL` and the static-context restriction for `TSTORE`.

## Changes

- Add `TempoPrecompileError::StateChangeInStaticCall` as a non-domain runtime error with no ABI selector.
- Route the new variant through the existing fatal precompile error path so it propagates as an exceptional halt rather than a Solidity-style revert.
- Guard `sstore`, `tstore`, `emit_event`, and `set_code` as the first statement in each method, before gas accounting or state access.
- Add `TestEvm::provider_static()` plus tests covering each rejected mutation and a positive control confirming `sload` still works in static context.

## Why the guard is first

The static-context check belongs before gas deduction and before loading state. That matches EVM execution semantics: mutation opcodes are rejected because the frame is static, not because execution partially proceeds and then rolls back.

## Verification

- `cargo fmt --all -- --check`
- `cargo test -p tempo-precompiles` — 926 passed, 1 ignored; doctests ok
- `cargo test -p tempo-precompiles static_context` — 5 passed
- `cargo clippy -p tempo-precompiles --all-targets -- -D warnings`

## Reviewer note

The new error currently maps to `PrecompileError::Fatal` for halt semantics. If maintainers prefer a dedicated `PrecompileHalt` variant mirroring revm's static-call halt reason, I can switch the representation in a follow-up commit.

### DIFF
--- a/crates/precompiles/src/error.rs
+++ b/crates/precompiles/src/error.rs
@@ -107,6 +107,14 @@ pub enum TempoPrecompileError {
     #[error("Gas limit exceeded")]
     OutOfGas,
 
+    /// A state-changing operation was attempted while the precompile was
+    /// invoked in a static call context (EIP-214 / EIP-1153).
+    ///
+    /// Returned by state-mutating storage provider methods when the provider
+    /// was constructed with `is_static = true`.
+    #[error("State change attempted in static call context (EIP-214)")]
+    StateChangeInStaticCall,
+
     /// The calldata's 4-byte selector does not match any known precompile function.
     #[error("Unknown function selector: {0:?}")]
     UnknownFunctionSelector([u8; 4]),
@@ -168,7 +176,7 @@ impl TempoPrecompileError {
             Self::StorageCreditsError(e) => e.selector(),
             Self::UnknownFunctionSelector(selector) => *selector,
             Self::Panic(_) | Self::StorageDeltaUnderflow(_) => Panic::SELECTOR,
-            Self::OutOfGas | Self::Fatal(_) => [0, 0, 0, 0],
+            Self::OutOfGas | Self::StateChangeInStaticCall | Self::Fatal(_) => [0, 0, 0, 0],
         }
         .into()
     }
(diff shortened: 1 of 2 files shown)


Zhekinmaksim:
Hi Tempo team, quick bump on this PR.

It is currently mergeable and has no review comments or requested changes. The remaining blocker appears to be review approval.

Local validation completed:

- `cargo fmt --all -- --check`
- `cargo test -p tempo-precompiles`
- `cargo test -p tempo-precompiles static_context`
- `cargo clippy -p tempo-precompiles --all-targets -- -D warnings`

Happy to adjust the error representation if you prefer a dedicated `PrecompileHalt` variant instead of the current `PrecompileError::Fatal` path.

PR #5141, opened 2026-06-01
fix: check 403 mint policy for blocked mints

0xalpharush:
To discuss ZELLIC-218


0xalpharush:
decided it is considered a transfer

PR #5143, opened 2026-06-01
perf(pool): use entry for expiring nonce inserts

mattsse:
Uses the expiring nonce transaction map entry API to reject duplicate expiring nonce hashes and insert through the vacant entry. Notification is delayed until after the pool indexes and metrics are updated, avoiding the extra `contains_key` lookup on successful imports.

### DIFF
--- a/crates/transaction-pool/src/tt_2d_pool.rs
+++ b/crates/transaction-pool/src/tt_2d_pool.rs
@@ -4,7 +4,7 @@ use crate::{
 };
 use alloy_primitives::{
     Address, B256, TxHash, U256,
-    map::{AddressMap, B256Map, HashMap, HashSet, U256Map},
+    map::{AddressMap, B256Map, HashMap, HashSet, U256Map, hash_map},
 };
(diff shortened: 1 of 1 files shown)


crates/transaction-pool/src/tt_2d_pool.rs
```diff
@@ -404,19 +405,24 @@ impl AA2dPool {
 
         // Create pending transaction
         let pending_tx = PendingTransaction {
-            submission_id: self.next_id(),
+            submission_id: {
+                let id = self.submission_id;
+                self.submission_id = self.submission_id.wrapping_add(1);
+                id
+            },
```
onbjerg, line 412: why this change

shekhirin (approved):

crates/transaction-pool/src/tt_2d_pool.rs
mattsse: becaus borrow checka doesnt allow &mut self simultaniously

PR #5154, opened 2026-06-01
fix(follower): don't hard fail when consensus block is missing

hamdiallam:
We can skip past epoch boundaries when we see a future finalization verified by the network identity. Thus on restart the last consensus boundary may not be available.

We should rely on the network identity fallback to continue chain progress rather than halting

### DIFF
--- a/crates/commonware-node/src/follow/driver.rs
+++ b/crates/commonware-node/src/follow/driver.rs
@@ -265,8 +265,8 @@ where
                 warn!(
                     "cannot heal finalization gap; consensus layer epoch {consensus_epoch} is ahead \
                     of execution layer epoch {execution_epoch}, but the consensus layer does not have \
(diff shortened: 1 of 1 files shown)


crates/commonware-node/src/follow/driver.rs
```diff
@@ -256,19 +257,20 @@ where
...
-                        `{last_consensus_boundary}`"
-                    )
-                })?;
+
+            let Some(boundary_block) = self.config.marshal.get_block(last_consensus_boundary).await
+            else {
+                let consensus_epoch = current_consensus_epoch.epoch();
+                let execution_epoch = current_execution_epoch.epoch();
+                warn!(
+                    "cannot heal finalization gap; consensus layer epoch {consensus_epoch} is ahead \
```
SuperFluffy, line 266: Mention that the node will try to fix this by using the network identity?

SuperFluffy (approved):

PR #5161, opened 2026-06-01
docs(tip-1063): clarify feature dependency activation

emmajam:
Documents dependency bitmaps, active feature audit order, activation-time quorum/dependency checks, and validator admission behavior for active features.


tips/tip-1063.md
```diff
@@ -51,42 +53,61 @@ contract FeatureRegistry {
...
+
+The dependency bitmap uses feature IDs as bit indexes. For a dependency on feature ID `x`, the bit at `dependency_bitmap[x / 256] & (1 << (x % 256))` is set. Missing words are treated as zero, and trailing zero words are omitted. A zero-length bitmap means the feature has no dependencies. A feature's dependencies are satisfied only when `(active_feature_bitmap & dependency_bitmap) == dependency_bitmap` for every bitmap word.
+
+The registry must reject a feature registration whose dependency bitmap includes its own feature ID, references an unregistered dependency, or creates a dependency cycle. This keeps the feature graph a DAG and ensures the active feature set can be validated as a closed set under the dependency relation.
+
+Pending feature ordering is not consensus state. Source-controlled metadata and binary feature evaluation order can change before a feature activates. If feature B activates before inactive feature A, the chain records B first in `active_feature_order`; a later release should reorder A after B before A activates, but B's release and activation are not blocked solely because A appeared earlier in an inactive release.
```
emmajam: @klkvr can you check this

tips/tip-1063.md
```diff
@@ -26,23 +26,25 @@ Feature flags separate code distribution from feature activation. Validators upg
...
 
-A feature can be scheduled for activation once enough active validators have reported a Tempo version that supports it. A designated multisig schedules the feature by setting a future activation epoch in an onchain feature registry. When the feature is scheduled, Tempo validates that enough active validators support the feature, and the activation epoch becomes the upgrade deadline. When the chain crosses into the scheduled epoch, the feature becomes active. Validators that have not upgraded must stop before processing blocks in epochs that require the active feature.
+A designated multisig schedules a feature by setting a future activation epoch in an onchain feature registry. Scheduling announces the intended activation boundary and makes the activation epoch the upgrade deadline. When the chain crosses into the scheduled epoch, Tempo rechecks the current active validator set for quorum and validates the feature's dependency bitmap against the active feature bitmap. If both checks pass, the feature becomes active. Validators that have not upgraded must stop before processing blocks in epochs that require the active feature.
```
jenpaff: > When the chain crosses into the scheduled epoch, Tempo rechecks the current active validator set for quorum and validates the feature's dependency bitmap against the active feature bitmap. 

still riffing, would re-phrase this slightly and make it super clear. 2 checks happen: 1) Has the feature reached quorum, 2) if a feature has dependencies, have they been resolved correctly

jenpaff:
```suggestion
A designated multisig schedules a feature by setting a future activation epoch in an onchain feature registry. Scheduling announces the intended activation boundary and makes the activation epoch the upgrade deadline. On activation, two requirements must be true: 1) The feature has reached quorum, 2) If a feature has configured dependencies, those features MUST be activated before. If both checks pass, the feature becomes active. Validators that have not upgraded must stop before processing blocks in epochs that require the active feature.
```

tips/tip-1063.md
```diff
@@ -26,23 +26,25 @@ Feature flags separate code distribution from feature activation. Validators upg
...
 
-A feature can be scheduled for activation once enough active validators have reported a Tempo version that supports it. A designated multisig schedules the feature by setting a future activation epoch in an onchain feature registry. When the feature is scheduled, Tempo validates that enough active validators support the feature, and the activation epoch becomes the upgrade deadline. When the chain crosses into the scheduled epoch, the feature becomes active. Validators that have not upgraded must stop before processing blocks in epochs that require the active feature.
+A designated multisig schedules a feature by setting a future activation epoch in an onchain feature registry. Scheduling announces the intended activation boundary and makes the activation epoch the upgrade deadline. At activation, Tempo evaluates two checks: the feature has reached activation quorum, and every configured dependency is already active. If both checks pass, the feature becomes active. Validators that have not upgraded must stop before processing blocks in epochs that require the active feature.
```
Zygimantass: designated multisig -> designated owner(?)

Zygimantass: at activation -> is this on the activation epoch or when we call activateFeature (or whatever the function is)?

Zygimantass: i would do when we call activateFeature

tips/tip-1063.md
```diff
@@ -51,42 +53,66 @@ contract FeatureRegistry {
   // --snip--
```
Zygimantass, line 53: why do we need the feature_id if we already have it in the mapping? dependency_bitmap also feels like a smell here, my thought would be to just have 

feature_bitmap and then the feature_id is the index in the bitmap

PR #5212, opened 2026-06-02
fix(dkg): preserve epoch logs on restart

joshieDo:
Fixes a restart path where marshal startup could set its sync floor to reth's finalized height. When that height is inside the active DKG epoch, marshal skips replaying headers that contain public DKG dealer logs, so the DKG actor can restart with incomplete local ceremony state and later fail finalization against a valid block outcome.

### DIFF
--- a/crates/commonware-node/src/alias.rs
+++ b/crates/commonware-node/src/alias.rs
@@ -22,7 +22,7 @@ pub(crate) mod marshal {
     use reth_node_builder::NodeTypesWithDBAdapter;
     use reth_provider::providers::BlockchainProvider;
     use tempo_node::{TempoFullNode, node::TempoNode};
-    use tracing::{info, instrument};
+    use tracing::instrument;
 
     use crate::{
         consensus::{Digest, block::Block},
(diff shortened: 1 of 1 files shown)


crates/commonware-node/src/alias.rs
```diff
@@ -172,25 +169,13 @@ pub(crate) mod marshal {
...
-        // re-sync history that the execution layer already finalized. The
-        // mailbox message is buffered until the actor starts; `set_floor` only
-        // ever advances, so sending it unconditionally is safe.
         let reth_finalized_height = execution_node
             .provider
             .canonical_in_memory_state()
             .get_finalized_num_hash()
             .map(|nh| nh.number)
             .unwrap_or(0);
         let last_finalized_height = marshal_stored_height.max(Height::new(reth_finalized_height));
```
joshieDo, line 178: unsure what returning this will do

hamdiallam: lets return the marshal reported height since we'll be replaying blocks from there. 

Since we were setting the floor previously then we needed the max which is where the updates would start from

PR #5215, opened 2026-06-02
docs(tip-1070): current committee state

SuperFluffy:


tips/tip-1070.md
```diff
@@ -0,0 +1,120 @@
...
+# Specification
+
+## State Location
+
+Committee state MUST live in dedicated execution-layer state for the current effective committee.
+
+This TIP defines a network parameter:
+
+```solidity
+address constant CURRENT_COMMITTEE_ADDRESS = TBD;
```
hamdiallam, line 53: Thoughts on extending ValidatorConfig so we only have 1 contract? 

ValidatorConfig#getActiveValidators -> Currently cctive Committee
ValidatroConfig#getNextEpochValidators -> Registration based

tips/tip-1070.md
```diff
@@ -0,0 +1,120 @@
...
+    returns (uint64 epoch, bytes32[] memory publicKeys);
+```
+
+`getCommitteeMembers()` MUST return the committee used by consensus for the epoch containing the block at which the call is evaluated.
+
+The returned `publicKeys` MUST be ordered exactly as the `players` vector in the DKG outcome selected for that epoch. If consensus falls back to a previous DKG outcome, the returned `publicKeys` MUST match the fallback outcome.
+
+## System Transaction Execution
+
+When proposing the last block of epoch `E`, the proposer already writes the DKG outcome for epoch `E+1` to the block header. At the same boundary, the proposer MUST execute a system transaction that persists the effective committee for epoch `E+1` to execution-layer state.
```
hamdiallam: If in the same boundary block, we should enforce that it's the last transaction in the block. Otherwise the state is visible in epoch E when it should only be visible in E+1

SuperFluffy: Good catch. Changed in [ee7b5db](https://github.com/tempoxyz/tempo/pull/5215/commits/ee7b5db67ad6dc1126a7293d796939e80b82e775)

tips/tip-1070.md
SuperFluffy: I prefer to not touch `ValidatorConfigV2` in this TIP - also because in principle this list can remain independent of any particular val cfg contract version.

I am going to leave this conversation unresolved in case there are other opinions.

tips/tip-1070.md
```diff
@@ -0,0 +1,122 @@
...
+/// @notice Emitted when the effective committee for an epoch is persisted.
+/// @param epoch Epoch for which the committee is effective.
+/// @param membersHash keccak256 hash of the ABI-encoded committee public keys in order.
+/// @param memberCount Number of committee members.
+event CommitteeUpdated(
+    uint64 indexed epoch,
+    bytes32 indexed membersHash,
+    uint64 memberCount
+);
+```
```
mattsse: we cant anchor a receipt without an actual transaction.

if we need events then we'd need an actual transaction.

SuperFluffy: I wasn't sure about this in the first place - after all, it's an extremely rare event which, at the same time, is known to take place on fixed blocks.

Your comment cements the decision that we should just remove them: [b2a6eee](https://github.com/tempoxyz/tempo/pull/5215/commits/b2a6eeeb5822d09e0ad5fd641358da95734d68a5)

tips/tip-1070.md
```diff
@@ -0,0 +1,101 @@
...
+/// @dev This returns the DKG outcome players selected by consensus, not the registry's active validator set.
+/// @return epoch Epoch for which the returned committee is effective.
+/// @return publicKeys Effective committee member public keys, in DKG outcome player order.
+function getCommitteeMembers()
+    external
+    view
+    returns (uint64 epoch, bytes32[] memory publicKeys);
+```
+
+`getCommitteeMembers()` MUST return the committee used by consensus for the epoch containing the block at which the call is evaluated.
```
grandizzy: should we be concerned re gas if the number of members increases? (would be good to have the gas cost/schedule for these reads specified in TIP) if so maybe let's add `getCommitteeSize()` + `getCommitteeMember(uint64 index)` to be able to read single member

tips/tip-1070.md
```diff
@@ -0,0 +1,101 @@
...
+
+The returned `publicKeys` MUST be ordered exactly as the `players` vector in the DKG outcome selected for that epoch. If consensus falls back to a previous DKG outcome, the returned `publicKeys` MUST match the fallback outcome.
+
+## System Call Execution
+
+When proposing the last block of epoch `E`, the proposer already writes the DKG outcome for epoch `E+1` to the block header. At the same boundary, the proposer MUST execute a system call that persists the effective committee for epoch `E+1` to execution-layer state.
+
+The system call MUST run from a post-execution hook, after all user transactions in the boundary block have executed. Transactions in the boundary block MUST continue to observe the committee for epoch `E`; the persisted committee for epoch `E+1` MUST only become visible to EVM execution in the first block of epoch `E+1`.
```
0xalpharush: Should we specify the following conditions like https://eips.ethereum.org/EIPS/eip-2935?
> This is a system operation following the same convention as [EIP-4788](https://eips.ethereum.org/EIPS/eip-4788) and therefore:
> 
> the call must execute to completion
> the call does not count against the block’s gas limit
> the call does not follow the [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559) burn semantics - no value should be transferred as part of the call
> if no code exists at HISTORY_STORAGE_ADDRESS, the call must fail silently

tips/tip-1070.md
```diff
@@ -0,0 +1,101 @@
...
+    returns (uint64 epoch, bytes32[] memory publicKeys);
+```
+
+`getCommitteeMembers()` MUST return the committee used by consensus for the epoch containing the block at which the call is evaluated.
+
+The returned `publicKeys` MUST be ordered exactly as the `players` vector in the DKG outcome selected for that epoch. If consensus falls back to a previous DKG outcome, the returned `publicKeys` MUST match the fallback outcome.
+
+## System Call Execution
+
+When proposing the last block of epoch `E`, the proposer already writes the DKG outcome for epoch `E+1` to the block header. At the same boundary, the proposer MUST execute a system call that persists the effective committee for epoch `E+1` to execution-layer state.
```
0xalpharush: Is it possible for this to be skipped or the following proposer would re-try?

tips/tip-1070.md
```diff
@@ -0,0 +1,101 @@
...
+- Validators verify the DKG outcome and a quorum of validators must arrive at the same outcome. This is ensured by Tempo's use of Commonware's Feldman/Desmedt-style DKG implementation.
+- If a DKG round fails, the effective committee for the next epoch is the previous successful DKG outcome selected by consensus.
+- The execution layer can persist enough committee state at the boundary block to answer current-committee reads without requiring callers to parse historical block headers.
+- DKG outcome `players` are ordered Ed25519 public keys.
+
+## Threat Model
+
+- **Consensus proposer**: trusted only to execute the system call required by the protocol at the boundary block. Invalid committee writes are consensus-invalid.
+- **ValidatorConfigV2 owner**: can change configured validator state, but cannot directly change the effective committee except through the normal DKG process.
+- **Contract callers and offchain tools**: untrusted readers. They rely on this query as a canonical view of effective committee membership and must not need to infer committee state from `getActiveValidators()`.
```
malleshpai:
```suggestion
- **Contract callers and offchain tools**: untrusted readers. They rely on this query as a canonical view of effective committee membership and should not  infer committee state from `getActiveValidators()`.
```

tips/tip-1070.md
dankrad: I don't like the formulation "the proposer MUST". Should it not just be part of the VM itself? The first block MUST include the system transaction as the first transaction, any node MUST reject a block that does not have it as the first transaction

dankrad:
I feel like this is a bit underspecified. Which contract does this system transaction call, and what exact state variables does it set?

tips/tip-1070.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Motivation
+
+`ValidatorConfigV2.getActiveValidators()` is a recurring source of confusion because it returns the configured validator registry, not necessarily the committee currently active in consensus.
+Registry changes such as `addValidator`, `rotateValidator`, and `deactivateValidator` immediately affect the return value of `getActiveValidators` returns even if the changes have not yet been propagated to the committee.
+Likewise, the committee is only updated if the relevant DKG round succeeds; if DKG fails, the committee remains that of the previous epoch, thus potentially containing validators that are marked deactivated and which would therefore be excluded from `getActiveValidators`.
+
+Callers that need the current effective committee must therefore inspect consensus data or reconstruct it from boundary block headers.
+This TIP gives contracts a stable execution-layer query for that value while keeping it separate from `ValidatorConfigV2`: the registry remains the input to future DKG player selection, and the new committee state records the DKG result consensus actually selected for the current epoch.
```
malleshpai:
```suggestion
This TIP gives contracts a stable execution-layer query for that value while keeping it separate from `ValidatorConfigV2`: that registry remains the input to future DKG player selection, while the new committee state records the DKG result consensus actually selected for the current epoch.
```

tips/tip-1070.md
malleshpai: i think we should resolve system call vs system tx per @dankrad comment as well. 

also call/ tx at end of epoch e vs first block of e+1.

tips/tip-1070.md
klkvr: i think it should be fine? like yes it will cost a couple million gas if we ever get past 1000 validators but it is still manageable and seems far away enough

tips/tip-1070.md
```diff
@@ -0,0 +1,126 @@
+---
+id: TIP-1070
+title: Current Committee State
+description: Add canonical execution-layer state for the current effective committee selected by DKG.
+authors: Richard Janis Goldschmidt (@superfluffy)
```
SuperFluffy: Add @klkvr

tips/tip-1070.md
```diff
@@ -0,0 +1,126 @@
...
+function getCommitteeMembers()
+    external
+    view
+    returns (uint64 epoch, bytes32[] memory publicKeys);
+```
+
+`getCommitteeMembers()` MUST return the committee used by consensus for the epoch containing the block at which the call is evaluated.
+The returned `epoch` MUST be computed from the block number at which the call is evaluated and the chain's epoch schedule.
+
+The returned `publicKeys` MUST be ordered exactly as the `players` vector in the DKG outcome selected for that epoch. If consensus falls back to a previous DKG outcome, the returned `publicKeys` MUST match the fallback outcome.
```
SuperFluffy: To tighten this: the implementation should not run any inference on whether the DKG failed. If DKG fell back to the previous outcome is decided at the consensus level and opaque to whatever routine writes the participants.

To avoid confusion. The `players` vector in the DKG outcome always encodes the participants (committee members) of the rising epoch.

tips/tip-1070.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Contract Storage
+
+The committee-state component MUST persist only the current committee member public keys:
+
+```solidity
+bytes32[] publicKeys;
+```
+
+Implementations MUST NOT persist the epoch alongside the committee members. The epoch returned by `getCommitteeMembers()` is derived from the current block context.
```
SuperFluffy: Should we justify why?

klkvr: hmm i think maybe we should just remove this line. we clearly define the layout and it's not really intuitive to put epoch into there as well

tips/tip-1070.md
klkvr:
```suggestion
authors: Richard Janis Goldschmidt (@superfluffy), Arsenii Kulikov (@klkvr)
```

🫡

klkvr (dismissed):

0xKitsune (dismissed):

tips/tip-1070.md
```diff
@@ -0,0 +1,124 @@
...
+This system call follows the same execution convention as EIP-4788-style system operations:
+
+- the call MUST execute to completion;
+- the call MUST NOT count against the block gas limit;
+- the call MUST NOT transfer value or apply EIP-1559 burn semantics.
+
+The system call MUST call `CURRENT_COMMITTEE_ADDRESS` from the system address with calldata equivalent to:
+
+```solidity
+function setCommitteeMembers(bytes32[] calldata publicKeys) external;
```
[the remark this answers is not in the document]
0xKitsune: @SuperFluffy this seems valid? Lmk if I am overlooking something but should be a simple spec change.

0xKitsune:
Lgtm pending patching the cyclops findings:
https://github.com/tempoxyz/tempo/pull/5215#pullrequestreview-4594509484

tips/tip-1070.md
```diff
@@ -0,0 +1,124 @@
...
+
+```solidity
+bytes32[] publicKeys;
+```
+
+## System Call Execution
+
+The VM MUST execute the committee update as a system call while processing the last block of epoch `E`.
+
+The system call MUST run from a post-execution hook after all user transactions in the last block of epoch `E` have executed. It is not a user transaction and MUST NOT appear in the transaction list. Transactions in the last block of epoch `E` MUST continue to observe the committee for epoch `E`; the persisted committee for epoch `E+1` MUST only become visible to EVM execution in the first block of epoch `E+1`.
```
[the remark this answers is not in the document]
0xKitsune: Seems valid cc @SuperFluffy @klkvr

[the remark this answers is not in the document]
SuperFluffy: So scenario is:

1. Boundary block B is built
2. `getCommitteeMembers` is run right after B is finalized but before B+1 is finalized
3. `getCommitteMembers` returns `(E, committee(E+1))`.

Yeah, that's valid and should be fixed because it's confusing. I don't think this is an actual security issue considering that any smart contract calls that follow the state write are done producing a block in E+1, and hence would take into account the correct committee.

tips/tip-1070.md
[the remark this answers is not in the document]
SuperFluffy: Seems valid and I tacitly assumed that this would be the case

tips/tip-1070.md
```diff
@@ -0,0 +1,124 @@
...
+
+The system call MUST call `CURRENT_COMMITTEE_ADDRESS` from the system address with calldata equivalent to:
+
+```solidity
+function setCommitteeMembers(bytes32[] calldata publicKeys) external;
+```
+
+The system call MUST NOT pass an epoch argument.
+
+The system call MUST read the DKG outcome from the boundary block's extra data and persist:
```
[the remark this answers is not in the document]
SuperFluffy: I initially wrote this spec having in mind that CL would a) write a DKG outcome, and b) that the EVM would write the committee if and only if it detected a DKG outcome written in `extra_data` - meaning that the EL would be oblivious to notions of "epoch" or "boundary" other than what CL defines.

If we decide to make the EVM epoch aware (as is done by the implementation in #6209), then we probably need to define this in this spec here as well.

tips/tip-1070.md
0xrusowsky: i personally like the UX benefits of integrators only having to deal with the val config address. We could still always surface the commitee via `ValidatorConfigVX`'s abi, why simply wrapping a `CurrentCommiteePrecompile.getCommiteeMembers()` call

legion2002: also think we should just extend ValidatorConfigV2

tips/tip-1070.md
```diff
@@ -0,0 +1,124 @@
...
+## Interface
+
+The committee-state component exposes a current-only read query:
+
+```solidity
+/// @notice Get the effective committee members for the current epoch.
+/// @dev This returns the DKG outcome players selected by consensus, not the registry's active validator set.
+/// @return epoch Epoch for which the returned committee is effective.
+/// @return publicKeys Effective committee member public keys, in DKG outcome player order.
+function getCommitteeMembers()
```
legion2002, line 67: nit: maybe we should add another helper fn `isCommitteeMember`?

legion2002 (dismissed):
lgtm, think we should just move this as an extension to the validator config contract, just simpler to manage that way

tips/tip-1070.md
legion2002: we decided in slack to keep the `CurrentCommittee` precompile

tips/tip-1070.md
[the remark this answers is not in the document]
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/3144304823c99f6006f3397fe0984e62d09e4764

tips/tip-1070.md
legion2002: this would require a mapping, or iterating through an array onchain. So maybe not worth it because we have a low number of validators rn, and you can trivially do this offchain.

tips/tip-1070.md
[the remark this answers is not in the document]
legion2002: seems like the decision was made here to not fix it in the impl - https://github.com/tempoxyz/tempo/pull/6209/changes#r3494166863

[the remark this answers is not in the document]
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/6ef4f538c9414c15fcdc4fc3c7b606fdb752979b

tips/tip-1070.md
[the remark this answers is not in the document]
legion2002: Addressed in https://github.com/tempoxyz/tempo/commit/6ef4f538c9414c15fcdc4fc3c7b606fdb752979b

tips/tip-1070.md
```diff
@@ -0,0 +1,124 @@
...
+
+```text
+committee(E+1) = selected_dkg_outcome(E).players
+```
+
+The system call MUST overwrite the stored `bytes32[] publicKeys` array with `selected_dkg_outcome(E).players`, converted to `bytes32[]` without reordering. Implementations MUST NOT append a new committee record per epoch or otherwise grow state as epochs advance. A block for the last slot of epoch `E` is invalid if processing it does not perform the system call with the committee implied by that block's extra data.
+
+## Activation
+
+At activation, implementations MUST deploy the committee contract at `CURRENT_COMMITTEE_ADDRESS`. Activation MUST NOT initialize the committee state from the latest DKG outcome.
```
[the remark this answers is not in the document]
legion2002: doesn't seem like it matters, it only happens for the first block activation. I don't think any consumers will really care about this.

0xrusowsky (approved):

PR #5224, opened 2026-06-02
feat(contracts): add feature tip ABI

emmajam:
Refs TIP-1063

Adds a feature tip ABI for the protocol feature registry. The feature tip is the highest active feature ID; validator support, scheduling, activation state, and quorum reads are exposed around that cursor instead of per-feature registration.

### DIFF
--- a/crates/contracts/src/precompiles/feature_registry.rs
+++ b/crates/contracts/src/precompiles/feature_registry.rs
@@ -3,182 +3,77 @@ pub use IFeatureRegistry::{
 };
 
 crate::sol! {
-    /// Registry for Tempo protocol feature activation.
+    /// Registry for Tempo protocol feature tip scheduling.
+    ///
+    /// The feature tip is the highest active protocol feature ID. A chain with feature tip `N`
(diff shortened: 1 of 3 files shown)


crates/contracts/src/precompiles/feature_registry.rs
```diff
@@ -0,0 +1,79 @@
...
+        function scheduleFeaturesTip(uint64 featuresTip, uint64 activationEpoch) external;
+
+        /// @notice Cancels the scheduled feature tip before activation.
+        function cancelScheduledFeaturesTip() external;
+
+        /// @notice Returns the latest feature tip reported by a validator.
+        function validatorSupportedFeaturesTip(address validator) external view returns (uint64);
+
+        /// @notice Returns current active-validator support for a feature tip.
+        function featuresTipSupportCount(uint64 featuresTip) external view returns (uint256 support, uint256 required);
```
emmajam:
```suggestion
        function featuresTipSupport(uint64 featuresTip) external view returns (uint256 support, uint256 required);
```

klkvr (approved):

PR #5226, opened 2026-06-02
docs(tip-1063): clarify feature flag activation model

emmajam:
Refs TIP-1063

Updates TIP-1063 to keep protocol feature flags as the public model while defining the active feature tip as the highest active feature ID. Validators report their supported feature tip, and scheduled activation advances that edge once quorum is reached.

### DIFF
--- a/tips/tip-1063.md
+++ b/tips/tip-1063.md
@@ -16,7 +16,7 @@ This TIP defines protocol feature flags for Tempo protocol changes, decoupling c
 
 Each feature ships inactive in a Tempo release and is assigned a monotonically increasing feature ID. The chain stores a single active feature tip: the edge of the active feature-flag set and the highest active feature ID. A chain with feature tip `N` treats every feature ID in `1..=N` as active.
 
-The registry owner schedules a higher feature tip for a future activation epoch. The scheduled tip becomes active through deterministic epoch processing once the target epoch has been reached and the current active validator set has quorum for that tip.
+The registry owner schedules a higher feature tip for a future activation epoch. Feature activation occurs in the first block of that consensus epoch, and the scheduled tip becomes active if the current active validator set has quorum for it.
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


tips/tip-1063.md
```diff
@@ -146,148 +112,58 @@ If an active feature needs to be fixed, disabled, or replaced, the change ships
...
 
-    /// @notice Returns the registered feature metadata and current status.
-    function getFeature(uint64 featureId) external view returns (Feature memory);
+    /// @notice Returns the scheduled feature tip and earliest activation epoch.
+    function scheduledFeaturesTip() external view returns (uint64 featuresTip, uint64 activationEpoch);
 
-    /// @notice Returns whether a feature is registered.
-    function isFeatureRegistered(uint64 featureId) external view returns (bool);
+    /// @notice Reports the highest protocol feature ID a validator supports.
+    function setSupportedFeaturesTip(address validator, uint64 featuresTip) external;
```
klkvr: the comment seems wrong

tips/tip-1063.md
```diff
@@ -12,132 +12,98 @@ protocolVersion: TBD
...
+Scheduling writes `scheduled_features_tip` and `scheduled_activation_epoch`, then emits an event to notify network participants. The scheduled epoch is the earliest epoch in which the chain may advance to the target tip. Operators should upgrade before that epoch if they are running a release that supports a lower feature tip.
 
-Required features activating in the same epoch do not count as active dependencies. A dependent feature must activate in a later epoch than the features it requires.
+At the first deterministic activation check in each epoch greater than or equal to `scheduled_activation_epoch`, the chain evaluates whether the active validator set has quorum for `scheduled_features_tip`. A scheduled tip has quorum when at least 80% of active validators support it. The required validator count is `ceil(4 * active_validator_count / 5)`.
```
klkvr: not sure if i understand this part `At the first deterministic activation check in each epoch`

the feature is activated and the quorum would be checked at the first block of specified activation epoch

tips/tip-1063.md
```diff
@@ -12,132 +12,98 @@ protocolVersion: TBD
...
+If quorum is satisfied, the chain sets `features_tip` to `scheduled_features_tip`, clears the scheduled tip fields, and emits an activation event. The block that records the new feature tip is the activation block; protocol behavior gated by newly active features applies to subsequent blocks unless a feature's own TIP specifies transition-block behavior.
 
-If the activation checks fail, the registry clears `activation_epoch`, leaves the feature `pending`, and emits an activation failure event. The feature can be scheduled again for a later epoch after the missing quorum or required-feature condition is resolved.
+If quorum is not satisfied, the scheduled tip remains pending. The chain can retry the deterministic activation check in later epochs, and the registry owner can cancel the schedule if the activation should no longer proceed.
```
klkvr: should we clear the `scheduled_features_tip` and `scheduled_activation_epoch` in the failure case?

klkvr (commented):
i think something that's worth clarifying is how exactly the quorum is checked. i.e we do track validator -> featureTip but it's not defined how we actually fetch the list of active validators

tips/tip-1063.md
```diff
@@ -12,132 +12,106 @@ protocolVersion: TBD
...
+
+Validators report the highest feature ID they support. A validator supports target tip `N` when its latest reported supported feature tip is greater than or equal to `N`. Feature support is monotonic by design: a validator that supports feature tip `N` also supports every lower feature ID.
 
-Tempo stores protocol feature data in an onchain feature registry. Each feature is keyed by feature ID, and the registry records the minimum supported version key, required feature bitmap, status, and activation epoch for that feature. Feature activation is determined by registry state.
+Activation eligibility is bound to epoch numbers rather than wall-clock timestamps. The registry owner schedules a target feature tip and an activation epoch. At the first block of that epoch, deterministic epoch processing advances `features_tip` if the current active validator set has quorum for the scheduled target. No external activation transaction is required.
```
klkvr: not sure i like the `deterministic epoch processing` term, i think we shouldn't focus much on the epoch itself and can just say that feature activations happen in the first block of the consensus epoch, and just refer to this moment as "at the feature activation" for the rest of the TIP

klkvr (approved):

PR #5228, opened 2026-06-02
feat(tip-1060): impl `StorageCredits` precompile

0xrusowsky:
spec: https://github.com/tempoxyz/tempo/pull/4016


crates/revm/src/evm.rs
```diff
@@ -219,6 +227,125 @@ where
...
+        }
+
+        // 0→x: slot create (charged EIP-8037 state gas today).
+        let is_create = values.new_values_changes_present()
+            && values.is_original_eq_present()
+            && values.is_original_zero();
+        // x→0: slot clear (present non-zero set to zero).
+        let is_clear = values.new_values_changes_present()
+            && values.is_new_zero()
+            && !values.is_present_zero();
```
rakita:
```suggestion
        // 0→x: slot create (charged EIP-8037 state gas today).
        let is_create = values.is_original_eq_present()
            && values.is_original_zero();
        // x→0: slot clear (present non-zero set to zero).
        let is_clear = values.is_new_zero()
            && !values.is_present_zero();
```

Check in line above is doing the same thing. Maybe fmt needed

crates/revm/src/instructions.rs
```diff
@@ -29,7 +32,12 @@ fn millis_timestamp<DB: Database>(
 pub(crate) fn tempo_instructions<DB: Database>(
     spec: TempoHardfork,
 ) -> EthInstructions<EthInterpreter, TempoContext<DB>> {
-    let mut instructions = EthInstructions::new_mainnet_with_spec(spec.into());
+    let evm_spec = spec.into();
+    let mut instructions = EthInstructions::new(
+        instruction_table_with_gas_state::<TempoGasState, EthInterpreter, TempoContext<DB>>(),
```
rakita: We probably dont gate it behind any spec.

rakita (dismissed):
lgtm

0xrusowsky:
cyclops audit note="reference spec: https://github.com/tempoxyz/tempo/blob/tip/1060/tips/tip-1060.md"

decofe:
cc @0xrusowsky

Cyclops audit event published. [View workflow run](https://github.com/tempoxyz/tempo/actions/runs/26938247075)

**Config:** config: `default`, iterations: `default`, hours: `default`, note: `reference spec: https://github.com/tempoxyz/tempo/blob/tip/1060/tips/tip-1060.md`

kuyziss:
cyclops audit note="reference spec: https://github.com/tempoxyz/tempo/blob/tip/1060/tips/tip-1060.md"

decofe:
cc @kuyziss

Cyclops audit event published. [View workflow run](https://github.com/tempoxyz/tempo/actions/runs/26940154707)

**Config:** config: `default`, iterations: `default`, hours: `default`, note: `reference spec: https://github.com/tempoxyz/tempo/blob/tip/1060/tips/tip-1060.md`

crates/revm/src/handler.rs
```diff
@@ -1534,17 +1535,33 @@ where
...
+        evm: &mut Self::Evm,
+        original_reservoir: u64,
+        frame_result: &mut FrameResult,
+    ) -> Result<(), Self::Error> {
+        MainnetHandler::<_, Self::Error, _>::default().last_frame_result(
+            evm,
+            original_reservoir,
+            frame_result,
+        )?;
+        tip1060::apply_refund(evm, frame_result.gas_mut())?;
```
rakita: True, with multi calls apply_refund would be called multiple times. We should move it later in different place in code

crates/precompiles/src/storage/evm.rs
```diff
@@ -193,17 +270,30 @@ impl<'a> PrecompileStorageProvider for EvmPrecompileStorageProvider<'a> {
...
 
-        // dynamic gas
-        self.deduct_gas(
-            self.gas_params
-                .sstore_dynamic_gas(true, &result.data, result.is_cold),
-        )?;
+        // TIP-1060 (T6+): run the storage credits policy so precompile-driven storage
+        // writes honor the same accounting as the opcode-level SSTORE hook.
+        let outcome = if self.spec.is_t6() {
+            sstore_storage_credits(self, address, &result)?
```
rakita, line 258: ~So system contracts, we could clear all changes made on CREDIT_CONTRACT at the end.~

  We will leave it as is to not have different moods

crates/precompiles/src/tip1060_storage_credits/gas_state.rs
```diff
@@ -0,0 +1,177 @@
...
+        return Ok(GasStateOutcome {
+            skip_gas: true,
+            skip_refund: true,
+        });
+    }
+
+    // 0→x: slot create (charged EIP-8037 state gas today).
+    let is_create = values.is_original_eq_present() && values.is_original_zero();
+    // x→0: slot clear (present non-zero set to zero).
+    let is_clear = values.is_new_zero() && !values.is_present_zero();
```
rakita: Few lines above we have: new != present (so there is a change)

And now we have
is_create =  original == present && original == 0
is_clear = new == 0 && present != 0

To say it with words:
is_create is true if original and present are zero and new one is not zero. 
is_clear is true if we go from present that is not zero to zero. Dont care for origin value.

What would we expect if:
origin value: 0
0 -> x create  0
x -> 0 clear (credit +1)
0 -> x create (credit -1)
x -> 0 clear (creati +1)

origin value: x
x -> 0 clear (credit +1) 
0 -> y create (credit -1)
y -> 0 clear (credit +1)
0 -> x create (credit -1)


I guess TIP-1060 does not need to check origin values. Previous logic was from SSTORE and how it interacts with origin value but i guess it is wrong here.

crates/precompiles/src/lib.rs
```diff
@@ -146,6 +147,8 @@ pub fn extend_tempo_precompiles(precompiles: &mut PrecompilesMap, cfg: &CfgEnv<T
             Some(SignatureVerifier::create_precompile(&cfg))
         } else if *address == RECEIVE_POLICY_GUARD_ADDRESS && cfg.spec.is_t6() {
             Some(ReceivePolicyGuard::create_precompile(&cfg))
+        } else if *address == STORAGE_CREDITS_ADDRESS {
+            Some(TIP1060StorageCredits::create_precompile(&cfg))
```
rakita, line 177: Yeah we should probably add this, if there are some changes to precompile it would be breaking change

0xrusowsky:
cyclops audit note="reference spec: https://github.com/tempoxyz/tempo/blob/tip/1060/tips/tip-1060.md"

decofe:
cc @0xrusowsky

Cyclops audit event published. [View workflow run](https://github.com/tempoxyz/tempo/actions/runs/26957811511)

**Config:** config: `default`, iterations: `default`, hours: `default`, note: `reference spec: https://github.com/tempoxyz/tempo/blob/tip/1060/tips/tip-1060.md`

crates/precompiles/src/tip1060_storage_credits/gas_state.rs
```diff
@@ -0,0 +1,172 @@
...
+                // Only if there is a credit available, skip gas
+                if storage_credits.balance > 0 {
+                    // Consume the storage credit and charge 20k for the SSTORE + cold access cost.
+                    if caller_state_load.is_cold {
+                        backend.charge_gas(backend.gas_params().cold_storage_cost())?;
+                    }
+                    backend.charge_gas(20_000)?;
+                    storage_credits.balance -= 1;
+                    was_changed = true;
+                    outcome.skip_gas = true;
```
0xrusowsky: EIP-8037 is disabled, so this is not a concern for the time being. Whenever we decide to activate it, we will revisit its interactions wqith TIP1060

crates/precompiles/src/storage/evm.rs
```diff
@@ -193,17 +270,30 @@ impl<'a> PrecompileStorageProvider for EvmPrecompileStorageProvider<'a> {
...
         }
 
-        // dynamic gas
-        self.deduct_gas(
-            self.gas_params
-                .sstore_dynamic_gas(true, &result.data, result.is_cold),
-        )?;
+        // TIP-1060 (T7+): run the storage credits policy so precompile-driven storage
+        // writes honor the same accounting as the opcode-level SSTORE hook.
+        let outcome = if self.spec.is_t7() {
```
0xrusowsky: should be addressed by [`597be8f` (this PR)](https://github.com/tempoxyz/tempo/pull/5228/commits/597be8f08e1f851ef03098154a28a72ec0331a76)

0xrusowsky:
cyclops audit note="reference spec: https://github.com/tempoxyz/tempo/blob/tip/1060/tips/tip-1060.md"

decofe:
cc @0xrusowsky

Cyclops audit event published. [View workflow run](https://github.com/tempoxyz/tempo/actions/runs/27057471126)

**Config:** config: `default`, iterations: `default`, hours: `default`, note: `reference spec: https://github.com/tempoxyz/tempo/blob/tip/1060/tips/tip-1060.md`

crates/precompiles/src/tip1060_storage_credits/gas_state.rs
```diff
@@ -0,0 +1,172 @@
...
+        });
+    }
+
+    // Load token_state for the contract
+    let warm_storage_read_cost = backend.gas_params().warm_storage_read_cost();
+    backend.charge_gas(warm_storage_read_cost)?;
+
+    backend.load_storage_credit_account()?;
+
+    let account_slot = TIP1060StorageCredits::slot(owner);
```
0xrusowsky, line 129: this is part of the design, individual contracts/precompiles can manager credit refunds on a per-user basis by switching the storage credit `Mode`.

there are follow-up TIPs will do this attribution for individual precompiles (i.e. TIP-1064 does it for the DEX)

crates/revm/src/tip1060.rs
```diff
@@ -0,0 +1,182 @@
...
+        refunds += settled;
+
+        let new_word = state.into_word();
+        debug_assert_ne!(new_word, old_word);
+
+        journal.sstore(STORAGE_CREDITS_ADDRESS, key, new_word)?;
+    }
+
+    // TODO use gas_params for proper constant.
+    gas.erase_cost(refunds * 230_000);
```
0xrusowsky: it will be properly accounted for when enabling TIP-1016

crates/precompiles/src/storage/evm.rs
```diff
@@ -43,11 +52,22 @@ impl<'a> EvmPrecompileStorageProvider<'a> {
             amsterdam_eip8037_enabled,
             is_static,
             gas_params,
+            tip1060_storage_credits_enabled: true,
```
0xrusowsky: afaiu this is not a concern, as we are actually clearing a slot in the ring buffer.

crates/precompiles/src/tip1060_storage_credits/gas_state.rs
```diff
@@ -0,0 +1,172 @@
...
+    // TIP-1060 removes the legacy storage-clearing gas refunds.
+    let mut outcome = GasStateOutcome {
+        skip_gas: false,
+        skip_refund: true,
+    };
+
+    // Only account for storage credits when the slot crosses the zero boundary
+    // (zero -> non-zero or non-zero -> zero). If both values are zero or both are
+    // non-zero, slot occupancy is unchanged, so skip storage credits accounting.
+    if values.is_present_zero() == values.is_new_zero() {
```
0xrusowsky, line 115: we believe this is not an issue as it doesn't compromise the soundness of the credit/refund attribution, as users pay for the creation.

crates/precompiles/src/tip20/mod.rs
```diff
@@ -1192,7 +1192,7 @@ impl TIP20Token {
         mint_total_supply: Option<U256>,
         memo: B256,
     ) -> Result<bool> {
-        if !self.storage.spec().is_t6() {
+        if !self.storage.spec().is_t7() {
```
0xrusowsky: fixed in [`fe2fcf0` (this PR)](https://github.com/tempoxyz/tempo/pull/5228/commits/fe2fcf02c21aa271bbb77c9640b5893f581953bd)

crates/revm/src/handler.rs
```diff
@@ -821,12 +822,14 @@ where
...
 
-        if let Some(tempo_tx_env) = evm.ctx().tx().tempo_tx_env.as_ref() {
+        let mut frame_result = if let Some(tempo_tx_env) = evm.ctx().tx().tempo_tx_env.as_ref() {
             let calls = tempo_tx_env.aa_calls.clone();
             self.execute_multi_call(evm, gas_limit, reservoir, calls)
         } else {
             self.execute_single_call(evm, gas_limit, reservoir)
-        }
+        }?;
+        tip1060::apply_refund(evm, frame_result.gas_mut())?;
```
klkvr: could we apply refunds in `reimburse_caller` intead?

crates/node/tests/it/eth_call.rs
```diff
@@ -300,7 +300,9 @@ async fn test_eth_estimate_gas(schedule: ForkSchedule) -> eyre::Result<()> {
     let gas = provider.estimate_gas(tx.clone()).await?;
     // gas estimation is calldata dependent, but should be consistent with same calldata
     // TIP-1000 (T1): gas includes 250k new account cost when nonce=0
-    let expected_gas = if schedule.is_active(TempoHardfork::T6) {
+    let expected_gas = if schedule.is_active(TempoHardfork::T7) {
+        547407
+    } else if schedule.is_active(TempoHardfork::T6) {
```
klkvr: seems like the extra branch is not necessary?

crates/revm/src/tip1060.rs
```diff
@@ -0,0 +1,182 @@
...
+where
+    DB: Database,
+{
+    fn sstore_gas_state(
+        context: &mut InstructionContext<'_, TempoContext<DB>, EthInterpreter>,
+        owner: Address,
+        values: &StateLoad<SStoreResult>,
+    ) -> Result<GasStateOutcome, InstructionResult> {
+        sstore_storage_credits(&mut InterpreterStorageCredits { context }, owner, values)
+    }
```
klkvr: do we actually need that? we have entire `TempoContext` here which is enough to construct `EvmPrecompileStorageProvider` i think? can we get rid of the `StorageCreditsBackend` trait?

crates/revm/src/tip1060.rs
```diff
@@ -0,0 +1,182 @@
...
+            pending.saturating_add(U256::from(1)),
+        );
+    }
+}
+
+/// Tempo SSTORE gas-state policy.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Hash)]
+pub(crate) struct TIP1060StorageCreditsState;
+
+impl<DB> GasStateTr<EthInterpreter, TempoContext<DB>> for TIP1060StorageCreditsState
```
klkvr: `GasStateTr` abstraction also feels odd. i'd rather have `tempo_instructions` do the SSTORE instruction override on its own and we could add something like `sstore_with_gas_accounting(ctx, |ctx, target, state_load| ...)` to revm

crates/precompiles/src/storage/evm.rs
```diff
@@ -193,17 +281,39 @@ impl<'a> PrecompileStorageProvider for EvmPrecompileStorageProvider<'a> {
...
 
-        // dynamic gas
-        self.deduct_gas(
-            self.gas_params
-                .sstore_dynamic_gas(true, &result.data, result.is_cold),
-        )?;
+        // TIP-1060 (T7+): run the storage credits policy so precompile-driven storage
+        // writes honor the same accounting as the opcode-level SSTORE hook.
+        let outcome = if self.spec.is_t7() {
+            if self.tip1060_storage_credits_enabled {
```
klkvr: is it not enough to just check `tip1060_storage_credits_enabled`? can it be `true` pre-t7?

crates/precompiles/src/tip1060_storage_credits/gas_state.rs
```diff
@@ -0,0 +1,172 @@
...
+        return Ok(outcome);
+    }
+
+    // Storage-credit precompile state is used for protocol bookkeeping. Because of that,
+    // always skips TIP-1000 + TIP-1060 self-accounting and charge only update gas.
+    if owner == STORAGE_CREDITS_ADDRESS {
+        if is_cold {
+            backend.charge_gas(backend.gas_params().cold_storage_cost())?;
+        }
+        if values.new_values_changes_present() && values.is_original_eq_present() {
```
klkvr:
```suggestion
        if values.is_original_eq_present() {
```

we've just checked that we are crossing zero boundary so new value can't be same as present

crates/precompiles/src/storage/evm.rs
```diff
@@ -20,9 +25,13 @@ pub struct EvmPrecompileStorageProvider<'a> {
     amsterdam_eip8037_enabled: bool,
     is_static: bool,
     gas_params: GasParams,
+    tip1060_storage_credits_enabled: bool,
     /// Debug-only LIFO checkpoint validator. See [`Self::assert_lifo`].
     #[cfg(debug_assertions)]
     checkpoint_stack: Vec<(usize, usize)>,
+    // TIP-1060
+    #[allow(dead_code)]
+    pending_refund_eligible_creations: AddressMap<u64>,
```
klkvr: this field is unused?

crates/precompiles/src/tip1060_storage_credits/gas_state.rs
```diff
@@ -0,0 +1,172 @@
...
+        // 0→x: slot create (charged EIP-8037 state gas today).
+        match storage_credits.mode {
+            CreditMode::Direct => {
+                // Only if there is a credit available, skip gas
+                if storage_credits.balance > 0 {
+                    // Consume the storage credit and charge 20k for the SSTORE + cold access cost.
+                    if caller_state_load.is_cold {
+                        backend.charge_gas(backend.gas_params().cold_storage_cost())?;
+                    }
+                    backend.charge_gas(20_000)?;
```
klkvr: this should use a const

crates/precompiles/src/tip1060_storage_credits/gas_state.rs
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    if was_changed {
+        // cold load is already checked above when we loaded the storage credits account.
+        let result = backend
+            .store_credits(account_slot, storage_credits.into_word())?
+            .data;
+
+        // Only when change happens charge additional gas.
+        // Creation of credit slot is compensated by creation of the contract creation.
```
klkvr:
```suggestion
        // Creation of credit slot is compensated by account creation cost.
```

crates/revm/src/handler.rs
0xrusowsky: i think `reimburse_caller` is too late because `ResultGas` is already built by then. if you dislike current placement, i think we should run it in `post_execution`, before `ResultGas`, floor checks, and reimbursement. wdyt @klkvr?

crates/precompiles/src/tip1060_storage_credits/gas_state.rs
0xrusowsky: addressed in [`15a8091` (this PR)](https://github.com/tempoxyz/tempo/pull/5228/commits/15a8091b7ea690bae75418e4e2904ece4c6d7cd4)

crates/precompiles/src/storage/evm.rs
0xrusowsky: addressed in [`15a8091` (this PR)](https://github.com/tempoxyz/tempo/pull/5228/commits/15a8091b7ea690bae75418e4e2904ece4c6d7cd4)

crates/precompiles/src/tip1060_storage_credits/gas_state.rs
0xrusowsky: addressed in [`15a8091` (this PR)](https://github.com/tempoxyz/tempo/pull/5228/commits/15a8091b7ea690bae75418e4e2904ece4c6d7cd4)

crates/precompiles/src/storage/evm.rs
0xrusowsky: addressed in [`15a8091` (this PR)](https://github.com/tempoxyz/tempo/pull/5228/commits/15a8091b7ea690bae75418e4e2904ece4c6d7cd4)

crates/revm/src/tip1060.rs
0xrusowsky: i kept the trait, but tweaked it to take `GasTracker` directly and impl it on `TempoContext` [`15a8091` (this PR)](https://github.com/tempoxyz/tempo/pull/5228/commits/15a8091b7ea690bae75418e4e2904ece4c6d7cd4). lmk if u still want changes

crates/revm/src/handler.rs
0xrusowsky: moved to `post_execution` in [`1d21271` (this PR)](https://github.com/tempoxyz/tempo/pull/5228/commits/1d212712f2fa4486bfd732b8da7afcb8fc3dbe00)

crates/precompiles/src/tip1060_storage_credits/gas_state.rs
```diff
@@ -0,0 +1,180 @@
...
+    let mut storage_credits =
+        AccountState::from_word(storage_credit_state_load.data).map_err(|_| B::fatal_external())?;
+
+    let mut was_changed = false;
+    if values.is_new_zero() {
+        // x→0: slot clear (present non-zero set to zero).
+        storage_credits.balance = storage_credits.balance.saturating_add(1);
+        was_changed = true;
+    } else {
+        // 0→x: slot create (charged EIP-8037 state gas today).
```
0xalpharush: remove mention of EIP-8037?

0xrusowsky:
cyclops audit note="reference spec: https://github.com/tempoxyz/tempo/blob/tip/1060/tips/tip-1060.md"

decofe:
cc @0xrusowsky

Cyclops audit event published. [View workflow run](https://github.com/tempoxyz/tempo/actions/runs/27340104628)

**Config:** config: `default`, iterations: `default`, hours: `default`, note: `reference spec: https://github.com/tempoxyz/tempo/blob/tip/1060/tips/tip-1060.md`

legion2002 (commented):
:eye: **Cyclops Review**

Manual repost from the failed Cyclops run on commit `f3585b7` because worker `pr-5228-w3` failed before consolidation. Source status comment: https://github.com/tempoxyz/tempo/pull/5228#issuecomment-4679541516

The verified findings are posted inline below. The duplicate/generalized system-precompile accounting finding is covered by the `enter_fee_collection` comment; the invariant-review item was a summary report rather than a separate inline finding.

crates/revm/src/handler.rs
```diff
@@ -828,6 +829,31 @@ where
...
+    /// Applies Tempo-specific post-execution accounting before the standard gas refund flow.
+    #[inline]
+    fn post_execution(
+        &self,
+        evm: &mut Self::Evm,
+        exec_result: &mut FrameResult,
+        init_and_floor_gas: InitialAndFloorGas,
+        eip7702_gas_refund: i64,
+    ) -> Result<ResultGas, Self::Error> {
+        tip1060::apply_refund(evm, exec_result.gas_mut())?;
```
0xrusowsky: i think its fine to leave as is. better to be protective.

crates/precompiles/src/storage/thread_local.rs
```diff
@@ -347,6 +347,27 @@ impl<'evm> StorageCtx {
         Self::enter(&mut provider, f)
     }
 
+    /// Enters storage for protocol-internal fee collection.
+    ///
+    /// Fee collection is exempt from TIP-1060 storage-credit accounting: it must not mint,
+    /// consume, or settle storage credits while updating fee/TIP20/keychain bookkeeping.
+    pub fn enter_fee_collection<J, R>(
```
0xrusowsky: this is intended behavior, so imo not an issue

crates/precompiles/src/tip1060_storage_credits/gas_state.rs
```diff
@@ -0,0 +1,184 @@
...
+    let storage_credit_state_load = backend.load_credits(account_slot, skip_cold)?;
+    if storage_credit_state_load.is_cold {
+        backend.charge_gas(additional_cold_cost)?;
+    }
+
+    let mut balance =
+        u64::from_word(storage_credit_state_load.data).map_err(|_| B::fatal_external())?;
+
+    let mut was_changed = false;
+    if values.is_new_zero() {
```
0xrusowsky, line 142: addressed in [`781e8d0` (this PR)](https://github.com/tempoxyz/tempo/pull/5228/commits/781e8d018fdd960d251a69efbfe1c4d8bc9cf3f2)

crates/precompiles/src/lib.rs
```diff
@@ -171,6 +173,8 @@ pub fn extend_tempo_precompiles(precompiles: &mut PrecompilesMap, cfg: &CfgEnv<T
             Some(SignatureVerifier::create_precompile(&cfg))
         } else if *address == RECEIVE_POLICY_GUARD_ADDRESS && cfg.spec.is_t6() {
             Some(ReceivePolicyGuard::create_precompile(&cfg))
+        } else if *address == STORAGE_CREDITS_ADDRESS && cfg.spec.is_t7() {
+            Some(TIP1060StorageCredits::create_precompile(&cfg))
```
0xrusowsky, line 177: dumb concern

crates/precompiles/src/tip1060_storage_credits/dispatch.rs
```diff
@@ -0,0 +1,69 @@
...
+            return err;
+        }
+
+        dispatch_call(
+            calldata,
+            &[],
+            ITIP1060StorageCreditsCalls::abi_decode,
+            |call| match call {
+                ITIP1060StorageCreditsCalls::balance(call) => {
+                    view(call, |_| self.balance_of(msg_sender))
```
legion2002: do we need 2 balance/mode functions? could just keep `balanceOf` and `modeOf` simillar to ERC20

crates/precompiles/src/tip1060_storage_credits/gas_state.rs
```diff
@@ -0,0 +1,184 @@
...
+
+    /// SSTORE a slot of the storage credits contract (cold load is never skipped here).
+    fn store_credits(
+        &mut self,
+        key: U256,
+        value: U256,
+    ) -> Result<StateLoad<SStoreResult>, Self::Error>;
+
+    /// Loads the transaction-local storage credit state word at `key`.
+    fn load_transient_state(&mut self, key: U256) -> U256;
```
legion2002: nit: `load_transient_state` can be a bit confusing in the context of storage. could be renamed to `load/store_transient_credit_state`

crates/precompiles/src/tip1060_storage_credits/gas_state.rs
```diff
@@ -0,0 +1,184 @@
...
+use crate::storage::FromWord;
+use alloy::primitives::{Address, U256};
+use revm::{
+    context_interface::cfg::GasParams,
+    interpreter::{SStoreResult, StateLoad, gas::GasTracker, instruction_context::GasStateOutcome},
+};
+use tempo_contracts::precompiles::STORAGE_CREDITS_ADDRESS;
+
+pub const STORAGE_CREDIT_VALUE: u64 = 230_000;
+const SSTORE_SET_WITHOUT_EXPANSION_COST: u64 = 20_000;
```
legion2002: we can just use this from revm, instead of hardcoding

Cargo.toml
```diff
@@ -144,66 +144,66 @@ tempo-telemetry-util = { path = "crates/telemetry-util", default-features = fals
...
-reth-stages = { git = "https://github.com/paradigmxyz/reth", rev = "72fafb5" }
-reth-static-file = { git = "https://github.com/paradigmxyz/reth", rev = "72fafb5" }
-reth-storage-api = { git = "https://github.com/paradigmxyz/reth", rev = "72fafb5" }
-reth-tasks = { git = "https://github.com/paradigmxyz/reth", rev = "72fafb5" }
-reth-tracing = { git = "https://github.com/paradigmxyz/reth", rev = "72fafb5" }
-reth-transaction-pool = { git = "https://github.com/paradigmxyz/reth", rev = "72fafb5" }
-reth-trie = { git = "https://github.com/paradigmxyz/reth", rev = "72fafb5" }
-reth-trie-common = { git = "https://github.com/paradigmxyz/reth", rev = "72fafb5" }
-reth-trie-db = { git = "https://github.com/paradigmxyz/reth", rev = "72fafb5" }
+reth-config = { git = "https://github.com/paradigmxyz/reth", branch = "ci/tip-1060" }
```
legion2002: we should pin this to a specific commit again, before merging the PR

crates/precompiles/src/tip1060_storage_credits/mod.rs
```diff
@@ -0,0 +1,148 @@
...
+
+impl TransientState {
+    /// Decodes a packed transient state word.
+    ///
+    /// Layout:
+    /// - bits `0..=63`: pending refund-eligible creations (`uint64`)
+    /// - bits `64..=71`: storage creation mode
+    /// - bits `72..=255`: reserved for future hardfork-gated extensions
+    #[inline]
+    pub fn from_word(value: U256) -> Result<Self> {
```
legion2002: we don't have precompile macros for transient storage packing?

legion2002: or is this also some non standard packing, like the gas state stuff described below?

crates/precompiles/src/tip1060_storage_credits/mod.rs
```diff
@@ -0,0 +1,148 @@
...
+
+    #[inline]
+    pub fn into_word(self) -> U256 {
+        U256::from_limbs([self.pending_refunds, self.mode as u64, 0, 0])
+    }
+}
+
+/// TIP-1060 storage credits precompile, which tracks per-account storage credit state.
+///
+/// Unlike the Solidity-compatible `Mapping<Address, GasState>` layout, persistent account state is
```
legion2002, line 62: this makes sense, but we should highlight in the docs/TIPs. Could be a footgun going forward, for eg. if we ever do a script to transition account mappings from keccak to blake3

crates/precompiles/src/tip1060_storage_credits/mod.rs
0xrusowsky: we can't use `Storable` cause `StorageCtx` is only available during precompile execution, but TIP1060's SSTORE hook also runs on regular evm calls.

added a note: [`30eaab9` (this PR)](https://github.com/tempoxyz/tempo/pull/5228/commits/30eaab96ba8eda11bd55eb73f4061c25b9d8f541)

crates/precompiles/src/tip1060_storage_credits/gas_state.rs
0xrusowsky: afaik these values are arbitrary and only related to TIP-1060.
we could define them in `chainspec` though

Cargo.toml
0xrusowsky: this will be fixed before merging. it is just to fix compilation issues cause reth < revm main

crates/precompiles/src/tip1060_storage_credits/gas_state.rs
0xrusowsky: moved there [`a5e2478` (this PR)](https://github.com/tempoxyz/tempo/pull/5228/commits/a5e24787e4d90301e8ffb609024ea9659df519be)

crates/precompiles/src/tip1060_storage_credits/gas_state.rs
0xrusowsky: addressed here [`c716eb3` (this PR)](https://github.com/tempoxyz/tempo/pull/5228/commits/c716eb32418b24537cb96944dd3b45d32d4a3e54)

crates/precompiles/src/tip1060_storage_credits/dispatch.rs
0xrusowsky: addressed in [`099203c` (this PR)](https://github.com/tempoxyz/tempo/pull/5228/commits/099203c30bb4e93230aeca6542502f9733897ac3)

crates/revm/src/evm.rs
```diff
@@ -528,15 +551,24 @@ mod tests {
...
             tx: TempoTransaction,
+        ) -> eyre::Result<tempo_primitives::AASigned> {
+            self.sign_tx_keychain_for_user(tx, self.address)
+        }
+
+        /// Sign a keychain transaction as this access key for the given root user.
+        fn sign_tx_keychain_for_user(
+            &self,
+            tx: TempoTransaction,
+            user: Address,
```
0xKitsune: nit: are these unrelated changes? If so can we remove from this PR?

crates/revm/src/evm.rs
```diff
@@ -1885,6 +1917,1094 @@ mod tests {
...
+    /// Only the create (0->x) leg is mode-sensitive, so the gas used differs per mode:
+    /// - `Preserve` charges the full EVM/state gas as normal,
+    /// - `Direct` has no storage credit to spend at create time (balance starts at 0), so it also
+    ///   charges the full create cost (identical to `Preserve`),
+    /// - `Refund` charges the full create gas but accrues a deferred storage credit that
+    ///   `apply_refund` settles at end-of-tx, erasing 230_000 gas, so it ends up the cheapest.
+    ///
+    /// The storage credit balance corroborates the mechanism: the clear (x->0) leg mints one
+    /// storage credit in every mode, but `Refund` consumes that minted storage credit against its
+    /// deferred create storage credit at end-of-tx, so it lands at 0 while the others stay at 1.
```
0xKitsune: very minor nit: can we make these less verbose?

crates/precompiles/src/tip1060_storage_credits/mod.rs
```diff
@@ -0,0 +1,177 @@
...
+    type Error = TempoPrecompileError;
+
+    fn try_from(mode: Mode) -> Result<Self> {
+        match mode {
+            Mode::Refund => Ok(Self::Refund),
+            Mode::Preserve => Ok(Self::Preserve),
+            Mode::Direct => Ok(Self::Direct),
+            _ => Err(TIP1060StorageCreditsError::invalid_mode().into()),
+        }
+    }
```
0xKitsune: nit: can we move all of these helper types after the main contract impl block?

crates/precompiles/src/tip1060_storage_credits/gas_state.rs
0xKitsune: Not sure this addresses the finding, looks like it fixes the opposite direction.

The commit added the `x->0->x` restore branch, which burns a credit when a slot that was nonzero at tx-start is cleared and then reset.

IIUC this doesnt cover the case of `0->x->0` where slot created within the same tx then cleared.

```rust
if values.is_new_zero() {
    // x->0: slot clear
    balance = balance.saturating_add(1);   // mints regardless of `original`
    was_changed = true;
}
```

Theres no `is_original_zero()` guard here, so clearing a slot that didn't exist at tx start still mints a credit even though no persistent storage was actually deleted.

Can we add a test for the `0->x->0` case?

0xrusowsky: `0->x->0` minting is intentional under the current TIP-1060 economics. The invariant is not "credits only come from deleting slots that existed at tx start" but instead "you cannot obtain storage credits without paying their storage-creation value"

for `0->x->0`, the `0->x` leg is a genuine fresh-slot creation and pays the full 250k cost. The later clear can mint/return one credit because that credit is economically backed by the creation cost already paid.

the finding is different though: `x->0->x` on a slot that was nonzero at tx start can mint through cheap dirty writes. The clear mints a credit, and the restore is only a dirty restore rather than a paid fresh creation. That is what the restore clawback fixes.

NOTE: we already cover the intentional `0->x->0` behavior in `test_tip1060_sstore_create_then_clear_modes`

crates/precompiles/src/storage/evm.rs
```diff
@@ -193,17 +261,30 @@ impl<'a> PrecompileStorageProvider for EvmPrecompileStorageProvider<'a> {
...
-        self.deduct_gas(
-            self.gas_params
-                .sstore_dynamic_gas(true, &result.data, result.is_cold),
-        )?;
+        // TIP-1060 (T7+): run the storage credits policy so precompile-driven storage
+        // writes honor the same accounting as the opcode-level SSTORE hook.
+        let outcome = if self.tip1060_storage_credits_enabled {
+            sstore_storage_credits(self, address, &result)?
+        } else {
+            Default::default()
```
0xalpharush: Please explicitly initialize struct

crates/revm/src/handler.rs
rakita: I would even say it is better, floor gas is there to protect against using tx.input for data availability, idea is you need to spend at least N amount of gas to cover input size

crates/revm/src/tip1060.rs
rakita: Changed in revm, will be integrated with next revm bump. Not a blocker, it is a code refactor change

0xrusowsky:
cyclops audit

decofe:
cc @0xrusowsky

Cyclops audit event published. [View workflow run](https://github.com/tempoxyz/tempo/actions/runs/27605857799)

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [workspace.package]
-version = "1.8.2"
+version = "1.8.1"
```
rakita: nit and reth/revm links needs to be resolved. And PR probably needs to be rebased on main.

rakita (dismissed):
lgtm, pending nits in Cargo.toml

crates/precompiles/src/tip_fee_manager/mod.rs
```diff
@@ -291,6 +291,10 @@ impl TipFeeManager {
     /// # Errors
     /// - `InvalidToken` — `token` does not have a valid TIP-20 prefix
     pub fn distribute_fees(&mut self, validator: Address, token: Address) -> Result<()> {
+        // +T7: Fee collection bypasses TIP-1060, so disable storage-credit accounting to avoid
+        // minting user-refundable storage credits when clearing fee-manager slots.
+        self.storage.set_tip1060_storage_credits(false);
+
```
klkvr: let's remove this

crates/primitives/src/reth_compat/transaction/envelope.rs
```diff
@@ -190,7 +190,9 @@ mod codec {
     }
 
     impl reth_db_api::table::Decompress for TempoTxEnvelope {
-        fn decompress(value: &[u8]) -> Result<Self, DecompressError> {
+        fn decompress(
+            value: &[u8],
+        ) -> core::result::Result<Self, reth_codecs::compress::DecompressError> {
```
klkvr: seems unnecessary to change this?

crates/primitives/src/reth_compat/header.rs
```diff
@@ -139,7 +139,9 @@ mod codec {
     }
 
     impl reth_db_api::table::Decompress for TempoHeader {
-        fn decompress(value: &[u8]) -> Result<Self, reth_codecs::DecompressError> {
+        fn decompress(
+            value: &[u8],
+        ) -> core::result::Result<Self, reth_codecs::compress::DecompressError> {
```
klkvr: same

crates/revm/src/gas_params.rs
```diff
@@ -55,6 +70,35 @@ pub fn tempo_gas_params_with_amsterdam(
...
+    // entries overridden below; everything else (tx_create_cost, create, new_account_cost,
+    // code_deposit_cost, eip7702 costs, auth refund) is exactly as in `t1_gas_params`.
+    let mut gas_params = t1_gas_params();
+    gas_params.override_gas([
+        // SSTORE (zero -> non-zero): only the 5k residual; the 245k creditable portion is governed
+        // by the TIP-1060 storage-credit hook (T1 charged the full `SSTORE_CREATE_COST` here).
+        (GasId::sstore_set_without_load_cost(), SSTORE_SET_COST),
+        // Restore (non-zero -> zero) refund must not exceed the T7 residual. Important with
+        // TIP-1060 because the refund cap is removed. Otherwise, 0→x→0 could be refund-positive.
+        (GasId::sstore_set_refund(), SSTORE_SET_COST),
```
klkvr, line 93: does this mean that for a 0 -> X -> 0 case a user would get full refund? makes me wonder if it could be problematic for transactions that do a lot of such storage writes

i.e it does seem fine to provide generous refunds for the "state gas" portion (245k), but making 0->X->0 entirely free feels more risky

crates/revm/src/tip1060.rs
```diff
@@ -0,0 +1,166 @@
...
+        let InstructionContext { interpreter, host } = context;
+        sstore_storage_credits(
+            &mut StorageCreditsContext {
+                context: host,
+                gas_tracker: interpreter.gas.tracker_mut(),
+            },
+            owner,
+            values,
+        )
+    })
```
klkvr, line 173: should this delegate to `sstore_default_gas_accounting` in some cases? right now this never goes through the regular gas codepaths afaict

crates/primitives/src/reth_compat/transaction/envelope.rs
0xrusowsky: maybe skill issue, but i was getting clippy warnings

crates/revm/src/tip1060.rs
0xrusowsky: i forgot to call it... when migrating to the new revm api.

since hook only handles state gas, we should still always call `sstore_default_gas_accounting`

0xrusowsky: addressed in [`1c24398` (this PR)](https://github.com/tempoxyz/tempo/pull/5228/commits/1c24398ddbd8c411c8c0933492a62e5f2b479fe8)

crates/revm/src/gas_params.rs
0xrusowsky: NOTE: all work happens in memory cause the journaled state is unchanged at the end of the tx + they are effectively paying the `WARM_SSLOAD` cost, so its never free.

crates/precompiles/src/tip_fee_manager/mod.rs
0xrusowsky: reactivated here [`3a88089` (this PR)](https://github.com/tempoxyz/tempo/pull/5228/commits/3a8808944db356c31918d3348aa04f03e0d8158f)

crates/precompiles/src/tip_fee_manager/mod.rs
```diff
@@ -291,6 +291,10 @@ impl TipFeeManager {
     /// # Errors
     /// - `InvalidToken` — `token` does not have a valid TIP-20 prefix
     pub fn distribute_fees(&mut self, validator: Address, token: Address) -> Result<()> {
+        // +T7: Fee collection bypasses TIP-1060, so disable storage-credit accounting to avoid
+        // minting user-refundable storage credits when clearing fee-manager slots.
+        self.storage.set_tip1060_storage_credits(false);
```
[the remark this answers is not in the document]
0xrusowsky: we decide to ignore cause validators are trusted

crates/node/tests/it/eth_call.rs
```diff
@@ -294,7 +294,7 @@ async fn test_eth_estimate_gas(schedule: ForkSchedule) -> eyre::Result<()> {
     // gas estimation is calldata dependent, but should be consistent with same calldata
     // TIP-1000 (T1): gas includes 250k new account cost when nonce=0
     let expected_gas = if schedule.is_active(TempoHardfork::T6) {
-        545190
+        547407
```
0xalpharush, line 297: shouldn't this not affect t6?

0xrusowsky: i still need to update all snapshots for T7

crates/revm/src/handler.rs
```diff
@@ -828,6 +835,51 @@ where
...
```
0xalpharush:
```suggestion
        if exec_result.instruction_result().is_ok() {      
            tip1060::apply_refund(evm, exec_result.gas_mut())?;
        }
```
reverting tx's should not have any `pending_refunds` so we can skip?

0xrusowsky: addressed in [`2974b3a` (this PR)](https://github.com/tempoxyz/tempo/pull/5228/commits/2974b3a8d8bcff83486d91ef981beb72f15a11f7)

crates/node/tests/it/eth_call.rs
klkvr: can we keep older value (545190) for T6?

crates/precompiles/src/tip_fee_manager/dispatch.rs
```diff
@@ -66,7 +66,10 @@ impl Precompile for TipFeeManager {
                 }
                 TipFeeManagerCall::FeeManager(IFeeManagerCalls::distributeFees(call)) => {
                     mutate_void(call, msg_sender, |_, c| {
-                        self.distribute_fees(c.validator, c.token)
+                        StorageCtx.set_tip1060_storage_credits(false);
+                        let result = self.distribute_fees(c.validator, c.token);
+                        StorageCtx.set_tip1060_storage_credits(true);
+                        result
```
klkvr: can we remove the toggle from here as well? i don't think this is necessary

crates/primitives/src/reth_compat/transaction/envelope.rs
```diff
@@ -190,7 +190,7 @@ mod codec {
     }
 
     impl reth_db_api::table::Decompress for TempoTxEnvelope {
-        fn decompress(value: &[u8]) -> Result<Self, DecompressError> {
+        fn decompress(value: &[u8]) -> core::result::Result<Self, DecompressError> {
```
klkvr:
```suggestion
        fn decompress(value: &[u8]) -> Result<Self, DecompressError> {
```

crates/primitives/src/reth_compat/header.rs
```diff
@@ -139,7 +139,7 @@ mod codec {
     }
 
     impl reth_db_api::table::Decompress for TempoHeader {
-        fn decompress(value: &[u8]) -> Result<Self, reth_codecs::DecompressError> {
+        fn decompress(value: &[u8]) -> core::result::Result<Self, reth_codecs::DecompressError> {
```
klkvr:
```suggestion
        fn decompress(value: &[u8]) -> Result<Self, reth_codecs::DecompressError> {
```

Cargo.toml
```diff
@@ -400,6 +399,20 @@ vergen-git2 = "9.1.0"
...
+revm-context-interface = { git = "https://github.com/bluealloy/revm", branch = "main" }
+revm-database = { git = "https://github.com/bluealloy/revm", branch = "main" }
+revm-database-interface = { git = "https://github.com/bluealloy/revm", branch = "main" }
+revm-handler = { git = "https://github.com/bluealloy/revm", branch = "main" }
+revm-inspector = { git = "https://github.com/bluealloy/revm", branch = "main" }
+revm-interpreter = { git = "https://github.com/bluealloy/revm", branch = "main" }
+revm-precompile = { git = "https://github.com/bluealloy/revm", branch = "main" }
+revm-primitives = { git = "https://github.com/bluealloy/revm", branch = "main" }
+revm-state = { git = "https://github.com/bluealloy/revm", branch = "main" }
+
```
klkvr: are those patches still necessary?

crates/revm/src/tip1060.rs
```diff
@@ -0,0 +1,174 @@
...
+        refunds += settled;
+
+        let new_word = U256::from(balance);
+        debug_assert_ne!(new_word, old_word);
+
+        journal.sstore(STORAGE_CREDITS_ADDRESS, key, new_word)?;
+    }
+
+    // Refund storage credit value per settled credit.
+    gas.erase_cost(refunds.saturating_mul(STORAGE_CREDIT_VALUE));
```
klkvr: actually i'm wondering why do we use `erase_cost` here instead of `record_refund`?

crates/precompiles/src/tip1060_storage_credits/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+            u64::MAX
+        } else {
+            0
+        };
+
+        self.write_mode_with_budget(msg_sender, mode, budget)?;
+        self.emit_event(TIP1060StorageCreditsEvent::mode_updated(
+            msg_sender,
+            mode.into(),
+        ))
```
klkvr, line 105: not sure if it makes perfect sense to me to emit this event, feels somewhat noisy. but also not strongly opposed

crates/precompiles/src/tip1060_storage_credits/gas_state.rs
```diff
@@ -0,0 +1,214 @@
...
+                    store_credit_state(backend, account_slot, transient_state)?;
+                    emit_mode_updated(backend, owner, CreditMode::Preserve)?;
+                }
+                backend.charge_gas(STORAGE_CREDIT_VALUE)?;
+            }
+            CreditMode::Preserve => {
+                // Always charge the 245k creditable portion as gas without consuming credits.
+                backend.charge_gas(STORAGE_CREDIT_VALUE)?;
+            }
+            CreditMode::Refund => {
```
klkvr, line 186: should/could we respect `budget` in `Refund` mode as well?

crates/revm/src/tip1060.rs
rakita: Here it is: https://github.com/tempoxyz/tempo/pull/5951 `record_refund` is more accurate here

crates/precompiles/src/tip1060_storage_credits/mod.rs
0xrusowsky: aims to satify the spec's request of auditability during gas tracking, so that users can see mode changes in the traces

crates/precompiles/src/tip1060_storage_credits/gas_state.rs
0xrusowsky: we can try to make it applicable for both `Direct` and `Refund`, although maybe not worth the effort unless we have a good usecase for it?

rakita: We could but this would be at the end of the execution (in tip1060::apply_refund). When you iterate over transient storage to check amount in budget.

But i dont know if we should do it, that is a harder question :D

crates/precompiles/src/tip_fee_manager/dispatch.rs
0xrusowsky: addressed in [`bd8488f` (this PR)](https://github.com/tempoxyz/tempo/pull/5228/commits/bd8488f0423a27609025af9723bf1b12ab25fd58)

Cargo.toml
0xrusowsky: addressed in [`45881c5` (this PR)](https://github.com/tempoxyz/tempo/pull/5228/commits/45881c55c8eff9c62094c7daeb61fa3f5731901b)

Cargo.toml
```diff
@@ -400,6 +399,7 @@ vergen-git2 = "9.1.0"
 # reth-trie-sparse = { path = "../reth/crates/trie/sparse" }
 
 [patch.crates-io]
+
```
klkvr:
```suggestion
```

klkvr (dismissed):

klkvr (approved):

PR #5229, opened 2026-06-02
feat(precompiles): implement feature registry tip read

emmajam:
Refs TIP-1063

Adds the feature registry precompile implementation for the active feature tip cursor. Only featuresTip() is live in this first implementation; the remaining registry ABI selectors continue to return unknown until their storage and quorum semantics are added.


crates/precompiles/src/lib.rs
```diff
@@ -142,6 +144,8 @@ pub fn extend_tempo_precompiles(precompiles: &mut PrecompilesMap, cfg: &CfgEnv<T
             Some(AccountKeychain::create_precompile(&cfg))
         } else if *address == VALIDATOR_CONFIG_V2_ADDRESS {
             Some(ValidatorConfigV2::create_precompile(&cfg))
+        } else if *address == FEATURE_REGISTRY_ADDRESS {
+            Some(FeatureRegistry::create_precompile(&cfg))
```
klkvr: we need to hardfork gate this

klkvr (approved):

PR #5231, opened 2026-06-02
perf(transaction-pool): cache fee payer recovery

i-m-aditya:
Caches `TempoPooledTransaction::fee_payer()` via `OnceLock<Option<Address>>`, matching the shape of the existing memoized fields on the same struct.
`fee_payer()` does ECDSA recovery + keccak for AA txs with a fee_payer_signature. It's called per-tx inside `evict_invalidated_transactions` (via is_sender_paid_fee) and once more during fee_balance_slot init — uncached, so the same transaction pays the cost on every invalidation event.
 In the captured profile it accounted for ~0.4% of the tempo-txpool-ma track; scales linearly with the fraction of pool txs carrying a fee_payer_signature.
Screenshot 2026-06-02 at 8 33 02 PM

### DIFF
--- a/crates/transaction-pool/src/transaction.rs
+++ b/crates/transaction-pool/src/transaction.rs
@@ -74,6 +74,8 @@ pub struct TempoPooledTransaction {
     /// Stores `(fee_token, balance_slot)` so the payload builder's state-aware iterator
     /// can check if the fee payer's balance was modified without recomputing the keccak.
     fee_balance_slot: OnceLock<Option<(Address, U256)>>,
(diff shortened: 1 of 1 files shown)


i-m-aditya:
Note: `RecoveryError` isn't `Clone`, so cache hits of `None` reconstruct a generic
  `RecoveryError::new()`. No current caller reads the error source; happy to wrap the cached `Err`
  in `Arc` and expose `Result<Address, &RecoveryError>` if preferred.

PR #5241, opened 2026-06-02
fix(localnet): pass consensus signing key secret

mediocregopher:
Passes the generated localnet signing-key secret into consensus node startup so encrypted validator signing keys can be read. This keeps the local consensus bench path aligned with the localnet generator output.

### DIFF
--- a/tempo.nu
+++ b/tempo.nu
@@ -33,11 +33,16 @@ def log-filter-args [loud: bool] {
     if $loud { [] } else { ["--log.stdout.filter" "info"] }
 }
 
-def ensure-localnet-consensus-secret [] {
-    let secret_path = $"($LOCALNET_DIR)/consensus-secret.txt"
-    $"($LOCALNET_SIGNING_KEY_SECRET)\n" | save -f $secret_path
+def prepare-localnet-consensus-secret-fifo [node_dir: string] {
+    let secret_path = $"($node_dir)/consensus-secret.fifo"
+    rm -f $secret_path
+    mkfifo $secret_path
     chmod 600 $secret_path
     $secret_path
+}
+
+def start-localnet-consensus-secret-writer [secret_path: string] {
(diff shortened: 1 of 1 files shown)


tempo.nu
```diff
@@ -32,6 +33,13 @@ def log-filter-args [loud: bool] {
     if $loud { [] } else { ["--log.stdout.filter" "info"] }
 }
 
+def ensure-localnet-consensus-secret [] {
+    let secret_path = $"($LOCALNET_DIR)/consensus-secret.txt"
+    $"($LOCALNET_SIGNING_KEY_SECRET)\n" | save -f $secret_path
+    chmod 600 $secret_path
+    $secret_path
+}
```
SuperFluffy, line 42: the recommended way is to do this via a fifo, even though files are provided. Any chance to change that?

mediocregopher: done 👍

tempo.nu
```diff
@@ -32,6 +33,18 @@ def log-filter-args [loud: bool] {
...
+def prepare-localnet-consensus-secret-fifo [node_dir: string] {
+    let secret_path = $"($node_dir)/consensus-secret.fifo"
+    rm -f $secret_path
+    mkfifo $secret_path
+    chmod 600 $secret_path
+    $secret_path
+}
+
+def start-localnet-consensus-secret-writer [secret_path: string] {
+    job spawn { $"($LOCALNET_SIGNING_KEY_SECRET)\n" | save -f $secret_path } | ignore
```
SuperFluffy, line 45: tbh, I have no idea how this works - does `save -f $PATH` write something to the path?

mediocregopher: I'm new to nuscript, but `save -f` opens for writing, it pipes the string into that, the `ignore` is so output doesn't go to stdout, `job spawn` spawns that whole block in the background.

mattsse (approved):

SuperFluffy (approved):
Looks good to me, but I'm not a nushell practitioner

PR #5250, opened 2026-06-02
test(tip1060): consume deferred credits, erase refund gas, add SSTORE mode tests

rakita:
## What

Builds on the storage-credits SSTORE work with a correctness fix and end-to-end tests.

- **Consume deferred credits + erase refund gas** (`tip1060.rs`, `handler.rs`): move `apply_refund` into `last_frame_result` so deferred storage-creation credits are flushed before the result gas is snapshotted, and erase the corresponding gas cost. `apply_refund` now *consumes* minted tokens against the accrued per-key credit instead of accumulating them.
- **Defer `Direct` skip flags** (`gas_state.rs`): only set `skip_refund`/`skip_gas` once a credit is actually available for the SSTORE.
- **Tests** (`evm.rs`): add T6 SSTORE create/clear coverage across all three credit modes (`Refund`, `Preserve`, `Direct`), plus a cross-transaction test showing minted credits change the cost of a later create.

## Notes

- A `// TODO use gas_params for proper constant.` remains on the `230_000` magic number in `apply_refund`.

## Test

`cargo test -p tempo-revm --lib tip1060` — both new tests pass.


crates/revm/src/evm.rs
```diff
@@ -381,6 +382,27 @@ mod tests {
         evm
     }
 
+    /// Create an EVM with T6 hardfork enabled and a funded account.
+    /// T6 activates the TIP-1060 SSTORE storage gas-token hook. TIP-1016 (Amsterdam
+    /// EIP-8037 state-gas split) is disabled in T6, matching production config.
```
0xrusowsky:
```suggestion
    /// Creates a T6-enabled EVM with a funded account.
    /// This activates the TIP-1060 SSTORE gas-token hook while keeping the
    /// TIP-1016 state-gas split disabled to match production.

```

0xrusowsky (approved):

PR #5251, opened 2026-06-02
feat(evm): activate scheduled feature tips

emmajam:
Adds the scheduled feature tip activation scaffold on top of the TIP-1063 node features branch. The block executor activates the scheduled tip once consensus epoch reaches the target, while quorum enforcement and public scheduling writes stay out of this pass.


crates/evm/src/block.rs
```diff
@@ -209,6 +211,21 @@ where
...
+        };
+
+        tempo_revm::activate_scheduled_features_tip(
+            self.evm_mut().ctx_mut(),
+            consensus_context.epoch,
+        )
+        .map_err(BlockExecutionError::other)?;
+
+        Ok(())
+    }
```
klkvr, line 248: this doesn't seem right, i think we should do an actual call into the evm so that we get state that we can commit and get it dispatched to state hook, etc. rn this would just update the journal but not the database state

klkvr (approved):

PR #5255, opened 2026-06-02
fix: decode packed TIP-1060 state during refund settlement

0xrusowsky:
this PR properly decodes TIP1060 account state during refund settlement.
it also disables legacy refunds, as required by the spec

### DIFF
--- a/crates/precompiles/src/tip1060_storage_credits/gas_state.rs
+++ b/crates/precompiles/src/tip1060_storage_credits/gas_state.rs
@@ -70,7 +70,9 @@ pub fn sstore_storage_credits<B: StorageCreditsBackend>(
     owner: Address,
     values: &SStoreResult,
 ) -> Result<GasStateOutcome, B::Error> {
+    // TIP-1060 removes the legacy storage-clearing gas refunds.
     let mut outcome = GasStateOutcome::default();
+    outcome.skip_refund = true;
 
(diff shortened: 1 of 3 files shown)


crates/precompiles/src/tip1060_storage_credits/gas_state.rs
```diff
@@ -110,25 +112,20 @@ pub fn sstore_storage_credits<B: StorageCreditsBackend>(
...
+                // Only if there is a credit available, skip gas
                 if storage_credits.balance > 0 {
                     // Consume the storage credit and charge 20k for the SSTORE.
                     backend.charge_gas(20_000)?;
                     storage_credits.balance -= 1;
                     was_changed = true;
-
-                    // only if there is a credit available, skip refund and gas
-                    outcome.skip_refund = true;
                     outcome.skip_gas = true;
```
0xrusowsky, line 123: do u think this is correct given that the spec says: https://tips.sh/1060?h=3259-3822,26908-26912,26949-26979

rakita (approved):

crates/precompiles/src/tip1060_storage_credits/gas_state.rs
rakita: Right, will take a look

PR #5153, opened 2026-06-01
feat(tip-1067): dynamic base fee

dankrad:
Adds TIP-1067, replacing TIP-1010's fixed `2 × 10^10` attodollars/gas base fee with an EIP-1559-style per-block controller bounded to `[BASE_FEE_CAP / 20, BASE_FEE_CAP]`. The current constants set `BASE_FEE_CAP = 1.2 × 10^10` attodollars/gas, `BASE_FEE_FLOOR = 6 × 10^8` attodollars/gas, and `GAS_TARGET = 10,000,000` gas/block.

When the chain is underutilized the base fee decays toward one twentieth of the cap; under load it rises back to the lowered cap and never exceeds it. Because the cap is below TIP-1010's current fixed fee, this can only make fees cheaper than today.

## Parameters

- Gas target: `10,000,000` gas/block
- Floor: `1.2 × 10^10 / 20 = 6 × 10^8` attodollars/gas
- Cap: `1.2 × 10^10` attodollars/gas
- Max change per block: `1/8` (12.5%), EIP-1559 denominator `8`

## Numerical examples

**TIP-20 transfer (50,000 gas) at the bounds**
- At the cap (`1.2 × 10^10`): `50,000 × 1.2 × 10^10 = 6 × 10^14` attodollars = `600` microdollars = **$0.0006 (0.06 cent)**.
- At the floor (`6 × 10^8`): `50,000 × 6 × 10^8 = 3 × 10^13` attodollars = `30` microdollars = **$0.00003 (0.003 cent)** — 20× cheaper than the cap.

**Decay when idle** (empty block multiplies base fee by `7/8`, a 12.5% drop, at ~500ms blocks)
- Cap → floor (20× decrease): ~23 consecutive empty blocks (~12s).
- Decay half-life: ~5 blocks (~2.6s).

**Rise under a 30M gas block** (general cap)
- `gas_delta = 20M`, step `= 20M / 10M / 8 = 0.25` → multiply by about `1.25`/block.
- Floor → cap: ~14 such blocks (~7s).

**Rise under a full 500M gas block** (block gas limit)
- `gas_delta = 490M`, raw step `= 490M / 10M / 8 = 6.125` → tries to multiply by **≈7.125×** in one block, then clamps to the cap.
- From the floor: one block → `4.275 × 10^9` (~36% of cap); a second full block saturates to the cap (~2 blocks, ~1s).
[...]

PR #5257, opened 2026-06-03
feat: TIP-1067

klkvr:
Implements https://github.com/tempoxyz/tempo/pull/5153


decofe:
cc @klkvr

❌ Benchmark failed. [View logs](https://github.com/tempoxyz/tempo/actions/runs/26878352481/job/79271599152)

crates/chainspec/src/spec.rs
```diff
@@ -356,8 +356,26 @@ impl EthChainSpec for TempoChainSpec {
         self.inner.get_final_paris_total_difficulty()
     }
 
-    fn next_block_base_fee(&self, _parent: &TempoHeader, target_timestamp: u64) -> Option<u64> {
-        Some(self.tempo_hardfork_at(target_timestamp).base_fee())
+    fn next_block_base_fee(&self, parent: &TempoHeader, target_timestamp: u64) -> Option<u64> {
+        let target_fork = self.tempo_hardfork_at(target_timestamp);
+        if target_fork.is_t6() && !self.tempo_hardfork_at(parent.inner.timestamp).is_t6() {
+            return Some(TEMPO_T6_BASE_FEE_CAP);
```
legion2002: think this can be removed, as at hardfork boundary the value will just be clamped to the cap anyways?

crates/chainspec/src/constants.rs
```diff
@@ -28,6 +30,37 @@ pub mod gas {
...
+    /// TIP-1067 base fee floor: one twentieth of the TIP-1067 cap.
+    pub const TEMPO_T6_BASE_FEE_FLOOR: u64 = TEMPO_T6_BASE_FEE_CAP / 20;
+
+    /// TIP-1067 gas target for the dynamic base fee controller.
+    pub const TEMPO_T6_BASE_FEE_GAS_TARGET: u64 = 10_000_000;
+
+    /// TIP-1067 uses EIP-1559's base-fee update formula with a fixed 10M gas target.
+    ///
+    /// The params are `(max_change_denominator = 8, elasticity_multiplier = 1)`: `8` keeps the
+    /// standard EIP-1559 maximum 12.5% per-block base-fee delta, while `1` prevents EIP-1559's
```
legion2002, line 45: At hardfork activation boundary, the max base-fee delta would be 40% with current params. But don't think it's a problem.

legion2002 (dismissed):
lgtm, pending resolution of the discussion around the TIP constants and CI

crates/chainspec/src/spec.rs
```diff
@@ -356,8 +356,23 @@ impl EthChainSpec for TempoChainSpec {
...
-    fn next_block_base_fee(&self, _parent: &TempoHeader, target_timestamp: u64) -> Option<u64> {
-        Some(self.tempo_hardfork_at(target_timestamp).base_fee())
+    fn next_block_base_fee(&self, parent: &TempoHeader, target_timestamp: u64) -> Option<u64> {
+        let target_fork = self.tempo_hardfork_at(target_timestamp);
+
+        if target_fork.is_t7() {
+            let parent_base_fee = parent
+                .inner
+                .base_fee_per_gas
+                .expect("tempo blocks are expected to have a base fee");
```
mattsse, line 366: should we treat this as None response and use `?`

crates/node/tests/it/payment_lane.rs
```diff
@@ -9,12 +9,33 @@ use alloy::{
...
+where
+    P: Provider,
+    R: ReceiptResponse,
+{
+    let block_number = receipt
+        .block_number()
+        .expect("mined receipt should have a block number");
+    let block = provider
+        .get_block_by_number(BlockNumberOrTag::Number(block_number))
+        .await?
```
mattsse, line 29: we only need the header here?
but this is just test code so we can keep it as is

0xrusowsky (commented):
makes sense to me

0xrusowsky (approved):

crates/chainspec/src/hardfork.rs
```diff
@@ -206,20 +206,6 @@ tempo_hardfork! (
...
-    /// - Pre-T1: 10 billion attodollars per gas
-    /// - T1+: 20 billion attodollars per gas (targets ~0.1 cent per TIP-20 transfer)
-    ///
-    /// Economic conversion: ceil(basefee × gas_used / 10^12) = cost in microdollars (TIP-20 tokens)
-    pub const fn base_fee(&self) -> u64 {
-        if self.is_t1() {
-            return gas::TEMPO_T1_BASE_FEE;
-        }
-        gas::TEMPO_T0_BASE_FEE
-    }
```
mattsse, line 221: good call to remove this so that we dont reuse this accidenally

mattsse (commented):
lgtm

one convern is that the current txpool validation might not respect the next base fee, so we could pin it to the floor value?

PR #5264, opened 2026-06-03
feat(precompiles): add feature tip scheduling

emmajam:
Adds owner-gated FeatureRegistry scheduling and cancellation, with ownership mirrored from ValidatorConfigV2.

Propagates the consensus epoch into precompile execution so scheduleFeaturesTip rejects non-future activation epochs while scheduled activation continues to run from the block executor.

### DIFF
--- a/crates/precompiles/src/feature_registry/mod.rs
+++ b/crates/precompiles/src/feature_registry/mod.rs
@@ -15,6 +15,7 @@ use tempo_precompiles_macros::contract;
 // Activation requires 80% support from the active validator set, represented exactly as 4/5.
 const ACTIVATION_QUORUM_NUMERATOR: u64 = 4;
 const ACTIVATION_QUORUM_DENOMINATOR: u64 = 5;
+const FEATURE_REGISTRY_EPOCH_LENGTH: u64 = 21_600;
 
 /// Protocol feature registry.
 ///
@@ -85,10 +86,6 @@ impl FeatureRegistry {
     ) -> Result<()> {
         self.check_owner(msg_sender)?;
 
-        if call.featuresTip == 0 {
-            return Err(FeatureRegistryError::invalid_features_tip().into());
-        }
-
         if call.featuresTip <= self.features_tip.read()? {
             return Err(FeatureRegistryError::features_tip_not_increasing().into());
         }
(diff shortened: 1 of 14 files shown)


crates/precompiles/src/feature_registry/mod.rs
```diff
@@ -65,6 +78,58 @@ impl FeatureRegistry {
...
+        self.check_owner(msg_sender)?;
+
+        if call.featuresTip == 0 {
+            return Err(FeatureRegistryError::invalid_features_tip().into());
+        }
+
+        if call.featuresTip <= self.features_tip.read()? {
+            return Err(FeatureRegistryError::features_tip_not_increasing().into());
+        }
+
```
klkvr:
```suggestion
        if call.featuresTip <= self.features_tip.read()? {
            return Err(FeatureRegistryError::features_tip_not_increasing().into());
        }

```

crates/precompiles/src/feature_registry/mod.rs
```diff
@@ -65,6 +78,58 @@ impl FeatureRegistry {
...
+        if call.activationEpoch == 0
+            || self
+                .storage
+                .consensus_epoch()
+                .is_some_and(|epoch| call.activationEpoch <= epoch)
+        {
+            return Err(FeatureRegistryError::activation_epoch_not_future().into());
+        }
```
klkvr:
```suggestion
        if self
                .storage
                .consensus_epoch()
                .is_some_and(|epoch| call.activationEpoch <= epoch)
        {
            return Err(FeatureRegistryError::activation_epoch_not_future().into());
        }
```

crates/revm/src/lib.rs
```diff
@@ -24,3 +24,7 @@ pub use evm::TempoEvm;
 pub use handler::{ValidationContext, calculate_aa_batch_intrinsic_gas};
 pub use revm::interpreter::instructions::utility::IntoAddress;
 pub use tx::{TempoBatchCallEnv, TempoTxEnv};
+
+pub fn enter_consensus_epoch<R>(consensus_epoch: Option<u64>, f: impl FnOnce() -> R) -> R {
+    tempo_precompiles::storage::StorageCtx::enter_consensus_epoch(consensus_epoch, f)
```
klkvr: not sure if i like propagating the epoch like this and having it optional on the storage provider

technically we could propagate it from `TempoBlockEnv` or just infer from block number. rn epoching is pretty much static tempo_precompile

tips/verify/test/TempoTest.t.sol
```diff
@@ -52,7 +52,6 @@ contract TempoTest is Tempo, Test {
     function setUp() public virtual {
         _requirePrecompile("AccountKeychain", KEYCHAIN);
         _requirePrecompile("TIP403Registry", TIP403_REGISTRY);
-        _requirePrecompile("AddressRegistry", ADDRESS_REGISTRY);
```
klkvr, line 55: accidental?

emmajam: fixed

klkvr (approved):

PR #5265, opened 2026-06-03
docs(tip-1067): preserve TIP-1059 discount

malleshpai:
Updates TIP-1067 so TIP-1059 eligible transactions keep paying 60% of the prevailing base fee, clamped at the dynamic fee floor.

Also clarifies pricing examples and test coverage for discounted transfers.


dankrad:
Closing in favour of simpler solution of just lowering the basefee

PR #5285, opened 2026-06-04
feat(evm): report supported feature tip

emmajam:
Adds block execution wiring to report the local node's highest supported protocol feature tip through FeatureRegistry when nonzero. The write is skipped when the validator has already reported that tip or a higher one, and scheduled activation now reuses the same system-call result handling.

### DIFF
--- a/crates/evm/src/block.rs
+++ b/crates/evm/src/block.rs
@@ -30,15 +30,18 @@ use tempo_chainspec::{
     TempoChainSpec, features::highest_supported_protocol_feature_id, hardfork::TempoHardforks,
 };
 use tempo_contracts::precompiles::{
-    ADDRESS_REGISTRY_ADDRESS, FEATURE_REGISTRY_ADDRESS, IFeatureRegistry, IValidatorConfigV2,
+    ADDRESS_REGISTRY_ADDRESS, FEATURE_REGISTRY_ADDRESS, IFeatureRegistry,
     RECEIVE_POLICY_GUARD_ADDRESS, SIGNATURE_VERIFIER_ADDRESS, SYSTEM_CALLER_ADDRESS,
     TIP20_CHANNEL_RESERVE_ADDRESS, VALIDATOR_CONFIG_V2_ADDRESS,
+};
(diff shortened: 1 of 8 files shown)


crates/evm/src/block.rs
```diff
@@ -213,40 +215,132 @@ where
...
+
+        let call = IValidatorConfigV2::validatorByPublicKeyCall {
+            publicKey: B256::from(&consensus_context.proposer),
+        };
+        let output = self.transact_precompile_system_call(
+            VALIDATOR_CONFIG_V2_ADDRESS,
+            call.abi_encode().into(),
+            "validator config proposer lookup",
+            false,
+        )?;
```
klkvr: i don't think we should need an entire `transact` call for this, for read-only call seems fine to just create a `StorageCtx` in place, not sure if i like this `commit: bool`

crates/evm/src/block.rs
```diff
@@ -299,15 +322,66 @@ where
...
+            .evm_mut()
+            .db_mut()
+            .with_read_only_storage_ctx(spec, || {
+                ValidatorConfigV2::new().validator_by_public_key(public_key)
+            })
+            .map_err(|_| {
+                BlockValidationError::msg("invalid feature registry system transaction")
+            })?;
+
+        if validator.deactivatedAtHeight != 0 || validator.validatorAddress != call.validator {
```
klkvr: looks like we could slightly simplify both this and the builder if we changed `setSupportedFeaturesTipCall` to take public key as input and then look it up in the validator regsitry internally?

i guess one downside is that within EVM passing address is kinda cleaner

klkvr (approved):

crates/evm/src/block.rs
emmajam: validation still has to resolve publicKey -> validator to enforce active validator + no-op rejection, so it doesn’t really delete much logic

Current model is explicit and harder to misuse: builder reports for a validator address, block validation proves that address matches the proposer pubkey

klkvr (approved):

PR #5286, opened 2026-06-04
fix(tips): allow T2 tests without address registry

decofe:
Fixes the invariant workflow failure where `test/TIP1015.t.sol:TIP1015Test` runs under its declared `tempo:T2` hardfork but inherits `TempoTest.setUp()`, which unconditionally required the T3 `AddressRegistry` precompile.

`AddressRegistry` is not active at T2, so the base fixture should not require it for all TIP verification tests. T3+ coverage still exercises the precompile through the tests that use it directly.

Validation:
- `/home/agent/branches/foundry-rs/foundry/target/debug/forge test --match-path test/TIP1015.t.sol --mt 'test_invariant' -vvvv`\n\nPrompted by: @grandizzy


decofe:
Closing in favor of the Foundry-side fix in foundry-rs/foundry#15064. The Argo failure is caused by Forge local genesis/warmup sentinel setup, not Tempo node runtime behavior.

PR #5288, opened 2026-06-04
fix: apply_refunds moved and clear/create flag does not touch origin

rakita:
Fixes for found bugs:
- apply_refund moved after exec loop
- is_clear and is_create now take into account only new value and not original value. (Checks for new != present is done few lines before).
- t6 check in precompile activation

### DIFF
--- a/crates/precompiles/src/tip1060_storage_credits/gas_state.rs
+++ b/crates/precompiles/src/tip1060_storage_credits/gas_state.rs
@@ -83,8 +83,10 @@ pub fn sstore_storage_credits<B: StorageCreditsBackend>(
         skip_refund: true,
     };
 
-    // if new and present value are equal or if both present value are not zero, skip storage credits accounting.
-    if values.is_new_eq_present() || (!values.is_present_zero() && !values.is_new_zero()) {
+    // Only account for storage credits when the slot crosses the zero boundary
+    // (zero -> non-zero or non-zero -> zero). If both values are zero or both are
+    // non-zero, slot occupancy is unchanged, so skip storage credits accounting.
+    if values.is_present_zero() == values.is_new_zero() {
         return Ok(outcome);
     }
 
--- a/crates/precompiles/src/tip1060_storage_credits/gas_state.rs
+++ b/crates/precompiles/src/tip1060_storage_credits/gas_state.rs
@@ -103,16 +103,6 @@ pub fn sstore_storage_credits<B: StorageCreditsBackend>(
         });
     }
 
-    // 0→x: slot create (charged EIP-8037 state gas today).
-    let is_create = values.is_original_eq_present() && values.is_original_zero();
-    // x→0: slot clear (present non-zero set to zero).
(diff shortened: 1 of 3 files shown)


crates/precompiles/src/tip1060_storage_credits/gas_state.rs
```diff
@@ -103,16 +103,6 @@ pub fn sstore_storage_credits<B: StorageCreditsBackend>(
...
 
-    // 0→x: slot create (charged EIP-8037 state gas today).
-    let is_create = values.is_original_eq_present() && values.is_original_zero();
-    // x→0: slot clear (present non-zero set to zero).
-    let is_clear = values.is_new_zero() && !values.is_present_zero();
-
-    // x→y: return from instruction, nothing to be done.
-    if !(is_create || is_clear) {
-        return Ok(outcome);
-    }
```
0xrusowsky, line 114: we should preserve the early return to avoid reading form the `StorageCredit` precompile if we go x→y

crates/precompiles/src/tip1060_storage_credits/gas_state.rs
```diff
@@ -131,10 +121,12 @@ pub fn sstore_storage_credits<B: StorageCreditsBackend>(
         AccountState::from_word(storage_credit_state_load.data).map_err(|_| B::fatal_external())?;
```
0xrusowsky, line 124: i guess we could turn this into a closure and call it inside each if branch

0xrusowsky (commented):
looks good, but i think we can do this perf optimization

crates/precompiles/src/tip1060_storage_credits/gas_state.rs
rakita: With only checking if new is zero or not, this will always be false

rakita: True, will add check for if new and present are not zeroes but different values

crates/precompiles/src/tip1060_storage_credits/gas_state.rs
rakita: I think it is fine as it is needed for both branches

crates/precompiles/src/tip1060_storage_credits/gas_state.rs
```diff
@@ -83,7 +83,10 @@ pub fn sstore_storage_credits<B: StorageCreditsBackend>(
         skip_refund: true,
     };
 
-    if values.is_new_eq_present() {
+    // Only account for storage credits when the slot crosses the zero boundary
+    // (zero -> non-zero or non-zero -> zero). If both values are zero or both are
+    // non-zero, slot occupancy is unchanged, so skip storage credits accounting.
+    if values.is_present_zero() == values.is_new_zero() {
```
0xrusowsky, line 89: nice

0xrusowsky (approved):

PR #5299, opened 2026-06-04
fix(rpc): reject 2D-nonce gas estimation when `from` is missing

crazywriter1:
`create_txn_env` defaulted a missing `from` to the zero address when a request
set a non-zero `nonceKey` without an explicit `nonce`, then read the zero-address
nonce slot and returned a bogus gas estimate / simulation result.

Now it returns `SignError::NoAccount` ("unknown account") when `from` is absent,
mirroring the existing handling in `next_available_nonce_for`.

Adds an integration test: `eth_estimateGas` with `nonceKey` but no `from` is
rejected as "unknown account".


mattsse (commented):
we can keep this as is

